PR #58479, opened 2025-04-28
[Enhancement] Prevent MV reload too many times to reduce FE's startup time

wxl24life:
## Why I'm doing:

When there are many materialized views and the dependency among them is complicated, it might cost a lot of time to restart FE.

## What I'm doing:

Add a `reloaded` flag to `MaterializedView` to indicate a materialized view has been reloaded before, and no need to repeatedly reload it again and again when it was a sub-mv of some other materialized view 

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
@@ -529,6 +529,10 @@ public class MaterializedView extends OlapTable implements GsonPreProcessable, G
 
     protected volatile ParseNode defineQueryParseNode = null;
 
+    // Use a flag to prevent MV reload too many times while recursively reloading every mv at FE start time
+    // and in each round of checkpoint
+    private boolean reloaded = false;
+
     public MaterializedView() {
         super(TableType.MATERIALIZED_VIEW);
         this.tableProperty = null;
@@ -732,6 +736,14 @@ public class MaterializedView extends OlapTable implements GsonPreProcessable, G
 
     public void setQueryOutputIndices(List<Integer> queryOutputIndices) {
         this.queryOutputIndices = queryOutputIndices;
+    }
+
+    public boolean isReloaded() {
+        return reloaded;
+    }
+
+    public void setReloaded(boolean reloaded) {
+        this.reloaded = reloaded;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/server/GlobalStateMgr.java
```diff
@@ -1700,16 +1700,27 @@ private void onReloadTables() {
...
+                mv.onReload(true);
+            }
+        }
+
+        // we should reset reloaded flags after each round of reloading for all materializedViews
+        int count = 0;
+        for (Database db : localMetastore.getIdToDb().values()) {
+            for (MaterializedView mv : db.getMaterializedViews()) {
+                count++;
+                mv.setReloaded(false);
```
LiShuMing, line 1718: Why needs to reset it again? (Of course this should be safer)

LiShuMing (dismissed):
+1

fe/fe-core/src/main/java/com/starrocks/server/GlobalStateMgr.java
wxl24life: The `onReload` method is called in some other scenarios such as - schema change of a materialize view.  The reloaded flag was introduced only to increase the speed of FE startup and checkpointing. It shouldn't affect the behavior of other operation which might indeed need to do reload.

gengjun-git (dismissed):

caneGuy (dismissed):

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
```diff
@@ -1055,14 +1074,14 @@ public void onReload() {
...
-        onReloadImpl();
+        onReloadImpl(false);
     }
 
     /**
      * @return active or not
      */
-    private boolean onReloadImpl() {
-        long startTime = System.currentTimeMillis();
+    private boolean onReloadImpl(boolean postLoadImage) {
```
kevincai, line 1083: The interfaces changed here may not be very clear if put all together into the whole MV context.

Maybe, introduce a concept of "ReloadOnce" and "ReloadRecursively" to explicit the different behavior and accidentally it will be `ReloadOnce` mode during the post-loadImage phase.

wxl24life: ok i have added these notices as comments

LiShuMing (approved):

gengjun-git (approved):

PR #58484, opened 2025-04-28
[BugFix] LargeStringLiterals in filters of JDBCScanNode/MySQLScanNode is truncated and appended to ellipsis

satanson:
## Why I'm doing:

Conjuncts in JDBCScanNode/MySQLScanNode is converted into string-type filters which are dispatched to other db server to process. the LargeStringLiteral(length exceeds 50) is truncated mistakenly and appended to ellipsis(...).  
```java
    public static StringLiteral create(String value) {
        if (value.length() > LargeStringLiteral.LEN_LIMIT) {
            return new LargeStringLiteral(value, NodePosition.ZERO);
        } else {
            return new StringLiteral(value);
        }
    }
```

```java
    @Override
    public String toSqlImpl() {
        if (shortSqlStr == null) {
            String fullSql = toFullSqlImpl();
            fullSql = fullSql.substring(0, LEN_LIMIT);
            shortSqlStr = fullSql + "...'";
        }
        return shortSqlStr;
    }
```
## What I'm doing:

Convert LargeStringLiteral to StringLiteral for JDBCScanNode/MySQLScanNode

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [x] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/analysis/Expr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/analysis/Expr.java
@@ -46,6 +46,7 @@ import com.starrocks.catalog.FunctionSet;
 import com.starrocks.catalog.ScalarType;
 import com.starrocks.catalog.Type;
 import com.starrocks.common.AnalysisException;
+import com.starrocks.common.Pair;
 import com.starrocks.common.TreeNode;
 import com.starrocks.common.io.Writable;
 import com.starrocks.planner.FragmentNormalizer;
@@ -84,8 +85,10 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Root of the expr node hierarchy.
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/com/starrocks/analysis/Expr.java
```diff
@@ -286,6 +289,32 @@ public Type getOriginType() {
...
+
+        for (int i = 0; i < this.children.size(); ++i) {
+            Pair<Expr, Optional<Expr>> childPair = childAndNewChildPairList.get(i);
+            Expr newChild = childPair.second.orElse(childPair.first);
+            setChild(i, newChild);
+        }
+        return Optional.of(this);
+    }
+
+    public Expr replaceLargeStringLiteral() {
```
LiShuMing, line 314: so `Expr` to SQL also needs to be fixed like this(eg:`AstToStringBuilder` )?

MV's defined sql is deduced from `AstToSQLBuilder#toSQL` which can also be wrong if it meets `LargeStringLiteral`?

satanson: yeah, it seems refresh sql and mv sql also have this issue.

Seaven (approved):

stephen-shelby (approved):

fe/fe-core/src/main/java/com/starrocks/planner/JDBCScanNode.java
```diff
@@ -140,6 +140,7 @@ private void createJDBCTableFilters() {
 
         ArrayList<Expr> jdbcConjuncts = Expr.cloneList(conjuncts, sMap);
         for (Expr p : jdbcConjuncts) {
+            p = p.replaceLargeStringLiteral();
```
alvin-celerdata, line 143: Could you add a comment on why this operation is added here? 
For me, I can't understand the reason, I think it is not understandable to others.
And why is this LargeStringLiteral.toString not the correct value? Maybe we should consider introducing another function to achieve it.
Like fix LargeStringValue.toString() to return a normal value, but introduce another function to give a readable string.

PR #58509, opened 2025-04-28
[Feature] Support iceberg partition transform

### LINKED ISSUE #58322: iceberg partition table transforms
https://docs.starrocks.io/docs/data_source/catalog/iceberg/iceberg_catalog/#partition_desc

Starrocks only supports identity transform and does not handle day, month etc https://iceberg.apache.org/spec/#partition-transforms

select count(*) from users where date(_cdc_tsms)='2032-01-01';
select count(*) from users where day(_cdc_tsms)=232222;

both queries scan the whole table and i cant select particular partition even if i calculated day outside sr

Wenjun7J:
## Why I'm doing:
Support the iceberg transform partition.
## What I'm doing:
Support the partition transform for iceberg, and supported column types are below:
| Transform  | SR Result Type | SR Argument Type   |
|-------|------|--------|
| year  | int   |  date, datetime   |
| month  | int   |  date, datetime  |
| day | date  | date, datetime   |
| hour | int   | datetime   |
| bucket | int   | int, bigint, date, datetime, varchar,  varbinary, decimal32, decimal64, decimal128  |
| truncate | origin column type   |  int, bigint, decimal32, decimal64, decimal128, varchar, varbinary   |

These transform are only supported in iceberg partition clause now.

Fixes #58322
## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/connector/connector_chunk_sink.cpp
+++ b/be/src/connector/connector_chunk_sink.cpp
@@ -16,6 +16,7 @@
 
 #include "column/chunk.h"
 #include "common/status.h"
+#include "connector/iceberg_chunk_sink.h"
 #include "connector/sink_memory_manager.h"
 #include "formats/file_writer.h"
 #include "runtime/runtime_state.h"
@@ -46,17 +47,35 @@ Status ConnectorChunkSink::init() {
 Status ConnectorChunkSink::add(Chunk* chunk) {
     std::string partition = DEFAULT_PARTITION;
     bool partitioned = !_partition_column_names.empty();
+    std::vector<bool> partition_field_null_list;
     if (partitioned) {
-        ASSIGN_OR_RETURN(partition,
-                         HiveUtils::make_partition_name(_partition_column_names, _partition_column_evaluators, chunk,
-                                                        _support_null_partition));
+        if (typeid(*this) == typeid(IcebergChunkSink)) {
+            ASSIGN_OR_RETURN(partition, HiveUtils::iceberg_make_partition_name(
+                                                _partition_column_names, _partition_column_evaluators,
+                                                dynamic_cast<IcebergChunkSink*>(this)->transform_expr(), chunk,
+                                                _support_null_partition, partition_field_null_list));
+        } else {
+            ASSIGN_OR_RETURN(partition,
+                             HiveUtils::make_partition_name(_partition_column_names, _partition_column_evaluators,
+                                                            chunk, _support_null_partition));
+        }
     }
 
-    auto it = _writer_stream_pairs.find(partition);
+    std::string null_fingerprint = "";
+    for (bool b : partition_field_null_list) {
+        null_fingerprint += (b ? '1' : '0');
+    }
+
+// null_fingerprint is used to distinguish with the secenario like NULL and string "null"
+// They are under the same dir path, but should not in the same data file.
+// We should record them in different files so that each data file could has its own meta info.
+// otherwise, the scanFileTask may filter data incorrectly.
+#define PartitionKey(partition, finger) partition + "&" + finger
+    auto it = _writer_stream_pairs.find(PartitionKey(partition, null_fingerprint));
     if (it != _writer_stream_pairs.end()) {
         Writer* writer = it->second.first.get();
         if (writer->get_written_bytes() >= _max_file_size) {
-            callback_on_commit(writer->commit());
+            callback_on_commit(writer->commit().set_extra_data(null_fingerprint));
             _writer_stream_pairs.erase(it);
             auto path = partitioned ? _location_provider->get(partition) : _location_provider->get();
             ASSIGN_OR_RETURN(auto new_writer_and_stream, _file_writer_factory->create(path));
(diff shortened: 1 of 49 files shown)


alvin-celerdata:
@Wenjun7J 
this should be a new feature, please change the title from [Enhancement] to [Feature]

be/src/connector/connector_chunk_sink.cpp
```diff
@@ -18,6 +18,7 @@
 #include "common/status.h"
 #include "connector/sink_memory_manager.h"
 #include "formats/file_writer.h"
+#include "iceberg_chunk_sink.h"
```
trueeyu, line 21: `connector/iceberg_chunk_sink.h`

gensrc/thrift/Descriptors.thrift
```diff
@@ -524,7 +530,7 @@ struct TIcebergTable {
     3: optional TIcebergSchema iceberg_schema
 
     // partition column names
-    4: optional list<string> partition_column_names
```
trueeyu, line 527: If we upgrade BE first and then upgrade the FE, will it be compatible?

be/src/runtime/descriptors.cpp
```diff
@@ -229,20 +229,28 @@ IcebergTableDescriptor::IcebergTableDescriptor(const TTableDescriptor& tdesc, Ob
     _table_location = tdesc.icebergTable.location;
     _columns = tdesc.icebergTable.columns;
     _t_iceberg_schema = tdesc.icebergTable.iceberg_schema;
-    _partition_column_names = tdesc.icebergTable.partition_column_names;
```
trueeyu, line 232: Should we use  `__is_set` to compatile this item?

fe/fe-core/src/test/java/com/starrocks/statistic/StatisticsCollectJobTest.java
```diff
@@ -1262,8 +1262,8 @@ public void testExternalFullStatisticsBuildCollectSQLWithException1() {
                         StatsConstants.ScheduleType.ONCE,
                         Maps.newHashMap());
 
-        expectedException.expect(StarRocksConnectorException.class);
-        expectedException.expectMessage("Partition transform BUCKET not supported to analyze, table: t0_bucket");
+        //expectedException.expect(StarRocksConnectorException.class);
```
trueeyu, line 1265: Why not remove these two lines of code directly?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/PruneHDFSScanColumnRule.java
```diff
@@ -165,7 +165,12 @@ private void checkPartitionColumnType(LogicalScanOperator scanOperator,
                 scanColumnRefOperators.stream().map(col -> context.getColumnRefFactory().getColumn(col)).
                         collect(Collectors.toList());
         partitionColumns.retainAll(scanColumns);
-        if (partitionColumns.stream().map(Column::getType).anyMatch(this::notSupportedPartitionColumnType)) {
+        if (table.isIcebergTable()) {
+            if (partitionColumns.stream().map(Column::getType).anyMatch(this::icebergNotSupportedPartitionColumnType)) {
+                throw new StarRocksPlannerException("Iceberg table partition by float/double/decimalV2 datatype is not supported",
```
trueeyu, line 170: decimalV2 or decimal?

fe/fe-core/src/main/java/com/starrocks/sql/ast/ListPartitionDesc.java
```diff
@@ -230,9 +231,15 @@ public void analyzeExternalPartitionColumns(List<ColumnDef> columnDefs, String e
...
+                    if (engineName.equalsIgnoreCase(Table.TableType.ICEBERG.name())) {
+                        if (columnDef.getType().isFloatingPointType() || columnDef.getType().isComplexType()
+                                || columnDef.getType().isDecimalV2()) {
+                            throw new SemanticException(String.format("Invalid iceberg partition column '%s': %s",
+                                    columnDef.getName(), "invalid data type " + columnDef.getType()));
+                        }
+                    } else if (columnDef.getType().isFloatingPointType() || columnDef.getType().isComplexType()
                             || columnDef.getType().isDecimalOfAnyVersion()) {
-                        throw new SemanticException(String.format("Invalid partition column '%s': %s",
+                        throw new SemanticException(String.format("Invalid hive partition column '%s': %s",
```
trueeyu, line 242: Why add `hive`?

Youngwb:
@Wenjun7J Could you split this PR into some small PRs, it's hard for me to review.
like these : 
1. support create iceberg table with partition transfrom (include parser/analyzer/optimizer)
2. support insert  iceberg table with partition transfrom
3. support partition transform function

Wenjun7J:
> @Wenjun7J Could you split this PR into some small PRs, it's hard for me to review. like these :
> 
> 1. support create iceberg table with partition transfrom (include parser/analyzer/optimizer)
> 2. support insert  iceberg table with partition transfrom
> 3. support partition transform function

To split the code through ddl and transform function definition may not be easy, because they have some dependencies.

Actually, the most of code are doing two things:
1. insert iceberg table with partition transform
2. support partition transform function

The code to create tables with partition transform is very simple.
You can check the logic in  `CreateTableAnalyzer.java`, `CreateTableStmt.java` and `IcebergMetadata.java createTable` Method.

The partition transform function are all implemented in `be/exprs`.

The rest of the code is mainly used to process insert.

At last, there is some code in fe's `optimizer` to process the query with filters for partition pruning.

So I think the structure can be clearer after the descriptions above. But if needed, I think I can try to refactor and break these logics into separate parts.

eshishki:
maybe close this?

PR #58517, opened 2025-04-28
[BugFix] Fix a bug in neon_select_if_common_implement

zhangyaqi1989:
## Why I'm doing:

There is a bug in neon_select_if_common_implement. `dst`, `a`, `b` are T*, we should not increment them by 16 elements.

## What I'm doing:

This PR fixes this bug.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/simd/selector.h
+++ b/be/src/simd/selector.h
@@ -441,13 +441,14 @@ inline void neon_select_if_common_implement(uint8_t*& selector, T*& dst, const T
             }
         }
 
-        dst += 16;
+        const int elements_per_iteration = neon_width / data_size;
+        dst += elements_per_iteration;
         selector += 16;
         if constexpr (!left_const) {
-            a += 16;
+            a += elements_per_iteration;
         }
         if constexpr (!right_const) {
-            b += 16;
+            b += elements_per_iteration;
         }
     }
 }


stdpain (approved):

before-Sunrise:
actually the big loop below will handle 16 selector masks every iteration, which means it will handle 16 row of column
`while (dst + neon_width < dst_end) `

for example, if data_size == 2, the big loop will load 16 selector mask(16* 8bit=128bit=neon rigestor width), and the inner loop will load 8 row * 2byte=16byte=128 bit every iteration and loop for 2 times. so no matter what data_size is, the big loop outside always handle 16 row

zhangyaqi1989:
> actually the big loop below will handle 16 selector masks every iteration, which means it will handle 16 row of column `while (dst + neon_width < dst_end) `
> 
> for example, if data_size == 2, the big loop will load 16 selector mask(16* 8bit=128bit=neon rigestor width), and the inner loop will load 8 row * 2byte=16byte=128 bit every iteration and loop for 2 times. so no matter what data_size is, the big loop outside always handle 16 row

Ah, you are right. Closed this PR.

PR #58654, opened 2025-05-02
[Doc] Add description for `iceberg.catalog.view-endpoints-supported`

ebongzzang:
## Why I'm doing:

Currently when using View-related features in Iceberg REST catalog, VIEW-related functions cannot be used.

Without the option `iceberg.catalog.view-endpoints-supported`, `Server does not support endpoint error occurs` and these features cannot be used.

Therefore, I would like to document this option so users are aware of it and can use views.

## What I'm doing:

Added option `iceberg.catalog.view-endpoints-supported` to the Iceberg catalog docs to ensure users are aware of this requirement.

Related sections: 
- https://github.com/apache/iceberg/blob/4a432839233f2343a9eae8255532f911f06358ef/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L264-L274
- https://github.com/apache/iceberg/blob/4a432839233f2343a9eae8255532f911f06358ef/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L447-L454

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [X] Doc
- [ ] Tool

- [] Yes, this PR will result in a change in behavior.

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/docs/en/data_source/catalog/iceberg/iceberg_catalog.md
+++ b/docs/en/data_source/catalog/iceberg/iceberg_catalog.md
@@ -261,6 +261,11 @@ Description: The type of metastore that you use for your Iceberg cluster. Set th
 Required: Yes
 Description: The URI of the REST service endpoint. Example: `https://api.tabular.io/ws`. 
 
+###### iceberg.catalog.view-endpoints-supported
+
+Required: No
+Description: Enables support for View-related operations. Default: `false`
+
 ###### iceberg.catalog.security
 
 Required: No
--- a/docs/en/data_source/catalog/iceberg/iceberg_catalog.md
+++ b/docs/en/data_source/catalog/iceberg/iceberg_catalog.md
@@ -264,7 +264,8 @@ Description: The URI of the REST service endpoint. Example: `https://api.tabular
 ###### iceberg.catalog.view-endpoints-supported
 
 Required: No
-Description: When the old REST service does not return endpoints in CatalogConfig, whether to use the view endpoints to support view-related operations. This parameter is used for backwards compatibility with older REST servers. Default: `false`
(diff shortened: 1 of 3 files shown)


docs/en/data_source/catalog/iceberg/iceberg_catalog.md
```diff
@@ -261,6 +261,11 @@ Description: The type of metastore that you use for your Iceberg cluster. Set th
 Required: Yes
 Description: The URI of the REST service endpoint. Example: `https://api.tabular.io/ws`. 
 
+###### iceberg.catalog.view-endpoints-supported
+
+Required: No
+Description: Enables support for View-related operations. Default: `false`
```
Youngwb, line 267:
```suggestion
Description: When the old REST service does not return endpoints in CatalogConfig, whether to use the view endpoints to support view-related operations. This parameter is used for backwards compatibility with older REST servers. Default: `false`
```

Is this better?

ebongzzang: I also think what you suggested is better. Thank you. I will reflect on it.

Youngwb (dismissed):

EsoragotoSpirit (approved):

Youngwb (approved):

Youngwb:
@ebongzzang  thanks for your contribution

PR #58710, opened 2025-05-06
[Enhancement] support adaptive RocksDB write buffer configuration

luohaha:
## Why I'm doing:
In the current implementation, the RocksDB instance storing metadata uses the default write buffer size of 64MB. In scenarios with heavy metadata write operations, this configuration may potentially cause metadata write blocking issues.

## What I'm doing:
Introduce an adaptive write buffer sizing strategy that defaults to using 5% of system memory for the write buffer, with a defined maximum upper limit.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1369,6 +1369,13 @@ CONF_String(rocksdb_cf_options_string, "block_based_table_factory={block_cache={
 
 CONF_String(rocksdb_db_options_string, "create_if_missing=true;create_missing_column_families=true");
 
+// It is the memory percent of write buffer for meta in rocksdb.
+// default is 5% of system memory.
+CONF_mInt64(rocksdb_write_buffer_memory_percent, "5");
+// It is the max size of the write buffer for meta in rocksdb.
+// default is 1GB.
+CONF_mInt64(rocksdb_max_write_buffer_memory_bytes, "1073741824");
+
 // limit local exchange buffer's memory size per driver
 CONF_Int64(local_exchange_buffer_mem_limit_per_driver, "134217728"); // 128MB
 // only used for test. default: 128M
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1371,10 +1371,12 @@ CONF_String(rocksdb_db_options_string, "create_if_missing=true;create_missing_co
 
 // It is the memory percent of write buffer for meta in rocksdb.
 // default is 5% of system memory.
-CONF_mInt64(rocksdb_write_buffer_memory_percent, "5");
+// However, aside from this, the final calculated size of the write buffer memory
+// will not be less than 64MB nor exceed 1GB (rocksdb_max_write_buffer_memory_bytes).
+CONF_Int64(rocksdb_write_buffer_memory_percent, "5");
 // It is the max size of the write buffer for meta in rocksdb.
 // default is 1GB.
-CONF_mInt64(rocksdb_max_write_buffer_memory_bytes, "1073741824");
+CONF_Int64(rocksdb_max_write_buffer_memory_bytes, "1073741824");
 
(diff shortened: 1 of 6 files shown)


be/src/storage/kv_store.cpp
```diff
@@ -80,6 +82,28 @@ KVStore::~KVStore() {
     }
 }
 
+int64_t KVStore::calc_rocksdb_write_buffer_size() {
+    // 1. Get the number of disks
+    std::vector<starrocks::StorePath> paths;
+    parse_conf_store_paths(config::storage_root_path, &paths);
+    int64_t disk_cnt = paths.size();
+    // 2. Get the total memory bytes of BE
+    int64_t total_mem_bytes = MemInfo::physical_mem();
```
trueeyu, line 91: better to use process_mem_limit() instead of MemInfo::physical_mem

luohaha: @trueeyu exec_env will be init after kvstore init...

luohaha: So I don't think we can use that.

kevincai: then it will be problematic in case of hybrid deployment in a single large VM. the mem_limit can be half of the physical mem. Possibly taking config::mem_limit into consideration here.

be/src/common/config.h
```diff
@@ -1369,6 +1369,13 @@ CONF_String(rocksdb_cf_options_string, "block_based_table_factory={block_cache={
 
 CONF_String(rocksdb_db_options_string, "create_if_missing=true;create_missing_column_families=true");
 
+// It is the memory percent of write buffer for meta in rocksdb.
+// default is 5% of system memory.
+CONF_mInt64(rocksdb_write_buffer_memory_percent, "5");
```
kevincai, line 1374: if the parameter won't be take effect in runtime, better define it as immutable, so it is clear that the parameter can only take effect after restart.

be/src/common/config.h
```diff
@@ -1369,6 +1369,13 @@ CONF_String(rocksdb_cf_options_string, "block_based_table_factory={block_cache={
...
+// It is the max size of the write buffer for meta in rocksdb.
+// default is 1GB.
+CONF_mInt64(rocksdb_max_write_buffer_memory_bytes, "1073741824");
```
kevincai, line 1377: Also update the BE_Configuration.md doc as well.

be/src/common/config.h
kevincai: shall also explain that the minimal size is 64MB even if `total_mem * percent < 64MB.`

be/src/storage/kv_store.cpp
luohaha: Sorry, I think i mixup the `GlobalEnv` and `exec_env`, process mem tracker will init in `GlobalEnv::init`, and `GlobalEnv::init` will happen before kvstore init, so I think I can use `process_mem_limit` instead of `physical_mem`.

luohaha: done

be/src/common/config.h
luohaha: done

be/src/common/config.h
luohaha: done

kevincai (dismissed):

wyb (approved):

kevincai (approved):

PR #58737, opened 2025-05-07
[BugFix] fix partial compaction's segment deleted by abort transaction

starrocks-xupeng:
## What I'm doing:

in partial compaction, some segments in rowset might be moved into new rowset, when compaction is aborted, those segments should not be deleted.

for upgrading, if compaction is aborted during upgrading, then its segments will not be deleted, which should be fine.
for downgrading, the behavior is the same, if compaction is aborted during downgrading, then its segments will be wrongly deleted, since this utility is turned off by default, so it's fine.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/storage/lake/transactions.cpp
+++ b/be/src/storage/lake/transactions.cpp
@@ -452,8 +452,7 @@ Status publish_log_version(TabletManager* tablet_mgr, int64_t tablet_id, std::sp
     return Status::OK();
 }
 
-static void collect_files_in_log(TabletManager* tablet_mgr, const TxnLog& txn_log,
-                                 std::vector<std::string>* files_to_delete) {
+void collect_files_in_log(TabletManager* tablet_mgr, const TxnLog& txn_log, std::vector<std::string>* files_to_delete) {
     auto tablet_id = txn_log.tablet_id();
     if (txn_log.has_op_write()) {
         for (const auto& segment : txn_log.op_write().rowset().segments()) {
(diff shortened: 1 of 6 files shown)


be/src/storage/lake/transactions.cpp
```diff
@@ -464,8 +463,12 @@ static void collect_files_in_log(TabletManager* tablet_mgr, const TxnLog& txn_lo
         }
     }
     if (txn_log.has_op_compaction()) {
-        for (const auto& segment : txn_log.op_compaction().output_rowset().segments()) {
-            files_to_delete->emplace_back(tablet_mgr->segment_location(tablet_id, segment));
+        // only delete actual new segments
+        size_t new_segment_offset = txn_log.op_compaction().new_segment_offset();
+        size_t new_segment_count = txn_log.op_compaction().new_segment_count();
+        auto& segments = txn_log.op_compaction().output_rowset().segments();
```
xiangguangyxg, line 469:
```suggestion
        const auto& segments = txn_log.op_compaction().output_rowset().segments();
```

gensrc/proto/lake_types.proto
```diff
@@ -188,6 +188,13 @@ message TxnLogPB {
         optional PersistentIndexSstablePB output_sstable = 4;
         // Base version of compaction task, used for conflict check with partial update
         optional int64 compact_version = 5;
+        // For partial compaction, we put compacted segments and uncompacted segments
+        // together, and when compaction is aborted, we need to avoid uncompacted segments
+        // being deleted.
+        // e.g., a b x y e f, in `output_rowset`, a,b,e,f might be uncompacted segments,
+        // x and y might be new segments
```
xiangguangyxg, line 195: Add some comments for normal compaction ?

gensrc/proto/lake_types.proto
```diff
@@ -188,6 +188,13 @@ message TxnLogPB {
...
+        optional int32 new_segment_offset = 6;
+        optional int32 new_segment_count = 7;
```
xiangguangyxg, line 197: Is there compatibility issues when upgrading or downgrading ?

xiangguangyxg (dismissed):

gensrc/proto/lake_types.proto
kevincai: is it safe to assume the new segments are all in sequence with [offset, offset +cout), is it better to have an array to record all the indexes of the new segments in the output_rowset, so the assumption of the sequence of new rowset will be removed.

kevincai (dismissed):

kevincai (approved):

xiangguangyxg (approved):

wangruin (approved):

starrocks-xupeng:
https://github.com/StarRocks/starrocks/pull/58866
3.3

PR #58826, opened 2025-05-09
[BugFix] fix wrong transform from inner join to semi

before-Sunrise:
## Why I'm doing:
for sql like distinct on inner join, we can transform inner join to semi join if inner join's output is only from one child. But if join OptExpression has projection, then join OptExpression's output columns may not same as join operator's output columns if projection is like "ColumnRef(1): ColumnRef(2)"

## What I'm doing:
1. if join OptExpression has projection, use projection's used column as join operator's output column
2. only transfrom inner to semi if inner join's output is only from one child. 
Fixes [#issue](https://github.com/StarRocks/StarRocksTest/issues/9593), it should transfrom inner to left semi, but tranform it into right semi which cause empty set

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/test/java/com/starrocks/sql/plan/InnerToSemiTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/sql/plan/InnerToSemiTest.java
@@ -17,6 +17,7 @@ import com.starrocks.catalog.OlapTable;
 import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.sql.optimizer.statistics.ColumnStatistic;
 import com.starrocks.sql.optimizer.statistics.StatisticStorage;
+import com.starrocks.utframe.StarRocksAssert;
 import mockit.Expectations;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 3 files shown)


kangkaisen (approved):

fe/fe-core/src/test/java/com/starrocks/sql/plan/InnerToSemiTest.java
```diff
@@ -90,5 +119,10 @@ public void testInnerToSemi() throws Exception {
         assertContains(plan, " LEFT SEMI JOIN (join-predicate [1: v1 = 7: v7] post-join-predicate [null])\n" +
                 "        LEFT SEMI JOIN (join-predicate [1: v1 = 4: v4] post-join-predicate [null])");
 
+        sql = "select distinct t1.col_struct.c_char from duplicate_table_struct t1 " +
```
stephen-shelby, line 122: please add a test to cover TPCHQ9

stephen-shelby (approved):

PR #58836, opened 2025-05-09
[Enhancement] Support materialized view concurrent prepare

kaijianding:
## Why I'm doing:
mv prepare costs more than 1 seconds where there are lots of mv, we can use threadPool to reduce time costs

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -3357,6 +3357,9 @@ public class Config extends ConfigBase {
             "materialized views(>10) or query is complex(multi table joins).")
     public static long mv_query_context_cache_max_size = 1000;
 
+    @ConfField(mutable = true)
+    public static boolean enable_materialized_view_concurrent_prepare = true;
+
     /**
      * Checking the connectivity of port opened by FE,
      * mainly used for checking edit log port currently.
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/MvRewritePreprocessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/MvRewritePreprocessor.java
@@ -46,6 +46,7 @@ import com.starrocks.catalog.mv.MVPlanValidationResult;
 import com.starrocks.common.AnalysisException;
 import com.starrocks.common.Config;
 import com.starrocks.common.Pair;
+import com.starrocks.common.ThreadPoolManager;
 import com.starrocks.common.profile.Timer;
 import com.starrocks.common.profile.Tracers;
 import com.starrocks.common.util.DebugUtil;
@@ -98,7 +99,7 @@ public class MvRewritePreprocessor {
     private static final Logger LOG = LogManager.getLogger(MvRewritePreprocessor.class);
 
     private static final Executor MV_PREPARE_EXECUTOR = Executors.newFixedThreadPool(
-            Runtime.getRuntime().availableProcessors(),
+            ThreadPoolManager.cpuIntensiveThreadPoolSize(),
             new ThreadFactoryBuilder().setDaemon(true).setNameFormat("mv-prepare-%d").build());
 
     private final ConnectContext connectContext;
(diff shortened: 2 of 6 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/MvRewritePreprocessor.java
```diff
@@ -83,12 +85,22 @@
...
 import java.util.stream.Collectors;
 
+import static com.google.common.util.concurrent.MoreExecutors.newDirectExecutorService;
 import static com.starrocks.sql.optimizer.OptimizerTraceUtil.logMVPrepare;
 
 public class MvRewritePreprocessor {
     private static final Logger LOG = LogManager.getLogger(MvRewritePreprocessor.class);
+
+    private static final Executor MV_PREPARE_EXECUTOR = Executors.newFixedThreadPool(
+            Runtime.getRuntime().availableProcessors(),
```
LiShuMing, line 101: This is a bit heavy for mv prepare processor tasks, can we reduce it for a less value?

LiShuMing: How about ` Math.max(4, Runtime.getRuntime().availableProcessors() / 4)`

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/MvRewritePreprocessor.java
```diff
@@ -661,36 +673,40 @@ public void prepareRelatedMVs(Set<Table> queryTables,
...
         }
 
+        List<Pair<MaterializedViewWrapper, MvUpdateInfo>> mvInfos =
+                Lists.newArrayListWithExpectedSize(mvWithPlanContexts.size());
         for (MaterializedViewWrapper wrapper : mvWithPlanContexts) {
             MaterializedView mv = wrapper.getMV();
-            MvPlanContext mvPlanContext = wrapper.getMvPlanContext();
             try {
                 // mv's partitions to refresh
                 MvUpdateInfo mvUpdateInfo = queryMaterializationContext.getOrInitMVTimelinessInfos(mv);
```
LiShuMing, line 682: This is also a heavy operation, can add it into the future too?

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3354,6 +3354,9 @@ public class Config extends ConfigBase {
             "materialized views(>10) or query is complex(multi table joins).")
     public static long mv_query_context_cache_max_size = 1000;
 
+    @ConfField(mutable = true)
+    public static boolean enable_materialized_view_concurrent_prepare = true;
```
LiShuMing, line 3358: Update the `FE Configuration` docs in english?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/MvRewritePreprocessor.java
kaijianding: > This is also a heavy operation, can add it into the future too?

getOrInitMVTimelinessInfos() has multiple implements, it may introduce more concurrent issues.
like `logMVPrepare` is not designed for multi-thread

fe/fe-core/src/main/java/com/starrocks/common/Config.java
kaijianding: done

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/MvRewritePreprocessor.java
kaijianding: done

LiShuMing (approved):
+1

satanson (approved):

Seaven (approved):

Issue #53117, opened 2024-11-22
[Feature] Support TTL for List Partition Tables and Materialized Views


LiShuMing:
- [x] https://github.com/StarRocks/starrocks/pull/53819
- [x] https://github.com/StarRocks/starrocks/pull/53488 
- [x] https://github.com/StarRocks/starrocks/pull/53478
- [x] https://github.com/StarRocks/starrocks/pull/53139
- [x] https://github.com/StarRocks/starrocks/pull/53118

## Feature request

**Is your feature request related to a problem? Please describe.**

1. If the partition method of a Native Table is List Partition, it currently does not support Partition TTL, which means partitions can only be deleted manually and cannot be automatically managed;  https://github.com/StarRocks/starrocks/pull/43539  has already supported batch deletion of partitions, which can improve ease of use to some extent, but there are still many inconvenient aspects.

2. MV supports multi-level partitioning by internally mapping to NativeTable as List Partition type; to achieve automatic partition management for the purpose of "hot-cold data tiering," it is necessary to clarify the semantics of Partition TTL, especially under the List Partition type.

**Describe the solution you'd like**

1. Support for `ALTER TABLE xxx DROP PARTITION WHERE xxxx ` universal partition expression;
2. Support for single-level and multi-level Partition TTL semantics in Native Table, and support for universal partition TTL expressions;
3. Support for single-level and multi-level Partition TTL semantics in MV when mapped to List Partition type inner tables, and support for universal partition TTL expressions;

```
CREATE TABLE t1 (
    dt varchar(20),
    province string,
    num int
)
DUPLICATE KEY(dt)
PARTITION BY (`dt`, `province`)
PROPERTIES (
--     'partition_ttl_number' = '1', -- NOT WORKS
--     'partition_ttl' = '1 day', --  NOT WORKS
    "partition_redention_condition" = "str2date(dt, '%Y-%m-%d') <= CURRENT_DATE() - INTERVAL 3 MONTH",
    "replication_num" = "1"
);

-- drop partitions supports common expressions
ALTER TABLE t1 DROP PARTITIONS WHERE str2date(dt, '%Y-%m-%d') <= CURRENT_DATE() - INTERVAL 3 MONTH;

-- MV supports common expressions
CREATE MATERIALIZED VIEW mv1 
PARTITION BY (dt, province)
REFRESH DEFERRED MANUAL 
PROPERTIES (
[...]

PR #58867, opened 2025-05-12
[BugFix] Fix bugs with partition_condition_retention

LiShuMing:
## Why I'm doing:
https://github.com/StarRocks/starrocks/issues/53117 has supported `partition_condition_retention` property for more generic partition TTL(eg: multi partition columns or more generic partition expressions).

But there are some bugs in using:
1. For `range-partitioned` table, scheduler may fail to drop expired partitions for needs of `ThreadLocal`'s `ConnectContext`;
2. For `list-partitioned` table, it may drop wrong partitions if it's a multi-values partition;
3. `ShowCreateTable` cannot display `partition_condition_retention`;

#### Problem 1

Add a `Guard` to protect this, since connect context is needed sometime, otherwise ConnectContext.get() is null.
```
    public void scheduleTTLPartition() {
        // connect context is needed sometime, otherwise ConnectContext.get() is null.
        ConnectContext connectContext = ConnectContext.buildInner();
        try (var  guard = connectContext.bindScope()) {
            doScheduleTTLPartition();
        } catch (Exception e) {
            LOG.warn("Failed to schedule ttl partition", e);
        }
    }
```
#### Problem 2
Distinguish `drop partition by where` and `partition condition retention` which may cause different selections by `isRecyclingCondition`.

for recycling condition(drop partition with where), keep partitions as less as possible：
```
  // T1, partitions:
  //  p1: [1, 2]
  //  p2: [2, 3]
  // eg: alter table T1 drop partitions where p > 1, only choose p1 when all values are satisfied.
  // p1: [1, 2] is not satisfied, so we need to return p2: [2, 3]
```

for retention condition, keep partitions as much as possible:
```

  // T1, partitions:
  //  p1: [1, 2]
  //  p2: [2, 3]
  // eg: partition_retention_condition = 'p > 1', choose partition if it once is satisfied.
  // p1: [1, 2]/[2, 3] are all satisfied, so we need to return p1/p2 both
```
#### Problem 3

- add  `partition_condition_retention` for  `ShowCreateTable` & support set  `partition_condition_retention` to empty.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/clone/DynamicPartitionScheduler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/clone/DynamicPartitionScheduler.java
@@ -467,14 +467,12 @@ public class DynamicPartitionScheduler extends FrontendDaemon {
                 for (Table table : GlobalStateMgr.getCurrentState().getLocalMetastore().getTables(dbId)) {
                     // register dynamic partition table
                     if (DynamicPartitionUtil.isDynamicPartitionTable(table)) {
-                        LOG.info("register table[{}] with dynamic partition", table.getName());
                         registerDynamicPartitionTable(db.getId(), table.getId());
                         dynamicPartitionTables.computeIfAbsent(db.getFullName(), k -> new ArrayList<>())
                                     .add(table.getName());
                     }
                     // register ttl partition table
                     if (DynamicPartitionUtil.isTTLPartitionTable(table)) {
-                        LOG.info("register table[{}] with ttl partition", table.getName());
                         // Table(MV) with dynamic partition enabled should not specify partition_ttl_number(MV) or
                         // partition_live_number property.
                         registerTtlPartitionTable(db.getId(), table.getId());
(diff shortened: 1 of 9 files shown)


satanson (approved):

fe/fe-core/src/main/java/com/starrocks/clone/DynamicPartitionScheduler.java
```diff
@@ -500,12 +500,13 @@ public void runOnceForTest() {
     protected void runAfterCatalogReady() {
         // Find all tables that need to be scheduled.
         long now = System.currentTimeMillis();
-        if ((now - lastFindingTime) > Math.max(300000, Config.dynamic_partition_check_interval_seconds)) {
+        long checkIntervalMs = Config.dynamic_partition_check_interval_seconds * 1000L;
+        if ((now - lastFindingTime) > Math.max(60000, checkIntervalMs)) {
```
LiShuMing, line 504: for better testing, reduce min schedule time from 5min to 1min but original config has not changed.

HangyuanLiu (approved):

Seaven (approved):

Issue #58316, opened 2025-04-23
[Feature] Cost optimization for shared-data arch in real-time scenarios


luohaha:
## Feature request

**Is your feature request related to a problem? Please describe.**

We found that the transaction model of StarRocks' current shared-data architecture generates excessive object storage PUT calls in real-time ingestion scenarios, leading to high API costs and negatively impacting import performance.

Therefore, the goal of this design is to minimize the cost of S3 API calls in real-time scenarios as much as possible. Ultimately, we aim to achieve:
1. S3 API costs approaching or even lower than storage costs, giving CelerData a cost-performance advantage 
    compared to competitors:
  > Data freshness is around 30 seconds.
  > Out-of-the-box, which means reducing API costs without any tuning.
 
2. Minimal impact on existing query and ingestion performance.

**Describe the solution you'd like**
1. We reduce API calls by aggregating files as much as possible:
  > We combine data file from per tablet to per BE/CN.
  > We combine TxnLog/Meta file from per tablet to per transaction.

2. Because transactions related to ingestion and compaction represent the highest proportion, our current optimizations target only these two transaction workflows.

**Additional context**
https://docs.google.com/document/d/1CYHwzer_57hbU-hcQ5UcCQshxaG0h1gkvFjyuMldAlw/edit?usp=sharing

PR #58892, opened 2025-05-12
[Enhancement] Submit write combined_txn_log/aggregate_tablet_metadata to thread pool to avoid block brpc thread

sevev:
## Why I'm doing:
The combined_txn_log and aggregated_tablet_metadata are written to remote storage directly in the BRPC threads, which may block the BRPC threads. This PR submits these tasks to a thread pool for execution to avoid blocking BRPC threads.

## What I'm doing:

Fixes #58316

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/runtime/exec_env.cpp
+++ b/be/src/runtime/exec_env.cpp
@@ -575,6 +575,11 @@ Status ExecEnv::init(const std::vector<StorePath>& store_paths, bool as_cn) {
     _lake_tablet_manager =
             new lake::TabletManager(_lake_location_provider, _lake_update_manager, config::lake_metadata_cache_limit);
     _lake_replication_txn_manager = new lake::ReplicationTxnManager(_lake_tablet_manager);
+    RETURN_IF_ERROR(ThreadPoolBuilder("put_aggregate_metadata_pool")
+                            .set_min_threads(1)
+                            .set_max_threads(1)
+                            .set_max_queue_size(std::numeric_limits<int>::max())
+                            .build(&_put_aggregate_metadata_thread_pool));
 #endif
 
     _agent_server = new AgentServer(this, false);
--- a/be/src/runtime/exec_env.cpp
+++ b/be/src/runtime/exec_env.cpp
@@ -557,7 +557,6 @@ Status ExecEnv::init(const std::vector<StorePath>& store_paths, bool as_cn) {
     }
     setenv(staros::starlet::fslib::kFslibCacheDir.c_str(), config::starlet_cache_dir.c_str(), 1);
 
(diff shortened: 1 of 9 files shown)


be/src/runtime/exec_env.cpp
```diff
@@ -789,13 +789,29 @@ Status ExecEnv::init(const std::vector<StorePath>& store_paths, bool as_cn) {
     }
     setenv(staros::starlet::fslib::kFslibCacheDir.c_str(), config::starlet_cache_dir.c_str(), 1);
 
+    std::unique_ptr<ThreadPool> put_aggregate_metadata_thread_pool;
```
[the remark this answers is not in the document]
sevev: > Use `put_aggregate_file_thread_pool` or other name? Because it's used for metadata and txnlog, no only for metadata.

The writing of `combined_txn_log` uses a separate thread pool, which already existed in the previous implementation. Therefore, the thread pools for writing combined_txn_log and aggregate_metadata are distinct.

Furthermore, since these two operations (combined_txn_log writing and aggregate_metadata writing) occur at different stages of the process, maintaining separate thread pools for them may be better.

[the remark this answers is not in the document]
luohaha: ok, got it

luohaha (dismissed):

be/src/runtime/exec_env.cpp
kevincai, line 792: what does this line for? seems useless.

be/src/runtime/exec_env.cpp
```diff
@@ -789,13 +789,29 @@ Status ExecEnv::init(const std::vector<StorePath>& store_paths, bool as_cn) {
...
+    int32_t max_thread_count = config::transaction_publish_version_worker_count;
```
kevincai, line 793: consider make this configvar dynamic changeable so it will be much easier to update the configvar and take effect without restarting the BE process.

kevincai (dismissed):

be/src/runtime/exec_env.cpp
sevev: > consider make this configvar dynamic changeable so it will be much easier to update the configvar and take effect without restarting the BE process.

done.

kevincai (dismissed):

luohaha (approved):

kevincai (approved):

PR #58911, opened 2025-05-13
[BugFix] reduce lock contention of TableMetricsManager

silverbullet233:
## What I'm doing:

1. Use parallel_flat_hash_map instead of flat_hash_map in TableMetricsManager to reduce contention by reducing the granularity of the lock
2. Use BThreadSharedMutex instead of std::shared_mutex, because I noticed that LakeChannel may acquire locks in bthread. **we should note that BThreadSharedMutex's performance is not as good as std::shared_mutex, but in this scenario we have no choice.**
3. Add _metric_labels map in MetricCollector to speed up deletion

Add two BE configurations
1. enable_table_metrics: used to control whether to enable table metrics
2. max_table_metrics_num: used to limit the number of metrics registered in MetricRegistry to avoid the metrics interface returning too much data.

before optimization
```
-----------------------------------------------------------------------------------------------------
Benchmark                                                           Time             CPU   Iterations
-----------------------------------------------------------------------------------------------------
TableMetricsBench/BM_RegisterTable/iterations:1/threads:1         284 ms          284 ms            1
TableMetricsBench/BM_RegisterTable/iterations:1/threads:2         319 ms          368 ms            2
TableMetricsBench/BM_RegisterTable/iterations:1/threads:4         267 ms          308 ms            4
TableMetricsBench/BM_RegisterTable/iterations:1/threads:8         244 ms          282 ms            8
TableMetricsBench/BM_RegisterTable/iterations:1/threads:16        225 ms          260 ms           16
TableMetricsBench/BM_RegisterTable/iterations:1/threads:32        217 ms          253 ms           32
TableMetricsBench/BM_UnregisterTable/threads:1                   2.44 ms         2.43 ms          287
TableMetricsBench/BM_UnregisterTable/threads:2                   4.17 ms         6.75 ms          100
TableMetricsBench/BM_UnregisterTable/threads:4                   5.66 ms         17.5 ms           40
TableMetricsBench/BM_UnregisterTable/threads:8                   8.20 ms         57.0 ms           16
TableMetricsBench/BM_UnregisterTable/threads:16                  7.95 ms          116 ms           16
TableMetricsBench/BM_UnregisterTable/threads:32                  8.10 ms          248 ms           32
TableMetricsBench/BM_GetTableMetrics/threads:1                   2.36 ms         2.36 ms          296
TableMetricsBench/BM_GetTableMetrics/threads:2                   1.77 ms         3.54 ms          188
TableMetricsBench/BM_GetTableMetrics/threads:4                   1.71 ms         6.85 ms          100
TableMetricsBench/BM_GetTableMetrics/threads:8                   1.72 ms         13.8 ms           48
TableMetricsBench/BM_GetTableMetrics/threads:16                  1.66 ms         26.6 ms           32
TableMetricsBench/BM_GetTableMetrics/threads:32                  1.43 ms         45.7 ms           32
TableMetricsBench/BM_Cleanup                                    16023 ms        16020 ms            1
```

after optimization
```
-----------------------------------------------------------------------------------------------------
Benchmark                                                           Time             CPU   Iterations
-----------------------------------------------------------------------------------------------------
TableMetricsBench/BM_RegisterTable/iterations:1/threads:1         292 ms          292 ms            1
TableMetricsBench/BM_RegisterTable/iterations:1/threads:2         197 ms          234 ms            2
TableMetricsBench/BM_RegisterTable/iterations:1/threads:4         105 ms          149 ms            4
TableMetricsBench/BM_RegisterTable/iterations:1/threads:8        63.3 ms          145 ms            8
TableMetricsBench/BM_RegisterTable/iterations:1/threads:16       54.2 ms          230 ms           16
TableMetricsBench/BM_RegisterTable/iterations:1/threads:32       32.8 ms          268 ms           32
TableMetricsBench/BM_UnregisterTable/threads:1                  0.707 ms        0.707 ms          974
TableMetricsBench/BM_UnregisterTable/threads:2                  0.623 ms         1.24 ms          512
TableMetricsBench/BM_UnregisterTable/threads:4                  0.752 ms         2.58 ms          296
TableMetricsBench/BM_UnregisterTable/threads:8                   2.04 ms         9.93 ms           56
TableMetricsBench/BM_UnregisterTable/threads:16                  3.00 ms         22.4 ms           32
TableMetricsBench/BM_UnregisterTable/threads:32                  3.69 ms         32.4 ms           32
TableMetricsBench/BM_GetTableMetrics/threads:1                  0.867 ms        0.867 ms          801
TableMetricsBench/BM_GetTableMetrics/threads:2                   2.00 ms         3.86 ms          216
TableMetricsBench/BM_GetTableMetrics/threads:4                   1.87 ms         6.88 ms          108
TableMetricsBench/BM_GetTableMetrics/threads:8                   2.35 ms         13.0 ms           56
TableMetricsBench/BM_GetTableMetrics/threads:16                  3.22 ms         25.8 ms           32
TableMetricsBench/BM_GetTableMetrics/threads:32                  3.60 ms         37.2 ms           32
TableMetricsBench/BM_Cleanup                                     37.9 ms         37.9 ms           19
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1


kevincai:
need bump this with high priority, and possibly add runtime config to turn off the table metrics completely.

be/CMakeLists.txt
```diff
@@ -268,7 +268,7 @@ set_target_properties(openssl PROPERTIES IMPORTED_LOCATION ${THIRDPARTY_DIR}/lib
 add_library(leveldb STATIC IMPORTED GLOBAL)
 set_target_properties(leveldb PROPERTIES IMPORTED_LOCATION ${THIRDPARTY_DIR}/lib/libleveldb.a)
 
-if ("${CMAKE_BUILD_TYPE}" STREQUAL "DEBUG" OR "${CMAKE_BUILD_TYPE}" STREQUAL "RELEASE")
+if (("${CMAKE_BUILD_TYPE}" STREQUAL "DEBUG" OR "${CMAKE_BUILD_TYPE}" STREQUAL "RELEASE") AND ${WITH_BENCH} STREQUAL "OFF")
```
kevincai: can set to static link to bench binary specifically to the binary instead of change this globally for all.

be/src/bench/table_metrics_mgr_bench.cpp
```diff
@@ -0,0 +1,68 @@
...
+    std::unique_ptr<TableMetricsManager> manager;
+};
+
+BENCHMARK_DEFINE_F(TableMetricsBench, BM_RegisterTable)(benchmark::State& state) {
+    const int thread_id = state.thread_index;
+
+    for (auto _ : state) {
+        uint64_t start_id = thread_id;
+        for (uint64_t i = 0; i < kMaxTableId; ++i) {
+            manager->register_table(table_ids[(i + start_id) % kMaxTableId]);
```
kevincai, line 59: need test with table registration and table unregistration and background cleanup together to reflect the real world usage.

kevincai, line 57: need test the following ops contention to reflect the real world usage.
* table metrics registration
* table metrics unregistration
* table metrics get
* table metrics background cleanup.

kevincai: and also the metrics registery may be pulled every 15 seconds by external metrics collection system such as prometheus.

be/src/util/table_metrics.cpp
```diff
@@ -40,24 +39,78 @@ void TableMetrics::uninstall(MetricRegistry* registry) {
...
-            ++iter;
+#endif
+    // metrics should be deleted from _metrics_map
+    std::vector<std::pair<uint64_t, TableMetricsPtr>> delete_metrics;
+    // metrics should be uninstalled from _metrics
+    std::vector<TableMetricsPtr> uninstalled_metrics;
+    _metrics_map.for_each([&](const auto& pair) {
+        if (pair.second->ref_count == 0) {
+            delete_metrics.emplace_back(pair.first, pair.second);
+            if (pair.second->installed) {
```
banmoy: could judge the flag when erasing the metric so that can save the cost of `uninstalled_metrics`?

be/src/util/table_metrics.cpp
```diff
@@ -40,24 +39,78 @@ void TableMetrics::uninstall(MetricRegistry* registry) {
     UNREGISTER_TABLE_METRIC(load_rows);
 }
 
-void TableMetricsManager::cleanup() {
+bool TableMetricsManager::can_install_metrics() {
+    int64_t old_num = 0;
+    do {
+        old_num = _installed_metrics_num.load();
```
banmoy, line 45: I think we should install the metrics of the top scan/load tables which are dynamic, rather than the first registered tables which are relatively static.  If select the top, the implementation might be complex. perhaps we can consider it later.

silverbullet233: > I think we should install the metrics of the top scan/load tables which are dynamic, rather than the first registered tables which are relatively static. If select the top, the implementation might be complex. perhaps we can consider it later.

@banmoy  good question. I have considered this before. Metrics is a scenario where writes far outweigh reads. The cost of maintaining the top is very high. I doubt the value of doing this in the kernel. I prefer to leave it to an external system.

be/src/common/config.h
```diff
@@ -1705,6 +1705,8 @@ CONF_mInt32(put_combined_txn_log_thread_pool_num_max, "64");
 CONF_mBool(enable_put_combinded_txn_log_parallel, "false");
 // used to control whether the metrics/ interface collects table metrics
 CONF_mBool(enable_collect_table_metrics, "true");
+CONF_Bool(enable_table_metrics, "true");
```
kevincai: 1. is it possible to make runtime mutable so it can be turned on/off without service restart.
2. add comments to explain the new configuration.
3. add doc about the new configuration items.

be/src/http/action/metrics_action.cpp
```diff
@@ -334,17 +328,21 @@ void MetricsAction::handle(HttpRequest* req) {
     if (type == "core") {
         SimpleCoreMetricsVisitor visitor;
         _metrics->collect(&visitor);
+        _collect_table_metrics(&visitor);
```
kevincai: should not collect table metrics if type == core.

be/src/util/table_metrics.cpp
kevincai: Per discussed with @silverbullet233 , this is a temp simple solution to prevent too many tables registered that may affect the system stability. Decent design and use of the table metrics will be revisited later without blocking the urgency of risk mitigation.

be/src/util/table_metrics.cpp
```diff
@@ -40,24 +39,78 @@ void TableMetrics::uninstall(MetricRegistry* registry) {
...
+                ctor(table_id, metrics_ptr);
+            });
+
+    if (is_created && metrics_ptr->installed) {
+        metrics_ptr->install(&_metrics, std::to_string(table_id));
+    }
+}
+
+void TableMetricsManager::unregister_table(uint64_t table_id) {
+    if (!config::enable_table_metrics) {
```
kevincai, line 77: no need to block unregister_table, because the table can be registered and then the configuration can be turned off. let the unregister always possible, just blocking the registering.

be/src/util/table_metrics.cpp
```diff
@@ -40,24 +39,78 @@ void TableMetrics::uninstall(MetricRegistry* registry) {
...
+        return;
+    }
+    _metrics_map.modify_if(table_id, [](TableMetricsPtr& metrics_ptr) { metrics_ptr->ref_count--; });
+}
+
+void TableMetricsManager::cleanup(bool force) {
+    if (!config::enable_table_metrics) {
```
kevincai, line 84: is it possible to do quick `clear()` of all the metrics if not enabled?

be/src/util/table_metrics.h
```diff
@@ -44,53 +51,43 @@ using TableMetricsPtr = std::shared_ptr<TableMetrics>;
...
-    }
-    void unregister_table(uint64_t table_id) {
-        std::unique_lock l(_mu);
-        DCHECK(_metrics_map.contains(table_id));
-        auto metrics_ptr = _metrics_map.at(table_id);
-        metrics_ptr->ref_count--;
-    }
+    TableMetricsManager() : _last_cleanup_ts(MonotonicSeconds()) {}
+
+    MetricRegistry* metrics() { return &_metrics; }
```
kevincai:
```suggestion
    MetricRegistry* metric_registry() { return &_metrics; }
```

kevincai: may not be a good practice to expose this metric registry directly to outside the TableMetricsManager, suggest and `visit()` interface so the visitor of the metrics in this metric registry can be `TableMetricsManager()->visit(visitor);`

be/src/util/table_metrics.cpp
```diff
@@ -40,24 +39,78 @@ void TableMetrics::uninstall(MetricRegistry* registry) {
     UNREGISTER_TABLE_METRIC(load_rows);
 }
 
-void TableMetricsManager::cleanup() {
+bool TableMetricsManager::can_install_metrics() {
```
kevincai, line 42: If considering this limit of number of table registration is just a protection of the system, the accuracy can be compromised a little bit, and the consequence is nearly nothing, here can just check the size of the phmap and determine reject or accept the registration. Would it be simple and easy to understand?

be/src/util/table_metrics.cpp
silverbullet233: fixed

be/src/common/config.h
silverbullet233: for 1, because the deletion of metrics is based on reference counts, making it a dynamic switch may miss some reference count updates, which will make cleanup impossible. so we have to keep it static and close it by default to avoid risks.  2 and 3 have been fixed.

be/src/http/action/metrics_action.cpp
silverbullet233: fixed

be/src/util/table_metrics.cpp
silverbullet233: since this configuration is static, we need to make sure it doesn't have any effect.

be/src/util/table_metrics.h
silverbullet233: MetricRegistry already has a `collect` method for different visitor, I think we can just use it directly.

be/src/util/table_metrics.cpp
silverbullet233: `parallel_flat_hash_map::size` method requires the read lock on each submap, and this way is not completely accurate (although this is not our goal). I think the current approach may be cheaper

kevincai (approved):

banmoy (approved):

Seaven (approved):

PR #58942, opened 2025-05-13
[Enhancement] support pagination for information_schema.task_runs

murphyatwork:
## Why I'm doing:

This view retrieves all historical task run records, which can be substantial. Fetching all records in a single RPC call may result in a large response size, potentially leading to memory issues.

## What I'm doing:

Introduce pagination for the `information_schema.task_runs`, returning 1000 records in each RPC call.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
@@ -19,7 +19,6 @@
 #include "runtime/datetime_value.h"
 #include "runtime/runtime_state.h"
 #include "runtime/string_value.h"
-#include "util/timezone_utils.h"
 
 namespace starrocks {
 
@@ -52,27 +51,26 @@ Status SchemaTaskRunsScanner::start(RuntimeState* state) {
     std::string task_name;
     std::string query_id;
     std::string task_run_state;
-    TGetTasksParams task_params;
     // task_name
     if (_parse_expr_predicate("TASK_NAME", task_name)) {
-        task_params.__set_task_name(task_name);
+        _task_params.__set_task_name(task_name);
     }
     // query_id
     if (_parse_expr_predicate("QUERY_ID", query_id)) {
-        task_params.__set_query_id(query_id);
+        _task_params.__set_query_id(query_id);
     }
     // task_run_state
     if (_parse_expr_predicate("STATE", task_run_state)) {
-        task_params.__set_state(task_run_state);
+        _task_params.__set_state(task_run_state);
     }
     if (nullptr != _param->current_user_ident) {
-        task_params.__set_current_user_ident(*(_param->current_user_ident));
+        _task_params.__set_current_user_ident(*(_param->current_user_ident));
     }
-    if (_param->limit > 0) {
-        task_params.__isset.pagination = true;
-        task_params.pagination.__set_limit(_param->limit);
-    }
-    RETURN_IF_ERROR(SchemaHelper::get_task_runs(_ss_state, task_params, &_task_run_result));
+    _task_params.__isset.pagination = true;
+    _task_params.pagination.__set_offset(0);
+    _task_params.pagination.__set_limit(kPaginationBatchSize);
+    RETURN_IF_ERROR(SchemaHelper::get_task_runs(_ss_state, _task_params, &_task_run_result));
+    _task_params.pagination.__set_offset(_task_run_result.task_runs.size());
     _task_run_index = 0;
     return Status::OK();
 }
@@ -284,12 +282,25 @@ Status SchemaTaskRunsScanner::get_next(ChunkPtr* chunk, bool* eos) {
     if (!_is_init) {
         return Status::InternalError("Used before initialized.");
     }
-    if (_task_run_index >= _task_run_result.task_runs.size()) {
-        *eos = true;
-        return Status::OK();
-    }
     if (nullptr == chunk || nullptr == eos) {
         return Status::InternalError("input pointer is nullptr.");
+    }
+
+    // Streamingly fetch the next batch. task_run_result will be empty if EOS is reached.
+    if (_task_run_index >= _task_run_result.task_runs.size()) {
+        if (_param->limit > 0 && _task_params.pagination.offset >= _param->limit) {
+            *eos = true;
+            return {};
+        }
+        _task_run_result.task_runs.clear();
+        RETURN_IF_ERROR(SchemaHelper::get_task_runs(_ss_state, _task_params, &_task_run_result));
+        if (_task_run_result.task_runs.empty()) {
+            *eos = true;
+            return {};
+        }
+        _task_run_index = 0;
+        size_t offset = _task_params.pagination.offset + kPaginationBatchSize;
+        _task_params.pagination.__set_offset(offset);
     }
     *eos = false;
     return fill_chunk(chunk);
--- a/be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
@@ -70,7 +70,13 @@ Status SchemaTaskRunsScanner::start(RuntimeState* state) {
     _task_params.pagination.__set_offset(0);
     _task_params.pagination.__set_limit(kPaginationBatchSize);
     RETURN_IF_ERROR(SchemaHelper::get_task_runs(_ss_state, _task_params, &_task_run_result));
-    _task_params.pagination.__set_offset(_task_run_result.task_runs.size());
+    size_t finished_count = 0;
+    for (auto& run : _task_run_result.task_runs) {
+        if (!(run.state == "PENDING" || run.state == "RUNNING")) {
+            finished_count++;
+        }
+    }
+    _task_params.pagination.__set_offset(finished_count);
     _task_run_index = 0;
     return Status::OK();
 }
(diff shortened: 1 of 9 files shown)


be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
```diff
@@ -284,13 +282,26 @@ Status SchemaTaskRunsScanner::get_next(ChunkPtr* chunk, bool* eos) {
...
-    if (_task_run_index >= _task_run_result.task_runs.size()) {
-        *eos = true;
-        return Status::OK();
-    }
     if (nullptr == chunk || nullptr == eos) {
         return Status::InternalError("input pointer is nullptr.");
     }
+
+    // Streamingly fetch the next batch. task_run_result will be empty if EOS is reached.
+    if (_task_run_index >= _task_run_result.task_runs.size()) {
```
alvin-celerdata, line 290: I think you should write a UT to cover this case to make sure that the right result happens when pagination is enabled. Otherwise, how can you guarantee that your code is correct?

be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
```diff
@@ -284,13 +282,26 @@ Status SchemaTaskRunsScanner::get_next(ChunkPtr* chunk, bool* eos) {
...
+        if (_param->limit > 0 && _task_params.pagination.offset >= _param->limit) {
+            *eos = true;
+            return {};
+        }
+        _task_run_result.task_runs.clear();
+        RETURN_IF_ERROR(SchemaHelper::get_task_runs(_ss_state, _task_params, &_task_run_result));
```
alvin-celerdata, line 296: Just suppose the worst-case scenario: a new task is generated between two RPC calls. How could this loop stop? And what's your expected behavior in that situation?

be/src/exec/schema_scanner/schema_task_runs_scanner.h
```diff
@@ -28,9 +28,12 @@ class SchemaTaskRunsScanner : public SchemaScanner {
     Status get_next(ChunkPtr* chunk, bool* eos) override;
 
 private:
+    static constexpr int64_t kPaginationBatchSize = 1000;
```
alvin-celerdata, line 31: The default chunk size is 4096, it means that there will be 5 RPC calls to fill a chunk. I don't think it is an efficient choice. 
What is the estimated size for each result when page is 1000? If it is not that big, we can just use default chunk size or maybe more larger chunk to fetch result from FE to reduce RPC calls.

fe/fe-core/src/main/java/com/starrocks/scheduler/history/TaskRunHistoryTable.java
```diff
@@ -210,8 +210,17 @@ public List<TaskRunStatus> lookup(TGetTasksParams params) {
             predicates.add(" task_state = " + Strings.quote(params.getState()));
         }
         sql += Joiner.on(" AND ").join(predicates);
-        if (params.isSetPagination() && params.getPagination().getLimit() > 0) {
-            sql += " LIMIT " + params.getPagination().getLimit();
+
+        if (params.isSetPagination()) {
+            if (params.getPagination().getLimit() > 0 || params.getPagination().getOffset() > 0) {
```
alvin-celerdata, line 215: If only the limit is given, there is no need to order all `task runs`.

alvin-phoenix-ai (commented):
More tests should be added

be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
murphyatwork: added test cases

be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
murphyatwork: This cursor-based approach does not guarantee a stable snapshot, meaning it may return additional records if new entries are added to the table during the RPC fetching process. 
However, the loop will eventually stop as long as the RPC fetching speed exceeds the rate of record insertion, which I assume is typically the case, because the insertion is way heavier than reading. Therefore, we don't need to be concerned about this issue.

fe/fe-core/src/main/java/com/starrocks/scheduler/history/TaskRunHistoryTable.java
murphyatwork: In the case of `LIMIT 200 OFFSET 0`, all records still need to be sorted. Essentially, there's always an `OFFSET`.

be/src/exec/schema_scanner/schema_task_runs_scanner.h
murphyatwork: The `DEFAULT_CHUNK_SIZE` has been adjusted. 
The size of the RPC response varies significantly based on the task, as it includes a task definition. In practice, memory issues can occur if the total number of records exceeds 100,000, so it's advisable to limit it to under 10,000. 

Therefore, 4,096 could be an optimal choice.

fe/fe-core/src/main/java/com/starrocks/scheduler/history/TaskRunHistoryTable.java
```diff
@@ -210,8 +210,17 @@ public List<TaskRunStatus> lookup(TGetTasksParams params) {
...
+                sql += " ORDER BY `task_id`, `task_run_id`, `create_time`";
+            }
+            if (params.getPagination().getLimit() > 0) {
+                sql += " LIMIT " + params.getPagination().getLimit();
```
LiShuMing, line 219: Can `Pagination` cause result duplication or unstable if history table is changing?

LiShuMing: Another question, can we always add a `limit` for the history table to avoid fetching too many results?

murphyatwork: 1. Will it cause duplication? Technically, no. The `task_runs` consists of three components: running, in-memory, and in-table. Therefore, it can be enumerated using `offset + limit`.  
2. Is it stable? Not entirely, but it’s not a critical issue.  
3. How? If we want to retrieve all records, we can query something like: select count(*) from `information_schema.task_runs`.  Then you cannot add an implicit limit for the query

be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
```diff
@@ -52,27 +51,32 @@ Status SchemaTaskRunsScanner::start(RuntimeState* state) {
...
+    _task_params.pagination.__set_limit(kPaginationBatchSize);
+    RETURN_IF_ERROR(SchemaHelper::get_task_runs(_ss_state, _task_params, &_task_run_result));
+    size_t finished_count = 0;
+    for (auto& run : _task_run_result.task_runs) {
+        if (!(run.state == "PENDING" || run.state == "RUNNING")) {
+            finished_count++;
+        }
     }
-    RETURN_IF_ERROR(SchemaHelper::get_task_runs(_ss_state, task_params, &_task_run_result));
+    _task_params.pagination.__set_offset(finished_count);
```
alvin-celerdata, line 79: I don't understand the logic here; why can finished_count be set as an offset? Could you add some comments to the codebase?
My question is why you skip task_runs whose `(run.state == "PENDING" || run.state == "RUNNING")`, but you don't skip these in the `get_next` function when you change the offset.

be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
```diff
@@ -284,13 +288,27 @@ Status SchemaTaskRunsScanner::get_next(ChunkPtr* chunk, bool* eos) {
...
-        return Status::OK();
-    }
     if (nullptr == chunk || nullptr == eos) {
         return Status::InternalError("input pointer is nullptr.");
     }
+
+    // Fetch the next batch of task runs in a streaming manner. If the end of the stream is reached, task_run_result will be empty.
+    // For the first batch, include both RUNNING and PENDING tasks, and fetch up to kPaginationBatchSize historical tasks starting from offset 0.
+    // For subsequent batches, fetch historical tasks starting from the appropriate offset.
+    if (_task_run_index >= _task_run_result.task_runs.size()) {
```
alvin-celerdata, line 298: I'm wondering why we just fill one row for each call to this function? This is not efficient.

be/src/exec/schema_scanner/schema_task_runs_scanner.cpp
```diff
@@ -284,13 +288,27 @@ Status SchemaTaskRunsScanner::get_next(ChunkPtr* chunk, bool* eos) {
...
-    if (_task_run_index >= _task_run_result.task_runs.size()) {
-        *eos = true;
-        return Status::OK();
-    }
     if (nullptr == chunk || nullptr == eos) {
         return Status::InternalError("input pointer is nullptr.");
     }
+
+    // Fetch the next batch of task runs in a streaming manner. If the end of the stream is reached, task_run_result will be empty.
+    // For the first batch, include both RUNNING and PENDING tasks, and fetch up to kPaginationBatchSize historical tasks starting from offset 0.
```
alvin-celerdata, line 296: 1. How can you guarantee that all PENDING and RUNNING can be returned in the first chunk? 
2. Why do the RUNNING and PENDING tasks need to be filtered? If someone want to see the running and pending tasks, how can they achieve it, if you skip all RUNNING and PENDING tasks. 

Also, this is a behavior change modification, which will cause incompatibility when users upgrade to the version with this change. I need to submit a change request.

alvin-phoenix-ai (changes requested):

PR #58943, opened 2025-05-14
[Enhancement] Rename oidc to jwt

HangyuanLiu:
## Why I'm doing:
Because we use the mysql client, in order to correspond to the naming of mysql, we divide the authentication methods into OAuth and OIDC. But in fact, in the OIDC authentication method of mysql, what needs to be provided by the client is access token or id token (choose 1 from 2), not the complete OIDC token (access token + id token), which is also confusing in the design and naming of mysql.
And in StarRocks' complete lake authentication solution, iceberg needs access token, so in the current OIDC authentication method, users should provide access token for authentication and login, which is not the concept of complete OIDC token.
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -3552,7 +3552,7 @@ public class Config extends ConfigBase {
      * The URL to a JWKS service or a local file in the conf dir
      */
     @ConfField(mutable = false)
-    public static String oidc_jwks_url = "";
+    public static String jwt_jwks_url = "";
 
     /**
      * String to identify the field in the JWT that identifies the subject of the JWT.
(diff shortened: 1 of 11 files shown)


fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3552,29 +3552,29 @@ public class Config extends ConfigBase {
      * The URL to a JWKS service or a local file in the conf dir
      */
     @ConfField(mutable = false)
-    public static String oidc_jwks_url = "";
+    public static String jwt_jwks_url = "";
```
alvin-celerdata, line 3555: Should the document be changed accordingly?

HangyuanLiu: yes, Chelsea is working on document

nshangyimingCD (approved):

nshangyiming (approved):

gengjun-git (approved):

PR #58965, opened 2025-05-14
[Feature] Support tokenize function

### LINKED ISSUE #45145: Support tokenize function
- [x] https://github.com/StarRocks/starrocks/pull/58965
## Why I'm doing:
StarRocks has implemented GIN (Generalized Inverted Index), which works by tokenizing fields into individual tokens and building a dictionary out of them. This allows users to perform different semantic searches on this dictionary. However, due to the presence of various tokenizers, the results of tokenization can differ, making it not very intuitive for users to understand how the original field text is tokenized into specific tokens.

## What I'm doing:
Support a tokenize function, to allow users to work with specific tokenizer and get results of tokenization easily.

#### Function definition
```
function tokenize(tokenizer_name: string, content: string) -> list of strings
```
#### Input and output
```
// input 
tokenizer_name: needs to be limited to the existing tokenizers. For now, only support chinese, english, standard.
content: text, but notice that the language corresponding to the text content only achieves the expected effect when work with the specified tokenizer.

// output 
tokens: splited and analyzed by tokenizer
```

#### Example
```
// tokenize with english
mysql> SELECT tokenize('english', 'Today is saturday');
+------------------------------------------+
| tokenize('english', 'Today is saturday') |
+------------------------------------------+
| ["today","is","saturday"]                |
+------------------------------------------+
1 row in set (0.00 sec)

// count word frequency
mysql> select unnest, count(*) as count 
mysql> from t_tokenized_table, unnest(tokenize('english', english_text)) as unnest
mysql> group by unnest order by count;
+----------+-------+
| unnest   | count |
+----------+-------+
| world    |     1 |
| comes    |     1 |
| tap      |     1 |
| the      |     1 |
| from     |     1 |
| sea      |     1 |
| shanghai |     1 |
| water    |     1 |
| hello    |     2 |
+----------+-------+
9 rows in set (0.06 sec)
```

#### Notice
This function can __work independently__ without building GIN for a column, however, it is not advisable to invoke this function to tokenize and construct a dictionary on the massive data during the query time due to the poor performance. In fact, there is no need for the user to explicitly call the tokenize function to build a dictionary at the time of writing. Since both actions have the same behavior, this function is more suitable to be used to troubleshoot some search results that are difficult to understand.

hongkunxu:
## Why I'm doing:
The different results of tokenization provided by various tokenizers are too vague to users, so we need a tokenize function to allow users to figure it out easily.

We are currently researching inverted indexes, so we need a tokenize function to obtain more detailed tokenization results from the parser. I noticed that the original author, dujijun007, is no longer actively maintaining the PR or following up with the StarRocks community. Therefore, I copied the original PR and made some modifications based on the community’s feedback.

## What I'm doing:
Support a tokenize function, like tokenize(<tokenizer_name>, <content>)

Here is the function describe doc: https://docs.google.com/document/d/1oOegC8Qwud0_z08tgytx8KyZB3lwRZwfQCmO1kMw784/edit?usp=sharing

Fixes https://github.com/StarRocks/starrocks/issues/45145 https://github.com/StarRocks/starrocks/issues/61827

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/exprs/gin_functions.cpp
+++ b/be/src/exprs/gin_functions.cpp
@@ -1,0 +1,117 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "exprs/gin_functions.h"
+
+#include <CLucene.h>
+#include <CLucene/analysis/LanguageBasedAnalyzer.h>
+
+#include <boost/locale/encoding_utf.hpp>
+
+#include "column/array_column.h"
+#include "column/column_viewer.h"
+#include "column/datum.h"
+#include "util/faststring.h"
+
+namespace starrocks {
+
+Status GinFunctions::tokenize_prepare(FunctionContext* context, FunctionContext::FunctionStateScope scope) {
+    if (scope != FunctionContext::THREAD_LOCAL) {
+        return Status::OK();
+    }
+
+    auto column = context->get_constant_column(0);
+    auto method = ColumnHelper::get_const_value<TYPE_VARCHAR>(column);
+
+    lucene::analysis::Analyzer* analyzer;
+
+    if (method == "english") {
+        analyzer = _CLNEW lucene::analysis::SimpleAnalyzer();
+    } else if (method == "standard") {
+        analyzer = _CLNEW lucene::analysis::standard::StandardAnalyzer();
+    } else if (method == "chinese") {
+        auto* canalyzer = _CLNEW lucene::analysis::LanguageBasedAnalyzer();
+        canalyzer->setLanguage(L"cjk");
+        canalyzer->setStem(false);
+        analyzer = canalyzer;
+    } else {
+        return Status::NotSupported("Unknown method '" + method.to_string() +
+                                    "'. Supported methods are: 'english', 'standard', 'chinese'.");
+    }
+
+    context->set_function_state(scope, analyzer);
+
+    return Status::OK();
+}
+
+Status GinFunctions::tokenize_close(FunctionContext* context, FunctionContext::FunctionStateScope scope) {
+    if (scope == FunctionContext::THREAD_LOCAL) {
+        auto* analyzer = reinterpret_cast<lucene::analysis::Analyzer*>(
+                context->get_function_state(FunctionContext::THREAD_LOCAL));
+        delete analyzer;
+    }
+    return Status::OK();
+}
+
+StatusOr<ColumnPtr> GinFunctions::tokenize(FunctionContext* context, const starrocks::Columns& columns) {
+    auto* analyzer =
+            reinterpret_cast<lucene::analysis::Analyzer*>(context->get_function_state(FunctionContext::THREAD_LOCAL));
+
+    if (columns.size() != 2) {
+        return Status::InvalidArgument("Tokenize function only call by tokenize('<index_type>', str_column)");
+    }
+
+    ColumnViewer<TYPE_VARCHAR> value_viewer(columns[1]);
+    size_t num_rows = value_viewer.size();
+
+    // Array Offset
+    int offset = 0;
+    UInt32Column::Ptr array_offsets = UInt32Column::create();
+    array_offsets->reserve(num_rows + 1);
+
+    // Array Binary
+    BinaryColumn::Ptr array_binary_column = BinaryColumn::create();
+
+    NullColumnPtr null_array = NullColumn::create();
+
+    for (int row = 0; row < num_rows; ++row) {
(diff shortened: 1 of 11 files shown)


be/src/exprs/gin_functions.h
```diff
@@ -0,0 +1,33 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <CLucene.h>
```
alvin-celerdata, line 17: This header is not used in this file; please move it into the CPP files. We should keep the header files included as few as possible for better compilation speed and reduced exposure of third-party code.

gensrc/script/functions.py
```diff
@@ -1398,4 +1398,7 @@
     # data synopse functions
     [181000, 'bar', True, False, 'VARCHAR', ['BIGINT', "BIGINT", 'BIGINT', 'BIGINT'], 'UtilityFunctions::bar'],
     [181001, 'equiwidth_bucket', True, False, 'BIGINT', ['BIGINT', 'BIGINT', 'BIGINT', 'BIGINT'], 'UtilityFunctions::equiwidth_bucket'],
+
+    # gin functions
+    [190000, 'tokenize', True, False, 'ARRAY_VARCHAR', ['VARCHAR', 'VARCHAR'], 'GinFunctions::tokenize', 'GinFunctions::tokenize_prepare', 'GinFunctions::tokenize_close']
```
alvin-celerdata, line 1403: could write an English document to describe the function behavior?

alvin-phoenix-ai (commented):
Please change the title from "Enhancement" to "Feature", because this PR will introduce a new function.

gensrc/script/functions.py
INNOCENT-BOY: I have write english google doc and put it in the "What I am doing" section.

be/src/exprs/gin_functions.h
INNOCENT-BOY: remove it.

gensrc/script/functions.py
alvin-celerdata: What I mean is to write a doc in the directory of "/docs/en/sql-reference/sql-functions". Once the feature is released, users can learn about these functions on the StarRocks doc website. Otherwise, no one will know StarRocks has this function.

test/sql/test_inverted_index/T/test_tokenize
```diff
@@ -0,0 +1,28 @@
...
+"replicated_storage" = "false",
+"compression" = "LZ4"
+);
+
+INSERT INTO t_tokenized_table VALUES
+(1, 'hello', 'Привет', '你好'),
+(2, 'hello world', 'Привет, мир', '你好世界'),
+(3, 'Shanghai tap water comes from the sea', 'Водопроводная вода в Шанхае поступает из моря', '上海自来水来自海上');
+
+select id, tokenize('english', english_text), tokenize('standard', standard_text), tokenize('chinese', chinese_text) from t_tokenized_table order by id;
```
alvin-celerdata, line 26: More tests should be added to guarantee the correct behavior, such as the parameter of "Null", empty strings.

INNOCENT-BOY: I have added more tests.

gensrc/script/functions.py
INNOCENT-BOY: Have write a tokenize.md in the sql-functions.

test/sql/test_inverted_index/T/test_tokenize
INNOCENT-BOY: @alvin-celerdata Please help review again.

Seaven (approved):

kangkaisen (approved):

be/src/exprs/gin_functions.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (scope != FunctionContext::THREAD_LOCAL) {
+        return Status::OK();
+    }
+
+    auto column = context->get_constant_column(0);
+    auto method = ColumnHelper::get_const_value<TYPE_VARCHAR>(column);
+
+    lucene::analysis::Analyzer* analyzer;
+
+    if (method == "english") {
```
apatrida, line 39: This is not an English tokenizer at all, completely wrong tokenization for "english"  

This will break English words in bad ways, doesn't stem, doesn't lem, doesn't deplural, doesn't handle possessives, doesn't handle conjunctions, etc.

completely wrong.

be/src/exprs/gin_functions.cpp
```diff
@@ -0,0 +1,117 @@
...
+        analyzer = _CLNEW lucene::analysis::SimpleAnalyzer();
+    } else if (method == "standard") {
+        analyzer = _CLNEW lucene::analysis::standard::StandardAnalyzer();
```
apatrida, line 42: This uses "english" stop words by default and is not documented that stop word removal happens, but only for English.  It should also be documented that this uses the Unicode standard for word separations which might alter some test such as possessives.

This is accidentally more English than the English analyzer above.

PR #59009, opened 2025-05-15
[Enhancement] Report tablet version in advance if publish cost too much time

sevev:
## Why I'm doing:
There is currently a bad case during majority quorum publish in SR, assuming the following scenario:

1. Table A contains 3 tablets (tablet_a, tablet_b, tablet_c), each with 3 replicas distributed across 3 Backend Nodes (BEs):
2. BE1: tablet_a1, tablet_b1, tablet_c1
BE2: tablet_a2, tablet_b2, tablet_c2
BE3: tablet_a3, tablet_b3, tablet_c3

​​Publish Failure Analysis:​​
1. tablet_a1's publish hangs on BE1 → BE1 fails to response publish completion to FE
2. tablet_b2's publish hangs on BE2 → BE2 fails to response publish completion to FE

The publish cannot achieve majority consensus (2/3 acknowledgments required) because only BE3 successfully reports tablets status
The system must wait for all BEs to proactively report tablet statuses before publish completion and the publish will cost too much time.

This issue mostly occurs in ​​primary key tables​​ because their publish operations are not lightweight enough and may take prohibitively long to complete.

## What I'm doing:
If some tablets have completed publishing, the (BE) will proactively report their status updates to the FE.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/storage/publish_version_manager.cpp
+++ b/be/src/storage/publish_version_manager.cpp
@@ -201,7 +201,6 @@ void PublishVersionManager::finish_publish_version_task() {
                                 update_tablet_version(request.tablet_versions);
                                 TNetworkAddress master_addr = get_master_address();
                                 TUpdateTabletVersionResult result;
-#ifndef BE_TEST
                                 auto st = ThriftRpcHelper::rpc<FrontendServiceClient>(
                                         master_addr.hostname, master_addr.port,
                                         [&request, &result](FrontendServiceConnection& client) {
@@ -210,7 +209,6 @@ void PublishVersionManager::finish_publish_version_task() {
                                 if (!st.ok()) {
                                     LOG(WARNING) << "updateTabletVersion failed: " << st;
                                 }
-#endif
                             });
                     if (!st.ok()) {
                         LOG(WARNING) << "submit report tablet version task failed";
--- a/be/src/storage/publish_version_manager.cpp
+++ b/be/src/storage/publish_version_manager.cpp
@@ -207,7 +207,8 @@ void PublishVersionManager::finish_publish_version_task() {
                                             client->updateTabletVersion(result, request);
                                         });
                                 if (!st.ok()) {
-                                    LOG(WARNING) << "updateTabletVersion failed: " << st;
+                                    LOG(WARNING) << "updateTabletVersion failed: " << st
+                                                 << ", signature: " << request.signature;
                                 }
                             });
                     if (!st.ok()) {
--- a/be/src/storage/publish_version_manager.cpp
+++ b/be/src/storage/publish_version_manager.cpp
@@ -15,12 +15,15 @@
 #include "publish_version_manager.h"
 
 #include "agent/finish_task.h"
+#include "agent/master_info.h"
 #include "agent/task_signatures_manager.h"
 #include "common/config.h"
+#include "runtime/client_cache.h"
 #include "storage/storage_engine.h"
 #include "storage/tablet.h"
 #include "storage/tablet_manager.h"
 #include "util/cpu_info.h"
+#include "util/thrift_rpc_helper.h"
 
 namespace starrocks {
 const int MIN_FINISH_PUBLISH_WORKER_COUNT = 8;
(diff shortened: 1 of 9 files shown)


gensrc/thrift/FrontendService.thrift
```diff
@@ -2121,6 +2121,16 @@ struct TUpdateFailPointResponse {
     1: optional Status.TStatus status;
 }
 
+struct TUpdateTabletVersionRequest {
```
banmoy, line 2124: Can we TReportRequest to report the tablet version so that reuse most codes on both BE and FE? A difference is that FE should not change tablet meta (add/delete) with this type of report

sevev: > Can we use TReportRequest to report the tablet version so that reuse most codes on both BE and FE? A difference is that FE should not change tablet meta (add/delete) with this type of report

I previously considered this approach but decided against it for two reasons:

1. ​​Inconsistency with Periodic Report Logic​​
 This behavior differs from the periodic reporting mechanism. Incorporating it into the report logic didn’t align well with the existing reporting flow.
2. ​​Limited Code Reusability​​
Placing it in the report logic didn’t offer significant code reuse. The only reusable part was the leader-checking logic, which wasn’t enough to justify the integration.

be/src/storage/publish_version_manager.cpp
```diff
@@ -173,6 +180,40 @@ void PublishVersionManager::finish_publish_version_task() {
...
+                                update_tablet_version(request.tablet_versions);
+                                TNetworkAddress master_addr = get_master_address();
+                                TUpdateTabletVersionResult result;
+                                auto st = ThriftRpcHelper::rpc<FrontendServiceClient>(
+                                        master_addr.hostname, master_addr.port,
+                                        [&request, &result](FrontendServiceConnection& client) {
+                                            client->updateTabletVersion(result, request);
+                                        });
+                                if (!st.ok()) {
+                                    LOG(WARNING) << "updateTabletVersion failed: " << st;
```
wyb, line 210: better to add `signature` in the log

fe/fe-core/src/main/java/com/starrocks/leader/LeaderImpl.java
```diff
@@ -1438,4 +1441,95 @@ public TTableReplicationResponse startTableReplication(TTableReplicationRequest
...
+
+    public TUpdateTabletVersionResult updateTabletVersion(TUpdateTabletVersionRequest request) {
+        TUpdateTabletVersionResult result = new TUpdateTabletVersionResult();
+        TStatus tStatus = new TStatus(TStatusCode.OK);
+        result.setStatus(tStatus);
+        if (!GlobalStateMgr.getCurrentState().isLeader()) {
+            LOG.warn("current node is not leader, update tablet version failed, signature: {}",
+                    request.getSignature());
+            tStatus.setStatus_code(TStatusCode.CANCELLED);
+            tStatus.setError_msgs(Lists.newArrayList("current fe is not master"));
```
wyb, line 1453:
```suggestion
            tStatus.setError_msgs(Lists.newArrayList("current fe is not leader"));
```

fe/fe-core/src/main/java/com/starrocks/leader/LeaderImpl.java
```diff
@@ -1438,4 +1441,95 @@ public TTableReplicationResponse startTableReplication(TTableReplicationRequest
...
+            tStatus.setStatus_code(TStatusCode.CANCELLED);
+            tStatus.setError_msgs(Lists.newArrayList("no replicas on backend"));
+            result.setStatus(tStatus);
+            return result;
+        }
+
+        List<TabletMeta> tabletMetaList = tablets.getTabletMetaList(tabletIds);
+        Long dbId = null;
+        Long tableId = null;
+        if (tabletMetaList == null || tabletMetaList.isEmpty()) {
```
wyb, line 1493: tabletMetaList is always not null.

fe/fe-core/src/main/java/com/starrocks/leader/LeaderImpl.java
```diff
@@ -1438,4 +1441,95 @@ public TTableReplicationResponse startTableReplication(TTableReplicationRequest
...
+            tStatus.setStatus_code(TStatusCode.CANCELLED);
+            tStatus.setError_msgs(Lists.newArrayList("no tabletMeta found"));
+            result.setStatus(tStatus);
+            return result;
+        }
+        for (TabletMeta tabletMeta : tabletMetaList) {
+            if (tabletMeta == null) {
```
wyb, line 1500:
```suggestion
            if (tabletMeta == NOT_EXIST_TABLET_META) {
```

wyb (approved):

HangyuanLiu (approved):

dirtysalt (approved):

PR #59052, opened 2025-05-17
[BugFix] Avoid polluting task run properties in multi task runs

LiShuMing:
## Why I'm doing:
it's weird to find that TaskRun's properties can be polluted sometimes.

```
2025-05-15 16:48:43.837-04:00 INFO (starrocks-taskrun-pool-1839|26764776) [TaskRunManager.submitTaskRun():56] submit task run:TaskRun{taskId=982129, type=PERIODICAL, uuid=fd6dbef0-31cd-11f0-a9ef-0affeef69e79, task_state=, properties={PARTITION_START=2025-03-18, query_mem_limit32212254720, FORCE=true, query_timeout=240000, mvId=982127, replication_num=1, enable_spill=false, PARTITION_END=2025-04-16, in_memory=false, warehouse_id=0, START_TASK_RUN_ID=94925bde-1b18-11f0-a9ef-0affeef69e79}, extra_message =}
```

For MV's task, `force` property should be only set in the command  and should not be for many task runs :
```
refresh materialized view xxx force;
```
which means `force` property maybe polluted by some codes.

In `Task`'s properties,  
- I found there's no way to change it since we always use `copy` rather changing it directly.

In `TaskRun`'s properties, There are some codes that are not safe:
1. `generateNextTaskRun` will always copy previous `force` properties:  If the base table's partitions always changed, it may always refresh(force property will refresh all the partitions)
```
    private void generateNextTaskRun() {
        TaskManager taskManager = GlobalStateMgr.getCurrentState().getTaskManager();
        Map<String, String> properties = mvContext.getProperties();
        long mvId = Long.parseLong(properties.get(MV_ID));
        String taskName = TaskBuilder.getMvTaskName(mvId);
        Map<String, String> newProperties = Maps.newHashMap();
        for (Map.Entry<String, String> proEntry : properties.entrySet()) {
            if (proEntry.getValue() != null) {
                newProperties.put(proEntry.getKey(), proEntry.getValue());
            }
        }

}
```

2. Where there are many pending task runs to merge, it's not safe to copy properties directly. 

```
    public void mergeProperties(ExecuteOption option) {
        if (option.taskRunProperties != null) {
            if (taskRunProperties == null) {
                taskRunProperties = Maps.newHashMap();
            }
            taskRunProperties.putAll(option.taskRunProperties);
        }
    }
```
## What I'm doing:
To avoid this:
1. When mv generates the next task run, skip to copy some uncopyable properties.
2. When merge pending task runs, skip to merge task runs if theirs properties are not the same.

Fixes 

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/ExecuteOption.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/ExecuteOption.java
@@ -19,8 +19,10 @@ import com.google.common.collect.Maps;
 import com.google.gson.annotations.SerializedName;
 import com.starrocks.common.Config;
 import com.starrocks.persist.gson.GsonUtils;
+import org.apache.commons.collections4.CollectionUtils;
 
 import java.util.Map;
+import java.util.Objects;
 
 public class ExecuteOption {
 
@@ -106,34 +108,37 @@ public class ExecuteOption {
         isReplay = replay;
     }
 
-    private boolean containsKey(String key) {
-        return taskRunProperties.containsKey(key) && taskRunProperties.get(key) != null;
+    /**
+     * Check if the current ExecuteOption can be merged with the given option:
+     * - Only all taskRunProperties are empty or the same with specific keys, we can merge two pending task runs.
+     */
+    public boolean isMergeableWith(ExecuteOption other) {
+        if (!isMergeRedundant || !other.isMergeRedundant) {
+            return false;
+        }
+        if (CollectionUtils.sizeIsEmpty(this.taskRunProperties)
+                && CollectionUtils.sizeIsEmpty(other.taskRunProperties)) {
+            return true;
+        }
+        return Objects.equals(this.getTaskRunComparableProperties(), other.getTaskRunComparableProperties());
     }
 
     /**
-     * If the execute option contains the properties that need to be merged into the task run, eg: it's an internal partition
-     * refresh, needs to merge it into the newer task run.
-     * task in mv refresh
-     * @return
+     * See @{code TaskRun#MV_COMPARABLE_PROPERTIES} for more details.
+     * @return a map of task run properties that are comparable for merging.
      */
-    public boolean containsToMergeProperties() {
-        if (taskRunProperties == null) {
-            return false;
+    public Map<String, String> getTaskRunComparableProperties() {
+        // For non-mv task, we don't need to check the task run properties.
+        if (taskRunProperties == null || !taskRunProperties.containsKey(TaskRun.MV_ID)) {
+            return taskRunProperties;
         }
-        if (containsKey(TaskRun.PARTITION_START) || containsKey(TaskRun.PARTITION_END)
-                || containsKey(TaskRun.START_TASK_RUN_ID) || containsKey(TaskRun.PARTITION_VALUES)) {
-            return true;
-        }
-        return false;
-    }
-
-    public void mergeProperties(ExecuteOption option) {
-        if (option.taskRunProperties != null) {
-            if (taskRunProperties == null) {
-                taskRunProperties = Maps.newHashMap();
+        Map<String, String> result = Maps.newHashMap();
+        for (Map.Entry<String, String> e : taskRunProperties.entrySet()) {
+            if (TaskRun.MV_COMPARABLE_PROPERTIES.contains(e.getKey())) {
+                result.put(e.getKey(), e.getValue());
             }
-            taskRunProperties.putAll(option.taskRunProperties);
         }
+        return result;
     }
 
     @Override
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/TaskRun.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/TaskRun.java
@@ -56,6 +56,8 @@ public class TaskRun implements Comparable<TaskRun> {
 
     private static final Logger LOG = LogManager.getLogger(TaskRun.class);
 
+    // NOTE: ALL task run properties should be defined here, and the associated properties configuration below should be
+    // added carefully.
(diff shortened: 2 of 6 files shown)


alvin-celerdata:
> What I'm doing:
To avoid this:

> When mv generates the next task run, skip to copy some uncopyable properties.
When merge pending task runs, skip to merge task runs if theirs properties are not the same.
Fixes https://github.com/StarRocks/StarRocksTest/issues/9658

It is better to explain why this happens in the original code, so that everyone can understand that the change you made in this PR addresses the issue. 
Otherwise, I have no idea why the change can resolve the issue.

fe/fe-core/src/main/java/com/starrocks/scheduler/ExecuteOption.java
```diff
@@ -106,34 +108,32 @@ public void setReplay(boolean replay) {
...
-     * If the execute option contains the properties that need to be merged into the task run, eg: it's an internal partition
-     * refresh, needs to merge it into the newer task run.
-     * task in mv refresh
-     * @return
+     * * Check if the current ExecuteOption can be merged with the given option.
+     * Only all taskRunProperties are empty or the same with specific keys, we can merge two pending task runs.
      */
-    public boolean containsToMergeProperties() {
-        if (taskRunProperties == null) {
+    public boolean isMergeableWith(ExecuteOption option) {
```
alvin-celerdata, line 115:
```suggestion
    public boolean isMergeableWith(ExecuteOption other) {
```
`other` will be a better name

fe/fe-core/src/main/java/com/starrocks/scheduler/ExecuteOption.java
```diff
@@ -106,34 +108,32 @@ public void setReplay(boolean replay) {
...
-        return false;
+        return Objects.equals(this.getTaskRunComparableProperties(), option.getTaskRunComparableProperties());
     }
 
-    public void mergeProperties(ExecuteOption option) {
-        if (option.taskRunProperties != null) {
-            if (taskRunProperties == null) {
-                taskRunProperties = Maps.newHashMap();
+    /**
+     * @return a map of task run properties that are comparable for merging.
```
alvin-celerdata, line 127: What does comparable mean?
I think you should write more comments here, perhaps providing some examples of what properties are comparable and what properties are not, and why.

fe/fe-core/src/main/java/com/starrocks/scheduler/TaskRun.java
```diff
@@ -62,15 +62,21 @@ public class TaskRun implements Comparable<TaskRun> {
     // list partition values to be refreshed
     public static final String PARTITION_VALUES = "PARTITION_VALUES";
     public static final String FORCE = "FORCE";
+    // properties that are not allowed to be copied from one task run to another
+    public static final Set<String> UNCOPYABLE_PROPERTIES = ImmutableSet.of(
+            PARTITION_START, PARTITION_END, PARTITION_VALUES, FORCE);
```
alvin-celerdata, line 67: Could you provide a more detailed explanation on why these are not copiable?

fe/fe-core/src/main/java/com/starrocks/scheduler/TaskRun.java
```diff
@@ -62,15 +62,21 @@ public class TaskRun implements Comparable<TaskRun> {
...
+    // properties that are used to check equality of task runs
+    public static final Set<String> COMPARABLE_PROPERTIES = ImmutableSet.of(
```
alvin-celerdata, line 69: same here

fe/fe-core/src/main/java/com/starrocks/scheduler/TaskRun.java
```diff
@@ -62,15 +62,21 @@ public class TaskRun implements Comparable<TaskRun> {
...
+            MV_ID, PARTITION_START, PARTITION_END, PARTITION_VALUES, FORCE);
     public static final String START_TASK_RUN_ID = "START_TASK_RUN_ID";
+    // Only used in FE's UT
+    public static final String IS_TEST = "__IS_TEST__";
     // All properties that can be set in TaskRun
     public static final Set<String> TASK_RUN_PROPERTIES = ImmutableSet.of(
-            MV_ID, PARTITION_START, PARTITION_END, FORCE, START_TASK_RUN_ID, PARTITION_VALUES, PROPERTIES_WAREHOUSE);
+            MV_ID, PARTITION_START, PARTITION_END, FORCE, START_TASK_RUN_ID, PARTITION_VALUES, PROPERTIES_WAREHOUSE, IS_TEST);
```
alvin-celerdata, line 76: And if someone adds a new property in the future, how can we ensure that the property is classified as uncopyable or comparable?

LiShuMing:
> > What I'm doing:
> > To avoid this:
> 
> > When mv generates the next task run, skip to copy some uncopyable properties.
> > When merge pending task runs, skip to merge task runs if theirs properties are not the same.
> > Fixes [StarRocks/StarRocksTest#9658](https://github.com/StarRocks/StarRocksTest/issues/9658)
> 
> It is better to explain why this happens in the original code, so that everyone can understand that the change you made in this PR addresses the issue. Otherwise, I have no idea why the change can resolve the issue.

updated, PTAL.

satanson (approved):

kangkaisen (approved):

PR #59056, opened 2025-05-18
[BugFix] Fix miss update job state when replaying restore log may cause restored table lost after FE restart

srlch:
## Why I'm doing:
If we set history_job_keep_max_second < checkpoint time interval (no matter by background daemon
or manlly create image). When we replay the log for the restore job, the final FINISHED state log
will be skipped to update the state job state in BackupHandler because this job is considered
as a expired job controlled by history_job_keep_max_second.

But the point here is that, Only FINISHED state is skipped to update but not other state log, the BackupHandler
will persistent the unfinished state for this job even the job is actually finished now.

When we restart FE in this case, we will get a unfinished restore job in FE and if the restart time is far away
from the restore job start time, the job will be cancelled uncorrectly(because FE treat it as a unfinished job)
and remove all table restored successfully before FE restart.

## What I'm doing:
Skip to check expire or not in replay.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/backup/BackupHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/backup/BackupHandler.java
@@ -812,10 +812,6 @@ public class BackupHandler extends FrontendDaemon implements Writable, MemoryTra
             // for example: In restore job, PENDING will transfer to SNAPSHOTING, not DOWNLOAD.
             job.replayRun();
         }
-        if (isJobExpired(job, System.currentTimeMillis())) {
-            LOG.warn("skip expired job {}", job);
-            return;
-        }
         dbIdToBackupOrRestoreJob.put(job.getDbId(), job);
         mvRestoreContext.addIntoMvBaseTableBackupInfo(job);
     }
--- a/fe/fe-core/src/test/java/com/starrocks/backup/RestoreJobTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/backup/RestoreJobTest.java
@@ -1008,4 +1008,31 @@ public class RestoreJobTest {
         job.run();
         job.run();
     }
+
+    @Test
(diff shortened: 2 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/backup/BackupHandler.java
```diff
@@ -813,6 +813,12 @@ public void replayAddJob(AbstractJob job) {
...
         if (isJobExpired(job, System.currentTimeMillis())) {
+            AbstractJob existingJob = dbIdToBackupOrRestoreJob.get(job.getDbId());
+            // remove the existed job info if the job is expired
+            if (existingJob != null) {
+                dbIdToBackupOrRestoreJob.remove(job.getDbId());
+                mvRestoreContext.discardExpiredBackupTableInfo(job);
+            }
             LOG.warn("skip expired job {}", job);
             return;
         }
```
xiangguangyxg, line 824: Is it better not to check whether the job is expired during replay?

srlch: yes, I think it is a better way actually. I am still checking if there is other problem if do not skip the check in replay.

srlch: Updated

gengjun-git (approved):

xiangguangyxg (approved):

wyb (approved):

PR #59102, opened 2025-05-19
[Feature] Support enabling partition aggregation for tables created in older versions

sevev:
## Why I'm doing:
In previous PRs, we implemented table-level `enable_partition_aggregation` property to reduce API calls in StarRocks shared-data mode. However, this property currently can only be specified during table creation. As a result, even after users upgrade to the latest version, they cannot enable this feature for existing tables unless they create new ones.

Therefore, we need to develop a method to activate this functionality for historical tables.

## What I'm doing:
Users can enable/disable this feature using the following SQL command when they upgrade to new version:
```sql
Alter table $table_name set ('enable_partition_property' = 'true');
Alter table $table_name set ('enable_partition_property' = 'false');
```

But there is one critical limitation:**consecutive modifications to enable_partition_aggregation are prohibited.**

e.g.
1. If you successfully execute:  
   ```sql
   ALTER TABLE test_tbl SET ('enable_partition_aggregation' = 'true');
2. You cannot immediately follow it with:
    ```sql
    ALTER TABLE test_tbl SET ('enable_partition_aggregation' = 'false');
    ```
   Otherwise, you may encounter errors like:
   ```
    Property enable_partition_aggregation cannot be updated now because this table contains mixed metadata types (both split and aggregate). Please wait until old metadata versions are vacuumed);
    ```

**Why:**
1.  `enable_partition_aggregation` is designed as a non-frequently toggled property. Modification support is primarily intended to allow enabling the feature on tables created in old version.
2. Modify `enable_partition_aggregation` will ​​modify the table's metadata storage format​​. So the system must support ​​coexistence of both metadata formats​​ during the transition period. After modify `enable_partition_aggregation`, the table metadata will keep a `metadataSwitchVersion` field to track the version of the format change. Frequent modifications will increased complexity in compatibility logic, So the modification is prohibited​​ when `metadataSwitchVersion` is not 0.

**When we can do the next modification of `enable_partition_aggregation property`?**

If you want to modify `enable_partition_aggregation`, you can check the `metadataSwitchVersion` first. You can run  `show '/proc/dbs/$db/$table/partitions' ` to check the `metadataSwitchVersion`.

**How can we efficiently reset the metadataSwitchVersion to 0 after completing the metadata format transition?**

​​`metadataSwitchVersion` is reset to 0 when the version before `metadataSwitchVersion` is vacuumed. So we can decrease `lake_autovacuum_grace_period_minutes` to speed up vacuum.

Fixes https://github.com/StarRocks/starrocks/issues/58316

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJobBase.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJobBase.java
@@ -62,6 +62,14 @@ import javax.validation.constraints.NotNull;
 
 public abstract class LakeTableAlterMetaJobBase extends AlterJobV2 {
     private static final Logger LOG = LogManager.getLogger(LakeTableAlterMetaJobBase.class);
+
+    public enum MetadataOp {
+        NO_OPERATION, // do nothing
+        AGGREGATE, // update enable_partition_aggregation from false to true
+        SPLIT // update enable_partition_aggregation from true to false
+    }
+
+
     @SerializedName(value = "watershedTxnId")
     private long watershedTxnId = -1;
     @SerializedName(value = "watershedGtid")
@@ -71,10 +79,8 @@ public abstract class LakeTableAlterMetaJobBase extends AlterJobV2 {
     private Table<Long, Long, MaterializedIndex> physicalPartitionIndexMap = HashBasedTable.create();
     @SerializedName(value = "commitVersionMap")
     private Map<Long, Long> commitVersionMap = new HashMap<>();
-    @SerializedName(value = "aggregateTabletMetadata")
-    private boolean aggregateTabletMetadata = false;
-    @SerializedName(value = "splitTabletMetadata")
-    private boolean splitTabletMetadata = false;
+    @SerializedName(value = "metadataOp")
+    private MetadataOp metadataOp;
     private AgentBatchTask batchTask = null;
     private boolean enablePartitionAggregation = false;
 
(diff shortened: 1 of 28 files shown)


gensrc/proto/lake_types.proto
```diff
@@ -162,6 +162,8 @@ message MetadataUpdateInfoPB {
     optional bool enable_persistent_index = 1;
     optional TabletSchemaPB tablet_schema = 2;
     optional PersistentIndexTypePB persistent_index_type = 3;
+    optional bool aggregate_tablet_metadata = 4;
+    optional bool split_tablet_metadata = 5;
```
luohaha, line 166: Is `split_tablet_metadata` useless?

sevev: > Is `split_tablet_metadata` useless?

removed.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJobBase.java
```diff
@@ -264,20 +270,20 @@ protected boolean lakePublishVersion() {
...
-            List<Tablet> tablets = new ArrayList<>();
+            boolean isEnablePartitionAggregation = isEnablePartitionAggregation();
             for (long partitionId : physicalPartitionIndexMap.rowKeySet()) {
                 long commitVersion = commitVersionMap.get(partitionId);
-                tablets.clear();
                 Map<Long, MaterializedIndex> dirtyIndexMap = physicalPartitionIndexMap.row(partitionId);
+                List<Tablet> tablets = new ArrayList<>();
                 for (MaterializedIndex index : dirtyIndexMap.values()) {
-                    if (!enablePartitionAggregation) {
+                    if (!enablePartitionAggregation && !isEnablePartitionAggregation) {
```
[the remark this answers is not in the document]
sevev: updated.

kevincai:
@Entelligence.AI

kevincai (approved):

wyb (approved):

gengjun-git (approved):

Seaven (approved):

PR #59104, opened 2025-05-19
[BugFix] persistent index rebuild when migrate between different cpu arch

### LINKED ISSUE #57952: Crash of starrocks BE on ARM
Waiting about 10 minutes, no query required to reproduce the crash

It does not crash

### Real behavior (Required)
It crashes

3.3.9, 3.4.1

### Description of the bug

Tested on both versions 3.3.9 and 3.4.1, in a setup with a single FE and single BE, nothing shared mode, about 160GB of database data, zstd compressed.

This setup is running on x86 architecture with no particular issue. Now, moving to an ARM machine, I observe the BE is regularly crashing with this stack trace :

```
W20250406 17:07:33.542100 281456209117312 stack_util.cpp:347] 2025-04-06 17:07:33.227865, query_id=00000000-0000-0000-0000-000000000000, fragment_instance_id=00000000-0000-0000-
0000-000000000000 throws exception: std::bad_alloc, trace:
     @          0x6549740  __wrap___cxa_throw
    @          0xd7d7298  operator new(unsigned long)
    @          0x6d5ae2c  starrocks::FixedMutableIndex<20ul>::load_snapshot(phmap::BinaryInputArchive&)
    @          0x6d17698  starrocks::ShardByLengthMutableIndex::load_snapshot(phmap::BinaryInputArchive&, std::set<unsigned int, std::less<unsigned int>, std::allocator<unsigned
 int> > const&)
    @          0x6d2f948  starrocks::ShardByLengthMutableIndex::load(starrocks::MutableIndexMetaPB const&)
    @          0x6d3a4c4  starrocks::PersistentIndex::_load(starrocks::PersistentIndexMetaPB const&, bool)
    @          0x6d3b06c  starrocks::PersistentIndex::load(starrocks::PersistentIndexMetaPB const&)
    @          0x6d42acc  starrocks::PersistentIndex::_load_by_loader(starrocks::TabletLoader*)
    @          0x6d441f0  starrocks::PersistentIndex::load_from_tablet(starrocks::Tablet*)
    @          0x6867f2c  starrocks::PrimaryIndex::_do_load(starrocks::Tablet*)
    @          0x6868bd8  starrocks::PrimaryIndex::load(starrocks::Tablet*)
    @          0x69373a8  starrocks::UpdateManager::on_rowset_finished(starrocks::Tablet*, starrocks::Rowset*)
    @          0x71ca5b8  starrocks::DeltaWriter::commit()
    @          0x8136388  starrocks::AsyncDeltaWriter::_execute(void*, bthread::TaskIterator<starrocks::AsyncDeltaWriter::Task>&)
    @          0x989e6ac  bthread::ExecutionQueueBase::_execute(bthread::TaskNode*, bool, int*)
    @          0x989f638  bthread::ExecutionQueueBase::_execute_tasks(void*)
    @          0x83236a4  starrocks::ThreadPool::dispatch_thread()
    @          0x831b69c  starrocks::Thread::supervise_thread(void*)
    @     0xffff9455d5c8  (/usr/lib/aarch64-linux-gnu/libc.so.6+0x7d5c7)
    @     0xffff945c5edc  (/usr/lib/aarch64-linux-gnu/libc.so.6+0xe5edb)
```
Before it crashes, I can access all the data, and there is no visible corruption. But after 10 minutes about it crashes like this.

One thing to note, the data volume of the BE was copied from the X86 machine to the ARM machine as is. Are the data architecture specific ? Maybe I should rebuild from scratch the data instead ?

luohaha:
## Why I'm doing:
Currently, persistent index does not support migration between different CPU architectures. For example, a persistent index generated on an x86 architecture cannot be reloaded and used on an arm64 architecture. The reason is that we have not properly handled architecture-specific inconsistencies such as byte order and padding. Therefore, to ensure reliability, when CPU architecture migration occurs, it is recommended to rebuild the persistent index to guarantee the correctness of the index. This check can be enabled by configuring ​​`pindex_cpu_arch_migration_check`​​.

TODO. In the future, we could implement a cross-platform persistent index, but currently this is not a top priority.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1


sevev (approved):

be/src/common/config.h
```diff
@@ -1353,6 +1353,9 @@ CONF_mInt32(pindex_load_thread_pool_num_max, "8");
 CONF_mInt32(pindex_rebuild_clone_wait_seconds, "120");
 // Max wait time to rebuild persistent index in load (preload update state)
 CONF_mInt32(pindex_rebuild_load_wait_seconds, "20");
+// Check if there is migration between different CPU architectures.
+// If true, the persistent index will be rebuilt when the CPU architecture is different.
+CONF_mBool(pindex_cpu_arch_migration_check, "false");
```
chaoyli, line 1358: Why not set this to be true?

be/src/storage/persistent_index.cpp
```diff
@@ -3434,6 +3434,12 @@ bool PersistentIndex::_need_rebuild_index(const PersistentIndexMetaPB& index_met
         // Make sure l2 version equal to l2 version merged flag
         return true;
     }
+    if (index_meta.cpu_arch_id() != CPU_ARCH_ID && config::pindex_cpu_arch_migration_check) {
+        // If we found the migration between different CPU arch, we need to rebuild index.
```
dengliu, line 3438: in this case, shall we delete the existing index file first to stop it from being used for updating/generate new tablet files?

be/src/storage/persistent_index.h
```diff
@@ -34,6 +34,20 @@ class Schema;
...
+static constexpr uint32_t CPU_ARCH_ID = 0x00000002; // x86_64 platform
+#elif defined(__powerpc64__)
+static constexpr uint32_t CPU_ARCH_ID = 0x00000003; // PowerPC64 platform
+#elif defined(__arm__)
+static constexpr uint32_t CPU_ARCH_ID = 0x00000004; // ARM platform
+#elif defined(__aarch64__)
+static constexpr uint32_t CPU_ARCH_ID = 0x00000005; // AArch64 platform
+#else
+static constexpr uint32_t CPU_ARCH_ID = 0x0000000F; // Unknown platform
+#endif
```
alvin-celerdata, line 49: If the binary is generated on an i386 platform but can execute on an x86_64 platform, what is the behavior? Is it expected?

be/src/storage/persistent_index.cpp
```diff
@@ -3434,6 +3434,12 @@ bool PersistentIndex::_need_rebuild_index(const PersistentIndexMetaPB& index_met
         // Make sure l2 version equal to l2 version merged flag
         return true;
     }
+    if (index_meta.cpu_arch_id() != CPU_ARCH_ID && config::pindex_cpu_arch_migration_check) {
```
alvin-celerdata, line 3437: Check if cpu_arch_id is set, and then verify if this is compatible. There is no need for an extra configuration.

gensrc/proto/persistent_index.proto
```diff
@@ -84,4 +84,8 @@ message PersistentIndexMetaPB {
     repeated EditVersionPB l2_versions = 8;
     // record if this version is generated by bg merge compaction
     repeated bool l2_version_merged = 9;
+    // cpu arch platform.
+    // PersistentIndex can't support cross platform,
+    // need to identify platform change and rebuild index.
+    uint32 cpu_arch_id = 10;
```
alvin-celerdata, line 90:
```suggestion
    optional uint32 cpu_arch_id = 10;
```

alvin-celerdata: I'm worried that only the CPU ID can't guarantee compatibility.
Suppose that arm can be configured into different endianness. If a chip with little-endian writes an index, but another chip with big-endian reads it, will it succeed?

alvin-phoenix-ai (commented):
Do you know the root cause for this bug?
Can we have other solution rather than using CPU_ID?
Because both x86_64 and arm64 are little endian by default, what caused the issue?
And I'm worried this can't resolve all issues without knowing the root cause.

be/src/common/config.h
luohaha: Because it will cause unnecessary rebuild

be/src/storage/persistent_index.cpp
luohaha: No, it you set `pindex_cpu_arch_migration_check` to true, index will be rebuilt after migration.

dengliu:
@luohaha @alvin-celerdata @chaoyli 
Can you please update what's the plan here?

PR #59098, opened 2025-05-19
[Refactor] Use UUIDUtil to generate queryId/loadId/executionId

dolfinus:
## Why I'm doing:

Preparation for changing UUID generation algorithm in #58962.

## What I'm doing:

For code related to queryId/loadId/executionId generation:
* Replace `UUID.randomUUID()` calls with `UUIDUtil.genUUID()`.
* Replace `new TUniqueId(uuid.getMostSignificantBits(), uuid.getLeastSignificantBits())` with `UUIDUtil.genTUniqueId()`

There are other places there `UUID.randomUUID()` is used (e.g. in FileSystem temp paths, in AWS secret handling and so on), which I left intact.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [X] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [X] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [ ] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

PR #59107, opened 2025-05-19
[Enhancement] Use UUIDv7 for QueryId

### LINKED ISSUE #58962: [Enhancement] Use UUIDv7 for queryId
### Description

Currently queryId is generated using [UUIDv4](https://github.com/StarRocks/starrocks/blob/3.4.3/be/src/util/uuid_generator.h). Although it provides great uniqueness, it can be beneficial to use [UUIDv7](https://datatracker.ietf.org/doc/rfc9562/) instead:
* UUIDv7 and UUIDv4 have the same number of bits, so no changes in field size.
* UUIDv7 is generated using timestamp (milliseconds) + counter (optional) + random part, so values can be sorted, unlike UUIDv4. This is important for storing audit logs, as queries started one after another will be stored as table records next to each other.
* The latter also greatly improves size of B-Tree indexes. For example, UUIDv4 indexes can take x10 size of UUIDv7 index because v4 values are not correlated with row position, and index bloat is a thing.
* It is possible to extract timestamp (approximate) the query was created. Currently this information can be extracted only from [current_queries](https://docs.starrocks.io/docs/using_starrocks/running_queries/) view. lso the view doesn't return queries which already finished, so currently the only way to know it is to capture all audit logs, and search for a start event.

### Solution

Change the algorithm used by `ThreadLocalUUIDGenerator::next_uuid()` to generate values based on curent timestamp. It doesn't require adding any dependencies, as UUID generation code is quire simple, and can be vendored.

As an option, boost 1.86+ already has such generator: https://www.boost.org/doc/libs/master/libs/uuid/doc/html/uuid.html#time_generator_v7

dolfinus:
## Why I'm doing:

Closes #58962

## What I'm doing:

Update `UUIDUtil` Java class to return UUIDv7 values instead of UUIDv4. Using fasterxml's time epoch based generator.

C++ implementation still generates UUIDv4.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [X] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/UUIDUtil.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/UUIDUtil.java
@@ -15,22 +15,32 @@
 
 package com.starrocks.common.util;
 
-import com.fasterxml.uuid.EthernetAddress;
 import com.fasterxml.uuid.Generators;
-import com.fasterxml.uuid.impl.TimeBasedGenerator;
+import com.fasterxml.uuid.impl.TimeBasedEpochRandomGenerator;
 import com.starrocks.thrift.TUniqueId;
 
+import java.security.SecureRandom;
+import java.util.Random;
 import java.util.UUID;
 
 public class UUIDUtil {
-    private static final EthernetAddress ETHERNET_ADDRESS = EthernetAddress.fromInterface();
-    private static final TimeBasedGenerator UUID_GENERATOR = Generators.timeBasedGenerator(ETHERNET_ADDRESS);
+    // Prefer UUIDv7 values as they are ordered, and include timestamp.
+    // Using ThreadLocal to avoid lock contention in multithread environment.
+    // See https://github.com/StarRocks/starrocks/issues/58962 for more details.
+    private static final SecureRandom TRUE_RANDOM = new SecureRandom();
+    private static final ThreadLocal<TimeBasedEpochRandomGenerator> UUID7_FACTORY =
+            new ThreadLocal<>() {
+                @Override protected TimeBasedEpochRandomGenerator initialValue() {
+                    // SecureRandom return truly random values, but it is 4x slower than Random.
+                    // Random is pseudo-random, so it has to be initialized with a random seed
+                    // to avoid producing same results in a different thread
+                    long seed = TRUE_RANDOM.nextLong();
+                    return Generators.timeBasedEpochRandomGenerator(new Random(seed));
+                }
+            };
 
-    // java.util.UUID.randomUUID() uses SecureRandom to generate random uuid,
-    // and SecureRandom hold locks when generate random bytes.
-    // This method is faster than java.util.UUID.randomUUID() in high concurrency environment
     public static UUID genUUID() {
-        return UUID_GENERATOR.generate();
+        return UUID7_FACTORY.get().generate();
     }
 
     public static UUID fromTUniqueid(TUniqueId id) {
@@ -42,7 +52,7 @@ public class UUIDUtil {
     }
 
     public static TUniqueId genTUniqueId() {
-        return toTUniqueId(UUID_GENERATOR.generate());
+        return toTUniqueId(genUUID());
     }
 
 }
--- a/fe/fe-core/src/test/java/com/starrocks/common/util/UUIDUtilTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/common/util/UUIDUtilTest.java
@@ -17,8 +17,11 @@ package com.starrocks.common.util;
 import org.junit.Assert;
 import org.junit.Test;
 
+import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
@@ -28,6 +31,22 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 public class UUIDUtilTest {
+    /**
+     * Extract timestamp from a UUIDv7.
+     * UUIDv7 has the timestamp in the first 48 bits
+     */
+    private long extractTimestampFromUuid(UUID uuid) {
+        long msb = uuid.getMostSignificantBits();
+        return (msb >>> 16) & 0xFFFFFFFFFFFL; // 48 bits
+    }
+
+    /**
+     * Extract random component from UUIDv7 for comparison
+     */
+    private long extractRandomFromUuid(UUID uuid) {
+        return uuid.getLeastSignificantBits();
+    }
+
     @Test
     public void testUniqueness() {
         final int numUuids = 10000;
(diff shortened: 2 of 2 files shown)


alvin-celerdata:
@dolfinus 
Since this PR would modify the third-party dependencies, I suggest splitting it into three separate PRs.
The first PR is only to bump up the boost version.
The second PR is to update the fastxml version.
The third one is to change the UUID generation methods.
In this way, if the UUID generator causes unexpected bugs, we can revert the last PR.

dolfinus:
Ok

dolfinus:
Splitted dependency upgrades to PRs #59217 and #59218. Converting current PR back to draft until those are merged

dolfinus:
Well, apparently Boost cannot be upgraded to 1.86 now. C++ UUID generator implementation could be vendored though.

Before changing the behavior, let's add test for existing behavior of Java #59262 and C++ #59267 generators.

alvin-celerdata:
@dolfinus 
Could you change the UUID generator to UUIDv7 in Java alone?
This change can benefit most UUIDs for now.
So that we can make backend-generated UUIDs in a separate PR, and I can wait for boost 1.86 until we make the change.

dolfinus:
Ok, done

alvin-phoenix-ai:
@mergify rebase

alvin-phoenix-ai:
@mergify rebase

alvin-phoenix-ai (approved):

fe/fe-core/src/main/java/com/starrocks/common/util/UUIDUtil.java
```diff
@@ -15,20 +15,16 @@
...
+import com.fasterxml.uuid.impl.TimeBasedEpochRandomGenerator;
 import com.starrocks.thrift.TUniqueId;
 
 import java.util.UUID;
 
 public class UUIDUtil {
-    private static final EthernetAddress ETHERNET_ADDRESS = EthernetAddress.fromInterface();
-    private static final TimeBasedGenerator UUID_GENERATOR = Generators.timeBasedGenerator(ETHERNET_ADDRESS);
+    // Prefer UUIDv7 to UUIDv4, as UUIDv7 values are sortable, but UUIDv4 are not
+    private static final TimeBasedEpochRandomGenerator UUID_GENERATOR = Generators.timeBasedEpochRandomGenerator();
```
alvin-phoenix-ai: >    // java.util.UUID.randomUUID() uses SecureRandom to generate random uuid,
>    // and SecureRandom hold locks when generate random bytes.
>    // This method is faster than java.util.UUID.randomUUID() in high concurrency environment

If the random is not specified, it will use SecureRandom to generate UUID, which is slow according to the comment.
Did you do some benchmarking testing between two UUID generators?
I can't merge it before we have a benchmark to proof that this change won't impact the performance.

alvin-phoenix-ai (changes requested):
Sorry, I need to block this PR to be merged before we have a solid performance benchmark result.

fe/fe-core/src/main/java/com/starrocks/common/util/UUIDUtil.java
dolfinus: Are there some benchmarks which I can run locally?

alvin-phoenix-ai: Unfortunately, there probably are no existing benchmarks. 
But this class has no dependence, and it is easy to build a benchmark for this function.
I suggest you make it one and compare two different versions, especially under the concurrent situation.

Seaven: I checked the implementation of TimeBasedEpochRandomGenerator, and by default, it uses SecureRandom to constructor it's _random. SecureRandom uses `synchronized` to generate random number, and since the `generate` function  of TimeBasedEpochRandomGenerator is already locked with a ReentrantLock, using the default SecureRandom may not be necessary. you could try constructing TimeBasedEpochRandomGenerator with a regular Random instead; this might offer better performance than the default TimeBasedEpochRandomGenerator. However, in high-concurrency scenarios, the ReentrantLock can still be a bottleneck.

dolfinus: I've added benchmark results & source code here https://github.com/StarRocks/starrocks/issues/58962#issuecomment-2945751685. Results:

* Default `TimeBasedEpochRandomGenerator(SecureRandom)` is not slower than `UUID.randomUUID()` used before #59098 merged. But it is ~3x slower than `TimeBasedGenerator(SecureRandom)` used in `UUIDUtil.genUUID()` currently.
* `TimeBasedEpochRandomGenerator(Random)` is ~3..4x faster than `TimeBasedEpochRandomGenerator(SecureRandom)`, in both single in multiple threads.
* Using the same generator object with 4 threads is 6..7x times slower than in single thread case. This is true both for UUIDv7 and UUIDv4 generators. This could be almost completely eliminated by wrapping the UUID generator with `ThreadLocal<T>`, with only ~5% penalty in single thread case.
* Using custom UUIDv7 generator with Random gives ~20% boost compairing to fasterxml implementation. This could be caused by a fact that fasterxml uses [random.nextBytes(byte[])](https://github.com/cowtowncoder/java-uuid-generator/blob/37d4ea4cfed174eecef5594df98918eb6b42b23d/src/main/java/com/fasterxml/uuid/impl/RandomBasedGenerator.java#L71-L84) when custom generator calls `random.nextLong()` twice. But this is a speculation, I haven't used any profiling tool here.

TL;DR:
* Combination of `ThreadLocal<TimeBasedEpochRandomGenerator>(new Random)` makes UUID generation ~30% faster in single thread case and ~4x in multithread case than now.
* Custom UUIDv7 generator with ThreadLocal + Random makes generation ~40% faster in single thread case and ~5x in multithread case.

Which option should be chosen?

alvin-phoenix-ai: I think we can use TimeBasedEpochRandomGenerator with ThreadLocal and a new Random instance. 
We should provide different Random numbers for each generator in a thread to avoid the same random values.

dolfinus: Done

Seaven: I also agree with the solution of using TimeBasedEpochRandomGenerator with ThreadLocal and a new Random instance.
and we can use the thread id as the seed, which can avoid the same random values

dolfinus: > we can use the thread id as the seed, which can avoid the same random values

According to Java docs:

https://docs.oracle.com/javase/8/docs/api/java/util/Random.html
> If two instances of Random are created with the same seed, and the same sequence of method calls is made for each, they will generate and return identical sequences of numbers.

https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#getId--
> When a thread is terminated, this thread ID may be reused.

So threadId is not globally unique - it's just a in-process counter starting with 1, which can be reused by another thread, and different processes may return the same threadId. Using it as a seed will produce same random values, which is worse than current implementation.
Instead I've altered generator initialization to be seeded with `SecureRandom`, to make it truly random.

fe/fe-core/src/main/java/com/starrocks/common/util/UUIDUtil.java
```diff
@@ -15,22 +15,32 @@
...
+    // Prefer UUIDv7 values as they are ordered, and include timestamp.
+    // Using ThreadLocal to avoid lock contention in multithread environment.
+    // See https://github.com/StarRocks/starrocks/issues/58962 for more details.
+    private static final ThreadLocal<TimeBasedEpochRandomGenerator> UUID7_FACTORY =
+            new ThreadLocal<>() {
+                @Override protected TimeBasedEpochRandomGenerator initialValue() {
+                    // SecureRandom return truly random values, but it is 4x slower than Random.
+                    // Random is pseudo-random, so it has to be initialized with a random seed
+                    // to avoid producing same results in a different thread
+                    SecureRandom trueRandom = new SecureRandom();
```
alvin-phoenix-ai: If this is a SecureRandom, is it possible to make it a static member of UUIDUtil rather than generating a new one for each thread?

dolfinus: Yes, but is it worth it? This influences only the time of first UUID generated within a thread

alvin-phoenix-ai: For the performance aspect, it is not necessary. 
Another potential benefit is that we can guarantee all seeds are unique.

According to the documents

> Many SecureRandom implementations are in the form of a pseudo-random number generator (PRNG), which means they use a deterministic algorithm to produce a pseudo-random sequence from a true random seed. Other implementations may produce true random numbers, and yet others may use a combination of both techniques.

Doing it can help avoid potential risks.

dolfinus: Sounds reasonable, done

alvin-phoenix-ai (approved):

Seaven (approved):

PR #30386, opened 2023-09-04
[Enhancement] optimize sending large results

fzhedu:
2 optimizations in `BufferControlBlock`, increasing 100 MB/s, from 300MB/s to 400MB/s during transfering data from BE to FE.
1. move serializing data from fetching thread to pushing thread;
2. limit queue by memory usage and block size, instead of by row num.

as no idea of whether results are from sorting, so can't use concurrentQueue.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function

## Bugfix cherry-pick branch check:

- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 3.1
  - [ ] 3.0
  - [ ] 2.5
  - [ ] 2.4

PR #59153, opened 2025-05-20
[BugFix] Fix result queue capacity of result sink

ZiheLiu:
## Why I'm doing:

Introduced by #30386.

The result queue capacity is `dop*2`. 

However, this dop is get by `state->query_options().pipeline_dop`, which is the raw value of the session variable  `pipeline_dop` (0 by default). 

This results that the queue capacity is 0, which means the queue will be full after putting one result batch. And therefore, result sink can only buffer one chunk.

### Test

16Core 64GB Memory BE.

```sql
select k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1,k1 
from d1.__row_util_base 
limit 100000000
```
ResultDeliverTime: 
- Before: 35s139ms
- After:    17s432ms

ResultDeliverTime is decreased, **but the total execution time between these two versions are the same.**

## What I'm doing:

Use actual DOP to set queue capacity.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/exec/data_sink.cpp
+++ b/be/src/exec/data_sink.cpp
@@ -291,7 +291,7 @@ Status DataSink::decompose_data_sink_to_pipeline(pipeline::PipelineBuilderContex
                                                            result_sink->get_file_opts(), dop, fragment_ctx);
         } else {
             op = std::make_shared<ResultSinkOperatorFactory>(
-                    context->next_operator_id(), result_sink->get_sink_type(), result_sink->isBinaryFormat(),
+                    context->next_operator_id(), dop, result_sink->get_sink_type(), result_sink->isBinaryFormat(),
                     result_sink->get_format_type(), result_sink->get_output_exprs(), fragment_ctx,
                     result_sink->get_row_desc());
         }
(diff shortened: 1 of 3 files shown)


alvin-phoenix-ai (approved):
LGTM
Could you elaborate a little more on the influence of query in the commit message? Will the performance be influenced?
I'm not very sure about the consequence of one slot in the buffer.

ZiheLiu:
@alvin-celerdata Added a performance test in the description. `ResultDeliverTime` is decreased, but **the total execution time between these two versions are the same**.

kangkaisen (approved):

satanson (approved):

PR #59161, opened 2025-05-20
[Enhancement] apply the task_runs_max_history_number to task_runs

murphyatwork:
## What I'm doing:

To apply `config.task_runs_max_history_number` to `information_schema.task_runs`:

1. **Objective**: Limit the number of records retrieved by this view to optimize memory usage effectively.
2. **Impact**: The `task_runs` table will enforce a defined maximum number of records, ensuring controlled resource consumption.

**Behavior change:**
1. Before
    - Before version 3.3, task_runs is stored in memory, which is limited by constraints, `task_runs_ttl_second` and `task_runs_max_history_number`
    -  Starting from version 3.3, the `information_schema.task_runs` table no longer has a limit on the number of records but is instead constrained by a TTL setting, `task_runs_ttl_second=7 days`, provided that the configuration `config.enable_task_history_archive=true` is enabled.  
3. Following this update, the `information_schema.task_runs` table will be subject to two restrictions:  
   - `task_runs_ttl_second`: Defines the maximum time-to-live (TTL) for records.  
   - `task_runs_max_history_number`: Specifies the maximum number of records allowed.  

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -400,8 +400,7 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true, comment = "task run ttl")
     public static int task_runs_ttl_second = 7 * 24 * 3600;     // 7 day
 
-    @Deprecated
-    @ConfField(mutable = true, comment = "[DEPRECATED as enable_task_history_archive] max number of task run history. ")
+    @ConfField(mutable = true, comment = "max number of task run history. ")
     public static int task_runs_max_history_number = 10000;
 
     @ConfField(mutable = true, comment = "Minimum schedule interval of a task")
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/history/TaskRunHistoryTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/history/TaskRunHistoryTable.java
@@ -210,8 +210,11 @@ public class TaskRunHistoryTable {
             predicates.add(" task_state = " + Strings.quote(params.getState()));
         }
         sql += Joiner.on(" AND ").join(predicates);
+        // If user explicitly specify the LIMIT in sql, we don't apply default limit
         if (params.isSetPagination() && params.getPagination().getLimit() > 0) {
             sql += " LIMIT " + params.getPagination().getLimit();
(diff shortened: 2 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -400,8 +400,7 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true, comment = "task run ttl")
     public static int task_runs_ttl_second = 7 * 24 * 3600;     // 7 day
 
-    @Deprecated
-    @ConfField(mutable = true, comment = "[DEPRECATED as enable_task_history_archive] max number of task run history. ")
+    @ConfField(mutable = true, comment = "max number of task run history. ")
```
alvin-celerdata, line 403: Could you please verify if the document needs to be updated accordingly?

murphyatwork: no need

murphyatwork: because I forgot to change the document when I deprecated this configuration last time, so i can reuse it right now without any modification of doc

LiShuMing: The current meaning is different from original setting:
Before: `task_runs_max_history_number` is used to limit the whole history task runs records  to avoid occupying too much memory.
After:  `task_runs_max_history_number` is used to limit querying history table's size.

Am I right?

alvin-phoenix-ai (approved):

fe/fe-core/src/main/java/com/starrocks/common/Config.java
murphyatwork: Yes, but from the user's perspective, it's the same—you can only retrieve a maximum of task_runs_max_history_number records from this table.

stephen-shelby (approved):

PR #59178, opened 2025-05-21
[BugFix] Fix query fail after rename column name if scan node contain unused output column name

srlch:
## Why I'm doing:
Query fail after rename column name if scan node contain unused output column name because the unused output column name use the new column name instead of the base name(old name) and fail in BE side.

## What I'm doing:
Set base column name(old name) for unused output column name in scan node.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
@@ -329,8 +329,7 @@ public class OlapScanNode extends ScanNode {
                 continue;
             }
             if (unUsedOutputColumnIds.contains(slot.getId().asInt())) {
-                Column col = slot.getColumn();
-                unUsedOutputStringColumns.add(col.isShadowColumn() ? col.getName() : col.getColumnId().getId());
+                unUsedOutputStringColumns.add(slot.getColumn().getColumnId().getId());
             }
         }
     }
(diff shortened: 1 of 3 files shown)


gengjun-git (dismissed):

fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
```diff
@@ -329,7 +329,8 @@ public void setUnUsedOutputStringColumns(Set<Integer> unUsedOutputColumnIds) {
                 continue;
             }
             if (unUsedOutputColumnIds.contains(slot.getId().asInt())) {
-                unUsedOutputStringColumns.add(slot.getColumn().getName());
+                Column col = slot.getColumn();
+                unUsedOutputStringColumns.add(col.isShadowColumn() ? col.getName() : col.getColumnId().getId());
```
wyb, line 333: why need shadow column check?

srlch: updated, remove the check is ok

wyb (approved):

Seaven (approved):

PR #59219, opened 2025-05-22
[BugFix] fix persistent index compatibility issue when migrate between different cpu arch

### LINKED ISSUE #57952: Crash of starrocks BE on ARM

luohaha:
## Why I'm doing:
In the current implementation, the FixedMutableIndex of the persistent index uses two different data structures depending on whether the CPU supports SSE instructions (https://github.com/StarRocks/starrocks/blob/0d19cb4f9bc58d0cab5237a469b9e4bd30c0eb31/be/src/util/phmap/phmap.h#L447). 
Since ARM architecture CPUs do not support SSE instructions, when migrating from x86 to ARM, the differing implementations cause the loaded FixedMutableIndex to become corrupted.

## What I'm doing:
1. For the subsequent generation of snapshots in the persistent index, all snapshots will be converted to the new format to support cross-CPU architectures.
2. For snapshots generated by older versions of StarRocks, a completeness_check will be performed to verify whether the snapshot file is complete and compatible with the current CPU architecture. If it passes the check, the old format will be used for loading.
3. If the completeness_check fails, it indicates that the file is either corrupted or was generated on a different CPU architecture. In this case, compatibility loading will be skipped, and the snapshot will be rebuilt.

Fix #57952

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1


be/src/storage/persistent_index.cpp
```diff
@@ -60,6 +60,12 @@ constexpr size_t kPackSize = 16;
 constexpr size_t kBucketSizeMax = 256;
 constexpr size_t kFixedMaxKeySize = 128;
 constexpr size_t kBatchBloomFilterReadSize = 4ULL << 20;
+/*
+The introduction of this magic number serves two purposes:
+1. To detect endianness mismatches in cross-platform scenarios
+2. To identify the new snapshot encoding format
+*/
```
alvin-celerdata, line 67:
```suggestion
// The introduction of this magic number serves two purposes:
// 1. To detect endianness mismatches in cross-platform scenarios
// 2. To identify the new snapshot encoding format
```
Prefer comment starts with`//`

be/test/storage/persistent_index_snapshot_load_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+        vector<IndexValue> values;
+        vector<size_t> idxes;
+        key_slices.reserve(N);
+        idxes.reserve(N);
+        for (int i = 0; i < N; i++) {
+            if constexpr (std::is_same_v<Key, uint32_t>) {
+                ASSERT_TRUE(GetParam().fixed_key_size)
+                        << "KeyType is uint32_t, but GetParam().fixed_key_size is false.";
+                keys[i] = i;
+                values.emplace_back(i * 2); // 假设 IndexValue 可以这样构造
```
alvin-celerdata, line 64: please use English

dengliu (approved):

be/src/storage/persistent_index.cpp
```diff
@@ -1006,7 +1027,26 @@ class FixedMutableIndex : public MutableIndex {
     // will use `sizeof(size_t)` as return value.
     size_t dump_bound() override { return _map.empty() ? sizeof(size_t) : _map.dump_bound(); }
 
-    bool dump(phmap::BinaryOutputArchive& ar) override { return _map.dump(ar); }
+    bool dump(phmap::BinaryOutputArchive& ar) override {
```
alvin-celerdata, line 1030: As far as I know, this dump should be compatible with phmap dump behavior in X86 platform.
You should add comments in the codebase to let others know why you implemented similar logic in this function, rather than just calling _map.dump() directly.

alvin-celerdata, line 1030: Can you change the function return type from bool to Status, which can have more meaningful information.

be/src/storage/persistent_index.cpp
```diff
@@ -1964,6 +2004,13 @@ Status ShardByLengthMutableIndex::append_wal(const Slice* keys, const IndexValue
 }
 
 Status ShardByLengthMutableIndex::load_snapshot(phmap::BinaryInputArchive& ar, const std::set<uint32_t>& idxes) {
+    uint32_t magic_num = 0;
+    RETURN_IF(!ar.load(&magic_num), Status::Corruption("ShardByLengthMutableIndex load snapshot magic num failed"));
+    if (magic_num != kSnapshotMagicNum) {
+        return Status::Corruption(
+                fmt::format("ShardByLengthMutableIndex load snapshot magic num failed, expect {}, but got {}",
+                            kSnapshotMagicNum, magic_num));
+    }
```
alvin-celerdata, line 2013: Will this new check cause all persistent index corruption? Because the previously generated snapshot has no magic number.

be/src/storage/persistent_index.cpp
```diff
@@ -1978,6 +2025,9 @@ size_t ShardByLengthMutableIndex::dump_bound() {
 }
 
 bool ShardByLengthMutableIndex::dump(phmap::BinaryOutputArchive& ar_out, std::set<uint32_t>& dumped_shard_idxes) {
+    if (!ar_out.dump(kSnapshotMagicNum)) {
+        return false;
+    }
```
alvin-celerdata, line 2030: prefer Status return type

be/test/storage/persistent_index_snapshot_load_test.cpp
luohaha: Ok, this is generared by AI...

be/src/storage/persistent_index.cpp
luohaha: Persistent index will be rebuilt

alvin-phoenix-ai (changes requested):
sorry, I leave a request changes because of behavior changes.

be/src/storage/persistent_index.cpp
luohaha: done

be/src/util/phmap/phmap.h
```diff
@@ -1541,6 +1541,9 @@ class raw_hash_set {
 
     template <typename InputArchive>
     bool load(InputArchive&);
+
+    template <typename InputArchive>
+    bool completeness_check(InputArchive& ar);
```
alvin-celerdata, line 1546:
```suggestion
    Status completeness_check(const InputArchive& ar);
```

be/src/util/phmap/phmap_dump.h
```diff
@@ -143,6 +143,39 @@ bool raw_hash_set<Policy, Hash, Eq, Alloc>::load(InputArchive& ar) {
...
+template <class Policy, class Hash, class Eq, class Alloc>
+template <typename InputArchive>
+bool raw_hash_set<Policy, Hash, Eq, Alloc>::completeness_check(InputArchive& ar) {
+    static_assert(type_traits_internal::IsTriviallyCopyable<value_type>::value,
+                  "value_type should be trivially copyable");
+    raw_hash_set<Policy, Hash, Eq, Alloc>().swap(*this); // clear any existing content
+    size_t size = 0;
+    if (!ar.load(&size)) {
+        std::cerr << "Failed to load size" << std::endl;
+        return false;
```
alvin-celerdata, line 155: 1. std::cerr/std::cout is not allowed in our codebase. Only LOG is allowed.
2. It is not a good design choice to print logs in a basic structure class. Image that std::vector will print log, it will make it difficult to handle it.
2. Considering returning Status with detail information.

be/src/storage/persistent_index.h
```diff
@@ -281,6 +281,10 @@ class MutableIndex {
 
     virtual Status pk_dump(PrimaryKeyDump* dump, PrimaryIndexDumpPB* dump_pb) = 0;
 
+    virtual void set_format_version(uint32_t ver) = 0;
+
+    virtual Status completeness_check(phmap::BinaryInputArchive& ar) = 0;
```
alvin-celerdata, line 286:
```suggestion
    virtual Status completeness_check(const phmap::BinaryInputArchive& ar) = 0;
```
use `const` as much as possible

be/src/storage/persistent_index.cpp
```diff
@@ -1977,17 +2098,22 @@ size_t ShardByLengthMutableIndex::dump_bound() {
                            [](size_t s, const auto& e) { return e->size() > 0 ? s + e->dump_bound() : s; });
 }
 
-bool ShardByLengthMutableIndex::dump(phmap::BinaryOutputArchive& ar_out, std::set<uint32_t>& dumped_shard_idxes) {
+Status ShardByLengthMutableIndex::dump(phmap::BinaryOutputArchive& ar_out, std::set<uint32_t>& dumped_shard_idxes) {
+    bool use_old_format = false;
+    TEST_SYNC_POINT_CALLBACK("ShardByLengthMutableIndex::dump::1", &use_old_format);
+    if (LIKELY(!use_old_format)) {
```
alvin-celerdata, line 2104: Please add more comments here to explain the reason why you do it this way.

alvin-celerdata: What will happen if a new version snapshot is generated? However, the image is rolled back due to other reasons. Will it cause the old version image to crash forever?

be/src/storage/persistent_index.cpp
```diff
@@ -1058,8 +1120,11 @@ class FixedMutableIndex : public MutableIndex {
 
     size_t memory_usage() override { return _map.capacity() * (1 + (KeySize + 3) / 4 * 4 + kIndexValueSize); }
 
+    void set_format_version(uint32_t ver) override { format_version = ver; }
+
 private:
     phmap::flat_hash_map<KeyType, IndexValue, FixedKeyHash<KeySize>> _map;
+    uint32_t format_version = kMutableIndexFormatVersion2;
```
alvin-celerdata, line 1127: Should follow the naming rules.
```suggestion
    uint32_t _format_version = kMutableIndexFormatVersion2;
```

chaoyli (dismissed):

be/src/storage/persistent_index.h
luohaha: It can't be const, there is `ifstream` in `BinaryInputArchive`, and `completeness_check` will lead to state change in `ifstream`.

be/src/util/phmap/phmap.h
luohaha: done

be/src/storage/persistent_index.cpp
luohaha: If BE rollback, then old BE can't recognize new format snapshot. So I increase the persistent index format version to `PERSISTENT_INDEX_VERSION_7`, so when rollback happen, it will be rebuilt.

be/src/storage/persistent_index.h
```diff
@@ -74,7 +74,8 @@ enum PersistentIndexFileVersion {
     PERSISTENT_INDEX_VERSION_3,
     PERSISTENT_INDEX_VERSION_4,
     PERSISTENT_INDEX_VERSION_5,
-    PERSISTENT_INDEX_VERSION_6
+    PERSISTENT_INDEX_VERSION_6,
+    PERSISTENT_INDEX_VERSION_7
```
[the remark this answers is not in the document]
luohaha: his is a legacy issue, and we will refactor it in subsequent PRs.

be/src/storage/persistent_index.cpp
```diff
@@ -1060,8 +1120,11 @@ class FixedMutableIndex : public MutableIndex {
 
     size_t memory_usage() override { return _map.capacity() * (1 + (KeySize + 3) / 4 * 4 + kIndexValueSize); }
 
+    void set_format_version(uint32_t ver) override { _format_version = ver; }
+
 private:
     phmap::flat_hash_map<KeyType, IndexValue, FixedKeyHash<KeySize>> _map;
+    uint32_t _format_version = kMutableIndexFormatVersion2;
```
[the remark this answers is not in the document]
luohaha: There is not format version in `MutableIndex`, previous version is in `ImmutableIndex`

be/src/storage/persistent_index.h
```diff
@@ -281,6 +282,10 @@ class MutableIndex {
 
     virtual Status pk_dump(PrimaryKeyDump* dump, PrimaryIndexDumpPB* dump_pb) = 0;
 
+    virtual void set_format_version(uint32_t ver) = 0;
```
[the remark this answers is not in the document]
luohaha: The new ​​PERSISTENT_INDEX_VERSION_xx​​ and the old ​​kMutableIndexFormatVersion1​​ may coexist because:

1. Each time the ​​persistent index​​ is updated, we record the latest ​​PERSISTENT_INDEX_VERSION_xx​​ in RocksDB's metadata.
2. However, not every write triggers a ​​dump​​, so the snapshot file may still be in the old format.
3. Therefore, an additional ​​format version​​ identifier is needed to distinguish between versions.

be/src/storage/persistent_index.cpp
```diff
@@ -1348,34 +1411,50 @@ class SliceMutableIndex : public MutableIndex {
     //  |total num|| data size ||  data    | ... | data size ||  data    |
     size_t dump_bound() override { return sizeof(size_t) * (1 + size()) + _total_kv_pairs_usage; }
 
-    bool dump(phmap::BinaryOutputArchive& ar) override {
+    Status dump(phmap::BinaryOutputArchive& ar) override {
```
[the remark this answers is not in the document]
luohaha: This is a legacy issue too, they were introduced from day one. We can submit a refactor PR later.

be/src/storage/persistent_index.h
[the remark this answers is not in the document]
luohaha: changed. done.

be/src/util/phmap/phmap_dump.h
```diff
@@ -143,6 +143,35 @@ bool raw_hash_set<Policy, Hash, Eq, Alloc>::load(InputArchive& ar) {
...
+template <class Policy, class Hash, class Eq, class Alloc>
+template <typename InputArchive>
+starrocks::Status raw_hash_set<Policy, Hash, Eq, Alloc>::completeness_check(InputArchive& ar) {
+    static_assert(type_traits_internal::IsTriviallyCopyable<value_type>::value,
+                  "value_type should be trivially copyable");
+    raw_hash_set<Policy, Hash, Eq, Alloc>().swap(*this); // clear any existing content
+    size_t size = 0;
+    if (!ar.load(&size)) {
+        return starrocks::Status::InternalError("Failed to load size");
+    }
```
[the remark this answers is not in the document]
luohaha: done

decster (dismissed):

be/src/exprs/agg/distinct.h
```diff
@@ -72,18 +72,18 @@ struct DistinctAggregateState<LT, SumLT, FixedLengthLTGuard<LT>> {
 
     void serialize(uint8_t* dst) const {
         phmap::InMemoryOutput output(reinterpret_cast<char*>(dst));
-        set.dump(output);
+        (void)set.dump(output);
```
alvin-celerdata, line 75:
``` suggestion
        [[maybe_unused]] set.dump(output);
```

alvin-celerdata: It will make compilation fail in CentOS. https://github.com/StarRocks/starrocks/pull/58419

be/src/exprs/agg/distinct.h
```diff
@@ -72,18 +72,18 @@ struct DistinctAggregateState<LT, SumLT, FixedLengthLTGuard<LT>> {
...
         DCHECK(output.length() == set.dump_bound());
     }
 
     void deserialize_and_merge(const uint8_t* src, size_t len) {
         phmap::InMemoryInput input(reinterpret_cast<const char*>(src));
         auto old_size = set.size();
         if (old_size == 0) {
-            set.load(input);
+            (void)set.load(input);
```
alvin-celerdata, line 83: ditto

be/src/exprs/agg/distinct.h
```diff
@@ -72,18 +72,18 @@ struct DistinctAggregateState<LT, SumLT, FixedLengthLTGuard<LT>> {
...
         } else {
             MyHashSet set_src;
-            set_src.load(input);
+            (void)set_src.load(input);
```
alvin-celerdata, line 86: ditto

be/src/storage/persistent_index.cpp
```diff
@@ -1965,7 +2040,45 @@ Status ShardByLengthMutableIndex::append_wal(const Slice* keys, const IndexValue
...
+    auto check_arch_fn = [&]() {
+        ar.reset();
+        for (const auto idx : idxes) {
+            if (!_shards[idx]->completeness_check(ar).ok()) {
+                return false;
+            }
+        }
+        // Must reach the end of the file.
+        return ar.eof();
+    };
```
alvin-celerdata, line 2056: Better to use Status as the return type to pass along the error message

alvin-phoenix-ai (approved):

chaoyli (approved):

luohaha:
https://github.com/Mergifyio backport branch-3.3

PR #59220, opened 2025-05-22
[Enhancement] support disable cloud partition compaction

starrocks-xupeng:
## What I'm doing:

add lake_compaction_disable_ids to supporting disabling table or partition compaction

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -2880,9 +2880,11 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true, comment = "True to start warehouse idle checker")
     public static boolean warehouse_idle_check_enable = false;
 
-    // e.g. "tableId1;tableId2"
-    @ConfField(mutable = true)
-    public static String lake_compaction_disable_tables = "";
+    // e.g. "tableId1;partitionId2"
+    // both table id and partition id are supported
+    @ConfField(mutable = true, comment = "disable table or partition compaction, format:'id1;id2'",
+            aliases = {"lake_compaction_disable_tables"})
+    public static String lake_compaction_disable_ids = "";
 
     @ConfField(mutable = true, comment = "the max number of threads for lake table publishing version")
     public static int lake_publish_version_max_threads = 512;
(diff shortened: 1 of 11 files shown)


fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -2881,9 +2881,14 @@ public class Config extends ConfigBase {
...
 
     // e.g. "tableId1;tableId2"
-    @ConfField(mutable = true)
+    @ConfField(mutable = true, comment = "Deprecated")
     public static String lake_compaction_disable_tables = "";
 
+    // e.g. "tableId1;partitionId2"
+    // both table id and partition id are supported
+    @ConfField(mutable = true, comment = "disable table or partition compaction, format:'id1;id2'")
+    public static String lake_compaction_disable_ids = "";
```
xiangguangyxg, line 2890: Why do not keep the old config item `lake_compaction_disable_tables`, and add a new config item `lake_compaction_disable_partitions` ?

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -2881,9 +2881,14 @@ public class Config extends ConfigBase {
     public static boolean warehouse_idle_check_enable = false;
 
     // e.g. "tableId1;tableId2"
-    @ConfField(mutable = true)
+    @ConfField(mutable = true, comment = "Deprecated")
```
wyb, line 2884: how about using `@ConfField(mutable = true, aliases = {"lake_compaction_disable_ids"})`

fe/fe-core/src/main/java/com/starrocks/common/Config.java
starrocks-xupeng: need to gradually discard lake_compaction_disable_tables, use just one config instead

fe/fe-core/src/main/java/com/starrocks/common/Config.java
starrocks-xupeng: i put this aliases to lake_compaction_disable_tables

starrocks-xupeng: i put this aliases to lake_compaction_disable_tables, so we can just use lake_compaction_disable_ids

wyb (approved):

xiangguangyxg (approved):

PR #59358, opened 2025-05-26
[Feature] (CNResourceProvider Part1) Introduce ComputeNode Resource and ResourceProvider

### LINKED ISSUE #59332: [Feature] Introduce ComputeNode Resource Provider
Currently, we only use `warehouseId` to mark the warehouse's resource which is limited for multi CNGroups' Warehouse.  Introduce `CNResource` and `CNResourceProvider` layer between `ComputeNode CNResourceProvider` and `Warehouse Manger`.

### CNResource
```
/**
 * 1. CNResource is a compute node resource acquired from a warehouse or some other resources.
 * 2. It can be a warehouse or a cngroup of warehouse which is to represent the resource that can be used for compute.
 */
public interface CNResource {

    /**
     * Get the id of the CNResource
     * @return: the id of the CNResource
     */
    long getWarehouseId();

    /**
     * Get the id of the worker group that this CNResource belongs to.
     * @return: the id of the worker group
     */
    long getWorkerGroupId();
}
```

### CNResourceProvider
```
/**
 * WorkerGroupProvider provides an available cngroup to a job scheduler based on the current warehouse's load status or strategy.
 */
public interface CNResourceProvider {
    /**
     * NOTE: prefer to call this infrequently, as it can come to dominate the execution time of a query in the
     *  frontend if there are many calls per request (e.g. one per partition when there are many partitions).
     * @param warehouse: the warehouse to get the worker group from
     * @param acquireContext: the context to acquire the worker group
     * @return: an available CNGroupResource for the warehouse by the strategy
     * @throws RuntimeException : if the warehouse is invalid or there is no available worker group
     */
    Optional<CNResource> acquireCNResource(Warehouse warehouse, CNAcquireContext acquireContext);

    /**
     * Check Whether the resource is available, this method will not throw exception
     * @param cnResource: the CNResource to check
     * @return: true if the resource is available, false otherwise
     */
    boolean isResourceAvailable(CNResource cnResource);

    /**
     * Get all compute node ids in the CNResource without checking the alive status
     * @param cnResource: the CNResource to get the compute node ids from
     * @return: a list of compute node ids, empty if the CNResource is not available
     */
    List<Long> getAllComputeNodeIds(CNResource cnResource);

    /**
     * Get all alive compute nodes in the CNResource
     * @param cnResource: the CNResource to get the alive compute nodes from
     * @return: a list of alive compute nodes, empty if the CNResource is not available
     */
    List<ComputeNode> getAliveComputeNodes(CNResource cnResource);
}

```

LiShuMing:
## Why I'm doing:
Currently, we only use `warehouseId` to mark the warehouse's resource which is limited in multi CNGroups' Warehouse env. 

## What I'm doing:
- Introduce `CNResource` and `CNResourceProvider` between ComputeNode CNResourceProvider and Warehouse Manger.

## NOTES:
For a specific CNResource, during the execution of a job, the CNResource schedule should 
- (Scheduling Integrity) The CNResource assigned for each job’s minimum execution unit must remain consistent:
  - It should not happen that within a single execution unit, some tablets are assigned to CNResource1 while others are assigned to CNResource2, as this breaks task integrity and leads to poor cache performance.
  - Even worse would be assigning the same tablet to different CNResources (e.g., CNResource1 and CNResource2) during execution,  which would cause significant scheduling and execution issues.

Fixes https://github.com/StarRocks/starrocks/issues/59332

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
@@ -42,8 +42,8 @@ import com.starrocks.system.ComputeNode;
 import com.starrocks.system.SystemInfoService;
 import com.starrocks.warehouse.DefaultWarehouse;
 import com.starrocks.warehouse.Warehouse;
-import com.starrocks.warehouse.cngroup.CNResource;
-import com.starrocks.warehouse.cngroup.CNWarehouseResource;
+import com.starrocks.warehouse.cngroup.ComputeResource;
+import com.starrocks.warehouse.cngroup.WarehouseComputeResource;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -67,7 +67,7 @@ public class WarehouseManager implements Writable {
     public static final long DEFAULT_WAREHOUSE_ID = 0L;
 
     // default compute resource
-    public static final CNResource DEFAULT_CNRESOURCE = CNWarehouseResource.DEFAULT;
+    public static final ComputeResource DEFAULT_CNRESOURCE = WarehouseComputeResource.DEFAULT;
 
     protected final Map<Long, Warehouse> idToWh = new HashMap<>();
     protected final Map<String, Warehouse> nameToWh = new HashMap<>();
--- a/fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
@@ -42,6 +42,8 @@ import com.starrocks.system.ComputeNode;
 import com.starrocks.system.SystemInfoService;
 import com.starrocks.warehouse.DefaultWarehouse;
 import com.starrocks.warehouse.Warehouse;
+import com.starrocks.warehouse.cngroup.CNResource;
+import com.starrocks.warehouse.cngroup.CNWarehouseResource;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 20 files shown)


HangyuanLiu (dismissed):

satanson (dismissed):

fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
```diff
@@ -42,6 +42,8 @@
 import com.starrocks.system.SystemInfoService;
 import com.starrocks.warehouse.DefaultWarehouse;
 import com.starrocks.warehouse.Warehouse;
+import com.starrocks.warehouse.cngroup.CNResource;
+import com.starrocks.warehouse.cngroup.CNWarehouseResource;
```
nshangyimingCD, line 46: CN = Compute Node? If that's case, the naming here may sound wierd.
Warehouse = a bunch of compute nodes or a bunch of compute node groups, so what does `CNWarehouseResource` mean?
I think maybe we just call it `ComputeResource` and `WarehouseComputeResource`/`CNGroupComputeResource`, how do you think?

fe/fe-core/src/main/java/com/starrocks/warehouse/cngroup/CNAcquireContext.java
```diff
@@ -0,0 +1,79 @@
...
+import com.starrocks.common.ErrorReportException;
+import com.starrocks.server.GlobalStateMgr;
+import com.starrocks.server.WarehouseManager;
+import com.starrocks.warehouse.Warehouse;
+
+/**
+ * CNAcquireContext is the context for acquiring CNResource from a warehouse.
+ */
+public class CNAcquireContext {
+    // The id of the warehouse which is must be specified.
```
nshangyimingCD, line 27:
```suggestion
    // The id of the warehouse which must be specified.
```

nshangyimingCD (approved):

nshangyiming (approved):

HangyuanLiu (approved):

Issue #59134, opened 2025-05-20
Dynamic tablet splitting and merging


xiangguangyxg:
- [x] https://github.com/StarRocks/starrocks/pull/60645
- [x] https://github.com/StarRocks/starrocks/pull/60423
- [x] https://github.com/StarRocks/starrocks/pull/59993
- [x] https://github.com/StarRocks/starrocks/pull/59441

## Feature request

StarRocks uses a partitioning + bucketing two-tier data distribution strategy, to evenly distribute data across BEs or CNs. A well-designed data distribution strategy is crucial to performance, cost and stability. 
However, it is difficult for users to design a good data distribution strategy for the following reasons:
- It is difficult to choose an appropriate bucket number when creating a new table or a new partition, even if the system automatically calculates it. 
- It is costly to modify the bucket number of an existing table dynamically when it is found to be inappropriate.
- The most appropriate bucket number of a table is not fixed and will change with the development of business and the cluster specifications.

The dynamic tablet splitting and merging feature achieves the following goals:
- When creating tables or partitions, there is no need to estimate the final bucket number. Users or systems could provide an initial bucket number, then the number of tablets in a partition can grow or shrink dynamically. In most cases, users could not care about the bucket number. The initial number is calculated by system, and then the system can split and merge buckets automatically.
- Too large tablets and too many tablets with small sizes can be avoided. 
- The performance should not degrade compared to the perfect data distribution manually.
- Prioritize support dynamic tablet splitting and merging in shared-data mode.

PR #59378, opened 2025-05-27
[Feature] Introduce virtual buckets metadata for dynamic tablet splitting and merging

xiangguangyxg:
## Why I'm doing:
This pr is a preliminary work for dynamic tablet splitting and merging.
We divide data into virtual buckets and then arrange these virtual buckets into physical buckets, which are tablets. Each virtual bucket is associated with a tablet. Multiple virtual buckets are allowed to be associated with the same tablet.
Before virtual buckets are introduced, logical buckets and physical buckets (tablets) are one-to-one, one bucket corresponds to one tablet, the bucket number is always equal to the tablet number.
After virtual buckets are introduced, logical buckets and physical buckets (tablets) are many-to-one, multiple buckets correspond to one tablet, the bucket number is always greater or equal to the tablet number. When a table or a partition is created, its bucket number and tablet number are the same at the begining, and they may be different after a while.
The concept of bucket number is already exists in starrocks. We keep it as the virtual bucket number. The concept of tablet number needs to be introduced. Users need to understand that bucket number and tablet number are no longer always the same.

## What I'm doing:
Introduce virtual buckets metadata for dynamic tablet splitting and merging.
The virtual buckets is not used in this pr, will be used in next pr.

This pr also add some fields to show virtual buckets.

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
@@ -102,9 +102,18 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
     @SerializedName(value = "rowCount")
     private long rowCount;
 
+    // Virtual buckets. Each virtual bucket saves a tablet id.
+    // We divide data into virtual buckets and then arrange these virtual buckets
+    // into physical buckets, which are tablets.
+    // Each virtual bucket is associated with a tablet. Multiple virtual buckets are
+    // allowed to be associated with the same tablet.
+    @SerializedName(value = "virtualBuckets")
+    private List<Long> virtualBuckets;
+
     private Map<Long, Tablet> idToTablets;
+
+    // Deprecated, virtual buckets keeps tablet order, keep this for compatibility
     @SerializedName(value = "tablets")
-    // this is for keeping tablet order
     private List<Tablet> tablets;
 
     @SerializedName(value = "shardGroupId")
@@ -143,6 +152,7 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
     public MaterializedIndex(long id, @Nullable IndexState state, long visibleTxnId, long shardGroupId) {
         this.id = id;
         this.state = state == null ? IndexState.NORMAL : state;
+        this.virtualBuckets = new ArrayList<>();
         this.idToTablets = new HashMap<>();
         this.tablets = new ArrayList<>();
         this.rowCount = 0;
@@ -184,6 +194,10 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
         return shardGroupId;
     }
 
+    public List<Long> getVirtualBuckets() {
+        return virtualBuckets;
+    }
+
     public List<Tablet> getTablets() {
         return tablets;
     }
@@ -201,6 +215,7 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
     }
 
     public void clearTabletsForRestore() {
+        virtualBuckets.clear();
         idToTablets.clear();
         tablets.clear();
     }
@@ -210,6 +225,7 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
     }
 
     public void addTablet(Tablet tablet, TabletMeta tabletMeta, boolean updateInvertedIndex) {
+        virtualBuckets.add(tablet.getId());
         idToTablets.put(tablet.getId(), tablet);
         tablets.add(tablet);
         if (updateInvertedIndex) {
@@ -280,6 +296,16 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
         }
     }
 
+    public List<Integer> getVirtualBucketsByTabletId(long tabletId) {
+        List<Integer> virtualBucketIndexes = new ArrayList<>();
+        for (int i = 0; i < virtualBuckets.size(); ++i) {
+            if (virtualBuckets.get(i).longValue() == tabletId) {
+                virtualBucketIndexes.add(i);
+            }
+        }
+        return virtualBucketIndexes;
+    }
+
     public int getTabletOrderIdx(long tabletId) {
         int idx = 0;
         for (Tablet tablet : tablets) {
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
```diff
@@ -102,9 +102,18 @@ public enum IndexExtState {
     @SerializedName(value = "rowCount")
     private long rowCount;
 
+    // Virtual buckets. Each virtual bucket saves a tablet id.
+    // We divide data into virtual buckets and then arrange these virtual buckets
+    // into physical buckets, which are tablets.
+    // Each virtual bucket is associated with a tablet. Multiple virtual buckets are
+    // allowed to be associated with the same tablet.
+    @SerializedName(value = "buckets")
+    private List<Long> buckets;
```
alvin-celerdata, line 111: If this stands for virtual bucket, I would like to name it virtualBuckets or vBuckets to eliminate the ambiguity.
Because the term "bucket" is also used in other contexts.

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
```diff
@@ -352,5 +380,12 @@ public void gsonPostProcess() {
         for (Tablet tablet : tablets) {
             idToTablets.put(tablet.getId(), tablet);
         }
+
+        // build "buckets" from "tablets" when upgrading from old versions
```
alvin-celerdata, line 384: // Before StarRocks didn't have virtual buckets, it would be empty when loading from the previous image. 
In this situation, the virtual bucket is equivalent to the physical tablet. So just fill the virtual buckets with the physical tablets.

It would be better to include the above explanation in the comment.

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
```diff
@@ -184,6 +194,10 @@ public long getShardGroupId() {
         return shardGroupId;
     }
 
+    public List<Long> getBuckets() {
```
alvin-celerdata, line 197:
```suggestion
    public List<Long> getVitualBuckets() {
```

fe/fe-core/src/main/java/com/starrocks/common/proc/LakeTabletsProcDir.java
```diff
@@ -42,7 +42,8 @@
  */
 public class LakeTabletsProcDir implements ProcDirInterface {
     public static final ImmutableList<String> TITLE_NAMES = new ImmutableList.Builder<String>()
-            .add("TabletId").add("BackendId").add("DataSize").add("RowCount").add("MinVersion")
+            .add("TabletId").add("BackendId").add("DataSize").add("RowCount")
```
alvin-celerdata, line 45: This will change the behavior; please change the documentation accordingly.

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
xiangguangyxg: changed to virtualBuckets

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
xiangguangyxg: done

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
```diff
@@ -280,6 +296,16 @@ public long getReplicaCount() {
         }
     }
 
+    public List<Integer> getVirtualBucketIndexes(long tablet) {
```
alvin-celerdata, line 299:
```suggestion
    public List<Integer> getVirtualBucketsByTabletId(long tablet) {
```

xiangguangyxg: done

alvin-phoenix-ai (dismissed):

fe/fe-core/src/main/java/com/starrocks/common/proc/IndicesProcDir.java
```diff
@@ -59,6 +59,7 @@
 public class IndicesProcDir implements ProcDirInterface {
     public static final ImmutableList<String> TITLE_NAMES = new ImmutableList.Builder<String>()
             .add("IndexId").add("IndexName").add("State").add("LastConsistencyCheckTime")
+            .add("Buckets").add("Tablets")
```
alvin-celerdata, line 62:
```suggestion
            .add("VirtualBuckets").add("Tablets")
```

alvin-phoenix-ai (approved):

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
```diff
@@ -102,9 +102,18 @@ public enum IndexExtState {
     @SerializedName(value = "rowCount")
     private long rowCount;
 
+    // Virtual buckets. Each virtual bucket saves a tablet id.
+    // We divide data into virtual buckets and then arrange these virtual buckets
+    // into physical buckets, which are tablets.
+    // Each virtual bucket is associated with a tablet. Multiple virtual buckets are
+    // allowed to be associated with the same tablet.
+    @SerializedName(value = "virtualBuckets")
+    private List<Long> virtualBuckets;
```
srlch, line 111: Shall we need to introduce the concept of virtual bucket id?

xiangguangyxg: Virtual bucket id is the index in the virtual bucket list.

gengjun-git (approved):

PR #59406, opened 2025-05-28
[BugFix] Add special handling for information_schema.tables to support Tableau's use of SELECT aliases in HAVING clause

### LINKED ISSUE #55626: tableau is broken
starrocks 3.4

using jdbc connection
Tableau looks for tables in database
```
SELECT TABLE_SCHEMA AS TABLE_CAT, NULL AS TABLE_SCHEM, TABLE_NAME, CASE WHEN TABLE_TYPE='BASE TABLE' THEN CASE WHEN TABLE_SCHEMA='mysql' OR TABLE_SCHEMA='performance_schema' OR TABLE_SCHEMA='sys' THEN 'SYSTEM TABLE' ELSE 'TABLE' END WHEN TABLE_TYPE='TEMPORARY' THEN 'LOCAL_TEMPORARY' ELSE TABLE_TYPE END AS TABLE_TYPE, TABLE_COMMENT AS REMARKS, NULL AS TYPE_CAT, NULL AS TYPE_SCHEM, NULL AS TYPE_NAME, NULL AS SELF_REFERENCING_COL_NAME, NULL AS REF_GENERATION FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'eshishkin' HAVING TABLE_TYPE IN ('LOCAL TEMPORARY','TABLE','VIEW',null,null) ORDER BY TABLE_TYPE, TABLE_SCHEMA, TABLE_NAME
```
and gets nothing

because we now return `TABLE_TYPE: BASE TABLE`

Zac-saodiseng:
## Why I'm doing:
Tableau issues queries on `information_schema.tables` that use SELECT aliases in the HAVING clause.
StarRocks by default disables `enable_groupby_use_output_alias`, causing query failures.

This PR adds special handling in `QueryAnalyzer` to detect queries on `information_schema.tables`
and temporarily enable `enable_groupby_use_output_alias` during query analysis.

The session variable is restored to its original value after analysis to avoid side effects on other queries.

This fix ensures compatibility with Tableau without changing the default behavior for other queries.
## What I'm doing:

Fixes #55626 

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1


fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
```diff
@@ -116,13 +116,36 @@ public class QueryAnalyzer {
...
+        // `enable_groupby_use_output_alias` to support Tableau compatibility.
+        if (containsInformationSchemaTables) {
+            session.getSessionVariable().setEnableGroupbyUseOutputAlias(true);
+
+            // Re-analyze the query with the session variable enabled.
+            // This is necessary because Tableau uses SELECT aliases in the HAVING clause.
+            new Visitor().process(node, new Scope(RelationId.anonymous(), new RelationFields()));
+            // Restore the original session variable to avoid affecting other queries.
+            session.getSessionVariable().setEnableGroupbyUseOutputAlias(originalGroupbyUseOutputAlias);
+        }
```
Seaven, line 148: don't need analyze again, can modify it in `visitSelect`

Astralidea (approved):

Seaven (approved):

PR #57819, opened 2025-04-09
[Enhancement] introduce show data distribution command to display data distribution

MatthewH00:
## Why I'm doing:
When data skew occurs, it can lead to cluster instability, but currently there are no good methods to detect it.

## What I'm doing:
introduce show data distribution command to display data distribution at the bucket level. 
notes : the command is facing the normal user(like data analyst), when occurs data skew , could use the command to detect data skew, then adjust table schema(bucket key) to fix.
syntax: `SHOW DATA DISTRIBUTION FROM [db_name.]tbl_name [PARTITION (p1, ...)];`
desc: 
 `PartitionName: when partition table display partition name, when unpartition table display table name`
 `BucketId: bucket id`
 `RowCount: row count`
 `RowCount%: at the current partition, current bucket's row count / total row count`
 `DataSize: data size`
 `DataSize%: at the current partition, current bucket's data size / total data size`

1.partition table：
 1)entire table
   `e.g. show data distribution from partition_tbl_name`
 2)single partition
   `e.g. show data distribution from partition_tbl_name partition(p1)`
 3)several partition
   `e.g. show data distribution from partition_tbl_name partition(p1,p2)`
1_N

2.unpartition table
 1)entire table
   `e.g. show data distribution from unpartition_tbl_name`
   `e.g. show data distribution from unpartition_tbl_name partition(unpartition_tbl_name)`
2_N

3.special case
 1)not exist db
   `return error like: Getting analyzing error. Detail message: Database db_name does not exsit.`
 2)not exist table
   `return error like: Getting analyzing error. Detail message: Table does not exist or is not native table: table_name.`
 3)not exist partition 
   `return error like: Getting analyzing error. Detail message: Partition does not exist: partition_name.`
 4)not privilege table
   `return error like: Access denied; you need (at least one of) the ANY privilege(s) on TABLE table_name for this operation.`
 5)invalid sql
   `return error like: Getting syntax error at line 1, column 15. Detail message: Unexpected input 'table_name', the most similar input is {'FROM'}.`
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [x] Yes, this PR will result in a change in behavior.
- [ ] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
[...]

PR #59503, opened 2025-05-30
[Feature] Prune tablets according to virtual bucket for dynamic tablet splitting and merging

xiangguangyxg:
## Why I'm doing:
https://github.com/StarRocks/starrocks/pull/59378 introduced virtual buckets metadata, which has not been used to prune tablets.
This pr uses virtual buckets metadata to prune tablets.

## What I'm doing:
Prune tablets according to virtual bucket for dynamic tablet splitting and merging.

This pr only focus on tablet pruning, and there are 2 more features that need to be adapted with virtual bucket:
- Colocate table
- Show data distribution

These feature will be adapted with virtual bucket in the subsequent pr.

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
@@ -102,7 +102,8 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
     @SerializedName(value = "rowCount")
     private long rowCount;
 
-    // Virtual buckets. There is a tablet id for each virtual bucket,
+    // Virtual buckets in order.
+    // There is a tablet id for each virtual bucket,
     // which means this virtual bucket's data is stored in this tablet.
     // We divide data into virtual buckets and then arrange these virtual buckets
     // into physical buckets, which are tablets.
@@ -113,7 +114,7 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
 
     private Map<Long, Tablet> idToTablets;
 
-    // Deprecated, virtual buckets keeps tablet order, keep this for compatibility
+    // Since virtual buckets keeps the order, this can be deprecated if idToTablets persists
     @SerializedName(value = "tablets")
     private List<Tablet> tablets;
 
@@ -195,6 +196,7 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
         return shardGroupId;
     }
 
+    // The virtual buckets are in order
     public List<Long> getVirtualBuckets() {
         return virtualBuckets;
     }
@@ -203,7 +205,8 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
         return tablets;
     }
 
-    public List<Long> getTabletIdsInOrder() {
+    // With virtual buckets, the order of tablets is irrelevant
+    public List<Long> getTabletIds() {
         List<Long> tabletIds = Lists.newArrayListWithCapacity(tablets.size());
         for (Tablet tablet : tablets) {
             tabletIds.add(tablet.getId());
@@ -307,6 +310,9 @@ public class MaterializedIndex extends MetaObject implements Writable, GsonPostP
         return virtualBucketIndexes;
     }
 
+    // With virtual buckets, the order index of tablets is irrelevant.
+    // Keep this method only for colocate table in shared-nothing mode,
+    // in which we do not implement tablet split and merge and virtual buckets are the same with tabletIds.
     public int getTabletOrderIdx(long tabletId) {
         int idx = 0;
         for (Tablet tablet : tablets) {
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/MetadataViewer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/MetadataViewer.java
@@ -357,7 +357,8 @@ public class MetadataViewer {
                 for (PhysicalPartition physicalPartition : partition.getSubPartitions()) {
                     long version = physicalPartition.getVisibleVersion();
                     for (MaterializedIndex index : physicalPartition.getMaterializedIndices(IndexExtState.VISIBLE)) {
-                        List<Long> tabletIds = index.getTabletIdsInOrder();
+                        // TODO(TackY): adapt the modification of bucket number from partition level to materialized index level
+                        List<Long> tabletIds = index.getTabletIds();
                         for (int i = 0; i < tabletIds.size(); i++) {
                             Tablet tablet = index.getTablet(tabletIds.get(i));
                             long rowCount = tablet.getRowCount(version);
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
@@ -2352,7 +2352,7 @@ public class OlapTable extends Table {
 
             short replicationNum = partitionInfo.getReplicationNum(partition.getId());
             MaterializedIndex baseIdx = physicalPartition.getBaseIndex();
-            for (Long tabletId : baseIdx.getTabletIdsInOrder()) {
+            for (Long tabletId : baseIdx.getTabletIds()) {
                 LocalTablet tablet = (LocalTablet) baseIdx.getTablet(tabletId);
                 List<Long> replicaBackendIds = tablet.getNormalReplicaBackendIds();
                 if (replicaBackendIds.size() < replicationNum) {
--- a/fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
@@ -66,20 +66,22 @@ import java.util.Set;
 public class HashDistributionPruner implements DistributionPruner {
     private static final Logger LOG = LogManager.getLogger(HashDistributionPruner.class);
 
-    // partition list, sort by the hash code
-    private List<Long> bucketsList;
-    // partition columns
-    private List<Column> distributionColumns;
-    // partition column filters
-    private Map<String, PartitionColumnFilter> distributionColumnFilters;
-    private int hashMod;
+    // virtual bucket list, sort by the hash code
+    private final List<Long> virtualBuckets;
+    // tablet id list (physical bucket list), has no order
+    private final List<Long> tabletIds;
+    // distribution columns
+    private final List<Column> distributionColumns;
+    // distribution column filters
+    private final Map<String, PartitionColumnFilter> distributionColumnFilters;
 
-    public HashDistributionPruner(List<Long> bucketsList, List<Column> columns,
-                                  Map<String, PartitionColumnFilter> filters, int hashMod) {
-        this.bucketsList = bucketsList;
+    // virtualBuckets must be in order, tabletIds has no order
+    public HashDistributionPruner(List<Long> virtualBuckets, List<Long> tabletIds,
+                                  List<Column> columns, Map<String, PartitionColumnFilter> filters) {
+        this.virtualBuckets = virtualBuckets;
+        this.tabletIds = tabletIds;
         this.distributionColumns = columns;
         this.distributionColumnFilters = filters;
-        this.hashMod = hashMod;
     }
 
     // columnId: which column to compute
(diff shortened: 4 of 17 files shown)


alvin-phoenix-ai:
@mergify rebase

fe/fe-core/src/main/java/com/starrocks/catalog/MetadataViewer.java
```diff
@@ -357,7 +357,8 @@ public static List<List<String>> getDataDistribution(
                 for (PhysicalPartition physicalPartition : partition.getSubPartitions()) {
                     long version = physicalPartition.getVisibleVersion();
                     for (MaterializedIndex index : physicalPartition.getMaterializedIndices(IndexExtState.VISIBLE)) {
-                        List<Long> tabletIds = index.getTabletIdsInOrder();
+                        // TODO: adapt the modification of bucket number from partition level to materialized index level
+                        List<Long> tabletIds = index.getVirtualBuckets();
```
alvin-phoenix-ai: The following code is working on a tablet id. I don't understand the reason to get virtual buckets here.
It seems to me that we need to get physical tablet IDs.

fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
```diff
@@ -2352,7 +2352,7 @@ public List<List<Long>> getArbitraryTabletBucketsSeq() throws DdlException {
 
             short replicationNum = partitionInfo.getReplicationNum(partition.getId());
             MaterializedIndex baseIdx = physicalPartition.getBaseIndex();
-            for (Long tabletId : baseIdx.getTabletIdsInOrder()) {
+            for (Long tabletId : baseIdx.getVirtualBuckets()) {
```
alvin-phoenix-ai: same here

fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
```diff
@@ -66,20 +66,21 @@
 public class HashDistributionPruner implements DistributionPruner {
     private static final Logger LOG = LogManager.getLogger(HashDistributionPruner.class);
 
-    // partition list, sort by the hash code
-    private List<Long> bucketsList;
+    // virtual partition list, sort by the hash code
+    private final List<Long> virtualBuckets;
+    // physical partition list
+    private final List<Long> physicalBuckets;
```
alvin-phoenix-ai: 1. it is tablet list? Isn't it? The comments seem to be mismatched.
2. Perhaps we can use tablet IDs, which can help mitigate some confusion here.

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
```diff
@@ -202,7 +202,8 @@ public List<Tablet> getTablets() {
         return tablets;
     }
 
-    public List<Long> getTabletIdsInOrder() {
+    // With virtual buckets, the order of tablets is irrelevant
+    public List<Long> getTabletIds() {
```
alvin-phoenix-ai, line 209: Do you know the reason why this is ordered?

fe/fe-core/src/main/java/com/starrocks/catalog/MetadataViewer.java
xiangguangyxg: This code here is for `show data distribution` introduced by  https://github.com/StarRocks/starrocks/pull/57819, which assumes that all materialized indexes in a partition have the same bucket number, and then merges the statistics of all materialized indexes with the same bucket sequence to show. Adaptation is needed here later because its assumption is no longer always true. Before adaptation, the code cannot be simply changed to get virtual buckets or get physical tablet IDs.  I used get virtual buckets here temporarily because it need the order, add a todo here for future modification.

fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
xiangguangyxg: Here is for another feature that needs to be adapted: colocate table. It will be adapted in the subsequent pr. Now the code cannot be simply changed to get virtual buckets or get physical tablet IDs. Here I use get virtual buckets temporarily because it need the order.

fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
xiangguangyxg: done

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
xiangguangyxg: Without virtual buckets, hash data distribution and bucket pruning need the order of tablets. 
With virtual buckets,  hash data distribution and bucket pruning use virtual buckets instead, the order of tablets should not be cared.

alvin-phoenix-ai: Why does hash data distribution and bucket pruning need the order of tablets?

fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
```diff
@@ -66,20 +66,21 @@
...
 
-    // partition list, sort by the hash code
-    private List<Long> bucketsList;
-    // partition columns
-    private List<Column> distributionColumns;
-    // partition column filters
-    private Map<String, PartitionColumnFilter> distributionColumnFilters;
-    private int hashMod;
+    // virtual bucket list, sort by the hash code
+    private final List<Long> virtualBuckets;
```
alvin-phoenix-ai, line 70: Oh? If the buckets should be ordered, I think we need to write the requirements as a comment on the constructor function, so that callers are aware of this requirement. Otherwise, they may use this class with an unexpected result.

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
alvin-phoenix-ai: And if virtualBuckets are required to be in an order, you should write down the implicit requirement for virtualBuckets and the getVirtualBuckets function.

fe/fe-core/src/main/java/com/starrocks/catalog/MetadataViewer.java
```diff
@@ -357,7 +357,8 @@ public static List<List<String>> getDataDistribution(
                 for (PhysicalPartition physicalPartition : partition.getSubPartitions()) {
                     long version = physicalPartition.getVisibleVersion();
                     for (MaterializedIndex index : physicalPartition.getMaterializedIndices(IndexExtState.VISIBLE)) {
-                        List<Long> tabletIds = index.getTabletIdsInOrder();
+                        // TODO: adapt the modification of bucket number from partition level to materialized index level
```
alvin-phoenix-ai:
```suggestion
                        // TODO(TachY): adapt the modification of bucket number from partition level to materialized index level
```
In this way, you can get all the TODOs you left in the code

fe/fe-core/src/main/java/com/starrocks/catalog/MetadataViewer.java
alvin-phoenix-ai: I don't catch up with you 100%. However, this PR handles the distribution prune. This part doesn't relate to the distribution prune, I suggest you leave it unchanged in this PR and change it in a following PR.

fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
alvin-phoenix-ai: Same here
You should change it with the following pull request rather than making the change in this pull request.

fe/fe-core/src/main/java/com/starrocks/catalog/MetadataViewer.java
xiangguangyxg: changed to getTabletIds

fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
xiangguangyxg: changed to getTabletIds

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndex.java
xiangguangyxg: added some comments for virtualBuckets and the getVirtualBuckets function

fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
xiangguangyxg: added some comments

fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
```diff
@@ -747,17 +747,17 @@ private void computeTabletInfo() throws StarRocksException {
...
                     scanTabletIds.addAll(tabletIds);
                 } else {
                     tablets.addAll(selectedIndex.getTablets());
-                    scanTabletIds.addAll(allTabletIds);
+                    scanTabletIds.addAll(selectedIndex.getTabletIds());
                 }
 
+                List<Long> allTabletIds = selectedIndex.getVirtualBuckets();
                 for (int i = 0; i < allTabletIds.size(); i++) {
                     tabletId2BucketSeq.put(allTabletIds.get(i), i);
```
wyb: one tablet may correspond to multiple bucket seqs

fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
```diff
@@ -747,17 +747,17 @@ private void computeTabletInfo() throws StarRocksException {
...
                         tablets.add(selectedIndex.getTablet(id));
                     }
                     scanTabletIds.addAll(tabletIds);
                 } else {
                     tablets.addAll(selectedIndex.getTablets());
-                    scanTabletIds.addAll(allTabletIds);
+                    scanTabletIds.addAll(selectedIndex.getTabletIds());
                 }
 
+                List<Long> allTabletIds = selectedIndex.getVirtualBuckets();
```
srlch: Why we need VirtualBuckets instead of physical one here?

fe/fe-core/src/main/java/com/starrocks/sql/plan/PlanFragmentBuilder.java
```diff
@@ -926,16 +926,16 @@ public PlanFragment visitPhysicalOlapScan(OptExpression optExpr, ExecPlan contex
                         selectedNonEmptyPartitionIds.add(partitionId);
                         Map<Long, Integer> tabletId2BucketSeq = Maps.newHashMap();
                         Preconditions.checkState(selectTabletIds != null && !selectTabletIds.isEmpty());
-                        final MaterializedIndex selectedTable = physicalPartition.getIndex(selectedIndexId);
-                        List<Long> allTabletIds = selectedTable.getTabletIdsInOrder();
-                        totalTabletsNum += selectedTable.getTablets().size();
+                        final MaterializedIndex selectedIndex = physicalPartition.getIndex(selectedIndexId);
+                        totalTabletsNum += selectedIndex.getTablets().size();
+                        List<Long> allTabletIds = selectedIndex.getVirtualBuckets();
```
srlch: same above

fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
xiangguangyxg: changed it back

fe/fe-core/src/main/java/com/starrocks/sql/plan/PlanFragmentBuilder.java
xiangguangyxg: changed it back

wyb (approved):

stephen-shelby (approved):

gengjun-git (approved):

PR #59510, opened 2025-05-30
[BugFix] fix LambdaFunction's isConst always return false

before-Sunrise:
## Why I'm doing:
 lambda exp like array_map's isConst() always return false, but actually lambda expr's isConst should rely only on childs except first one: LambdaFunction which always return false right now because it has fake ColumnRef

## What I'm doing:
LambdaFunction's isConst can return ture.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/be/src/exprs/lambda_function.h
+++ b/be/src/exprs/lambda_function.h
@@ -64,7 +64,7 @@ public:
 
     bool is_lambda_function() const override { return true; }
     bool is_lambda_expr_independent() const { return _is_lambda_expr_independent; }
-    bool is_constant() const {return true;}
+    bool is_constant() const { return true; }
 
     Expr* get_lambda_expr() const { return _children[0]; }
     std::string debug_string() const override;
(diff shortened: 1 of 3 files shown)


be/src/exprs/lambda_function.h
```diff
@@ -64,6 +64,7 @@ class LambdaFunction final : public Expr {
 
     bool is_lambda_function() const override { return true; }
     bool is_lambda_expr_independent() const { return _is_lambda_expr_independent; }
+    bool is_constant() const { return true; }
```
alvin-celerdata, line 67:
```suggestion
    bool is_constant() const override { return true; }
```

alvin-phoenix-ai (approved):

silverbullet233 (approved):

satanson (approved):

PR #59562, opened 2025-06-03
[Refactor] Rename enable_partiton_aggregate to file_bundling

sevev:
## Why I'm doing:
We support `enable_partition_aggregate` in shared-data mode to reduce api calls, but the property name is not accurate enough. The new feature merge data file, txn log file and metadata file, so change the property name to `file_bundling`.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1


gensrc/proto/lake_service.proto
```diff
@@ -296,7 +296,7 @@ message VacuumRequest {
     // The tablet infos to be vacuumed.
     repeated TabletInfoPB tablet_infos = 7;
     // Whether enable partition aggregation
-    optional bool enable_partition_aggregation = 8;
+    optional bool enable_io_merge = 8;
```
alvin-phoenix-ai: Should change the comment accordingly.

alvin-phoenix-ai (commented):
IMO, io_merge is so general. I take it as merging local disk IO operations intuitively. 
And io_merge is often used to try to merge READ IO operations and issue them into one.

Image that now this feature is released, and we talk to one customer about io_merge. I'm not sure that what they perceive is what we want them to.

As far as I know this feature is more related to the way of generating files, for example, try to write a bigger file rather than multi small files. 
I would recommend `enable_file_bundling`, or introduce a new format like `enable_table_format_v2`, but again this is just an example.
I suggest that you should think about the name more from the users' perspective.

sevev:
> enable_file_bundling

OK, I'll discuss with other team members on how to name it.

sevev:
> IMO, io_merge is so general. I take it as merging local disk IO operations intuitively. And io_merge is often used to try to merge READ IO operations and issue them into one.
> 
> Image that now this feature is released, and we talk to one customer about io_merge. I'm not sure that what they perceive is what we want them to.
> 
> As far as I know this feature is more related to the way of generating files, for example, try to write a bigger file rather than multi small files. I would recommend `enable_file_bundling`, or introduce a new format like `enable_table_format_v2`, but again this is just an example. I suggest that you should think about the name more from the users' perspective.

I think `enable_table_format_v2` is not suitable for this feature. This is more like adopting a new storage format to store data (e.g., converting from row-based to column-based storage), but this feature doesn't make such a major change. After discussion, we believe `file_bundling` is a better choice.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3243,7 +3243,7 @@ public class Config extends ConfigBase {
     public static boolean enable_fast_schema_evolution_in_share_data_mode = true;
 
     @ConfField(mutable = true)
-    public static boolean enable_partition_aggregation = false;
+    public static boolean lake_enable_file_bundling = false;
```
alvin-phoenix-ai: I saw that in the FE configuration, it is `lake_enable_file_bundling`, but when creating tables, it is `file_bundling`.
What is the reason behind that?

Configs that start with `lake_` is a legacy, I don't think we should still follow it now. Because it will cause confusion in our customers, who, in their mind, lake stands for Open Lake like Iceberg.
I think `enable_file_bundling` is sufficient. If you want to make it more specific in shared-data mode, maybe you can consider `enable_file_bundling_in_shared_data`.

sevev: > I saw that in the FE configuration, it is `lake_enable_file_bundling`, but when creating tables, it is `file_bundling`. What is the reason behind that?
> 
> Configs that start with `lake_` is a legacy, I don't think we should still follow it now. Because it will cause confusion in our customers, who, in their mind, lake stands for Open Lake like Iceberg. I think `enable_file_bundling` is sufficient. If you want to make it more specific in shared-data mode, maybe you can consider `enable_file_bundling_in_shared_data`.

1. I reviewed the naming conventions in our table properties and found that most properties do not use the `enable_` prefix to denote a status. Additionally, when executing `ALTER` to modify properties, our code convention uses` enable_xx` to represent the action of modifying a property. Therefore, I have chosen `file_bundling` as the display item for the property.
However, in our configuration code, it is a convention to use the `enable_` prefix to indicate the activation or deactivation of a feature or optimization. Therefore, I added `enable_` to the configuration to maintain consistency with the overall code style.
2. I will remove `lake_` prefix.

luohaha (approved):

HangyuanLiu (approved):

stephen-shelby (approved):

Issue #58963, opened 2025-05-14
[Feature] Identifying user session in audit logs


dolfinus:
- [x] https://github.com/StarRocks/starrocks/pull/59754

## Feature request

**Is your feature request related to a problem? Please describe.**

Currently [audit log](https://docs.starrocks.io/docs/administration/management/audit_loader/) contains information like client IP, user name, statement, status and so on. But it is unknown if a set of queries are a part of some specific session/connection, or are executed independently.

Inspection of audit logs can give information about specific user and query which was executed by StarRocks. But it don't give any information about application or script which executed this query, so this becomes hard to track down falling query source.

Other databases like Postgres allow to track each session specifically:
* There is a [pg_stat_activity](https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW) view which returns state of currently executed sessions in the system.
* Sessions have a name which is set using [ApplicationName](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-APPLICATION-NAME) client parameter. So session can be tracked by explicit name given by a user, and not by some random ID (which user doesn't know or log on client side in most cases).
* Audit extensions (e.g. pg_proaudit) include [session identifier column](https://postgrespro.ru/docs/postgrespro/14/pg-proaudit?lang=en#PG-PROAUDIT-VIEWING-SECURITY-EVENT-LOG)

Finally, the combination of session_id + timestamp + query_id can tell everything which was done by specific session, and in which order.

But currently I don't see how this can be implemented in StarRocks:
* There is no session/application/connection state concept. I don't see any views which can return this information. There is a [SHOW PROCESSLIST](https://docs.starrocks.io/docs/sql-reference/sql-statements/cluster-management/nodes_processes/SHOW_PROCESSLIST/) command which can return connectionId, user and command type, but it returns only connections executing queries just now, while audit log can contain information about already finished queries, so currently there is no way to get connection info by finished queryId.
* There is no client/protocol option which accepts user-specified session/connection name, and saves it into some database object/view which can be used for introspection or audit.
* There is no sessionId/connectionId column in audit event, so queries cannot be grouped by connection, they all seems to be totally independent.

**Describe the solution you'd like**

Please consider implementing:
* connection/session name can be explicitly set by user
* audit logs should contain connection/session id and name

**Describe alternatives you've considered**

**Additional context**

[...]

PR #59563, opened 2025-06-03
[Enhancement] Add sessionId to audit log

dolfinus:
## Why I'm doing:

Currently audit log has queryId, statementId and so on. But there is no information if queries were created by the same session or different ones.

## What I'm doing:

Partially solves #58963

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [X] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [X] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/plugin/AuditEvent.java
+++ b/fe/fe-core/src/main/java/com/starrocks/plugin/AuditEvent.java
@@ -151,6 +151,9 @@ public class AuditEvent {
     @AuditField(value = "QueryFEAllocatedMemory")
     public long queryFeMemory = 0;
 
+    @AuditField(value = "SessionId")
+    public String sessionId = "";
+
     public static class AuditEventBuilder {
 
         private AuditEvent auditEvent = new AuditEvent();
@@ -386,6 +389,11 @@ public class AuditEvent {
 
         public AuditEventBuilder setQueryFeMemory(long queryFeMemory) {
             auditEvent.queryFeMemory = queryFeMemory;
+            return this;
+        }
+
+        public AuditEventBuilder setSessionId(String sessionId) {
+            auditEvent.sessionId = sessionId;
             return this;
         }
 
--- a/fe/fe-core/src/main/java/com/starrocks/qe/ConnectProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/ConnectProcessor.java
@@ -227,7 +227,8 @@ public class ConnectProcessor {
                 .setReturnRows(ctx.getReturnRows())
                 .setStmtId(ctx.getStmtId())
                 .setIsForwardToLeader(isForwardToLeader)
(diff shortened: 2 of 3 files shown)


docs/en/administration/management/audit_loader.md
```diff
@@ -26,6 +26,7 @@ CREATE TABLE starrocks_audit_db__.starrocks_audit_tbl__ (
   `queryId`           VARCHAR(64)                COMMENT "Unique query ID",
   `timestamp`         DATETIME         NOT NULL  COMMENT "Query start time",
   `queryType`         VARCHAR(12)                COMMENT "Query type (query, slow_query, connection）",
+  `sessionId`         VARCHAR(16)                COMMENT "Session Id",
```
alvin-phoenix-ai: As the CAUTION above states, modifying this change will cause incompatibility issues between different StarRocks versions.

dolfinus: If audit log with 30 fields will be inserted to a table containing 25 column, wouldn't it fail?

If not, can user with new AuditEvent class fields just add new column with name `sessionId`, and it will be automatically filled up with expected value?

dolfinus: Also, almost all columns here are NULL, so wouldn't default value for missing columns be NULL?

alvin-phoenix-ai: I don't understand how the audit loader works internally.
So, it is @gengjun-git who is best suited to check how to make it work.

dolfinus: I may revert this part, and create a dedicated issue

dolfinus: Reverted

alvin-phoenix-ai (approved):

nshangyiming (approved):

jaogoy:
@dolfinus Where is the code to print the session id into Audit Log.

dolfinus:
@jaogoy If I understand correctly, AuditLogBuilder uses Java reflection to get all fields of AuditEvent annotated with `@AuditField`, and then add them to log (either JSON or plaintext):
https://github.com/StarRocks/starrocks/blob/30c79c94241044019fc6137d7c03d0a02e786ddd/fe/fe-core/src/main/java/com/starrocks/qe/AuditLogBuilder.java#L88-L137

Only few fields (e.g. BigQuery-related or `features`) are handled in a special way.

docs/en/administration/management/audit_loader.md
dolfinus: Created #59790

dolfinus:
Should I backport this to 3.5.0 release candidate?

PR #59651, opened 2025-06-05
[Feature] Create table supports decimal256 column type

stephen-shelby:
## What I'm doing:
support create table with decimal precision in the range (38, 76]. 
eg:
```
CREATE TABLE test_decimal (k1 int, k2 int, k3 int , d1 decimal(50, 24));
```
now, we can support decimal's precision in the range [1, 76]. decimal's scale in the range [0, 76]. 
Also, this patch removes some unused code.

Fixes #issue
https://github.com/StarRocks/starrocks/issues/59645

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/PrimitiveType.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/PrimitiveType.java
@@ -76,6 +76,7 @@ public enum PrimitiveType {
     DECIMAL32("DECIMAL32", 4, TPrimitiveType.DECIMAL32),
     DECIMAL64("DECIMAL64", 8, TPrimitiveType.DECIMAL64),
     DECIMAL128("DECIMAL128", 16, TPrimitiveType.DECIMAL128),
+    DECIMAL256("DECIMAL256", 32, TPrimitiveType.DECIMAL256),
 
     JSON("JSON", 16, TPrimitiveType.JSON),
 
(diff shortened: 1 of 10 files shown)


gensrc/thrift/Types.thrift
```diff
@@ -98,6 +98,7 @@ enum TPrimitiveType {
   DECIMAL32,
   DECIMAL64,
   DECIMAL128,
+  DECIMAL256,
```
alvin-phoenix-ai: This will break the compatibility when the cluster is under rolling upgrade. All new items should be added to the end. Sorry I need to give a change-request because of it.

fe/fe-core/src/main/java/com/starrocks/catalog/PrimitiveType.java
```diff
@@ -295,10 +279,6 @@ public static List<TPrimitiveType> toThrift(PrimitiveType[] types) {
         return result;
     }
 
-    public static int getMaxSlotSize() {
```
alvin-phoenix-ai: Can you make all these refactors in a separate PR?

fe/fe-core/src/main/java/com/starrocks/catalog/PrimitiveType.java
```diff
@@ -347,6 +329,8 @@ public static int getDefaultScaleOfDecimal(PrimitiveType t) {
                 return 18;
             case DECIMAL128:
                 return 38;
+            case DECIMAL256:
+                return 76;
```
alvin-phoenix-ai, line 369: The magic number should be a static definition rather than a hand-coded number.

alvin-phoenix-ai (changes requested):

alvin-phoenix-ai (dismissed):

kangkaisen (approved):

alvin-phoenix-ai (approved):

PR #36837, opened 2023-12-12
[Enhancement] Thansform 'json_extract' to 'get_json_string' for trino parser

lipenglin:
Why I'm doing:

**badcase**:
- starrocks: `json_query(json_object_expr, json_path)` returns a JSON value.
But
- trino: `json_extract(json, json_path)` return a JSON string
```sql
SELECT
  json_extract(c1, '$.address') AS addr,
  sum(c2),
  count(*) AS cnt
FROM
  (
    SELECT
      '{"id":1,"name":"John","address":{"city":"New York","zip":"10001"},"contacts":[{"type":"email","value":"john@example.com"},{"type":"phone","value":"123-456-7890"}]}' AS c1,
      1 AS c2
    UNION
    ALL
    SELECT
      '{"id":2,"name":"John","address":{"city":"New York","zip":"10002"},"contacts":[{"type":"email","value":"john@example.com"},{"type":"phone","value":"123-456-7890"}]}' AS c1,
      2 AS c2
    UNION
    ALL
    SELECT
      '{"id":2,"name":"John","address":{"city":"New York","zip":"10002"},"contacts":[{"type":"email","value":"john@example.com"},{"type":"phone","value":"123-456-7890"}]}' AS c1,
      3 AS c2
  ) AS tbl
GROUP BY
  addr;
---ERROR 1064 (HY000): Getting analyzing error. Detail message: Type (nested) percentile/hll/bitmap/json not support group-by.
SELECT
  json_extract(c1, '$.id') AS id,
  sum(c2),
  count(*) AS cnt
FROM
  (
    SELECT
      '{"id":1,"name":"John","address":{"city":"New York","zip":"10001"},"contacts":[{"type":"email","value":"john@example.com"},{"type":"phone","value":"123-456-7890"}]}' AS c1,
      1 AS c2
    UNION
    ALL
    SELECT
      '{"id":2,"name":"John","address":{"city":"New York","zip":"10002"},"contacts":[{"type":"email","value":"john@example.com"},{"type":"phone","value":"123-456-7890"}]}' AS c1,
      2 AS c2
    UNION
    ALL
    SELECT
      '{"id":2,"name":"John","address":{"city":"New York","zip":"10002"},"contacts":[{"type":"email","value":"john@example.com"},{"type":"phone","value":"123-456-7890"}]}' AS c1,
      3 AS c2
  ) AS tbl
GROUP BY
  id;
[...]

PR #59718, opened 2025-06-06
[Enhancement] Make json_extract return json type in trino dialect

duanyyyyyyy:
[Enhancement] Make json_extract return json type in trino dialect
## Why I'm doing:
In trino the json_extract function will return json type.
But in SR trino dialect it will return string.
In some cases it will lead query result mismatch.
image
For example:
   select 
                    cast(json_extract(t.ext,'$.group_id') as int) as group_id
                    , null as merchant_id_joined
                    ,*
                from food.foody_promotion_db__promotion_tab__reg_daily_s0_live
                cross join unnest (cast(json_extract(extra_data,'$.merchant_condition.apply_merchant_groups') as ARRAY<JSON>)) as t(ext)

It may have some issue that will return wrong result.
The SQL
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/parser/trino/Trino2SRFunctionCallTransformer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/parser/trino/Trino2SRFunctionCallTransformer.java
@@ -320,8 +320,8 @@ public class Trino2SRFunctionCallTransformer {
         registerFunctionTransformer("json_parse", 1, "parse_json",
                 List.of(Expr.class));
 
-        // json_extract -> get_json_string
-        registerFunctionTransformer("json_extract", 2, "get_json_string",
+        // json_extract -> json_query
+        registerFunctionTransformer("json_extract", 2, "json_query",
                 List.of(Expr.class, Expr.class));
 
         // json_size -> json_length
--- a/fe/fe-core/src/test/java/com/starrocks/connector/parser/trino/TrinoFunctionTransformTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/connector/parser/trino/TrinoFunctionTransformTest.java
@@ -339,10 +339,10 @@ public class TrinoFunctionTransformTest extends TrinoTestBase {
         assertPlanContains(sql, "parse_json('{\"a\": {\"b\": 1}}')");
 
         sql = "select json_extract(json_parse('{\"a\": {\"b\": 1}}'), '$.a.b')";
-        assertPlanContains(sql, "get_json_string(parse_json('{\"a\": {\"b\": 1}}'), '$.a.b')");
+        assertPlanContains(sql, "json_query(parse_json('{\"a\": {\"b\": 1}}'), '$.a.b')");
 
         sql = "select json_extract(JSON '{\"a\": {\"b\": 1}}', '$.a.b');";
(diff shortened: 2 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/connector/parser/trino/Trino2SRFunctionCallTransformer.java
```diff
@@ -321,8 +321,15 @@ private static void registerJsonFunctionTransformer() {
                 List.of(Expr.class));
 
         // json_extract -> get_json_string
-        registerFunctionTransformer("json_extract", 2, "get_json_string",
-                List.of(Expr.class, Expr.class));
+        registerFunctionTransformer("json_extract", 2,
+                new FunctionCallExpr("parse_json",
```
Youngwb: It looks like we can use **json_query** directly

duanyyyyyyy: It looks that the first version is using json_query, but someone change it to get_json_string. Maybe in that version the json_extract in trino will return STRING type.
And I also checked the latest Trino documentation. Currently they will return JSON type, do we need to follow Trino current behaivor?

duanyyyyyyy: I changed the codebase.
If community want to keep the behavior with Trino can merge.
Otherwise can just close the ticket....

duanyyyyyyy: This is the documentation
https://trino.io/docs/current/functions/json.html#json_extract

Youngwb: OK，you can refer to https://github.com/StarRocks/starrocks/pull/36837, we use get_json_string beacuse of SR can not group by json type

duanyyyyyyy: I reviewed the issue carefully, and I still think we need to keep it return JSON type as wrong result is more harmful compare to return error.

duanyyyyyyy: Any feedback?
For example, if the group by column is JSON type the query will be failed, and the error message will be 
`Getting analyzing error. Detail message: Type (nested) percentile/hll/bitmap/json not support group-by.`
But for my case, when we cast the String to ARRAY<JSON> it will return wrong result without any error message.

Youngwb (approved):

dirtysalt (approved):

Youngwb:
@mergify backport branch-4.0

Youngwb:
@mergify backport branch-3.5

murphyatwork:
@Mergifyio backport branch-3.4

PR #59754, opened 2025-06-09
[Feature] Introduce customSessionName

### LINKED ISSUE #58963: [Feature] Identifying user session in audit logs

dolfinus:
## Why I'm doing:

In #59563 I've added `sessionId` field to AuditEvent, so audit plugins can understand that some queries were executed within the same session. But there is no session name concept, which may be beneficial for some integrations to track user-defined session name, like `applicatioName` in Postgres.

## What I'm doing:

Introduce `custom_session_name` session variable which can be set like this:

```sql
SET SESSION custom_session_name = 'my_session';
```

It works similar to `custom_query_id`, but `custom_session_name` doesn't reset after query is finished. For now, this value is added only to AuditEvent, but it also can be added to various logs within StarRocks.

This partially solves #58963.

## What type of PR is this:

- [ ] BugFix
- [X] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [X] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/docs/en/sql-reference/System_variable.md
+++ b/docs/en/sql-reference/System_variable.md
@@ -408,6 +408,13 @@ Used for MySQL client compatibility. No practical usage.
 * **Data type**: String
 * **Introduced in**: v3.4.0
 
+### custom_session_name (session)
+
+* **Description**: Used to specify custom name of current session, analog of `applicationName` or `program_name` in DMBS like MySQL or PostgreSQL. Can be set using `SET SESSION custom_session_name = 'my session name';`. Value can be found in audit logs in `customSessionName` field.
+* **Default**: ""
+* **Data type**: String
+* **Introduced in**: v4.0.0
+
 ### datacache_sharing_work_period
 
 * **Description**: The period of time that Cache Sharing takes effect. After each cluster scaling operation, only the requests within this period of time will try to access the cache data from other nodes if the Cache Sharing feature is enabled.
--- a/docs/en/sql-reference/System_variable.md
+++ b/docs/en/sql-reference/System_variable.md
@@ -413,7 +413,7 @@ Used for MySQL client compatibility. No practical usage.
 * **Description**: Used to specify custom name of current session, analog of `applicationName` or `program_name` in DMBS like MySQL or PostgreSQL. Can be set using `SET SESSION custom_session_name = 'my session name';`. Value can be found in audit logs in `customSessionName` field.
 * **Default**: ""
 * **Data type**: String
-* **Introduced in**: v4.0.0
+* **Introduced in**: v4.1.0
 
 ### datacache_sharing_work_period
 
--- a/fe/fe-core/src/main/java/com/starrocks/qe/AuditLogBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/AuditLogBuilder.java
@@ -70,19 +70,20 @@ public class AuditLogBuilder extends Plugin implements AuditPlugin {
                 if (af == null) {
                     continue;
                 }
+                String fieldName = af.value();
 
                 // fields related to big queries are not written into audit log by default,
                 // they will be written into big query log.
-                if (af.value().equals("BigQueryLogCPUSecondThreshold") ||
-                        af.value().equals("BigQueryLogScanBytesThreshold") ||
-                        af.value().equals("BigQueryLogScanRowsThreshold")) {
+                if (fieldName.equals("BigQueryLogCPUSecondThreshold") ||
+                        fieldName.equals("BigQueryLogScanBytesThreshold") ||
+                        fieldName.equals("BigQueryLogScanRowsThreshold")) {
                     continue;
                 }
-                if (af.value().equalsIgnoreCase("features")) {
+                if (fieldName.equalsIgnoreCase("features")) {
                     continue;
                 }
 
-                if (af.value().equals("Time")) {
+                if (fieldName.equals("Time")) {
                     queryTime = (long) f.get(event);
                 }
 
@@ -117,9 +118,15 @@ public class AuditLogBuilder extends Plugin implements AuditPlugin {
                 }
 
                 if (Config.audit_log_json_format) {
-                    logMap.put(af.value(), value);
+                    logMap.put(fieldName, value);
                 } else {
-                    sb.append("|").append(af.value()).append("=").append(value);
+                    // Prevent inserting strings which may break log format
+                    String strValue = value == null ? "null" : value.toString();
+                    strValue = strValue.replace("\n", "\\n")
+                                    .replace("\r", "\\r")
+                                    .replace("|", "")
+                                    .replace("=", "");
+                    sb.append("|").append(fieldName).append("=").append(strValue);
                 }
             }
 
--- a/fe/fe-core/src/main/java/com/starrocks/qe/AuditLogBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/AuditLogBuilder.java
@@ -120,13 +120,7 @@ public class AuditLogBuilder extends Plugin implements AuditPlugin {
                 if (Config.audit_log_json_format) {
                     logMap.put(fieldName, value);
                 } else {
-                    // Prevent inserting strings which may break log format
-                    String strValue = value == null ? "null" : value.toString();
-                    strValue = strValue.replace("\n", "\\n")
-                                    .replace("\r", "\\r")
-                                    .replace("|", "")
-                                    .replace("=", "");
-                    sb.append("|").append(fieldName).append("=").append(strValue);
+                    sb.append("|").append(af.value()).append("=").append(value);
                 }
             }
 
--- a/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
@@ -734,6 +734,9 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
     public static final String OPTIMIZE_DISTINCT_AGG_OVER_FRAMED_WINDOW =
             "optimize_distinct_agg_over_framed_window";
 
+    public static final String CUSTOM_SESSION_NAME = "custom_session_name";
+    public static final int CUSTOM_SESSION_NAME_MAX_LENGTH = 64;
+
     // Flag to control whether to proxy follower's query statement to leader/follower.
     public enum FollowerQueryForwardMode {
         DEFAULT,    // proxy queries by the follower's replay progress (default)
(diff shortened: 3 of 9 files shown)


fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
```diff
@@ -613,6 +613,8 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
 
     public static final String ENABLE_PIPELINE_EVENT_SCHEDULER = "enable_pipeline_event_scheduler";
 
+    public static final String CUSTOM_SESSION_NAME = "custom_session_name";
```
alvin-phoenix-ai, line 737: Could you change the document accordingly to let users know how to use it?

alvin-phoenix-ai: 1. A new session variable is added.
2. A new field is added into audit log.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
```diff
@@ -4811,6 +4816,14 @@ public void setCustomQueryId(String customQueryId) {
         this.customQueryId = customQueryId;
     }
 
+    public String getCustomSessionName() {
+        return customSessionName;
+    }
+
+    public void setCustomSessionName(String customSessionName) {
+        this.customSessionName = customSessionName;
+    }
```
alvin-phoenix-ai, line 5553: It is better to do some sanitization verification on the input session name.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
dolfinus: I thought about dedicated PR for documentation, as `custom_query_id` is not documented as well

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
dolfinus: What should it look like? `custom_query_id` doesn't have any, so I can't use it as an example. Should it be limited by length, or allowed characters?

alvin-phoenix-ai: Yeah, we didn't do well with customer_query_id.
I think `COMMON_NAME_REGEX` in FeNameFormat.java is sufficient to use with the session name.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
alvin-phoenix-ai: Yeah, something that we need to improve.
But I'd prefer to have the document with the same PR.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
dolfinus: As I see, `COMMON_NAME_REGEX` requires value to not have any spaces or characters other than alphanumeric. This is okay for user-defined variable _names_, but why prohibit _values_ like `my-session`, `my session`, `my/project/abc`?

dolfinus: IMHO it's better to truncate input value, like max 64 characters, but not validate name against some regexp, as this is an externally set up name, which can contain values provided by third-party system.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
dolfinus: Added an entry to System_variable.md. I don't know next release version number, so I've set 3.5.1 here

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
alvin-phoenix-ai: My concern comes from the fact that we are outputting our log in the format like "[custom_session_name=xxx]". If we accept any value, it may cause the parser to fail, for example: [custom_session_name=xx[y]]. 
This is why I think we should establish a rule for the value.
However, for what kind of rule to apply to the value, I don't think too much. My thinking is that it is easier to start from a strict way. And it is easier to change the rule from more strict to less strict.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
alvin-phoenix-ai: We don't want to introduce a new feature in the released version. 
This feature will be ready in our next minor version, maybe 4.0.0

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
dolfinus: Instead of failing with assertion, we can remove symbols which may interfere with audit log format (`|=\n\r`). Maybe only for delimited format, but keep original value in JSON log. What do you think?

alvin-phoenix-ai: Looks good to me

dolfinus: Updated audit log builder to remove `|`, `=` and escape `\r`, `\n`. But I don't see any tests for log entry format, should I add them?

I can move this change to a dedicated PR.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
dolfinus: Ok, set 4.0.0 here

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
alvin-phoenix-ai: Better to change the audit log part in a separate PR to make it easy to cherry-pick in case of need in the future.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
```diff
@@ -613,6 +613,9 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
 
     public static final String ENABLE_PIPELINE_EVENT_SCHEDULER = "enable_pipeline_event_scheduler";
 
+    public static final String CUSTOM_SESSION_NAME = "custom_session_name";
+    public static final int CUSTOM_SESSION_NAME_MAX_LENGTH = 64;
```
alvin-phoenix-ai, line 738: You can implement a VariableVarConverterI to check if the input value of `custom_session_name` is valid.

fe/fe-core/src/main/java/com/starrocks/qe/AuditLogBuilder.java
```diff
@@ -131,9 +132,15 @@ public void exec(AuditEvent event) {
                 }
 
                 if (Config.audit_log_json_format) {
-                    logMap.put(af.value(), value);
+                    logMap.put(fieldName, value);
                 } else {
-                    sb.append("|").append(af.value()).append("=").append(value);
+                    // Prevent inserting strings which may break log format
```
alvin-phoenix-ai: As mentioned in the comment, you can make the change with the corresponding test in a separate PR.

StarRocks-Reviewer:
@cursor review

StarRocks-Reviewer:
@cursor review

docs/en/sql-reference/System_variable.md
```diff
@@ -408,6 +408,13 @@ Used for MySQL client compatibility. No practical usage.
 * **Data type**: String
 * **Introduced in**: v3.4.0
 
+### custom_session_name (session)
+
+* **Description**: Used to specify custom name of current session, analog of `applicationName` or `program_name` in DMBS like MySQL or PostgreSQL. Can be set using `SET SESSION custom_session_name = 'my session name';`. Value can be found in audit logs in `customSessionName` field.
+* **Default**: ""
+* **Data type**: String
+* **Introduced in**: v4.0.0
```
dolfinus: 4.0.0 is already released, should I change this line to 4.0.5?

stdpain: this feature will be supported since 4.1

dolfinus: Ok, bumped version to 4.1.0 here

stdpain (approved):

alvin-phoenix-ai (approved):

stdpain:

PR #59758, opened 2025-06-09
[Enhancement] Remove alias check in subquery

duanyyyyyyy:
[Enhancement] Remove alias check in subquery
## Why I'm doing:
We can set alias in the subquery relation to make it success when we forget to set alias.
For example in some view, especially hive view.
Users can set SQL like:
select * from (select * from a union all select * from b)
for the view logic.
And here we can add the alias for it, as the alias can be generated by ourselves.
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
@@ -990,7 +990,8 @@ public class QueryAnalyzer {
         @Override
         public Scope visitSubqueryRelation(SubqueryRelation subquery, Scope context) {
             if (subquery.getResolveTableName() != null && subquery.getResolveTableName().getTbl() == null) {
-                ErrorReport.reportSemanticException(ErrorCode.ERR_DERIVED_MUST_HAVE_ALIAS);
+                String catalog = ConnectContext.get().getCurrentCatalog();
+                subquery.setAlias(new TableName(catalog, "alias_db", "alias_" + System.nanoTime()));
             }
 
             Scope queryOutputScope = process(subquery.getQueryStatement(), context);
--- a/fe/fe-core/src/test/java/com/starrocks/sql/analyzer/AnalyzeJoinTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/sql/analyzer/AnalyzeJoinTest.java
@@ -26,6 +26,7 @@ import org.junit.Test;
 
 import static com.starrocks.sql.analyzer.AnalyzeTestUtil.analyzeFail;
 import static com.starrocks.sql.analyzer.AnalyzeTestUtil.analyzeSuccess;
+import static com.starrocks.sql.analyzer.AnalyzeTestUtil.analyzeWithoutTestView;
 
 public class AnalyzeJoinTest {
 
@@ -197,7 +198,7 @@ public class AnalyzeJoinTest {
         analyzeFail(sql, "Not unique table/alias: 'a'");
 
         sql = "select * from (t0 a, (select * from t1))";
-        analyzeFail(sql, "Every derived table must have its own alias");
+        analyzeWithoutTestView(sql);
 
         sql = "select * from t0, t0";
         analyzeFail(sql, "Not unique table/alias: 't0'");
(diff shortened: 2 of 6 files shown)


duanyyyyyyy:
@wanpengfei-git I wonder why the CI stoped?

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
```diff
@@ -990,7 +990,8 @@ private void analyzeJoinHints(JoinRelation join) {
         @Override
         public Scope visitSubqueryRelation(SubqueryRelation subquery, Scope context) {
             if (subquery.getResolveTableName() != null && subquery.getResolveTableName().getTbl() == null) {
-                ErrorReport.reportSemanticException(ErrorCode.ERR_DERIVED_MUST_HAVE_ALIAS);
+                String catalog = ConnectContext.get().getCurrentCatalog();
+                subquery.setAlias(new TableName(catalog, "alias_db", "alias_" + System.nanoTime()));
```
alvin-phoenix-ai, line 994: It is a little tricky.
Could you explain more details in the commit message about how it will work by changing in this way?

duanyyyyyyy: Changed

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
```diff
@@ -990,7 +990,8 @@ private void analyzeJoinHints(JoinRelation join) {
         @Override
         public Scope visitSubqueryRelation(SubqueryRelation subquery, Scope context) {
             if (subquery.getResolveTableName() != null && subquery.getResolveTableName().getTbl() == null) {
-                ErrorReport.reportSemanticException(ErrorCode.ERR_DERIVED_MUST_HAVE_ALIAS);
+                String catalog = ConnectContext.get().getCurrentCatalog();
```
kangkaisen, line 993: I think  we shouldn't change the default behavior
The SQL standard requires that derived tables (subqueries) must have aliases.

we could add a new session variable for  whether auto generate aliases.

fe/fe-core/src/test/java/com/starrocks/sql/analyzer/AnalyzeJoinTest.java
```diff
@@ -197,7 +198,7 @@ public void testParenthesizedRelation() {
         analyzeFail(sql, "Not unique table/alias: 'a'");
 
         sql = "select * from (t0 a, (select * from t1))";
-        analyzeFail(sql, "Every derived table must have its own alias");
+        analyzeWithoutTestView(sql);
```
kangkaisen, line 201: I think it should pass `analyzeSuccess `

fe/fe-core/src/test/java/com/starrocks/sql/analyzer/AnalyzeSubqueryTest.java
```diff
@@ -46,7 +48,7 @@ public void testSimple() {
         analyzeFail("select a.k1 from (select k1, k2 from (select v1 as k1, v2 as k2 from t0) a) b");
 
         analyzeSuccess("select * from (select count(v1) from t0) a");
-        analyzeFail("select * from (select count(v1) from t0)");
+        analyzeWithoutTestView("select * from (select count(v1) from t0)");
```
kangkaisen, line 51: we should add more test cases for this feature, like multi subqueries, same subqueries.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
kangkaisen, line 994: I think would better use Auto-increment id in query level

eshishki:
@duanyyyyyyy the motivation (Hive views whose body has an unaliased derived table like `select * from (select * from a union all select * from b)`) is legit, but as kangkaisen noted this PR drops the standard `ERR_DERIVED_MUST_HAVE_ALIAS` check for everyone, which we probably don't want — the SQL standard requires derived tables to be aliased, and a user's own query should still fail.

Instead of a session variable, could this be scoped to the external-view path? External view bodies are expanded through the `ConnectorView` branch in `QueryAnalyzer` (`connectorView.getQueryStatement()`), so that `QueryStatement` could carry a "from external view" flag and `visitSubqueryRelation` would only auto-generate the alias when it's set. That fixes the Hive-view case for everyone without a flag and keeps the default standard-compliant. kangkaisen's other asks (more test cases, query-level auto-increment id for the generated alias) still apply.

It's been ~11 months and is now CONFLICTING — are you still up for reworking it this way? Otherwise probably best to close.

PR #59771, opened 2025-06-10
[BugFix] Transform all exception for some operation in HdfsFsMgr into StarRocksException to prevent Exception leak for the caller

### LINKED ISSUE #59770: Automated Cluster Snapshot can not set off because of network failure
1. Turn on automated cluster snapshot
2. Turn off automated cluster snapshot if network has some problem.

Set off successfully

### Real behavior (Required)
Get exception and fail

3.4.2

srlch:
## Why I'm doing:
Some operation in HdfsFsMgr does not catch all kinds of exception which cause some non-StarRocksException will be throw for the caller. But the caller usually only catch StarRocksException and lead Exception leak

## What I'm doing:
Make sure all exception will be transformed into StarRocksException

Fixes #59770

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
+++ b/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
@@ -1311,9 +1311,6 @@ public class HdfsFsManager {
             Thread.interrupted(); // clear interrupted flag
             LOG.error("Interrupted while delete path: " + path, e);
             throw new StarRocksException("Failed to delete path: " + path, e); // throw unified user exception
-        } catch (IOException e) {
-            LOG.error("errors while delete path " + path, e);
-            throw new StarRocksException("delete path " + path + "error", e);
         } catch (Exception e) {
             LOG.error("errors while delete path " + path, e);
             throw new StarRocksException("delete path " + path + "error", e);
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
```diff
@@ -1314,6 +1314,9 @@ public void deletePath(String path, Map<String, String> loadProperties) throws S
         } catch (IOException e) {
             LOG.error("errors while delete path " + path, e);
             throw new StarRocksException("delete path " + path + "error", e);
+        } catch (Exception e) {
```
wyb, line 1314: directly replace above `IOException` with `Exception`？

srlch: Fixed

wyb (approved):

hongkunxu (approved):

gengjun-git (approved):

PR #59778, opened 2025-06-10
[Feature] Support write and read decimal256

stephen-shelby:
## What I'm doing:
the changes of this PR are all to ensure that decimal256 can be correctly compiled and read-written. some traits aren't implemented yet because we need other features to test them. the main change of this patch is that we define int256 as the storage for decimal256. This PR is only the simplest version for int256 arithmetic operations. We will optimize it in the subsequent PR.

Currently, StarRocks uses multiple widths to represent decimals
1.  Decimal (P <= 18, S), LogicalType is Decimal64, stored using int64, delegate LogicalType is BIGINT 
2.  Decimal (P > 18 & P <= 38, S), LogicalType is Decimal128, stored using int128, delegate LogicalType is LARGEINT 
For the newly added Decimal(P > 38 & P <= 76, S), the LogicalType is Decimal256, which is represented by the custom int256, and its delegate LogicalType is INT_256. 

### Storage Layer
use int256 as storage https://github.com/StarRocks/starrocks/pull/59808

### Computing layer
Currently, StarRocks has abstracted a set of good computing layer expressions for Decimal, `DecimalV3Column`, which is well adapted to the precision and scale unique to decimal. For Decimal256, it can be reused directly. 
`DecimalV3Column` is a fixed length column. We can reuse the serialized and deserialized implement. And we need to adapt the corresponding hash function for int256. 
Some core definitions as follow up:
```
template class DecimalV3Column<int32_t>;
template class DecimalV3Column<int64_t>;
template class DecimalV3Column<int128_t>;
template class DecimalV3Column<int256_t>;

template <>
struct RunTimeTypeTraits<TYPE_INT256> {
    using CppType = starrocks::int256_t;
    using ColumnType = FixedLengthColumn<starrocks::int256_t>;
};

template <>
struct RunTimeTypeTraits<TYPE_DECIMAL256> {
    using CppType = int256_t;
    using ColumnType = Decimal256Column;
};

using Decimal256Column = DecimalV3Column<int256_t>;
```

### Read-Written behavior
The read and write behavior is consistent with the previous decimal.
use `decimal(50, 48)` for example
- `insert into test_decimal(d1) select 1.222222;`  work normal
- `insert into test_decimal(d1) select 11111111111111111111111111111111111111111111.222222;` Successfully written. Since the scale is 48, the integer part plus 48 has exceeded the maximum integer that can be represented by 256 bits, So fe directly casts the value to null.
- `insert into test_decimal(d1) select 333`;  throw an exception on the be side`ERROR 1064 (HY000): Insert has filtered data`.  Because the integer part plus the scale exceeds the maximum value represented by a 50-digit integer.

### Others behavior changes
- When inserting, you can write a integer that exceeds 38 and has a maximum of 76 digits. However, you cannot write the maximum 77-digit decimal integer that can be represented by 256 bits. This is the same as decimal128. 
- TYPE_INT is only used for TYPE_DECIMAL256 storage and won't be exposed to users.

Fixes #issue
https://github.com/StarRocks/starrocks/issues/59645

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


be/src/types/int256.cpp
```diff
@@ -0,0 +1,140 @@
+#include "types/int256.h"
```
alvin-phoenix-ai: license header

be/src/types/int256.h
```diff
@@ -0,0 +1,569 @@
+// src/types/int256.h
```
alvin-phoenix-ai: You should create a separate PR to introduce int256 with its UT to avoid such a big PR.

stephen-shelby: separated PR https://github.com/StarRocks/starrocks/pull/59808

be/src/util/decimal_types.h
```diff
@@ -133,6 +174,8 @@ template <typename T>
 inline constexpr T get_max() {
     if constexpr (std::is_same_v<T, int128_t>) {
         return INT128_MAX;
+    } else if constexpr (std::is_same_v<T, int256_t>) {
+        return INT256_MAX;
```
satanson, line 178: where is INT256_MAX's definition?

satanson (dismissed):

be/src/util/decimal_types.h
stephen-shelby: [here](https://github.com/StarRocks/starrocks/blob/main/be/src/types/int256.h#L572)

satanson (approved):

Seaven (approved):

kangkaisen (approved):

PR #59814, opened 2025-06-11
[BugFix] Let submitted tasks without execution can be awared in starrocks::LakeServiceImpl to set a correct response and status

srlch:
## Why I'm doing:
Currently, some task submitted by starrocks::LakeServiceImpl can not aware whether the task is executed or not,
which will cause the response or status to be set incorrectly.

## What I'm doing:
Change AutoCleanRunnable to CancellableRunnable, which has a canceller to be called when threadpool is shutdown
to handle the correct response and status. Remove the cleaner and caller should do the proper clean-up work
in run() or when the task submitted failed.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/be/src/service/service_be/lake_service.cpp
+++ b/be/src/service/service_be/lake_service.cpp
@@ -179,8 +179,9 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
     }
     bool skip_write_tablet_metadata = request->has_enable_aggregate_publish() && request->enable_aggregate_publish();
     for (auto tablet_id : request->tablet_ids()) {
-        auto task = std::make_shared<AutoCleanRunnable>(
+        auto task = std::make_shared<CancellableRunnable>(
                 [&, tablet_id] {
+                    DeferOp defer([&] { latch.count_down(); });
                     scoped_refptr<Trace> child_trace(new Trace);
                     Trace* sub_trace = child_trace.get();
                     trace->AddChildTrace("PublishTablet", sub_trace);
@@ -262,7 +263,18 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
                     TRACE("finished");
                     g_publish_tablet_version_latency << (butil::gettimeofday_us() - run_ts);
                 },
-                [&] { latch.count_down(); });
+                [&] {
+                    g_publish_version_failed_tasks << 1;
+                    Status st = Status::Cancelled(
+                            fmt::format("publish version task has been cancelled, tablet_id={}", tablet_id));
+                    LOG(WARNING) << st;
+                    std::lock_guard l(response_mtx);
+                    response->add_failed_tablets(tablet_id);
+                    if (response->status().status_code() == 0) {
+                        st.to_protobuf(response->mutable_status());
+                    }
+                    latch.count_down();
+                });
 
         auto st = thread_pool_token.submit(std::move(task), timeout_deadline);
         if (!st.ok()) {
@@ -272,6 +284,7 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
             std::lock_guard l(response_mtx);
             response->add_failed_tablets(tablet_id);
             st.to_protobuf(response->mutable_status());
+            latch.count_down();
         }
     }
 
@@ -338,13 +351,21 @@ struct AggregatePublishContext {
                 publish_status = Status::InternalError("can not find put_aggregate_metadata thread pool");
             } else {
                 auto latch = BThreadCountDownLatch(1);
-                auto task = std::make_shared<AutoCleanRunnable>(
-                        [&] { publish_status = env->lake_tablet_manager()->put_bundle_tablet_metadata(tablet_metas); },
-                        [&] { latch.count_down(); });
+                auto task = std::make_shared<CancellableRunnable>(
+                        [&] {
+                            DeferOp defer([&] { latch.count_down(); });
+                            publish_status = env->lake_tablet_manager()->put_bundle_tablet_metadata(tablet_metas);
+                        },
+                        [&] {
+                            publish_status = Status::Cancelled("put_bundle_tablet_metadata task has been cancelled");
+                            LOG(WARNING) << publish_status;
+                            latch.count_down();
+                        });
                 Status submit_st = thread_pool->submit(std::move(task));
                 if (!submit_st.ok()) {
                     LOG(WARNING) << "Fail to submit put_bundle_tablet_metadata task";
                     publish_status = submit_st;
+                    latch.count_down();
                 }
                 latch.wait();
             }
@@ -429,8 +450,9 @@ void LakeServiceImpl::_submit_publish_log_version_task(const int64_t* tablet_ids
 
     for (int i = 0; i < tablet_size; i++) {
         auto tablet_id = tablet_ids[i];
-        auto task = std::make_shared<AutoCleanRunnable>(
+        auto task = std::make_shared<CancellableRunnable>(
                 [&, tablet_id] {
+                    DeferOp defer([&] { latch.count_down(); });
                     auto st = lake::publish_log_version(_tablet_mgr, tablet_id, txn_infos, log_versions);
                     if (!st.ok()) {
                         g_publish_version_failed_tasks << 1;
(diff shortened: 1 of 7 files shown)


be/src/service/service_be/lake_service.cpp
```diff
@@ -178,9 +178,12 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
         rebuild_pindex_tablets.insert(id);
     }
     bool skip_write_tablet_metadata = request->has_enable_aggregate_publish() && request->enable_aggregate_publish();
+    std::unordered_set<int64> no_execution_tablets(request->tablet_ids().begin(), request->tablet_ids().end());
     for (auto tablet_id : request->tablet_ids()) {
         auto task = std::make_shared<AutoCleanRunnable>(
                 [&, tablet_id] {
+                    DeferOp defer([&]() { no_execution_tablets.erase(tablet_id); });
```
xiangguangyxg: need lock ?

srlch: Fixed

be/src/util/threadpool.h
```diff
@@ -79,15 +79,22 @@ class Runnable {
...
+            : _runnable(std::move(runner)), _cleaner(std::move(cleaner)), _st(Status::Unknown("Not yet executed")) {}
+    virtual ~AutoCleanRunnable() { _cleaner(_st); }
 
-    virtual void run() override { _runnable(); }
+    virtual void run() override {
+        set_status(Status::OK());
+        _runnable();
+    }
+
+    void set_status(Status st) { _st = st; }
```
xiangguangyxg: need thread safe ?

xiangguangyxg: Would it be simpler to execute only one of the runner or cleaner ?

srlch: Fixed

srlch: Could you give some detail about how to just run one of them? It seems that we always need two functions here.

xiangguangyxg: one is for normal execution, another is for abort execution. Use an atomic bool flag to garanteen only one will be executed.

srlch: Updated

be/src/util/threadpool.cpp
```diff
@@ -707,9 +730,21 @@ Status ConcurrencyLimitedThreadPoolToken::submit(std::shared_ptr<Runnable> task,
...
-    return _pool->submit(std::move(token_task));
+    auto token_task = std::make_shared<CancellableRunnable>(
+            [t = task, sem = _sem] {
+                t->run();
+                sem->release();
+            },
+            [t = task, sem = _sem] {
+                t->cancel();
+                sem->release();
+            });
```
xiangguangyxg: Using a specific class implemented runable here could avoid coping task.

be/src/util/threadpool.h
```diff
@@ -68,26 +68,30 @@ class Runnable {
...
-//  The function or the runnable submit to a ThreadPool successfully, may not be running
-// at all because of threadpool shutdown. The caller who depends on the function/runnable
-// execution to signal or clean resources, needs to inject a `cleaner` to ensure the clean-
-// up work can be done under no condition.
-class AutoCleanRunnable : public Runnable {
+// There are three states in the life cycle of this class and caller must handle the
+// correct clean-up work in the corrsponding case:
+// 1. Schedule by the threadpool and execute run().
+// 2. Cancelled by the threadpool when it has been shutdown(), runnable will be called cancel().
+// 3. Submit runnable task failed into the threadpool.
```
kevincai:
```suggestion
// 3. Submit runnable task failed into the threadpool. Neither `run()` nor `cancel()` will be invoked.
```

be/src/util/threadpool.cpp
```diff
@@ -129,6 +130,15 @@ void ThreadPoolToken::shutdown() {
     // are destructed after the lock is released. This is important because the task's
     // destructors may acquire locks, etc., so this also prevents lock inversions.
     PriorityQueue<ThreadPool::NUM_PRIORITY, ThreadPool::Task> to_release;
+    DeferOp defer([&]() {
+        while (!to_release.empty()) {
+            (to_release.front().runnable)->cancel();
```
kevincai: make it exception safe, so every task will have chance to invoke `cancel()`

be/src/util/threadpool.h
```diff
@@ -68,26 +68,30 @@ class Runnable {
...
+class CancellableRunnable : public Runnable {
```
kevincai, line 82: need to fix the comments of ThreadPool::submit and ThreadPool::submit_func

be/src/service/service_be/lake_service.cpp
```diff
@@ -1168,12 +1273,22 @@ void LakeServiceImpl::vacuum_full(::google::protobuf::RpcController* controller,
...
-    auto task = std::make_shared<AutoCleanRunnable>([&] { lake::vacuum_full(_tablet_mgr, *request, response); },
-                                                    [&] { latch.count_down(); });
+    auto task = std::make_shared<CancellableRunnable>(
+            [&] {
+                DeferOp defer([&] { latch.count_down(); });
+                lake::vacuum_full(_tablet_mgr, *request, response);
+            },
+            [&] {
+                Status st = Status::Cancelled("full vacuum task has been cancelled");
+                LOG(WARNING) << st;
```
kevincai: this logline has no useful information, shall print at least the tablet id, or remove this logline

be/src/util/threadpool.h
srlch: Fixed

be/src/util/threadpool.cpp
srlch: Fixed

be/src/service/service_be/lake_service.cpp
srlch: Fixed

be/src/util/threadpool.h
srlch: Fixed

be/src/util/threadpool.h
```diff
@@ -68,26 +68,30 @@ class Runnable {
...
+    virtual ~CancellableRunnable() = default;
 
     virtual void run() override { _runnable(); }
 
+    virtual void cancel() override { _canceller(); }
+
 protected:
     std::function<void()> _runnable;
-    std::function<void()> _cleaner;
+    std::function<void()> _canceller;
```
xiangguangyxg, line 94: Will it be simpler if there are 3 tasks? runnable, canceller and cleaner, then you only need to add the canceller to handle the cancellation, runnable and cleaner are keep unchanged.

srlch: I think keep it only runnable and canceller is more simpler for the user.

kevincai: This PR can be considered as a re-do of the AutoCleanRunnable PR, the cleaner is not necessary, since there will be clearly 3 different cases. 1. submit failed, the caller need to handle the failure and do clean up. 2. submit succeeded and executed, _runnable will be executed, 3. submit succeeded but not executed, the _canceller will be executed. A `_cleaner` might be just a sugar to put the common things into both _runnable() and _cancelled() and usually this can be done by defining a lambda and call the lambda in both _runnable() and _cancelled().

be/src/util/threadpool.cpp
```diff
@@ -296,6 +310,21 @@ void ThreadPool::shutdown() {
...
+                try {
+                    (pq.front().runnable)->cancel();
+                } catch (...) {
+                    LOG(WARNING) << "Exception while cancelling runnable";
+                }
+                // PriorityQueue is not iterateable unless we pop the front element.
+                // But it is safe to do that because to_release will be destroyed just
+                // after the defer.
+                pq.pop_front();
+            }
```
kevincai: Extract the code snippets into a simple static function `void cancelTasksInQueue(PriorityQueue<...> &)`

```
void cancelTasksInQueue(PriorityQueue<NUM_PRIORITY, Task> &queue) {
    ...
}
```

So both ThreadPool::shutdown and ThreadPoolToken::shutdown can reuse the code, only one place need to be changed if this part needs update in future.

be/src/util/threadpool.h
xiangguangyxg: Yes but without the _cleaner, there are so many explicit `count_down()` added in many places, they should be eliminated, or the code is hard to maintain.

be/src/util/threadpool.cpp
srlch: Fixed

kevincai (dismissed):

be/src/util/threadpool.h
xiangguangyxg: Can the explicit `count_down()` be eliminated by using `DeferOP`?

srlch: Seems `count_down()` is too specified for Runnable and its derived class. I think we need a simpler interface for the general purpose.

kevincai (dismissed):

xiangguangyxg (approved):

wyb (approved):

kevincai (approved):

Seaven (approved):

PR #59845, opened 2025-06-12
[Enhancement] Introduce ClusterSnapshotInfo for cluster snapshot

srlch:
## What I'm doing:
This pr introduce ClusterSnapshotInfo for cluster snapshot which will be used in the future pr to save the cluster snapshot information(version, object id.... etc.) for more precise gc control.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
@@ -1,0 +1,210 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.lake.snapshot;
+
+import com.google.gson.annotations.SerializedName;
+import com.starrocks.catalog.Database;
+import com.starrocks.catalog.MaterializedIndex;
+import com.starrocks.catalog.OlapTable;
+import com.starrocks.catalog.Partition;
+import com.starrocks.catalog.PhysicalPartition;
+import com.starrocks.catalog.Table;
+import com.starrocks.server.GlobalStateMgr;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.starrocks.server.GlobalStateMgr.isCheckpointThread;
+
+public class ClusterSnapshotInfo {
+    // layer struct begin from db
+    @SerializedName(value = "dbInfos")
+    private Map<Long, DatabaseSnapshotInfo> dbInfos;
+
+    public ClusterSnapshotInfo() {
+        this.dbInfos = new HashMap<>();
+
+        if (!isCheckpointThread()) {
+            return;
+        }
+
+        List<Long> dbIds = GlobalStateMgr.getCurrentState().getLocalMetastore().getDbIds();
+        for (Long dbId : dbIds) {
+            Database db = GlobalStateMgr.getCurrentState().getLocalMetastore().getDb(dbId);
+            if (db == null) {
+                continue;
+            }
+            dbInfos.put(dbId, new DatabaseSnapshotInfo(db));
+        }
+    }
+
+    public boolean isEmpty() {
+        return this.dbInfos.isEmpty();
+    }
+
+    public long getVersion(long dbId, long tableId, long partId, long physicalPartId) {
+        PhysicalPartitionSnapshotInfo physicalPartInfo = getPhysicalPartitionInfo(dbId, tableId, partId, physicalPartId);
+        if (physicalPartInfo == null) {
+            return 0;
+        }
+        return physicalPartInfo.version;
+    }
+
+    public boolean isDbExisted(long dbId) {
+        return getDbInfo(dbId) != null;
+    }
+
+    public boolean isTableExisted(long dbId, long tableId) {
+        return getTableInfo(dbId, tableId) != null;
+    }
+
+    public boolean isPartitionExisted(long dbId, long tableId, long partId) {
+        return getPartitionInfo(dbId, tableId, partId) != null;
+    }
+
+    public boolean isMaterializedIndexExisted(long dbId, long tableId, long partId, long physicalPartId, long indexId) {
+        return getIndexInfo(dbId, tableId, partId, physicalPartId, indexId) != null;
+    }
+
+    public boolean isMaterializedIndexExisted(long dbId, long tableId, long partId, long indexId) {
+        PartitionSnapshotInfo partInfo = getPartitionInfo(dbId, tableId, partId);
+        if (partInfo == null) {
+            return false;
+        }
+
+        for (PhysicalPartitionSnapshotInfo physicalPartInfo : partInfo.physicalPartInfos.values()) {
+            if (isMaterializedIndexExisted(dbId, tableId, partId, physicalPartInfo.physicalPartitionId, indexId)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private DatabaseSnapshotInfo getDbInfo(long dbId) {
+        return dbInfos.get(dbId);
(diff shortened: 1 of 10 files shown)


fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
```diff
@@ -0,0 +1,210 @@
...
+import com.starrocks.catalog.Table;
+import com.starrocks.server.GlobalStateMgr;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.starrocks.server.GlobalStateMgr.isCheckpointThread;
+
+public class ClusterSnapshotInfo {
```
alvin-phoenix-ai, line 21: You should create a dedicated UT file for this class.
Rather than test it with other cases.

srlch: Fixed

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
```diff
@@ -0,0 +1,210 @@
...
+            return;
+        }
+
+        List<Long> dbIds = GlobalStateMgr.getCurrentState().getLocalMetastore().getDbIds();
+        for (Long dbId : dbIds) {
+            Database db = GlobalStateMgr.getCurrentState().getLocalMetastore().getDb(dbId);
+            if (db == null) {
+                continue;
+            }
+            dbInfos.put(dbId, new DatabaseSnapshotInfo(db));
```
alvin-phoenix-ai: Oh, it causes circular dependency, which is bad in program design.
ClusterSnapshotInfo should not depend on GlobalStateMgr. Because GlobalStateMgr is the top of the dependency tree. It is not reasonable to let a basic class to depend on it.
You should design this class without depending on GlobalStateMgr, to construct this info, you should let constructors give the necessary materials, rather than constructing in the constructors.

alvin-phoenix-ai: If you want ClusterSnapshotInfo be immutable, you can create a Builder class. If this information is mutable, you can add mutation interfaces to modify the class.
But it is not reasonable to depend on GlobalStateMgr in the Constructor function, which means that when you create one class, the whole process should be available.

alvin-phoenix-ai (dismissed):

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
srlch: Fixed, this class is not necessary to be immutable. I introduce a public interface `rebuildInfo` to fill the info by the given database list. Make constructor decoupling with GlobalStateMgr.

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
```diff
@@ -0,0 +1,202 @@
...
+        public DatabaseSnapshotInfo(Database db) {
+            this.dbId = db.getId();
+            this.tableInfos = new HashMap<>();
+
+            for (Table table : db.getTables()) {
+                if (table.isCloudNativeTableOrMaterializedView()) {
+                    tableInfos.put(table.getId(), new TableSnapshotInfo((OlapTable) table));
+                }
+            }
+        }
```
alvin-phoenix-ai:
```suggestion
        public DatabaseSnapshotInfo(long dbId, Map<Long, TableSnapshotInfo> tableInfos) {
            this.dbId = dbId;
            this.tableInfos = tableInfos;
        }
```
1. I think constructors should be kept as simple as possible.
2. You can write another Visitor Class to finish this logic in that class
Then, ClusterSnapshotInfo will depend on no Table, Db, and MaterialIndex

srlch: Fixed

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
```diff
@@ -0,0 +1,236 @@
...
+
+        public PhysicalPartitionSnapshotInfo(long physicalPartId, long visibleVersion,
+                                             Map<Long, MaterializedIndexSnapshotInfo> indexInfos) {
+            this.physicalPartitionId = physicalPartId;
+            this.version = visibleVersion;
+            this.indexInfos = indexInfos;
+        }
+    }
+
+    private static class MaterializedIndexSnapshotInfo {
```
alvin-phoenix-ai: How about creating a dedicated file for each class? 
Besides table-related objects, there will be other objects, such as Auth-related objects.
If all classes are included in one file, it will not be very readable.

srlch: Fixed

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
srlch: Use builder instead of visitor to make ClusterSnapshotInfo immutable.

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
```diff
@@ -0,0 +1,171 @@
...
+            return null;
+        }
+        return physicalPartInfo.indexInfos.get(indexId);
+    }
+
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    public static class Builder {
```
alvin-phoenix-ai: This class should be moved to a dedicated file with a better name.
Usually, Builder is used to build one class.
Maybe this class can be named SnapshotInfoHelper?
And all the functions will be static functions like `static generateDb()`

srlch: Fixed

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/SnapshotInfoHelper.java
```diff
@@ -0,0 +1,79 @@
...
+import com.starrocks.catalog.PhysicalPartition;
+import com.starrocks.catalog.Table;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class SnapshotInfoHelper {
+    public SnapshotInfoHelper() {
+    }
```
xiangguangyxg: remove

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public long getVersion(long dbId, long tableId, long partId, long physicalPartId) {
+        PhysicalPartitionSnapshotInfo physicalPartInfo = getPhysicalPartitionInfo(dbId, tableId, partId, physicalPartId);
+        if (physicalPartInfo == null) {
+            return 0;
+        }
+        return physicalPartInfo.version;
+    }
+
+    public boolean isDbExisted(long dbId) {
```
xiangguangyxg:
```suggestion
    public boolean containsDb(long dbId) {
```

xiangguangyxg: isXxxExisted -> containsXxx

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/SnapshotInfoHelper.java
srlch: Fixed

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotInfo.java
srlch: Fixed

xiangguangyxg (approved):

gengjun-git (approved):

PR #59859, opened 2025-06-12
[Enhancement] Introduce distribution_key_column_names in TabletSchema

srlch:
## What I'm doing:
This is a prerequisite for tablet spliting. We introduce distribution_key_column_names in TabletSchema to ensure the ability to use them for data filtering base on hash in the future pr.

We choose physical column names instead of column unique id because column names are easier to handle compatibility issues for the tables which create in the older version without unique id in FE.

Fixes #59134 

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/metadata_util.cpp
+++ b/be/src/storage/metadata_util.cpp
@@ -399,7 +399,7 @@ Status convert_t_schema_to_pb_schema(const TTabletSchema& tablet_schema, uint32_
         schema->set_bf_fpp(tablet_schema.bloom_filter_fpp);
     }
 
-    if (tablet_schema.__isset.distribution_key_column_ids) {
+    if (tablet_schema.__isset.distribution_key_column_names) {
         for (const auto& col_name : tablet_schema.distribution_key_column_names) {
             schema->add_distribution_key_column_names(col_name);
         }
--- a/be/src/storage/metadata_util.cpp
+++ b/be/src/storage/metadata_util.cpp
@@ -398,6 +398,10 @@ Status convert_t_schema_to_pb_schema(const TTabletSchema& tablet_schema, uint32_
     if (has_bf_columns && tablet_schema.__isset.bloom_filter_fpp) {
         schema->set_bf_fpp(tablet_schema.bloom_filter_fpp);
     }
+
+    for (const auto& col_name : tablet_schema.distribution_key_column_names) {
+        schema->add_distribution_key_column_names(col_name);
+    }
     return validate_tablet_schema(*schema);
 }
 
--- a/be/src/storage/metadata_util.cpp
+++ b/be/src/storage/metadata_util.cpp
@@ -399,8 +399,10 @@ Status convert_t_schema_to_pb_schema(const TTabletSchema& tablet_schema, uint32_
         schema->set_bf_fpp(tablet_schema.bloom_filter_fpp);
     }
 
-    for (const auto& col_name : tablet_schema.distribution_key_column_names) {
-        schema->add_distribution_key_column_names(col_name);
+    if (tablet_schema.__isset.distribution_key_column_ids) {
+        for (const auto& col_name : tablet_schema.distribution_key_column_names) {
+            schema->add_distribution_key_column_names(col_name);
+        }
     }
     return validate_tablet_schema(*schema);
 }
--- a/gensrc/proto/tablet_schema.proto
+++ b/gensrc/proto/tablet_schema.proto
@@ -96,6 +96,7 @@ message TabletSchemaPB {
     repeated uint32 sort_key_unique_ids = 13;
     repeated TabletIndexPB table_indices = 14;
     optional int32 compression_level = 15 [default=-1];
+    // physical column name
     repeated string distribution_key_column_names = 16;
     optional int64 id = 50;
 }
--- a/gensrc/proto/tablet_schema.proto
+++ b/gensrc/proto/tablet_schema.proto
@@ -96,6 +96,7 @@ message TabletSchemaPB {
     repeated uint32 sort_key_unique_ids = 13;
     repeated TabletIndexPB table_indices = 14;
     optional int32 compression_level = 15 [default=-1];
+    repeated string distribution_key_column_names = 16;
     optional int64 id = 50;
 }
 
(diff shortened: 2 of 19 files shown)


gensrc/thrift/AgentService.thrift
```diff
@@ -55,6 +55,7 @@ struct TTabletSchema {
     10: optional list<i32> sort_key_idxes
     11: optional list<i32> sort_key_unique_ids
     12: optional i32 schema_version;
+    13: optional list<string> distribution_key_column_names;
```
alvin-phoenix-ai, line 59: Why not use column id? 
For now, we support altering column names. 
The column name will be updated, but the column ID will remain unchanged.

alvin-phoenix-ai (changes requested):

gensrc/thrift/AgentService.thrift
srlch: Actually i use base/physical column name here which keeps unchanged even after rename column. But it seems that cid is much more simpler. Cid for distribution keys will not affected by fast schema evolution. I will change it into cid later.

srlch: Fixed

gensrc/proto/tablet_schema.proto
```diff
@@ -96,6 +96,7 @@ message TabletSchemaPB {
     repeated uint32 sort_key_unique_ids = 13;
     repeated TabletIndexPB table_indices = 14;
     optional int32 compression_level = 15 [default=-1];
+    repeated uint32 distribution_key_column_ids = 16;
```
xiangguangyxg: id or unique id ?

srlch: It seems that column id is enough? We can not change the order of distribution key by fast schema evolution unless we run alter job in non-fast way. It means that the cid keep unchanged during the life cycle of the tablet.

wyb: why not use `List<ColumnId.id>`?
`ColumnId.id` is not changed after table is created.

srlch: Yes, this is what i have done before, save the base/physical column name `(ColumnId.id)`. But it seems that cid of distribution keys can not be changed by fast schema evolution so use cid will be more simpler than using physical col name. Any suggestion for this?

alvin-phoenix-ai: Just to confirm, what is the difference between ColumnId.Id and cid? And why is it simpler to use cid?

srlch: `ColumnId.Id` is actually a column name but not numeric id. Since v3.3, we get two kinds of column names says `Column.getName` and `Column.getColumnId.id()` (original col name, TabletSchema actually saves `Column.getColumnId.id`). We build some mapping between them and rename columns will always change the first one but not the other one. So `Column.getColumnId.id` is unique and immutable after the table is created.

srlch: Simpler may not be very accurate, using cid would be easier to understand what this field actually represents.

alvin-phoenix-ai: Just curious, why was `Column.getColumnId.id()` introduced to resolve renaming column rather than just using cid?

srlch: Not 100% sure, but I guess it is more easy to implement to introduce `Column.getColumnId.id` for renaming because for BE side, nothing need to be changed.

alvin-phoenix-ai: From the existing document, there is no logic to find the column by `cid` in BE. If only the map from `Column. getColumnId. id` to column is maintained, I would prefer to use `Column. getColumnId. id`.
If

be/src/storage/metadata_util.cpp
```diff
@@ -398,6 +398,10 @@ Status convert_t_schema_to_pb_schema(const TTabletSchema& tablet_schema, uint32_
     if (has_bf_columns && tablet_schema.__isset.bloom_filter_fpp) {
         schema->set_bf_fpp(tablet_schema.bloom_filter_fpp);
     }
+
+    for (const auto& cid : tablet_schema.distribution_key_column_ids) {
```
wyb: check `tablet_schema.__isset.distribution_key_column_ids`

gensrc/proto/tablet_schema.proto
srlch: Fixed, use `Column.getColumnId.id` instead

be/src/storage/metadata_util.cpp
srlch: Fixed

gensrc/thrift/AgentService.thrift
```diff
@@ -55,6 +55,8 @@ struct TTabletSchema {
     10: optional list<i32> sort_key_idxes
     11: optional list<i32> sort_key_unique_ids
     12: optional i32 schema_version;
+    // physical column name
+    13: optional list<string> distribution_key_column_names;
```
alvin-phoenix-ai, line 59: This class is TabletSchema. 
A tablet is generated after the table is partitioned and distributed.
So, distributed column information would be a TABLE attribute, rather than a tablet attribute.
Could you explain the necessity of persisting this attribute in TabletSchema?
Why not get the information from the corresponding table?

srlch: Because we need to use this information to compute the predicate in segment-level. So we have to pass this information to pushdown to segment-level. We can pass it on runtime or persist it into a tablet-level, even rowset-level metadata not necessary in TabletSchema. So you means it prefer to pass this information in runtime by some tablet-level param? Or still persist it but in other tablet-level metadata or even in rowst-level?

alvin-phoenix-ai: How will the prediction be generated? If predications are generated on the frontend, you can get the distributed information from the table schema.

PR #59883, opened 2025-06-13
[Enhancement] change a few metrics fe leader awareness

kevincai:
* add label `is_leader={"true"|"false"}` for those metrics that distinguish between leader and follower

## What I'm doing:

Fixes #59887

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/FeConstants.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/FeConstants.java
@@ -126,6 +126,8 @@ public class FeConstants {
     public static final String DOCUMENT_ALTER_ROUTINE_LOAD =
             "https://docs.starrocks.io/docs/sql-reference/sql-statements/data-manipulation/ALTER_ROUTINE_LOAD";
 
+    public static final String METRIC_LABEL_IS_LEADER = "is_leader";
+
     public static String getNodeNotFoundError(boolean chooseComputeNode) {
         return chooseComputeNode ? COMPUTE_NODE_NOT_FOUND_ERROR : BACKEND_NODE_NOT_FOUND_ERROR;
     }
--- a/fe/fe-core/src/main/java/com/starrocks/metric/JsonMetricVisitor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/metric/JsonMetricVisitor.java
@@ -228,7 +228,7 @@ public class JsonMetricVisitor extends MetricVisitor {
                 String.valueOf(systemInfoService.getAliveComputeNodeNumber()),
                 Arrays.asList(new MetricLabel(TYPE, CN_NODE_NUM), new MetricLabel(STATUS, "alive")));
 
-        // only master FE has this metrics, to help the Grafana knows who is the leader
+        // only the leader FE has this metric, to help the Grafana knows who is the leader
         if (GlobalStateMgr.getCurrentState().isLeader()) {
             buildMetric(NODE_INFO, NOUNIT, String.valueOf(1),
                     Collections.singletonList(new MetricLabel(TYPE, "is_master")));
--- a/fe/fe-core/src/main/java/com/starrocks/metric/LeaderAwareGaugeMetric.java
+++ b/fe/fe-core/src/main/java/com/starrocks/metric/LeaderAwareGaugeMetric.java
@@ -1,0 +1,51 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.metric;
+
+import com.starrocks.common.FeConstants;
+import com.starrocks.server.GlobalStateMgr;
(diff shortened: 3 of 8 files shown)


fe/fe-core/src/main/java/com/starrocks/metric/JsonMetricVisitor.java
```diff
@@ -231,7 +232,7 @@ public void getNodeInfo() {
         // only master FE has this metrics, to help the Grafana knows who is the leader
```
wyb:
```suggestion
        // only leader FE has this metrics, to help the Grafana knows who is the leader
```

fe/fe-core/src/main/java/com/starrocks/metric/LeaderAwareGaugeMetric.java
```diff
@@ -0,0 +1,51 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.metric;
+
+import com.starrocks.common.FeConstants;
+import com.starrocks.server.GlobalStateMgr;
+
+/**
+ * Gauge metric is updated every time it is visited. Additionally, label {"is_leader"="true|false"} will be added automatically
```
wyb:
```suggestion
 * Gauge metric is updated every time it is visited. Additionally, label {"is_master"="true|false"} will be added automatically
```

fe/fe-core/src/test/java/com/starrocks/metric/LeaderAwareGaugeMetricTest.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        };
+
+        // setup Expectations before creating the metrics
+        createMetrics();
+        {
+            metricLongValue = 3;
+            MetricVisitor visitor = new PrometheusMetricVisitor("");
+            visitor.visit(metricLong);
+            // _metricLong{is_leader="true"} 3
```
wyb, line 69:
```suggestion
            // _metricLong{is_master="true"} 3
```

kevincai: done

fe/fe-core/src/main/java/com/starrocks/metric/JsonMetricVisitor.java
kevincai: done

fe/fe-core/src/main/java/com/starrocks/metric/LeaderAwareGaugeMetric.java
kevincai: done

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
```diff
@@ -260,13 +260,10 @@ public static synchronized void init() {
             }
 
             for (JobState state : JobState.values()) {
-                GaugeMetric<Long> gauge = new GaugeMetric<Long>("job",
+                Metric<Long> gauge = new LeaderAwareGaugeMetricLong("job",
```
alvin-phoenix-ai, line 263: Is there any documentation for these metrics? If there is, it's better to change the doc accordingly.

kevincai: let me check

kevincai: we have some explanation to the metrics, but without the details of each metric, about tags/labels to the metric. For example: https://docs.starrocks.io/docs/administration/management/monitoring/metrics/#starrocks_fe_tablet_num

fe/fe-core/src/main/java/com/starrocks/common/FeConstants.java
```diff
@@ -126,6 +126,9 @@ public class FeConstants {
     public static final String DOCUMENT_ALTER_ROUTINE_LOAD =
             "https://docs.starrocks.io/docs/sql-reference/sql-statements/data-manipulation/ALTER_ROUTINE_LOAD";
 
+    // Historic reason naming as "is_master" rather than "is_leader"
+    public static final String METRIC_LABEL_IS_LEADER = "is_master";
```
chaoyli: It's better to rename it to `is_leader` and change the previous name gradually.

fe/fe-core/src/main/java/com/starrocks/common/FeConstants.java
```diff
@@ -126,6 +126,9 @@ public class FeConstants {
     public static final String DOCUMENT_ALTER_ROUTINE_LOAD =
             "https://docs.starrocks.io/docs/sql-reference/sql-statements/data-manipulation/ALTER_ROUTINE_LOAD";
 
+    // Historic reason naming as "is_master" rather than "is_leader"
```
chaoyli: It doesn't meet the requirement of the customers. They want to get the consistent metrics between FE leader and FE followers. 
1. In current situation, I think we should forward to the leader to get the metrics.
2. In final, we should sync the metrics between FE leader and followers.

kevincai: it is a matter of fact how the system is designed, not how the system is required by the customer.

fe/fe-core/src/main/java/com/starrocks/common/FeConstants.java
kevincai: ok

wyb (approved):

alvin-phoenix-ai (approved):

PR #59919, opened 2025-06-16
[BugFix] Fix the `disk_device_id` function to aquire the device_id of an non-existent entry by its ancestor entries.

GavinMar:
## Why I'm doing:
We use the `device_id` to identify whether two disk paths locate in same disks. 
The old `disk_device_id` function only aquires stats for the accurate path parameter, which may be not exist and cause the function failed.

## What I'm doing:
Get the `device_id` of the given path from the current entry to its ancestor entries. If the current path does not exist, check its parent and ancestor entries.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/cache/datacache_utils.cpp
+++ b/be/src/cache/datacache_utils.cpp
@@ -163,8 +163,19 @@ Status DataCacheUtils::change_disk_path(const std::string& old_disk_path, const 
 }
 
 dev_t DataCacheUtils::disk_device_id(const std::string& disk_path) {
+    std::filesystem::path cur_path(disk_path);
+    cur_path = std::filesystem::absolute(cur_path);
+
+    // Traverse from the current path to the ancestor node and find the first existing path
+    while (!cur_path.empty()) {
+        if (std::filesystem::exists(cur_path) || cur_path == cur_path.root_path()) {
+            break;
+        }
+        cur_path = cur_path.parent_path();
+    }
+
     struct stat s;
-    if (stat(disk_path.c_str(), &s) != 0) {
+    if (stat(cur_path.c_str(), &s) != 0) {
         return 0;
     }
     return s.st_dev;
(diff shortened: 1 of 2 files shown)


dirtysalt (dismissed):

be/src/cache/datacache_utils.cpp
```diff
@@ -163,8 +163,19 @@ Status DataCacheUtils::change_disk_path(const std::string& old_disk_path, const
 }
 
 dev_t DataCacheUtils::disk_device_id(const std::string& disk_path) {
+    std::filesystem::path cur_path(disk_path);
+    cur_path = std::filesystem::absolute(cur_path);
+
+    // Traverse from the current path to the ancestor node and find the first existing path
+    while (!cur_path.empty()) {
+        if (std::filesystem::exists(cur_path) || cur_path == cur_path.root_path()) {
```
trueeyu, line 171: Will doing this encounter permission issues?

trueeyu (approved):

be/src/cache/datacache_utils.cpp
GavinMar: > Will doing this encounter permission issues?

Good point! From the function itself, it really may encounter the permission issues because the `stat` need the search permission for the target entry. However, from our use cases, it should not happen. When renaming an old cache path to a new one, we should have the write permission for the source and dest directory, and the search permission also is needed to locate the paths in os path_walk process.  So, the permissions for this function is a subset of the permissions we should have.

dirtysalt (approved):

Issue #58868, opened 2025-05-12
turn on profile in default


before-Sunrise:
## Enhancement
right now profile is disabled in default, but it's hard to positioning online problems without profile. So we need to turn on profile  in default. To enable profile in default, we need to do two things:
1. make profile async in BE
2. optimize profile's performance, maybe add more levels and only report necessary counters in default level

PR #59962, opened 2025-06-17
[Feature] make  profile async and turn on query profile on default

before-Sunrise:
## What I'm doing:
1. make profile report async in BE, profile merge and report will in other thread pool instead of pipeline-exec query pool
2. split profile from ExecStateReport, use another thrift rpc.  Right now Exec state report has too much information including query exec state/ load state etc, better to split profile from this interface since profile is less important
3. optimize profile merge and ToThrift method. 
   3.1 ：Using a Block-grained mempool to let counters in same level can be near by each other, this is cache-friendly for our top-down counter-merge algorithm and ToThrift method
   3.2  optmize counter-merge algorithm, only touch every counter once, and prefetch child counters when visit parent.
4. for counter which value is zero, don't display it. For unique metrics, display threshold is controled by be config:counter_display_threshold, default value is 50. These thresholds can delete useless counter
5. in high qps, we don't record profile if query's total time is less then profileThresHold, default is 100ms
6. remove collect profile time from profile's total time, since now profile collecting is totally async, sometimes profile collect time can be longer then query execution time.

Fixes #https://github.com/StarRocks/starrocks/issues/58868

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exec/pipeline/exec_state_reporter.cpp
+++ b/be/src/exec/pipeline/exec_state_reporter.cpp
@@ -45,7 +45,8 @@ std::string to_http_path(const std::string& token, const std::string& file_name)
 }
 
 std::unique_ptr<TReportExecStatusParams> ExecStateReporter::create_report_exec_status_params(
-        QueryContext* query_ctx, FragmentContext* fragment_ctx, const Status& status, bool done) {
+        QueryContext* query_ctx, FragmentContext* fragment_ctx, RuntimeProfile* profile,
+        RuntimeProfile* load_channel_profile, const Status& status, bool done, bool enable_async_profile_in_be) {
     auto res = std::make_unique<TReportExecStatusParams>();
     TReportExecStatusParams& params = *res;
     auto* runtime_state = fragment_ctx->runtime_state();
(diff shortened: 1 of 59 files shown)


wangsimo0:
ignore merge 4.0.0-rc01

wangsimo0:

be/src/exec/pipeline/exec_state_reporter.cpp
```diff
@@ -46,12 +46,11 @@ std::string to_http_path(const std::string& token, const std::string& file_name)
 
 std::unique_ptr<TReportExecStatusParams> ExecStateReporter::create_report_exec_status_params(
         QueryContext* query_ctx, FragmentContext* fragment_ctx, RuntimeProfile* profile,
-        RuntimeProfile* load_channel_profile, const Status& status, bool done) {
+        RuntimeProfile* load_channel_profile, const Status& status, bool done, bool enable_async_profile_in_be) {
```
lxynov, line 49: In the case of `enable_async_profile_in_be=true`, `profile` will be `nullptr`, otherwise it's a non-null. Should we use this to determine whether to report profiles or not? It'll also be safer as it eliminates the possibility of calling `profile->to_thrift()` when `profile` is `nullptr`

be/src/exec/pipeline/profile_manager.cpp
```diff
@@ -0,0 +1,191 @@
...
+                auto* process_timer = ADD_TIMER(new_instance_profile, "BackendProfileMergeTime");
+                COUNTER_SET(process_timer, process_raw_timer);
+            }
+        });
+
+        SCOPED_RAW_TIMER(&process_raw_timer);
+        std::vector<RuntimeProfile*> pipeline_profiles;
+        instance_profile->get_children(&pipeline_profiles);
+
+        std::vector<RuntimeProfile*> merged_driver_profiles;
```
lxynov, line 33: nit: use the same reserve-then-emplace_back pattern as in other places?

be/src/exec/pipeline/profile_manager.cpp
```diff
@@ -0,0 +1,191 @@
...
+        Status submit_report_status = _report_thread_pool->submit_func(std::move(profile_report_task));
+        if (!submit_report_status.ok()) {
+            profile_report_task();
+        }
+    };
+
+    Status status = _merge_thread_pool->submit_func(profile_merge_task);
+    if (!status.ok()) {
+        LOG(WARNING) << "Cannot submit merge-profile task: " << status.to_string();
+        profile_merge_task();
```
lxynov, line 166: wondering whether we want to just skip reporting this profile in case of asynchronous failure, since it may indicate workload pressure

fe/fe-core/src/main/java/com/starrocks/common/util/RunningProfileManager.java
```diff
@@ -0,0 +1,278 @@
...
+            status.addToError_msgs("query id " + DebugUtil.printId(request.getQuery_id()) + ", fragment instance id " +
+                    DebugUtil.printId(request.fragment_instance_id) + " not found");
+            return status;
+        }
+
+        boolean shouldProcessDone = false;
+        if (request.isDone()) {
+            shouldProcessDone = fragmentInstanceProfile.isDone.compareAndSet(false, true);
+        } else {
+            // if the fragment is already done, we can return directly
```
lxynov, line 121: Curious why will `fragmentInstanceProfile` be marked done already but the BE still sends an in-progress request?

fe/fe-core/src/main/java/com/starrocks/common/util/RunningProfileManager.java
```diff
@@ -0,0 +1,278 @@
...
+            LOG.info(
+                    "asyncProfileReport, queryid: {}, instanceIdx: {}, isDone:{}, shouldProcessDone:{}, total instance num:{}",
+                    DebugUtil.printId(request.getQuery_id()), DebugUtil.printId(request.fragment_instance_id),
+                    fragmentInstanceProfile.isDone.toString(), shouldProcessDone,
+                    queryProfile.fragmentInstanceProfiles.size());
+            queryProfile.tryToTriggerRuntimeProfileUpdate();
+            fragmentInstanceProfile.updateRunningProfile(request);
+            queryProfile.updateLoadChannelProfile(request);
+            
+            if (shouldProcessDone) {
```
lxynov, line 138: wondering is it that the FE won't process in-progress profile report at all? does this match previous behavior? if so, shall we change BE to not report in-progress profile at all?

lxynov (commented):
thank you for working on this. added a few minor comments/questions. thanks

PR #59975, opened 2025-06-17
[Feature] introduce RowIdColumn

### LINKED ISSUE #60015: support global late materialization
## Background
StarRocks provides relatively limited support for late materialization. Currently, only the Scan and HashJoinProbe operators implement some localized late materialization. 
In many scenarios, this can still result in excessive resource consumption, with Join and TopN being typical examples. Take the following query as an example:
```sql
select * from lineorder order by lo_partkey limit 1000000, 5;
```
In the current execution plan, all columns are read and participate in the top N calculation before the top N operator is executed, yet only 5 rows are ultimately returned.

In fact, we only need to read the `lo_partkey` column first. After completing the topN calculation, the other columns' data can be read, thereby significantly reducing CPU/IO overhead.

## Goal

Implement a global late materialization strategy to delay reading columns unrelated to computation as much as possible, thus improving query performance.

## Basic Design
```sql
select A.a, A.b, B.a, B.b from A join B on A.a = B.a;
```
Take this query as an example. After introducing global late materialization, the execution plan would be transformed as follows. 

Image

The columns A.b and B.b, which do not participate in the join computation, will be read after the Join operator executes. The Scan operators only need to return columns A.a, B.a, and the rowid for each row. After the Join operation executes, the A.b and B.b columns can then be read using their respective rowids.

## TODO
To achieve this goal, we need to address three key issues: 
1. introducing a rowid to globally identify each row of data; 
2. enabling the optimizer to generate execution plans for global late materialization;
3. enhancing the execution engine with Fetch and LookUp operators to retrieve corresponding data based on the rowid.

silverbullet233:
## Why I'm doing:

This is the first part of #60015. We need a method to uniquely identify each row of data in order to implement delayed reading. 

## What I'm doing:

I introduce a new data type, `ROW_ID`, used to globally identify a row. 
The `ROW_ID` is composed of three parts: 
* `be_id`: identifies a specific BE node
* `seg_id`: identifies a specific segment.
* `ord_id`: identifies the position of the data within the segment

By using these three components, we can locate a specific row of data.

In this PR, I have only implemented some basic functionality for ROW_ID, including RowIdColumn, the in-memory representation of ROW_ID, and GlobalRowIdColumnIterator, which is used to return the corresponding rowid for each row. Subsequent PRs will depend on these components to implement global late materialization.

Fixes #60015

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/column/CMakeLists.txt
+++ b/be/src/column/CMakeLists.txt
@@ -40,6 +40,7 @@ add_library(Column STATIC
         column_visitor_mutable.cpp
         json_column.cpp
         map_column.cpp
+        row_id_column.cpp
         struct_column.cpp
         stream_chunk.cpp
         column_view/column_view_base.cpp
--- a/be/src/column/chunk.cpp
+++ b/be/src/column/chunk.cpp
@@ -413,18 +413,44 @@ std::string Chunk::rebuild_csv_row(size_t index, const std::string& delimiter) c
 
 std::string Chunk::debug_columns() const {
     std::stringstream os;
-    os << "nullable[";
-    for (size_t col = 0; col < _columns.size() - 1; ++col) {
-        os << _columns[col]->is_nullable();
-        os << ", ";
+    if (_columns.empty()) {
+        os << "empty columns[]";
+    } else {
+        os << "nullable[";
+        for (size_t col = 0; col < _columns.size() - 1; ++col) {
+            os << _columns[col]->is_nullable();
+            os << ", ";
+        }
+        os << _columns[_columns.size() - 1]->is_nullable() << "]";
+        os << " const[";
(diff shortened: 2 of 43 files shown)


alvin-phoenix-ai (changes requested):
Hi, @silverbullet233 This is a new feature, please use the right title.
And a detailed design is required for new feature.

PR #59993, opened 2025-06-17
[Feature] Introduce Column Hash Predicate as a Record Predicate for data filtering

srlch:
## What I'm doing:
This is a prerequisite for tablet splitting. We introduce `RecordPredicate` and its derived class `ColumnHashIsCongruent` into `RowsetMetadataPB` to provide ability to filter data in all segments of the rowset using this predicate. `RecordPredicate` can be considered as a storage layer only predicate to filter data and `ColumnHashIsCongruent` calculate the available row for a given chunk in the following way:
1. Find columns by hash_column_names in chunk and calculate their combined `crc32` value.
2. `select[i] = (hashes[i] % modulus == remainder)`

Fixes #59134

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/record_predicate/column_hash_is_congruent.cpp
+++ b/be/src/storage/record_predicate/column_hash_is_congruent.cpp
@@ -1,0 +1,72 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "storage/record_predicate/column_hash_is_congruent.h"
+
+#include "column/chunk.h"
+#include "storage/record_predicate/record_predicate_helper.h"
+
+namespace starrocks {
+
+ColumnHashIsCongruent::ColumnHashIsCongruent(const RecordPredicatePB& record_predicate_pb)
+        : RecordPredicate(record_predicate_pb), _modulus(0), _remainder(0) {}
+
+Status ColumnHashIsCongruent::evaluate(Chunk* chunk, uint8_t* selection) const {
+    return evaluate(chunk, selection, 0, chunk->num_rows());
+}
+
+Status ColumnHashIsCongruent::evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const {
+    RETURN_IF_ERROR(RecordPredicateHelper::check_valid_schema(*this, *chunk->schema()));
+    size_t size = to - from;
+    std::vector<uint32_t> hashes(size, 0);
+    for (const auto& col_name : _column_names) {
+        chunk->get_column_by_name(col_name)->crc32_hash(&(hashes)[0], from, size);
+    }
+
+    const uint32_t* hashes_data = hashes.data();
+    for (size_t i = from; i < to; ++i) {
+        selection[i] = (hashes_data[i - from] % _modulus == _remainder);
+    }
+    return Status::OK();
+}
+
+Status ColumnHashIsCongruent::init(const RecordPredicatePB& record_predicate_pb) {
+    const auto& column_hash_is_congruent_pb = record_predicate_pb.column_hash_is_congruent();
+    RETURN_IF_ERROR(_check_valid_pb(column_hash_is_congruent_pb));
+    _modulus = column_hash_is_congruent_pb.modulus();
+    _remainder = column_hash_is_congruent_pb.remainder();
+    _column_names.insert(_column_names.end(), column_hash_is_congruent_pb.column_names().begin(),
+                         column_hash_is_congruent_pb.column_names().end());
+    return Status::OK();
+}
+
+Status ColumnHashIsCongruent::_check_valid_pb(const ColumnHashIsCongruentMeta& column_hash_is_congruent_meta) const {
+    if (column_hash_is_congruent_meta.column_names().empty()) {
+        return Status::InternalError("column hash congruent predicate has no hash columns defined");
+    }
+    for (const auto& column_name : column_hash_is_congruent_meta.column_names()) {
+        if (column_name.empty()) {
+            return Status::InternalError("column hash congruent predicate has empty hash column name");
+        }
+    }
+
+    int64_t modulus = column_hash_is_congruent_meta.modulus();
+    int64_t remainder = column_hash_is_congruent_meta.remainder();
+    if (modulus <= 0 || remainder < 0 || remainder >= modulus) {
+        return Status::InternalError("Invalid modulus or remainder in column hash congruent predicate");
+    }
+    return Status::OK();
+}
+
+} // namespace starrocks
--- a/be/src/storage/record_predicate/column_hash_is_congruent.h
+++ b/be/src/storage/record_predicate/column_hash_is_congruent.h
@@ -1,0 +1,48 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "storage/record_predicate/record_predicate.h"
+
+namespace starrocks {
+/*
+ * ColumnHashIsCongruentPB is used to evaluate a chunk of data as following:
+ * 1. Compute the hash value of the specified columns.
+ * 2. Check if the hash value modulo a given modulus equals a specified remainder.
+*/
+class ColumnHashIsCongruent final : public RecordPredicate {
+public:
+    using ColumnHashIsCongruentMeta = RecordPredicatePB::ColumnHashIsCongruentPB;
+
+    ColumnHashIsCongruent(const RecordPredicatePB& record_predicate_pb);
+    ~ColumnHashIsCongruent() override = default;
+
+    Status evaluate(Chunk* chunk, uint8_t* selection) const override;
+
+    Status evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const override;
+
+    Status init(const RecordPredicatePB& record_predicate_pb) override;
+
+    const std::vector<std::string>* column_names() const override { return &_column_names; }
+
+private:
+    Status _check_valid_pb(const ColumnHashIsCongruentMeta& column_hash_is_congruent_meta) const;
+
+    int64_t _modulus;
+    int64_t _remainder;
+    std::vector<std::string> _column_names;
+};
+
+} // namespace starrocks
--- a/be/src/storage/record_predicate/record_predicate.h
+++ b/be/src/storage/record_predicate/record_predicate.h
@@ -29,7 +29,8 @@ using RecordPredicateUPtr = std::unique_ptr<RecordPredicate>;
 
 class RecordPredicate {
 public:
-    RecordPredicate(const RecordPredicatePB& record_predicate_pb) : _pred_type(record_predicate_pb.type()), _children() {}
+    RecordPredicate(const RecordPredicatePB& record_predicate_pb)
+            : _pred_type(record_predicate_pb.type()), _children() {}
     virtual ~RecordPredicate() = default;
 
     // @Param chunk: the chunk to be evaluated
--- a/be/src/storage/record_predicate/record_predicate.h
+++ b/be/src/storage/record_predicate/record_predicate.h
@@ -1,0 +1,56 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "gen_cpp/types.pb.h"
+
+namespace starrocks {
+class Chunk;
+class Status;
+/*
+ * RecordPredicate is an abstract base class for predicates on chunk of data. All RecordPredicate
+ * can be only evaluated in storage layer.
+*/
+class RecordPredicate;
+using RecordPredicateSPtr = std::shared_ptr<RecordPredicate>;
+using RecordPredicateUPtr = std::unique_ptr<RecordPredicate>;
+
+class RecordPredicate {
+public:
+    RecordPredicate(const RecordPredicatePB& record_predicate_pb) : _pred_type(record_predicate_pb.type()), _children() {}
+    virtual ~RecordPredicate() = default;
+
+    // @Param chunk: the chunk to be evaluated
+    // @Param selection: the selection vector to be filled, which indicates whether each row in the chunk
+    // is selected (1) or not (0).
+    // @Param from: the starting index of the chunk to evaluate
+    // @Param to: the ending index of the chunk to evaluate
+    // @Return: Status indicating success or failure of the evaluation.
+    // The function without `from` and `to` parameters evaluates the entire chunk.
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection) const = 0;
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const = 0;
+
+    virtual Status init(const RecordPredicatePB& record_predicate_pb) = 0;
+
+    virtual const std::vector<std::string>* column_names() const = 0;
+
+    RecordPredicatePB::RecordPredicateTypePB type() { return _pred_type; }
+
+protected:
+    RecordPredicatePB::RecordPredicateTypePB _pred_type;
+    std::vector<RecordPredicateUPtr> _children;
+};
+
+} // namespace starrocks
--- a/gensrc/proto/lake_types.proto
+++ b/gensrc/proto/lake_types.proto
@@ -118,6 +118,8 @@ message RowsetMetadataPB {
     repeated int64 bundle_file_offsets = 14;
     // Whether segments are shared by multiple tablets
     repeated bool shared_segments = 15;
+    // Record Predicate use for filtering all segments for the current rowset
+    optional RecordPredicatePB record_predicate = 16;
 }
 
 // At present, the lake persistent index reuses the logic of the persistent index,
--- a/gensrc/proto/types.proto
+++ b/gensrc/proto/types.proto
@@ -181,6 +181,25 @@ message IsNullPredicatePB {
     optional bool is_not_null = 2;
 }
 
+message RecordPredicatePB {
+    enum RecordPredicateTypePB {
+        UNKNOWN = 0;
+        COLUMN_HASH_IS_CONGRUENT = 1;
+    }
+
+    message ColumnHashIsCongruentPB {
+        optional int64 modulus = 1;
+        optional int64 remainder = 2;
+        repeated string column_names = 3;
+    }
+
+    optional RecordPredicateTypePB type = 1;
+    // set by some type of predicate (RecordPredicateTypePB is AND/OR/NOT...) which will be implemented in the future
+    repeated RecordPredicatePB children = 2;
+    // The following member should be set one of them by predicate type
+    optional ColumnHashIsCongruentPB column_hash_is_congruent = 3;
+}
+
 message IndexValueWithVerPB {
     // will be set after mvcc support
     optional int64 version = 1;
--- a/be/src/storage/CMakeLists.txt
+++ b/be/src/storage/CMakeLists.txt
@@ -270,7 +270,9 @@ set(STORAGE_FILES
     index/vector/tenann_index_reader.cpp
     index/vector/tenann/del_id_filter.cpp
     index/vector/tenann/tenann_index_builder.cpp
-    index/vector/tenann/tenann_index_utils.cpp)
+    index/vector/tenann/tenann_index_utils.cpp
+    record_predicate/record_predicate_helper.cpp
+    record_predicate/column_hash_is_congruent.cpp)
 
 if (NOT BUILD_FORMAT_LIB)
     list(APPEND STORAGE_FILES  lake/starlet_location_provider.cpp)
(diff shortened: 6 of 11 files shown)


xiangguangyxg:
does it support using a hidden generated column to save the hash value ?

srlch:
> does it support using a hidden generated column to save the hash value ?

I think this predicate can support it as long as we give the correct col_name of the generated column, we can skip the hash value calculation and just using the column itself.

gensrc/proto/lake_types.proto
```diff
@@ -110,6 +110,7 @@ message RowsetMetadataPB {
     // only be set > 0 if `overlapped` is true
     optional uint32 next_compaction_offset = 13;
     repeated int64 bundle_file_offsets = 14;
+    optional ColumnHashModulusPredicatePB column_hash_modulus_predicate = 15;
```
alvin-phoenix-ai: Could you add comments for this field?
1. Is it a select predicate or a filter predicate? Better to name it selection_predicate or filter_predicate.
2. Better to have a general predicate for later extensibility.

gensrc/proto/types.proto
```diff
@@ -181,6 +181,18 @@ message IsNullPredicatePB {
...
+        UNKNOWN = 0;
+        CRC32 = 1;
+    }
+
+    optional ColumnHashTypePB type = 1;
+    optional int64 modulus = 2;
+    optional int64 remainder = 3;
+    repeated string hash_column_names = 4;
+}
+
```
alvin-phoenix-ai, line 202:
```suggestion
message RecordSelectPredicatePB {
    enum TypePB {
        UNKNOWN = 0;
        XXX = 1;
    }

    optional XXXType xxx = 1;

}
```

gensrc/proto/types.proto
```diff
@@ -181,6 +181,18 @@ message IsNullPredicatePB {
...
+    enum ColumnHashTypePB {
+        UNKNOWN = 0;
+        CRC32 = 1;
+    }
+
+    optional ColumnHashTypePB type = 1;
+    optional int64 modulus = 2;
+    optional int64 remainder = 3;
+    repeated string hash_column_names = 4;
+}
```
alvin-phoenix-ai: This class is too specific.
If a rowset can handle multiple remainders, how can you achieve it?

alvin-phoenix-ai (dismissed):
After the definition has been agreed, I will dismiss the request-change.

gensrc/proto/lake_types.proto
srlch: 1. This is a select predicate, i will get a better name.
2. I'm not pretty sure what general means here. You mean i should introduce general select predicate here instead of a specified one for hash modulus?

gensrc/proto/types.proto
srlch: How about define a wrapper class out of this class if we need multiple remainders or even modulus.

gensrc/proto/lake_types.proto
srlch: `RecordSelectPredicatePB::ColumnHashModulusPredicatePB` something like this? Define ColumnHashModulusPredicatePB as a subtype of RecordSelectPredicatePB?

gensrc/proto/types.proto
srlch: I think another option maybe introduce repeated `<modulus, remainder>` pair and `OpType` for `AND/OR`

gensrc/proto/lake_types.proto
alvin-phoenix-ai: > I'm not pretty sure what general means here. You mean i should introduce general select predicate here instead of a specified one for hash modulus?

Yes. A base class should be put here.

srlch: Get it, i will change the definition and introduce the base class

srlch: Updated, introduce `RecordPredicatePB` as a base class which is a kinds of predicate only for storage layer. `SELECT/FILTER` means it expect record should be selected or filtered by this predicate. Besides, i introduce `RecordPredicateTreeNodePB` structure to represent any kinds of combination of `RecordPredicatePB`.

gensrc/proto/types.proto
srlch: Introduce tree structure for such predicate, i think keep very specified predicate simpler is a better choice? We can build a tree to achieve this goal.

gensrc/proto/types.proto
```diff
@@ -181,6 +181,40 @@ message IsNullPredicatePB {
     optional bool is_not_null = 2;
 }
 
+message RecordPredicateTreeNodePB {
+    NodeType {
+        AND = 0;
+        OR = 1;
+        LEAF 2;
+    }
```
alvin-phoenix-ai: This should be one kind of predicate and be placed in RecordPredicatePB

gensrc/proto/types.proto
```diff
@@ -181,6 +181,40 @@ message IsNullPredicatePB {
...
+
+    optional NodeType type = 1;
+    repeated RecordPredicatePB children = 2;
+}
+
+message RecordPredicatePB {
```
alvin-phoenix-ai, line 184:
```suggestion
message RecordPredicateNodePB {
```

gensrc/proto/types.proto
```diff
@@ -181,6 +181,40 @@ message IsNullPredicatePB {
     optional bool is_not_null = 2;
 }
 
+message RecordPredicateTreeNodePB {
```
alvin-phoenix-ai:
```suggestion
message RecordPredicatePB {
```

alvin-phoenix-ai (commented):
You can reference how TExpr is defined.

gensrc/proto/types.proto
srlch: Oh, i miss a commit here, updated. I try to use recursive definition here to define tree node, and only set its record predicate when the node is leaf. I think it is more readable.

srlch: Besides, i remove the SELECT/FILTER type and introduce NOT nodeType to represent it

srlch: Any suggestion for this updated?

srlch: Updated, simplify the definition.

gensrc/proto/types.proto
```diff
@@ -181,6 +181,31 @@ message IsNullPredicatePB {
     optional bool is_not_null = 2;
 }
 
+message RecordPredicatePB {
+    enum RecordPredicateTypePB {
+        UNKNOWN = 0;
+        HASH_MODULUS = 1;
+    }
+
+    message ColumnHashModulusPredicatePB {
```
alvin-phoenix-ai: image

image
`ColumnisCongruent` may be a better name

srlch: `ColumnHashIsCongruentPredicatePB`? Seems should emphasize that the dividend is the hash of columns.

alvin-phoenix-ai: ColumnHashIsCongruentPB

srlch: Updated

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+    virtual Status get_column_ids(const Schema& schema, std::set<ColumnId>* column_ids) const { return Status::OK(); }
+
+protected:
+    RecordPredicatePB::RecordPredicateTypePB _pred_type;
+    std::vector<RecordPredicateSPtr> _children;
+    bool _inited;
+};
+
+using RecordPredicateSPtr = std::shared_ptr<RecordPredicate>;
```
alvin-phoenix-ai: I don't think shared_ptr is necessary. std::unique_ptr is sufficient.

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+
+    virtual bool empty() const = 0;
+
+    // Some type of RecordPredicate may depand on a specific chunk schema. Such type of predicate need
+    // override the following method.
+    virtual bool check_valid_schema(const Schema& chunk_schema) const { return true; }
+    virtual Status get_column_ids(const TabletSchemaCSPtr& tablet_schema, std::set<ColumnId>* column_ids) const {
+        return Status::OK();
+    }
+    virtual Status get_column_ids(const Schema& schema, std::set<ColumnId>* column_ids) const { return Status::OK(); }
```
alvin-phoenix-ai: We'd better to keep base class/interface as simple as possible, so these functions are better moved to the derived class.

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+    RecordPredicate(const RecordPredicatePB& record_predicate_pb);
+    virtual ~RecordPredicate() = default;
+
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection) const = 0;
+
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const = 0;
+
+    virtual Status init(const RecordPredicatePB& record_predicate_pb) = 0;
+
+    virtual bool empty() const = 0;
```
alvin-phoenix-ai: What does empty mean for a predicate?

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+#include "storage/tablet_schema.h"
+
+namespace starrocks {
+/*
+ * RecordPredicate is an abstract base class for predicates on chunk of data. All RecordPredicate
+ * can be only evaluated in storage layer.
+*/
+class RecordPredicate {
+public:
+    using RecordPredicateSPtr = std::shared_ptr<RecordPredicate>;
```
alvin-phoenix-ai: This is declared twice.

be/src/storage/rowset/segment_iterator.cpp
```diff
@@ -1716,6 +1719,32 @@ StatusOr<uint16_t> SegmentIterator::_filter_by_expr_predicates(Chunk* chunk, vec
     return chunk_size;
 }
 
+StatusOr<uint16_t> SegmentIterator::_filter_by_record_predicate(Chunk* chunk, vector<rowid_t>* rowid) {
```
alvin-phoenix-ai: I think you can move this change in read path to a separate PR and to keep the current PR as the introduction of RecordPredicate

gensrc/proto/types.proto
```diff
@@ -181,6 +181,31 @@ message IsNullPredicatePB {
...
+
+        optional ColumnHashTypePB type = 1;
+        optional int64 modulus = 2;
+        optional int64 remainder = 3;
+        repeated string column_names = 4;
+    }
+
+    optional RecordPredicateTypePB type = 1;
+    // set by some type of predicate (RecordPredicateTypePB is AND/OR/NOT...) which will be implemented in the future
+    repeated RecordPredicatePB children = 2;
```
xiangguangyxg, line 198: not used now, could be added in future when needed.

be/src/storage/record_predicate/record_predicate.h
srlch: I think shared_ptr will be more easier? Because this pointer will be set in `SegmentOption` which is a copyable class. But if this pointer is unique_ptr, `SegmentOption` will be  un-copyable and we need to handle this case.

be/src/storage/record_predicate/record_predicate.h
srlch: Removed from base class. But i add a general interface in Helper class to these three function which make the caller more easier to handle.

be/src/storage/record_predicate/record_predicate.h
srlch: Removed, empty is `ColumnPredicate` style API which means all member of this predicate is empty. But it seems that it doest not need it because record predicate will always inited from a valid PB.

be/src/storage/record_predicate/record_predicate.h
srlch: Fixed

be/src/storage/rowset/segment_iterator.cpp
srlch: Fixed

gensrc/proto/types.proto
srlch: Fixed

be/src/storage/record_predicate/record_predicate.h
alvin-phoenix-ai: In SegmentOption it can be a shared-ptr, but children should be unique_ptr

srlch: Fixed for the children definition

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,66 @@
...
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const = 0;
+
+    virtual Status init(const RecordPredicatePB& record_predicate_pb) = 0;
+
+    RecordPredicatePB::RecordPredicateTypePB type() { return _pred_type; }
+
+protected:
+    RecordPredicatePB::RecordPredicateTypePB _pred_type;
+    std::vector<RecordPredicateUPtr> _children;
+    bool _inited;
```
xiangguangyxg: remove ?

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,66 @@
...
+};
+
+class RecordPredicateHelper {
+public:
+    static StatusOr<RecordPredicateSPtr> create(const RecordPredicatePB& record_predicate_pb);
```
xiangguangyxg:
```suggestion
    static StatusOr<RecordPredicateUPtr> create(const RecordPredicatePB& record_predicate_pb);
```

be/src/storage/record_predicate/record_predicate.h
srlch: Updated

be/src/storage/record_predicate/record_predicate.h
srlch: Updated

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+    virtual std::vector<std::string> column_names() const = 0;
+
+    RecordPredicatePB::RecordPredicateTypePB type() { return _pred_type; }
+
+protected:
+    RecordPredicatePB::RecordPredicateTypePB _pred_type;
+    std::vector<RecordPredicateUPtr> _children;
+};
+
+class RecordPredicateHelper {
```
alvin-phoenix-ai: Could you move it to another file? The base class won't change too much, but the helper will change a lot.

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <vector>
+
+#include "common/status.h"
+#include "storage/tablet_schema.h"
```
alvin-phoenix-ai: Try to use include headers in the header file as little as possible; please use declarations.

be/src/storage/record_predicate/column_hash_is_congruent.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    const uint32_t* hashes_data = hashes.data();
+    for (size_t i = from; i < to; ++i) {
+        selection[i] = (hashes_data[i - from] % _modulus == _remainder);
+    }
+    return Status::OK();
+}
+
+Status ColumnHashIsCongruent::init(const RecordPredicatePB& record_predicate_pb) {
+    const ColumnHashIsCongruentMeta& column_hash_is_congruent_pb = record_predicate_pb.column_hash_is_congruent();
```
alvin-phoenix-ai:
```suggestion
    const auto& column_hash_is_congruent_pb = record_predicate_pb.column_hash_is_congruent();
```

be/src/storage/record_predicate/column_hash_is_congruent.h
```diff
@@ -0,0 +1,52 @@
...
+    ColumnHashIsCongruent(const RecordPredicatePB& record_predicate_pb);
+    ~ColumnHashIsCongruent() = default;
+
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection) const override;
+
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const override;
+
+    virtual Status init(const RecordPredicatePB& record_predicate_pb) override;
+
+    virtual std::vector<std::string> column_names() const override { return _column_names; }
```
alvin-phoenix-ai: `virtual` is not necessary.

be/src/storage/record_predicate/column_hash_is_congruent.h
```diff
@@ -0,0 +1,52 @@
...
+ * 1. Compute the hash value of the specified columns.
+ * 2. Check if the hash value modulo a given modulus equals a specified remainder.
+*/
+class ColumnHashIsCongruent final : public RecordPredicate {
+public:
+    using ColumnHashIsCongruentMeta = RecordPredicatePB::ColumnHashIsCongruentPB;
+    using HashType = RecordPredicatePB::ColumnHashIsCongruentPB::ColumnHashTypePB;
+
+    ColumnHashIsCongruent(const RecordPredicatePB& record_predicate_pb);
+    ~ColumnHashIsCongruent() = default;
```
alvin-phoenix-ai:
```suggestion
    ~ColumnHashIsCongruent() override = default;
```

be/test/storage/record_predicate/record_predicate_test.cpp
alvin-phoenix-ai, line 1: You should have at least two UT files.
1. for column-hash-is-congruent
2. for helper.

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+protected:
+    RecordPredicatePB::RecordPredicateTypePB _pred_type;
+    std::vector<RecordPredicateUPtr> _children;
+};
+
+class RecordPredicateHelper {
+public:
+    static StatusOr<RecordPredicateUPtr> create(const RecordPredicatePB& record_predicate_pb);
+
+    static bool check_valid_schema(const RecordPredicate& predicate, const Schema& chunk_schema);
```
alvin-phoenix-ai: Status is preferred because more information can be returned

be/src/storage/record_predicate/record_predicate.h
```diff
@@ -0,0 +1,67 @@
...
+using RecordPredicateUPtr = std::unique_ptr<RecordPredicate>;
+
+class RecordPredicate {
+public:
+    RecordPredicate(const RecordPredicatePB& record_predicate_pb);
+    virtual ~RecordPredicate() = default;
+
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection) const = 0;
+
+    virtual Status evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const = 0;
```
alvin-phoenix-ai: Please add comments for these two functions to explain how to use them

be/src/storage/record_predicate/record_predicate.cpp
```diff
@@ -0,0 +1,101 @@
...
+    RETURN_IF_ERROR(RecordPredicateHelper::_check_valid_pb(record_predicate_pb));
+    RecordPredicateUPtr predicate;
+    switch (record_predicate_pb.type()) {
+    case RecordPredicatePB::COLUMN_HASH_IS_CONGRUENT: {
+        predicate = std::make_unique<ColumnHashIsCongruent>(record_predicate_pb);
+        RETURN_IF_ERROR(predicate->init(record_predicate_pb));
+        break;
+    }
+    default:
+        return Status::InternalError("Unknown record predicate type: " +
```
alvin-phoenix-ai: should use fmt::format to construct string rather than `operator+`

be/src/storage/record_predicate/record_predicate.h
srlch: Fixed

be/src/storage/record_predicate/record_predicate.h
srlch: Fixed, try to use declarations can using include in `cpp` if it is necessary.

be/src/storage/record_predicate/column_hash_is_congruent.cpp
srlch: Fixed

be/src/storage/record_predicate/column_hash_is_congruent.h
srlch: Fixed

be/src/storage/record_predicate/column_hash_is_congruent.h
srlch: Fixed

be/test/storage/record_predicate/record_predicate_test.cpp
srlch: Fixed

be/src/storage/record_predicate/record_predicate.h
srlch: Fixed

be/src/storage/record_predicate/record_predicate.cpp
srlch: Fixed

gensrc/proto/types.proto
```diff
@@ -181,6 +181,31 @@ message IsNullPredicatePB {
...
+        COLUMN_HASH_IS_CONGRUENT = 1;
+    }
+
+    message ColumnHashIsCongruentPB {
+        enum ColumnHashTypePB {
+            UNKNOWN = 0;
+            CRC32 = 1;
+        }
+
+        optional ColumnHashTypePB type = 1;
```
xiangguangyxg: not used ?

be/src/storage/record_predicate/column_hash_is_congruent.h
```diff
@@ -0,0 +1,50 @@
...
+    Status evaluate(Chunk* chunk, uint8_t* selection, uint16_t from, uint16_t to) const override;
+
+    Status init(const RecordPredicatePB& record_predicate_pb) override;
+
+    const std::vector<std::string>* column_names() const override { return &_column_names; }
+
+private:
+    Status _check_valid_pb(const ColumnHashIsCongruentMeta& column_hash_is_congruent_meta) const;
+
+    HashType _hash_type;
```
xiangguangyxg: not used ?

gensrc/proto/types.proto
srlch: fixed

be/src/storage/record_predicate/column_hash_is_congruent.h
srlch: fixed

xiangguangyxg (dismissed):

gensrc/proto/lake_types.proto
```diff
@@ -110,6 +110,7 @@ message RowsetMetadataPB {
     // only be set > 0 if `overlapped` is true
     optional uint32 next_compaction_offset = 13;
     repeated int64 bundle_file_offsets = 14;
+    optional RecordPredicatePB record_predicate = 15;
```
xiangguangyxg: should it be repeated ?

alvin-phoenix-ai: I think optional is good enough, if there are more predicates, we can implement it by a compound predicate.

luohaha (approved):

wangruin (approved):

xiangguangyxg (approved):

wyb (approved):

PR #60001, opened 2025-06-17
[BugFix] Support switch Auth Plugin base on user's authentication type

### LINKED ISSUE #56994: Support switch Auth Plugin base on user's authentication type
## Feature request

**Additional context**

xhumanoid:
## Why I'm doing:

similar issue for ldap user described in #56994

 in current logic we need to ask users to specify connection parameter `defaultAuthenticationPlugin=mysql_clear_password`
 otherwise we will have password hash instead of cleartext password for ldap

without parameter account will be blocked after few attempts to check in ldap,
because we try to use native_password string as password for ldap

with patch we detect ldap users early and ask user to switch to `mysql_clear_password`
and auth finished correctly with/without explicit parameter

3.5+ version have reworked logic for handshake where this issue was fixed
so need to fix only in 3.3 and 3.4

## What I'm doing:

users with AUTHENTICATION_LDAP_SIMPLE and AUTHENTICATION_LDAP_SIMPLE_FOR_EXTERNAL now detected as ldap in `isLDAPUser` function

Fixes #56994

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/mysql/MysqlProto.java
+++ b/fe/fe-core/src/main/java/com/starrocks/mysql/MysqlProto.java
@@ -42,6 +42,7 @@ import com.starrocks.common.DdlException;
 import com.starrocks.common.ErrorCode;
 import com.starrocks.common.ErrorReport;
 import com.starrocks.common.Pair;
+import com.starrocks.mysql.privilege.AuthPlugin;
 import com.starrocks.mysql.ssl.SSLContextLoader;
 import com.starrocks.qe.ConnectContext;
 import com.starrocks.server.GlobalStateMgr;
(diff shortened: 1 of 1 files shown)


alvin-phoenix-ai:
@xhumanoid 
All PR should be merged into main branch first. Could you create a new PR for main branch?

xhumanoid:
@alvin-celerdata this code doesn't exist in main and 3.5+, because was reworked to support Security Integrations

issue still presented in 3.4 and 3.5 only

code was refactored in #56556 and later fully reimplemented with fixes

alvin-phoenix-ai:
@xhumanoid 
> this code doesn't exist in main and 3.5+, because was reworked to support Security Integrations
> issue still presented in 3.4 and 3.5 only

Got cha, Because v3.4 is released version, we won't introduce new feature to a released version.
If you want to have this new feature, I suggest you to upgrade to v3.5

xhumanoid:
not sure is a new feature

mark as feature just because similar functionality in 3.5 marked as feature
but in reality 3.5+ didn't have isLDAP check after refactoring, so it was fully new functionality
for 3.3 + 3.4 we have isLDAP check, but it work incorrect

it's fix of isLDAP function when we try to detect isLDAP user or not

we do only check
```
        // If the user can not be found in local, and there is more than 1 auth type in authentication_chain.
        // It is speculated that the user may be a ldap user.
        return localUser == null && Config.authentication_chain.length > 1;
```

but user can be created with 

```
CREATE USER tom IDENTIFIED WITH authentication_ldap_simple
```

and `isLDAP` function will return is not LDAP user, don't do authPlugin switch, but later detect user is `authentication_ldap_simple` and try to go into ldap

but because plugin stay `native_password` you will try hash instead of password and lock the users

HangyuanLiu (approved):

PR #60022, opened 2025-06-18
[Enhancement] optimize count(1) for iceberg table

### LINKED ISSUE #46525: Optimize Iceberg table count
## Enhancement

One can obtain the `count(*)` for an iceberg table from the Iceberg metadata without having to do a full scan of the data. Currently, Starrocks performs a full scan of the Iceberg Table data when doing a count(*) query on external Iceberg lake table. This should be optimized to just use the Iceberg metadata (this is already available via the statistics).

E.g.

```
StarRocks > explain select count(*) as cnt from tbl1;
+-------------------------------------------+
| Explain String                            |
+-------------------------------------------+
| PLAN FRAGMENT 0                           |
|  OUTPUT EXPRS:19: count                   |
|   PARTITION: UNPARTITIONED                |
|                                           |
|   RESULT SINK                             |
|                                           |
|   4:AGGREGATE (merge finalize)            |
|   |  output: count(19: count)             |
|   |  group by:                            |
|   |                                       |
|   3:EXCHANGE                              |
|                                           |
| PLAN FRAGMENT 1                           |
|  OUTPUT EXPRS:                            |
|   PARTITION: RANDOM                       |
|                                           |
|   STREAM DATA SINK                        |
|     EXCHANGE ID: 03                       |
|     UNPARTITIONED                         |
|                                           |
|   2:AGGREGATE (update serialize)          |
|   |  output: count(*)                     |
|   |  group by:                            |
|   |                                       |
|   1:Project                               |
|   |  <slot 21> : 1                        |
|   |                                       |
|   0:IcebergScanNode                       |
|      TABLE: iceberg.db.tbl1 |
|      cardinality=13219153                 |
|      avgRowSize=2.0                       |
+-------------------------------------------+
```

The `cardinality` in the `IcebergScanNode` already has the result it does not need to perform any scan.

dirtysalt:
## Why I'm doing:

Since we have `file_record_count` of each file from iceberg metadata, we can use that info instead of reading data from file.

## What I'm doing:

This PR does:
1. pass `file_record_count` from iceberg metadata down to scan range
2. if we spot the `count(1)` sql pattern and we have `file_record_count` field, then we can use that.
3. but only when scan range offset is 0, otherwise we will double-count for a file being splitted into multiple scan ranges.
4. remove `canUseAnyColumn` since we don't need it any more. 

Fixes #46525 

--------

Take following query for example on tpcds1T
> select count(1) from iceberg.zz_iceberg_tpcds_sf1000_iceberg_parquet_lz4.store_sales;

Before this pr, we can see from the profile that we read like 800MB data

image

And after this pr, we don't read any data

image

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exec/hdfs_scanner.cpp
+++ b/be/src/exec/hdfs_scanner.cpp
@@ -190,8 +190,11 @@ Status HdfsScanner::_build_scanner_context() {
             new RuntimeScanRangePruner(predicate_parser, ctx.conjuncts_manager->unarrived_runtime_filters()));
 
     RETURN_IF_ERROR(ctx.update_return_count_columns());
-    if (ctx.scan_range->__isset.file_record_count && ctx.scan_range->delete_files.empty()) {
+    if (ctx.scan_range->__isset.record_count && ctx.scan_range->delete_files.empty()) {
         ctx.can_use_file_record_count = true;
+    }
+    if (ctx.scan_range->__isset.is_first_split && ctx.scan_range->is_first_split) {
+        ctx.is_first_split = true;
     }
     return Status::OK();
 }
@@ -206,8 +209,8 @@ Status HdfsScanner::get_next(RuntimeState* runtime_state, ChunkPtr* chunk) {
             return Status::EndOfFile("");
         }
         int64_t file_record_count = 0;
-        if (_scanner_ctx.scan_range->offset == 0) {
-            file_record_count = _scanner_ctx.scan_range->file_record_count;
+        if (_scanner_ctx.is_first_split) {
+            file_record_count = _scanner_ctx.scan_range->record_count;
         }
         _scanner_ctx.append_or_update_count_column_to_chunk(chunk, file_record_count);
         _scanner_ctx.append_or_update_partition_column_to_chunk(chunk, 1);
--- a/be/src/exec/hdfs_scanner.cpp
+++ b/be/src/exec/hdfs_scanner.cpp
@@ -193,8 +193,8 @@ Status HdfsScanner::_build_scanner_context() {
     if (ctx.scan_range->__isset.record_count && ctx.scan_range->delete_files.empty()) {
         ctx.can_use_file_record_count = true;
     }
-    if (ctx.scan_range->__isset.is_first_split && ctx.scan_range->is_first_split) {
-        ctx.is_first_split = true;
+    if (ctx.scan_range->__isset.is_first_split) {
+        ctx.is_first_split = ctx.scan_range->is_first_split;
     }
     return Status::OK();
 }
@@ -203,11 +203,12 @@ Status HdfsScanner::get_next(RuntimeState* runtime_state, ChunkPtr* chunk) {
     SCOPED_RAW_TIMER(&_total_running_time);
     RETURN_IF_CANCELLED(_runtime_state);
 
+    if (_scanner_ctx.no_more_chunks) {
+        return Status::EndOfFile("");
+    }
+
     // short circuit for ___count___ optimization.
     if (_scanner_ctx.return_count_column && _scanner_ctx.can_use_file_record_count) {
-        if (_scanner_ctx.emit_count_column) {
-            return Status::EndOfFile("");
-        }
         int64_t file_record_count = 0;
         if (_scanner_ctx.is_first_split) {
             file_record_count = _scanner_ctx.scan_range->record_count;
@@ -215,7 +216,7 @@ Status HdfsScanner::get_next(RuntimeState* runtime_state, ChunkPtr* chunk) {
         _scanner_ctx.append_or_update_count_column_to_chunk(chunk, file_record_count);
         _scanner_ctx.append_or_update_partition_column_to_chunk(chunk, 1);
         _scanner_ctx.append_or_update_extended_column_to_chunk(chunk, 1);
-        _scanner_ctx.emit_count_column = true;
+        _scanner_ctx.no_more_chunks = true;
         _app_stats.rows_read += 1;
         return Status::OK();
     }
--- a/be/src/exec/hdfs_scanner.cpp
+++ b/be/src/exec/hdfs_scanner.cpp
@@ -124,19 +124,13 @@ Status HdfsScanner::_build_scanner_context() {
     // build columns of materialized and partition.
     for (size_t i = 0; i < _scanner_params.materialize_slots.size(); i++) {
         auto* slot = _scanner_params.materialize_slots[i];
-
-        // if `can_use_any_column`, we can set this column to non-existed column without reading it.
-        if (_scanner_params.can_use_any_column) {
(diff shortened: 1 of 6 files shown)


be/src/exec/hdfs_scanner.cpp
```diff
@@ -549,35 +575,27 @@ Status HdfsScannerContext::append_or_update_not_existed_columns_to_chunk(ChunkPt
...
         col->assign(row_count, 0);
         ck->append_or_update_column(std::move(col), slot_desc->id());
-    } else {
-        for (auto* slot_desc : not_existed_slots) {
-            auto col = ColumnHelper::create_column(slot_desc->type(), slot_desc->is_nullable());
-            if (row_count > 0) {
-                col->append_default(row_count);
-            }
-            ck->append_or_update_column(std::move(col), slot_desc->id());
+        ck->set_num_rows(row_count);
```
zombee0, line 593: how to deal with position delete in iceberg?

dirtysalt: @zombee0  you are right. this solution does not work well with position delete files. I need to add a rule at FE side to avoid that.

it works well with eq-delete because eq-delete file requires some materialized column, so FE will not rewrite using this rule.

dirtysalt: @zombee0 fixed

gensrc/thrift/PlanNodes.thrift
```diff
@@ -426,6 +426,8 @@ struct THdfsScanRange {
     31:optional TDeletionVectorDescriptor deletion_vector_descriptor
 
     32: optional string candidate_node
+
+    33: optional i64 file_record_count
```
alvin-phoenix-ai: Could you add some comments for this field?
I'm wondering why you chose `file_record_count` instead of `record_count`?

dirtysalt: OK. 
because there is a `file_path`. I don't mind changing name.  I can change it if you think it's more concise.

zombee0 (approved):

gensrc/thrift/PlanNodes.thrift
alvin-phoenix-ai: I like record_count, but it doesn't matter.
However, I think you should add some comments, such as the meaning and whether it is available all the time.

Samrose-Ahmed:
This PR is great improvement and simple. but we can maybe even do it completely in FE?, 

because iirc this is  getting the list of files from the iceberg metadata, then creating kinda dummy hdfs scan ranges that include the file record count and sending them to BE's which just short circuit. The overhead might be negligent but idk for many files and creating sending bunch of scan ranges for whats just a metadata entry often cached in FE memory.

Another approach since we have metadata tables  implemented is you can basically rewrite `select count(1) from mytbl` to `select sum(record_count) from mytbl$files` right?

dirtysalt:
> This PR is great improvement and simple. but we can maybe even do it completely in FE?,
> 
> because iirc this is getting the list of files from the iceberg metadata, then creating kinda dummy hdfs scan ranges that include the file record count and sending them to BE's which just short circuit. The overhead might be negligent but idk for many files and creating sending bunch of scan ranges for whats just a metadata entry often cached in FE memory.

That will complicate the whole architecture: embedding some computation into FE.

> Another approach since we have metadata tables implemented is you can basically rewrite `select count(1) from mytbl` to `select sum(record_count) from mytbl$files` right?

Even by that, it's still done by doing execution at BE sides.

be/src/exec/hdfs_scanner.cpp
```diff
@@ -196,12 +189,37 @@ Status HdfsScanner::_build_scanner_context() {
     ctx.rf_scan_range_pruner = opts.obj_pool->add(
             new RuntimeScanRangePruner(predicate_parser, ctx.conjuncts_manager->unarrived_runtime_filters()));
 
+    RETURN_IF_ERROR(ctx.update_return_count_columns());
+    if (ctx.scan_range->__isset.record_count && ctx.scan_range->delete_files.empty()) {
+        ctx.can_use_file_record_count = true;
+    }
+    if (ctx.scan_range->__isset.is_first_split && ctx.scan_range->is_first_split) {
+        ctx.is_first_split = true;
```
trueeyu:
```   
 if (ctx.scan_range->__isset.is_first_split) {
        ctx.is_first_split = ctx.scan_range->is_first_split;
    }
```

be/src/exec/hdfs_scanner.h
```diff
@@ -343,11 +342,12 @@ struct HdfsScannerContext {
 
     bool orc_use_column_names = false;
 
-    bool can_use_any_column = false;
-
     bool can_use_min_max_count_opt = false;
 
+    bool is_first_split = false;
     bool return_count_column = false;
+    bool emit_count_column = false;
```
trueeyu: What is the purpose of the emit_count_column?

dirtysalt: it's just local variable.

because when calling get_next, we have two states:
1. output chunk (with count value)
2. output eof

be/src/exec/hdfs_scanner.cpp
dirtysalt: updated

gensrc/thrift/PlanNodes.thrift
dirtysalt: updated

feipool (approved):

trueeyu (dismissed):

trueeyu (approved):

be/src/exec/hdfs_scanner.cpp
```diff
@@ -196,12 +189,38 @@ Status HdfsScanner::_build_scanner_context() {
...
 Status HdfsScanner::get_next(RuntimeState* runtime_state, ChunkPtr* chunk) {
     SCOPED_RAW_TIMER(&_total_running_time);
     RETURN_IF_CANCELLED(_runtime_state);
+
+    if (_scanner_ctx.no_more_chunks) {
+        return Status::EndOfFile("");
+    }
+
+    // short circuit for ___count___ optimization.
+    if (_scanner_ctx.return_count_column && _scanner_ctx.can_use_file_record_count) {
```
stephen-shelby, line 211: I think we could provide a session variables to control it. record_count from metadata may be incorrect.

dirtysalt: `record_count` in iceberg metadata is required and correct.

stephen-shelby (approved):

PR #60035, opened 2025-06-18
[Enhancement] MV union self support multiple expressions

### LINKED ISSUE #59203: Using UNION ALL for self-joins may produce inaccurate results.
When performing a self-join query, the results are sometimes incorrect.

```sql
CREATE TABLE `par_tbl1` (
  `datekey` datetime NOT NULL COMMENT "",
  `item_id` varchar(65533) NOT NULL COMMENT "",
  `v1` int(11) NULL COMMENT ""
) ENGINE=OLAP 
PRIMARY KEY(`datekey`, `item_id`)
COMMENT "OLAP"
PARTITION BY date_trunc('day', `datekey`)
DISTRIBUTED BY HASH(`datekey`, `item_id`)

INSERT INTO par_tbl1 (datekey, item_id, v1) VALUES ('2025-01-02 00:00:00', '1', 2);
INSERT INTO par_tbl1 (datekey, item_id, v1) VALUES ('2025-01-01 00:00:00', '1', 2);

# union all
select date_trunc("day", a.datekey) as p_time, sum(a.v1) as value
from par_tbl1 a
group by p_time, a.item_id
union all
select date_trunc('day', date_add(b.datekey, INTERVAL 1 DAY)) AS p_time, sum(b.v1) as value
from par_tbl1 b
group by p_time, item_id
;
```

| p_time              | value |
|---------------------|-------|
| 2025-01-01 00:00:00 |     2 |
| 2025-01-02 00:00:00 |     2 |
| 2025-01-02 00:00:00 |     2 |
| 2025-01-03 00:00:00 |     2 |

### Real behavior (Required)

Image

| p_time              | value |
|---------------------|-------|
| 2025-01-01 00:00:00 |     2 |
| 2025-01-02 00:00:00 |     2 |
| 2025-01-02 00:00:00 |     2 |

```sql
select date_trunc("day", a.datekey) as p_time, sum(a.v1) as value
from par_tbl1 a
group by p_time, a.item_id
```

Image

```sql
select date_trunc('day', date_add(b.datekey, INTERVAL 1 DAY)) AS p_time, sum(b.v1) as value
from par_tbl1 b
group by p_time, item_id
```

Image

If I execute these two SQL statements separately by hand, the results are correct, but if I use `union all`, the result is incorrect.

----

Image

If I manually clear the data and then rewrite it, the result is correct.

 - 3.3.x
 - 3.4.x

crossoverJie:
## Why I'm doing:

```sql
CREATE TABLE `par_tbl1` (
  `datekey` datetime NOT NULL COMMENT "",
  `item_id` varchar(65533) NOT NULL COMMENT "",
  `v1` int(11) NULL COMMENT ""
) ENGINE=OLAP 
PRIMARY KEY(`datekey`, `item_id`)
COMMENT "OLAP"
PARTITION BY date_trunc('day', `datekey`)
DISTRIBUTED BY HASH(`datekey`, `item_id`)

INSERT INTO par_tbl1 (datekey, item_id, v1) VALUES ('2025-01-02 00:00:00', '1', 2);
INSERT INTO par_tbl1 (datekey, item_id, v1) VALUES ('2025-01-01 00:00:00', '1', 2);


CREATE
MATERIALIZED VIEW mv_test99
REFRESH ASYNC EVERY(INTERVAL 60 MINUTE)
PARTITION BY p_time
PROPERTIES (
"partition_refresh_number" = "1"
)
AS
select date_trunc("day", a.datekey) as p_time, sum(a.v1) as value
from par_tbl1 a
group by p_time, a.item_id
union all
select date_trunc('day', date_add(b.datekey, INTERVAL 1 DAY)) AS p_time, sum(b.v1) as value
from par_tbl1 b
group by p_time, item_id
;
```

The query result of this view is wrong because one partition is refreshed less.

```sql
show partitions from mv_test99;
```

image

Partition `p20250103_20250104` is missing.

## What I'm doing:

Fixes #59203

1. When creating an MV, if a `UnionRelation` exists, save these expressions for calculating **virtual partitions**.  
2. When refreshing the MV, manually add the previously saved virtual partitions.

`Virtual partitions`: Partitions calculated based on the base table's partitions and `date_add`/`date_sub` function expressions, used for refreshing additional partitions. 

They only apply to materialized views  and do not affect the base table.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


crossoverJie:
@LiShuMing Please take a look.

fe/fe-core/src/main/java/com/starrocks/sql/common/SyncPartitionUtils.java
```diff
@@ -518,31 +525,75 @@ public static LocalDateTime getUpperDateTime(LocalDateTime upperDateTime, String
...
     @NotNull
-    public static LocalDateTime nextUpperDateTime(LocalDateTime upperDateTime, String granularity) {
+    public static LocalDateTime nextUpperDateTime(LocalDateTime upperDateTime, String granularity, String dateFunc) {
         LocalDateTime truncUpperDateTime;
         switch (granularity) {
             case MINUTE:
-                truncUpperDateTime = upperDateTime.plusMinutes(1).withNano(0).withSecond(0);
+                if (dateFunc.equalsIgnoreCase(FunctionSet.DATE_ADD)) {
+                    truncUpperDateTime = upperDateTime.plusMinutes(1).withNano(0).withSecond(0);
+                } else if (dateFunc.equalsIgnoreCase(FunctionSet.DATE_SUB)) {
```
LiShuMing: It's better to split different `dateFunc`function implementations into different functions.

fe/fe-core/src/main/java/com/starrocks/sql/common/SyncPartitionUtils.java
```diff
@@ -122,20 +122,25 @@ public static PartitionDiff getRangePartitionDiffOfExpr(Map<String, Range<Partit
...
         } else if (functionCallExpr.getFnName().getFunction().equalsIgnoreCase(FunctionSet.STR2DATE)) {
             rollupRange = mappingRangeListForDate(baseRangeMap);
         }
-        return getRangePartitionDiff(mvRangeMap, rollupRange, differ);
+        return getRangePartitionDiff(mvRangeMap, rollupRange, virtualRange, differ);
     }
 
-    public static Map<String, Range<PartitionKey>> toMappingRanges(Map<String, Range<PartitionKey>> baseRangeMap,
+    public static MVRangePartitionMapper.PartitionRangeWrapper toMappingRanges(Map<String, Range<PartitionKey>> baseRangeMap,
                                                                    String granularity,
```
LiShuMing, line 140: to align?

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
```diff
@@ -560,6 +561,12 @@ public String toString() {
     // and in each round of checkpoint
     private boolean reloaded = false;
 
+    // This is used to store the virtual partition of union expressions.
+    private transient Map<String, Range<PartitionKey>> virtualPartitionMapping;
```
LiShuMing, line 565: What's the concept of `virtualPartition`? It may be a bit confused and complex for mv refresh.

LiShuMing: Can you add more documents about this, and is this concept only for `union`? How about self-join or other scenes?

crossoverJie: > What is the concept of `virtualPartition`? It may be a bit confusing and complex for MV refresh.

The background of this PR is that our union table uses date expressions (`DATE_SUB(b.datekey, INTERVAL 3 DAY)/ date_add(b.datekey, INTERVAL 4 DAY)`) which results in the inability to refresh data for `+/-N` partitions during MV refresh.

To solve this problem, when synchronizing `MV` partitions, we calculate the partitions that need to be added using date expressions and then store them in a separate field; since this partition is not an `actual` partition that exists in the MV, it is named as a `virtual` partition.


When calculating the partitions that need to be refreshed for `MVToRefreshedPartitions`, manually add the virtual partitions that were just saved.

https://github.com/StarRocks/starrocks/pull/60035/files#diff-92f7115b4de0c80e52de50e97a67e4d8370aec550d0cc43bf3d66b39e9b9a630R131-R132

This allows for the extra refreshing of newly created partitions through time expressions, while not affecting the MV's original partitions.

crossoverJie: > Can you add more documents about this

Once we reach agreement, I'll add more detailed documentation.

crossoverJie: > and is this concept only for `union`?

This concept can be applied not only to `union`, but also to any scenario where an additional partition refresh is needed. Data can be appended to the `virtualPartition` in these cases.

Currently, other commonly used join operations do not require the use of `virtualPartition`.

> How about self-join or other scenes?

Specifically: 

For `inner join`, the result is the intersection of the two tables. Therefore, adding a date using `date_add` will not affect the query results.

For `left join`, the query result remains based on the left table, regardless of how many new date `date_add` are added to the right table, so it won't affect the query results.

Therefore, I believe it can currently meet the needs of the vast majority of scenarios.

fe/fe-core/src/main/java/com/starrocks/sql/common/SyncPartitionUtils.java
crossoverJie: Done

fe/fe-core/src/main/java/com/starrocks/sql/common/SyncPartitionUtils.java
crossoverJie: What does this mean?

eshishki:
@LiShuMing could you take another look here? @crossoverJie addressed your review on 2025-08-21 — applied the dateFunc split, and replied with the rationale for the `virtualPartition` concept (it's a generic mechanism for "extra refresh partitions beyond what base tables expose", not union-specific). The thread didn't get a follow-up, so it's not clear whether the concept landed for you or whether you'd prefer a different shape.

Worth deciding before a rebase — the IVM refactor (#61857) renamed `MVPCTBasedRefreshProcessor` and reshuffled the partition mapper layer, so porting this PR to today's main is a real chunk of work.

Underlying bug (#59203) is a real correctness issue — MV with `UNION ALL` + `date_add(±N day)` silently drops a partition.

crossoverJie:
@eshishki Thanks for the reminder, I will continue to push this PR later.

PR #60052, opened 2025-06-19
[BugFix] Fix assertion failure for insert stmt planning of shuffle for primary key table

srlch:
## Why I'm doing:
If we create a table satisify the following condition: 1. primary key table, 2. replication_num > 1
3. without replicate storage. In this case, if we execute a partial update insert stmt, we will get
assertion failure because we need to build the shuffle property in this case but if insert stmt is
in partial update mode, the list of ColumnRefOperator has not the same size of full schema.

## What I'm doing:
We just use outputFullSchema which generated by partial update checking instead of Table::getFullSchema()

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/test/sql/test_insert_empty/R/test_insert_partial_update
+++ b/test/sql/test_insert_empty/R/test_insert_partial_update
@@ -68,3 +68,24 @@ insert into test4 (pk, v0) values(1, 'v0_1'), (3, 'v3_1');
 -- result:
 [REGEX].*partial update on table with sort key must provide all sort key columns.*
 -- !result
+-- name: test_non_replicated_storage_multi_replica_pk
+CREATE TABLE `t_non_replicated_storage_multi_replica_pk` (
+  `k` STRING NOT NULL COMMENT "",
+  `v1` int,
+  `v2` int
+) ENGINE=OLAP 
+PRIMARY KEY(`k`)
+DISTRIBUTED BY HASH(`k`) BUCKETS 1
+PROPERTIES (
+"replication_num" = "2",
+"replicated_storage" = "false"
+);
+-- result:
+-- !result
+insert into t_non_replicated_storage_multi_replica_pk (k,v2) select "abc", 10;
+-- result:
+-- !result
+select * from t_non_replicated_storage_multi_replica_pk;
+-- result:
+abc	10	10
+-- !result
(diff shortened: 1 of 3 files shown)


test/sql/test_insert_empty/T/test_insert_partial_update
```diff
@@ -35,3 +35,18 @@ primary KEY (pk) DISTRIBUTED BY HASH(pk) BUCKETS 1 order by (v1) PROPERTIES("rep
...
+  `v2` int
+) ENGINE=OLAP 
+PRIMARY KEY(`k`)
+DISTRIBUTED BY HASH(`k`) BUCKETS 1
+PROPERTIES (
+"replication_num" = "2",
+"replicated_storage" = "false"
+);
+
+insert into t_non_replicated_storage_multi_replica_pk (k,v2) select "abc", 10;
```
wyb, line 52: missing select sql

test/sql/test_insert_empty/R/test_insert_partial_update
```diff
@@ -67,4 +67,25 @@ insert into test4 values(1, 'v0', 1), (2, 'v2', 2);
 insert into test4 (pk, v0) values(1, 'v0_1'), (3, 'v3_1');
 -- result:
 [REGEX].*partial update on table with sort key must provide all sort key columns.*
+-- !result
+-- name: test_non_replicated_storage_multi_replica_pk
+CREATE TABLE `t_non_replicated_storage_multi_replica_pk` (
+  `k` STRING NOT NULL COMMENT "",
+  `v1` int,
```
wyb:
```suggestion
  `v1` int DEFAULT "10",
```

srlch: Fixed

test/sql/test_insert_empty/T/test_insert_partial_update
srlch: `select "abc", 10;` is enough, select two literal

banmoy (approved):

test/sql/test_insert_empty/T/test_insert_partial_update
wyb: `select * from t_non_replicated_storage_multi_replica_pk;`
this sql

srlch: Fixed

wyb (approved):

xiangguangyxg (approved):

PR #60068, opened 2025-06-19
[Feature] support recording last transaction warehouse id in OlapTable

starrocks-xupeng:
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
@@ -90,6 +90,7 @@ import com.starrocks.common.util.Util;
 import com.starrocks.common.util.WriteQuorum;
 import com.starrocks.common.util.concurrent.MarkedCountDownLatch;
 import com.starrocks.lake.DataCacheInfo;
+import com.starrocks.lake.LastTransactionWarehouseInfo;
 import com.starrocks.lake.StarOSAgent;
 import com.starrocks.lake.StorageInfo;
 import com.starrocks.persist.ColocatePersistInfo;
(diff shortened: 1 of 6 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
```diff
@@ -272,6 +274,11 @@ public enum OlapTableState {
     @SerializedName(value = "sessionId")
     protected UUID sessionId = null;
 
+    // stores warehouse info of last transaction for this table,
+    // used in shared-data cluster for lake table and lake mv
+    @SerializedName(value = "lastTransactionWarehouseInfo")
+    protected LastTransactionWarehouseInfo lastTransactionWarehouseInfo = new LastTransactionWarehouseInfo();
```
alvin-phoenix-ai, line 280:
```suggestion
    // stores warehouse info of last transaction for this table,
    // used in shared-data cluster for lake table and lake mv
    @SerializedName(value = "lastTransactionWarehouseInfo")
    protected LastTransactionWarehouseInfo lastTransactionWarehouseInfo = new LastTransactionWarehouseInfo();
```
I don't think recording transaction information in a table is a suitable approach.
The relationship between Transaction and warehouse should be kept in another place rather than in Table struct.

alvin-phoenix-ai (changes requested):

fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
alvin-phoenix-ai: Things in table are all information related to the definition. 
However, warehouse running information is runtime information, which should not be recorded here.

starrocks-xupeng: ok, i will put a map in WarehouseManager to record the last transaction warehouse info for each table, in that case, i will only need to submit this pr to celerdata

PR #60072, opened 2025-06-19
[BugFix] fix partition key shuffle error

Wenjun7J:
## Why I'm doing:

In pipeline, partition key exchanger for iceberg sink should generate the key according to the transform expr.

For example, the day transform result is an integer, but the current implemention will treat the result value as the origin partition column type, such as datetime. Thus to read the result with incorrect type will be undefined.

## What I'm doing:

In local exchange, replace the column_value with the iceberg_column_value interface.

Fixes https://github.com/StarRocks/StarRocksTest/issues/9844

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


be/src/exec/pipeline/exchange/local_exchange.cpp
```diff
@@ -266,10 +266,12 @@ size_t OrderedPartitionExchanger::_find_min_channel_id() {
...
 KeyPartitionExchanger::KeyPartitionExchanger(const std::shared_ptr<ChunkBufferMemoryManager>& memory_manager,
                                              LocalExchangeSourceOperatorFactory* source,
-                                             std::vector<ExprContext*> partition_expr_ctxs, const size_t num_sinks)
+                                             std::vector<ExprContext*> partition_expr_ctxs, const size_t num_sinks,
+                                             const std::vector<std::string>& transform_exprs)
         : LocalExchanger(strings::Substitute("KeyPartition"), memory_manager, source),
           _source(source),
-          _partition_expr_ctxs(std::move(partition_expr_ctxs)) {}
+          _partition_expr_ctxs(std::move(partition_expr_ctxs)),
+          _transform_exprs(std::move(transform_exprs)) {}
```
trueeyu, line 274: useless to `std::move (const auto&)` ?

be/src/connector/utils.cpp
```diff
@@ -112,22 +112,23 @@ StatusOr<std::string> HiveUtils::iceberg_make_partition_name(
 }
 
 StatusOr<std::string> HiveUtils::iceberg_column_value(const TypeDescriptor& type_desc, const ColumnPtr& column,
-                                                      const std::string& transform_expr, int8_t& is_null) {
+                                                      const int i, const std::string& transform_expr, int8_t& is_null) {
```
trueeyu: Why not use `int8_t is_null`?

be/src/connector/utils.cpp
```diff
@@ -112,22 +112,23 @@ StatusOr<std::string> HiveUtils::iceberg_make_partition_name(
...
     std::string value;
     is_null = false;
-    if (column->size() < 1) {
-        return Status::InternalError("column size of extra chunk in make partition name is less than one");
-    } else if (column->is_null(0)) {
+    if (column->size() <= i) {
```
trueeyu: column->size() <= (i+1) ?

be/src/connector/utils.cpp
trueeyu: index or idx may be better?

Wenjun7J: changed to idx

be/src/connector/utils.cpp
Wenjun7J: `0 <= idx < column->size()` is valid,  so `idx>=column->size()` is invalid.

be/src/connector/utils.cpp
Wenjun7J: is_null flag is an output, and will be used outside.

be/src/exec/pipeline/exchange/local_exchange.cpp
Wenjun7J: delete `const &` symbol

be/src/exec/pipeline/exchange/local_exchange.cpp
```diff
@@ -297,15 +299,23 @@ Status KeyPartitionExchanger::accept(const ChunkPtr& chunk, const int32_t sink_d
         partition_columns.push_back(std::move(partition_column));
     }
 
-    std::unordered_map<std::vector<std::string>, std::vector<uint32_t>> key2indices;
-    std::unordered_map<std::vector<std::string>, std::vector<std::pair<TypeDescriptor, ColumnPtr>>> key2datum;
+    std::map<std::vector<std::optional<std::string>>, std::vector<uint32_t>> key2indices;
```
trueeyu, line 302: Why changed to `std::map`?

trueeyu (approved):

Seaven (approved):

PR #60073, opened 2025-06-19
[BugFix] LakeRollup should use physical partition ID instead of partition ID.

sevev:
## Why I'm doing:
In shared-data mode, the allocation of `shardGroup` and `shard` should correspond to the physical partition ID, rather than the partition ID.

## What I'm doing:

Fixes https://github.com/StarRocks/starrocks/issues/60184

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
@@ -74,7 +74,7 @@ public class LakeTableRollupBuilder extends AlterJobV2Builder {
                 long physicalPartitionId = physicalPartition.getId();
                 // create shard group
                 long shardGroupId = GlobalStateMgr.getCurrentState().getStarOSAgent().
-                        createShardGroup(dbId, olapTable.getId(), partitionId, rollupIndexId);
+                        createShardGroup(dbId, olapTable.getId(), physicalPartitionId, rollupIndexId);
                 // index state is SHADOW
                 MaterializedIndex mvIndex = new MaterializedIndex(rollupIndexId,
                         MaterializedIndex.IndexState.SHADOW, shardGroupId);
@@ -98,8 +98,8 @@ public class LakeTableRollupBuilder extends AlterJobV2Builder {
                         .collect(Collectors.toList());
                 List<Long> shadowTabletIds = GlobalStateMgr.getCurrentState().getStarOSAgent().createShards(
                         originTablets.size(),
-                        olapTable.getPartitionFilePathInfo(partitionId),
-                        olapTable.getPartitionFileCacheInfo(partitionId),
+                        olapTable.getPartitionFilePathInfo(physicalPartitionId),
+                        olapTable.getPartitionFileCacheInfo(physicalPartitionId),
                         shardGroupId, originTableIds, shardProperties, computeResource);
                 Preconditions.checkState(originTablets.size() == shadowTabletIds.size());
 
--- a/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
@@ -74,7 +74,7 @@ public class LakeTableRollupBuilder extends AlterJobV2Builder {
                 long physicalPartitionId = physicalPartition.getId();
                 // create shard group
                 long shardGroupId = GlobalStateMgr.getCurrentState().getStarOSAgent().
-                        createShardGroup(dbId, olapTable.getId(), physicalPartitionId, rollupIndexId);
+                        createShardGroup(dbId, olapTable.getId(), partitionId, rollupIndexId);
                 // index state is SHADOW
                 MaterializedIndex mvIndex = new MaterializedIndex(rollupIndexId,
                         MaterializedIndex.IndexState.SHADOW, shardGroupId);
--- a/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
@@ -69,12 +69,11 @@ public class LakeTableRollupBuilder extends AlterJobV2Builder {
         for (Partition partition : olapTable.getPartitions()) {
             long partitionId = partition.getId();
             TStorageMedium medium = olapTable.getPartitionInfo().getDataProperty(partitionId).getStorageMedium();
-
+            // create shard group
+            long shardGroupId = GlobalStateMgr.getCurrentState().getStarOSAgent().
+                        createShardGroup(dbId, olapTable.getId(), partitionId, rollupIndexId);
             for (PhysicalPartition physicalPartition : partition.getSubPartitions()) {
                 long physicalPartitionId = physicalPartition.getId();
-                // create shard group
-                long shardGroupId = GlobalStateMgr.getCurrentState().getStarOSAgent().
-                        createShardGroup(dbId, olapTable.getId(), partitionId, rollupIndexId);
                 // index state is SHADOW
                 MaterializedIndex mvIndex = new MaterializedIndex(rollupIndexId,
                         MaterializedIndex.IndexState.SHADOW, shardGroupId);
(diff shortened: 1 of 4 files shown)


wyb (commented):
Add some UT?

sevev:
> Add some UT?

I will add some test case.

HangyuanLiu (dismissed):

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
```diff
@@ -98,8 +98,8 @@ public AlterJobV2 build() throws StarRocksException {
                         .collect(Collectors.toList());
                 List<Long> shadowTabletIds = GlobalStateMgr.getCurrentState().getStarOSAgent().createShards(
                         originTablets.size(),
-                        olapTable.getPartitionFilePathInfo(partitionId),
-                        olapTable.getPartitionFileCacheInfo(partitionId),
+                        olapTable.getPartitionFilePathInfo(physicalPartitionId),
```
kevincai, line 100: this may cause the new shadowTablets storage path differs from the original ones, consider before/after upgrading, will there be compatible issues?

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
```diff
@@ -74,7 +74,7 @@ public AlterJobV2 build() throws StarRocksException {
                 long physicalPartitionId = physicalPartition.getId();
                 // create shard group
                 long shardGroupId = GlobalStateMgr.getCurrentState().getStarOSAgent().
-                        createShardGroup(dbId, olapTable.getId(), partitionId, rollupIndexId);
+                        createShardGroup(dbId, olapTable.getId(), physicalPartitionId, rollupIndexId);
```
kevincai: also with this change, we need to keep in mind that in future if anyone want to use the "partitionId" attached to the shardGroup property, it can be `partitionId` or `physicalPartitionId` based on when the shardGroup is created. May need to record this info somewhere.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
sevev: > shadowTablets storage path differs from the original

Upgrading may indeed cause path switching, but the logic will not be problematic. This is because during actual operation, it relies on the shard information returned by starmgr to locate the tablet and the change will only affect the save path when creating the tablet.

kevincai: yes, the tablet itself is not a concern. However, when doing the vacuum, it holds the assumption that the tablets in the same partition stored in the same root location, and hence use the first tablet's fs to perform the vacuum for all the tablets.

sevev: > yes, the tablet itself is not a concern. However, when doing the vacuum, it holds the assumption that the tablets in the same partition stored in the same root location, and hence use the first tablet's fs to perform the vacuum for all the tablets.

Yes. However, the root location provided by starmgr is not the real physical path. When actually executing file deletion, the system will find the physical path of the corresponding tablet to perform the deletion, so the files can be correctly deleted in the end.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
sevev: > also with this change, we need to keep in mind that in future if anyone want to use the "partitionId" attached to the shardGroup property, it can be `partitionId` or `physicalPartitionId` based on when the shardGroup is created. May need to record this info somewhere.

Yes. But I think that since we are not currently using the partitionId in shardinfo, there is no need to introduce additional complexity. If we need to use partitionId in the future, we will need to consider historical issues.

kevincai: That's what I am worried about. Need to put notes somewhere to reminder the "future" developer who will make the change, otherwise who knows there is an issue here due to historic change of the partitionId.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
kevincai: That's not the real case, check the following implementation in `delete_tablets` vacuum.cpp, only the first tablet's root location is retrieved, and it is used to clean all the tablets in the batch without considering some of the tablets may have a different root_location.

```
    auto root_dir = tablet_mgr->tablet_root_location(tablet_ids[0]);
    auto st = delete_tablets_impl(tablet_mgr, root_dir, tablet_ids);
```

luohaha: > this may cause the new shadowTablets storage path differs from the original ones, consider before/after upgrading, will there be compatible issues?

This is a bug, should we put so much effect to make this bug compatible?

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableRollupBuilder.java
```diff
@@ -98,8 +98,8 @@ public AlterJobV2 build() throws StarRocksException {
                         .collect(Collectors.toList());
                 List<Long> shadowTabletIds = GlobalStateMgr.getCurrentState().getStarOSAgent().createShards(
                         originTablets.size(),
-                        olapTable.getPartitionFilePathInfo(partitionId),
```
wyb, line 101: create shard group once before traverse all sub partitions
```
// line 76
// create shard group
long shardGroupId = GlobalStateMgr.getCurrentState().getStarOSAgent().createShardGroup(dbId, olapTable.getId(), partitionId, rollupIndexId);
```

fe/fe-core/src/main/java/com/starrocks/lake/StarOSAgent.java
```diff
@@ -451,6 +451,9 @@ private long getWorkerIdByNodeIdInternal(long nodeId) {
         return workerId;
     }
 
+    // ATTN
+    // the partitionId is physical partitionId right now, but the partitionId is logical partitonId in
```
wyb: still use logical partition id in new version

fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java
```diff
@@ -1729,6 +1729,7 @@ Partition createPartition(Database db, OlapTable table, long partitionId, String
                               ComputeResource computeResource) throws DdlException {
         PartitionInfo partitionInfo = table.getPartitionInfo();
         Map<Long, MaterializedIndex> indexMap = new HashMap<>();
+        long physicalPartitionId = GlobalStateMgr.getCurrentState().getNextId();
```
wyb: no need to change

kevincai (approved):

fe/fe-core/src/test/java/com/starrocks/alter/LakeRollupJobTest.java
```diff
@@ -213,6 +217,25 @@ public void sendAgentTask(AgentBatchTask batchTask) {
...
+                shardGroupInfos = GlobalStateMgr.getCurrentState().getStarOSAgent().listShardGroup();
+                Assert.assertTrue(shardGroupInfos != null && !shardGroupInfos.isEmpty());
+                Optional<ShardGroupInfo> targetGroup = shardGroupInfos.stream()
+                        .filter(group -> group.getGroupId() == physicalPartition.getShardGroupId())
+                        .findFirst();
+                Assert.assertTrue(targetGroup.isPresent());
+                Map<String, String> labels = targetGroup.get().getLabelsMap();
+                if (!labels.containsKey("partitionId")) {
+                    Assert.assertTrue(false);
+                }
```
kevincai, line 234:
```suggestion
                Assert.assertTrue(labels.containsKey("partitionId"));
```

kevincai (approved):

wyb (approved):

HangyuanLiu (approved):

PR #60112, opened 2025-06-20
[Enhancement] Collect ClusterSnapshotInfo via cluster snapshot checkpoint

srlch:
## What I'm doing:
In this pr #59845, we introduce ClusterSnapshotInfo. This info is used to save some snapshot info for a single cluster snapshot. This pr introduce a mechanism that collect such information during cluster snapshot checkpoint.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
@@ -55,10 +55,10 @@ public class GlobalStateCheckpointWorker extends CheckpointWorker {
             if (needClusterSnapshotInfo) {
                 this.clusterSnapshotInfo = SnapshotInfoHelper.buildClusterSnapshotInfo(
                         globalStateMgr.getLocalMetastore().getAllDbs());
-                LOG.info("get cluster snapshot info successfully");
             }
 
-            LOG.info("checkpoint finished save image.{}", replayedJournalId);
+            LOG.info("checkpoint finished save image.{}, needClusterSnapshotInfo: {}",
+                     replayedJournalId, needClusterSnapshotInfo);
         } finally {
             GlobalStateMgr.destroyCheckpoint();
         }
--- a/fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
@@ -14,6 +14,8 @@
 
 package com.starrocks.journal;
 
+import com.starrocks.lake.snapshot.ClusterSnapshotInfo;
+import com.starrocks.lake.snapshot.SnapshotInfoHelper;
 import com.starrocks.leader.CheckpointController;
 import com.starrocks.metric.MetricRepo;
 import com.starrocks.persist.EditLog;
@@ -26,7 +28,7 @@ public class GlobalStateCheckpointWorker extends CheckpointWorker {
     }
 
     @Override
-    void doCheckpoint(long epoch, long journalId) throws Exception {
+    void doCheckpoint(long epoch, long journalId, boolean needClusterSnapshotInfo) throws Exception {
         long replayedJournalId = -1;
         // generate new image file
         LOG.info("begin to generate new image: image.{}", journalId);
@@ -50,6 +52,13 @@ public class GlobalStateCheckpointWorker extends CheckpointWorker {
                 MetricRepo.COUNTER_IMAGE_WRITE.increase(1L);
             }
             servingGlobalState.setImageJournalId(journalId);
+
+            if (needClusterSnapshotInfo) {
+                this.clusterSnapshotInfo = SnapshotInfoHelper.buildClusterSnapshotInfo(
+                        globalStateMgr.getLocalMetastore().getAllDatabases());
+                LOG.info("get cluster snapshot info successfully");
+            }
+
             LOG.info("checkpoint finished save image.{}", replayedJournalId);
         } finally {
             GlobalStateMgr.destroyCheckpoint();
--- a/fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
@@ -14,7 +14,6 @@
 
 package com.starrocks.journal;
 
-import com.starrocks.lake.snapshot.ClusterSnapshotInfo;
 import com.starrocks.lake.snapshot.SnapshotInfoHelper;
 import com.starrocks.leader.CheckpointController;
 import com.starrocks.metric.MetricRepo;
@@ -55,7 +54,7 @@ public class GlobalStateCheckpointWorker extends CheckpointWorker {
 
             if (needClusterSnapshotInfo) {
                 this.clusterSnapshotInfo = SnapshotInfoHelper.buildClusterSnapshotInfo(
-                        globalStateMgr.getLocalMetastore().getAllDatabases());
+                        globalStateMgr.getLocalMetastore().getAllDbs());
                 LOG.info("get cluster snapshot info successfully");
             }
 
--- a/fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshot.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshot.java
@@ -98,10 +98,6 @@ public class ClusterSnapshot {
         return type == ClusterSnapshotType.AUTOMATED;
     }
 
-    public boolean needClusterSnapshotInfo() {
-        return false;
-    }
-
     public void setClusterSnapshotInfo(ClusterSnapshotInfo clusterSnapshotInfo) {
         return;
     }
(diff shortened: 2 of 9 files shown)


xiangguangyxg (dismissed):

gengjun-git (dismissed):

fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
```diff
@@ -50,6 +51,13 @@ void doCheckpoint(long epoch, long journalId) throws Exception {
                 MetricRepo.COUNTER_IMAGE_WRITE.increase(1L);
             }
             servingGlobalState.setImageJournalId(journalId);
+
+            if (needClusterSnapshotInfo) {
+                this.clusterSnapshotInfo = SnapshotInfoHelper.buildClusterSnapshotInfo(
+                        globalStateMgr.getLocalMetastore().getAllDbs());
+                LOG.info("get cluster snapshot info successfully");
```
alvin-phoenix-ai: Could you print the information in the original log?

fe/fe-core/src/main/java/com/starrocks/leader/CheckpointController.java
```diff
@@ -96,14 +97,19 @@ public class CheckpointController extends FrontendDaemon {
     private volatile String workerNodeName;
     private volatile long workerSelectedTime;
     private volatile long journalId;
-    private volatile BlockingQueue<Pair<Boolean, String>> result;
+    private volatile BlockingQueue<FinishCheckpointResult> result;
+
+    private volatile boolean needClusterSnapshotInfo;
+    private volatile ClusterSnapshotInfo clusterSnapshotInfo;
```
alvin-phoenix-ai: Could you add a comment to explain how these fields will be used?

alvin-phoenix-ai (commented):
could you explain the mechanism you designed to collect SnapshotInfo during checkpoint?
I feel a little confused from reading the code.

fe/fe-core/src/main/java/com/starrocks/leader/CheckpointController.java
srlch: updated

fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
srlch: You mean use `System.out.println`?

srlch:
> could you explain the mechanism you designed to collect SnapshotInfo during checkpoint? I feel a little confused from reading the code.

When cluster snapshot begin, the snapshot scheduler will try to get checkpoint ids and send it to checkpoint controller to trigger a new round of checkpoint. After finishing, we will get a new image file corresponding to the checkpoint ids which is what we needed in cluster snapshot. Cluster snapshot info can be considered  as part of the snapshoted image. It contains the necessary info to tell us what we actually snapshoted.  So if we want to get it, we can build this info from checkpoint worker just after the image file has been flushed into disk.

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotCheckpointScheduler.java
```diff
@@ -73,6 +73,7 @@ protected void runAfterCatalogReady() {
             runCheckpointScheduler(runningJob);
         } finally {
             runningJob = null;
+            feController.setNeedClusterSnapshotInfo(false);
```
alvin-phoenix-ai: How about using a function interface `submit`, which mean to submit a request to get a ClusterSnapshotInfo. 
The return of this function can be a Future, allowing the function caller to wait for the result.

alvin-phoenix-ai: It is challenging to synchronize by setting and getting status.

srlch: Updated, i try to use a simpler way to avoid potential concurrency issues. I changed `NeedClusterSnapshotInfo` from a member of the CheckpointController to a runtime function param to let Controller know to collect the information.

fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
alvin-phoenix-ai: you can change `LOG.info("checkpoint finished save image.{}", replayedJournalId);` to have this information

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshot.java
```diff
@@ -94,6 +94,18 @@ public long getId() {
...
+        return type == ClusterSnapshotType.AUTOMATED;
+    }
+
+    public boolean needClusterSnapshotInfo() {
+        return false;
+    }
+
+    public void setClusterSnapshotInfo(ClusterSnapshotInfo clusterSnapshotInfo) {
+        return;
+    }
```
alvin-phoenix-ai: What are these two functions used for

fe/fe-core/src/main/java/com/starrocks/leader/CheckpointController.java
```diff
@@ -471,4 +486,20 @@ public void workerRestarted(String nodeName, long startTime) {
     public Journal getJournal() {
         return journal;
     }
+
+    public ClusterSnapshotInfo getClusterSnapshotInfo() {
+        return this.clusterSnapshotInfo;
+    }
+
+    static class FinishCheckpointResult {
```
alvin-phoenix-ai: Will CheckpointCompletionStatus be a suitable name?

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotJob.java
```diff
@@ -148,6 +154,18 @@ public void setDetailInfo(String detailInfo) {
...
+        return snapshot.needClusterSnapshotInfo();
+    }
+
+    public boolean isAutomated() {
+        return snapshot.isAutomated();
+    }
+
+    public void setClusterSnapshotInfo(ClusterSnapshotInfo clusterSnapshotInfo) {
+        snapshot.setClusterSnapshotInfo(clusterSnapshotInfo);
+    }
```
alvin-phoenix-ai: Why do these functions get information from snapshot?
And what does snapshot mean? 
Does it mean a snapshot request? Or an outcome snapshot?

fe/fe-core/src/main/java/com/starrocks/journal/GlobalStateCheckpointWorker.java
srlch: Updated

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshot.java
srlch: This two functions will be overwrite in the future and this is the base class.

fe/fe-core/src/main/java/com/starrocks/leader/CheckpointController.java
srlch: Updated

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotJob.java
srlch: It means a outcome snapshot. `snapshot` can be consider as a result of a finished `clusterJob`. In current impl, a `clusterjob` contains a `snapshot` member which is visible until job is finised. When the job is running, new status information will be injected into the ``snapshot` via `clusterjob`.

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshot.java
alvin-phoenix-ai: The name 'needClusterSnapshotInfo' seems to me to be a property that relates to the snapshot job, which should not be included in ClusterSnapshot.
A snapshot request can have a property that indicates if ClusterSnapshotInfo is needed or not, but Snapshot itself should have no idea about it.

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotJob.java
alvin-phoenix-ai: Who will be the caller of this function?

srlch: Snapshot Checkpoint Scheduler will call these functions during cluster snapshot execution because the scheduler needs to modify the status of the running job.

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshot.java
srlch: Updated, remove it from snapshot, keep it only in snapshotJob

xiangguangyxg (approved):

gengjun-git (approved):

PR #60146, opened 2025-06-20
[UT] update unstable casee

Seaven:
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/test/sql/test_query_history/R/test_query_history
+++ b/test/sql/test_query_history/R/test_query_history
@@ -36,7 +36,7 @@ select sleep(2);
 -- result:
 100
 -- !result
-select count(1) from _statistics_.query_history where sql_digest like "%count%"; 
+select count(1) > 0 from _statistics_.query_history where sql_digest like "%count%"; 
 
 set global enable_query_history=false;
 -- result:
--- a/test/sql/test_query_history/T/test_query_history
+++ b/test/sql/test_query_history/T/test_query_history
@@ -23,7 +23,7 @@ select count(1) from t1;
 
 select sleep(2);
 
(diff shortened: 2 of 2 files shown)


test/sql/test_query_history/R/test_query_history
```diff
@@ -36,7 +36,7 @@ select sleep(2);
 -- result:
 100
 -- !result
-select count(1) from _statistics_.query_history where sql_digest like "%count%"; 
+select count(1) > 0 from _statistics_.query_history where sql_digest like "%count%";
```
alvin-phoenix-ai, line 39: I'm wondering why this change can mitigate the unstable test.

Seaven: this feature is related to the number of queries being execute and the task schedule, i don't replay the problem in other env

alvin-phoenix-ai: Got it, because the test cases run in parallel, this should be the reason why the count is not what you want.

alvin-phoenix-ai: Follow up question is why the query changes but the result has no change.

Seaven: the expect result is `1` row , and `count(1) > 0` will be `1` when has any rows

alvin-phoenix-ai (approved):

PR #59904, opened 2025-06-14
[BugFix] Fix third-party build failure for pulsar-client-cpp-3.3.0 mo…

rohitrs1983:
Fixes #59903 

**Issue 1:**
===== begin build pulsar-client-cpp-3.3.0
-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is GNU 14.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Pulsar Client version: 3.3.0
-- Pulsar Client version macro: 3003000
-- LINK_STATIC:  OFF
-- Using CCache
-- ARCHITECTURE: x86_64
-- BUILD_DYNAMIC_LIB:  OFF
-- BUILD_STATIC_LIB:  ON
-- BUILD_TESTS:  OFF
-- BUILD_PERF_TOOLS:  OFF
-- CMAKE_BUILD_TYPE:  RelWithDebInfo
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Threads library: 
-- Found Boost: /home/ubuntu/starrocks/thirdparty/installed/include (found version "1.80.0")  
Boost_INCLUDE_DIRS: /home/ubuntu/starrocks/thirdparty/installed/include
-- Found OpenSSL: /home/ubuntu/starrocks/thirdparty/installed/lib/libcrypto.a (found version "1.1.1m")  
OPENSSL_INCLUDE_DIR: /home/ubuntu/starrocks/thirdparty/installed/include
OPENSSL_LIBRARIES: /home/ubuntu/starrocks/thirdparty/installed/lib/libssl.a/home/ubuntu/starrocks/thirdparty/installed/lib/libcrypto.adl
-- Found Protobuf: /home/ubuntu/starrocks/thirdparty/installed/lib/libprotobuf.a (found version "3.14.0") 
Protobuf_INCLUDE_DIRS: /home/ubuntu/starrocks/thirdparty/installed/include
[...]

PR #60156, opened 2025-06-22
[Enhancement] Enhance thirdparty build script

Mesut-Doner:
## Why I'm doing:
When some specific thirdparty build scenario, I had to wait all thirdparties build before that spesific lib. This is takes too much time. 
## What I'm doing:
I've added build parameters to make it easier for adding new thirdparty lib etc. 
I have referenced from Doris' build-thirdparty.sh script, added j and clean parameters also...

image

Example syntaxes:
`./build-thirdparty.sh -j16 --continue libdeflate`
`./build-thirdparty.sh -j16 --clean`
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/thirdparty/build-thirdparty.sh
+++ b/thirdparty/build-thirdparty.sh
@@ -26,7 +26,7 @@
 # This script will run *download-thirdparty.sh* once again
 # to check if all thirdparties have been downloaded, unpacked and patched.
 #################################################################################
-set -e
+set -eo pipefail
 
 curdir=`dirname "$0"`
 curdir=`cd "$curdir"; pwd`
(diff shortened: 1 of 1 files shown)


thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,102 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
+
+if [[ "${CONTINUE}" -eq 1 ]]; then
+    if [[ -z "${start_package}" ]] || [[ "${#}" -ne 0 ]]; then
+        usage
+    fi
+fi
+
+read -r -a packages <<<"${@}"
+
+if [[ "${HELP}" -eq 1 ]]; then
```
kevincai: shall put this help check just after the while loop.

thirdparty/build-thirdparty.sh
```diff
@@ -1475,83 +1569,102 @@ export CPPFLAGS=$GLOBAL_CPPFLAGS
...
+        command="build_${package}"
+        ${command}
+    fi
+done
+
 # strip unnecessary debug symbol for binaries in thirdparty
 strip_binary
 
-echo "Finished to build all thirdparties"
+echo "Finished to build all thirdparties"
```
kevincai: fix the eol mark.

thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,102 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
 . ${TP_DIR}/vars.sh
 
+# Check args
+usage() {
+    echo "
+Usage: $0 [options...] [packages...]
+  Optional options:
+     -j <num>               build thirdparty parallel
+     --clean                clean the extracted data
+     --continue <package>   continue to build the remaining packages (starts from the specified package)
```
kevincai: provide a few examples in the usage, so the helper message can guide the user to use the script correctly.

thirdparty/build-thirdparty.sh
Mesut-Doner: Many thanks! @kevincai I have modified.

thirdparty/build-thirdparty.sh
Mesut-Doner: thanks, I've fixed.

thirdparty/build-thirdparty.sh
Mesut-Doner: thanks you're right @kevincai I tried to improve help section like below: 
 
image

thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,128 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
+    $0 -j8
+
+    # Clean and rebuild everything with 16 parallel jobs
+    $0 --clean -j16
+
+    # Continue building from rocksdb (useful after build failure)
+    $0 --continue rocksdb
+
+    # Build only specific packages
+    $0 openssl curl protobuf
```
kevincai, line 83: specific packages should be still built by the same order where these packages are when all the packages are built. some of the packages depend on other packages and must be built after the depending packages.

thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,128 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
+    CLEAN               -- ${CLEAN}
+    PACKAGES            -- ${packages[*]}
+    CONTINUE            -- ${start_package}
+"
+
 cd $TP_DIR
 
+if [[ "${CLEAN}" -eq 1 ]] && [[ -d "${TP_SOURCE_DIR}" ]]; then
+    echo 'Clean the extracted data ...'
+    find "${TP_SOURCE_DIR}" -mindepth 1 -maxdepth 1 -type d -exec rm -rf {} \;
```
kevincai: Potentially it will clean all the directories including directories that are not extracted from the downloaded archives. It will be safer to iterate all the package sources and clean them, or just clean the packages' sources provided by the commandline options.

thirdparty/build-thirdparty.sh
Mesut-Doner: hi @kevincai I added a comment about it like warn

thirdparty/build-thirdparty.sh
Mesut-Doner: many thanks! @kevincai Hope I understood correctly. Done the changes. Could you please review again?

thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,264 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
+
+    # Clean and build only specific packages
+    $0 --clean boost thrift
+  "
+    exit 1
+}
+
+get_all_package_sources() {
+    local sources=(
+        "$LIBEVENT_SOURCE"
```
kevincai: these long list can be easily out of sync with the actual package list, better to use `$TP_ARCHIVES` to create the list.

thirdparty/build-thirdparty.sh
kevincai: thought line 222-224 should be moved to here to replace the `find` command to clean the source, right?

thirdparty/build-thirdparty.sh
kevincai: ok

thirdparty/build-thirdparty.sh
Mesut-Doner: thanks sir! I've modified as using $TP_ARCHIVES

thirdparty/build-thirdparty.sh
Mesut-Doner: sorry @kevincai you're right. I missed removing this. Now It should be ok.

thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,186 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
+
+        if [[ -n "$source_name" ]]; then
+            echo "$source_name"
+        fi
+    done
+    # llvm is a special case where the build directory is named differently
+    # for example
+    # llvm-9.0.0.src (the source)
+    # llvm-build (the build directory)
+    echo "llvm-build"
```
kevincai: Think it should be fixed by https://github.com/StarRocks/starrocks/pull/59904

thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,186 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
+    for archive in ${TP_ARCHIVES}; do
+        # getting source name from archive
+        # removes .tar.gz, .tar.bz2, .tgz, .zip, etc.
+        local source_name="${archive%.tar.*}"
+        source_name="${source_name%.tgz}"
+        source_name="${source_name%.zip}"
+
+        if [[ -n "$source_name" ]]; then
+            echo "$source_name"
+        fi
```
kevincai: source name can be extracted from those *_SOURCE var.
```
for archive in $TP_ARCHIVES
do
    local source_var="${archive}_SOURCE"
    if [[ -n "${!source_var} ]] ; then
        echo ${!source_var}
    fi
done
```

thirdparty/build-thirdparty.sh
```diff
@@ -50,8 +50,186 @@ if [ ! -f ${TP_DIR}/vars.sh ]; then
...
+        ;;
+    esac
+done
+
+# checking for help first, before processing other arguments
+if [[ "${HELP}" -eq 1 ]]; then
+    usage
+fi
+
+read -r -a packages <<<"${@}"
```
kevincai: not quite sure what's the bash version required from this `<<<` in place here doc. Wondering if there is alternative to replace this one that might be compatible in a lower bash version.

thirdparty/build-thirdparty.sh
Mesut-Doner: hello @kevincai I think this is still needed. Maybe @rohitrs1983 can double check.

thirdparty/build-thirdparty.sh
Mesut-Doner: thanks you're right. That was wrong. I've fixed.

thirdparty/build-thirdparty.sh
Mesut-Doner: I've changed <<< operator to make compatible with lower bash versions.

thirdparty/build-thirdparty.sh
kevincai: ok

kevincai: if that is the case, the LLVM should be skipped from the for-loop when iterating TP_ARCHIVES.

Mesut-Doner: hello again @kevincai thanks! . We've checked it. And I tested it. You are right, that part is unnecessary. It already removes llvm source without this. I removed unnecessary code.

kevincai (approved):

alvin-phoenix-ai (approved):

PR #60180, opened 2025-06-23
[Enhancement] from_unixtime: return datetime type

murphyatwork:
## Why I'm doing:

The `from_unixtime` function converts a Unix timestamp into a datetime based on the specified format.

The current implementation always returns a VARCHAR type regardless of the input parameter. This behavior is inconsistent with the MySQL protocol and introduces unnecessary overhead during string formatting.

## What I'm doing:

The signature of the `from_unixtime` function is updated to return a `DATETIME` directly when the format parameter is omitted.

To ensure compatibility during version upgrades, we have introduced a new function signature, `from_unixtime_v2`. This approach allows the replacement to occur only within the AST, enabling both versions of the implementation to coexist seamlessly.

Performance benefit: reduce function latency by ~30%.

Behavior change:
- `from_unixtime` returns `DATETIME` type when the format parameter is omitted
- previously it returns `VARCHAR` 

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exprs/time_functions.cpp
+++ b/be/src/exprs/time_functions.cpp
@@ -26,6 +26,7 @@
 #include "runtime/datetime_value.h"
 #include "runtime/runtime_state.h"
 #include "types/date_value.h"
+#include "types/logical_type.h"
 
 namespace starrocks {
 // index as day of week(1: Sunday, 2: Monday....), value as distance of this day and first day(Monday) of this week.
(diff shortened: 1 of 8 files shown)


stdpain:
prefer add a new session variables to control the behaviour.

PR #60213, opened 2025-06-24
[Feature] Add manual split tablet sql ast for dynamic tablet splitting and merging

xiangguangyxg:
## What I'm doing:
Manual split tablet sql:
```
ALTER TABLE <table_name> SPLIT
    [(TABLET | TABLETS) (PARTITION | PARTITIONS) (<partiton_name1>, <partition_name2>, ...)]
    [(TABLET | TABLETS) (tablet_id1, tablet_id2, tablet_id3, ...)]
[PROPERTIES (
    "dynamic_tablet_split_size"="<split_size>")]
```

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


fe/fe-core/src/main/java/com/starrocks/sql/parser/StarRocks.g4
```diff
@@ -2449,10 +2458,18 @@ keyPartitions
     : PARTITION '(' keyPartition (',' keyPartition)* ')'                              #keyPartitionList
     ;
 
+rollupNames
+    : (ROLLUP | ROLLUPS) '(' identifierOrString (',' identifierOrString)* ')'
```
alvin-phoenix-ai: Rollup is a legacy concept that we want to deprecate. 
We'd better not to introduce it in this feature.

fe/fe-core/src/main/java/com/starrocks/sql/parser/StarRocks.g4
```diff
@@ -2449,10 +2458,18 @@ keyPartitions
...
+    ;
+
 tabletList
     : TABLET '(' INTEGER_VALUE (',' INTEGER_VALUE)* ')'
     ;
 
+virtualBucketList
+    : VIRTUAL BUCKETS '(' INTEGER_VALUE (',' INTEGER_VALUE)* ')'
+    ;
```
alvin-phoenix-ai: if the command name is "split tablet", why should users specify "virtual bucket", rather than tablet id?

fe/fe-core/src/main/java/com/starrocks/sql/parser/StarRocks.g4
```diff
@@ -1197,6 +1198,14 @@ dropPersistentIndexClause
     : DROP PERSISTENT INDEX ON TABLETS integer_list
     ;
 
+splitTabletClause
+    : partitionNames?
+      rollupNames?
+      SPLIT TABLET
+      virtualBucketList?
```
alvin-phoenix-ai: Split Tablet should be put in the first place, so that the parser will know it is splitTabletClause, otherwise, it will introduce ambiguity.
```
alter table xxx split tablet;
alter table xxx split tablet id=1,2,3;
alter table xxx split tablet partition=yyy, tablet-id=1,2,3;
```

alvin-phoenix-ai (dismissed):
I just to put a request-changes to make sure we will have agreement on the syntax

fe/fe-core/src/main/java/com/starrocks/sql/parser/StarRocks.g4
xiangguangyxg: How about using materialized index instead of rollup?

alvin-phoenix-ai: My suggestion will be that we don't support it for now to make things easier.

fe/fe-core/src/main/java/com/starrocks/sql/parser/StarRocks.g4
xiangguangyxg: changed to tablet id

fe/fe-core/src/main/java/com/starrocks/sql/parser/StarRocks.g4
xiangguangyxg: done

fe/fe-core/src/main/java/com/starrocks/sql/parser/StarRocks.g4
```diff
@@ -1197,6 +1198,13 @@ dropPersistentIndexClause
     : DROP PERSISTENT INDEX ON TABLETS integer_list
     ;
 
+splitTabletClause
+    : SPLIT TABLET
+      partitionNames?
+      tabletList?
```
alvin-phoenix-ai: When the tablet is specified, is the partition name mandatory or optional?

xiangguangyxg: optional

alvin-phoenix-ai: How about making these two options mutually exclusive?
And if these parititon and tablet is mutually exclusive, 
we can let users do `alter table xxx split tablet 1,2,3` in this way without an extra `tablet(1,2,3)`

xiangguangyxg: good suggestion

xiangguangyxg: done

kangkaisen (approved):

gengjun-git (approved):

PR #60216, opened 2025-06-24
[Enhancement] Spill PartitionWise aggregation

satanson:
## Why I'm doing:

**Motivation**

Introduce spillable partitionwise aggregate/distinct operator to replace  its spillable sorted-based counterpart.

In Sort-based streamin agg implementation, SpillerWriter append input data to OrderedMemTable, when OrderedMemTable's size reach pre-configured limit, the data is sorted by group-by columns, then materialize chunk and convert it into serialized form and flush to disk. The Sort-based streaming agg would create an SpillerReader which uses sort merger to create a totally ordered run from these sorted blocks, then the agg would read from SpillerReader sequentially and merge the same adjascent group by columns. For complex types, both write-time
Sorting and read-time merging is time-consuming.
sort_spill

In Partition-wise agg, input data is shuffled into dozens of SpillPartition by group-by columns, then each SpillPartition flush a block into durable device. and Partitionwise agg would process each SpillPartition one by one, it read the data from the current SpillPartition and push the data into it, then the final aggregating results can be pulled from it, after all the data of the current SpillPartition is processed, Partitionwise agg reset its state then process the next SpillPartition. In this implementation, write-time shuffle cost is introduced but write-time sorting and read-time merging in sorted-based streaming agg are eliminated.

hash_spill

**Test**
Test Environment
1. Cluster: 1 FE x 3 BE(ecs.c6.4xlarge, 16c32g)
2. Dataset: SSB_100g

Test query
```sql
-- Q1  (group by int,int)
select lo_custkey,lo_orderkey, sum(lo_quantity) as sum_qty from lineorder group by lo_custkey,lo_orderkey order by sum_qty desc limit 10;

-- Q2  (group by array[struct])
with cte as (
select 
[named_struct("key1", lo_orderkey, "key2", lo_custkey), named_struct("key1",lo_partkey, "key2", lo_suppkey)] as group_by,
lo_quantity
from lineorder
)
select group_by, sum(lo_quantity) as sum_qty from cte
group by group_by
order by sum_qty desc
limit 10;
```

Test Result

image
From the test result, partitionwise agg outperforms sort-based streaming agg. when the agregation's group by columns are simple type(Q1), partitionwise agg speedup 1.26X(dop=1) and 1.35X(dop=0); for complex group by columns(Q2), partitionwise agg is 2.X times as fast as sort-based, the root cause is that method compare_at of complex times is poor performance. Let's inspect Q2(pipeline_dop=0)'s profiles as follows.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exec/aggregator.cpp
+++ b/be/src/exec/aggregator.cpp
@@ -1206,6 +1206,19 @@ void Aggregator::_destroy_state(AggDataPtr __restrict state) {
     }
 }
 
+const std::vector<SlotId>& Aggregator::output_group_by_slot_ids() const {
+    if (!_opt_partition_by_slot_ids.has_value()) {
+        std::vector<SlotId> slot_ids;
+        const auto num_group_by_columns = _group_by_expr_ctxs.size();
+        slot_ids.resize(num_group_by_columns);
+        auto slots_begin = _output_tuple_desc->slots().begin();
+        std::transform(slots_begin, slots_begin + num_group_by_columns, slot_ids.begin(),
+                       [](auto slot) { return slot->id(); });
+        _opt_partition_by_slot_ids = std::move(slot_ids);
+    }
+    return _opt_partition_by_slot_ids.value();
+}
+
 ChunkPtr Aggregator::_build_output_chunk(const Columns& group_by_columns, const Columns& agg_result_columns,
                                          bool use_intermediate_as_output) {
     ChunkPtr result_chunk = std::make_shared<Chunk>();
--- a/be/src/exec/aggregator.cpp
+++ b/be/src/exec/aggregator.cpp
@@ -1206,19 +1206,6 @@ void Aggregator::_destroy_state(AggDataPtr __restrict state) {
     }
 }
 
-const std::vector<SlotId>& Aggregator::output_group_by_slot_ids() const {
-    if (!_opt_partition_by_slot_ids.has_value()) {
-        std::vector<SlotId> slot_ids;
-        const auto num_group_by_columns = _group_by_expr_ctxs.size();
-        slot_ids.resize(num_group_by_columns);
-        auto slots_begin = _output_tuple_desc->slots().begin();
-        std::transform(slots_begin, slots_begin + num_group_by_columns, slot_ids.begin(),
-                       [](auto slot) { return slot->id(); });
-        _opt_partition_by_slot_ids = std::move(slot_ids);
-    }
-    return _opt_partition_by_slot_ids.value();
-}
-
 ChunkPtr Aggregator::_build_output_chunk(const Columns& group_by_columns, const Columns& agg_result_columns,
                                          bool use_intermediate_as_output) {
     ChunkPtr result_chunk = std::make_shared<Chunk>();
--- a/be/src/exec/pipeline/aggregate/spillable_partitionwise_aggregate_operator.cpp
+++ b/be/src/exec/pipeline/aggregate/spillable_partitionwise_aggregate_operator.cpp
@@ -332,7 +332,7 @@ Status SpillablePartitionWiseAggregateSinkOperatorFactory::prepare(RuntimeState*
     RETURN_IF_ERROR(OperatorFactory::prepare(state));
 
     // init spill options
-    _spill_options = std::make_shared<spill::SpilledOptions>(config::spill_init_partition);
+    _spill_options = std::make_shared<spill::SpilledOptions>(state->spill_partitionwise_agg_partition_num());
 
     _spill_options->spill_mem_table_bytes_size = state->spill_mem_table_size();
     if (state->enable_agg_spill_preaggregation()) {
--- a/be/src/exec/pipeline/aggregate/spillable_partitionwise_aggregate_operator.cpp
+++ b/be/src/exec/pipeline/aggregate/spillable_partitionwise_aggregate_operator.cpp
@@ -17,9 +17,6 @@
 
 namespace starrocks::pipeline {
 
-DEFINE_FAIL_POINT(spill_always_streaming);
-DEFINE_FAIL_POINT(spill_always_selection_streaming);
-
 bool SpillablePartitionWiseAggregateSinkOperator::need_input() const {
     return !is_finished() && !_agg_op->aggregator()->is_full() && !_agg_op->aggregator()->spill_channel()->has_task();
 }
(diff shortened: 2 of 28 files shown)


be/src/exec/aggregator.cpp
```diff
@@ -1206,6 +1206,19 @@ void Aggregator::_destroy_state(AggDataPtr __restrict state) {
     }
 }
 
+const std::vector<SlotId>& Aggregator::output_group_by_slot_ids() const {
```
stdpain: Why do we need this property? It looks like group_by_columns is enough.

be/src/exec/pipeline/aggregate/spillable_partitionwise_aggregate_operator.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+ChunkPtr& SpillablePartitionWiseAggregateSinkOperator::_append_hash_column(ChunkPtr& chunk) {
+    auto slot_ids = _agg_op->aggregator()->output_group_by_slot_ids();
+    size_t num_rows = chunk->num_rows();
+    auto hash_column = spill::SpillHashColumn::create(num_rows);
+    auto& hash_values = hash_column->get_data();
+    // TODO: use different hash method
+    for (auto& slot_id : slot_ids) {
+        auto column = chunk->get_column_by_slot_id(slot_id);
+        column->fnv_hash(hash_values.data(), 0, num_rows);
```
stdpain, line 273: I found a problem with FNV hash at seed=0. This is called FNV-0 hash, if some key is 0, then the output will always be 0. Please use another seed. https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_ hash_function#FNV-0_hash_(deprecated)

be/src/exec/spill/spill_components.cpp
```diff
@@ -579,6 +581,174 @@ Status PartitionedSpillerWriter::_split_input_partitions(workgroup::YieldContext
     return Status::OK();
 }
 
+Status PartitionedSpillerWriter::_pick_and_compact_skew_partitions(std::vector<SpilledPartition*>& partitions) {
```
stdpain, line 588: Can you comment the pick strategy?

be/src/exec/spill/spiller.hpp
```diff
@@ -314,6 +324,13 @@ Status PartitionedSpillerWriter::flush(RuntimeState* state, bool is_final_flush,
         _need_final_flush = true;
         return Status::OK();
     }
+
+    std::vector<SpilledPartition*> flushing_partitions;
+    flushing_partitions.reserve(spilling_partitions.size() + splitting_partitions.size());
+    flushing_partitions.insert(flushing_partitions.end(), splitting_partitions.begin(), splitting_partitions.end());
+    flushing_partitions.insert(flushing_partitions.end(), spilling_partitions.begin(), spilling_partitions.end());
+    RETURN_IF_ERROR(_pick_and_compact_skew_partitions(flushing_partitions));
+
```
stdpain, line 333: why these code is not in L318 ?

be/src/exec/spill/spill_components.cpp
```diff
@@ -579,6 +581,174 @@ Status PartitionedSpillerWriter::_split_input_partitions(workgroup::YieldContext
...
+
+    if (mode_elem_it->second < num_rows * 0.25) {
+        return Status::OK();
+    }
+    auto merger = std::make_shared<Aggregator>(aggregator_params);
+    merger->set_aggr_mode(AM_STREAMING_POST_CACHE);
+    RuntimeProfile* profile = _runtime_state->runtime_profile()->create_child("spillable_pw_skew_elimination", true);
+    RETURN_IF_ERROR(merger->prepare(_runtime_state, _runtime_state->obj_pool(), profile));
+    RETURN_IF_ERROR(merger->open(_runtime_state));
+    DeferOp defer([&merger, this]() { merger->close(_runtime_state); });
```
stdpain, line 716: Will it cause leak when prepare in L693 failed?

satanson: aggregator.prepare/open methods are called in AggregateXXXOperator.prepare method, if  AggregateXXXOperator.prepare method fails, FragmentExecutor.execute returns prematurely, the PipelineDriver's dtor is called. in PipelineDriver::~PipelineDriver(), operator's close status is checked without invoking close method.
```
PipelineDriver::~PipelineDriver() noexcept {
    if (_workgroup != nullptr) {
        _workgroup->decr_num_running_drivers();
    }
    check_operator_close_states("deleting pipeline drivers");
}

void PipelineDriver::check_operator_close_states(const std::string& func_name) {
    if (_driver_id == -1) { // in test cases
        return;
    }
    for (auto& op : _operators) {
        auto& op_state = _operator_stages[op->get_id()];
        if (op_state > OperatorStage::PREPARED && op_state != OperatorStage::CLOSED) {
            std::stringstream ss;
            ss << "query_id=" << (this->_query_ctx == nullptr ? "None" : print_id(this->query_ctx()->query_id()))
               << " fragment_id="
               << (this->_fragment_ctx == nullptr ? "None" : print_id(this->fragment_ctx()->fragment_instance_id()));
            auto msg = fmt::format(
                    "{} close operator {}-{} failed, may leak resources when {}, please report an issue at "
                    "https://github.com/StarRocks/starrocks/issues/new/choose.",
                    ss.str(), op->get_raw_name(), op->get_plan_node_id(), func_name);
            LOG(ERROR) << msg;
            DCHECK(false) << msg;
        }
    }
}
```

so aggregator's close functions should be invoked only after its prepare/close methods are invoked successfully.

be/src/exec/spill/spiller.hpp
satanson: _pick_and_compact_skew_partitions introduces some overhead, if it is put in L318,  it would be invoked excessively more than once for the same mem table. so only when a mem table is flushing, we try to compact the skew data.

stdpain (approved):

silverbullet233 (approved):

ZiheLiu (approved):

colagy:
I have two questions:

Will enabling partitioned aggregation consume more memory?

Can partitioned aggregation cause data skew? Does it support multi-stage aggregation? I didn’t see related information in the diagram.

PR #60228, opened 2025-06-24
[BugFix] Clean the stale datacache data to free disk spaces immediately.

GavinMar:
## Why I'm doing:
Now we have deprecated the `datacache_disk_path` configuration and use the `${storage_root_path}/datacache` as the datacache path.
However, if users configured their own datacache paths before, after upgrading to current version, the old datacache data still remains until users manually clean it.

## What I'm doing:
Automatically check and clean the stale datacache data to free the disk spaces immediately.

For safety, we only clean cache directories that meet the following conditions:
* Only clean the data files that prefix with `blockfile_` in the directory, does not remove the cache directory.
* If there is a `meta` directory in the cache directory, means it is a persistent cache directory and be generated by StarRocks 3.4 or later, we will ignore the cache directory.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1305,6 +1305,10 @@ CONF_Bool(datacache_unified_instance_enable, "true");
 // * lru: the typical `Least Recently Used` eviction policy.
 // * slru: segment lru eviction policies, which can better reduce cache pollution problem.
 CONF_String(datacache_eviction_policy, "slru");
+// Only used to clean the old and stale datacache files.
+// Users usually do not need to configure it.
+CONF_String(datacache_stale_data_path, "");
+CONF_Alias(datacache_stale_data_path, datacache_disk_path);
 
 // The following configurations will be deprecated, and we use the `datacache` prefix instead.
 // But it is temporarily necessary to keep them for a period of time to be compatible with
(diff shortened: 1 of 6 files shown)


be/src/common/config.h
```diff
@@ -1305,6 +1305,10 @@ CONF_Bool(datacache_unified_instance_enable, "true");
 // * lru: the typical `Least Recently Used` eviction policy.
 // * slru: segment lru eviction policies, which can better reduce cache pollution problem.
 CONF_String(datacache_eviction_policy, "slru");
+// Only used to clean the old and stale datacache files.
+// Users usually do not need to configure it.
+CONF_String(datacache_stale_data_path, "");
+CONF_Alias(datacache_stale_data_path, datacache_disk_path);
```
trueeyu: Why not alias to `block_cache_disk_path`

GavinMar: > Why not alias to `block_cache_disk_path`

Good suggestion, fixed.

trueeyu (approved):

PR #60240, opened 2025-06-24
[Enhancement] Convert nan/inf to null to avoid client error

### LINKED ISSUE #51233: [Bug] Parsing of double result fails when using JDBC
In the lake warehouse query scenario, StarRocks is used to read data in hive/iceberg/.... When the double type of the external table stores `nan` or `infinity` (note that it is not `NaN` and `Infinity`), when querying the results through JDBC connection, the final result is parsed through the `getDouble` method of `ResultSet`. However, `getDouble` throws an exception when encountering lowercase `nan` or `infinity`, causing the query to fail.

JDBC getDouble call path：

ResultSet#getDouble -> ResultSetImpl#getDouble -> getDoubleInternal -> Double.parseDouble -> readJavaFormatString

image

image

- 3.2.8

zenoyang:
## Why I'm doing:

fix: https://github.com/StarRocks/starrocks/issues/51233

## What I'm doing:

Added the switch `enable_inf_nan_convert_to_null` to convert `inf` and `nan` to `NULL`, thus avoiding client errors.

Before:
image

After:
image

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/column/adaptive_nullable_column.cpp
+++ b/be/src/column/adaptive_nullable_column.cpp
@@ -301,9 +301,10 @@ int64_t AdaptiveNullableColumn::xor_checksum(uint32_t from, uint32_t to) const {
     return NullableColumn::xor_checksum(from, to);
 }
 
-void AdaptiveNullableColumn::put_mysql_row_buffer(MysqlRowBuffer* buf, size_t idx, bool is_binary_protocol) const {
+void AdaptiveNullableColumn::put_mysql_row_buffer(MysqlRowBuffer* buf, size_t idx, bool is_binary_protocol,
+                                                  bool is_inf_nan_convert_to_null) const {
     materialized_nullable();
-    NullableColumn::put_mysql_row_buffer(buf, idx, is_binary_protocol);
+    NullableColumn::put_mysql_row_buffer(buf, idx, is_binary_protocol, is_inf_nan_convert_to_null);
 }
 
 } // namespace starrocks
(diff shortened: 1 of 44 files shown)


eshishki:
@zenoyang let me restate the bug as I read it from #51233, please correct me if I'm wrong:

1. Source file (Parquet/ORC in Hive/Iceberg) holds a regular `double` — real IEEE-754 values, some of which are `NaN` / `±Inf` (special bit patterns, not strings).
2. StarRocks reads them as `double` correctly — no strings involved on the storage side.
3. When the result is serialized for the MySQL text protocol, the C/C++ formatter (`printf` / `fmt`) prints those special values as `"nan"` / `"inf"` / `"infinity"` — lower-case by default.
4. On the client, JDBC's `ResultSet.getDouble` → `Double.parseDouble` accepts only `"NaN"` / `"Infinity"` (camel-case). Lower-case → `NumberFormatException` → query fails.

If that's the actual root cause, the minimal fix is in the BE result writer's float/double formatting — emit `"NaN"` / `"Infinity"` with the case Java expects. One spot in the formatter.

Replacing `nan`/`inf` with NULL + a session flag + threading the flag through ~44 files of column / pipeline / result-sink code feels heavier than the underlying problem, and silently dropping NaN/Inf to NULL is a behavior change users may not want. Could you reconsider the approach, or — given this has been idle ~11 months with no review, CONFLICTING, and several CI reds (Clang-Format / FE UT / DCO / behavior-check / backport-check) — close it?

PR #60243, opened 2025-06-24
[BugFix]Fix backend stream callback loss.

### LINKED ISSUE #60244: After backend stream callback loss,  flink-connector commit failed
1. CREATE TABLE '...'
2. INSERT INTO '....'
3. SELECT  '....'

 - You can get the StarRocks version by executing SQL `select current_version()`

zhangheihei:
## Why I'm doing:
A flink-connector task imports multiple tables, restarts leader FE, current checkpoint A fails, and there are 2 states of transactions in CPA: prepare/prepared.

After flink task restart, playback starts from CPA, rollback is performed for transactions in prepare state, and commit is performed for transactions in prepared state. but the commit operation will fail because the callback is not persisted, and the callback is lost after FE restart(beginLoadTaskFromFrontend don't edit log).

There are two solutions:
1 Replace the label, but the topic offset of the prepared transaction is already from x--->y, replacing the label is consumed from y, and the data between x and y will be lost.
2 Playback from A-1, so that if it is a DUPLICATE KEY table, will lead to duplicate data import.

## What I'm doing:

BACKEND_STREAMING does not need to rely on callbacks:
1 beginLoadTaskFromBackend does not do the logCreateStreamLoadJob action and the BACKEND StreamLoadTask is not persisted.
2 StreamLoadMgr does not load BACKEND StreamLoadTask when leader FE is restarted

Fixes #60244

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionState.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionState.java
@@ -645,6 +645,9 @@ public class TransactionState implements Writable, GsonPreProcessable {
             // GlobalTransactionMgr between beforeStateTransform and afterStateTransform
             TxnStateChangeCallback callback = GlobalStateMgr.getCurrentState().getGlobalTransactionMgr()
                     .getCallbackFactory().getCallback(callbackId);
+            if (callback == null && this.sourceType == LoadJobSourceType.BACKEND_STREAMING) {
+                callback = GlobalStateMgr.getCurrentState().getStreamLoadMgr().getTaskByLabel(this.label);
+            }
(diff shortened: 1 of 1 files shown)


banmoy:
add a test to cover the case

meegoo (approved):

gengjun-git (approved):

PR #60297, opened 2025-06-25
[BugFix] Prevent approx_cosine_similarity from Returning NaN When Input Vector Norm is Zero

cutiechi:
## Why I'm doing:

The ‎`approx_cosine_similarity` function could previously return NaN when one or both input vectors had a zero norm. This led to unexpected results and potential downstream errors in queries relying on this function.

## What I'm doing:

This PR updates the ‎`approx_cosine_similarity` implementation to check if either input vector has a zero norm. If so, the function now returns 0 instead of performing a division that would result in NaN. This change improves the robustness and reliability of cosine similarity calculations.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exprs/math_functions.cpp
+++ b/be/src/exprs/math_functions.cpp
@@ -1216,7 +1216,11 @@ StatusOr<ColumnPtr> MathFunctions::cosine_similarity(FunctionContext* context, c
             }
         }
         if constexpr (!isNorm) {
-            result_value = sum / (std::sqrt(base_sum) * std::sqrt(target_sum));
+            if (base_sum == 0 || target_sum == 0) {
+                result_value = 0;
+            } else {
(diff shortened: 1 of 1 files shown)


cutiechi:
Before:
企业微信截图_7169736f-8742-45e6-a661-ca1502515802

After:
企业微信截图_78fb8ff0-68e9-4494-b736-3cc43297c29b

dirtysalt (approved):

ZiheLiu (approved):

stdpain (approved):

PR #60331, opened 2025-06-26
[Enhancement] function hour_from_unixtime

murphyatwork:
## Why I'm doing:

The `hour(from_unixtime(xxx))` function is slow because `from_unixtime` needs to extract the full `yyyy-MM-dd HH:mm:ss` format from the timestamp, which takes approximately 30 nanoseconds.

However, in many cases, not all fields are necessary. By calculating only the required fields directly from the timestamp, the process can be significantly faster.

## What I'm doing:

Add a function `hour_from_unixtime(ts)` to extract the hour from a unix timestamp, which is timezone aware.

Implementation:
1. Lookup the timezone offset with `cctz` library 
2. Calculate the hour from local unixtime
```
int64_t local_unixtime = unixtime + timezone.lookup_offset(unixtime);
int hour = (local_unixtime % (24 * 3600)) / 3600
```

Details:
1. Why not use a lookup table? Creating a table to quickly map `seconds of the day -> hour` could serve the same purpose. However, this approach still requires a `%84600` operation, making it not significantly more efficient than the formula mentioned above.
2. How to speedup the hour calculation? We use the `libdivide` for fast `mod` and `div` instruction.

According to benchmark result, it's 4x faster than the `from_unixtime`:
```
// Run on (104 X 3200 MHz CPU s)
// CPU Caches:
//   L1 Data 32 KiB (x52)
//   L1 Instruction 32 KiB (x52)
//   L2 Unified 1024 KiB (x52)
//   L3 Unified 36608 KiB (x2)
// Load Average: 36.03, 17.40, 21.87
// ----------------------------------------------------------------------
// Benchmark                            Time             CPU   Iterations
// ----------------------------------------------------------------------
// BM_HourFromUnixtime             147895 ns       147841 ns         4147
// BM_FromUnixtime_HourExtract     656650 ns       651969 ns         1445
```

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3


satanson (dismissed):

thirdparty/patches/cctz_02_lookup_offset.patch
```diff
@@ -0,0 +1,223 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index e8b2f34..97b34ac 100644
+--- a/CMakeLists.txt
```
dirtysalt, line 3: why need to add this patch? any critical code in this patch?

dirtysalt (dismissed):

be/src/exprs/time_functions.cpp
```diff
@@ -1566,6 +1568,55 @@ StatusOr<ColumnPtr> TimeFunctions::from_unix_to_datetime_ms_64(FunctionContext*
     return _t_from_unix_to_datetime_ms<TYPE_BIGINT>(context, columns);
 }
 
+static inline int64_t impl_hour_from_unixtime(int64_t unixtime) {
+    // return (unixtime % 86400) / 3600;
+    static const libdivide::divider<int64_t> fast_div_3600(3600);
+    static const libdivide::divider<int64_t> fast_div_86400(86400);
+    int64_t hour = (unixtime - unixtime / fast_div_86400 * 86400) / fast_div_3600;
```
dirtysalt, line 1575: what's performance between regular div and this lib?

silverbullet233 (dismissed):

be/src/exprs/time_functions.cpp
murphyatwork: according to this [doc](https://github.com/ridiculousfish/libdivide), it's 5~10x faster compared with regular DIV instruction.

thirdparty/patches/cctz_02_lookup_offset.patch
murphyatwork: - Before this patch, the `from_unixtime` function relied on `cctz::convert` to transform timestamps into civil time. This process involved two main steps: first, consulting the timezone table to determine the offset for a specific timestamp, and second, converting the timestamp into civil time. Both steps were relatively slow.  
- After this patch, a new interface `lookup_offset` is introduced in the cctz library. This allows the second step to be simplified by converting the timestamp directly into hours instead of the full `yymmdd-hhMMss` format, significantly reducing the number of required instructions.

wangsimo0 (approved):

silverbullet233 (approved):

kevincai (approved):

stdpain (approved):

yingtingdong:
ignore backport check: 3.5.2

PR #60348, opened 2025-06-26
[Feature] Add splitting and merging tablets info in the publish version rpc request for dynamic tablet splitting and merging

xiangguangyxg:
## Why I'm doing:
This is a preliminary work for later PRs for tablet splitting and merging.
This PR is for 2 cases:
1. In the split or merge tablet transaction, the publish version rpc request need to carry the distributed columns and the splitting or merging tablets info to do tablet splitting or merging.
2. In the normal write transaction during tablet splitting or merging, the publish version rpc request need to carry the splitting or merging tablets info to do cross-commit (Write data to old tablets but apply to new tablets).

## What I'm doing:
Add splitting and merging tablets info in the publish version rpc request for dynamic tablet splitting and merging.

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/gensrc/proto/lake_service.proto
+++ b/gensrc/proto/lake_service.proto
@@ -39,6 +39,16 @@ message PublishVersionRequest {
     repeated TxnInfoPB txn_infos = 7;
     repeated int64 rebuild_pindex_tablet_ids = 8;
     optional bool enable_aggregate_publish = 9;
+    // Distribution key column physical names (column id)
+    repeated string distribution_columns = 10;
+    // This field is used for 2 cases to specify the splitting tablets info (not included in tablet_ids):
+    // 1. The split tablet transaction, in which the txn type in txn_infos is TXN_SPLIT_TABLET.
+    // 2. The nornal write transaction but need cross-commit during tablet splitting.
+    repeated SplittingTabletInfoPB splitting_tablet_infos = 11;
+    // This field is used for 2 cases to specify the merging tablets info (not included in tablet_ids):
+    // 1. The merge tablet transaction, in which the txn type in txn_infos is TXN_MERGE_TABLET.
+    // 2. The nornal write transaction but need cross-commit during tablet merging.
+    repeated MergingTabletInfoPB merging_tablet_infos = 12;
 }
 
 message PublishVersionResponse {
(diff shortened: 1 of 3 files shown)


gensrc/proto/lake_service.proto
```diff
@@ -39,6 +39,12 @@ message PublishVersionRequest {
     repeated TxnInfoPB txn_infos = 7;
     repeated int64 rebuild_pindex_tablet_ids = 8;
     optional bool enable_aggregate_publish = 9;
+    // Distribution key column physical names
+    repeated string distribution_columns = 10;
```
alvin-phoenix-ai, line 43: One question: How can we achieve this if we want to support splitting by transform columns?
like iceberg https://iceberg.apache.org/spec/?h=transfor#partition-transforms

xiangguangyxg: If we want to support distributing and splitting by transform columns, we can introduce an internal generated column to save the evaluated value of the transform columns, and let it be the distribution key. This will make things   easier, like the unified expression partitioning.

gensrc/proto/lake_types.proto
```diff
@@ -252,5 +252,9 @@ message TxnInfoPB {
     optional bool force_publish = 5; // only used for compaction
     optional bool rebuild_pindex = 6;
     optional int64 gtid = 7 [default=0];
-};
+}
 
+message SplitTabletPB {
```
alvin-phoenix-ai: I think we will support merging tablets, too. Could we create a unified structure to include both? 
original-tablets,
new-tablets
?

xiangguangyxg: I think creating another MergeTabletPB for merging tablets will be better, because the tasks of split and merge in BE are so different that they probably use two different codes.

alvin-phoenix-ai: Split Tablet is a command; we can name it SplitTabletInfoPB or SplitTabletReqPB.

xiangguangyxg: Changed to SplittingTabletInfoPB, I think the ing form is more accurate

srlch (approved):

kevincai (approved):

wangruin (approved):

PR #60355, opened 2025-06-26
[BugFix] Reserve leading inline comments

satanson:
## Why I'm doing:

Leading inline comments is removed in audit log  since 3.4.2 exceptionally, it is unexpected. so we must reserve inline comments

For an example
image

in fe.audit.log
image
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AstToSQLBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AstToSQLBuilder.java
@@ -46,6 +46,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AstToSQLBuilder.java
```diff
@@ -74,12 +76,24 @@ public static String toSQLWithCredential(ParseNode statement) {
         return new AST2SQLBuilderVisitor(false, false, false).visit(statement);
     }
 
+    private static final Pattern LEADING_INLINE_COMMENT_PATTERN =
```
stephen-shelby, line 79: I can't find these code before 3.4.2? do you know why it was removed before?

PR #60366, opened 2025-06-27
[Enhancement] Support different compaction strategy for different table in shared-data mode

sevev:
## Why I'm doing:
After enabling `file_bundling`, we have reduced the number of API calls generated during imports and compaction by 60%. However, in real-time scenarios, compaction is triggered very frequently. In test scenarios, the number of compactions is roughly equivalent to the number of imports (with a 10-second import interval). Since all tables in StarRocks use a unified compaction strategy, it is difficult to handle multiple different user scenarios simultaneously. For example, the compaction strategy preferences for large batch imports and real-time import scenarios differ. Therefore, this PR supports setting different compaction strategies for different tables, enabling support for more scenarios.

## What I'm doing:
1. Introduced the compaction_strategy property to support configuring different compaction strategies. Currently supports two strategies: DEFAULT and REAL_TIME, with DEFAULT as the default to maintain historical behavior.
2. REAL_TIME is suitable for real-time import scenarios. Under the REAL_TIME strategy, there are several behavioral changes compared to historical implementations:
    - Modified the compaction score calculation strategy for PK tables. In `REAL_TIME` strategy, the impact of `update_compaction_delvec_file_io_amp_ratio` on score calculation is removed.
    -  When calculating the compaction score for PK tables under the `REAL_TIME` strategy, scores are also computed hierarchically based on rowset size.
    - Under the `REAL_TIME` strategy, PK tables allow partial rowset cross-level merging. When executing a compaction task, the rowset level with the highest score is selected first, and rowsets from levels with smaller rowset sizes are also merged into it.

## Test Result
use table `store_sales` in TPCDS as test table:
```
CREATE TABLE `store_sales` (
  `ss_item_sk` int(11) NOT NULL COMMENT "",
  `ss_ticket_number` int(11) NOT NULL COMMENT "",
  `ss_sold_date_sk` int(11) NULL COMMENT "",
  `ss_sold_time_sk` int(11) NULL COMMENT "",
  `ss_customer_sk` int(11) NULL COMMENT "",
  `ss_cdemo_sk` int(11) NULL COMMENT "",
  `ss_hdemo_sk` int(11) NULL COMMENT "",
  `ss_addr_sk` int(11) NULL COMMENT "",
  `ss_store_sk` int(11) NULL COMMENT "",
  `ss_promo_sk` int(11) NULL COMMENT "",
  `ss_quantity` int(11) NULL COMMENT "",
  `ss_wholesale_cost` decimal(7, 2) NULL COMMENT "",
  `ss_list_price` decimal(7, 2) NULL COMMENT "",
  `ss_sales_price` decimal(7, 2) NULL COMMENT "",
  `ss_ext_discount_amt` decimal(7, 2) NULL COMMENT "",
  `ss_ext_sales_price` decimal(7, 2) NULL COMMENT "",
  `ss_ext_wholesale_cost` decimal(7, 2) NULL COMMENT "",
  `ss_ext_list_price` decimal(7, 2) NULL COMMENT "",
  `ss_ext_tax` decimal(7, 2) NULL COMMENT "",
  `ss_coupon_amt` decimal(7, 2) NULL COMMENT "",
  `ss_net_paid` decimal(7, 2) NULL COMMENT "",
  `ss_net_paid_inc_tax` decimal(7, 2) NULL COMMENT "",
  `ss_net_profit` decimal(7, 2) NULL COMMENT ""
) ENGINE=OLAP
PRIMARY KEY(`ss_item_sk`, `ss_ticket_number`)
COMMENT "OLAP"
DISTRIBUTED BY HASH(`ss_item_sk`, `ss_ticket_number`) BUCKETS 3
PROPERTIES (
"compression" = "LZ4",
"datacache.enable" = "true",
"enable_async_write_back" = "false",
"enable_persistent_index" = "true",
"persistent_index_type" = "CLOUD_NATIVE",
"replication_num" = "1",
"storage_volume" = "builtin_storage_volume",
"compaction_strategy" = "REAL_TIME" 
);
``` 
Import 6,000 rows of data every 10 seconds for 60 consecutive minutes, and execute the query during the import:
```
select count(1) from store_sales2 where ss_sold_date_sk=2451634;
```
Statistic the query time consumption, the number of compactions, and the write amplification of compactions.

The following is test result:

### query cost time：
1. enable starlet cache:
    
image

| compaction strategy | compaction times | compaction write bytes |
| ---- | ---- | ---- |
| default | 272 | 633174015 |
| real_time | 35 | 228015297 |

2. disable starlet cache:

image
| compaction strategy | compaction times | compaction write bytes |
| ---- | ---- | ---- |
| default | 258 | 576481031 |
| real_time | 36 | 228015297 |

#### summary
1. The `REAL_TIME` strategy results in fewer compaction executions and less byte writes in real-time scenarios, reducing API calls and performance consumption.  

2. The query performance under the `REAL_TIME` strategy in real-time scenarios is basically consistent with that of the `DEFAULT` strategy.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -396,6 +396,9 @@ CONF_Bool(enable_event_based_compaction_framework, "true");
 
 CONF_Bool(enable_size_tiered_compaction_strategy, "true");
 CONF_mBool(enable_pk_size_tiered_compaction_strategy, "true");
+// We support real-time compaction strategy for primary key tables in shared-data mode.
+// This real-time compaction strategy enables compacting rowsets across multiple levels simultaneously.
+// The parameter `size_tiered_max_compaction_level` defines the maximum compaction level allowed in a single compaction task.
 CONF_mInt32(size_tiered_max_compaction_level, "3");
 CONF_mInt64(size_tiered_min_level_size, "131072");
 CONF_mInt64(size_tiered_level_multiple, "5");
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -396,6 +396,7 @@ CONF_Bool(enable_event_based_compaction_framework, "true");
 
 CONF_Bool(enable_size_tiered_compaction_strategy, "true");
 CONF_mBool(enable_pk_size_tiered_compaction_strategy, "true");
+CONF_mInt(size_tiered_max_compaction_level, "3");
 CONF_mInt64(size_tiered_min_level_size, "131072");
 CONF_mInt64(size_tiered_level_multiple, "5");
 CONF_mInt64(size_tiered_level_multiple_dupkey, "10");
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -396,7 +396,7 @@ CONF_Bool(enable_event_based_compaction_framework, "true");
 
 CONF_Bool(enable_size_tiered_compaction_strategy, "true");
 CONF_mBool(enable_pk_size_tiered_compaction_strategy, "true");
-CONF_mInt(size_tiered_max_compaction_level, "3");
+CONF_mInt32(size_tiered_max_compaction_level, "3");
 CONF_mInt64(size_tiered_min_level_size, "131072");
 CONF_mInt64(size_tiered_level_multiple, "5");
 CONF_mInt64(size_tiered_level_multiple_dupkey, "10");
--- a/be/src/storage/lake/compaction_policy.cpp
+++ b/be/src/storage/lake/compaction_policy.cpp
@@ -80,6 +80,12 @@ private:
 
 StatusOr<uint32_t> primary_compaction_score_by_policy(TabletManager* tablet_mgr,
                                                       const std::shared_ptr<const TabletMetadataPB>& metadata) {
+    bool is_real_time = false;
+    if (metadata->has_compaction_strategy() && metadata->compaction_strategy() == CompactionStrategyPB::REAL_TIME) {
+        is_real_time = true;
+    }
+    uint32_t update_compaction_delvec_file_io_amp_ratio =
+            is_real_time ? 1 : config::update_compaction_delvec_file_io_amp_ratio;
     PrimaryCompactionPolicy policy(tablet_mgr, metadata, false /* force_base_compaction */);
     std::vector<bool> has_dels;
     ASSIGN_OR_RETURN(auto pick_rowset_indexes, policy.pick_rowset_indexes(metadata, true, &has_dels));
@@ -90,7 +96,7 @@ StatusOr<uint32_t> primary_compaction_score_by_policy(TabletManager* tablet_mgr,
         auto current_score = pick_rowset.overlapped() ? pick_rowset.segments_size() : 1;
         if (has_del) {
             // if delvec file exist, expand score by config.
-            current_score *= config::update_compaction_delvec_file_io_amp_ratio;
+            current_score *= update_compaction_delvec_file_io_amp_ratio;
         }
         segment_num_score += current_score;
     }
--- a/be/src/storage/lake/compaction_policy.cpp
+++ b/be/src/storage/lake/compaction_policy.cpp
@@ -79,11 +79,7 @@ private:
 };
 
 bool CompactionPolicy::is_real_time_compaction_strategy(const std::shared_ptr<const TabletMetadataPB>& metadata) {
-    bool is_real_time = false;
-    if (metadata->has_compaction_strategy() && metadata->compaction_strategy() == CompactionStrategyPB::REAL_TIME) {
-        is_real_time = true;
-    }
-    return is_real_time;
+    return metadata->has_compaction_strategy() && metadata->compaction_strategy() == CompactionStrategyPB::REAL_TIME;
 }
 
 StatusOr<uint32_t> primary_compaction_score_by_policy(TabletManager* tablet_mgr,
--- a/be/src/storage/lake/compaction_policy.cpp
+++ b/be/src/storage/lake/compaction_policy.cpp
@@ -78,17 +78,21 @@ private:
     };
 };
 
-StatusOr<uint32_t> primary_compaction_score_by_policy(TabletManager* tablet_mgr,
-                                                      const std::shared_ptr<const TabletMetadataPB>& metadata) {
+bool CompactionPolicy::is_real_time_compaction_strategy(const std::shared_ptr<const TabletMetadataPB>& metadata) {
     bool is_real_time = false;
     if (metadata->has_compaction_strategy() && metadata->compaction_strategy() == CompactionStrategyPB::REAL_TIME) {
         is_real_time = true;
     }
-    uint32_t update_compaction_delvec_file_io_amp_ratio =
-            is_real_time ? 1 : config::update_compaction_delvec_file_io_amp_ratio;
+    return is_real_time;
+}
+
+StatusOr<uint32_t> primary_compaction_score_by_policy(TabletManager* tablet_mgr,
+                                                      const std::shared_ptr<const TabletMetadataPB>& metadata) {
     PrimaryCompactionPolicy policy(tablet_mgr, metadata, false /* force_base_compaction */);
+    uint32_t update_compaction_delvec_file_io_amp_ratio =
+            policy.is_real_time_compaction_strategy(metadata) ? 1 : config::update_compaction_delvec_file_io_amp_ratio;
     std::vector<bool> has_dels;
-    ASSIGN_OR_RETURN(auto pick_rowset_indexes, policy.pick_rowset_indexes(metadata, true, &has_dels));
+    ASSIGN_OR_RETURN(auto pick_rowset_indexes, policy.pick_rowset_indexes(metadata, &has_dels));
     uint32_t segment_num_score = 0;
     for (int i = 0; i < pick_rowset_indexes.size(); i++) {
         const auto& pick_rowset = metadata->rowsets(pick_rowset_indexes[i]);
--- a/be/src/storage/lake/compaction_policy.h
+++ b/be/src/storage/lake/compaction_policy.h
@@ -44,7 +44,7 @@ public:
                                                 std::shared_ptr<const TabletMetadataPB> tablet_metadata,
                                                 bool force_base_compaction);
 
-    bool is_real_time_compaction_strategy(const std::shared_ptr<const TabletMetadataPB>& metadata);
+    static bool is_real_time_compaction_strategy(const std::shared_ptr<const TabletMetadataPB>& metadata);
 
 protected:
     explicit CompactionPolicy(TabletManager* tablet_mgr, std::shared_ptr<const TabletMetadataPB> tablet_metadata,
--- a/be/src/storage/lake/compaction_policy.h
+++ b/be/src/storage/lake/compaction_policy.h
@@ -44,6 +44,8 @@ public:
                                                 std::shared_ptr<const TabletMetadataPB> tablet_metadata,
                                                 bool force_base_compaction);
 
+    bool is_real_time_compaction_strategy(const std::shared_ptr<const TabletMetadataPB>& metadata);
+
 protected:
     explicit CompactionPolicy(TabletManager* tablet_mgr, std::shared_ptr<const TabletMetadataPB> tablet_metadata,
                               bool force_base_compaction)
--- a/be/src/storage/lake/primary_key_compaction_policy.cpp
+++ b/be/src/storage/lake/primary_key_compaction_policy.cpp
@@ -67,8 +67,7 @@ StatusOr<std::unique_ptr<PKSizeTieredLevel>> PrimaryCompactionPolicy::pick_max_l
     if (!calc_score && top_level_ptr->rowsets.size() == 1 &&
         !top_level_ptr->rowsets.top().multi_segment_with_overlapped() && !order_levels.empty()) {
         auto second_level_ptr = std::make_unique<PKSizeTieredLevel>(order_levels.top());
-        //second_level_ptr->merge_top(*top_level_ptr);
-        top_level_ptr->merge_top(*second_level_ptr);
+        top_level_ptr->merge_level(*second_level_ptr);
         order_levels.pop();
     }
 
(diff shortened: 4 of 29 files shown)


be/src/storage/lake/compaction_policy.cpp
```diff
@@ -80,6 +80,12 @@ class SizeTieredCompactionPolicy : public CompactionPolicy {
 
 StatusOr<uint32_t> primary_compaction_score_by_policy(TabletManager* tablet_mgr,
                                                       const std::shared_ptr<const TabletMetadataPB>& metadata) {
+    bool is_real_time = false;
```
luohaha: Make it become a function.
```
bool is_real_time = false;
    if (metadata->has_compaction_strategy() && metadata->compaction_strategy() == CompactionStrategyPB::REAL_TIME) {
        is_real_time = true;
    }
```

be/src/storage/lake/primary_key_compaction_policy.cpp
```diff
@@ -121,13 +133,14 @@ StatusOr<std::vector<int64_t>> PrimaryCompactionPolicy::pick_rowset_indexes(
         rowset_vec.emplace_back(&rowset_pb, stat, i);
     }
     // 2. pick largest score level
-    ASSIGN_OR_RETURN(auto pick_level_ptr, pick_max_level(calc_score, rowset_vec));
+    ASSIGN_OR_RETURN(auto pick_level_ptr, pick_max_level(calc_score && !is_real_time, rowset_vec));
```
luohaha: I think we can remove `calc_score` and always do score calculation within one level?

be/src/storage/lake/primary_key_compaction_policy.cpp
```diff
@@ -58,24 +57,31 @@ StatusOr<std::unique_ptr<PKSizeTieredLevel>> PrimaryCompactionPolicy::pick_max_l
...
-    } else {
-        return top_level_ptr;
+        top_level_ptr->merge_level(*second_level_ptr);
+        order_levels.pop();
     }
+
+    while (!order_levels.empty()) {
+        auto next_level_ptr = std::make_unique<PKSizeTieredLevel>(order_levels.top());
+        order_levels.pop();
+        if (next_level_ptr->get_compact_level() < top_level_ptr->get_compact_level()) {
```
luohaha, line 79: What is `get_compact_level` means? And it seems that `set_compact_level` never been called?

be/src/storage/lake/primary_key_compaction_policy.h
```diff
@@ -75,41 +75,54 @@ class RowsetCandidate {
...
+    // Add other level's rowset into candidate rowsets
+    void add_candidate_rowsets(PKSizeTieredLevel& other) {
+        while (!other.rowsets.empty()) {
+            const auto& top_rowset = other.rowsets.top();
+            candidate_rowsets.push_back(top_rowset);
+            other.rowsets.pop();
+        }
+    }
+
+    void set_compact_leve(int64_t compact_level) { this->compact_level = compact_level; }
```
luohaha: You can remove this.

be/src/storage/lake/primary_key_compaction_policy.cpp
```diff
@@ -58,24 +57,31 @@ StatusOr<std::unique_ptr<PKSizeTieredLevel>> PrimaryCompactionPolicy::pick_max_l
...
         auto second_level_ptr = std::make_unique<PKSizeTieredLevel>(order_levels.top());
-        second_level_ptr->merge_top(*top_level_ptr);
-        return second_level_ptr;
-    } else {
-        return top_level_ptr;
+        top_level_ptr->merge_level(*second_level_ptr);
+        order_levels.pop();
     }
+
+    while (!order_levels.empty()) {
```
luohaha: Maybe it's better to add a BE config which limit the max compact levels.

be/src/storage/lake/primary_key_compaction_policy.h
```diff
@@ -75,41 +75,54 @@ class RowsetCandidate {
...
+
+    int64_t get_compact_level() { return compact_level; }
+
     bool operator<(const PKSizeTieredLevel& other) const { return score < other.score; }
 
     std::priority_queue<RowsetCandidate> rowsets;
+    std::vector<RowsetCandidate> candidate_rowsets;
```
luohaha: I think `other_level_candidate_rowsets` might be clearer?

be/src/storage/lake/primary_key_compaction_policy.cpp
sevev: > Maybe it's better to add a BE config which limit the max compact levels.

done.

be/src/storage/lake/primary_key_compaction_policy.h
sevev: > I think `other_level_rowsets` might be clearer?

update.

be/src/storage/lake/primary_key_compaction_policy.h
sevev: > You can remove this.

done.

luohaha (dismissed):

be/src/storage/lake/compaction_policy.cpp
```diff
@@ -78,19 +78,29 @@ class SizeTieredCompactionPolicy : public CompactionPolicy {
     };
 };
 
+bool CompactionPolicy::is_real_time_compaction_strategy(const std::shared_ptr<const TabletMetadataPB>& metadata) {
+    bool is_real_time = false;
+    if (metadata->has_compaction_strategy() && metadata->compaction_strategy() == CompactionStrategyPB::REAL_TIME) {
+        is_real_time = true;
+    }
+    return is_real_time;
```
kevincai:
```suggestion
    return metadata->has_compaction_strategy() && metadata->compaction_strategy() == CompactionStrategyPB::REAL_TIME;
```

be/src/storage/lake/compaction_policy.h
```diff
@@ -44,6 +44,8 @@ class CompactionPolicy {
                                                 std::shared_ptr<const TabletMetadataPB> tablet_metadata,
                                                 bool force_base_compaction);
 
+    bool is_real_time_compaction_strategy(const std::shared_ptr<const TabletMetadataPB>& metadata);
```
kevincai: the result can be determined by the metadata, so it is perfectly a static function.

be/src/common/config.h
```diff
@@ -396,6 +396,7 @@ CONF_Bool(enable_event_based_compaction_framework, "true");
 
 CONF_Bool(enable_size_tiered_compaction_strategy, "true");
 CONF_mBool(enable_pk_size_tiered_compaction_strategy, "true");
+CONF_mInt32(size_tiered_max_compaction_level, "3");
```
kevincai, line 402: Add comment and possibly a separate PR to add doc to explain the configuration.

be/src/storage/lake/primary_key_compaction_policy.cpp
```diff
@@ -58,24 +57,34 @@ StatusOr<std::unique_ptr<PKSizeTieredLevel>> PrimaryCompactionPolicy::pick_max_l
...
-        second_level_ptr->merge_top(*top_level_ptr);
-        return second_level_ptr;
-    } else {
-        return top_level_ptr;
+        top_level_ptr->merge_level(*second_level_ptr);
+        order_levels.pop();
+        compaction_level++;
     }
+
+    while (!order_levels.empty() && compaction_level <= config::size_tiered_max_compaction_level) {
```
kevincai: shall use a local variable to capture the config var in case its value changed in the middle of the while loop.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java
```diff
@@ -415,6 +415,23 @@ public Void visitModifyTablePropertiesClause(ModifyTablePropertiesClause clause,
...
+        } else if (properties.containsKey(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY)) {
+            if (!properties.get(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY).equalsIgnoreCase("default") &&
+                    !properties.get(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY).equalsIgnoreCase("real_time")) {
+                ErrorReport.reportSemanticException(ErrorCode.ERR_COMMON_ERROR, "unknown compaction strategy: " + 
+                            properties.get(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY));
+            }
+            if (!table.isCloudNativeTable()) {
+                ErrorReport.reportSemanticException(ErrorCode.ERR_COMMON_ERROR,
+                            "Property " + PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY +
+                                    " only support cloud native table");
```
kevincai:
```suggestion
                                    " can be only set to the cloud native table");
```

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java
```diff
@@ -415,6 +415,23 @@ public Void visitModifyTablePropertiesClause(ModifyTablePropertiesClause clause,
                                     " haven't been enabled");
                 }
             }
+        } else if (properties.containsKey(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY)) {
+            if (!properties.get(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY).equalsIgnoreCase("default") &&
+                    !properties.get(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY).equalsIgnoreCase("real_time")) {
+                ErrorReport.reportSemanticException(ErrorCode.ERR_COMMON_ERROR, "unknown compaction strategy: " + 
+                            properties.get(PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY));
+            }
+            if (!table.isCloudNativeTable()) {
```
kevincai: cloudnativetable or materialized view?

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java
```diff
@@ -415,6 +415,23 @@ public Void visitModifyTablePropertiesClause(ModifyTablePropertiesClause clause,
...
+                ErrorReport.reportSemanticException(ErrorCode.ERR_COMMON_ERROR,
+                            "Property " + PropertyAnalyzer.PROPERTIES_COMPACTION_STRATEGY +
+                                    " only support cloud native table");
+            }
+
+            OlapTable olapTable = (OlapTable) table;
+            if (olapTable.getKeysType() != KeysType.PRIMARY_KEYS) {
+                ErrorReport.reportSemanticException(ErrorCode.ERR_COMMON_ERROR, "only primary key table support change " +
```
kevincai: The compaction strategy can be only updated for a primary key table.

fe/fe-core/src/main/java/com/starrocks/server/OlapTableFactory.java
```diff
@@ -378,6 +379,21 @@ public Table createTable(LocalMetastore metastore, Database db, CreateTableStmt
...
+                TCompactionStrategy compactionStrategy;
+                try {
+                    compactionStrategy = PropertyAnalyzer.analyzecompactionStrategy(properties);
+                } catch (AnalysisException e) {
+                    throw new DdlException(e.getMessage());
+                }
+                // REAL_TIME strategy only work for primary key table right now, so set compaction strategy to DEFAULT
+                // if table is not primary key table.
+                if (table.getKeysType() != KeysType.PRIMARY_KEYS && compactionStrategy != TCompactionStrategy.DEFAULT) {
+                    throw new DdlException("non-pk table only support default compaction strategy right now");
```
kevincai: Only `default` compaction strategy is allowed for non-pk table.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJob.java
```diff
@@ -43,6 +43,9 @@ public class LakeTableAlterMetaJob extends LakeTableAlterMetaJobBase {
     @SerializedName(value = "enableFileBundling")
     private boolean enableFileBundling;
 
+    @SerializedName(value = "compactionStrategy")
+    private String compactionStrategy;
```
kevincai, line 47: is `gsonPostProcess()` required to set this to "default" in case the job is replayed from an old version without the persistent info of compactionStrategy, this `compactionStrategy` will be `null` by default.

be/src/common/config.h
sevev: > Add comment and possibly a separate PR to add doc to explain the configuration.

done.

be/src/storage/lake/compaction_policy.cpp
sevev: > 

done.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJob.java
sevev: > is `gsonPostProcess()` required to set this to "default" in case the job is replayed from an old version without the persistent info of compactionStrategy, this `compactionStrategy` will be `null` by default.

It is doesn't matter. If the job is replayed from old version, we don't access the `compactionStrategy`.

kevincai (approved):

luohaha (approved):

sevev:
@mergify rebase

HangyuanLiu (approved):

stephen-shelby (approved):

PR #60368, opened 2025-06-27
[Refactor] enable skip pk preload by default

luohaha:
## Why I'm doing:
When `skip_pk_preload` is false, we will do PK state preload in delta writer finish step, which will cause `Reached Timeout` issue when:
1. PK index rebuild cause too much time.
2. Read from PK index wait for lock for a long time.
3. Preload logic itself cause too much time.

## What I'm doing:
This pull request updates several configurations in the `be/src/common/config.h` file, primarily altering default values for specific settings.

Configuration changes:

* Updated the default value of `CONF_mBool(skip_pk_preload)` from `"false"` to `"true"`, which changes the behavior of preloading primary keys.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1583,7 +1583,7 @@ CONF_Bool(report_python_worker_error, "true");
 CONF_Bool(python_worker_reuse, "true");
 CONF_Int32(python_worker_expire_time_sec, "300");
 CONF_mBool(enable_pk_strict_memcheck, "true");
-CONF_mBool(skip_pk_preload, "false");
+CONF_mBool(skip_pk_preload, "true");
 // Reduce core file size by not dumping jemalloc retain pages
 CONF_mBool(enable_core_file_size_optimization, "true");
 // Current supported modules:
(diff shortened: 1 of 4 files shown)


sevev (dismissed):

wyb (dismissed):

kevincai (dismissed):

kevincai:
do we want to introduce this in v3.3? BugFix only for v3.3, this is a behavior change to v3.3

luohaha:
> do we want to introduce this in v3.3?

Yes, most of our user are using v3.3, we need to fix it for all.

kevincai (approved):

sevev (approved):

alvin-phoenix-ai:
@luohaha 
I don't think it is good to change the default behavior for v3.3 because it is a stable version. 
If this can be fixed by changing the configuration values, we can update the document to let users know how to mitigate when errors happen.

luohaha:
> @luohaha 
> 
> I don't think it is good to change the default behavior for v3.3 because it is a stable version. 
> 
> If this can be fixed by changing the configuration values, we can update the document to let users know how to mitigate when errors happen.
> 
> 

It is very hard to identity these errors, and this change already become the best practice of 3.3

luohaha:
> @luohaha 
> 
> I don't think it is good to change the default behavior for v3.3 because it is a stable version. 
> 
> If this can be fixed by changing the configuration values, we can update the document to let users know how to mitigate when errors happen.
> 
> 

I know your concern, I won't backport it to 3.3

luohaha:
https://github.com/Mergifyio backport branch-3.5

luohaha:
https://github.com/Mergifyio backport branch-3.4

PR #60385, opened 2025-06-27
[Enhancement] use lower_bound/upper_bound to optimize min/max

dirtysalt:
## Why I'm doing:

Icberg provides lower/upper bound for each field.  Maybe we can use that info to do optimization

## What I'm doing:

How to achieve that:
- parse lowe/upper bound at fe side to object
- pass this object down to BE side
- if it's some sql pattern like `select min(a)`, and if we find we have lower_bound value of a, we just output that

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3


be/src/exec/hdfs_scanner.cpp
```diff
@@ -28,9 +29,10 @@
 #include "storage/runtime_range_pruner.hpp"
 #include "util/compression/compression_utils.h"
 #include "util/compression/stream_compression.h"
-
 namespace starrocks {
 
+static const std::string kCountOptColumnName = "___count___";
```
trueeyu, line 34: Should we add the name to https://github.com/StarRocks/starrocks/blob/2e76ca471fc4f771a7a655734fd431fe18d87788/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/FeNameFormat.java#L61

be/src/exec/hdfs_scanner.cpp
```diff
@@ -617,6 +663,90 @@ void HdfsScannerContext::append_or_update_count_column_to_chunk(ChunkPtr* chunk,
...
+        data.emplace_back((double)value.min_int_value * 1e-6);
+        data.emplace_back((double)value.max_int_value * 1e-6);
+        break;
+    default:
+        break;
+    }
+
+    // if this is the first split, we use null/min/max order
+    // otherwise, we reverse it. In that way, we can make sure
+    // null/min/max values all output from this file.
```
trueeyu, line 722: I still don't understand why reverse

be/src/exec/hdfs_scanner.cpp
dirtysalt: OK.

be/src/exec/hdfs_scanner.cpp
dirtysalt: Here's the interesting part: the min/max values of certain columns are not present in the Iceberg metadata.

Currently, we support a fallback mechanism where if a column's min/max values are missing, we can read the original content for that column. However, for columns where min/max values are available, we need to populate the data accordingly.

If a split contains more than 3 rows, we can confidently fill in the min, max, and null values without issues. But what if a split only has 1 or 2 rows? How should we handle the min/max/null values in such cases?

Here’s a proposed approach:

- ​First split: Fill in the values in the order [null, min, max].
- ​Non-first splits: Fill in the values in the order [max, min, null].

This ensures that even if a file has only 1 or 2 rows but is read as multiple splits, we can still output [null, min, max] across all splits. This is somewhat similar to a coordination problem.

trueeyu (dismissed):

be/src/exec/hdfs_scanner.cpp
dirtysalt: Suppose there are two rows: [null, 5],  but this file has been split to 2 scan ranges.

if we don't reverse, the scan ranges output
- first: [null]
- second: [null]

Thats' wrong. but if we reverse it:
- first: [null]
- second:[5]

Tha'ts correct.

This is just for extreme case.  In normal case, it does not matter at all.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/MinMaxOptOnScanRule.java
```diff
@@ -32,23 +33,29 @@
...
-// 'select min(c1),max(c2),count(c3) from table',
+// 'select min(c1),max(c2) from table',
 // we add a label on scan node to indicates that pattern for further optimization
 
-public class MinMaxCountOptOnScanRule extends TransformationRule {
-    public MinMaxCountOptOnScanRule() {
-        // agg -> project -> scan[checked in `check`]
+public class MinMaxOptOnScanRule extends TransformationRule {
+    private static final Set<OperatorType> SUPPORTED = Set.of(
+            OperatorType.LOGICAL_ICEBERG_SCAN
```
Youngwb, line 41: this rule not only used for iceberg, if you chenage like this, how about hive?

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
```diff
@@ -2358,6 +2356,9 @@ public long getConnectorSinkTargetMaxFileSize() {
     @VarAttr(name = ENABLE_COUNT_STAR_OPTIMIZATION, flag = VariableMgr.INVISIBLE)
     private boolean enableCountStarOptimization = true;
 
+    @VarAttr(name = ENABLE_MIN_MAX_OPTIMIZATION)
+    private boolean enableMinMaxOptimization = true;
```
stephen-shelby, line 2360: I think default value should be false. I remember that trino won't record min/max by default

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteSimpleAggToHDFSScanRule.java
```diff
@@ -232,13 +232,6 @@ public boolean check(final OptExpression input, OptimizerContext context) {
             return false;
         }
 
-        if (scanOperator.getOpType() == OperatorType.LOGICAL_ICEBERG_SCAN) {
```
stephen-shelby, line 235: why remove this block?

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
dirtysalt: OK.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteSimpleAggToHDFSScanRule.java
dirtysalt: @stephen-shelby  because I find a way that can handle transformed partition, which is covered in sql-test case

```
-- name: testIcebergCatalog

select count(*) from iceberg_sql_test_${uuid0}.iceberg_oss_db.test_trans_part;

select count(*), col_date from iceberg_sql_test_${uuid0}.iceberg_oss_db.test_trans_part group by col_date order by col_date;

function: assert_explain_costs_contains("select * from iceberg_sql_test_${uuid0}.iceberg_ci_db.day_ts_to_ts;","partitions=2")

```

fe/fe-core/src/main/java/com/starrocks/planner/HdfsScanNode.java
```diff
@@ -203,8 +203,9 @@ public static void appendDataCacheOptionsInExplain(StringBuilder output, String
     public static void setScanOptimizeOptionToThrift(THdfsScanNode tHdfsScanNode, ScanNode scanNode) {
         ScanOptimizeOption option = scanNode.getScanOptimizeOption();
         tHdfsScanNode.setCan_use_any_column(option.getCanUseAnyColumn());
-        tHdfsScanNode.setCan_use_min_max_count_opt(option.getCanUseMinMaxCountOpt());
+        tHdfsScanNode.setCan_use_min_max_opt(option.getCanUseMinMaxOpt());
```
stephen-shelby, line 206: where do you get min/max values about hdfs_scan_node

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergUtil.java
```diff
@@ -14,10 +14,159 @@
...
+        public void toThrift(Map<Integer, TExprMinMaxValue> tExprMinMaxValueMap, SlotDescriptor slot) {
+            TExprMinMaxValue texpr = new TExprMinMaxValue();
+            if (toThrift(slot, texpr)) {
+                tExprMinMaxValueMap.put(slot.getId().asInt(), texpr);
+            }
+        }
+    }
+
+    private static final Set<Type.TypeID> MIN_MAX_SUPPORTED_TYPES = Set.of(
+            Type.TypeID.BOOLEAN,
```
stephen-shelby, line 100: nit: please add a TODO here

stephen-shelby (approved):

be/src/exec/hdfs_scanner.cpp
```diff
@@ -221,6 +234,18 @@ Status HdfsScanner::get_next(RuntimeState* runtime_state, ChunkPtr* chunk) {
         return Status::OK();
     }
 
+    // short circuit for min/max optimization.
+    if (_scanner_ctx.can_use_min_max_optimization()) {
+        // 3 means we output 3 values: min, max, and null
```
Youngwb, line 239: Why output min, max, and null?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/MinMaxOptOnScanRule.java
dirtysalt: hive does not have this metadata, so hive does not use this optimization

we don't mark this to true. 
```
        scanOperator.getScanOptimizeOption().setCanUseMinMaxOpt(true);

```

be/src/exec/hdfs_scanner.cpp
dirtysalt: Because to get right result of min(a)/max(a), you just need to output null(if has null value), min and max value (3 rows).

fe/fe-core/src/main/java/com/starrocks/planner/HdfsScanNode.java
dirtysalt: it's just a general writitng: set that flag to each node to indicate that "if this node can use min/max optimization or not".

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergUtil.java
dirtysalt: OK. will updated in the next pr.

be/src/exec/hdfs_scanner.cpp
```diff
@@ -570,34 +614,36 @@ Status HdfsScannerContext::update_materialized_columns(const std::unordered_set<
...
-        auto* slot_desc = not_existed_slots[0];
-        TypeDescriptor desc;
-        desc.type = TYPE_BIGINT;
-        auto col = ColumnHelper::create_column(desc, slot_desc->is_nullable());
-        col->append_datum(int64_t(1));
-        col->assign(row_count, 0);
-        ck->append_or_update_column(std::move(col), slot_desc->id());
-        ck->set_num_rows(row_count);
-        return Status::OK();
+    if (use_min_max_opt) {
```
Youngwb, line 626: why need to deal use_min_max_opt?  there is short circuit for min/max optimization in getNext

be/src/exec/hdfs_scanner.cpp
```diff
@@ -570,34 +614,36 @@ Status HdfsScannerContext::update_materialized_columns(const std::unordered_set<
...
+        }
+
         auto col = ColumnHelper::create_column(slot_desc->type(), slot_desc->is_nullable());
         if (row_count > 0) {
-            col->append_default(row_count);
+            if (slot_desc->col_name() == kCountOptColumnName) {
+                TypeDescriptor desc;
+                desc.type = TYPE_BIGINT;
+                col = ColumnHelper::create_column(desc, slot_desc->is_nullable());
+                col->append_datum(int64_t(1));
```
Youngwb, line 642: add 1 for `__count__`?

be/src/exec/hdfs_scanner.cpp
dirtysalt: here is the quick check for fast path.

Because if we don't use min/max opt in query(and in most cases we don't), then we don't need to call that function any more.

be/src/exec/hdfs_scanner.cpp
dirtysalt: yes. so in agg node `sum(__count__)` can get the right result.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/MinMaxOptOnScanRule.java
dirtysalt: this rule has neved been used before actually. I wanted to make it work before, but I gave up.

So actually this is the firs time min/max works.

trueeyu (approved):

PR #60400, opened 2025-06-27
[Enhancement] Outer-Join Elimination for duplicate-insensitive aggreg…

### LINKED ISSUE #58945: Outer-Join Elimination for duplicate-insensitive aggregates
**Background**  
StarRocks’ logical-rewrite stage already converts some `LEFT/RIGHT` joins to `INNER` joins but still keeps the join operator even when the inner table contributes *zero semantic value*.  
Other engines (e.g., Oracle, DB2) go one step further and **remove the join entirely** when the query only needs columns from the preserved side and uses aggregates that ignore duplicates.  This optimization is often called *outer-join elimination*.

> Example of the missed opportunity  
> ```sql
> SELECT  MIN(t1.c1), MAX(t1.c1)
> FROM    t1
> LEFT JOIN t2 ON t1.id = t2.id;
> ```  
> Today: both tables are scanned  
> Desired: rewrite to `SELECT MIN(c1), MAX(c1) FROM t1;`

**duplicate-insensitive**
-  `MIN`, `MAX`, `COUNT(DISTINCT)`, `ANY_VALUE`, etc.).  
- But not `SUM`, `AVG`, etc.

**Why it matters**  
* Cuts one full-table scan (and the hash-table build) → less CPU, memory, and network.  
* Measurable in real workloads: removing a large dimension table often gives a 1-2× speed-up.

**Safe rewrite conditions**  
1. Join type is `LEFT` (or `RIGHT` after symmetry).  
2. Join predicates are equal-join, no other filters on the inner side.  
3. The query does **not** reference inner-table columns after the join.  
4. Every aggregate is **duplicate-insensitive** (`MIN`, `MAX`, `COUNT(DISTINCT)`, `ANY_VALUE`, etc.).  
5. (Optional) If the FK can be `NULL`, add `WHERE fk IS NOT NULL` to preserve semantics.

**Implementation sketch**  
* **Rule trigger**: `LogicalAggregate → LogicalJoin(outer, inner)`  
* **Check** conditions above using existing relation metadata and `AggFn.isDuplicateInsensitive()`.  
* **Transform**: replace the `Join` node with the outer child; push a non-null filter on the FK when required.  
* **Cost**: always cheaper (fewer operators, smaller row-set).  

StarRocks’ Cascades-style memo already hosts similar rules (e.g., *Outer Join → Inner*), so the new rule should slot in cleanly.

Mesut-Doner:
## Why I'm doing:

when queries use LEFT/RIGHT JOINs but only reference columns from the preserved (outer) side with duplicate-insensitive aggregates. This resulted in unnecessary table scans and JOIN operations

## What I'm doing:
Implemented `OuterJoinEliminationRule` that eliminates outer JOINs when:
- Query uses duplicate-insensitive aggregates (MIN, MAX, COUNT DISTINCT, etc.)
- Only outer table columns are referenced
- JOIN conditions are simple equalities (FK = PK style)
- No post-JOIN filters exist

image

image

Fixes #58945

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/RuleSet.java
```diff
@@ -85,6 +85,7 @@
 import com.starrocks.sql.optimizer.rule.transformation.MergeTwoFiltersRule;
 import com.starrocks.sql.optimizer.rule.transformation.MergeTwoProjectRule;
 import com.starrocks.sql.optimizer.rule.transformation.MinMaxCountOptOnScanRule;
+import com.starrocks.sql.optimizer.rule.transformation.OuterJoinEliminationRule;
```
stephen-shelby, line 88: please add some unit tests for this patch first

Mesut-Doner: Thanks @stephen-shelby I've tried to add UT.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * OuterJoinEliminationRule implements the outer-join elimination optimization.
+ */
+public class OuterJoinEliminationRule extends TransformationRule {
+    private static final Logger LOG = LogManager.getLogger(OuterJoinEliminationRule.class);
+
+    public static final OuterJoinEliminationRule INSTANCE = new OuterJoinEliminationRule();
+
+    private static final Set<String> DUPLICATE_INSENSITIVE_FUNCTIONS = ImmutableSet.of(
+            "min", "max", "any_value", "first_value", "last_value",
```
murphyatwork: Please refer to the variables within `FunctionSet`

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
```diff
@@ -0,0 +1,190 @@
...
+import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
+import com.starrocks.sql.optimizer.rule.RuleType;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * OuterJoinEliminationRule implements the outer-join elimination optimization.
```
murphyatwork, line 40: Please add an example here

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
```diff
@@ -0,0 +1,190 @@
...
+                    ColumnRefSet usedColumns = argument.getUsedColumns();
+                    if (!outerColumns.containsAll(usedColumns)) {
+                        return false;
+                    }
+                }
+            }
+
+            return true;
+        } catch (Exception e) {
+            LOG.warn("Failed to check outer join elimination applicability", e);
```
murphyatwork: Why? Generally, we don't recommend catching all exceptions in an optimizer, as it can obscure potential bugs.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
Mesut-Doner: Thanks! I've modified...

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
Mesut-Doner: Thanks! I've added...

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
Mesut-Doner: Thanks!. You're right, I got it. I've inspected other rules and found no reason for adding try catching. So I've removed unnecessary try-catch blocks.

murphyatwork:
@Mesut-Doner could you please rebase the main branch and fix the code conflicts ? We've just upgrade the junit 4 to junit 5, which cause a compilation error to this PR

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
```diff
@@ -0,0 +1,197 @@
...
+
+        return false;
+    }
+
+    /**
+     * only strict foreign key equality is allowed.
+     * for example
+     * o.customer_id = c.customer_id allowed
+     */
+    private boolean isStrictForeignKeyEquality(ScalarOperator joinCondition) {
```
stephen-shelby: why is it called isStrictForeignKeyEquality?

stephen-shelby:
@mergifyio rebase

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/OuterJoinEliminationRule.java
Mesut-Doner: hi @stephen-shelby you're right. Current situation it sound like its new type of fk. 
I've changed function name.

Mesut-Doner:
> @Mesut-Doner could you please rebase the main branch and fix the code conflicts ? We've just upgrade the junit 4 to junit 5, which cause a compilation error to this PR

thanks, I rebased locally.

stdpain (approved):

murphyatwork:
@Cursor review

murphyatwork (approved):

stephen-shelby (approved):

Issue #56855, opened 2025-03-13
[Partition Prune] Wrong result for partition predicate


LiShuMing:
### Steps to reproduce the behavior (Required)
```
CREATE TABLE t0(
 ts datetime,
 v1 INT,
 v2 INT)
 DUPLICATE KEY(ts)
 PARTITION BY date_trunc('day', ts)
DISTRIBUTED BY HASH(ts);
-- result:
-- !result
INSERT INTO t0 VALUES 
  ('2020-01-22 12:12:12', 0,1),
  ('2020-02-23 12:12:12',1,1),
  ('2020-03-24 12:12:12',1,2),
  ('2020-04-25 12:12:12',3,3),
  ('2020-05-22 12:12:12', 0,1),
  ('2020-06-23 12:12:12',1,1),
  ('2020-07-24 12:12:12',1,2),
  ('2020-08-24 12:12:12',1,2),
  ('2020-09-24 12:12:12',1,2),
  ('2020-10-25 12:12:12',3,3);

-- ! | 2020-02-23 12:12:12 | 1|1| should be hit?

mysql> select * from t0 where ts > '2020-02-23 12:12:00' and date_trunc('hour', ts) <= '2020-02-23 12:12:00'  order by 1;
Empty set (0.06 sec)

mysql> select  ts > '2020-02-23 12:12:00' and date_trunc('hour', ts) <= '2020-02-23 12:12:00' from t0;
+--------------------------------------------------------------------------------------+
| (ts > '2020-02-23 12:12:00') AND ((date_trunc('hour', ts)) <= '2020-02-23 12:12:00') |
+--------------------------------------------------------------------------------------+
|                                                                                    1 |
|                                                                                    0 |
|                                                                                    0 |
|                                                                                    0 |
|                                                                                    0 |
|                                                                                    0 |
|                                                                                    0 |
|                                                                                    0 |
|                                                                                    0 |
|                                                                                    0 |
+--------------------------------------------------------------------------------------+

```

- maybe there is some bug in partition pruning.
```
mysql> explain select * from t0 where ts > '2020-02-23 12:12:00' and date_trunc('hour', ts) <= '2020-02-23 12:12:00'  order by 1;
+----------------------------------------------------------------------------------------------------+
| Explain String                                                                                     |
+----------------------------------------------------------------------------------------------------+
| PLAN FRAGMENT 0                                                                                    |
|  OUTPUT EXPRS:1: ts | 2: v1 | 3: v2                                                                |
[...]

PR #60467, opened 2025-07-01
[BugFix] fix error tablet pruning

stephen-shelby:
## Why I'm doing:

When the low and upper bounds derived from bucket column predicates are equal, we can precisely calculate whether a bucket can be hit through the hash function. However, currently for predicates where the left-hand side is func(bucket_column), we directly use the right-hand value as the predicate for bucket_column to derive low and upper bounds. This is incorrect and can easily hit wrong buckets, causing incorrect results.

## What I'm doing:
PartitionColumnFilter records whether it is derived from a function call expression. If true, tablet pruning will be disabled.
Since we have applied some base rewrite rule before DistributionPruneRule, we can consider that when the left-hand side of a predicate is not a columnRefOperator, tablet pruning should not be performed.

Fixes #issue
https://github.com/StarRocks/starrocks/issues/56855

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/HashDistributionPruner.java
@@ -108,7 +108,8 @@ public class HashDistributionPruner implements DistributionPruner {
             // equal one value
             if (filter.lowerBoundInclusive && filter.upperBoundInclusive
                     && lowerBound != null && upperBound != null
-                    && 0 == lowerBound.compareLiteral(upperBound)) {
+                    && 0 == lowerBound.compareLiteral(upperBound)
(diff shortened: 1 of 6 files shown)


stephen-shelby:
@mergifyio rebase

stephen-shelby:
@mergifyio rebase

LiShuMing (approved):

Seaven (approved):

PR #60484, opened 2025-07-01
[Enhancement] min/max optimization can be applied when group partition column

dirtysalt:
## Why I'm doing:

Original PR: #60385

Actually `min/max` optimization can be used when group partition column.

Because for each iceberg data file: partition column is a fixed value, and when doing group by this partition column,  we can still use (null/min/max) rows.

## What I'm doing:

Fixes #60385

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/IcebergMetadataScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/IcebergMetadataScanNode.java
@@ -138,7 +138,6 @@ public class IcebergMetadataScanNode extends ScanNode {
         msg.node_type = TPlanNodeType.HDFS_SCAN_NODE;
         THdfsScanNode tHdfsScanNode = new THdfsScanNode();
         tHdfsScanNode.setTuple_id(desc.getId().asInt());
-        tHdfsScanNode.setCan_use_min_max_opt(false);
 
         String explainString = getExplainString(conjuncts);
         LOG.info("Explain string: " + explainString);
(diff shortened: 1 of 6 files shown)


fe/fe-core/src/main/java/com/starrocks/planner/IcebergMetadataScanNode.java
```diff
@@ -138,7 +138,6 @@ protected void toThrift(TPlanNode msg) {
         msg.node_type = TPlanNodeType.HDFS_SCAN_NODE;
         THdfsScanNode tHdfsScanNode = new THdfsScanNode();
         tHdfsScanNode.setTuple_id(desc.getId().asInt());
-        tHdfsScanNode.setCan_use_min_max_opt(false);
```
dirtysalt, line 141: no need to set it.

fe/fe-core/src/main/java/com/starrocks/planner/OdpsScanNode.java
```diff
@@ -172,7 +172,6 @@ protected void toThrift(TPlanNode msg) {
         msg.node_type = TPlanNodeType.HDFS_SCAN_NODE;
         THdfsScanNode tHdfsScanNode = new THdfsScanNode();
         tHdfsScanNode.setTuple_id(desc.getId().asInt());
-        tHdfsScanNode.setCan_use_min_max_opt(false);
```
dirtysalt, line 175: no need to set it.

Youngwb (approved):

stephen-shelby (approved):

PR #60517, opened 2025-07-02
[Enhancement] Add lake service stub cache

sevev:
## Why I'm doing:
After support `file_bundling`,  we will create brpc channels between CN nodes during each publish operation which may affect publish performance. 

## What I'm doing:
Add lake service stub cache to avoid creating brpc channels on each publish

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/service/service_be/lake_service.cpp
+++ b/be/src/service/service_be/lake_service.cpp
@@ -290,9 +290,10 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
     TEST_SYNC_POINT("LakeServiceImpl::publish_version:return");
 }
 
-struct PublishRequestCtx {
+template <typename ResponseType>
+struct RequestContext {
     std::unique_ptr<brpc::Controller> cntl;
-    std::unique_ptr<PublishVersionResponse> resp;
+    std::unique_ptr<ResponseType> resp;
 };
 
 struct AggregatePublishContext {
@@ -303,6 +304,7 @@ struct AggregatePublishContext {
     PublishVersionResponse* response;
     Status publish_status = Status::OK();
 
+    using PublishRequestCtx = RequestContext<PublishVersionResponse>;
     std::vector<PublishRequestCtx> publish_request_ctx;
 
     void handle_failure(const std::string& error) {
@@ -968,18 +970,14 @@ void LakeServiceImpl::compact(::google::protobuf::RpcController* controller, con
     _tablet_mgr->compaction_scheduler()->compact(controller, request, response, guard.release());
 }
 
-struct PendingCompactRequestCtx {
-    std::unique_ptr<brpc::Controller> cntl;
-    std::unique_ptr<CompactResponse> resp;
-};
-
 struct AggregateCompactContext {
     bthread::Mutex response_mtx;
     Status final_status = Status::OK();
     std::unique_ptr<BThreadCountDownLatch> latch;
     CombinedTxnLogPB combined_txn_log;
 
-    std::vector<PendingCompactRequestCtx> compact_request_ctx;
+    using CompactRequestCtx = RequestContext<CompactResponse>;
+    std::vector<CompactRequestCtx> compact_request_ctx;
 
     void handle_failure(const std::string& error) {
         std::lock_guard l(response_mtx);
--- a/be/src/service/service_be/lake_service.cpp
+++ b/be/src/service/service_be/lake_service.cpp
@@ -36,6 +36,7 @@
 #include "storage/lake/transactions.h"
 #include "storage/lake/vacuum.h"
 #include "testutil/sync_point.h"
+#include "util/brpc_stub_cache.h"
 #include "util/countdown_latch.h"
 #include "util/defer_op.h"
 #include "util/thread.h"
@@ -378,20 +379,20 @@ void LakeServiceImpl::aggregate_publish_version(::google::protobuf::RpcControlle
     ctx.latch = std::make_unique<BThreadCountDownLatch>(request->publish_reqs_size());
 
     for (int i = 0; i < request->publish_reqs_size(); ++i) {
+        if (ctx.has_failure) {
+            ctx.count_down();
+            continue;
+        }
+
         const auto timeout_ms = request->publish_reqs(i).has_timeout_ms() ? request->publish_reqs(i).timeout_ms()
                                                                           : kDefaultTimeoutForPublishVersion;
         const auto& compute_node = request->compute_nodes(i);
         const auto& single_req = request->publish_reqs(i);
 
-        butil::EndPoint endpoint;
-        std::string brpc_url = fmt::format("{}:{}", compute_node.host(), compute_node.brpc_port());
-        if (str2endpoint(brpc_url.c_str(), &endpoint)) {
-            LOG(WARNING) << "unknown endpoint, host=" << compute_node.host();
-#ifndef BE_TEST
-            ctx.handle_failure(fmt::format("unknown endpoint, host={}", compute_node.host()));
-#endif
-        }
-        if (ctx.has_failure) {
+        auto res = LakeServiceBrpcStubCache::getInstance()->get_stub(compute_node.host(), compute_node.brpc_port());
+        if (!res.ok()) {
+            LOG(WARNING) << "aggregate publish failed because get stub failed: " << res.status();
+            ctx.handle_failure(fmt::format("get stub failed: ", res.status().to_string()));
             ctx.count_down();
             continue;
         }
@@ -399,15 +400,8 @@ void LakeServiceImpl::aggregate_publish_version(::google::protobuf::RpcControlle
         auto* node_cntl = new brpc::Controller();
         auto* node_resp = new PublishVersionResponse();
         node_cntl->set_timeout_ms(timeout_ms);
-
-        std::unique_ptr<brpc::Channel> channel(new brpc::Channel());
-        brpc::ChannelOptions options;
-        options.connect_timeout_ms = config::rpc_connect_timeout_ms;
-        channel->Init(endpoint, &options);
-        auto stub = std::make_shared<starrocks::LakeService_Stub>(channel.release(),
-                                                                  google::protobuf::Service::STUB_OWNS_CHANNEL);
-        stub->publish_version(node_cntl, &single_req, node_resp,
-                              brpc::NewCallback(aggregate_publish_cb, node_cntl, node_resp, &ctx));
+        (*res)->publish_version(node_cntl, &single_req, node_resp,
+                                brpc::NewCallback(aggregate_publish_cb, node_cntl, node_resp, &ctx));
     }
 
     // wait for publish task finish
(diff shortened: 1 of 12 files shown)


andyziye:
@Mergify rebase

be/src/util/brpc_stub_cache.h
```diff
@@ -179,4 +180,57 @@ class HttpBrpcStubCache {
...
 };
 
+class LakeServiceBrpcStubCache {
+public:
+    static LakeServiceBrpcStubCache* getInstance() {
+        static LakeServiceBrpcStubCache cache;
+        return &cache;
+    }
+
+    StatusOr<std::shared_ptr<starrocks::LakeService_Stub>> get_stub(const std::string& host, int port) {
```
alvin-phoenix-ai: Implementation should be written in cpp files.

be/src/util/brpc_stub_cache.h
```diff
@@ -179,4 +180,57 @@ class HttpBrpcStubCache {
...
+        // we use `connection_group` to create multiple single connections to remove this bottleneck.
+        options.connection_group = std::to_string(_stub_map.size());
+        options.connection_type = config::brpc_connection_type;
+        std::unique_ptr<brpc::Channel> channel(new brpc::Channel());
+        if (channel->Init(endpoint, &options)) {
+            return Status::RuntimeError("brpc channel init failed");
+        }
+        auto stub = std::make_shared<starrocks::LakeService_Stub>(channel.release(),
+                                                                  google::protobuf::Service::STUB_OWNS_CHANNEL);
+        _stub_map.insert(endpoint, stub);
```
alvin-phoenix-ai: The lock should be released when trying to create another link.

be/src/util/brpc_stub_cache.h
```diff
@@ -179,4 +180,57 @@ class HttpBrpcStubCache {
...
+        brpc_url = get_host_port(realhost, port);
+        if (str2endpoint(brpc_url.c_str(), &endpoint)) {
+            return Status::RuntimeError("unknown endpoint, host = " + host);
+        }
+        // get if exist
+        std::lock_guard<SpinLock> l(_lock);
+        auto stub_ptr = _stub_map.seek(endpoint);
+        if (stub_ptr != nullptr) {
+            return *stub_ptr;
+        }
```
alvin-phoenix-ai:
```suggestion
        {
            std::lock_guard<SpinLock> l(_lock);
            auto stub_ptr = _stub_map.seek(endpoint);
            if (stub_ptr != nullptr) {
                return *stub_ptr;
            }
        }
```
Keep the lock as minimal as possible.

be/src/util/brpc_stub_cache.h
sevev: > Implementation should be written in cpp files.

done, split `brpc_stub_cache.h` to `brpc_stub_cache.h` and `brpc_stub_cache.cpp`

be/src/util/brpc_stub_cache.h
sevev: > The lock should be released when trying to create another link.

update.

be/src/service/service_be/lake_service.cpp
```diff
@@ -378,36 +379,29 @@ void LakeServiceImpl::aggregate_publish_version(::google::protobuf::RpcControlle
...
-
-        std::unique_ptr<brpc::Channel> channel(new brpc::Channel());
-        brpc::ChannelOptions options;
-        options.connect_timeout_ms = config::rpc_connect_timeout_ms;
-        channel->Init(endpoint, &options);
-        auto stub = std::make_shared<starrocks::LakeService_Stub>(channel.release(),
-                                                                  google::protobuf::Service::STUB_OWNS_CHANNEL);
-        stub->publish_version(node_cntl, &single_req, node_resp,
-                              brpc::NewCallback(aggregate_publish_cb, node_cntl, node_resp, &ctx));
+        (*res)->publish_version(node_cntl, &single_req, node_resp,
```
kevincai: not sure if this is right. but I sense a potential mem leak here if something wrong with the publish_version call, where the brpc::NewCallback won't trigger the `aggregate_publish_cb` callback and then the  `node_cntl` and `node_resp` will be lost, no one will be responsible for release the mem.

be/src/util/brpc_stub_cache.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "gen_cpp/internal_service.pb.h"
+#include "gen_cpp/lake_service.pb.h"
+#include "util/failpoint/fail_point.h"
+#include "util/starrocks_metrics.h"
+
+namespace starrocks {
+
+BrpcStubCache::BrpcStubCache() {
+    _stub_map.init(239);
+    REGISTER_GAUGE_STARROCKS_METRIC(brpc_endpoint_stub_count, [this]() {
```
kevincai, line 27: need to avoid the race condition that the metric is pulling and the BrpcStubCache is destroyed.

Also the metric should be registered only once and here the BrpcStubCache is not designed to be singleton from the class interfaces.

be/src/util/brpc_stub_cache.cpp
```diff
@@ -0,0 +1,193 @@
...
+    brpc::ChannelOptions options;
+    options.connect_timeout_ms = config::rpc_connect_timeout_ms;
+    options.max_retry = 3;
+    options.connection_group = std::to_string(_stub_map.size());
+    options.connection_type = config::brpc_connection_type;
+    std::unique_ptr<brpc::Channel> channel(new brpc::Channel());
+    if (channel->Init(endpoint, &options)) {
+        return Status::RuntimeError("brpc channel init failed");
+    }
+    auto stub = std::make_shared<starrocks::LakeService_Stub>(channel.release(),
```
kevincai: need to consider the similar ability of RecoverableStub by the PInternalService, it will be anonying the stub hanging and can't recover itself.

be/src/util/brpc_stub_cache.h
```diff
@@ -32,99 +32,37 @@
 // specific language governing permissions and limitations
 // under the License.
 
-#pragma once
```
kevincai: removed by mistake?

be/src/service/service_be/lake_service.cpp
sevev: > not sure if this is right. but I sense a potential mem leak here if something wrong with the publish_version call, where the brpc::NewCallback won't trigger the `aggregate_publish_cb` callback and then the `node_cntl` and `node_resp` will be lost, no one will be responsible for release the mem.

According to the official documentation of brpc(https://github.com/apache/brpc/blob/master/docs/cn/client.md#%E5%BC%82%E6%AD%A5%E8%AE%BF%E9%97%AE), the callback will always be invoked. However, as you've pointed out, if there are latent issues in brpc that prevent the callback from being invoked, memory leaks will indeed occur.  I will add a defensive mechanism to ensure that memory leaks won't occur.

be/src/util/brpc_stub_cache.cpp
sevev: > need to consider the similar ability of RecoverableStub by the PInternalService, it will be anonying the stub hanging and can't recover itself.

update.

be/src/service/service_be/lake_service.cpp
sevev: > not sure if this is right. but I sense a potential mem leak here if something wrong with the publish_version call, where the brpc::NewCallback won't trigger the `aggregate_publish_cb` callback and then the `node_cntl` and `node_resp` will be lost, no one will be responsible for release the mem.

update.

be/src/service/service_be/lake_service.cpp
```diff
@@ -289,6 +290,11 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
     TEST_SYNC_POINT("LakeServiceImpl::publish_version:return");
 }
 
+struct PublishRequestCtx {
```
kevincai: `PublishRequestCtx` and the following `PendingCompactRequestCtx` are both for resource management,  it can be designed as a generic class. One is to use template

```
template<typename PBType>
struct RequestCtx {
    std::unique_ptr<brpc::Controller> cntl;
    std::unique_ptr<PBType> resp;
};
```

or using base message type

```
struct RequestCtx {
    std::unique_ptr<brpc::Controller> cntl;
    std::unique_ptr<protobuf::Message> resp;
};
```

be/src/util/base_recoverable_stub.h
```diff
@@ -0,0 +1,98 @@
...
+        return reset_channel_impl(std::move(channel), next_connection_group);
+    }
+
+protected:
+    virtual Status reset_channel_impl(std::unique_ptr<brpc::Channel> channel, int64_t next_connection_group) = 0;
+
+    butil::EndPoint _endpoint;
+    std::string _protocol;
+    std::atomic<int64_t> _connection_group;
+    mutable std::shared_mutex _mutex;
```
kevincai: not used in base class, a good sign to move it to derived classes.

be/src/util/base_recoverable_stub.h
```diff
@@ -0,0 +1,98 @@
...
+        if (channel->Init(_endpoint, &options)) {
+            LOG(WARNING) << "Fail to init channel " << _endpoint;
+            return Status::InternalError("Fail to init channel");
+        }
+
+        return reset_channel_impl(std::move(channel), next_connection_group);
+    }
+
+protected:
+    virtual Status reset_channel_impl(std::unique_ptr<brpc::Channel> channel, int64_t next_connection_group) = 0;
```
kevincai: This interface is not very clear what to do for derived class.

The `reset_channel()` call is split into `reset_channel_base()` + `options_customizer()` + `reset_channel_impl()` just for code reuse. It is not every intuitive.

One way is, base class provide a few small functions and the sub class uses these shared functions to assemble its reset_channel() implementation.
The other way is the base class implements the `reset_channel()`, leave a few virtual interfaces to customizing derived classes behavior change. However, the expectations of the virtual interfaces should be very clear.

be/src/service/service_be/lake_service.cpp
sevev: > `PublishRequestCtx` and the following `PendingCompactRequestCtx` are both for resource management, it can be designed as a generic class. One is to use template
> 
> ```
> template<typename PBType>
> struct RequestCtx {
>     std::unique_ptr<brpc::Controller> cntl;
>     std::unique_ptr<PBType> resp;
> };
> ```
> 
> or using base message type
> 
> ```
> struct RequestCtx {
>     std::unique_ptr<brpc::Controller> cntl;
>     std::unique_ptr<protobuf::Message> resp;
> };
> ```

done.

kevincai (approved):

alvin-phoenix-ai (approved):

PR #60536, opened 2025-07-02
[BugFix] Fix vacuum across metadata switch version

sevev:
## Why I'm doing:
If we change table property `file_bundling`, we will set  `metadataSwitchVersion` to the latest version and vacuum should not across `metadataSwitchVersion` because the metadata formats  is different before and after the `metadataSwitchVersion`.  But when the `minRetainVersion` is 0, we will skip `metadataSwitchVersion` and use partition visible version as the vacuum version which across `metadataSwitchVersion`. And the result is some data file will not be gc successfully.

## What I'm doing:
Fix the minRetainVersion selection bug.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
@@ -214,10 +214,12 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
 
     public long getMinRetainVersion() {
         long retainVersion = minRetainVersion;
-        if (metadataSwitchVersion != 0 && retainVersion != 0) {
-            retainVersion = Math.min(retainVersion, metadataSwitchVersion);
-        } else if (metadataSwitchVersion != 0) {
-            retainVersion = metadataSwitchVersion;
+        if (metadataSwitchVersion != 0) {
+            if (retainVersion != 0) {
+                retainVersion = Math.min(retainVersion, metadataSwitchVersion);
+            } else {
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
```diff
@@ -214,8 +214,10 @@ public void setLastVacuumTime(long lastVacuumTime) {
 
     public long getMinRetainVersion() {
         long retainVersion = minRetainVersion;
-        if (metadataSwitchVersion != 0) {
+        if (metadataSwitchVersion != 0 && retainVersion != 0) {
             retainVersion = Math.min(retainVersion, metadataSwitchVersion);
+        } else if (metadataSwitchVersion != 0) {
+            retainVersion = metadataSwitchVersion;
         }
```
xiangguangyxg: could refactor ? there are two `metadataSwitchVersion != 0`

sevev: > could refactor ? there are two `metadataSwitchVersion != 0`

done.

luohaha (approved):

xiangguangyxg (approved):

wyb (approved):

PR #60550, opened 2025-07-03
[BugFix] Fix incorrect matching when multiple global UDFs with the same name exist

stephen-shelby:
## Why I'm doing:
Currently, we use `IS_NONSTRICT_SUPERTYPE_OF` mode to match global udf. This means using the most lenient strategy for matching.  If a global udf is `test_udf(string)`, selectItem like `udf_test(123)` will match it. If user create two global udf with same function_name but different parameter types. we may match incorrectly. for example, user create two udfs contains `test_udf(string)` and `test_udf(int)`. the query `select test_udf(123), test_udf('123')` match test_udf(string).
```
|   |  <slot 2> : udf_test('123')         |
|   |  <slot 3> : udf_test('123')         | // all use the string version
```
This will lead to incorrect results 

## What I'm doing:
since NumericType can be converted to StringType and StringType can't be converted to NumericType when matching udf, Considering that the function matching process is particularly common, we handle it when creating global UDFs. So when creating UDFs, we sort UDFs with the same name, with numeric types preceding varchar types. This patch only UDFs with a single parameter are processed. 

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/GlobalFunctionMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/GlobalFunctionMgr.java
@@ -32,6 +32,7 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * global function manager
@@ -97,6 +98,7 @@ public class GlobalFunctionMgr {
             }
             assignIdToUserDefinedFunction(function);
         }
+
         name2Function.put(functionName, addOrReplaceFunction(function, existFuncs));
     }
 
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/GlobalFunctionMgr.java
```diff
@@ -109,12 +111,27 @@ private void addFunction(Function function, boolean isReplay, boolean allowExist
...
-                        .collect(ImmutableList.toImmutableList()))
-                .add(function)
-                .build();
+        List<Function> filteredFuncs = existFuncs.stream()
+                .filter(f -> !function.compare(f, Function.CompareMode.IS_IDENTICAL))
+                .collect(Collectors.toList());
+
+        filteredFuncs.add(function);
+
+        filteredFuncs.sort((f1, f2) -> {
```
trueeyu, line 120: If the parameter types are float and Int, how does this match?

stephen-shelby: According to the order in which UDFs are added

trueeyu: That doesn't meet expectations either?

Seaven (approved):

HangyuanLiu (approved):

fe/fe-core/src/main/java/com/starrocks/catalog/GlobalFunctionMgr.java
stephen-shelby: I will handle it in the next PR

PR #60608, opened 2025-07-04
[BugFix]fix ngram_search use after free 

before-Sunrise:
## Why I'm doing:
ngram_search has a map in function's state, key is driver id and value is per-thread vector, every thread will get or create its own vector and read/modify this vector. Everything goes well if in pipeline thread pool. But if ngram_search is pushed down into storage engine, driver id is not unique for each thread. Two thread can have same thread id, and both create a new vector, so one thread's vector will be replaced and this thread will still visit this vector, cause use-after-free
  
## What I'm doing:
1. use thread-id instead of driver-id as map's key
2. use phmap's lazy_emplace_l, lazy_emplace_l first arg is key, second arg is lambda which will be called if this key is already exist in map, third arg is lambda  which will be called if this key doens't exist in map

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/be/src/exprs/ngram.cpp
+++ b/be/src/exprs/ngram.cpp
@@ -45,7 +45,6 @@ struct Ngramstate {
         std::thread::id current_thread_id = std::this_thread::get_id();
 
         std::vector<NgramHash>* result = nullptr;
-        driver_maps.if_contains(current_thread_id, [&](const auto& value) { result = value.get(); });
         driver_maps.lazy_emplace_l(
                 current_thread_id, [&](const auto& value) { result = value.get(); },
                 [&](auto build) {
--- a/be/src/exprs/ngram.cpp
+++ b/be/src/exprs/ngram.cpp
@@ -211,8 +211,7 @@ private:
             DCHECK(needle_not_overlap_with_haystack <= needle_gram_count);
 
             // now get the result
-            double row_result =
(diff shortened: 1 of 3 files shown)


satanson (approved):

be/src/exprs/ngram.cpp
```diff
@@ -42,17 +42,17 @@ struct Ngramstate {
     float result = -1;
 
     std::vector<NgramHash>* get_or_create_driver_hashmap() {
-        int32_t driver_id = CurrentThread::current().get_driver_id();
+        std::thread::id current_thread_id = std::this_thread::get_id();
```
kangkaisen, line 45: if so, why not thread local?

before-Sunrise: > if so, why not thread local?
this is stored in Function State, and in Function state, we create a vector for one thread, use thread id as key. But Function State is a instance level object, so different query has different Function State. If using thread local variable, then different query can share the same vector, which is not what we want

kangkaisen (approved):

PR #60613, opened 2025-07-04
[Refactor] Refactor dropTable method to accept Table object directly and remove isSetIfExists parameter

crossoverJie:
## What I'm doing:

- Use `table` instead of `tableName`
	- 	When dropping the `MV`, the `table` has already been obtained, so there is no need to acquire the table again through locking.
- Remove the `isSetIfExists` parameter
	- 	When dropping the `MV`, the table must exist when calling the `com.starrocks.catalog.Database#dropTable` function.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/Database.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/Database.java
@@ -256,17 +256,15 @@ public class Database extends MetaObject implements Writable {
         }
     }
 
-    public void dropTable(String tableName, boolean isSetIfExists, boolean isForce) throws DdlException {
-        Table table;
+    public void dropTable(Table table, boolean isSetIfExists, boolean isForce) throws DdlException {
+        if (table == null && isSetIfExists) {
+            return;
+        }
         Locker locker = new Locker();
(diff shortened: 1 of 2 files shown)


alvin-phoenix-ai (changes requested):
The current interface is more general, it doesn't need the caller to have the Table object. So I don't know the necessity to make this change.

crossoverJie:
@alvin-celerdata The two parameters, `tableName` and `isSetIfExists`, are confusing; I believe the readability is improved after the modification.

alvin-phoenix-ai:
> @alvin-celerdata The two parameters, `tableName` and `isSetIfExists`, are confusing; I believe the readability is improved after the modification.

Hmmm, I still think we should keep the original way. Even the table is given this function still needs to check if this table is belong to this database.

PR #60637, opened 2025-07-05
[Feature] add new function to_datetime and to_datetime_ntz

### LINKED ISSUE #57988: [Function] unix_to_datetime, unix_ms_to_datetime
## Feature request

StarRocks now supports milliseconds, but we do not have a way to convert unix ms time to datetime. 

when it comes to unix time, there are two ways to specify milliseconds. most system will be either using float like 159830640.123 , or append millisecond in the last like 1598306400123 to represent datetime with milliseconds.

In StarRocks, we use the standard UNIX time without any milli or micro , nano seconds. However, there may be some case that user import data from other system who use other kinds of UNIX time format that starrocks cannot recognize and need further transformation.

### unix_to_datetime, unix_ms_to_datetime
```
unix_to_datetime(unix_time)
unix_ms_to_datetime(unix_time)

```

example:

```
select unix_to_datetime(1598306400)

2007-11-30 16:30:19
```

2.  
```
select unix_ms_to_datetime(1598306400123)

2007-12-01 12:30:19.123000

-- when it comes to float unix time, user should manually times 1000
select unix_ms_to_datetime(1598306400.123 * 1000)

```

liubotao:
## Why I'm doing:

https://github.com/StarRocks/starrocks/issues/57988

## What I'm doing:

> Convert a Unix epoch timestamp to a `DATETIME` value with timezone-aware and timezone-neutral options.

StarRocks provides two functions for converting Unix timestamps into `DATETIME` values, offering different timezone handling behaviors:

### `to_datetime` (Timezone-aware)
Converts a Unix timestamp—the number of seconds elapsed since **1970-01-01 00:00:00 UTC**—into a StarRocks `DATETIME` using the **current session timezone**.

### `to_datetime_ntz` (Timezone-neutral) 
Converts a Unix timestamp into a StarRocks `DATETIME` treating it as **UTC+0 time** without any timezone conversion.

## Function Signatures:

Both functions accept the same parameters:

| Argument  | Description                                                                                                           |
| --------- | --------------------------------------------------------------------------------------------------------------------- |
| `unix_ts` | **Required.** The Unix timestamp. Accepts `BIGINT`.                                                                   |
| `scale`   | **Optional.** The time-unit scale:<br>  • `0` → seconds (default)<br>  • `3` → milliseconds<br>  • `6` → microseconds |

If `scale` is omitted, the value is treated as seconds. Use `3` or `6` to interpret the input as milliseconds or microseconds.

## Key Differences:

| Function | Timezone Handling | Example Result (Session TZ: Asia/Shanghai) |
|----------|-------------------|---------------------------------------------|
| `to_datetime` | Uses session timezone | `2020-08-25 06:00:00` (UTC+8) |
| `to_datetime_ntz` | Always UTC+0 | `2020-08-24 22:00:00` (UTC+0) |

## Return Values:

Both functions return **`NULL`** when either of the following is true:
* `scale` is not one of `{0, 3, 6}`
* The resulting datetime falls outside the valid range `0001-01-01` – `9999-12-31`

## Usage Examples:

```sql
-- Set session timezone to Asia/Shanghai (UTC+8)
SET time_zone = 'Asia/Shanghai';

-- Same Unix timestamp, different results
SELECT to_datetime(1598306400, 0);     -- 2020-08-25 06:00:00 (UTC+8)
SELECT to_datetime_ntz(1598306400, 0); -- 2020-08-24 22:00:00 (UTC+0)

-- With microsecond precision
SELECT to_datetime(1598306400123456, 6);     -- 2020-08-25 06:00:00.123456
SELECT to_datetime_ntz(1598306400123456, 6); -- 2020-08-24 22:00:00.123456
```

## When to Use Which:

- **Use `to_datetime`** when you need timezone-aware conversion that respects user's local timezone settings
- **Use `to_datetime_ntz`** when you need consistent UTC-based results regardless of session timezone, or when working with systems that expect UTC timestamps

Fixes #57988 

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


be/src/exprs/time_functions.cpp
```diff
@@ -1886,6 +1888,176 @@ StatusOr<ColumnPtr> TimeFunctions::from_unix_to_datetime_with_format_32(Function
...
+            }
+        } else {
+            scale_viewer.emplace(columns[1]);
+        }
+    }
+
+    ColumnViewer<TIMESTAMP_TYPE> timestamp_viewer(columns[0]);
+    ColumnBuilder<TYPE_DATETIME> result(size);
+
+    if (has_scale_column && scale_is_const) {
```
kangkaisen: you could do scale_is_const in the prepare function. then, you could do more optimizations for convert_timestamp_to_seconds_microseconds

alvin-phoenix-ai:
@liubotao v3.5 is released, so we don't cherry-pick new features to it.

LiShuMing (dismissed):

alvin-phoenix-ai (dismissed):
I just put a request change here for waiting the confirmation about this function timezone-related behavior. 
And also the timezone behavior should be described in the user document.
cc @wangsimo0

LiShuMing (dismissed):

be/src/exprs/time_functions.cpp
```diff
@@ -1886,6 +1888,280 @@ StatusOr<ColumnPtr> TimeFunctions::from_unix_to_datetime_with_format_32(Function
...
+    int64_t quotient = dividend / divisor;
+    int64_t remainder = dividend % divisor;
+    if (remainder < 0) {
+        quotient -= 1;
+        remainder += divisor;
+    }
+    return {quotient, remainder};
+}
+
+inline void normalize_microseconds(int64_t& seconds, int64_t& microseconds) {
```
murphyatwork, line 1971: I'm a little bit concerned about the performance of this function; it seems like it requires a significant number of CPU cycles to perform a conversion.

murphyatwork (dismissed):

be/src/exprs/time_functions.cpp
```diff
@@ -1886,6 +1888,280 @@ StatusOr<ColumnPtr> TimeFunctions::from_unix_to_datetime_with_format_32(Function
...
+        return 1000;
+    case 6:
+        return 1000000;
+    default:
+        return -1;
+    }
+}
+
+inline std::pair<int64_t, int64_t> safe_divmod(int64_t dividend, int64_t divisor) {
+    int64_t quotient = dividend / divisor;
```
murphyatwork, line 1962: the `/` and `%` requires more cpu cycles

murphyatwork (approved):

LiShuMing (approved):

murphyatwork:

PR #60638, opened 2025-07-06
[Enhancement] fix missing call to hms client

dirtysalt:
## What I'm doing:

Fixes #

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
+++ b/fe/fe-core/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
@@ -1215,83 +1215,84 @@ public class HiveMetaStoreClient implements IMetaStoreClient, AutoCloseable {
             throws MetaException, UnknownTableException, NoSuchObjectException, TException {
         return client.get_partition_with_auth(dbName, tableName, pvals, userName, groupNames);
     }
+
     @Override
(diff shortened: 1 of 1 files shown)


dirtysalt:
@stephen-shelby can you just replace this file with

https://github.com/apache/hive/blob/master/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java

fe/fe-core/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
```diff
@@ -818,55 +837,52 @@ public synchronized Object invoke(Object proxy, Method method, Object [] args)
             }
         }
     }
-
+    @Override
     public void setMetaConf(String key, String value) throws MetaException, TException {
-
-        throw new TException("method not implemented");
+        client.setMetaConf(key, value);
```
stephen-shelby, line 842: I suggest only adding the interface that we have already know.

PR #60726, opened 2025-07-09
[BugFix] fix stream load does not use correct CN group

starrocks-xupeng:
## What I'm doing:

stream load always uses default warehouse for now

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
@@ -309,7 +309,7 @@ public abstract class RoutineLoadJob extends AbstractTxnStateChangeCallback
     @SerializedName("warehouseId")
     protected long warehouseId = WarehouseManager.DEFAULT_WAREHOUSE_ID;
 
-    // no needs to persist
+    @SerializedName("wcr")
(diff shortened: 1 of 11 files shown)


LiShuMing (dismissed):

kevincai (dismissed):

fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
```diff
@@ -309,7 +309,7 @@ public enum JobSubstate {
     @SerializedName("warehouseId")
     protected long warehouseId = WarehouseManager.DEFAULT_WAREHOUSE_ID;
 
-    // no needs to persist
+    @SerializedName("wcr")
```
kevincai, line 312: create an json string without `wcr` in the json, check what the computeResource value will be after deserialize. Could be possibly handle the null/empty value in gsonPostProcess() if needed to avoid unexpected NPE exception.

murphyatwork (approved):

kevincai (approved):

gengjun-git (approved):

PR #60737, opened 2025-07-09
[Enhancement] disable colocate balancer and tablet scheduler in shared-data

starrocks-xupeng:
## Why I'm doing:

it will print some useless logs in shared-data

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/clone/TabletChecker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/clone/TabletChecker.java
@@ -62,7 +62,6 @@ import com.starrocks.common.util.FrontendDaemon;
 import com.starrocks.common.util.concurrent.lock.LockType;
 import com.starrocks.common.util.concurrent.lock.Locker;
 import com.starrocks.server.GlobalStateMgr;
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/clone/ColocateTableBalancer.java
```diff
@@ -251,6 +252,9 @@ public String toString() {
      */
     @Override
     protected void runAfterCatalogReady() {
+        if (RunMode.isSharedDataMode()) {
```
kevincai: shall disable it in GlobalStateMgr, so the Daemon thread will not be started at all.

kevincai (approved):

gengjun-git (approved):

PR #60743, opened 2025-07-09
[Enhancement] Filter out keys using SstablePredicate for sstable after compaction

srlch:
## What I'm doing:
This pr is prerequisite pr for tablet split.
In this pr, we use SstablePredicate to filter out some keys which do not meet the condition after the sstable compaction.

Fixes #59134

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/lake/sstable_predicate.h
+++ b/be/src/storage/lake/sstable_predicate.h
@@ -14,16 +14,13 @@
 
 #pragma once
 
-#include "storage/lake/key_to_chunk_converter.h"
+#include "storage/lake/sstable_predicate_fwd.h"
+#include "storage/lake/sstable_predicate_utils.h"
 #include "storage/record_predicate/record_predicate.h"
 
 namespace starrocks {
 
 namespace lake {
-
-class SstablePredicate;
-using SstablePredicateUPtr = std::unique_ptr<SstablePredicate>;
-using SstablePredicateSPtr = std::shared_ptr<SstablePredicate>;
 
 class SstablePredicate {
 public:
@@ -34,6 +31,8 @@ public:
                                                  const PersistentIndexSstablePredicatePB& predicate_pb);
 
     Status evaluate(const std::string& row, uint8_t* selection);
+
+    bool equals(const SstablePredicate& other) const { return _record_predicate->equals(*other._record_predicate); }
 
 private:
     Status _init(const PersistentIndexSstablePredicatePB& predicate_pb, KeyToChunkConverterUPtr& converter);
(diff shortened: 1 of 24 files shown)


be/src/storage/lake/sstable_predicate.h
```diff
@@ -14,17 +14,14 @@
...
 
 namespace starrocks {
 
 namespace lake {
 
-class SstablePredicate;
-using SstablePredicateUPtr = std::unique_ptr<SstablePredicate>;
-using SstablePredicateSPtr = std::shared_ptr<SstablePredicate>;
-
 class SstablePredicate {
```
luohaha, line 25: I think maybe we need to move `SstablePredicate` from `storage/lake` dir to `storage/sstable`, because `SstablePredicate` may be reused by shard-nothing storage one day.

srlch: Updated, has been removed into `sstable/`

luohaha (dismissed):

be/src/storage/sstable/iterator.h
```diff
@@ -63,6 +64,19 @@ class Iterator {
     // Return the max rss_rowid the iterator contains.
     virtual uint64_t max_rss_rowid() const { return 0; };
 
+    // Return predicate the iterator contains.
+    /*
+     * Currently, predicate is available for TwoLevelIterator and MergingIterator, because such
```
wyb:
```suggestion
    /*
     * Return predicate the iterator contains.
     * Currently, predicate is available for TwoLevelIterator and MergingIterator, because such
```

gensrc/proto/types.proto
```diff
@@ -227,6 +227,9 @@ message PersistentIndexSstablePB {
     optional bytes encryption_meta = 5;
     // Whether this file shared by multiple tablets
     optional bool shared = 6 [default=false];
+    // Filter out rows which do not satisfy the predicate
+    // when reading this specific sstable
+    optional PersistentIndexSstablePredicatePB predicate = 7;
```
wyb, line 232: when is the predicate cleared?

be/src/storage/sstable/iterator.h
srlch: fixed

gensrc/proto/types.proto
srlch: After the sstable has been compacted, this predicate will be removed. It has the same life cycle as sstable itself

wyb (approved):

wangruin (approved):

PR #60949, opened 2025-07-15
[BugFix] Fix txn log not exist when batch publish for shared-data arch

banmoy:
## Why I'm doing:
In shared-data architecture, concurrent execution of a batch publish and a schema change can lead to a publish failure, with the corresponding exception being txn log not found
```
Fail to publish log version: Not found: starlet err Object s3://starrocks-qa-test-cloud-data/daily/fe_InstanceId/f9a9c798-f208-48f1-a056-4ae85a26e385/db11101/11219/13574/log/000000000005B4D4_000000000000AF28.log does not exist
```
Scenario that Triggers the Issue:
* The initial table has only a normal index, one partition (partition 0), and one tablet (Tablet 1).
* An import transaction, TXN 1, is initiated. This import involves partition 0 and only contains Tablet 1.
* A schema change is initiated to add a bloom filter. This creates a shadow index and a corresponding Tablet 2 for Tablet 1. The process then enters the WAITING_TXN state, waiting for TXN 1 to complete.
* Another import transaction, TXN 2, is initiated. This import involves partition 0 and contains both Tablet 1 and Tablet 2.
* TXN 2 commits first, creating version 1. A .log file for TXN 2 exists on both Tablet 1 and Tablet 2.
* TXN 1 commits later, creating version 2. A .log file for TXN 1 exists only on Tablet 1.
* A batch publish is performed for TXN 1 and TXN 2. However, due to a logic error in the code, a task to publish TXN 1 is also sent to Tablet 2. Because TXN 1 does not involve Tablet 2, Tablet 2 cannot find the corresponding .log file, and the publish operation fails.

image

## What I'm doing:

The core issue is the lack of differentiation between ​normal indexes​ and ​shadow indexes ( https://github.com/StarRocks/starrocks/blob/branch-3.5.2/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java#L532)
* When all transactions include the same partition, batch publishing can be performed for these transactions.
* While partition inclusion guarantees all transactions cover tablets of the normal index, due to the special mechanism of schema changes, ​there's no guarantee​ that all transactions include tablets of the shadow index (as illustrated above).
* When processing shadow indexes, the code selects transactions to publish from the txnInfos list using the ​same logic​ as for normal indexes. This logic assumes that if the first transaction in a batch contains shadow index tablets, ​subsequent transactions must also contain them. If a later transaction misses a tablet, its publish task fails with a "transaction log not found" error when pushed to that tablet.
image

The solution is to ​individually check each transaction​ for the presence of a shadow index. ​It cannot be assumed​ that just because the previous committed transaction had a shadow index, ​all subsequent transactions must necessarily have one too.​​

Fixes https://github.com/StarRocks/StarRocksTest/issues/9986

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
@@ -505,7 +505,9 @@ public class PublishVersionDaemon extends FrontendDaemon {
         final List<Long> versions = publishVersionData.getCommitVersions();
         final List<TxnInfoPB> txnInfos = publishVersionData.getTxnInfos();
 
-        Map<Long, Set<Tablet>> shadowTabletsMap = new HashMap<>();
+        //  Record which transactions can be published in a batch for each shadow index.
+        //  The mapping is shadow index id -> ShadowIndexTxnBatch
+        Map<Long, ShadowIndexTxnBatch> shadowIndexTxnBatches = null;
         Set<Tablet> normalTablets = null;
 
         Locker locker = new Locker();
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -538,18 +540,22 @@ public boolean publishPartitionBatch(Database db, long tableId, PartitionPublish
...
-                            Set<Tablet> tabletsNew = new HashSet<>(index.getTablets());
-                            shadowTabletsMap.put(versions.get(i), tabletsNew);
+                        // sanity check. should not happen
+                        if (!index.visibleForTransaction(txnState.getTransactionId())) {
+                            LOG.error("Shadow index is included in the transaction but not visible, " +
+                                    "partitionId: {}, partitionName: {}, txnId: {} indexId: {}, indexName: {}",
+                                    partition.getId(), partition.getName(), txnState.getTransactionId(),
+                                    index.getId(), table.getIndexNameById(index.getId()));
+                            return false;
+                        }
```
xiangguangyxg: continue? otherwise if it encounters a bug, publish will not be able to recover

banmoy: fixed

xiangguangyxg (dismissed):

wyb (dismissed):

gengjun-git (dismissed):

gengjun-git (approved):

xiangguangyxg (approved):

wangruin:
https://github.com/Mergifyio backport branch-3.5.2

PR #60973, opened 2025-07-16
[BugFix] fix cache might not be used when upgraded from 3.3

starrocks-xupeng:
## What I'm doing:

if user set `starlet_cache_dir` in 3.3, the cache will not be used when upgraded to 3.4+

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/cache/datacache.cpp
+++ b/be/src/cache/datacache.cpp
@@ -165,6 +165,22 @@ StatusOr<CacheOptions> DataCache::_init_cache_options() {
     cache_options.engine = config::datacache_engine;
 
     if (config::datacache_engine == "starcache") {
+#ifdef USE_STAROS
+        std::vector<string> corresponding_starlet_dirs;
+        if (config::datacache_unified_instance_enable && !config::starlet_cache_dir.empty()) {
+            // in older versions, users might set `starlet_cache_dir` instead of `storage_root_path` for starlet cache,
+            // we need to move starlet cache into storage_root_path/datacache
+            auto s = DataCacheUtils::get_corresponding_starlet_cache_dir(_store_paths, config::starlet_cache_dir);
+            if (!s.ok()) {
+                LOG(WARNING) << s.status().message() << ", change config::datacache_unified_instance_enable to false";
+                config::datacache_unified_instance_enable = false;
+            } else {
+                corresponding_starlet_dirs = *s;
+            }
+        }
+#endif
+
+        int idx = 0;
         for (auto& root_path : _store_paths) {
             // Because we have unified the datacache between datalake and starlet, we also need to unify the
             // cache path and quota.
@@ -173,9 +189,14 @@ StatusOr<CacheOptions> DataCache::_init_cache_options() {
             // we do not automatically rename it when the source and destination directories are on different disks.
             // In this case, users should manually remount the directories and restart them.
             std::string datacache_path = root_path.path + "/datacache";
-            std::string starlet_cache_path = root_path.path + "/starlet_cache/star_cache";
 #ifdef USE_STAROS
             if (config::datacache_unified_instance_enable) {
+                std::string starlet_cache_path;
+                if (idx < corresponding_starlet_dirs.size()) {
+                    starlet_cache_path = corresponding_starlet_dirs[idx++];
+                } else {
+                    starlet_cache_path = root_path.path + "/starlet_cache/star_cache";
+                }
                 RETURN_IF_ERROR(DataCacheUtils::change_disk_path(starlet_cache_path, datacache_path));
             }
 #endif
--- a/be/src/cache/datacache_utils.cpp
+++ b/be/src/cache/datacache_utils.cpp
@@ -19,6 +19,8 @@
 
 #include <filesystem>
 
+#include "absl/status/statusor.h"
+#include "absl/strings/str_split.h"
 #include "fs/fs.h"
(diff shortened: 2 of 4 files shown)


be/src/cache/datacache_utils.cpp
```diff
@@ -181,4 +183,59 @@ dev_t DataCacheUtils::disk_device_id(const std::string& disk_path) {
...
+            if (id == id2) {
+                match = true;
+                corresponding_starlet_dirs.push_back(*iter);
+                starlet_paths.erase(iter);
+                break;
+            }
+        }
+        if (!match) {
+            std::string error_str = "can not find corresponding starlet cache dir for storage path " + root_path;
+            return Status::InternalError(error_str);
```
GavinMar: This means even if users configure a subset of `storage_root_path` as the `star_cache_dir`, we will skip to rename and force to use non-unified mode, is it expected?

starrocks-xupeng: yes, generally they are the same, if not, i think it's fine to use seperate starcache

GavinMar: If the `storage_root_path` is configured to `/data1/storage;/data2/storage` and `star_cache_dir` is `/data1/storage`, why can't we handle this one? I think the seperate mode is only need to be used to handle some unexpected cases in emergencies. We need to use the unified cache as our final goal and choose it as much as possible. Otherwise, we still need to maintain these two modes in future versions.

be/src/cache/datacache.cpp
```diff
@@ -173,9 +189,14 @@ StatusOr<CacheOptions> DataCache::_init_cache_options() {
...
             std::string datacache_path = root_path.path + "/datacache";
-            std::string starlet_cache_path = root_path.path + "/starlet_cache/star_cache";
 #ifdef USE_STAROS
             if (config::datacache_unified_instance_enable) {
+                std::string starlet_cache_path;
+                if (idx < corresponding_starlet_dirs.size()) {
+                    starlet_cache_path = corresponding_starlet_dirs[idx++] + "/star_cache";
+                } else {
+                    starlet_cache_path = root_path.path + "/starlet_cache/star_cache";
+                }
```
GavinMar, line 199: If the old starlet cache locates in `/data1/starlet_cache` and `/data3/starlet_cache`, this will cause the old cache data is moved to `/data1/${storage_root_path}/datacache` and `/data2/${storage_root_path}/datacache` ?

be/test/cache/datacache_utils_test.cpp
```diff
@@ -192,4 +192,83 @@ TEST_F(DataCacheUtilsTest, change_cache_path_from_nonexist_src) {
     fs::remove_all(new_dir);
 }
 
+TEST_F(DataCacheUtilsTest, get_corresponding_starlet_cache_dir) {
```
kevincai, line 196: need condition compile with USE_STAROS

be/src/cache/datacache.cpp
```diff
@@ -165,6 +165,22 @@ StatusOr<CacheOptions> DataCache::_init_cache_options() {
...
 
     if (config::datacache_engine == "starcache") {
+#ifdef USE_STAROS
+        std::vector<string> corresponding_starlet_dirs;
+        if (config::datacache_unified_instance_enable && !config::starlet_cache_dir.empty()) {
+            // in older versions, users might set `starlet_cache_dir` instead of `storage_root_path` for starlet cache,
+            // we need to move starlet cache into storage_root_path/datacache
+            auto s = DataCacheUtils::get_corresponding_starlet_cache_dir(_store_paths, config::starlet_cache_dir);
+            if (!s.ok()) {
+                LOG(ERROR) << s.status().message() << ", change config::datacache_unified_instance_enable to false";
```
kevincai: should be a WARN log because the process can still run with degraded configuration

starrocks-xupeng: done

be/src/cache/datacache.cpp
```diff
@@ -173,9 +189,20 @@ StatusOr<CacheOptions> DataCache::_init_cache_options() {
...
+                    if (!corresponding_starlet_dirs[idx].empty()) {
+                        starlet_cache_path = corresponding_starlet_dirs[idx] + "/star_cache";
+                    } else {
+                        // empty means we can't find an old starlet cache dir for this root_path, use the default one
+                        starlet_cache_path = root_path.path + "/starlet_cache/star_cache";
+                    }
+                    idx++;
+                } else {
+                    starlet_cache_path = root_path.path + "/starlet_cache/star_cache";
+                }
```
GavinMar, line 199: It seems assign the same concatenated string twice is redundant and error-prone. You can directly use `root_path.path + "/starlet_cache/star_cache"` to initialize the default value, and then adjust it to `corresponding_starlet_dirs[idx] + "/star_cache"`only when necessary.

starrocks-xupeng: done

GavinMar (approved):

murphyatwork (approved):

kevincai (approved):

PR #60986, opened 2025-07-16
[Feature] Support MATCH_ANY operator

hongkunxu:
## Why I'm doing:
Currently, users lack a clean and intuitive way to perform "match any keyword" queries against tokenized columns using GIN indexes. While similar behavior can be emulated using multiple MATCH clauses with OR, this approach is verbose, error-prone, and may result in inconsistent behavior depending on the parser configuration. Moreover, it complicates query optimization and planning. Introducing a native MATCH_ANY operator helps clarify intent and improves both correctness and performance.

## What I'm doing:
This PR introduces a new keyword predicate: MATCH_ANY, which allows users to match any token from a given string against a GIN-indexed column.

**Usage example:**
```
WHERE (NOT) <col_name> MATCH_ANY "keyword1 keyword2 keyword3"
```

**Behavior:**
1. The right-hand string is tokenized according to the column's configured parser (e.g., standard, english, chinese).
2. The expression evaluates to true if any of the resulting tokens match the column value.

**Constraints:**

1. MATCH_ANY can only be used on columns with GIN indexes.
2. It must be used as a pushdown predicate; it is not a general-purpose expression and cannot be evaluated outside of index scan filtering.
3. MATCH_ANY does not support wildcard characters such as '%' or '*'. The input string is treated as space-separated tokens, not as a pattern.

Fixes[ #61827](https://github.com/StarRocks/starrocks/issues/61827)

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


be/src/storage/index/inverted/clucene/clucene_inverted_util.h
```diff
@@ -38,4 +41,19 @@ StatusOr<std::unique_ptr<lucene::analysis::Analyzer>> get_analyzer(InvertedIndex
     }
 }
 
+inline Status tokenize_text(InvertedIndexParserType parser_type, std::string search_str,
```
srlch, line 44: It seems that this function is only used for tokenizing the keywords given by MATCH_ANY. It always should be tokenized by space. If user use CHINESE or STANDARD analyzer, such analyzer may not tokenize text by space. Is there any problem in such case?

hongkunxu: Hi @srlch , thanks for your comment. For the standard/Chinese analyzers, this method is still effective. As for spaces, they also work well. The underlying implementation logic is the same as the tokenize function that I proposed in previous merged PR— you can try testing it yourself. The two screenshots below illustrate this clearly. I’ve tested both the Chinese and standard analyzers, but only included the Chinese one in the screenshots below.
The first screenshot shows the result of tokenization. As you can see, the tokens “持架” and “构快” do not appear, which indicates that the space is effective.
The second screenshot shows that line 21 was not retrieved, even though it contains the content “构快”.
image
image

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/operator/scalar/MatchExprOperator.java
```diff
@@ -24,20 +25,33 @@
 import java.util.Objects;
 
 public class MatchExprOperator extends ArgsScalarOperator {
-    public MatchExprOperator(ScalarOperator... arguments) {
+    private List<ScalarOperator> arguments;
```
Seaven: duplicate field

hongkunxu: Thanks for your comment. I've removed duplicated field now. @Seaven

srlch (dismissed):

be/src/storage/index/inverted/clucene/clucene_inverted_reader.cpp
```diff
@@ -53,6 +54,8 @@ Status FullTextCLuceneInvertedReader::query(OlapReaderStatistics* stats, const s
     auto act_len = strnlen(search_query->data, search_query->size);
     std::string search_str(search_query->data, act_len);
     std::wstring search_wstr = boost::locale::conv::utf_to_utf<TCHAR>(search_str);
+    std::vector<std::wstring> tokens;
+    RETURN_IF_ERROR(starrocks::tokenize_text(_parser_type, search_str, tokens));
```
kangkaisen, line 58: seems only InvertedIndexQueryType::MATCH_ANY_QUERY branch need tokens, so would better move the starrocks::tokenize_text to InvertedIndexQueryType::MATCH_ANY_QUERY branch

be/src/storage/index/inverted/clucene/match_operator.cpp
```diff
@@ -34,6 +34,28 @@ Status MatchTermOperator::_match_internal(lucene::search::HitCollector* hit_coll
...
+    std::vector<lucene::index::Term*> terms;
+    std::vector<lucene::search::TermQuery*> queries;
+
+    for (const auto& token : _tokens) {
+        auto* term = _CLNEW lucene::index::Term(_field_name.c_str(), token.c_str());
+        terms.push_back(term);
+        queries.push_back(_CLNEW lucene::search::TermQuery(term));
+    }
+
+    lucene::search::BooleanQuery boolean_query;
```
kangkaisen: do we need some try catch here? to make sure _CLDELETE be called.

be/src/storage/index/inverted/clucene/clucene_inverted_reader.cpp
hongkunxu: Hi @kangkaisen  Next, I will be working on the match_all operator, so I need to extract tokenize_text separately.

be/src/storage/index/inverted/clucene/match_operator.cpp
hongkunxu: Sure. Thanks for your good advice. I 've added try catch to make sure _CLDELETE be called.  Please help to review again. @kangkaisen

Seaven (dismissed):

hongkunxu:
Hi @kangkaisen please help to review again.

srlch (dismissed):

kangkaisen (approved):

PR #61027, opened 2025-07-17
[BugFix] Executable segments generated by JIT are not released when it is evicted from JIT cache

satanson:
## Why I'm doing:
The JIT cache design is flawed, primarily because it confuses the two distinct stages:
1. Compile: IR → Object
2. Link: Object → StandardSegments (executable memory)

In reality, two separate caches are needed:
1. An Object cache — LLVM provides API support for this.
2. A StandardSegments (executable memory) cache — LLVM does not encourage this, but it can be worked around.

There is a fundamental misunderstanding of the purpose of the ObjectCache. Its role is to avoid recompiling IR to Object — not to cache executable segments. However, many developers mistakenly treat it as a cache for executable memory and create a separate ObjectCache for each JIT expression. The correct design should treat this as a global table.

As for JITLinkingMemoryManager, which manages executable segments:
This component should not be global, but has incorrectly been made global. In fact, its lifecycle should be managed by the second cache — the one responsible for executable memory. JITLinkingMemoryManager's derived class InProcessMemoryManager is used in link-time, it never releases executable segments when InProcessMemoryManager goes out its scope. 

## What I'm doing:
1. Use object cache and callable cache to save compiling result(object) and linking result(executable segments) respectively.
2. Introduce a subclass of InProcessMemoryManager, which named CustomizedInProcessMemoryManager, it can release executable segments when it goes out of scope.
3. refactor JITEngine.

**Test Result**

Use query format based on ssb
```
SELECT SUM(
    CASE 
        WHEN x < {param1} THEN x / 1000000 + 0 * x
        WHEN x < {param2} THEN x / 2000000 + 1 * x
        WHEN x < {param3} THEN x / 3000000 + 2 * x
        WHEN x < 400000   THEN x / 4000000 + 3 * x
        WHEN x < 500000   THEN x / 5000000 + 4 * x
        WHEN x < 600000   THEN x / 6000000 + 5 * x
        WHEN x < 700000   THEN x / 7000000 + 6 * x
        WHEN x < 800000   THEN x / 8000000 + 7 * x
        WHEN x < 900000   THEN x / 9000000 + 8 * x
        WHEN x < 1000000  THEN x / 10000000 + 9 * x
        ELSE 10
    END
) 
FROM (
    SELECT lo_partkey AS x FROM lineorder
    UNION ALL
    SELECT lo_partkey AS x FROM lineorder
    UNION ALL
    SELECT lo_partkey AS x FROM lineorder
) A;
```

{params1} ranges from 1000 ~ 1010
{params2} ranges from 2000 ~ 2010
{params3} ranges from 3000 ~ 3020

use different combinations of params1, params2 and params3 to generate 2000 queries, which queries would be JITed into different executable segments. 

Then set capacity of JIT cache to be 1048576 bytes(jit_lru_cache_size = 1048576),  so it can keep 256 pages, 8 pages for each shard of JIT cache.

in the previous version,  readonly executable private(r-xp)mapped pages are 4001,  it exceeds 256 pages so far.
in the current version, readonly executable private mapped pages are 95, it is below 256 pages.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1557,10 +1557,15 @@ CONF_mInt32(olap_string_max_length, "1048576");
 // Skip get from pk index when light pk compaction publish is enabled
 CONF_mBool(enable_light_pk_compaction_publish, "true");
 
+// jit LRU object cache size for total 32 shards, it will be an auto value if it < 0
+// mem_limit = system memory or process memory limit if set.
+// if mem_limit < 16 GB, disable JIT.
+// else it  = min(mem_limit*0.01, 4MB);
+CONF_mInt64(jit_lru_object_cache_size, "0");
 // jit LRU cache size for total 32 shards, it will be an auto value if it <=0:
 // mem_limit = system memory or process memory limit if set.
 // if mem_limit < 16 GB, disable JIT.
-// else it = min(mem_limit*0.01, 1GB)
+// else it = min(mem_limit*0.01, 4MB)
 CONF_mInt64(jit_lru_cache_size, "0");
 
 CONF_mInt64(arrow_io_coalesce_read_max_buffer_size, "8388608");
--- a/be/src/exprs/jit/jit_engine.cpp
+++ b/be/src/exprs/jit/jit_engine.cpp
@@ -69,136 +69,9 @@
 
 namespace starrocks {
 
-struct JitCacheEntry {
-    JitCacheEntry(std::shared_ptr<llvm::MemoryBuffer> buff, JITScalarFunction f)
-            : obj_buff(std::move(buff)), func(std::move(f)) {}
-    std::shared_ptr<llvm::MemoryBuffer> obj_buff;
-    JITScalarFunction func;
-};
-
-JitObjectCache::JitObjectCache(const std::string& expr_name, Cache* cache)
-        : _cache_key(std::move(expr_name)), _lru_cache(std::move(cache)) {}
-
-void JitObjectCache::notifyObjectCompiled(const llvm::Module* M, llvm::MemoryBufferRef Obj) {
-    std::unique_ptr<llvm::MemoryBuffer> obj_buffer =
-            llvm::MemoryBuffer::getMemBufferCopy(Obj.getBuffer(), Obj.getBufferIdentifier());
-    _obj_code = std::move(obj_buffer);
-}
-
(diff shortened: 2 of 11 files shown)


alvin-phoenix-ai:
@satanson This change is so big, and JIT is not necessary for most scenarios. To avoid introducing potential risks to the released versions, I believe this PR should not be backported to the released versions.

satanson:
> @satanson This change is so big, and JIT is not necessary for most scenarios. To avoid introducing potential risks to the released versions, I believe this PR should not be backported to the released versions.

@alvin-celerdata  The root cause of the issue is clear, I have testify this PR rigorously. additionally,  JIT on machines with 16GB(exactly speaking, size of query pool exceeds 16GB ) is enabled in default.

kevincai:
> > @satanson This change is so big, and JIT is not necessary for most scenarios. To avoid introducing potential risks to the released versions, I believe this PR should not be backported to the released versions.
> 
> @alvin-celerdata The root cause of the issue is clear, I have testify this PR rigorously. additionally, JIT on machines with 16GB(exactly speaking, size of query pool exceeds 16GB ) is enabled in default.

so we should disable it by default in v3.4/v3.3 as a bugfix?

alvin-phoenix-ai:
> > > @satanson This change is so big, and JIT is not necessary for most scenarios. To avoid introducing potential risks to the released versions, I believe this PR should not be backported to the released versions.
> > 
> > 
> > @alvin-celerdata The root cause of the issue is clear, I have testify this PR rigorously. additionally, JIT on machines with 16GB(exactly speaking, size of query pool exceeds 16GB ) is enabled in default.
> 
> so we should disable it by default in v3.4/v3.3 as a bugfix?

I like this suggestion

ZiheLiu (approved):

stdpain (approved):

kevincai (approved):

silverbullet233 (approved):

satanson:
@mergifyio backport branch-3.5

PR #61071, opened 2025-07-18
[Refactor] Refactor the historical node manager interfaces based on compute resource instead of the combination of warehouse id and worker group id.

GavinMar:
## Why I'm doing:
Now we use the `warehouseId` and `workerGroupId` in historical node manager interfaces to visit historical node information. However, this breaks down the ComputeResource concept, and we should be compute resource oriented design and implementation. It means that all the interfaces should be interacted as ComputeResource whenever there is a warehouseId + workerGroupId in the parameter list. 

## What I'm doing:
* Refactor the historical node manager interfaces based on compute resource instead of the combination of warehouse id and worker group id.
* Catch some exceptions related to historical node management to avoid affecting the main logic.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/proc/HistoricalNodeProcNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/proc/HistoricalNodeProcNode.java
@@ -15,13 +15,13 @@
 package com.starrocks.common.proc;
 
 import com.starrocks.common.AnalysisException;
-import com.starrocks.common.Pair;
 import com.starrocks.common.util.TimeUtils;
 import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.server.WarehouseManager;
 import com.starrocks.system.HistoricalNodeMgr;
 import com.starrocks.system.HistoricalNodeSet;
 import com.starrocks.warehouse.Warehouse;
+import com.starrocks.warehouse.cngroup.ComputeResource;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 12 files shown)


fe/fe-core/src/main/java/com/starrocks/common/proc/HistoricalNodeProcNode.java
```diff
@@ -45,16 +45,16 @@ public ProcResult fetchResult() throws AnalysisException {
...
         WarehouseManager warehouseManager = GlobalStateMgr.getCurrentState().getWarehouseMgr();
         HistoricalNodeMgr historicalNodeMgr = globalStateMgr.getHistoricalNodeMgr();
-        ConcurrentHashMap<Pair<Long, Long>, HistoricalNodeSet> resourceToNodeSet = historicalNodeMgr.getAllHistoricalNodeSet();
-        for (Pair<Long, Long> resourceId : resourceToNodeSet.keySet()) {
-            HistoricalNodeSet nodeSet = resourceToNodeSet.get(resourceId);
-            Warehouse warehouse = warehouseManager.getWarehouse(resourceId.first);
+        ConcurrentHashMap<ComputeResource, HistoricalNodeSet> resourceToNodeSet = historicalNodeMgr.getAllHistoricalNodeSet();
+        for (ComputeResource computeResource : resourceToNodeSet.keySet()) {
+            HistoricalNodeSet nodeSet = resourceToNodeSet.get(computeResource);
+            Warehouse warehouse = warehouseManager.getWarehouse(computeResource.getWarehouseId());
```
kevincai: `getWarehouse()` will throw exception if the given id not exist. turn to `getWarehouseAllowNull` to get a `null` result if not exist.

fe/fe-core/src/main/java/com/starrocks/persist/UpdateHistoricalNodeLog.java
```diff
@@ -27,11 +28,8 @@ public class UpdateHistoricalNodeLog implements Writable {
     @SerializedName(value = "warehouse")
     private String warehouse;
 
-    @SerializedName(value = "warehouseId")
-    private long warehouseId;
-
-    @SerializedName(value = "workerGroupId")
-    private long workerGroupId;
+    @SerializedName("computeResource")
+    private ComputeResource computeResource;
```
kevincai, line 28: Please double check if this can be a problem serializing and deserializing the object, because the ComputeResource is an interface, can't be instantialized by gson reflection, it must be delegated to a subclass for instantiation.

wangsimo0:
ignore merge 4.0.0-rc01

wangsimo0:

fe/fe-core/src/main/java/com/starrocks/common/proc/HistoricalNodeProcNode.java
GavinMar: > `getWarehouse()` will throw exception if the given id not exist. turn to `getWarehouseAllowNull` to get a `null` result if not exist.

Fixed

fe/fe-core/src/main/java/com/starrocks/persist/UpdateHistoricalNodeLog.java
GavinMar: > Please double check if this can be a problem serializing and deserializing the object, because the ComputeResource is an interface, can't be instantialized by gson reflection, it must be delegated to a subclass for instantiation.

Yes, the ComputeResource has been registered to GsonUtils, and can handle the serializing and deserializing process.

alvin-phoenix-ai:
@cursor review

PR #61083, opened 2025-07-18
[Enhancement] Improve error handling in TabletStatMgr and WarehouseManager

crossoverJie:
## Why I'm doing:
When the function `com.starrocks.server.WarehouseManager#getComputeNodeAssignedToTablet` throws an exception, the entire task will exit.

## What I'm doing:

- When the function `com.starrocks.server.WarehouseManager#getComputeNodeAssignedToTablet` throws an exception, skip the current Tablet and continue executing the task.
- Print exception.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3


fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
```diff
@@ -340,11 +341,13 @@ public List<Long> getAllComputeNodeIdsAssignToTablet(ComputeResource computeReso
             return GlobalStateMgr.getCurrentState().getStarOSAgent()
                     .getAllNodeIdsByShard(tabletId, computeResource.getWorkerGroupId());
         } catch (StarRocksException e) {
+            LOG.warn("get all compute node ids assign to tablet {} fail.", tabletId, e);
```
kevincai: these two lines's warn may be over chatty when the system has no node at all. Need to consider something similar to GLOG LOG_EVERY_N(), or LogLimiter to reduce the logs.

fe/fe-core/src/main/java/com/starrocks/catalog/TabletStatMgr.java
```diff
@@ -372,10 +373,17 @@ private void sendTasks() {
...
+                ComputeNode node;
+                try {
+                    node = warehouseManager.getComputeNodeAssignedToTablet(computeResource, tablet.getId());
+                    if (node == null) {
+                        LOG.warn("Skip sending tablet stat task for partition {} because no alive node", debugName());
+                        continue;
+                    }
+                } catch (ErrorReportException e) {
+                    LOG.warn("Skip sending tablet stat task for partition {} because exception: {}",
+                            debugName(), e);
```
kevincai: print e.getMessage(), instead of print the entire stack trace, although the stack trace can be turned off in log4j configuration, but it is on by default.

fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
crossoverJie: It seems there isn't a ready-made `LogLimiter` available right now?

fe/fe-core/src/main/java/com/starrocks/catalog/TabletStatMgr.java
crossoverJie: Done.

fe/fe-core/src/main/java/com/starrocks/server/WarehouseManager.java
kevincai: No, there isn't any.

crossoverJie: Perhaps we can use this [library](https://github.com/google/flogger), although I think this is a common scenario and can be implemented in a separate PR.

kevincai (approved):

HangyuanLiu (approved):

kevincai:
It appears that this will affect the tablet stats collection in shared-data mode, should be treated as a bugfix instead of enhancement.

backport to branch-3.4.

@mergifyio backport branch-3.4

kevincai:
@mergifyio backport branch-3.4

PR #61154, opened 2025-07-22
[Enhancement] Collect FE's machine info including cpu and mac address

gengjun-git:
## What I'm doing:
Refer to https://github.com/StarRocks/starrocks/blob/main/be/src/util/cpu_info.cpp

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/MachineInfo.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/MachineInfo.java
@@ -1,0 +1,192 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


fe/fe-core/src/main/java/com/starrocks/common/util/MachineInfo.java
```diff
@@ -0,0 +1,192 @@
...
+
+public class MachineInfo {
+    private static final Logger LOG = LogManager.getLogger(MachineInfo.class);
+
+    /**
+     * Get the number of CPU cores.
+     *
+     * @return the number of CPU cores.
+     */
+    public static int getCpuCores() {
```
kevincai: may need to implement in a single instance mode, so the cpucore and mem info and the mac address will be only intialized once, repeated invoke of the getXXX() won't cause the FE process to parse procfs again and again giving the info are pretty immutable after process started.

gengjun-git: done

kevincai (approved):

HangyuanLiu (approved):

PR #61182, opened 2025-07-23
[BugFix] Fix inconsistent status between aliveStatus and becomeDead for BE

ZiheLiu:
## Why I'm doing:

When replay `handleHbResponse`, the status `alive` will be overrided by `hbResponse.aliveStatus `, and the reason is as follows:
> Override the aliveStatus detected by the counter `heartbeatRetryTimes`, in two cases
> 1. the follower has a different `heartbeatRetryTimes` value compared to the Leader, the follower  must follow the leader's aliveStatus decision.
 > 2. editLog replay in leader FE's startup, where the value of `heartbeatRetryTimes` is changed.

However, `becomeDead` is not set for this scenario. And then the status of this BE becomes `alive`, but `becomeDead` will be notified to `CoordinatorMonitor` and kill all the queries related to this BE.

---

The related content in `fe.log` is as follows
```
2025-07-23 13:46:16.911+08:00 INFO (replayer|99) [ComputeNode.handleHbResponse():655] Backend [id=10265, host=127.0.0.1, heartbeatPort=9050, alive=false, status=OK] is dead due to exceed heartbeatRetryTimes
2025-07-23 13:46:16.911+08:00 INFO (replayer|99) [ComputeNode.handleHbResponse():689] Backend [id=10265, host=127.0.0.1, heartbeatPort=9050, alive=true, status=OK] alive status is changed to true
2025-07-23 13:46:16.911+08:00 INFO (replayer|99) [CoordinatorMonitor.addDeadBackend():57] add backend 10265 to dead backend queue
```

---

The related codes:

```java
  if (needSetAlive) {
      if (isAlive.compareAndSet(true, false)) {     // ========== Set `alive` to false.
          LOG.info("{} is dead due to {}", this, deadMessage);
      }
      heartbeatErrMsg = hbResponse.getMsg() == null ? "Unknown error" : hbResponse.getMsg();
      Status targetStatus = isShutdown ? Status.SHUTDOWN : Status.DISCONNECTED;
      if (status != targetStatus) {
          status = targetStatus;
          switch (targetStatus) {
              case DISCONNECTED:
                  becomeDead = true;          // ========== Set `becomeDead` to true.
                  break;
          }
      }
  }

  if (!isReplay) {
      hbResponse.aliveStatus = isAlive.get() ?
              HeartbeatResponse.AliveStatus.ALIVE : HeartbeatResponse.AliveStatus.NOT_ALIVE;
  } else {
      if (hbResponse.aliveStatus != null) {
          // Override the aliveStatus detected by the counter `heartbeatRetryTimes`, in two cases
          // 1. the follower has a different `heartbeatRetryTimes` value compared to the Leader, the follower
          //    must follow the leader's aliveStatus decision.
          // 2. editLog replay in leader FE's startup, where the value of `heartbeatRetryTimes` is changed.
          boolean newIsAlive = hbResponse.aliveStatus == HeartbeatResponse.AliveStatus.ALIVE;
          if (setAlive(newIsAlive)) {.  // ========== Set `alive` to true, but `becomeDead` is still true.
              LOG.info("{} alive status is changed to {}", this, newIsAlive);
          }
      }
  }
```

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/test/java/com/starrocks/system/ComputeNodeTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/system/ComputeNodeTest.java
@@ -29,7 +29,6 @@ import org.json.JSONObject;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
--- a/fe/fe-core/src/test/java/com/starrocks/system/ComputeNodeTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/system/ComputeNodeTest.java
@@ -23,12 +23,16 @@ import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.system.HeartbeatResponse.HbStatus;
 import com.starrocks.thrift.TStatusCode;
 import mockit.Expectations;
+import mockit.Mock;
+import mockit.MockUp;
 import org.json.JSONObject;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class ComputeNodeTest {
 
@@ -359,6 +363,7 @@ public class ComputeNodeTest {
         public VerifyComputeNodeStatus() {
             // nothing
         }
+
         public VerifyComputeNodeStatus(boolean isAlive, HeartbeatResponse.AliveStatus aliveStatus, ComputeNode.Status status) {
             this.isAlive = isAlive;
             this.aliveStatus = aliveStatus;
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/test/java/com/starrocks/system/ComputeNodeTest.java
```diff
@@ -610,4 +614,68 @@ public void testComputeNodeStatusUpgradeCompatibility() {
...
+        };
+
+        long nodeId = 1000;
+
+        {
+            // status is ERROR, but aliveStatus is ALIVE.
+
+            ComputeNode node = new ComputeNode(nodeId, "127.0.0.1", 9050);
+            BackendHbResponse hbResponse = generateHbResponse(node, TStatusCode.THRIFT_RPC_ERROR, 0);
+            hbResponse.aliveStatus = HeartbeatResponse.AliveStatus.ALIVE;
```
kevincai, line 636: so it is weird that the leader heartbeat probing fails with RPC error but the alive status keeps as ALIVE without change, this should not be expected?

What I can think of is that the heartbeat retry failure times configuration is different between FE leader and FE follower, so the Leader think it is not the time to set dead but the follower thinks it is time to take the node as dead.

kevincai:
succeeded by #61249

PR #61233, opened 2025-07-24
[BugFix] Fix antlr conflict in gradle

decster:
## Why I'm doing:
when running ut with gradle, got this error:

```
ShowCreateRoutineLoadStmtTest > testBackquote() STANDARD_ERROR
    ANTLR Tool version 4.9.3 used for code generation does not match the current runtime version 4.11.1
```

trino-parser:385 must use antlr4-runtime 4.9.3, otherwise runtime error.
but some other packages uses antlr4-runtime 4.11.1, so upgrade to trino-parser:396(latest version which not break current compile), which is compactable with antlr4-runtime:4.11.1

Note: should upgrade trino-parser to newer version in the futures(current version is too old), but this requires code change, mainly in `fe/fe-core/src/main/java/com/starrocks/connector/parser/trino/AstBuilder.java`

Update:

unfortunately, spark 3.5.5 uses antlr4-runtime:4.9.3, and there is no new version to upgrade, so we cannot upgrade our antlr4-runtime, stuck at 4.9.3

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/build.gradle.kts
+++ b/fe/build.gradle.kts
@@ -39,6 +39,7 @@ subprojects {
     ext {
         set("starrocks.home", "${rootDir}/../")
         // var sync start
+        set("antlr.version", "4.9.3")
         set("arrow.version", "18.0.0")
         set("async-profiler.version", "4.0")
         set("avro.version", "1.12.0")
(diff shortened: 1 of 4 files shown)


fe/build.gradle.kts
```diff
@@ -154,7 +154,7 @@ subprojects {
             implementation("javax.validation:validation-api:1.1.0.Final")
             implementation("javax.xml.ws:jaxws-api:2.3.0")
             implementation("net.sourceforge.czt.dev:java-cup:0.11-a-czt02-cdh")
-            implementation("org.antlr:antlr4:4.9.2")
+            implementation("org.antlr:antlr4:4.11.1")
```
alvin-phoenix-ai: Please make it a variable to avoid handwriting many times.

decster: done

alvin-phoenix-ai (approved):

PR #61251, opened 2025-07-24
[Refactor] Change the data type of data_size column in the partitions_meta table to bigint.

sevev:
## Why I'm doing:

The column type of the `data_size` field in the `partitions_meta` table  is string and the `storage_size` is `bigint` which is not consistent. 
We should keep the unit consistency of these two fields and `bigint` is more programmer-friendly.

```
mysql> select data_size, storage_size, MAX_CS, storage_path from information_schema.partitions_meta where db_name = 'ssb_100g_3'; 
+-----------+--------------+--------+-----------------------------------------------------------------------------------------------------------+
| data_size | storage_size | MAX_CS | storage_path                                                                                              | 
+-----------+--------------+--------+-----------------------------------------------------------------------------------------------------------+
| 237.9MB   |    483805555 |     40 | s3://xxxxxx/db518214/567560/567562 | 
+-----------+--------------+--------+-----------------------------------------------------------------------------------------------------------+
```

## What I'm doing:
Change the data type of `data_size` column in the `partitions_meta` table to bigint.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exec/pipeline/scan/schema_chunk_source.cpp
+++ b/be/src/exec/pipeline/scan/schema_chunk_source.cpp
@@ -84,6 +84,13 @@ Status SchemaChunkSource::prepare(RuntimeState* state) {
                 slot_desc->type().type = TYPE_DOUBLE;
             }
         }
+    } else if (schema_table->schema_table_type() == TSchemaTableType::SCH_PARTITIONS_META) {
+        for (auto* slot_desc : dest_slot_descs) {
+            const auto& col_name = slot_desc->col_name();
+            if (slot_desc->type().type == TYPE_VARCHAR && boost::iequals(col_name, "DATA_SIZE")) {
+                slot_desc->type().type = TYPE_BIGINT;
+            }
+        }
     }
 
     int slot_num = dest_slot_descs.size();
--- a/be/src/exec/schema_scanner/schema_partitions_meta_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_partitions_meta_scanner.cpp
@@ -19,8 +19,8 @@
 #include "common/logging.h"
 #include "exec/schema_scanner/schema_helper.h"
 #include "runtime/runtime_state.h"
-#include "types/logical_type.h"
 #include "storage/utils.h"
+#include "types/logical_type.h"
 
 namespace starrocks {
 
--- a/be/src/exec/schema_scanner/schema_partitions_meta_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_partitions_meta_scanner.cpp
@@ -20,6 +20,7 @@
 #include "exec/schema_scanner/schema_helper.h"
 #include "runtime/runtime_state.h"
 #include "types/logical_type.h"
+#include "storage/utils.h"
 
 namespace starrocks {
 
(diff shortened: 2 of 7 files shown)


gensrc/thrift/FrontendService.thrift
```diff
@@ -1517,7 +1517,7 @@ struct TPartitionMetaInfo {
     26: optional i64 data_version
     27: optional i64 version_epoch
     28: optional Types.TTxnType version_txn_type = Types.TTxnType.TXN_NORMAL
-    29: optional i64 storage_size
+    29: optional string storage_size
```
alvin-phoenix-ai: Type of existing fields CAN NOT be changed because it will cause a compatibility issue when upgrading.
So a change is required.

alvin-phoenix-ai (dismissed):
Changing it from int64 to string is not necessary to me.
If this is bigint, we can apply functions to it to get what users want, but if it is varchar the output is unpredicatable so that it is very difficult to apply functions.

alvin-phoenix-ai:
@curosr review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai (dismissed):

be/src/exec/schema_scanner/schema_partitions_meta_scanner.cpp
```diff
@@ -261,8 +262,8 @@ Status SchemaPartitionsMetaScanner::fill_chunk(ChunkPtr* chunk) {
         }
         case 22: {
             // DATA_SIZE
-            Slice data_size = Slice(info.data_size);
-            fill_column_with_slot<TYPE_VARCHAR>(column.get(), (void*)&data_size);
+            int64_t data_size_bytes = parse_data_size(info.data_size);
```
wyb, line 265: better to remove `ByteSizeValue` conversion in FE, so than we can remove such complex conversion in later version.

sevev: > better to remove `ByteSizeValue` conversion in FE, so than we can remove such complex conversion in later version.

update.

wyb (dismissed):

fe/fe-core/src/main/java/com/starrocks/service/InformationSchemaDataSource.java
```diff
@@ -455,8 +454,7 @@ private static void genPartitionMetaInfo(long dbId, OlapTable table,
         // REPLICATION_NUM
         partitionMetaInfo.setReplication_num(partitionInfo.getReplicationNum(partition.getId()));
         // DATA_SIZE
-        ByteSizeValue byteSizeValue = new ByteSizeValue(physicalPartition.storageDataSize());
-        partitionMetaInfo.setData_size(byteSizeValue.toString());
+        partitionMetaInfo.setData_size(String.valueOf(physicalPartition.storageDataSize()));
```
kevincai, line 460: add comment to explain why here is the String value and then parse the string value back to bigint on BE side.

be/src/storage/utils.cpp
```diff
@@ -406,4 +406,57 @@ int caculate_delta_writer_thread_num(int thread_num_from_config) {
     return std::max<int>(CpuInfo::num_cores() / 2, 16);
 }
 
+int64_t parse_data_size(const std::string& value_str) {
+    if (value_str.empty()) return 0;
+
+    // Trim leading/trailing spaces
+    size_t start = value_str.find_first_not_of(" \t\n\r");
+    size_t end = value_str.find_last_not_of(" \t\n\r");
+    if (start == std::string::npos) return 0;
```
kevincai:
```suggestion
    if (start == std::string::npos) return 0;
    size_t end = value_str.find_last_not_of(" \t\n\r");
```

be/src/exec/pipeline/scan/schema_chunk_source.cpp
```diff
@@ -84,6 +84,13 @@ Status SchemaChunkSource::prepare(RuntimeState* state) {
                 slot_desc->type().type = TYPE_DOUBLE;
             }
         }
+    } else if (schema_table->schema_table_type() == TSchemaTableType::SCH_PARTITIONS_META) {
+        for (auto* slot_desc : dest_slot_descs) {
+            const auto& col_name = slot_desc->col_name();
+            if (slot_desc->type().type == TYPE_VARCHAR && boost::iequals(col_name, "DATA_SIZE")) {
```
kevincai, line 90: what's the purpose? compatibility during upgrade?

sevev: > what's the purpose? compatibility during upgrade?

yes, for the compatibility.

be/src/storage/utils.cpp
sevev: > if (start == std::string::npos) return 0;

done.

fe/fe-core/src/main/java/com/starrocks/service/InformationSchemaDataSource.java
sevev: > add comment to explain why here is the String value and then parse the string value back to bigint on BE side.

done.

kevincai (approved):

alvin-phoenix-ai (approved):

PR #61255, opened 2025-07-25
[BugFix] Fix set null value for auto_increment column will reject the valid data if they are in the same chunk

srlch:
## Why I'm doing:
If a null value set by user for auto_increment, during the validation in tablet sink, it will reject the entire chunk if the some row is not null.

## What I'm doing:
Just reject the rows with null.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/be/src/exec/tablet_sink.cpp
+++ b/be/src/exec/tablet_sink.cpp
@@ -1008,7 +1008,7 @@ void OlapTableSink::_validate_data(RuntimeState* state, Chunk* chunk) {
                         }
 #endif
                         // If enable_log_rejected_record is true, we need to log the rejected record.
-                        if (nullable->is_null(j) && state->enable_log_rejected_record()) {
+                        if (state->enable_log_rejected_record()) {
                             state->append_rejected_record_to_file(chunk->rebuild_csv_row(j, ","), ss.str(), "");
                         }
                     }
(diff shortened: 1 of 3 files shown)


be/src/exec/tablet_sink.cpp
```diff
@@ -991,26 +991,28 @@ void OlapTableSink::_validate_data(RuntimeState* state, Chunk* chunk) {
...
 #else
-                if (!state->has_reached_max_error_msg_num()) {
-                    state->append_error_msg_to_file("", ss.str());
-                }
+                        if (!state->has_reached_max_error_msg_num()) {
+                            state->append_error_msg_to_file(chunk->debug_row(j), ss.str());
+                        }
 #endif
+                        // If enable_log_rejected_record is true, we need to log the rejected record.
+                        if (nullable->is_null(j) && state->enable_log_rejected_record()) {
```
wyb:
```suggestion
                        if (state->enable_log_rejected_record()) {
```

srlch: Fixed

wyb (approved):

banmoy (approved):

PR #61271, opened 2025-07-25
[BugFix] fix the pruned column size to be same with the unpruned one

dbjoa:
## Why I'm doing:
The SQL query having `MATCH` op over a to-be pruned column and `non-MATCH` op over an unpruned one like the sample query below meets the chunk size error (i.e., `ERROR 1064`) if the number of records becomes larger than a specific number (e.g., about 100k for the sample case if `chunk_size` is `4096`). 

In detail, if the to-be pruned column is not needed no longer after applying index filter, for improving the query performance, the existing implementation does not read the column data into the chunk but only sets its size to a “predefined value” given by the caller.

However, if the predefined size is different from the “actual size” of the unpruned column read to the chunk, the chunk size error can be occurred in a subsequent operation. This is because it expects the sizes of the columns in the chunk are same one another.  

Here is a sample query and a table schema to reproduce the issue:
```
select count(id) from t where message match "starrocks" and country = 'Korea';
```
```
CREATE TABLE t (
    id bigint(20) NOT NULL COMMENT "",
    message varchar(1024) NOT NULL COMMENT "",
    country varchar(255) NOT NULL COMMENT "",
    INDEX idx_message (message) USING GIN("imp_lib" = "clucene", "parser" = "english") COMMENT 'english index',
    INDEX idx_country (country) USING BITMAP COMMENT ''
) ENGINE=OLAP
DUPLICATE KEY(id)
DISTRIBUTED BY HASH(id)
PROPERTIES (
    "compression" = "LZ4",
    "fast_schema_evolution" = "true",
    "replicated_storage" = "false",
    "replication_num" = "1"
);
```

```
ERROR 1064 (HY000): Intermediate chunk size must not be greater than 4096, actually 4235 after 0-th operator olap_scan_0_0x7fda032fa710(O) { full:false iostasks:3 has_active:true num_chunks:1 morsel:fixed_morsel_queue empty:true has_output:true} in query_id=8381a23a-6779-11f0-9179-12e0d4e7a05d fragment_id=8381a23a-6779-11f0-9179-12e0d4e7a05f driver=driver_0_4 addr=0x7fd98fb62410, status=RUNNING, operator-chain: [olap_scan_0_0x7fda032fa710(O) { full:false iostasks:3 has_active:true num_chunks:1 morsel:fixed_morsel_queue empty:
```

## What I'm doing:
The PR sets the size of a pruned column to be same with the `actual size` of a unpruned one in the same chunk.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/rowset/segment_iterator.cpp
+++ b/be/src/storage/rowset/segment_iterator.cpp
@@ -146,15 +146,25 @@ private:
 
         Status read_columns(Chunk* chunk, const SparseRange<>& range) {
             bool may_has_del_row = chunk->delete_state() != DEL_NOT_SATISFIED;
+            std::vector<size_t> pruned_cols;
+            size_t pruned_col_size = 0;
             for (size_t i = 0; i < _column_iterators.size(); i++) {
                 ColumnPtr& col = chunk->get_column_by_index(i);
                 if (_prune_column_after_index_filter && _prune_cols.count(i)) {
-                    // make sure each pruned column has the same size as the unpruneable one.
-                    col->resize(range.span_size());
+                    pruned_cols.push_back(i);
                     continue;
                 }
                 RETURN_IF_ERROR(_column_iterators[i]->next_batch(range, col.get()));
+                if (pruned_col_size == 0) {
+                    pruned_col_size = col->size();
+                }
+                DCHECK_EQ(pruned_col_size, col->size());
                 may_has_del_row |= (col->delete_state() != DEL_NOT_SATISFIED);
+            }
+            for (size_t i : pruned_cols) {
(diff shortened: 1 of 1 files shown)


srlch (approved):

stdpain (approved):

wyb (approved):

wyb:
https://github.com/Mergifyio backport branch-3.4

PR #61283, opened 2025-07-25
[Enhancement] Disable the inline mode when writing data to datacache because it may cause a performance degradation. (backport #60530)

GavinMar:
## Why I'm doing:
The starcache engine supports inline a cache item that smaller than the threshold with its metadata into memory. This can help reduce the waste of disk space.
However, as the inline cache item only uses memory, sometimes although the disk space is sufficient, we cannot ensure all target data can be cached if reached the inline memory quota. For datalake queries, this will cause a performance degradation because we read the remote data with a coalesce io buffer and may read more data from remote storage and cause a high io latency.

## What I'm doing:
Disable the inline mode when writing data to datacache because it may cause a performance degradation.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/block_cache/starcache_wrapper.cpp
+++ b/be/src/block_cache/starcache_wrapper.cpp
@@ -76,6 +76,7 @@ Status StarCacheWrapper::write_buffer(const std::string& key, const IOBuffer& bu
         opts.mode = starcache::WriteOptions::WriteMode::WRITE_THROUGH;
     }
     opts.evict_probability = options->evict_probability;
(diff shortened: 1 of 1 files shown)


jaogoy:
ignore merge 3.4.6

PR #61326, opened 2025-07-28
[Feature] support creating iceberg table with sort order.

GavinMar:
## Why I'm doing:
Sorted Iceberg tables can provide a huge increase in query performance and can also lead to a decrease in cloud object storage request costs. 
Iceberg spec introduces sort_order of tables for users to sort their data during the writing process.  In order to further improve the query performance of the Iceberg table, it is also necessary for StarRocks to improve its support for sort_order. 

## What I'm doing:
We support specifying the sorting sequence when creating a table.

As the `ORDER BY` clause has been introduced for StarRocks native table to specify the sorting columns. So we consider reusing the same syntax. And we further expand the existing order by syntax to also support specifying the sort direction and null-order for each sort field separately. For example:

`ORDER BY (action DESC NULLS FIRST, id ASC NULLS LAST)`

  The sort_direction and null_order are optional, so we can also create an iceberg table with the same `ORDER BY` syntax as the native table:

`ORDER BY (action, id)`

In addition, we call the iceberg api to set the `sort-order` property that parsed from `ORDER BY` clause.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/analysis/OrderByElement.java
+++ b/fe/fe-core/src/main/java/com/starrocks/analysis/OrderByElement.java
@@ -143,6 +143,19 @@ public class OrderByElement implements ParseNode {
         return pos;
     }
 
+    /**
+     * Try to extract the column name from the `expr`.
+     * If the `expr` represents a valid column, the column name will be returned.
+     * Otherwise, it will return null.
+     */
+    public String castAsSlotRef() {
+        if (!(expr instanceof SlotRef)) {
+            return null;
+        }
+        SlotRef slotRef = (SlotRef) expr;
+        return slotRef.getColumnName();
+    }
+
     public String explain() {
         StringBuilder strBuilder = new StringBuilder();
         strBuilder.append(expr.explain());
--- a/fe/fe-core/src/main/java/com/starrocks/server/OlapTableFactory.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/OlapTableFactory.java
@@ -18,6 +18,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.starrocks.analysis.BloomFilterIndexUtil;
+import com.starrocks.analysis.OrderByElement;
 import com.starrocks.binlog.BinlogConfig;
 import com.starrocks.catalog.ColocateTableIndex;
 import com.starrocks.catalog.Column;
@@ -181,10 +182,14 @@ public class OlapTableFactory implements AbstractTableFactory {
 
         short shortKeyColumnCount = 0;
         List<Integer> sortKeyIdxes = new ArrayList<>();
-        if (stmt.getSortKeys() != null) {
+        if (stmt.getOrderByElements() != null) {
             Set<Integer> addedSortKey = new HashSet<>();
             List<String> baseSchemaNames = baseSchema.stream().map(Column::getName).collect(Collectors.toList());
-            for (String column : stmt.getSortKeys()) {
+            for (OrderByElement orderByElement : stmt.getOrderByElements()) {
+                String column = orderByElement.castAsSlotRef();
+                if (column == null) {
+                    throw new DdlException("Unknown column '" + orderByElement.getExpr().toSql() + "' in order by clause");
+                }
                 int idx = IntStream.range(0, baseSchemaNames.size())
                         .filter(i -> baseSchemaNames.get(i).equalsIgnoreCase(column))
                         .findFirst().orElse(-1);
@@ -633,7 +638,7 @@ public class OlapTableFactory implements AbstractTableFactory {
             Util.checkColumnSupported(baseSchema);
             int schemaHash = Util.schemaHash(schemaVersion, baseSchema, bfColumns, bfFpp);
 
-            if (stmt.getSortKeys() != null) {
+            if (stmt.getOrderByElements() != null) {
                 table.setIndexMeta(baseIndexId, tableName, baseSchema, schemaVersion, schemaHash,
                         shortKeyColumnCount, baseIndexStorageType, keysType, null, sortKeyIdxes,
                         sortKeyUniqueIds);
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CreateTableAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CreateTableAnalyzer.java
@@ -22,6 +22,7 @@ import com.google.common.collect.Sets;
 import com.starrocks.analysis.Expr;
 import com.starrocks.analysis.FunctionCallExpr;
 import com.starrocks.analysis.LiteralExpr;
+import com.starrocks.analysis.OrderByElement;
 import com.starrocks.analysis.ParseNode;
 import com.starrocks.analysis.SlotRef;
 import com.starrocks.analysis.TableName;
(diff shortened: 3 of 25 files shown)


fe/fe-core/src/main/java/com/starrocks/server/OlapTableFactory.java
```diff
@@ -180,10 +182,15 @@ public Table createTable(LocalMetastore metastore, Database db, CreateTableStmt
...
+        if (stmt.getOrderByElements() != null) {
             Set<Integer> addedSortKey = new HashSet<>();
             List<String> baseSchemaNames = baseSchema.stream().map(Column::getName).collect(Collectors.toList());
-            for (String column : stmt.getSortKeys()) {
+            for (OrderByElement orderByElement : stmt.getOrderByElements()) {
+                if (!(orderByElement.getExpr() instanceof SlotRef)) {
+                    throw new DdlException("Should order by column");
+                }
+                SlotRef slotRef = (SlotRef) orderByElement.getExpr();
+                String column = slotRef.getColumnName();
```
alvin-phoenix-ai: 1. Same logic has been replicated many times; we should use a function.
2. And it should be one place to validate the input.

fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
```diff
@@ -2084,8 +2080,15 @@ public ParseNode visitCreateMaterializedViewStatement(
...
-                        .stream().map(Identifier::getValue).collect(toList());
+                sortKeys = new ArrayList<>();
+                List<OrderByElement> orderByElements = visit(desc.orderByDesc().sortItem(), OrderByElement.class);
+                for (OrderByElement orderByElement : orderByElements) {
+                    if (!(orderByElement.getExpr() instanceof SlotRef)) {
+                        throw new ParsingException("Should order by column");
+                    }
+                    SlotRef slotRef = (SlotRef) orderByElement.getExpr();
+                    sortKeys.add(slotRef.getColumnName());
+                }
```
alvin-phoenix-ai: This is not a good place to valid the syntax. Suppose we support creating a materialized view with an Iceberg Table. How would you make the change?

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/server/OlapTableFactory.java
GavinMar: > 1. Same logic has been replicated many times; we should use a function.
> 2. And it should be one place to validate the input.

Good suggestion! Fixed.

fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
GavinMar: > This is not a good place to valid the syntax. Suppose we support creating a materialized view with an Iceberg Table. How would you make the change?

Fixed.

fe/fe-core/src/main/java/com/starrocks/analysis/OrderByElement.java
```diff
@@ -162,6 +162,14 @@ public NodePosition getPos() {
         return pos;
     }
 
+    public String getColumnName() {
```
alvin-phoenix-ai: A more appropriate name should be given, like castAsColumnRef.
And you can add some comments for the context of this function.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CreateTableAnalyzer.java
```diff
@@ -418,12 +419,16 @@ private static void analyzeSortKeys(CreateTableStmt stmt) {
...
         List<String> columnNames = columnDefs.stream().map(ColumnDef::getName).collect(Collectors.toList());
-        if (sortKeys != null) {
+        if (orderByElements != null) {
             // we should check sort key column type if table is primary key table
             if (keysType == KeysType.PRIMARY_KEYS) {
-                for (String column : sortKeys) {
+                for (OrderByElement orderByElement : orderByElements) {
+                    String column = orderByElement.getColumnName();
+                    if (column == null) {
+                        throw new SemanticException("Should order by column");
```
alvin-phoenix-ai: More clear information is needed to indicate to users what's wrong. it should includes
1. Which element
2. why is not support

fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
```diff
@@ -2084,8 +2080,8 @@ public ParseNode visitCreateMaterializedViewStatement(
 
             // Order By
             if (desc.orderByDesc() != null) {
-                sortKeys = visit(desc.orderByDesc().identifierList().identifier(), Identifier.class)
-                        .stream().map(Identifier::getValue).collect(toList());
+                List<OrderByElement> orderByElements = visit(desc.orderByDesc().sortItem(), OrderByElement.class);
+                sortKeys = orderByElements.stream().map(e -> e.getExpr().toSql().replace("`", "")).collect(toList());
```
alvin-phoenix-ai: I think you should change CreateMaterializedViewStatement constructor too with List<OrderByElement>

fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
```diff
@@ -4777,8 +4773,8 @@ public ParseNode visitOptimizeClause(StarRocksParser.OptimizeClauseContext conte
                 context.partitionDesc() == null ? null : getPartitionDesc(context.partitionDesc(), null),
                 context.distributionDesc() == null ? null : (DistributionDesc) visit(context.distributionDesc()),
                 context.orderByDesc() == null ? null :
-                        visit(context.orderByDesc().identifierList().identifier(), Identifier.class)
```
alvin-phoenix-ai, line 4790: The same, OptimizeClause should be changed too.

fe/fe-core/src/main/java/com/starrocks/analysis/OrderByElement.java
```diff
@@ -162,6 +162,19 @@ public NodePosition getPos() {
         return pos;
     }
 
+    /**
+     * Try to extract the column name from the `expr`.
+     * If the `expr` represents a valid column, the column name will be returned.
+     * Otherwise, it will return null.
+     */
+    public String getColumnNameAsSlotRef() {
```
alvin-phoenix-ai:
```suggestion
    public String castAsSlotRef() {
```

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/sql/ast/CreateMaterializedViewStatement.java
```diff
@@ -62,7 +63,12 @@ public class CreateMaterializedViewStatement extends DdlStmt {
     private DistributionDesc distributionDesc;
     private final int queryStartIndex;
     private final int queryStopIndex;
-    private final List<String> sortKeys;
+    // It saves the original sort order elements parsing from the order by clause.
+    // Because other sort properties, such as sort-direction and null-orders, are not supported in materialized view now.
+    // We extract its sort columns to `sortKeys` in analyze phase and use the `sortKeys` instead of it in most places.
+    private final List<OrderByElement> orderByElements;
+    // It will be set based on `orderByElements` in analyze
+    private List<String> sortKeys;
```
Seaven, line 71: unused?

GavinMar: > unused?

Yes, because now the other sorting properties, such as sort-direction and nulls-order, are only supported in creating iceberg table. The MV statement only supports the sort-keys as before. So, we extract the sort-keys from the order by elements to the original `sortKeys` in analyze phase, to ensures that this part of the logic is consistent with before.

Youngwb (approved):

Seaven (approved):

wangsimo0 (approved):

PR #61332, opened 2025-07-28
[Enhancement] assign a large but configurable row count to unknown stats table

dirtysalt:
## Why I'm doing:

For some tables which we can not collect table stats/column stats, we will assign a default row count.

But default row count is 1 is not good:
- suppoer A join B, A's count is 100, But B's count is unknown.  and If we assign row count 1 to B, then we will put B at the right side. Bad thing happens if B is large table.
- small table will more likley to trigger broadcast join which is more likely to have OOM.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/statistics/StatisticsUtils.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/statistics/StatisticsUtils.java
@@ -18,6 +18,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.starrocks.catalog.Database;
 import com.starrocks.catalog.Table;
+import com.starrocks.common.Config;
 import com.starrocks.qe.ConnectContext;
 import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.sql.analyzer.SemanticException;
@@ -79,7 +80,7 @@ public class StatisticsUtils {
 
     public static Statistics buildDefaultStatistics(Set<ColumnRefOperator> columns) {
         Statistics.Builder statisticsBuilder = Statistics.builder();
-        statisticsBuilder.setOutputRowCount(1);
+        statisticsBuilder.setOutputRowCount(Config.default_statistics_output_row_count);
         statisticsBuilder.addColumnStatistics(
                 columns.stream().collect(Collectors.toMap(column -> column, column -> ColumnStatistic.unknown())));
         return statisticsBuilder.build();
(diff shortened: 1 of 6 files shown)


alvin-phoenix-ai:
@cursor review

Seaven (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/Statistics.java
```diff
@@ -201,7 +203,7 @@ public static Builder buildFrom(Statistics other) {
     }
 
     public static Builder builder() {
-        return new Builder();
+        return new Builder().setOutputRowCount(Config.default_statistics_output_row_count);
     }
```
Seaven: too many place use the method, don't set it in here

Seaven (changes requested):
don't change all default value

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/Statistics.java
Seaven: I think you just want to modify scan's default output rows?

dirtysalt: yes. change default output rows instead of 1.

wangsimo0:
ignore merge 4.0.0-rc01

wangsimo0:

Seaven (dismissed):

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/connector/statistics/StatisticsUtils.java
```diff
@@ -79,7 +80,7 @@ public static List<String> getTableNameByUUID(String tableUUID) {
 
     public static Statistics buildDefaultStatistics(Set<ColumnRefOperator> columns) {
         Statistics.Builder statisticsBuilder = Statistics.builder();
-        statisticsBuilder.setOutputRowCount(1);
+        statisticsBuilder.setOutputRowCount(Config.default_statistics_output_row_count);
```
dirtysalt, line 83: this is only called by lake table like hudi/iceberg/delta/hive etc.

stephen-shelby (approved):

Seaven (approved):

dirtysalt:
@mergify backport branch-4.0

PR #61370, opened 2025-07-29
[Enhancement]DataCache supports TTL

jiangjiguang:
## Why I'm doing:
   When using SR on HIve with datacache，some users want to custom datacache TTL instead of LRU.

## What I'm doing:
   The StarCache has implemented TTL functionality in fact, the pr just control whether to enable this TTL function
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1318,8 +1318,9 @@ CONF_mInt64(datacache_mem_adjust_period, "20");
 CONF_mInt64(datacache_mem_adjust_interval_seconds, "10");
 
 CONF_Int32(datacache_inline_item_count_limit, "130172");
-CONF_Int32(datacache_ttl_check_interval_ms, "0");
-CONF_Int32(datacache_ttl_remove_batch, "100");
+// The interval for ttl reaper to check the expired cache item.
+// If set to 0, the ttl dose not take effect.
+CONF_Int32(datacache_ttl_check_interval_ms, "1000");
 // Whether use an unified datacache instance.
 CONF_Bool(datacache_unified_instance_enable, "true");
 // The eviction policy for datacache, alternatives: [lru, slru].
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1318,6 +1318,8 @@ CONF_mInt64(datacache_mem_adjust_period, "20");
 CONF_mInt64(datacache_mem_adjust_interval_seconds, "10");
 
 CONF_Int32(datacache_inline_item_count_limit, "130172");
+CONF_Int32(datacache_ttl_check_interval_ms, "0");
+CONF_Int32(datacache_ttl_remove_batch, "100");
 // Whether use an unified datacache instance.
 CONF_Bool(datacache_unified_instance_enable, "true");
 // The eviction policy for datacache, alternatives: [lru, slru].
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1320,7 +1320,7 @@ CONF_mInt64(datacache_mem_adjust_interval_seconds, "10");
 CONF_Int32(datacache_inline_item_count_limit, "130172");
 // The interval for ttl reaper to check the expired cache item.
 // If set to 0, the ttl dose not take effect.
-CONF_Int32(datacache_ttl_check_interval_ms, "1000");
+CONF_Int32(datacache_ttl_check_interval_ms, "0");
 // Whether use an unified datacache instance.
 CONF_Bool(datacache_unified_instance_enable, "true");
 // The eviction policy for datacache, alternatives: [lru, slru].
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/DataCacheStmtAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/DataCacheStmtAnalyzer.java
@@ -154,7 +154,7 @@ public class DataCacheStmtAnalyzer {
             // Duration for cache remains active.
             // Use PT0M to prevent expiration of the rule.
             // Use duration specified in ISO-8601 duration format (PnDTnHnMn).
-            long ttlSeconds = context.getSessionVariable().getDatacacheTTLSeconds();;
+            long ttlSeconds = context.getSessionVariable().getDatacacheTTLSeconds();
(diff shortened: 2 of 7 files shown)


wangsimo0:
can you illustrate more about why you want to use ttl?
ttl is not a safe guarantee. so this product concept may raise ambiguities in my opinion.

be/src/common/config.h
```diff
@@ -1318,6 +1318,8 @@ CONF_mInt64(datacache_mem_adjust_period, "20");
 CONF_mInt64(datacache_mem_adjust_interval_seconds, "10");
 
 CONF_Int32(datacache_inline_item_count_limit, "130172");
+CONF_Int32(datacache_ttl_check_interval_ms, "0");
+CONF_Int32(datacache_ttl_remove_batch, "100");
```
GavinMar: These configurations are the underlying configuration of cache library, and users usually do not need to modify it in starrocks, and most users do not know how to modify it, so there is no need to expose it.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/DataCacheStmtAnalyzer.java
```diff
@@ -154,13 +154,15 @@ public Void visitDataCacheSelectStatement(DataCacheSelectStatement statement, Co
...
             // Use duration specified in ISO-8601 duration format (PnDTnHnMn).
-            long ttlSeconds = 0L;
-            try {
-                ttlSeconds = Duration.parse(properties.getOrDefault("ttl", "PT0M")).toSeconds();
-            } catch (DateTimeParseException e) {
-                throw new SemanticException(String.format(
-                        "Illegal ttl format, use duration specified in ISO-8601 duration format (PnDTnHnMn). Error msg: %s",
-                        e.getMessage()));
+            long ttlSeconds = context.getSessionVariable().getDatacacheTTLSeconds();
+            if (properties.containsKey("ttl")) {
```
GavinMar, line 158: If the ttl only tacks effect for `cache select` population, I think the session variable is also unnecessary, because users can config it in `cache select`. This session makes users confused that all cached data, including populated by query, will be affected by the TTL.

GavinMar:
Thank you very much for this PR. However, I don't think it's necessary to expose these configurations to users now. The TTL function itself belongs to the beta function.

be/src/common/config.h
jiangjiguang: The cache production speed is different.  If users can't custom the two configurations. This may cause the cache production rate to be faster than the deletion rate

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/DataCacheStmtAnalyzer.java
jiangjiguang, line 158: Some users will active warmup datacache with `cache select`,  However, other users only passively populate the cache with query. So it is necessary to session variable

GavinMar: > Some users will active warmup datacache with `cache select`, However, other users only passively populate the cache with query. So it is necessary to session variable

I think you can make it as an invisible session first because the underlying TTL function has not been fully tested yet.

be/src/common/config.h
GavinMar: > The cache production speed is different. If users can't custom the two configurations. This may cause the cache production rate to be faster than the deletion rate

I think how to coordinate these two parameters is quite confusing for most users, so we can set a reasonable default value for `datacache_ttl_remove_batch` and only keep `datacache_ttl_check_interval_ms` for adjustment. This can satisfy most user cases.

jiangjiguang:
> ttl is not a safe guarantee. so this product concept may raise ambiguities in my opinion.

@wangsimo0 
Because some of our users are confused about the current LRU, which caches have been deleted and which caches still exist?  
If the cache is deleted, it will have a negative impact on query performance, 
A sudden slower in queries can cause unnecessary disputes.
Our users are very clear about what hive data needs to be cached and When the cache will be deteled. 
So they want to custom cache TTL themselves.

Besides after my test , a large amount of cache stored on disk will have a negative impact on query.
So We want unnecessary cache to be deleted as soon as possible, obviously LRU cannot achieve it. 
So We need cache TTL.

be/src/common/config.h
jiangjiguang: > > The cache production speed is different. If users can't custom the two configurations. This may cause the cache production rate to be faster than the deletion rate
> 
> I think how to coordinate these two parameters is quite confusing for most users, so we can set a reasonable default value for `datacache_ttl_remove_batch` and only keep `datacache_ttl_check_interval_ms` for adjustment. This can satisfy most user cases.

Good advice, fixed

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/DataCacheStmtAnalyzer.java
jiangjiguang: fixed

wangsimo0:
> > ttl is not a safe guarantee. so this product concept may raise ambiguities in my opinion.
> 
> @wangsimo0 Because some of our users are confused about the current LRU, which caches have been deleted and which caches still exist? If the cache is deleted, it will have a negative impact on query performance, A sudden slower in queries can cause unnecessary disputes. Our users are very clear about what hive data needs to be cached and When the cache will be deteled. So they want to custom cache TTL themselves.
> 
> Besides after my test , a large amount of cache stored on disk will have a negative impact on query. So We want unnecessary cache to be deleted as soon as possible, obviously LRU cannot achieve it. So We need cache TTL.

ttl won't guarantee the existence of the cache either. I think this will cause more confusion than SLRU, SLRU is very simple to explain, the more you query the higher the possibility the cache will exist. 
and I don't quite understand why cache will have a negative impact on query. Which version are you using? SR have async datacache populate, which will make the cache population less impact on query.

PR #61417, opened 2025-07-29
[Enhancement] Add a cleaner for BrpcStubCache to cleanup unused connections

duanyyyyyyy:
[Enhancement] Add a cleaner for BrpcStubCache to cleanup unused connections
## Why I'm doing:
The BRPC Stub cache will never be evicted and will always stored in the cache.
Therefore there will be some bug here.
In the case that when users are using K8s to deploy the StarRocks backend, after sometime running there will be many log like this.
image
After check the BRPC code in https://github.com/apache/brpc/blob/1.8.0/src/brpc/socket.cpp#L1340
And from the grafana monitor 
image
This cluster have 148 BE nodes, normally it will be about 148 BRPC stub but after some times running, some pods restarted and the IP changed, some of the node BRPC stub creased to over 200.
And as we all known the endpoint in BRPC are made of ip and host, that means if the IP is changed the endpoint will still in the cache.
## What I'm doing:
Here I introduce a BrpcStubManager to periodically check and cleanup expired BRPC stub for BrpcStubCache, HttpBrpcStubCache and LakeServiceBrpcStubCache 
1. Record the last access time when get the stub for given endpoint
2. Loop all the endpoint in `_stub_map` or `_last_access_time_map` check if now - last_access_time is larger than expire time or not
3. If less, will not clean up the endpoint
4. If larger or equal, will try to clean up the endpoint
## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -58,6 +58,9 @@ CONF_String(ssl_private_key_path, "");
 // The max number of single connections maintained by the brpc client and each server.
 // These connections are created during the first few access and will be used thereafter
 CONF_Int32(brpc_max_connections_per_server, "1");
+// BRPC stub cache expire configurations
+// The expire time of BRPC stub cache, default 60 minutes.
+CONF_mInt32(brpc_stub_expire_s, "3600"); // 60 minutes
 
 // Declare a selection strategy for those servers have many ips.
 // Note that there should at most one ip match this list.
--- a/be/src/util/brpc_stub_cache.cpp
+++ b/be/src/util/brpc_stub_cache.cpp
@@ -25,7 +25,6 @@ namespace starrocks {
 
 BrpcStubCache::BrpcStubCache(ExecEnv* exec_env) : _pipeline_timer(exec_env->pipeline_timer()) {
     _stub_map.init(239);
-    _pipeline_timer = exec_env->pipeline_timer();
     REGISTER_GAUGE_STARROCKS_METRIC(brpc_endpoint_stub_count, [this]() {
         std::lock_guard<SpinLock> l(_lock);
         return _stub_map.size();
--- a/be/src/util/brpc_stub_cache.cpp
+++ b/be/src/util/brpc_stub_cache.cpp
@@ -14,17 +14,16 @@
 
 #include "util/brpc_stub_cache.h"
 
-#include <runtime/exec_env.h>
-
 #include "common/config.h"
 #include "gen_cpp/internal_service.pb.h"
 #include "gen_cpp/lake_service.pb.h"
+#include "runtime/exec_env.h"
 #include "util/failpoint/fail_point.h"
 #include "util/starrocks_metrics.h"
 
 namespace starrocks {
 
-BrpcStubCache::BrpcStubCache(ExecEnv* exec_env) {
+BrpcStubCache::BrpcStubCache(ExecEnv* exec_env) : _pipeline_timer(exec_env->pipeline_timer()) {
     _stub_map.init(239);
     _pipeline_timer = exec_env->pipeline_timer();
     REGISTER_GAUGE_STARROCKS_METRIC(brpc_endpoint_stub_count, [this]() {
@@ -61,11 +60,12 @@ std::shared_ptr<PInternalService_RecoverableStub> BrpcStubCache::get_stub(const 
         stub_pool = _stub_map.seek(endpoint);
     }
 
-    _pipeline_timer->unschedule((*stub_pool)->_cleanup_task);
-    timespec tm = butil::seconds_from_now(config::brpc_stub_expire_s);
-    auto status = _pipeline_timer->schedule((*stub_pool)->_cleanup_task, tm);
-    if (!status.ok()) {
-        LOG(WARNING) << "Failed to schedule brpc cleanup task: " << endpoint;
+    if (_pipeline_timer->unschedule((*stub_pool)->_cleanup_task) != 1) {
+        timespec tm = butil::seconds_from_now(config::brpc_stub_expire_s);
+        auto status = _pipeline_timer->schedule((*stub_pool)->_cleanup_task, tm);
+        if (!status.ok()) {
+            LOG(WARNING) << "Failed to schedule brpc cleanup task: " << endpoint;
+        }
     }
 
     return (*stub_pool)->get_or_create(endpoint);
@@ -134,7 +134,6 @@ HttpBrpcStubCache* HttpBrpcStubCache::getInstance() {
 
 HttpBrpcStubCache::HttpBrpcStubCache() {
     _stub_map.init(500);
-    _task_map.init(500);
     _pipeline_timer = ExecEnv::GetInstance()->pipeline_timer();
 }
 
@@ -143,8 +142,8 @@ HttpBrpcStubCache::~HttpBrpcStubCache() {
 
     {
         std::lock_guard<SpinLock> l(_lock);
-        for (auto& stub : _task_map) {
-            task_to_cleanup.push_back(stub.second);
+        for (auto& stub : _stub_map) {
+            task_to_cleanup.push_back(stub.second.second);
         }
     }
 
@@ -153,7 +152,6 @@ HttpBrpcStubCache::~HttpBrpcStubCache() {
     }
 
     _stub_map.clear();
-    _task_map.clear();
 }
 
 StatusOr<std::shared_ptr<PInternalService_RecoverableStub>> HttpBrpcStubCache::get_http_stub(
@@ -176,32 +174,29 @@ StatusOr<std::shared_ptr<PInternalService_RecoverableStub>> HttpBrpcStubCache::g
     // get is exist
     std::lock_guard<SpinLock> l(_lock);
 
-    // schedule clean up task
-    auto task = _task_map.seek(endpoint);
-    if (task == nullptr) {
+    auto stub_pair_ptr = _stub_map.seek(endpoint);
+    if (stub_pair_ptr == nullptr) {
+        // create
         auto new_task = std::make_shared<HttpEndpointCleanupTask>(this, endpoint);
-        _task_map.insert(endpoint, new_task);
-        task = _task_map.seek(endpoint);
-    }
-    _pipeline_timer->unschedule((*task).get());
-    timespec tm = butil::seconds_from_now(config::brpc_stub_expire_s);
-    auto status = _pipeline_timer->schedule((*task).get(), tm);
-    if (!status.ok()) {
-        LOG(WARNING) << "Failed to schedule http brpc cleanup task: " << endpoint;
+        auto stub = std::make_shared<PInternalService_RecoverableStub>(endpoint, "http");
+        if (!stub->reset_channel().ok()) {
+            return Status::RuntimeError("init brpc http channel error on " + taddr.hostname + ":" +
+                                        std::to_string(taddr.port));
+        }
+        _stub_map.insert(endpoint, std::make_pair(stub, new_task));
+        stub_pair_ptr = _stub_map.seek(endpoint);
     }
 
-    auto stub_ptr = _stub_map.seek(endpoint);
-    if (stub_ptr != nullptr) {
-        return *stub_ptr;
+    // schedule clean up task
+    if (_pipeline_timer->unschedule((*stub_pair_ptr).second.get()) != 1) {
+        timespec tm = butil::seconds_from_now(config::brpc_stub_expire_s);
+        auto status = _pipeline_timer->schedule((*stub_pair_ptr).second.get(), tm);
+        if (!status.ok()) {
+            LOG(WARNING) << "Failed to schedule http brpc cleanup task: " << endpoint;
+        }
     }
-    // create
-    auto stub = std::make_shared<PInternalService_RecoverableStub>(endpoint, "http");
-    if (!stub->reset_channel().ok()) {
-        return Status::RuntimeError("init brpc http channel error on " + taddr.hostname + ":" +
-                                    std::to_string(taddr.port));
-    }
-    _stub_map.insert(endpoint, stub);
-    return stub;
+
+    return (*stub_pair_ptr).first;
 }
 
 void HttpBrpcStubCache::cleanup_expired(const butil::EndPoint& endpoint) {
@@ -209,7 +204,6 @@ void HttpBrpcStubCache::cleanup_expired(const butil::EndPoint& endpoint) {
 
     LOG(INFO) << "cleanup http stubs from endpoint:" << endpoint;
     _stub_map.erase(endpoint);
-    _task_map.erase(endpoint);
 }
 
 LakeServiceBrpcStubCache* LakeServiceBrpcStubCache::getInstance() {
@@ -219,7 +213,6 @@ LakeServiceBrpcStubCache* LakeServiceBrpcStubCache::getInstance() {
 
 LakeServiceBrpcStubCache::LakeServiceBrpcStubCache() {
     _stub_map.init(500);
-    _task_map.init(500);
     _pipeline_timer = ExecEnv::GetInstance()->pipeline_timer();
 }
 
@@ -228,8 +221,8 @@ LakeServiceBrpcStubCache::~LakeServiceBrpcStubCache() {
 
     {
         std::lock_guard<SpinLock> l(_lock);
-        for (auto& stub : _task_map) {
-            task_to_cleanup.push_back(stub.second);
+        for (auto& stub : _stub_map) {
+            task_to_cleanup.push_back(stub.second.second);
         }
     }
 
@@ -238,7 +231,6 @@ LakeServiceBrpcStubCache::~LakeServiceBrpcStubCache() {
     }
 
     _stub_map.clear();
-    _task_map.clear();
 }
 
 DEFINE_FAIL_POINT(get_stub_return_nullptr);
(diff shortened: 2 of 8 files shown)


duanyyyyyyy:
Hi @kevincai master, are there any suggestion for this case? Or any comment for this PR?

kevincai:
> Hi @kevincai master, are there any suggestion for this case? Or any comment for this PR?

will take a look today.

be/src/util/brpc_stub_cache.h
```diff
@@ -56,18 +58,26 @@ class BrpcStubCache {
...
         std::vector<std::shared_ptr<PInternalService_RecoverableStub>> _stubs;
         int64_t _idx;
     };
 
+    StubPool* get_stub_pool_readonly(const butil::EndPoint& endpoint);
+
     SpinLock _lock;
     butil::FlatMap<butil::EndPoint, StubPool*> _stub_map;
+    std::atomic<bool> _stopped{false};
+    std::thread _cleanup_thread;
```
kevincai: there are `LakeServiceBrpcStubCache` and `HttpBrpcStubCache` which may have the same issue as described in this PR, should apply the same strategy to the StubCache.

be/src/util/internal_service_recoverable_stub.cpp
```diff
@@ -60,6 +60,67 @@ Status PInternalService_RecoverableStub::reset_channel(int64_t next_connection_g
     return Status::OK();
 }
 
+Status PInternalService_RecoverableStub::check_health() {
+#ifdef BE_TEST
+    if (_endpoint.port == 125) {
+        LOG(INFO) << "BE_TEST mode: Port 125 simulated as unhealthy for " << _endpoint;
+        return Status::InternalError("Simulated failure for port 111");
```
kevincai: Log and error msg inconsistent, is it port 125 or port 111?

be/src/util/internal_service_recoverable_stub.cpp
```diff
@@ -60,6 +60,67 @@ Status PInternalService_RecoverableStub::reset_channel(int64_t next_connection_g
...
+    // extract host
+    std::string host = butil::endpoint2str(_endpoint).c_str();
+    size_t colon_pos = host.find(':');
+    if (colon_pos != std::string::npos) {
+        host = host.substr(0, colon_pos);
+    }
+
+    // generate endpoint
+    butil::EndPoint http_endpoint;
+    if (butil::str2endpoint((host + ":" + std::to_string(config::be_http_port)).c_str(), &http_endpoint) != 0) {
```
kevincai: theoretically, it is allowed that different BE has different http port configured, this implementation will add the restriction that all the BEs/CNs have the same http port configured which should be avoided.

be/src/util/internal_service_recoverable_stub.cpp
```diff
@@ -60,6 +60,67 @@ Status PInternalService_RecoverableStub::reset_channel(int64_t next_connection_g
...
+    brpc::Channel channel;
+    if (channel.Init(http_endpoint, &options) != 0) {
+        LOG(WARNING) << "Failed to init HTTP channel for " << http_endpoint;
+        return Status::InternalError("Failed to init HTTP channel");
+    }
+
+    // generate controller
+    brpc::Controller cntl;
+    cntl.http_request().set_method(brpc::HTTP_METHOD_GET);
+    cntl.http_request().uri() = "/api/health";
```
kevincai: Instead of depending on the brpc server `/api/health`, shall it be good enough to take socket connectable as the healthy indicator, and if a few consecutive socket connection failed, take the service as down and remove the from stub cache?

Another thing is to expire the stub by its creating time, abandon the stub after X seconds and replace with a new one no matter it is healthy or not. will it be much easier and also can help to remedy some weird errors where the brpc channel can't be recovered after a long usage.

be/src/util/internal_service_recoverable_stub.cpp
duanyyyyyyy: Sorry for typo

be/src/util/internal_service_recoverable_stub.cpp
duanyyyyyyy: For the first question, seems no way to get the socket ptr either from the channel ptr or from the channelBase ptr in the BRPC. The second one would be much easier.

For the second question. Use expire time will be much easier, but it will make the BRPC channel have some cold start if there is a long time we have not send RPC request. I am not sure it will be a good design or not.

kevincai: If channel interface is not sufficient, maybe we can use brpc Socket interface directly. Personally I am preferred with the expiration time. As long as the expiration is not that frequent (tens of mins up to hours maybe?) the cold start shall be acceptable.

duanyyyyyyy: Got it, let me commit the version with expiration. Actually speaking, I tested two version code in our internal service, both can solve the issue.

xhumanoid:
@duanyyyyyyy gentle ping =)

duanyyyyyyy:
> @duanyyyyyyy gentle ping =)

sry for forget.

duanyyyyyyy:
> > @duanyyyyyyy gentle ping =)
> 
> sry for forget.

duanyyyyyyy:
> @duanyyyyyyy gentle ping =)

Seems some unrelated test failed, let me retry for it

be/src/util/internal_service_recoverable_stub.cpp
duanyyyyyyy: @kevincai Finished, R U free to review?

be/src/util/brpc_stub_cache.cpp
```diff
@@ -169,7 +234,50 @@ StatusOr<std::shared_ptr<starrocks::LakeService_RecoverableStub>> LakeServiceBrp
...
+}
+
+BrpcStubManager::BrpcStubManager(ExecEnv* exec_env) : _exec_env(exec_env) {
+    _cleanup_thread = std::thread([this] {
+#ifdef GOOGLE_PROFILER
+        ProfilerRegisterThread();
+#endif
+        while (!_is_stopped.load()) {
+            LOG(INFO) << "Start to clean up expired brpc stubs";
+            _exec_env->brpc_stub_cache()->check_and_cleanup_expired_stubs(config::brpc_stub_expire_ms);
```
kevincai: better capture the configvar into a local var before doing the actual cleanup, to avoid inconsistent expiration in the same round.

be/src/common/config.h
```diff
@@ -59,6 +59,13 @@ CONF_String(ssl_private_key_path, "");
 // These connections are created during the first few access and will be used thereafter
 CONF_Int32(brpc_max_connections_per_server, "1");
 
+// BRPC stub cache cleanup configurations
+// Interval in seconds to run cleanup for expired BRPC stubs
+CONF_mInt32(brpc_stub_cleanup_interval_s, "600"); // 10 minutes
```
kevincai: would you please also add the new configurations to `docs/en/administration/management/BE_configuration.md`

kevincai (commented):
overall implementation looks good to me. Thanks!

kevincai:
@duanyyyyyyy looks like you have to do "git commit -s" for all the commits in order to pass the DCO check.

duanyyyyyyy:
@kevincai All comments are resolved

kevincai (dismissed):

alvin-phoenix-ai:
@cursor review

duanyyyyyyy:
@alvin-celerdata Any other comments?

be/src/util/brpc_stub_cache.cpp
```diff
@@ -169,7 +235,52 @@ StatusOr<std::shared_ptr<starrocks::LakeService_RecoverableStub>> LakeServiceBrp
...
+
+    for (const auto& endpoint : expired_endpoints) {
+        LOG(INFO) << "cleanup lake service stub from endpoint:" << endpoint;
+        _stub_map.erase(endpoint);
+        _last_access_time_map.erase(endpoint);
+    }
+}
+
+BrpcStubManager::BrpcStubManager(ExecEnv* exec_env) : _exec_env(exec_env) {
+    _cleanup_thread = std::thread([this] {
```
stdpain: We can use a better implementation instead of adding extra threads to do this.
We can update the timeout task each time we get a stub. You can use a class PipelineTimer to implement this.

```
std::shared_ptr<PInternalService_RecoverableStub> BrpcStubCache::get_stub(const butil::EndPoint& endpoint) {
if (_task ) timer->unschedule(_task);
_task = timer->schedule(timeout, callback)
// 
......
```

duanyyyyyyy: Let me try to implement like that

duanyyyyyyy: I thought abut your suggestion.
I can commit one version in another commit but I think if we introduce PipelineTimer here, the logic will become complicated....

stdpain: > I thought abut your suggestion. I can commit one version in another commit but I think if we introduce PipelineTimer here, the logic will become complicated....

I don't think this is a complicated modification. This class can help you finish this task.

https://github.com/StarRocks/starrocks/blob/main/be/src/exec/pipeline/schedule/pipeline_timer.h#L63C1-L75C1

kevincai: I vote NOT introducing pipelineTimer here directly, shall do some refactor to move the PipelineTimer out of the exec/pipeline before it can be referenced in utility modules.

If timer is suggested, how about the PeriodicTask from brpc library? PeriodicTaskManager::StartTaskAt(task, abstime) or use bthread_timer_add() interface directly.

stdpain: > I vote NOT introducing pipelineTimer here directly, shall do some refactor to move the PipelineTimer out of the exec/pipeline before it can be referenced in utility modules.
> 
> If timer is suggested, how about the PeriodicTask from brpc library? PeriodicTaskManager::StartTaskAt(task, abstime) or use bthread_timer_add() interface directly.

I prefer to put the pipeline timer in the public component. The interface provided by brpc requires additional adaptation (careful handling of details).

kevincai: @stdpain Will it be risky to make the PipelineTimer public, which potentially a bad timer task may block the dedicated timer thread for a long time and consequently blocks timed tasks for pipeline scheduling and bring down the entire pipeline engine?

stdpain: > @stdpain Will it be risky to make the PipelineTimer public, which potentially a bad timer task may block the dedicated timer thread for a long time and consequently blocks timed tasks for pipeline scheduling and bring down the entire pipeline engine?

There is no need to worry about this. Based on the following two pieces of information.
1. Stub timeouts are low-frequency events, so in most cases, the callback function will not be executed.
2. The callback function is a lightweight operation; we will only remove the stub from the map. This critical section is very small, and we believe it will not cause any blocking.

duanyyyyyyy: > @stdpain Will it be risky to make the PipelineTimer public, which potentially a bad timer task may block the dedicated timer thread for a long time and consequently blocks timed tasks for pipeline scheduling and bring down the entire pipeline engine?

It will not be risk for this case, as cleanup operation will be very light

kevincai: The PipelineTimer itself is implemented as dedicated bthread for the tasks, as long as the tasks are not reuse the same PipelineTimer instance, it should be OK.

The concern about the PipelineTimer reuse, is that every TimerTask added into the Timer must be a behaved citizen, it is safe to say that this stub cleanup is lightweight enough that will not introduce risks blocking the bthread. However the risk is if the PipelineTimer used by the pipeline Engine is public accessible, it could be potentially abused used if code review is overlooked. Better put that in private access so the Timer for the pipeline engine can be assured in good shape.

kevincai: Since the PipelineTimer is already exposed through ExecEnv, I think there is no blocker to go with the direction with the evaluation of low risk of adding this cleanup into the Timer. @duanyyyyyyy please go ahead making necessary change. This PR shall not be blocked by the design discuss related to the restriction of the PipelineTimer usage.

be/src/util/brpc_stub_cache.cpp
```diff
@@ -34,10 +38,24 @@ BrpcStubCache::~BrpcStubCache() {
...
+    timespec tm = butil::seconds_from_now(config::brpc_stub_expire_s);
+    auto task = std::make_shared<EndpointCleanupTask>(this, endpoint);
+    auto st = _pipeline_timer->schedule(task.get(), tm);
+    if (st.ok()) {
+        _cleanup_task_map[endpoint] = task;
+    }
+
     auto stub_pool = _stub_map.seek(endpoint);
     if (stub_pool == nullptr) {
         StubPool* pool = new StubPool();
```
stdpain: how about use the same map ?.
```
    butil::FlatMap<butil::EndPoint, std::pair<StubPool*, std::shared_ptr<EndpointCleanupTask>>> _stub_map;
```

duanyyyyyyy:
@kevincai @stdpain Finished

duanyyyyyyy:
> @kevincai @stdpain Finished

Sry for miss one comment from stdpain, let me fix it.

be/src/util/brpc_stub_cache.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include "util/brpc_stub_cache.h"
 
+#include <runtime/exec_env.h>
```
kevincai:
```suggestion
#include "runtime/exec_env.h"
```

be/src/util/brpc_stub_cache.h
```diff
@@ -34,6 +34,8 @@
 
 #pragma once
 
+#include <exec/pipeline/schedule/pipeline_timer.h>
```
kevincai:
```suggestion
#include "exec/pipeline/schedule/pipeline_timer.h"
```

be/src/util/brpc_stub_cache.cpp
```diff
@@ -22,28 +24,50 @@
 
 namespace starrocks {
 
-BrpcStubCache::BrpcStubCache() {
+BrpcStubCache::BrpcStubCache(ExecEnv* exec_env) {
```
kevincai:
```suggestion
BrpcStubCache::BrpcStubCache(ExecEnv* exec_env) : _pipeline_timer(exec_env->pipeline_timer()) {
```

duanyyyyyyy:
@kevincai @stdpain R U free to take a look? I think already finished all comments.

be/src/util/brpc_stub_cache.cpp
```diff
@@ -22,28 +24,50 @@
...
-        StubPool* pool = new StubPool();
-        _stub_map.insert(endpoint, pool);
-        return pool->get_or_create(endpoint);
+        auto new_pool = std::make_shared<StubPool>();
+        new_pool->_cleanup_task = new EndpointCleanupTask(this, endpoint);
+        _stub_map.insert(endpoint, new_pool);
+        stub_pool = _stub_map.seek(endpoint);
     }
+
+    _pipeline_timer->unschedule((*stub_pool)->_cleanup_task);
```
kevincai: what if the unschedule failed?

be/src/util/brpc_stub_cache.h
```diff
@@ -48,54 +50,105 @@
...
 
 private:
     HttpBrpcStubCache();
     HttpBrpcStubCache(const HttpBrpcStubCache&) = delete;
     HttpBrpcStubCache& operator=(const HttpBrpcStubCache&) = delete;
+    ~HttpBrpcStubCache();
 
     SpinLock _lock;
-    butil::FlatMap<butil::EndPoint, std::shared_ptr<PInternalService_RecoverableStub>> _stub_map;
+    butil::FlatMap<butil::EndPoint, std::pair<std::shared_ptr<PInternalService_RecoverableStub>,
```
kevincai, line 106: Every endpoint is registered with a TimerTask, not sure if this is feasible for a cluster with large scale of nodes, e.g. hundreds of backends/compute nodes, will this be carefully evaluated?

be/src/util/brpc_stub_cache.cpp
duanyyyyyyy: Do not need to worry about unschedule failed.
As we have lock in the cleanup_expire, that means even if the task is scheduling but the lock got by the get_stub there will not be any dead lock or NPE issue.

kevincai: assume the execute sequence is, 
1. thread-1: get lock, 
2. thread-2: timer task wait for lock, 
3. thread-1: get the endpoint, cancel the timer task (fail), reschedule the task, return the stub
4. thread-2: get the lock, clean the endpoint from the map, deref the stub and task shared_ptr (task shared_ptr shall be count down to zero)
5. thread-2: next time the task (registered in step 3 by raw pointer) timed out and get executed, will result an use-after-free.

be/src/util/brpc_stub_cache.h
duanyyyyyyy: I think can.
From the documentation we can see that hundreds of task can run without waiting for lock.
https://github.com/apache/brpc/blob/master/docs/cn/timer_keeping.md

duanyyyyyyy: @stdpain Any suggestions master?

stdpain: You can use it with confidence. They will be stored in the heap after being bucketed by the timer. So there's no need to worry about memory or other overhead.

be/src/util/brpc_stub_cache.cpp
```diff
@@ -17,33 +17,57 @@
...
+#include "runtime/exec_env.h"
 #include "util/failpoint/fail_point.h"
 #include "util/starrocks_metrics.h"
 
 namespace starrocks {
 
-BrpcStubCache::BrpcStubCache() {
+BrpcStubCache::BrpcStubCache(ExecEnv* exec_env) : _pipeline_timer(exec_env->pipeline_timer()) {
     _stub_map.init(239);
+    _pipeline_timer = exec_env->pipeline_timer();
```
kevincai: _pipeline_timer is initialized in the init list, this line is redundant.

duanyyyyyyy: Any other comments?
Let's wait for the CI but I think the CI will not be problem.

duanyyyyyyy:
@kevincai Master, any other comments?

be/src/util/brpc_stub_cache.cpp
```diff
@@ -71,10 +94,22 @@ std::shared_ptr<PInternalService_RecoverableStub> BrpcStubCache::get_stub(const
     return get_stub(endpoint);
 }
 
+void BrpcStubCache::cleanup_expired(const butil::EndPoint& endpoint) {
+    std::lock_guard<SpinLock> l(_lock);
+
+    LOG(INFO) << "cleanup stubs from endpoint:" << endpoint;
```
kevincai:
```suggestion
    LOG(INFO) << "cleanup stubs to endpoint:" << endpoint;
```

be/src/util/brpc_stub_cache.h
```diff
@@ -48,54 +49,105 @@
...
+
+class EndpointCleanupTask : public starrocks::pipeline::PipelineTimerTask {
+public:
+    EndpointCleanupTask(BrpcStubCache* cache, const butil::EndPoint& endpoint) : _cache(cache), _endpoint(endpoint){};
+    void Run() override;
+
+private:
+    BrpcStubCache* _cache;
+    butil::EndPoint _endpoint;
+};
```
kevincai: these *CleanupTask can be simplified with templating as 

```suggestion
template<typename StubCacheT>
class EndpointCleanupTask : public starrocks::pipeline::PipelineTimerTask {
public:
    EndpointCleanupTask(StubCacheT* cache, const butil::EndPoint& endpoint) : _cache(cache), _endpoint(endpoint){};
    void Run() override {
        _cache->cleanup_expired(_endpoint);
    }

private:
    StubCacheT* _cache;
    butil::EndPoint _endpoint;
};
```

kevincai (commented):
overall looks good to me.

kevincai (approved):

duanyyyyyyy:
@stdpain Master, any comments?

stdpain (approved):

PR #61489, opened 2025-07-31
[Enhancement] support config page cache shard count

starrocks-xupeng:
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/cache/lrucache_engine.cpp
+++ b/be/src/cache/lrucache_engine.cpp
@@ -18,7 +18,7 @@
 
 namespace starrocks {
 Status LRUCacheEngine::init(const CacheOptions& options) {
-    _cache = std::make_unique<ShardedLRUCache>(options.mem_space_size);
+    _cache = std::make_unique<ShardedLRUCache>(options.mem_space_size, config::page_cache_shard_bits);
     _initialized.store(true, std::memory_order_relaxed);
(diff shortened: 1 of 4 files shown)


be/src/cache/lrucache_engine.cpp
```diff
@@ -18,7 +18,7 @@
 
 namespace starrocks {
 Status LRUCacheEngine::init(const CacheOptions& options) {
-    _cache = std::make_unique<ShardedLRUCache>(options.mem_space_size);
+    _cache = std::make_unique<ShardedLRUCache>(options.mem_space_size, config::page_cache_shard_bits);
```
kevincai, line 21: shall be able to calculate a proper shard bits based on the cache capacity and average page size (64KB). If the configuration is given as `0`

kevincai (commented):
prefer to add benchmark test against the LRUCache to know exactly the number of shards affecting the performance of the LRUCache insertion/eviction

PR #61498, opened 2025-07-31
[Enhancement] Simplify chunks_sorter_full_sort by removing the merge

griebelsi:
**!This is currently still a draft, because the changes only support 32bit indexes!**

## Why I'm doing:
(If you want I can also create an issue for the topics mentioned here, that are not touched by this PR. However you prefer it.)
The sort is currently quite slow. There are multiple ways to improve this, this is just a first step. There are two separate parts to this:
1. Large partitions (partition by partitions) are routed through a single chunks_sorter_full_sort node, leading to all elements being sorted by a single thread. This probably needs more fundamental changes in the task scheduling system.
2. The sort code itself is not optimal. This is split into the part of the code the code related to the ChunksSorter and the part that is done in parallel_merge_source_operators. This PR only touches the former and here only the chunks_sorter_full_sort. When looking at the code I was wondering why the sort is not completely handled by the parallel_merge_source operators. Another point would be to generate the comparison function with llvm (and probably add fixed template instantiations for sorting 1 or two columns). This would reduce the amount of indirections to a minimum (one indirect function call per row vs. more than one virtual function call per row and column).

## What I'm doing:

The chunks_sorter_full_sort is split into two steps. First batches of chunks are aggregated and then (after a threshold is hit) sorted by _partial_sort. Once all data is read for the node, these batches are then merged. As we need to wait for all chunks to arrive anyway, I would say that there is no reason to split the sort like this. This is why I removed the merge part completly and just used _partial_sort (I just realized I probably should rename this). In local tests, this improved the runtime for 100mio strings in very skewed partitions from 90s to 75s and it made the code much simpler. This should allow future improvements to be implemented much quicker.

## Open points
- [ ] Make the code work for more than 2**32 elements.
I tried to use sort_vertical_chunks for that, but it made the code ~5 times slower. I removed some shared pointer copies, but I still could not make it faster, which is why I want to stay with the current approach.
- [ ] Ensure that the spilling changes are correct.
As I do not understand the spilling code, it would be good if you can check whether the changes are fine.
- [ ] Check whether (parts of) the merge related code can be removed.
I did not check yet whether other code still uses the merge code that was used in chunks_sorter_full_sort. If not this code can be removed as well.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/bench/chunks_sorter_bench.cpp
+++ b/be/src/bench/chunks_sorter_bench.cpp
@@ -227,7 +227,6 @@ static void do_bench(benchmark::State& state, SortAlgorithm sorter_algo, Logical
     int64_t data_size = 0;
     int64_t mem_usage = 0;
     const int64_t max_buffered_rows = 1024 * 1024;
-    const int64_t max_buffered_bytes = max_buffered_rows * 256;
     const std::vector<SlotId> early_materialized_slots;
 
     for (auto _ : state) {
@@ -241,8 +240,7 @@ static void do_bench(benchmark::State& state, SortAlgorithm sorter_algo, Logical
         switch (sorter_algo) {
         case FullSort: {
             sorter = std::make_unique<ChunksSorterFullSort>(suite._runtime_state.get(), &sort_exprs, &asc_arr,
-                                                            &null_first, "", max_buffered_rows, max_buffered_bytes,
-                                                            early_materialized_slots);
+                                                            &null_first, "");
             expected_rows = total_rows;
             break;
         }
--- a/be/src/exec/chunks_sorter_full_sort.cpp
+++ b/be/src/exec/chunks_sorter_full_sort.cpp
@@ -20,6 +20,7 @@
 #include "exprs/expr.h"
 #include "gutil/strings/substitute.h"
 #include "runtime/runtime_state.h"
+#include "sorting/sort_permute.h"
 
 namespace starrocks {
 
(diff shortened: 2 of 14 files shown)


satanson:
@griebelsi 
full sort is two staged, the first stage is local partition sort which is  processed distributedly on several compute node (backend),  the each local partition sort outputs a sorted run, all of sorted runs are delivered to remote compute node which gather all and merge them.

the code you working on is first stage, local partiton sort.  it has two implementations.
1. enable_parallel_merge=false:   the pipeline is PartitionSortSinkOperatorFactory->LocalMergeSortSourceOperatorFactory,  merge is required in chunks_sorter_full_sort, since LocalMergeSortSourceOperatorFactory depends on ordered output of PartitionSortSinkOperatorFactory.
2. enable_parallel_merge=true,  the pipeline is PartitionSortSinkOperatorFactory->LocalParallelMergeSortSourceOperatorFactory; in this situtation, merge operations on partial_sort's outcomes can removed.

However, for legacy reason,  both pipelines reuse the PartitionSortSinkOperatorFactory. It seems that you want improve enable_parallel_merge mode,  I suggest to implement a new PartitionSortSinkOperatorFactory version for LocalParallelMergeSortSourceOperatorFactory, preventing disrupting the LocalMergeSortSourceOperatorFactory.

griebelsi:
Hi, just so that I understand you correctly:
If `enable_parallel_merge=true`, we can create a new SortSinkOperator(Factory), that just passes the chunks unsorted to the LocalParallelMergeSortSourceOperator and then implement the whole sort algorithm there, while we leave the other code paths alone?

satanson:
> Hi, just so that I understand you correctly: If `enable_parallel_merge=true`, we can create a new SortSinkOperator(Factory), that just passes the chunks unsorted to the LocalParallelMergeSortSourceOperator and then implement the whole sort algorithm there, while we leave the other code paths alone?

SortSinkOperator in this situation should accumulate enough chunks then sorted them to form a sorted run, it need not  merge all sorted runs into a final sorted run; instead, each SortSinkOperator can offer these sorted runs to LocalParallelMergeSortSourceOperator to merge all at once.

However,  LocalParallelMergeSortSourceOperator adopts two-way merging algorithm, so the height of merge tiers are approximately  logN (N is number of initial sorted runs). height would become greater as number of intial sorted runs increases, merge tiers increasing would result into more memory copy/moves, cost of which would be significant and can not be neglected.

So the solution should be testified rigorously, if inital sorted runs are too many, should LocalParallelMergeSortSourceOperator leverage multi-way merging instead of 2-way merging?

griebelsi:
> SortSinkOperator in this situation should accumulate enough chunks then sorted them to form a sorted run, it need not merge all sorted runs into a final sorted run; instead, each SortSinkOperator can offer these sorted runs to LocalParallelMergeSortSourceOperator to merge all at once.

As the chunks are still completely unsorted when they reach the SinkOperator, it has to accumulate all unsorted chunks, before it can continue. For the output it can then indeed stream the sorted chunks, but if I understand it correctly, the case, where you only need to produce a part of the output (due to the limit) is already handled by other classes. So in my opinion, the ChunksSorterFullSort class can always assume that it needs to forward the whole sorted output to the SourceOperator.

> However, LocalParallelMergeSortSourceOperator adopts two-way merging algorithm, so the height of merge tiers are approximately logN (N is number of initial sorted runs). height would become greater as number of intial sorted runs increases, merge tiers increasing would result into more memory copy/moves, cost of which would be significant and can not be neglected.
>
> So the solution should be testified rigorously, if inital sorted runs are too many, should LocalParallelMergeSortSourceOperator leverage multi-way merging instead of 2-way merging?

If I understand it correctly, the height of the merge tree is not dependent on the number of input chunks/runs, but on the number of (sorted) "chunk providers". Each chunk provider can provide multiple sorted chunks/runs, and if I understand correctly every SinkOperator instance is mapped to one chunk provider.
I looked into improving the ParallelMergeSortSourceOperator by implementing a direct k-way merge sort, but since the inputs are streamed to the operator and since they are split into chunks, it becomes quite complex to keep track of everything. In my first benchmark with very low merge tree heights, the performance of my implementation was the same as the current one. I still have to test it for larger merge trees, but due to the complexity of the implementation, I wanted to improve the single threaded part first.

griebelsi:
With my last commits, I "finished" the "logical" part of this PR (i.e. only tests and clean-ups left from my side) , but I wanted to get some feedback first, because the amount of lines needed to be changed became more than I expected. This is mainly due to making the existing structure more flexible. The question is now whether you feel confident to merge it like this or how we should continue. I think this PR is the minimum change, which achieves a substantial performance improvement of the sort. Other possibilities too improve the sort performance that I see are:
1. Not routing a single partition through a single chunks_sorter_full_sort operator
2. Using the sort_vertical_chunks function, while also improving its performance (mainly removing unnecessary pointer companies). I tried this, but came to the conclusion that this would be an even bigger change (but I might have overlooked something).
3. Generating the sort function with llvm and only having one sort pass.
4. Do not do any sort in the sort_sink_operator, but move the complete responsibility to the parallel_merge_sort_source_operator.
Not all of these points are exclusive, but in my mind my current approach is the quickest solution for now, as the others require larger changes in the current code structure according to my current understanding.

To make my changes easier to review I could split them up into three parts:

1. Make materialize_by_permutation work on SmallPermutations and use the sort_and_tie function, that directly works with with a SmallPermutation for the chunks_sorter_full_sort.
2. Remove the unnecessary pointer copies in materialize_by_permutation (and VerticalColumnSorter). This is a smaller PR which should increase the performance of the components quite a bit.
3. Refactor the chunks_sorter_full_sort to use the 64bit permutation vector.

griebelsi:
This PR will be split up into multiple smaller PRs.

PR #61530, opened 2025-08-01
[BugFix] fix builtin_function fuzzy test

murphyatwork:
## What I'm doing:

1. bits_function: the implementation is wrong
2. change the static `DCHECK` to dynamic argument validation for some functions
3. fix some type mapping error in the `logical_type.cpp`

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/column/column_helper.h
+++ b/be/src/column/column_helper.h
@@ -280,47 +280,30 @@ public:
      */
     template <LogicalType Type>
     static inline typename RunTimeColumnType<Type>::Ptr cast_to(const ColumnPtr& value) {
-#ifdef NDEBUG
-        down_cast<const RunTimeColumnType<Type>*>(value.get());
-#else
+#ifndef NDEBUG
         auto* result = dynamic_cast<const RunTimeColumnType<Type>*>(value.get());
-        if (result == nullptr) {
-            const auto* ptr = value.get();
-            std::cerr << "Cast failed for column: "
-                      << " (expected type: " << Type << ", actual type: " << ptr->get_name() << ")" << std::endl;
-        }
-        assert(result != nullptr);
+        DCHECK(result) << "Cast failed for column: "
+                       << " (expected type: " << Type << ", actual type: " << value->get_name() << ")";
 #endif
         return RunTimeColumnType<Type>::static_pointer_cast(value);
     }
 
     template <LogicalType Type>
     static inline typename RunTimeColumnType<Type>::Ptr cast_to(ColumnPtr&& value) {
-#ifdef NDEBUG
-        down_cast<const RunTimeColumnType<Type>*>(value.get());
-#else
+#ifndef NDEBUG
         auto* result = dynamic_cast<const RunTimeColumnType<Type>*>(value.get());
-        if (result == nullptr) {
-            const auto* ptr = value.get();
-            std::cerr << "Cast failed for column: "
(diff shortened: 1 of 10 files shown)


be/src/column/column_helper.h
```diff
@@ -280,19 +280,48 @@ class ColumnHelper {
      */
     template <LogicalType Type>
     static inline typename RunTimeColumnType<Type>::Ptr cast_to(const ColumnPtr& value) {
+#ifdef NDEBUG
         down_cast<const RunTimeColumnType<Type>*>(value.get());
+#else
+        auto* result = dynamic_cast<const RunTimeColumnType<Type>*>(value.get());
+        if (result == nullptr) {
+            const auto* ptr = value.get();
+            std::cerr << "Cast failed for column: "
```
alvin-phoenix-ai: std::cerr is not a good option, why not just use CHECK() or LOG(FATAL) <<

be/src/column/column_helper.h
```diff
@@ -280,19 +280,48 @@ class ColumnHelper {
...
     static inline typename RunTimeColumnType<Type>::MutablePtr cast_to(MutableColumnPtr&& value) {
+#ifdef NDEBUG
         down_cast<const RunTimeColumnType<Type>*>(value.get());
+#else
+        auto* result = dynamic_cast<const RunTimeColumnType<Type>*>(value.get());
+        if (result == nullptr) {
+            std::cerr << "Cast failed for column: "
+                      << " (expected type: " << Type << ", actual type: " << value->get_name() << ")" << std::endl;
+        }
+        assert(result != nullptr);
```
stdpain: I think down_cast is enough

be/src/column/column_helper.h
murphyatwork: fixed

be/src/column/column_helper.h
murphyatwork: the assert in the down_cast would only report `From=Column To=FixedLengthColumn`, which you don't know the real type of `From` column, that's why I introduce another type check here

stdpain: We could consider modifying the implementation of down_cast directly when assert failed .

stdpain:
```
class Base {
public:
    virtual ~Base() = default;
};

class Derived : public Base {
public:
};

void printActualType(Base* obj) {
    const std::type_info& typeInfo = typeid(*obj);

    int status;
    std::unique_ptr<char, void(*)(void*)> demangledName(
        abi::__cxa_demangle(typeInfo.name(), nullptr, nullptr, &status),
        std::free
    );

    if (status == 0) {
        std::cout << "Actual type: " << demangledName.get() << std::endl;
    } else {
        std::cout << "Actual type: " << typeInfo.name() << " (demangling failed)" << std::endl;
    }
}
```

murphyatwork: yeah, that's a good idea, but not conflicted.
you can get the basic type info with `typeid`, but you can get more debug information if you already know it's a `Column`

satanson (approved):

Seaven (approved):

PR #61594, opened 2025-08-05
[Tool] share build result between build and ut

murphyatwork:
## What I'm doing:

**Previously**, we relied on two separate scripts for building the code:
- `build.sh`: Responsible for compiling the `starrocks_be` executable.
- `run-be-ut.sh`: Handles the compilation of the `starrocks_test` executable and executes unit tests.

Since these scripts operate independently and do not share resources, the code must be compiled twice, significantly increasing the development cycle time.

**An option** has been added to the build process to allow sharing of the build state:
- `build.sh --with-ut` to build `starrocks_be` along with the unit tests. The output will be located at `be/build_DEBUG_with_ut/`.
- `run-be-ut.sh` to build and execute the tests, with the results also stored at `be/build_DEBUG_with_ut/`.

By enabling shared build states, the code no longer needs to be compiled twice.

**Why** dont we create a single script to build all code? 
- The `MAKE_TEST` variable differentiates between `starrocks_be` and `starrocks_test`, as they utilize distinct linker and compiler options.
- Ideally we can remove this variable and make them use the exactly same compiling options

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [x] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -95,6 +95,7 @@ Usage: $0 <options>
      --without-java-ext build Backend without java-extensions(default with java-extensions)
      --without-starcache
                         build Backend without starcache library
+     --with-ut          build Backend with unit tests in the same build directory
      -j                 build Backend parallel
      --output-compile-time 
                         save a list of the compile time for every C++ file in ${ROOT}/compile_times.txt.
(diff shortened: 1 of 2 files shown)


kevincai (changes requested):
this will not work. There is a `-DMAKE_TEST` which will define `BE_TEST` in the code base.

Regular `starrocks_be` shall not be generated with this macro definition, otherwise a lot functions will be broken.

murphyatwork:
> this will not work. There is a `-DMAKE_TEST` which will define `BE_TEST` in the code base.
> 
> Regular `starrocks_be` shall not be generated with this macro definition, otherwise a lot functions will be broken.

This codebase is incredibly frustrating.

kevincai:
@murphyatwork if BE_TEST can be removed completely from be/src codebase, your proposal will work. lol

PR #61621, opened 2025-08-05
[Enhancement] Add header setting support to http_client

Nboxff:
## Why I'm doing:

Currently, the HTTPClient class lacks the ability to set custom headers, which limits its functionality for various API interactions such as calling large language models' service.

## What I'm doing:

add methods in http_client.h and http_client.cpp

```
// Set HTTP header
void set_header(const std::string& key, const std::string& value);

// Set multiple headers at once
void set_headers(const std::unordered_map<std::string, std::string>& headers);

// Clear all headers
void clear_headers();

// Set Authorization header with Bearer token (useful for LLM APIs)
void set_bearer_token(const std::string& token) { set_header("Authorization", "Bearer " + token); }
```
## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


be/src/http/http_client.h
```diff
@@ -133,12 +137,18 @@ class HttpClient {
...
     const HttpCallback* _callback = nullptr;
     char _error_buf[CURL_ERROR_SIZE];
     curl_slist* _header_list = nullptr;
+
+    // Store headers for easy management
+    std::unordered_map<std::string, std::string> _headers;
 };
 
-} // namespace starrocks
+} // namespace starrocks
```
kevincai: properly setup your IDE to avoid introducing this unnecessary EOL change.

Nboxff: @kevincai Thanks for catching this! I have fixed the unnecessary EOL change.

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@Nboxff could you sign-off it by `git commit -s` to make the DCO check pass

be/src/http/http_client.cpp
```diff
@@ -138,6 +138,44 @@ void HttpClient::set_method(HttpMethod method) {
     }
 }
 
+void HttpClient::set_header(const std::string& key, const std::string& value) {
+    _headers[key] = value;
```
kevincai:
```suggestion
    _headers.insert_or_assign(key, value);
```

be/src/http/http_client.cpp
```diff
@@ -138,6 +138,44 @@ void HttpClient::set_method(HttpMethod method) {
...
+    _apply_headers();
+}
+
+void HttpClient::clear_headers() {
+    _headers.clear();
+    if (_header_list != nullptr) {
+        curl_slist_free_all(_header_list);
+        _header_list = nullptr;
+    }
+    _apply_headers();
```
kevincai: why apply headers here when doing clear_headers? if need to clear curl headers, the _header_list free op is taken care in _apply_headers, could be just `_headers.clear()` and then `_apply_headers()`?

be/test/http/http_client_test.cpp
```diff
@@ -69,8 +69,42 @@ class HttpClientTestSimplePostHandler : public HttpHandler {
...
+    void handle(HttpRequest* req) override {
+        // Check the custom test header set by client
+        const auto& test_header = req->header("X-Custom-Header");
+        if (!test_header.empty()) {
+            HttpChannel::send_reply(req, test_header);
+            return;
+        }
+
+        // Check authentication header type
+        const auto& auth_header = req->header(HttpHeaders::AUTHORIZATION);
```
kevincai, line 83: how about echo the auth header in response body, let client to check if the response body matches its auth header?

be/src/http/http_client.cpp
Nboxff: Thanks for the thorough review!I’ve replaced `_headers[key] = value` with `insert_or_assigrn()` as suggested in latest commit.

be/src/http/http_client.cpp
Nboxff: Thanks for the thorough review! `_apply_headers()` is unnecessary and I have removed it in the latest commit.

be/test/http/http_client_test.cpp
Nboxff: Thanks! I have updated tests to echo Authorization header in response body and update the testcases.

kevincai (dismissed):

Nboxff:
> @Nboxff could you sign-off it by `git commit -s` to make the DCO check pass你能通过 `git commit -s` 签字，让 DCO 检查通过吗

Thanks for the reminder! I've signed off all commits and the DCO check should now pass.

kevincai (dismissed):

be/src/http/http_client.cpp
```diff
@@ -168,6 +205,10 @@ Status HttpClient::execute_delete_request(const std::string& payload, std::strin
 
 Status HttpClient::execute(const std::function<bool(const void* data, size_t length)>& callback) {
     _callback = &callback;
+
+    // Apply headers before executing
+    _apply_headers();
```
alvin-phoenix-ai: _apply_headers is called when set_headers is executed every time. So there is no need to call _apply_headers in this place.

And init() will be called when call execute, so _headers should be reset in init function to be consistent with _header_list.

Nboxff: Thank you for your review. I have reset `_headers` in init function and removed `_apply_headers();` in `execute` method. 
To ensure that the execute method can run correctly after removing the _apply_headers() method call, I have added the following statement to the clear_headers method:
```cpp
curl_easy_setopt(_curl, CURLOPT_HTTPHEADER, _header_list);
```

alvin-phoenix-ai (approved):

alvin-phoenix-ai:
@Nboxff Because v3.5 is released, and we don't need to cherry-pick this enhancement to it.

kevincai (approved):

PR #61747, opened 2025-08-09
[Enhancement] Support standard bitmap32 without type code in bitmap_from_binary function

2k124:
## Why I'm doing:
StarRocks cannot parse standard RoaringBitmap binary data during queries due to type encapsulation mismatch.

## What I'm doing:
Extend the RoaringBitmap type and implement read logic for correct parsing.

Reference standard:  https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java (In the serialization, BITMAP32 out.writeInt(Integer.reverseBytes("12346"),first char is 58)

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/types/bitmap_value.cpp
+++ b/be/src/types/bitmap_value.cpp
@@ -751,8 +751,8 @@ bool BitmapValue::valid_and_deserialize(const char* src, size_t max_bytes) {
         _type = EMPTY;
         return true;
     }
-
-    if (*src < BitmapTypeCode::EMPTY || *src > BitmapTypeCode::BITMAP64_SERIV2) {
+    if ((*src < BitmapTypeCode::EMPTY || *src > BitmapTypeCode::BITMAP64_SERIV2) &&
+        *src != BitmapTypeCode::STANDARD_BITMAP32 && *src != BitmapTypeCode::STANDARD_BITMAP32_HAS_RUN) {
         return false;
     } else {
         bool valid = true;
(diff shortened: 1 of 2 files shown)


alvin-phoenix-ai:
@2k124 
This is a big change. Previously, StarRocks encode the bitmap type with its own format, which is not compatible with RoaringBitmap. So bitmap_from_binary means to transform a starrocks encoded binary to a bitmap.
If you are going to make this function compatible with roaringbitmap encoded binary, it will influence the behavior of this function. 
So you should create a GitHub issue ticket and explain why this change makes sense and has no influence on the current users.

kangkaisen:
you should add more test cases

eshishki:
@2k124 still interested? alvin-phoenix-ai asked back in August whether this should be a new function vs changing `bitmap_from_binary`'s contract (existing callers rely on the StarRocks-encoded format), and kangkaisen asked for more test cases. Neither got a follow-up, behavior-check is still red, and there are no reviews. If you're not picking it back up, could you close it?

PR #61749, opened 2025-08-09
[Feature] Remove table path after clear table

2k124:
## Why I'm doing:
Currently, after executing a clear table, the system does not physically delete the underlying files or directories. This leads to an excessive number of subdirectories in real-business scenarios, which impacts metadata read performance in distributed storage systems

## What I'm doing:
I'm implementing physical deletion logic for files and directories after a  clear table operation. This ensures that all associated data and directory structures are removed, preventing unnecessary metadata accumulation and improving distributed storage performance.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/service/service_be/lake_service.cpp
+++ b/be/src/service/service_be/lake_service.cpp
@@ -829,6 +829,53 @@ void LakeServiceImpl::drop_table(::google::protobuf::RpcController* controller,
     latch.wait();
 }
 
+void LakeServiceImpl::remove_table_path(::google::protobuf::RpcController* controller,
+                                        const ::starrocks::RemoveTablePathRequest* request,
+                                        ::starrocks::RemoveTablePathResponse* response,
+                                        ::google::protobuf::Closure* done) {
+    brpc::ClosureGuard guard(done);
+    auto cntl = static_cast<brpc::Controller*>(controller);
+
+    if (!request->has_path()) {
+        cntl->SetFailed("missing path");
+        return;
+    }
+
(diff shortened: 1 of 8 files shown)


be/src/service/service_be/lake_service.cpp
```diff
@@ -829,6 +829,53 @@ void LakeServiceImpl::drop_table(::google::protobuf::RpcController* controller,
...
+
+    LOG(INFO) << "Received remove table path request. path=" << request->path()
+              << " queued_tasks=" << thread_pool->num_queued_tasks();
+
+    auto latch = BThreadCountDownLatch(1);
+    auto task = [&]() {
+        TEST_SYNC_POINT("LakeService::remove_table_path:task_run");
+        DeferOp defer([&]() { latch.count_down(); });
+
+        auto st = fs::remove_all(request->path());
```
kevincai, line 860: This will not work. the request->path() from `table.getDefaultFilePathInfo().getFullPath()` is only a physical path from remote storage, without any credential info, fs::remove_all() will be most likely fail for EACESS.

kevincai (changes requested):

kevincai:
close this PR as it won't work at all.

PR #61753, opened 2025-08-10
[BugFix] fix cast string to number bug

2k124:
## Why I'm doing:
In StarRocks, when a table is list-partitioned on a STRING-typed column and a query uses a numeric literal predicate (e.g., WHERE p = 1), the Frontend (FE) attempts to implicitly cast all string-based partition values to the target numeric type (e.g., BIGINT or DECIMAL) during list-based partition pruning.
However, if any partition value cannot be safely converted to a number—such as 'a', 'xyz', or '1.2.3'—the cast operation throws a runtime exception (e.g., NumberFormatException). This causes the FE to fail during query planning instead of gracefully excluding the incompatible partitions.
This behavior breaks query robustness, especially in real-world scenarios where partition values may contain non-numeric strings due to flexible ingestion pipelines or data quality issues. According to standard SQL semantics, 'a' = 1 should evaluate to FALSE (not raise an error), and the query should proceed by pruning only valid matching partitions.

## What I'm doing:
I’m adding type-safe validation during partition pruning for string-typed list partitions when the query predicate uses a numeric literal. Specifically:
Before casting a string partition value to a numeric type, we now validate whether the value is convertible.
If conversion is not possible (e.g., CAST('a' AS DECIMAL) would fail), the partition is safely skipped—treated as non-matching.No exception is thrown, pruning continues for other partitions.
This ensures the query plan is generated successfully, and execution proceeds only on relevant, compatible partitions.

For example:
Given partitions with values '01', '1', '001', and 'a', and a query:

SELECT * FROM tb WHERE p = 1;

Only '01', '1', and '001' are considered (as they numerically equal 1), while 'a' is silently excluded—without throwing any exception.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Prevents failures during list-partition pruning when casting string partition keys to numeric types.
> 
> - In `ListPartitionPruner#getCastPartitionValueMap`, skip entries where `key` is a `StringLiteral` and not numeric when target `CastOperator` type is numeric (uses `NumberUtils.isNumber`) to avoid cast exceptions
> - Adds `testExternalTableBinaryPredicate5` to validate pruning with mixed string values (`'1','2','ab'`) when casting to `INT`
> - Minor imports added for `StringLiteral` and `NumberUtils`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 52dd67b90e64e4fb921a860b7deaca84bf802a10. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


stephen-shelby:
hi, thanks for your patch. could you provide a complete reproduce case？

caneGuy (commented):
add  an example in the description

2k124:
> hi, thanks for your patch. could you provide a complete reproduce case？

ok

dirtysalt (dismissed):

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-celerdata:
@cursor review

alvin-celerdata:
@codex review

miomiocat (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/ListPartitionPruner.java
```diff
@@ -551,6 +553,11 @@ private ConcurrentNavigableMap<LiteralExpr, Set<Long>> getCastPartitionValueMap(
 
         for (Map.Entry<LiteralExpr, Set<Long>> entry : partitionValueMap.entrySet()) {
             LiteralExpr key = entry.getKey();
+            if (castOperator.getType().isNumericType() && key instanceof StringLiteral &&
+                    !NumberUtils.isNumber(key.getStringValue())) {
```
stephen-shelby: You can reuse or adapt the logic of `castLiteralExpr`.
Also, I have a question: why would an int type partition have partition values like "ab"?

2k124: Thanks for the review!
To clarify the scenario:
When the partition column is of STRING type, but the query predicate uses a numeric literal (e.g., WHERE p = 1), StarRocks performs an implicit cast from the string partition values to the numeric type during list-based partition pruning. This is needed to build a correct mapping for pruning.

For example:
query: select * from tb where p=1
eg. string partition value cast to int
// string_col = '01'  → 1
// string_col = '1'   → 2
// string_col = '001' → 3
// will generate new partition value map
// int_col = 1  [1, 2, 3]

 if a partition value like 'a' exists, the cast (CAST('a' AS DECIMAL)) throws an exception and crashes the FE during planning.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/ListPartitionPruner.java
```diff
@@ -551,6 +553,11 @@ private ConcurrentNavigableMap<LiteralExpr, Set<Long>> getCastPartitionValueMap(
...
+                // Partition values (p='1','2','a'), select * from tb where p=1, cast(p as decimal). An error occurs when the value is 'a'.
+                continue;
+            }
             LiteralExpr literalExpr = castLiteralExpr(key, castOperator.getType());
             Set<Long> partitions = newPartitionValueMap.computeIfAbsent(literalExpr, k -> Sets.newHashSet());
```
stephen-shelby, line 563: Could you check the AI comments above?

2k124: Fixed.

PR #61756, opened 2025-08-11
[Enhancement]support bucket-aware execution for iceberg

zombee0:
## What I'm doing:

Fixes #issue
#61287 

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/BucketProperty.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/BucketProperty.java
@@ -15,10 +15,15 @@
 package com.starrocks.connector;
 
 import com.starrocks.catalog.Column;
+import com.starrocks.sql.common.StarRocksPlannerException;
 import com.starrocks.thrift.TBucketFunction;
 import com.starrocks.thrift.TBucketProperty;
 
+import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
+
+import static com.starrocks.sql.common.ErrorType.INTERNAL_ERROR;
 
 public class BucketProperty {
     private final TBucketFunction bucketFunction;
@@ -56,6 +61,28 @@ public class BucketProperty {
         tBucketProperty.setBucket_func(bucketFunction);
         tBucketProperty.setBucket_num(bucketNum);
         return tBucketProperty;
+    }
+
+    public static Optional<List<BucketProperty>> checkAndGetBucketProperties(
+            List<List<BucketProperty>> bucketProperties) {
+        if (bucketProperties.isEmpty()) {
+            return Optional.empty();
+        }
+        List<BucketProperty> bp0 = bucketProperties.get(0);
+        for (int i = 1; i < bucketProperties.size(); i++) {
+            List<BucketProperty> bp = bucketProperties.get(i);
+            boolean pass = bp.size() == bp0.size();
+            for (int j = 0; j < bp0.size(); j++) {
+                if (!bp.get(j).satisfy(bp0.get(j))) {
+                    pass = false;
+                    break;
+                }
+            }
+            if (!pass) {
+                throw new StarRocksPlannerException("Error when using bucket-aware execution", INTERNAL_ERROR);
+            }
+        }
+        return Optional.of(bucketProperties.get(0));
     }
 
     @Override
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/RemoveUselessScanOutputPropertyRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/RemoveUselessScanOutputPropertyRule.java
@@ -1,0 +1,76 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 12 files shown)


wangsimo0:
ignore merge 4.0.0-rc01

wangsimo0:

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/RemoveUselessScanOutputPropertyRule.java
```diff
@@ -0,0 +1,76 @@
...
+// limitations under the License.
+
+package com.starrocks.sql.optimizer.rule.tree;
+
+import com.starrocks.sql.optimizer.OptExpression;
+import com.starrocks.sql.optimizer.OptExpressionVisitor;
+import com.starrocks.sql.optimizer.base.PhysicalPropertySet;
+import com.starrocks.sql.optimizer.task.TaskContext;
+
+public class RemoveUselessScanOutputPropertyRule implements TreeRewriteRule {
```
Seaven, line 22: add comment

zombee0: This rule is used to remove Scan Node Output Properties requested by the `RequiredPropertyDriver` but ultimately unused. If there are no nodes between the Exchange and the Scan Node requesting these properties, Optional Output Properties on the Scan Node are unused.
For example, if both the left and right tables are bucket column partitioned tables, but the number of buckets differs, a bucket shuffle join is required. And then the upstream node of the Scan Node for the right table is an Exchange, the Output Property on the right table is unused and should be removed.

zombee0: I will add this comment with other changes or next pr.

Youngwb: why need to remove the Output Property of scan node? what will happend if you do not remove?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/RemoveUselessScanOutputPropertyRule.java
```diff
@@ -0,0 +1,76 @@
...
+            } else {
+                for (OptExpression opt : optExpression.getInputs()) {
+                    opt.getOp().accept(this, opt, context);
+                }
+            }
+            return null;
+        }
+
+        @Override
+        public Void visitPhysicalHashAggregate(OptExpression optExpression, Boolean context) {
```
Youngwb, line 48: please give context a more meaningful name

fe/fe-core/src/main/java/com/starrocks/connector/BucketProperty.java
```diff
@@ -58,6 +63,28 @@ public TBucketProperty toThrift() {
...
+        for (int i = 1; i < bucketProperties.size(); i++) {
+            List<BucketProperty> bp = bucketProperties.get(i);
+            boolean pass = bp.size() == bp0.size();
+            for (int j = 0; j < bp0.size(); j++) {
+                if (!bp.get(j).satisfy(bp0.get(j))) {
+                    pass = false;
+                    break;
+                }
+            }
+            if (!pass) {
```
Youngwb, line 81: how could  this happen? colocate join plan should check satisfy in optimizer, not in planfragment builder

zombee0: This means something wrong with bucket-aware execution, in normal case, it's impossible to access here

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/RemoveUselessScanOutputPropertyRule.java
zombee0: If we don't remove Output Property, scan ranges will be scheduled by bucket aware scheduler, even there is no need to do that, and that may lead to more disk usage for cache and cache miss.

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/RemoveUselessScanOutputPropertyRule.java
zombee0: If we don't remove Output Property, scan ranges will be scheduled by bucket aware scheduler, even there is no need to do that, and that may lead to more disk usage for cache and cache miss.

fe/fe-core/src/main/java/com/starrocks/connector/BucketProperty.java
```diff
@@ -58,6 +63,28 @@ public TBucketProperty toThrift() {
...
+                throw new StarRocksPlannerException("Error when using bucket-aware execution", INTERNAL_ERROR);
+            }
+        }
```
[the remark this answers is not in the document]
zombee0: Good catch, even this is impossible in normal case, but I will fixed it in next pr.

dirtysalt (approved):

Seaven (approved):

Youngwb (approved):

zombee0:
@Mergify backport branch-4.0

PR #61795, opened 2025-08-11
[BugFix](starrocks-python-client) Correct reflection of server_default and generated columns

### LINKED ISSUE #61793: starrocks-python-client: ArgumentError when reflecting tables with server_default columns
**Summary:**

Reflecting a table with a `server_default` column (e.g., `DEFAULT CURRENT_TIMESTAMP`) using `MetaData.reflect()` or Alembic's `autogenerate` fails with `sqlalchemy.exc.ArgumentError: A generated column cannot specify a server_default or a server_onupdate argument`.

This appears to be a regression in the new `information_schema`-based reflection logic, which incorrectly flags columns with a `COLUMN_DEFAULT` as also being `computed`.

1.  **Create this table in StarRocks:**
    ```sql
    CREATE TABLE `repro_bug_table` (
      `id` INT,
      `created_at` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
    ) PRIMARY KEY(`id`) DISTRIBUTED BY HASH(`id`);
    ```

2.  **Run this Python script:**
    ```python
    from sqlalchemy import create_engine, MetaData
    
    # Replace with your StarRocks connection string
    engine = create_engine("starrocks://root:@127.0.0.1:9030/test")
    
    metadata = MetaData()
    metadata.reflect(bind=engine, only=['repro_bug_table']) # This line will fail
    
    print("Success! (Bug not reproduced)")
    ```

The Python script should run without errors and print "Success!".

### Real behavior (Required)

The script fails with the following traceback:
```
Traceback (most recent call last):
  ...
sqlalchemy.exc.ArgumentError: A generated column cannot specify a server_default or a server_onupdate argument
```

3.5.2-69de616

janlink:
## Why I'm doing:
The current SQLAlchemy dialect for StarRocks has a bug in its reflection logic when using information_schema. It incorrectly identifies columns with a server_default (like DEFAULT CURRENT_TIMESTAMP) as "computed" columns.
This causes SQLAlchemy to raise an ArgumentError: A generated column cannot specify a server_default..., which breaks schema reflection tools like MetaData.reflect() and Alembic's autogenerate feature, making automated schema management for such tables impossible.

## What I'm doing:
This PR fixes the reflection logic in starrocks/reflection.py to correctly distinguish between true generated columns and columns that only have a server-side default value.

Fixes #61793

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/contrib/starrocks-python-client/starrocks/reflection.py
+++ b/contrib/starrocks-python-client/starrocks/reflection.py
@@ -107,17 +107,22 @@ class StarRocksTableDefinitionParser(object):
         Parse column from information_schema.columns table.
         It returns dictionary with column informations expected by sqlalchemy.
         """
-        return {
+        col_data = {
             "name": column["COLUMN_NAME"],
             "type": self._parse_column_type(column=column),
             "nullable": column["IS_NULLABLE"] == "YES",
             "default": column["COLUMN_DEFAULT"],
             "autoincrement": None,  # TODO: This is not specified
-            "computed": {
-                "sqltext": column["GENERATION_EXPRESSION"]
-            },
             "comment": column["COLUMN_COMMENT"],
         }
+
+        generation_expression = column.get("GENERATION_EXPRESSION")
+        if generation_expression:
+            col_data["computed"] = {
+                "sqltext": generation_expression
+            }
+
+        return col_data
 
     def _get_key_columns(self, columns: list[_DecodingRow]) -> list[str]:
         """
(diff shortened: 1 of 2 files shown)


Astralidea (approved):

jaogoy:
@janlink Your fix is correct.
And, as I looked into the code, I saw that it's because the process way of `computed` in func `get_column_specification`:
```py
if column.computed is not None:
        colspec += " " + self.process(column.computed)  
```
Then, it will always append such a non-existent `computed` into a column's specification if the `computed` field is always not None.

jaogoy (approved):

janlink:
@jaogoy How to merge this PR? Is there anything left to do?

jaogoy:
@janlink I think PRs for the starrocks-python-client doesn't need to be merge into other branches than main, because the release is not related to the StarRocks core release plan.
You can abandon the backport prs.

PR #61799, opened 2025-08-11
[Enhancement] Emit routine load lag time

coderfreestyle:
## Why I'm doing:

Measuring the lag in time for routine Kafka loads is essential for real-time and near–real-time analytics at scale. Existing metrics may track lag in terms of Kafka offsets or rows, but they don’t reflect how far behind the system is in actual wall-clock time. Without this metric, operators and developers cannot easily assess whether data ingestion is keeping up with data production, identify performance bottlenecks, or promptly detect incidents causing ingest delays. This metric addresses an important gap in observability for production data pipelines using StarRocks routine load.

## What I'm doing:

This PR introduces a new time-based metric that measures the end-to-end lag of StarRocks’ routine load jobs from Kafka. The metric is calculated by comparing the timestamp of the latest processed Kafka record against the current system time, providing a real-time view of the ingestion delay. This value is exported as an observable metric, allowing users to monitor it through StarRocks’ metrics systems. With this addition, users can track how stale the ingested data is and respond quickly to unexpected increases in load lag.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -2483,7 +2483,7 @@ public class Config extends ConfigBase {
      * Whether to collect routine load latency metrics.
      */
     @ConfField(mutable = true)
-    public static boolean enable_routine_load_lag_time_metrics = true;
+    public static boolean enable_routine_load_lag_time_metrics = false;
 
     @ConfField(mutable = true)
     public static boolean enable_collect_query_detail_info = false;
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -2479,6 +2479,12 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true)
     public static boolean enable_routine_load_lag_metrics = false;
 
+    /**
+     * Whether to collect routine load latency metrics.
+     */
+    @ConfField(mutable = true)
+    public static boolean enable_routine_load_lag_time_metrics = true;
+
     @ConfField(mutable = true)
     public static boolean enable_collect_query_detail_info = false;
 
--- a/fe/fe-core/src/main/java/com/starrocks/load/routineload/KafkaRoutineLoadJob.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/routineload/KafkaRoutineLoadJob.java
@@ -502,9 +502,7 @@ public class KafkaRoutineLoadJob extends RoutineLoadJob {
                 Long.valueOf((totalRows - errorRows - unselectedRows) * 1000 / totalTaskExcutionTimeMs));
         summary.put("committedTaskNum", Long.valueOf(committedTaskNum));
         summary.put("abortedTaskNum", Long.valueOf(abortedTaskNum));
-        if (Config.enable_routine_load_lag_time_metrics) {
-            summary.put("routineLoadLagTime", getRoutineLoadLagTime());
-        }
+        summary.put("routineLoadLagTime", getRoutineLoadLagTime());
         Gson gson = new GsonBuilder().disableHtmlEscaping().create();
         return gson.toJson(summary);
     }
--- a/fe/fe-core/src/main/java/com/starrocks/load/routineload/KafkaRoutineLoadJob.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/routineload/KafkaRoutineLoadJob.java
@@ -68,6 +68,7 @@ import com.starrocks.common.util.concurrent.lock.LockType;
 import com.starrocks.common.util.concurrent.lock.Locker;
 import com.starrocks.load.Load;
 import com.starrocks.load.RoutineLoadDesc;
+import com.starrocks.metric.RoutineLoadLagTimeMetricMgr;
 import com.starrocks.qe.OriginStatement;
 import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.server.RunMode;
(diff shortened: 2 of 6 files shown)


fe/fe-core/src/main/java/com/starrocks/metric/RoutineLoadLagTimeMetricMgr.java
```diff
@@ -0,0 +1,238 @@
...
+                RoutineLoadLagTimeMetric lagTimeMetric = entry.getValue();
+                
+                // Skip metrics that don't have data yet
+                if (!lagTimeMetric.hasData()) {
+                    LOG.debug("Skipping job {} - no data available yet", jobKey);
+                    continue;
+                }
+                
+                // Remove stale data
+                if (lagTimeMetric.isStale(now, staleThresholdMs)) {
```
banmoy, line 182: could move the cleanup to MetricCalculator, and run it backgroud

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -2479,6 +2479,12 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true)
     public static boolean enable_routine_load_lag_metrics = false;
 
+    /**
+     * Whether to collect routine load latency metrics.
+     */
+    @ConfField(mutable = true)
+    public static boolean enable_routine_load_lag_time_metrics = true;
```
banmoy: default should be false

fe/fe-core/src/main/java/com/starrocks/metric/RoutineLoadLagTimeMetricMgr.java
```diff
@@ -0,0 +1,238 @@
...
+ * Manager for Routine Load Lag Time Metrics
+ * Handles collection and emission of lag metrics for Kafka routine load jobs
+ */
+public class RoutineLoadLagTimeMetricMgr {
+    private static final Logger LOG = LogManager.getLogger(RoutineLoadLagTimeMetricMgr.class);
+    
+    private static volatile RoutineLoadLagTimeMetricMgr instance;
+    
+    // Use composite key (dbId_jobName) to uniquely identify jobs
+    private final Map<String, RoutineLoadLagTimeMetric> jobLagTimeMap = Maps.newConcurrentMap();
```
banmoy, line 41: KafkaRoutineLoadJob could maintain the RoutineLoadLagTimeMetric itself. A manager is unnecessary

fe/fe-core/src/main/java/com/starrocks/load/routineload/KafkaRoutineLoadJob.java
```diff
@@ -501,6 +502,9 @@ protected String getStatistic() {
                 Long.valueOf((totalRows - errorRows - unselectedRows) * 1000 / totalTaskExcutionTimeMs));
         summary.put("committedTaskNum", Long.valueOf(committedTaskNum));
         summary.put("abortedTaskNum", Long.valueOf(abortedTaskNum));
+        if (Config.enable_routine_load_lag_time_metrics) {
```
banmoy: could always put it in the summary? The lag could be calculated at once no matter if the metric is enabled

fe/fe-core/src/main/java/com/starrocks/metric/RoutineLoadLagTimeMetricMgr.java
```diff
@@ -0,0 +1,238 @@
...
+            this.updateTimestamp = newUpdateTimestamp;
+            this.partitionMetrics.clear();
+
+            // Calculate max lag time from partition lag times
+            long newMaxLagTime = newPartitionLagTimes.values().stream()
+                    .mapToLong(Long::longValue)
+                    .max()
+                    .orElse(0L);
+
+            for (Map.Entry<Integer, Long> entry : newPartitionLagTimes.entrySet()) {
```
banmoy, line 91: We could only expose the max lag metric, and get partiton-level lag by `SHOW ROUTINE LOAD`. After you find the max lag metric is high, then check the partition-level lags by show command. The consideration is if there are many partitions, it will increase the cost of metrics, but most of them are useless.

fe/fe-core/src/main/java/com/starrocks/metric/RoutineLoadLagTimeMetricMgr.java
coderfreestyle: KafkaRoutineLoadJob could but the manager is here to provide the simple interface to update and emit the metrics without exposing the details. The intention behind this is to keep the single responsibility

fe/fe-core/src/main/java/com/starrocks/load/routineload/KafkaRoutineLoadJob.java
coderfreestyle: will update.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
coderfreestyle: will update

fe/fe-core/src/main/java/com/starrocks/metric/RoutineLoadLagTimeMetricMgr.java
coderfreestyle: Thanks for the call out and it makes sense, although I can also think of scenarios when the trend of partition-level lag metrics can be helpful for debugging. This is a trade off between observability and efficiency, and will need data, e.g., how many partitions do clients usually have and what's their use patterns of this metrics to debug, to make the decision. At this time I can make the change as there's the lack of data and assume the suggested pattern  - see max lag first then use `SHOW ROUTINE LOAD` to further dive deep - is sufficient enough.

fe/fe-core/src/main/java/com/starrocks/metric/RoutineLoadLagTimeMetricMgr.java
coderfreestyle: sg.

coderfreestyle:
Closing this PR in favor of https://github.com/StarRocks/starrocks/pull/62048

PR #61854, opened 2025-08-12
[Enhancement] Separate path id from physical partition id

xiangguangyxg:
## Why I'm doing:
Physical partition id is used as path id to construct storage path in shared-data mode.
In tablet splitting and merging, a new physical partition will replace the old physical partition,
and the new physical partition will share the same path id with the old one.
So we need to separate path id from physical partition id.

## What I'm doing:
Introduce path id and separate it from physical partition id for physical partition.
Path id is the same as physical partition id when a new physical partition is created, 
but after tablet splitting or merging, path id is no longer the same as physical partition id.

Note that path id is different from physical partition id only after a tablet splitting.
Tablet splitting does not support downgrading cluster to an unsupported version.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
@@ -60,6 +60,14 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
     @SerializedName(value = "shardGroupId")
     private long shardGroupId = INVALID_SHARD_GROUP_ID;
 
+    // Path id is only for shared-data mode, used to construct storage path.
+    // Default value is physical partition id, but in tablet splitting or merging,
+    // a new physical partition will replace the old physical partition,
+    // the new physical partition will share the same path id with the old one,
+    // and the path id is no longer the same as physical partition id.
+    @SerializedName(value = "pathId")
+    private long pathId;
+
     /* Physical Partition Member */
     @SerializedName(value = "isImmutable")
     private AtomicBoolean isImmutable = new AtomicBoolean(false);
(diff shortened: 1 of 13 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
```diff
@@ -60,6 +60,14 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
     @SerializedName(value = "shardGroupId")
     private long shardGroupId = INVALID_SHARD_GROUP_ID;
 
+    // Path id is only for shared-data mode, used to construct storage path.
```
kevincai, line 63: didn't see new physical partition creation with the new path id, suppose it is included in other PRs, right?

kevincai (dismissed):

fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
xiangguangyxg: Yes, I updated the code to add some comments.

kevincai (approved):

fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
```diff
@@ -60,6 +60,14 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
...
+    // Default value is physical partition id, but in tablet splitting or merging,
+    // a new physical partition will replace the old physical partition,
+    // the new physical partition will share the same path id with the old one,
+    // and the path id is no longer the same as physical partition id.
+    @SerializedName(value = "pathId")
+    private long pathId;
```
wyb, line 69:
```suggestion
    private long pathId = 0L;
```

wyb (approved):

gengjun-git (approved):

Seaven (approved):

PR #61856, opened 2025-08-12
[Feature] Support call procedure statement

Youngwb:
## Why I'm doing:
#61855

## What I'm doing:
This pull request introduces a framework for supporting procedures within the Iceberg connector, including infrastructure for registering, finding, and executing procedures, as well as a sample `register_table` procedure. The changes span the connector, metadata, and procedure registry, and also add integration points for invoking procedures via SQL statements.

**SQL Statement and Execution Integration**

* Extended the SQL analyzer and executor to support `CALL PROCEDURE` statements, including argument analysis and invocation through the new procedure framework.

**Iceberg Procedure Framework Integration**

* Added the `Procedure` class, which defines the structure for connector procedures, including argument specification and an `execute` method stub.
* Implemented `IcebergProcedureRegistry` for registering and retrieving procedures by name, and integrated it into `IcebergConnector` and `IcebergMetadata` to enable procedure lookup. 

**Procedure Registration and Example**

* Registered a sample `RegisterTableProcedure` in the Iceberg connector, which defines required arguments and a stub for execution logic. 

**Connector API Extension**

* Added the `getProcedure` method to the `ConnectorMetadata` interface and implemented it in `IcebergMetadata` to support procedure lookup by name. 

**Metadata Manager Support**

* Updated `MetadataMgr` to provide procedure lookup and execution capabilities, bridging SQL statements to connector procedures. 

These changes lay the groundwork for extensible, connector-specific procedures, starting with Iceberg, and provide the necessary plumbing for future procedure implementations.
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/connector/Procedure.java
```diff
@@ -0,0 +1,75 @@
...
+    public String getDatabaseName() {
+        return this.databaseName;
+    }
+
+    public String getProcedureName() {
+        return this.procedureName;
+    }
+
+    public void execute(ConnectContext context, Map<String, ConstantOperator> args) {
+        throw new UnsupportedOperationException("Procedure execution not implemented");
```
xxubai: Should we add more informations in the exception message?

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/procedure/RegisterTableProcedure.java
```diff
@@ -0,0 +1,55 @@
...
+                        new Argument(DATABASE_NAME, Type.VARCHAR, true),
+                        new Argument(TABLE_NAME, Type.VARCHAR, true),
+                        new Argument(TABLE_LOCATION, Type.VARCHAR, true)
+                )
+        );
+    }
+
+    @Override
+    public void execute(ConnectContext context, Map<String, ConstantOperator> args) {
+        // todo: Implement the logic to register an Iceberg table
```
xxubai, line 53: The execute implementation is not in this PR? 
Could you give a simple explanation of how execution works?

Samrose-Ahmed:
Good feature! Nice work.

fe/fe-core/src/main/java/com/starrocks/connector/Procedure.java
Youngwb: Actually, It should not run here, it's a hint to let developer to implements  this method.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/procedure/RegisterTableProcedure.java
Youngwb: Yes, this PR is focus on call statement, we can implement this procedure later.
You can get database_name, table_name, table_locatiion from args, then use iceberg table metadata parser to read snapshot json file and create the iceebrg table in hms/glue/hadoop/rest/jdbc catalog

fe/fe-core/src/main/java/com/starrocks/connector/Procedure.java
```diff
@@ -0,0 +1,75 @@
...
+package com.starrocks.connector;
+
+import com.starrocks.catalog.Type;
+import com.starrocks.qe.ConnectContext;
+import com.starrocks.sql.optimizer.operator.scalar.ConstantOperator;
+
+import java.util.List;
+import java.util.Map;
+
+public class Procedure {
```
zombee0: shall we add interface to check auth?

Youngwb: Yes, I think we need add different auth for different procedure, like need create tabler pressions  for register table procedure. 
I will add check permissions next PR which will implements the register table

Seaven (approved):

wangsimo0 (approved):

dirtysalt (approved):

PR #61899, opened 2025-08-13
[BugFix] avoid hold tablet shard lock to get compaction score

luohaha:
## Why I'm doing:
`tablet_ptr->updates()->get_compaction_score()` may be time-cost, avoid to call it while holding tablet shard lock.

## What I'm doing:
This pull request refactors the way tablets are iterated over in the `TabletManager::find_best_tablet_to_do_update_compaction` method to improve code clarity and potentially thread safety. Instead of directly locking and iterating over the shard's map, it now retrieves all tablets from the shard using a helper function.

**Refactoring and Code Simplification:**

* Replaced direct iteration and locking of `tablets_shard.tablet_map` with a call to `_get_all_tablets_from_shard`, which returns a vector of all tablets in the shard for iteration.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4

### DIFF
--- a/be/src/storage/tablet_manager.cpp
+++ b/be/src/storage/tablet_manager.cpp
@@ -838,8 +838,8 @@ TabletSharedPtr TabletManager::find_best_tablet_to_do_update_compaction(DataDir*
     int64_t highest_score = 0;
     TabletSharedPtr best_tablet;
     for (const auto& tablets_shard : _tablets_shards) {
-        std::shared_lock rlock(tablets_shard.lock);
-        for (const auto& [tablet_id, tablet_ptr] : tablets_shard.tablet_map) {
+        std::vector<TabletSharedPtr> all_tablets_by_shard = _get_all_tablets_from_shard(tablets_shard);
+        for (const auto& tablet_ptr : all_tablets_by_shard) {
             if (tablet_ptr->keys_type() != PRIMARY_KEYS) {
                 continue;
             }


kevincai (approved):

be/src/storage/tablet_manager.cpp
```diff
@@ -838,8 +838,8 @@ TabletSharedPtr TabletManager::find_best_tablet_to_do_update_compaction(DataDir*
     int64_t highest_score = 0;
     TabletSharedPtr best_tablet;
     for (const auto& tablets_shard : _tablets_shards) {
-        std::shared_lock rlock(tablets_shard.lock);
-        for (const auto& [tablet_id, tablet_ptr] : tablets_shard.tablet_map) {
+        std::vector<TabletSharedPtr> all_tablets_by_shard = _get_all_tablets_from_shard(tablets_shard);
```
kevincai, line 841: wondering if there is any other places that has the similar code structure and can be replaced with lock+snapshot+unlock+iterating

kevincai (approved):

be/src/storage/tablet_manager.cpp
luohaha: I had checked, this one is the most risky one.

wyb (approved):

luohaha:
https://github.com/Mergifyio backport branch-3.3

PR #61965, opened 2025-08-15
[Enhancement] create adaptive zonemap index for strings

### LINKED ISSUE #61960: Adaptive ZoneMap Index Creation for string type based on overlap quality
## Summary
Introduce an adaptive ZoneMap index creation mechanism in SegmentWriter that decides whether to persist page-level ZoneMaps based solely on page overlap structure, without relying on value spans, point-query fanout estimation, or runtime statistics. This makes the algorithm applicable to all comparable types, including strings, while remaining O(P) time complexity.

## Background
StarRocks automatically generates ZoneMap indexes for supported columns during segment flush. However, these indexes are only created under specific conditions:
- The column is a sort key, primary key, or duplicate key.
- The column has a fixed-length type, excluding string types.

As a result, non-key string columns do not automatically receive ZoneMap indexes. Nevertheless, there are scenarios where queries could benefit from having ZoneMap indexes for such columns. For example:
```sql
CREATE TABLE t1 (
  id bigint,
  data JSON,
  sort_key varbinary as encode_sort_key(
    get_json_string(data, 'operation'),
    get_json_string(data, 'name')
  )
)
ORDER BY (sort_key);
```
In this scenario, while `get_json_string(data, 'operation')` is not a KEY column, it exhibits a functional dependency on the `sort_key`, meaning it is inherently sorted. As a result, `data.operation` demonstrates strong clustering, which can effectively leverage the zonemap index.

Generally speaking, a string column can exhibit good clustering, even if it is not explicitly sorted.

## Goals
1.	Writer-only decision: Determine at flush time whether to persist ZoneMaps for a column in a segment.
2.	Type-agnostic: Works for numeric, datetime, and string types (lexicographic order).
3.	Low overhead: O(P) in number of pages, constant memory.
4.	No runtime dependency: Can be deployed immediately without query feedback.

## Intuition  
Greater overlap implies reduced clustering. Therefore, clustering can be quantified using the overlap ratio of zonemaps.

To calculate:  
- **s_k(i)**: Sum of overlaps with page(i) across all pages.  
- **overlap_ratio**: Total sum of `s_k` divided by the square of the number of pages: `num_pages * num_pages / 2`.  
- **Quality**: Defined as `1 - overlap_ratio`, where higher values indicate better clustering.

- If pages are perfectly separated: **Quality ≈ 1**.  
- If all pages overlap completely: **Quality = 0**.  

## Implementation

1. If `ColumnWriterOptions::need_zone_map` is set to false and the column is of string type, an adaptive strategy will be applied as follows:  
2. Within `StringColumnWriter::append`, the first `NUM_SAMPLES = 64` pages will be sampled. Using the specified algorithm, the necessity of building the zonemap index will be evaluated. If the index proves to be non-beneficial, the `ZoneMapIndexWriter` will be discarded, and the sampling process will be terminated.

## Others
- Long string columns are inefficient for constructing zonemap indexes due to their high overhead and minimal advantages. Currently, we prevent zonemap creation for such cases, with the potential to explore truncating long strings to a fixed length in the future.
- Numeric columns always generate zonemaps, regardless of their actual benefit. Introducing an adaptive quality measurement strategy could help optimize this process.

murphyatwork:
## Why I'm doing:
Currently, ZoneMap indexes for string types are always created, even when the data distribution leads to high overlap between page ranges. This can result in ineffective indexes that consume storage without providing significant query pruning benefits.

## What I'm doing:
This PR introduces adaptive ZoneMap index creation for string types (CHAR/VARCHAR) based on the overlap quality of page ranges.

- **New Configuration Flags**:
    - `string_zonemap_overlap_threshold` (default 0.8): If the estimated overlap ratio of consecutive page ranges exceeds this, the ZoneMap index for the column will be skipped.
    - `string_zonemap_min_pages_for_adaptive_check` (default 16): Minimum number of non-empty pages required to apply the adaptive check.
- **Adaptive Heuristic**: For CHAR/VARCHAR columns, the system now calculates the overlap ratio between consecutive page ZoneMaps. If this ratio is too high (above the configured threshold) and there are enough pages, the page-level ZoneMap index will not be written.

Fixes #61960

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---
<a href="https://cursor.com/background-agent?bcId=bc-e40a1ba3-294f-4381-be12-42d7b1a470e7">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">
    Open in Cursor
  </picture>
</a>
<a href="https://cursor.com/agents?id=bc-e40a1ba3-294f-4381-be12-42d7b1a470e7">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">
    Open in Web
  </picture>
</a>


satanson (approved):

dirtysalt (approved):

meegoo (approved):

be/src/storage/rowset/zone_map_index.cpp
```diff
@@ -411,4 +438,95 @@ size_t ZoneMapIndexReader::mem_usage() const {
...
+        // since null values can exist alongside any non-null values
+        if (zone_map.has_null || other.zone_map.has_null) {
+            return true;
+        }
+
+        // For non-null zones, check value range overlap
+        return (other.zone_map.min_value.value >= zone_map.min_value.value &&
+                other.zone_map.min_value.value <= zone_map.max_value.value) ||
+               (other.zone_map.max_value.value >= zone_map.min_value.value &&
+                other.zone_map.max_value.value <= zone_map.max_value.value);
```
wyb, line 499: this is incomplete, for example: `other_zonemap.min <= zonemap.min && other_zonemap.max >= zonemap.max` is also overlapped.

be/src/storage/rowset/column_writer.cpp
```diff
@@ -594,6 +595,22 @@ Status ScalarColumnWriter::_write_data_page(Page* page) {
 Status ScalarColumnWriter::finish_current_page() {
     if (_zone_map_index_builder != nullptr) {
         RETURN_IF_ERROR(_zone_map_index_builder->flush());
+        if (_zone_map_index_quality_judger != nullptr) {
+            std::optional<ZoneMapPB> last_zonemap = _zone_map_index_builder->get_last_zonemap();
+            if (last_zonemap.has_value()) {
+                _zone_map_index_quality_judger->feed(last_zonemap.value());
+            }
+            CreateIndexDecision decision = _zone_map_index_quality_judger->make_decision();
```
wyb, line 603: `make decision` iterates all previous zonemaps. 
executing it on every finished page will cause performance issue?

PR #62000, opened 2025-08-17
[BugFix] Avoid double precission issue while displaying resource group

ikishore:
## Why I'm doing:
Precission issue while dispaying resource group info, ex: when mem limit is set to 0.55, 0.54, 0.56 etc

## What I'm doing:
Using Big decimal API to resolve precission issue and to match with create resource command. 

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/ResourceGroup.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/ResourceGroup.java
@@ -22,6 +22,8 @@ import com.starrocks.sql.analyzer.SemanticException;
 import com.starrocks.thrift.TWorkGroup;
 import com.starrocks.thrift.TWorkGroupType;
 
+import java.math.BigDecimal;
+import java.math.RoundingMode;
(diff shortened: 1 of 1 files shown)


eshishki:
@ikishore the fix looks right for `mem_limit`. The same double-precision pattern is present one column over for `spill_mem_limit_threshold` (`ResourceGroup.java:135`: `... * 100) + "%"`), so it'd display the same garbage tail — worth applying the BigDecimal fix there too while you're at it.

The PR has also gone stale and now has merge conflicts (`DIRTY`) — a rebase onto current `main` would be needed to move it forward.

PR #62096, opened 2025-08-19
[Enhancement] Optimize the iceberg sink local sorting based on the spill partition writer

GavinMar:
## Why I'm doing:
StarRocks currently supports local sorting of Iceberg tables with sort-order property during writing. When writing to an Iceberg table, a `PhysicalTopNOperator` will be appended during the enfore period, in order to sort data before sink operator.

So, there are obvious performance and memory problems for this implementation:
* The sort operation sorts all the data to be written on the BE node, which will greatly increase the sorting time. Moreover, sorting all the data will occupy a large amount of memory and easily trigger OOM.
* During the whole sort procedure, the sink operator needs to wait synchronously, which cannot take advantage of the performance of the pipeline architecture. 

## What I'm doing:
Optimize the iceberg sink local sorting based on the spill partition writer.  We sort the data to be written in each file writer to ensure the data that written to one file is sorted. If the memory is insufficient, we spill the data to disk or remote storage and then merge sort them before writing to iceberg files.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/connector/sink_memory_manager.cpp
+++ b/be/src/connector/sink_memory_manager.cpp
@@ -35,7 +35,11 @@ bool SinkOperatorMemoryManager::kill_victim() {
     // For spillable partition writer, choose the the writer with the largest memory size that can be spilled.
     PartitionChunkWriterPtr victim = nullptr;
     for (auto& [key, writer] : *_candidates) {
-        if (victim && victim->get_flushable_bytes() > writer->get_flushable_bytes()) {
+        int64_t flushable_bytes = writer->get_flushable_bytes();
+        if (flushable_bytes == 0) {
+            continue;
+        }
+        if (victim && flushable_bytes < victim->get_flushable_bytes()) {
             continue;
         }
         victim = writer;
@@ -60,7 +64,7 @@ int64_t SinkOperatorMemoryManager::update_releasable_memory() {
 int64_t SinkOperatorMemoryManager::update_writer_occupied_memory() {
     int64_t writer_occupied_memory = 0;
     for (auto& [_, writer] : *_candidates) {
-        writer_occupied_memory += writer->get_written_bytes();
+        writer_occupied_memory += writer->get_flushable_bytes();
     }
     _writer_occupied_memory.store(writer_occupied_memory);
     return _writer_occupied_memory;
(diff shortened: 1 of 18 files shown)


be/src/connector/sink_memory_manager.cpp
```diff
@@ -35,7 +35,11 @@ bool SinkOperatorMemoryManager::kill_victim() {
     // For spillable partition writer, choose the the writer with the largest memory size that can be spilled.
     PartitionChunkWriterPtr victim = nullptr;
     for (auto& [key, writer] : *_candidates) {
-        if (victim && victim->get_flushable_bytes() > writer->get_flushable_bytes()) {
+        int64_t flushable_bytes = writer->get_flushable_bytes();
+        if (flushable_bytes == 0) {
+            continue;
+        }
+        if (victim && flushable_bytes < victim->get_flushable_bytes()) {
```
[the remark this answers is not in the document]
GavinMar: The current code checks if the current writer has less flushable bytes than the candidate victim, it it is, skip it and check the next one. This has the same effect with the suggested code, and will not skip writers with  larger flushable bytes.

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

gensrc/thrift/Descriptors.thrift
```diff
@@ -528,6 +528,12 @@ struct TIcebergPartitionInfo {
     4: optional Exprs.TExpr partition_expr
 }
 
+struct TSortOrder {
```
trueeyu, line 531: Why not use `TSortInfo`?

trueeyu (approved):

gensrc/thrift/Descriptors.thrift
GavinMar: > Why not use `TSortInfo`?

Mainly considering the following points:
1. The sort column is also currently passed in the native table's structure using sort_key_idexes. The Iceberg table's creation statement and sort processing are essentially the same logic as the native table's, with the addition of sort-direction and nulls-order. Therefore, a similar structure to the native table is used for transmission, making it more convenient for the BE side to handle.
2. Iceberg sorting may require further support for Iceberg transform sort in the future, requiring further structure expansion. The existing TSortInfo may not meet these requirements.
3. TSortInfo is currently stored in PlanNodes.thrift. Directly using it will cause cross-inclusion issues with Descriptor.thrift, so additional adjustments are required.

gengjun-git (approved):

Seaven (approved):

meegoo (approved):

PR #62210, opened 2025-08-22
[Enhancement] Add be jvm memory metrics

nillin0:
## Why I'm doing:

Some JNI calls are used to interact with Java in the BE.

## What I'm doing:

Adding JVM metrics to the BE, collecting various memory usage statistics during JVM runtime.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -584,7 +584,7 @@ CONF_mBool(enable_token_check, "true");
 // to open/close system metrics
 CONF_Bool(enable_system_metrics, "true");
 
-CONF_Bool(enable_jvm_metrics, "true");
+CONF_Bool(enable_jvm_metrics, "false");
 
 CONF_mBool(enable_prefetch, "true");
 
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -584,6 +584,8 @@ CONF_mBool(enable_token_check, "true");
 // to open/close system metrics
 CONF_Bool(enable_system_metrics, "true");
 
+CONF_Bool(enable_jvm_metrics, "true");
+
 CONF_mBool(enable_prefetch, "true");
 
 // Number of cores StarRocks will used, this will effect only when it's greater than 0.
--- a/be/src/tools/mock_jvm.c
+++ b/be/src/tools/mock_jvm.c
@@ -30,7 +30,8 @@ _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_CreateJavaVM(JavaVM** pvm, void** penv, 
 
 _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_GetCreatedJavaVMs(JavaVM** vm, jsize sz, jsize* output) {
     fprintf(stderr, "unsupported call jni function, JAVA_HOME is required\n ");
-    return JNI_ERR;
+    exit(1);
+    return 1;
 }
 
 /* Defined by native libraries. */
--- a/be/src/tools/mock_jvm.c
+++ b/be/src/tools/mock_jvm.c
@@ -30,8 +30,7 @@ _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_CreateJavaVM(JavaVM** pvm, void** penv, 
 
 _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_GetCreatedJavaVMs(JavaVM** vm, jsize sz, jsize* output) {
     fprintf(stderr, "unsupported call jni function, JAVA_HOME is required\n ");
-    exit(1);
-    return 1;
+    return JNI_ERR;
 }
 
 /* Defined by native libraries. */
--- a/be/src/util/jvm_metrics.cpp
+++ b/be/src/util/jvm_metrics.cpp
@@ -1,3 +1,17 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #include "util/jvm_metrics.h"
 
 #include <vector>
--- a/be/src/util/jvm_metrics.cpp
+++ b/be/src/util/jvm_metrics.cpp
@@ -36,7 +36,7 @@ Status JVMMetrics::init() {
     _get_memory_mxbean = env->GetStaticMethodID(management_factory_cls, "getMemoryMXBean", "()Ljava/lang/management/MemoryMXBean;");
     DCHECK(_get_memory_mxbean != nullptr);
     jobject memory_mxbean_obj = env->CallStaticObjectMethod(management_factory_cls, _get_memory_mxbean);
-    CHECK_JNI_EXCEPTION(env, "get memory mxbean failed");
+    CHECK_JNI_EXCEPTION(env, "get memory mxbean failed")
     _memory_mxbean_obj = JavaGlobalRef(env->NewGlobalRef(memory_mxbean_obj));
     LOCAL_REF_GUARD_ENV(env, memory_mxbean_obj);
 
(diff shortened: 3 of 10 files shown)


dirtysalt (dismissed):

be/src/util/jvm_metrics.cpp
```diff
@@ -0,0 +1,233 @@
...
+    _get_name = env->GetMethodID(memory_pool_mxbean_cls, "getName", "()Ljava/lang/String;");
+    DCHECK(_get_name != nullptr);
+
+    return Status::OK();
+}
+
+void JVMMetrics::install(MetricRegistry* registry) {
+    if (!registry->register_hook("jvm_metrics", [this] { 
+            if (!update().ok()) {
+                LOG(WARNING) << "update jvm metrics failed: " << update().to_string();
```
kevincai: doesn't look right to call update() repeatedly here

nillin0: fixed, please take another look

be/src/common/config.h
```diff
@@ -568,6 +568,8 @@ CONF_mBool(enable_token_check, "true");
 // to open/close system metrics
 CONF_Bool(enable_system_metrics, "true");
 
+CONF_Bool(enable_jvm_metrics, "true");
```
kevincai: Consider disable it by default. BE JVM is not a must-running module.

be/src/util/jvm_metrics.cpp
```diff
@@ -0,0 +1,234 @@
...
+        env->ExceptionDescribe();                                                                  \
+        env->ExceptionClear();                                                                     \
+        env->DeleteLocalRef(thr);                                                                  \
+        return Status::InternalError(fmt::format("{}, error: {}", message, jni_error_message));    \
+    }                                                                                              \
+
+namespace starrocks {
+
+Status JVMMetrics::init() {
+    RETURN_IF_ERROR(detect_java_runtime());
```
kevincai, line 37: this will just check if JAVA_HOME set or not. It doesn't guarantee that the JVM env is properly set.

be/src/util/jvm_metrics.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    JNIEnv* env = JVMFunctionHelper::getInstance().getEnv();
```
kevincai, line 44: env could be nullptr.

be/src/util/jvm_metrics.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    jclass memory_pool_mxbean_cls = env->FindClass("java/lang/management/MemoryPoolMXBean");
+    DCHECK(memory_pool_mxbean_cls != nullptr);
+    LOCAL_REF_GUARD_ENV(env, memory_pool_mxbean_cls);
+    _get_usage = env->GetMethodID(memory_pool_mxbean_cls, "getUsage", "()Ljava/lang/management/MemoryUsage;");
+    DCHECK(_get_usage != nullptr);
+    _get_peak_usage = env->GetMethodID(memory_pool_mxbean_cls, "getPeakUsage", "()Ljava/lang/management/MemoryUsage;");
+    DCHECK(_get_peak_usage != nullptr);
+    _get_name = env->GetMethodID(memory_pool_mxbean_cls, "getName", "()Ljava/lang/String;");
+    DCHECK(_get_name != nullptr);
```
kevincai: DCHECK doesn't help anything in release mode.

be/src/util/jvm_metrics.h
```diff
@@ -0,0 +1,86 @@
...
+    StatusOr<std::vector<MemoryPool>> get_memory_pools();
+
+private:
+    std::unique_ptr<JVMClass> _management_factory_cls = nullptr;
+    JavaGlobalRef _memory_mxbean_obj = nullptr;
+
+    jmethodID _get_memory_mxbean = nullptr;
+    jmethodID _get_heap_memory_usage = nullptr;
+    jmethodID _get_non_heap_memory_usage = nullptr;
+    jmethodID _get_init = nullptr;
```
kevincai, line 88: consider graceful shutdown scenario, can the memory referenced by these variables be properly released?

alvin-phoenix-ai:
@cursor review

be/src/common/config.h
nillin0: Done.

be/src/util/jvm_metrics.cpp
nillin0: env had been checked in JVMFunctionHelper::getInstance().
```cpp
JVMFunctionHelper& JVMFunctionHelper::getInstance() {
    if (_env == nullptr) {
        _env = getJNIEnv();
        CHECK(_env != nullptr) << "couldn't got a JNIEnv";
    }
    static JVMFunctionHelper helper;
    return helper;
}
```

be/src/util/jvm_metrics.cpp
nillin0: Had replaced DCHECK with CHECK_JNI_EXCEPTION to catch exception when nullptr returns.

be/src/util/jvm_metrics.h
nillin0: I think that jmethodID values don’t need to be released manually—they’re internal JVM function pointers whose lifetime matches the JVM itself.

be/src/util/jvm_metrics.cpp
kevincai: CHECK will crash the process in release mode. shouldn't be implemented this way.

kevincai (dismissed):

be/src/util/jvm_metrics.cpp
nillin0: I had added JNIEnv check in init() and modified JNI_GetCreatedJavaVMs in mock_jvm.c for it would exit program.

kevincai:
@nillin0 there is a code conflict with the current main branch, can you rebase the code to solve the conflict.

nillin0:
> @nillin0 there is a code conflict with the current main branch, can you rebase the code to solve the conflict.

done.

be/src/tools/mock_jvm.c
```diff
@@ -30,8 +30,7 @@ _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_CreateJavaVM(JavaVM** pvm, void** penv,
 
 _JNI_IMPORT_OR_EXPORT_ jint JNICALL JNI_GetCreatedJavaVMs(JavaVM** vm, jsize sz, jsize* output) {
     fprintf(stderr, "unsupported call jni function, JAVA_HOME is required\n ");
-    exit(1);
-    return 1;
+    return JNI_ERR;
```
kevincai: this mock_jvm.so is not supposed to be dynamically linked during the runtime, it is just a mock to workaround the binary link issue. Shall cry loudly if it is wrongly loaded in runtime.

be/src/util/jvm_metrics.cpp
```diff
@@ -0,0 +1,243 @@
+#include "util/jvm_metrics.h"
```
kevincai, line 15: Please add license header.

be/src/util/jvm_metrics.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
```
kevincai, line 15: add license header for new files.

kevincai (dismissed):

be/src/util/jvm_metrics.cpp
```diff
@@ -0,0 +1,257 @@
...
+            }
+        })) {
+        return;
+    }
+
+#define REGISTER_JVM_METRIC(name) registry->register_metric("jvm_" #name "_bytes", &jvm_##name##_bytes)
+
+    REGISTER_JVM_METRIC(heap_used);
+    REGISTER_JVM_METRIC(heap_committed);
+    REGISTER_JVM_METRIC(heap_max);
```
wyb: why not use label like FE JVM metrics?

```
jvm_heap_size_bytes{type="max"} 1024
jvm_heap_size_bytes{type="committed"} 1024
jvm_heap_size_bytes{type="used"} 1024
```

nillin0: have modified, please take a review. @wyb

wyb (approved):

kevincai (approved):

PR #62225, opened 2025-08-22
[Enhancement] Rewrite MIN(f(col)) to f(MIN(col)) for Monotonic Functions

### LINKED ISSUE #62222: Optimization: Rewrite MIN(f(col)) to f(MIN(col)) for Monotonic Functions
## Background

In StarRocks, expressions inside aggregation functions are currently evaluated for every row. For example:

```sql
SELECT MIN(TO_DATETIME(ts)) FROM t;
```

This evaluates `TO_DATETIME(ts)` N times (once per row) before applying `MIN`. However, in many real workloads, this can be optimized.

## Optimization

We propose an optimizer rule that rewrites:

```sql
MIN(TO_DATETIME(ts)) → TO_DATETIME(MIN(ts))
MAX(TO_DATETIME(ts)) → TO_DATETIME(MAX(ts))
```

This reduces function evaluations from **O(N)** to **O(1)** while preserving correctness.

## Correctness Conditions

1. **Aggregation must be MIN or MAX.**
2. **Function must be deterministic and monotonic.**

   * For the initial implementation, only certain date-related functions are supported.
   * Example: `TO_DATETIME(BIGINT)`, `TO_DATE(DATETIME)`.
3. **NULLs are handled consistently:** both forms ignore NULL values.
4. **Domain must be valid:** inputs must not produce overflow/invalid casts.
   * Example counter-case: `MAX(TO_DATETIME('2024-21-23'))`. In the original form, illegal values become `NULL` and are ignored by `MAX`. In the rewritten form, `MAX('2024-21-23')` could yield the illegal string and then `TO_DATETIME()` returns `NULL`, changing semantics.
   * Therefore, rewrite is only safe when the function is **total** (defined for all possible inputs of the column). For date functions, this usually means numeric timestamp inputs.

## Scope (Initial)

* This optimization will **only apply to a subset of  functions**.
* Functions will be explicitly marked in `FunctionSet` as **`can_push_through_minmax`**.
* Examples:
  * `TO_DATETIME(BIGINT)`
  * `TO_DATE(DATETIME)`
* Not applied to string-based conversions that may return `NULL` for invalid inputs.

## Implementation Detail

* Before applying the rewrite, StarRocks will first check the \*\*raw ****`MIN(col)`**** or \*\***`MAX(col)`** on the original data. The optimizer then checks if the resulting value is within the **legal datetime .
* If the value is valid, apply the rewrite and evaluate the outer function once.
* If the value is out of range, **fallback to the original plan** (evaluate `f(col)` per row), ensuring semantic correctness.

## Performance Considerations

* **Biggest gain**: expensive per-row casts (e.g., bigint → datetime parsing).
* **Comparison cost**: comparing `BIGINT` is as cheap (or cheaper) than comparing `DATETIME`.
* **Index/ordering benefit**: when `ts` is sorted or indexed, `MIN(ts)`/`MAX(ts)` can be fetched quickly, requiring only one final function call.

## Impact

* Significant CPU savings for large tables with timestamp-to-date conversions.
* Transparent to users, no semantic changes (under the domain restriction).
* Safe fallback path ensures no correctness issues when invalid values exist.
* Provides a foundation for future expansions beyond date functions.

murphyatwork:
## Why I'm doing:

This PR resolves #62222 by optimizing queries involving `MIN(f(col))` or `MAX(f(col))` where `f` is a monotonic function. The goal is to reduce per-row function cost by pushing the aggregate function inside the monotonic function, i.e., rewriting `MIN(f(col))` to `f(MIN(col))`.

## What I'm doing:

- Implemented a new optimizer transformation rule, `RewriteMinMaxByMonotonicFunctionRule`.
- This rule rewrites `MIN(f(col))` to `f(MIN(col))` and `MAX(f(col))` to `f(MAX(col))`.
- Initial support is limited to `to_datetime(BIGINT)`
- Registered the new rule enum `TF_REWRITE_MINMAX_BY_MONOTONIC_FUNCTION` in `RuleType`.

Fixes #62222

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---
<a href="https://cursor.com/background-agent?bcId=bc-e9c95739-6bdd-46e6-9319-d64e3bf6aad4">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">
    Open in Cursor
  </picture>
</a>
<a href="https://cursor.com/agents?id=bc-e9c95739-6bdd-46e6-9319-d64e3bf6aad4">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">
    Open in Web
  </picture>
</a>


murphyatwork:
@Cursor review

stdpain (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteMinMaxByMonotonicFunctionRule.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+        // Add rewritten outputs
+        postProj.putAll(rewrittenPostProj);
+
+        // Build the new expression
+        LogicalOlapScanOperator newScan =
+                new LogicalOlapScanOperator.Builder()
+                        .withOperator(scan).setProjection(new Projection(newPreProj))
+                        .build();
+        LogicalAggregationOperator newAgg =
```
stdpain, line 234: It seems newAgg should also apply the predicates.

Seaven (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteMinMaxByMonotonicFunctionRule.java
murphyatwork: disable the support of aggregation with `HAVING`

gengjun-git (approved):

Seaven (approved):

stdpain (approved):

PR #62253, opened 2025-08-25
[Enhancement] Fast schema evolution supports adding key columns for shared-data

banmoy:
### Why I’m doing this
- Reduce schema change latency and avoid full data rewrite when adding key columns under the following conditions.
  - The new key column’s default is a constant or NULL (deterministic backfill).
  - The short key remains unchanged, so prefix index layout is not affected.
- Duplicate/Aggregate/Unique tables are supported. (Primary table does not support adding key column)
- Only shared-data is supported in this PR
   -  Will support shared-nothing in another PR which is based on this PR and also needs to modify BE codes, otherwise ingestions for aggregate/unique table will fail with error `sort key columns is different with key columns` 

### What I’m doing
- Remove the blanket ban on fast schema evolution when adding KEY columns; allow the fast path when conditions are met (removed the one-size-fits-all `newColumn.isKey()` disable).
- Extract and refine short-key computation: add `calculateShortKey(...)` to compute the new short-key columns and compare them with the original ones.
- Add a `shortKeyChanged` flag to `SchemaChangeData`, and propagate it via `withNewIndexShortKeyCount(..., shortKeyChanged)` to decide fast-path eligibility.
- In `analyzeAndCreateJob(...)`, if `isShortKeyChanged()` is true, automatically fall back to the normal schema change; otherwise, keep using fast schema evolution (shared-nothing: direct metadata update; shared-data: async fast job).

- Changes are focused in:
  - `com.starrocks.alter.SchemaChangeHandler`
  - `com.starrocks.alter.SchemaChangeData`

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [X] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [X] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -1152,11 +1152,6 @@ public class SchemaChangeHandler extends AlterHandler {
             throw new DdlException("PERCENTILE_UNION must be used in AGG_KEYS");
         }
 
-        //type key column do not allow light schema change.
-        if (newColumn.isKey()) {
-            fastSchemaEvolution = false;
-        }
-
         // check if the new column already exist in base schema.
         // do not support adding new column which already exist in base schema.
         Optional<Column> foundColumn = olapTable.getBaseSchema().stream()
@@ -1513,7 +1508,6 @@ public class SchemaChangeHandler extends AlterHandler {
             dataBuilder.withComputeResource(computeResource);
         }
 
-        long baseIndexId = olapTable.getBaseIndexId();
         Map<Integer, Column> columnUniqueIdToColumn = Maps.newHashMap();
         // begin checking each table
         // ATTN: DO NOT change any meta in this loop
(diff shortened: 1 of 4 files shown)


sevev (dismissed):

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -1661,6 +1621,85 @@ private SchemaChangeData finalAnalyze(Database db, OlapTable olapTable,
...
+    }
+
+    private boolean isShortKeyChanged(List<Column> originShortKeyColumns, List<Column> newShortKeyColumns) {
+        if (originShortKeyColumns.size() != newShortKeyColumns.size()) {
+            return true;
+        }
+        for (int i = 0; i < originShortKeyColumns.size(); i++) {
+            Column originColumn = originShortKeyColumns.get(i);
+            Column newColumn = newShortKeyColumns.get(i);
+            if (!originColumn.equals(newColumn)) {
```
xiangguangyxg: Whether colunm comments are included in this equals here?

banmoy: Only comparing column name and type should be enough.

xiangguangyxg: So the column comment should not be compared, but I find it compared in the equals function.

banmoy: Yes, just fixed it. Do not depend on `Column.equals` anymore

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -1661,6 +1621,86 @@ private SchemaChangeData finalAnalyze(Database db, OlapTable olapTable,
...
+    }
+
+    private boolean isShortKeyChanged(List<Column> originShortKeyColumns, List<Column> newShortKeyColumns) {
+        if (originShortKeyColumns.size() != newShortKeyColumns.size()) {
+            return true;
+        }
+        for (int i = 0; i < originShortKeyColumns.size(); i++) {
+            Column originColumn = originShortKeyColumns.get(i);
+            Column newColumn = newShortKeyColumns.get(i);
+            if (!originColumn.getName().equalsIgnoreCase(newColumn.getName())
```
xiangguangyxg, line 1696: should compare column id not column name ?

banmoy: It should be safe to compare column names here because ADD/MODIFY/DROP do not change column name. Actually `SchemaChangeHandler` has been always using column names to find the original columns.

xiangguangyxg: What will happend if a original column has been renamed

banmoy: The rename could not happen concurrently with alter operations in`SchemaChangeHandler` because they must get table write lock before starting. The rename is processed in `AlterJobExecutor.visitColumnRenameClause`

xiangguangyxg: Not mean that rename could be concurrently with alter. Just want to know if a key column has been renamed before, its column name is no longer the same as column id, what will happend here? Could it happen that a column name is equal but column id is not equal?

xiangguangyxg (dismissed):

chaoyli:
How about add value column in the primary key table?

verbose
chaoyli, line 1: This file is unused. It's better to remove.

banmoy: Fixed

banmoy:
> How about add value column in the primary key table?

@chaoyli  Fast schema evolution has already supported it

xiangguangyxg (approved):

alvin-phoenix-ai:
@cursor review

gengjun-git (approved):

PR #62330, opened 2025-08-26
[Feature] (IVM Part6) Support IVM based mv refresh for basic operators

### LINKED ISSUE #61789: [Feature] [MV] Incremental Materialized View Refresh
- [x] https://github.com/StarRocks/starrocks/pull/62330
- [x] https://github.com/StarRocks/starrocks/pull/62122

# Feature request

StarRocks supports partition-based materialized view refresh (https://docs.starrocks.io/docs/using_starrocks/async_mv/Materialized_view/) , but it has a number of deficiencies or shortcomings:
- Cost:
  - If base table's partition is updated, you need to refresh the affected MV's partitions fully by Insert Overwrite;
- Modeling:
  - Partitioning:
    - MV is strongly aligned/coupled with the base table's partition; e.g., MV's partition must be date_trunc if base table's date_trunc, cannot be other partition expressions.
    - If the base table is not a partitioned table, you cannot create a partitioned MV.
  - Refreshing:
    - Refresh MV OOM when MV's base table contains too much data (especially when MV contains many group by keys).
- Competitiveness:
  - Highlights and selling points

To reduce the refreshing cost of materialized views and attract more users to utilize materialized views, we can introduce a new materialized view refresh method `IVMBasedMVRefreshProcessor`.

# Concepts

- PCT(Partition Change Tracking): Tracks changes at the partition level of a table to enable efficient incremental processing.
- IVM(Incremental View Maintenance): Maintains materialized views by applying only data deltas instead of full recomputation.
- CDC(Change Data Capture): Captures and propagates row-level changes from source tables to downstream systems in real time.
- TVR(Time-Varying Relation): Represents a logical relation that evolves over time, supporting versioned and temporal queries.
- Iceberg Snapshot/Version: Each snapshot in Apache Iceberg captures a table’s state at a point in time, enabling time travel, rollback, and incremental reads.
- Incremental Read: Reads only newly added or changed data since the last read, reducing scan cost and latency.

# Objects

Introduce incremental compute in StarRocks and 
- Cost:
  - Reduce the refreshing cost of materialized views.
- Modeling:
  - Incremental refresh MV to avoid OOM when MV's base table contains too much data;
  - Expand the capability scope of materialized views to address the partition dependency issues that asynchronous views cannot resolve.

We may split this basic framework into multiple phases, and in this current phase we can support incremental changes for Iceberg append-only tables:
  - Build a basic incremental framework for incremental compute;
  - Support major operators like Select/Project/Join/Aggregate/UnionAll;

# Design

## Incremental Reads
We can use Iceberg format's IncrementalScan API to implement simple incremental reads, which can work with both V1 and V2 format versions. 

For retractable changes, we may use Iceberg V3 to produce CDC changes later.

Image

## Incremental Compute

We can use relational algebra to compute incremental results to reduce redundant computations and improve performance while reusing existing query engine optimizations as much as possible, which also means preferring to use recompute to compute incremental results for retractable changes. And for all operations involving state (aggregation/join/window), use result tables as state tables instead of maintaining additional state tables.

Image

# Roadmap
1. (Phase 1) Build a basic incremental refresh framework to support incremental changes for Iceberg  append-only data (while considering future extensibility), supporting major operators like Select/Project/Join/Aggregate:
2. (Phase 2) Add more smart strategies to make incremental materialized view refresh better used:
  - Choose how many incremental snapshots/versions to refresh;
  - Choose whether to use incremental or full strategy to refresh;
  - Be compatible with old Partition Based Refresh；
  - Be compatible with mv rewrite;

LiShuMing:
## What I'm doing:
Support IVM for some basic operators:
- Scan/Filter/Project
- Inner Join
- Union All

This pull request introduces support for a new "refresh mode" property for materialized views, enabling more flexible and fine-grained control over how materialized views are refreshed. The changes include defining the new property, updating relevant classes to handle it, and adding supporting logic and data structures for incremental and partition-based refresh operations.

### Materialized View Refresh Mode Support

* Added a new `refresh_mode` property (`PROPERTIES_REFRESH_MODE`) for materialized views, allowing users to specify how the view should be refreshed (e.g., AUTO, PCT, FULL, INCREMENTAL). This includes validation and parsing logic in `PropertyAnalyzer` and integration into materialized view creation and alteration workflows. [[1]](diffhunk://#diff-61be4f97e7d1b1a47bd2246fa0a233592132c888cebba90077216db336661ba1R218) [[2]](diffhunk://#diff-61be4f97e7d1b1a47bd2246fa0a233592132c888cebba90077216db336661ba1R679-R696) [[3]](diffhunk://#diff-00aa3dcf5aaa2f891fbf8aeb40f13d30052ce4e51658d21e2c23cbac02dbdcd7R150-R153) [[4]](diffhunk://#diff-00aa3dcf5aaa2f891fbf8aeb40f13d30052ce4e51658d21e2c23cbac02dbdcd7R346-R350) [[5]](diffhunk://#diff-61be4f97e7d1b1a47bd2246fa0a233592132c888cebba90077216db336661ba1R1726-R1732)
* Introduced the `RefreshMode` enum in `MaterializedView`, with methods to determine the current mode and defaults, and exposed getter/setter methods for this property in `TableProperty` and `MaterializedView`. [[1]](diffhunk://#diff-30af13c1a1cdd2dcedc3200a0a9a3497645f60a2a173ec4b2ca17217af93afb0R170-R208) [[2]](diffhunk://#diff-52d5477d33c44a0027256c5ecb673a76d7ee203626a4d2ae0126530827fe8389R218-R221) [[3]](diffhunk://#diff-52d5477d33c44a0027256c5ecb673a76d7ee203626a4d2ae0126530827fe8389R997-R1004) [[4]](diffhunk://#diff-30af13c1a1cdd2dcedc3200a0a9a3497645f60a2a173ec4b2ca17217af93afb0R900-R903)

TODO:
- Support more operators
- Add more checks
- Add more documents

Fixes https://github.com/StarRocks/starrocks/issues/61789

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
```diff
@@ -492,6 +552,9 @@ public String toString() {
     // This is the original user's view define SQL which can be used to generate ast key in text based rewrite.
     @SerializedName(value = "originalViewDefineSql")
     private String originalViewDefineSql;
+    // This is the rewritten view define SQL which is used to generate IVM refresh tasks.
+    @SerializedName(value = "ivmDefineSql")
```
kangkaisen: why iMV couldn't use the viewDefineSql?

fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMBasedMVRefreshProcessor.java
```diff
@@ -0,0 +1,471 @@
...
+            mvRefreshContext.getTempBaseTableInfoTvrDeltaMap().putAll(tempMvTvrVersionRangeMap);
+            executor.executePlan(execPlan, insertStmt);
+        }
+
+        try (Timer ignored = Tracers.watchScope("MVRefreshUpdateMeta")) {
+            try {
+                updatePCTMeta(execPlan, pctMVToRefreshedPartitions, pctRefTableRefreshPartitions);
+            } catch (Exception e) {
+                // if the update meta failed, we should not throw exception here
+                // because this meta update only affects mv refresh rather than mv refresh.
```
kangkaisen: typo?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tvr/common/TvrLazyOptExpression.java
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+package com.starrocks.sql.optimizer.rule.tvr.common;
+
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
+
+/**
+ * TvrLazyOptExpression is a wrapper for TvrOptExpression that allows lazy loading of the expression.
```
kangkaisen, line 21: what's the meaning of 'lazy loading of the expression.'?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tvr/TvrTransformationRule.java
```diff
@@ -0,0 +1,135 @@
...
+import com.starrocks.sql.optimizer.rule.tvr.common.TvrOptMeta;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * TvrTransformationRule is an abstract class that provides common functionality for tvr transformation rules.
+ */
+public abstract class TvrTransformationRule extends TransformationRule {
```
kangkaisen, line 36: why do we need to separate TransformationRule  framework for TVR?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/OptExpression.java
```diff
@@ -66,6 +67,9 @@ public class OptExpression {
     // the flag if its parent has required data distribution property for this expression
     private boolean existRequiredDistribution = true;
 
+    // the TvrOptMeta of this expression
+    private TvrOptMeta tvrOptMeta;
```
kangkaisen, line 71: why optexpression need to care the TvrOptMeta?

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3383,6 +3383,9 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true)
     public static String default_mv_partition_refresh_strategy = "strict";
 
+    @ConfField(mutable = true)
+    public static String default_mv_refresh_mode = "pct";
```
satanson, line 3390: why not use "auto" for default refresh mode? then introduce a decision-maker to determine which real refresh mode takes effect.

fe/fe-core/src/main/java/com/starrocks/connector/CatalogConnectorMetadata.java
```diff
@@ -149,6 +151,28 @@ public Table getTable(ConnectContext context, String dbName, String tblName) {
         return metadata.getTable(context, dbName, tblName);
     }
 
+    @Override
+    public TvrTableSnapshot getCurrentTvrSnapshot(String dbName, Table table) {
```
satanson, line 155: what does Tvr mean? time-varying relation?  how about using TVR instead of Tvr, Ivm instead of IVM

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
```diff
@@ -514,6 +520,50 @@ public Table getView(ConnectContext context, String dbName, String viewName) {
...
+        final long toSnapshotIdInclusive =
+                toSnapshotInclusive.end().orElseThrow(() -> new StarRocksConnectorException(
+                        "toSnapshotInclusive must have a valid snapshot ID"));
+        final IcebergTable icebergTable = (IcebergTable) table;
+        final org.apache.iceberg.Table nativeTable = icebergTable.getNativeTable();
+        long lastSnapshotId = toSnapshotIdInclusive;
+
+        final List<TvrTableDeltaTrait> tvrDeltaTraits = Lists.newArrayList();
+        // snapshot from the last to the first
+        final Iterable<Snapshot> snapshots = SnapshotUtil.ancestorsBetween(
```
satanson, line 549: Is it possible that returned snapshots are overlapping?

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
```diff
@@ -951,20 +1001,29 @@ private Iterator<FileScanTask> buildFileScanTaskIterator(IcebergTable icebergTab
...
-                .planWith(jobPlanningExecutor);
-
+        Scan scan;
+        if (tvrVersionRange.start() != null && tvrVersionRange.start().isPresent()) {
+            IncrementalAppendScan incrementalAppendScan = nativeTbl.newIncrementalAppendScan();
+            incrementalAppendScan =
+                    incrementalAppendScan.fromSnapshotExclusive(tvrVersionRange.from.getVersion());
+            incrementalAppendScan =
+                    incrementalAppendScan.toSnapshot(snapshotId);
+            scan = incrementalAppendScan;
```
satanson: Could Obtaining incrementalAppendScan trigger access a great deal of metadata ? why not construct a plan to fetch these data just like prevous?

fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
LiShuMing: `viewDefineSql` is used for `aggregate operator` in IVM which will be used later.

For `aggregate operator`, we need to generate an incremental refresh plan by rewritting its ast tree for better anlysis and validation.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tvr/common/TvrLazyOptExpression.java
LiShuMing: An TvrOptExpression may contain multi time versions(eg: from/to snapshot's opt expression) but those time-varying OptExpressions may not be visited for the parent OptExpressions.
eg: If parent OptExpression is JoinOpt, it may visit Children' time-varying OptExpressions, otherwise it may not visit.

To avoid early wasting computes, use `TvrLazyOptExpression` to save the TvrLazyOptExpression compute, and only trigger to compute for parent's needs.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tvr/TvrTransformationRule.java
LiShuMing: update, no sperate indeed. Just use `TvrTransformationRule` to provide some common functionality for tvr transform rules.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/OptExpression.java
LiShuMing: `TvrOptMeta` is used to represent children's time-varying OptExpressions which can contain `current/from/to`'s plan which can be used for current transformation.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
LiShuMing: Good question.

I want to use `auto` for more intelligent strategy for automatic switches between `pct` and `increment` later.

And for current's implementions, we are using `pct` (partition change tracking) by default.

fe/fe-core/src/main/java/com/starrocks/connector/CatalogConnectorMetadata.java
LiShuMing: yes. it means for `time-varying relation` Since this is just a choose for IVM's implementation.

> `how about using TVR instead of Tvr`?
Since `Tvr` is already used for current codes, can we keep to use it `Tvr` then?

satanson (approved):

satanson (dismissed):

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
LiShuMing: It will not overlapping from Iceberg SDK.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
LiShuMing: It's a good idea. 

> Could Obtaining incrementalAppendScan trigger access a great deal of metadata ? 

Yes, it may be, and this is associated with Iceberg's compaction from user side.

> why not construct a plan to fetch these data just like prevous?
This seems fine by now because original mv refresh will still fetch those metadata either. I can take a look to optimize by metada distributed plan later.

fe/fe-core/src/main/java/com/starrocks/common/util/PropertyAnalyzer.java
```diff
@@ -214,6 +215,7 @@ public class PropertyAnalyzer {
     public static final String PROPERTIES_PARTITION_REFRESH_NUMBER = "partition_refresh_number";
     public static final String PROPERTIES_EXCLUDED_TRIGGER_TABLES = "excluded_trigger_tables";
     public static final String PROPERTIES_EXCLUDED_REFRESH_TABLES = "excluded_refresh_tables";
+    public static final String PROPERTIES_REFRESH_MODE = "refresh_mode";
```
stephen-shelby: if we need to add prefix like `mv_refresh_mode` ? Since `refresh_mode` is relatively common

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
```diff
@@ -951,20 +1001,29 @@ private Iterator<FileScanTask> buildFileScanTaskIterator(IcebergTable icebergTab
...
                 "Snapshot ID is not present in tvrVersionRange: " + tvrVersionRange));
-        TableScan scan = icebergCatalog.getTableScan(nativeTbl, scanContext)
-                .useSnapshot(snapshotId)
-                .metricsReporter(metricsReporter)
-                .planWith(jobPlanningExecutor);
-
+        Scan scan;
+        if (tvrVersionRange.start() != null && tvrVersionRange.start().isPresent()) {
+            IncrementalAppendScan incrementalAppendScan = nativeTbl.newIncrementalAppendScan();
+            incrementalAppendScan =
```
stephen-shelby, line 1007: I think you could add following properties like 
```
.metricsReporter(metricsReporter)
.planWith(jobPlanningExecutor);
```
so we can get some metrics in the query profile

stephen-shelby (approved):

Seaven (approved):

gengjun-git (approved):

kangkaisen (approved):

PR #62396, opened 2025-08-27
[BugFix] fix the privilege issue of refresh mv

murphyatwork:
## What I'm doing:

MV currently employs creator-based authorization, where the system records the creator (user) of an MV and uses that user for refreshing the MV. However, this approach poses challenges:
1. If the user lacks a default role, the MV refresh process will attempt to use the user's default role, leading to privilege issues.
2. In LDAP systems, this method is unreliable because the user-based approach may not function consistently.

To address these issues, we implemented the following improvements:
- During MV refresh, all roles of the user are activated instead of relying on the default role.
- A new configuration, `mv_use_creator_based_authorization`, has been introduced. When set to false, the system uses the ROOT user to refresh the MV, which is more appropriate for LDAP environments.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -3411,6 +3411,13 @@ public class Config extends ConfigBase {
     public static boolean mv_auto_analyze_async = true;
 
     /**
+     * Creator-based: record the creator(user) of MV, refresh the MV with same user
+     * Root-based: always use the ROOT to refresh the MV
+     */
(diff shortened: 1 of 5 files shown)


LiShuMing (dismissed):
+1

HangyuanLiu (approved):

LiShuMing (approved):

PR #62484, opened 2025-08-28
[Enhancement] Add support for simple partition_by, buckets clauses

dasdy:
## Why I'm doing:

We want to create tables of following format using alembic:
```sql
CREATE TABLE IF NOT EXISTS account (
       /* .... */
      )
      PRIMARY KEY (column1, column2)
      PARTITION BY (column1)
      DISTRIBUTED BY HASH (column2) BUCKETS 3
      ORDER BY (column3)
      PROPERTIES (
       /* .... */
      );
```

Unfortunately, this is not possible even using latest version of the library on `main` branch. The missing parts are:
1. `BUCKETS <num>` clause
2. `PARTITION BY <column>` - full support for this clause is complicated, but for us just a simple column expression will be enough.

## What I'm doing:

I'm adding basic support for those statements using `__table_args__`, similar to how `DISTRIBUTED BY HASH` statement is supported at the moment.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/contrib/starrocks-python-client/README.md
+++ b/contrib/starrocks-python-client/README.md
@@ -87,6 +87,10 @@ class table1(Base):
         "starrocks_PRIMARY_KEY": "id",
         "starrocks_engine": "OLAP",
         "starrocks_comment": "table comment",
+        "starrocks_distributed_by": "id",
+        "starrocks_distributed_by_buckets": 5,
+        "starrocks_partition_by": "name",
+        "starrocks_order_by": "created_at",
         "starrocks_properties": (
             ("storage_medium", "SSD"),
             ("storage_cooldown_time", "2025-06-04 00:00:00"),
--- a/contrib/starrocks-python-client/test/test_suite.py
+++ b/contrib/starrocks-python-client/test/test_suite.py
@@ -19,7 +19,7 @@ from textwrap import dedent
 from sqlalchemy.testing.suite import *
 
 from sqlalchemy.testing.assertions import AssertsCompiledSQL
-from sqlalchemy import VARCHAR, Table, Column, Integer, MetaData
+from sqlalchemy import VARCHAR, Table, Column, Integer, MetaData, DateTime
 from sqlalchemy import schema
 
 from sqlalchemy.testing import fixtures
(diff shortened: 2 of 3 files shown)


contrib/starrocks-python-client/README.md
```diff
@@ -87,6 +87,10 @@ class table1(Base):
         "starrocks_PRIMARY_KEY": "id",
         "starrocks_engine": "OLAP",
         "starrocks_comment": "table comment",
+        "starrocks_distributed_by": "id",
+        "starrocks_distributed_by_buckets": 5,
+        "starrocks_partition_by": "name",
```
jaogoy, line 92: Will it support `RANGE(id, name)`, `LIST(id, name)`, `data_trunc('day', dt)`?
So, here is a expresion partitioning?

contrib/starrocks-python-client/test/test_suite.py
```diff
@@ -56,6 +56,54 @@ def test_create_primary_key_table(self):
             schema.CreateTable(tbl),
             "CREATE TABLE btable (id BIGINT NOT NULL AUTO_INCREMENT)PRIMARY KEY(id) DISTRIBUTED BY HASH(id)")
 
+    def test_create_primary_key_with_hash_by_buckets(self):
+        m = MetaData()
+        tbl = Table(
+            'btable', m, Column("id", Integer, primary_key=True),
+            starrocks_primary_key="id",
+            starrocks_distributed_by="id",
```
jaogoy, line 64: `id`  or `HASH(id)`?
SR support `RANDOM` too.

contrib/starrocks-python-client/README.md
```diff
@@ -87,6 +87,10 @@ class table1(Base):
         "starrocks_PRIMARY_KEY": "id",
         "starrocks_engine": "OLAP",
         "starrocks_comment": "table comment",
+        "starrocks_distributed_by": "id",
+        "starrocks_distributed_by_buckets": 5,
```
jaogoy, line 91: we can directly use: `"starrocks_distributed_by": "HASH(id) BUCKETS 8"`.

dasdy: I don't think this is how it currently works on `main`: even without my changes, the `DISTRIBUTED BY` expression is built like this:

```python
         if 'DISTRIBUTED_BY' in opts:
            table_opts.append(f'DISTRIBUTED BY HASH({opts["DISTRIBUTED_BY"]})')
```

so the `by hash` is hardcoded. I did not want to mess with the logic of the library: changing the behavior to one where we can just provide the whole SQL expression in the parameter (like `"HASH(id) ...."`) is backwards-incompatible change, and my plan was to extend already-existing behavior without problems for others.

contrib/starrocks-python-client/test/test_suite.py
dasdy: hash, it is currently hardcoded in `main`, I did not make changes to that logic.

contrib/starrocks-python-client/README.md
dasdy: Yes - I've added a test for that and did a small fix regarding that functionality.

PR #62537, opened 2025-08-29
[Enhancement] Prune unnecessary cte consume when prune empty join

kisshot288:
## Why I'm doing:
When prune a empty join, it just replace join to a logical value, but do not decrease cte consume.
So if a cte consume join with a empty set, join will be execute even the it is unnecessary, 
## What I'm doing:
When prune a empty join, decrease possible cte consume. So it's possible to prune cte produce(since there is no cte consume)
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/CTEContext.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/CTEContext.java
@@ -96,6 +96,12 @@ public class CTEContext {
         this.consumeNums.put(cteId, i + 1);
     }
 
+    public void decCTEConsume(int cteId) {
+        int i = this.consumeNums.getOrDefault(cteId, 0) - 1;
+        Preconditions.checkState(i >= 0);
+        this.consumeNums.put(cteId, i);
+    }
+
     public void addCTEStatistics(int cteId, Statistics statistics) {
         produceStatistics.put(cteId, statistics);
     }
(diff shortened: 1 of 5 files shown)


eshishki:
@Seaven you own most of `CTEContext`, so flagging this for your judgment. The PR addresses a real but narrow inefficiency: `PruneEmptyJoinRule` replaces an empty join with `LogicalValues` but never decrements the CTE consume count, so a CTE that's provably no longer consumed still gets materialized (and a redundant branch may execute). Result is correct either way — this is purely perf/plan cleanup, not a correctness bug.

The catch is the fix touches CTE accounting directly: it adds `decCTEConsume` with `Preconditions.checkState(i >= 0)` and flips `needInline` to a count-based check. So the perf win comes at the cost of poking a sensitive counter — if the consume/produce balance doesn't hold on every path, that assertion turns into a planner crash.

Given it's been open ~9 months with no review (and now conflicting), could you take a look and judge whether it's worth pursuing? If the approach is sound it's a small change worth rescuing; if the accounting risk isn't worth the narrow gain, better to say so. Thanks.

PR #62539, opened 2025-08-30
[Enhancement] Enable skew join rewrite for NULL-only skew

murphyatwork:
## What I'm doing:

Extend stats-based skew join optimization to trigger when the join column’s NULL fraction exceeds the threshold, even if MCVs are empty.
- Consider nullsFraction in skew detection (uses skewJoinDataSkewThreshold)
- Avoid mutating incoming skewValues; add IN branch only when non-empty
- Ensure right-side salt table builds when skewValues is empty/NULL-only

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SkewJoinOptimizeRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SkewJoinOptimizeRule.java
@@ -171,8 +171,19 @@ public class SkewJoinOptimizeRule extends TransformationRule {
             double nullsFraction = leftColumnStats.getNullsFraction();
             if (isDataSkew(leftChildMCV, leftRowCount, nullsFraction, context.getSessionVariable())) {
                 joinOperator.setSkewColumn(skewJoinColumn);
-                joinOperator.setSkewValues(leftChildMCV.stream().map(pair -> ConstantOperator.createVarchar(pair.first)).
-                        collect(Collectors.toList()));
+
+                // Handle NULL-only skew case: when MCV is empty but NULL fraction indicates skew
+                List<ScalarOperator> skewValues;
+                if (leftChildMCV.isEmpty() &&
+                        nullsFraction > context.getSessionVariable().getSkewJoinDataSkewThreshold()) {
+                    // Create a special NULL skew value for NULL-only skew cases
+                    skewValues = Lists.newArrayList(ConstantOperator.createNull(skewJoinColumn.getType()));
+                } else {
+                    // Use MCV-based skew values
+                    skewValues = leftChildMCV.stream().map(pair -> ConstantOperator.createVarchar(pair.first)).
+                            collect(Collectors.toList());
+                }
+                joinOperator.setSkewValues(skewValues);
                 return true;
             }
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SkewJoinOptimizeRule.java
```diff
@@ -416,6 +441,10 @@ private OptExpression createSkewValueSaltTable(List<ScalarOperator> skewValues,
     private OptExpression addSaltForRightChild(LogicalJoinOperator oldJoinOperator, OptExpression input,
                                                ScalarOperator rightSkewColumn, OptimizerContext context) {
         List<ScalarOperator> skewValues = oldJoinOperator.getSkewValues();
+        // If skew values is empty or contains only NULL, still proceed to create salt table.
```
m-selmi, line 444: would this mean that even if nulls are not skewed, we trigger the salting mechanism (if other values are skewed)?

fe/fe-core/src/test/java/com/starrocks/sql/plan/SkewJoinTest.java
```diff
@@ -324,6 +330,56 @@ public void testIntSkewColumnVarchar() throws Exception {
         assertCContains(sqlPlan, "C_NATIONKEY IN (22, 23, 24, 10, 11)");
     }
 
+    @Test
+    public void testSkewJoinWithNullOnlySkewByStats() throws Exception {
```
m-selmi, line 334: We could add a test where both null and some other values are skewed.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SkewJoinOptimizeRule.java
```diff
@@ -167,22 +167,42 @@ public boolean check(OptExpression input, OptimizerContext context) {
...
+        
+        // Check MCV-based skew
         long mcvRowCount = mcvList.stream().mapToLong(pair -> pair.second).sum();
-        return ((double) mcvRowCount / rowCount) > sessionVariable.getSkewJoinDataSkewThreshold();
+        boolean mcvSkew = ((double) mcvRowCount / rowCount) > sessionVariable.getSkewJoinDataSkewThreshold();
+        
+        // Check NULL-based skew - treat high NULL concentration as skew
+        boolean nullSkew = nullsFraction > sessionVariable.getSkewJoinDataSkewThreshold();
+        
+        return mcvSkew || nullSkew;
```
m-selmi, line 205: why are other values handled by summing their occurrence fractions but null is handled separately?

m-selmi (commented):
Thanks for creating this :+1:  is there anything missing to move this from draft to open?

PR #62542, opened 2025-08-31
[BugFix] Fix redundant replica handling after clone

### LINKED ISSUE #62541: [Bug] New BE added to an existing rack cannot get balanced data
**Environment / Setup**

- Cluster with 3 racks: rack:rack1, rack:rack2, rack:rack3
- Each rack contains 4 BEs initially.
- A new BE is added to rack:rack1.

**Problem**

- When the new BE is added to rack1, data cannot be rebalanced to the new BE.
- Although the scheduler triggers clone tasks to move data to the new BE, the cloned replicas are always deleted afterwards.

**Expected Behavior**

The new BE in rack1 should eventually receive replicas from other BEs in the same rack and participate in balancing, instead of having all cloned replicas removed.

**Actual Behavior**

- Cloned replicas are created on the new BE.
- Shortly after, those replicas are deleted as redundant.
- The new BE remains empty and cannot join the balance.

**Steps to Reproduce**

- Prepare a cluster with multiple racks (e.g., 3 racks with 4 BEs each).
- Add a new BE to an existing rack (e.g., rack1).
- Observe the balance process — cloned replicas on the new BE are deleted and data never gets rebalanced onto it.

hongkunxu:
## Why I'm doing:
Under the current rebalance + deletion mechanism, when a new BE node is added, data cannot be successfully migrated to it.

Here’s why:
- During rebalance, a tablet replica is cloned from a source BE to the new BE.
- At this moment, the source BE still holds the replica, so the system immediately treats the situation as redundant.
- Since the source BE’s replica is already recorded in cachedReplicaId and scheduled for deletion, the cloned replica on the new BE is also incorrectly judged as redundant and then removed.

As a result, the new BE node never keeps the migrated data, making the rebalance ineffective.

## What I'm doing:
I added an extra check in the redundant replica detection logic:
1. Before deciding whether a replica is redundant, we now verify if its tabletId already exists in cachedReplicaId.
2. If it does, we skip the redundant replica judgment because the replica is already scheduled for cleanup.

Fixes [#62541](https://github.com/StarRocks/starrocks/issues/62541)

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/clone/TabletScheduler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/clone/TabletScheduler.java
@@ -1176,8 +1176,14 @@ public class TabletScheduler extends FrontendDaemon {
 
         Set<Pair<String, String>> matchedLocations = new HashSet<>();
         Replica dupReplica = null;
-        //1. delete the unmatched replica
-        for (Replica replica : tabletCtx.getReplicas()) {
+        // 1. delete the unmatched replica
+        // To ensure the correctness of the cleanup process after a rebalance, we iterate the replicas in reverse order.
+        // After cloning a replica from the source BE to a newly added destination BE, the replica on the new BE will
+        // always be at the end of the list. By traversing the list in reverse, we process newer replicas first
+        // and preserve the newly added replica, preventing it from being mistakenly deleted.
+        List<Replica> replicas = tabletCtx.getReplicas();
+        for (int i = replicas.size() - 1; i >= 0; i--) {
+            Replica replica = replicas.get(i);
             if (!TabletChecker.isLocationMatch(replica.getBackendId(), tabletCtx.getRequiredLocation())) {
                 deleteReplicaInternal(tabletCtx, replica, "location mismatch", force);
                 return true;


fe/fe-core/src/main/java/com/starrocks/clone/TabletScheduler.java
```diff
@@ -1185,7 +1185,13 @@ private boolean deleteLocationMismatchReplica(TabletSchedCtx tabletCtx, boolean
                 if (backend != null) {
                     Pair<String, String> location = backend.getSingleLevelLocationKV();
                     if (location != null && matchedLocations.contains(location)) {
-                        dupReplica = replica;
```
wyb: how about iterating replicas backwards and remove the replica with duplicate location?

hongkunxu: Hi @wyb Iterating replicas backwards and removing the replica with duplicate location is indeed one solution, but in that case, wouldn’t this mean that
```
// tabletId -> replicaId
// used to delete src replica after copy task success
private final Map<Long, Long> cachedReplicaId = new ConcurrentHashMap<>();
```

becomes useless? This set operation only happens in completeSchedCtx inside createBalanceTask. And since rebalance will inevitably cause duplicate replicas, if we clean them up directly here, doesn’t that mean the subsequent **deleteReplicaChosenByRebalancer** can also be removed?

wyb: current implement doesn't guarantee that the source replica must be deleted via `deleteReplicaChosenByRebalancer` after balancing.

hongkunxu: Hi @wyb I have adopted your suggestion and tested it in the production environment. Both approaches work well.

fe/fe-core/src/main/java/com/starrocks/clone/TabletScheduler.java
```diff
@@ -1176,8 +1176,14 @@ private boolean deleteLocationMismatchReplica(TabletSchedCtx tabletCtx, boolean
...
         Set<Pair<String, String>> matchedLocations = new HashSet<>();
         Replica dupReplica = null;
-        //1. delete the unmatched replica
-        for (Replica replica : tabletCtx.getReplicas()) {
+        // 1. delete the unmatched replica
+        // To ensure the correctness of the cleanup process after a rebalance, we iterate the replicas in reverse order.
+        // After cloning a replica from the source BE to a newly added destination BE, the replica on the new BE will
+        // always be at the end of the list. By traversing the list in reverse, we process older replicas first
+        // and preserve the newly added replica, preventing it from being mistakenly deleted.
+        List<Replica> replicas = tabletCtx.getReplicas();
```
gengjun-git, line 1184: The larger the index, the newer the replica should be when it joins, right? Why is it older instead?

hongkunxu: Hi @gengjun-git 
This replica set is managed with a list, where each new addition is added to the end.
For example, the original List<Replica> is:
- be0 (rack1)
- be1 (rack2)
- be2 (rack3)

After adding a new backend be10 (rack1) and performing rebalance, the List<Replica> becomes:

be0, be1, be2, be10

With the original traversal order, the result is:

- rack1: be0
- rack2: be1
- rack3: be2

Then, when reaching be10 (rack1), it detects that rack1 already exists, so the newly migrated replica on be10 gets removed.

gengjun-git: OK, I get you, But I think the comments mislead me:
`By traversing the list in reverse, we process older replicas first and preserve the newly added replica, preventing it from being mistakenly deleted.`
should be:
`By traversing the list in reverse, we process newer replicas first and preserve the newly added replica, preventing it from being mistakenly deleted.`

older => newer

hongkunxu: HI @gengjun-git ，I've modified the comment. Please help review again.

gengjun-git (approved):

alvin-phoenix-ai:
@cursor review

wyb (approved):

PR #62690, opened 2025-09-03
[Feature] Mapped ldap groups with db roles if exists

### LINKED ISSUE #62689: Map db roles with LDAP User groups if exists
## Feature request

Currently, StarRocks requires creating users manually and then mapping them to roles for authorization when using LDAP integration. There is a Ranger plugin for authorization, but if StarRocks could directly map existing database roles to LDAP user groups (when the names match), it would significantly reduce the need for manual setup.

Allow StarRocks to automatically associate LDAP user groups with existing StarRocks roles (by name or a configurable mapping). This way, when a user logs in through LDAP, they inherit the correct role permissions without requiring manual user creation and role assignment.

	•	Using Apache Ranger for role and authorization management (heavier setup).
	•	Continuing with manual user creation and mapping, which is error-prone and doesn’t scale well.

**Additional context**
	•	This feature would simplify LDAP integration and improve security management.
	•	It provides a lightweight alternative to Ranger for organizations that only need group-to-role mapping.

vivek807:
## Why I'm doing:
Currently, StarRocks requires manual user creation and role assignment when using LDAP integration. This is not scalable and creates overhead for managing security. While Apache Ranger can solve this, it’s a heavy solution for teams who only need a simple mapping between LDAP groups and StarRocks roles.

By enabling direct mapping of LDAP user groups to StarRocks roles (when the group exists in StarRocks), we can simplify security integration and remove the need for repetitive manual steps.

## What I'm doing:
	•	Adding support to map LDAP user groups directly to StarRocks roles if they exist.
	•	Ensuring that when a user authenticates via LDAP, their roles in StarRocks are automatically inferred from their LDAP group memberships.

Fixes #62689

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/authorization/AuthorizationMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/authorization/AuthorizationMgr.java
@@ -54,6 +54,7 @@ import com.starrocks.sql.ast.RevokePrivilegeStmt;
 import com.starrocks.sql.ast.RevokeRoleStmt;
 import com.starrocks.sql.ast.UserRef;
 import com.starrocks.sql.parser.NodePosition;
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
@@ -67,6 +68,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


eshishki:
@vivek807 this has merge conflicts now (`DIRTY`) and hasn't had a review since it was opened — could you rebase onto current `main` when you get a chance?

@HangyuanLiu this proposes auto-mapping LDAP groups to StarRocks roles by name as a lightweight alternative to Ranger, and it's a behavior-changing feature (auto-enabled). Could you give a direction on whether this is something StarRocks wants before the code gets reviewed in detail? Tests and docs are still missing, so a yes/no on the approach first would save @vivek807 the effort if it's not wanted.

PR #53916, opened 2024-12-13
[Enhancement] check partition scan number limit in resource group

kaijianding:
## Why I'm doing:

Want to limit the number of scanned partitions to prevent starrocks from heavy load

## What I'm doing:
Fixes #issue

Add a limit rule in resource group.
eg :
image

`partition_scan_number_limit_rule: {"db_1.tbl_1":7, "db_1.tbl_2":10}`

If table scan in a  single sub-query exceeds the limit, the query will be rejected.

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 4.1
  - [x] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1
  - [ ] 3.0

PR #53747, opened 2024-12-10
[Enhancement] add partition scan num limit when query internal olap table

MatthewH00:
## Why I'm doing:
when query big size internal olap table with full table scan or scan too many partitions, would cause BE/CN node high load, lead to cluster instability.

## What I'm doing:
add a new FE session variable `scan_olap_partition_num_limit` to limit partition scan num when query internal olap table.
(default value is 0, means no limitation)

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [x] Yes, this PR will result in a change in behavior.
- [ ] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [x] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1
  - [ ] 3.0

PR #62705, opened 2025-09-03
[Enhancement] Add partition limit functionality for RESOURCE GROUP

crossoverJie:
## What I'm doing:

Fixes https://github.com/StarRocks/starrocks/issues/62704

Add two parameters to the `RESOURCE GROUP`:
- `partition_num`: Limit on the Number of Partitions in a `Single Table`.
- `total_partition_num`: This query provides the cumulative partition limit for all tables.

```sql
ALTER RESOURCE GROUP rg1 WITH (
      'partition_num' = '3',
      'total_partition_num' = '5',
);
```

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3


crossoverJie:
@ZiheLiu Please take a look.

alvin-phoenix-ai:
@cursor review

kevincai:
A similar PR: #53916, and PR already merged: #53747

crossoverJie:
> A similar PR: #53916, and PR already merged: #53747

This proposal is different from #53747, but similar to #53916.

We can set different resource group for different users, but in our scenario, we only need to limit single tables or partitions of the entire query.


I think setting limits based on specific tables is a bit complicated. A more common approach would be to assign resource group when creating a new user, rather than adding limits after problems have already occurred.

Also, the tables a user accesses can change dynamically, so setting limits per table seems somewhat complex.

@kaijianding WDYT

kaijianding:
> > A similar PR: #53916, and PR already merged: #53747
> 
> This proposal is different from #53747, but similar to #53916.
> 
> We can set different resource group for different users, but in our scenario, we only need to limit single tables or partitions of the entire query.
> 
> I think setting limits based on specific tables is a bit complicated. A more common approach would be to assign resource group when creating a new user, rather than adding limits after problems have already occurred.
> 
> Also, the tables a user accesses can change dynamically, so setting limits per table seems somewhat complex.
> 
> @kaijianding WDYT

In our senario, some tables are big which should be limitted when scan no matter which user, but some tables are small table which can be scanned freely. The purpose of setting limitation to big tables is to encourage user to create materialized view for purticular sqls with big tables to avoid accessing the big tables directly

eshishki:
@ZiheLiu there are two open resource-group PRs adding partition-scan limits that overlap and have been stuck without a direction — #62705 (this one, @crossoverJie) and #53916 (@kaijianding). Both differ from the already-merged session variable `scan_olap_partition_num_limit` (#53747) by enforcing the limit at the **resource-group** level, so users can't disable it themselves; all three throw an explicit error on exceed (no silent truncation).

The two differ only in granularity:
- **#62705**: uniform numbers per RG — `partition_num` (any single table) + `total_partition_num` (whole query).
- **#53916**: a per-named-table rule (`{table: limit}`) — limit specific big tables, leave small ones free.

Having two separate RG properties for the same concept seems undesirable. Could you give a direction on which shape to take (or whether a single more general one subsumes both)? Both authors are active and waiting — #53916 has been open ~17 months and this one ~9.

PR #62717, opened 2025-09-03
[Enhancement] Support HH:mm format for cast string to time

duanyyyyyyy:
## Why I'm doing:
Currently when we cast('HH:mm' as time) will return null, but actually speaking we can make it return HH:mm:00.
Here I will implement it.
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exprs/cast_expr.cpp
+++ b/be/src/exprs/cast_expr.cpp
@@ -981,25 +981,29 @@ static ColumnPtr cast_from_string_to_time_fn(ColumnPtr& column) {
         int hour = 0, minute = 0, second = 0;
         char* first_colon = (char*)memchr(first_char, ':', time.size);
         if (first_colon != nullptr) {
+            // Parse HH format
+            StringParser::ParseResult parse_result = StringParser::PARSE_SUCCESS;
+            auto int_value = StringParser::string_to_unsigned_int<uint64_t>(reinterpret_cast<char*>(first_char),
+                                                                            first_colon - first_char, &parse_result);
+            if (UNLIKELY(parse_result != StringParser::PARSE_SUCCESS)) {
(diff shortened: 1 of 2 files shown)


alvin-phoenix-ai:
@cursor review

duanyyyyyyy:
@alvin-celerdata Any other comments?

wangsimo0 (changes requested):
Thanks for working on this, but I think this PR needs to step back and provide a proper product/semantic design before we decide on the implementation.

Right now the PR only changes one specific behavior: `CAST('HH:mm' AS TIME)` becomes `HH:mm:00`. However, I do not see the design thinking behind this change, and I am worried that we are patching one case without defining the full contract of string-to-TIME conversion.

My concern is not the implementation complexity of this PR. The concern is that we are changing a SQL type conversion rule without first defining the product contract. I would prefer that we pause this PR, write down the intended TIME parsing semantics, review that design, and then decide whether `CAST` is the right surface or whether this should be handled by an explicit parsing function.

eshishki:
@duanyyyyyyy what's the motivation here? `cast('12:30:45' as time)` already works, so this only adds the seconds-less `HH:mm` form. Is it for compatibility with MySQL / another engine (if so, a reference to the expected behavior would help), or driven by a specific workload? A concrete use case in the description would help a reviewer decide whether to take it.

PR #62826, opened 2025-09-08
[BugFix] mutate input columns in functions' returning value

silverbullet233:
## What I'm doing:

Fixes #62828
some functions may return the input column directly as result, because the reused result column and its source are the same underlying Column, the filter can be applied multiple times to the same data. Since filtering is in-place and not idempotent, this produces wrong results. 

we should return MutatePtr in such cases.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3


LiShuMing (approved):

satanson (approved):

silverbullet233:
@mergify backport branch-3.4

PR #62860, opened 2025-09-09
[BugFix] call addWorker on non-leaders to handle stale cached workerid issues when CN are restarted

ctbrennan:
## Why I'm doing:
In shared-data deployments, Observer nodes experienced stale `StarOSAgent.workerToId` issues that caused queries to use backup Compute Nodes instead of optimal primary nodes, resulting in performance degradation.
Problem Timeline:
1. Compute Nodes restart and receive new WorkerID assignments from StarMgr
2. Leader FE processes heartbeats and updates its internal cache with new WorkerIDs
3. Non-leader FEs replay heartbeat events from the journal but skip cache updates due to replay logic
4. Non-leader FE cache contains outdated WorkerID mappings
5. Query scheduling uses stale WorkerIDs, causing primary CNs to appear unavailable
6. System falls back to backup CN selection, impacting query performance
Symptoms observed:
* Queries routed to Observer FEs perform poorly due to backup CN usage
* Issue persists until Observer FE restart (which clears the stale cache)
* Inconsistent WorkerID values between Leader and Observer FEs
## What I'm doing:
Modified heartbeat processing logic to allow FEs to update their WorkerID cache during journal replay, ensuring all FE nodes maintain synchronized compute node mappings.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/system/HeartbeatMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/system/HeartbeatMgr.java
@@ -220,10 +220,10 @@ public class HeartbeatMgr extends FrontendDaemon {
                     } else {
                         if (RunMode.isSharedDataMode() && 
                                 (!isReplay || !GlobalStateMgr.getCurrentState().isLeader())) {
-                            // Observer FEs maintain StarOSAgent worker map fields via journal replay.
-                            // When CNs restart, they get new WorkerIds. Observer FEs (which can never
-                            // become leaders) must refresh their cached workers during replay to avoid incorrectly
-                            // identifying primary CNs as unavailable and triggering backup CN selection.
+                            // Non-leader FEs maintain StarOSAgent worker map fields via journal replay.
+                            // When CNs restart, they get new WorkerIds. Non-leader FEs must refresh
+                            // their cached workers during replay to avoid incorrectly identifying primary
+                            // CNs as unavailable and triggering backup CN selection.
                             int starletPort = computeNode.getStarletPort();
                             if (starletPort != 0) {
                                 String workerAddr = NetUtils.getHostPortInAccessibleFormat(computeNode.getHost(),
--- a/fe/fe-core/src/main/java/com/starrocks/system/HeartbeatMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/system/HeartbeatMgr.java
@@ -188,7 +188,8 @@ public class HeartbeatMgr extends FrontendDaemon {
         setInterval(Math.max(1L, Config.heartbeat_timeout_second * 1000L - (System.currentTimeMillis() - startTime)));
     }
 
-    private boolean handleHbResponse(HeartbeatResponse response, boolean isReplay) {
+    // package-private for testing
+    boolean handleHbResponse(HeartbeatResponse response, boolean isReplay) {
         switch (response.getType()) {
             case FRONTEND: {
                 FrontendHbResponse hbResponse = (FrontendHbResponse) response;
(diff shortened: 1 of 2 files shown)


alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/system/HeartbeatMgr.java
```diff
@@ -217,8 +218,12 @@ private boolean handleHbResponse(HeartbeatResponse response, boolean isReplay) {
                                     .abortTxnWhenCoordinateBeDown(computeNode.getHost(), 100);
                         }
                     } else {
-                        if (RunMode.isSharedDataMode() && !isReplay) {
-                            // addWorker
+                        if (RunMode.isSharedDataMode() && 
+                                (!isReplay || !GlobalStateMgr.getCurrentState().isLeader())) {
+                            // Observer FEs maintain StarOSAgent worker map fields via journal replay.
```
kevincai: is it just for Observer, or the Follower will have the same issue?

starrocks-xupeng:
why would step 1 occur? if worker restarts, it worker id should not change

fe/fe-core/src/main/java/com/starrocks/system/HeartbeatMgr.java
ctbrennan: Good point, I think the followers would have the same issue.

ctbrennan:
> why would step 1 occur? if worker restarts, it worker id should not change
When are new worker ids assigned? I thought it might be assigned a new worker id by StarMgr in `init_staros_worker` https://github.com/StarRocks/starrocks/blob/77a4dac33b199576aa70b77a5b9e29db1c1e187a/be/src/service/staros_worker.cpp#L510

I don't have a perfect understanding, but I've observed the following
1. I restart a CN. FE start using backup CN (due to unavailability I think)
2. I restart 1 FE. That FE starts showing a different workerid for the restarted CN compared to other FE, it stops needing to use backup CN
3. I restart the other FE, they start showing the same workerId for the restarted CN

kevincai:
> > why would step 1 occur? if worker restarts, it worker id should not change
> > When are new worker ids assigned? I thought it might be assigned a new worker id by StarMgr in `init_staros_worker` https://github.com/StarRocks/starrocks/blob/77a4dac33b199576aa70b77a5b9e29db1c1e187a/be/src/service/staros_worker.cpp#L510
> 
> I don't have a perfect understanding, but I've observed the following
> 
> 1. I restart a CN. FE start using backup CN (due to unavailability I think)
> 2. I restart 1 FE. That FE starts showing a different workerid for the restarted CN compared to other FE, it stops needing to use backup CN
> 3. I restart the other FE, they start showing the same workerId for the restarted CN

If running in kubernetes environment, it is because the CN node is automatically dropped by the cn_entrypoint.sh when the POD is killed, and the same FQDN name will be added when the POD is started. This is a specific case in kubernetes environment.

fe/fe-core/src/main/java/com/starrocks/system/HeartbeatMgr.java
```diff
@@ -217,8 +218,12 @@ private boolean handleHbResponse(HeartbeatResponse response, boolean isReplay) {
                                     .abortTxnWhenCoordinateBeDown(computeNode.getHost(), 100);
                         }
                     } else {
-                        if (RunMode.isSharedDataMode() && !isReplay) {
-                            // addWorker
+                        if (RunMode.isSharedDataMode() &&
```
kevincai, line 221: This is not right. The worker shall not be added again into starmgr since the leader already did it.

And during the replayDropBackends/replayDropComputeNodes, the `removeWorkerFromMap()` will be called to invalidate the `workerId<->backendId` map. So actually I don't still get the root cause.

ctbrennan: Right, this bug isn't about adding the worker to starmgr, it's about maintaining the `workerToId` and `workerToNode` maps correctly in StarOSAgent, that's what calling `StarOSAgent.addWorker()` here is doing. We expect `StarOSAgent.addWorker()` handles `ALREADY_EXIST` gracefully. 

> And during the replayDropBackends/replayDropComputeNodes, the removeWorkerFromMap() will be called to invalidate the workerId<->backendId map. So actually I don't still get the root cause.

This is probably the crux of the issue. When I restart the CN process, we're not executing a `DROP COMPUTE NODE`. In general I think it would be better if Starrocks were flexible about allowing compute node processes to restart without requiring them to be dropped and re-added to the cluster with `ALTER SYSTEM` statements. But if it's the explicit policy that any time the CN process stops it must be dropped explicitly, then we can make a change on our end.

kevincai: If that is the case, I can't think of any reason the workerId changed after CN restarted. There is NO such design decision that every time the CN service get restarted, the assigned workerId must be rotated.

Need to find out why the workerId changed and then fix the issue. This band-aid code change seems unreasonable to me for now.

kevincai:
@ctbrennan please let me know if you have any comments on this, otherwise this PR shall be closed as invalid.

kevincai:
close as no update.

PR #62916, opened 2025-09-10
[BugFix] Gracefully Shutdown Compute Node on Exit

### LINKED ISSUE #62915: Compute Node does not gracefully exit and deregister from FE Node on shutdown
Run Compute Node deployment [with StarRocks operator](https://github.com/StarRocks/starrocks-kubernetes-operator) using the default configuration (i.e no overrides on pre-stop hooks or other scripts).

Scale down the number of compute nodes running by changing the value of `starRocksCnSpec.replicas` to something less on the StarRocksCluster object definition, for example from `3` → `2`.

Wait for the node to shutdown and terminate. Observe that the Compute Node that was just terminated is still registered with the FE leader node but with StatusCode `DISCONNECTED`. This can be seen by running `SHOW COMPUTE NODES`. Additionally, the StarRocks operator node_info metric for the `cn_node_num` type dimension will report that the number of nodes in the `total` state is greater than the number of nodes in the `alive` state indicating there is an issue with one or more compute nodes. Finally, observe the repetitive logs from the FE leader node indicating that it is trying to and cannot resolve the address for the Compute Node that was just shut down. This should not happen as the FE should recognise that the Compute Node was gracefully shut down and hence it should not try and connect to it.

```sh
 WARNING: [Channel<313102>: (a-starrocks-in-share-data-mode-cn-2.a-starrocks-in-share-data-mode-cn-search.starrocks.
     at io.grpc.internal.DnsNameResolver.resolveAddresses(DnsNameResolver.java:223)                                 
     at io.grpc.internal.DnsNameResolver.doResolve(DnsNameResolver.java:282)                                        
     at io.grpc.grpclb.GrpclbNameResolver.doResolve(GrpclbNameResolver.java:63)                                     
     at io.grpc.internal.DnsNameResolver$Resolve.run(DnsNameResolver.java:318)                                      
     at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)                   
     at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)                   
     at java.base/java.lang.Thread.run(Thread.java:829)                                                              
 Caused by: java.net.UnknownHostException: a-starrocks-in-share-data-mode-cn-2.a-starrocks-in-share-data-mode-cn-sea 
     at java.base/java.net.InetAddress$CachedAddresses.get(InetAddress.java:797)                                     
     at java.base/java.net.InetAddress.getAllByName0(InetAddress.java:1533)                                          
     at java.base/java.net.InetAddress.getAllByName(InetAddress.java:1386)                                           
     at java.base/java.net.InetAddress.getAllByName(InetAddress.java:1307)                                           
     at io.grpc.internal.DnsNameResolver$JdkAddressResolver.resolveAddress(DnsNameResolver.java:632)                 
     at io.grpc.internal.DnsNameResolver.resolveAddresses(DnsNameResolver.java:219)                                  
     ... 6 more                                                                                                      
 }
```

We expect that the compute node does a clean shutdown and de-registers itself from the leader FE Node. The FE node should not try and poll the compute node as it should be aware that it was removed and the operator metrics should indicate that all running nodes are healthy.

On a graceful exit of the compute node, when inspecting the logs, we should expect to see the entrypoint cleanup script run which does the deregistering https://github.com/StarRocks/starrocks/blob/main/docker/dockerfiles/be/cn_entrypoint.sh#L140 and hence see the logs indicating so.

```sh
E20250904 03:52:07.084132 139650349562560 daemon.cpp:259] got signal: Terminated from pid: 1130481248, is going to exit
...
[Thu Sep  4 03:52:16 UTC 2025[] Receives signal to exit ...
[Thu Sep  4 03:52:16 UTC 2025[] Can't find /opt/starrocks/cn/bin/cn.pid!
[Thu Sep  4 03:52:16 UTC 2025[] try to drop myself(a-starrocks-in-share-data-mode-cn-2.a-starrocks-in-share-data-mode-cn-search.starrocks.svc.cluster.local) from FE ...
[Thu Sep  4 03:52:16 UTC 2025[] drop my self a-starrocks-in-share-data-mode-cn-2.a-starrocks-in-share-data-mode-cn-search.starrocks.svc.cluster.local:9050 ...
stream closed EOF for starrocks/a-starrocks-in-share-data-mode-cn-2 (cn)
```

### Real behavior (Required)

In versions 3.4.6+ we observe that the main application receives the `SIGTERM` but the entrypoint process does not and hence we are missing those logs indicating it is deregistering itself from the FE leader node.

```sh
E20250904 03:34:21.817762 134151544224512 daemon.cpp:259] got signal: Terminated from pid: 1606223648, is going to exit
...
stream closed EOF for starrocks/a-starrocks-in-share-data-mode-cn-2 (cn)
```

The bug was seemingly introduced in the following PR https://github.com/StarRocks/starrocks/pull/60481 which introduces a while loop to hang in the `stop_cn.sh` script which is run as a preStop hook in k8s.

In particular, the behaviour exhibited is that the script signals the main application to shut down, then enters a while loop that checks if the application has exited and `sleep`s for a second in between each check. The expected behaviour is that once the main application exits then [the preStop hook exits, thereby sending a SIGTERM to PID 1 ](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks)(the entrypoint) to shut down as the child process has shut down. However, what actually happens is that as soon as the child process (main application) shuts down, it immediately returns and exits the entrypoint https://github.com/StarRocks/starrocks/blob/main/docker/dockerfiles/be/cn_entrypoint.sh#L188, thereby never encountering the SIGTERM and triggering the trap that does a graceful shutdown https://github.com/StarRocks/starrocks/blob/main/docker/dockerfiles/be/cn_entrypoint.sh#L178. The suspicion is that the 1 second delay in the while loop between each check makes the logic error-prone to this race condition. There is not enough time for the preStopHook to finish running and send the `SIGTERM` to the entrypoint. Instead the entrypoint immediately exits and the preStopHook never finishes running.

You can confirm this behaviour as well by checking the system events of the pod which indicates that the PreStopHook fails. This is because once the main application is terminated and the entrypoint exits, the preStopHook is forcefully terminated as the container will no longer be running.

```
Events:
  Type     Reason             Age              From               Message
  ----     ------             ----             ----               -------
  Normal   Scheduled          37s              default-scheduler  Successfully assigned starrocks/a-starrocks-in-share-data-mode-cn-1 to ip-10-7-48-17.ap-southeast-2.compute.internal
  Normal   Pulled             38s              kubelet            Container image "starrocks/cn-ubuntu:3.4.7" already present on machine
  Normal   Created            38s              kubelet            Created container: cn
  Normal   Started            38s              kubelet            Started container cn
  Normal   Pulled             38s              kubelet            Container image "644975362233.dkr.ecr.ap-southeast-2.amazonaws.com/kitt/iam-sidecar:2.2.0" already present on machine
  Normal   Created            38s              kubelet            Created container: iam-sidecar
  Normal   Started            37s              kubelet            Started container iam-sidecar
  Normal   Killing            10s              kubelet            Stopping container cn
  Normal   Killing            10s              kubelet            Stopping container iam-sidecar
  Warning  Unhealthy          2s (x2 over 7s)  kubelet            Readiness probe failed: Get "http://10.7.47.109:8040/api/health": dial tcp 10.7.47.109:8040: connect: connection refused
  Warning  FailedPreStopHook  1s               kubelet            PreStopHook failed
```

Previously, before the introduced change, the preStop hook script would signal to terminate the main application then immediately exit, sending a SIGTERM to PID 1 which is the `cn_entrypoint.sh`. As this occurs immediately and the shutdown of the main application may take some time, the `trap` is always triggered. The `exit_clean` function which is triggered waits for the child process to shutdown before proceeding with the rest of the shut down process https://github.com/StarRocks/starrocks/blob/main/docker/dockerfiles/be/cn_entrypoint.sh#L154 and so gracefully shuts down.

### Suggested Fix

I will raise a PR per the suggested fix here but am open to other suggestions. The proposal here would be to modify the trap to also respect clean exits i.e trap exit_clean SIGTERM EXIT. This should be the expected behaviour as the trap is injected after registering itself with the FE node https://github.com/StarRocks/starrocks/blob/main/docker/dockerfiles/be/cn_entrypoint.sh#L177 so on any exits, it should always do a cleanup.

The change has been tested in our cluster and resolves the issue described above.

3.4.7 - The bug looks like it was introduced in 3.4.6 based on testing versions prior and after and observing the behaviour.

Tenaria:
## Why I'm doing:

See full context in below issue.

## What I'm doing:

Fixes https://github.com/StarRocks/starrocks/issues/62915

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/docker/dockerfiles/be/cn_entrypoint.sh
+++ b/docker/dockerfiles/be/cn_entrypoint.sh
@@ -175,7 +175,7 @@ fi
 update_conf_from_configmap
 collect_env_info
 add_self $svc_name || exit $?
-trap exit_clean SIGTERM
+trap exit_clean SIGTERM EXIT
 
 log_stderr "run start_cn.sh"
 


alvin-phoenix-ai:
@cursor review

kevincai (approved):

alvin-phoenix-ai (approved):

kevincai:
This PR will introduce  behavior change, before the PR, only SIGTERM causes the deregister of the cn node from FE, after the change, no matter what reason, the CN will be registered every time the POD exits.

There are two different scenarios of CN pod exit.
1. POD restart (due to config update or version upgrade) or crash 
2. POD scale in

It is not necessary to deregister the CN POD in the first scenario while it is needed in the second scenario.

Unnecessary deregister of CN POD causes the CN POD workerId/backendId change, and causes unnecessary tablets scheduling and shuffling, and also cache eviction.

So will revert this PR and rework it on operator side to distinguish between the regular restarts and scaling in. Operator will take over the deregister responsibility.

cc @yandongxiao

PR #62929, opened 2025-09-10
[Enhancement] Full sort use german string for comparison

satanson:
## Why I'm doing:

Full sort using german string for comparison can accelerate queries if the string-type order-by column has different prefix 4 bytes.

use sv enable_full_sort_use_german_string to control this behavior, it is true in default.

for queries constructed on click_bench as follows
```
-- Q1
with cte as( select MobilePhoneModel, count(1) over(partition by MobilePhoneModel) as cnt from hits) 
select /*+SET_VAR(enable_full_sort_use_german_string=true)*/ sum(murmur_hash3_32(MobilePhoneModel, cnt)) from cte
-- Q2
with cte as( select URL, count(1) over(partition by URL) as cnt from hits) 
select /*+SET_VAR(enable_full_sort_use_german_string=true)*/ sum(murmur_hash3_32(URL, cnt)) from cte
-- Q3
with cte as( select SearchPhrase, count(1) over(partition by SearchPhrase) as cnt from hits) 
select /*+SET_VAR(enable_full_sort_use_german_string=true)*/ sum(murmur_hash3_32(SearchPhrase, cnt)) from cte
-- Q4
with cte as( select Title, count(1) over(partition by Title) as cnt from hits) 
select /*+SET_VAR(enable_full_sort_use_german_string=true)*/ sum(murmur_hash3_32(Title, cnt)) from cte
-- Q5
with cte as( select Referer, count(1) over(partition by Referer) as cnt from hits) 
select /*+SET_VAR(enable_full_sort_use_german_string=true)*/ sum(murmur_hash3_32(Referer, cnt)) from cte
-- Q6
with cte as( select MobilePhoneModel, SearchPhrase, count(1) over(partition by MobilePhoneModel, SearchPhrase) as cnt from hits) 
select /*+SET_VAR(enable_full_sort_use_german_string=true)*/ sum(murmur_hash3_32(MobilePhoneModel, SearchPhrase, cnt)) from cte
-- Q7
with cte as( select MobilePhoneModel, URL, count(1) over(partition by MobilePhoneModel, URL) as cnt from hits) 
select /*+SET_VAR(enable_full_sort_use_german_string=true)*/ sum(murmur_hash3_32(MobilePhoneModel, URL, cnt)) from cte
```

The test result 
image

Performances of Q2, Q5, Q7 decrease significantly since the order-by column URL and Referer are almost the same.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/column/binary_column.h
+++ b/be/src/column/binary_column.h
@@ -17,6 +17,7 @@
 #include "column/bytes.h"
 #include "column/column.h"
 #include "column/datum.h"
+#include "column/german_string.h"
 #include "column/vectorized_fwd.h"
 #include "common/statusor.h"
 #include "gutil/strings/fastmem.h"
@@ -48,6 +49,7 @@ public:
     };
 
     using Container = Buffer<Slice>;
+    using GermanStringContainer = Buffer<GermanString>;
     using ProxyContainer = BinaryDataProxyContainer;
     using ImmContainer = BinaryDataProxyContainer;
 
@@ -304,6 +306,20 @@ public:
         return _slices;
     }
 
+    GermanStringContainer& get_german_strings() {
+        if (!_german_strings_cache) {
+            _build_german_strings();
+        }
+        return _german_strings;
+    }
+
+    const GermanStringContainer& get_german_strings() const {
+        if (!_german_strings_cache) {
+            _build_german_strings();
+        }
+        return _german_strings;
+    }
+
     const BinaryDataProxyContainer& get_proxy_data() const { return _immuable_container; }
 
     Bytes& get_bytes() { return _bytes; }
@@ -342,7 +358,10 @@ public:
         _slices_cache = false;
     }
 
-    void invalidate_slice_cache() { _slices_cache = false; }
+    void invalidate_slice_cache() {
+        _slices_cache = false;
+        _german_strings_cache = false;
+    }
 
     std::string debug_item(size_t idx) const override;
 
@@ -366,12 +385,16 @@ public:
 
 private:
     void _build_slices() const;
+    void _build_german_strings() const;
 
     Bytes _bytes;
     Offsets _offsets;
 
     mutable Container _slices;
     mutable bool _slices_cache = false;
+    mutable GermanStringContainer _german_strings;
+    mutable bool _german_strings_cache = false;
+
     BinaryDataProxyContainer _immuable_container = BinaryDataProxyContainer(*this);
 };
 
--- a/be/src/column/german_string.cpp
+++ b/be/src/column/german_string.cpp
@@ -1,0 +1,103 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "column/german_string.h"
+
+#include "gutil/strings/fastmem.h"
+#include "util/hash_util.hpp"
+#include "util/misc.h"
+#include "util/raw_container.h"
+#include "util/slice.h"
+
+namespace starrocks {
+
+GermanString::GermanString() {
+    auto* p = reinterpret_cast<char*>(this);
+    std::fill(p, p + sizeof(GermanString), 0);
+}
+
+GermanString::GermanString(const starrocks::GermanString& rhs) {
+    strings::memcpy_inlined(this, &rhs, sizeof(GermanString));
+}
+
+GermanString::GermanString(const char* str, size_t len, void* ptr) {
+    if (len <= INLINE_MAX_LENGTH) {
+        auto* p = reinterpret_cast<char*>(this);
+        std::fill(p, p + sizeof(GermanString), 0);
+        strings::memcpy_inlined(short_rep.str, str, len);
+    } else {
+        strings::memcpy_inlined(long_rep.prefix, str, PREFIX_LENGTH);
+        strings::memcpy_inlined(ptr, str, len);
+        long_rep.ptr = reinterpret_cast<uintptr_t>(ptr);
+    }
+    this->len = len;
+}
+
+GermanString::GermanString(const void* str, size_t len) {
+    if (len <= INLINE_MAX_LENGTH) {
+        auto* p = reinterpret_cast<char*>(this);
+        std::fill(p, p + sizeof(GermanString), 0);
+        strings::memcpy_inlined(short_rep.str, str, len);
+    } else {
+        strings::memcpy_inlined(long_rep.prefix, str, PREFIX_LENGTH);
+        long_rep.ptr = reinterpret_cast<uintptr_t>(str);
+    }
+    this->len = len;
+}
+GermanString& GermanString::operator=(const Slice& slice) {
+    *this = GermanString(slice);
+    return *this;
+}
+
+GermanString::GermanString(const GermanString& rhs, void* ptr) {
+    strings::memcpy_inlined(this, &rhs, sizeof(GermanString));
+    if (!rhs.is_inline()) {
+        // NOLINTNEXTLINE(performance-no-int-to-ptr)
+        const auto* rhs_ptr = reinterpret_cast<const char*>(rhs.long_rep.ptr);
+        strings::memcpy_inlined(ptr, rhs_ptr, rhs.len);
+        long_rep.ptr = reinterpret_cast<uintptr_t>(ptr);
+    }
+}
+
+GermanString::operator std::string() const {
+    if (len <= INLINE_MAX_LENGTH) {
+        return std::string(short_rep.str, len);
+    } else {
+        std::string s;
+        raw::make_room(&s, len);
+        char* data = s.data();
+        // NOLINTNEXTLINE(performance-no-int-to-ptr)
+        strings::memcpy_inlined(data, reinterpret_cast<const char*>(long_rep.ptr), len);
+        return s;
+    }
+}
+
+uint32_t GermanString::fnv_hash(uint32_t seed) const {
+    if (is_inline()) {
+        return HashUtil::fnv_hash(short_rep.str, len, seed);
+    } else {
+        // // NOLINTNEXTLINE(performance-no-int-to-ptr)
+        return HashUtil::fnv_hash(reinterpret_cast<const char*>(long_rep.ptr), len, seed);
+    }
+}
+uint32_t GermanString::crc32_hash(uint32_t seed) const {
+    if (is_inline()) {
+        return HashUtil::zlib_crc_hash(short_rep.str, len, seed);
+    } else {
+        // NOLINTNEXTLINE(performance-no-int-to-ptr)
+        return HashUtil::zlib_crc_hash(reinterpret_cast<const char*>(long_rep.ptr), len, seed);
+    }
+}
+
+} // namespace starrocks
--- a/be/src/column/german_string.h
+++ b/be/src/column/german_string.h
@@ -1,0 +1,139 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <glog/logging.h>
+
+#include <cstdint>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "runtime/mem_pool.h"
+#include "runtime/memory/column_allocator.h"
+#include "util/memcmp.h"
+#include "util/raw_container.h"
+#include "util/slice.h"
+
+namespace starrocks {
+class Slice;
+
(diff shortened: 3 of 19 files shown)


be/src/column/german_string.h
```diff
@@ -0,0 +1,246 @@
...
+    virtual void incorporate(std::shared_ptr<GermanStringExternalAllocator>&& other);
+    bool is_from_binary() const { return false; }
+
+private:
+    std::vector<std::vector<char>> medium_string_pages;
+    std::vector<std::string> large_strings;
+};
+
+using Bytes = starrocks::raw::RawVectorPad16<uint8_t, ColumnAllocator<uint8_t>>;
+class GermanStringBinaryColumnExternalAllocator : public GermanStringExternalAllocator {
```
stdpain: seems unused ?

be/src/column/german_string.h
```diff
@@ -0,0 +1,246 @@
...
+    inline bool operator>(const GermanString& rhs) const { return compare(rhs) > 0; }
+    inline bool operator>=(const GermanString& rhs) const { return compare(rhs) >= 0; }
+    inline bool operator<=>(const GermanString& rhs) const { return compare(rhs); }
+
+    friend std::ostream& operator<<(std::ostream& os, const GermanString& gs) {
+        os << gs.to_string();
+        return os;
+    }
+    union {
+        uint32_t len;
```
silverbullet233, line 120: this len is not necessary?

satanson: no overhead

be/src/column/binary_column.h
```diff
@@ -342,7 +358,10 @@ class BinaryColumnBase final : public CowFactory<ColumnFactory<Column, BinaryCol
         _slices_cache = false;
     }
 
-    void invalidate_slice_cache() { _slices_cache = false; }
+    void invalidate_slice_cache() {
+        _slices_cache = false;
+        _german_strings_cache = false;
```
ZiheLiu, line 363: Do the methods like `append` also need call `_german_strings_cache = false`?

be/src/column/german_string.h
```diff
@@ -0,0 +1,217 @@
...
+    explicit GermanString(const Slice& slice) : GermanString(slice.data, slice.size){};
+    GermanString& operator=(const Slice& slice);
+
+    GermanString(const std::string_view& str, void* ptr) : GermanString(str.data(), str.size(), ptr){};
+    GermanString(const std::string& str);
+    GermanString& operator=(const std::string& str);
+
+    explicit operator std::string() const;
+
+    void append(size_t pos, const void* str, size_t len) {
```
silverbullet233: why we need write interface in GermanString？I think it's just a string ref only for read?

be/src/exec/sorting/sort_column.cpp
```diff
@@ -170,16 +171,29 @@ class ColumnSorter final : public ColumnVisitorAdapter<ColumnSorter<R>> {
...
+            auto cmp = [&](const ItemType& lhs, const ItemType& rhs) -> int {
+                return lhs.inline_value.compare(rhs.inline_value);
+            };
 
-        auto inlined = create_inline_permutation<Slice, IS_RANGES>(_permutation, column.get_proxy_data());
-        RETURN_IF_ERROR(sort_and_tie_helper(_cancel, &column, _sort_desc.asc_order(), inlined, _tie, cmp,
-                                            _range_or_ranges, _build_tie));
-        restore_inline_permutation(inlined, _permutation);
+            auto inlined =
+                    create_inline_permutation<GermanString, IS_RANGES>(_permutation, column.get_german_strings());
```
silverbullet233, line 181: How can we know the exact cost of constructing the German string itself? should we add some profile metrics?

be/src/column/german_string.h
```diff
@@ -0,0 +1,217 @@
...
+            return (t2 << 8u) | (t3 >> 8u);
+        };
+
+        auto lhs_prefix = *reinterpret_cast<const uint32_t*>(this->long_rep.prefix);
+        auto rhs_prefix = *reinterpret_cast<const uint32_t*>(rhs.long_rep.prefix);
+        lhs_prefix = byte_swap(lhs_prefix);
+        rhs_prefix = byte_swap(rhs_prefix);
+
+        auto r = lhs_prefix - rhs_prefix;
+        if (r != 0) {
```
silverbullet233: if (lhs_prefix != rhs_prefix) ？it seems r is not necessary here？

be/src/column/german_string.h
```diff
@@ -0,0 +1,217 @@
...
+    std::string to_string() const { return static_cast<std::string>(*this); }
+    inline bool operator==(const GermanString& rhs) const {
+        if (this->len != rhs.len) {
+            return false;
+        }
+        const auto lhs_h0 = *reinterpret_cast<const uint64_t*>(this);
+        const auto rhs_h0 = *reinterpret_cast<const uint64_t*>(&rhs);
+        if (lhs_h0 != rhs_h0) {
+            return false;
+        }
```
silverbullet233: we can directly know whether length and prefix  both are equal by comparing lhs_h0 and rhs_h0？L144 is not necessary?

be/src/column/german_string.cpp
```diff
@@ -0,0 +1,156 @@
...
+    auto* p = reinterpret_cast<char*>(this);
+    std::fill(p, p + sizeof(GermanString), 0);
+    this->len = len;
+    if (len <= INLINE_MAX_LENGTH) {
+        strings::memcpy_inlined(short_rep.str, str, len);
+    } else {
+        strings::memcpy_inlined(long_rep.prefix, str, PREFIX_LENGTH);
+        strings::memcpy_inlined(ptr, str, len);
+        long_rep.ptr = reinterpret_cast<uintptr_t>(ptr);
+    }
```
silverbullet233:
```suggestion
    short_rep.len = len;
    if (len <= INLINE_MAX_LENGTH) {
        memset(short_rep.str, 0, INLINE_MAX_LENGTH);
        strings::memcpy_inlined(short_rep.str, str, len);
    } else {
        strings::memcpy_inlined(long_rep.prefix, str, PREFIX_LENGTH);
        strings::memcpy_inlined(ptr, str, len);
        long_rep.ptr = reinterpret_cast<uintptr_t>(ptr);
    }
```

be/src/column/german_string.h
satanson: deleted

be/src/column/binary_column.h
satanson: yes, constructing slices and german_strings from binary_column is on-command,  when mutating binary_column,  should invalidate cached results.

be/src/column/german_string.h
satanson: it is used by hash-agg and hash-join, so new is unused, so it can be deleted.

be/src/exec/sorting/sort_column.cpp
satanson: From benchmarking german string vs slice,  the conversion cost difference is very small can be ignored.

be/src/column/german_string.h
satanson: yes

be/src/exec/sorting/sort_column.cpp
satanson: From benchmarking german string vs slice,  the conversion cost difference is very small can be ignored.

be/src/column/german_string.h
satanson: yes, fixed

be/src/column/german_string.h
```diff
@@ -0,0 +1,156 @@
...
+        if (lhs_prefix != rhs_prefix) {
+            // if prefix not equal, we can determine the result by prefix
+            lhs_prefix = byte_swap(lhs_prefix);
+            rhs_prefix = byte_swap(rhs_prefix);
+            return (lhs_prefix > rhs_prefix) - (lhs_prefix < rhs_prefix);
+        }
+
+        auto min_len = std::min<uint32_t>(this->len, rhs.len);
+        const auto* lhs_str =
+                this->is_inline() ? this->short_rep.str : reinterpret_cast<const char*>(this->long_rep.ptr);
```
silverbullet233: maybe we can add a method `get_data()`? Then all interfaces that access data do not need to check is_Inlined, only `get_data()` and `get_size()` are needed.

be/src/column/german_string.cpp
satanson: fixed

be/src/column/binary_column.h
ZiheLiu: @satanson  So we should modify these methods to call `_german_strings_cache = false`? It seems that they do not call it for now, like
https://github.com/StarRocks/starrocks/pull/62929/files#diff-b1b0122ec118c02609a167a3a6ae6b6d806926fc24211126d7240f74b08797a4R406

be/src/column/german_string.h
```diff
@@ -0,0 +1,156 @@
...
+        } short_rep;
+        struct {
+            uint32_t len;
+            char prefix[PREFIX_LENGTH];
+            uintptr_t ptr;
+        } long_rep;
+    };
+};
+
+class GermanStringExternalAllocator {
```
stdpain: seems only used in UTs. How about port to be/test dir?

be/src/column/german_string.h
```diff
@@ -0,0 +1,156 @@
...
+    union {
+        uint32_t len;
+        struct {
+            uint32_t len;
+            char str[INLINE_MAX_LENGTH];
+        } short_rep;
+        struct {
+            uint32_t len;
+            char prefix[PREFIX_LENGTH];
+            uintptr_t ptr;
```
stdpain, line 128: The address size on x64 is 48 bits. If pointers only retain 48 bits, we can inline more characters.

be/src/column/binary_column.h
```diff
@@ -366,12 +385,16 @@ class BinaryColumnBase final : public CowFactory<ColumnFactory<Column, BinaryCol
...
 private:
     void _build_slices() const;
+    void _build_german_strings() const;
 
     Bytes _bytes;
     Offsets _offsets;
 
     mutable Container _slices;
     mutable bool _slices_cache = false;
+    mutable GermanStringContainer _german_strings;
```
stdpain, line 395: Observe whether the German string container leads to increased memory usage.
in #17362, we find that container data may double the memory usage of string columns.

be/src/column/german_string.h
```diff
@@ -0,0 +1,156 @@
...
+    inline bool operator<=(const GermanString& rhs) const { return compare(rhs) <= 0; }
+    inline bool operator>(const GermanString& rhs) const { return compare(rhs) > 0; }
+    inline bool operator>=(const GermanString& rhs) const { return compare(rhs) >= 0; }
+    inline bool operator<=>(const GermanString& rhs) const { return compare(rhs); }
+
+    friend std::ostream& operator<<(std::ostream& os, const GermanString& gs) {
+        os << gs.to_string();
+        return os;
+    }
+    union {
```
kangkaisen: there is a Memory alignment issue. should update to

```
class GermanString {
    union {
        struct {
            uint32_t len;
            char data[12];  
        } raw;
        struct {
            uint32_t len;
            char str[12];
        } short_rep;
        struct {
            uint32_t len;
            char prefix[4];
            uintptr_t ptr;
        } long_rep;
    };
   

};
```

or 

```
    alignas(8) union { ... };

```

be/src/runtime/runtime_state.h
```diff
@@ -362,6 +362,10 @@ class RuntimeState {
         return _spill_options->spill_partitionwise_agg_skew_elimination;
     }
 
+    bool enable_full_sort_use_german_string() const {
+        return _query_options.__isset.enable_full_sort_use_german_string;
```
kangkaisen: bool enable_full_sort_use_german_string() const {
      return _query_options.__isset.enable_full_sort_use_german_string &&
             _query_options.enable_full_sort_use_german_string;
  }

be/src/column/binary_column.h
satanson: fixed

be/src/column/german_string.h
satanson: fixed

be/src/column/german_string.h
satanson: fixed

be/src/column/german_string.h
satanson: fixed

be/src/runtime/runtime_state.h
satanson: fixed

be/src/column/binary_column.h
satanson: Constructing GermanString from BinaryColumn costs more then constructing Slice, in partial_sort and merge_sort, an element of BinaryColumn take participant in comparison multiple times, so GermanString should construct in-advance. if there are a memory OOM, we call turn off german string in sv.

be/src/column/german_string.h
satanson: Accoding to GermanString definition, for long string, only save prefix 4 bytes,   I guess the motivation is that both compare and  equal method can be optimized based on this. for compare method, first comparing 4 bytes, we can turn these 4 bytes into uint32 to compare them.

ZiheLiu (approved):

silverbullet233 (approved):

kangkaisen (approved):

be/src/util/misc.h
```diff
@@ -22,4 +24,21 @@ namespace starrocks {
 // take a sleep with small intervals until time out by `sleep_secs` or the `stop_condition()` is true
 void nap_sleep(int32_t sleep_secs, const std::function<bool()>& stop_condition);
 
+#if defined(__GNUC__) || defined(__clang__)
+#define NOT_SUPPORT()                                                                        \
+    do {                                                                                     \
+        throw std::runtime_error(std::string("Not support method '") + __PRETTY_FUNCTION__ + \
```
kevincai, line 30: `runtime_error` will be printed by the stack_util as always, is it expected. `RuntimeException` is introduced in https://github.com/starRocks/StarRocks/pull/62098 , worth a double check.

kevincai (approved):

satanson:
ignore backport check: 4.0.1

PR #63042, opened 2025-09-12
[BugFix] Fix null exception during remove expired load job

sevev:
## Why I'm doing:
The `unprotectedRemoveJobReleatedMeta` method in `LoadMgr `was throwing NullPointerException when trying to remove expired load jobs. The issue occurred when sameLabelJobs was null, expired LoadJobs fail to be cleaned up, which in turn leads to high memory usage on the FE.
### Root Cause Analysis
The null pointer exception was caused by a race condition between two concurrent operations:
1. Periodic cleanup thread: LoadLabelCleaner daemon that runs every label_clean_interval_second seconds
2. Journal replay thread: replayer daemon that processes journal logs without proper synchronization
Specific scenarios that triggered the issue:
+ Duplicate cleanup: The same job could be cleaned up by both the periodic cleaner and journal replay simultaneously
+ Data inconsistency: idToLoadJob and dbIdToLabelToLoadJobs could become inconsistent during concurrent modifications
+ Missing synchronization: Journal replay operations (replayEndLoadJob, replayUpdateLoadJobStateInfo) were not protected by locks, while periodic cleanup used write locks

Timeline of the race condition:
```
T1: Cleanup thread acquires write lock, removes job from idToLoadJob
T2: Cleanup thread removes job from dbIdToLabelToLoadJobs  
T3: Journal replay thread processes END_LOAD_JOB log for the same job
T4: Journal replay calls unprotectedRemoveJobReleatedMeta(job)
T5: sameLabelJobs = labelToLoadJobs.get(label) returns null ❌
T6: NullPointerException when calling sameLabelJobs.remove(job)
```

## What I'm doing:
Added null safety checks in unprotectedRemoveJobReleatedMeta method to prevent NullPointerException:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/loadv2/LoadMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/loadv2/LoadMgr.java
@@ -391,14 +391,15 @@ public class LoadMgr implements MemoryTrackable {
         // 2. remove from dbIdToLabelToLoadJobs
         Map<String, List<LoadJob>> labelToLoadJobs = dbIdToLabelToLoadJobs.get(dbId);
         List<LoadJob> sameLabelJobs = labelToLoadJobs.get(label);
-        sameLabelJobs.remove(job);
-        if (sameLabelJobs.isEmpty()) {
-            labelToLoadJobs.remove(label);
+        if (sameLabelJobs != null) {
+            sameLabelJobs.remove(job);
+            if (sameLabelJobs.isEmpty()) {
+                labelToLoadJobs.remove(label);
+            }
+            if (labelToLoadJobs.isEmpty()) {
+                dbIdToLabelToLoadJobs.remove(dbId);
+            }
         }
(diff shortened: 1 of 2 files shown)


wyb:
does this issue happen in follower FE?
how about adding lock in replay？

sevev:
> does this issue happen in follower FE? how about adding lock in replay？

Follower FE could also occur in theoretically. Already add lock in replay.

wyb (approved):

gengjun-git (approved):

sevev:
@mergify backport branch-3.3

PR #63085, opened 2025-09-12
[BugFix] Remove the deregister logic from container

### LINKED ISSUE #62915: Compute Node does not gracefully exit and deregister from FE Node on shutdown

yandongxiao:
## Why I'm doing:

There are two different scenarios of CN pod exit.

1. POD restart (due to config update or version upgrade) or crash
2. POD scale in

It is not necessary to deregister the CN POD in the first scenario while it is needed in the second scenario.

The operator will know the scale in op and do the deregistration, user should use the >= v1.11.1 operator version.

## What I'm doing:

Fixes #62915

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/docker/dockerfiles/be/cn_entrypoint.sh
+++ b/docker/dockerfiles/be/cn_entrypoint.sh
@@ -163,9 +163,13 @@ if [[ "x$LOG_CONSOLE" == "x1" ]] ; then
     addition_args="--logconsole"
 fi
 $STARROCKS_HOME/bin/start_cn.sh $addition_args
+ret=$?
 
 # The reason why we need to sleep here is to avoid the pod being killed by k8s
 # before the preStop hook is exited.
 # If the CN subprocess fails to start, we also want the entrypoint script to exit as soon as possible.
 # Taking all the above into consideration, we set it to five seconds.
 sleep 5
+
+# keep the same return code from start_cn.sh
+exit $ret
--- a/docker/dockerfiles/be/cn_entrypoint.sh
+++ b/docker/dockerfiles/be/cn_entrypoint.sh
@@ -164,7 +164,8 @@ if [[ "x$LOG_CONSOLE" == "x1" ]] ; then
 fi
 $STARROCKS_HOME/bin/start_cn.sh $addition_args
 
-# the reason why we need to sleep here is to avoid the pod being killed by k8s
+# The reason why we need to sleep here is to avoid the pod being killed by k8s
 # before the preStop hook is exited.
-# 120 is the default value of terminationGracePeriodSeconds
-sleep 120
+# If the CN subprocess fails to start, we also want the entrypoint script to exit as soon as possible.
+# Taking all the above into consideration, we set it to five seconds.
+sleep 5
--- a/docker/dockerfiles/be/cn_entrypoint.sh
+++ b/docker/dockerfiles/be/cn_entrypoint.sh
@@ -165,11 +165,11 @@ fi
 $STARROCKS_HOME/bin/start_cn.sh $addition_args
 ret=$?
 
-# The reason why we need to sleep here is to avoid the pod being killed by k8s
-# before the preStop hook is exited.
-# If the CN subprocess fails to start, we also want the entrypoint script to exit as soon as possible.
-# Taking all the above into consideration, we set it to five seconds.
-sleep 5
+if [[ $ret -eq 0 || $ret -eq 137 ]] ; then
+    # The reason why we need to sleep here is to avoid the pod being killed by k8s before the preStop hook is exited.
+    # If the CN subprocess fails to start, we also want the entrypoint script to exit as soon as possible.
+    sleep 5
+fi
 
 # keep the same return code from start_cn.sh
 exit $ret
--- a/docker/dockerfiles/be/cn_entrypoint.sh
+++ b/docker/dockerfiles/be/cn_entrypoint.sh
@@ -139,30 +139,8 @@ drop_my_self()
 
 exit_clean()
 {
-    log_stderr "Receives signal to exit ..."
-    pidfile=$STARROCKS_HOME/bin/cn.pid
-    if ! test -f $pidfile ; then
-        log_stderr "Can't find $pidfile!"
-    else
-        pid=`cat $pidfile`
-        if [[ "x$pid" == "x" ]] ; then
-            log_stderr "Empty pid file!"
-        else
-            log_stderr "detect cn pid $pid exists ..."
-            while true
-            do
-                if ps -p $pid &>/dev/null ; then
-                    log_stderr "cn process $pid is still alive ..."
-                    sleep $PROBE_INTERVAL
-                else
-                    log_stderr "cn process $pid dead."
-                    break;
-                fi
-            done
-        fi
-    fi
-    # remove myself from FE
-    drop_my_self $svc_name
+    log_stderr "Got SIGTERM, exit ..."
+    exit 143
 }
 
 svc_name=$1
@@ -185,12 +163,8 @@ if [[ "x$LOG_CONSOLE" == "x1" ]] ; then
     addition_args="--logconsole"
 fi
 $STARROCKS_HOME/bin/start_cn.sh $addition_args
-ret=$?
-if [[ $ret -ne 0 && "x$LOG_CONSOLE" != "x1" ]] ; then
-    nol=50
-    log_stderr "Last $nol lines of cn.INFO ..."
-    tail -n $nol $STARROCKS_HOME/log/cn.INFO
-    log_stderr "Last $nol lines of cn.out ..."
-    tail -n $nol $STARROCKS_HOME/log/cn.out
-fi
-exit $ret
+
+# the reason why we need to sleep here is to avoid the pod being killed by k8s
+# before the preStop hook is exited.
+# 120 is the default value of terminationGracePeriodSeconds
+sleep 120


alvin-phoenix-ai:
@cursor review

yandongxiao:
> @cursor review

I need to check how SIGKILL is handled

yandongxiao:
> > @cursor review
> 
> I need to check how SIGKILL is handled

Regardless of whether the `preStop` configuration is `/opt/starrocks/cn/bin/stop_cn.sh` or `/opt/starrocks/cn/bin/stop_cn.sh -g`, after k8s uses SIGKILL or SIGTERM to kill the BE process, k8s will immediately send the SIGTERM signal to the entrypoint script, so it can always exit in a timely manner.

docker/dockerfiles/be/cn_entrypoint.sh
```diff
@@ -139,30 +139,8 @@ drop_my_self()
...
-                else
-                    log_stderr "cn process $pid dead."
-                    break;
-                fi
-            done
-        fi
-    fi
-    # remove myself from FE
-    drop_my_self $svc_name
+    log_stderr "Got SIGTERM, exit ..."
```
kevincai, line 142: this makes `exit_clean` trap useless, so a cleaner way is remove exit_clean() and the corresponding signal trap.

docker/dockerfiles/be/cn_entrypoint.sh
```diff
@@ -185,12 +163,8 @@ if [[ "x$LOG_CONSOLE" == "x1" ]] ; then
...
-    tail -n $nol $STARROCKS_HOME/log/cn.INFO
-    log_stderr "Last $nol lines of cn.out ..."
-    tail -n $nol $STARROCKS_HOME/log/cn.out
-fi
-exit $ret
+
+# the reason why we need to sleep here is to avoid the pod being killed by k8s
+# before the preStop hook is exited.
+# 120 is the default value of terminationGracePeriodSeconds
+sleep 120
```
kevincai: this makes the pod termination wait for 120seconds under no condition, brutal in prod env to hold a pod exit for additional 120 seconds.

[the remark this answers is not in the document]
yandongxiao: If the CN subprocess fails to start, there will indeed be an issue where the container cannot exit immediately, which is not a good practice. Our original intention was that the `entrypoint` script should only exit after the `preStop` script (whose main purpose is to kill the CN subprocess) has been successfully executed; otherwise, there will be an error indicating that the preStop script execution has failed. Determining whether the process was actively killed or encountered a startup exception based on the return code of the CN subprocess also does not seem to be a good option. In summary, I think we should adjust the sleep time to 5 seconds.

kevincai (commented):
The key thing is, how to take the compatibility into consideration. The deregistration of CN node is only available since v1.11+, so there are two cases:
1. operator v1.11+, starrocks core removed the deregister logic completely, or
2. operator below v1.111, starrocks core still do the best to deregister the node.

For a long term view, option 1 is preferred. so the cn deregister logic shall be completely removed from starrocks core. However, if the user doesn't upgrade the operator to v1.11, it means the deregistration will be completely lost with only the entrypoint change.

Also consider remove the preStop hook for the pod, if the SIGTERM can be properly propagated to the CN/BE process, then exit. So the preStop hook can be returned to the user to do whatever they need to do before stop

docker/dockerfiles/be/cn_entrypoint.sh
yandongxiao: Without a trap, script would have to execute the sleep operation, causing the container's exit time to increase.

docker/dockerfiles/be/cn_entrypoint.sh
yandongxiao: Good catch. I have changed it to 5s.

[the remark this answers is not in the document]
yandongxiao: @kevincai here is my think.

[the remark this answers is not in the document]
kevincai: under what condition, the cn subprocess failed to start, but the entrypoint doesn't exit?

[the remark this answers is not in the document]
kevincai: the original preStop script is introduced because CN/BE had a bad exit behavior since v2.4/v2.5, a SIGTERM can rarely kill the be/cn process to exit quickly, that's why a prestop is introduced to stop with -g or stop directly without -g (kill -9) to force stop the be/cn process.

Since v3.2, the situation changes a lot, the exit behavior is properly maintained for most of the time, so we shall turn to the regular way to think about the exit procedure in k8s env.

docker/dockerfiles/be/cn_entrypoint.sh
```diff
@@ -185,12 +163,9 @@ if [[ "x$LOG_CONSOLE" == "x1" ]] ; then
...
-    log_stderr "Last $nol lines of cn.out ..."
-    tail -n $nol $STARROCKS_HOME/log/cn.out
-fi
-exit $ret
+
+# The reason why we need to sleep here is to avoid the pod being killed by k8s
+# before the preStop hook is exited.
+# If the CN subprocess fails to start, we also want the entrypoint script to exit as soon as possible.
+# Taking all the above into consideration, we set it to five seconds.
+sleep 5
```
kevincai: this will override the exit code, no matter it is a normal exit or a crash or whatever, better keep the sam return code from start_cn.sh

kevincai: maybe we can do this:

```
if [[ $ret -eq 0 ]] ; then
    sleep 5
else
    exit $ret
fi
```

exit immediately if exit code is non-zero, otherwise sleep for a while to be favor of the preStop hook.

yandongxiao: I think this suggestion is fine. Previously, I always executed sleep 5 because I was unsure whether the normal exit logic of BE would return 0.

alvin-phoenix-ai:
@cursor review

kevincai (approved):

wangruin (approved):

PR #63143, opened 2025-09-15
[BugFix] Fix incompatible zonemap reuse for fast schema evolution in shared-data

banmoy:
## Why I'm doing:
Fast schema evolution in shared-data modifies a column's type without rewriting data files, and the existing zonemap index is reused for the new type. However, this reuse is not always safe. For certain type conversions, the min/max values stored in the zonemap are no longer valid under the new type's comparison rules, leading to incorrect query results.

For example, when changing a column from `INT` to `STRING`, the numeric ordering (`9 < 123`) differs from the lexicographical ordering (`"123" < "9"`). Reusing the integer-based zonemap for string-based comparisons will cause the query engine to prune data incorrectly. This issue can be reproduced with the following steps:
```
create table t (c0 int, c1 int) duplicate key (c0);

insert into t values (1, 123), (2, 9);

select * from t where c1 = 123;             -- expect one row. correct
+------+------+
| c0   | c1   |
+------+------+
|    1 |  123 |
+------+------+
1 row in set (0.02 sec)

alter table t modify column c1 string;

select * from t where c1 = "123";           -- expect one row, but is empty. wrong
Empty set (0.01 sec)
```

From the profile, we can observe that the zonemap filter unexpectedly pruned all rows.

image

This PR wants to fix the issue.

### Conditions for Zonemap Index Reuse
The zonemap index mechanism is briefly illustrated below. The type conversion that wants to safely reuse the existing zonemap index must meet two conditions simultaneously:
* The type conversion must be monotonically non-decreasing. This ensures that the original min/max values remain valid boundaries after the conversion.
* The type conversion must not change non-null values to null. This is required to keep the has_null and has_not_null metadata accurate.

image

Based on these conditions, the following is the compatibility matrix defining which type conversions can safely reuse the zonemap index.
- Schema Change Conversion Prohibited: ⚪
- Zonmap Index can be reused:  ✅
- Zonmap Index can not be reused:  ❌
image

### Solution

To address this issue, two potential solutions were considered:

**Option 1: Prohibit Incompatible Fast Schema Evolution (FE-side Check)**

This approach involves enhancing the validation logic in the Frontend (FE). Before a fast schema evolution is executed, the system checks if the proposed type change is compatible with zonemap index reuse. If the conversion is not on the approved list, fast schema evolution is disallowed, and the operation falls back to the traditional method of rewriting data and rebuilding the index.

   *   **Pros:** Guarantees that zonemap indexes are always valid and can be used by the query optimizer, ensuring consistent and predictable performance.
   *   **Cons:** A subset of type changes (primarily conversions to/from `VARCHAR`) will not be able to use the fast schema evolution feature. Based on experience, these types of conversions are infrequent.

**Option 2: Conditionally Disable Zonemap Index at Query Time (BE-side Logic)**

This alternative would allow all type conversions to proceed with fast schema evolution. The Backend (BE) would then be responsible for detecting incompatible conversions at query time and disabling the use of the zonemap index for those specific columns.

   *   **Pros:** Allows all type modifications to benefit from fast schema evolution, avoiding data rewrites.
   *   **Cons:** May lead to significant and unpredictable performance degradation for queries, as the optimizer would be unable to use the zonemap for pruning.

**Decision**

Query performance is a critical priority. The zonemap index is a widely used and highly effective optimization, and disabling it could have a severe, hard-to-predict impact. Therefore, **choose Option 1** to guarantee index correctness and consistent query performance. The trade-off of disallowing fast schema evolution for a small number of less common type conversions is acceptable to avoid this risk.

### Code Changes
- Add `ZONEMAP_SCHEMA_CHANGE_MATRIX` and `canReuseZonemapIndex(fromType, toType)` to `SchemaChangeTypeCompatibility`, with clear docs on reuse principles.
- Integrate reuse check into FE fast schema evolution path:
  - In schema change planning, for shared-data fast path, validate each altered column via `canReuseZonemapIndex`.
  - If any column is incompatible, disable fast schema evolution and fall back to data rewrite (rebuild zonemap index).
- Tests:
  - Add/extend unit tests for the compatibility matrix and edge cases.
  - Add/adjust schema change tests to verify fallback behavior under incompatible conversions (e.g., VARCHAR↔INT).

Fixes #issue

## What type of PR is this:

- [X] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [X] 4.0
  - [X] 3.5
  - [X] 3.4
  - [X] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/Type.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/Type.java
@@ -903,6 +903,10 @@ public abstract class Type implements Cloneable {
                 !isJsonType() && !isOnlyMetricType() && !isFunctionType() && !isBinaryType();
     }
 
+    public boolean supportZonemap() {
+        return isScalarType() && (isNumericType() || isDateType() || isStringType());
+    }
+
     public static final String NOT_SUPPORT_JOIN_ERROR_MSG =
             "Type (nested) percentile/hll/bitmap/json not support join";
 
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/Type.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/Type.java
@@ -903,7 +903,7 @@ public abstract class Type implements Cloneable {
                 !isJsonType() && !isOnlyMetricType() && !isFunctionType() && !isBinaryType();
     }
 
-    public boolean supportZonemap() {
+    public boolean supportZoneMap() {
         return isScalarType() && (isNumericType() || isDateType() || isStringType());
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -61,6 +61,7 @@ import com.starrocks.catalog.OlapTable.OlapTableState;
 import com.starrocks.catalog.Partition;
 import com.starrocks.catalog.PhysicalPartition;
 import com.starrocks.catalog.Replica;
+import com.starrocks.catalog.SchemaChangeTypeCompatibility;
 import com.starrocks.catalog.SchemaInfo;
 import com.starrocks.catalog.StructField;
 import com.starrocks.catalog.StructType;
@@ -910,6 +911,10 @@ public class SchemaChangeHandler extends AlterHandler {
             if (column.isKey()) {
                 return false;
             }
+        }
+
+        if (!SchemaChangeTypeCompatibility.canReuseZonemapIndex(oriColumn.getType(), modColumn.getType())) {
+            fastSchemaEvolution = false;
         }
 
         return fastSchemaEvolution;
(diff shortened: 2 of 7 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/Type.java
```diff
@@ -903,6 +903,10 @@ public boolean supportBloomFilter() {
                 !isJsonType() && !isOnlyMetricType() && !isFunctionType() && !isBinaryType();
     }
 
+    public boolean supportZonemap() {
```
xiangguangyxg:
```suggestion
    public boolean supportZoneMap() {
```

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/catalog/Type.java
banmoy: fixed

xiangguangyxg (approved):

HangyuanLiu (approved):

alvin-phoenix-ai:
@cursor review

banmoy:
ignore merge 3.3.19

PR #63182, opened 2025-09-16
[BugFix] support clear corrupted lake data cache

starrocks-xupeng:
## What I'm doing:

add another config to control the behavior when reading corrupted cache from star cache for lake table

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1187,7 +1187,10 @@ CONF_mInt32(lake_pk_index_sst_max_compaction_versions, "100");
 // When the ratio of cumulative level to base level is greater than this config, use base merge.
 CONF_mDouble(lake_pk_index_cumulative_base_compaction_ratio, "0.1");
 CONF_Int32(lake_pk_index_block_cache_limit_percent, "10");
-CONF_mBool(lake_clear_corrupted_cache, "true");
+// clear *.meta cache for lake table
+CONF_mBool(lake_clear_corrupted_cache_meta, "true");
+// clear *.data cache for lake table
+CONF_mBool(lake_clear_corrupted_cache_data, "false");
 // The maximum number of files which need to rebuilt in cloud native pk index.
 // If files which need to rebuilt larger than this, we will flush memtable immediately.
 CONF_mInt32(cloud_native_pk_index_rebuild_files_threshold, "50");
--- a/be/src/storage/rowset/page_io.cpp
+++ b/be/src/storage/rowset/page_io.cpp
@@ -42,6 +42,7 @@
 #include "column/column.h"
 #include "common/logging.h"
 #include "fs/fs.h"
+#include "fs/fs_starlet.h"
 #include "gutil/strings/substitute.h"
 #include "runtime/current_thread.h"
 #include "storage/rowset/storage_page_decoder.h"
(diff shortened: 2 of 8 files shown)


be/src/storage/rowset/page_io.cpp
```diff
@@ -144,8 +145,27 @@ std::string encode_cache_key(const std::string& fname, int64_t offset) {
...
 }
 
-Status PageIO::read_and_decompress_page(const PageReadOptions& opts, PageHandle* handle, Slice* body,
-                                        PageFooterPB* footer) {
+#if defined(USE_STAROS) && !defined(BUILD_FORMAT_LIB)
+Status drop_local_cache_data(const std::string& fname) {
+    if (!is_starlet_uri(fname)) {
+        return Status::NotSupported("only support starlet file");
+    }
+    if (!config::lake_clear_corrupted_cache_data) {
```
kevincai, line 150: do this check first, it's cheaper than the uri check.

be/test/fs/fs_starlet_test.cpp
```diff
@@ -430,6 +431,16 @@ TEST_P(StarletFileSystemTest, test_tag) {
     config::starlet_write_file_with_tag = old;
 }
 
+TEST_P(StarletFileSystemTest, test_drop_cache) {
+    bool old = config::lake_clear_corrupted_cache_data;
```
kevincai, line 439:
```suggestion
    SCOPED_UPDATE(bool, config::lake_clear_corrupted_cache_data, false);
```

be/src/common/config.h
```diff
@@ -1187,7 +1187,10 @@ CONF_mInt32(lake_pk_index_sst_max_compaction_versions, "100");
 // When the ratio of cumulative level to base level is greater than this config, use base merge.
 CONF_mDouble(lake_pk_index_cumulative_base_compaction_ratio, "0.1");
 CONF_Int32(lake_pk_index_block_cache_limit_percent, "10");
-CONF_mBool(lake_clear_corrupted_cache, "true");
+// clear *.meta cache for lake table
+CONF_mBool(lake_clear_corrupted_cache_meta, "true");
+// clear *.data cache for lake table
+CONF_mBool(lake_clear_corrupted_cache_data, "false");
```
kevincai, line 1193: add doc

starrocks-xupeng: done

kevincai (dismissed):

be/src/common/config.h
wyb, line 1193: why add a new config?
`lake_clear_corrupted_cache` can be used for all types, like meta, segment, delete vector.

starrocks-xupeng: meta is known issue, they will be corrupted, but data is unknown issue, keep a seperate config might help us find out why data is corrupted

kevincai (approved):

wyb (approved):

PR #63202, opened 2025-09-16
[Enhancement]Optimize performance for the alter publish version in the lake table fast schema change.

zhangheihei:
## Why I'm doing:
Currently, lakePublishVersion executes the publishVersion serially by partition, which can be extremely time-consuming if there are many partitions.

## What I'm doing:
Serial actions are optimized for parallel execution
## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobV2.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobV2.java
@@ -45,7 +45,6 @@ import com.starrocks.catalog.OlapTable.OlapTableState;
 import com.starrocks.catalog.TabletInvertedIndex;
 import com.starrocks.catalog.UserIdentity;
 import com.starrocks.common.Config;
-import com.starrocks.common.ThreadPoolManager;
 import com.starrocks.common.TraceManager;
 import com.starrocks.common.io.Writable;
 import com.starrocks.common.util.concurrent.lock.LockType;
@@ -69,8 +68,6 @@ import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
-import java.util.concurrent.ThreadPoolExecutor;
-import javax.validation.constraints.NotNull;
 
 /*
  * Version 2 of AlterJob, for replacing the old version of AlterJob.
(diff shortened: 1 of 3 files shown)


eshishki:
@zhangheihei the direction is still valid — the per-partition loop in `lakePublishVersion()` is indeed serial today, and `file_bundling`'s aggregate path doesn't fix that: `aggregatePublishVersion` only collapses the *indexes within one partition* into a single batched call; the outer `for (physicalPartitionId ...)` loop is still sequential. So N partitions = N sequential round-trips regardless.

A few things to reconsider before this can move forward, though:

1. **Heads-up on a structural change since you opened this.** The aggregate-publish path (`useAggregatePublish`, #59102) landed after this PR, and your change only parallelizes the non-bundling branch — the bundling branch is left serial. A current fix would need to cover both.

2. **Double publish in the diff.** In the `!useAggregatePublish` branch the original synchronous `Utils.publishVersion(...)` call is still there, and you also submit the same call inside the `CompletableFuture`. As written each index gets published twice.

3. **The win is a trade-off, not unconditional — and it can be made adaptive.** `Utils.publishVersion`/`aggregatePublishVersion` already fan out in parallel *across compute nodes* (they group tablets by node, send the RPCs concurrently, then join). So the only thing the outer parallelism buys you is overlapping the per-partition round-trips. That helps when nodes are under-utilized by a single partition, and does almost nothing when a single partition already saturates the cluster. The clean signal is `tablets_per_partition` (≈ `bucket_num × dirty_index_count`) vs `numComputeNodes`: parallelize the outer loop only when `numPartitions > 1 && tablets_per_partition < numComputeNodes`, otherwise keep it serial and avoid the future/executor overhead. Both numbers are already available right there (`physicalPartitionIndexMap.rowKeySet()`, `index.getTablets()`, system info).

Also needs a rebase — it's currently conflicting with `main`. If you'd like to pick this back up, happy to take another look.

PR #63203, opened 2025-09-16
[Doc] Add tip about TPC-DS benchmark query example

DanRoscigno:
Added a tip regarding the source of the example query.

@murphyatwork , a reader left a comment saying:

> The example query plan and the corresponding explanation of the plan are not in sync, so it's difficult to understand

I am guessing that this is because the plan is a simplified version, but I do not know. Can you have a look please?

While I was reading through the page I noticed that you mentioned "query 96," so I added a tip that the query is from the TPC-DS benchmark.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/docs/en/best_practices/query_tuning/query_planning.md
+++ b/docs/en/best_practices/query_tuning/query_planning.md
@@ -103,6 +103,10 @@ You can control this merging behavior through the session variable `pipeline_pro
 
 ### Example Query Plan
 
+:::tip
+This is query 96 from the TPC-DS benchmark.
+:::
+
 ```sql
 EXPLAIN select  count(*)
 from store_sales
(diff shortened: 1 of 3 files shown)


murphyatwork:
> I am guessing that this is because the plan is a simplified version, but I do not know. 

yeah, the actual query plan is too long to put here, i should use a simple query with a simple plan instead.

murphyatwork (approved):

EsoragotoSpirit (approved):

PR #63218, opened 2025-09-17
[UT] Fix memory leak in test LocalTabletsChannelTest::test_add_chunk_not_exist_tablet

banmoy:
## Why I'm doing
- ASAN reported a memory leak in BE unit test `LocalTabletsChannelTest::test_add_chunk_not_exist_tablet`
  - Direct leak from `SecondaryReplicasWaiter::_send_replica_status_request()` allocating `ReusableClosure<PLoadReplicaStatusResult>`
```
=================================================================
==752488==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 792 byte(s) in 1 object(s) allocated from:
    #0 0x1a830f48 in operator new(unsigned long) (/root/celerdata/be/ut_build_ASAN/test/starrocks_test+0x1a830f48)
    #1 0x30153219 in starrocks::SecondaryReplicasWaiter::_send_replica_status_request(int) /root/celerdata/be/src/runtime/local_tablets_channel.cpp:1342
    #2 0x30152d52 in starrocks::SecondaryReplicasWaiter::_try_check_replica_status_on_primary(int) /root/celerdata/be/src/runtime/local_tablets_channel.cpp:1327
    #3 0x301520e3 in starrocks::SecondaryReplicasWaiter::wait() /root/celerdata/be/src/runtime/local_tablets_channel.cpp:1285
    #4 0x3013603a in starrocks::LocalTabletsChannel::add_chunk(starrocks::Chunk*, starrocks::PTabletWriterAddChunkRequest const&, starrocks::PTabletWriterAddBatchResult*, bool*) /root/celerdata/be/src/runtime/local_tablets_channel.cpp:380
    #5 0x1eefff41 in starrocks::LocalTabletsChannelTest_test_add_chunk_not_exist_tablet_Test::TestBody() /root/celerdata/be/test/runtime/local_tablets_channel_test.cpp:255
    #6 0x37d95703 in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) googletest-release-1.10.0/googletest/src/gtest.cc:2433
    #7 0x37d95703 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) googletest-release-1.10.0/googletest/src/gtest.cc:2469
    #8 0x37d8856d in testing::Test::Run() googletest-release-1.10.0/googletest/src/gtest.cc:2508
    #9 0x37d8856d in testing::Test::Run() googletest-release-1.10.0/googletest/src/gtest.cc:2498
    #10 0x37d886cd in testing::TestInfo::Run() googletest-release-1.10.0/googletest/src/gtest.cc:2684
    #11 0x37d886cd in testing::TestInfo::Run() googletest-release-1.10.0/googletest/src/gtest.cc:2657
    #12 0x37d887b6 in testing::TestSuite::Run() googletest-release-1.10.0/googletest/src/gtest.cc:2816
    #13 0x37d887b6 in testing::TestSuite::Run() googletest-release-1.10.0/googletest/src/gtest.cc:2795
    #14 0x37d88d13 in testing::internal::UnitTestImpl::RunAllTests() googletest-release-1.10.0/googletest/src/gtest.cc:5338
    #15 0x37d88f23 in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) googletest-release-1.10.0/googletest/src/gtest.cc:2433
    #16 0x37d88f23 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) googletest-release-1.10.0/googletest/src/gtest.cc:2469
    #17 0x37d88f23 in testing::UnitTest::Run() googletest-release-1.10.0/googletest/src/gtest.cc:4925
    #18 0x1a893ea2 in RUN_ALL_TESTS() /var/local/thirdparty/installed/include/gtest/gtest.h:2473
    #19 0x1a87d8a4 in starrocks::init_test_env(int, char**) /root/celerdata/be/src/testutil/init_test_env.h:115
    #20 0x1a87e332 in main /root/celerdata/be/test/test_main.cpp:18
    #21 0x7f5813e29d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f)

```
  
- Root cause:
  - The send path does `closure->ref()` twice (one for “self-hold”, one for “RPC-hold”).
  - The test does not register a sync-point callback for `LocalTabletsChannel::rpc::get_load_replica_status` to simulate the RPC, so no one calls `closure->Run()`, and the “RPC-hold” ref is never released.
  - The waiter’s release path only dropped one ref, leaving one outstanding reference and causing the leak.

## What I'm doing
- Introduce and use a sync-point primitive `TEST_SYNC_POINT_CALLBACK_OR_DEFAULT` with a default action, so tests without a registered callback still release the “RPC-hold” reference.
- At `LocalTabletsChannel::rpc::get_load_replica_status`, switch to a callback-with-default that calls `closure->Run()` when no test callback is installed. This balances the double `ref()` even in no-callback tests, while keeping production behavior unchanged.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [X] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [X] 4.0
  - [X] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/test/runtime/local_tablets_channel_test.cpp
+++ b/be/test/runtime/local_tablets_channel_test.cpp
@@ -234,9 +234,17 @@ using RpcLoadDisagnosePair = std::pair<PLoadDiagnoseRequest*, ReusableClosure<PL
 
 TEST_F(LocalTabletsChannelTest, test_add_chunk_not_exist_tablet) {
     _create_tablets(1);
-    // open as a secondary replica of 3 replicas
+
     ReplicaInfo replica_info{_tablets[0]->tablet_id(), _nodes};
-    _open_channel(_nodes[1].node_id(), {replica_info});
+    PTabletWriterOpenRequest request;
+    _create_open_request(_nodes[1].node_id(), {replica_info}, &request);
+    // turn off _is_replicated_storage to avoid launching secondary waiter
+    request.set_is_replicated_storage(false);
+
+    std::shared_ptr<OlapTableSchemaParam> schema_param(new OlapTableSchemaParam());
+    ASSERT_OK(schema_param->init(request.schema()));
+    PTabletWriterOpenResult response;
+    ASSERT_OK(_tablets_channel->open(request, &response, schema_param, false));
 
     PTabletWriterAddChunkRequest add_chunk_request;
     add_chunk_request.mutable_id()->CopyFrom(_load_id);


be/src/runtime/local_tablets_channel.cpp
```diff
@@ -1365,7 +1365,12 @@ void SecondaryReplicasWaiter::_send_replica_status_request(int unfinished_tablet
 #else
     std::pair<PLoadReplicaStatusRequest*, ReusableClosure<PLoadReplicaStatusResult>*> rpc_pair{&request,
                                                                                                _replica_status_closure};
-    TEST_SYNC_POINT_CALLBACK("LocalTabletsChannel::rpc::get_load_replica_status", &rpc_pair);
+    std::function<void(void*)> default_cb = [](void* arg) {
```
kevincai: shall be a `{bool continue, request, closure}` passed down to the CALLBACK, if `continue == true`, continue call the stub and run the following code, else return directly.

This shall be easier than skipping some important steps in UT scenario.

banmoy: don't quite understand, could you give an example? How to set `continue`?

kevincai:
```suggestion
#ifdef BE_TEST
    bool need_return = false;
    std::vector<void *> args = {(void *)&continue, (void *) &request, (void *) _replica_status_closure};
    TEST_SYNC_POINT_CALLBACK("LocalTabletsChannel::rpc::get_load_replica_status", &args);
    // sync point may set `*need_return = true`, so the execution will break immediately.
    if (need_return) {
        return;
    }
#endif
    stub->get_load_replica_status(&_replica_status_closure->cntl, &request, &_replica_status_closure->result,
                                  _replica_status_closure);
```

banmoy: There are two considerations
1.  If the sync point is not the last of the function, `return` directly will not reach the left codes
2.  the callback needs to care about `continue` which is complex

I've considered another option, perhaps a compromise can be made. see 1a2823ecd28e39d74e6bb10ec00693b307910bea

kevincai: 1. the continue/return is specific to a SYNC point definition, the user who defines the sync point shall know clearly the behavior of the callback because of the `(void *) args`, it must be casted to a type that is know in unit test which is an implementation details.

2. The code must be absolutely safe when the sync point is turned off which is the most case. The sync point may be only turned on in some specific cases, but for all other cases, they are turned off. So the code must be favor of most cases, dirty things shall be done in the code path when the sync point is turned on.

be/src/runtime/local_tablets_channel.cpp
```diff
@@ -1359,14 +1359,14 @@ void SecondaryReplicasWaiter::_send_replica_status_request(int unfinished_tablet
...
-    stub->get_load_replica_status(&_replica_status_closure->cntl, &request, &_replica_status_closure->result,
-                                  _replica_status_closure);
-#else
+#ifdef BE_TEST
     std::pair<PLoadReplicaStatusRequest*, ReusableClosure<PLoadReplicaStatusResult>*> rpc_pair{&request,
-                                                                                               _replica_status_closure};
-    TEST_SYNC_POINT_CALLBACK("LocalTabletsChannel::rpc::get_load_replica_status", &rpc_pair);
+                                                                                           _replica_status_closure};
 #endif
+    TEST_SYNC_POINT_CALLBACK_OR_STMT("LocalTabletsChannel::rpc::get_load_replica_status", &rpc_pair,
```
kevincai: This pattern is hard to understand, it takes the regular production code to be nested inside a macro which is purposely defined for unit test.

be/src/runtime/local_tablets_channel.cpp
kevincai: > If the sync point is not the last of the function, return directly will not reach the left codes

The `need_return` is just for this specific case, but it opens the possibility to handle the code specific by the conditions, e.g.

```
Status status;
SYNC_POINT .... (&status);
if (status.Is_xxxx()) {
    // do something;
} else if (status.Is_yyy()) {
    // do something else
} 
if (status.is_ok()) {
    // OK status
    production code here
}
```
Should make sure that the default code path is always safe when sync point is not defined or turned on.

be/src/runtime/local_tablets_channel.cpp
kevincai: The asan memleak can be fixed by adjusting the test case. The sync point issue that the closure doesn't get called if sync point is not enabled in UT mode, will be addressed in a separate PR.

kevincai (approved):

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai (approved):

PR #63297, opened 2025-09-19
[Enhancement] remove the adaptive zonemap creation

murphyatwork:
## What I'm doing:

Remove the "Adaptive ZoneMap creation" function #61960, due to robustness and predictability.

A corner case like this:
```
[A, A]
[A, A]
[A, A]
[A, A]
[A, A]
[A, A]
[B, B]
```

The current adaptive strategy tends to reject this zonemap due to excessive overlap among pages. However, this zonemap can still prove valuable in certain scenarios:
- For queries like `WHERE value = 'B'`, the zonemap can effectively prune most pages, even if the majority of pages contain `A`.
- While the page-level zonemap might be ineffective, the segment-level zonemap can still assist in pruning pages.

In summary, the overlap ratio based strategy is sufficiently robust to handle such cases, especially when the data distribution is skewed.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/rowset/column_writer.cpp
+++ b/be/src/storage/rowset/column_writer.cpp
@@ -393,11 +393,6 @@ Status ScalarColumnWriter::init() {
         if (_opts.zone_map_truncate_string) {
             _zone_map_index_builder->enable_truncate_string();
         }
-        if (is_string_type(_type_info->type())) {
-            _zone_map_index_quality_judger =
-                    ZoneMapIndexQualityJudger::create(_type_info.get(), config::string_zonemap_overlap_threshold,
-                                                      config::string_zonemap_min_pages_for_adaptive_check);
-        }
     }
     if (_opts.need_bitmap_index) {
         _has_index_builder = true;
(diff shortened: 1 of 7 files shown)


be/src/storage/rowset/column_writer.cpp
```diff
@@ -393,11 +393,6 @@ Status ScalarColumnWriter::init() {
         if (_opts.zone_map_truncate_string) {
             _zone_map_index_builder->enable_truncate_string();
         }
-        if (is_string_type(_type_info->type())) {
-            _zone_map_index_quality_judger =
-                    ZoneMapIndexQualityJudger::create(_type_info.get(), config::string_zonemap_overlap_threshold,
```
wyb, line 398: remove `ZoneMapIndexQualityJudger` related code?

wyb (approved):

ZiheLiu (approved):

be/src/storage/rowset/column_writer.cpp
murphyatwork: It’s still worthwhile; I plan to reintroduce it someday when I have a clearer idea.

kevincai (approved):

Issue #62686, opened 2025-09-03
[BugFix] The JOIN USING syntax does not comply with the SQL standard.


trueeyu:
### Steps to reproduce the behavior (Required)
StarRocks
```
mysql> select * from t1 a join t1 b using (c1);
+------+------+
| c1   | c1   |
+------+------+
|    1 |    1 |
+------+------+
1 row in set (0.08 sec)

mysql> select c1 from t1 a join t1 b using (c1);
ERROR 1064 (HY000): Getting analyzing error. Detail message: Column 'c1' is ambiguous.
```

Mysql and other db
```
mysql> select * from t1 a join t1 b using (c1);
+------+
| c1   |
+------+
|    1 |
+------+
1 row in set (0.00 sec)

mysql> select c1 from t1 a join t1 b using (c1);
+------+
| c1   |
+------+
|    1 |
+------+
1 row in set (0.00 sec)
```

### StarRocks version (Required)
 - You can get the StarRocks version by executing SQL `select current_version()`

PR #63312, opened 2025-09-19
[Enhancement] Implement SQL standard JOIN USING with MySQL compatibility

stephen-shelby:
## Why I'm doing:

StarRocks' current JOIN USING implementation does not comply with SQL standard and [MySQL](https://dev.mysql.com/doc/refman/8.4/en/join.html) behavior. The main issues were:

1. **Column ordering**: JOIN USING columns appeared twice in SELECT * results instead of once
2. **Column resolution**: Unqualified USING column references caused ambiguity errors  
3. **RIGHT JOIN semantics**: RIGHT JOIN USING did not follow the correct column ordering and value selection rules
4. **COALESCE behavior**: USING columns did not implement proper COALESCE semantics based on JOIN type

This made StarRocks incompatible with standard SQL and MySQL, causing migration issues for users.

### Part 1: Unqualified USING Column Resolution

Fixed MySQL-compatible behavior when selecting USING columns without table qualification:

**The Issue:**
```sql
-- Test setup:
CREATE TABLE t1 (id INT, name VARCHAR(20));
CREATE TABLE t2 (id INT, dept VARCHAR(20));
INSERT INTO t1 VALUES (1, 'Alice'), (2, 'Bob');
INSERT INTO t2 VALUES (1, 'Engineering'), (3, 'Sales');

-- Before fix (StarRocks error):
SELECT id FROM t1 JOIN t2 USING(id);
-- Error: Column 'id' is ambiguous

-- After fix (MySQL compatible):
SELECT id FROM t1 JOIN t2 USING(id);
-- Works! Returns coalesced id value
-- For INNER/LEFT JOIN: prefers left table's id value
-- For RIGHT JOIN: prefers right table's id value

-- Example with RIGHT JOIN:
SELECT id FROM t1 RIGHT JOIN t2 USING(id);
-- Returns right table's id value when both exist
-- This matches MySQL's COALESCE(t1.id, t2.id) behavior for RIGHT JOIN
```

**Key Point:** Unqualified USING column references (`id`) now resolve to a single coalesced column instead of causing ambiguity errors, following MySQL's COALESCE semantics based on JOIN type.

### Part 2: SELECT * Column Ordering

Implemented MySQL-standard column ordering where USING columns appear only once:

**The Issue:**
```sql
-- Test with different column counts:
CREATE TABLE left_tbl (name VARCHAR(20), id INT, age INT, city VARCHAR(20));  -- 4 columns
CREATE TABLE right_tbl (id INT, dept VARCHAR(20), bonus DECIMAL(8,2));        -- 3 columns

-- Before fix (StarRocks incorrect):
SELECT * FROM left_tbl JOIN right_tbl USING(id);
-- Output: name, id, age, city, id, dept, bonus  (7 columns, id appears twice - WRONG)

-- After fix (MySQL compatible):
SELECT * FROM left_tbl JOIN right_tbl USING(id);
-- Output: id, name, age, city, dept, bonus  (6 columns, id appears once - CORRECT)
-- Order: [USING columns] + [left non-USING] + [right non-USING]

-- RIGHT JOIN has different ordering per MySQL standard:
SELECT * FROM left_tbl RIGHT JOIN right_tbl USING(id);
-- Output: id, dept, bonus, name, age, city  (6 columns)
-- Order: [USING columns] + [right non-USING] + [left non-USING]
```

**Key Point:** SELECT * now follows MySQL's exact column ordering rules:
- **Standard JOINs**: `[USING columns, left table non-USING, right table non-USING]`
- **RIGHT JOIN**: `[USING columns, right table non-USING, left table non-USING]`
- **USING columns appear exactly once**, not duplicated from both tables

### Technical Implementation:

1. **`CoalescedJoinFields`**: Resolves unqualified USING column references to single coalesced fields
2. **`getFieldsForJoinUsingStar()`**: Constructs properly ordered field lists for SELECT * in JOIN USING context
3. **Complete MySQL compatibility**: Both behaviors now match MySQL exactly

Fixes #issue
https://github.com/StarRocks/starrocks/issues/62686

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CoalescedJoinFields.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CoalescedJoinFields.java
@@ -1,0 +1,83 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.sql.analyzer;
+
+import com.google.common.collect.ImmutableList;
+import com.starrocks.sql.ast.expression.JoinOperator;
+import com.starrocks.sql.ast.expression.SlotRef;
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Handles column resolution for JOIN USING clauses per SQL standard.
+ * 
+ * In JOIN USING, each USING column appears only once in the result (coalesced),
+ * unlike JOIN ON where both L.col and R.col are visible. For unqualified 
+ * references to USING columns, this class resolves to the appropriate table's
+ * field based on JOIN type (RIGHT JOIN prefers right table values).
+ * 
+ * Examples:
+ * - SELECT id FROM t1 JOIN t2 USING(id)     -> returns coalesced id column
+ * - SELECT t1.id FROM t1 JOIN t2 USING(id)  -> returns original t1.id column
+ */
+public class CoalescedJoinFields extends RelationFields {
+    private final Set<String> usingColumns;
+    private final JoinOperator joinOperator;
+
+    public CoalescedJoinFields(List<Field> fields, List<String> usingColNames, JoinOperator joinOperator) {
+        super(fields);
+        this.usingColumns = usingColNames.stream()
+                .map(String::toLowerCase)
+                .collect(Collectors.toSet());
+        this.joinOperator = joinOperator;
+    }
+
+    @Override
+    public List<Field> resolveFields(SlotRef name) {
+        String columnName = name.getColumnName().toLowerCase();
+        
+        // For unqualified USING columns, return single coalesced field per SQL standard
+        // This implements the "appears only once" semantics of JOIN USING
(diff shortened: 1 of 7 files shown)


alvin-phoenix-ai:
@cursor review

Seaven:
need compatibility config？

stephen-shelby:
> need compatibility config？

This is a standard behavior. I have checked some databases like mysql/pg/oracle/trino. I don't think it's necessary to use config for compatibility?

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CoalescedJoinFields.java
```diff
@@ -0,0 +1,83 @@
...
+     * This implements COALESCE semantics: 
+     * - RIGHT JOIN: COALESCE(L.col, R.col) -> prefer non-null R.col
+     * - Other JOINs: COALESCE(L.col, R.col) -> prefer non-null L.col
+     */
+    private Field selectCoalescedField(List<Field> allMatches) {
+        if (joinOperator != null && joinOperator.isRightJoin()) {
+            // RIGHT JOIN: right table is primary, prefer its field for coalesced column
+            return allMatches.get(allMatches.size() - 1);
+        } else {
+            // Other JOINs: left table is primary, prefer its field for coalesced column
```
kangkaisen, line 79: what about full outer join?

stephen-shelby: mysql doesn't support full outer join. But I have tested it with other db like trino, and it should use real coalesce semantics, like coalesce(left.id, right.id). The changes are a bit complex. I have recorded a TODO and I'll address this in the next PR

kangkaisen (approved):

Seaven (approved):

PR #63315, opened 2025-09-19
[BugFix] Fix incompatible bitmap index reuse for fast schema evolution in shared-data 

banmoy:
## Why I'm doing:
Fast schema evolution in shared-data modifies a column's type without rewriting data files, and the existing bitmap index is reused for the new type. However, this reuse is not universally applicable because some type conversions alter critical semantics, such as orderliness. For type conversions where the index can be reused, the backend (BE) must perform explicit type casting before querying the index. The current fast schema evolution mechanism does not consider these requirements. This leads to incorrect query results, where data may be erroneously filtered or not filtered. This issue can be reproduced with the following steps:

```
create table t (
    c0 int,
    int_to_bigint int,
    date_to_datetime date,
    int_to_string int,
    index idx1 (int_to_bigint) using bitmap,
    index idx2 (date_to_datetime) using bitmap,
    index idx3 (int_to_string) using bitmap
)
duplicate key(c0)
distributed by hash(c0) buckets 1;

-- disable zonemap index filter to eliminate interference
update information_schema.be_configs set value=false where name='enable_index_page_level_zonemap_filter';
update information_schema.be_configs set value=false where name='enable_index_segment_level_zonemap_filter';

insert into t values (1, -1, '2025-09-12', 5);
insert into t values (2, -2147483648, '2025-09-13', 6);

alter table t modify column int_to_bigint bigint,
    modify column date_to_datetime datetime,
    modify column int_to_string string;

select * from t where int_to_bigint = 21474836479;    -- expect empty. wrong
    - BitmapIndexFilterRows: 1
+------+---------------+---------------------+---------------+
| c0   | int_to_bigint | date_to_datetime    | int_to_string |
+------+---------------+---------------------+---------------+
|    1 |            -1 | 2025-09-12 00:00:00 | 5             |
+------+---------------+---------------------+---------------+


select * from t where date_to_datetime = '2025-09-12 00:00:00';    -- expect one row. wrong
Empty set (0.04 sec)
   - BitmapIndexFilterRows: 2

select * from t where int_to_string = "5";    -- expect one row. wrong
Empty set (0.03 sec)
   - BitmapIndexFilterRows: 2
```

This PR wants to fix the issue.

## What I'm doing:

The issue has existed since version 3.3. Therefore, the fix needs to be backported to earlier versions. To minimize the impact of significant changes on older versions, propose a two-step approach:

1. **Forbid fast schema evolution if a bitmap index is present**: This immediate solution will be applied to older versions. It prevents the incorrect behavior by disabling fast schema evolution in scenarios where bitmap index reuse could lead to errors.
2. **Allow fast schema evolution for specific type conversions**: This step will be implemented in later versions only and will not be backported. It will enable fast schema evolution for certain type conversions once the backend (BE) is fully prepared to handle the associated semantic changes and ensure correct index querying.

This PR implements Step 1, which forbids fast schema evolution when a bitmap index is involved, and will be backported to older versions.

Fixes #issue

## What type of PR is this:

- [X] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [X] 4.0
  - [X] 3.5
  - [X] 3.4
  - [X] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -917,6 +917,23 @@ public class SchemaChangeHandler extends AlterHandler {
             fastSchemaEvolution = false;
         }
 
+        // Check whether manually created indexes support fast schema evolution. The rules are as follows:
+        // 1. The index can be reused after type conversion, and BE has made necessary changes,
+        //    such as casting new type values to old type values before querying the index.
+        // 2. The index cannot be reused, and BE has made changes to skip it during queries.
+        // Currently, BLOOMFILTER and NGRAMBF indexes use rule 2 to support fast schema evolution. BE-side changes
+        // for other indexes are not yet ready, so fast schema change is temporarily disabled for them. This feature
+        // will be gradually enabled for these indexes once BE support is in place.
+        List<Index> existedIndexes = olapTable.getIndexes();
+        for (Index index : existedIndexes) {
+            if (index.getColumns().contains(oriColumn.getColumnId())) {
+                if (index.getIndexType() != IndexDef.IndexType.NGRAMBF) {
+                    fastSchemaEvolution = false;
+                    break;
+                }
+            }
+        }
+
         return fastSchemaEvolution;
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -914,7 +914,7 @@ public class SchemaChangeHandler extends AlterHandler {
         }
 
         if (!SchemaChangeTypeCompatibility.canReuseZonemapIndex(oriColumn.getType(), modColumn.getType())) {
-            fastSchemaEvolution = false;
+            return false;
         }
 
         // Check whether manually created indexes support fast schema evolution. The rules are as follows:
@@ -928,8 +928,7 @@ public class SchemaChangeHandler extends AlterHandler {
         for (Index index : existedIndexes) {
             if (index.getColumns().contains(oriColumn.getColumnId())) {
                 if (index.getIndexType() != IndexDef.IndexType.NGRAMBF) {
-                    fastSchemaEvolution = false;
-                    break;
+                    return false;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -917,6 +917,23 @@ private boolean processModifyColumn(ModifyColumnClause alterClause, OlapTable ol
...
+        //    such as casting new type values to old type values before querying the index.
+        // 2. The index cannot be reused, and BE has made changes to skip it during queries.
+        // Currently, BLOOMFILTER and NGRAMBF indexes use rule 2 to support fast schema evolution. BE-side changes
+        // for other indexes are not yet ready, so fast schema change is temporarily disabled for them. This feature
+        // will be gradually enabled for these indexes once BE support is in place.
+        List<Index> existedIndexes = olapTable.getIndexes();
+        for (Index index : existedIndexes) {
+            if (index.getColumns().contains(oriColumn.getColumnId())) {
+                if (index.getIndexType() != IndexDef.IndexType.NGRAMBF) {
+                    fastSchemaEvolution = false;
```
xiangguangyxg: ealy return false will be better?

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -917,6 +917,23 @@ private boolean processModifyColumn(ModifyColumnClause alterClause, OlapTable ol
             fastSchemaEvolution = false;
```
xiangguangyxg: same as following

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
banmoy: fixed

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
banmoy: fixed

xiangguangyxg (approved):

gengjun-git (approved):

PR #61975, opened 2025-08-15
[Enhancement] create string column zonemap with prefix truncation

murphyatwork:
## Why I'm doing:

To optimize the storage footprint of ZoneMap metadata for string columns by reducing the serialized size of min/max values, addressing issue #61967.

## What I'm doing:

-   Implemented prefix truncation (to 16 bytes) for string min/max values during ZoneMap serialization.
-   For truncated max values, a `0xFF` byte is appended to ensure the value remains a valid upper bound for pruning.
-   Updated existing string ZoneMap tests to reflect and validate the new prefix truncation and padding behavior.

Fixes #61967

## What type of PR is this:

-   [ ] BugFix
-   [ ] Feature
-   [x] Enhancement
-   [ ] Refactor
-   [ ] UT
-   [ ] Doc
-   [ ] Tool

Does this PR entail a change in behavior?

-   [ ] Yes, this PR will result in a change in behavior.
-   [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

-   [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
-   [ ] Parameter changes: default values, similar parameters but with different default values
-   [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
-   [ ] Feature removed
-   [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

-   [x] I have added test cases for my bug fix or my new feature
-   [ ] This pr needs user documentation (for new or modified features or behaviors)
    -   [ ] I have added documentation for my new feature or new function
-   [ ] This is a backport pr

## Bugfix cherry-pick branch check:
-   [x] I have checked the version labels which the pr will be auto-backported to the target branch
    -   [x] 4.0
    -   [ ] 3.5
    -   [ ] 3.4
    -   [ ] 3.3

---
<a href="https://cursor.com/background-agent?bcId=bc-eeef66af-38ad-4c18-8f36-afad98e4b775">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">
    Open in Cursor
  </picture>
</a>
<a href="https://cursor.com/agents?id=bc-eeef66af-38ad-4c18-8f36-afad98e4b775">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg">
[...]

PR #63377, opened 2025-09-22
[BugFix] Fix zone map incorrect filtering after CHAR to VARCHAR fast schema evolution in shared-data

banmoy:
## Why I'm doing

**The Problem**
Zone map filtering fails after fast schema evolution from CHAR to VARCHAR, causing data may be erroneously filtered

**How to reproduce**
```
create table t (c0 int, c1 char(20)) duplicate key(c0) distributed by hash(c0) buckets 1;
insert into t values (1, "abc");
alter table t modify column c1 varchar(200);
select * from t where c1 = 'abc';              -- expect one row. wrong
     - SegmentZoneMapFilterRows: 1
Empty set (0.04 sec)
```

**Why is it happening**
- CHAR values are fixed-length and padded with zeros (e.g., "abc\0\0\0\0\0\0\0")
- VARCHAR values are variable-length without padding (e.g., "abc")
- After fast schema evolution, zone maps still contain CHAR values with padding bytes
- The zone map parsing logic used the predicate type (VARCHAR) instead of the stored data type (CHAR)
- When parsing CHAR values as VARCHAR, `datum_from_string` preserves padding bytes, causing incorrect comparisons with VARCHAR predicates

**Note this problem is introduced since 4.0** by https://github.com/StarRocks/starrocks/pull/61975 which supports to build CHAR/VARCHAR zonemap for non-key columns. **3.5 and previous versions are not affected** because modifying a CHAR/VARCHAR key column would not use fast schema evolution.

## What I'm doing

**The solution:**
Introduce a type-aware zone map parsing mechanism that uses the stored data type (CHAR) when parsing zone map values, even when the predicate type is VARCHAR, ensuring consistent comparison semantics.
- Added `_get_zone_map_parse_type()` in `ColumnReader` to determine the correct parsing type for zone map values
- When the stored column type is CHAR and the predicate type is VARCHAR, force CHAR parsing
- CHAR parsing via `datum_from_string` strips padding bytes, ensuring consistent comparison with VARCHAR predicates
- Applied this logic to both page-level and segment-level zone map filtering

**Key Code Changes:**
- **`be/src/storage/rowset/column_reader.h`**: Added `_get_zone_map_parse_type()` method declaration
- **`be/src/storage/rowset/column_reader.cpp`**: 
  - Implemented `_get_zone_map_parse_type()` with CHAR→VARCHAR handling logic
  - Refactored `_zone_map_filter()` and `segment_zone_map_filter()` to use the new type determination method
- **`be/test/storage/rowset/segment_iterator_test.cpp`**: Added `testCharToVarcharZoneMapFilter()` test covering equality, non-matching, and range predicate scenarios

**How was this fix tested**
- **Unit tests**: Added `testCharToVarcharZoneMapFilter()` covering:
  - VARCHAR equality predicates matching CHAR data
  - VARCHAR predicates that should not match any CHAR data
  - VARCHAR range predicates with CHAR data
- **E2E Tests**:  https://github.com/StarRocks/StarRocksTest/pull/10318

Fixes #issue

## What type of PR is this:

- [X] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [X] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/rowset/column_reader.cpp
+++ b/be/src/storage/rowset/column_reader.cpp
@@ -632,6 +632,17 @@ Status ColumnReader::_zone_map_filter(const std::vector<const ColumnPredicate*>&
         return Status::OK();
     }
 
+    // This addresses a zone map filtering issue that occurs when a column's type is CHAR but the
+    // predicate's type is VARCHAR, often seen after fast schema evolution converts a CHAR column
+    // to VARCHAR. The zone map may still contain CHAR values with padding bytes (e.g., "abc\0\0\0\0\0\0\0").
+    // If these values are parsed as VARCHAR, the padding bytes are preserved, leading to incorrect
+    // comparisons with VARCHAR predicates (e.g., "abc"). By forcing the parsing type to CHAR,
+    // `datum_from_string` in `_parse_zone_map()` strips these padding bytes, ensuring consistent
+    // comparison semantics between zone map entries and predicate values.
+    if (_column_type == TYPE_CHAR && lt == TYPE_VARCHAR) {
+        lt = TYPE_CHAR;
+    }
+
     auto page_satisfies_zone_map_filter = [&](const ZoneMapDetail& detail) {
         if constexpr (PredRelation == CompoundNodeType::AND) {
             return std::ranges::all_of(predicates, [&](const auto* pred) { return pred->zone_map_filter(detail); });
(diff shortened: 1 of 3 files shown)


alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

wyb (approved):

xiangguangyxg (approved):

PR #63575, opened 2025-09-25
[BugFix] fix exception-safety of json extraction

murphyatwork:
## What I'm doing:

The `extract_number`, `extract_bool`, and `extract_string` functions are not exception-safe. They modify `data_column` and `null_column` independently, which can lead to an inconsistent state in the `NullableColumn` if an exception occurs midway.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/util/json_converter.h
+++ b/be/src/util/json_converter.h
@@ -183,10 +183,10 @@ static Status cast_vpjson_to(const vpack::Slice& slice, ColumnBuilder<ResultType
             }
         }
     } catch (const vpack::Exception& e) {
-        LOG(WARNING) << "vpack::Exception in cast_vpjson_to: " << e.what();
         if constexpr (AllowThrowException) {
             return Status::JsonFormatError(fmt::format("cast from JSON to {} failed", type_to_string(ResultType)));
         }
+        LOG(INFO) << "vpack::Exception in cast_vpjson_to: " << e.what();
         result.append_null();
     }
     return Status::OK();
--- a/be/src/util/json_converter.h
+++ b/be/src/util/json_converter.h
@@ -96,6 +96,9 @@ static StatusOr<RunTimeCppType<ResultType>> get_number_from_vpjson(const vpack::
     return Status::JsonFormatError("not a number");
 }
 
+// Converts a JSON value to the specified type and appends it to the result column.
+// Performs type conversion as needed to match the target type.
+// TODO(murphy): it's duplicated with extract_number/extract_bool/extract_string in json_flattener.cpp
 template <LogicalType ResultType, bool AllowThrowException>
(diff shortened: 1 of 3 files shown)


be/src/util/json_converter.h
```diff
@@ -180,6 +183,7 @@ static Status cast_vpjson_to(const vpack::Slice& slice, ColumnBuilder<ResultType
             }
         }
     } catch (const vpack::Exception& e) {
+        LOG(WARNING) << "vpack::Exception in cast_vpjson_to: " << e.what();
```
kevincai: print message only when `allowthrowException` is false.

be/src/util/json_flattener.cpp
```diff
@@ -60,35 +59,44 @@ namespace starrocks {
...
         } else if (json->isBool()) {
-            result->null_column()->append(0);
-            down_cast<RunTimeColumnType<TYPE>*>(result->data_column().get())->append(json->getBool());
+            datum.set(json->getBool());
         } else {
-            result->append_nulls(1);
+            datum.set_null();
         }
     } catch (const vpack::Exception& e) {
+        LOG(WARNING) << "vpack::Exception in extract_number: " << e.what();
```
kevincai: set log level to info or vlog, the exception is not taken as error to the upper level, so it is not worth in the WARN level.

Also avoid a bad json extract causes BE warning log flood.

be/src/util/json_flattener.cpp
```diff
@@ -97,42 +105,50 @@ void extract_bool(const vpack::Slice* json, NullableColumn* result) {
...
         } else if (json->isNumber()) {
             auto res = json->getNumber<double>();
-            result->append_datum(res != 0);
+            datum.set(res != 0);
         } else {
-            result->append_nulls(1);
+            datum.set_null();
         }
     } catch (const vpack::Exception& e) {
+        LOG(WARNING) << "vpack::Exception in extract_bool: " << e.what();
```
kevincai: ditto

be/src/util/json_converter.h
murphyatwork: fixed

be/src/util/json_flattener.cpp
murphyatwork: fixed

kevincai (approved):

satanson (approved):

Seaven (approved):

wyb (approved):

PR #63593, opened 2025-09-26
[Enhancement] Optionally resolve hostname to IP address for errorURL

### LINKED ISSUE #63517: Return BE IP instead of HostName in ErrorURL of StreamLoad
## Enhancement
StreamLoad is the HTTP Request Method to ingest data. When it fails, we will check the response body which contains ErrorURL. However, since our StarRocks Cluster is deployed in K8s env, the ErrorURL cannot be accessed via live network env as it is generated with that internal pod hostname.

chewbum:
## What I did:

This PR introduces a new configuration flag to control hostname resolution in error URLs for Stream Load operations. The enhancement allows operators to choose between using original hostnames from FE heartbeat or forcing resolution to IP addresses based on their environment needs.

**Key Changes:**
- Added `enable_resolve_hostname_to_ip_in_load_error_url` configuration flag (defaults to false)
- Implemented hostname-to-IP resolution logic in BackendOptions with new `get_resolved_ip()` method
- Updated error URL generation in fragment manager and execution state reporter to use resolved IPs when enabled

Fixes #63517 

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [X] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


kevincai:
> Currently, StarRocks always attempts to resolve backend hostnames into IP addresses when generating the ErrorURL in Stream Load responses.

I don't think this is true in current code base. BE uses the `BackendOptions::get_localhost()` which is set by the FE's heartbeat, should be the IP address in non-FQDN mode and should be the dns name in FQDN mode.

chewbum:
@kevincai 
Thanks for pointing this out. In the current codebase:
- In non-FQDN mode, BackendOptions::get_localhost() is already an IP set by the FE’s heartbeat, so no resolution is needed.
- In FQDN mode, BackendOptions::get_localhost() is a hostname (also from the FE heartbeat), and BE will use that directly unless we explicitly resolve it.

My change here is just add configurability:

- Some environments (K8s for example.g.) prefer to keep the hostname in the error URL, since the resolved IP may not be reachable externally
- Other environments (debugging outside the cluster where hostnames are not resolvable) prefer the resolved IP

So the new flag gives operators the choice between using the original hostname from FE heartbeat or forcing resolution to an IP.

fe/fe-core/src/main/java/com/starrocks/common/proc/BackendsProcDir.java
```diff
@@ -69,7 +72,7 @@ public class BackendsProcDir implements ProcDirInterface {
     public static final ImmutableList<String> TITLE_NAMES_SHARED_DATA;
     static {
         ImmutableList.Builder<String> builder = new ImmutableList.Builder<String>()
-                .add("BackendId").add("IP").add("HeartbeatPort")
+                .add("BackendId").add("Address").add("IP").add("HeartbeatPort")
```
duanyyyyyyy: Please name it as IP and realIP

kevincai:
> @kevincai Thanks for pointing this out. In the current codebase:
> 
> * In non-FQDN mode, BackendOptions::get_localhost() is already an IP set by the FE’s heartbeat, so no resolution is needed.
> * In FQDN mode, BackendOptions::get_localhost() is a hostname (also from the FE heartbeat), and BE will use that directly unless we explicitly resolve it.
> 
> My change here is just add configurability:
> 
> * Some environments (K8s for example.g.) prefer to keep the hostname in the error URL, since the resolved IP may not be reachable externally
> * Other environments (debugging outside the cluster where hostnames are not resolvable) prefer the resolved IP
> 
> So the new flag gives operators the choice between using the original hostname from FE heartbeat or forcing resolution to an IP.

Please update the issue and PR description to make it match the actual implementation.

be/src/common/config.h
```diff
@@ -62,6 +62,16 @@ CONF_Int32(brpc_max_connections_per_server, "1");
...
 
+// Whether to resolve backend hostnames to IP addresses in generated error URLs.
+// - true  (default): StarRocks will attempt to resolve hostnames to IPs.
+//  Useful in debugging scenarios where internal hostnames (e.g., K8s pod names)
+//  are not resolvable from the client’s network, so you can curl or open the
+//   error URL directly.
+// - false: Keep the original hostname in the error URL.
+//  Recommended if your DNS setup already allows resolving backend hostnames
+//  across your network, or if IP-based access is unstable (e.g., NAT environments).
+CONF_Bool(enable_resolve_hostname, "false");
```
kevincai:
```
CONF_mBool(enable_resolve_hostname_to_ip_in_load_error_url, "false")
```

allow mutable change during runtime whenever possible unless there is absolute reason for not enabling it.

be/src/exec/pipeline/exec_state_reporter.cpp
```diff
@@ -32,8 +33,25 @@ std::string to_load_error_http_path(const std::string& file_name) {
     if (file_name.empty()) {
         return "";
     }
+
+    std::string host = BackendOptions::get_localhost();
```
kevincai: possibly save the ip address in BackendOptions as well, so it won't need to resolve the hostname everytime.

fe/fe-core/src/main/java/com/starrocks/common/proc/BackendsProcDir.java
```diff
@@ -130,6 +133,7 @@ public static List<List<String>> getClusterBackendInfos() {
         Stopwatch watch = Stopwatch.createUnstarted();
         List<List<Comparable>> comparableBackendInfos = new LinkedList<>();
         long tabletNum = 0;
+        DomainResolver domainResolver = new DomainResolver();
```
kevincai: DnsCache should be the right one to resolve hostname to IP address.

fe/fe-core/src/main/java/com/starrocks/catalog/DomainResolver.java
```diff
@@ -64,6 +64,7 @@ public class DomainResolver extends FrontendDaemon {
     private AuthenticationMgr authenticationManager;
     private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
 
+    public DomainResolver() {}
```
kevincai: this change is not necessary.

fe/fe-core/src/main/java/com/starrocks/common/proc/FrontendsProcNode.java
```diff
@@ -60,7 +63,7 @@ public class FrontendsProcNode implements ProcNodeInterface {
     private static final Logger LOG = LogManager.getLogger(FrontendsProcNode.class);
 
     public static final ImmutableList<String> TITLE_NAMES = new ImmutableList.Builder<String>()
-            .add("Id").add("Name").add("IP").add("EditLogPort").add("HttpPort").add("QueryPort").add("RpcPort")
+            .add("Id").add("Name").add("HostAddress").add("IP").add("EditLogPort").add("HttpPort").add("QueryPort").add("RpcPort")
```
kevincai: this change will break the fe_entrypoint.sh, try not add fields in the middle.

kevincai (changes requested):

kevincai:
I think BE code change is enough, FE side show proc change is not necessary.

be/src/common/config.h
```diff
@@ -62,6 +62,16 @@ CONF_Int32(brpc_max_connections_per_server, "1");
 // The expire time of BRPC stub cache, default 60 minutes.
 CONF_mInt32(brpc_stub_expire_s, "3600"); // 60 minutes
 
+// Whether to resolve backend hostnames to IP addresses in generated error URLs.
+// - true  (default): StarRocks will attempt to resolve hostnames to IPs.
```
kevincai: this is not the default value.

be/src/exec/pipeline/exec_state_reporter.cpp
```diff
@@ -20,6 +20,7 @@
 #include <memory>
 
 #include "agent/master_info.h"
+#include "common/status.h"
```
kevincai: seems not ncessary include header

be/src/runtime/fragment_mgr.cpp
```diff
@@ -43,6 +43,7 @@
 
 #include "agent/master_info.h"
 #include "common/object_pool.h"
+#include "common/status.h"
```
kevincai: not necessary include

be/src/runtime/fragment_mgr.cpp
```diff
@@ -76,8 +77,24 @@ std::string to_load_error_http_path(const std::string& file_name) {
     if (file_name.empty()) {
         return "";
     }
+
+    std::string host = BackendOptions::get_localhost();
+    std::string resolved_ip = host;
+
+    if (config::enable_resolve_hostname_to_ip_in_load_error_url) {
```
kevincai: why here try to resolve the host again? thought it should be good enough to call `BackendOptions::get_resolved_ip();`

fe/fe-core/src/main/java/com/starrocks/catalog/DomainResolver.java
```diff
@@ -195,3 +195,4 @@ public boolean resolveWithBNS(String domainName, Set<String> resolvedIPs) {
     }
 
 }
+
```
kevincai: remove this unnecessary change.

fe/fe-core/src/main/java/com/starrocks/common/proc/BackendsProcDir.java
```diff
@@ -69,7 +70,7 @@ public class BackendsProcDir implements ProcDirInterface {
     public static final ImmutableList<String> TITLE_NAMES_SHARED_DATA;
     static {
         ImmutableList.Builder<String> builder = new ImmutableList.Builder<String>()
-                .add("BackendId").add("IP").add("HeartbeatPort")
+                .add("BackendId").add("IP").add("ResolvedIP").add("HeartbeatPort")
```
kevincai: don't try to add new columns in the middle, will cause unexpected compatible issue if external tools parsing the output as text format.

fe/fe-core/src/main/java/com/starrocks/common/proc/FrontendsProcNode.java
```diff
@@ -190,3 +189,4 @@ private static boolean isJoin(List<InetSocketAddress> allFeHosts, Frontend fe) {
     }
 }
 
+
```
kevincai: revert the change, blank line changes are not necessary for the file.

kevincai (changes requested):

fe/fe-core/src/main/java/com/starrocks/common/proc/BackendsProcDir.java
```diff
@@ -146,7 +148,10 @@ public static List<List<String>> getClusterBackendInfos() {
             watch.stop();
             List<Comparable> backendInfo = Lists.newArrayList();
             backendInfo.add(String.valueOf(backendId));
-            backendInfo.add(backend.getHost());
+            String hostname = backend.getHost();
+            String resolvedIP = DnsCache.tryLookup(hostname);
+            backendInfo.add(hostname);
```
kevincai: `backendInfo` is a list, the position matters. the element in the list should match the position in the Column meta.

Can we revert all the java code change, just keep the cpp code?

be/src/common/config.h
```diff
@@ -62,6 +62,16 @@ CONF_Int32(brpc_max_connections_per_server, "1");
...
 
+// Whether to resolve backend hostnames to IP addresses in generated error URLs.
+// - true: StarRocks will attempt to resolve hostnames to IPs.
+//  Useful in debugging scenarios where internal hostnames (e.g., K8s pod names)
+//  are not resolvable from the client’s network, so you can curl or open the
+//   error URL directly.
+// - false (default) : Keep the original hostname in the error URL.
+//  Recommended if your DNS setup already allows resolving backend hostnames
+//  across your network, or if IP-based access is unstable (e.g., NAT environments).
+CONF_mBool(enable_resolve_hostname_to_ip_in_load_error_url, "false");
```
kevincai, line 73: please update `docs/en/administration/management/BE_configuration.md`, add the new configvar.

docs/en/administration/management/BE_configuration.md
```diff
@@ -5394,3 +5394,12 @@ When this value is set to less than `0`, the system uses the product of its abso
...
 - Introduced in: 3.5.1
+
+##### enable_resolve_hostname_to_ip_in_load_error_url
+
+- Default: false
+- Type: Boolean
+- Unit: -
+- Is mutable: Yes
+- Description: For `error_urls` debugging, whether to allow operators to choose between using original hostnames from FE heartbeat or forcing resolution to IP addresses based on their environment needs.
+- Introduced in: -
```
EsoragotoSpirit: Please specify the introduced-in version.

kevincai: shall be  from v4.0.1

EsoragotoSpirit (approved):

kevincai (approved):

trueeyu (approved):

PR #63694, opened 2025-09-29
[BugFix] add forced exit fatal signal handlers for CN running as PID 1(in dock…

cbcbq:
…er) to guarantee termination on fatal signals

## Why I'm doing:

When CN runs as PID 1 inside a Kubernetes container, fatal signals (SIGSEGV, SIGABRT, SIGILL, SIGFPE, SIGBUS) may not terminate the process reliably after glog’s handler re-raises them. This can leave the process in a corrupted, semi-dead state, preventing the container from restarting and impairing recovery and observability.

## What I'm doing:

For CN processes detected as PID 1:
Preserve and invoke the original glog handler once.
Call _exit(128 + sig) unconditionally to guarantee immediate termination.
Apply only to a fixed fatal signal set.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/common/daemon.cpp
+++ b/be/src/common/daemon.cpp
@@ -35,6 +35,9 @@
 #include "common/daemon.h"
 
 #include <gflags/gflags.h>
+#include <unistd.h>
+
+#include <csignal>
 
 #include "column/column_helper.h"
 #include "common/config.h"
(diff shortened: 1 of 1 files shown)


be/src/common/daemon.cpp
```diff
@@ -264,6 +266,43 @@ void init_signals() {
     if (ret < 0) {
         exit(-1);
     }
+
+    // For CN running as PID 1 inside minimal containers: default fatal signal
```
kevincai, line 271: can you elaborate a little bit about the repro step?

cbcbq:
``` Dockerfile
FROM rockylinux:8-minimal
Workdir /usr/local/java

RUN microdnf install -y bash util-linux procps-ng numactl which wget tar \
    && microdnf clean all

# install your jdk here

# ENV JAVA_HOME=/usr/local/java/*****

# starrocks-*.tar.gz is the tarball contains compiled starrocks be
ADD starrocks-*.tar.gz /data
RUN ln -sf /data/starrocks-* /data/starrocks && mkdir -p /data/starrocks/fe/meta /data/starrocks/be/storage && touch /.dockerenv
Workdir /data/starrocks

ENTRYPOINT ["/data/starrocks/be/bin/start_cn.sh"]
CMD ["--logconsole"]
```

docker build && docker run -d this image
then `docker exec cn1 readlink /proc/1/exe` will be starrocks_be
send SIGABRT to this process by `docker kill --signal=SIGABRT cn1`
you will see something like below showing the process is not terminated but in a sick state

image

kevincai: I am able to repro the issue, thinking of proper fixes. Trying to check if this is container specific or will be the same in VM-based process. Wondering if we have something missing in daemon.cpp when installing signal handler.

kevincai: Thinking about if [tini](https://github.com/krallin/tini) can solve the problem if the PID 1 is tini instead of starrocks_be itself.

cbcbq: yes, infact tini is the best way to handle all signal or subprocess issues
if you prefer to let user to handle this by themselves but not to implement a minimal default behavior wrapper?
I will close this pr

kevincai: yeah, I think we shall follow the best practice, let tini do whatever it should as PID 1. I am working on a PR to turn to tini as the entrypoint for official docker image.

kevincai: https://github.com/StarRocks/starrocks/pull/63781

PR #63726, opened 2025-09-30
[Enhancement] use crc hash for exchange op when hash partition

before-Sunrise:
## Why I'm doing:
fnv hash is much slower then crc hash.

## What I'm doing:
use crc hash when sv is set. So upgrade is safe

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/exec/pipeline/exchange/exchange_sink_operator.cpp
+++ b/be/src/exec/pipeline/exchange/exchange_sink_operator.cpp
@@ -581,9 +581,17 @@ Status ExchangeSinkOperator::push_chunk(RuntimeState* state, const ChunkPtr& chu
 
             // Compute hash for each partition column
             if (_part_type == TPartitionType::HASH_PARTITIONED) {
-                _hash_values.assign(num_rows, HashUtil::FNV_SEED);
-                for (const ColumnPtr& column : _partitions_columns) {
-                    column->fnv_hash(&_hash_values[0], 0, num_rows);
+                // Check if we should use crc32_hash instead of fnv_hash based on session variable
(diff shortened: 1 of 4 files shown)


be/src/exec/pipeline/exchange/exchange_sink_operator.cpp
```diff
@@ -581,9 +581,17 @@ Status ExchangeSinkOperator::push_chunk(RuntimeState* state, const ChunkPtr& chu
 
             // Compute hash for each partition column
             if (_part_type == TPartitionType::HASH_PARTITIONED) {
-                _hash_values.assign(num_rows, HashUtil::FNV_SEED);
-                for (const ColumnPtr& column : _partitions_columns) {
-                    column->fnv_hash(&_hash_values[0], 0, num_rows);
+                // Check if we should use crc32_hash instead of fnv_hash based on session variable
```
alvin-phoenix-ai, line 584: What hash function does starrocks use for aggregation. 
The hash functions of aggregation and shuffle should be different, otherwise, it will cause a lot of hash conflicts in aggregation or hash-join. It will cause a much bigger hash table than it should be.
So when you change it, you must make sure that they are using two different functions.

alvin-phoenix-ai (changes requested):
I put a request changes first before the concerns are resolved.

PR #63841, opened 2025-10-09
[BugFix] Fix incorrect errmsg when dictionary has error

srlch:
## What I'm doing:
This pr fix two issues:
1. Fix FE throw NPE if some CN/BE node has been dead when SHOW DICTIONARY.
2. Fix errmsg loss when the refresh dictionary cache has failed.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Propagates dictionary cache RPC errors via return values and surfaces them in SHOW DICTIONARY, preventing NPEs when nodes fail.
> 
> - **Catalog/DictionaryMgr**:
>   - Change `processDictionaryCacheInteranl` to return `Pair<Boolean, String>` with explicit error message; update all call sites (`clearDictionaryCache`, refresh worker `begin/commit`).
>   - Update `getDictionaryStatistic` to return `Pair<Map<TNetworkAddress, PProcessDictionaryCacheResult>, String>`; propagate errors and validate results.
>   - Enhance `getAllInfo` to handle missing stats and append error message instead of risking NPE.
> - **Tests**:
>   - Add unit tests for error surfacing and robustness: show dictionary error handling, clear cache, statistic success, process exception path, and refresh worker flow; update mocks accordingly.
> 
[...]

PR #64002, opened 2025-10-14
[BugFix] Fix show dictionary when some BE/CN nodes are disconnected #63982

jlon:
## Why I'm doing:
- show dictionary currently fails when any BE/CN node is disconnected
- users receive an opaque “Unknown error” instead of useful diagnostics
- `mysql> show dictionary;
ERROR 1064 (HY000): Getting analyzing error. Detail message: begin 0, end -2, length 0.`

## What I'm doing:
- skip unavailable BE/CN nodes when collecting dictionary cache statistics
- surface error messages for nodes that fail RPC calls instead of aborting the whole command
- adjust show dictionary output formatting to tolerate missing per-node data
- add unit test coverage for available/unavailable node handling

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/DictionaryMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/DictionaryMgr.java
@@ -63,6 +63,7 @@ import com.starrocks.sql.parser.ParsingException;
 import com.starrocks.sql.plan.ExecPlan;
 import com.starrocks.system.Backend;
 import com.starrocks.system.ComputeNode;
+import com.starrocks.system.SystemInfoService;
 import com.starrocks.thrift.TNetworkAddress;
 import com.starrocks.thrift.TStatusCode;
 import com.starrocks.thrift.TUniqueId;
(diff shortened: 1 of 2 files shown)


srlch:
https://github.com/StarRocks/starrocks/pull/63841 fixed in this pr

eshishki:
The core issue here — `show dictionary` failing with the opaque `begin 0, end -2, length 0` error when a node is down — was already fixed by @srlch's #63841 (merged the same day this was opened, as noted above). On current `main` the command no longer aborts; it surfaces a readable `Can not get memory info, errMsg: ...` per dictionary instead.

The remaining delta in this PR is returning partial stats from the live nodes rather than a clear "can't get memory info" message — and that's debatable, since silently-partial cache stats can be more misleading than an explicit error when a node is unreachable. Given the original bug is fixed and this has gone stale (`DIRTY`), probably cleanest to close it. Thanks @jlon.

PR #64133, opened 2025-10-16
[Enhancement] support global late materialization on iceberg v3

### LINKED ISSUE #60015: support global late materialization

silverbullet233:
## What I'm doing:

Fixes #60015

In this PR, I implemented global late materialization based on iceberg v3. The basic idea can be referred to the description in the above issue.

Next, I'll explain the basic design and implementation, focusing on several key issues.

### Key Points of the Global Late Materialization Framework

Given that StarRocks supports querying data from various data sources, global late materialization theoretically needs to work on any data source, so the design needs to be easily extensible.

#### How to Represent RowPosition?

RowPosition describes how to find a row. Although different data sources may represent RowPosition differently, they all represent a combination of one or more columns. We introduce the `RowPositionDescriptor` to describe this relationship.

`RowPositionDescriptor` consists of two parts: one row source slot  and some ref slots. 

If a table has columns that require lazy reads, the table generates several columns to describe the position during the early read phase. One column describes the compute node where the data originated (the column pointed to by the row source slot). During a lazy read, the request is sent to the corresponding compute node for processing. The columns pointed to by the ref slots are combined to logically uniquely identify a row. For example, suppose we want to design a RowPositionDescriptor for a Primary Key Table with primary keys a and b. The ref slots should include slots corresponding to columns a and b.

This information allows us to determine how to find a particular row of data.

#### How to generate a global late materialization execution plan?

During the final stage of query optimization, the `LateMaterializationRewriter` is responsible for this task.

The rewriting process is roughly divided into two steps:

1. The `ColumnCollector` collects all columns that require lazy reads, along with their first usage locations, and determines the timing of lazy reads. This information is recorded in the `CollectorContext`.

2. The `PlanRewriter`, based on the `CollectorContext` information, performs the actual physical plan rewriting, inserting Fetch and LookUp operators at specific locations to perform lazy reads.

#### How are Fetch and LookUp operators implemented?

During the execution phase, the Fetch operator initiates  LookUpRequests to the LookUp operator, retrieving specific data based on the rowPosition. The LookUp operator processes the LookUpRequest. To ensure scalability, we introduced several new abstractions. Each data source needs to implement its own implementation for each of these:

- FetchTask: Responsible for executing fetch operations and sending LookupRequests to specific lookup operators.

- FetchTaskContext: Used to record the context required by the FetchTask.

- LookUpTask: Responsible for executing lookup operations and reading data from a specific data source.

- LookUpTaskContext: Used to record the context required by the LookUpTask.

The core logic of the Fetch and LookUp operators is implemented by FetchProcessor and LookupProcessor. To minimize IO costs, they each have specific batching process strategies.

### how to implement the global lazy materialization on Iceberg?

Based on the above design, the key issues are how to represent row position and how to perform lazy reads.

In Iceberg v3, row position consists of three columns:

- row_source_id: The compute node that generated the row data.

- scan_range_id: The identifier of the ScanRange that generated the row data. During the early read phase, the `HiveConnector` generates a unique ID for each scan range. 

- Row id: A reserved field introduced in Iceberg v3 that uniquely identifies a row. 

with the row_source_id and scan_range_id, we know how to create a `HiveDataSource` for lazy reads. We also create predicates for the _row_id columns we want to read and pass them into the HiveDataSource. 
Predicates for multiple - _row_id columns are concatenated using OR. For example, if we need to read _row_ids in [1, 10) and [20, 30), the generated predicate would be `(_row_id >= 1 and _row_id < 10) or (_row_id >= 20 and _row_id < 30)`.

### Performance Improvements

I constructed three typical queries based on a SSB 100g dataset for testing. They all have a common feature: they return a large number of columns, but only a few columns are involved in joins/sorts, resulting in a small number of rows returned.

The test environment included one 8-core 16GB FE node and three 16-core 64GB BE nodes.
baseline is the result of executing with default parameters.
opt is the result of enabling global late materialization (set enable_global_late_materialization=true).

```sql
--baseline: 12.74s
--opt 3.49s
select * from lineorder join part on lo_partkey = p_partkey where (p_size + lo_linenumber) > 60;

--baseline: 9.02s
--opt: 1.71s
select * from lineorder order by lo_partkey limit 100000, 5;

--baseline: 9.88s
--opt: 2.85s
select * from lineorder, customer where lineorder.lo_custkey = customer.c_custkey order by lineorder.lo_custkey limit 10;
```

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Introduce a global late materialization framework (initially for Iceberg v3), adding Fetch/LookUp operators, planner rewrites, execution processors/tasks, RPCs, and configs to defer wide-column reads until needed.
> 
> - **Planner/Frontend**:
>   - Add `RowPositionDescriptor`, session var `enable_global_late_materialization`, and plan rewriter (`LateMaterializationRewriter`).
>   - Introduce physical ops `PhysicalFetch`/`PhysicalLookUp` and plan nodes `FetchNode`/`LookUpNode`; integrate into fragment assignment and explain.
>   - Extend Hive/Iceberg planning; minor debug/printing updates.
> - **Execution/Backend**:
>   - Implement Fetch/LookUp pipeline operators (`fetch_sink/source`, `lookup_operator`) with `FetchProcessor` and `LookUpProcessor` plus `FetchTask`.
>   - Add global LM context manager (`glm_manager`) in query context.
>   - Add `NoopSink` and support in `DataSink`.
>   - Enhance `Chunk::debug_columns` for empty and extra metadata.
> - **Connector/Scan**:
>   - Wire GLM through Hive connector; assign and propagate `scan_range_id`; initialize global LM context.
>   - HDFS scanner marks reserved fields (`_row_source_id`, `_scan_range_id`).
> - **Formats/Parquet**:
>   - Add `RowSourceReader`; pass `scan_range_id`; enable `FixedValueColumnReader::read_range`.
> - **Runtime/RPC**:
>   - Add lookup dispatcher/manager and new RPC `lookup`; service and recoverable stub support.
> - **Config**:
>   - New GLM-related knobs (e.g., `fetch_max_buffer_chunk_num`, `max_lookup_batch_request`, etc.).
> - **Proto/Thrift**:
>   - Define `TRowPositionDescriptor`, `FETCH_NODE`/`LOOKUP_NODE`, `NOOP_SINK`, and `PLookUp{Request,Response}`.
> - **Tests**:
>   - Add SQL tests for GLM (TopN/Join on Iceberg v3) and minor plan feature updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b343b9481bbaca2718721ddd68e69e5dde407946. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


fe/fe-core/src/main/java/com/starrocks/planner/NoopSink.java
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.planner;
+
+import com.starrocks.thrift.TDataSink;
+import com.starrocks.thrift.TDataSinkType;
+import com.starrocks.thrift.TExplainLevel;
+
+public class NoopSink extends DataSink {
```
LiShuMing, line 21: what's this used for?

fe/fe-core/src/main/java/com/starrocks/planner/LookUpNode.java
```diff
@@ -0,0 +1,65 @@
...
+import com.starrocks.thrift.TPlanNode;
+import com.starrocks.thrift.TPlanNodeType;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class LookUpNode extends PlanNode {
```
LiShuMing, line 33: 1. How about add GLM `prefix` for `LookUp/Fetch` Node to ensure those nodes are used for `glm`?

2. is LookUpNode only used for `glm`, can we extend it for `index seek` or `index scan` later?

fe/fe-core/src/main/java/com/starrocks/sql/ast/expression/RowPositionDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+import com.starrocks.planner.SlotId;
+import com.starrocks.thrift.TRowPositionDescriptor;
+import com.starrocks.thrift.TRowPositionType;
+
+import java.util.List;
+
+// describe how to find a row in specific table, only used by global late materialization
+public class RowPositionDescriptor {
+    public enum Type {
+        ICEBERG_V3
```
LiShuMing, line 37: will we support OLAP later?

fe/fe-core/src/main/java/com/starrocks/planner/FetchNode.java
```diff
@@ -0,0 +1,122 @@
...
+import com.starrocks.warehouse.cngroup.ComputeResource;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class FetchNode extends PlanNode {
```
LiShuMing, line 38: What's the FetchNode/LookUpNode's uses and differences, maybe we can add some docs?

be/src/exec/pipeline/lookup_request.cpp
```diff
@@ -0,0 +1,643 @@
...
+        literal_end.__isset.int_literal = true;
+        nodes.push_back(literal_end);
+    }
+
+    return expr;
+}
+
+// Scans the Iceberg storage engine for the calculated row-id ranges and
+// accumulates the resulting columns.
+StatusOr<ChunkPtr> IcebergV3LookUpTask::_get_data_from_storage(
```
LiShuMing, line 521: Is it better to split Iceberg Lookup implementations from this since we may support Olap lookup later?

be/src/exec/pipeline/lookup_request.cpp
```diff
@@ -0,0 +1,643 @@
...
+    int_literal_10.value = end;
+    literal_10.int_literal = int_literal_10;
+    literal_10.__isset.int_literal = true;
+
+    nodes.push_back(literal_10);
+
+    return expr;
+}
+
+TExpr IcebergV3LookUpTask::create_row_id_filter_expr(SlotId slot_id, const SparseRange<int64_t>& row_id_range) {
```
LiShuMing, line 357: do we need to construct this expr for each lookup call?

gensrc/thrift/InternalService.thrift
```diff
@@ -350,6 +350,9 @@ struct TQueryOptions {
   191: optional i64 column_view_concat_bytes_limit;
 
   200: optional bool enable_full_sort_use_german_string;
+  
+  210: optional bool enable_global_late_materialization;
```
satanson, line 354: why need this option?  it seems that plan supports GLM always contain FetchNode and LookUpNode, so can BE
recognize GLM plan?

gensrc/thrift/DataSinks.thrift
```diff
@@ -58,7 +58,8 @@ enum TDataSinkType {
     BLACKHOLE_TABLE_SINK,
     DICTIONARY_CACHE_SINK,
     MULTI_OLAP_TABLE_SINK,
-    SPLIT_DATA_STREAM_SINK
+    SPLIT_DATA_STREAM_SINK,
+    NOOP_SINK
```
satanson, line 62: there already exists pipeline-level operator noop_sink_operator, it is confusing.  Dose Fragment containing LookUpNode always use NOOP_SINK as DataSink?

gensrc/proto/internal_service.proto
```diff
@@ -740,6 +740,26 @@ message PUpdateTransactionStateResponse {
...
+    optional int64 data_size = 2;
+    optional bytes data = 3;
+}
+
+message PLookUpRequest {
+    optional PUniqueId query_id = 1;
+    optional int32 lookup_node_id = 2;
+    optional int32 request_tuple_id = 3;
+    repeated PColumn request_columns = 4;
+    repeated PSlotDescriptor lookup_slots = 5;
```
satanson, line 760: can you add some comment to explain which fields are row_source_id, scan_range_id, row_id respectively? 

Is PColumn.slot_id is row_id?  if the row_ids sent to each BE is very sparse, how to batch these sparse row_ids？

gensrc/proto/internal_service.proto
```diff
@@ -740,6 +740,26 @@ message PUpdateTransactionStateResponse {
...
+}
+
+message PLookUpResponse {
+    optional StatusPB status = 1;
+    repeated PColumn columns = 2;
```
satanson, line 765: can we use column-oriented compression or encoding/decoding to minimize both requests and responses?

fe/fe-core/src/main/java/com/starrocks/sql/ast/expression/RowPositionDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    private Type type;
+    private SlotId rowSourceSlot;
+    private List<SlotId> fetchRefSlots;
+    private List<SlotId> lookupRefSlots;
```
satanson, line 54: why fetchRefSlots and lookupRefSlots are different?  scan_range_id is not required actually?

satanson: why fetchRefSlots and lookupRefSlots are different?  scan_range_id is not required actually?

be/src/exec/fetch_node.cpp
```diff
@@ -0,0 +1,90 @@
...
+        auto* desc = RowPositionDescriptor::from_thrift(row_pos_desc, pool);
+        _row_pos_descs.emplace(tuple_id, desc);
+    }
+
+    _nodes_info = std::make_shared<StarRocksNodesInfo>(tnode.fetch_node.nodes_info);
+}
+
+FetchNode::~FetchNode() {
+    if (runtime_state() != nullptr) {
+        close(runtime_state());
```
satanson, line 47: runtime_state is a object of fragment-instance lifecycle, it should be closed in fragment_context.

be/src/exec/fetch_node.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+Status FetchNode::init(const TPlanNode& tnode, RuntimeState* state) {
+    RETURN_IF_ERROR(ExecNode::init(tnode, state));
+    std::vector<TupleId> tuple_ids;
```
satanson, line 53: here tuple_ids is a vector, in real cases, can this vector contain more than one tuples, can a FetchNode send requests to multiple LookupNodes on more than one tables?

be/src/exec/lookup_node.cpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+    ExecNode::close(state);
+}
+
+pipeline::OpFactories LookUpNode::decompose_to_pipeline(pipeline::PipelineBuilderContext* context) {
+    int32_t max_io_tasks = context->degree_of_parallelism();
+    auto lookup_op = std::make_shared<pipeline::LookUpOperatorFactory>(context->next_operator_id(), id(),
+                                                                       _row_pos_descs, _dispatcher, max_io_tasks);
+
+    lookup_op->set_degree_of_parallelism(1);
```
satanson, line 76: Dop of LookUp is one, if there or many scan ranges,  can only one driver service these scan ranges?

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/planner/NoopSink.java
silverbullet233: In our design, each Fragment must have a destination. The fragment where the LookUpNode is located actually has no upstream or downstream. The NoopSink is added here just to construct a valid fragment, you can treat it as a placeholder:)

fe/fe-core/src/main/java/com/starrocks/planner/LookUpNode.java
silverbullet233: 1. The prefix isn't necessary; I added a comment to clarify their purpose.
2. maybe yes... In theory, we could simply extend the LookUpTask implementation to support different lazy read methods.

fe/fe-core/src/main/java/com/starrocks/planner/FetchNode.java
silverbullet233: added some descriptions

be/src/exec/pipeline/lookup_request.cpp
silverbullet233: this is a specific implementation for Iceberg v3. If we want to support olap table, we need to implement a special OlapTableLookUpTask.

be/src/exec/pipeline/lookup_request.cpp
silverbullet233: Yes, because each lookup requires the different row ids. This is done just to avoid hacking into the HiveDataSource, but it's not elegant. In the future, we may consider adding a row_id_ranges option to avoid constructing thrift.

gensrc/thrift/InternalService.thrift
silverbullet233: during the prepare phase, we need to create a glm context for the query that enables global late materialization. This option makes it much easier.

gensrc/thrift/DataSinks.thrift
silverbullet233: In our design, each Fragment must have a destination. The fragment where the LookUpNode is located actually has no upstream or downstream. The NoopSink is added here just to construct a valid fragment. I know that there is already an implementation of noop_sink_operator on BE, but it is only used by olap prepare itself and not exposed to FE

be/src/exec/fetch_node.cpp
silverbullet233: A FetchNode only sends  requests to one LookUpNode, but a FetchNode may get data from multiple tables, that is, there may be multiple tuple_ids. A typical example is the FetchNode after join, which may read data from both the build side and the probe side tables at the same time.

be/src/exec/lookup_node.cpp
silverbullet233: The lookup operator is only responsible for scheduling the lookup task. The actual heavy operation is done in the lookup task, i think dop=1 is acceptable

gensrc/proto/internal_service.proto
silverbullet233: yes, I will support it in later pr.

fe/fe-core/src/main/java/com/starrocks/sql/ast/expression/RowPositionDescriptor.java
silverbullet233: Logically, they are the same column, and we do not allow duplicate slot IDs, so different slot IDs are needed during fetch and lookup. `scan_range_id` is simply a fetch ref column in Iceberg v3, which can be obtained from fetch ref slots.

be/src/exec/fetch_node.cpp
silverbullet233: runtime_state is an input parameter of close method

satanson (dismissed):

LiShuMing (dismissed):

stephen-shelby (dismissed):

alvin-phoenix-ai:
@cursor review

trueeyu (dismissed):

alvin-phoenix-ai:
@cursor review

stephen-shelby (dismissed):

silverbullet233:
@mergify rebase

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

silverbullet233:
@mergify rebase

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai (commented):
This PR is huge, could you split this PR into several smaller PRs?

LiShuMing (approved):
+1

stephen-shelby (approved):

ZiheLiu (approved):

trueeyu (approved):

satanson (approved):

PR #64319, opened 2025-10-21
[BugFix] Consider the replica status when choosing a replica for backup

srlch:
# Why I'm doing:
In some cases, if the backend node is not available, the replica status will be DEAD.
But the replica has not any invalid version info.
Some the backup job will choose this replica for backup, but the backup job will fail.

# What I'm doing:
Add a check to the replica status when choosing a replica for backup.
If the replica status is DEAD, skip this replica.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/Replica.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/Replica.java
@@ -52,7 +52,8 @@ public class Replica implements Writable {
 
     public static final int DEPRECATED_PROP_SCHEMA_HASH = 0;
 
-    public static ReplicaStatus computeReplicaStatus(Replica replica, SystemInfoService infoService, long visibleVersion, int schemaHash) {
+    public static ReplicaStatus computeReplicaStatus(Replica replica, SystemInfoService infoService,
+                                                     long visibleVersion, int schemaHash) {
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/Replica.java
```diff
@@ -50,6 +52,21 @@ public class Replica implements Writable {
 
     public static final int DEPRECATED_PROP_SCHEMA_HASH = 0;
 
+    public static ReplicaStatus computeReplicaStatus(Replica replica, SystemInfoService infoService,
```
xiangguangyxg: Use a member function instead of a static function will be better?

srlch: Fixed

xiangguangyxg (approved):

wyb (approved):

nshangyimingCD (approved):

PR #64443, opened 2025-10-23
[Enhancement] disable collecting table stats when single iceberg/deltalake table

dirtysalt:
## Why I'm doing:

Since we enable collecting table stats from iceberg by default. https://github.com/StarRocks/starrocks/pull/64140. Every query will trigger collecting table stats.

But somehow, it's not reasonable if you just query a single table. And this is especially necessary if user just wants to preview data using `select * from T limit 10`.  if T is a very large table, then it  will take a long time to collect table stats but it's unnecessary.

## What I'm doing:

This PR is to:
- analyze query how many updated tables and source tables
- when query table stats of a iceberg table, and # of source table is 1, just return default row count.
- this could also works for delta lake table I think.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AnalyzerUtils.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AnalyzerUtils.java
@@ -839,7 +839,7 @@ public class AnalyzerUtils {
         new AnalyzerUtils.OlapTableCollector(olapTables).visit(statementBase);
     }
 
-    public static void collectSourceTables(StatementBase statementBase, Set<Table> sourceTables) {
+    public static void collectSourceTables(StatementBase statementBase, List<Table> sourceTables) {
         new SourceTablesCollector(sourceTables).visit(statementBase);
     }
 
@@ -1031,9 +1031,9 @@ public class AnalyzerUtils {
     }
 
     private static class SourceTablesCollector extends TableCollector {
-        private Set<Table> sourceTables;
+        private List<Table> sourceTables;
 
-        public SourceTablesCollector(Set<Table> sourceTables) {
+        public SourceTablesCollector(List<Table> sourceTables) {
             super();
             this.sourceTables = sourceTables;
         }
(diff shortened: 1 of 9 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AnalyzerUtils.java
```diff
@@ -1026,6 +1030,22 @@ private Table copyTable(Table originalTable) {
...
+
+        public SourceTablesCollector(Set<Table> sourceTables) {
+            super();
+            this.sourceTables = sourceTables;
+        }
+
+        @Override
+        public Void visitTable(TableRelation node, Void context) {
+            super.visitTable(node, context);
+            sourceTables.add(node.getTable());
```
stephen-shelby, line 1044: should we also collect stats when joining the same table?

LiShuMing: 1. `collect stats` seems fine if it is a load rather a query?
2. is that possible to trigger load stats async rather than sync for latter use?

dirtysalt: yes, I hope so.

dirtysalt: we have a background stats collecting job.  and this job will be triggered always.

but what I hope is not to collect stats from iceberg metaata, it's time consuming.

dirtysalt: @stephen-shelby @LiShuMing I changed to lis not set.

LiShuMing (approved):

wyb (approved):

Seaven (approved):

stephen-shelby (approved):

PR #64680, opened 2025-10-28
[Enhancement] Add session variable to convert decimal multiplication overflow to double

stephen-shelby:
## Why I'm doing:
When performing decimal multiplication operations, if the resulting precision exceeds the maximum decimal precision (38), StarRocks currently returns NULL as the default behavior. This can cause queries to lose meaningful results in scenarios involving high-precision decimal calculations.

Users may prefer to get an approximate result rather than NULL when precision overflow occurs. If users can accept some precision loss, they should have the option to convert the result to double type to maintain a meaningful calculation result instead of getting NULL.

## What I'm doing:
**Added new session variable**: `decimal_overflow_to_double` (default: false)
   - When enabled, decimal multiplication operations that would cause precision overflow (precision > 38) are automatically converted to double type instead of returning NULL
   - When disabled, maintains the original behavior (returns NULL on precision overflow)
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/ScalarOperatorEvaluator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/ScalarOperatorEvaluator.java
@@ -199,6 +199,12 @@ public enum ScalarOperatorEvaluator {
                     operator.getType().getPrimitiveType() != fn.getReturnType().getPrimitiveType()) {
                 Preconditions.checkState(operator.getType().isDecimalOfAnyVersion());
                 Preconditions.checkState(fn.getReturnType().isDecimalOfAnyVersion());
+                if (operator.getType().isDecimal256()) {
+                    Optional<ConstantOperator> res = operator.castTo(fn.getReturnType());
+                    if (res.isPresent() && res.get().isConstantNull()) {
+                        return root;
+                    }
+                }
                 operator.setType(fn.getReturnType());
             }
             return operator;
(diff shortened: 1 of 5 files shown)


satanson (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/ScalarOperatorEvaluator.java
```diff
@@ -199,6 +199,12 @@ public ScalarOperator evaluation(CallOperator root, boolean needMonotonic) {
                     operator.getType().getPrimitiveType() != fn.getReturnType().getPrimitiveType()) {
                 Preconditions.checkState(operator.getType().isDecimalOfAnyVersion());
                 Preconditions.checkState(fn.getReturnType().isDecimalOfAnyVersion());
+                if (operator.getType().isDecimal256()) {
```
kangkaisen, line 202: why need this?

stephen-shelby: Actually, this modification is fixing another bug. After calculating the multiplication of two constants, the result's type will be automatically inferred. When neither of the two constants is decimal256, if the calculation result is decimal256, since we don't support automatic scaling for 128 * 128 = 256, the type will be converted back to 128. Subsequently, when checking the decimal value length, if it exceeds 128, an exception will be thrown.For now, we will maintain the behavior of not automatically scaling. When we uniformly support automatic scaling in the future, we will make the necessary adaptations.

kangkaisen (approved):

PR #61012, opened 2025-07-17
[Enhancement] share file io in iceberg metadata scanner

dirtysalt:
## Why I'm doing:

We came across such excpetion when querying a huge iceberg table

```
Caused by: software.amazon.awssdk.core.exception.SdkClientException: Failed to load credentials from IMDS.
        at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:111)
        at software.amazon.awssdk.core.exception.SdkClientException.create(SdkClientException.java:47)
       ...
        at software.amazon.awssdk.services.s3.DefaultS3Client.getObject(DefaultS3Client.java:5848)
        at org.apache.iceberg.aws.s3.S3InputStream.openStream(S3InputStream.java:240)
        at org.apache.iceberg.aws.s3.S3InputStream.openStream(S3InputStream.java:225)
        at org.apache.iceberg.aws.s3.S3InputStream.positionStream(S3InputStream.java:221)
        at org.apache.iceberg.aws.s3.S3InputStream.read(S3InputStream.java:143)
        at org.apache.iceberg.avro.AvroIO$AvroInputStreamAdapter.read(AvroIO.java:117)
        at org.apache.avro.file.DataFileReader.openReader(DataFileReader.java:66)
        at org.apache.iceberg.avro.AvroIterable.newFileReader(AvroIterable.java:102)
        at org.apache.iceberg.avro.AvroIterable.iterator(AvroIterable.java:77)
        at org.apache.iceberg.io.CloseableIterable$7$1.<init>(CloseableIterable.java:188)
        at org.apache.iceberg.io.CloseableIterable$7.iterator(CloseableIterable.java:187)
        at org.apache.iceberg.io.CloseableIterable.lambda$filter$0(CloseableIterable.java:109)
        at org.apache.iceberg.io.CloseableIterable$2.iterator(CloseableIterable.java:72)
        at org.apache.iceberg.io.CloseableIterable.lambda$filter$1(CloseableIterable.java:136)
        at org.apache.iceberg.io.CloseableIterable$2.iterator(CloseableIterable.java:72)
        at org.apache.iceberg.io.CloseableIterable$7$1.<init>(CloseableIterable.java:188)
        at org.apache.iceberg.io.CloseableIterable$7.iterator(CloseableIterable.java:187)
        at org.apache.iceberg.ManifestReader.iterator(ManifestReader.java:305)
[...]

PR #64999, opened 2025-11-05
[Enhancement] share a single file io in iceberg metadata scanner 

dirtysalt:
## Why I'm doing:

This is the further enhancement of https://github.com/StarRocks/starrocks/pull/61012

Because `get` is not single thread, there is a time window between
- io has been cached already.
- but  `cache.getIfPresent` returns null

and during that time window, io could be reused/shared but is not.

## What I'm doing:

But using `synchronized`, we can guarantee to access the same table, a single IO will be used.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/java-extensions/iceberg-metadata-reader/src/main/java/com/starrocks/connector/iceberg/TableFileIOCache.java
+++ b/java-extensions/iceberg-metadata-reader/src/main/java/com/starrocks/connector/iceberg/TableFileIOCache.java
@@ -18,10 +18,13 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.io.FileIO;
+import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.TimeUnit;
 
 public class TableFileIOCache {
+    public static final org.slf4j.Logger LOG = LoggerFactory.getLogger(AbstractIcebergMetadataScanner.class);
+
     private final Cache<String, FileIO> cache;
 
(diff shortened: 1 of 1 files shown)


stephen-shelby (approved):

Youngwb (approved):

java-extensions/iceberg-metadata-reader/src/main/java/com/starrocks/connector/iceberg/TableFileIOCache.java
```diff
@@ -18,10 +18,13 @@
 import com.google.common.cache.CacheBuilder;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.io.FileIO;
+import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.TimeUnit;
 
 public class TableFileIOCache {
+    public static final org.slf4j.Logger LOG = LoggerFactory.getLogger(AbstractIcebergMetadataScanner.class);
```
murphyatwork, line 26: why using slf4j

dirtysalt: I don't know the difference. But I've tested that it works well with log4j2.properties.

PR #65030, opened 2025-11-06
[Enhancement] Support distinct aggregation over window without sliding frames

satanson:
## Why I'm doing:
Rewrite count(distinct) over window without sliding frames into null-safe-equal join + aggregation
for an exmaple:

```sql
select v1,v2,v3,count(distinct v3) over(partition by v1,v2)
from t;
```

can be rewritten into 
```sql
with cte as(
select v1, v2, count(distinct v3) cd
from t0
group by v1,v2
)
select t0.v1, t0.v2, t0.v3, cte.cd
from t0 inner join cte on t0.v1 <=> cte.v1 and t0.v2 <=> cte.v2;
```

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/DistinctAggregationOverWindowRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/DistinctAggregationOverWindowRule.java
@@ -1,0 +1,192 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.sql.optimizer.rule.transformation;
+
+import com.google.common.base.Functions;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
+import com.starrocks.catalog.FunctionSet;
+import com.starrocks.sql.ast.expression.JoinOperator;
+import com.starrocks.sql.optimizer.OptExpression;
+import com.starrocks.sql.optimizer.OptExpressionVisitor;
+import com.starrocks.sql.optimizer.Utils;
+import com.starrocks.sql.optimizer.operator.AggType;
+import com.starrocks.sql.optimizer.operator.Projection;
+import com.starrocks.sql.optimizer.operator.logical.LogicalAggregationOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalCTEAnchorOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalCTEConsumeOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalCTEProduceOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalJoinOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalWindowOperator;
+import com.starrocks.sql.optimizer.operator.scalar.BinaryPredicateOperator;
+import com.starrocks.sql.optimizer.operator.scalar.CallOperator;
+import com.starrocks.sql.optimizer.operator.scalar.ColumnRefOperator;
+import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
+import com.starrocks.sql.optimizer.rewrite.ReplaceColumnRefRewriter;
+import com.starrocks.sql.optimizer.rule.tree.TreeRewriteRule;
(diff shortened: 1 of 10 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/DistinctAggregationOverWindowRule.java
```diff
@@ -0,0 +1,176 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class DistinctAggregationOverWindowRule implements TreeRewriteRule {
```
murphyatwork, line 65: Could you include an example in the description of this rule? This way, when I read the code, I can intuitively understand the purpose of the rule.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/transformer/WindowTransformer.java
```diff
@@ -223,6 +224,10 @@ public static WindowOperator standardize(AnalyticExpr analyticExpr) {
                 partitions.add(partitionExpression);
             }
         }
+        // AnalyticExpr with the iso-window can merged into one WindowOperator;
+        // partition exprs are unordered, the same partition exprs may have
+        // different permutations, so we sort them for normalization.
+        partitions.sort(Comparator.comparing(ExprToSql::toSql));
```
murphyatwork: for performance purpose, it would be better to sort by cardinality

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/DistinctAggregationOverWindowRule.java
murphyatwork: what if the sql is ` select count(distinct x) over (partition by c1) from t group by c2` ?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/transformer/WindowTransformer.java
satanson: fixed, add a new field just for equals and hashCode computation.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/DistinctAggregationOverWindowRule.java
satanson: fixed

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/DistinctAggregationOverWindowRule.java
```diff
@@ -0,0 +1,192 @@
...
+                joinOp.setProjection(windowOp.getProjection());
+            } else {
+                joinOp.setProjection(new Projection(joinColRefMap));
+            }
+            OptExpression cteProduceOptExpr = OptExpression.create(cteProduceOp, child);
+            OptExpression aggOptExpr = OptExpression.create(aggOp, OptExpression.create(cteConsumeOp));
+            OptExpression joinLhsOptExpr = OptExpression.create(joinLhsCteConsumeOp);
+            OptExpression joinOptExpr = OptExpression.create(joinOp, joinLhsOptExpr, aggOptExpr);
+            OptExpression cteAnchorOptExpr = OptExpression.create(cteAnchorOp, cteProduceOptExpr, joinOptExpr);
+
```
LiShuMing, line 173:
```
select v1,v2,v3,count(distinct v3) over(partition by v1,v2) from t;

```
The rewritten result seems as below:

```
with cte as(
  select v1, v2, v3 from t0
)
select t0.v1, t0.v2, t0.v3, cte.cd
from cte 
inner join (select v1, v2, count(distinct v3) cd
  from cte
  group by v1,v2) as agg 
on t0.v1 <=> agg.v1 and t0.v2 <=> agg.v2;
```

Rather than
```
with cte as(
  select v1, v2, count(distinct v3) cd
  from t0
  group by v1,v2
)
select t0.v1, t0.v2, t0.v3, cte.cd
from t0 inner join cte on t0.v1 <=> cte.v1 and t0.v2 <=> cte.v2;
```
am I right?

LiShuMing: They should be equivalient?

LiShuMing (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/DistinctAggregationOverWindowRule.java
satanson: yes, they are equivalent; the sql example in the description is succinct,  they real logical is just as  you write.

Seaven (approved):

PR #65276, opened 2025-11-11
[Enhancement] Make `TXN_BEGIN ` operation idempotency to prevent duplicate node allocation

wxl24life:
## Why I'm doing:

When network issues occur, clients may retry the TXN_BEGIN operation with the same label. Without proper idempotency handling, this could lead to allocating different nodes for the same transaction label, causing inconsistent transaction states and potential resource waste.

## What I'm doing:

Modified the node allocation logic in TransactionLoadAction to check if a label already exists in the coordinator manager cache before allocating a new node for TXN_BEGIN operations. When a TXN_BEGIN request arrives and the label already exists in cache, the system now reuses the existing node instead of allocating a new one

This ensures that network retries of TXN_BEGIN operations are handled idempotently, maintaining transaction consistency

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/http/rest/TransactionLoadAction.java
+++ b/fe/fe-core/src/main/java/com/starrocks/http/rest/TransactionLoadAction.java
@@ -261,9 +261,22 @@ public class TransactionLoadAction extends RestBaseAction {
         TNetworkAddress redirectAddress = result.getRedirectAddress();
         if (null == redirectAddress) {
             // TODO This must be a transaction stream load. Should refactor the code to make it clearer.
-            ComputeNode node = TXN_BEGIN.equals(txnOperation) ?
-                    coordinatorMgr.allocate(label, txnOperationParams.getWarehouseName()) :
-                    coordinatorMgr.get(label, txnOperationParams.getDbName());
+            // For TXN_BEGIN operation, check if the label already exists in coordinatorMgr.
+            // If it does, return the existing node to ensure idempotency in case of network retries.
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/http/rest/TransactionLoadAction.java
```diff
@@ -261,9 +261,22 @@ protected void executeTransaction(BaseRequest request, BaseResponse response) th
         TNetworkAddress redirectAddress = result.getRedirectAddress();
         if (null == redirectAddress) {
             // TODO This must be a transaction stream load. Should refactor the code to make it clearer.
-            ComputeNode node = TXN_BEGIN.equals(txnOperation) ?
-                    coordinatorMgr.allocate(label, txnOperationParams.getWarehouseName()) :
-                    coordinatorMgr.get(label, txnOperationParams.getDbName());
+            // For TXN_BEGIN operation, check if the label already exists in coordinatorMgr.
```
banmoy, line 264: could move this logic to `TransactionLoadCoordinatorMgr::get` ?
```
ComputeNode get(String label, String dbName, boolean allocateIfNotExist)
```

PR #65558, opened 2025-11-14
[BugFix] Query Id Collision on Arrow Flight

chris-celerdata:
## Why I'm doing:
**Issue**: Duplicate query IDs when using multiple threads with Arrow Flight SQL
When multiple threads share a single Arrow Flight SQL connection and execute queries concurrently, they encounter "queryId already exists" errors and receive incorrect/incomplete results.
**Root Cause**:
The `ArrowFlightSqlConnectContext` stores the current query string in a single shared field (query). When multiple concurrent requests call `createPreparedStatement`, they overwrite each other's queries in the shared context:

Thread 1 calls `createPreparedStatement` → sets `ctx.query = query1, ctx.queryId = UUID1`
Thread 2 calls `createPreparedStatement` → overwrites `ctx.query = query2, ctx.queryId = UUID2`
Thread 1 calls `getFlightInfoPreparedStatement` → executes with `query2` instead of `query1`, `UUID2` instead of `UUID1`
Thread 2 calls `getFlightInfoPreparedStatement` → executes with `query2`, `UUID2` (but queryId already exists)
Both threads generate the same execution ID, causing "queryId already exists" errors.

Additionally, `ctx.reset()` calls `removeAllResults()` which clears the result cache while other queries may still be storing results, leading to incomplete data.
## What I'm doing:
**Solution**: Throw an error when query is already set. 

Changes:

Add a lock `queryLock` to `query` in `ArrowFlightSqlConnectContext`. 
`ArrowFlightSqlConnectContext.reset` checks for existence of a currently running query before resetting. If it exists, throws an error. 

Result: On concurrent queries being submitted, only the first will execute. The rest will fail with a descriptive error about a query already being in progress on the connection. 

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Serializes query lifecycle with a lock, introducing initWithStatement/reset, updating service calls, and clearing state on cancel/error; adds tests for new behavior.
> 
> - **ArrowFlightSqlConnectContext**:
>   - Add `ReentrantLock` to guard `query` access; make `getQuery()` thread-safe.
>   - Replace `reset(String)` with `initWithStatement(String)` and parameterless `reset()`; generate IDs in `initWithStatement` and throw `IllegalStateException` if a query is already in progress.
>   - On `cancelQuery()` and error paths, call `reset()`; keep result cache management.
> - **ArrowFlightSqlServiceImpl**:
>   - Use `ctx.initWithStatement(...)` instead of `ctx.reset(query)` in `createPreparedStatement` and `getFlightInfoStatement`.
>   - Call `ctx.reset()` on `closePreparedStatement` and on exceptions in `getFlightInfoFromQuery`.
> - **Tests**:
>   - Update and add unit tests to cover `initWithStatement`, `reset`, single in-flight query enforcement, and state clearing behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 96692adbc1c811835dcb5207072b55479ba758c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlConnectContext.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlConnectContext.java
@@ -88,7 +88,7 @@ public class ArrowFlightSqlConnectContext extends ConnectContext {
         this.returnResultFromFE = true;
     }
 
-    public void reset(String query) {
+    public void initWithStatement(String query) {
         queryLock.lock(); 
         try {
             if (!this.query.isEmpty()) {
@@ -110,7 +110,7 @@ public class ArrowFlightSqlConnectContext extends ConnectContext {
         }
     }
 
-    public void clearQuery() {
+    public void reset() {
         queryLock.lock(); 
         try {
             this.query = ""; 
@@ -181,7 +181,7 @@ public class ArrowFlightSqlConnectContext extends ConnectContext {
     public void cancelQuery() {
         if (executor != null) {
             executor.cancel("Arrow Flight SQL client disconnected");
-            clearQuery();
+            reset();
         }
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlConnectContext.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlConnectContext.java
@@ -45,6 +45,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.locks.ReentrantLock;
 
 // one connection will create one ArrowFlightSqlConnectContext
 public class ArrowFlightSqlConnectContext extends ConnectContext {
@@ -55,6 +56,8 @@ public class ArrowFlightSqlConnectContext extends ConnectContext {
     private StatementBase statement;
 
     private String query;
+
+    private final ReentrantLock queryLock = new ReentrantLock();
 
     private CompletableFuture<Coordinator> coordinatorFuture;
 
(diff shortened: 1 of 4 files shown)


alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlConnectContext.java
```diff
@@ -86,14 +89,34 @@ public ArrowFlightSqlConnectContext(String arrowFlightSqlToken) {
...
+            if (!this.query.isEmpty()) {
+                throw new IllegalStateException(
+                        "Query already in progress on this connection"
+                );
+            }
+
+            removeAllResults();
+            statement = null;
+            coordinatorFuture.complete(null);
+            coordinatorFuture = new CompletableFuture<>();
```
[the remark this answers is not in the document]
chris-celerdata: This is incorrect, the query is only cleared after query execution or if there is an error processing the query.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlConnectContext.java
```diff
@@ -86,14 +89,34 @@ public ArrowFlightSqlConnectContext(String arrowFlightSqlToken) {
     }
 
     public void reset(String query) {
```
alvin-phoenix-ai: Is reset is a good name for this function? what do you think?
I think reset is to reset the class to a fresh state, so it should not have a input argument.
I would like name it `initWithStatement()`, statement is the terminolgy in Arrow Flight proto.
And then you can rename clearQuery to `reset` or `closeStatement`

chris-celerdata: Do you think `reset` (formerly `clearQuery`) should then reset the `statement`, `coordinatorFuture`, `cache`, etc. as well?

alvin-phoenix-ai: Yes, I believe so. after `reset`, the state should be ready for next query as a fresh one. And we need to try to keep all information consistent all the time.

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -169,6 +169,9 @@ public void createPreparedStatement(
     @Override
     public void closePreparedStatement(FlightSql.ActionClosePreparedStatementRequest request, CallContext context,
                                        StreamListener<Result> listener) {
+        String token = context.peerIdentity();
+        ArrowFlightSqlConnectContext ctx = sessionManager.validateAndGetConnectContext(token);
+        ctx.reset();
```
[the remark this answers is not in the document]
chris-celerdata: `closePreparedStatement` is not called until `getStreamResult` is finished according to the [documentation](https://arrow.apache.org/docs/format/FlightSql.html#id3). This is a non-issue.

alvin-phoenix-ai (approved):

ZiheLiu (approved):

PR #65756, opened 2025-11-19
[BugFix] avoid inline non-deterministic CTE

murphyatwork:
## What I'm doing:

```
DROP TABLE IF EXISTS t;
CREATE TABLE t(id INT);

INSERT INTO t (id)
SELECT seq
FROM (
    SELECT 1 AS seq UNION ALL
    SELECT 2 UNION ALL
    SELECT 3 UNION ALL
    SELECT 4 UNION ALL
    SELECT 5 UNION ALL
    SELECT 6 UNION ALL
    SELECT 7 UNION ALL
    SELECT 8 UNION ALL
    SELECT 9 UNION ALL
    SELECT 10
) AS x;

WITH c AS (
    SELECT id
    FROM t
    LIMIT 5
)
SELECT 'A' AS src, id FROM c
UNION ALL
SELECT 'B' AS src, id FROM c;
```

Inlining a non-deterministic CTE (such as one with a LIMIT) may lead to unexpected query results; therefore, we recommend avoiding its inlining.

**Behavior change:**
- Before: CTE with `LIMIT X` can be inlined
- After: CTE with `LIMIT X` will not be inlined

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/ForceCTEReuseRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/ForceCTEReuseRule.java
@@ -17,18 +17,22 @@ package com.starrocks.sql.optimizer.rule.transformation;
 
 import com.starrocks.sql.optimizer.CTEContext;
 import com.starrocks.sql.optimizer.OptExpression;
+import com.starrocks.sql.optimizer.OptExpressionVisitor;
 import com.starrocks.sql.optimizer.OptimizerContext;
 import com.starrocks.sql.optimizer.operator.OperatorType;
 import com.starrocks.sql.optimizer.operator.logical.LogicalCTEProduceOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalLimitOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalTopNOperator;
 import com.starrocks.sql.optimizer.operator.pattern.Pattern;
 import com.starrocks.sql.optimizer.rule.NonDeterministicVisitor;
 import com.starrocks.sql.optimizer.rule.RuleType;
-
 import java.util.Collections;
 import java.util.List;
 
 /**
- * If the opt expression contains non-deterministic function, force cte reuse to avoid producing wrong result.
+ * Force cte reuse to avoid producing wrong result in the following cases:
+ * 1. The opt expression contains non-deterministic function
+ * 2. The opt expression contains LIMIT without ORDER BY (unstable result order)
  */
 public class ForceCTEReuseRule extends TransformationRule {
     public ForceCTEReuseRule() {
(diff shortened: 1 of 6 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/ForceCTEReuseRule.java
```diff
@@ -51,4 +65,46 @@ public List<OptExpression> transform(OptExpression input, OptimizerContext conte
...
+            // This is unstable and should force CTE reuse
+            return true;
+        }
+
+        @Override
+        public Boolean visitLogicalTopN(OptExpression optExpression, Void context) {
+            // LogicalTopNOperator has ORDER BY, so it's stable
+            // Continue checking children but don't mark as unstable
+            return visit(optExpression, context);
+        }
```
Seaven: Limit will merge with TopN, so check limit is enough

murphyatwork: fixed

LiShuMing (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/ForceCTEReuseRule.java
```diff
@@ -38,7 +40,19 @@ public ForceCTEReuseRule() {
...
     public List<OptExpression> transform(OptExpression input, OptimizerContext context) {
+        boolean shouldForceReuse = false;
+
+        // Always force reuse for non-deterministic functions
         if (hasNonDeterministicFunction(input)) {
+            shouldForceReuse = true;
+        }
+
+        // Force reuse for LIMIT without ORDER BY if enabled by session variable
+        if (hasLimitWithoutOrderBy(input) && context.getSessionVariable().isCboCTEForceReuseLimitWithoutOrderBy()) {
```
stephen-shelby: please check session variable first

Seaven (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/ForceCTEReuseRule.java
murphyatwork: doen

stephen-shelby (approved):

satanson (approved):

PR #65793, opened 2025-11-20
[Enhancement] reduce remote storage access when vaccum after file bundle enabled

luohaha:
## Why I'm doing:
1. After file bundle is enabled, when publish version, we will put latest tablet meta in meta cache on every BE. But when auto vacuum run, each partition's vacuum job will only be handled by one BE, so some tablets which aren't belong to this BE will be cache miss when access by vacuum job. And also when call `get_tablet_metadata` in vacuum job, we won't do any cache refill, so there will be lots of data cache and meta cache miss events when handle vacuum job.
2. Whether a tablet uses the file-bundle format is a state stored in the metadata cache. In the current implementation, after the cache is evicted, the logic to refill this state during subsequent reads is missing.

## What I'm doing:
1. refill datacache when call `get_tablet_metadata` in vacuum job, so that tablets belong to one partition can reuse this datacache to reduce remote storage access.
3. Refill file bundle state after access.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/lake/tablet_manager.cpp
+++ b/be/src/storage/lake/tablet_manager.cpp
@@ -446,6 +446,13 @@ StatusOr<TabletMetadataPtr> TabletManager::get_tablet_metadata(int64_t tablet_id
     return get_tablet_metadata(tablet_id, version, cache_opts, expected_gtid, fs);
 }
 
+StatusOr<TabletMetadataPtr> TabletManager::get_tablet_metadata(int64_t tablet_id, int64_t version, bool fill_meta_cache,
+                                                               bool fill_data_cache, int64_t expected_gtid,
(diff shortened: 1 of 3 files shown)


sevev (approved):

wyb (approved):

xiangguangyxg (approved):

kevincai (approved):

luohaha:
https://github.com/Mergifyio backport branch-4.0.2

PR #65838, opened 2025-11-21
[Enhancement] Add ConvertTimeZonePredicateRule to support partition prune

jiangjiguang:
## Why I'm doing:
The convert_tz function causes a full table scan.

## What I'm doing:

select * from t where convert_tz(__time, '+08:00', '+09:00') > '2025-11-21 10:00:00'
->
select * from t where __time > convert_tz('2025-11-21 10:00:00', '+09:00', '+08:00')

**before plan:**
```
0:OlapScanNode
     TABLE: test_time
     PREAGGREGATION: ON
     PREDICATES: convert_tz(1: __time, '+08:00', '+09:00') >= '2025-11-16 10:00:00'
     partitions=0/1
     rollup: test_time
     tabletRatio=0/0
     tabletList=
     cardinality=1
     avgRowSize=2.0
```

**after plan:**
```
0:OlapScanNode
     TABLE: test_time
     PREAGGREGATION: ON
     PREDICATES: 1: __time >= '2025-11-16 09:00:00'
     partitions=0/1
     rollup: test_time
     tabletRatio=0/0
     tabletList=
     cardinality=1
     avgRowSize=2.0
```
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/ScalarOperatorRewriter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/ScalarOperatorRewriter.java
@@ -22,6 +22,7 @@ import com.starrocks.sql.optimizer.operator.scalar.ColumnRefOperator;
 import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
 import com.starrocks.sql.optimizer.rewrite.scalar.ArithmeticCommutativeRule;
 import com.starrocks.sql.optimizer.rewrite.scalar.ConsolidateLikesRule;
+import com.starrocks.sql.optimizer.rewrite.scalar.ConvertTimeZonePredicateRule;
 import com.starrocks.sql.optimizer.rewrite.scalar.ExtractCommonPredicateRule;
 import com.starrocks.sql.optimizer.rewrite.scalar.FoldConstantsRule;
 import com.starrocks.sql.optimizer.rewrite.scalar.ImplicitCastRule;
@@ -55,6 +56,8 @@ public class ScalarOperatorRewriter {
             new SimplifiedDateColumnPredicateRule(),
             new ExtractCommonPredicateRule(),
             new ArithmeticCommutativeRule(),
+            new ConvertTimeZonePredicateRule(),
+
             ConsolidateLikesRule.INSTANCE
     );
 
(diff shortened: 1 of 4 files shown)


eshishki:
The problem here is real — `convert_tz` wrapping a partition column blocks partition pruning and forces a full scan, so rewriting it onto the constant to expose the bare column is a worthwhile optimization. The direction is good.

One correctness concern though: the rewrite assumes `convert_tz` is an invertible constant shift, which only holds for **offset-stable** zones. For zones with DST it isn't — the offset depends on the date of the value:

```sql
-- America/New_York: UTC-5 in winter, UTC-4 in summer
convert_tz(col, 'UTC', 'America/New_York') > '2025-07-01 10:00:00'
  -->  col > convert_tz('2025-07-01 10:00:00', 'America/New_York', 'UTC')
```

The rewritten constant computes a single offset (based on the constant's own date, here summer/-4h), but `col` rows can fall on either side of a DST transition (winter rows need -5h). The rule runs at planning time, where the per-row value is unknown, so it can only bake in one offset — which is correct for fixed-offset zones but wrong near DST boundaries for named zones. (BE's parquet converter handles exactly this split: a fast fixed-`_offset` path vs a per-value `get_timezone_offset_by_timestamp(timestamp, ctz)` path, and the fast one is gated behind `parquet_fast_timezone_conversion` with a "potentially wrong" caveat.)

The repo already has the classification you'd need for a guard:
- `_match_cctz_time_zone` distinguishes fixed-offset formats (`+HH:MM`, `GMT±HH:MM`, `CST`) -> `cctz::fixed_time_zone`, which are always offset-stable;
- `timezone.dat` vs `othertimezone.dat` split DST-free zones from DST zones (only the former go into the precomputed `_s_cached_offsets` cache, with the explicit comment "We cannot add a time zone to the cache that contains both daylight saving time and winter time"), and `TimezoneUtils::timezone_offsets()` returns a constant offset only for that DST-free set.

Suggest gating the rewrite to offset-stable zone pairs (fixed-offset format, or both named zones in the DST-free set) and skipping it otherwise. With that guard it's always sound.

PR #65858, opened 2025-11-21
[Enhancement] support output timezone in be log

murphyatwork:
## What I'm doing:

1. Output timezone like `+0800` in the be.INFO, controlled by switch `sys_log_timezone`.
2. ~Use a shorter thread_id in the log~

```
I20251121 21:26:32.484570 +0800 67435 tablet_updates.cpp:583]  redo edit version log tablet:113028 version:2 rowsets:0 deltas:0 rows
```

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Adds an optional timezone offset to BE log prefixes controlled by `sys_log_timezone`, installing a custom glog prefix formatter.
> 
> - **Logging**:
>   - **Timezone-aware prefix**: Implement custom glog prefix formatter in `be/src/common/logconfig.cpp` to include timezone offset, thread id, and file:line in log prefix.
>     - Computes offset via `get_timezone_offset_string(...)` with thread-safe caching.
>     - Installs formatter only when `config::sys_log_timezone` is enabled.
>   - **Config**: Introduce `CONF_Bool(sys_log_timezone, "false")` in `be/src/common/config.h`.
>   - **Tests/Config**: Enable `sys_log_timezone = true` in `conf/be_test.conf`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 34bae9a96a30872e3560b225f47c4391b6c4d657. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -222,6 +222,8 @@ CONF_Strings(sys_log_verbose_modules, "");
 CONF_Int32(sys_log_verbose_level, "10");
 // The log buffer level.
 CONF_String(log_buffer_level, "");
+// Whether to show timezone in log prefix.
+CONF_Bool(sys_log_timezone, "false");
 
 // Pull load task dir.
 CONF_String(pull_load_task_dir, "${STARROCKS_HOME}/var/pull_load");
--- a/be/src/common/logconfig.cpp
+++ b/be/src/common/logconfig.cpp
@@ -215,8 +215,8 @@ static void failure_function() {
 static std::string get_timezone_offset_string(const google::LogMessageTime& time) {
     // Cache the last offset and its string representation
     static std::mutex cache_mutex;
-    static std::atomic<long> cached_offset_seconds(0);
-    static std::shared_ptr<std::string> cached_tz_str_ptr(std::make_shared<std::string>("+0000"));
+    static std::atomic<long> cached_offset_seconds(-1);
+    static std::shared_ptr<std::string> cached_tz_str_ptr(std::make_shared<std::string>(""));
 
     // Get timezone offset from LogMessageTime
     long offset_seconds = time.gmtoffset().count();
--- a/be/src/common/logconfig.cpp
+++ b/be/src/common/logconfig.cpp
@@ -25,6 +25,7 @@
 #include <pthread.h>
 #endif
 
+#include <atomic>
 #include <cerrno>
 #include <cstdio>
 #include <cstdlib>
@@ -32,6 +33,7 @@
 #include <functional>
 #include <iomanip>
 #include <iostream>
+#include <memory>
 #include <mutex>
 
 #include "cache/datacache.h"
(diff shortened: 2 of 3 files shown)


alvin-phoenix-ai:
@cursor review

be/src/common/logconfig.cpp
```diff
@@ -207,6 +209,58 @@ static void failure_function() {
...
+          << ':' << std::setw(2) << time.sec() << '.' << std::setw(6) << time.usec();
+    } else {
+        // Format: MMDD HH:MM:SS.uuuuuu +HHMM
+        s << std::setfill('0') << std::setw(2) << (1 + time.month()) << std::setw(2) << time.day() << ' '
+          << std::setw(2) << time.hour() << ':' << std::setw(2) << time.min() << ':' << std::setw(2) << time.sec()
+          << '.' << std::setw(6) << time.usec();
+    }
+
+    // Timezone offset (calculated dynamically)
+    s << ' ' << get_timezone_offset_string(time);
```
kevincai, line 284: this calculation will do for every line, wondering if there is perf bench due to this logformat change.

Also this logline format change should be marked as "behavior change".

murphyatwork: 1. This function adds only a few cycles, which is not significant for most cases. And for most performance-critical cases (like evaluating expression on a chunk), it should not use LOG.
2. done

stdpain (dismissed):

be/src/common/config.h
```diff
@@ -222,6 +222,8 @@ CONF_Strings(sys_log_verbose_modules, "");
 CONF_Int32(sys_log_verbose_level, "10");
 // The log buffer level.
 CONF_String(log_buffer_level, "");
+// Whether to show timezone in log prefix.
+CONF_Bool(sys_log_timezone, "false");
```
kevincai, line 226: shall add doc for the new config

be/src/common/logconfig.cpp
```diff
@@ -207,6 +211,87 @@ static void failure_function() {
...
+
+    // Thread ID (simplified, use hash to get a shorter ID)
+    std::hash<std::thread::id> hasher;
+    size_t thread_hash = hasher(message.thread_id());
```
kevincai: the thread_id() can be mapped to the thread hex address if using gdb reading the core. Either keep the thread_id or use the pthread id, this hasher will make it harder to look up logs by the thread id.

murphyatwork: In my opinion, it's better to use gdb to find the correct thread_id rather than relying on the log. The primary purpose of thread_id in logs is to distinguish between logs from different threads, and for that, the hashed thread_id is sufficient.

be/src/common/config.h
murphyatwork: The CI can add the doc automatically

be/src/common/logconfig.cpp
kevincai: It is not rely on log to find thread_id, but find the thread_id from gdb and then grep the log to find the thread activities. The hashed thread_id could mix a few threads to the same id, which could be living at the same time, it is confusing when analysis a thread activity.

kevincai: the key difference is, both thread_id and pthread_id, It could be only one thread alive at any time given the same number. But after hash, it could be multiple threads alive that produces the same hasher number, which looks like a single thread produces multiple logs from unrelated source files.

murphyatwork: ok, let's keep the annoying thread_id

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

kevincai (approved):

stdpain (approved):

murphyatwork:
docgen

PR #65889, opened 2025-11-24
[BugFix] Fix bugs of Arrow Flight SQL

ZiheLiu:
### Sequence diagrams for interactions between client and FE/BE

For native Arrow Flight SQL clients, such as `FlightSqlClient` in Java:

1. `getFlightInfoStatement`: Send SQL to the FE. The FE generates the plan, dispatches fragment instances to BEs, and returns the endpoint from which results can be fetched.
   - For queries: the endpoint is the address of the BE that hosts the result sink.
   - For non-queries: the endpoint is the address of the FE itself.
2. `getStreamStatement`: The client pulls results from the endpoint.

image

For ADBC/JDBC drivers, SQL is executed using the following sequence:

1. `createPreparedStatement`: Create a prepared statement and return a handle (`preparedStmtId`) to the client as its identifier.
2. Execute the query. Note that a single connection can hold multiple prepared statements at the same time.
   1. `acceptPutStatement`: Update the parameter values of a prepared statement.
      - We currently do not implement this RPC, so only non-parameterized SQL is supported.
   2. `getFlightInfoPreparedStatement`: The FE starts executing the query associated with the given handle. The FE generates the plan, dispatches fragment instances to BEs, and returns the endpoint from which results can be fetched.
      - For queries: the endpoint is the address of the BE that hosts the result sink.
      - For non-queries: the endpoint is the address of the FE itself.
   3. `getStreamStatement`: Fetch results from the endpoint returned in the previous step.
3. `closePreparedStatement`: Close a prepared statement.

On the client side, a single cursor only maintains one prepared statement. For the same cursor, when calling `cursor.execute(SQL)`:

- If the SQL is the same as the previous one, it directly runs the three “execute query” steps.
- Otherwise, it first issues `closePreparedStatement`, then creates a new prepared statement via `acceptPutStatement`.

image

### FE/BE Internal Execution Flow

There are four different execution flows, determined by two dimensions:

- Whether the request is forwarded to the leader.
- Whether it is a query or a non-query.

#### Non-forward Query

1. The client sends a `getFlightInfoPreparedStatement` RPC to the FE. After the FE generates the plan and dispatches fragment instances to BEs, it returns the result sink BE endpoint to the client.
2. The client uses the endpoint to send a `getStreamStatement` RPC to the BE and pulls results from the BE.

image

#### Non-forward Non-Query

1. The client sends a `getFlightInfoPreparedStatement` RPC to the FE. The FE’s behavior:
   - For `SHOW xxx` statements, the FE executes everything locally; there is no BE-side execution.
   - For DDL and DML, execution must run on BEs. The FE waits until execution fully completes, and only then returns to the client.
2. The client uses the endpoint to send `getStreamStatement` to the FE and fetch results from the FE.

image

#### Forward Query

1. The follower FE executes in a dedicated executor thread, forwards the request to the leader FE, blocks waiting for the response, and then writes `audit.log` and query detail.
2. After the leader FE finishes deploying fragment instances to BEs, it notifies the follower FE via a new RPC. The follower FE immediately returns the result sink BE endpoint to the client without waiting for step 1’s forward flow to complete.

image

#### Forward Non-Query

1. The follower FE executes in a dedicated executor thread, forwards the request to the leader FE, blocks waiting for the response, and then writes `audit.log` and query detail.
2. The follower FE waits until step 1 fully completes, writes the result into its local cache, and then returns its own FE endpoint to the client.

image

### 1. Column type conversion

| SR Type           | Arrow-Before | Arrow-PR                                        |
|-------------------|--------------|-------------------------------------------------|
| `TYPE_LARGEINT`   | utf8         | `Decimal128Type(38, 0)`                         |
| `TYPE_DATE`       | utf8         | `date32`                                        |
| `TYPE_DATETIME`   | utf8         | `timestamp(arrow::TimeUnit::MICRO, "UTC")`      |
| `TYPE_VARBINARY`  | unsupported  | `binary`                                        |
| `TYPE_DECIMAL256` | unsupported  | `Decimal256`                                    |
| `TYPE_HLL`        | utf8         | `binary` (always returns `NULL`, same as MySQL) |
| `TYPE_OBJECT(BITMAP)` | unsupported | `binary` (always returns `NULL`, same as MySQL) |
| `TYPE_PERCENTILE` | unsupported  | `binary` (always returns `NULL`, same as MySQL) |

The mapping from SR types to Arrow types is used not only by Arrow Flight SQL, but also by import/export. For compatibility reasons, only Arrow Flight SQL changes the mapping above.

See `row_batch.h/cpp`, `starrocks_column_to_arrow.cpp`.

### 2. A single connection cannot handle multiple queries concurrently

**Problem**

Our `ConnectContext` is not thread-safe.

There are two scenarios where a single connection can be handling multiple queries at once:

- First, the ADBC driver allows a single connection to initiate multiple queries concurrently, as long as each cursor is only executing one query at a time.
- Second, even if the client issues queries serially, this can still occur because the client and the FE do not consider a query “finished” at exactly the same time. A query executes as follows:
  1. The client sends a `getFlightInfoPreparedStatement` RPC to the FE. The FE generates a plan and deploys it to BEs, then returns the result sink BE endpoint to the client.
  2. The client sends a `getStreamStatement` RPC to the BE and pulls results directly from the BE.
  3. After finishing result fetching, the client issues the next query. However, at this point the FE may still not have received all BE profile reports. As a result, two queries may be using the same `ConnectContext` simultaneously.

**Fix**

Ideally, we would split `ConnectContext` into `ConnectContext` and `StatementContext`. Only a small subset of `ConnectContext` fields would need to be thread-safe, while `StatementContext` would not.

However, this change would be too large; `ConnectContext` is used in hundreds of places. Therefore, the current approach is: each `ConnectContext` can execute only one SQL at a time, and subsequent SQLs wait up to `arrow_flight_sql_connection_query_wait_timeout_ms` before timing out.

Implementation-wise, we maintain a `RunningToken` (a semaphore) that can be held by only one owner at a time. Others block and wait for up to `arrow_flight_sql_connection_query_wait_timeout_ms`. If it times out, an error is returned.

- Before a SQL starts, acquire the `RunningToken`.
- After the SQL fully completes, release the `RunningToken`. “Fully completes” means `StmtExecutor::execute` finishes, including the extra 2 seconds waiting for profile reports and writing audit logs and query detail.

See `ArrowFlightSqlConnectProcessor::execute`.

### 3. No statistics collected for BE execution

**Problem**

BE fragments report statistics when `ResultSink` finishes (on `ResultSink::close`), piggybacked on the last batch sent to the FE. However, Arrow Flight SQL doesn’t have a `ResultSink` between FE and BE.

**Fix**

We still let the FE issue a `fetch_data` RPC to the BE’s `ResultSink`. When `ResultSink::close` is called, the BE responds to this `fetch_data` RPC and sends the statistics to the FE.

This also implicitly lets the FE know that the query has finished.

### 4. One `ConnectionContext` maintaining multiple prepared queries

The client’s RPC sequence for a single query is:

`[createPreparedStatement] -> [getFlightInfoPreparedStatement(preparedStmtId) -> getStreamStatement]`

1. `createPreparedStatement`: Send the query to the FE, and the FE returns a `preparedStmtId`.
2. `getFlightInfoPreparedStatement(preparedStmtId)`: Send a request to the FE to execute the query corresponding to the provided `preparedStmtId`. The FE returns the endpoint from which results can be fetched (BE for queries, FE for others).
3. `getStreamStatement`: Fetch query results from the FE or BE.

**Problem**

`ArrowFlightSQLConnectionContext` currently stores only a single `query` field. In reality, a single `ArrowFlightSQLConnectionContext` may hold multiple different prepared statements at the same time.

More specifically, each client cursor maintains one prepared statement. For a given cursor:

- If `cursor.execute(sql)` uses the same SQL as before, it reuses the existing prepared statement and issues `getFlightInfoPreparedStatement(preparedStmtId) -> getStreamStatement`.
- Otherwise, it first closes the previous prepared statement and then creates a new one, issuing `closePreparedStatement(prevPreparedStmtId) -> getFlightInfoPreparedStatement -> getStreamStatement`.

**Fix**

`ArrowFlightSQLConnectionContext` should store a collection of `preparedStatements` instead of a single `query`.

### 5. Forwarding to the leader for execution

**Flow changes**

- For non-queries, the flow is the same as the MySQL protocol: wait for the response from the RPC forwarded to the leader, and then set the result.
- For queries, since the client fetches results directly from BEs, we cannot wait for the leader FE to return results to the follower FE. Instead, once the leader FE finishes deploying fragment instances to BEs, it sends a `notifyForwardDeploymentFinished` RPC (including BE information) to the follower FE. The follower FE directly returns this information to the client.

**Kill-related fixes**

- For MySQL, when canceling a query, it automatically sends a `KILL connection_id` command. The follower FE forwards this `KILL connection_id` to the leader FE.
- For Arrow Flight SQL, there are two ways to cancel, both requiring the FE to actively send a `KILL connection_id` command to the leader FE:
  - Closing the connection: `closeSession` RPC.
  - Canceling the connection/query: `cancelFlightInfo` RPC (not yet implemented).

### 6. Passing the cancel reason to the BE

**Problem**

When one fragment fails, it passes the error message to the FE. The FE then cancels the other fragments, but only sends an `INTERNAL_ERROR` without the error message.

However, since the client pulls data directly from BEs, it cannot obtain the error message from the FE. It only sees `INTERNAL_ERROR` from the BE.

**Fix**

When the FE cancels other fragments, it also passes along the error message.

### 7. Other issues

**FE**

1. Exceptions thrown when closing a connection prevented the connection context from being released.
2. `showExternalBasicStatsMeta` has one more column than `StatsMeta` for internal tables. MySQL results ignore this extra column, so there is no issue.
3. Parser retry does not support large `IN` value lists.

**BE**

1. The map from `query_id` to schema has unsynchronized read/write access (no locking).
2. The `get_arrow_batch` method had incorrect checks on the terminal state `_status`, causing occasional deadlocks.

## SQL Test
- Adding `--arrow` to `python run.py` will run the cases using the Arrow Flight SQL protocol.
- Adding `@no_arrow_flight_sql` to a case will cause that case to be skipped when running with `--arrow`.
- Adding `@arrow_flight_sql` to a case will make that case run using the Arrow Flight SQL protocol even when `--arrow` is not specified.

## TODO

1. Parameterized prepared SQL is not supported.
2. A single connection cannot initiate multiple SQLs concurrently.
3. SQL retries are not supported, especially when BEs restart/fail.
   - Since the client pulls data directly from BEs, the FE cannot re-plan and retry. Once we support fetching data via an FE proxy, we can consider adding this feature.
4. `enable_short_circuit` is not supported.
5. `enable_constant_execute_in_fe` is not supported.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Stabilizes Arrow Flight SQL by adding correct Arrow type mappings, FE/BE execution and leader-forwarding flows with error propagation, connection concurrency control, and comprehensive test/client updates.
> 
> - **Arrow Flight SQL execution (FE/BE)**:
>   - Add `ArrowFlightSqlConnectContext/Processor`, `LeaderOpExecutor` enhancements, `ArrowFlightSqlResultDescriptor`, and `ArrowFlightSqlProxyQueryManager` for endpoint return and forward-to-leader deployment notification (`notifyForwardDeploymentFinished`).
>   - Enforce single in-flight query per connection via `RunningToken` with configurable timeout `arrow_flight_sql_connection_query_wait_timeout_ms`.
>   - Support FE-returned results for non-queries; BE endpoint for queries; handle explain and SHOW.
> - **Type mapping and Arrow conversions**:
>   - Implement Arrow Flight SQL–specific schema/field conversions: `DATE→date32`, `DATETIME→timestamp(micro,UTC)`, `VARBINARY→binary`, `LARGEINT→Decimal128(38,0)`, `DECIMAL256` support; map `HLL/OBJECT/PERCENTILE→binary` (always null).
>   - Extend row/column converters (date/timestamp units, decimal128/256, varbinary), and add guards for unsupported MAP null keys.
> - **Runtime/IO fixes**:
>   - Propagate cancel error messages to BEs (`PCancelPlanFragmentRequest.error_message`) and through FE (`DefaultCoordinator`, `ExecutionDAG`, `BackendServiceClient`).
>   - Fix `BufferControlBlock::get_arrow_batch` status handling to avoid deadlocks; `ResultBufferMgr` returns underlying status.
>   - Track written rows in `ArrowResultWriter`; carry `arrow_flight_sql_version` via `RuntimeState` and fragment params.
> - **Core utilities**:
>   - Add `DateValue::to_days_since_unix_epoch` and `TimestampValue::to_unix_microsecond`.
>   - Add `DECIMAL256` to arrow type traits.
> - **FE service/proxy**:
>   - Update `FrontendServiceImpl.forward` to handle Arrow Flight SQL sessions and deployment-finished callback; improve cancel/log messages.
>   - Make `ConnectContext` Arrow-aware (`isArrowFlightSql`, `updateReturnRows` uses long).
> - **Tests and tooling**:
>   - Introduce Python Arrow Flight SQL client/helpers and connection pooling; upgrade `pyarrow>=21`.
>   - Add arrow-mode test runner flags, tagging (`@arrow_flight_sql`, `@no_arrow_flight_sql`), and numerous SQL test adjustments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 964ff4357525e2d9296e84c22ab18414f17484b0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/exec/arrow_flight_batch_reader.cpp
+++ b/be/src/exec/arrow_flight_batch_reader.cpp
@@ -25,21 +25,21 @@ ArrowFlightBatchReader::ArrowFlightBatchReader(ResultBufferMgr* result_buffer_mg
 arrow::Status ArrowFlightBatchReader::init() {
     _schema = _result_buffer_mgr->get_arrow_schema(_query_id);
     if (_schema == nullptr) {
-        return arrow::Status::ExecutionError("Failed to fetch schema for query ID", print_id(_query_id));
+        return arrow::Status::ExecutionError("Failed to fetch schema for query ID: ", print_id(_query_id));
     }
     return arrow::Status::OK();
 }
 
 arrow::Status ArrowFlightBatchReader::ReadNext(std::shared_ptr<arrow::RecordBatch>* out) {
     if (!_schema) {
-        return arrow::Status::IOError("Failed to fetch schema for query ID ", print_id(_query_id));
+        return arrow::Status::IOError("Failed to fetch schema for query ID: ", print_id(_query_id));
     }
 
     *out = nullptr;
     auto status = ExecEnv::GetInstance()->result_mgr()->fetch_arrow_data(_query_id, out);
     if (!status.ok()) {
-        return arrow::Status::IOError("Failed to fetch arrow data for query ID ", print_id(_query_id), ": ",
-                                      status.to_string());
+        return arrow::Status::IOError("Failed to fetch arrow data for query ID: ", print_id(_query_id),
+                                      ", error: ", status.to_string());
     }
 
     return arrow::Status::OK();
--- a/be/src/exec/arrow_type_traits.h
+++ b/be/src/exec/arrow_type_traits.h
@@ -49,6 +49,7 @@ M_ArrowTypeIdToTypeStruct(ArrowTypeId::FIXED_SIZE_BINARY, arrow::FixedSizeBinary
 M_ArrowTypeIdToTypeStruct(ArrowTypeId::LARGE_BINARY, arrow::LargeBinaryType);
 M_ArrowTypeIdToTypeStruct(ArrowTypeId::LARGE_STRING, arrow::LargeStringType);
 M_ArrowTypeIdToTypeStruct(ArrowTypeId::DECIMAL, arrow::Decimal128Type);
+M_ArrowTypeIdToTypeStruct(ArrowTypeId::DECIMAL256, arrow::Decimal256Type);
 M_ArrowTypeIdToTypeStruct(ArrowTypeId::DATE32, arrow::Date32Type);
 M_ArrowTypeIdToTypeStruct(ArrowTypeId::DATE64, arrow::Date64Type);
 M_ArrowTypeIdToTypeStruct(ArrowTypeId::TIMESTAMP, arrow::TimestampType);
@@ -81,6 +82,10 @@ struct ArrowTypeIdToCppTypeStruct<AT, BinaryATGuard<AT>> {
 };
 template <>
 struct ArrowTypeIdToCppTypeStruct<ArrowTypeId::DECIMAL, guard::Guard> {
+    using type = const uint8_t*;
+};
+template <>
+struct ArrowTypeIdToCppTypeStruct<ArrowTypeId::DECIMAL256, guard::Guard> {
     using type = const uint8_t*;
 };
 template <>
--- a/be/src/exec/pipeline/fragment_executor.cpp
+++ b/be/src/exec/pipeline/fragment_executor.cpp
@@ -259,6 +259,10 @@ Status FragmentExecutor::_prepare_runtime_state(ExecEnv* exec_env, const Unified
     runtime_state->set_func_version(func_version);
     runtime_state->init_mem_trackers(query_mem_tracker);
     runtime_state->set_be_number(request.backend_num());
+    const int arrow_flight_sql_version = request.common().__isset.arrow_flight_sql_version
+                                                 ? request.common().arrow_flight_sql_version
+                                                 : TArrowFlightSQLVersion::type::V0;
+    runtime_state->set_arrow_flight_sql_version(arrow_flight_sql_version);
 
     // RuntimeFilterWorker::open_query is idempotent
     const TRuntimeFilterParams* runtime_filter_params = nullptr;
--- a/be/src/runtime/arrow_result_writer.cpp
+++ b/be/src/runtime/arrow_result_writer.cpp
@@ -63,8 +63,9 @@ Status ArrowResultWriter::init(RuntimeState* state) {
     }
 
     std::unordered_map<int64_t, std::string> temp_id_to_col_name;
+
     RETURN_IF_ERROR(convert_to_arrow_schema(_row_desc, temp_id_to_col_name, &_arrow_schema, _output_expr_ctxs,
-                                            &_output_column_names));
+                                            &_output_column_names, state->arrow_flight_sql_version()));
 
     state->exec_env()->result_mgr()->set_arrow_schema(state->fragment_instance_id(), _arrow_schema);
 
@@ -96,6 +97,8 @@ StatusOr<TFetchDataResultPtrs> ArrowResultWriter::process_chunk(Chunk* chunk) {
     RETURN_IF_ERROR(convert_chunk_to_arrow_batch(chunk, _output_expr_ctxs, _arrow_schema, arrow::default_memory_pool(),
                                                  &result));
     RETURN_IF_ERROR(_sinker->add_arrow_batch(result));
+
+    _written_rows += chunk->num_rows();
 
     return TFetchDataResultPtrs{};
 }
--- a/be/src/runtime/buffer_control_block.cpp
+++ b/be/src/runtime/buffer_control_block.cpp
@@ -330,9 +330,8 @@ void BufferControlBlock::get_batch(GetResultBatchCtx* ctx) {
 Status BufferControlBlock::get_arrow_batch(std::shared_ptr<arrow::RecordBatch>* result) {
     auto notify = defer_notify();
     std::unique_lock<std::mutex> l(_lock);
-    if (!_status.ok()) {
-        return _status;
-    }
+
+    RETURN_IF_ERROR(_status);
 
     if (_is_cancelled) {
         return Status::Cancelled("Cancelled BufferControlBlock::get_arrow_batch");
@@ -341,6 +340,8 @@ Status BufferControlBlock::get_arrow_batch(std::shared_ptr<arrow::RecordBatch>* 
     while (_arrow_batch_queue.empty() && !_is_close && !_is_cancelled) {
         _data_arriaval.wait(l);
     }
+
+    RETURN_IF_ERROR(_status);
 
     if (_is_cancelled) {
         return Status::Cancelled("Cancelled BufferControlBlock::get_arrow_batch");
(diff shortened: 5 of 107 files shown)


alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

satanson (approved):

dirtysalt (approved):

Seaven (approved):

gengjun-git (approved):

kangkaisen (approved):

PR #65911, opened 2025-11-25
[BugFix] disable rewrite min/max to meta scan in shared-data mode if schema change happened

silverbullet233:
## What I'm doing:

Fixes https://github.com/StarRocks/StarRocksTest/issues/10622

Root Cause

when changing column type happens, data won't be rewrote under fast schema evolution in shared-data mode, in this case, meta_reader need Segment::_column_readers to read zone_map, column_reader's type will be wrong and cause type mismatch error.
but we can't distinguish which type of schema change happened from olap table meta, so I choose to disable rewriting min/max under shared-data mode if schema change happened

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteSimpleAggToMetaScanRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteSimpleAggToMetaScanRule.java
@@ -22,10 +22,12 @@ import com.starrocks.catalog.Column;
 import com.starrocks.catalog.Function;
 import com.starrocks.catalog.FunctionSet;
 import com.starrocks.catalog.KeysType;
+import com.starrocks.catalog.MaterializedIndexMeta;
 import com.starrocks.catalog.OlapTable;
 import com.starrocks.catalog.Partition;
 import com.starrocks.catalog.Table;
 import com.starrocks.server.GlobalStateMgr;
+import com.starrocks.server.RunMode;
 import com.starrocks.sql.ast.expression.ExprUtils;
 import com.starrocks.sql.optimizer.OptExpression;
 import com.starrocks.sql.optimizer.OptimizerContext;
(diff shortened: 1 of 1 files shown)


Seaven (approved):

stephen-shelby (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteSimpleAggToMetaScanRule.java
```diff
@@ -221,6 +223,8 @@ public boolean check(final OptExpression input, OptimizerContext context) {
         if (aggregationOperator.getPredicate() != null) {
             return false;
         }
+        MaterializedIndexMeta currentIndexMeta = table.getIndexMetaByIndexId(table.getBaseIndexId());
+        boolean hasSchemaChange = currentIndexMeta.getSchemaVersion() > 0;
```
murphyatwork, line 227: why schemaVersion > 0  indicates schema change? what if a schema change finished ?

silverbullet233: > why schemaVersion > 0 indicates schema change?

this is by design. the initial value of schemaVersion is 0, each schema change increases it by 1.

murphyatwork: so it means, once a schema change happens, we can never use META_SCAN anymore?
i feel like it would narrow the scope of this optimization a lot, because schema change is pretty common

silverbullet233:
@mergify backport branch-4.0.2

PR #65917, opened 2025-11-25
[BugFix] fix iceberg table cache bug

Wenjun7J:
## Why I'm doing:
image

In our Iceberg catalog, we observed that Guava’s LocalCache can enter a weird state. Specifically:
	1.	cache.size() becomes 0, indicating that the cache believes it contains no active entries.
	2.	However, some Segment in the cache still retains stale StrongAccessWriteEntry objects, which remain in memory for over 10 hours.
	3.	A call for get(key) will return the value associated with this stale entry via LoadingValueReference.futureValue.
while getIfPresent(key) returns null, because the entry is already considered inactive by the cache’s bookkeeping.

This condition leads to two major issues:
	1.	Refresh becomes ineffective, since refresh will call getIfPresent to judge if continue.
	2.	Reads always return stale results, and async reload judge the value is loading, become ineffective too.

Some unusual behaviors of the CachingIcebergCatalog may have triggered this cache bug:
	1.	Iceberg caches multiple table entries for different snapshots of the same table,
while IcebergTableName.equals() does not strictly satisfy the commutative property,
which may cause the Guava bugs.
	2.	Guava’s cache appears to suffer from several long-standing concurrency issues that still persist today, as referenced in
https://github.com/google/guava/issues/1881 and https://github.com/google/guava/issues/2971.

## What I'm doing:

1.	The table cache should store only the newest object per table. The snapshotId does not necessarily need to be part of the cache key.
2.	Migrate the Guava cache implementation to Caffeine.

Fixes [#65827](https://github.com/StarRocks/starrocks/issues/65827)

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x]  3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Replaces Guava caches with Caffeine, caches only the latest table per name (no snapshot in key), refines refresh/invalidation and counting, and adds tests including REST catalog auth-token bypass.
> 
> - **Iceberg cache migration (Caffeine)**
>   - Replace Guava `Cache/LoadingCache` with Caffeine equivalents across `CachingIcebergCatalog`.
>   - Add executor-backed loaders and removal listeners; use `estimatedSize()` where applicable.
>   - Implement weighted caches for data/delete files with Caffeine `weigher` and logging.
> - **Table cache semantics**
>   - Cache only the latest `Table` per `db.table` (exclude snapshotId from key for table lookups).
>   - Update refresh flow to put updated table under snapshot-agnostic key; adjust partition cache refresh (`get` vs `getUnchecked`).
>   - Simplify invalidation to a single key; remove snapshot-scoped invalidation logic.
> - **Behavioral guards**
>   - Bypass table cache when using `IcebergRESTCatalog` with auth token.
> - **Metrics/estimation**
>   - Use `estimatedSize()` for `Database`/`Table` counts; sample methods updated accordingly.
> - **Tests**
>   - Update existing tests for Caffeine behavior differences and add tests for count estimation and REST catalog cache-bypass.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2adb5fac98ec7038265c0c8a7b9d2eb334d690e5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
@@ -17,11 +17,6 @@ package com.starrocks.connector.iceberg;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.github.benmanes.caffeine.cache.Weigher;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
-import com.google.common.cache.RemovalNotification;
 import com.google.common.collect.Lists;
 import com.starrocks.catalog.Database;
 import com.starrocks.catalog.IcebergTable;
@@ -65,7 +60,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.stream.Collectors;
 
-import static com.google.common.cache.CacheLoader.asyncReloading;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class CachingIcebergCatalog implements IcebergCatalog {
@@ -76,8 +70,8 @@ public class CachingIcebergCatalog implements IcebergCatalog {
     private static final int MEMORY_FILE_SAMPLES = 100;
     private final String catalogName;
     private final IcebergCatalog delegate;
-    private final LoadingCache<IcebergTableCacheKey, Table> tables;
-    private final Cache<String, Database> databases;
+    private final com.github.benmanes.caffeine.cache.LoadingCache<IcebergTableCacheKey, Table> tables;
+    private final com.github.benmanes.caffeine.cache.Cache<String, Database> databases;
     private final ExecutorService backgroundExecutor;
 
     private final IcebergCatalogProperties icebergProperties;
@@ -87,7 +81,7 @@ public class CachingIcebergCatalog implements IcebergCatalog {
     private final Map<IcebergTableName, Long> tableLatestAccessTime = new ConcurrentHashMap<>();
     private final Map<IcebergTableName, Long> tableLatestRefreshTime = new ConcurrentHashMap<>();
 
-    private final LoadingCache<IcebergTableName, Map<String, Partition>> partitionCache;
+    private final com.github.benmanes.caffeine.cache.LoadingCache<IcebergTableName, Map<String, Partition>> partitionCache;
 
     public CachingIcebergCatalog(String catalogName, IcebergCatalog delegate, IcebergCatalogProperties icebergProperties,
                                  ExecutorService executorService) {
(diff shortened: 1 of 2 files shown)


alvin-phoenix-ai:
@cursor review

dirtysalt (dismissed):

stephen-shelby:
could you consistently reproduce this issue？

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
```diff
@@ -368,19 +371,19 @@ private void refreshTable(BaseTable currentTable, BaseTable updatedTable,
...
         // so when refreshing partition cache, `getTables` can return the latest one.
         // another way to fix is to call `delegate.getTables` when refreshing partition cache.
         synchronized (this) {
-            tables.put(updateKey, updatedTable);
+            tables.put(keyWithoutSnap, updatedTable);
         }
 
         partitionCache.invalidate(baseIcebergTableName);
-        partitionCache.getUnchecked(updatedIcebergTableName);
+        partitionCache.get(updatedIcebergTableName);
```
stephen-shelby, line 386: you could consider this scenario. When a query fetch a table, which means obtaining the corresponding snapshot id, if a background refresh is completed before getting the partition, and this refresh has different base and updated metadata location,  can you guarantee that the values retrieved from partitionCache in this query are correct?

stephen-shelby: the timeline is 
get table of query thread -> background refresh -> get partitions of query thread

alvin-phoenix-ai:
@cursor review

Wenjun7J:
> could you consistently reproduce this issue？

It’s not very easy to reproduce. It only happens occasionally under some concurrent conditions. After running for a sufficient amount of time, the issue will surface.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
Wenjun7J: The partitionCache key is still composed of (dbName, tableName, snapshotId), and queries use this snapshotId to retrieve the entry. Nothing has changed here. Only tables cache key's snapshotId should always be -1 now.

zhaohehuhu:
```
    public List<String> getPartitionNames() {
        if (table.isUnPartitioned()) {
            return Lists.newArrayList(table.getName());
        }

        IcebergTable icebergTable = (IcebergTable) table;
        Optional<Long> snapshotId = Optional.ofNullable(icebergTable.getNativeTable().currentSnapshot())
                .map(Snapshot::snapshotId);
        ConnectorMetadatRequestContext requestContext = new ConnectorMetadatRequestContext();
        requestContext.setQueryMVRewrite(isQueryMVRewrite());
        requestContext.setTableVersionRange(TableVersionRange.withEnd(snapshotId));
        return GlobalStateMgr.getCurrentState().getMetadataMgr().listPartitionNames(
                table.getCatalogName(), getCatalogDBName(), getTableName(), requestContext);
    }
  ```

zhaohehuhu: I believe it should be ok for mv refresh as per the code

mixermt:
> > could you consistently reproduce this issue？
> 
> It’s not very easy to reproduce. It only happens occasionally under some concurrent conditions. After running for a sufficient amount of time, the issue will surface.

If you need a real-world testing I think we can help with it. 
In our env we encountered this issue pretty fast.

Youngwb (approved):

Wenjun7J:
> > > could you consistently reproduce this issue？
> > 
> > 
> > It’s not very easy to reproduce. It only happens occasionally under some concurrent conditions. After running for a sufficient amount of time, the issue will surface.
> 
> If you need a real-world testing I think we can help with it. In our env we encountered this issue pretty fast.

Oh, thanks a lot, that would be very helpful! Do you have an environment where you can build the project?
If you have Docker available, you can quickly set up a build environment using this guide: https://docs.starrocks.io/docs/developers/build-starrocks/Build_in_docker/
If that doesn’t work for you, we can think of other ways to proceed.
Looking forward to your reply 🙂

stephen-shelby (approved):

dirtysalt (approved):

Wenjun7J:
> > > > could you consistently reproduce this issue？
> > > 
> > > 
> > > It’s not very easy to reproduce. It only happens occasionally under some concurrent conditions. After running for a sufficient amount of time, the issue will surface.
> > 
> > 
> > If you need a real-world testing I think we can help with it. In our env we encountered this issue pretty fast.
> 
> Oh, thanks a lot, that would be very helpful! Do you have an environment where you can build the project? If you have Docker available, you can quickly set up a build environment using this guide: https://docs.starrocks.io/docs/developers/build-starrocks/Build_in_docker/ If that doesn’t work for you, we can think of other ways to proceed. Looking forward to your reply 🙂

@mixermt How did it go? Let me know if you need any help.

PR #65951, opened 2025-11-25
[BugFix] Fix connector sink hang when writer's initial write fails

### LINKED ISSUE #65924: [BUG] file table doesn't abort in case of an invalid time zone set

* launch allin1-ubuntu:4.0.1 env.
* create a table with date_time column
* insert into the table t with time zone info
* `set global time_zone = '+0800'` -- set an invalid time zone
* data unloading by `insert into files("path"="/opt/", "format"="parquet") select * from t`

The `insert into files` should fail if the time zone is incorrect and the data unloading can't be processed

### Real behavior (Required)

The `insert into files` statement stuck for long time

`starrocks-4.0.1`

kevincai:
## Why I'm doing:

Fix a race condition in ConnectorChunkSink where writer registration
happened after initialization and writing. If writer->init() or
writer->write() failed, the writer would never be added to the partition
map, causing subsequent operations (error handlers, commit callbacks) to
hang waiting for a writer that was never registered.

Added regression test using invalid timezone format which triggers
writer initialization failure to reproduce the hang condition.

## What I'm doing:

Fixes #65924 

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/connector/connector_chunk_sink.cpp
+++ b/be/src/connector/connector_chunk_sink.cpp
@@ -58,9 +58,14 @@ Status ConnectorChunkSink::write_partition_chunk(const std::string& partition,
         writer->set_commit_callback(commit_callback);
         writer->set_error_handler(error_handler);
         writer->set_io_poller(_io_poller);
-        RETURN_IF_ERROR(writer->init());
-        RETURN_IF_ERROR(writer->write(chunk));
+        auto st = writer->init();
+        if (!st.ok()) {
+            set_status(st);
+            return st;
+        }
+        // save the writer to the map, so errors from subsequent write() can be correctly handled.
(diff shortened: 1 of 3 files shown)


trueeyu (approved):

GavinMar (approved):

wyb (approved):

kevincai:
Ignore backport to branch 3.5/3.4/3.3, exhibit different behavior in these branches.

PR #65976, opened 2025-11-26
[Tool] Python SQLAlchemy 1.4 Compatibility

rad-pat:
## Why I'm doing:
SQLAlchemy client was built alongside SQLAlchemy 2.0, but it needs to also be compatible with version 1.4.*
This is for people/projects who have not yet been able to make the upgrade to SQLAlchemy 2.0, such as Apache Superset which quite a number of people use for visualization. Having it compatible there is a good way to improve visibility.

## What I'm doing:
Tweaks to the dialect to allow it to run in SQLAlchemy 1.4. This involves removing some of the typing on methods which seems to be not fully supported by the `reflection.cache` decorator in 1.4. Also needed to make the starrocks_properties dictionary a hashable item so that it worked in 1.4

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [x] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


rad-pat:
@jaogoy, please take a look - these changes should make compatible with SQLAlchemy 1.4. 
I ran dialect tests against 1.4 and only experienced failures due to tests that are excluded in 2.0 due to higher granularity of test requirements.
I failed to check your tests against SQLAlchemy 1.4, I could not even get running against 2.0 - errors such as below
```
Table 'user': Attribute 'ENGINE' in database has non-default value 'CLOUD_NATIVE' (default: 'OLAP'), but not specified in metadata. Please specify this attribute explicitly in your table definition to avoid unexpected behavior. No ALTER statement will be generated automatically.
```

contrib/starrocks-python-client/starrocks/dialect.py
```diff
@@ -1425,8 +1422,8 @@ def _read_from_information_schema(
 
     @reflection.cache
     def _setup_parser(
-        self, connection: Connection, table_name: str, schema: Optional[str] = None, **kwargs: Any
-    ) -> ReflectedState:
+        self, connection, table_name, schema = None, **kwargs
```
jaogoy, line 1428: What's the resone of removing the type annotations?

contrib/starrocks-python-client/starrocks/dialect.py
```diff
@@ -1664,26 +1661,23 @@ def get_show_alter_table(connection: Connection, table_name: str, alter_type: st
...
         Overrides the mysql's implementation, which will use 'mysql_comment' as the key,
         here the `comment` supports lower case only.
         """
         parsed_state = self._parsed_state_or_create(connection, table_name, schema, **kw)
         comment = parsed_state.table_options.get(TableInfoKeyWithPrefix.COMMENT, None)
-        if comment is not None:
-            return {"text": comment}
-        else:
-            return ReflectionTableDefaults.table_comment()
+        return {"text": comment}
```
jaogoy: Why change it?

contrib/starrocks-python-client/starrocks/reflection.py
```diff
@@ -487,7 +495,7 @@ def _parse_general_table_options(self, table_name: str, schema: Optional[str] =
         if properties := table_config.get(TableConfigKey.PROPERTIES):
             # logger.debug("table_config.%s: %s", TableConfigKey.PROPERTIES, properties)
             try:
-                opts[TableInfoKeyWithPrefix.PROPERTIES] = dict(json.loads(properties or "{}").items())
+                opts[TableInfoKeyWithPrefix.PROPERTIES] = hashdict(json.loads(properties or "{}").items())
```
jaogoy, line 552: What's the reason?

jaogoy (commented):
Maybe it's better to leave the code as 2.x style with a try/except for 1.4 compatibility for most of the imports?

contrib/starrocks-python-client/starrocks/dialect.py
rad-pat: It's in the PR description. The reflection.cache() decorator in 1.4 cannot handle the typing. I asked about getting it fixed so we did not have to change, but there will be no more 1.4 releases.

contrib/starrocks-python-client/starrocks/dialect.py
rad-pat: No ReflectionTableDefaults.table_comment() in 1.4 - this code simply returns {"text": None} which is what would be achieved anyway

contrib/starrocks-python-client/starrocks/reflection.py
rad-pat: This one is kind of interesting because the reflection.cache decorator did not handle caching in 1.4 when properties (stored as dict) were part of the cache key (because dict is not hashable). 2.0 handles this, hence no error, but I don't think the cache key can be assumed to be the same each time. The hashdict ensure that the cache key for those properties is the same, so probably a valid change for 2.0 anyway.

contrib/starrocks-python-client/starrocks/dialect.py
rad-pat: Actually, this can be added easily to the fake ReflectionDefaults class defined in defaults.py, so make this change not necessary

rad-pat:
> Maybe it's better to leave the code as 2.x style with a try/except for 1.4 compatibility for most of the imports?

@jaogoy - I think changing the imports to work with 1.4 and 2.0 is inconsequential. Try/except should be avoided if possible.

Once a PyPIrelease is made to support SQLAlchemy 1.4, support can be dropped afterwards - just flagging that the PyPI version will be last to support SQLAlchemy 1.4. This then gives people using Superset an option to use a more up to date dialect version, until Superset finalizes the support for SQLAlchemy 2.0. I have asked Superset for an expectation of when that may be completed, but no answer yet.

jaogoy:
> > Maybe it's better to leave the code as 2.x style with a try/except for 1.4 compatibility for most of the imports?
> 
> @jaogoy - I think changing the imports to work with 1.4 and 2.0 is inconsequential. Try/except should be avoided if possible.
> 
> Once a PyPIrelease is made to support SQLAlchemy 1.4, support can be dropped afterwards - just flagging that the PyPI version will be last to support SQLAlchemy 1.4. This then gives people using Superset an option to use a more up to date dialect version, until Superset finalizes the support for SQLAlchemy 2.0. I have asked Superset for an expectation of when that may be completed, but no answer yet.

@rad-pat IMO, compatibility code will be dropped in the future. And in our code, it's better to use newer version SQLAlchemy sytle import.
Of course it's not so important and urgent, meybe we will support old versions for a long time.

jaogoy (dismissed):

inviscid:
@jaogoy Is this at a point that it can be approved for merging?

jaogoy (approved):

alvin-phoenix-ai (approved):

Issue #10651, opened 2022-08-31
[Bug] IndexOutOfBoundException when concurrent AlterTable with Insert


decster:
```
java.lang.IndexOutOfBoundsException: Index: 3, Size: 3
    at java.util.ArrayList.rangeCheck(ArrayList.java:659) ~[?:1.8.0_322]
    at java.util.ArrayList.get(ArrayList.java:435) ~[?:1.8.0_322]
    at com.starrocks.sql.InsertPlanner.castLiteralToTargetColumnsType(InsertPlanner.java:167) ~[classes/:?]
    at com.starrocks.sql.InsertPlanner.plan(InsertPlanner.java:82) ~[classes/:?]
    at com.starrocks.sql.StatementPlanner.plan(StatementPlanner.java:77) ~[classes/:?]
    at com.starrocks.sql.StatementPlanner.plan(StatementPlanner.java:40) ~[classes/:?]
    at com.starrocks.qe.StmtExecutor.execute(StmtExecutor.java:342) ~[classes/:?]
    at com.starrocks.qe.ConnectProcessor.handleQuery(ConnectProcessor.java:312) ~[classes/:?]
    at com.starrocks.qe.ConnectProcessor.dispatch(ConnectProcessor.java:427) ~[classes/:?]
    at com.starrocks.qe.ConnectProcessor.processOnce(ConnectProcessor.java:665) ~[classes/:?]
    at com.starrocks.mysql.nio.ReadListener.lambda$handleEvent$0(ReadListener.java:55) ~[classes/:?]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_322]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_322]
    at java.lang.Thread.run(Thread.java:750) [?:1.8.0_322]

```

### StarRocks version (Required)
 - You can get the StarRocks version by executing SQL `select current_version()`

PR #66097, opened 2025-11-28
[BugFix] fix connector chunk sink bug

Wenjun7J:
## Why I'm doing:
The original question is described in https://github.com/StarRocks/starrocks/pull/65951

But the pr introduces a new problem:
1. the _file_writer in PartitionChunkWriter init failed
2. PartitionChunkWriter is still put into the map
3. PartitionChunkWriter close will also call _file_writer close, but it is nullptr.

## What I'm doing:

Fixes [#10651](https://github.com/StarRocks/StarRocksTest/issues/10651)

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/connector/partition_chunk_writer.cpp
+++ b/be/src/connector/partition_chunk_writer.cpp
@@ -69,6 +69,7 @@ void PartitionChunkWriter::commit_file() {
 }
 
 Status BufferPartitionChunkWriter::init() {
+    RETURN_IF_ERROR(create_file_writer_if_needed());
     return Status::OK();
 }
 
@@ -125,6 +126,7 @@ Status SpillPartitionChunkWriter::init() {
     RETURN_IF_ERROR(_load_spill_block_mgr->init());
     _load_chunk_spiller = std::make_unique<LoadChunkSpiller>(_load_spill_block_mgr.get(),
                                                              _fragment_context->runtime_state()->runtime_profile());
+    RETURN_IF_ERROR(create_file_writer_if_needed());
     return Status::OK();
 }
 
--- a/be/src/formats/orc/orc_file_writer.cpp
+++ b/be/src/formats/orc/orc_file_writer.cpp
@@ -115,7 +115,9 @@ FileWriter::CommitResult ORCFileWriter::commit() {
     FileWriter::CommitResult result{
             .io_status = Status::OK(), .format = ORC, .location = _location, .rollback_action = _rollback_action};
     try {
-        _writer->close();
+        if (_writer != nullptr) {
+            _writer->close();
+        }
     } catch (const std::exception& e) {
         result.io_status.update(Status::IOError(fmt::format("{}: {}", "close file error", e.what())));
     }
--- a/be/src/formats/parquet/parquet_file_writer.cpp
+++ b/be/src/formats/parquet/parquet_file_writer.cpp
@@ -68,7 +68,9 @@ FileWriter::CommitResult ParquetFileWriter::commit() {
     FileWriter::CommitResult result{
             .io_status = Status::OK(), .format = PARQUET, .location = _location, .rollback_action = _rollback_action};
     try {
-        _writer->Close();
+        if (_writer != nullptr) {
+            _writer->Close();
+        }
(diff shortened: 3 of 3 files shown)


trueeyu (dismissed):

be/src/connector/connector_chunk_sink.cpp
```diff
@@ -63,9 +63,21 @@ Status ConnectorChunkSink::write_partition_chunk(const std::string& partition,
             set_status(st);
             return st;
         }
-        // save the writer to the map, so errors from subsequent write() can be correctly handled.
+
+        st = writer->write(chunk);
```
kevincai: thought the error should be handled inside the writer properly.

Reason:
1. once the `writer->init()` is done correctly without error, the `writer` instance should be well behalfed  including error handling. Shouldn't take the caller this complexity to handle the error of the first write() call.
2. idempotent implementation, the first write() is significantly different than the non-first write() call, why? Compared to the line 53, why it doesn't need this kind of error handling? The first and non-first error handling differentiation should be encapsulated into the writer instance.
3. Coupled with interface, not implementation details. This error handling code directly implements the out_stream() of the writer bypass the writer itself, apparently this is an implementation details leakage to the caller rather than interface contract.

be/src/connector/connector_chunk_sink.cpp
```diff
@@ -63,9 +63,21 @@ Status ConnectorChunkSink::write_partition_chunk(const std::string& partition,
...
+        if (!st.ok()) {
```
GavinMar: I think we can consider moving the file writer init for the first write to writer->init.

be/src/connector/partition_chunk_writer.cpp
```diff
@@ -307,7 +306,6 @@ Status SpillPartitionChunkWriter::_write_chunk(Chunk* chunk) {
     if (!_sort_ordering && _file_writer->get_written_bytes() >= _max_file_size) {
         commit_file();
     }
-    RETURN_IF_ERROR(create_file_writer_if_needed());
```
GavinMar, line 310: If the above `commit_file` executed, the file_writer will be reset to null, which will cause the next write crash.

GavinMar (approved):

be/src/connector/partition_chunk_writer.cpp
```diff
@@ -69,6 +69,7 @@ void PartitionChunkWriter::commit_file() {
 }
 
 Status BufferPartitionChunkWriter::init() {
+    RETURN_IF_ERROR(create_file_writer_if_needed());
```
kevincai, line 72: looks like this call is not necessary, because the `create_file_writer_if_needed` could be called anyway if a new file is needed, the file_writer will be switched to a new instance and the new writer's init() will be invoked anyway.

be/src/connector/partition_chunk_writer.cpp
```diff
@@ -125,6 +126,7 @@ Status SpillPartitionChunkWriter::init() {
     RETURN_IF_ERROR(_load_spill_block_mgr->init());
     _load_chunk_spiller = std::make_unique<LoadChunkSpiller>(_load_spill_block_mgr.get(),
                                                              _fragment_context->runtime_state()->runtime_profile());
+    RETURN_IF_ERROR(create_file_writer_if_needed());
```
kevincai, line 129: The error handling could be consistent with the `SpillPartitionChunkWriter`, where the error handling will be called if the underlying IO op fails.

Otherwise it could be that the caller properly set the writer's commit_callback, error_handling, but still some of the errors can't be handled cleanly in this framework. The error handling logic could be encapsulated into limited locations with well-defined framework.

kevincai (approved):

trueeyu (approved):

PR #66117, opened 2025-11-29
[Tool] Fix build failure on Ubuntu 24.04 by conditionally linking libsframe

arvindKandpal-ksolves:
## Why I'm doing:

On newer Linux distributions (specifically Ubuntu 24.04 LTS with GCC 12.4.0 and Binutils 2.42), the backend (BE) build fails during the final linking stage of `starrocks_be`.

The error reported is:
`undefined reference to sframe_encoder_write`

This occurs because `libbfd` now has a dependency on `libsframe` in newer binutils versions, but `libsframe` was not included in the linker flags in `be/CMakeLists.txt`.

## What I'm doing:

I have modified `be/CMakeLists.txt` to introduce a conditional check for the `libsframe` library.

- Used `find_library(SFRAME_LIB sframe)` to detect if the library exists on the host system.
- If found (e.g., on Ubuntu 24.04), `-lsframe` is explicitly appended to `STARROCKS_STATIC_LINK_LIBS` after `-lbfd`.
- If not found (e.g., on older CentOS 7 systems), the build proceeds with the original linker flags to ensure backward compatibility.

Fixes #66116

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [x] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Conditionally link `libsframe` in `be/CMakeLists.txt` when detected to satisfy newer binutils dependencies.
> 
> - **Build/CMake**:
>   - Detect `libsframe` via `find_library(SFRAME_LIB sframe)` in `be/CMakeLists.txt`.
>   - If found, append `-lsframe` to `STARROCKS_STATIC_LINK_LIBS` (after `-lbfd`); otherwise retain existing linker flags.
>   - Add status messages indicating whether `libsframe` was found and linked.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d66a92533337cbfc0ad4350e8774cfa07ef8c636. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/CMakeLists.txt
+++ b/be/CMakeLists.txt
@@ -612,7 +612,16 @@ FUNCTION(ADD_BE_LIB LIB_NAME)
     set(ALL_BE_LIBS "${ALL_BE_LIBS};${LIB_NAME}" CACHE INTERNAL "All BE libraries")
 ENDFUNCTION()
 
-set (STARROCKS_STATIC_LINK_LIBS ${WL_LINK_STATIC} -lgomp -lbfd)
+# Check if libsframe exists (needed for binutils 2.40+ on newer distros like Ubuntu 24.04)
+find_library(SFRAME_LIB sframe)
+
+if(SFRAME_LIB)
+    message(STATUS "Found libsframe: ${SFRAME_LIB}, linking it explicitly.")
+    set (STARROCKS_STATIC_LINK_LIBS ${WL_LINK_STATIC} -lgomp -lbfd -lsframe)
+else()
+    message(STATUS "libsframe not found, skipping explicit linkage.")
+    set (STARROCKS_STATIC_LINK_LIBS ${WL_LINK_STATIC} -lgomp -lbfd)
+endif()
 
 if (NOT ("${MAKE_TEST}" STREQUAL "ON" AND "${BUILD_FOR_SANITIZE}" STREQUAL "ON"))
     # In other words, turn to dynamic link when MAKE_TEST and BUILD_TYPE == *SAN


arvindKandpal-ksolves:
Hi @kevincai ,

Could you please take a look at this PR when you get a chance?

This fixes the backend build failure on newer distributions like Ubuntu 24.04 (GCC 12 / Binutils 2.42) by conditionally linking `libsframe`. I have verified that it works on Ubuntu 24.04 and should maintain backward compatibility for older systems.

Thanks!

alvin-phoenix-ai:
@cursor review

be/CMakeLists.txt
```diff
@@ -612,7 +612,16 @@ FUNCTION(ADD_BE_LIB LIB_NAME)
     set(ALL_BE_LIBS "${ALL_BE_LIBS};${LIB_NAME}" CACHE INTERNAL "All BE libraries")
 ENDFUNCTION()
 
-set (STARROCKS_STATIC_LINK_LIBS ${WL_LINK_STATIC} -lgomp -lbfd)
+# Check if libsframe exists (needed for binutils 2.40+ on newer distros like Ubuntu 24.04)
+find_library(SFRAME_LIB sframe)
```
kevincai, line 616: considering if there is a way to check the bfd version, and conditionally link the libsframe only when the version greater than x.y.

or if the `pkg-config` can do the tricky, using pkg-config to get the link dependency of libbfd, and add it together with `-lbfd`

arvindKandpal-ksolves:
@kevincai Thank you for the review and suggestions.

I investigated both approaches on the standard Ubuntu 24.04 build environment, and here are the findings:

**1. Regarding `pkg-config`:**
This approach is unfortunately not viable because the `libbinutils-dev` package on Ubuntu 24.04 **does not provide a `bfd.pc` metadata file**.
I verified this by searching the library paths (`/usr/lib`, `/usr/share/pkgconfig`), and the config file is completely missing. Without this file, `pkg-config` cannot detect the dependencies of libbfd.

**2. Regarding explicit version check vs. `find_library`:**
The current `find_library` approach serves as an effective, implicit version check:
* **On newer Binutils (e.g., v2.42 on Ubuntu 24.04):** The `libsframe` library is present on the OS. `find_library` detects it, and we link it. This fixes the build.
* **On older Binutils (e.g., CentOS 7):** The `libsframe` library does not exist. `find_library` returns nothing, and we skip the linkage. This preserves backward compatibility.

Since we cannot rely on `pkg-config` due to missing system files, detecting the existence of the library itself seems to be the most robust and lightweight solution to handle this dependency.

kevincai (approved):

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai (approved):

PR #66288, opened 2025-12-04
[Enhancement] Remove TabletinvertedIndex lock to improve metadata performance

HangyuanLiu:
## What I'm doing:

This pull request introduces several improvements and refactorings to how tablet and replica metadata are managed and accessed, as well as some robustness enhancements around tablet state handling. The main themes are API simplification for tablet/replica lookups, improved error handling when tablet metadata is missing, and the addition of a new concurrent map utility.

**Tablet and Replica Metadata API Simplification:**

* Refactored `TabletInvertedIndex` usage across the codebase to replace `Map<Long, Replica>` return types with more direct methods such as `getReplicasByTabletId`, `getTabletIdsByBackendId`, and `getReplica`, simplifying how replicas and tablets are accessed and iterated. This reduces unnecessary map wrapping and improves clarity. [[1]](diffhunk://#diff-c7bc9884ee4f299876f46e65a0c5384c91b0011e3606aac12631f2ed309f7b18L334-R333) [[2]](diffhunk://#diff-c7bc9884ee4f299876f46e65a0c5384c91b0011e3606aac12631f2ed309f7b18L345-R344) [[3]](diffhunk://#diff-8e506bc2b6e93440302f02affa31475e3192c3d35b8cf2c1f7ed3824c7947a30L578-R601) [[4]](diffhunk://#diff-8e506bc2b6e93440302f02affa31475e3192c3d35b8cf2c1f7ed3824c7947a30L732-L734) [[5]](diffhunk://#diff-539cb709034beb058dad765d09d48ae01c3f590f0140a4fee3d031d58c5063e6L408-R409)

**Robustness and Error Handling:**

* Added checks and logging for cases where tablet metadata is missing during replica creation and binlog scan range building, ensuring that such cases are handled gracefully and do not cause failures or inconsistencies. [[1]](diffhunk://#diff-cbfb0bf4c740dd4b26fa0f6093b3119661df21c7e950b68f62b5b4224c9ca95dR1905-R1910) [[2]](diffhunk://#diff-7128e65db8ef34d0256fd83ff0000714a28288e421535bcac7a162c7c3d941b8R51-R54)

**Concurrency Utility Addition:**

* Introduced a new utility class `ConcurrentLong2ObjectHashMap` that provides a segmented, thread-safe map for long keys, improving performance and concurrency for metadata storage. This is backed by `Long2ObjectOpenHashMap` to avoid auto-boxing and supports efficient reads, writes, and resizing.

**Logging Improvements:**

* Added a logger to `BinlogConsumeStateVO` and improved logging for missing tablet scenarios, aiding debugging and operational visibility. [[1]](diffhunk://#diff-7128e65db8ef34d0256fd83ff0000714a28288e421535bcac7a162c7c3d941b8R25-R26) [[2]](diffhunk://#diff-7128e65db8ef34d0256fd83ff0000714a28288e421535bcac7a162c7c3d941b8R40-R41)

**Code Cleanup:**

* Removed unused imports from files that previously referenced the old tablet/replica APIs. [[1]](diffhunk://#diff-c7bc9884ee4f299876f46e65a0c5384c91b0011e3606aac12631f2ed309f7b18L83) [[2]](diffhunk://#diff-539cb709034beb058dad765d09d48ae01c3f590f0140a4fee3d031d58c5063e6L52) [[3]](diffhunk://#diff-56b066b3b46593e4852ca1f60db30cf9578c72924f5fd00ead96d71f36646786R26)

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Replaces lock-based TabletInvertedIndex with concurrent data structures and new APIs, updates callers accordingly, adds a concurrent long-key map, and improves missing-metadata handling/logging.
> 
> - **Metadata/Concurrency**:
>   - **TabletInvertedIndex**: Replaced global RW lock and map-of-maps with concurrent structures (`ConcurrentLong2ObjectHashMap`, `CopyOnWriteArrayList`), added mutation lock only for writes. Introduced APIs: `getReplicasByTabletId`, `getReplica`, `getTabletIdsByBackendId`, path-hash grouping, and replica counts; removed replica/id cross maps and read/write lock methods.
>   - **Utility**: Added `ConcurrentLong2ObjectHashMap` (segmented, thread-safe long→object map) to back metadata indexes.
> - **Callers updated to new APIs**:
>   - `ReportHandler`: Reworked tablet diff to iterate `getTabletIdsByBackendId`, use `getReplica`, handle missing `TabletMeta`, and retain migration/txn sync logic; minor logging tweaks.
>   - `SystemHandler`: Uses `getReplicasByTabletId` in decommission-drop checks.
>   - `TabletScheduler`: Validate `TabletMeta` before logging add-replica edit.
>   - `OlapDeleteJob`: Build `TabletCommitInfo` directly from `TabletDeleteInfo` (no inverted-index reverse lookup).
>   - `BinlogConsumeStateVO`: Added logger and explicit error when tablet meta is missing.
> - **Tests**: Updated to new list-based replica APIs and path-hash grouping; adjusted FE test utilities to new replica map type.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 61b5112b9d594537a98f9137cd7eb8b1d4370e67. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
@@ -35,20 +35,15 @@
 package com.starrocks.catalog;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
-import com.starrocks.common.Config;
-import com.starrocks.common.util.concurrent.QueryableReentrantReadWriteLock;
+import com.starrocks.common.util.concurrent.ConcurrentLong2ObjectHashMap;
 import com.starrocks.lake.LakeTablet;
 import com.starrocks.memory.MemoryTrackable;
 import com.starrocks.memory.estimate.Estimator;
 import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.thrift.TStorageMedium;
-import it.unimi.dsi.fastutil.longs.Long2LongOpenHashMap;
-import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -56,8 +51,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.locks.ReentrantLock;
 
 /*
  * this class stores an inverted index
@@ -73,71 +69,36 @@ public class TabletInvertedIndex implements MemoryTrackable {
     public static final TabletMeta NOT_EXIST_TABLET_META = new TabletMeta(NOT_EXIST_VALUE, NOT_EXIST_VALUE,
             NOT_EXIST_VALUE, NOT_EXIST_VALUE, TStorageMedium.HDD);
 
-    private final QueryableReentrantReadWriteLock lock = new QueryableReentrantReadWriteLock();
-
     // tablet id -> tablet meta
-    private final Map<Long, TabletMeta> tabletMetaMap = new Long2ObjectOpenHashMap<>();
+    private final ConcurrentLong2ObjectHashMap<TabletMeta> tabletMetaMap = new ConcurrentLong2ObjectHashMap<>();
 
-    // replica id -> tablet id
-    private final Map<Long, Long> replicaToTabletMap = new Long2LongOpenHashMap();
+    // tablet id -> replicas
+    private final ConcurrentLong2ObjectHashMap<CopyOnWriteArrayList<Replica>> tabletToReplicaList =
+            new ConcurrentLong2ObjectHashMap<>();
+
+    // backend id -> tablet id list
+    private final ConcurrentLong2ObjectHashMap<Set<Long>> backendToTabletIdList =
+            new ConcurrentLong2ObjectHashMap<>();
 
     // tablet id -> backend set
-    private final Map<Long, Set<Long>> forceDeleteTablets = new Long2ObjectOpenHashMap<>();
+    private final ConcurrentLong2ObjectHashMap<Set<Long>> forceDeleteTablets = new ConcurrentLong2ObjectHashMap<>();
 
-    // tablet id -> (backend id -> replica)
-    private final Map<Long, Map<Long, Replica>> replicaMetaTable = new Long2ObjectOpenHashMap<>();
-    // backing replica table, for visiting backend replicas faster.
-    // backend id -> (tablet id -> replica)
-    private final Map<Long, Map<Long, Replica>> backingReplicaMetaTable = new Long2ObjectOpenHashMap<>();
+    // Protects multi-map mutations; readers rely on higher level synchronization.
+    private final ReentrantLock mutationLock = new ReentrantLock();
 
     public TabletInvertedIndex() {
     }
 
-    public void readLock() {
-        lock.sharedLockDetectingSlowLock(Config.slow_lock_threshold_ms, TimeUnit.MILLISECONDS);
-    }
-
-    public void readUnlock() {
-        lock.sharedUnlock();
-    }
-
-    private void writeLock() {
-        lock.exclusiveLockDetectingSlowLock(Config.slow_lock_threshold_ms, TimeUnit.MILLISECONDS);
-    }
-
-    private void writeUnlock() {
-        lock.exclusiveUnlock();
-    }
-
-    public Long getTabletIdByReplica(long replicaId) {
-        readLock();
-        try {
-            return replicaToTabletMap.get(replicaId);
-        } finally {
-            readUnlock();
-        }
-    }
-
     public TabletMeta getTabletMeta(long tabletId) {
-        readLock();
-        try {
-            return tabletMetaMap.get(tabletId);
-        } finally {
-            readUnlock();
-        }
+        return tabletMetaMap.get(tabletId);
     }
 
     public List<TabletMeta> getTabletMetaList(List<Long> tabletIdList) {
         List<TabletMeta> tabletMetaList = new ArrayList<>(tabletIdList.size());
-        readLock();
-        try {
-            for (Long tabletId : tabletIdList) {
-                tabletMetaList.add(tabletMetaMap.getOrDefault(tabletId, NOT_EXIST_TABLET_META));
-            }
-            return tabletMetaList;
-        } finally {
-            readUnlock();
+        for (Long tabletId : tabletIdList) {
+            tabletMetaList.add(tabletMetaMap.getOrDefault(tabletId, NOT_EXIST_TABLET_META));
         }
+        return tabletMetaList;
     }
 
     // always add tablet before adding replicas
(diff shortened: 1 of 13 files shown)


fe/fe-core/src/main/java/com/starrocks/scheduler/mv/BinlogConsumeStateVO.java
```diff
@@ -44,6 +48,10 @@ public List<TBinlogScanRange> toThrift() {
         binlogMap.forEach((key, value) -> {
             TBinlogScanRange scan = new TBinlogScanRange();
             TabletMeta meta = tabletIndex.getTabletMeta(key.getTabletId());
+            if (meta == null) {
```
kevincai, line 51: shall throw exception? from the original code, it is possible the tablet id doesn't exist during the foreach op, and it will be an NPE exception, now we can turn it to specific MetaNotFound or TabletNotFound exception.

fe/fe-core/src/main/java/com/starrocks/load/OlapDeleteJob.java
```diff
@@ -405,16 +404,9 @@ public boolean commitImpl(Database db, long timeoutMs) throws StarRocksException
...
-                // the inverted index contains rolling up replica
-                Long tabletId = invertedIndex.getTabletIdByReplica(replica.getId());
-                if (tabletId == null) {
-                    LOG.warn("could not find tablet id for replica {}, the tablet maybe dropped", replica);
-                    continue;
-                }
-                tabletCommitInfos.add(new TabletCommitInfo(tabletId, replica.getBackendId()));
+        for (TabletDeleteInfo tabletDeleteInfo : getTabletDeleteInfo()) {
+            for (Replica replica : tabletDeleteInfo.getFinishedReplicas()) {
+                tabletCommitInfos.add(new TabletCommitInfo(tabletDeleteInfo.getTabletId(), replica.getBackendId()));
```
kevincai, line 419: this doesn't look like identical code change, need to double check that the caller can handle the case when the tablet id doesn't exist in the TabletCommitInfo.

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -73,132 +69,77 @@ public class TabletInvertedIndex implements MemoryTrackable {
...
-            }
-        } finally {
-            writeUnlock();
+        Set<Long> backendSets = forceDeleteTablets.get(tabletId);
+        if (backendSets != null) {
+            backendSets.add(backendId);
+        } else {
+            Set<Long> s = ConcurrentHashMap.newKeySet();
+            s.add(backendId);
+            forceDeleteTablets.put(tabletId, s);
```
kevincai: shall be protected by write lock, in case concurrent call with the same tabletId which doesn't exist yet in the map. The later put will overwrite the former put, which resulting data loss.

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -73,132 +69,77 @@ public class TabletInvertedIndex implements MemoryTrackable {
...
-        writeLock();
-        try {
-            forceDeleteTablets.put(tabletId, backendIds);
-        } finally {
-            writeUnlock();
-        }
+
+        Set<Long> s = ConcurrentHashMap.newKeySet();
+        s.addAll(backendIds);
+        forceDeleteTablets.put(tabletId, s);
```
kevincai: ditto

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -210,110 +151,92 @@ public void markTabletForceDelete(Tablet tablet) {
...
-        try {
-            if (forceDeleteTablets.containsKey(tabletId)) {
-                forceDeleteTablets.get(tabletId).remove(backendId);
-                if (forceDeleteTablets.get(tabletId).isEmpty()) {
-                    forceDeleteTablets.remove(tabletId);
-                }
+        Set<Long> backendSets = forceDeleteTablets.get(tabletId);
+        if (backendSets != null) {
+            backendSets.remove(backendId);
+            if (backendSets.isEmpty()) {
```
kevincai: also need lock, in case of remove and add conflict. the isEmpty() op and the remove() op is not atomic.

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/load/OlapDeleteJob.java
HangyuanLiu: The code here will be checked again externally to see if the tablet still exists.

fe/fe-core/src/main/java/com/starrocks/scheduler/mv/BinlogConsumeStateVO.java
HangyuanLiu: In fact, this code has no meaning and is not called. Let the developers define it themselves when it is actually used.

kevincai: then an exception is more proper, otherwise the caller may not even notice that the null meta is ignore silently.

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -73,132 +69,68 @@ public class TabletInvertedIndex implements MemoryTrackable {
...
-        try {
-            if (forceDeleteTablets.containsKey(tabletId)) {
-                forceDeleteTablets.get(tabletId).add(backendId);
-            } else {
-                forceDeleteTablets.put(tabletId, Sets.newHashSet(backendId));
-            }
-        } finally {
-            writeUnlock();
-        }
+        forceDeleteTablets.computeIfAbsent(tabletId, k -> ConcurrentHashMap.newKeySet()).add(backendId);
```
kevincai, line 133: this doesn't look right. If the tabletId exists, the backendId is not added into the keySet at all

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -73,132 +69,68 @@ public class TabletInvertedIndex implements MemoryTrackable {
...
         if (backendIds.isEmpty()) {
             return;
         }
-        writeLock();
-        try {
-            forceDeleteTablets.put(tabletId, backendIds);
-        } finally {
-            writeUnlock();
-        }
+
```
kevincai, line 140: the same issue

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -210,110 +142,89 @@ public void markTabletForceDelete(Tablet tablet) {
...
-            writeUnlock();
+        if (tabletMetaMap.get(tabletId) == null) {
+            return;
+        }
+        CopyOnWriteArrayList<Replica> replicaList = tabletToReplicaList.get(tabletId);
+        if (replicaList != null) {
+            replicaList.removeIf(replica -> replica.getBackendId() == backendId);
+        }
+
+        Set<Long> tabletIds = backendToTabletIdList.get(backendId);
```
kevincai: Modification of `tabletToReplicaList` and `backendToTabletIdList` are in two statements. Considering concurrent `deleteReplica()` and `addReplica()` for the same tablet and backendId, the final result may be inconsistent because of the two ops are not atomic.

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -210,110 +142,89 @@ public void markTabletForceDelete(Tablet tablet) {
...
+        LOG.debug("remove tablet {} in backend {}", tabletId, backendId);
     }
 
     public Replica getReplica(long tabletId, long backendId) {
-        readLock();
-        try {
-            return getReplica(replicaMetaTable, tabletId, backendId);
-        } finally {
-            readUnlock();
+        CopyOnWriteArrayList<Replica> replicas = tabletToReplicaList.get(tabletId);
```
kevincai, line 233: consider returns an immutableList() instead, so that it is guaranteed that the list will not be modified in this method.

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -324,217 +235,140 @@ public List<Replica> getReplicasByTabletId(long tabletId) {
...
-                }
-                return replicas;
+        List<Replica> replicas = Lists.newArrayListWithCapacity(tabletIds.size());
+        for (Long tabletId : tabletIds) {
+            Replica replica = getReplica(tabletId, backendId);
+            if (replica != null) {
+                replicas.add(replica);
             }
+        }
+        if (replicas.isEmpty()) {
```
kevincai, line 269: the interface contract is changed, original implementation return null only when backend doesn't exist, if the backend exists but doesn't have any replica, an empty list will be returned. After refactor, null will be returned in second case. Need carefully check if the behavior should be changed. If yes, update the comment as well.

HangyuanLiu: This should return NULL, confirmed.

alvin-phoenix-ai:
@cursor review

kevincai (dismissed):

StarRocks-Reviewer:
@cursor review

kevincai (dismissed):

gengjun-git (dismissed):

kevincai (approved):

gengjun-git (approved):

meegoo (approved):

wyb (approved):

PR #66365, opened 2025-12-05
[BugFix] Retry in RPC-level for RuntimeException Unable to validate object

ZiheLiu:
## Why I'm doing:

When the FE sends bRPC requests `sendPlanFragmentAsync` and `cancelPlanFragmentAsync` to the BE, an RPC-level retry is triggered if a `NoSuchElementException` is thrown. However, for a stale connection (for example, when a BE crashes and quickly comes back up, and the newly added BE has the same IP as before), a `RuntimeException("Unable to validate object")` is thrown instead of `NoSuchElementException`, which we currently do not handle.

The reason for the `RuntimeException("Unable to validate object")` is that jprotobuf catches the `NoSuchElementException("Unable to validate object")` thrown by `GenericObjectPool.borrowObject` and then wraps it inside a `RuntimeException`.
img_v3_02sm_b74c88c8-330e-4606-a9b7-5b45843de37g
img_v3_02sm_e3081946-c9e9-4bbf-8795-0ac1b89ebdfg

## What I'm doing:

We should also perform a retry when encountering a `RuntimeException` whose cause is a `NoSuchElementException`.
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

---

> [!NOTE]
> Centralizes retry logic for `execPlanFragmentAsync` and `cancelPlanFragmentAsync` to handle `NoSuchElementException` and its `RuntimeException` wrapper; adds unit tests.
> 
> - **RPC client (`fe/fe-core/src/main/java/com/starrocks/rpc/BackendServiceClient.java`)**:
>   - Introduce `retryNoSuchElement` helper with `RETRY_TIMES=2` to retry on `NoSuchElementException` or when wrapped in `RuntimeException` (via `getCause`).
>   - Refactor `sendPlanFragmentAsync` and `cancelPlanFragmentAsync` to use the new retry helper; preserve timing/timeout setup for exec path.
>   - Improve sleep interruption handling by restoring the interrupt flag.
> - **Tests (`fe/fe-core/src/test/java/com/starrocks/rpc/BackendServiceClientTest.java`)**:
>   - Add tests covering retry success/failure for both `execPlanFragmentAsync` and `cancelPlanFragmentAsync`, including cases where `NoSuchElementException` is wrapped in `RuntimeException`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23f70ac0cf9816a4ce41cc8921159002826654b4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/rpc/BackendServiceClient.java
+++ b/fe/fe-core/src/main/java/com/starrocks/rpc/BackendServiceClient.java
@@ -69,10 +69,13 @@ import org.apache.logging.log4j.Logger;
 import org.apache.thrift.TException;
 
 import java.util.NoSuchElementException;
+import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
 
 public class BackendServiceClient {
     private static final Logger LOG = LogManager.getLogger(BackendServiceClient.class);
+
+    private static final int RETRY_TIMES = 2;
 
     private BackendServiceClient() {
     }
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/rpc/BackendServiceClient.java
```diff
@@ -84,30 +87,17 @@ public static BackendServiceClient getInstance() {
...
-            }
-        } catch (Throwable e) {
-            LOG.warn("Execute plan fragment catch a exception, address={}:{}",
-                    address.getHostname(), address.getPort(), e);
-            throw new RpcException(address.hostname, e.getMessage());
-        }
+
+        return retryNoSuchElement(
+                () -> {
+                    try (Timer ignored = Tracers.watchScope(Tracers.Module.SCHEDULER, "DeployAsyncSendTime")) {
```
kevincai, line 93: should this scope time be the outside, wrapping the entire retry op?

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/rpc/BackendServiceClient.java
```diff
@@ -314,6 +289,30 @@ public Future<PProcessDictionaryCacheResult> processDictionaryCache(
         }
     }
 
+    private <T> T retryNoSuchElement(Callable<T> action, TNetworkAddress address, String actionName) throws RpcException {
```
alvin-phoenix-ai, line 292: Is it better to make retry time as an argument of this function?

ZiheLiu:
Initially, we thought the issue was caused by missing retry logic on our side, but in fact the GenericObjectPool we use already checks whether existing connections are still alive. If a connection is not alive, it will create a new one. The “Unable to validate object” error is thrown while re-establishing a connection to this BE IP, rather than when sending a packet over the old connection.

AWS has an ARP (Address Resolution Protocol) cache that stores mappings from IP to MAC addresses.  
The MAC address cached for a given IP is only refreshed when a request to that IP times out once, or when the cache is full and entries are evicted.

PR #66368, opened 2025-12-05
[Doc] Update connection_id example in Query_management.md

Lordeath:
## Why I'm doing:
this connect id might be wrong
## What I'm doing:
fix it

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [x] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3


DanRoscigno:
Please sign the CLA @Lordeath

docs/zh/administration/management/resource_management/Query_management.md
alvin-phoenix-ai, line 133: I think the case is right, because it is going to kill 114 session which is listed in the result of `show processlist`
And the current session is 98. it is not going to kill itself.

Lordeath: Alright, this is my mistake. I saw 114 in SHOW PROCESSLIST, and below it the example also demonstrated KILL 114.

What I was trying to say is that the document first demonstrates the command KILL connection_id;, and then later demonstrates how to obtain the connection ID.

So does that mean the ID shown in SHOW PROCESSLIST is the connection ID?

Maybe I’m just being slow, because I found this a bit hard to understand at first. But I understand it now.

PR #66399, opened 2025-12-06
[Feature] Enable external catalog discovery for MySQL-compatible clients

### LINKED ISSUE #40287: Show external catalog with their schemas in IDEs like Dbeaver, Datagrip etc
Hi, is there a way  to make external catalogs show up in IDEs like DBeaver, Datagrip etc in the schema views?
currently i only see the "internal" SR managed tables/DB and external catalogs must be manually called by SHOW CATALOGS

 The IDEs do the schema intropection which should show the full schema of external catalogs just like how it shows for internal catalogs

Example:- 
image

yushesp:
## Changes:

Add opt-in feature (enable_cross_catalog_database_list) that allows MySQL clients like DBeaver, Metabase, DataGrip etc to discover databases across external catalogs (Hive, Polaris, etc.).

Key features:
- Connect with catalog name in database field to browse that catalog
- SHOW DATABASES from default_catalog shows all catalogs in catalog.db format
- information_schema tables work with catalog.database format

IMPORTANT: If an external catalog has the same name as a database in the internal catalog, the database takes precedence. Ensure external catalog names don't conflict with internal database names when using this feature.

All changes gated behind enable_cross_catalog_database_list (default: false) for full backwards compatibility.

## What I'm doing:

Fixes #40287

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

### DIFF
--- a/be/src/exec/schema_scanner/schema_schemata_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_schemata_scanner.cpp
@@ -68,8 +68,12 @@ Status SchemaSchemataScanner::fill_chunk(ChunkPtr* chunk) {
             // catalog
             {
                 ColumnPtr column = (*chunk)->get_column_by_slot_id(1);
-                const char* str = "def";
-                Slice value(str, strlen(str));
+                // Use actual catalog name from response if available, otherwise fallback to "def"
+                std::string catalog_name = "def";
+                if (_db_result.__isset.catalogs && _db_index < _db_result.catalogs.size()) {
+                    catalog_name = _db_result.catalogs[_db_index];
+                }
+                Slice value(catalog_name.c_str(), catalog_name.length());
                 fill_column_with_slot<TYPE_VARCHAR>(column.get(), (void*)&value);
             }
             break;
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/system/information/TablesSystemTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/system/information/TablesSystemTable.java
@@ -19,6 +19,7 @@ import com.starrocks.authentication.UserIdentityUtils;
 import com.starrocks.catalog.Column;
 import com.starrocks.catalog.Table;
 import com.starrocks.catalog.system.SystemId;
+import com.starrocks.common.Config;
 import com.starrocks.catalog.system.SystemTable;
 import com.starrocks.qe.ConnectContext;
 import com.starrocks.service.InformationSchemaDataSource;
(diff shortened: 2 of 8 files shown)


alvin-phoenix-ai:
@yushesp 
First of all, thank you for the proposal. Although you have considered backward compatibility, it is still a significant change for StarSrocks, especially if we want users to use it in a production environment.
FE configuration is a system change, even though this change can make DBeaver work well. But at the same time, the other uses or applications will exhibit unexpected behavior, getting all databases from different catalogs. From my perspective, it will mislead users.

yushesp:
> @yushesp First of all, thank you for the proposal. Although you have considered backward compatibility, it is still a significant change for StarSrocks, especially if we want users to use it in a production environment. FE configuration is a system change, even though this change can make DBeaver work well. But at the same time, the other uses or applications will exhibit unexpected behavior, getting all databases from different catalogs. From my perspective, it will mislead users.

Thank you for the feedback @alvin-celerdata, it makes sense. A system-wide config does affect all connections, which could surprise existing users/applications.

Is it possible to make this a session variable instead? That way:
- Users explicitly opt-in per connection: SET enable_cross_catalog_database_list = true
- BI tools can pass it via JDBC URL: sessionVariables=enable_cross_catalog_database_list=true
 - Existing applications remain completely unaffected
- Admins could still SET GLOBAL as a default if they want cluster-wide behavior

Would this approach address your concerns? Happy to update the PR if so.

alvin-phoenix-ai:
> > @yushesp First of all, thank you for the proposal. Although you have considered backward compatibility, it is still a significant change for StarSrocks, especially if we want users to use it in a production environment. FE configuration is a system change, even though this change can make DBeaver work well. But at the same time, the other uses or applications will exhibit unexpected behavior, getting all databases from different catalogs. From my perspective, it will mislead users.
> 
> Thank you for the feedback @alvin-celerdata, it makes sense. A system-wide config does affect all connections, which could surprise existing users/applications.
> 
> Is it possible to make this a session variable instead? That way:
> 
> * Users explicitly opt-in per connection: SET enable_cross_catalog_database_list = true
> * BI tools can pass it via JDBC URL: sessionVariables=enable_cross_catalog_database_list=true
> * Existing applications remain completely unaffected
> * Admins could still SET GLOBAL as a default if they want cluster-wide behavior
> 
> Would this approach address your concerns? Happy to update the PR if so.

@yushesp 
This approach addresses the concern about the influence of other users. However, this change is so fundamental that I can't make a decision right now. 
This should be discussed in a separate GitHub issue or Google Document, as it will affect many aspects of the system. And the team needs more time to evaluate the solution.

PR #66566, opened 2025-12-10
[BugFix] Support FE_PASSWORD in BE and CN docker entrypoints 

arvindKandpal-ksolves:
## Why I'm doing:

The current Docker entrypoint scripts (`be_entrypoint.sh` and `cn_entrypoint.sh`) hardcode the command `mysql -u root` without a password option.
If a user changes the Frontend (FE) root password for security reasons (e.g., `SET PASSWORD ...`), the Backend (BE) and Compute Nodes (CN) fail to register with the FE or send heartbeats, resulting in `ERROR 1045 (28000): Access denied`.

## What I'm doing:

1. Modified `docker/dockerfiles/be/be_entrypoint.sh` and `docker/dockerfiles/cn/cn_entrypoint.sh`.
2. Added logic to check for an optional environment variable `FE_PASSWORD`.
3. If `FE_PASSWORD` is set, the script injects the `-p${FE_PASSWORD}` flag into the `mysql` commands used for node registration and health checks.
4. If the variable is not set, it retains the default behavior (connecting without a password).

Fixes #66551

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/docker/dockerfiles/be/be_entrypoint.sh
+++ b/docker/dockerfiles/be/be_entrypoint.sh
@@ -17,6 +17,10 @@ STARROCKS_ROOT=${STARROCKS_ROOT:-"/opt/starrocks"}
 export STARROCKS_HOME=${STARROCKS_ROOT}/be
 BE_CONFIG=$STARROCKS_HOME/conf/be.conf
 
+MYSQL_PWD_OPT=""
+if [[ -n "$FE_PASSWORD" ]]; then
+    MYSQL_PWD_OPT="-p${FE_PASSWORD}"
+fi
 
 log_stderr()
 {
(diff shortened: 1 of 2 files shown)


kevincai (changes requested):
no need to explicit the password, `MYSQL_PWD` env did the same trick, try to set `MYSQL_PWD` to the docker env, everything will work again.

arvindKandpal-ksolves:
@kevincai You are absolutely right.

I verified this locally by setting the `MYSQL_PWD` environment variable in my Docker Compose configuration without modifying the entrypoint scripts. The backend successfully registered with the Frontend, and I did not encounter the Access Denied error.

It seems I missed that the standard mysql client automatically picks up `MYSQL_PWD`. Thanks for pointing this out!

I am closing this PR as the code change is unnecessary.

PR #66640, opened 2025-12-11
[Enhancement] Use more sophisticated statistics to infer skew for `GroupByCountDistinctDataSkewEliminateRule`

OliLay:
## Why I'm doing:

Given a query like
```sql
select
    `TABLE`.`GROUP`,
    count(distinct `TABLE`.`DISTINCT`) as cnt
from
    `TABLE`
group by
    `TABLE`.`GROUP`
```
where `GROUP` may be skewed heavily, the actual aggregation of the skewed group is executed single-threaded. For this reason there is the `GroupByCountDistinctDataSkewEliminateRule`, which actually did not consider sophisticated column statistics. This would yield to the rule only being activated very seldomly. 
For example, for a case with 100M rows and the `GROUP` column being skewed heavily:
```
select `TABLE`.`GROUP`, count(`TABLE`.`GROUP`) from `TABLE` group by 1 order by 2 desc limit 5;
+-------+--------------------+
| GROUP | count(TABLE.GROUP) |
+-------+--------------------+
|     0 |           98175725 |
|     1 |              68181 |
|     2 |              40059 |
|     3 |              28474 |
|     4 |              22114 |
+-------+--------------------+
```
executing the above query without the change takes >50s ([profile](https://gist.github.com/OliLay/1f631023b42e127168a97e1d28c5fb2d)), while with the change it takes <12s ([profile](https://gist.github.com/OliLay/6c085c59f38e32144b6721d54dd3a4fb)) due to the rewrite. 

Note the whole rule is behind `enable_distinct_column_bucketization` (which is default `false`).

## What I'm doing:
- Weighting cost of `GroupByCountDistinctDataSkewEliminateRule` was previously only done by looking at the cardinality of the group by columns; now it is also done based on histograms and null fractions. The old logic is still there, I just added a new condition in which the rule may trigger.
- Created `DataSkew` utility which can be re-used to check a column for data skew instead of implementing it in every rule itself.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Enhances skew inference for GROUP BY COUNT(DISTINCT) by leveraging histograms/null fractions in cost model, adds a configurable skew threshold, introduces reusable DataSkew utility, and adds tests.
> 
> - **Optimizer/Cost Model**:
>   - Improve `computeDataSkewPenaltyOfGroupByCountDistinct` in `CostModel` to detect skew via column histograms and null fractions; adds helpers to evaluate group-by skew and cardinality; adjusts penalty factors (e.g., 0.2/0.5/1.5).
>   - Refactor imports to use `skew` package and `ColumnRefOperator` where needed.
> - **Rule**:
>   - Keep trigger condition for `GroupByCountDistinctDataSkewEliminateRule` but refactor `check` for clarity; continues to honor `enable_distinct_column_bucketization`.
> - **Session Variables**:
>   - Add `data_skew_row_percentage_threshold` (default `0.2`) with getters/setters to tune skew detection.
> - **Utilities**:
>   - Add reusable `sql/optimizer/skew/DataSkew` with configurable thresholds and MCV-based/null-fraction skew checks.
>   - Move `DataSkewInfo` to `sql/optimizer/skew` and streamline fields.
> - **Tests**:
>   - Add `DataSkewTest` for null- and histogram-based skew detection and threshold handling.
>   - Add `DistinctAggSkewTest` asserting plan rewrites for skewed vs non-skewed cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 15224e78885c837733017f22bae73b81f27e771d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
```diff
@@ -281,52 +283,96 @@ public CostEstimate visitPhysicalHashAggregate(PhysicalHashAggregateOperator nod
...
+            final var groupBys = getGroupBysWithoutDistinctColumn(node.getGroupBys(), countDistinctColumnRef);
+            final var groupByStatistics = groupBys.stream()
+                    .map(inputStatistics::getColumnStatistic) //
+                    .filter(Objects::nonNull) //
+                    .collect(Collectors.toList());
+
+            // Basic skew detection on the group by columns. We are weighting this heavily if we know
+            // that group by columns are skewed since this optimization has significant speed up then.
+            if (isGroupBySkewed(groupByStatistics, inputStatistics)) {
+                return 0.2;
```
OliLay, line 375: If we know that the group by is skewed we know that the aggregation is going to be executed single-threadedly for the large groups. Hence, we have to take the skew into consideration when determining the cost. In my case, the speedup was 5xish, so I choose this factor for the agg as well.

alvin-phoenix-ai:
@cursor review

stephen-shelby:
please fix the code style first

OliLay:
> please fix the code style first

Done [458a2b6](https://github.com/StarRocks/starrocks/pull/66640/commits/458a2b61d95b0612690151ec3862f4c30d72dbdc)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/skew/DataSkew.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+
+public class DataSkew {
+    private static final int DEFAULT_MCV_LIMIT = 5;
+    private static final double DEFAULT_RELATIVE_ROW_THRESHOLD = 0.2;
+    private static final Thresholds DEFAULT_THRESHOLDS = new Thresholds(DEFAULT_MCV_LIMIT, DEFAULT_RELATIVE_ROW_THRESHOLD);
+
+    public static class Thresholds {
```
murphyatwork: how about using `Record`

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/skew/DataSkew.java
```diff
@@ -0,0 +1,85 @@
...
+import com.starrocks.sql.optimizer.statistics.Statistics;
+
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+
+public class DataSkew {
+    private static final int DEFAULT_MCV_LIMIT = 5;
+    private static final double DEFAULT_RELATIVE_ROW_THRESHOLD = 0.2;
```
murphyatwork, line 28: does it make more sense to make it configurable ? i don't understand why it's `0.2`

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/skew/DataSkew.java
```diff
@@ -0,0 +1,85 @@
...
+                List<Pair<String, Long>> mcvs = Lists.newArrayList();
+                int mcvLimit = Math.min(thresholds.mcvLimit, mcv.size());
+                mcv.entrySet().stream()
+                        .sorted(Map.Entry.comparingByValue(Comparator.reverseOrder())) //
+                        .limit(mcvLimit)
+                        .forEach(entry -> {
+                            mcvs.add(Pair.create(entry.getKey(), entry.getValue()));
+                        });
+
+                long rowCountOfMcvs = mcvs.stream().mapToLong(pair -> pair.second).sum();
```
murphyatwork, line 72: Does it mean that these common values make up most of the data?

murphyatwork (dismissed):

satanson (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/skew/DataSkew.java
OliLay: [15224e7](https://github.com/StarRocks/starrocks/pull/66640/commits/15224e78885c837733017f22bae73b81f27e771d)
Initially did not do it because I want to backport it to 3.3 (only for us) and there I can't use it. But I can change it in the backport itself then again.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/skew/DataSkew.java
OliLay: I added a new session variable which can be taken into account for more skew use cases in the future: [15224e7](https://github.com/StarRocks/starrocks/pull/66640/commits/15224e78885c837733017f22bae73b81f27e771d)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/skew/DataSkew.java
OliLay: Yes, with the default limits this means that the top 5 MCV make up >= 20% of the rows.

alvin-phoenix-ai:
@cursor review

stephen-shelby:
The profiles in the pr description are the same. all take 56s

OliLay:
> The profiles in the pr description are the same. all take 56s

Sorry, linked the wrong gist, should now be the correct one

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
```diff
@@ -281,52 +283,104 @@ public CostEstimate visitPhysicalHashAggregate(PhysicalHashAggregateOperator nod
...
-                    .stream().map(inputStatistics::getColumnStatistic).collect(Collectors.toList());
+            final var dataSkewRowPercentageThreshold =
+                    ConnectContext.get().getSessionVariable().getDataSkewRowPercentageThreshold();
+            final var skewThresholds = DataSkew.Thresholds.withRelativeRowThreshold(dataSkewRowPercentageThreshold);
 
-            if (distColStat.isUnknownValue() || distColStat.isUnknown() ||
-                    groupByStats.stream().anyMatch(groupStat -> groupStat.isUnknown() || groupStat.isUnknownValue())) {
-                return 1.5;
+            return groupByStats.stream()
+                    .allMatch(groupByStat -> {
```
stephen-shelby, line 305: `allMatch` logic may be too restrictive

OliLay: Its restrictive, but if at least one column is not skewed then most likely the group by also isn't skewed. So I think its necessary, else you would have false-positive activations of this rule

murphyatwork (approved):

OliLay:
Hey @satanson , can you have another look? I had to dismiss your approval because I incorporated some of Murphys suggested changes.

satanson (approved):

stephen-shelby (approved):

PR #66699, opened 2025-12-14
[Enhancement] Add BE TableSchemaService for Fast Schema Evolution in shared-data

banmoy:
## Why I'm doing:

Fast Schema Evolution v2 (shared-data / lake) allows FE to evolve table schema without propagating schema metas to BEs. In this mode, BE cannot reliably get the correct schema from schema file/tablet metadata alone, which can break:

- **LOAD** (DeltaWriter building writer schema / partial update / auto-inc, etc.)
- **SCAN** (LakeDataSource building reader schema)
- **TxnLog apply** (updating tablet metadata schema when publish/apply sees newer schema)

This PR introduces a BE-side schema fetch service so BE can **pull schema on-demand from FE** with caching and deduplication, and adds compatibility fallbacks for mixed-version upgrade.

## What I'm doing:

- Add BE `TableSchemaService` for shared-data (lake) to retrieve `TabletSchema` for both **LOAD** and **SCAN**:
  - Two-level local lookup: global schema cache (TabletManager) + tablet metadata (current/historical schema).
  - Remote fetch from FE via thrift RPC `getTableSchema` when local miss.
  - Deduplicate concurrent requests using SingleFlight (grouped by schema + FE; last retry isolates by query/txn to avoid cross-query/txn interference).
  - **LOAD-only fallback**: when FE doesn’t support `getTableSchema` (method not found), fallback to schema-file / tablet schema path for compatibility.

- Wire schema_meta-driven schema update for fast schema evolution v2 in txn-log apply:
  - If `op_write.schema_meta.schema_version` is newer than tablet metadata schema version, fetch new schema via `TableSchemaService` and update tablet metadata schema.
  - Backfill `rowset_to_schema` and archive old schema into `historical_schemas` when needed.

- Add/adjust configs & protocol:
  - Add `config::table_schema_service_max_retries` (default `3`).
  - Update lake txn log proto to carry schema meta needed by the apply path.

- **UT**:
  - `be/test/storage/lake/table_schema_service_test.cpp`
  - `be/test/storage/lake/fast_schema_evolution_v2_test.cpp`
  - Updated: `be/test/storage/lake/delta_writer_test.cpp`, `be/test/storage/lake/lake_data_source_test.cpp`

Fixes #65706

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Introduce a BE TableSchemaService with caching and FE RPC to fetch table schemas on demand, integrate schema_key across load/scan paths, and update txn log/apply to persist and upgrade schemas.
> 
> - **Backend (Lake/BE)**:
>   - **New `TableSchemaService`** (`storage/lake/table_schema_service.*`): local cache + FE RPC (`getTableSchema`) with singleflight dedup and retries; config `table_schema_service_max_retries` added.
>   - **Schema fetch integration**:
>     - `connector/lake_connector`: fetch scan schema via `schema_key`; fallback to tablet schema if absent.
>     - `lake/async_delta_writer` & `lake/delta_writer`: plumb `db_id`, fetch load schema via service; record `schema_key` in `TxnLogPB.OpWrite`.
>     - `lake/txn_log_applier`: on publish/apply, update tablet metadata schema via service; backfill `rowset_to_schema` and archive historical schemas.
>     - `lake/tablet_manager`: own/expose `TableSchemaService`; add schema cache helpers.
>   - Build: add `table_schema_service.cpp` to storage.
> - **Protocol/FE**:
>   - Replace `TTableSchemaMeta` with `TTableSchemaKey`; add `TableSchemaKeyPB`; extend `TxnLogPB.OpWrite` with `schema_key`.
>   - Update FE `TableSchemaService` and `OlapScanNode` to use `schema_key`.
> - **Docs**: document `table_schema_service_max_retries` in BE config (EN/JA/ZH).
> - **Tests**: add comprehensive UTs for schema service, fast schema evolution v2, and adapt lake writer/data source tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 411e15259276c7d002a2e624b8e34af8ca1a96c0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -186,7 +186,7 @@ Status LakeDataSource::get_tablet(const TInternalScanRange& scan_range) {
         schema_meta_pb.set_schema_id(schema_meta.schema_id);
         schema_meta_pb.set_db_id(schema_meta.db_id);
         schema_meta_pb.set_table_id(schema_meta.table_id);
-        ASSIGN_OR_RETURN(_tablet_schema, tablet_manager->table_schema_service()->get_scan_schema(
+        ASSIGN_OR_RETURN(_tablet_schema, tablet_manager->table_schema_service()->get_schema_for_scan(
                                                  schema_meta_pb, tablet_id, _runtime_state->query_id(),
                                                  _runtime_state->fragment_ctx()->fe_addr(), _tablet.metadata()));
     } else {
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -24,6 +24,7 @@
 #include "storage/chunk_helper.h"
 #include "storage/column_predicate_rewriter.h"
 #include "storage/lake/tablet.h"
+#include "storage/lake/table_schema_service.h"
 #include "storage/predicate_parser.h"
 #include "storage/predicate_tree/predicate_tree.hpp"
 #include "storage/projection_iterator.h"
@@ -175,8 +176,24 @@ Status LakeDataSource::get_next(RuntimeState* state, ChunkPtr* chunk) {
 Status LakeDataSource::get_tablet(const TInternalScanRange& scan_range) {
     int64_t tablet_id = scan_range.tablet_id;
     int64_t version = strtoul(scan_range.version.c_str(), nullptr, 10);
-    ASSIGN_OR_RETURN(_tablet, ExecEnv::GetInstance()->lake_tablet_manager()->get_tablet(tablet_id, version));
-    _tablet_schema = _tablet.get_schema();
+    auto tablet_manager = ExecEnv::GetInstance()->lake_tablet_manager();
+    ASSIGN_OR_RETURN(_tablet, tablet_manager->get_tablet(tablet_id, version));
+    auto& lake_scan_node = _provider->_t_lake_scan_node;
+    if (lake_scan_node.__isset.schema_meta) {
+        auto& schema_meta = lake_scan_node.schema_meta;
+        auto schema_info = lake::TableSchemaService::TableSchemaInfo{
+                .schema_id = schema_meta.schema_id,
+                .db_id = schema_meta.db_id,
+                .table_id = schema_meta.table_id,
+                .tablet_id = tablet_id};
+        ASSIGN_OR_RETURN(_tablet_schema, tablet_manager->table_schema_service()->get_scan_schema(
+                schema_info, _runtime_state->query_id(),
+                _runtime_state->fragment_ctx()->fe_addr(), _tablet.metadata()));
+    } else {
+        // no table schema meta indicates FE has not been upgraded to use fast schema evolution v2,
+        // so fallback to the old way to get schema from tablet metadata
+        _tablet_schema = _tablet.get_schema();
+    }
     return Status::OK();
 }
 
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -20,6 +20,7 @@
 #include "exec/connector_scan_node.h"
 #include "exec/olap_scan_prepare.h"
 #include "exec/pipeline/fragment_context.h"
+#include "gen_cpp/lake_types.pb.h"
 #include "runtime/global_dict/parser.h"
 #include "storage/chunk_helper.h"
 #include "storage/column_predicate_rewriter.h"
@@ -180,13 +181,13 @@ Status LakeDataSource::get_tablet(const TInternalScanRange& scan_range) {
     ASSIGN_OR_RETURN(_tablet, tablet_manager->get_tablet(tablet_id, version));
     auto& lake_scan_node = _provider->_t_lake_scan_node;
     if (lake_scan_node.__isset.schema_meta) {
-        auto& schema_meta = lake_scan_node.schema_meta;
-        auto schema_info = lake::TableSchemaService::TableSchemaInfo{.schema_id = schema_meta.schema_id,
-                                                                     .db_id = schema_meta.db_id,
-                                                                     .table_id = schema_meta.table_id,
-                                                                     .tablet_id = tablet_id};
+        const auto& schema_meta = lake_scan_node.schema_meta;
+        TableSchemaMetaPB schema_meta_pb;
+        schema_meta_pb.set_schema_id(schema_meta.schema_id);
+        schema_meta_pb.set_db_id(schema_meta.db_id);
+        schema_meta_pb.set_table_id(schema_meta.table_id);
         ASSIGN_OR_RETURN(_tablet_schema, tablet_manager->table_schema_service()->get_scan_schema(
-                                                 schema_info, _runtime_state->query_id(),
+                                                 schema_meta_pb, tablet_id, _runtime_state->query_id(),
                                                  _runtime_state->fragment_ctx()->fe_addr(), _tablet.metadata()));
     } else {
         // no table schema meta indicates FE has not been upgraded to use fast schema evolution v2,
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -179,12 +179,12 @@ Status LakeDataSource::get_tablet(const TInternalScanRange& scan_range) {
     auto tablet_manager = ExecEnv::GetInstance()->lake_tablet_manager();
     ASSIGN_OR_RETURN(_tablet, tablet_manager->get_tablet(tablet_id, version));
     auto& lake_scan_node = _provider->_t_lake_scan_node;
-    if (lake_scan_node.__isset.schema_meta) {
-        const auto& schema_meta = lake_scan_node.schema_meta;
+    if (lake_scan_node.__isset.schema_key) {
+        const auto& t_schema_key = lake_scan_node.schema_key;
         TableSchemaKeyPB schema_key_pb;
-        schema_key_pb.set_schema_id(schema_meta.schema_id);
-        schema_key_pb.set_db_id(schema_meta.db_id);
-        schema_key_pb.set_table_id(schema_meta.table_id);
+        schema_key_pb.set_schema_id(t_schema_key.schema_id);
+        schema_key_pb.set_db_id(t_schema_key.db_id);
+        schema_key_pb.set_table_id(t_schema_key.table_id);
         ASSIGN_OR_RETURN(_tablet_schema, tablet_manager->table_schema_service()->get_schema_for_scan(
                                                  schema_key_pb, tablet_id, _runtime_state->query_id(),
                                                  _runtime_state->fragment_ctx()->fe_addr(), _tablet.metadata()));
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -23,8 +23,8 @@
 #include "runtime/global_dict/parser.h"
 #include "storage/chunk_helper.h"
 #include "storage/column_predicate_rewriter.h"
-#include "storage/lake/tablet.h"
 #include "storage/lake/table_schema_service.h"
+#include "storage/lake/tablet.h"
 #include "storage/predicate_parser.h"
 #include "storage/predicate_tree/predicate_tree.hpp"
 #include "storage/projection_iterator.h"
@@ -181,14 +181,13 @@ Status LakeDataSource::get_tablet(const TInternalScanRange& scan_range) {
     auto& lake_scan_node = _provider->_t_lake_scan_node;
     if (lake_scan_node.__isset.schema_meta) {
         auto& schema_meta = lake_scan_node.schema_meta;
-        auto schema_info = lake::TableSchemaService::TableSchemaInfo{
-                .schema_id = schema_meta.schema_id,
-                .db_id = schema_meta.db_id,
-                .table_id = schema_meta.table_id,
-                .tablet_id = tablet_id};
+        auto schema_info = lake::TableSchemaService::TableSchemaInfo{.schema_id = schema_meta.schema_id,
+                                                                     .db_id = schema_meta.db_id,
+                                                                     .table_id = schema_meta.table_id,
+                                                                     .tablet_id = tablet_id};
         ASSIGN_OR_RETURN(_tablet_schema, tablet_manager->table_schema_service()->get_scan_schema(
-                schema_info, _runtime_state->query_id(),
-                _runtime_state->fragment_ctx()->fe_addr(), _tablet.metadata()));
+                                                 schema_info, _runtime_state->query_id(),
+                                                 _runtime_state->fragment_ctx()->fe_addr(), _tablet.metadata()));
     } else {
         // no table schema meta indicates FE has not been upgraded to use fast schema evolution v2,
         // so fallback to the old way to get schema from tablet metadata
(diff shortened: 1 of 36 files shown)


alvin-phoenix-ai:
@cursor review

be/src/storage/lake/table_schema_service.h
```diff
@@ -0,0 +1,199 @@
...
+ *    - Uses a SingleFlight mechanism to merge concurrent requests for the same schema, reducing pressure on the FE.
+ */
+class TableSchemaService {
+public:
+    struct TableSchemaInfo {
+        int64_t schema_id;
+        int64_t db_id;
+        int64_t table_id;
+        int64_t tablet_id;
+    };
```
alvin-phoenix-ai: if we have tablet_id and schema_id, is it necessary to identify one schema? Are db_id and table_id necessary?

gensrc/proto/lake_types.proto
```diff
@@ -222,6 +222,14 @@ message BundleTabletMetadataPB {
 }
 
 message TxnLogPB {
+
+    message TableSchemaMetaPB {
+        optional int64 db_id = 1;
+        optional int64 table_id = 2;
+        optional int64 schema_id = 3;
+        optional int64 schema_version = 4; 
+    }
```
alvin-phoenix-ai: I have some questions.
1. Why is this different from `TableSchemaInfo`?
2. is db_id and table_id necessary if there would be a tablet id?

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -1411,4 +1415,20 @@ StatusOr<TabletAndRowsets> TabletManager::capture_tablet_and_rowsets(int64_t tab
...
+    // GlobalTabletSchemaMap and metadata cache overlap in functionality, but because many places
+    // previously relied on GlobalTabletSchemaMap, caching is still performed in GlobalTabletSchemaMap
+    // here. In the future, it may be possible to refactor and remove GlobalTabletSchemaMap.
+    auto [cached_schema, inserted] = GlobalTabletSchemaMap::Instance()->emplace(schema);
+    auto cache_key = global_schema_cache_key(cached_schema->id());
+    auto cache_size = inserted ? cached_schema->mem_usage() : 0;
+    _metacache->cache_tablet_schema(cache_key, cached_schema, cache_size);
+}
+
+TabletSchemaPtr TabletManager::get_cached_global_schema(int64_t schema_id) {
```
alvin-phoenix-ai: Is this schema ID globally unique?

alvin-phoenix-ai: And what is the difference between `global_schema`, `tablet_schema` and `TableSchema`.
I think we need to make it clearer, to keep using the same terminology all the time.

be/src/storage/lake/table_schema_service.h
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * @brief Retrieves the table schema for a LOAD operation.
+     * 
+     * @param schema_info Basic information to identify the schema.
+     * @param txn_id The transaction ID associated with the load.
+     * @param tablet_meta Optional pointer to tablet metadata for local lookup.
+     * @return A shared pointer to the TabletSchema on success, or an error status.
+     */
+    StatusOr<TabletSchemaPtr> get_load_schema(const TableSchemaInfo& schema_info, int64_t txn_id,
```
alvin-phoenix-ai: Getting **Tablet**Schema with **Table**Schema is a little confusing.

be/src/storage/lake/table_schema_service.h
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * 2. Remote Fetch via Frontend (FE):
+ *    - If a schema is missing locally, it fetches explicitly from the FE via RPC.
+ *
+ * 3. Request Deduplication (SingleFlight):
+ *    - Uses a SingleFlight mechanism to merge concurrent requests for the same schema, reducing pressure on the FE.
+ */
+class TableSchemaService {
+public:
+    struct TableSchemaInfo {
```
alvin-phoenix-ai: Maybe changing 'Info' to 'Locator' or "ScopedSchemaId" or "SchemaKey" would be better for understanding

be/src/storage/lake/table_schema_service.h
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    /**
+     * @brief Result of a SingleFlight execution, shared among all waiters.
+     */
+    struct SingleFlightResult {
+        SingleFlightExecutionContext execution_ctx;
+        StatusOr<TabletSchemaPtr> rpc_result;
+    };
+    using SingleFlightResultPtr = std::shared_ptr<SingleFlightResult>;
```
alvin-phoenix-ai:
```suggestion
    using SingleFlightResultPtr = std::shared_ptr<const SingleFlightResult>;
```
It is safe to use a const restrict.

be/src/storage/lake/table_schema_service.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+    if (tablet_meta != nullptr) {
+        const TabletSchemaPB* schema_pb = nullptr;
+        if (schema_id == tablet_meta->schema().id()) {
+            schema_pb = &tablet_meta->schema();
+        } else {
+            auto it = tablet_meta->historical_schemas().find(schema_id);
+            if (it != tablet_meta->historical_schemas().end()) {
+                schema_pb = &it->second;
+            }
```
alvin-phoenix-ai, line 158: It is not safe because of no lock protection. 
The `tablet_meta->historical_schemas` can be changed between find and verify. Actually, the `tablet_meta->historical_schemas` may be changed during `find`
Please have a function in tablet_meta and do it under a shared lock protection.

be/src/storage/lake/table_schema_service.h
banmoy: schema_id is allocated by FE, and globally unique. We can identify a schema just by id. But FE does not maintain a global mapping from schema_id -> schema, so we need to first find the OlapTable in the catalog according to db_id/table_id, then get the schema according to schema_id. db_id/table_id is for the efficiency to lookup the schema without introducing a global schema mapping

be/src/storage/lake/tablet_manager.cpp
banmoy: 1. schema id is allocated by FE, and is globally unique
2. remove the confusing `global_schema`. add an explanation about `tablet_schema` and `TableSchema`  in class `TableSchemaService`. see https://github.com/StarRocks/starrocks/pull/66699/commits/a3ccaec5a187dde9d1eeffa543646672e3c18933.

be/src/storage/lake/table_schema_service.h
banmoy: * `TabletSchema` class is actually the in-memory representation of a table schema version. It shouldn't be tied to "Tablet" or "Table". Something like `Schema` or `PhysicalSchema` would be more appropriate. However, `Schema` is already taken by `schema.h`.
*  Given the legacy reason and the extensive usage of this class, suggest addressing this refactor in a dedicated PR if must do it

be/src/storage/lake/table_schema_service.cpp
banmoy: The `tablet_meta` here is a `TabletMetadataPB`. In our system its usage has two modes:
  - After a meta is successfully published, it’s immutable and can be read concurrently without risk of modification.
  - While a meta is being built (during publish), only the publish thread can access it, and calls to `TableSchemaService` in that phase also run on the publish thread.

Therefore, the `tablet_meta` seen here is either immutable or accessed only by the single publish thread. In both cases, concurrent modification of `historical_schemas` cannot happen, so this code path is safe.

gensrc/proto/lake_types.proto
banmoy: 1. almost same, have removed `TableSchemaInfo`.
2. As said in the above comment, need db_id/table_id for looking up the schema on FE efficiently. Actually, we can also get the db_id/table_id according to tablet id on FE via TabletInvertedIndex. But a tablet can be dropped after BE sends the rpc request and before FE processes the request, so it can happen that the tablet can not be found in TabletInvertedIndex. We need to deal with such case, for example, BE should retry with another tablet id. This makes error handling more complex. So I use db_id/table_id which is unchanging, and if can't find the table, just fail the request on BE side

be/src/storage/lake/table_schema_service.h
banmoy: have removed it, just reuse TableSchemaMetaPB

be/src/storage/lake/table_schema_service.h
banmoy: fixed

alvin-phoenix-ai:
@cursor review

gensrc/proto/lake_types.proto
```diff
@@ -221,7 +221,14 @@ message BundleTabletMetadataPB {
     map<int64, PagePointerPB> tablet_meta_pages = 3;
 }
 
+message TableSchemaMetaPB {
```
alvin-phoenix-ai: I would like `ScopedTableSchemaIdPB` or `TableSchemaKeyPB`. Because it is only a key or id to locate a schema, rather than information in the schema. Naming it "schema meta" will cause confusion.

be/src/storage/lake/table_schema_service.cpp
alvin-phoenix-ai: Is it possible that this tablet_meta will be destroyed by others? To answer this question, we should know how this tablet_meta is reclaimed in our system. Will it happen when this function is running?

gensrc/proto/lake_types.proto
banmoy: fixed. rename both proto and thrift  to `TableSchemaKeyPB`/`TTableSchemaKey`

be/src/storage/lake/table_schema_service.cpp
banmoy: We use it synchronously, and not access it after the function finishes. The caller holds a reference before calling the function, and not release the ref before the function finishes. Do you think it should copy the shared_ptr to ensure more safe?

alvin-phoenix-ai:
@cursor review

be/src/storage/lake/table_schema_service.cpp
alvin-phoenix-ai: Got it, if the caller has the shared_ptr reference, it will make sense.

be/src/connector/lake_connector.cpp
```diff
@@ -175,8 +176,23 @@ Status LakeDataSource::get_next(RuntimeState* state, ChunkPtr* chunk) {
...
+    auto& lake_scan_node = _provider->_t_lake_scan_node;
+    if (lake_scan_node.__isset.schema_key) {
+        const auto& t_schema_key = lake_scan_node.schema_key;
+        TableSchemaKeyPB schema_key_pb;
+        schema_key_pb.set_schema_id(t_schema_key.schema_id);
+        schema_key_pb.set_db_id(t_schema_key.db_id);
+        schema_key_pb.set_table_id(t_schema_key.table_id);
+        ASSIGN_OR_RETURN(_tablet_schema, tablet_manager->table_schema_service()->get_schema_for_scan(
+                                                 schema_key_pb, tablet_id, _runtime_state->query_id(),
+                                                 _runtime_state->fragment_ctx()->fe_addr(), _tablet.metadata()));
```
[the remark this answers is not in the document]
banmoy: not necessary according to the style in other  places of lake_connector

alvin-phoenix-ai:
@cursor review

be/src/storage/lake/delta_writer.cpp
```diff
@@ -407,18 +410,14 @@ inline Status DeltaWriterImpl::init_tablet_schema() {
...
-        return res.status();
-    }
+    TabletMetadataPtr latest_metadata = _tablet_manager->get_latest_cached_tablet_metadata(_tablet_id);
+    TableSchemaKeyPB schema_key;
+    schema_key.set_schema_id(_schema_id);
+    schema_key.set_db_id(_db_id);
+    schema_key.set_table_id(_table_id);
+    ASSIGN_OR_RETURN(_tablet_schema, _tablet_manager->table_schema_service()->get_schema_for_load(
+                                             schema_key, _tablet_id, _txn_id, latest_metadata));
+    return Status::OK();
```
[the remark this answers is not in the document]
banmoy: @cursoragent 
1. not reading from schema-file is expected because the PR wants to abandon it
2. TableSchemaService still checks the latest tablet metadata 
3. FE is the source of truth, so send RPC to FE is expected if can not find the schema locally
4. `db_id/table_id` have been set in the normal load path, so RPC could not fail because of this

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -1428,4 +1432,20 @@ StatusOr<TabletAndRowsets> TabletManager::capture_tablet_and_rowsets(int64_t tab
...
+
+void TabletManager::cache_schema(const TabletSchemaPtr& schema) {
+    // GlobalTabletSchemaMap and metadata cache overlap in functionality, but because many places
+    // previously relied on GlobalTabletSchemaMap, caching is still performed in GlobalTabletSchemaMap
+    // here. In the future, it may be possible to refactor and remove GlobalTabletSchemaMap.
+    auto [cached_schema, inserted] = GlobalTabletSchemaMap::Instance()->emplace(schema);
+    auto cache_key = global_schema_cache_key(cached_schema->id());
+    auto cache_size = inserted ? cached_schema->mem_usage() : 0;
+    _metacache->cache_tablet_schema(cache_key, cached_schema, cache_size);
+}
```
xiangguangyxg, line 1444: Why not keep using GlobalTabletSchemaMap

banmoy: GlobalTabletSchemaMap uses `weak_ptr`. If no one holds a shared_ptr outside map, it's gone, so it does not implement a real lru cache

alvin-phoenix-ai (approved):
The changes in protobuf and thrift look good to me

alvin-phoenix-ai:
@cursor review

be/test/storage/lake/table_schema_service_test.cpp
```diff
@@ -0,0 +1,764 @@
...
+    for (const auto& c : cases) {
+        // Each sub-case runs in its own SyncPoint scope to avoid cross-test interference.
+        ScopedSyncPoint sp;
+        SCOPED_TRACE(c.name);
+        int64_t schema_id = next_id();
+        c.setup(this);
+        // Trigger LOAD path to exercise _get_remote_schema() error handling.
+        auto result = _schema_service->get_schema_for_load(create_schema_info(schema_id, 100, 101), 1000, 1);
+        c.check(result.status());
+    }
```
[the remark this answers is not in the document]
banmoy: load will fallback to schema file if not supported, then return not found because the schema file/table does not exist

xiangguangyxg (approved):

wyb (approved):

Issue #64986, opened 2025-11-05
New multi-tenant data management


xiangguangyxg:
- [x] https://github.com/StarRocks/starrocks/pull/68591
- [x] https://github.com/StarRocks/starrocks/pull/68569
- [x] https://github.com/StarRocks/starrocks/pull/68497
- [x] https://github.com/StarRocks/starrocks/pull/68342
- [x] https://github.com/StarRocks/starrocks/pull/67441
- [x] https://github.com/StarRocks/starrocks/pull/66672
- [x] https://github.com/StarRocks/starrocks/pull/67082
- [x] https://github.com/StarRocks/starrocks/pull/65389
- [x] https://github.com/StarRocks/starrocks/pull/66775
- [x] https://github.com/StarRocks/starrocks/pull/67415
- [x] https://github.com/StarRocks/starrocks/pull/67595
- [x] https://github.com/StarRocks/starrocks/pull/68207
- [x] https://github.com/StarRocks/starrocks/pull/68593
- [x] https://github.com/StarRocks/starrocks/pull/67056
- [x] https://github.com/StarRocks/starrocks/pull/66743

## Feature request

StarRocks provides strong capabilities for large-scale analytical workloads, its current design introduces usability challenges, lacks adaptive mechanisms for handling data skew, and may deliver suboptimal performance for small-tenant in multi-tenant scenarios.

A new multi-tenant data management is needed for StarRocks, aiming to:

**Fewer concepts and simpler usage**
- In most cases, table creation should require only column definitions and the specification of ORDER BY clauses.
- Users don't need to define DISTRIBUTED BY clauses.

**Ability to handle data skew**
- Tablets support automatic splitting and merging, enabling the system to rebalance data dynamically and mitigate skew.

**Balanced data locality and distribution**
- Data from a small tenant can be located on a single compute node for optimal locality.
- Data from a large tenant can be distributed across multiple compute nodes to maintain scalability and performance.

**No dependency on time-based partitioning**
- System can no longer rely on time-based partitioning for data management.
- A single partition will be capable of storing large data volumes, and the amount of data in different partitions can vary greatly, allowing users to decide partitioning strategies based on their specific needs.

PR #66743, opened 2025-12-15
[Feature] Add tablet range pruning for range-distribution Lake tables for shared segments and sstables.

srlch:
## Summary

Adds tablet-range based pruning for Lake range-distribution tablets across:
Query path (Rowset → SegmentReadOptions.tablet_range → SegmentIterator::_apply_tablet_range), only for shared segments.

PK index compaction path (both serial LakePersistentIndex::major_compact and parallel LakePersistentIndexParallelCompactMgr) via TabletRangeHelper::create_sst_seek_range_from.

Centralizes TabletRangePB parsing in TabletRangeHelper and introduces a standalone SstSeekRange type under lake/, cleaning up includes and ownership between helpers and the compaction manager.

UTs cover single/multi-column PK, inclusive/exclusive bounds, invalid metadata, CHAR/VARCHAR, and both serial/parallel compaction with tablet range filtering.

Fixes #64986

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Introduces tablet-range pruning across query and compaction paths for Lake range-distribution tablets.
> 
> - Query path: propagate `SegmentReadOptions.tablet_range` from `Rowset` and apply in `SegmentIterator::_apply_tablet_range()` to filter shared segments only
> - PK index compaction: honor tablet range in both `LakePersistentIndex::major_compact` and `LakePersistentIndexParallelCompactMgr`, seeking/limiting via `SstSeekRange`; detect shared sstables
> - New helpers/types: `TabletRangeHelper` centralizes `TabletRangePB` parsing; `SstSeekRange` encapsulates [seek, stop) logic and intersection
> - API/struct changes: add optional `tablet_range` into `SegmentReadOptions`; adjust `merge_sstables`/`prepare_merging_iterator` signatures to pass metadata and shared flags; add `TabletRangePB range` to `TabletMetadataPB`
> - FE: enforce sort key equals primary key (including order) for PK tables (range distribution context); add analyzer tests
> - Build/tests: add `tablet_range_helper.cpp`, `sst_seek_range.*`; extensive UTs for single/multi-column keys, inclusive/exclusive bounds, CHAR/VARCHAR, invalid metadata, and serial/parallel compaction behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1909b9c0030324d706e8cdfff33a91614c256ce9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -16,12 +16,15 @@
 
 #include <future>
 
+#include "column/datum_convert.h"
 #include "runtime/current_thread.h"
+#include "runtime/types.h"
 #include "storage/chunk_helper.h"
 #include "storage/delete_predicates.h"
 #include "storage/lake/column_mode_partial_update_handler.h"
 #include "storage/lake/lake_delvec_loader.h"
 #include "storage/lake/tablet.h"
+#include "storage/lake/tablet_range_helper.h"
 #include "storage/lake/tablet_writer.h"
 #include "storage/lake/update_manager.h"
 #include "storage/projection_iterator.h"
@@ -31,6 +34,7 @@
 #include "storage/rowset/segment.h"
 #include "storage/rowset/segment_options.h"
 #include "storage/rowset/short_key_range_option.h"
+#include "storage/seek_range.h"
 #include "storage/tablet_schema_map.h"
 #include "storage/union_iterator.h"
 #include "types/logical_type.h"
@@ -280,11 +284,24 @@ StatusOr<std::vector<ChunkIteratorPtr>> Rowset::read(const Schema& schema, const
         options.stats->segments_read_count += num_segments();
     }
 
+    SeekRange tablet_range;
+    if (_tablet_metadata != nullptr && _tablet_metadata->has_range()) {
+        // do not use mem_pool here which means SeekRange is the reference of the data in tablet_metadata->range()
+        // and the data in tablet_metadata->range() has the same lifetime as the rowset
+        ASSIGN_OR_RETURN(tablet_range,
+                         TabletRangeHelper::create_seek_range_from(_tablet_metadata->range(), _tablet_schema, nullptr));
+    }
+
     std::vector<SegmentPtr> segments;
     RETURN_IF_ERROR(load_segments(&segments, seg_options, nullptr));
-    for (auto& seg_ptr : segments) {
+    for (int i = 0; i < segments.size(); i++) {
+        auto& seg_ptr = segments[i];
         if (seg_ptr->num_rows() == 0) {
             continue;
+        }
+        seg_options.tablet_range = std::nullopt;
+        if (i < _metadata->shared_segments_size() && _metadata->shared_segments(i)) {
+            seg_options.tablet_range = tablet_range;
         }
 
         if (options.rowid_range_option != nullptr) { // physical split.
@@ -359,7 +376,20 @@ StatusOr<std::vector<ChunkIteratorPtr>> Rowset::get_each_segment_iterator(const 
         RETURN_IF_ERROR(RecordPredicateHelper::check_valid_schema(*seg_options.record_predicate, schema));
     }
 
-    for (auto& seg_ptr : segments) {
+    SeekRange tablet_range;
+    if (_tablet_metadata != nullptr && _tablet_metadata->has_range()) {
+        // do not use mem_pool here which means SeekRange is the reference of the data in tablet_metadata->range()
+        // and the data in tablet_metadata->range() has the same lifetime as the rowset
+        ASSIGN_OR_RETURN(tablet_range,
+                         TabletRangeHelper::create_seek_range_from(_tablet_metadata->range(), _tablet_schema, nullptr));
+    }
+
+    for (int i = 0; i < segments.size(); i++) {
+        auto& seg_ptr = segments[i];
+        seg_options.tablet_range = std::nullopt;
+        if (i < _metadata->shared_segments_size() && _metadata->shared_segments(i)) {
+            seg_options.tablet_range = tablet_range;
+        }
         auto res = seg_ptr->new_iterator(schema, seg_options);
         if (res.status().is_end_of_file()) {
             continue;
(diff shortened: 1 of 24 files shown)


alvin-phoenix-ai:
@cursor review

be/src/storage/lake/rowset.h
```diff
@@ -145,6 +151,8 @@ class Rowset : public BaseRowset {
     // default is 0 means every segment will be used.
     // only used for compaction
     size_t _compaction_segment_limit;
+    // represents the seek range of the rowset, this member is only used for range distribution table
+    std::unique_ptr<SeekRange> _seek_range;
```
alvin-phoenix-ai: This will be a significant change for Rowset.
Before this change, the rowset could be reused for all requests because every reader would have the same range.
But when a rowset has a range, it becomes hard to reuse, especially when you are trying to record the seek range in this class.
If a rowset will be reused for different ranges, how do you differentiate the various ranges for the same rowset? 
I will set a request change before we have an alignment on how you are going to handle it.

alvin-phoenix-ai (dismissed):
Need to know how to reuse the same rowset with different seek ranges.

be/src/storage/lake/rowset.h
srlch: I have remove the `_seek_range` from rowset and rename it to avoid ambiguity. `_seek_range` means `tablet_range` which is a immutable property for a range distribution tablet. `tablet_range` is totally different from the query range. It means that even a rowset is reused by different reader with different query ranges, this rowset must belong to the same tablet so all reader share this rowset must satisfy the same `tablet_range` to meet the tablet definition

be/src/storage/lake/rowset.cpp
```diff
@@ -264,12 +267,19 @@ StatusOr<std::vector<ChunkIteratorPtr>> Rowset::read(const Schema& schema, const
...
     std::vector<SegmentPtr> segments;
     RETURN_IF_ERROR(load_segments(&segments, seg_options, nullptr));
-    for (auto& seg_ptr : segments) {
+    for (int i = 0; i < segments.size(); i++) {
+        auto& seg_ptr = segments[i];
         if (seg_ptr->num_rows() == 0) {
             continue;
         }
+        seg_options.tablet_range = std::nullopt;
+        if (_metadata->has_shared_segments() && _metadata->shared_segments(i)) {
```
xiangguangyxg: need check shard_segments_size() ?

alvin-phoenix-ai:
@cursor review

be/src/storage/lake/rowset.cpp
srlch: Fixed

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-phoenix-ai:
@cursor review

alvin-celerdata:
@cursor review

be/src/storage/lake/rowset.h
```diff
@@ -22,6 +22,10 @@
 #include "storage/options.h"
 #include "storage/rowset/base_rowset.h"
 
+namespace starrocks {
+class SeekRange;
+} // namespace starrocks
```
alvin-phoenix-ai: This file has no SeekRange, so please remove this forward declaration.

be/src/storage/lake/sst_seek_range.h
```diff
@@ -0,0 +1,76 @@
...
+namespace starrocks::lake {
+
+// SstSeekRange represents a basic key range unit for compaction tasks split.
+// The scan range is [seek_key, stop_key). stop_key is exclusive.
+struct SstSeekRange {
+    std::string seek_key;
+    std::string stop_key; // could be empty meaning infinity
+
+    // Return true when [seek_key, stop_key) and [range.start_key(), range.end_key()] overlap.
+    bool has_overlap(const PersistentIndexSstableRangePB& range) const {
```
alvin-phoenix-ai: Please move implementation into CPP file

be/src/storage/lake/sst_seek_range.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include "gen_cpp/lake_types.pb.h"
+#include "storage/sstable/comparator.h"
+#include "util/slice.h"
+
+namespace starrocks::lake {
+
+// SstSeekRange represents a basic key range unit for compaction tasks split.
+// The scan range is [seek_key, stop_key). stop_key is exclusive.
+struct SstSeekRange {
```
alvin-phoenix-ai: Please use class instead, because it is not a simple struct

alvin-phoenix-ai: And also need dedicated UT for this class

be/src/storage/lake/rowset.h
srlch: Fixed

be/src/storage/lake/sst_seek_range.h
srlch: Fixed

be/src/storage/lake/sst_seek_range.h
srlch: Fixed

srlch: Fixed

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CreateTableAnalyzer.java
```diff
@@ -492,9 +492,20 @@ private static void analyzeSortKeys(CreateTableStmt stmt) {
...
+                }
+
                 if (!res) {
-                    throw new SemanticException("The sort columns must be same with key columns");
+                    if (keysType == KeysType.PRIMARY_KEYS) {
+                        throw new SemanticException("The sort columns must be same with primary key columns " +
+                                "and the order must be consistent");
+                    } else {
+                        throw new SemanticException("The sort columns must be same with key columns");
+                    }
```
xiangguangyxg: two `if (keysType == KeysType.PRIMARY_KEYS)` could refactor to one

luohaha (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CreateTableAnalyzer.java
srlch: Fixed

be/src/storage/lake/tablet_range_helper.cpp
```diff
@@ -0,0 +1,200 @@
...
+        values.reserve(n);
+        for (int i = 0; i < n; i++) {
+            const int idx = sort_key_idxes[i];
+            schema.append_sort_key_idx(idx);
+            auto f = std::make_shared<Field>(ChunkHelper::convert_field(idx, tablet_schema->column(idx)));
+            schema.append(std::move(f));
+
+            const auto& v = tuple.values(i);
+            if (!v.has_type()) {
+                return Status::Corruption("Missing type in TabletRangePB bound value");
```
wyb, line 96:
```suggestion
                return Status::Corruption("Missing type in TabletRangePB_VariantPB bound value");
```

wyb (approved):

Seaven (approved):

be/src/storage/lake/tablet_range_helper.cpp
srlch: Fix in next pr

xiangguangyxg (approved):

PR #66748, opened 2025-12-16
[BugFix] fix LocalTabletsChannel and LakeTabletsChannel dead lock

kevincai:
* fix the circular lock wait by moving `remove_tablets_channel()` to LoadChannel, TabletsChannel will not deregister itself from LoadChannel any more.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

---

> [!NOTE]
> Move tablets channel deregistration to LoadChannel on close; update Lake/Local tablets channels and constructors accordingly.
> 
> - **LoadChannel**
>   - Handles channel removal via private `_remove_tablets_channel()` when `add_chunk()` signals `close_channel` and updates profile if needed.
>   - Initializes `close_channel` explicitly and removes public `remove_tablets_channel()` API.
> - **LakeTabletsChannel**
>   - Drops `LoadChannel*` dependency: constructor now `LakeTabletsChannel(tablet_manager, key, mem_tracker, parent_profile)`; member `_load_channel` removed.
>   - Factory `new_lake_tablets_channel(...)` keeps `load_channel` param (unused) and constructs with new signature.
>   - Stops self-deregistering on close (no call to `remove_tablets_channel`).
> - **LocalTabletsChannel**
>   - Removes self-deregistration on close; channel lifecycle now managed by `LoadChannel`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 642c88d65993f9d8aa27ecdc9136af12fe27ee65. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/runtime/lake_tablets_channel.cpp
+++ b/be/src/runtime/lake_tablets_channel.cpp
@@ -60,8 +60,8 @@ class LakeTabletsChannel : public TabletsChannel {
     using TxnLogPtr = AsyncDeltaWriter::TxnLogPtr;
 
 public:
-    LakeTabletsChannel(LoadChannel* load_channel, lake::TabletManager* tablet_manager, const TabletsChannelKey& key,
-                       MemTracker* mem_tracker, RuntimeProfile* parent_profile);
+    LakeTabletsChannel(lake::TabletManager* tablet_manager, const TabletsChannelKey& key, MemTracker* mem_tracker,
+                       RuntimeProfile* parent_profile);
 
     ~LakeTabletsChannel() override;
 
@@ -246,7 +246,6 @@ private:
         }
     };
 
-    LoadChannel* _load_channel;
     lake::TabletManager* _tablet_manager;
 
     TabletsChannelKey _key;
@@ -315,11 +314,9 @@ private:
     std::unique_ptr<RuntimeProfile> _tablets_profile;
 };
 
-LakeTabletsChannel::LakeTabletsChannel(LoadChannel* load_channel, lake::TabletManager* tablet_manager,
-                                       const TabletsChannelKey& key, MemTracker* mem_tracker,
-                                       RuntimeProfile* parent_profile)
+LakeTabletsChannel::LakeTabletsChannel(lake::TabletManager* tablet_manager, const TabletsChannelKey& key,
+                                       MemTracker* mem_tracker, RuntimeProfile* parent_profile)
         : TabletsChannel(),
-          _load_channel(load_channel),
           _tablet_manager(tablet_manager),
           _key(key),
           _mem_tracker(mem_tracker),
@@ -646,7 +643,6 @@ void LakeTabletsChannel::add_chunk(Chunk* chunk, const PTabletWriterAddChunkRequ
     COUNTER_UPDATE(_wait_writer_timer, wait_writer_ns);
 
     if (close_channel) {
-        _load_channel->remove_tablets_channel(_key);
         if (_finish_mode == lake::DeltaWriterFinishMode::kDontWriteTxnLog) {
             _txn_log_collector.notify();
         }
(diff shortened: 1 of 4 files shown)


alvin-phoenix-ai:
@cursor review

be/src/runtime/lake_tablets_channel.cpp
alvin-phoenix-ai:
```suggestion
LakeTabletsChannel::LakeTabletsChannel(lake::TabletManager* tablet_manager,
```
load_channel can be removed from the constructor

trueeyu (dismissed):

be/src/runtime/load_channel.cpp
```diff
@@ -189,12 +189,15 @@ void LoadChannel::_add_chunk(Chunk* chunk, const MonotonicStopWatch* watch, cons
     }
     bool close_channel;
```
alvin-phoenix-ai: Better to have a default value.

be/src/runtime/load_channel.cpp
alvin-phoenix-ai, line 333: Will it cause the same issue if it removes itself from load_mgr?

kevincai: the lock is unlocked before call remove_load_channel(), should be fine.

be/src/runtime/load_channel.cpp
kevincai: ok

be/src/runtime/lake_tablets_channel.cpp
kevincai: ok

kevincai: Fixed the constructor, but leave the parameter there for the creator, with consideration of
1. keep input parameter consistent with LocalTabletsChannel
2. keep code change as minimal as possible in a bugfix.

wyb (dismissed):

alvin-phoenix-ai:
@cursor review

be/src/runtime/load_channel.cpp
```diff
@@ -187,14 +187,17 @@ void LoadChannel::_add_chunk(Chunk* chunk, const MonotonicStopWatch* watch, cons
...
+    bool close_channel = false;
     channel->add_chunk(chunk, request, response, &close_channel);
-    if (close_channel &&
-        (_should_enable_profile() || (watch != nullptr && watch->elapsed_time() > request.timeout_ms() * 1000000))) {
-        // If close_channel is true, the channel has been removed from _tablets_channels
-        // in TabletsChannel::add_chunk, so there will be no chance to get the channel to
-        // update the profile later. So update the profile here
-        channel->update_profile();
+    if (close_channel) {
+        _remove_tablets_channel(key);
```
alvin-phoenix-ai, line 193: in `_remove_tablets_channel`, there is a message "// Do NOT touch |this| since here, it could have been deleted."
And after this function call, `_should_enable_profile()` is still called; it seems dangerous.

be/src/runtime/load_channel.cpp
alvin-phoenix-ai: I still think we should call `_load_mgr->remove_load_channel(_load_id);` elsewhere.

be/src/runtime/load_channel.cpp
kevincai: Will remove the comment.

It is added previously due to the interface is public, the invocation chain is uncertain.

Now the interface is private, the invocation location is inside this LoadChannel which is under control.

The `channel->update_profile()` invocation is safe, because the `channel` is a shared_ptr, even the LoadChannel is invalid, the `channel` is still valid.

Actually the LoadChannel itself is safe now after moving the `_remove_tablets_channel()` to `add_chunk()` because in LoadChannelMgr::add_chunk(), a temp shared_ptr is held to invoke `LoadChannel::add_chunk()`, so inside LoadChannel::add_chunk , the `this` is guaranteed to be valid.

wyb (approved):

alvin-phoenix-ai (approved):

PR #67027, opened 2025-12-19
[BugFix] Change deallocation order in exec_env->destroy to prevent Use-After-Free in MemTracker::~MemTracker

### LINKED ISSUE #67026: SEGSEV in MemTracker::~MemTracker during BE shutdown when query is executed
When a query is executed during a BE shutdown there is a chance to run into a SEGFAULT in `MemTracker::~MemTracker`.

Stacktrace:
```
@     0xfcd0f67553dc (/usr/lib/aarch64-linux-gnu/libc.so.6+0x853db)
    @          0xa156c6c google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @     0xfcd0f73d81a8 os::Linux::chained_handler(int, siginfo_t*, void*)
    @     0xfcd0f73dda68 JVM_handle_linux_signal
    @     0xfcd0f73d125c signalHandler(int, siginfo_t*, void*)
    @     0xfcd0f7ab87a0 ([vdso]+0x79f)
    @          0x595de14 starrocks::MemTracker::~MemTracker()
    @          0x587ad40 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release()
    @          0x5c0b8b8 starrocks::DataStreamRecvr::~DataStreamRecvr()
    @          0x5c08d8c std::_Sp_counted_ptr<starrocks::DataStreamRecvr*, (__gnu_cxx::_Lock_policy)2>::_M_dispose()
    @          0x5c0808c starrocks::DataStreamMgr::~DataStreamMgr()
    @          0x594c718 starrocks::ExecEnv::destroy()
    @          0x8aa9700 starrocks::start_be(std::vector<starrocks::StorePath, std::allocator<starrocks::StorePath> > const&, bool)
    @          0x586d1b8 main
    @     0xfcd0f66f7400 (/usr/lib/aarch64-linux-gnu/libc.so.6+0x273ff)
    @     0xfcd0f66f74d8 __libc_start_main
    @          0x5868030 _start
```

It does not occur everytime. I found that executing the query below around 50-100 and triggering a shutdown makes it likely to run into the SEGFAULT.

#### Steps to reproduce:
Setup 
```sql
-- 1. Create Seed Table (0-9)
CREATE TABLE seed_nums (n INT) PROPERTIES("replication_num" = "1");

INSERT INTO
    seed_nums
VALUES
    (0),
    (1),
    (2),
    (3),
    (4),
    (5),
    (6),
    (7),
    (8),
    (9);

-- 2. Create Users Table
CREATE TABLE dim_users (
    user_id BIGINT,
    user_name VARCHAR(64),
    country_code VARCHAR(10),
    registration_date DATE
) ENGINE = OLAP PRIMARY KEY(user_id) DISTRIBUTED BY HASH(user_id) BUCKETS 10 PROPERTIES ("replication_num" = "1");

-- 3. Create Traffic Table (Partitioned for scale)
CREATE TABLE fact_web_traffic (
    event_id VARCHAR(64),
    user_id BIGINT,
    url_path VARCHAR(255),
    session_duration_ms INT,
    event_timestamp DATETIME
) ENGINE = OLAP DUPLICATE KEY(event_id) -- More buckets for larger data to increase parallelism
DISTRIBUTED BY HASH(user_id) BUCKETS 64 PROPERTIES ("replication_num" = "1");

INSERT INTO
    dim_users
SELECT
    (
        t1.n + t2.n * 10 + t3.n * 100 + t4.n * 1000 + t5.n * 10000 + t6.n * 100000
    ) AS user_id,
    concat(
        'User_',
        cast(
            (
                t1.n + t2.n * 10 + t3.n * 100 + t4.n * 1000 + t5.n * 10000 + t6.n * 100000
            ) as string
        )
    ) AS user_name,
    CASE
        (t1.n % 5)
        WHEN 0 THEN 'US'
        WHEN 1 THEN 'DE'
        WHEN 2 THEN 'CN'
        WHEN 3 THEN 'JP'
        ELSE 'UK'
    END AS country_code,
    date_add('2023-01-01', INTERVAL (t1.n * 10) DAY) AS registration_date
FROM
    seed_nums t1,
    seed_nums t2,
    seed_nums t3,
    seed_nums t4,
    seed_nums t5,
    seed_nums t6;

INSERT INTO
    fact_web_traffic
SELECT
    uuid() AS event_id,
    u.user_id,
    -- Randomize URL slightly
    concat(
        '/page/',
        cast((s1.n * 10 + s2.n) as string),
        '/product'
    ) AS url_path,
    -- Random duration
    cast(100 + (rand() * 1000) as int) AS session_duration_ms,
    -- Random timestamps over the last year
    date_add(
        now(),
        INTERVAL - cast((rand() * 365) as int) DAY
    ) AS event_timestamp
FROM
    dim_users u
    CROSS JOIN seed_nums s1,
    seed_nums s2 -- Multiplying 1M users by 100
WHERE
    u.user_id <= 1000000;
```
Execute a query.
```sql
SELECT
    u.country_code,
    CONCAT(
        substring(t.url_path, 1, 6),
        ' - ',
        u.country_code
    ) as segment_key,
    COUNT(DISTINCT t.user_id) as unique_active_users,
    AVG(t.session_duration_ms) as avg_duration
FROM
    fact_web_traffic t
    JOIN dim_users u ON t.user_id = u.user_id
WHERE
    t.session_duration_ms > 100
GROUP BY
    u.country_code,
    segment_key
ORDER BY
    unique_active_users DESC;
```

Shutdown backend
```
root@e222851cf1f6:/celostar-starrocks# ./be/output/bin/start_be.sh
^CSegmentation fault
```
3.3 but it should also occur for the most recent one.

martinr0x:
### Problem
MemTrackers are organized hierarchically. A typical query hierarchy looks like this:
1. `_query_pool_mem_tracker` (root)
2. `WorkGroup::_mem_tracker` (parent)
3. `QueryContext::_query_mem_tracker` (child)
4. `RuntimeState::_instance_mem_tracker` (grandchild - held by `DataStreamRecvr`)

### The Issue

During BE shutdown while a query is still running, the destruction order in `ExecEnv::destroy` causes a Use-After-Free:
1. `WorkGroupManager` is destroyed first
2. `QueryContextManager` is deleted next. This is safe because it holds a shared_ptr ref to WorkGroup. WorkGroup along its MemTrackers can get deallocated now.
3. When `DataStreamMgr` is destroyed, its `_query_mem_tracker` destructor attempts to release the remaining memory quota to its parent.
4. Since the parent (`WorkGroup` tracker) was already freed in step 2, accessing the raw pointer results in a **SEGFAULT**.

## What I'm doing:
I have modified the deallocation order in `ExecEnv::destroy`.
I moved the destruction of `WorkGroupManager` to occur **after** `QueryContextManager` and `DataStreamMgr`. This ensures that the parent `WorkGroup` tracker remains valid until all child query trackers have been safely destroyed.

Fixes #67026

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Prevents potential use-after-free during BE shutdown by ensuring child query trackers are destroyed before their parent workgroup trackers.
> 
> - Reorders teardown in `ExecEnv::destroy`: delete `DataStreamMgr` and `QueryContextManager` before calling `WorkGroupManager::destroy()`
> - Adds DEBUG-only guard in `MemTracker::~MemTracker` to assert `_child_trackers` is empty (with locking)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8dc45cc8e6158f9edb72b12402f0af9c0c0aafc9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/runtime/mem_tracker.cpp
+++ b/be/src/runtime/mem_tracker.cpp
@@ -203,7 +203,8 @@ MemTracker::~MemTracker() {
 #ifdef DEBUG
     {
         std::unique_lock<std::mutex> lock(_child_trackers_lock);
-        DCHECK(_child_trackers.size() == 0) << err_msg("Child mem trackers have not been released, may cause corruption" );
+        DCHECK(_child_trackers.size() == 0)
+                << err_msg("Child mem trackers have not been released, may cause corruption");
     }
 #endif
 
--- a/be/src/runtime/mem_tracker.cpp
+++ b/be/src/runtime/mem_tracker.cpp
@@ -203,8 +203,7 @@ MemTracker::~MemTracker() {
 #ifdef DEBUG
     {
         std::unique_lock<std::mutex> lock(_child_trackers_lock);
-        DCHECK(_child_trackers.empty())
-                << err_msg("Child mem trackers have not been released, may cause corruption");
+        DCHECK(_child_trackers.empty()) << err_msg("Child mem trackers have not been released, may cause corruption");
     }
 #endif
 
--- a/be/src/runtime/mem_tracker.cpp
+++ b/be/src/runtime/mem_tracker.cpp
@@ -199,6 +199,14 @@ MemTracker::~MemTracker() {
     if (parent()) {
         unregister_from_parent();
     }
+
+#ifdef DEBUG
+    {
+        std::unique_lock<std::mutex> lock(_child_trackers_lock);
+        DCHECK(_child_trackers.size() == 0) << err_msg("Child mem trackers have not been released, may cause corruption" );
+    }
+#endif
+
     // When the mem_tracker is destroyed, manually setting _consumption to null can easily
     // trigger a use-after-free bug in MemTracker.
     _consumption = nullptr;
--- a/be/src/runtime/mem_tracker.cpp
+++ b/be/src/runtime/mem_tracker.cpp
@@ -203,7 +203,7 @@ MemTracker::~MemTracker() {
 #ifdef DEBUG
     {
         std::unique_lock<std::mutex> lock(_child_trackers_lock);
-        DCHECK(_child_trackers.size() == 0)
+        DCHECK(_child_trackers.empty())
                 << err_msg("Child mem trackers have not been released, may cause corruption");
     }
 #endif
--- a/be/src/runtime/exec_env.cpp
+++ b/be/src/runtime/exec_env.cpp
@@ -864,6 +864,8 @@ void ExecEnv::destroy() {
     SAFE_DELETE(_query_rpc_pool);
     SAFE_DELETE(_datacache_rpc_pool);
     _load_rpc_pool.reset();
+    SAFE_DELETE(_stream_mgr);
+    SAFE_DELETE(_query_context_mgr);
     _workgroup_manager->destroy();
     _workgroup_manager.reset();
     SAFE_DELETE(_thread_pool);
@@ -874,7 +876,6 @@ void ExecEnv::destroy() {
         _lake_tablet_manager->prune_metacache();
     }
 
-    SAFE_DELETE(_query_context_mgr);
     // WorkGroupManager should release MemTracker of WorkGroups belongs to itself before deallocate
     // _query_pool_mem_tracker.
     SAFE_DELETE(_runtime_filter_cache);
@@ -893,7 +894,6 @@ void ExecEnv::destroy() {
     SAFE_DELETE(_backend_client_cache);
     SAFE_DELETE(_result_queue_mgr);
     SAFE_DELETE(_result_mgr);
-    SAFE_DELETE(_stream_mgr);
     SAFE_DELETE(_lookup_dispatcher_mgr);
     SAFE_DELETE(_batch_write_mgr);
     SAFE_DELETE(_external_scan_context_mgr);


be/src/runtime/mem_tracker.cpp
```diff
@@ -199,6 +199,15 @@ MemTracker::~MemTracker() {
     if (parent()) {
         unregister_from_parent();
     }
+
+#ifdef DEBUG
+    {
+        std::unique_lock<std::mutex> lock(_child_trackers_lock);
+        DCHECK(_child_trackers.size() == 0)
```
alvin-phoenix-ai:
```suggestion
        DCHECK(_child_trackers.empty())
```
`empty()` may be more efficient.

alvin-phoenix-ai (dismissed):
LGTM

alvin-phoenix-ai:
@cursor review

be/src/runtime/mem_tracker.cpp
```diff
@@ -199,6 +199,15 @@ MemTracker::~MemTracker() {
     if (parent()) {
         unregister_from_parent();
     }
+
+#ifdef DEBUG
+    {
+        std::unique_lock<std::mutex> lock(_child_trackers_lock);
+        DCHECK(_child_trackers.empty())
+                << err_msg("Child mem trackers have not been released, may cause corruption");
```
alvin-phoenix-ai:
```suggestion
        DCHECK(_child_trackers.empty()) << err_msg("Child mem trackers have not been released, may cause corruption");
```

alvin-phoenix-ai:
@mergify rebase

alvin-phoenix-ai (approved):

trueeyu (approved):

be/src/runtime/mem_tracker.cpp
```diff
@@ -199,6 +199,14 @@ MemTracker::~MemTracker() {
     if (parent()) {
         unregister_from_parent();
     }
+
+#ifdef DEBUG
```
kevincai, line 203: shall use `#ifndef NDEBUG` instead of using `#ifdef DEBUG`

martinr0x: @kevincai That would cause the lock to be obtained in release builds, no?

kevincai: `NDEBUG` is more widely used than `DEBUG` macro.  

And it is `#ifndef NDEBUG` which means "no definition of Non-DEBUG", equivalent to current `#ifdef DEBUG`.

martinr0x: Ah apologies, I read `DEBUG`, not `NDEBUG`.

PR #67151, opened 2025-12-23
[BugFix] Fix NPE in Iceberg metadata table query due to missing Configuration propagation

j1wonpark:
## Why I'm doing:

When querying Iceberg metadata tables (history, snapshots, refs, etc.),
NPE occurs with message: "Cannot invoke Configuration.size() because conf is null".

This happens because:
1. `IcebergCachingFileIO.getWrappedIO()` returns `ResolvingFileIO` without Configuration set
2. When `SerializableTable` serializes the FileIO, it calls `serializeConfWith()`
3. `SerializableConfSupplier` constructor calls `conf.size()` on null Configuration
  
## What I'm doing:

Fix `getWrappedIO()` to propagate Configuration to wrappedIO before returning.
- If `conf` is set via `setConf()`, use it
- If `conf` is null (e.g., REST catalog), use default `new Configuration()`

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Fixes missing Hadoop `Configuration` propagation to the underlying `ResolvingFileIO`, preventing NPE during metadata table operations.
> 
> - In `IcebergCachingFileIO.initialize`, if `conf` is already set, forward it to `wrappedIO`
> - In `setConf`, also forward the config to `wrappedIO` when it exists
> - Adds `testWrappedIOConfigurationPropagation` to verify `HadoopConfigurable.serializeConfWith` works and no NPE occurs
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bb2649854f40a618a9c616ba02e4af3956db90f9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
@@ -196,10 +196,8 @@ public class IcebergCachingFileIO implements FileIO, HadoopConfigurable {
     }
 
     public FileIO getWrappedIO() {
-        if (wrappedIO instanceof HadoopConfigurable) {
-            Configuration configuration = conf != null ? conf.get() : new Configuration();
-            ((HadoopConfigurable) wrappedIO).setConf(configuration);
-        }
+        Configuration configuration = conf != null ? conf.get() : new Configuration();
+        wrappedIO.setConf(configuration);
         return wrappedIO;
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
@@ -126,6 +126,10 @@ public class IcebergCachingFileIO implements FileIO, HadoopConfigurable {
         wrappedIO = new ResolvingFileIO();
         wrappedIO.initialize(properties);
 
+        if (conf != null) {
+            wrappedIO.setConf(conf.get());
+        }
+
         if (ENABLE_DISK_CACHE) {
             this.fileContentCache = TwoLevelCacheHolder.INSTANCE;
         } else {
@@ -155,6 +159,9 @@ public class IcebergCachingFileIO implements FileIO, HadoopConfigurable {
     @Override
     public void setConf(Configuration conf) {
         this.conf = new SerializableConfiguration(conf)::get;
+        if (wrappedIO != null) {
+            wrappedIO.setConf(conf);
+        }
     }
 
     @Override
@@ -196,8 +203,6 @@ public class IcebergCachingFileIO implements FileIO, HadoopConfigurable {
     }
 
     public FileIO getWrappedIO() {
-        Configuration configuration = conf != null ? conf.get() : new Configuration();
-        wrappedIO.setConf(configuration);
         return wrappedIO;
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
@@ -196,6 +196,10 @@ public class IcebergCachingFileIO implements FileIO, HadoopConfigurable {
     }
 
     public FileIO getWrappedIO() {
+        if (wrappedIO instanceof HadoopConfigurable) {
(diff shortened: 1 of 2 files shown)


alvin-phoenix-ai:
@cursor review

j1wonpark:
Hi @dirtysalt @Youngwb 

Could you please review this PR when you have a chance? 

All CI checks have passed with 100% coverage on the changed lines.

Thanks!

alvin-phoenix-ai:
@mergify rebase

alvin-phoenix-ai:
@cursor review

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
```diff
@@ -196,6 +196,8 @@ public void deleteFile(String path) {
     }
 
     public FileIO getWrappedIO() {
+        Configuration configuration = conf != null ? conf.get() : new Configuration();
```
Youngwb: you can use `wrappedIO.setConf(buildConfFromProperties(properties, path));`  I'm curious about why newInputFile wasn't used before getConf. Here, the conf is set, and before reading the table metadata, newInputFile should be used.

j1wonpark: Thanks for the review!

Regarding your question about why newInputFile wasn't used before getConf:

This bug occurs in a specific code path where `getWrappedIO()` is called directly without going through `newInputFile()`.


In IcebergMetadata.getSerializedMetaSpec():

```
FileIO fileIO = nativeTable.io();
if (fileIO instanceof IcebergCachingFileIO) {
    fileIO = ((IcebergCachingFileIO) fileIO).getWrappedIO();  // called directly
}
String serializedTable = SerializationUtil.serializeToBase64(new SerializableTable(nativeTable, fileIO));
```

When querying Iceberg metadata tables, this path is taken and newInputFile() is never called beforehand. During serialization of SerializableTable, getConf() is invoked on the ResolvingFileIO, which causes NPE if Configuration is not set.

As for using `buildConfFromProperties(properties, path)` - this method requires a path parameter (needed for Azure Blob SAS token handling), but `getWrappedIO()` doesn't have access to a specific path. So I used the base configuration `(conf.get())` or a default `new Configuration()` as a fallback.

Youngwb: You can debug the process of querying Iceberg. When parsing an Iceberg table, it uses newInputFileto read remote metadata files. Querying Iceberg metadata tables also follows this process, so newInputFile will indeed be invoked; otherwise, Iceberg metadata tables using vending credentials would also encounter issues.

j1wonpark: Thanks for the insightful feedback! After digging deeper, I found the real issue.

You're right that `newInputFile()` sets the conf to `wrappedIO`. But here's the thing - in the metadata table query path, `newInputFile()` is never called.

Looking at the stack trace:

```
NPE: Cannot invoke "Configuration.size()" because "conf" is null
    at SerializableTable$SerializableConfSupplier.<init>
    at ResolvingFileIO.serializeConfWith
    at SerializableTable.fileIO
    at SerializableTable.<init>
    at IcebergMetadata.getSerializedMetaSpec  ← starts here
```

This path is for queries like `SELECT * FROM iceberg_table$files`. Here we're just serializing the table to send to BE - no actual file reading happens at this point. So `newInputFile()` never gets a chance to set the conf.

The flow is:
```
getSerializedMetaSpec()
  → getWrappedIO()        // get ResolvingFileIO
  → new SerializableTable // serialize for BE
    → serializeConfWith() // conf is null!
```

Also, I found that Iceberg's `CatalogUtil.loadFileIO()` calls `setConf()` **before** `initialize()`. So when `setConf()` is called, `wrappedIO` doesn't exist yet.

The fix handles both cases:
- `setConf()`: store conf, propagate to wrappedIO if it exists
- `initialize()`: after creating wrappedIO, apply the stored conf

This way, `wrappedIO.conf` is always properly set regardless of call order or code path.

Youngwb (approved):

dirtysalt (approved):

PR #67207, opened 2025-12-24
[Feature] Add UTF-8 support for ngram_search functions

eshishki:
This patch adds proper UTF-8 character-based n-gram computation for
  ngram_search and ngram_search_case_insensitive functions.

  Previously, n-grams were computed byte-by-byte, which produced incorrect
  results for non-ASCII text (Cyrillic, Chinese, etc.). Now n-grams are
  computed based on UTF-8 characters using the existing UTF8_BYTE_LENGTH_TABLE.

  Changes:
  - Add session variable `ngram_search_support_utf8` (default: false)
  - Add utf8_tolower() utility function using ICU for proper Unicode case folding
  - Fix ngram_search to iterate by UTF-8 characters when enabled
  - Fix bloom filter index writer to use UTF-8 case folding
  - Fix bloom filter query to use UTF-8 case folding

  Note: Bloom filter index already used UTF-8 for n-gram extraction,
  but case-insensitive mode used ASCII tolower. This is now fixed.

## What I'm doing:

Fixes #67208

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Adds optional UTF-8 character–based n-gram processing and Unicode case folding to string similarity and bloom-filter flows.
> 
> - **UTF-8 mode (gated by `ngram_search_support_utf8`)** for `ngram_search` and `ngram_search_case_insensitive`; FE `SessionVariable`, Thrift `TQueryOptions`, and BE `RuntimeState` wiring included
> - **Unicode lowercase via `utf8_tolower` (ICU)**; added `util/utf8.cpp` and integrated into n-gram processing and bloom filter index writer/query paths, with ASCII fast-path preserved
> - **N-gram logic updated** to iterate by UTF-8 characters when enabled; case-insensitive handling adjusted; runtime dispatch via `Ngramstate::use_utf8`
> - **Bloom filter**: LIKE/ngram paths now validate UTF-8 and apply proper case folding for non-ASCII
> - **Tests**: new SQL cases for Cyrillic/Chinese; build updated to compile new UTF-8 utility
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 894c3f2ab94af96cd15bb037a34c06ab6642ade0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/storage/rowset/bloom_filter_index_writer.cpp
+++ b/be/src/storage/rowset/bloom_filter_index_writer.cpp
@@ -217,6 +217,9 @@ public:
         for (int i = 0; i < count; ++i) {
             std::vector<size_t> index;
             size_t slice_gram_num = get_utf8_index(*cur_slice, &index);
+            // slice_gram_num can be used to judge if the cur_slice is an ASCII string or not.
+            // If slice_gram_num == cur_slice->get_size(), then it is.
+            bool is_ascii = (slice_gram_num == cur_slice->get_size());
 
             size_t j;
             for (j = 0; j + gram_num <= slice_gram_num; j++) {
@@ -230,9 +233,14 @@ public:
                     if (this->_bf_options.case_sensitive) {
                         _values.insert(get_value<field_type>(&cur_ngram, this->_typeinfo, &this->_type_info_allocator));
                     } else {
-                        // todo::exist two copy of ngram, need to optimize
                         std::string lower_ngram;
-                        Slice lower_ngram_slice = cur_ngram.tolower(lower_ngram);
+                        Slice lower_ngram_slice;
+                        if (is_ascii) {
+                            lower_ngram_slice = cur_ngram.tolower(lower_ngram);
+                        } else {
+                            utf8_tolower(cur_ngram.get_data(), cur_ngram.get_size(), lower_ngram);
+                            lower_ngram_slice = Slice(lower_ngram.data(), lower_ngram.size());
+                        }
                         _values.insert(get_value<field_type>(&lower_ngram_slice, this->_typeinfo,
                                                              &this->_type_info_allocator));
                     }
--- a/be/src/storage/rowset/bloom_filter_index_writer.cpp
+++ b/be/src/storage/rowset/bloom_filter_index_writer.cpp
@@ -215,35 +215,33 @@ public:
         size_t gram_num = this->_bf_options.gram_num;
         const auto* cur_slice = reinterpret_cast<const Slice*>(values);
         for (int i = 0; i < count; ++i) {
-            std::vector<size_t> index;
-            size_t slice_gram_num = get_utf8_index(*cur_slice, &index);
-            // slice_gram_num can be used to judge if the cur_slice is an ASCII string or not.
-            // If slice_gram_num == cur_slice->get_size(), then it is.
-            bool is_ascii = (slice_gram_num == cur_slice->get_size());
+            // For a case-insensitive index, lowercase the whole value once and build ngrams from the
+            // folded copy. The reader lowercases the whole needle before splitting it, so the writer
+            // must split in the same order: folding each ngram after slicing would disagree with the
+            // reader for context-dependent or length-changing case mappings (e.g. Turkish 'İ').
+            Slice value = *cur_slice;
+            std::string lower_buf;
+            if (!this->_bf_options.case_sensitive) {
+                if (validate_ascii_fast(value.get_data(), value.get_size())) {
+                    value.tolower(lower_buf);
+                } else {
+                    utf8_tolower(value.get_data(), value.get_size(), lower_buf);
+                }
+                value = Slice(lower_buf.data(), lower_buf.size());
+            }
 
-            size_t j;
-            for (j = 0; j + gram_num <= slice_gram_num; j++) {
+            std::vector<size_t> index;
+            size_t slice_gram_num = get_utf8_index(value, &index);
+
+            for (size_t j = 0; j + gram_num <= slice_gram_num; j++) {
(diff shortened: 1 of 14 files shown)


alvin-phoenix-ai:
@cursor review

be/src/util/utf8.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <algorithm>
+#include <cctype>
+
+#include "common/logging.h"
+
+namespace starrocks {
+
+void utf8_tolower(const char* src, size_t src_len, std::string& dst) {
+    UErrorCode err_code = U_ZERO_ERROR;
+    UCaseMap* case_map = ucasemap_open("", U_FOLD_CASE_DEFAULT, &err_code);
```
satanson: Is this UCaseMap real map like std::map?
the map should be readonly, so it can be loaded only once before using it.

be/src/storage/rowset/bloom_filter_index_writer.cpp
```diff
@@ -226,9 +226,11 @@ class NgramBloomFilterIndexWriterImpl<field_type, std::enable_if_t<is_slice_type
                     if (this->_bf_options.case_sensitive) {
                         _values.insert(get_value<field_type>(&cur_ngram, this->_typeinfo, &this->_pool));
                     } else {
-                        // todo::exist two copy of ngram, need to optimize
+                        // TODO: exist two copies of ngram, need to optimize
+                        // Use UTF-8 aware tolower for proper Unicode case folding
                         std::string lower_ngram;
-                        Slice lower_ngram_slice = cur_ngram.tolower(lower_ngram);
+                        utf8_tolower(cur_ngram.get_data(), cur_ngram.get_size(), lower_ngram);
```
satanson: utf8_tolower costs more time than its ASCII's version, so we should check the string if it is an ASCII string or not, if so, we use ASCII to_lower instead.

slice_gram_num from LN#215 can used to judge if the cur_slice is an ASCII string.  if slice_gram_num == cur_slice.size, then it is.

alvin-phoenix-ai:
@cursor review

alvin-celerdata:
@cursor review

alvin-celerdata:
@cursor review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/rowset/bloom_filter_index_writer.cpp
```diff
@@ -230,9 +233,14 @@ class NgramBloomFilterIndexWriterImpl<field_type, std::enable_if_t<is_slice_type
...
                     } else {
-                        // todo::exist two copy of ngram, need to optimize
                         std::string lower_ngram;
-                        Slice lower_ngram_slice = cur_ngram.tolower(lower_ngram);
+                        Slice lower_ngram_slice;
+                        if (is_ascii) {
+                            lower_ngram_slice = cur_ngram.tolower(lower_ngram);
+                        } else {
+                            utf8_tolower(cur_ngram.get_data(), cur_ngram.get_size(), lower_ngram);
+                            lower_ngram_slice = Slice(lower_ngram.data(), lower_ngram.size());
```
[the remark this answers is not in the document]
eshishki: Thanks — fixed. The writer now lowercases the whole value once (same predicate the reader uses) and builds the n-grams from the folded copy, so the writer and reader stay byte-identical even for length-changing / context-dependent mappings. Added a regression test (`test_ngram_utf8_case_insensitive_length_changing_fold`) covering the U+0130 → `i` + U+0307 case.

eshishki:
@stdpain @satanson — rebased this onto the latest `main`. Could you take another look when you get a chance?

Quick recap of where it stands:

1. **Ngram bloom filter index (writer + probe) now supports UTF-8 case-insensitive build and lookup** — n-grams are formed over whole UTF-8 characters and lowercased with `utf8_tolower`, and the writer and probe use the same fold-the-whole-value-then-split order so stored and probed n-grams stay byte-identical.

2. **The `ngram_search_support_utf8` gate is wired only to the `ngram_search` / `ngram_search_case_insensitive` functions** (default off), so ASCII-only workloads don't pay the per-row UTF-8 cost — same rationale as the existing `lower_upper_support_utf8` flag. The index path itself is always UTF-8-aware: the writer runs at build time (no session flag to read) and the probe only folds a single constant needle, so there's no ASCII hot loop worth gating there.

3. **Docs are not updated yet.** Compatibility note: the writer's case-folding changed, so an existing `case_sensitive=false` ngram bloom filter index built on non-ASCII data would need a rebuild for lookups to match (the new probe folds the needle, while old segments still hold the un-folded n-grams). In practice this should affect ~nobody — case-insensitive matching never actually worked on non-ASCII before: the old ASCII-only fold left non-ASCII bytes untouched, so such an index silently behaved case-sensitively and wasn't doing what its name implies anyway.

stdpain (dismissed):

stdpain:
@copilot resolve the merge conflicts in this pull request

stdpain:
@eshishki Could u fix conflicts again?

eshishki:
@stdpain rebased

stdpain (approved):

trueeyu (approved):

PR #67321, opened 2025-12-29
[BugFix] use null-safe equals for Iceberg equality delete columns

### LINKED ISSUE #67326: [Bug] iceberg eq delete null handling
Problem

  When reading Iceberg tables with equality delete files containing NULL values in identity columns, StarRocks incorrectly fails to delete matching rows.

  The IcebergEqualityDeleteRewriteRule transforms equality deletes into a LEFT ANTI JOIN between data files and delete files. The join predicate uses BinaryType.EQ (regular equals) for comparing identity columns:

  BinaryType binaryType = icebergIdentifierColumnName.equals(DATA_SEQUENCE_NUMBER)
      ? BinaryType.LT
      : BinaryType.EQ;  // problem here

  In SQL semantics, NULL = NULL evaluates to UNKNOWN, which is treated as FALSE in join predicates. This means rows with NULL values in identity columns are never matched and therefore never deleted.

  Iceberg Specification

  According to the https://iceberg.apache.org/spec/#equality-delete-files:

  Each row of the delete file produces one equality predicate that matches any row where the delete columns are equal. Multiple columns can be thought of as an AND of equality predicates. A null value in a delete column matches a row if the row's value is null, equivalent to col IS NULL.

  The spec explicitly states that NULL values in delete columns should match NULL values in data rows, similar to how PostgreSQL handles REPLICA IDENTITY FULL in logical replication where tuples_equal() treats NULL == NULL as equal.

  Example

  Given a table with data:
```
  id | category | name
  ---|----------|--------
  1  | NULL     | Alice
  2  | NULL     | Bob
  3  | toys     | Charlie
```
  And an equality delete file with equality_ids=[id, category]:
```  
id | category
  ---|----------
  1  | NULL
```
  Expected behavior: Row (1, NULL, Alice) should be deleted.

  Actual behavior (before fix): Row is NOT deleted because NULL = NULL → UNKNOWN → no match in LEFT ANTI JOIN.

  Solution

  Replace BinaryType.EQ with BinaryType.EQ_FOR_NULL (null-safe equals, equivalent to <=> operator) for identity column comparisons:

  BinaryType binaryType = icebergIdentifierColumnName.equals(DATA_SEQUENCE_NUMBER)
      ? BinaryType.LT
      : BinaryType.EQ_FOR_NULL;  // NULL <=> NULL evaluates to TRUE

  This ensures that:
  - NULL <=> NULL → TRUE (rows match, deletion applied)
  - NULL <=> value → FALSE (rows don't match)
  - value <=> value → TRUE/FALSE based on actual comparison

eshishki:
## Problem

When reading Iceberg tables with equality delete files containing NULL values in identity columns, StarRocks incorrectly fails to delete matching rows.

`IcebergEqualityDeleteRewriteRule` transforms equality deletes into a LEFT ANTI JOIN between data files and delete files. The join predicate on identity columns used `BinaryType.EQ` (plain `=`):

```java
BinaryType binaryType = icebergIdentifierColumnName.equals(DATA_SEQUENCE_NUMBER)
    ? BinaryType.LT
    : BinaryType.EQ;  // problem here
```

In SQL semantics, `NULL = NULL` evaluates to UNKNOWN, treated as FALSE in join predicates. Rows with NULL values in identity columns are never matched and therefore never deleted.

## Iceberg specification

[Iceberg spec — equality delete files](https://iceberg.apache.org/spec/#equality-delete-files):

> A null value in a delete column matches a row if the row's value is null, equivalent to `col IS NULL`.

## Example

Data:
```
id | category | name
---|----------|--------
1  | NULL     | Alice
2  | NULL     | Bob
3  | toys     | Charlie
```
Equality delete file (`equality_ids=[id, category]`):
```
id | category
---|----------
1  | NULL
```
Expected: row `(1, NULL, Alice)` is deleted. Actual before fix: not deleted, because `NULL = NULL → UNKNOWN`.

## Solution

Switch identity-column comparison to `BinaryType.EQ_FOR_NULL` (null-safe equals, equivalent to `<=>`); `$data_sequence_number` bound stays `BinaryType.LT`.

- `NULL <=> NULL → TRUE` (rows match, deletion applied)
- `NULL <=> value → FALSE` (rows don't match)
- `value <=> value → TRUE/FALSE` (regular comparison)

## Tests

- `IcebergEqualityDeleteRewriteRuleTest` — plan test asserts the rewritten anti-join uses `EQ_FOR_NULL` for identity columns and keeps `LT` for `$data_sequence_number`.
- `test_iceberg_equality_delete_null_key` — SQL fixture over a v2 hadoop catalog whose equality-delete file is keyed on NULL; the NULL-id rows must be deleted (6 visible rows out of 10).

Caveat on the SQL fixture: it's a local `file://` hadoop catalog copied into `/tmp`, so multi-node CI fails with `Unknown database` (the `shell: cp` lands on the runner, FE/BE read from their own node). The fixture works end-to-end on a single-node deployment, where the behavior was validated. Hosting it in the CI HMS+OSS external catalog (alongside the existing `iceberg_oss_db.eq_del_tbl` cases) would make it green in CI; that needs access to the test catalog.

## Related

The float/double `<=>` NaN-canonicalization fix, originally bundled into this PR, has been split out per review feedback into a separate PR (#74000). It's required for an Iceberg equality-delete file keyed on a NaN identity column but applies to any null-safe equal on float/double — separate review scope (hash-join key constructors, ONE_KEY fast-path routing, scalar `<=>` interpreter and JIT).

Fixes #67326

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/IcebergEqualityDeleteRewriteRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/IcebergEqualityDeleteRewriteRule.java
@@ -236,7 +236,11 @@ public class IcebergEqualityDeleteRewriteRule extends TransformationRule {
                 throw new StarRocksConnectorException("can not find iceberg identifier column {}",
                         rightColRef.getName());
             }
-            BinaryType binaryType = icebergIdentifierColumnName.equals(DATA_SEQUENCE_NUMBER) ? BinaryType.LT : BinaryType.EQ;
+            // Use EQ_FOR_NULL (null-safe equals) for identity columns to match Iceberg spec:
+            // "A null value in a delete column matches a row if the row's value is null"
+            BinaryType binaryType = icebergIdentifierColumnName.equals(DATA_SEQUENCE_NUMBER)
+                    ? BinaryType.LT
+                    : BinaryType.EQ_FOR_NULL;
             BinaryPredicateOperator binaryPredicateOperator = new BinaryPredicateOperator(binaryType,
                     List.of(leftColRef, rightColRef));
             onPredicates.add(binaryPredicateOperator);
--- a/fe/fe-core/src/test/java/com/starrocks/sql/optimizer/rule/transformation/IcebergEqualityDeleteRewriteRuleTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/sql/optimizer/rule/transformation/IcebergEqualityDeleteRewriteRuleTest.java
@@ -1,0 +1,73 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.sql.optimizer.rule.transformation;
+
+import com.starrocks.common.jmockit.Deencapsulation;
+import com.starrocks.sql.ast.expression.BinaryType;
+import com.starrocks.sql.optimizer.operator.scalar.BinaryPredicateOperator;
+import com.starrocks.sql.optimizer.operator.scalar.ColumnRefOperator;
+import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
+import com.starrocks.type.IntegerType;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.starrocks.catalog.IcebergTable.DATA_SEQUENCE_NUMBER;
+
+public class IcebergEqualityDeleteRewriteRuleTest {
+
+    // The LEFT ANTI JOIN that applies equality deletes must compare identity columns with null-safe
+    // equals (EQ_FOR_NULL / <=>), not plain EQ: per the Iceberg spec a NULL value in a delete column
+    // matches a NULL row value, and plain EQ would evaluate NULL = NULL -> UNKNOWN and silently keep
+    // the row alive. The $data_sequence_number bound must stay a strict LT.
+    @Test
+    public void testBuildOnPredicateUsesNullSafeEqualsForIdentityColumns() {
+        ColumnRefOperator leftId = new ColumnRefOperator(1, IntegerType.INT, "id", true);
+        ColumnRefOperator leftSeq = new ColumnRefOperator(2, IntegerType.BIGINT, DATA_SEQUENCE_NUMBER, true);
+        ColumnRefOperator rightId = new ColumnRefOperator(3, IntegerType.INT, "id", true);
+        ColumnRefOperator rightSeq = new ColumnRefOperator(4, IntegerType.BIGINT, DATA_SEQUENCE_NUMBER, true);
+
+        Map<String, ColumnRefOperator> leftCols = new HashMap<>();
+        leftCols.put("id", leftId);
+        leftCols.put(DATA_SEQUENCE_NUMBER, leftSeq);
+
+        IcebergEqualityDeleteRewriteRule rule = new IcebergEqualityDeleteRewriteRule();
+        ScalarOperator onPredicate = Deencapsulation.invoke(
+                rule, "buildOnPredicate", leftCols, List.of(rightId, rightSeq));
+
+        Map<String, BinaryType> byColumn = new HashMap<>();
+        collectBinaryTypes(onPredicate, byColumn);
+
+        Assertions.assertEquals(BinaryType.EQ_FOR_NULL, byColumn.get("id"),
+                "identity column must use null-safe equals so NULL-key rows are deleted");
+        Assertions.assertEquals(BinaryType.LT, byColumn.get(DATA_SEQUENCE_NUMBER),
+                "data sequence number bound must remain a strict less-than");
+    }
+
+    // Walk the AND tree and index each leaf comparison by the name of its right-hand (delete-table) column.
+    private void collectBinaryTypes(ScalarOperator op, Map<String, BinaryType> out) {
+        if (op instanceof BinaryPredicateOperator binary) {
+            ColumnRefOperator rightCol = (ColumnRefOperator) binary.getChild(1);
+            out.put(rightCol.getName(), binary.getBinaryType());
+            return;
+        }
+        for (ScalarOperator child : op.getChildren()) {
+            collectBinaryTypes(child, out);
+        }
+    }
+}
--- a/test/sql/test_iceberg/R/test_iceberg_equality_delete_null_key
+++ b/test/sql/test_iceberg/R/test_iceberg_equality_delete_null_key
@@ -1,0 +1,31 @@
+-- name: test_iceberg_equality_delete_null_key
+create external catalog iceberg_eqnull_${uuid0} PROPERTIES ("type"="iceberg", "iceberg.catalog.type"="hive", "iceberg.catalog.hive.metastore.uris"="${iceberg_catalog_hive_metastore_uris}", "aws.s3.access_key"="${oss_ak}", "aws.s3.secret_key"="${oss_sk}", "aws.s3.endpoint"="${oss_endpoint}");
+-- result:
+-- !result
+create database iceberg_eqnull_${uuid0}.eq_delete_db_${uuid0};
+-- result:
+-- !result
+call iceberg_eqnull_${uuid0}.system.register_table("eq_delete_db_${uuid0}", "eq_del_convert", "oss://${oss_bucket}/iceberg_ci_db/eq_del_convert/metadata/v4.metadata.json");
+-- result:
+-- !result
+set new_planner_optimize_timeout = 30000;
+-- result:
+-- !result
+select * from iceberg_eqnull_${uuid0}.eq_delete_db_${uuid0}.eq_del_convert order by id;
+-- result:
+1	n1
+2	n2
+4	n4
+5	n5
+7	n7
+8	n8
+-- !result
+drop table iceberg_eqnull_${uuid0}.eq_delete_db_${uuid0}.eq_del_convert;
+-- result:
+-- !result
+drop database iceberg_eqnull_${uuid0}.eq_delete_db_${uuid0};
+-- result:
+-- !result
+drop catalog iceberg_eqnull_${uuid0};
+-- result:
+-- !result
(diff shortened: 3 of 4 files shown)


alvin-celerdata:
@cursor review

eshishki:
Addressed the review feedback in the latest push:

**Fixed compile errors flagged by the Copilot review:**
- `RunTimeTypeLimits<T>::logical_type` doesn't exist — replaced with a compile-time `LogicalType` selected via `std::is_same_v<T, float>`.
- `reinterpret_cast` through a local pointer is UB in a `constexpr` context and didn't compile — switched to `std::bit_cast` (we're on `-std=gnu++20`).
- Deduplicated the nullable / non-nullable branches so `canonical_nan` is computed once.

**Added BE unit tests:** new `JoinHashMapTest.NormalizeNanInColumn` in `be/test/exec/join_hash_map_test.cpp` covering float/double with and without NaN, nullable column null-mask preservation, empty column short-circuit, and the `normalize_float_nan` type dispatcher. The key invariant — that all NaN entries collapse to identical canonical bits (`0x7FC00000` / `0x7FF8000000000000`) — is verified via `std::bit_cast`, since that's what null-safe joins actually rely on for `NaN <=> NaN`.

**What I didn't add and why:**
- **FE rule unit test** for `IcebergEqualityDeleteRewriteRule` — there are no existing tests for this rule in the repo, and writing one requires setting up `OptExpression` / `MultiJoinNode` mocks from scratch. Happy to do this if it's a hard requirement, but it's a separate piece of work.
- **End-to-end SQL tests** for NULL / NaN in equality delete files — the existing `test/sql/test_iceberg/T/test_iceberg_eq_delete_table` runs against an external HMS+OSS-backed Iceberg catalog (`iceberg_oss_db.eq_del_tbl`). Adding NULL/NaN cases would require new fixture tables (`eq_del_tbl_null`, `eq_del_tbl_nan`) in that external catalog, which is outside the repo. If someone with access to the test catalog can stage those tables, I can add the SQL test file.

eshishki:
CI failure here is unrelated — only `SQL-Tester` fails on `sql/test_materialized_view_rewrite/R/test_mv_rewrite_bugfix2.sql` (materialized-view rewrite chose 1 MV instead of 2). CI itself flags it as `UNSTABLE SQL TEST`. This PR touches only the equality-delete LEFT-ANTI-JOIN path (`join_hash_map_helper.h`, `join_key_constructor.{cpp,hpp}`, `IcebergEqualityDeleteRewriteRule.java`) and has no overlap with MV rewrite.

Youngwb:
Thanks for catching this — `NULL = NULL → UNKNOWN` silently keeping deleted rows alive is exactly the kind of bug that doesn't show up in monitoring.

A few things I'd want to see before this merges:

**1. The FE change has no test.** This is a correctness fix for a specific Iceberg semantic, but the `EQ → EQ_FOR_NULL` switch in `IcebergEqualityDeleteRewriteRule.buildOnPredicate` isn't asserted anywhere:
- No plan test that captures the rewritten anti-join uses `EQ_FOR_NULL` for identifier columns
- No SQL integration test with an eq-delete file containing NULL identifier values, verifying the row actually gets filtered

Without the second one, the bug stays "the code looks right" rather than "we verified the broken case is now fixed". `NormalizeNanInColumn` exercises the helper but doesn't cover join semantics end-to-end either.

**2. The BE NaN normalization is a separate behavior change.** It runs on every `is_null_safe_equal == true` join path, not just Iceberg eq-delete. User SQL written as `t1.f <=> t2.f` on float/double columns will go from current behavior to "all NaN match each other" — a `<=>` operator semantic change with no release note. What is the current StarRocks behavior of `NaN <=> NaN` on float/double? If it was `FALSE` (IEEE-754), this PR silently flips it for everyone.

Worth either splitting the BE NaN normalization into its own PR, or at minimum:
- Calling out the `<=>` behavior change in the PR description
- Flipping the "Does this PR entail a change in behavior" checkbox to Yes

**3. Coverage question on the BE side.** The changes are in `BuildKeyConstructorForSerialized*` / `ProbeKeyConstructorForSerialized*` (the serialized/multi-column key paths). Do single-column null-safe equal joins on float/double go through the same path, or is there a fast path that wasn't touched? Worth confirming, otherwise 1-col vs N-col `<=>` joins on float would behave differently.

Given this has been open since December, splitting into two PRs (FE `EQ_FOR_NULL` as one, BE NaN normalization as another) might be the fastest way to land the FE half — the two halves carry quite different backport risk profiles.

eshishki:
@Youngwb thanks — genuinely useful review, the point-by-point breakdown made it easy to act on. Addressed below.

**1. FE test.** Added both:
- A plan test (`IcebergEqualityDeleteRewriteRuleTest`) asserting the rewritten anti-join uses `EQ_FOR_NULL` for identity columns and keeps `LT` for `$data_sequence_number`.
- A SQL integration test over a v2 fixture whose equality-delete file is keyed on NULL — the NULL-id rows must be deleted (6 visible rows), so the broken case is *verified*, not just "looks right".

Caveat: these Iceberg fixtures are local (a `file://` hadoop catalog copied into `/tmp`). We'd already validated the behavior in a separate PR; here they're checked in properly. I'm not sure the CI external-catalog suite picks up local fixtures like these — ideally we'd host them in the CI external catalog, pointers welcome.

**3 (coverage) — you nailed it.** A single-column null-safe float/double key was taking the `ONE_KEY` fast path, which compares raw float bits and never normalized NaN, so 1-col and N-col `<=>` really did diverge. Good catch. Fixed: float/double keys stay on the null-safe (serialized) path that canonicalizes NaN instead of falling back to `ONE_KEY`.

**2 — agreed it's a behavior change; digging in, it's a bug fix.** The old `NaN <=> NaN = FALSE` wasn't a deliberate policy, it was a default (`std::equal_to` / `FCmpOEQ`). StarRocks' **own** sort comparator already collapses NaN to one value (`column_sorter_comparator.h`), so `<=>` and sort disagreed. We audited the others to be sure:
- **Spark** — `NormalizeFloatingNumbers` canonicalizes NaN for join/group/distinct keys; `NaN` equals `NaN`.
- **ClickHouse** — hash keys are compared bit-wise *specifically* so NaN keys stay consistent; their comment notes the hash-table probing invariants otherwise break when NaNs are present. So for hash joins it's a correctness point, not only semantics.
- **Iceberg** — equality deletes match NaN against NaN, and the spec canonicalizes every NaN bit pattern to `0x7ff8000000000000` — exactly the value we normalize to.

So we fixed the inconsistency in both the hash-join path and the scalar `<=>` evaluator (interpreted + JIT), so they agree with each other and with sort. Scope: only `<=>`; plain `=` is unchanged and still IEEE (`NaN = NaN` stays false). I've flipped the change-in-behavior checkbox to Yes and spelled this out in the description.

I'd prefer to keep the NaN fix in this PR — it's required for the Iceberg equality-delete case this PR is about, and the canonical NaN value is literally from the spec. But if you'd rather land the FE `EQ_FOR_NULL` half first, I'm happy to split the BE NaN work into its own PR, no problem.

**Tests added:** BE unit + end-to-end join tests for NaN (match and non-match control), scalar `<=>` NaN tests, and an Iceberg equality-delete fixture keyed on NaN.

eshishki:
Follow-up on the two SQL integration tests (`test_iceberg_equality_delete_null_key` / `_nan_key`) — they're **red in CI and I can't turn them green from my side**. Here's exactly why.

Both fail at analysis with `(5501) Unknown database 'eq_delete_db'` — the query never runs. The fixtures are a local `file://` hadoop catalog: the case does `shell: cp … /tmp/sr_eq_fixture/warehouse`, then `create external catalog … "iceberg.catalog.warehouse"="file:///tmp/sr_eq_fixture/warehouse"`.

In SQL-Tester CI the cluster is multi-node — FE and BE are separate nodes (`FE_NODE` ≠ `BE_NODE`), and `shell:` runs via `subprocess` on the GitHub runner (a third machine). So the `cp` lands on the runner's `/tmp`, the FE reads `file:///tmp/...` from its own node where the namespace dir doesn't exist → "Unknown database". Not a code bug — it works end-to-end on a single-node deployment (FE/BE/runner sharing one FS), which is where I validated it.

No test tag fixes this: `@native`/`@cloud` select storage mode (shared-nothing vs shared-data), not node count; every iceberg fixture test that passes in CI (e.g. `test_iceberg_variant_query_*`) builds its table over an `oss://${oss_bucket}` warehouse, not a local `file://` tree. Making these green needs the fixtures where all nodes can read them — the CI external HMS+OSS catalog (`iceberg_oss_db`), staged as e.g. `eq_del_tbl_null` / `eq_del_tbl_nan` — and that needs access to the test catalog, which I don't have.

Youngwb:
The NaN audit is convincing — `<=>` was the StarRocks outlier vs sort/group, and aligning to the Iceberg spec's canonical NaN is hard to argue with. I'm with you on this being a correctness fix, not a fresh semantic choice.

That said, I'd still ask for the split, for review attention rather than correctness.

The BE NaN work touches three independent hot paths:
1. Hash join key constructors (build + probe, multi-col + fixed-size)
2. Key-constructor selector / `ONE_KEY` routing (`join_hash_table.cpp`)
3. Scalar `<=>` evaluator — both interpreted and JIT'd LLVM IR

Each has its own subsystem owners who should look at the change separately from the Iceberg eq-delete semantics the FE one-liner is about. Bundled, Iceberg reviewers may LGTM the BE changes by trust, and BE join / JIT reviewers may skim the FE rule changes. Split:

- **FE PR:** `EQ_FOR_NULL` + plan test + Iceberg NULL-key fixture. Iceberg connector reviewers.
- **BE PR:** NaN canonicalization + fast-path routing + scalar `<=>` (interpreted + JIT) + binary predicate tests + NaN fixture. BE join / JIT reviewers, independent rollout.

eshishki:
@Youngwb done — kept the Iceberg part in this PR (FE `EQ → EQ_FOR_NULL` one-liner + plan test + NULL-key SQL fixture) and moved the BE half to #74000 (NaN canonicalization, ONE_KEY routing, scalar `<=>` interp+JIT, reserved-null-byte fix, plus the NaN-key SQL fixture).

Heads-up on #74000: its `test_iceberg_equality_delete_nan_key` will fail in CI for the same reason `test_iceberg_equality_delete_null_key` fails here — both are local `file://` hadoop-catalog fixtures, so `shell: cp` lands on the GitHub runner and the FE/BE on other nodes don't see it. Both work end-to-end on a single-node deployment, where I validated them. BE unit tests on #74000 cover the NaN paths inline.

Youngwb:
@eshishki Re: the SQL fixture caveat (multi-node CI `Unknown database`, because the `shell: cp` lands on the runner while FE/BE read their own nodes) — I've hosted the fixture on the shared CI OSS bucket and opened a small PR against this branch with the test change:

**alifcapital/starrocks#5**

Merging it pulls the change in here. It makes the test read the fixture through a hadoop-on-OSS catalog (no `shell: cp`, no `@sequential`), keeps R at the 6 surviving rows, and drops the 14 binary fixture files. The data is already uploaded to OSS with every embedded `file:/tmp` path rewritten to `oss://`, and verified readable on a live cluster — the `{3,6}` equality-delete applies and, with the `EQ_FOR_NULL` fix in this PR, the two NULL-id rows are deleted → 6 rows. So CI should go green once it's merged.

CelerData-Reviewer:
@codex review

eshishki:
@Youngwb I switched the test to the `register_table` pattern (per `test_iceberg_register_table`) — no hadoop catalog, no namespace listing. The [latest run](https://github.com/StarRocks/starrocks/actions/runs/27016779817) shows the access path works end-to-end now; the only failure left (1 of 1238) is the missing fixture itself:

```
Failed to register table: File does not exist:
oss://${oss_bucket}/iceberg_ci_db/eq_del_convert/metadata/v4.metadata.json
```

Could you re-host the fixture under `iceberg_ci_db/eq_del_convert/` (embedded paths rewritten to that location — your existing tooling)? Or, if a different prefix is preferable, tell me which path to point the test at and I'll update it.

Youngwb (approved):

stephen-shelby (approved):

trueeyu (approved):

PR #67323, opened 2025-12-29
[BugFix] Fallback to non-short-circuit execution in share-data mode

ZiheLiu:
## Why I'm doing:

```sql
select idx from __row_util where idx = 1
 Compute node not found. Check if any compute node is down. nodeId: -1 compute node
```

When `enable_short_circuit` is enabled, in share-data mode, queries that hit short-circuit execution fail with`Compute node not found. Check if any compute node is down. nodeId: -1 compute node`.

**Root cause:**
- On the BE side, share-data mode doesn’t support short-circuit execution because `lake::TabletReader` doesn’t implement the core interface `multi_get`. Only `storage::TabletReader` (for share-nothing mode) implements it.
- On the FE side, it looks up the `compute_node_id` of the tablet only from `backends` **except `compute_nodes`**. When it can’t find a match, it throws the error.

**Changes:**
1. Fall back to non-short-circuit execution in share-data mode.
2. Improve the error message so we can tell from the message that this is a short-circuit query:
   - Before: `Compute node not found. Check if any compute node is down. nodeId: -1 compute node`
   - After:  `No alive backend for short-circuit query. Compute node not found. Check if any compute node is down. nodeId: 1 compute node`

**TODO**
In the future, we need to implement `lake::TabletReader::multi_get` to enable short-circuit execution for compute–storage separation.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

---

> [!NOTE]
> Implements fallback to non-short-circuit execution in shared-data mode and improves failure signaling when no backend is available.
> 
> - Gate `SessionVariable.isEnableShortCircuit()` by run mode: returns `false` in `RunMode.SHARED_DATA`
> - In `ShortCircuitHybridExecutor.assignTablet2Backends()`, wrap worker lookup and rethrow `NonRecoverableException` with a clear "No alive backend for short-circuit query" message
> - Tests: add shared-data fallback coverage and no-alive-backend execution error; minor test refactors/renames
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9cfea6d550ead76c7aae57bd1265f6b3adfa0624. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
@@ -5236,7 +5236,7 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
     }
 
     public boolean isEnableShortCircuit() {
-        return enableShortCircuit && RunMode.getCurrentRunMode() != RunMode.SHARED_DATA;
+        return enableShortCircuit && !RunMode.isSharedDataMode();
     }
 
     public boolean isEnablePrepareStmt() {
--- a/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
@@ -5236,7 +5236,7 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
     }
 
     public boolean isEnableShortCircuit() {
-        return enableShortCircuit;
+        return enableShortCircuit && RunMode.getCurrentRunMode() != RunMode.SHARED_DATA;
     }
 
     public boolean isEnablePrepareStmt() {
--- a/fe/fe-core/src/main/java/com/starrocks/qe/ShortCircuitHybridExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/ShortCircuitHybridExecutor.java
@@ -201,8 +201,6 @@ public class ShortCircuitHybridExecutor extends ShortCircuitExecutor {
 
     /**
      * compute all tablets per be
-     *
-     * @return
      */
     private SetMultimap<TNetworkAddress, TabletWithVersion> assignTablet2Backends() throws NonRecoverableException {
         SetMultimap<TNetworkAddress, TabletWithVersion> backend2Tablets = HashMultimap.create();
@@ -220,7 +218,11 @@ public class ShortCircuitHybridExecutor extends ShortCircuitExecutor {
 
             Optional<Backend> be = pick(scanBackendIds, aliveIdToBackends);
             if (be.isEmpty()) {
-                workerProvider.get().reportWorkerNotFoundException();
+                try {
+                    workerProvider.get().reportWorkerNotFoundException();
+                } catch (NonRecoverableException e) {
+                    throw new NonRecoverableException("No alive backend for short-circuit query. " + e.getMessage());
+                }
             }
             be.ifPresent(backend -> backend2Tablets.put(be.get().getBrpcAddress(), tabletWithVersion));
         }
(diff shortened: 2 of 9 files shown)


alvin-phoenix-ai (commented):
I think we need to improve the error message, too. From the current message, others can't know it has a relation to short-circuit read.

And BTW, I believe we need short-circuit read in shared-data mode

ZiheLiu:
> I think we need to improve the error message, too. From the current message, others can't know it has a relation to short-circuit read.

Make sense, let me pretty the error message about short-circuit read.

> And BTW, I believe we need short-circuit read in shared-data mode

Sure, let me support it instead of fallback to non-short-circuit execution.

alvin-celerdata:
@ZiheLiu 

> Sure, let me support it instead of fallback to non-short-circuit execution.

I suggest that you fix the error log and rollback in one PR and support it in another PR.

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
```diff
@@ -5244,7 +5236,7 @@ public void setEnableShortCircuit(boolean enableShortCircuit) {
     }
 
     public boolean isEnableShortCircuit() {
-        return enableShortCircuit;
+        return enableShortCircuit && RunMode.getCurrentRunMode() != RunMode.SHARED_DATA;
```
alvin-phoenix-ai:
```suggestion
        return enableShortCircuit && !RunMode.isSharedDataMode();
```

fe/fe-core/src/main/java/com/starrocks/qe/ShortCircuitHybridExecutor.java
```diff
@@ -220,7 +218,11 @@ private SetMultimap<TNetworkAddress, TabletWithVersion> assignTablet2Backends()
 
             Optional<Backend> be = pick(scanBackendIds, aliveIdToBackends);
             if (be.isEmpty()) {
-                workerProvider.get().reportWorkerNotFoundException();
+                try {
+                    workerProvider.get().reportWorkerNotFoundException();
+                } catch (NonRecoverableException e) {
+                    throw new NonRecoverableException("No alive backend for short-circuit query. " + e.getMessage());
+                }
```
alvin-phoenix-ai: Why not throw reportWorkerNotFoundException directly?
This non-straightforward way makes it hard for me to understand.

ZiheLiu: > I think we need to improve the error message, too. From the current message, others can't know it has a relation to short-circuit read.

Because `reportWorkerNotFoundException` only reports `Compute node not found. Check if any compute node is down. nodeId: -1 compute node`, which doesn't contain the information about `short-circuit`

alvin-phoenix-ai: How about calling `throw new NonRecoverableException("No alive backend for short-circuit query. " + e.getMessage());` directly?
Or add a new function `reportWorkerNotFoundException` which will accept error message

ZiheLiu: Make sense. Introduce a new function `reportWorkerNotFoundException(String errorMessagePrefix) `.

kangkaisen (approved):

stephen-shelby (approved):

PR #67363, opened 2025-12-30
[Enhancement] syntax sugar for accessing field in variant

dirtysalt:
## What I'm doing:

This is just a POC.

With this feature, you can access like

> select data:commit:cid::string from bluesky limit 10;

instead of 

> select get_variant_string(data, "$.commit.cid") from bluesky limit 10;

TODO:
- access array index
- how to handle quote character

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-grammar/src/main/antlr/com/starrocks/grammar/StarRocks.g4
+++ b/fe/fe-grammar/src/main/antlr/com/starrocks/grammar/StarRocks.g4
@@ -2675,6 +2675,7 @@ primaryExpression
     | primaryExpression COLLATE (identifier | string)                                     #collate
     | literalExpression                                                                   #literal
     | columnReference                                                                     #columnRef
+    | variantReference                                                                    #variantRef
     | base = primaryExpression (DOT_IDENTIFIER | '.' fieldName = identifier )             #dereference
     | left = primaryExpression CONCAT right = primaryExpression                           #concat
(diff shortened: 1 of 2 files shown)


fe/fe-grammar/src/main/antlr/com/starrocks/grammar/StarRocks.g4
```diff
@@ -2752,6 +2753,10 @@ columnReference
     : identifier
     ;
 
+variantReference
+    : identifier ( ':' identifier )* ('::' type)?
```
alvin-phoenix-ai, line 2757: `('::' type)?` This syntax can be applied to many places, not only to variants. 
When we do select data from `FILE`, do get field from JSON, we could use this syntax to give a hint for the type.
So I would suggest that you split this reference into two parts. One is a reference for the field of variant, and the other is the given type

alvin-phoenix-ai: https://www.postgresql.org/docs/8.3/sql-expressions.html
image

PR #67370, opened 2025-12-31
[Enhancement] Support partitioned prefix for shared-data cross cluster migration

wxl24life:
## What I'm doing:

1. When partitioned prefix is enabled on source cluster, we use raw path while replicating files from source cluster's storage.

2. Add a fallback strategy while replicating tables from source cluster with legacy version, where partition root path was different with the current implementation

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Adds end-to-end S3 partitioned-prefix support for shared-data migration while keeping non-S3 behavior unchanged.
> 
> - FE: `LakeReplicationJob` fetches `FilePathInfo` for source table, computes per-partition full S3 path (incl. partitioned prefix), serializes it for recovery, and sends via new `TReplicateSnapshotRequest.src_partition_full_path`; new `StarOSAgent.allocateFilePathFromOtherService` helper
> - BE (lake): `LakeReplicationTxnManager` detects `src_partition_full_path`, converts to `staros://` via `convert_s3_path_to_starlet_uri`, and creates FS with S3 raw-path mode; falls back to `RemoteStarletLocationProvider` and normal mode for non-S3
> - BE (fs/starlet): add `new_fs_starlet(shard_id, bool use_s3_raw_path_mode)` with separate LRU caches; enable `kS3UseRawPathWithScheme` when raw-path is requested; keep legacy overload using raw-path by default
> - BE (worker): `StarOSWorker::build_conf_from_shard_info` now merges an initial conf, allowing the above S3 flag to propagate
> - Thrift: extend `TReplicateSnapshotRequest` with `src_partition_full_path`
> - Tests: add unit tests for S3 path conversion, RemoteStarletLocationProvider paths, FE job serialization/recovery, task propagation, and StarOSAgent new API
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 083f10032c3c8d7a914023eaa6ae93b5271bd6bd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/fs/fs_starlet.cpp
+++ b/be/src/fs/fs_starlet.cpp
@@ -30,6 +30,7 @@
 #include "base/concurrency/stopwatch.hpp"
 #include "base/container/lru_cache.h"
 #include "base/string/string_parser.hpp"
+#include "base/testutil/sync_point.h"
 #include "base/utility/defer_op.h"
 #include "common/config.h"
 #include "fs/encrypt_file.h"
@@ -637,33 +638,58 @@ static void shard_fs_cache_deleter(const CacheKey& /*key*/, void* value) {
     delete static_cast<std::shared_ptr<staros::starlet::fslib::FileSystem>*>(value);
 }
 
-std::shared_ptr<FileSystem> new_fs_starlet(int64_t shard_id) {
-    // The cache here is used to store fslib's shard fs which is used for cross cluster migration,
-    // where each shard fs correspond to one storage volume on source cluster.
-    // We chose a small cache capacity size here, with expect that normally very few storage volumes are used.
+// Internal helper to get or create shard filesystem cache
+static Cache* get_shard_fs_cache() {
     constexpr size_t kDefaultCacheCapacity = 1024 * 10;
     static std::unique_ptr<Cache> g_shard_fs_cache(new_lru_cache(kDefaultCacheCapacity));
+    return g_shard_fs_cache.get();
+}
 
-    // Build cache key from shard_id
-    std::string key_str = std::to_string(shard_id);
+std::shared_ptr<FileSystem> new_fs_starlet(int64_t shard_id, bool use_raw_path) {
+    // The cache here is used to store fslib's shard fs which is used for cross cluster migration,
+    // where each shard fs correspond to one storage volume on source cluster.
+    Cache* cache = get_shard_fs_cache();
+
+    // Build cache key from shard_id, appending "_1" suffix for raw path mode
+    // to differentiate from normal mode entries within the same cache instance.
+    std::string key_str = use_raw_path ? fmt::format("{}_1", shard_id) : std::to_string(shard_id);
     CacheKey cache_key(key_str);
 
     // Try cache lookup
-    Cache::Handle* handle = g_shard_fs_cache->lookup(cache_key);
+    Cache::Handle* handle = cache->lookup(cache_key);
     if (handle != nullptr) {
-        auto* cached_ptr =
-                static_cast<std::shared_ptr<staros::starlet::fslib::FileSystem>*>(g_shard_fs_cache->value(handle));
+        auto* cached_ptr = static_cast<std::shared_ptr<staros::starlet::fslib::FileSystem>*>(cache->value(handle));
         std::shared_ptr<staros::starlet::fslib::FileSystem> shard_fs = *cached_ptr;
-        g_shard_fs_cache->release(handle);
+        cache->release(handle);
         return std::make_shared<StarletFileSystem>(shard_fs);
     }
 
     // Cache miss - create new shard filesystem
     staros::starlet::fslib::Configuration conf;
-    absl::StatusOr<std::shared_ptr<staros::starlet::fslib::FileSystem>> fs_st =
-            g_worker->get_shard_filesystem(shard_id, conf);
+    if (use_raw_path) {
+        // S3 raw path mode: use the input path as-is without normalize_path processing
+        // This is required for S3 storage type to support partitioned prefix feature
+        //
+        // The configuration is passed to StarOSWorker::build_conf_from_shard_info() which
+        // forwards it to ShardInfo::fslib_conf_from_this() as initial configuration.
+        // When cache is enabled, fslib_conf_from_this() will automatically add "cachefs."
+        // prefix to all configuration keys.
+        conf[staros::starlet::fslib::kS3UseRawPathWithScheme] = "true";
+    }
+    // For non-S3 storage types (use_raw_path=false), use default configuration
+    // Starlet will use normalize_path to combine sys.root with the relative path
+#ifdef BE_TEST
+    absl::StatusOr<std::shared_ptr<staros::starlet::fslib::FileSystem>> fs_st(absl::UnimplementedError(""));
+    TEST_SYNC_POINT_CALLBACK("new_fs_starlet::get_shard_filesystem", &fs_st);
+    if (absl::IsUnimplemented(fs_st.status())) {
+        fs_st = g_worker->get_shard_filesystem(shard_id, conf);
+    }
+#else
+    auto fs_st = g_worker->get_shard_filesystem(shard_id, conf);
+#endif
     if (!fs_st.ok()) {
-        LOG(WARNING) << "Failed to get shard filesystem, shard_id: " << shard_id << ", error: " << fs_st.status();
+        LOG(WARNING) << "Failed to get shard filesystem, shard_id: " << shard_id << ", use_raw_path: " << use_raw_path
+                     << ", error: " << fs_st.status();
         return nullptr;
     }
 
@@ -671,19 +697,20 @@ std::shared_ptr<FileSystem> new_fs_starlet(int64_t shard_id) {
 
     // Insert into cache
     auto* cache_value = new std::shared_ptr<staros::starlet::fslib::FileSystem>(shard_fs);
-    handle = g_shard_fs_cache->insert(cache_key, cache_value, 1, shard_fs_cache_deleter);
+    handle = cache->insert(cache_key, cache_value, 1, shard_fs_cache_deleter);
     if (handle != nullptr) {
-        g_shard_fs_cache->release(handle);
+        cache->release(handle);
     } else {
         delete cache_value;
     }
 
     LOG(INFO) << "Created new shard filesystem, shard_id: " << shard_id
-              << ", cache_inserts: " << g_shard_fs_cache->get_insert_count()
-              << ", cache_memory: " << g_shard_fs_cache->get_memory_usage() << " bytes";
+              << ", cache_inserts: " << cache->get_insert_count() << ", use_raw_path: " << use_raw_path
+              << ", cache_memory: " << cache->get_memory_usage() << " bytes";
 
     return std::make_shared<StarletFileSystem>(shard_fs);
 }
+
 } // namespace starrocks
 
 #endif // USE_STAROS
--- a/be/src/storage/lake/lake_replication_txn_manager.cpp
+++ b/be/src/storage/lake/lake_replication_txn_manager.cpp
@@ -35,6 +35,65 @@
 
 namespace starrocks::lake {
 
+#ifdef USE_STAROS
+std::string convert_s3_path_to_starlet_uri(std::string_view s3_path, int64_t shard_id) {
+    // S3 URI format: s3://bucket/path...
+    // Starlet URI format: staros://shard_id/path...
+    // We need to replace "s3://bucket/" with "staros://shard_id/"
+    std::string_view path = s3_path;
+    if (path.find("s3://") == 0) {
+        // Remove "s3://" prefix
+        path.remove_prefix(5);
+        // Find the first "/" after bucket name and skip the bucket part
+        size_t first_slash_pos = path.find('/');
+        if (first_slash_pos != std::string_view::npos) {
+            path.remove_prefix(first_slash_pos + 1);
+        } else {
+            path = std::string_view();
+        }
+        // Build starlet URI: staros://shard_id/path...
+        return build_starlet_uri(shard_id, path);
+    }
+
+    // Not a valid S3 path - log warning
+    LOG(WARNING) << "S3 path does not start with 's3://': " << s3_path;
+    return build_starlet_uri(shard_id, path);
+}
+#endif // USE_STAROS
+
+std::string remove_last_path_component(const std::string& path) {
+    // Find the last "/" which separates the directory name (meta or data)
+    size_t last_slash = path.find_last_of('/');
+    if (last_slash == std::string::npos) {
+        return path;
+    }
+
+    // Get the directory name (meta or data)
+    std::string dir_name = path.substr(last_slash + 1);
+
+    // Get the path before the directory name
+    std::string base_path = path.substr(0, last_slash);
+
+    // Find the second-to-last "/"
+    size_t second_last_slash = base_path.find_last_of('/');
+    if (second_last_slash == std::string::npos) {
+        return path;
+    }
+
+    // Remove the component between second_last_slash and last_slash
+    return base_path.substr(0, second_last_slash + 1) + dir_name;
+}
+
+std::string remove_db_id_component(const std::string& path, int64_t db_id) {
+    std::string db_pattern = "/db" + std::to_string(db_id) + "/";
+    size_t pos = path.find(db_pattern);
+    if (pos == std::string::npos) {
+        return path;
+    }
+    // Remove "db{db_id}/" but keep the "/" before it
+    return path.substr(0, pos + 1) + path.substr(pos + db_pattern.length());
+}
+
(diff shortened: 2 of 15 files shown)


alvin-celerdata:
@cursor review

wxl24life:
this pr relies on staros, will amend it after starlet version released

be/src/fs/fs_starlet.cpp
```diff
@@ -637,53 +638,83 @@ static void shard_fs_cache_deleter(const CacheKey& /*key*/, void* value) {
...
+        // forwards it to ShardInfo::fslib_conf_from_this() as initial configuration.
+        // When cache is enabled, fslib_conf_from_this() will automatically add "cachefs."
+        // prefix to all configuration keys.
+        conf[staros::starlet::fslib::kS3UseRawPathWithScheme] = "true";
+    }
+    // For non-S3 storage types (use_s3_raw_path_mode=false), use default configuration
+    // Starlet will use normalize_path to combine sys.root with the relative path
     absl::StatusOr<std::shared_ptr<staros::starlet::fslib::FileSystem>> fs_st =
             g_worker->get_shard_filesystem(shard_id, conf);
+    TEST_SYNC_POINT_CALLBACK("new_fs_starlet::get_shard_filesystem", &fs_st);
```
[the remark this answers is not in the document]
wxl24life: this shouldn't be a problem, as virtual_tablet_id can only be bound to a unique storage volume

be/src/fs/fs_starlet.cpp
```diff
@@ -637,53 +638,83 @@ static void shard_fs_cache_deleter(const CacheKey& /*key*/, void* value) {
...
+    // Each cache corresponds to different starlet configurations
+    constexpr size_t kDefaultCacheCapacity = 1024 * 10;
+    if (use_s3_raw_path_mode) {
+        static std::unique_ptr<Cache> g_shard_fs_cache_raw_path(new_lru_cache(kDefaultCacheCapacity));
+        return g_shard_fs_cache_raw_path.get();
+    } else {
+        static std::unique_ptr<Cache> g_shard_fs_cache_normal(new_lru_cache(kDefaultCacheCapacity));
+        return g_shard_fs_cache_normal.get();
+    }
+}
```
xiangguangyxg: It's not good to have "s3" in the name.

xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/replication/LakeReplicationJob.java
```diff
@@ -137,4 +243,21 @@ public long getReplicationReplicaCount() {
...
         return count;
     }
+
+    @Override
+    public void gsonPreProcess() throws IOException {
+        // Convert FilePathInfo (protobuf) to byte array for serialization
+        if (srcTableFilePathInfo != null) {
+            srcTableFilePathInfoBytes = srcTableFilePathInfo.toByteArray();
+        }
+    }
```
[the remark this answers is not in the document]
wxl24life: This suggestion is incorrect. The parent class ReplicationJob implements GsonPostProcessable (has gsonPostProcess()), but does NOT implement GsonPreProcessable. There is no super.gsonPreProcess() to call — attempting to do so would cause a compilation error. Note that gsonPostProcess() correctly calls super.gsonPostProcess() because the parent does implement that interface.

fe/fe-core/src/main/java/com/starrocks/replication/LakeReplicationJob.java
```diff
@@ -55,13 +83,39 @@ public LakeReplicationJob(TTableReplicationRequest request) throws MetaNotFoundE
...
     protected LakeReplicationJob(LakeReplicationJob job) {
         super(job);
         this.virtualTabletId = job.virtualTabletId;
         this.srcDatabaseId = job.srcDatabaseId;
         this.srcTableId = job.srcTableId;
+        // Copy srcTableFilePathInfo and its serialized bytes for persistence
+        // Without this, gsonPreProcess() on the copy won't serialize the FilePathInfo,
+        // and after restart srcTableFilePathInfo will be null, breaking S3 partitioned-prefix replication
+        this.srcTableFilePathInfo = job.srcTableFilePathInfo;
+        this.srcTableFilePathInfoBytes = job.srcTableFilePathInfoBytes;
```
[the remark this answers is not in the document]
wxl24life: Thanks for the observation. The race condition described here won't actually cause issues because:
1. The gsonPreProcess() is called on the copy object (created by copyForPersist()), not on the original — ProcessHookTypeAdapterFactory synchronizes on the serialized object.
2. srcTableFilePathInfo is effectively immutable after construction — it's set once in the constructor and never modified.
3. Even if srcTableFilePathInfoBytes were null in the copy, gsonPreProcess() running on the copy would regenerate it from srcTableFilePathInfo, so serialization is always correct.

be/src/fs/fs_starlet.cpp
```diff
@@ -637,53 +638,89 @@ static void shard_fs_cache_deleter(const CacheKey& /*key*/, void* value) {
...
+
+std::shared_ptr<FileSystem> new_fs_starlet(int64_t shard_id, bool use_raw_path) {
     // The cache here is used to store fslib's shard fs which is used for cross cluster migration,
     // where each shard fs correspond to one storage volume on source cluster.
-    // We chose a small cache capacity size here, with expect that normally very few storage volumes are used.
-    constexpr size_t kDefaultCacheCapacity = 1024 * 10;
-    static std::unique_ptr<Cache> g_shard_fs_cache(new_lru_cache(kDefaultCacheCapacity));
+    Cache* cache = get_shard_fs_cache(use_raw_path);
```
kevincai: is it possible to have `use_raw_path` encoded into the CacheKey, so the LRU cache instance can be still a single one.

e.g.

```
if (!use_raw_path) {
     std::string key_str = std::to_string(shard_id);
} else {
    std::string key_str = absl::StrFormat("%lld_1", shard_id);
}

File: be/src/fs/fs_starlet.cpp [hunk 6]
```diff
@@ -637,53 +638,89 @@ static void shard_fs_cache_deleter(const CacheKey& /*key*/, void* value) {
...
+    }
+    // For non-S3 storage types (use_raw_path=false), use default configuration
+    // Starlet will use normalize_path to combine sys.root with the relative path
+    //
+    // SyncPoint allows tests to inject mock filesystem result without calling real g_worker.
+    // The sentinel kUnimplemented status means "not overridden by test", so we proceed with
+    // the real g_worker->get_shard_filesystem() call.
+    absl::StatusOr<std::shared_ptr<staros::starlet::fslib::FileSystem>> fs_st(absl::UnimplementedError(""));
+    TEST_SYNC_POINT_CALLBACK("new_fs_starlet::get_shard_filesystem", &fs_st);
+    if (fs_st.status().code() == absl::StatusCode::kUnimplemented) {
```
kevincai:
```suggestion
    if (absl::IsUnimplementedError(fs_st.status())) {
```

kevincai: this is a bit odd in a production code, try to wrap it with BE_TEST to condition compile it, and only go with this weird code in unit test code path.

be/src/storage/lake/lake_replication_txn_manager.cpp
```diff
@@ -29,10 +28,70 @@
 #include "storage/lake/tablet_manager.h"
 #include "storage/segment_stream_converter.h"
 #include "util/dynamic_cache.h"
+#include "util/trace.h"
 #include "vacuum.h"
 
 namespace starrocks::lake {
 
+#ifdef USE_STAROS
+std::string convert_s3_path_to_starlet_uri(std::string_view s3_path, int64_t shard_id) {
```
kevincai, line 39:
```suggestion
static std::string convert_s3_path_to_starlet_uri(std::string_view s3_path, int64_t shard_id) {
```

be/src/storage/lake/lake_replication_txn_manager.cpp
```diff
@@ -29,10 +28,70 @@
...
+        return build_starlet_uri(shard_id, path);
+    }
+
+    // Not a valid S3 path - log warning
+    LOG(WARNING) << "S3 path does not start with 's3://': " << s3_path;
+    return build_starlet_uri(shard_id, path);
+}
+#endif // USE_STAROS
+
+std::string remove_last_path_component(const std::string& path) {
```
kevincai, line 64:
```suggestion
static std::string remove_last_path_component(const std::string& path) {
```

be/src/storage/lake/lake_replication_txn_manager.cpp
```diff
@@ -29,10 +28,70 @@
...
+    size_t second_last_slash = base_path.find_last_of('/');
+    if (second_last_slash == std::string::npos) {
+        return path;
+    }
+
+    // Remove the component between second_last_slash and last_slash
+    return base_path.substr(0, second_last_slash + 1) + dir_name;
+}
+
+std::string remove_db_id_component(const std::string& path, int64_t db_id) {
```
kevincai, line 87:
```suggestion
static std::string remove_db_id_component(const std::string& path, int64_t db_id) {
```

be/src/storage/lake/lake_replication_txn_manager.cpp
```diff
@@ -65,28 +132,55 @@ Status LakeReplicationTxnManager::replicate_lake_remote_storage(const TReplicate
...
+    if (has_full_path) {
+        // S3 storage type: FE provides full S3 path (supports partitioned prefix feature)
+        // Use S3 raw path mode - starlet will use the path as-is without normalize_path
+        std::string src_partition_starlet_uri = convert_s3_path_to_starlet_uri(src_partition_full_path, src_tablet_id);
+
+        // Append metadata and segment directory names
+        src_meta_dir = join_path(src_partition_starlet_uri, kMetadataDirectoryName);
+        src_data_dir = join_path(src_partition_starlet_uri, kSegmentDirectoryName);
+
+        LOG(INFO) << "S3 storage: converted S3 full path to starlet URI, original: " << src_partition_full_path
```
kevincai: how frequent is this logline expected? will it necessary to be LOG_EVERY_N()?

be/src/storage/lake/lake_replication_txn_manager.cpp
```diff
@@ -65,28 +132,55 @@ Status LakeReplicationTxnManager::replicate_lake_remote_storage(const TReplicate
...
-    auto src_meta_dir = "test_lake_replication/meta";
-    auto src_data_dir = "test_lake_replication/data";
-    auto shared_src_fs_st_or = FileSystem::CreateSharedFromString(src_data_dir);
-    if (!shared_src_fs_st_or.ok()) {
-        return Status::Corruption("Failed to create virtual starlet filesystem");
-    }
-    auto shared_src_fs = shared_src_fs_st_or.value();
-    ASSIGN_OR_RETURN(auto src_tablet_meta,
-                     _tablet_manager->get_tablet_metadata(src_tablet_id, src_visible_version, false, 0, nullptr));
+    return Status::NotSupported("Lake replication remote storage requires StarOS (USE_STAROS)");
```
kevincai:
```suggestion
    return Status::NotSupported("Lake replication remote storage requires build with shared-data support!");
```

be/src/storage/lake/lake_replication_txn_manager.cpp
```diff
@@ -65,28 +132,55 @@ Status LakeReplicationTxnManager::replicate_lake_remote_storage(const TReplicate
...
+#ifdef USE_STAROS
+    std::string src_meta_dir;
+    std::string src_data_dir;
+    std::shared_ptr<FileSystem> shared_src_fs;
+    TabletMetadataPtr src_tablet_meta;
+
+    if (has_full_path) {
+        // S3 storage type: FE provides full S3 path (supports partitioned prefix feature)
+        // Use S3 raw path mode - starlet will use the path as-is without normalize_path
+        std::string src_partition_starlet_uri = convert_s3_path_to_starlet_uri(src_partition_full_path, src_tablet_id);
```
kevincai, line 150: this may fail silently if the provided fullpath is non-s3 style path, will it be expected, or should check this inconsistency and fail the request directly?

wxl24life: fixed

be/src/fs/fs_starlet.cpp
wxl24life: fixed

fe/fe-core/src/main/java/com/starrocks/replication/LakeReplicationJob.java
```diff
@@ -122,6 +181,53 @@ private void sendReplicateLakeRemoteStorageTasks() throws Exception {
...
+     * - Non-S3 storage types (OSS/Azure/HDFS/GFS): FE returns null.
+     *   BE will use RemoteStarletLocationProvider to construct the relative path,
+     *   and starlet will use normalize_path to build the final path.
+     * 
+     * The partitioned prefix feature is ONLY supported for S3 storage type.
+     * 
+     * @param physicalPartitionId the physical partition ID
+     * @return the full S3 path if storage type is S3, null otherwise
+     */
+    private String getS3PartitionFullPath(long physicalPartitionId) {
```
xiangguangyxg:
```suggestion
    private String getPartitionFullPath(long physicalPartitionId) {
```

fe/fe-core/src/main/java/com/starrocks/replication/LakeReplicationJob.java
```diff
@@ -55,13 +83,39 @@ public LakeReplicationJob(TTableReplicationRequest request) throws MetaNotFoundE
...
+            if (srcStorageVolume == null) {
+                throw new MetaNotFoundException("Source storage volume not found: " + srcStorageVolumeName);
+            }
+            StarOSAgent starOSAgent = GlobalStateMgr.getCurrentState().getStarOSAgent();
+            this.srcTableFilePathInfo = starOSAgent.allocateFilePathFromOtherService(
+                    srcStorageVolume.getId(), srcDatabaseId, srcTableId, srcServiceId);
+            LOG.info("LakeReplicationJob created, src_storage_volume: {}, src_table_file_path: {}",
+                    srcStorageVolumeName, srcTableFilePathInfo.getFullPath());
+        } catch (DdlException e) {
+            throw new MetaNotFoundException("Failed to allocate file path for source table: " + e.getMessage());
```
xiangguangyxg:
```suggestion
            throw new MetaNotFoundException("Failed to allocate file path for source table: " + srcTableId, e);
```

fe/fe-core/src/main/java/com/starrocks/replication/LakeReplicationJob.java
wxl24life: Good catch! You're right — since this method handles all storage types (not just S3), the name `getS3PartitionFullPath` is misleading. Renamed to `getPartitionFullPath`. Fixed in the latest commit. Thanks!

fe/fe-core/src/main/java/com/starrocks/replication/LakeReplicationJob.java
wxl24life: Great suggestion! Including `srcTableId` in the error message makes it much easier to identify the problematic table during debugging. Also preserved the original cause exception for full stack trace. Fixed in the latest commit. Thanks!

xiangguangyxg (dismissed):

be/src/fs/fs_starlet.cpp
```diff
@@ -637,53 +638,83 @@ static void shard_fs_cache_deleter(const CacheKey& /*key*/, void* value) {
...
     LOG(INFO) << "Created new shard filesystem, shard_id: " << shard_id
-              << ", cache_inserts: " << g_shard_fs_cache->get_insert_count()
-              << ", cache_memory: " << g_shard_fs_cache->get_memory_usage() << " bytes";
+              << ", cache_inserts: " << cache->get_insert_count() << ", use_raw_path: " << use_raw_path
+              << ", cache_memory: " << cache->get_memory_usage() << " bytes";
 
     return std::make_shared<StarletFileSystem>(shard_fs);
 }
+
+std::shared_ptr<FileSystem> new_fs_starlet(int64_t shard_id) {
```
kevincai: should this interface be removed? looks like no one will use it, replication txn manager will use the new interface with 2 arguments.

kevincai (approved):

xiangguangyxg (approved):

HangyuanLiu (approved):

kangkaisen (approved):

PR #67496, opened 2026-01-05
[Enhancement] support multi-threaded publish version daemon

starrocks-xupeng:
## What I'm doing:

introduce FE config `publish_version_poller_threads`, default will use the single thread publish daemon, if set to larger than 0, each DB will be assigned to a publish version thread.

preparation
```
1. set publish_version_poller_threads to 8
2. create 5 databases, each has 1 table
create table t (k int, v int) distributed BY HASH(`k`) BUCKETS 6 properties ('replication_num'='1');
```

test command, run  these 2 at the same time
```
sysbench test_multi_db.lua \
  --mysql-host=172.26.34.240 \
  --mysql-port=9030 \
  --mysql-user=root \
  --mysql-password= \
  --mysql-db=test \
  --percentile=95 \
  --threads=100 \
  --time=120 \
  run

sysbench test_multi_read_create.lua \
  --mysql-host=172.26.34.240 \
  --mysql-port=9030 \
  --mysql-user=root \
  --mysql-password= \
  --mysql-db=test \
  --percentile=95 \
  --threads=20 \
  --time=120 \
  run
```

test script `test_multi_db.lua`
```
function event()
  local id1 = math.random(1, 5)
  local id2 = math.random(1, 10000)
  connections[sysbench.tid]:query(
  string.format("insert into test%d.t values (%d,%d)", id1, id2, id2)
  )
end
```

test script `test_multi_read_create.lua`
```
function event()
  local conn = connections[sysbench.tid]

  local id1 = math.random(1, 5)

  local p = math.random(1, 100)
  if p <= 5 then
    local tbl_id = math.random(1, 1000000)
    local tbl_name = string.format("tmp_sb_%d_%d", sysbench.tid, tbl_id)

    conn:query(string.format([[
      CREATE TABLE test%d.%s (
        id INT,
        v  INT
      ) distributed BY HASH(`id`) BUCKETS 6
    ]], id1, tbl_name))

    conn:query(string.format("DROP TABLE if exists test%d.%s", id1, tbl_name))

  else
    local id2 = math.random(1, 10000)

    conn:query(
      string.format(
        "SELECT * FROM test%d.t WHERE k = %d",
        id1, id2
      )
    )
  end
end
```

before
```
Latency (ms):
         min:                                   99.40
         avg:                                 3565.94
         max:                                42975.98
         95th percentile:                 14047.70
         sum:                             15312162.14
```

after
```
Latency (ms):
         min:                                   78.26
         avg:                                  847.02
         max:                                 8601.03
         95th percentile:                  1973.38
         sum:                             12024346.54
```

截屏2026-01-06 上午11 25 54

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Enables parallel publish-version processing by database with a configurable worker pool.
> 
> - Adds immutable `Config.publish_version_poller_threads` (default 1) to toggle multi-thread mode
> - Refactors `PublishVersionDaemon`:
>   - New `prepare()` and `start()` flow initializing executors; per-DB poller thread pool and `processingDbs` guard
>   - Splits logic into `runWithSingleThread()` vs `runWithMultiThreads()`; per-DB `processDbTransactions()` handling batch/non-batch and shared-nothing/shared-data paths
>   - Initializes lake/delete executors in `prepare()`; exposes some methods as `protected`
> - Adds tests covering executor resizing/validation, multi-thread per-DB scheduling, and batch/non-batch processing paths
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d1db0658030ee5918f3765a5333ab7373689f4e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -3174,6 +3174,10 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true, comment = "the max number of threads for lake table publishing version")
     public static int lake_publish_version_max_threads = 512;
 
+    @ConfField(mutable = false, comment = "the number of threads for per-DB publish version polling. " +
+            "Set to > 1 to use the multi-thread model.")
+    public static int publish_version_poller_threads = 1;
+
     @ConfField(mutable = true, comment = "the max number of threads for lake table delete txnLog when enable batch publish")
     public static int lake_publish_delete_txnlog_max_threads = 16;
 
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
@@ -82,6 +82,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.stream.Collectors;
@@ -98,8 +99,22 @@ public class PublishVersionDaemon extends FrontendDaemon {
     // each thread under the default configurations
     private static final int LAKE_PUBLISH_MAX_QUEUE_SIZE = 4096;
 
+    // Default number of poller threads
+    private static final int POLLER_THREADS_DEFAULT = 4;
+    // Hard limit for poller threads
+    private static final int POLLER_THREADS_HARD_LIMIT = 128;
+
     private ThreadPoolExecutor lakeTaskExecutor;
     private ThreadPoolExecutor deleteTxnLogExecutor;
+
+    // Thread pool for per-DB publish version polling
+    private ThreadPoolExecutor pollerExecutor;
+
+    // Track which DBs are currently being processed to avoid duplicate submissions
+    private final Set<Long> processingDbs = ConcurrentHashMap.newKeySet();
+
+    // Whether to use multi-thread mode, determined at initialization and cannot be changed at runtime
+    private boolean useMultiThreadMode;
 
     @VisibleForTesting
     protected Set<Long> publishingLakeTransactions;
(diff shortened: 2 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3171,6 +3171,10 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true, comment = "the max number of threads for lake table publishing version")
     public static int lake_publish_version_max_threads = 512;
 
+    @ConfField(mutable = false, comment = "the number of threads for per-DB publish version polling. " +
```
kevincai, line 3177: can this be mutable?

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -109,52 +124,202 @@ public class PublishVersionDaemon extends FrontendDaemon {
...
-                    publishVersionForLakeTableBatch(readyTransactionStatesBatch);
-                }
-                return;
+            if (useMultiThreadMode) {
+                runWithMultiThreads();
+            } else {
+                runWithSingleThread();
+            }
+        } catch (Throwable t) {
+            LOG.error("errors while publish version to all backends", t);
```
kevincai, line 225: this error message is misleading.

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -109,52 +124,202 @@ public class PublishVersionDaemon extends FrontendDaemon {
...
-            List<Long> allBackends =
-                    GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getBackendIds(false);
-            if (RunMode.isSharedDataMode()) {
-                allBackends.addAll(
-                        GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getComputeNodeIds(false));
+            // Submit task to executor
+            try {
+                executor.execute(() -> {
+                    try {
+                        processDbTransactions(dbId, dbTxnMgr);
```
kevincai, line 311: this will cause the publish for the db occupying a thread until it finishes. Consider change each publish to a future, and every dbtxnmgr shall submit all the ready txns and wait for the future, and then run next round. The granularity is still large to run `processDbTransaction()` as a task in thread pool.

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
starrocks-xupeng: this is the message before

fe/fe-core/src/main/java/com/starrocks/common/Config.java
starrocks-xupeng: it's safer to put this immutable

kevincai: it will be bothersome to adjust the thread and restart FE leader when the thread pool is not enough. Try to make it mutable.

kevincai: the thread pool size can be adjusted in every runAfterCatalogReady()

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -109,52 +124,238 @@ public class PublishVersionDaemon extends FrontendDaemon {
...
-            GlobalTransactionMgr globalTransactionMgr = GlobalStateMgr.getCurrentState().getGlobalTransactionMgr();
-            if (Config.lake_enable_batch_publish_version && RunMode.isSharedDataMode()) {
-                // batch publish
-                List<TransactionStateBatch> readyTransactionStatesBatch = globalTransactionMgr.
-                        getReadyPublishTransactionsBatch();
-                if (readyTransactionStatesBatch.size() != 0) {
-                    publishVersionForLakeTableBatch(readyTransactionStatesBatch);
+        if (lakeTaskExecutor == null) {
+            /*
+             * Create a thread pool executor for LakeTable synchronizing publish.
```
kevincai, line 154: should not do this if is not a shared-data cluster.

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -109,52 +124,238 @@ public class PublishVersionDaemon extends FrontendDaemon {
...
+        // TODO: need to refactor after be split into cn + dn
+        List<Long> allBackends =
+                GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getBackendIds(false);
+        if (RunMode.isSharedDataMode()) {
+            allBackends.addAll(
+                    GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getComputeNodeIds(false));
+        }
+
+        if (allBackends.isEmpty()) {
+            LOG.warn("some transaction state need to publish, but no backend exists");
```
kevincai, line 260: this will be very chatty if truly no node available. make it debug for now.

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -109,52 +124,238 @@ public class PublishVersionDaemon extends FrontendDaemon {
...
+                executor.execute(() -> {
+                    try {
+                        processDbTransactions(dbId, dbTxnMgr);
+                    } catch (Throwable t) {
+                        LOG.error("Error processing transactions for db {}", dbId, t);
+                    } finally {
+                        processingDbs.remove(dbId);
+                    }
+                });
+            } catch (Exception e) {
```
kevincai, line 318: `Exception` is not good enough, put it in finally to make sure the id is removed from `processingDbs`, or use closable resource, something like

```
 try (closable xx = new deferOp(() -> processingDbs.remove(dbId))) {
}

Comment (starrocks-xupeng) [reply fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java:318]: there is a finally when the task has been submitted to the pool, we can't put it in outer try

File: fe/fe-core/src/main/java/com/starrocks/common/Config.java [hunk 7]
```diff
@@ -3174,6 +3174,10 @@ public class Config extends ConfigBase {
...
+            "Each DB is assigned to a thread via hash. Set to 0 to use the single-thread model.")
```
kevincai: why set to 0 for single-thread, it is more intuitive to set 1 as single thread, the detailed implementation shall take both 0 and 1 as single-thread model.

starrocks-xupeng: set to 1 by default

fe/fe-core/src/main/java/com/starrocks/common/Config.java
kevincai: I would insist make this mutable, otherwise this parameter is rarely useful.

Issue #66457, opened 2025-12-08
Tablet/Bucket support holding larger amounts of data under shared-data arch


luohaha:
## Feature request
Under the current best practices, the recommended size of a single tablet in StarRocks is approximately 1 GB (compressed). While this configuration works well in many cases, it introduces several scalability and performance challenges when the number of tablets grows large. 

#### Limitations of Small Tablets
1. Non-partitioned PK tables typically involve a large number of tablets during ingestion, leading to:
  a. High memory overhead — Each tablet maintains its own ingest memtable and PK index memtable.
  b. High CPU & I/O overhead — Each tablet independently triggers frequent compaction.
  c. Worsened realtime ingestion performance — In high-throughput or realtime CDC scenarios, the above overhead compounds and becomes a bottleneck.

2. Query performance degradation due to excessive IO
Even for partitioned tables, a single query may need to scan a large number of tablets. This results in:
  1. Significant I/O amplification, especially when scanning cold data.
  2. Poor cache resistance — Query latency deteriorates rapidly when cache hit rate drops, making performance highly unstable.

3. Increased FE metadata pressure
The Frontend (FE) must maintain a very large number of shards and tablet metadata, which leads to:
  a. Higher FE CPU and memory usage.
  b. Increased GC pressure and RPC overhead.
  c. A major stability risk, as FE becomes the bottleneck before BE capacity is fully utilized.

4. It cannot handle data skew because each tablet is processed with a single thread for both ingestion and compaction. If data skew occurs and one tablet holds significantly more data than others, it will become a clear performance bottleneck.

So we want to expand the per-tablet capacity from ~1 GB to  100 GB (shared-data architecture).

#### Benefits of Large Tablet
1. Significantly reduced ingestion resource cost
  a. Fewer tablets mean fewer memtables and PK index memtables.
  b. Memory usage during ingestion drops dramatically.
  c. Ingestion stability improves, especially for realtime and high-throughput workloads.
2. Lower compaction amplification
  a. Compaction no longer fans out across thousands of tablets.
  b. CPU, I/O, and write amplification are greatly reduced.
  c. Overall ingestion throughput becomes smoother and more predictable.
3. Improved query performance
  a. Fewer tablets to scan → lower I/O overhead.
  b. Performance becomes far more cache-insensitive; even with low cache hit rates, query latency does not deteriorate sharply.
  c. More efficient short-circuit pruning and metadata lookups.
4. Reduced FE metadata burden
[...]

PR #67554, opened 2026-01-07
[Enhancement] Support remote storage mapper files for lake primary key compaction

luohaha:
## Why I'm Doing This

  This PR enhances the Primary Key compaction workflow in Lake (shared-data) storage engine to support parallel pk index execution with remote storage mapper files. Currently, when `enable_pk_index_parallel_execution` is enabled, rows mapper files (.crm) are stored on local disk which limits scalability. This change introduces remote storage support for mapper files (.lcrm) on S3/HDFS, enabling:

  1. **Multi-node accessibility** - Multiple compute nodes can access the same mapper file during parallel execution without file replication
  2. **Light publish optimization** - Recording mapper file metadata in transaction logs eliminates expensive re-computation during publish phase
  3. **Performance improvement** - Avoiding repeated get_size() calls to remote storage saves ~10-50ms per file access

  ## What I'm Doing

  **High-level changes:**

  This PR introduces Lake Compaction Rows Mapper (.lcrm) files stored on remote storage (S3/HDFS) to support distributed parallel pk index execution, with metadata tracking in transaction logs for performance optimization.

  **Technical details:**

  - **New file format**: Added `.lcrm` (Lake Compaction Rows Mapper) files stored on remote storage alongside existing local `.crm` files
    - Local `.crm`: Fast I/O (1-5ms), single-node only, deleted immediately after use
    - Remote `.lcrm`: Slower I/O (50-200ms), multi-node accessible, managed by metadata GC

  - **Metadata optimization**: Record `lcrm_file` (name + size) in `TxnLogPB_OpCompaction` proto (gensrc/proto/lake_types.proto:281) to avoid expensive get_size() HEAD requests to S3/HDFS during publish phase

  - **Interface refactoring**: Changed `PrimaryKeyCompactionConflictResolver::filename()` from returning `string` to `FileInfo` struct containing both path and optional size

  - **Storage decision logic** (be/src/storage/rows_mapper.cpp:165-184):
    - If `enable_pk_index_parallel_execution` enabled → use remote storage (.lcrm)
    - Otherwise → use local disk (.crm) for faster I/O

  - **Light publish enhancement** (be/src/storage/lake/update_manager.cpp:1265-1296):
    - Check `op_compaction.has_lcrm_file()` first for remote mapper metadata
    - Fallback to local file existence check for backward compatibility
    - Enables publish optimization that reduces publish time from minutes to seconds for large compactions

  - **File lifecycle management**:
    - Local `.crm` files: Deleted immediately in `RowsMapperIterator` destructor (be/src/storage/rows_mapper.cpp:78-95)
    - Remote `.lcrm` files: Added to `orphan_files` list in metadata for asynchronous GC cleanup (be/src/storage/lake/meta_file.cpp:292-305)

  **Modified components:**
  - `compaction_task.cpp` - Record lcrm metadata in transaction log
  - `filenames.h` - Add `gen_lcrm_filename()` and `is_lcrm()` helper functions
  - `location_provider.h` - Add `lcrm_location()` path builder
  - `meta_file.*` - Add `remove_lcrm_file()` for GC integration
  - `pk_tablet_writer.cpp` - Capture lcrm file info after finalization
  - `tablet_manager.*` - Expose `lcrm_location()` API
  - `update_manager.*` - Enhance light publish detection logic
  - `rows_mapper.*` - Core implementation with dual storage strategy
  - `lake_types.proto` - Add `lcrm_file` field to `OpCompaction` message
  - Test files updated to use new FileInfo interface

  **Testing:**
  - Updated unit tests in `primary_key_compaction_task_test.cpp`, `rows_mapper_test.cpp`, `tablet_updates_test.cpp`
  - Backward compatibility maintained - existing `.crm` files continue to work

  **Performance benefits:**
  - Eliminates 10-50ms get_size() latency per mapper file access (critical when processing hundreds of files)
  - Light publish optimization leverages pre-computed mapper to skip re-reading/re-processing compaction data
  - Parallel pk execution gains outweigh slower remote I/O

  **Risks:**
  - Proto change (`lcrm_file` field addition) requires careful rollout in mixed-version clusters
  - Remote storage dependency introduces potential failure modes (S3 throttling, network issues)
  - Increased remote storage costs for mapper files (mitigated by GC cleanup)

Fix #66457

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Enables remote `.lcrm` rows mapper files for parallel PK index compaction with metadata-driven light publish optimization and proper lifecycle management.
> 
> - Record `lcrm_file` (name + size) in `TxnLogPB_OpCompaction` and serialize in txn logs; add proto field `lcrm_file` in `OpCompaction`
> - Writers generate mapper via `new_lake_rows_mapper_filename(...)`; capture file info on finalize and attach to compaction log
> - Conflict resolvers now return `FileInfo` (path + optional size); lake resolver resolves remote vs local paths using `TabletManager`, local resolver adapted
> - `RowsMapperIterator::open` accepts `FileInfo`; avoids get_size() when provided; destructor deletes only `.crm`, not `.lcrm`; builder exposes `file_info()`
> - Add helpers: `is_lcrm`, `gen_lcrm_filename`, and `LocationProvider::lcrm_location`; `TabletManager::lcrm_location` wrapper
> - Update light publish decision to prefer `op_compaction.lcrm_file` and fall back to local `.crm` existence
> - GC: `MetaFileBuilder` marks `lcrm_file` as orphan on publish for deferred cleanup
> - Tests updated to new interfaces and behaviors; expectations adjusted (extra orphan file)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fc90c2ef37f0b25694263478ff55e95af1687522. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/be/src/storage/rows_mapper.cpp
+++ b/be/src/storage/rows_mapper.cpp
@@ -17,7 +17,9 @@
 #include <fmt/format.h>
 
 #include "fs/fs.h"
+#include "lake/filenames.h"
 #include "storage/data_dir.h"
+#include "storage/lake/tablet_manager.h"
 #include "storage/storage_engine.h"
 #include "util/coding.h"
 #include "util/crc32c.h"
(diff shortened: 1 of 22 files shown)


be/src/storage/rows_mapper.cpp
```diff
@@ -129,6 +166,28 @@ Status RowsMapperIterator::status() {
     return Status::OK();
 }
 
+StatusOr<std::string> new_lake_rows_mapper_filename(lake::TabletManager* mgr, int64_t tablet_id, int64_t txn_id) {
+    // DESIGN DECISION: Storage location depends on execution mode
+    if (config::enable_pk_index_parallel_execution) {
```
smartlxh, line 171: Is it possible to put lcrm file in the remote  regardless of the execution mode, and without incurring significant overhead?

luohaha: The S3 PUT api cost can't be avoid

meegoo (approved):

be/src/storage/rows_mapper.cpp
smartlxh: Or add a separate switch to allow users to control whether to upload to S3？ In more elastic scenarios,  users might be more willing to pay the cost of api，compared to the performance loss caused by missing CRM files

smartlxh (approved):

wyb (approved):

PR #67677, opened 2026-01-09
[UT] Add basic test for iceberg delete

Youngwb:
## Why I'm doing:
#66944

## What I'm doing:
This pull request introduces significant changes to the Parquet position reader logic and its associated tests, as well as adds a comprehensive new SQL test for Iceberg partition transform deletes. The main update in the Parquet reader is the removal of filter-based row selection, leading to all positions in a range being output regardless of any filter. The tests are updated to reflect this new behavior. Additionally, a new end-to-end test script is added for verifying delete operations on Iceberg tables partitioned with various transforms.

The most important changes are:

### Parquet Position Reader Logic

* Removed filtering logic from `ParquetPosReader::read_range`, so it now always outputs positions for all rows in the specified range, ignoring any provided filter.

### Parquet Reader Tests

* Updated `TestReadRangeWithFilter` to expect all rows in the range instead of only filtered rows, and adjusted value checks accordingly.
* Updated `TestReadRangeWithNoSelectedRows` to expect all rows in the range, even when the filter would previously have selected none.

### Iceberg Partition Transform Delete Testing

* Added a new SQL test script `test_iceberg_delete_partition_transform` that creates Iceberg tables with various partition transforms (bucket, truncate, year, month, day, hour, and combinations), performs delete operations on them, and verifies the results. This ensures that delete functionality works correctly with different partitioning schemes.
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/formats/parquet/parquet_pos_reader.cpp
+++ b/be/src/formats/parquet/parquet_pos_reader.cpp
@@ -20,24 +20,10 @@
 namespace starrocks::parquet {
 
 Status ParquetPosReader::read_range(const Range<uint64_t>& range, const Filter* filter, ColumnPtr& dst) {
-    if (filter == nullptr) {
-        // No filter, generate positions for all rows in the range
-        for (uint64_t i = range.begin(); i < range.end(); ++i) {
-            // Generate position based on the absolute row position in the file.
-            int64_t pos = i;
-            dst->as_mutable_raw_ptr()->append_datum(Datum(pos));
-        }
-    } else {
-        // Apply filter, only generate positions for selected rows
-        DCHECK_EQ(filter->size(), range.span_size()) << "Filter size must match range size";
-        for (uint64_t i = range.begin(); i < range.end(); ++i) {
-            size_t filter_index = i - range.begin();
-            if ((*filter)[filter_index]) {
-                // Generate position based on the absolute row position in the file.
-                int64_t pos = i;
-                dst->as_mutable_raw_ptr()->append_datum(Datum(pos));
-            }
-        }
+    for (uint64_t i = range.begin(); i < range.end(); ++i) {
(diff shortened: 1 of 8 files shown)


be/src/formats/parquet/parquet_pos_reader.cpp
```diff
@@ -20,24 +20,10 @@
...
-        DCHECK_EQ(filter->size(), range.span_size()) << "Filter size must match range size";
-        for (uint64_t i = range.begin(); i < range.end(); ++i) {
-            size_t filter_index = i - range.begin();
-            if ((*filter)[filter_index]) {
-                // Generate position based on the absolute row position in the file.
-                int64_t pos = i;
-                dst->as_mutable_raw_ptr()->append_datum(Datum(pos));
-            }
-        }
+    for (uint64_t i = range.begin(); i < range.end(); ++i) {
```
dirtysalt, line 23: if you don't use filter, remove it from parameter list.

dirtysalt (commented):
some comments

be/src/formats/parquet/parquet_pos_reader.cpp
Youngwb: It's override from columnReader 

Image

GavinMar (approved):

gengjun-git (approved):

dirtysalt (approved):

PR #67778, opened 2026-01-12
[Enhancement] support adding FE observer

zhaohehuhu:
## Why I'm doing:
Support adding a read-only Observer role in FE to scale read capacity and improve query load balancing.

## What I'm doing:

as title

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

---

> [!NOTE]
> Adds observer-mode support to FE bootstrap.
> 
> - New `IS_FE_OBSERVER` env var (default `false`) in `fe_entrypoint.sh`
> - When set, FE registers with leader via `ALTER SYSTEM ADD OBSERVER "<host:port>"` instead of adding as follower
> - Leader probing updated: pod index 0 only uses pod0 probing when not an observer; observers follow podX probing path
> - No changes when `IS_FE_OBSERVER` is unset/false; follower behavior preserved
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1d161da3750e54faa839f62488b52dfe78a21904. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docker/dockerfiles/fe/fe_entrypoint.sh
+++ b/docker/dockerfiles/fe/fe_entrypoint.sh
@@ -24,6 +24,7 @@ STARROCKS_HOME=${STARROCKS_ROOT}/fe
 FE_CONFFILE=$STARROCKS_HOME/conf/fe.conf
 META_DIR=$STARROCKS_HOME/meta
 EXIT_IN_PROGRESS=false
+IS_FE_OBSERVER=${IS_FE_OBSERVER:-"false"}
 
 log_stderr()
 {
@@ -218,8 +219,13 @@ start_fe_no_meta()
         local start=`date +%s`
         while true
         do
-            log_stderr "Add myself($MYSELF:$EDIT_LOG_PORT) to leader as follower ..."
-            mysql --connect-timeout 2 -h $FE_LEADER -P $QUERY_PORT -u root --skip-column-names --batch -e "ALTER SYSTEM ADD FOLLOWER \"$MYSELF:$EDIT_LOG_PORT\";"
+            if [[ "$IS_FE_OBSERVER" == 'true' ]]; then
+              log_stderr "Add myself($MYSELF:$EDIT_LOG_PORT) to leader as observer ..."
+              mysql --connect-timeout 2 -h $FE_LEADER -P $QUERY_PORT -u root --skip-column-names --batch -e "ALTER SYSTEM ADD OBSERVER \"$MYSELF:$EDIT_LOG_PORT\";"
+            else
+              log_stderr "Add myself($MYSELF:$EDIT_LOG_PORT) to leader as follower ..."
+              mysql --connect-timeout 2 -h $FE_LEADER -P $QUERY_PORT -u root --skip-column-names --batch -e "ALTER SYSTEM ADD FOLLOWER \"$MYSELF:$EDIT_LOG_PORT\";"
+            fi
             # check if added successful
             if show_frontends $svc | grep -q -w "$MYSELF" &>/dev/null ; then
                 break;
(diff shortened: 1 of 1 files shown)


docker/dockerfiles/fe/fe_entrypoint.sh
```diff
@@ -24,6 +24,7 @@ STARROCKS_HOME=${STARROCKS_ROOT}/fe
 FE_CONFFILE=$STARROCKS_HOME/conf/fe.conf
 META_DIR=$STARROCKS_HOME/meta
 EXIT_IN_PROGRESS=false
+IS_FE_OBSERVER=${IS_FE_OBSERVER:-"false"}
```
kevincai, line 27: better have a separate entrypoint so that for an observer, it will never take itself as leader and starts after detecting timeout, instead it will wait forever or timeout and exit with failure start.

zhaohehuhu: Got it. I will add a new entrypoint for observers only.

kevincai: it's ok to reuse the same entrypoint, but need to carefully go through entire workflow and handle observer specifically.

zhaohehuhu: Thanks for your review. @kevincai

alvin-celerdata:
@cursor review

docker/dockerfiles/fe/fe_entrypoint.sh
```diff
@@ -218,8 +219,13 @@ start_fe_no_meta()
         local start=`date +%s`
         while true
         do
-            log_stderr "Add myself($MYSELF:$EDIT_LOG_PORT) to leader as follower ..."
-            mysql --connect-timeout 2 -h $FE_LEADER -P $QUERY_PORT -u root --skip-column-names --batch -e "ALTER SYSTEM ADD FOLLOWER \"$MYSELF:$EDIT_LOG_PORT\";"
+            if [[ "$IS_FE_OBSERVER" == 'true' ]]; then
+              log_stderr "Add myself($MYSELF:$EDIT_LOG_PORT) to leader as observer ..."
```
kevincai: keep the same indention in this file.

zhaohehuhu: Fixed. Thanks!

kevincai (dismissed):

docker/dockerfiles/fe/fe_entrypoint.sh
kevincai, line 27: both `true` and `false` are used in the script for this variable, in order to improve the script robust, need to check the value of this variable, only take `1/true/True` as `true`, all other values including invalid ones or empty as 'false' and only check `IS_FE_OBSERVER == false` or `IS_FE_OBSERVER != false` to avoid unexpected behavior. For example: "IS_FE_OBSERVER = 1" will fail both `== true' and `== false` check.

zhaohehuhu: Thanks. The code was refactored. @kevincai

docker/dockerfiles/fe/fe_entrypoint.sh
```diff
@@ -218,8 +224,13 @@ start_fe_no_meta()
         local start=`date +%s`
         while true
         do
+          if [[ "$IS_FE_OBSERVER" != "false" ]]; then
```
kevincai: looks like the indention is not correct. please fix the indetion.

zhaohehuhu: Fixed. Thanks @kevincai

kevincai (approved):

xiangguangyxg (approved):

PR #67786, opened 2026-01-12
[Enhancement] Reduce allocations in Statistics.getUsedColumns

stephen-shelby:
## Why I'm doing:
`Statistics.getUsedColumns()` is frequently invoked during planning and sits on a hot path. The previous implementation created a new `ColumnRefSet` per column via `ColumnRefOperator.getUsedColumns()` and unioned sets repeatedly, causing avoidable allocations and overhead (especially noticeable with many columns / lambda arguments).

## What I'm doing:
- Rework `Statistics.getUsedColumns()` to iterate `columnStatistics.keySet()` directly, **skip `LAMBDA_ARGUMENT`**, and union by **column id** to avoid per-column `ColumnRefSet` allocations.
- Minor hot-path follow-ups:
  - Replace a stream/lambda in `PredicateStatisticsCalculator` with a simple loop for unknown-stat checks.
  - Optimize cross join stats building in `StatisticsCalculator` by copying only required column statistics via `ColumnRefSet` id lookups and avoiding cloning when histogram is already null.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/PredicateStatisticsCalculator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/PredicateStatisticsCalculator.java
@@ -86,9 +86,12 @@ public class PredicateStatisticsCalculator {
                 return false;
             }
             // extract range predicate scalar operator with unknown column statistics will not eval
-            if (predicate.isFromPredicateRangeDerive() &&
-                    statistics.getColumnStatistics().values().stream().anyMatch(ColumnStatistic::isUnknown)) {
-                return false;
+            if (predicate.isFromPredicateRangeDerive()) {
+                for (ColumnStatistic cs : statistics.getColumnStatistics().values()) {
+                    if (cs.isUnknown()) {
+                        return false;
+                    }
+                }
             }
             return true;
         }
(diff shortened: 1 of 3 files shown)


andyziye:
@cursor review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/PredicateStatisticsCalculator.java
```diff
@@ -86,9 +86,12 @@ protected boolean checkNeedEvalEstimate(ScalarOperator predicate) {
                 return false;
             }
             // extract range predicate scalar operator with unknown column statistics will not eval
-            if (predicate.isFromPredicateRangeDerive() &&
-                    statistics.getColumnStatistics().values().stream().anyMatch(ColumnStatistic::isUnknown)) {
-                return false;
+            if (predicate.isFromPredicateRangeDerive()) {
```
kangkaisen, line 89: I think two ways are equal？

stephen-shelby: They are functionally equivalent but not performance-equal. The for-each version is ~30-50% faster due to:
- Less object allocation (no Stream/Sink creation)
- Better JIT inlining
- Simpler bytecode

Since this is in the optimizer hot path, the micro-optimization is worthwhile.

Seaven (approved):

kangkaisen (approved):

PR #67828, opened 2026-01-13
[Enhancement] Improve pruning in DP join reorder

stephen-shelby:
## Why I'm doing:
In DP join reorder, `calculateStatistics()` is expensive and is invoked for each candidate join expression. Even with existing pruning, we still end up calling `calculateStatistics()` for many candidates that are provably worse than the current best plan. Adding tighter lower-bound pruning **before** stats estimation can significantly reduce the number of expensive `calculateStatistics()` calls.

## What I'm doing:
- Add a child-cost lower bound (`saturatingAdd(leftCost, rightCost)`) to prune candidates early before building/estimating join expressions.
- Add an additional pre-stats lower bound that incorporates CROSS JOIN penalty (and pure cross row-count lower bound) and the non-equi (nestloop) penalty via `existsEqOnPredicate`, to skip expensive candidates before calling `calculateStatistics()`.
- The CROSS JOIN / non-equi penalty handling and saturation rules are **aligned with the existing `computeCost()` implementation**, so the pruning is semantically consistent (it only skips candidates that cannot beat `bestCostSoFar` under the same cost model).
- Use `BitSet.nextSetBit(0)` for the single-atom base case to simplify and speed up index lookup.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/join/JoinOrder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/join/JoinOrder.java
@@ -531,7 +531,7 @@ public abstract class JoinOrder {
         return true;
     }
 
-    private boolean existsEqOnPredicate(OptExpression optExpression) {
+    protected boolean existsEqOnPredicate(OptExpression optExpression) {
         LogicalJoinOperator joinOp = optExpression.getOp().cast();
         List<ScalarOperator> onPredicates = Utils.extractConjuncts(joinOp.getOnPredicate());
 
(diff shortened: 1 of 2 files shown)


andyziye:
@cursor review

Seaven (approved):

kangkaisen (approved):

PR #67873, opened 2026-01-14
[BugFix] change max string length to be configuable (from 1 to 1M) when unknown

jaogoy:
## Why I'ms doing:
Currently, SR can't infer a correct max length for a complex expression, such as with a function.
For some BI tools (such as Qlik) and using ODBC, they rely on the returned max string length in metadata of each query. Then, when a query contains some string functions, the BI tools may get empty string value.

(But, according to my test, Excel, pyodbc, PowerBI return right values)

## What I'm doing:
So, we can set the max string length from 1 to 1048576 according to users' need when it's -1 (can't be infered by SR correctly).
Default: 64 (the same as before)

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/qe/GlobalVariable.java
```diff
@@ -222,6 +223,9 @@ public final class GlobalVariable {
     @VariableMgr.VarAttr(name = ENABLE_TDE, flag = VariableMgr.GLOBAL | VariableMgr.READ_ONLY)
     public static boolean enableTde = KeyMgr.isEncrypted();
 
+    @VariableMgr.VarAttr(name = ENABLE_LARGE_STRING_METADATA_LENGTH, flag = VariableMgr.GLOBAL)
+    private static boolean enableLargeStringMetadataLength = false;
```
kevincai: should this be more proper as a FE configuration variable instead of a session variable?

jaogoy: IMO, it's related to query protocol, not to FE or BE.
So, a global variable is better.

StarRocks-Reviewer:
@cursor review

fe/fe-core/src/main/java/com/starrocks/qe/GlobalVariable.java
kevincai: FE delegates the mysql protocol, so it is fine to be an FE configuration.

Or can this be like the `@@max_allowed_packet` , make it integer instead of true/false switch, `@@max_unknown_string_meta_length`, set it default to 64 * 3, and for the specific user, it can be changed to a large value as needed. Just in case the hard coded 1M causes other problems and then the length can't be changed.

kevincai (approved):

Seaven (approved):

kangkaisen (approved):

PR #67944, opened 2026-01-14
[Enhancement] Optimize window functions with skewed partition keys by splitting into UNION

martinr0x:
## Why I'm doing:
Window functions involving `PARTITION BY` and `ORDER BY` currently generate an execution plan structured as: 
`Scan -> Exchange (Hash Partition) -> Local Sort -> Analytic`
When data is partitioned, all rows with the same partition key are sent to the same execution instance. If the partition expression has data skew (e.g., a large number of NULLs or a specific high-frequency value), a single node receives the clear majority of the data. This forces the Local Sort and Analytic calculation for that partition to run effectively single-threaded on that one node, causing performance bottlenecks.

## What I'm doing:
I introduced a new optimization rule `SplitWindowSkewToUnionRule` (`RuleType TF_SPLIT_WINDOW_SKEW`).
This rule detects skew in the window partition column (using column statistics for NULL skew or MCV/Most Common Value skew) and rewrites the plan to process the skewed value separately from the rest of the data.
The query is effectively transformed into a `UNION ALL` of two branches:
Skewed Branch: Filters for the skewed value (e.g., p IS NULL or p = 1). Since the partition key is constant here, the `PARTITION BY` clause is removed, leaving only the `ORDER BY`. This allows the expensive sorting to happen distributed (`filtering -> local sort -> merge exchange`) rather than on a single node.
Unskewed Branch: Filters out the skewed value (e.g., p IS NOT NULL or p != 1) and proceeds with the original window execution plan.

Example, say p has NULL skew:
Original:
```sql
with final AS (
  select p, s, sum(x) over (partition by p order by s)
  from t
)
```
Rewritten:
```sql
with unskewed AS (
  select p, s, sum(x) over (partition by p order by s)
  from t
  where p is not NULL
), skewed AS (
  select p, s, sum(x) over (order by s)
  from t
  where p is NULL
), final AS (
  select *
  from unskewed
  union all
  select *
  from skewed
)
```
For the skewed partition, the plan changes to 
`Scan -> Local Sort (Distributed) -> Merge Exchange -> Analytic`, significantly improving parallel processing capabilities for the skewed data portion.
## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

---

> [!NOTE]
> Adds a targeted optimizer rewrite for window functions with skewed partition keys.
> 
> - Introduces `SplitWindowSkewToUnionRule` (`TF_SPLIT_WINDOW_SKEW`) to split a window with a single partition key and `ORDER BY` into two branches: skewed value (no partition) and remaining data (original partition), based on NULL/MCV stats
> - Integrates the rule in `QueryOptimizer` before predicate pushdown and triggers stats calculation prior to application
> - Adds session variable `enable_split_window_skew_to_union` (with getters/setters) to control the optimization
> - Updates `RuleType`; adds unit tests in `WindowSkewTest` to cover NULL and MCV skew, multiple columns, disabled flag, and non-applicable cases
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8a8173528b8ae58c9994b45f07b259a613b2770e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


StarRocks-Reviewer:
@cursor review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -0,0 +1,237 @@
...
+            ColumnStatistic colStat = statistics.getColumnStatistic(col);
+            var skewInfo = DataSkew.getColumnSkewInfo(statistics, colStat);
+            if (skewInfo.isSkewed()) {
+
+                if (skewInfo.type() == DataSkew.SkewType.SKEWED_NULL) {
+                    return List.of(new SkewedInfo(col, ConstantOperator.createNull(col.getType())));
+                }
+                if (skewInfo.maybeMcvs().isPresent()) {
+                    return skewInfo.maybeMcvs().get().stream().map(p -> (new SkewedInfo(col,
+                            ConstantOperator.createVarchar(p.first)))).collect(Collectors.toList());
```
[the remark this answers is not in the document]
martinr0x: Lets decide this with the reviewer.

fe/fe-core/src/test/java/com/starrocks/sql/plan/WindowSkewTest.java
```diff
@@ -0,0 +1,267 @@
...
+                        "nullable: true], ]\n" +
+                        "  |  partition by: [1: p, INT, true]\n" +
+                        "  |  order by: [2: s, INT, true] ASC\n" +
+                        "  |  window: RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW\n" +
+                        "  |  cardinality: 700\n" +
+                        "  |  column statistics: \n" +
+                        "  |  * p-->[-Infinity, Infinity, 0.0, NaN, NaN] MCV: [[1:300]] ESTIMATE\n" +
+                        "  |  * s-->[-Infinity, Infinity, 0.0, 1.0, 1.0] UNKNOWN\n" +
+                        "  |  * x-->[-Infinity, Infinity, 0.0, 1.0, 1.0] UNKNOWN\n" +
+                        "  |  * sum(3: x)-->[-Infinity, Infinity, 0.0, 1.0, 1.0] UNKNOWN");
```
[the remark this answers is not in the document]
martinr0x: Fixed.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -0,0 +1,240 @@
...
+
+        // 3. Build Unskewed Branch (where p != skewed_value)
+        // We keep the original partition expressions for the rest of the data.
+        BranchResult unskewedBranch = buildBranch(
+                context,
+                child,
+                window,
+                window.getPartitionExpressions(),
+                unskewedPredicate
+        );
```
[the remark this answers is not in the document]
martinr0x: I dont think this is necessary here.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -0,0 +1,243 @@
...
+
+    private BranchResult buildBranch(OptimizerContext context,
+                                     OptExpression child,
+                                     LogicalWindowOperator originalWindow,
+                                     List<ScalarOperator> partitionExprs,
+                                     ScalarOperator predicate) {
+
+        // Create Filter
+        LogicalFilterOperator filterOp = new LogicalFilterOperator(predicate);
+        OptExpression filterExpr = OptExpression.create(filterOp, child);
```
LiShuMing: 1. For two different branches (with or without skew value), the child is computed twice, which may be expensive if the child is heavy; maybe we can use a CTEAnchor for reuse.
2. And It's not safe for the two branches with the same child, whose column ref is the same, since there are some optimizations which are using the column ref directly and the two same column refs may cause potential bugs.

fe/fe-core/src/test/java/com/starrocks/sql/plan/WindowSkewTest.java
```diff
@@ -0,0 +1,276 @@
...
+import com.starrocks.statistic.StatisticsMetaManager;
+import com.starrocks.thrift.TExplainLevel;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+
+class WindowSkewTest extends PlanTestBase {
```
LiShuMing, line 31: it's better to add an end-to-end sql test for this feature to cover/find more bugs.

LiShuMing (commented):
It's a good idea generally, and we may even extend this idea to other skew cases.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -0,0 +1,243 @@
...
+import com.starrocks.sql.optimizer.skew.DataSkew;
+import com.starrocks.sql.optimizer.statistics.ColumnStatistic;
+import com.starrocks.sql.optimizer.statistics.Statistics;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class SplitWindowSkewToUnionRule extends TransformationRule {
```
kangkaisen, line 107: would better add a comment for this class,  Draw the plan tree before and after the transformation.

martinr0x: added

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
martinr0x: 1.  If we create a `LogicalCTEAnchorOperator` for child, can the filter expression we apply on both branches still be pushed down?
2. I added duplication for one of the branches, I used `SplitJoinORToUnionRule` as inspiration.

StarRocks-Reviewer:
@cursor review

fe/fe-core/src/test/java/com/starrocks/sql/plan/WindowSkewTest.java
martinr0x: @LiShuMing added some e2e tests ([5791d03](https://github.com/StarRocks/starrocks/pull/67944/commits/5791d038bffaae86b4739683e159abdb7ec73218))

alvin-celerdata:
@codex review

StarRocks-Reviewer:
@cursor review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -0,0 +1,340 @@
...
+                                     boolean needsDuplication) {
+
+        OptExpression filterExpr = OptExpression.create(new LogicalFilterOperator(predicate), child);
+        Map<ColumnRefOperator, ColumnRefOperator> mapping = Maps.newHashMap();
+
+        LogicalWindowOperator.Builder windowBuilder = new LogicalWindowOperator.Builder()
+                .setAnalyticWindow(originalWindow.getAnalyticWindow())
+                .setUseHashBasedPartition(partitionExprs.isEmpty() && originalWindow.isUseHashBasedPartition())
+                .setIsSkewed(partitionExprs.isEmpty() && originalWindow.isSkewed())
+                .setInputIsBinary(originalWindow.isInputIsBinary());
```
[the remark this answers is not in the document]
kangkaisen: +1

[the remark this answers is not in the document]
martinr0x: [1ce264c](https://github.com/StarRocks/starrocks/pull/67944/commits/1ce264c7d45b24d078aa58ce343b906161ad9487)

stephen-shelby:
please fix the failed sql test `test_window_skew_rewrite_with_mcv`

kangkaisen (dismissed):

test/lib/skip.py
```diff
@@ -66,5 +66,6 @@
 ]
 
 skip_files = set([
+    'test_window_skew_rewrite_with_mcv',
```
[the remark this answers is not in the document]
martinr0x: @kangkaisen @stephen-shelby the test relies on histograms and for some reason the histograms are not available after they are computed in the test. I could not reproduce this issue locally and also adding wait sleeps has not resolved the issue. 
The other test `test_window_skew_rewrite_with_null` works as expected.
Unless you have any suggestions what goes wrong here I would skip this case for now.

kangkaisen (approved):

test/lib/skip.py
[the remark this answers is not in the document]
kangkaisen: OK, I think we could merge this PR first

stephen-shelby (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -0,0 +1,335 @@
...
+                                     List<ScalarOperator> partitionExprs,
+                                     ScalarOperator predicate,
+                                     boolean needsDuplication) {
+
+        OptExpression filterExpr = OptExpression.create(new LogicalFilterOperator(predicate), child);
+        Map<ColumnRefOperator, ColumnRefOperator> mapping = Maps.newHashMap();
+
+        LogicalWindowOperator.Builder windowBuilder = new LogicalWindowOperator.Builder()
+                .withOperator(originalWindow)
+                .setUseHashBasedPartition(partitionExprs.isEmpty() && originalWindow.isUseHashBasedPartition())
```
stephen-shelby, line 244: Thanks for you contribution!
NIT: 
In `buildBranch()`, `useHashBasedPartition` should not be gated by `partitionExprs.isEmpty()`.   `useHashBasedPartition` is only meaningful when `ORDER BY` is empty (see BE `AnalyticNode::_use_hash_based_partition`), while this rule only applies when `ORDER BY` is present. It should be `false` (or simply inherit `originalWindow`, but it will be `false` in this rule anyway).  you can check this one.

martinr0x: Ah okay, make sense. I will anyways open a follow-up PR to add more advanced skew handling :)

PR #67947, opened 2026-01-14
[BugFix] Fix variadic functions returning wrong dates

### LINKED ISSUE #67946: Variadic functions return wrong dates when mixing DATE and DATETIME
### Steps to reproduce the behavior

  ```sql
  -- Returns wrong date: 2026-01-11 instead of 2026-01-14
  SELECT date(greatest(date('2026-01-14'), timestamp('2026-01-11 00:00:00')));
```

  Should return 2026-01-14 (the greater date).

  Real behavior (Required)

  Returns 2026-01-11 (the wrong date).

### Root cause: 
Variadic functions greatest(DATE, DATETIME), least(DATE, DATETIME), and coalesce(DATE, DATETIME) return BIGINT instead of DATETIME. The values are compared as numbers:
  - date('2026-01-14') → 20260114 (BIGINT)
  - timestamp('2026-01-11 00:00:00') → 20260111000000 (BIGINT)
  - greatest(20260114, 20260111000000) = 20260111000000 (larger number)
  - date(20260111000000) = 2026-01-11 (wrong result)

  This happens because in gensrc/script/functions.py, BIGINT signatures are registered before DATETIME signatures for variadic functions. Function resolution picks the first matching signature (BIGINT), losing temporal semantics.

  Note: Fixed-arity functions (IF, IFNULL, NULLIF) are not affected.

### 
StarRocks version:

At least these:
  - StarRocks 3.5.0
  - StarRocks 4.0.3

### PR DESCRIPTION
**Pull request by fresh-borzoni**
## Why I'm doing:

  Variadic functions (greatest, least, coalesce) return wrong dates when mixing DATE and DATETIME types.
 
 Root Cause:
Function resolution in `fe/fe-core/src/main/java/com/starrocks/catalog/Function.java` uses "first match wins" strategy.  The `getFunction()` method (lines 778-817) tries each signature in order and returns the first one where all arguments can implicitly cast to the signature's types via `IS_NONSTRICT_SUPERTYPE_OF` mode.

The Bug:
  Before this fix, BIGINT signatures were registered before DATETIME signatures:
  greatest(BIGINT, ...) <- tried first
  greatest(DATETIME, ...) <- never reached

  For variadic functions, `Function.java` lines 584-593 check if trailing arguments can cast to the vararg type using `TypeManager.isImplicitlyCastable()`. Since both DATE and DATETIME can cast to BIGINT (for internal storage), the BIGINT signature matched first, causing temporal values to be compared as numbers instead of dates.

  ## What I'm doing:

  Reorder signatures of variadic functions working with date/datetimes in `functions.py` to prioritize DATETIME before numeric types:
  greatest(DATETIME, ...) <- tried first, matches DATE+DATETIME 
  greatest(BIGINT, ...) <- only tried if DATETIME fails

  **Why No Regressions:**
  The type casting check in  `Function.java:561-597` prevents false positives:
  - DATE/DATETIME can cast to DATETIME (temporal semantics preserved)
  - INT/BIGINT cannot cast to DATETIME (pure numbers skip to numeric signatures)

  Closes #67946

  ## What type of PR is this:

  - [x] BugFix
  - [ ] Feature
  - [ ] Enhancement
  - [ ] Refactor
  - [ ] UT
  - [ ] Doc
  - [ ] Tool

  - [ ] Feature removed

  ## Checklist:

  - [ ] This is a backport pr

    - [x] 4.1
    - [x] 4.0
    - [x] 3.5
    - [x] 3.4

### DIFF
--- a/test/sql/test_function/R/test_greatest_least_coalesce_datetime
+++ b/test/sql/test_function/R/test_greatest_least_coalesce_datetime
@@ -1,0 +1,68 @@
+-- name: test_greatest_date_datetime_wrong_result @system
+-- result:
+2026-01-14
+-- !result
+select date(greatest(date('2026-01-14'), timestamp('2026-01-11 00:00:00')));
+-- result:
+2026-01-14
+-- !result
+-- name: test_greatest_date_datetime_returns_datetime @system
+-- result:
+2026-01-14 00:00:00
+-- !result
+select greatest(date('2026-01-11'), timestamp('2026-01-14 00:00:00'));
+-- result:
+datetime
+-- !result
+select typeof(greatest(date('2026-01-11'), timestamp('2026-01-14 00:00:00')));
+-- result:
+datetime
+-- !result
+-- name: test_least_date_datetime_returns_datetime @system
+-- result:
+2026-01-11 00:00:00
+-- !result
+select least(date('2026-01-14'), timestamp('2026-01-11 00:00:00'));
+-- result:
+datetime
+-- !result
+select typeof(least(date('2026-01-14'), timestamp('2026-01-11 00:00:00')));
+-- result:
+datetime
+-- !result
+-- name: test_coalesce_date_datetime_returns_datetime @system
+-- result:
+2026-01-11 00:00:00
+-- !result
+select coalesce(null, date('2026-01-11'), timestamp('2026-01-14 00:00:00'));
+-- result:
+datetime
+-- !result
+select typeof(coalesce(null, date('2026-01-11'), timestamp('2026-01-14 00:00:00')));
+-- result:
+datetime
+-- !result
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by StarRocks-Reviewer** [comment]
@cursor review

File: test/sql/test_function/R/test_greatest_least_coalesce_datetime [hunk 1]
```diff
@@ -0,0 +1,68 @@
+-- name: test_greatest_date_datetime_wrong_result @system
+-- result:
```
stephen-shelby: there are only the results, without the query. Please use the results generated from the T file
for example:
`python3 run.py -r--file_filter="test_greatest_date_datetime_wrong_result" -a system`

stephen-shelby (approved):

kangkaisen (approved):

PR #67690, opened 2026-01-09
[BugFix] Fix identifier quoting for non-MySQL JDBC tables

killme2008:
## Why I'm doing:

Modified `getIdentifierSymbol()` to return database-specific quote characters based on JDBC URL prefix, otheriwise the query will fail when using keywords as column names, see #67115

Core Fix: JDBCScanNode.java
 - Modified `getIdentifierSymbol()` to return database-specific quote characters based on JDBC URL prefix
 - MySQL/MariaDB/ClickHouse: backtick
 - PostgreSQL/Oracle/SQL Server: double quote (ANSI SQL)

Tests: `MySqlAndJDBCScanNodeTest.java`
  - Added unit tests for PostgreSQL, Oracle, SQL Server identifier quoting

Tests: `JDBCIdentifierQuoteTest.java (new)`
  - Catalog-level integration tests for PostgreSQL and MySQL quoting

## What I'm doing:

Fixes #67115

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

---

> [!NOTE]
> Aligns JDBC identifier quoting with target databases to prevent failures when using reserved keywords or mixed cases.
> 
[...]

PR #67965, opened 2026-01-15
[BugFix] Fix sql server and oracle identifier symbol

Youngwb:
## Why I'm doing:
Fix https://github.com/StarRocks/StarRocksTest/issues/10865, this issue is introduced by https://github.com/StarRocks/starrocks/pull/67690

SQL Server Issues with Double Quotes:
**Setting Dependency**: Double quotes only work when the QUOTED_IDENTIFIER session setting is ON (enabled). If this setting is OFF, double quotes will cause syntax errors.
**Non-Standard Convention**: SQL Server primarily uses square brackets [] as its native identifier quoting mechanism. Double quotes are supported for SQL standard compliance but are not the preferred syntax.

Oracle Issues with Double Quotes:
**Case Sensitivity Trap**: Double-quoted identifiers are case-sensitive, while unquoted identifiers are automatically converted to uppercase. This creates a common pitfall where developers expect case-insensitive behavior but get case-sensitive matching instead.
**Mandatory Consistent Usage**: If an object (table, column, etc.) is created with double quotes, every subsequent reference to that object must use double quotes with the exact same case. This creates maintenance overhead and breaks standard querying patterns.
**Storage Mismatch:** Most Oracle objects are created without double quotes (stored as uppercase in data dictionaries). Attempting to query them with double quotes and lowercase names will fail because Oracle looks for exact case matches.

## What I'm doing:
This pull request updates how identifier symbols are handled for Oracle and SQL Server JDBC sources. Previously, table names were wrapped in double quotes for these databases; now, they are left unquoted. Corresponding unit tests have been updated to reflect this change.

JDBC identifier symbol handling:

* Modified `JDBCScanNode.getIdentifierSymbol()` to stop returning double quotes for Oracle and SQL Server JDBC URIs, so these databases now use unquoted identifiers.

Test updates:

* Updated `testFiltersInOracleJDBCScanNode` and `testFiltersInSqlServerJDBCScanNode` in `MySqlAndJDBCScanNodeTest.java` to expect unquoted table names in the explain output.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/JDBCScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/JDBCScanNode.java
@@ -158,9 +158,7 @@ public class JDBCScanNode extends ScanNode {
             return "`";
         }
         if (jdbcUri.startsWith("jdbc:postgresql") ||
-                jdbcUri.startsWith("jdbc:postgres") ||
-                jdbcUri.startsWith("jdbc:oracle") ||
-                jdbcUri.startsWith("jdbc:sqlserver")) {
+                jdbcUri.startsWith("jdbc:postgres")) {
             return "\"";
         }
         return "";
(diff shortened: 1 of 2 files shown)


StarRocks-Reviewer:
@cursor review

trueeyu (approved):

alvin-phoenix-ai (approved):

PR #56691, opened 2025-03-07
[Enhancement] Delete shards meta from starmgr after recycling partition for cloud-native cluster

wxl24life:
## Why I'm doing:

Currently, the RecycleBin is responsible for removing partition directories from remote storage, but it relies on StarMgrMetaSyncer to clean up shard metadata from StarMgr. However, since StarMgrMetaSyncer was not designed to handle heavy cleanup tasks, it would be more efficient to remove the metadata immediately after the RecycleBin completes its work.
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 3.4
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1
  - [ ] 3.0

PR #68039, opened 2026-01-16
[BugFix] Delete shard group meta from StarManager when dropping cloud-native table

wxl24life:
## Why I'm doing:
When executing DROP TABLE FORCE on a cloud-native (shared_data) table, the table's shard group metadata in StarManager was not being deleted immediately. This leads to orphan shard groups accumulating in `StarManager`.

The reason is that `LakeTableHelper.deleteTableFromRecycleBin()` only calls `LakeTableCleaner.cleanTable()` to delete remote storage data. It doesn't call `deleteShardGroup()` to clean up shard group metadata in `StarManager`. The orphan shard groups are only cleaned up by the background `StarMgrMetaSyncer` thread.

For clusters with frequent table creation/deletion operations, this results in a large number of orphan shard/shardGroup entries accumulating in `StarManager`, consuming memory and potentially affecting FE's performance.

## What I'm doing:

This PR is a supplement to #56691. In that PR, we proactively deleted the metadata in `StarManager` at the end of the drop partition operation.

Similarly,  in this PR I add `deleteTableShardGroupMeta()` method to `LakeTableHelper` and call it in `deleteTableFromRecycleBin()` after successfully cleaning table data. This ensures shard group metadata is deleted immediately when a table is dropped, consistent with the behavior of partition deletion (which already calls `deleteShardGroupMeta`).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableCleaner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableCleaner.java
@@ -15,16 +15,12 @@
 package com.starrocks.lake;
 
 import com.staros.client.StarClientException;
-import com.staros.proto.ShardInfo;
 import com.starrocks.catalog.OlapTable;
-import com.starrocks.catalog.PhysicalPartition;
+import com.starrocks.catalog.Partition;
 import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.warehouse.cngroup.ComputeResource;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-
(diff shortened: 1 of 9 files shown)


kevincai:
May need to consider break down the Table DROP into partitions drop first, clean the table meta after no partition remains.

PR #68075, opened 2026-01-19
[BugFix] change max string length to be configuable (from 1 to 1M) when unknown (backport #67873)

jaogoy:
## Why I'ms doing:
Currently, SR can't infer a correct max length for a complex expression, such as with a function.
For some BI tools (such as Qlik) and using ODBC, they rely on the returned max string length in metadata of each query. Then, when a query contains some string functions, the BI tools may get empty string value.

(But, according to my test, Excel, pyodbc, PowerBI return right values)

## What I'm doing:
So, we can set the max string length from 1 to 1048576 according to users' need when it's -1 (can't be infered by SR correctly).
Default: 64 (the same as before)

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


jaogoy:

kevincai (approved):

PR #68124, opened 2026-01-19
[Enhancement] Add segment metadata filter for lake tables to skip segments based on sort key range

meegoo:
## Why I'm doing:

In customer-facing scenarios, there are many queries that filter by tenant_id or similar prefix columns of the primary key (e.g., WHERE tenant_id=123 on a table with primary key (tenant_id, order_id)). Currently, lake tables have to scan all segments in a rowset even when the query only involves sort key prefix columns, causing unnecessary I/O and computation overhead.

## What I'm doing:

Lake tables already record sort_key_min/max in SegmentMetadataPB during write, but these metadata are not utilized during query. This commit leverages the existing sort_key_min/max metadata to filter segments before reading. By reusing the zone map filter mechanism from ColumnPredicate, it supports various predicate types including =, !=, <, <=, >, >=, IN, and NOT IN on sort key columns. A new config option enable_lake_segment_metadata_filter (default true) is added to control this feature, and a profile counter SegmentMetadataFilterRows is added for monitoring the filter effectiveness.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/olap_common.h
+++ b/be/src/storage/olap_common.h
@@ -296,6 +296,11 @@ struct OlapReaderStatistics {
     int64_t read_pk_index_ns = 0;
 
     // ------ for lake tablet ------
+    // Rows skipped by segment metadata filter (sort key range filtering).
+    int64_t segment_metadata_filtered = 0;
+    // Number of segments skipped by segment metadata filter.
+    int64_t segments_metadata_filtered = 0;
+
     int64_t pages_from_local_disk = 0;
 
     int64_t compressed_bytes_read_local_disk = 0;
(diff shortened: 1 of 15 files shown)


be/src/storage/olap_common.h
```diff
@@ -296,6 +296,11 @@ struct OlapReaderStatistics {
     int64_t read_pk_index_ns = 0;
 
     // ------ for lake tablet ------
+    // Rows skipped by segment metadata filter (sort key range filtering).
+    int64_t segment_metadata_filtered = 0;
+    // Number of segments skipped by segment metadata filter.
+    int64_t segments_metadata_filtered = 0;
```
wyb, line 302:
```suggestion
    int64_t segments_filtered = 0;
```

wyb (approved):

dirtysalt (approved):

Seaven (approved):

PR #68125, opened 2026-01-19
[BugFix] Fix shuffle join key order mismatch with EQ_FOR_NULL (<=>)

stephen-shelby:
## Why I'm doing:

This PR fixes a **wrong-result** issue that can happen in multi-join queries when at least one join uses **`<=>` (EQ_FOR_NULL)**.

In StarRocks, the **order** of hash-shuffle keys matters: hashing by `(k1, k2)` is generally different from hashing by `(k2, k1)`. If a parent join’s equal conjunct list implies join keys in order `(k1, k2)`, but a child subtree is already partitioned by `(k2, k1)`, matching rows can be routed to different nodes, causing **missing matches** (outer joins often make this visible as unexpected NULLs / missing rows).

A common confusion is: “the parent already derived the correct required key order, why didn’t the child get an Exchange enforcer after we adjusted the required property?”

This is expected in the current architecture, because:

- **Required properties are advisory** (used to select/lookup a child plan), not an immediate “force Exchange now” command.
- The parent first asks the child group for the best expression under `childRequiredProperty`, then obtains `childOutputProperty`, and later phases may treat the child as satisfying requirements based on property checks.
- **An Exchange enforcer is only inserted when the optimizer observes a concrete mismatch** between required distribution and the child’s output distribution.

The bug is that under `<=>` (join keys become `nullStrict=true`), the optimizer’s “adjust required shuffle-join keys to match parent order” logic may be triggered by a **permutation match** (same columns, different order). That adjustment can rewrite the required key order *without guaranteeing the child’s physical partition expr order is rewritten accordingly*. This can make the plan appear to satisfy requirements and skip enforcement, while the physical plan still contains swapped shuffle keys, leading to wrong results.

Replacing `<=>` with `=` often avoids the issue because `=` uses `nullStrict=false`, which changes how keys compare/match and reduces the chance of triggering the problematic permutation-based adjustment.

## What I'm doing:

- **Apply a minimal, case-by-case fix for `<=>` joins**:
  - Keep the existing set-based matching behavior in general.
  - When `nullStrict=true` is involved (typical for `<=>`), **do not allow a “same columns, different order” match to trigger shuffle-join key adjustment**. Only adjust if the key order is already consistent.
- **Make key index mapping stable when `nullStrict=true`**:
  - Use colId-based lookup to map required keys to join key positions (avoid `DistributionCol.equals` strictness differences affecting `indexOf`).

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


StarRocks-Reviewer:
@cursor review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/PropertyDeriverBase.java
```diff
@@ -55,15 +54,30 @@ protected static List<PhysicalPropertySet> computeShuffleJoinRequiredProperties(
...
                 List<DistributionCol> requiredRight = Lists.newArrayList();
 
                 for (DistributionCol cid : requiredColumns) {
-                    int idx = adjustBasedOnLeft ? leftShuffleColumns.indexOf(cid) : rightShuffleColumns.indexOf(cid);
+                    int idx;
+                    if (hasNullStrict) {
+                        idx = adjustBasedOnLeft ? indexOfByColId(leftShuffleColumns, cid.getColId())
+                                : indexOfByColId(rightShuffleColumns, cid.getColId());
+                    } else {
+                        idx = adjustBasedOnLeft ? leftShuffleColumns.indexOf(cid) : rightShuffleColumns.indexOf(cid);
```
kangkaisen, line 79: When hasNullStrict is true, the new indexOfByColId lookup ignores nullStrict and duplicate occurrences. If the join predicates repeat the same column with different null semantics (e.g., t1.a <=> t2.b AND t1.a = t2.c) or reuse the same column across multiple keys, indexOfByColId always returns the first match. This collapses distinct key positions, can drop the nullStrict variant, and can misalign left/right key pairs, making the required distribution incorrect and potentially skipping an Exchange, yielding wrong results for <=> joins.

please check this case.

```
Map<DistributionCol, Deque<Integer>> leftPos = new HashMap<>();
for (int i = 0; i < leftShuffleColumns.size(); i++) {
    leftPos.computeIfAbsent(leftShuffleColumns.get(i), k -> new ArrayDeque<>()).add(i);
}
Map<DistributionCol, Deque<Integer>> rightPos = new HashMap<>();
for (int i = 0; i < rightShuffleColumns.size(); i++) {
    rightPos.computeIfAbsent(rightShuffleColumns.get(i), k -> new ArrayDeque<>()).add(i);
}

for (DistributionCol cid : requiredColumns) {
    int idx = adjustBasedOnLeft ? leftPos.get(cid).removeFirst() : rightPos.get(cid).removeFirst();
    requiredLeft.add(leftShuffleColumns.get(idx));
    requiredRight.add(rightShuffleColumns.get(idx));
}

```

PR #68142, opened 2026-01-20
[BugFix] Prevent aggregate pushdown when group keys involve JSON type

stephen-shelby:
## Why I'm doing:

Aggregate pushdown may rewrite DISTINCT/group-by expressions to their used columns when pushing below Project/Join.  
For expressions that depend on a JSON column, this can end up pushing down an aggregate that **groups by a JSON column**, which is not supported in BE and fails at runtime with:
`group by type JSON is not supported backend`.

## What I'm doing:

- Add a safety check in `PushDownAggregateCollector` to block aggregate pushdown when any group-key used column type cannot be grouped in BE (e.g. JSON).

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/pdagg/PushDownAggregateCollector.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/pdagg/PushDownAggregateCollector.java
@@ -701,6 +701,13 @@ public class PushDownAggregateCollector extends OptExpressionVisitor<Void, Aggre
         ColumnRefSet allGroupByColumns = new ColumnRefSet();
         context.groupBys.values().forEach(c -> allGroupByColumns.union(c.getUsedColumns()));
 
+        for (int colId : allGroupByColumns.getColumnIds()) {
+            ColumnRefOperator colRef = factory.getColumnRef(colId);
+            if (colRef.getType() != null && !colRef.getType().canGroupBy()) {
+                return false;
+            }
+        }
+
         ColumnRefSet allAggregateColumns = new ColumnRefSet();
         context.aggregations.values().forEach(c -> allAggregateColumns.union(c.getUsedColumns()));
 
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/pdagg/PushDownAggregateCollector.java
```diff
@@ -701,6 +701,13 @@ private boolean canPushDown(OptExpression optExpression, AggregatePushDownContex
         ColumnRefSet allGroupByColumns = new ColumnRefSet();
         context.groupBys.values().forEach(c -> allGroupByColumns.union(c.getUsedColumns()));
 
+        for (int colId : allGroupByColumns.getColumnIds()) {
```
kangkaisen, line 704: The group by result expr result couldn't be json, but the group by  expr used column could be json.

stephen-shelby: yes, I agree that in SQL semantics the **GROUP BY key is the result of the expression** (e.g. `CAST(json_query(...) AS VARCHAR)`), and that result type itself is not JSON.  

However, with the current aggregate pushdown implementation, **GROUP BY expressions are not pushed down as expressions**. When the rule rewrites through `Project`, it eventually **rebuilds pushdown group keys from `expr.getUsedColumns()` (column refs only)**, which can turn `GROUP BY CAST(json_query(json_col) AS VARCHAR)` into pushing down `GROUP BY json_col` (JSON) or other used columns. and if the group-by is an expression it would require additional support to be safely pushed down.  
  
Therefore, the check in `canPushDown()` is to prevent generating a pushdown plan that ends up grouping by unsupported types on BE (like JSON).

kangkaisen (approved):

Seaven (approved):

PR #68170, opened 2026-01-20
[Enhancement] Implement metrics reporting for `MemTrackerManager`

arin-mirza:
## Why I'm doing:

There are currently no backend metrics reporting for memory pools.

I previously tried to add them by extending the workgroup metrics, but this turned out to be an incorrect approach: 
- https://github.com/StarRocks/starrocks/pull/67701

## What I'm doing:

This PR implements metric reporting for `MemTrackerManager` and adds the following new metrics:
- mem_pool_mem_limit_bytes
- mem_pool_mem_usage_bytes
- mem_pool_mem_usage_ratio 
- mem_pool_workgroup_count

The implementation follows the same locking structure that is present in `WorkGroupManager`. 

- It was necessary to add a new mutex for `MemTrackerManager` because the update_metrics callback hook passed to `MetricRegistry` needs to be a closure which captures a write lock.
- The unlocked gap inside add_metrics method is unavoidable to AB-BA deadlock scenario with the metrics collector.
- Metrics entries are never deleted as it would complicate the thread synchronization even further. This is also the case for the existing implementation in `WorkGroupManager`.

Minor: Changed `list_mem_trackers()` method to not return the default memory pool name.

## Tests and Docs
- I did **not** add any test cases as there were not any for workgroup metrics either. Let me know if this is necessary.
- I did **not** verify that the new metrics are being reported correctly by building and running the starrocks fe/be, as I am currently unable to build the engine locally. 
- I updated the user documentation.
  - I am not a Chinese or Japanese speaker so I used AI for the translation. I would appreciate it if a native speaker could review my additions to ensure the tone is correct. :)

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


StarRocks-Reviewer:
@cursor review

be/src/exec/workgroup/mem_tracker_manager.cpp
```diff
@@ -74,13 +79,100 @@ void MemTrackerManager::deregister_workgroup(const std::string& mem_pool) {
...
+            metrics->mem_limit->set_value(mem_tracker->limit());
+            metrics->mem_usage_bytes->set_value(mem_tracker->consumption());
+            metrics->mem_usage_ratio->set_value(divide(mem_tracker->consumption(), mem_tracker->limit()));
+            metrics->workgroup_count->set_value(child_count);
+        } else {
+            // Metrics entries for deleted shared_mem_trackers are never deleted, but simply set to 0.
+            metrics->mem_limit->set_value(0);
+            metrics->mem_usage_bytes->set_value(0);
+            metrics->mem_usage_ratio->set_value(0);
+            metrics->workgroup_count->set_value(0);
```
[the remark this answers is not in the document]
arin-mirza: Good catch, fixed in f7a13d9f6d711be548dcde7948586c45dd72f179.
The same issue exists in the update method of `WorkGroupManager` so it should be fixed there too. I will try to produce the segfault in a test and then open a new PR for the fix.

**Details** 
- We cannot solve the problem inside `_add_metrics_unlocked` method.
  - Any thread which successfully registered at least one metric needs run to completion and initialize those metrics with the gauges. If the thread returns prematurely, those metrics will never get initialized.
- The solution was to add if guards to `_update_metrics_unlocked` method, which prevents nullptr dereferencing. 
  - All metrics are registered and initialized by some thread, so we eventually have a complete metrics object.
  - No segfaults can occur during the period where the metrics object is only partially initialized.

be/src/exec/workgroup/mem_tracker_manager.h
```diff
@@ -45,7 +52,22 @@ struct MemTrackerManager {
...
     void deregister_workgroup(const std::string& mem_pool);
 
+    /**
+     * Returns a list of currently registered non-default memory trackers.
+     */
+    std::vector<std::string> list_mem_trackers() const;
+
 private:
+    using MutexType = std::shared_mutex;
+    using UniqueLockType = std::unique_lock<MutexType>;
```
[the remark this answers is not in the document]
arin-mirza: Fixed in 3d4792ac9f6eff5f63dfcf54f9c9fd52b2d75a9f

docs/en/administration/management/monitoring/metrics.md
```diff
@@ -1627,6 +1627,26 @@ For more information on how to build a monitoring service for your StarRocks clu
 - Unit: -
 - Description: Ratio of internal table scan thread time slices used by each resource group to the total used by all resource groups. This is an average value over the time interval between two metric retrievals.
 
+### mem_pool_mem_limit_bytes
```
kevincai: the actual metric name will be prefixed by `starrocks_be_`, so to end user it is actually `starrocks_be_mem_pool_mem_limit_bytes`, shall use the final name since this is the doc to end user.

be/src/exec/workgroup/mem_tracker_manager.cpp
```diff
@@ -74,13 +79,116 @@ void MemTrackerManager::deregister_workgroup(const std::string& mem_pool) {
...
+    if (workgroup_count_registered) {
+        metrics->workgroup_count = std::move(workgroup_count);
+    }
+
+    // Note: It is also theoretically possible that the mem_tracker was deleted during the unlocked period.
+    // We are ok with leaking a metric in this case and do not try to unregister.
+}
+
+void MemTrackerManager::_update_metrics() {
+    std::unique_lock write_lock(_mutex);
```
kevincai: this can be a shared lock, since it only update the value of the metric, doesn't modify the two unordered map, right?

be/src/exec/workgroup/mem_tracker_manager.cpp
```diff
@@ -74,13 +79,116 @@ void MemTrackerManager::deregister_workgroup(const std::string& mem_pool) {
...
+                metrics->mem_usage_bytes->set_value(mem_tracker->consumption());
+            }
+            if (metrics->mem_usage_ratio) {
+                metrics->mem_usage_ratio->set_value(divide(mem_tracker->consumption(), mem_tracker->limit()));
+            }
+            if (metrics->workgroup_count) {
+                metrics->workgroup_count->set_value(child_count);
+            }
+        } else {
+            // Metrics entries for deleted shared_mem_trackers are never deleted, but simply set to 0.
```
kevincai, line 171: in a long running system with Frequent workergroup creation and deletion, will thes garbage metrics accumulated and cause memory occupation and long useless serialization to the /metrics interface.

arin-mirza: Not really.

Workgroups by default belong to `DEFAULT_MEM_POOL` , these are not tracked in `MemTrackerManager` and it does not report mempool statistics for such work groups. Frequent creation and deletion of workgroups under a memory pool is a special case.

In the general case, we are already paying the price of not deleting the metrics for workgroups. Compare my implementation with `_update_metrics_unlocked()` method in `work_group.cpp`: https://github.com/StarRocks/starrocks/blob/cd9ba3245de35cb77717e5d8fff5f3982a815be3/be/src/exec/workgroup/work_group.cpp#L475-L490

Your concern of garbage metrics being accumulated applies here even more. I believe the reason it was implemented this way is that deleting metrics entries make handling race conditions more complicated and error-prone.

I can change my implementation so that unused metrics are deleted instead of being set to 0. However, I am not sure if it is worth the extra effort as `work_group.cpp` does not delete its own metrics anyway.

Let me know which one you prefer.

be/src/exec/workgroup/mem_tracker_manager.cpp
arin-mirza: Yes, this should've been a shared lock. Fixed in commit e0aa5a8362ec3fed166e61a2acf08bbc576b4ead

docs/en/administration/management/monitoring/metrics.md
arin-mirza: I updated the documentation to use `starrocks_be_` prefix.

---

I think the documentation is inconsistent in this regard. 

Most **resource group** related metrics are duplicated, a description exists with and without the `starrocks_be` prefix. I also checked Datadog metrics to see if they are actually reported twice, but no, they are always reported under a name with the `starrocks_be` prefix.

In [metrics.md](https://github.com/StarRocks/starrocks/blob/cd9ba3245de35cb77717e5d8fff5f3982a815be3/docs/en/administration/management/monitoring/metrics.md): 

| Without Prefix | With Prefix | in Datadog |
| --- | --- | --- |
| resource_group_mem_limit_bytes | starrocks_be_resource_group_mem_limit_bytes | with prefix |
| resource_group_mem_inuse_bytes | - | with prefix |
| resource_group_cpu_limit_ratio | starrocks_be_resource_group_cpu_limit_ratio | with prefix |
| resource_group_cpu_use_ratio | starrocks_be_resource_group_cpu_use_ratio | with prefix |
| resource_group_scan_use_ratio | - | with prefix |
| resource_group_inuse_cpu_cores | - | with prefix |
| resource_group_connector_scan_use_ratio | - | with prefix |
| - | starrocks_be_resource_group_mem_allocated_bytes | does not exist |

I believe all metrics above should only have one entry with the `starrocks_be` prefix.

Moreover, I could not find `starrocks_be_resource_group_mem_allocated_bytes` in Datadog. This metric was renamed to `resource_group_mem_inuse_bytes` in 6204611d5846963c97a39486fcf9d1466b5418fb on 2022-06-29. The entry should be removed from the user documentation if safe to do so.

kevincai: Yeah, I think the doc is messed up somehow, the actual metrics name produced via `/metrics` should be the canonical one in this doc.

Just keep the new added ones prefixed with the `starrocks_be_`, other inconsistent ones will be fixed in a dedicated PR.

be/src/exec/workgroup/mem_tracker_manager.cpp
kevincai: ok. will see if other reviewers have some thought. I am ok to keep as is for now.

conf/be.conf
```diff
@@ -58,4 +58,5 @@ starlet_port = 9070
 # eg:
 # JAVA_OPTS="-Djava.security.krb5.conf=/etc/krb5.conf"
 # used for query iceberg metadata and Java UDF for JDK17
-JAVA_OPTS="--add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED"
\ No newline at end of file
+JAVA_OPTS="--add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED"
+priority_networks = 127.0.0.1/24
```
kevincai: these conf/be.conf and conf/fe.conf changes should not be included.

docs/zh/administration/management/monitoring/metrics.md
```diff
@@ -1627,6 +1651,26 @@ displayed_sidebar: docs
 - 单位：-
 - 描述：每个资源组的内部表扫描线程时间片与所有资源组总使用时间片的比值（两次统计之间的平均值）。
 
+### mem_pool_mem_limit_bytes
```
kevincai: remove these repeated ones.

docs/zh/administration/management/monitoring/metrics.md
```diff
@@ -579,6 +579,30 @@ displayed_sidebar: docs
 - 类型：瞬时值
 - 描述：该资源组内存使用率瞬时值
 
+### starrocks_be_mem_pool_mem_limit_bytes
+
+- 単位: Bytes
+- タイプ: Instantaneous
+- 説明: メモリプールのメモリクォータの瞬時値。
```
kevincai: hmm, why ja translation in zh doc.

docs/ja/administration/management/monitoring/metrics.md
```diff
@@ -579,6 +579,30 @@ StarRocks クラスタのモニタリングサービスの構築方法につい
 - タイプ: Instantaneous
 - 説明: リソースグループのメモリ使用量の瞬時値。
 
+### starrocks_be_mem_pool_mem_limit_bytes
+
+- 单位：Byte
+- 类型：瞬时值
+- 描述：该内存池内存配额的瞬时值
```
kevincai: and zh translation in ja doc

docs/zh/administration/management/monitoring/metrics.md
arin-mirza: My bad, I mixed up the files while copy-pasting :sweat_smile: Fixed.

conf/be.conf
arin-mirza: Accidentally pushed these, fixed.

docs/ja/administration/management/monitoring/metrics.md
arin-mirza: Fixed the documentation and removed the accidentally pushed `conf` files

kangkaisen (dismissed):

arin-mirza:
@kevincai @kangkaisen Sorry I accidentally requested re-reviews while trying to notify you.
I rebased the branch, CI passes fully. Can you please approve again? 
It would be nice to merge this before the branch goes out-of-date with the main again.

kevincai (approved):

kangkaisen (approved):

PR #68171, opened 2026-01-20
[Enhancement] prevent mv holding large external tables

Wenjun7J:
## Why I'm doing:
MaterializeView will use MVPlanCacheBuilder to cache logical plan, the logical plan may hold the lake table with large amounts metadata. The logical plan is used to do mv rewrite, and holding the large memory of metadata are useless most of the time.

So does the ast defined query node kept in MaterializedView, it may be cached when the MaterializedView is cache key in mv cache. The lake table hold in ast tree is aslo useless.
## What I'm doing:
Remove these heavy memory node to lower the cache memory usage under mv on lake.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/DeltaLakeTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/DeltaLakeTable.java
@@ -197,6 +197,10 @@ public class DeltaLakeTable extends Table {
         return Joiner.on(":").join(tableName, uuid == null ? "" : uuid);
     }
 
+    public MetastoreTable getMetastoreTable() {
+        return metastoreTable;
+    }
+
     @Override
     public int hashCode() {
         return com.google.common.base.Objects.hashCode(getCatalogName(), dbName, getTableIdentifier());
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/IcebergTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/IcebergTable.java
@@ -275,6 +275,7 @@ public class IcebergTable extends Table {
     public int getFormatVersion() {
         return ((BaseTable) getNativeTable()).operations().current().formatVersion();
     }
+
     /**
      * <p>
      * In the Iceberg Partition Evolution scenario, 'org.apache.iceberg.PartitionField#name' only represents the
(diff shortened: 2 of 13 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/LightIcebergTable.java
```diff
@@ -0,0 +1,73 @@
...
+
+package com.starrocks.catalog;
+
+import com.google.common.base.Joiner;
+import com.starrocks.catalog.IcebergTable;
+import org.apache.iceberg.BaseTable;
+
+import java.util.List;
+
+public class LightIcebergTable extends IcebergTable {
```
alvin-phoenix-ai: Please provide more comments on this class, including when and why we need it.

fe/fe-core/src/main/java/com/starrocks/catalog/LightIcebergTable.java
```diff
@@ -0,0 +1,73 @@
...
+
+    String uuid;
```
alvin-phoenix-ai:
```suggestion
    private String uuid;
```

fe/fe-core/src/main/java/com/starrocks/catalog/LightIcebergTable.java
alvin-phoenix-ai:
```suggestion
public class LightweightIcebergTable extends IcebergTable {
```
 It is better to use Lightweight

LiShuMing (approved):
LGTM

dirtysalt (approved):

Youngwb:
Why not implement a LightweightTable class that includes the required attributes such as catalog, db, and name, and then support a utility class to complete the conversion of all external tables to LightweightTable?
 In this way, all conversion logic can be unified in one place, instead of using instanceof for type checks of IcebergTable and DeltaLakeTable everywhere as it is currently done.

Youngwb (approved):

Seaven (approved):

stephen-shelby (approved):

HangyuanLiu (approved):

PR #68191, opened 2026-01-21
[Enhancement] Support Big-endian encoding for primary key in range-distribution table in share data mode.

srlch:
## Why I'm doing:
In current implementation, we always encode the primary key without BIG_ENDIAN transformation if the there is only one pk column and it is not string type. Such encoding method can not perserve the key order after encoding which is bad for range-distribution table in share data mode.

## What I'm doing:
We use PrimaryKeyEncodingType in tabletschema to store the encoding type.
NONE: for non-primary key tablet for the compatibility issue and should not be used in any primary key encode context.
V1: encoding as the previous way.
v2: always encoding as big-endian way,
Currently, we only support V2 encoding type for range-distribution table in share data mode.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/pk_tablet_sst_writer.cpp
+++ b/be/src/storage/lake/pk_tablet_sst_writer.cpp
@@ -37,17 +37,19 @@ Status PkTabletSSTWriter::append_sst_record(const Chunk& data) {
     if (_pk_sst_builder == nullptr) {
         return Status::InternalError("pk sst writer not initialized");
     }
+    ASSIGN_OR_RETURN(auto pk_encoding_type, _tablet_schema_ptr->primary_key_encoding_type_or_error());
     if (_pk_column == nullptr) {
         vector<uint32_t> pk_columns;
         for (size_t i = 0; i < _tablet_schema_ptr->num_key_columns(); i++) {
             pk_columns.push_back((uint32_t)i);
         }
         _pkey_schema = ChunkHelper::convert_schema(_tablet_schema_ptr, pk_columns);
-        RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(_pkey_schema, &_pk_column));
-        _key_size = PrimaryKeyEncoder::get_encoded_fixed_size(_pkey_schema);
+        RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(_pkey_schema, &_pk_column, pk_encoding_type));
+        _key_size = PrimaryKeyEncoder::get_encoded_fixed_size(_pkey_schema, pk_encoding_type);
     }
     auto clone_pk_column = _pk_column->clone_empty();
-    TRY_CATCH_BAD_ALLOC(PrimaryKeyEncoder::encode(_pkey_schema, data, 0, data.num_rows(), clone_pk_column.get()));
+    TRY_CATCH_BAD_ALLOC(
+            PrimaryKeyEncoder::encode(_pkey_schema, data, 0, data.num_rows(), clone_pk_column.get(), pk_encoding_type));
     std::vector<Slice> keys;
     const Slice* vkeys =
             PrimaryIndex::build_persistent_keys(*clone_pk_column, _key_size, 0, clone_pk_column->size(), &keys);
--- a/be/src/storage/primary_key_compaction_conflict_resolver.cpp
+++ b/be/src/storage/primary_key_compaction_conflict_resolver.cpp
@@ -27,9 +27,10 @@ namespace starrocks {
 
 Status PrimaryKeyCompactionConflictResolver::execute() {
     Schema pkey_schema = generate_pkey_schema();
+    ASSIGN_OR_RETURN(auto encoding_type, primary_key_encoding_type());
 
     MutableColumnPtr pk_column;
-    RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(pkey_schema, &pk_column, true));
+    RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(pkey_schema, &pk_column, encoding_type, true));
 
     // init rows mapper iter
     ASSIGN_OR_RETURN(auto filename, filename());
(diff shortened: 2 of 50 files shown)


StarRocks-Reviewer:
@cursor review

be/src/storage/sstable/sstable_predicate_utils.cpp
```diff
@@ -31,7 +31,10 @@ StatusOr<KeyToChunkConverterUPtr> KeyToChunkConverter::create(const TabletSchema
     auto pkey_schema = ChunkHelper::convert_schema(tablet_schema, pk_columns);
     auto pk_chunk = ChunkHelper::new_chunk(pkey_schema, 1);
     MutableColumnPtr pk_column;
-    RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(pkey_schema, &pk_column));
+    // use V1 encoding type to decode the column for simplicity
+    // TODO: Should use V2 encoding type to decode the column for range-distribution table in share data mode.
+    // But for now, such table do not use sst predicate at all. Support it if necessary in the future.
+    RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(pkey_schema, &pk_column, PrimaryKeyEncodingType::V1));
```
luohaha: sstable_predicate is useless now?

srlch: Yes it is. Such predicate was introduced for the previous version of tablet split/ merge and the sst predicate has not been used for now. Maybe i can remove the the code which create the sst predicate and make it safer.

srlch: Updated

StarRocks-Reviewer:
@cursor review

be/src/storage/lake/pk_tablet_sst_writer.cpp
```diff
@@ -37,17 +37,20 @@ Status PkTabletSSTWriter::append_sst_record(const Chunk& data) {
     if (_pk_sst_builder == nullptr) {
         return Status::InternalError("pk sst writer not initialized");
     }
+    ASSIGN_OR_RETURN(auto tablet, _tablet_mgr->get_tablet(_tablet_id));
```
luohaha: This code will get tablet each time when call `append_sst_record`. It's inefficient.

luohaha (dismissed):

be/src/storage/primary_key_compaction_conflict_resolver.cpp
```diff
@@ -29,7 +29,8 @@ Status PrimaryKeyCompactionConflictResolver::execute() {
     Schema pkey_schema = generate_pkey_schema();
 
     MutableColumnPtr pk_column;
-    RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(pkey_schema, &pk_column, true));
+    RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(pkey_schema, &pk_column,
```
luohaha: It could be used by cloud native pk table too.

be/src/storage/primary_key_recover.cpp
```diff
@@ -28,7 +28,7 @@ Status PrimaryKeyRecover::recover() {
     // 2. generate primary key schema
     MutableColumnPtr pk_column;
     auto pkey_schema = generate_pkey_schema();
-    if (!PrimaryKeyEncoder::create_column(pkey_schema, &pk_column).ok()) {
+    if (!PrimaryKeyEncoder::create_column(pkey_schema, &pk_column, PrimaryKeyEncodingType::PK_ENCODING_TYPE_V1).ok()) {
```
luohaha: Cloud native pk table will use this class too.

be/src/storage/primary_key_compaction_conflict_resolver.cpp
srlch: Fixed

be/src/storage/primary_key_recover.cpp
srlch: Fixed

luohaha (approved):

be/src/storage/primary_key_encoder.h
```diff
@@ -106,42 +108,73 @@ void encode_slice(const Slice& s, std::string* dst, bool is_last);
...
+        case PrimaryKeyEncodingTypePB::PK_ENCODING_TYPE_V1:
+            return PrimaryKeyEncodingType::PK_ENCODING_TYPE_V1;
+        case PrimaryKeyEncodingTypePB::PK_ENCODING_TYPE_V2:
+            return PrimaryKeyEncodingType::PK_ENCODING_TYPE_V2;
+        default:
+            return PrimaryKeyEncodingType::PK_ENCODING_TYPE_NONE;
+        }
+    }
+
+    static PrimaryKeyEncodingTypePB pb_from_encoding_type(PrimaryKeyEncodingType encoding_type) {
```
wyb, line 124:
```suggestion
    static PrimaryKeyEncodingTypePB encoding_type_to_pb(PrimaryKeyEncodingType encoding_type) {
```

be/src/storage/tablet_schema.h
```diff
@@ -425,7 +432,7 @@ class TabletSchema {
     mutable std::once_flag _init_schema_once_flag;
     int32_t _schema_version = -1;
 
-    PrimaryKeyEncodingTypePB _primary_key_encoding_type = PrimaryKeyEncodingTypePB::PK_ENCODING_TYPE_NONE;
+    PrimaryKeyEncodingType _primary_key_encoding_type = PrimaryKeyEncodingType::PK_ENCODING_TYPE_NONE;
```
wyb, line 435: Why not use PB directly?

wyb (approved):

Issue #64340, opened 2025-10-21
Wrongfully unescaped partitions values in hive metastore


scSenicourt:
This is a bug that was introduced https://github.com/StarRocks/starrocks/pull/59032. If I understand this correctly, the bug that this PR was claiming to fix never existed. Additionally the examples mentioned in the PR are not correct, since partition names are not the same as storage locations because in partition names you only need to escape characters like `:`, but not whitespace. Hive metastore expects partition names in the get_partitions_by_names function, not storage locations.

### Steps to reproduce the behavior (Required)
```sql
create external catalog hive_catalog PROPERTIES ("type"="hive", <your-hive-metastore>);
create database hive_catalog.my_db properties (
    "location" = "s3a://my-s3-bucket.com/example-prefix"
);
CREATE TABLE hive_catalog.my_db.broken_partition_name (
       column_a int,
       column_b varchar(16))
       partition by(column_b);


-- This will not require any escaping in the HMS anyways, partition names are NOT the same thing as partition locations. That's why the test succeeded for https://github.com/StarRocks/starrocks/pull/59032
-- Partition locations will only escape characters like `:`.
insert into hive_catalog.my_db.broken_partition_name values(1, "4-NOT SPECI");

-- This will actually break
insert into hive_catalog.my_db.broken_partition_name values(2, "broken:partition");

```
Until here everything still succeeds. But now if you restart the starrocks frontend and query

```
select * from broken_partition_name;
```

### Expected behavior (Required)

The query returns:

```
+-------------+------------------------+
|       column_a | colum_b    |
+-------------+------------------------+
|           1 | 4-NOT SPECI        |
|           2 | broken:partition |
+-------------+------------------------+
```

### Real behavior (Required)

The select query will fail with: `ERROR 1064 (HY000): com.google.common.util.concurrent.UncheckedExecutionException: java.lang.NullPointerException: null value in entry: column_b=broken:partition=null`

### StarRocks version (4.0.0)
 - You can get the StarRocks version by executing SQL `select current_version()`

### Analysis
The reason why this issue never occurs when using a

lfdversluis:
We are running in this one too. Happens when we partition on a timestamp, the ":" in the time is escaped, not matching the folder name on HDFS anymore when using hive meta store.

duanyyyyyyy:
@scSenicourt @lfdversluis can try to backport my patch to fix your issue.

lfdversluis:
Commenting to avoid the staleness! Any update on the fix?

PR #68221, opened 2026-01-21
[BugFix] Fix a bug that some kind of partitions can not be written

duanyyyyyyy:
## Why I'm doing:
Fix the bug that mentioned in https://github.com/StarRocks/starrocks/issues/64340

When write the partition with ':' twice will have the following issue.
image
The partition name we used to get the partition from HMS was unescaped to 'lo_orderpriority=4-NOT SPECI:13123'
But we do not register it into HMS we will only register 'lo_orderpriority=4-NOT SPECI%3A213123' in HMS.
## What I'm doing:
Remove the unescape in HiveMetaClient and regenerate partitionName in PartitionUpdate.java
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/hive/HiveMetaClient.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/hive/HiveMetaClient.java
@@ -45,14 +45,12 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 import static com.starrocks.common.profile.Tracers.Module.EXTERNAL;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_CONNECTION_POOL_SIZE;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_TIMEOUT;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_TYPE;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_URIS;
-import static org.apache.hadoop.hive.common.FileUtils.unescapePathName;
 
 public class HiveMetaClient {
     private static final Logger LOG = LogManager.getLogger(HiveMetaClient.class);
@@ -317,10 +315,8 @@ public class HiveMetaClient {
             RecyclableClient client = null;
             StarRocksConnectorException connectionException = null;
             try {
-                List<String> decodedPartitionNames = partitionNames.stream().
-                        map(name -> unescapePathName(name)).collect(Collectors.toList());
                 client = getClient();
-                partitions = client.hiveClient.getPartitionsByNames(dbName, tblName, decodedPartitionNames);
+                partitions = client.hiveClient.getPartitionsByNames(dbName, tblName, partitionNames);
                 if (partitions.size() != partitionNames.size()) {
                     LOG.warn("Expect to fetch {} partition on [{}.{}], but actually fetched {} partition",
                             partitionNames.size(), dbName, tblName, partitions.size());
--- a/fe/fe-core/src/main/java/com/starrocks/connector/hive/PartitionUpdate.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/hive/PartitionUpdate.java
@@ -97,17 +97,20 @@ public class PartitionUpdate {
         OVERWRITE,
     }
 
-    public static PartitionUpdate get(THiveFileInfo fileInfo, String stagingDir, String tableLocation) {
+    public static PartitionUpdate get(THiveFileInfo fileInfo, String stagingDir,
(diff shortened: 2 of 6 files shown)


duanyyyyyyy:
@bugbot run

duanyyyyyyy:
@alvin-celerdata can you help start the CI

fe/fe-core/src/main/java/com/starrocks/connector/hive/HiveMetaClient.java
```diff
@@ -317,10 +315,8 @@ public List<Partition> getPartitionsByNames(String dbName, String tblName, List<
             RecyclableClient client = null;
             StarRocksConnectorException connectionException = null;
             try {
-                List<String> decodedPartitionNames = partitionNames.stream().
```
dirtysalt, line 320: @Wenjun7J does this revert your change? 

@duanyyyyyyy why we don't need unescape pathe names here?

dirtysalt: I see your explanation in issue.

Wenjun7J: now my change logic will apply in `PartitionUtil.toPartitionValues` during committing.

fe/fe-core/src/main/java/com/starrocks/connector/hive/PartitionUpdate.java
```diff
@@ -97,17 +97,20 @@ public enum UpdateMode {
...
+        List<String> partitionValues = PartitionUtil.toPartitionValues(partitionNameFromPath);
+        String partitionName = PartitionUtil.toHivePartitionName(partitionColNames, partitionValues);
 
         String tableLocationWithSlash = getPathWithSlash(tableLocation);
         String stagingDirWithSlash = getPathWithSlash(stagingDir);
 
-        String writePath = stagingDirWithSlash + partitionName + "/";
-        String targetPath = tableLocationWithSlash + partitionName + "/";
+        String writePath = stagingDirWithSlash + partitionNameFromPath + "/";
+        String targetPath = tableLocationWithSlash + partitionNameFromPath + "/";
```
Wenjun7J, line 113: Will the output path be different when using Hive？

fe/fe-core/src/main/java/com/starrocks/connector/hive/HiveMetaClient.java
Wenjun7J: The logic is moved from reading to writing.

duanyyyyyyy: Yes the logic is moved from reading to writing.

Because for the reading part, we just need to use the partitionNames that we can from the HMS, no need to do some unescape.

fe/fe-core/src/main/java/com/starrocks/connector/hive/PartitionUpdate.java
duanyyyyyyy: Actually speaking here also will have a bug as the for example some external table will have partition path not generate from the partition names.

For example for some hive external table the partition name can be k1=123 but the path can be some other path.

But I do not think it is related to this issue. We can open some other patch for those cases.

duanyyyyyyy: Oh you mean the hive partition name.

Yes it will be different as partitionNameFromPath will be escaped that can be store in HDFS but as you can see for the partition name like 'c1=a%20b' but we need to use the 'c1=a b'.

But for the last version if the partition have ':', and if we do the unescape when we are reading, the ':' will be in unescaped mode and we can not get the partition from HMS.

Wenjun7J: In my perspective, hms will keep the 'a b' as the key value, and the partition path should be 'a%20b'.
But for "a:b", it should be kept as 'a%3Ab' in hms, so as the path. so my pr introduced the problem for the ":" 
So your fix will make hms call using 'a b' and 'a%3Ab' instead of 'a%20b' and 'a:b', it will fix the both problems, am I correct?

duanyyyyyyy: yes your understand is correct

Wenjun7J:
LGTM

Wenjun7J (approved):

dirtysalt (approved):

Youngwb (approved):

Issue #68184, opened 2026-01-21
[Bug] Seg Fault caused by enabling `cbo_json_v2_dict_opt`


brent-statsig:
### Steps to reproduce the behavior (Required)
I have a JSON column with flat_json enabled. It is unclear how to repro this.

Stack:
```
*** SIGSEGV (@0x3) received by PID 27 (TID 0x7dbe405d4640) LWP(852) from PID 3; stack trace: ***
    @     0x7dbf41043ee8 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x99ee7)
    @         0x11a38c48 google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @     0x7dbf421e8706 PosixSignals::chained_handler(int, siginfo_t*, void*) [clone .part.0]
    @     0x7dbf421e917e JVM_handle_linux_signal
    @     0x7dbf40fec520 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x4251f)
    @          0xd36c865 starrocks::SegmentIterator::_switch_context(starrocks::SegmentIterator::ScanContext*)
    @          0xd37e14b starrocks::SegmentIterator::_init_context()
    @          0xd37e99e starrocks::SegmentIterator::_init()
    @          0xd387573 starrocks::SegmentIterator::do_get_next(starrocks::Chunk*)
    @          0xd93e5a8 starrocks::SegmentIteratorWrapper::do_get_next(starrocks::Chunk*)
    @          0xd6428dc starrocks::TimedChunkIterator::do_get_next(starrocks::Chunk*)
    @          0xd6754c0 starrocks::UnionIterator::do_get_next(starrocks::Chunk*)
    @          0xd62f46f starrocks::TabletReader::do_get_next(starrocks::Chunk*)
    @          0xc3c12de starrocks::pipeline::OlapChunkSource::_read_chunk_from_storage(starrocks::RuntimeState*, starrocks::Chunk*)
    @          0xc3c1dcf starrocks::pipeline::OlapChunkSource::_read_chunk(starrocks::RuntimeState*, std::shared_ptr<starrocks::Chunk>*)
    @          0xbe7eb6f starrocks::pipeline::ChunkSource::buffer_next_batch_chunks_blocking(starrocks::RuntimeState*, unsigned long, starrocks::workgroup::WorkGroup const*)
[...]

PR #68279, opened 2026-01-22
[BugFix] Fix duplicate column unique_id conflict for flat JSON extended columns

happycastle114:
## Summary

Fix a bug where flat JSON extended columns (subfields like `source.page`) can get duplicate `unique_id` values that conflict with existing regular columns, causing the error:

```
Duplicate column id=101 found between column 'tags_new' and column 'source.page' in tablet schema
```

## Root Cause

In `lake_meta_reader.cpp`, when querying with flat JSON subfields, extended columns are dynamically added to the tablet schema. These columns were assigned `unique_id` starting from `num_columns()`:

```cpp
int field_number = tmp_schema->num_columns();  // BUG
column.set_unique_id(++field_number);
```

However, `num_columns()` returns the current column count, while `unique_id` values can have gaps after ADD/DROP COLUMN operations (unique_ids are never reused). This causes extended columns to potentially receive the same `unique_id` as existing columns.

**Example scenario:**
| Operation | num_columns | next_column_unique_id | 
|-----------|-------------|----------------------|
| Create table (100 cols) | 100 | 100 |
| ADD COLUMN tags_new | 101 | 101 |
| DROP COLUMN old_col | 100 | 101 |
| Query flat JSON → extended col gets `unique_id = 101` | **CONFLICT** |

## Fix

close #68519, #68184

Use `next_column_unique_id()` instead of `num_columns()`, which correctly tracks the next available unique_id regardless of column count:

```cpp
int field_number = tmp_schema->next_column_unique_id();  // FIXED
```

## What type of PR is this?

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] Test
- [ ] Doc
- [ ] Build

## Testing

This fix addresses the duplicate column id error reported in production with StarRocks 4.0.x when using flat JSON feature with tables that have undergone schema changes (ADD/DROP COLUMN).

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/test/sql/test_semi/R/test_flat_json_schema_change
+++ b/test/sql/test_semi/R/test_flat_json_schema_change
@@ -1,0 +1,119 @@
+-- name: test_flat_json_after_schema_change @system
+CREATE DATABASE IF NOT EXISTS test_flat_json_schema_change;
+-- result:
+-- !result
+USE test_flat_json_schema_change;
+-- result:
+-- !result
+CREATE TABLE `t_json_schema_change` (
+  `id` bigint(20) NOT NULL,
+  `data` json NULL,
+  `col1` int NULL,
+  `col2` int NULL,
+  `col3` int NULL
+) ENGINE=OLAP
+DUPLICATE KEY(`id`)
+DISTRIBUTED BY HASH(`id`) BUCKETS 1
+PROPERTIES (
+"replication_num" = "1",
+"fast_schema_evolution" = "true"
+);
+-- result:
+-- !result
+INSERT INTO t_json_schema_change VALUES
+(1, parse_json('{"name": "alice", "age": 30, "city": "seattle"}'), 1, 2, 3),
+(2, parse_json('{"name": "bob", "age": 25, "city": "boston"}'), 4, 5, 6),
+(3, parse_json('{"name": "charlie", "age": 35, "city": "chicago"}'), 7, 8, 9);
+-- result:
+-- !result
+ALTER TABLE t_json_schema_change ADD COLUMN col4 int;
+-- result:
+-- !result
+ALTER TABLE t_json_schema_change ADD COLUMN col5 int;
+-- result:
+-- !result
+ALTER TABLE t_json_schema_change ADD COLUMN col6 int;
+-- result:
+-- !result
+ALTER TABLE t_json_schema_change DROP COLUMN col1;
+-- result:
+-- !result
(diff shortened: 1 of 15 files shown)


test/sql/test_semi/T/test_flat_json_schema_change
```diff
@@ -0,0 +1,64 @@
+-- name: test_flat_json_after_schema_change @system
```
kevincai:
```suggestion
-- name: test_flat_json_after_schema_change @cloud
```

or remove the tag so this case will be run for both shared-nothing cluster and shared-data cluster.

test/sql/test_semi/T/test_flat_json_schema_change
```diff
@@ -0,0 +1,64 @@
...
+-- Test flat JSON query after ADD/DROP COLUMN operations
+-- This tests the fix for duplicate column unique_id conflict
+-- where extended columns (flat JSON subfields) could get the same
+-- unique_id as existing columns after schema changes.
+
+CREATE DATABASE IF NOT EXISTS test_flat_json_schema_change;
```
kevincai:
```suggestion
CREATE DATABASE IF NOT EXISTS test_flat_json_${uuid0};
```

avoid database name conflict when repeatedly run the case or concurrent run the case.

kevincai (commented):
the comment from cursor is also valid.

test/sql/test_semi/T/test_flat_json_schema_change
```diff
@@ -0,0 +1,64 @@
+-- name: test_flat_json_after_schema_change @cloud
```
kevincai: I think this case can be run for both shared-nothing and shared-data, right?

happycastle114: yes

kevincai: so the @cloud annotation shall be removed.

test/sql/test_semi/R/test_flat_json_schema_change
```diff
@@ -0,0 +1,119 @@
+-- name: test_flat_json_after_schema_change @system
```
kevincai: this is not consistent with the T file. shall update the R file automatically with the `-r` option.

stdpain (approved):

PR #68300, opened 2026-01-22
[BugFix] Fix Arrow Flight Proxy with Multiple FE

chris-celerdata:
## Why I'm doing:
When the Arrow Flight proxy is directed towards a load balancer, requests from a single client may be routed to multiple FEs. This causes problems with authentication, as the token responsible for authentication is local to an FE. 
## What I'm doing:
- Add support for grpcs:// and grpc:// scheme prefixes in proxy configuration for TLS connections
- Include originating FE host:port in proxy tickets to enable routing back to the correct FE
- Forward bearer tokens when proxying requests to remote FE nodes for authentication
- Add local FE detection to avoid unnecessary network hops when proxy routes back to same node  
- Refactor FE endpoint and ticket building logic to match BE pattern
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/docs/en/unloading/arrow_flight.md
+++ b/docs/en/unloading/arrow_flight.md
@@ -176,6 +176,8 @@ SET GLOBAL arrow_flight_proxy = 'your-proxy-hostname:Port';
 
 - The proxy feature is enabled by default, which may result in 8-10% lower throughput compared to direct BE connections. If your clients have direct network access to BE nodes, you can disable the proxy to achieve optimal performance: `SET GLOBAL arrow_flight_proxy_enabled = false;`
 - When `arrow_flight_proxy` is empty, tickets will automatically route through the FE node that the client initially connected to.
+- **Important**: The `arrow_flight_proxy` and `arrow_flight_proxy_enabled` settings should be configured globally using `SET GLOBAL`. Session-level settings are not supported.
+- **Session restart required**: Changing the proxy settings only affects new sessions. Existing Arrow Flight SQL sessions will continue using their original settings until they reconnect.
 
 :::
 
--- a/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
@@ -505,6 +505,7 @@ public class ArrowFlightSqlServiceImpl implements FlightSqlProducer, AutoCloseab
     private void getStreamResultFromBE(String queryId, String fragmentInstanceId,
                                         String beHost, int bePort, ServerStreamListener listener) {
         ByteString ticketHandle = buildBETicket(queryId, fragmentInstanceId);
+        // no token needed for be connection
         getStreamResultFromRemoteNode(beHost, bePort, ticketHandle, null, "BE", listener);
     }
 
@@ -653,10 +654,12 @@ public class ArrowFlightSqlServiceImpl implements FlightSqlProducer, AutoCloseab
 
             // FE task will return FE (or proxy) as endpoint.
             if (!result.isBackendResultDescriptor()) {
-                final ByteString handle = buildFETicket(ctx);
+                Pair<Location, ByteString> feEndpoint = getFEEndpoint(ctx);
+                Location endpoint = feEndpoint.first;
+                ByteString handle = feEndpoint.second;
+
                 FlightSql.TicketStatementQuery ticketStatement =
                         FlightSql.TicketStatementQuery.newBuilder().setStatementHandle(handle).build();
-                Location endpoint = getFEEndpoint();
                 return buildFlightInfo(ticketStatement, descriptor, result.getSchema(), endpoint);
             }
 
@@ -681,18 +684,32 @@ public class ArrowFlightSqlServiceImpl implements FlightSqlProducer, AutoCloseab
         }
     }
 
-    private ByteString buildFETicket(ArrowFlightSqlConnectContext ctx) {
+    protected Pair<Location, ByteString> getFEEndpoint(ArrowFlightSqlConnectContext ctx) throws InvalidConfException {
+        ByteString handle;
+        Location endpoint;
         SessionVariable globalSv = getGlobalProxySessionVariable();
+
         if (globalSv.isArrowFlightProxyEnabled()) {
+            String arrowFlightProxy = globalSv.getArrowFlightProxy();
             // FE Proxy Ticket: <Token>|<QueryId>|<FEHost>:<FEPort>
             String feHost = feEndpoint.getUri().getHost();
             int fePort = feEndpoint.getUri().getPort();
-            return ByteString.copyFromUtf8(ctx.getArrowFlightSqlToken() + "|"
+            handle = ByteString.copyFromUtf8(ctx.getArrowFlightSqlToken() + "|"
                     + DebugUtil.printId(ctx.getExecutionId()) + "|"
                     + feHost + ":" + fePort);
+
+            if (!arrowFlightProxy.isEmpty()) {
+                endpoint = getProxyLocation(arrowFlightProxy);
+            } else {
+                endpoint = feEndpoint;
+            }
+        } else {
+            // FE Local Ticket: <Token>|<QueryId>
+            handle = ByteString.copyFromUtf8(ctx.getArrowFlightSqlToken() + "|" + DebugUtil.printId(ctx.getExecutionId()));
+            endpoint = feEndpoint;
         }
-        // FE Local Ticket: <Token>|<QueryId>
-        return ByteString.copyFromUtf8(ctx.getArrowFlightSqlToken() + "|" + DebugUtil.printId(ctx.getExecutionId()));
+
+        return new Pair<>(endpoint, handle);
     }
 
     private static ByteString buildBETicket(String queryId, String rootFragmentInstanceId) {
@@ -716,7 +733,7 @@ public class ArrowFlightSqlServiceImpl implements FlightSqlProducer, AutoCloseab
                                                                  Schema schema, CallContext context) {
         try {
             sessionManager.validateAndGetConnectContext(context.peerIdentity());
-            Location endpoint = getFEEndpoint();
+            Location endpoint = getFELocation();
             return buildFlightInfo(request, descriptor, schema, endpoint);
         } catch (InvalidConfException e) {
             throw CallStatus.INTERNAL.withDescription(e.getMessage()).toRuntimeException();
@@ -815,7 +832,7 @@ public class ArrowFlightSqlServiceImpl implements FlightSqlProducer, AutoCloseab
         return GlobalStateMgr.getCurrentState().getVariableMgr().getDefaultSessionVariable();
     }
 
-    protected Location getFEEndpoint() throws InvalidConfException {
+    protected Location getFELocation() throws InvalidConfException {
         SessionVariable globalSv = getGlobalProxySessionVariable();
         if (globalSv.isArrowFlightProxyEnabled()) {
             String arrowFlightProxy = globalSv.getArrowFlightProxy();
--- a/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
@@ -55,6 +55,8 @@ import org.apache.arrow.flight.SchemaResult;
 import org.apache.arrow.flight.SetSessionOptionsRequest;
 import org.apache.arrow.flight.SetSessionOptionsResult;
 import org.apache.arrow.flight.Ticket;
+import org.apache.arrow.flight.auth2.BearerCredentialWriter;
+import org.apache.arrow.flight.grpc.CredentialCallOption;
 import org.apache.arrow.flight.sql.FlightSqlProducer;
 import org.apache.arrow.flight.sql.SqlInfoBuilder;
 import org.apache.arrow.flight.sql.impl.FlightSql;
@@ -87,8 +89,8 @@ public class ArrowFlightSqlServiceImpl implements FlightSqlProducer, AutoCloseab
     private final ArrowFlightSqlSessionManager sessionManager;
     private final Location feEndpoint;
     private final SqlInfoBuilder sqlInfoBuilder;
-    private final Cache<String, FlightClient> beClientCache = CacheBuilder.newBuilder()
-            .maximumSize(128)
+    private final Cache<String, FlightClient> clientCache = CacheBuilder.newBuilder()
+            .maximumSize(256)
             .expireAfterAccess(10, TimeUnit.MINUTES)
             .removalListener((RemovalNotification<String, FlightClient> notification) -> {
                 FlightClient client = notification.getValue();
@@ -129,7 +131,7 @@ public class ArrowFlightSqlServiceImpl implements FlightSqlProducer, AutoCloseab
 
     @Override
     public void close() throws Exception {
-        beClientCache.invalidateAll();
+        clientCache.invalidateAll();
         AutoCloseables.close(rootAllocator);
     }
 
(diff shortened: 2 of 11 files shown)


docs/en/unloading/arrow_flight.md
```diff
@@ -176,6 +176,8 @@ SET GLOBAL arrow_flight_proxy = 'your-proxy-hostname:Port';
 
 - The proxy feature is enabled by default, which may result in 8-10% lower throughput compared to direct BE connections. If your clients have direct network access to BE nodes, you can disable the proxy to achieve optimal performance: `SET GLOBAL arrow_flight_proxy_enabled = false;`
 - When `arrow_flight_proxy` is empty, tickets will automatically route through the FE node that the client initially connected to.
+- **Important**: The `arrow_flight_proxy` and `arrow_flight_proxy_enabled` settings should be configured globally using `SET GLOBAL`. Session-level settings are not supported.
```
jaogoy, line 179: If there are global variables, why not declear them in GlobalVariable.java?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/auth2/ArrowFlightSqlAuthenticator.java
```diff
@@ -62,12 +64,23 @@ AuthResult validateBearerToken(String token) {
         try {
             sessionManager.validateToken(token);
         } catch (IllegalArgumentException e) {
+            // Token not found locally - if proxy is enabled, allow through.
+            // Validation will happen on the remote FE when we forward the request.
```
jaogoy: Does it mean that, even if it's an invalid token when doing the `getFlightInfo`, the FE will recoganize it as valid, and accept the sql, parse and execute the sql, return the endpoint?

Is it possible for a client that a client directly do a `getFlightInfo` without the first basic auth?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -473,100 +475,151 @@ protected static ByteBuffer serializeMetadata(final Schema schema) {
...
+                                                String bearerToken,
+                                                String targetType,
+                                                ServerStreamListener listener) {
+        FlightStream stream = null;
+        String nodeKey = host + ":" + port;
 
         try {
-            FlightClient beClient = getOrCreateBeClient(beKey, beHost, bePort);
-            // Reconstruct the original BE ticket (without BE host/port)
+            FlightClient client = getOrCreateClient(nodeKey, host, port);
```
jaogoy, line 587: Does it mean that an FE can fetch data from a BE with only one connection (a client) for different queries over http2/gPRC?
(Normally, it's OK too.)

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -473,100 +475,151 @@ protected static ByteBuffer serializeMetadata(final Schema schema) {
...
+                                ticketParts.length)).toRuntimeException();
         }
+    }
 
-        if (ticketParts.length == 2) {
-            getStreamResultFromFE(ticketParts[0], ticketParts[1], listener); 
-            return;
+    private void getStreamResultFromBE(String queryId, String fragmentInstanceId,
+                                        String beHost, int bePort, ServerStreamListener listener) {
+        ByteString ticketHandle = buildBETicket(queryId, fragmentInstanceId);
```
jaogoy: We need to add some desin info in the PR description.
And it'll be better to add some important comments.

Here, you can comment: `queryId` is used as a token amon FE, BE, and client for validation, because it's a random UUID.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -473,100 +475,151 @@ protected static ByteBuffer serializeMetadata(final Schema schema) {
...
+    private FlightStream getStreamWithRetry(FlightClient client, Ticket ticket,
+                                            String key, String host, int port,
+                                            String bearerToken) throws Exception {
         try {
-            return beClient.getStream(ticket);
+            if (bearerToken != null) {
+                CredentialCallOption authOption = new CredentialCallOption(new BearerCredentialWriter(bearerToken));
+                return client.getStream(ticket, authOption);
+            }
+            return client.getStream(ticket);
```
jaogoy, line 650: We need to add some important comments.
For example, here, `bearerToken == null` is only used when getting stream from remote BE,  becasue the ticket contains the token (queryId).

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -473,100 +475,151 @@ protected static ByteBuffer serializeMetadata(final Schema schema) {
...
         } catch (Exception e) {
-            LOG.warn("[ARROW] Failed to get stream from BE {}:{}, retrying with new client",
-                    beHost, bePort, e);
+            LOG.warn("[ARROW] Failed to get stream from {}:{}, retrying with new client", host, port, e);
 
-            beClientCache.invalidate(beKey);
-            FlightClient freshClient = getOrCreateBeClient(beKey, beHost, bePort);
+            clientCache.invalidate(key);
+            FlightClient freshClient = getOrCreateClient(key, host, port);
+            if (bearerToken != null) {
```
jaogoy, line 656: A little ungraceful: double code, without retry number.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -722,27 +806,51 @@ private static void validateProxyFormat(String arrowFlightProxy) throws InvalidC
...
     private Location getProxyLocation(String arrowFlightProxy) throws InvalidConfException {
         validateProxyFormat(arrowFlightProxy);
-        String[] split = arrowFlightProxy.split(":");
-        return Location.forGrpcInsecure(split[0], Integer.parseInt(split[1]));
+
+        boolean useTls = arrowFlightProxy.startsWith("grpcs://");
+        String hostPort = arrowFlightProxy;
+
+        // Strip scheme if present
+        if (arrowFlightProxy.startsWith("grpcs://")) {
```
jaogoy: useTls?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -722,27 +806,51 @@ private static void validateProxyFormat(String arrowFlightProxy) throws InvalidC
...
+            hostPort = arrowFlightProxy.substring("grpcs://".length());
+        } else if (arrowFlightProxy.startsWith("grpc://")) {
+            hostPort = arrowFlightProxy.substring("grpc://".length());
+        }
+
+        String[] split = hostPort.split(":");
```
jaogoy: Any existing utils to handle the parse of host:port, including for IPv6?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -631,9 +684,32 @@ protected FlightInfo getFlightInfoFromQuery(ArrowFlightSqlConnectContext ctx, Fl
...
-        // FETicket: <Token> : <QueryId>
-        return ByteString.copyFromUtf8(ctx.getArrowFlightSqlToken() + "|" + DebugUtil.printId(ctx.getExecutionId()));
+    protected Pair<Location, ByteString> getFEEndpoint(ArrowFlightSqlConnectContext ctx) throws InvalidConfException {
+        ByteString handle;
+        Location endpoint;
+        SessionVariable globalSv = getGlobalProxySessionVariable();
+
+        if (globalSv.isArrowFlightProxyEnabled()) {
+            String arrowFlightProxy = globalSv.getArrowFlightProxy();
+            // FE Proxy Ticket: <Token>|<QueryId>|<FEHost>:<FEPort>
```
jaogoy: It's better to extract the ticket code into a class, all the format, parse, generation will be much clear and concentrated.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
chris-celerdata: Not exactly, the stream created just after this fetches the data. The stream is a separate connection, but one client can create many streams.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
chris-celerdata: Yes, NetUtils contains some functions. I have changed the approach to use those.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/auth2/ArrowFlightSqlAuthenticator.java
```diff
@@ -64,10 +63,12 @@ AuthResult validateBearerToken(String token) {
...
             sessionManager.validateToken(token);
         } catch (IllegalArgumentException e) {
-            // Token not found locally - if proxy is enabled, allow through.
-            // Validation will happen on the remote FE when we forward the request.
+            // Token not found locally - if proxy is enabled and token is from a valid FE, allow through.
             if (isProxyEnabled()) {
-                return createAuthResult(token);
+                String feHost = ArrowFlightSqlSessionManager.extractFeHost(token);
+                if (feHost != null && ArrowFlightSqlSessionManager.isValidFeHost(feHost)) {
+                    return createAuthResult(token);
```
jaogoy, line 70: Does it mean to directly pass the validation, without proxy to the feHost for real validation?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/session/ArrowFlightSqlSessionManager.java
```diff
@@ -131,4 +133,61 @@ public ArrowFlightSqlConnectContext validateAndGetConnectContext(String token) t
...
+
+    /**
+     * Generate a token. When proxy is enabled, includes FE host prefix.
+     * Format: "FE_HOST:UUID" (proxy enabled) or "UUID" (proxy disabled)
+     */
+    private String generateToken() {
+        String uuid = UUIDUtil.genUUID().toString();
+        if (GlobalVariable.isArrowFlightProxyEnabled()) {
+            String selfHost = GlobalStateMgr.getCurrentState().getNodeMgr().getSelfNode().first;
+            return selfHost + ":" + uuid;
```
jaogoy: why not using `uuid + ":" + selfHost for compatibility?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -143,7 +144,15 @@ public void close() throws Exception {
...
-        ArrowFlightSqlConnectContext ctx = sessionManager.validateAndGetConnectContext(context.peerIdentity());
+        String token = context.peerIdentity();
+
+        // When proxy is enabled and token is from another FE, forward the request
+        if (GlobalVariable.isArrowFlightProxyEnabled() && !sessionManager.isLocalToken(token)) {
+            forwardCloseSessionToRemoteFE(token, listener);
+            return;
+        }
+
+        ArrowFlightSqlConnectContext ctx = sessionManager.validateAndGetConnectContext(token);
```
jaogoy, line 155: If the token is invalid, what will happen?

If a client closes a already closed session, for example it is retrying because the network exception , what will happen?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
```diff
@@ -639,6 +668,156 @@ private void getStreamResultFromFE(String token, String queryId, ServerStreamLis
...
+
+        int fePort = Config.arrow_flight_port;
+        String nodeKey = feHost + ":" + fePort;
+
+        try {
+            FlightClient client = getOrCreateClient(nodeKey, feHost, fePort);
+            CredentialCallOption authOption = new CredentialCallOption(new BearerCredentialWriter(token));
+
+            CloseSessionResult result = client.closeSession(new CloseSessionRequest(), authOption);
+            listener.onNext(result);
```
jaogoy, line 815: What's the reason to call `listener.onNext(...)`, which is not called for direct closeSession?

To a client's perspective, will there be some differences between direct call and forward call?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/session/ArrowFlightSqlSessionManager.java
```diff
@@ -131,4 +133,63 @@ public ArrowFlightSqlConnectContext validateAndGetConnectContext(String token) t
...
+    /**
+     * Generate a token. When proxy is enabled, includes FE host prefix.
+     * Format: "FE_HOST|UUID" (proxy enabled) or "UUID" (proxy disabled)
+     * Uses '|' as delimiter to support IPv6 addresses (which contain ':').
+     */
+    private String generateToken() {
+        String uuid = UUIDUtil.genUUID().toString();
+        if (GlobalVariable.isArrowFlightProxyEnabled()) {
+            String selfHost = GlobalStateMgr.getCurrentState().getNodeMgr().getSelfNode().first;
+            return selfHost + "|" + uuid;
```
jaogoy, line 191: Will `uuid | fehost` be more compatible?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
```diff
@@ -0,0 +1,366 @@
...
+            case 4:
+                return parseBEProxyTicket(ticketParts);
+            default:
+                throw CallStatus.INVALID_ARGUMENT.withDescription(
+                        String.format("Invalid ticket format: expected 2, 3, or 4 parts, got [%d]",
+                                ticketParts.length)).toRuntimeException();
+        }
+    }
+
+    private ParsedTicket parseFEProxyTicket(String[] ticketParts) {
```
jaogoy, line 158: A comment to indicate components of the ticketParts, more clear.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
```diff
@@ -0,0 +1,366 @@
...
+
+import java.net.URI;
+
+
+/**
+ * Manages Arrow Flight SQL ticket creation, parsing, and endpoint resolution.
+ *
+ * <p>Ticket formats:
+ * <ul>
+ *   <li>FE Local: {@code <Token>|<QueryId>} - 2 parts, handled by this FE</li>
```
jaogoy, line 38: It seems the part num is very important, and is the only indicator of the meaning to parse?

Even for `FEHost:FEPort` and `BEHost|BEPort`, we can't change the former to use `|` as the separator?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    public ByteString buildBETicket(String queryId, String fragmentInstanceId) {
+        return ByteString.copyFromUtf8(queryId + BE_TICKET_DELIMITER + fragmentInstanceId);
+    }
+
+    public ByteString buildFEProxyTicketForBE(TUniqueId queryId, TUniqueId fragmentInstanceId,
+                                               ComputeNode worker) {
+        return ByteString.copyFromUtf8(
+                hexStringFromUniqueId(queryId) + "|"
```
jaogoy, line 219: Not use the const variable?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
```diff
@@ -0,0 +1,366 @@
...
+                            arrowFlightProxy));
+        }
+
+        if (hostPort[0].isEmpty()) {
+            throw new InvalidConfException(
+                    String.format("Hostname cannot be empty, got '%s'", arrowFlightProxy));
+        }
+
+        try {
+            int port = Integer.parseInt(hostPort[1]);
```
jaogoy, line 359: Is port a must?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/auth2/ArrowFlightSqlAuthenticator.java
chris-celerdata: The validation will happen later, when the RPC is made to the FE.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/session/ArrowFlightSqlSessionManager.java
chris-celerdata: What do you mean here? This format would also be incompatible with local tokens.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
chris-celerdata: If the token is invalid, an exception will be thrown in `forwardCloseSessionToRemoteFE`. This will say that the token is invalid.
If it tries to close an already closed session, an exception will be thrown: "cannot find connect arrow context of the [token]". Arrow clients usually handle this.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/session/ArrowFlightSqlSessionManager.java
chris-celerdata: I am not sure what you mean by compatible. Do you have any more details?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
chris-celerdata: We could, but then there would be 4 parts, conflicting with the BE proxy. A different ticket design could explicitly state the format (i.e. `FEProxy|<Token>|<QueryId>|<FEHost>:<FEPort>`). Then the first part of the ticket denotes what format it is following, except for BE direct.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
chris-celerdata: What const variable?

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
chris-celerdata: Yes.

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
chris-celerdata: The direct call should also return a result to the listener. I have added it. There should not be a difference between a direct call and a forwarded call.

DanRoscigno:
/translate

**Maintainer:** monitor the translation in the [GitHub Actions tab](https://github.com/StarRocks/starrocks/actions/runs/21764226516)

ZiheLiu (dismissed):

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
jaogoy: If you use the original FlightClient, it will use the same TCP connection over http/2 (if we don't care about the TCP connection pool).

fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlTicketManager.java
jaogoy: You've decleared a TICKET_DELIMITER, but you also directly use "|" in your code.

jaogoy (dismissed):

ZiheLiu (approved):

jaogoy (approved):

PR #68434, opened 2026-01-26
[Enhancement] Unified cloud native table drop process with partition

### LINKED ISSUE #68267: Unified Partition-Level Deletion for Lake Tables in CatalogRecycleBin
## Refactoring request

**Is your refactoring request related to a problem? Please describe.**

When a Lake Table is dropped in StarRocks, it is first moved to the `CatalogRecycleBin`. After a configurable retention period (default 1 day), the table is permanently erased by the `eraseTable` method. The current implementation of `eraseTable` for Lake Tables handles the deletion logic independently, using a separate thread pool (`ASYNC_REMOVE_TABLE_EXECUTOR`) from the one used for partition deletion (`ASYNC_REMOVE_PARTITION_EXECUTOR`).

The current implementation has the following issues:

1. **Two Thread Pools**: Two separate thread pools are maintained for table and partition deletion:
   - `ASYNC_REMOVE_TABLE_EXECUTOR`: for Lake Table deletion
   - `ASYNC_REMOVE_PARTITION_EXECUTOR`: for partition deletion

2. **Coarse-Grained Retry**: When a Lake Table with multiple partitions fails to delete, the entire table deletion is retried. A finer-grained approach would retry only the failed partitions.

3. **Monolithic Deletion**: The table is deleted as a whole unit, even though the underlying data is organized by partitions.

Convert Lake Table deletion to partition-level deletion: When erasing a Lake Table, instead of deleting it as a whole, we:
1. Extract all partitions from the table
2. Add them to `idToPartition` (the same map used for dropped partitions)
3. Let `erasePartition()` handle the actual deletion using the shared `ASYNC_REMOVE_PARTITION_EXECUTOR`
4. After all partitions are deleted, clean up the table metadata and record the table-level edit log

This approach:
- **Reuses existing deletion infrastructure**: Both standalone partitions and table partitions go through `erasePartition()`
- **Uses single thread pool for Lake data cleanup**: `ASYNC_REMOVE_PARTITION_EXECUTOR` handles all Lake data cleanup
- **Enables fine-grained retry**: Failed partitions can be retried individually without affecting successfully deleted partitions
- **Maintains edit log consistency**: Table-level edit log is recorded only after all partitions are successfully deleted

**Additional context**
#### Decision 1: No Individual Partition Edit Logs for Table Deletion

**Decision**: Partitions from table deletion do NOT record individual partition erase edit logs.

**Rationale**:
1. The table-level edit log semantics are preserved
2. Reduces edit log volume (one table log vs N partition logs)
3. Matches the original table deletion behavior

**Risk Analysis**:

| Scenario | Risk | Mitigation |
|----------|------|------------|
| FE restart during deletion | Partitions may be re-processed | BE deletion is **idempotent** (see below) |
| Partial partition failure | Some partitions deleted, some not | Table stays in recycle bin, failed partitions retry individually |
| Follower FE sync | Follower doesn't have temp partitions | Follower only needs table erase edit log |

**BE Deletion Idempotency Verification**:

From `be/src/service/service_be/lake_service.cpp`:
```cpp
auto st = fs::remove_all(location);
if (!st.ok() && !st.is_not_found()) {
    // Only fail if error is NOT "not found"
    st.to_protobuf(response->mutable_status());
} else {
    // Success - includes "not found" case
    LOG(INFO) << "Removed " << location << " is_not_found=" << st.is_not_found();
}
```

**Conclusion**: BE treats "directory not found" as success, making re-deletion safe.

#### Decision 2: Lake vs Non-Lake Table Separation

**Decision**: Keep Lake and Non-Lake table deletion paths separate.

**Rationale**:
1. Non-Lake tables use a lightweight synchronous `deleteFromRecycleBin()` that directly returns `true`
2. Non-Lake table data cleanup is handled by BE tablet GC, not explicit RPC calls
3. Non-Lake tables have `isDeleteRetryable() = false`, so they don't need the retry mechanism
4. Unifying would add complexity without benefit

#### Decision 3: Memory-Only State for Tracking

**Decision**: `lakeTableToPartitions` and `partitionsFromTableDeletion` are not persisted.

**Rationale**:
1. Table is still in `idToTableInfo` (persisted), so it will be re-processed after restart
2. BE deletion is idempotent, so re-processing is safe
3. Avoids adding new edit log types or modifying existing persistence format

**Trade-off**: After FE restart, some partitions may be deleted again (no-op on BE side), causing minor RPC overhead.

### Comparison: Before vs After

| Aspect | Before | After |
|--------|--------|-------|
| Thread Pools for Lake | `ASYNC_REMOVE_TABLE_EXECUTOR` | `ASYNC_REMOVE_PARTITION_EXECUTOR` |
| Retry Granularity | Table-level | Partition-level |
| Progress Tracking | None | Per-partition |
| Code Complexity | Simple but less flexible | Slightly more complex but more robust |
| Edit Log | One table log | One table log (unchanged) |

wxl24life:
## Why I'm doing:

When a Lake Table is dropped and permanently erased from `CatalogRecycleBin`, the current implementation handles table deletion as a monolithic operation using a dedicated thread pool (`ASYNC_REMOVE_TABLE_EXECUTOR`), separate from the one used for partition deletion (`ASYNC_REMOVE_PARTITION_EXECUTOR`). This leads to:

1. **Redundant thread pools**: Two separate pools maintained for essentially the same type of work (Lake data cleanup).
2. **Coarse-grained retry**: A multi-partition table failure retries the entire table, rather than only the failed partitions.
3. **Shared directory skipping**: When dropping individual partitions, shared directories (used by multiple partitions of the same table) are intentionally skipped to avoid data loss. However, when the entire table is being dropped, all partitions are removed and these shared directories should be cleaned up as well. The original monolithic deletion handled this implicitly, but a partition-level approach requires explicit handling.

## What I'm doing:

Fixes #68267

Converts Lake Table deletion from a monolithic table-level operation to partition-level deletion, reusing the existing partition deletion infrastructure:

1. **Partition-level decomposition**: When erasing a Lake Table, extracts all its partitions into `idToPartition` and lets `erasePartition()` process them through the shared `ASYNC_REMOVE_PARTITION_EXECUTOR`, enabling per-partition retry on failure.
2. **Forced shared directory removal**: Introduces a `fromTableDeletion` flag on `RecyclePartitionInfo` to force removal of shared directories when partitions are deleted as part of a table drop, and to skip individual partition-level edit logs (only the table-level log is recorded).
3. **Eliminates the dedicated table deletion thread pool** (`ASYNC_REMOVE_TABLE_EXECUTOR`).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -68,6 +68,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.TestOnly;
```
kevincai: remove this annotation.

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -608,41 +712,38 @@ void eraseTable(long currentTimeMs) {
...
-                    finished = false;
-                    LOG.warn("erase table failed in Recycle Bin, DB id: {}, table name: {}, error message: {}",
-                             info.getDbId(), info.getTable().getName(), e.getMessage());
-                }
-
-                if (!finished) {
-                    // finish with error, re-submit in next round
-                    asyncDeleteForTables.remove(info);
+            // For Lake Table, convert to partition-level deletion
+            Preconditions.checkState(!info.isRecoverable());
```
kevincai: This is weird to use `isDeleteRetryable()` to distinguish the Table is a OlapTable or a LakeTable, we can add one more check here to make sure it is a LakeTable or MV if `isDeleteRetryable()` is true

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -608,41 +712,38 @@ void eraseTable(long currentTimeMs) {
...
+            long tableId = info.table.getId();
+
+            // Check if partitions have already been added to idToPartition
+            if (!lakeTableToPartitions.containsKey(tableId)) {
+                // First time processing this table, add its partitions to idToPartition
+                if (!addLakeTablePartitionsToRecycleBin(info)) {
+                    // No partitions, table is ready to be erased
+                    // Note: removeTableBinds() was already called in addLakeTablePartitionsToRecycleBin()
+                    OlapTable table = (OlapTable) info.getTable();
```
kevincai: looks like these lines are almost the same as `cleanupLakeTableAfterPartitionsDeletion()`

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -701,20 +801,34 @@ protected synchronized void erasePartition(long currentTimeMs) {
             }
 
             if (finished) {
-                GlobalStateMgr.getCurrentState().getEditLog().logErasePartition(partitionId, wal -> {
+                // Check if this partition comes from table deletion
+                boolean fromTableDeletion = partitionsFromTableDeletion.contains(partitionId);
+                if (fromTableDeletion) {
```
kevincai: need consider if it is good to write the WAL as well for those partitions, considering that if a table contains a large number of partitions, if the deletion is just partially done and the FE get restarted, those partitions will be deleted again after restart.

wxl24life: Good point. We intentionally do not write per-partition erase WAL for table-deletion-owned partitions, and only persist table-level erase WAL after all partitions are cleaned up. This keeps WAL volume manageable for large partitioned tables and preserves table-level semantic consistency. After FE restart, some already-cleaned partitions may be retried, but the partition delete path is designed to be idempotent (missing/cleaned paths are treated as safe), so correctness is not affected—only potential extra retry cost.

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
wxl24life: not exactly the same

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
kevincai: usually this is not a problem, but in case a table with large number of partitions, e.g. thousands of partitions, the retry cost is huge. Wondering if we can archive it by doing some intermittent WAL, e.g. has done hundreds of partitions cleanup from a table, a WAL will be persistent, so that even if it is a restart, up to this number of works will be redo at the most.

fe/fe-core/src/main/java/com/starrocks/persist/EraseTablePartitionsLog.java
```diff
@@ -0,0 +1,55 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
```
starrocks-xupeng: from design doc, i don't see there will be new log added, why here new type of log is added

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
wxl24life: > usually this is not a problem, but in case a table with large number of partitions, e.g. thousands of partitions, the retry cost is huge. Wondering if we can archive it by doing some intermittent WAL, e.g. has done hundreds of partitions cleanup from a table, a WAL will be persistent, so that even if it is a restart, up to this number of works will be redo at the most.


@starrocks-xupeng FYI

fe/fe-core/src/main/java/com/starrocks/persist/EraseTablePartitionsLog.java
wxl24life: prevent reloading cost too much time while table with large number of partitions failed to erase before fe restart

wxl24life: the docs need to update, thanks for the reminder~

wxl24life: as discussed offline, i will discard the batch WAL operation for now

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -564,7 +568,7 @@ void logAddApplyEraseTables(List<Long> tableIds) {
         LOG.info("Finished log erase tables: {}", StringUtils.join(tableIds, ","));
     }
 
-    List<RecycleTableInfo> removeTableFromRecycleBin(List<Long> tableIds) {
+    synchronized List<RecycleTableInfo> removeTableFromRecycleBin(List<Long> tableIds) {
```
kevincai, line 601: need to double check on the affection of this `synchronized` interface, it is a synchronized lock that may block all the operations.

kevincai (commented):
need fix the build break.

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -580,6 +584,18 @@ List<RecycleTableInfo> removeTableFromRecycleBin(List<Long> tableIds) {
             }
             idToRecycleTime.remove(tableId);
             enableEraseLater.remove(tableId);
+            // Clean up Lake Table partition tracking if exists
+            Set<Long> partitionIds = lakeTableToPartitions.remove(tableId);
```
starrocks-xupeng: i think if we reach here, it means partitionIds should be empty? otherwise it means there is bug

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -97,14 +97,17 @@ public class CatalogRecycleBin extends FrontendDaemon implements Writable, Memor
...
 
     private Map<RecyclePartitionInfo, CompletableFuture<Boolean>> asyncDeleteForPartitions;
-    private Map<RecycleTableInfo, CompletableFuture<Boolean>> asyncDeleteForTables;
+
+    // Track the relationship between Lake Table and its partitions for deletion
+    // Key: tableId, Value: Set of partitionIds that belong to this table
+    private final Map<Long, Set<Long>> lakeTableToPartitions;
+
+    // Track partitions that come from table deletion (should not log individual partition erase)
+    private final Set<Long> partitionsFromTableDeletion;
```
starrocks-xupeng: can you put `fromTableDeletion` into `RecyclePartitionInfo` or `idToPartition`, so you don't need another set

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
wxl24life: > i think if we reach here, it means partitionIds should be empty? otherwise it means there is bug

yes,  @starrocks-xupeng In the normal flow, by the time we erase the table, this mapping should already be gone (or empty) because partition-level deletion has finished.  I kept this block as a defensive cleanup for replay/recovery edge cases, so we don’t leak partition states if an inconsistent state appears. I can add a warning when `partitionIds` is non-empty to make the invariant explicit.

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -589,56 +608,140 @@ private synchronized void setNextEraseMinTime(long id, long eraseTime) {
...
+            recyclePartitionInfo.setRecoverable(false);
+            // Force remove shared directories since the entire table is being deleted
+            recyclePartitionInfo.setForceRemoveDirectory(true);
+
+            // Add to idToPartition so erasePartition() can process it
+            idToPartition.put(partitionId, recyclePartitionInfo);
+            // Use the table's recycle time to maintain consistency with ClusterSnapshot safety checks
+            idToRecycleTime.put(partitionId, partitionRecycleTime);
+            // Mark this partition as coming from table deletion
+            recyclePartitionInfo.setFromTableDeletion(true);
```
starrocks-xupeng: put this after setForceRemoveDirectory

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -589,56 +608,140 @@ private synchronized void setNextEraseMinTime(long id, long eraseTime) {
...
-                    asyncDeleteForTables.remove(info);
+            // Use async deletion with tracking for lake(cloud-native) tables
+            Preconditions.checkState(!info.isRecoverable() && info.table.isCloudNativeTableOrMaterializedView());
+            long tableId = info.table.getId();
+
+            // For lake table, we will convert it to partition-level deletion
+            // First, we should check if all of table's partitions have already been added to idToPartition
+            if (!lakeTableToPartitions.containsKey(tableId)) {
+                // First time processing this table, add its partitions to idToPartition
+                if (!addLakeTablePartitionsToRecycleBin(info)) {
```
starrocks-xupeng, line 771: this if seems not necessary, because if `addLakeTablePartitionsToRecycleBin` does not add any partition  `areLakeTablePartitionsDeleted` below will return true

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -684,7 +787,6 @@ protected synchronized void erasePartition(long currentTimeMs) {
                 try {
                     finished = future.get();
                 } catch (Exception e) {
-                    finished = false;
```
starrocks-xupeng, line 721: why remove this line

fe/fe-core/src/main/java/com/starrocks/catalog/RecyclePartitionInfo.java
```diff
@@ -46,6 +46,28 @@ public abstract class RecyclePartitionInfo extends JsonWriter {
...
+    /**
+     * When true, force remove the partition directory even if it's a shared directory.
+     * This is set when the partition is being deleted as part of a table deletion,
+     * where all partitions are being removed and shared directories should also be cleaned up.
+     * This field is transient and not serialized.
+     *
+     * Marked volatile because it is written in synchronized eraseTable() but read by
+     * async delete tasks submitted in erasePartition() on a different thread.
+     */
+    protected transient volatile boolean forceRemoveDirectory = false;
```
starrocks-xupeng: why this is needed if it's the same value as `fromTableDeletion`

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
starrocks-xupeng: so
```
cleanupLakeTableAfterPartitionsDeletion(info);
finishedTables.add(tableId);
```
will still be called

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -589,56 +608,140 @@ private synchronized void setNextEraseMinTime(long id, long eraseTime) {
...
+            partitionIds.add(partitionId);
+        }
+
+        if (partitionIds.isEmpty()) {
+            return false;
+        }
+
+        // Track the relationship between table and partitions
+        lakeTableToPartitions.put(tableId, partitionIds);
+        LOG.info("Added {} partitions from Lake table '{}' (tableId: {}) to recycle bin for deletion",
```
starrocks-xupeng: should be debug level

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -580,6 +580,25 @@ List<RecycleTableInfo> removeTableFromRecycleBin(List<Long> tableIds) {
...
             idToRecycleTime.remove(tableId);
             enableEraseLater.remove(tableId);
+            // Clean up Lake Table partition tracking if exists
+            Set<Long> partitionIds = lakeTableToPartitions.remove(tableId);
+            if (partitionIds != null) {
+                if (!partitionIds.isEmpty()) {
+                    // This is expected in follower replay scenario: leader has already deleted
+                    // partitions physically, follower just replays the table erase edit log
+                    // and cleans up the in-memory tracking data.
+                    LOG.info("Cleaning up lakeTableToPartitions for tableId {} during table erase. "
```
starrocks-xupeng: should be debug level

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
wxl24life: `finished` is initialized as false value. ok, we can revert this change to make logic cleare

fe/fe-core/src/main/java/com/starrocks/catalog/RecyclePartitionInfo.java
```diff
@@ -101,6 +113,22 @@ public long getRetentionPeriod() {
         return retentionPeriod;
     }
 
+    public boolean isForceRemoveDirectory() {
+        return fromTableDeletion;
+    }
+
+    public void setForceRemoveDirectory(boolean forceRemoveDirectory) {
```
starrocks-xupeng: why still need setForceRemoveDirectory if variable is removed

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -97,14 +97,14 @@ public class CatalogRecycleBin extends FrontendDaemon implements Writable, Memor
...
+
+    // Track the relationship between Lake Table and its partitions for deletion
+    // Key: tableId, Value: Set of partitionIds that belong to this table
+    private final Map<Long, Set<Long>> lakeTableToPartitions;
 
     private static final ExecutorService ASYNC_REMOVE_PARTITION_EXECUTOR = ThreadPoolManager.newDaemonFixedThreadPool(
                 Config.lake_remove_partition_thread_num, Integer.MAX_VALUE, "lake-remove-partition-pool", true);
 
-    private static final ExecutorService ASYNC_REMOVE_TABLE_EXECUTOR = ThreadPoolManager.newDaemonFixedThreadPool(
-                Config.lake_remove_table_thread_num, Integer.MAX_VALUE, "lake-remove-table-pool", true);
```
starrocks-xupeng, line 107: lake_remove_table_thread_num, need to remove this from Config

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -1198,19 +1316,58 @@ synchronized boolean isDeletingPartition(long id) {
...
+    }
+
+    @VisibleForTesting
+    synchronized boolean isPartitionFromTableDeletion(long partitionId) {
+        RecyclePartitionInfo info = idToPartition.get(partitionId);
+        return info != null && info.isFromTableDeletion();
+    }
+
+    @VisibleForTesting
+    synchronized boolean isPartitionForceRemoveDirectory(long partitionId) {
```
starrocks-xupeng: not needed?

wxl24life: done

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -589,56 +608,138 @@ private synchronized void setNextEraseMinTime(long id, long eraseTime) {
...
+     */
+    private void cleanupLakeTableAfterPartitionsDeletion(RecycleTableInfo info) {
+        long tableId = info.getTable().getId();
+        lakeTableToPartitions.remove(tableId);
+        // Clean up table-level resources
+        // Note: removeTableBinds() was already called in addLakeTablePartitionsToRecycleBin()
+        OlapTable table = (OlapTable) info.getTable();
+        table.removeTabletsFromInvertedIndex();
+        GlobalStateMgr.getCurrentState().getWarehouseMgr().removeTableWarehouseInfo(tableId);
+        LOG.info("All partitions of Lake table '{}' (tableId: {}) have been deleted, cleaned up table resources",
```
starrocks-xupeng: is this log necessary? in `logAddApplyEraseTables` right after, we will record a log as well

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -589,56 +608,138 @@ private synchronized void setNextEraseMinTime(long id, long eraseTime) {
...
+        }
+        return true;
+    }
+
+    /**
+     * Clean up Lake Table resources after all its partitions have been deleted.
+     *
+     * @param info the table info to clean up
+     */
+    private void cleanupLakeTableAfterPartitionsDeletion(RecycleTableInfo info) {
```
starrocks-xupeng, line 734: i suggest you also call `LakeTableHelper.deleteTableFromRecycleBin` here
```
    static boolean deleteTableFromRecycleBin(long dbId, OlapTable table, boolean replay) {
        Preconditions.checkState(table.isCloudNativeTableOrMaterializedView());
        if (replay) { // if not replay, removeTableBinds is already called in CatalogRecycleBin
            table.removeTableBinds(replay);
        }
        table.removeTabletsFromInvertedIndex();
        GlobalStateMgr.getCurrentState().getWarehouseMgr().removeTableWarehouseInfo(table.getId());
         return true;
    }
```

fe/fe-core/src/test/java/com/starrocks/catalog/CatalogRecycleBinLakeTableDropTableTest.java
```diff
@@ -0,0 +1,1511 @@
...
+        Assertions.assertTrue(recycleBin.isPartitionFromTableDeletion(p2.getId()));
+        Assertions.assertTrue(recycleBin.isPartitionFromTableDeletion(p3.getId()));
+
+        // erasePartition processes the partitions
+        recycleBin.erasePartition(futureTime);
+        // Wait for async deletion to complete
+        Thread.sleep(500);
+        // Second erasePartition call to process completed async tasks
+        recycleBin.erasePartition(futureTime);
+
```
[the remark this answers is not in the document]
starrocks-xupeng: do not rely on this `Thread.sleep(500);`, CI will not be stable

kevincai (dismissed):

starrocks-xupeng (approved):

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -614,6 +620,25 @@ List<RecycleTableInfo> removeTableFromRecycleBin(List<Long> tableIds) {
             }
             idToRecycleTime.remove(tableId);
             enableEraseLater.remove(tableId);
+            // Clean up Lake Table partition tracking if exists
```
kevincai: to keep consistent, the above line621-622 can be replaced by `removeRecycleMarkers(tableId)`

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -1232,19 +1366,52 @@ synchronized boolean isDeletingPartition(long id) {
     }
 
     @VisibleForTesting
-    synchronized boolean isDeletingTable(long id) {
-        RecycleTableInfo info = getRecycleTableInfo(id);
-        return info != null && asyncDeleteForTables.get(info) != null;
+    synchronized boolean isLakeTablePartitionsDeletionInProgress(long tableId) {
```
kevincai: no need to expose the internal implementation details in the interface.

name can be just `isLakeTableDeletingInProgress(long tableId)`

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -4101,9 +4101,6 @@ public class Config extends ConfigBase {
     @ConfField(mutable = false)
     public static int lake_remove_partition_thread_num = 8;
 
-    @ConfField(mutable = false)
```
kevincai, line 4104: mark it as deprecated and update comments and doc about it if exists

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -614,6 +620,25 @@ List<RecycleTableInfo> removeTableFromRecycleBin(List<Long> tableIds) {
             }
             idToRecycleTime.remove(tableId);
             enableEraseLater.remove(tableId);
+            // Clean up Lake Table partition tracking if exists
+            Set<Long> partitionIds = lakeTableToPartitions.remove(tableId);
+            if (partitionIds != null) {
+                if (!partitionIds.isEmpty()) {
+                    // This is expected in follower replay scenario: leader has already deleted
```
kevincai: this comment is a little bit confusing. on the follower, the `lakeTableToPartitions` should be empty because the `runAfterCatalogReady()` never runs, and partitions won't be populated to the `lakeTableToPartition` map, my understanding correct?

fe/fe-core/src/main/java/com/starrocks/catalog/RecyclePartitionInfo.java
```diff
@@ -46,6 +46,22 @@ public abstract class RecyclePartitionInfo extends JsonWriter {
...
+     *
+     * Persisted so that after FE restart, partitions mid-way through table deletion still
+     * have forceRemoveDirectory=true (avoiding missed shared-dir cleanup) and still suppress
+     * per-partition WAL logs (avoiding spurious individual erase logs).
+     *
+     * Marked volatile because it is written in synchronized eraseTable() but read by
+     * async delete tasks submitted in erasePartition() on a different thread.
+     */
+    @SerializedName(value = "fromTableDeletion")
+    protected volatile boolean fromTableDeletion = false;
```
kevincai, line 63: what's the default value if it is loaded from the gson persistent data without this field?

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
wxl24life: greate catch. the comment is misleading. it can only happened on a leader→follower switchover node, the comment will be updated with more info describing the secenrios.

fe/fe-core/src/main/java/com/starrocks/catalog/RecyclePartitionInfo.java
wxl24life: The default value would be false. When Gson deserializes a JSON object that does not contain the fromTableDeletion field, it leaves the Java field at its field initializer value, which is explicitly set to false

kevincai (dismissed):

xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -612,8 +612,28 @@ List<RecycleTableInfo> removeTableFromRecycleBin(List<Long> tableIds) {
...
+                            tableId, partitionIds.size(), partitionIds);
+                }
+                for (Long partitionId : partitionIds) {
+                    RecyclePartitionInfo partitionInfo = removePartitionFromRecycleBinInternal(partitionId);
+                    if (partitionInfo != null) {
+                        asyncDeleteForPartitions.remove(partitionInfo);
+                    }
+                    removeRecycleMarkers(partitionId);
+                }
+            }
```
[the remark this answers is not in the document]
wxl24life: make removeTableFromRecycleBin() derive the partition list from the table object itself (which IS persisted in the checkpoint), rather than relying solely on lakeTableToPartitions. The table's getAllPartitions() returns exactly the same partitions that addLakeTablePartitionsToRecycleBin() would have converted.

kevincai (dismissed):

caneGuy (approved):

kevincai (approved):

gengjun-git (approved):

srlch (approved):

wyb (approved):

PR #68477, opened 2026-01-27
[Enhancement] Add config to enable composite estimate for DataSketches HLL

zenoyang:
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1910,4 +1910,8 @@ CONF_Int32(cow_optimization_diagnose_level, "0");
 // to potentially find a better predicate order. When selectivity is already good (low), sampling
 // is unlikely to help and will be skipped.
 CONF_mDouble(predicate_sampling_trigger_selectivity_threshold, "0.2");
+
+// When enabled, HllSketchAggregateFunction will use get_composite_estimate() instead of get_estimate()
+// for more stable estimation results. Default is false to maintain current behavior.
+CONF_mBool(enable_data_sketches_hll_estimate_composite, "false");
 } // namespace starrocks::config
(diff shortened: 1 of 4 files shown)


be/src/common/config.h
```diff
@@ -1910,4 +1910,8 @@ CONF_Int32(cow_optimization_diagnose_level, "0");
 // to potentially find a better predicate order. When selectivity is already good (low), sampling
 // is unlikely to help and will be skipped.
 CONF_mDouble(predicate_sampling_trigger_selectivity_threshold, "0.2");
+
+// When enabled, HllSketchAggregateFunction will use get_composite_estimate() instead of get_estimate()
+// for more stable estimation results. Default is false to maintain current behavior.
+CONF_mBool(enable_data_sketches_hll_estimate_composite, "false");
```
alvin-phoenix-ai, line 1916: This is not a good option to have a configuration to control the behavior for a single function.
Suppose the scenarios that there are two tables, table A and table B
Users want a stable result for table A, but can accept a rough result for table B.
With a single configuration, users can't achieve it in a single cluster.

It is acceptable to have a new function.

alvin-phoenix-ai (changes requested):
To introduce a configuration to control a function's behavior is not good. So I put a change request for this PR.

be/src/common/config.h
ViggoC: IMHO, we can add an `isStable` function argument, with a default value of `false`.

PR #68480, opened 2026-01-27
[Enhancement] add partition scan num limit when query external table

MatthewH00:
## Why I'm doing:
when query big size paimon table or other external table with full table scan or scan too many partitions, would cause BE/CN node high load, lead to cluster instability.

## What I'm doing:
add a new FE session variable scan_lake_partition_num_limit to limit partition scan num when query paimon table.
(default value is 0, means no limitation)

note: the new session variable `scan_lake_partition_num_limit` replace the old variable `scan_hive_partition_num_limit`(still remain) to uniformly handle the lake partition limit . 

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
@@ -942,6 +942,8 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
 
     public static final String SCAN_OLAP_PARTITION_NUM_LIMIT = "scan_olap_partition_num_limit";
 
+    public static final String SCAN_PAIMON_PARTITION_NUM_LIMIT = "scan_paimon_partition_num_limit";
+
     public static final String ENABLE_CROSS_JOIN = "enable_cross_join";
 
     public static final String ENABLE_NESTED_LOOP_JOIN = "enable_nested_loop_join";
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
```diff
@@ -942,6 +942,8 @@ public static MaterializedViewRewriteMode parse(String str) {
 
     public static final String SCAN_OLAP_PARTITION_NUM_LIMIT = "scan_olap_partition_num_limit";
 
+    public static final String SCAN_PAIMON_PARTITION_NUM_LIMIT = "scan_paimon_partition_num_limit";
```
kevincai: I think the solution shall be considered with a broader use case. This session var is added specifically for paimon catalog, what if needed for hive/iceberg or other future catalog, will we redo the code change for every other external catalog?

MatthewH00: @kevincai Adjust to `scan_lake_partition_num_limit` would be ok?
And this PR is only for paimon, other catalog like iceberg could use this session var add the same limit on demond if needed later.

StarRocks-Reviewer:
@cursor review

Marvin5181:
@MatthewH00 @kevincai Excuse me, when the feature would be merged into v4.0? We also need the feature in our prod env.

kevincai (dismissed):

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
stephen-shelby, line 941: Please add an alias for SCAN_HIVE_PARTITION_NUM_LIMIT. it's best to reuse this session var to uniformly handle the lake partition limit.

MatthewH00: adjust done

kevincai (approved):

stephen-shelby (approved):

PR #68488, opened 2026-01-27
[Enhancement] Add optimizer trace logging for planning failures

stephen-shelby:
## Why I'm doing:
StarRocks supports `TRACE TIMES OPTIMIZER` to help diagnose slow optimizer rules/tasks, but it requires users to **explicitly add TRACE** to the SQL. In production incidents (especially **planning failures / CBO timeout / planner exceptions**), users often only have the failed SQL and an error message, and cannot easily reproduce the exact environment to run `TRACE TIMES OPTIMIZER` again. As a result, it’s hard to quickly identify **which optimizer stage/rule is slow** or where the planning time is spent.

## What I'm doing:
- Add a new FE config **`enable_dump_optimizer_trace_on_error`** (mutable, default `false`).
- When this config is enabled, FE will **collect optimizer TIMER trace during plan generation** and **dump a `trace times optimizer`-style timing tree** into FE logs **only when plan generation fails** (e.g. CBO timeout / planner exception), including:
  - `query_id`
  - redacted `sql`
  - `err` (exception type + message)
  - timing tree output (with truncation protection)

log example
default:
image

enable enable_dump_optimizer_trace_on_error:
image

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
@@ -688,6 +688,14 @@ public class StmtExecutor {
 
     private ExecPlan generateExecPlan() throws Exception {
         ExecPlan execPlan = null;
+
+        // Collect optimizer timing only when explicitly enabled to avoid overhead on normal queries.
+        // When enabled, we only dump the trace to logs on planning failure.
+        if (Config.enable_dump_optimizer_trace_on_error) {
+            Tracers.enableTraceMode(Tracers.Mode.TIMER);
+            Tracers.enableTraceModule(Tracers.Module.BASE);
+            Tracers.enableTraceModule(Tracers.Module.OPTIMIZER);
+        }
         try (Timer ignored = Tracers.watchScope("Total")) {
             if (!isForwardToLeader()) {
                 if (context.shouldDumpQuery()) {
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
```diff
@@ -688,6 +688,14 @@ public Pair<String, Integer> getTableQueryTimeoutInfo() {
 
     private ExecPlan generateExecPlan() throws Exception {
         ExecPlan execPlan = null;
+
+        // Collect optimizer timing only when explicitly enabled to avoid overhead on normal queries.
+        // When enabled, we only dump the trace to logs on planning failure.
+        if (Config.enable_dump_optimizer_trace_on_error) {
```
LiShuMing, line 694: I think we don't need a config to support collecting traces even if it meets an exception? how about supporting `trace <mode> <module> ` by default even it meets exception?

stephen-shelby: The main use case is production incidents where queries fail unexpectedly. users can't reproduce these failures. so they can't add trace beforehand. once a query fails, it's often hard to reproduce the exact same conditions.

kangkaisen (approved):

Seaven (approved):

PR #68505, opened 2026-01-27
[BugFix] pk table char column no padding during alter

sevev:
## Why I'm doing:
when pk alter table, char column miss padding, and it will generate an error zone map index.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [x] 3.3

### DIFF
--- a/be/src/storage/tablet_updates.cpp
+++ b/be/src/storage/tablet_updates.cpp
@@ -4400,6 +4400,7 @@ Status TabletUpdates::_convert_from_base_rowset(const Schema& base_schema, const
                                                 const std::unique_ptr<RowsetWriter>& rowset_writer) {
     ChunkPtr base_chunk = ChunkHelper::new_chunk(base_schema, config::vector_chunk_size);
     ChunkPtr new_chunk = ChunkHelper::new_chunk(new_schema, config::vector_chunk_size);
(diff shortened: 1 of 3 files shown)


wyb (approved):
The fix looks correct. Explicitly padding CHAR columns during schema change (conversion and reordering) ensures that the stored data complies with the expected fixed-length format (padded with nulls/zeros in this case). This is critical for correctness of storage indexes like ZoneMap, as demonstrated by the test case where ZoneMap filtering on unpadded data caused the row to be effectively lost in queries.

I verified that `ChunkHelper::padding_char_columns` pads with zeros, which matches the expected storage format for CHAR columns in StarRocks BE. Adding this step before adding chunks to the `RowsetWriter` prevents the inconsistency.

The changes in `_convert_from_base_rowset` and `reorder_from` cover the necessary paths for schema change operations.

wyb (approved):

luohaha (approved):

PR #68525, opened 2026-01-27
[Doc] Fix bloom filter index documentation example

rickxz:
## Why I'm doing:

The bloom filter index documentation contains an example that contradicts its own explanation. The docs state that bloom filters are for columns **not used as a sort key**, but the example creates bloom filter indexes on `k1, k2` — the PRIMARY KEY columns, which ARE the sort key. This makes the bloom filter useless (confirmed: `BloomFilterFilterRows: 0`).

Additionally, the example uses `REPLACE` aggregate type in a `PRIMARY KEY` table, which is confusing/misleading syntax.

## What I'm doing:

Updated the bloom filter index documentation (EN, ZH, JA) to fix the broken example:

1. **Bolded key concept**: emphasized that bloom filters are for non-sort-key, high-cardinality columns
2. **Fixed CREATE TABLE example**:
   - Changed from `PRIMARY KEY` to `DUPLICATE KEY` (more realistic use case)
   - Bloom filter now on `v1` (non-key VARCHAR column) instead of `k1, k2` (sort key)
   - Removed misleading `REPLACE` aggregate type
3. **Updated ALTER TABLE examples**: made them consistent with the new CREATE TABLE example

Fixes #68524

## What type of PR is this:
- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [x] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:
- [x] This PR needs user documentation (for new or modified features or behaviors)
- [ ] This is a backport PR

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4

---

## Additional Context

The old example would show `BloomFilterFilterRows: 0` because bloom filters on sort key columns are redundant — the prefix index already handles filtering for those columns. The new example demonstrates the actual intended use case: a bloom filter on a non-key, high-cardinality column like a VARCHAR field.

This documentation fix should be backported to all active release branches since users on those versions may copy the broken example.

### DIFF
--- a/docs/en/table_design/indexes/Bloomfilter_index.md
+++ b/docs/en/table_design/indexes/Bloomfilter_index.md
@@ -9,7 +9,7 @@ This topic describes how to create and modify bloom filter indexes, along with h
 
 A bloom filter index is a space-efficiency data structure that is used to detect the possible presence of filtered data in data files of a table. If the bloom filter index detects that the data to be filtered are not in a certain data file, StarRocks skips scanning the data file. Bloom filter indexes can reduce response time when the column (such as ID) has a relatively high cardinality.
 
-If a query hits a sort key column, StarRocks efficiently returns the query result by using the [prefix index](./Prefix_index_sort_key.md). However, the prefix index entry for a data block cannot exceed 36 bytes in length. If you want to improve the query performance on a column, which is not used as a sort key and has a relatively high cardinality, you can create a bloom filter index for the column.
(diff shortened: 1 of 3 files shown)


StarRocks-Reviewer:
@cursor review

DanRoscigno:
Hi @rickxz , sorry for the delay. I will get this reviewed.

PR #68659, opened 2026-01-29
[Refactor] deprecate `null_indicator_byte` and `null_indicator_bit` fields

dirtysalt:
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/runtime/descriptor_helper.h
+++ b/be/src/runtime/descriptor_helper.h
@@ -79,7 +79,7 @@ public:
         return this->type(TYPE_DECIMALV2).precision(precision).scale(scale);
     }
     TSlotDescriptorBuilder& nullable(bool nullable) {
-        _slot_desc.nullIndicatorByte = (nullable) ? 0 : -1;
+        _slot_desc.__set_isNullable(nullable);
         return *this;
     }
     TSlotDescriptorBuilder& is_materialized(bool is_materialized) {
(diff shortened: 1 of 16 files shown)


stdpain (approved):

gensrc/proto/descriptors.proto
```diff
@@ -46,13 +46,14 @@ message PSlotDescriptor {
     required PTypeDesc slot_type = 3;
     required int32 column_pos = 4;   // in originating table
     required int32 byte_offset = 5;  // into tuple
-    required int32 null_indicator_byte = 6;
-    required int32 null_indicator_bit = 7;
+    required int32 null_indicator_byte = 6; // deprecated
+    required int32 null_indicator_bit = 7; // deprecated
```
alvin-phoenix-ai, line 50: You can change them from required into optional in this PR

be/src/runtime/descriptor_helper.h
```diff
@@ -121,13 +121,13 @@ class TTupleDescriptorBuilder {
         _tuple_id = tb->next_tuple_id();
         int num_nullables = 0;
         for (auto& slot_desc : _slot_descs) {
-            if (slot_desc.nullIndicatorByte >= 0) {
+            bool is_nullable = slot_desc.__isset.isNullable ? slot_desc.isNullable : true;
+            if (is_nullable) {
```
alvin-phoenix-ai, line 125: This is only used internally, and in SlotDesc it is set `_slot_desc.__set_isNullable(nullable);`
So I think you can use `slot_desc.isNullable` directly

alvin-phoenix-ai (commented):
others look good to me

gensrc/proto/descriptors.proto
dirtysalt: maybe it will cause backward compatiblity problem?

new BE will use it as optional and there is no value in this field.
but old BE will expect value in this field.

be/src/runtime/descriptor_helper.h
dirtysalt: yes, but since this is a very typical form, I think it does not make any harm to write in this way.

alvin-phoenix-ai (approved):

PR #68680, opened 2026-01-29
[Enhancement] Improve skew join v2 rewrite with stats-based detection and NULL skew

stephen-shelby:
## Why I'm doing:
Skew join v2 works well when join keys are simple column references or when skew values are manually provided in a comparable form. In production, join keys frequently contain expressions (casts/functions/projection-derived keys), and we also want to leverage histogram MCV/NULL statistics to automatically identify skew values. Without these improvements, the rewrite may not trigger, may miss real skew, or may generate ineffective split predicates.

## What I'm doing:
- Add **NULL skew** support in split predicates using `IS NULL` / `IS NOT NULL` (instead of putting NULL into `IN` lists) to avoid SQL three-valued logic pitfalls.
- Integrate **stats-based skew detection** (MCV/NULL) with session-variable controls and safety guardrails.
- Enhance skew join v2 rewrite to better handle **expression join keys**, including rewriting skew values to match derived join-key expressions.

## Skew values selection strategy (parameters & defaults)
This change introduces **two new session variables** specifically for **MCV-based skew-value evaluation**. The remaining parameters below **already existed** and are reused to gate the rewrite.

### New parameters (added in this change)
- `skew_join_mcv_single_threshold` (default: 0.1)
  - Per-value MCV filter: keep a value `v` only if `count(v) / rowCount >= 0.1`.
  - This prevents many small MCVs from being treated as skew values.

- `skew_join_mcv_min_input_rows` (default: 10000000)
  - Minimum input rows gating for enabling MCV-based rewrite on a join side.
  - Skip a side if `Statistics.getOutputRowCount() < 10000000`.

### Existing parameters (already present)
- `enable_stats_to_optimize_skew_join` (default: false)
  - Master switch for stats-based skew detection. If false, only hint-based skew values are used.

- `skew_join_use_mcv_count` (default: 5)
  - Top-K limit: how many MCV entries are considered as skew-value candidates.

- `skew_join_data_skew_threshold` (default: 0.2)
  - Overall skew gating threshold.
  - MCV skew triggers when:
    - `mcvSkewFactor = sum_{v in topK} count(v) / rowCount` is greater than `skew_join_data_skew_threshold`.
  - NULL skew triggers when:
    - `nullFraction >= skew_join_data_skew_threshold`.

### Side selection and predicate construction 
- Evaluate both join children (subject to join-type constraints) and pick the side with larger:
  - `score = max(nullSkewFactor, mcvSkewFactor)`.
- LEFT OUTER JOIN constraint:
  - Right-side skew rewrite is forbidden, so the right child is excluded from candidates.
- Predicate construction:
  - MCV candidates become `IN (...)` values (after type casting and expression-key rewriting when needed).
  - NULL candidate is represented explicitly via `IS NULL` / `IS NOT NULL` (never `IN (NULL)`).
  - Final split predicate patterns:
    - MCV-only:
      - skew branch: `k IN (S)`
      - non-skew branch: `k NOT IN (S) OR k IS NULL`
    - NULL-only:
      - skew branch: `k IS NULL`
      - non-skew branch: `k IS NOT NULL`
    - MCV + NULL:
      - skew branch: `k IN (S) OR k IS NULL`
      - non-skew branch: `k NOT IN (S) AND k IS NOT NULL`

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/SkewShuffleJoinEliminationRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/SkewShuffleJoinEliminationRule.java
@@ -17,6 +17,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.starrocks.common.LocalExchangerType;
 import com.starrocks.common.Pair;
+import com.starrocks.qe.SessionVariable;
 import com.starrocks.sql.ast.HintNode;
 import com.starrocks.sql.optimizer.JoinHelper;
 import com.starrocks.sql.optimizer.OptExpression;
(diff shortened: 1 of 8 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/SkewShuffleJoinEliminationRule.java
```diff
@@ -491,17 +585,105 @@ private boolean isSkew(OptExpression opt) {
             return joinOperator.getJoinHint().equals(HintNode.HINT_JOIN_SKEW);
         }
 
-        private class ColumnRefReplacer extends BaseScalarOperatorShuttle {
+        private Optional<SkewJoinSplitInfo> detectSkewFromStats(OptExpression joinOpt) {
```
LiShuMing, line 588: Can we unify this detect skew from stats rule with `SkewJoinOptimzeRule.java ` even the inner operators are differnt(physical/logical)

stephen-shelby: The primary goal of optimization v2 is to disable v1, so we can revisit this after the remaining tests are completed.

before-Sunrise (approved):

LiShuMing (approved):

Seaven (approved):

PR #35486, opened 2023-11-21
[Enhancement] Support skew hint for window partition clause

liuyehcf:
Why I'm doing:

For window function with partition by clause, if the partition is skewed, the sort operator before and after the window operator will also be affected. And this pr is trying to address the skew problem.

What I'm doing:

1. Introduce a skewed hint to help be add additional local exchange operators to regulate the data distribution in order to get better performence of the sort operator.
    * Add passthrough exchange before sort node
    * Add ordered partition exchange before analytic node.

syntax like this
```
sum(v1) over([skewed] partition by v2 order by v3)
```

3. Add an invisiable session variable `enable_pipeline_level_shuffle` to control whether use `pipeline_level_shuffle`.

Fixes #issue

## Experiment

```sql
DROP DATABASE IF EXISTS `analytic`;
CREATE DATABASE `analytic`;
USE `analytic`;
DROP TABLE IF EXISTS `t0`;
CREATE TABLE IF NOT EXISTS `t0` (
  `id_int` INT(11) NOT NULL,
  `id_tinyint` TINYINT NOT NULL,
  `id_smallint` SMALLINT NOT NULL,
  `id_bigint` BIGINT NOT NULL,
  `id_largeint` LARGEINT NOT NULL,
  `id_float` FLOAT NOT NULL,
  `id_double` DOUBLE NOT NULL,
  `id_char` CHAR(10) NOT NULL,
  `id_varchar` VARCHAR(100) NOT NULL,
  `id_date` DATE NOT NULL,
  `id_datetime` DATETIME NOT NULL,
  `id_decimal` DECIMAL(9,0) NOT NULL
) ENGINE=OLAP
DUPLICATE KEY(`id_int`)
DISTRIBUTED BY HASH(`id_int`) BUCKETS 10
PROPERTIES (
 "replication_num" = "1"
);

INSERT INTO `t0` (id_int, id_tinyint, id_smallint, id_bigint, id_largeint, id_float, id_double, id_char, id_varchar, id_date, id_datetime, id_decimal) VALUES
    (1, 10, 100, 1000, 10000, 100000.1, 1000000.1, 'some_char', 'some_varchar', '2020-07-24', '2020-07-24', 123.78),
    (1, 20, 200, 2000, 20000, 200000.1, 2000000.1, 'some_char2', 'some_varchar2', '2020-07-23', '1020-07-24', 123.56),
[...]

PR #68739, opened 2026-01-30
[Enhancement] Support explicit skew hint for window functions

martinr0x:
## Why I'm doing:
Currently `SplitWindowSkewToUnionRule` is only triggered if statistics indicate skew. We need a hint to trigger this rule similiar to how it is done in [skew_join_v2](https://docs.starrocks.io/docs/using_starrocks/skew_join_v2/).

## What I'm doing:
This PR introduces support for explicit skew hints in window functions, allowing users to manually specify skewed values in the partition clause. This hint guides the optimizer to use the `SplitWindowSkewToUnionRule` optimization, which splits the window operator into a `UNION` of two branches: one handling the skewed value and another for the remaining data.
### Syntax
The syntax follows the pattern: `[skew|<column_name>(<value>)]` placed before the `PARTITION BY` clause.

**Example:**
```sql
SELECT sum(v) OVER ([skew|p(NULL)] PARTITION BY p ORDER BY o) FROM t;
SELECT sum(v) OVER ([skew|p(1)] PARTITION BY p ORDER BY o) FROM t;
```

If we need documentation I would add it in another PR.

## Benchmarks:
```sql
CREATE TABLE window_bench (
    o BIGINT,
    -- The ordering column
    v DOUBLE,
    -- The value to sum
    p1 INT,  -- Skewed towards NULL
    p2 INT, -- Skewed towards specific keys (Partition Skew)
    p3 INT -- No skew
) ENGINE = OLAP DUPLICATE KEY(o) DISTRIBUTED BY HASH(o) BUCKETS 32 PROPERTIES ("replication_num" = "1");

INSERT INTO
    window_bench
SELECT
    generate_series as o,
    rand() * 100 as v,
    IF(rand() < 0.95, NULL, floor(rand() * 1000)) as p1,
    CASE
        WHEN rand() < 0.5 THEN 1
        WHEN rand() < 0.8 THEN 2
        ELSE floor(rand() * 1000) + 3
    END as p2,
    floor(rand() * 1000) as p3
FROM
    TABLE(generate_series(1, 100000000));

set enable_split_window_skew_to_union = true;
set enable_pipeline_level_shuffle =true;

-- Q1 
-- without Optimizations
SELECT p1,v, o, sum(v) OVER (PARTITION BY p1 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;

-- with skew optimization on NULLs
SELECT p1,v, o, sum(v) OVER ([skew|p1(NULL)] PARTITION BY p1 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;

-- with skewed optimization 
SELECT p1,v, o, sum(v) OVER ([skewed] PARTITION BY p1 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;


-- Q2
-- without Optimizations
SELECT p2,v, o, sum(v) OVER (PARTITION BY p2 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;

-- with skew optimization on NULLs
SELECT p2,v, o, sum(v) OVER ([skew|p2(1)] PARTITION BY p2 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;

-- with skewed optimization 
SELECT p2,v, o, sum(v) OVER ([skewed] PARTITION BY p2 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;

-- Q3
-- without Optimizations
SELECT p3,v, o, sum(v) OVER (PARTITION BY p3 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;

-- with skew optimization on NULLs
SELECT p3,v, o, sum(v) OVER ([skew|p3(NULL)] PARTITION BY p3 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;

-- with skewed optimization 
SELECT p3,v, o, sum(v) OVER ([skewed] PARTITION BY p3 ORDER BY o) as running_total FROM window_bench order by running_total DESC LIMIT 1;
```
| Query | No Optimization | Skew | Skewed |
| :--- | :--- | :--- | :--- |
| **Q1** | 14.855s | 7.179s | 10.609s |
| **Q2** | 11.463s | 9.101s | 10.274s |
| **Q3** | 3.112s | 3.398s | 6.011s |

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/sql/ast/expression/AnalyticExpr.java
```diff
@@ -205,6 +213,22 @@ public String getSqlString() {
         return sqlString;
     }
 
+    public Expr getSkewColumn() {
+        return skewColumn;
+    }
+
+    public void setSkewColumn(Expr skewColumn) {
+        this.skewColumn = skewColumn;
+    }
```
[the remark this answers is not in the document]
martinr0x: removed setter

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/transformer/WindowTransformer.java
```diff
@@ -681,8 +699,12 @@ public WindowOperator(AnalyticExpr analyticExpr, List<Expr> partitionExprs,
             }
             if (!partitionExprs.isEmpty()) {
                 this.isSkewed = analyticExpr.isSkewed();
+                this.skewColumn = analyticExpr.getSkewColumn();
+                this.skewValue = analyticExpr.getSkewValue();
```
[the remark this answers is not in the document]
martinr0x: added fix and added more tests

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/transformer/WindowTransformer.java
```diff
@@ -720,6 +742,22 @@ public void setSkewed() {
...
+            return skewColumn;
+        }
+
+        public Expr getSkewValue() {
+            return skewValue;
+        }
+
         @Override
         public boolean equals(Object o) {
             if (this == o) {
```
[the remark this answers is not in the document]
martinr0x: good catch, done

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/operator/physical/PhysicalWindowOperator.java
```diff
@@ -43,6 +43,10 @@ public class PhysicalWindowOperator extends PhysicalOperator {
     private final boolean useHashBasedPartition;
     private final boolean isSkewed;
 
+    // Skew hint with explicit column and value: [skew|t.column(value)]
+    private final ScalarOperator skewColumn;
```
kangkaisen, line 47: fe/fe-core/src/main/java/com/starrocks/sql/spm/SPMAstCheckVisitor.java:182

Issue: SPM AST equivalence ignores skewColumn/skewValue. Queries that differ only by explicit skew hint can be treated as identical, which risks reusing an incompatible plan (or failing to reuse a correct one).

```
check = check && Objects.equals(node.getSkewHint(), other.getSkewHint());
check = check && Objects.equals(node.isUseHashBasedPartition(), other.isUseHashBasedPartition());
check = check && Objects.equals(node.isSkewed(), other.isSkewed());
```

fe/fe-parser/src/main/java/com/starrocks/sql/ast/HintNode.java
```diff
@@ -30,7 +30,7 @@ public abstract class HintNode implements Comparable<HintNode>, ParseNode {
     public static final String HINT_ANALYTIC_SORT = "sort";
     public static final String HINT_ANALYTIC_HASH = "hash";
     public static final String HINT_ANALYTIC_SKEW = "skewed";
-
+    public static final String HINT_SKEW_EXPLICIT = "skew";
```
kangkaisen: There are   public static final String HINT_JOIN_SKEW = "SKEW"; and  public static final String HINT_JOIN_SKEW = "SKEW";

do we still need HINT_SKEW_EXPLICIT = "skew";?

fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
```diff
@@ -8103,10 +8102,27 @@ private AnalyticExpr buildOverClause(FunctionCallExpr functionCallExpr,
...
+        if (context.bracketHint() != null) {
+            if (context.bracketHint().primaryExpression() != null) {
+                skewColumn = (Expr) visit(context.bracketHint().primaryExpression());
+            }
+            if (context.bracketHint().generalLiteralExpressionList() != null) {
+                List<Expr> skewValues = visit(
+                        context.bracketHint().generalLiteralExpressionList().generalLiteralExpression(),
+                        Expr.class);
+                if (!skewValues.isEmpty()) {
+                    skewValue = skewValues.get(0);
```
stephen-shelby: Why do we only use the first value here?

fe/fe-parser/src/main/java/com/starrocks/sql/ast/HintNode.java
```diff
@@ -30,7 +30,7 @@ public abstract class HintNode implements Comparable<HintNode>, ParseNode {
     public static final String HINT_ANALYTIC_SORT = "sort";
     public static final String HINT_ANALYTIC_HASH = "hash";
     public static final String HINT_ANALYTIC_SKEW = "skewed";
```
LiShuMing, line 32: It seems `HINT_ANALYTIC_SKEW` is not used before; maybe we can change it into `skew` and reuse this definition. Otherwise there are two skew hints: `skew`/`skewed`?

fe/fe-parser/src/main/java/com/starrocks/sql/ast/HintNode.java
martinr0x: Oh right, we could reuse the HINT_JOIN_SKEW.

fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
martinr0x: Because for now the rewrite rule only supports one skewed value

fe/fe-parser/src/main/java/com/starrocks/sql/ast/HintNode.java
martinr0x: I think it is used if enabled. HINT_ANALYTIC_SKEW is another rewrite introduced some long time ago (https://github.com/StarRocks/starrocks/pull/35486)

The change is mostly on BE hence there are not many references in FE for it

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/operator/physical/PhysicalWindowOperator.java
martinr0x: Thats a good catch [b8355c2](https://github.com/StarRocks/starrocks/pull/68739/commits/b8355c2800f2b58ed726f0ea2b8b012205d79960)

fe/fe-parser/src/main/java/com/starrocks/sql/ast/HintNode.java
martinr0x: [6be1c0a](https://github.com/StarRocks/starrocks/pull/68739/commits/6be1c0a5ae1999b7eb8d783772fb41e5fc5a7837)

fe/fe-parser/src/main/java/com/starrocks/sql/ast/HintNode.java
LiShuMing: The `SKEW` and `SKEWED` hints both appear to address skew issues in window operators, which seems a bit confusing. 
Could you test the performance difference between the `SKEWED` and `SKEW` hints?  So we can decide which one to be used later.

martinr0x: Hi @LiShuMing ,
I added some very basic benchmarks.

 Generally `[skewed]` is a valid optimization if data is slightly skewed. With this optimizations the analytical operators are more optimally distributed among worker thread. However, this optimization does not address the single threaded sort if one partition is very skewed, which the new skew optimization does.

martinr0x: Is it ok for you to keep both optimizations for now?

fe/fe-core/src/main/java/com/starrocks/sql/ast/expression/AnalyticExpr.java
```diff
@@ -138,8 +142,12 @@ public AnalyticExpr(FunctionCallExpr fnCall, List<Expr> partitionExprs, List<Ord
                 } else if (HintNode.HINT_ANALYTIC_SKEW.equalsIgnoreCase(hint)) {
                     this.skewHint = hint;
                     this.isSkewed = true;
+                } else if (HintNode.HINT_ANALYTIC_SKEW_EXPLICIT.equalsIgnoreCase(hint)) {
+                    this.skewHint = hint;
+                    this.skewColumn = skewColumn;
+                    this.skewValue = skewValue;
```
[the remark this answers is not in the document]
martinr0x: [d2b5cad](https://github.com/StarRocks/starrocks/pull/68739/commits/d2b5cad95a1e681b13c93eb774d373dfc1a6eb50)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -308,6 +315,34 @@ public BranchResult(OptExpression root, Map<ColumnRefOperator, ColumnRefOperator
...
+            return Collections.emptyList();
+        }
+
+        if (!(skewValue instanceof ConstantOperator val)) {
+            return Collections.emptyList();
+        }
+
+        // Verify that the skewed column is part of the partition expressions
+        if (!window.getPartitionExpressions().contains(col)) {
+            throw new StarRocksConnectorException("Can't find skew column");
```
stephen-shelby: Using StarRocksConnectorException here is inappropriate

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
```diff
@@ -2054,6 +2054,21 @@ public Expr visitAnalyticExpr(AnalyticExpr analyticExpr, Void context) {
...
             }
+
+            if (HintNode.HINT_ANALYTIC_SKEW_EXPLICIT.equalsIgnoreCase(analyticExpr.getSkewHint())) {
+                if (analyticExpr.getSkewColumn() == null) {
+                    throw new SemanticException("Window skew column must be specified when using explicit skew hint");
+                }
+                ExpressionAnalyzer.analyzeExpression(analyticExpr.getSkewColumn(), new AnalyzeState(), sourceScope, session);
+                if (analyticExpr.getSkewValue() != null) {
+                    if (!analyticExpr.getSkewValue().isConstant()) {
+                        throw new SemanticException("Window skew join value must be constant");
```
stephen-shelby: skew join is incorrect. and you should check skew value num. if exceeds 1, throw an exception

fe/fe-parser/src/main/java/com/starrocks/sql/ast/HintNode.java
stephen-shelby: Do we really need a new skew hint keyword? Could we reuse skewed and support an explicit form like [skewed|col(value)]? Then treat [skewed] as the existing boolean hint, and [skewed|...] as explicit (validated by presence of column/value). This would avoid introducing an extra keyword and keep syntax consistent.

martinr0x: Wouldn't it be more consistent with other hints(e.g join skew) to use [skew|col] here? 

Activating both optimisation with the same hint is a bit confusing in my opinion. Especially because two different session variables have to be active in order for the optimisation to get triggered.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
```diff
@@ -308,6 +315,33 @@ public BranchResult(OptExpression root, Map<ColumnRefOperator, ColumnRefOperator
...
+        if (!(skewColumn instanceof ColumnRefOperator col)) {
+            return Collections.emptyList();
+        }
+
+        // Verify that the skewed column is part of the partition expressions
+        if (!window.getPartitionExpressions().contains(col)) {
+            throw new SemanticException("Can't find skew column");
+        }
+
+        return skewValues.stream()
```
kangkaisen, line 341: do you think we need to add a cast check here? if the hint value type is different from the column type, will we get a wrong result?

```
return skewValues.stream()
        .map(v -> {
            if (!(v instanceof ConstantOperator c)) {
                throw new SemanticException("Window skew values must be constant");
            }
            Optional<ConstantOperator> cast = c.castTo(col.getType());
            if (cast.isEmpty()) {
                throw new SemanticException("Window skew value type mismatch: " +
                        c.getType() + " vs " + col.getType());
            }
            return new SkewedInfo(col, cast.get());
        })
        .toList();

```

martinr0x: Good point, for now it was failing silently as the rule would not be applied, but this is more explicit.

stephen-shelby (approved):

kangkaisen (approved):

PR #68752, opened 2026-01-30
[Enhancement] Add SQL expressions to project operator profile

lxynov:
## Why I'm doing:
- Generally useful.
- Particularly useful to us for profiling custom native function calls.

## What I'm doing:
- Add SQL expressions to project operator profile as info strings.

Example SQL:
```
select s_store_name, length(s_store_name) from store
```

Profile output:
```
...
        PROJECT (plan_node_id=1):
          CommonMetrics:
             - OperatorTotalTime: 83.235us
...
          UniqueMetrics:
             - ProjectExpressions: 30->DictDecode(31: s_store_name, [length()]), 31->31: s_store_name
...
```

EXPLAIN output (existing behavior):
```
mysql> explain select s_store_name, length(s_store_name) from store;
+-------------------------------------------------------------------------+
| Explain String                                                          |
+-------------------------------------------------------------------------+
| PLAN FRAGMENT 0                                                         |
|  OUTPUT EXPRS:6: s_store_name | 30: length                              |
|   PARTITION: RANDOM                                                     |
|                                                                         |
|   RESULT SINK                                                           |
|                                                                         |
|   2:Decode                                                              |
|   |  <dict id 31> : <string id 6>                                       |
|   |                                                                     |
|   1:Project                                                             |
|   |  <slot 30> : DictDecode(31: s_store_name, [length(<place-holder>)]) |
|   |  <slot 31> : 31: s_store_name                                       |
|   |                                                                     |
|   0:OlapScanNode                                                        |
|      TABLE: store                                                       |
|      PREAGGREGATION: ON                                                 |
|      partitions=1/1                                                     |
|      rollup: store                                                      |
|      tabletRatio=1/1                                                    |
|      tabletList=11745                                                   |
|      cardinality=12                                                     |
|      avgRowSize=5.25                                                    |
+-------------------------------------------------------------------------+
22 rows in set (0.00 sec)

mysql> 
```

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/project_node.cpp
+++ b/be/src/exec/project_node.cpp
@@ -53,6 +53,16 @@ ProjectNode::~ProjectNode() {
 
 Status ProjectNode::init(const TPlanNode& tnode, RuntimeState* state) {
     RETURN_IF_ERROR(ExecNode::init(tnode, state));
+
+    if (tnode.project_node.__isset.sql_project_exprs) {
+        _runtime_profile->add_info_string("ProjectExpressions", tnode.project_node.sql_project_exprs);
+        _sql_project_exprs = tnode.project_node.sql_project_exprs;
+    }
+    if (tnode.project_node.__isset.sql_common_exprs) {
+        _runtime_profile->add_info_string("CommonExpressions", tnode.project_node.sql_common_exprs);
+        _sql_common_exprs = tnode.project_node.sql_common_exprs;
+    }
+
     size_t column_size = tnode.project_node.slot_map.size();
     _expr_ctxs.reserve(column_size);
     _slot_ids.reserve(column_size);
@@ -94,7 +104,6 @@ Status ProjectNode::prepare(RuntimeState* state) {
 
     _expr_compute_timer = ADD_TIMER(runtime_profile(), "ExprComputeTime");
     _common_sub_expr_compute_timer = ADD_TIMER(runtime_profile(), "CommonSubExprComputeTime");
-
     return Status::OK();
 }
 
@@ -294,7 +303,8 @@ pipeline::OpFactories ProjectNode::decompose_to_pipeline(pipeline::PipelineBuild
 
     operators.emplace_back(std::make_shared<ProjectOperatorFactory>(
             context->next_operator_id(), id(), std::move(_slot_ids), std::move(_expr_ctxs),
-            std::move(_type_is_nullable), std::move(_common_sub_slot_ids), std::move(_common_sub_expr_ctxs)));
+            std::move(_type_is_nullable), std::move(_common_sub_slot_ids), std::move(_common_sub_expr_ctxs),
+            std::move(_sql_project_exprs), std::move(_sql_common_exprs)));
     // Initialize OperatorFactory's fields involving runtime filters.
     this->init_runtime_filter_for_operator(operators.back().get(), context, rc_rf_probe_collector);
     if (limit() != -1) {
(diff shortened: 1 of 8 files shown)


stdpain (dismissed):

be/src/exec/project_node.cpp
```diff
@@ -53,6 +53,16 @@ ProjectNode::~ProjectNode() {
 
 Status ProjectNode::init(const TPlanNode& tnode, RuntimeState* state) {
     RETURN_IF_ERROR(ExecNode::init(tnode, state));
+
+    if (tnode.project_node.__isset.sql_project_exprs) {
+        _runtime_profile->add_info_string("ProjectExpressions", tnode.project_node.sql_project_exprs);
+        _sql_project_exprs = tnode.project_node.sql_project_exprs;
```
kangkaisen, line 59: do you think we could get the _sql_project_exprs and _sql_common_exprs from _expr_ctxs and _common_sub_expr_ctxs, then we don't need to pass the sql_project_exprs and sql_common_exprs in the thrift, which could reduce the network cost.

lxynov: It's doable but has some engineering overhead. Basically we'd need to implement the same logic of `ExprExplainVisitor` on the BE. It'd be some non-trivial code since there are a few dozens of Expr types. Also we'll need to keep the implementations in sync. I also agree that sometimes project expressions can be huge so it's worth optimizing. So I'm not very sure. What do you think?

LiShuMing (dismissed):

fe/fe-core/src/test/java/com/starrocks/analysis/SelectStmtWithDecimalTypesNewPlannerTest.java
```diff
@@ -151,8 +151,11 @@ public void testMultiply() throws Exception {
                         " has_nullable_child:false, is_nullable:true, is_monotonic:true, is_index_only_filter:false)," +
                         " TExprNode(node_type:DECIMAL_LITERAL, type:TTypeDesc(types:[TTypeNode(type:SCALAR, scalar_type:TScalarType(type:DECIMAL128, precision:3, scale:2))])," +
                         " num_children:0, decimal_literal:TDecimalLiteral(value:3.14, integer_value:3A 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00), " +
-                        "output_scale:-1, has_nullable_child:false, is_nullable:false, is_monotonic:true, is_index_only_filter:false)])})";
+                        "output_scale:-1, has_nullable_child:false, is_nullable:false, is_monotonic:true, is_index_only_filter:false)])}, " +
+                        "sql_project_exprs:6->5: col_decimal128p20s3 * 3.14)";
+
         String plan = UtFrameUtils.getPlanThriftString(ctx, sql);
+        System.out.println(plan);
```
before-Sunrise: delete this

lxynov: good catch. done

lxynov:
Superseded by https://github.com/StarRocks/starrocks/pull/73005

PR #68788, opened 2026-02-02
[Enhancement] introduce memory allocator

silverbullet233:
## What I'm doing:

Part of the work at https://github.com/StarRocks/starrocks/issues/69128 
introduces the Allocator interface to manage memory allocation for custom data structures.

TODO: An Allocator interface was previously defined in `be/src/runtime/memory/allocator.h`, but its semantics are too redundant. It will be replaced with `memory::Allocator` in the future.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


xhumanoid:
@silverbullet233 do you have any plans to investigate performance of latest jemalloc ? 

starrocks use old stable version 5.3.0, but CH use latest available version with optimizations 
as example https://github.com/jemalloc/jemalloc/pull/2842

greatly improve for some type of workloads

be/src/runtime/memory/memory_allocator.h
```diff
@@ -0,0 +1,144 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+#include <atomic>
+#include <concepts>
+#include <cstdlib>
+
+#include "common/compiler_util.h"
```
alvin-phoenix-ai: please use base/compiler_util.h, common is just wrapper.

be/src/runtime/memory/memory_allocator.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+#include "runtime/memory/memory_allocator.h"
+
+#include <jemalloc/jemalloc.h>
+
+#include <cstdlib>
+#include <cstring>
+
+#include "gutil/strings/fastmem.h"
+#include "runtime/current_thread.h"
```
alvin-phoenix-ai: 1. split this into several cpp files, and put the native ones as low as possible, like base module.
2. Should not depend on current_thread.h, this belongs to runtime, which is higher level.

be/src/base/memory/counting_allocator.h
```diff
@@ -0,0 +1,126 @@
...
+    int64_t v{0};
+    ALWAYS_INLINE void add(int64_t delta) { v += delta; }
+    ALWAYS_INLINE int64_t value() const { return v; }
+};
+
+struct AtomicIntCounter {
+    std::atomic<int64_t> v{0};
+    ALWAYS_INLINE void add(int64_t delta) { v.fetch_add(delta, std::memory_order_relaxed); }
+    ALWAYS_INLINE int64_t value() const { return v.load(std::memory_order_relaxed); }
+};
```
satanson, line 44: Why not re-use RuntimeProfile::Counter?

be/src/base/memory/counting_allocator.h
```diff
@@ -0,0 +1,126 @@
...
+
+template <class BaseAllocator, class Counter>
+class CountingAllocator : public BaseAllocator {
```
satanson, line 47: Why not use CRTP pattern to define hierarchical classes?

be/src/base/memory/jemalloc_allocator.h
```diff
@@ -0,0 +1,37 @@
...
+    ~JemallocAllocator() override = default;
+
+    void* alloc(size_t size, size_t alignment = 0) override;
+    void* realloc(void* ptr, size_t old_size, size_t new_size, size_t alignment = 0) override;
+    void free(void* ptr, size_t size) override;
+    int64_t nallox(size_t size, int flags = 0) const override;
+
+    Allocator::MemoryKind memory_kind() const override { return Allocator::MemoryKind::kJemalloc; }
+
+    static constexpr bool throw_bad_alloc_on_failure() { return false; }
```
satanson, line 34: why is jemalloc  quiet when bad_alloc happens?

be/src/base/memory/malloc_allocator.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <cstdlib>
+#include <cstring>
+
+#include "base/compiler_util.h"
+#include "gutil/strings/fastmem.h"
+
+namespace starrocks::memory {
+
+template <bool clear_memory>
+void* MallocAllocator<clear_memory>::alloc(size_t size, size_t alignment) {
```
satanson, line 26: Will both MallocAllocator and JemallocAlloctor be used? or only one of Allocator  takes effect, if so how to choose which one two use?

be/src/base/memory/malloc_allocator.cpp
```diff
@@ -0,0 +1,93 @@
...
+            ret = std::calloc(size, 1);
+        } else {
+            ret = std::malloc(size);
+        }
+    } else {
+        int res = posix_memalign(&ret, alignment, size);
+        if (UNLIKELY(res != 0)) {
+            return nullptr;
+        }
+        if constexpr (clear_memory) {
```
satanson, line 39: clear_memoy means zero-out memory? how about using zero_out instead of clear_memory?

be/src/runtime/memory/tracked_allocator.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    return ptr;
+}
+
+template <class BaseAllocator>
+void* TrackedAllocator<BaseAllocator>::realloc(void* ptr, size_t old_size, size_t new_size, size_t alignment) {
+    int64_t old_alloc_size = BaseAllocator::nallox(old_size, 0);
+    int64_t new_alloc_size = BaseAllocator::nallox(new_size, 0);
+    int64_t delta = new_alloc_size - old_alloc_size;
+    if (delta > 0) {
```
satanson, line 81: if old_size is larger than new_size and they are of different size classes,  memory would be released, so delta <=0 should be also taken into consideration.

be/src/base/memory/counting_allocator.h
silverbullet233: > Why not re-use RuntimeProfile::Counter?

 Counter is too complex and take too much memory. in this scenario, we don't need a so much heavy component.

be/src/base/memory/jemalloc_allocator.h
silverbullet233: > why is jemalloc quiet when bad_alloc happens?

this is the behavior of malloc interface, when memory allocation failed, it returns nullptr instead of throwing exception

be/src/base/memory/malloc_allocator.cpp
silverbullet233: > Will both MallocAllocator and JemallocAlloctor be used? or only one of Allocator takes effect, if so how to choose which one two use?

MallocAllocator is used in ASAN mode to ensure that the address sanitizer functions properly; it is not used in production environments.

be/src/base/memory/malloc_allocator.cpp
silverbullet233: > clear_memoy means zero-out memory? how about using zero_out instead of clear_memory?

yeah. I think both names are acceptable, and clear_memory is also quite clear.

be/src/runtime/memory/tracked_allocator.cpp
silverbullet233: > if old_size is larger than new_size and they are of different size classes, memory would be released, so delta <=0 should be also taken into consideration.


This doesn't involve actual memory allocation, it only handles memory statistics. Memory limit checking will only be triggered when delta > 0,

satanson (approved):

trueeyu (approved):

PR #68816, opened 2026-02-02
[Enhancement] Support delete predicate handling in cloud-native tablet merge operations 

xiangguangyxg:
## Why I'm doing:                                                                                                                                                                                                                                                          
  When merging tablets in cloud-native (shared-data) mode, delete predicates from different                                                                                                                                                                                  
  tablets need to be handled correctly. Without proper handling:                                                                                                                                                                                                             
  1. Duplicate delete predicates (same version from different tablets) would be redundantly stored                                                                                                                                                                           
  2. Rowsets may not be ordered correctly relative to delete predicates, causing incorrect query results                                                                                                                                                                     
  3. DELETE transactions batched together could result in multiple predicates sharing the same version                                                                                                                                                                       
                                                                                                                                                                                                                                                                             
  ## What I'm doing:                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                             
  ### BE Changes:                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                             
  1. **Track rowset versions in txn_log_applier.cpp**                                                                                                                                                                                                                        
     - Set version field on rowsets during write, compaction, schema change, and replication operations                                                                                                                                                                      
     - This allows tracking when rowsets were created for proper delete predicate application                                                                                                                                                                                
                                                                                                                                                                                                                                                                             
  2. **Add delete predicate reordering during tablet merge (tablet_reshard.cpp)**                                                                                                                                                                                            
     - Implement `reorder_rowsets_by_delete_predicate` with helper functions:                                                                                                                                                                                               
       - `build_rowset_index_offsets`: Build offset mapping for tablets                                                                                                                                                                                                      
       - `build_tablet_group_states`: Build tablet states and collect predicate versions                                                                                                                                                                                     
       - `build_reordered_rowsets`: Reorder rowsets by predicate version                                                                                                                                                                                                     
       - `cleanup_orphan_schema_mappings`: Clean up orphan schema mappings after deduplication                                                                                                                                                                               
     - Deduplicate identical delete predicates from different tablets                                                                                                                                                                                                        
     - Ensure data rowsets are ordered correctly relative to their protecting predicates                                                                                                                                                                                     
                                                                                                                                                                                                                                                                             
  3. **Add comprehensive tests (tablet_reshard_test.cpp)**                                                                                                                                                                                                                   
     - Test tablets with same predicate versions (deduplication)                                                                                                                                                                                                             
     - Test tablets with different predicate versions                                                                                                                                                                                                                        
     - Test tablets with no predicates                                                                                                                                                                                                                                       
     - Test single tablet with predicates                                                                                                                                                                                                                                    
     - Test all-predicate edge case                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             
  ### FE Changes:                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                             
  4. **Prevent DELETE transaction batching (DatabaseTransactionMgr.java)**                                                                                                                                                                                                   
     - Add DELETE to the list of transaction types that should not be batched                                                                                                                                                                                                
     - Each DELETE transaction gets its own version for proper ordering during tablet merge                                                                                                                                                                                  
     - Refactor condition to check LoadJobSourceType instead of TransactionType                                                                                                                                                                                              
                                                                                                                                                                                                                                                                             
  5. **Add unit test for DELETE batching behavior (DatabaseTransactionMgrTest.java)**                                                                                                                                                                                        
     - Verify DELETE transactions are isolated in their own batch                                                                                                                                                                                                            
     - Verify normal transactions after DELETE can still be batched together 

Fixes #64986

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


be/src/storage/lake/tablet_reshard.cpp
```diff
@@ -119,6 +122,383 @@ static void set_all_data_files_shared(TabletMetadataPB* tablet_metadata) {
...
+            }
+        }
+
+        if (predicate_rowset_indices.empty()) {
+            return Status::Corruption("Delete predicate version not found during merge reorder");
+        }
+
+        // Keep a single predicate rowset for this version across tablets.
+        reordered_rowsets->emplace_back(new_metadata.rowsets(static_cast<int>(predicate_rowset_indices.front())));
+        *removed_predicate_count += predicate_rowset_indices.size() - 1;
```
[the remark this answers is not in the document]
xiangguangyxg: delete predicates end up with the same version have the same contents

srlch (approved):

be/src/storage/lake/txn_log_applier.cpp
```diff
@@ -522,6 +522,7 @@ class PrimaryKeyTxnLogApplier : public TxnLogApplier {
             DCHECK(rowset.has_id());
             auto new_rowset = _metadata->add_rowsets();
             new_rowset->CopyFrom(rowset);
+            new_rowset->set_version(_new_version);
```
wyb, line 525: 1. primary key tablet does not have delete predicate.
2. why apply write/compaction log does not set new version?

be/src/storage/lake/txn_log_applier.cpp
```diff
@@ -1007,6 +1011,7 @@ class NonPrimaryKeyTxnLogApplier : public TxnLogApplier {
             auto output_rowset = _metadata->mutable_rowsets(first_idx);
             output_rowset->CopyFrom(op_compaction.output_rowset());
             output_rowset->set_id(_metadata->next_rowset_id());
+            output_rowset->set_version(_new_version);
```
wyb, line 1014: `_new_version` is not the real data version in compaction, but the delete predicate filter needs real data version?

be/src/storage/lake/txn_log_applier.cpp
xiangguangyxg: Both pk tablet and non-pk tablet are set rowset version even pk tablet does not have delete predicate.
Pk tablet apply write/compaction log already set rowset version in MetaFileBuilder.

be/src/storage/lake/txn_log_applier.cpp
xiangguangyxg: delete predicate filter uses rowset index, and tablet merge only use the version to identify same delete predicate from different tablet.

wyb (approved):

PR #68821, opened 2026-02-02
[Feature] Add Debian (.deb) packaging support for FE and BE

### LINKED ISSUE #68205: [Feature Request] Provide official Debian (.deb) and RPM packages for standard installation
- [x] https://github.com/StarRocks/starrocks/pull/68821
## Feature request

Currently, StarRocks is distributed primarily via binary tarballs or Docker images. While this works for testing, it creates friction for bare-metal production deployments and standard Linux administration:

1.  **Non-Standard Paths:** When using the tarball, logs and configurations reside inside the installation directory (e.g., `StarRocks-x.x/fe/log`) instead of standard Linux Standard Base (LSB) paths like `/var/log/starrocks` and `/etc/starrocks`.
2.  **Service Management:** There is no native `systemd` integration out of the box. Administrators have to manually write unit files to manage the `fe` and `be` services or use start/stop scripts manually.
3.  **Upgrade Complexity:** Upgrading requires downloading a new tarball and manually migrating config files, whereas standard packages allow seamless updates via `apt-get upgrade` or `yum update`.

I would like StarRocks to provide official package repositories for major Linux distributions (Debian/Ubuntu and RHEL/CentOS).

Similar to how **Apache Cassandra** handles installation, we should be able to:
1.  Add a StarRocks repository key and URL.
2.  Run commands like `sudo apt-get install starrocks-fe starrocks-be`.
3.  Have the package automatically:
    * Place configs in `/etc/starrocks/`.
    * Place logs in `/var/log/starrocks/`.
    * Place data in `/var/lib/starrocks/`.
    * Register and enable `systemd` services (e.g., `systemctl start starrocks-fe`).

* **Docker:** While Docker is available, it adds an abstraction layer that isn't always desirable for high-performance bare-metal setups where we want direct OS control.
* **Ansible/Chef/Puppet:** We currently have to write complex playbooks to download tarballs, move files to `/etc/` and `/var/`, and create custom systemd files to simulate a package installation. This is high maintenance.

**Additional context**
Most enterprise-grade data systems (Cassandra, MySQL, PostgreSQL, ClickHouse) provide native package management support. Adding this would significantly lower the barrier to entry for SysAdmins and DevOps engineers accustomed to standard Linux workflows.

Reference examples:
* Cassandra Debian Repo: https://cassandra.apache.org/_/download.html

sunghj1118:
## Why I'm doing:
Currently, StarRocks installation on Debian-based systems (Ubuntu/Debian) relies on manual tarball extraction and manual environment configuration. This is error-prone and does not follow the Linux Standard Base (LSB) hierarchy. Providing native .deb packages simplifies deployment ensures consistent directory structures across clusters and enables modern process management via `systemd`.

## What I'm doing:
Introduced a Debian packaging infrastructure for both Frontend (FE) and Backend (BE) components.

**Implementation Details:**
* Standardized Path Mapping:
    * Binaries & Libraries: Moved to `/usr/lib/starrocks/` to keep the application isolated.
    * Configuration: Relocated to `/etc/starrocks/` and registered as `conffiles` to ensure user settings persist through package upgrades.
    * Logs: Directed to `/var/log/starrocks/`.
    * Metadata/Data: Directed to `/var/lib/starrocks/`.

* Automated Lifecycle Management: * Added a `postinst` script to automate the creation of a dedicated starrocks system user/group.
    * Automated directory permission handling (chown/chmod) for log and data paths.

* Systemd Integration: * Created `starrocks-fe.service` and `starrocks-be.service` units.
    * Configured services to run as the starrocks user with appropriate resource limits (`LimitNOFILE`).
    * Integrated `--daemon` to allow `systemd` to track process health correctly.

* Build Automation: Provided a cross-platform `build.sh` script (compatible with macOS/BSD and Linux/GNU `sed`) to generate the `.deb` files from the official build output.

Fixes #68205 

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [x] Miscellaneous: Introduces a new installation method and standardized directory hierarchy for Debian-based distributions.

## Checklist:

  - [x] This pr needs auto generate documentation
- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


packaging/debian/build.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+set -e
+
+# --- Configuration ---
+VERSION="4.0.4"
+ARCH="amd64"
+TARBALL="../../build_output/StarRocks-4.0.4-ubuntu-amd64.tar.gz"
```
kevincai: FE and BE build artifacts are in `out/fe` and `out/be` respectively, we can either assume these are the default location for the artifacts or provide commandline argument to specify a different location.

packaging/debian/control.be
```diff
@@ -0,0 +1,11 @@
+Package: starrocks-be
```
kevincai, line 1: is it possible to auto generate this control file and match the real release version?

packaging/debian/systemd/starrocks-be.service
```diff
@@ -0,0 +1,17 @@
...
+
+[Service]
+Type=simple
+User=starrocks
+Group=starrocks
+Environment="STARROCKS_HOME=/usr/lib/starrocks/be"
+Environment="PID_DIR=/var/run/starrocks"
+ExecStart=/usr/lib/starrocks/be/bin/start_be.sh --conf_dir /etc/starrocks/be --daemonize 0
+Restart=on-failure
+LimitNOFILE=65536
```
kevincai: raise this limit to 4x65536 for BE node.

packaging/debian/build.sh
sunghj1118: Done. I've updated build.sh to support positional arguments for the FE and BE source directories (defaulting to `../../out/fe` and `../../out/be`).

packaging/debian/control.be
sunghj1118: Yes, I've implemented this. The control files now use a placeholder version (0.0.0), and the `build.sh` script dynamically injects the correct release version using sed during the build process.

packaging/debian/systemd/starrocks-be.service
sunghj1118: Done. I've increased `LimitNOFILE` to 262144.

packaging/debian/systemd/starrocks-be.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=StarRocks Backend (BE)
+After=network.target
+
+[Service]
+Type=simple
+User=starrocks
+Group=starrocks
+RuntimeDirectory=starrocks
+Environment="STARROCKS_HOME=/usr/lib/starrocks/be"
```
kevincai: not sure if it is a good idea to set HOME to this location, usually `/usr/lib` sits the same partition/device as the `/`, and starrocks will write logs to $STARROCKS_HOME/log and write data to $STARROCKS_HOME/data by default, if user doesn't change the default configuration, it can easily cause root disk full which will take down the entire system.

packaging/debian/systemd/starrocks-be.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=StarRocks Backend (BE)
```
kevincai, line 2: need to consider how to add a CN dpkg with the same artifacts but different home and configurations for shared-data deployment

kevincai (commented):
The overall looks good to me, can you run `dpkg -L` with the built `.deb` and paste the result here. Would like to see the content of the .deb for an overview.

packaging/debian/systemd/starrocks-be.service
sunghj1118: That's a great point. To prevent the root partition from filling up, I've included sed patches in the `build.sh` script that automatically update the default be.conf and fe.conf during the build process.

Currently, the package redirects these paths to LSB-compliant locations:
- logs: Redirected from `$STARROCKS_HOME/log` to `/var/log/starrocks/`
- data/meta: Redirected from `$STARROCKS_HOME/meta` or storage_root_path to `/var/lib/starrocks/`

This ensures that even with default settings, StarRocks writes its heavy data to `/var`.

packaging/debian/systemd/starrocks-be.service
sunghj1118: I see your point about Compute Nodes (CN). Since CN shares the same artifacts as BE but uses `cn.conf`, I can update the build.sh loop to support a third component `cn`. I have fixed the `build.sh` such that it packages the `cn.conf` (if present) into `/etc/starrocks/cn/` and provides a separate `starrocks-cn.service` for systemd.

kevincai (commented):
@sunghj1118 AI reviewer provides some valid comments, need to address them before merge.

packaging/debian/postinst
```diff
@@ -0,0 +1,151 @@
...
+        # - If a deployment requires auto-generated or runtime-writable configuration,  
+        #   use a separate writable location (e.g. under /var/lib/starrocks) or adjust  
+        #   permissions for the specific files as part of local administration.  
+        if [ -n "$COMP" ]; then  
+            CONF_DIR="/etc/starrocks/$COMP"  
+            if [ -d "$CONF_DIR" ]; then  
+                chown -R root:starrocks "$CONF_DIR"  
+                find "$CONF_DIR" -type d -exec chmod 750 {} \;  
+                find "$CONF_DIR" -type f -exec chmod 640 {} \;  
+            fi
```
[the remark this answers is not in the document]
sunghj1118: StarRocks FE/BE/CN are designed to store all runtime state (metadata, storage, and PIDs) in /var/lib, /var/log, or /run. The /etc directory is treated as read-only configuration, which is why 640/750 is preferred to maintain a higher security posture and prevent the service from self-modifying its configuration.

packaging/debian/control.be
```diff
@@ -0,0 +1,14 @@
...
+Version: 0.0.0
+Section: database
+Priority: optional
+Architecture: amd64
+Depends: bash, procps, libssl1.0.0 | libssl1.1 | libssl3, libibverbs1
```
kevincai: thought BE/CN doesn't explicit require libssl runtime, isn't it?

kevincai: and what does this `libibverbs1` for?

sunghj1118: while it doesn't use libssl for core compute, it is required for object storage and ssl connections. I added varying libssl versions to support portability across different debian/ubuntu versions and common LTS versions

sunghj1118: meanwhile libibverbs1 is required for RDMA support or links in the storage engine or data shuffle layer

kevincai: these dependency shall not be exactly required by starrocks, better remove it. For cases that willing to bundle install packages along with starrocks, a dummy package can be created with just dependencies, so that it can pull all desired pkgs installed all the same time.

kevincai:
@sunghj1118 thanks for the effort, the AI code reviewer still provides some comments, please check if they are still valid ones.

packaging/debian/control.be
sunghj1118: Got it. I'll remove the libssl and libiverbs1 requirements as you recommended.

kevincai (approved):

alvin-phoenix-ai (approved):

PR #68860, opened 2026-02-03
[BugFix] fix the issue of not being able to find LookUpDispatcher

silverbullet233:
## Why I'm doing:

unstable case: sql/test_global_late_mterialization/R/test_glm_topn@test_glm_topn_iceberg_v3

## What I'm doing:

In multi-BE lookup/fetch execution, remote LookUp RPCs can fail with "can't find LookUpDispatcher" (NotFound), causing queries to fail or retry.

**Phenomenon:**
- Queries with enabling global late materialization send LookUp RPCs from producer BEs to the BE that runs the LookUp/Fetch plan.
- The RPC handler `lookup()` used `get_dispatcher()` and only looked up an existing dispatcher. If the dispatcher did not exist, it returned `Status::NotFound` and the request failed.
- The dispatcher is created in `LookUpNode::init()` / `FetchNode::init()` when the fragment that contains the LookUp/Fetch operator is set up. Due to execution order and parallelism, the first LookUp RPC can arrive **before** that fragment has run `init()` on the target BE, so the dispatcher does not exist yet and the lookup fails.

**Root cause:**
1. **Ordering assumption:** The code assumed the dispatcher was always created locally before any remote LookUp RPC arrived. When RPCs arrived first, `get_dispatcher()` returned NotFound.
2. **Rigid request queues:** `add_request()` used `DCHECK(_request_queues.contains(request_tuple_id))` and `_request_queues.at(request_tuple_id)`, so only pre-registered tuple_ids were allowed. A lazily created dispatcher (created on first RPC) would only have one tuple_id; later requests with other tuple_ids would hit the check or wrong queue.
3. **Unsafe second lookup in try_get:** `try_get()` found `target_tuple_id` by iterating, then did `_request_queues.at(target_tuple_id)` to get the queue. With concurrent updates to `_request_queues`, this second lookup could be inconsistent or race with the map; the map was also not thread-safe for concurrent add/try_get.

## What I'm doing:

1. **Lazy creation of dispatcher on first LookUp RPC**  
   - Add `get_or_create_dispatcher(ctx)` that creates a dispatcher under `(query_id, lookup_node_id)` when it does not exist, using the first request’s `request_tuple_id`.  
   - In `lookup()`, use `get_or_create_dispatcher(ctx)` instead of `get_dispatcher()`, so the first incoming RPC creates the dispatcher if the local fragment has not yet run `init()`.

2. **Dynamic request queues in LookUpDispatcher**  
   - In `add_request()`, use `_request_queues.lazy_emplace(request_tuple_id, ...)` to create a queue for any `request_tuple_id` on demand, instead of DCHECK + `at()`.  
   - This allows both “dispatcher created by init() with full tuple_ids” and “dispatcher created by first RPC with a single tuple_id” to accept subsequent requests for any tuple_id.

3. **Safe use of queue in try_get()**  
   - In `try_get()`, keep the selected queue pointer `target_queue` from the iteration and use it directly for `try_dequeue_bulk()`, instead of calling `_request_queues.at(target_tuple_id)` again.  
   - Switch `_request_queues` to `phmap::parallel_flat_hash_map` with `bthread::Mutex` so concurrent `add_request()` and `try_get()` are safe.

4. **Cleanup**  
   - Remove unused `RuntimeState*` from `LookUpDispatcher` and `create_dispatcher()` (including constructor and call sites).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/fetch_node.cpp
+++ b/be/src/exec/fetch_node.cpp
@@ -55,8 +55,8 @@ Status FetchNode::init(const TPlanNode& tnode, RuntimeState* state) {
         tuple_ids.emplace_back(tuple_id);
     }
 
-    _dispatcher = state->exec_env()->lookup_dispatcher_mgr()->create_dispatcher(state, state->query_id(),
-                                                                                _target_node_id, tuple_ids);
+    _dispatcher = state->exec_env()->lookup_dispatcher_mgr()->create_dispatcher(state->query_id(), _target_node_id,
+                                                                                tuple_ids);
     for (const auto& tuple_id : tuple_ids) {
         const auto& tuple_desc = state->desc_tbl().get_tuple_descriptor(tuple_id);
         for (const auto& slot : tuple_desc->slots()) {
--- a/be/src/exec/lookup_node.cpp
+++ b/be/src/exec/lookup_node.cpp
@@ -47,8 +47,7 @@ Status LookUpNode::init(const TPlanNode& tnode, RuntimeState* state) {
     for (const auto& [tuple_id, row_pos_desc] : _row_pos_descs) {
         tuple_ids.emplace_back(tuple_id);
     }
-    _dispatcher =
-            state->exec_env()->lookup_dispatcher_mgr()->create_dispatcher(state, state->query_id(), id(), tuple_ids);
+    _dispatcher = state->exec_env()->lookup_dispatcher_mgr()->create_dispatcher(state->query_id(), id(), tuple_ids);
 
     return Status::OK();
 }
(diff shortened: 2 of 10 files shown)


alvin-phoenix-ai (changes requested):
This is not a UT, please use the correct title for this PR.

LiShuMing (approved):

satanson (approved):

PR #68877, opened 2026-02-03
[Enhancement] Support fast cancel for lake DeltaWriter in shared-data mode

### LINKED ISSUE #68782: Support fast cancel for lake DeltaWriter in shared-data mode
## Enhancement

in #15514 , an enhancement for load support fast cancel  was introduced

```
If the load task times out or the user terminates early, the resources used by load cannot be released immediately.
So we need to support FastCancel and release resources as soon as possible.
This is the first pr: Modify the implement of cancel, only set status, will not be blocked by I/O, queue or other behavior.
```

However, the fast cancel only works in shared-nothing cluster. We need to add this enhancement in shared-data mode, which is a must-need feature for resource usage.

wxl24life:
## Why I'm doing:
After transaction abort, import tasks in shared-data (lake) mode did not support fast cancel, which caused resource waste. This commit implements fast cancel support for lake::DeltaWriter to address this issue.

## What I'm doing:

Fixes #68782

- Add cancel() method to lake::DeltaWriter and lake::DeltaWriterImpl
- Add _cancelled atomic flag and _cancel_status to track cancellation state
- Check cancellation status in write() and flush_async() methods
- Implement LakeTabletsChannel::cancel() to invoke delta writer cancel

The fast cancel mechanism allows:
1. Immediate rejection of subsequent write/flush operations after cancel
2. Cancellation of pending flush tasks via FlushToken::cancel()
3. Thread-safe cancellation that can be called concurrently

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/runtime/lake_tablets_channel.cpp
+++ b/be/src/runtime/lake_tablets_channel.cpp
@@ -78,7 +78,7 @@ public:
     Status incremental_open(const PTabletWriterOpenRequest& params, PTabletWriterOpenResult* result,
                             std::shared_ptr<OlapTableSchemaParam> schema) override;
 
-    void cancel() override;
+    void cancel(const std::string& reason) override;
 
     void abort() override;
 
@@ -850,8 +850,12 @@ void LakeTabletsChannel::abort() {
     }
 }
 
-void LakeTabletsChannel::cancel() {
-    //TODO: Current LakeDeltaWriter don't support fast cancel
+void LakeTabletsChannel::cancel(const std::string& reason) {
+    std::shared_lock<bthreads::BThreadSharedMutex> l(_rw_mtx);
+    auto cancel_status = Status::Cancelled(reason.empty() ? "tablet channel cancelled" : reason);
+    for (auto& it : _delta_writers) {
+        it.second->cancel(cancel_status);
+    }
 }
 
 StatusOr<std::unique_ptr<LakeTabletsChannel::WriteContext>> LakeTabletsChannel::_create_write_context(
--- a/be/src/storage/lake/delta_writer.cpp
+++ b/be/src/storage/lake/delta_writer.cpp
@@ -18,6 +18,7 @@
 #include <fmt/format.h>
 
 #include <memory>
+#include <shared_mutex>
 #include <utility>
 
 #include "agent/master_info.h"
@@ -143,6 +144,8 @@ public:
 
     void close();
 
+    void cancel(const Status& st);
+
     [[nodiscard]] int64_t partition_id() const { return _partition_id; }
 
     [[nodiscard]] int64_t tablet_id() const { return _tablet_id; }
@@ -204,6 +207,8 @@ public:
     bool already_finished() const { return _already_finished; }
 
 private:
+    Status current_cancel_status() const;
+
     Status reset_memtable();
 
     Status fill_auto_increment_id(Chunk& chunk);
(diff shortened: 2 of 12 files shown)


be/src/runtime/lake_tablets_channel.cpp
```diff
@@ -851,7 +851,12 @@ void LakeTabletsChannel::abort() {
 }
 
 void LakeTabletsChannel::cancel() {
-    //TODO: Current LakeDeltaWriter don't support fast cancel
+    std::shared_lock<bthreads::BThreadSharedMutex> l(_rw_mtx);
+    for (auto& it : _delta_writers) {
+        if (auto* writer = it.second->delta_writer(); writer != nullptr) {
+            writer->cancel(Status::Cancelled("transaction cancelled"));
```
kevincai: is this error message accurate? LoadChannel can also cancel the tablet channel requested by external RPC.

be/src/storage/lake/delta_writer.cpp
```diff
@@ -922,6 +942,31 @@ void DeltaWriterImpl::close() {
...
+        return;
+    }
+
+    // Only set cancelled once
+    bool expected = false;
+    if (!_cancelled.compare_exchange_strong(expected, true, std::memory_order_acq_rel)) {
+        return;
+    }
+
+    {
```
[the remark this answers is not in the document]
kevincai: This is a valid point

be/src/storage/lake/delta_writer.cpp
```diff
@@ -293,6 +294,12 @@ class DeltaWriterImpl {
     // 2. After finish completes, txnlog is generated with all data files. Any subsequent write
     //    tasks will have their data discarded, resulting in data loss.
     bool _already_finished = false;
+
+    // Whether cancel() has been called. Used to support fast cancel when transaction is aborted.
+    // Once set to true, subsequent write/flush operations will fail quickly with the cancel status.
+    std::atomic<bool> _cancelled{false};
+    mutable std::mutex _cancel_lock;
+    Status _cancel_status;
```
xiangguangyxg: could remove cancelled atomic flag, only use a cancel status and a **shared** lock to protect it.

wxl24life: Thanks for the suggestion. I’d prefer to keep the atomic flag. write() and flush_async() are hot paths; relying only on a lock would add contention and overhead even when not cancelled. The current design uses the atomic as a fast-path to avoid taking the lock, while the mutex only protects the Status payload. This keeps the common path lock-free and still guarantees correct cancellation semantics. If we really want to simplify, we could consider an atomic shared_ptr<Status>, but that’s more complex with limited benefit.

xiangguangyxg: using a shared lock will not increase extra overhead in the read only path than a atomic flag

be/src/runtime/lake_tablets_channel.cpp
```diff
@@ -851,7 +851,12 @@ void LakeTabletsChannel::abort() {
 }
 
 void LakeTabletsChannel::cancel() {
-    //TODO: Current LakeDeltaWriter don't support fast cancel
+    std::shared_lock<bthreads::BThreadSharedMutex> l(_rw_mtx);
+    for (auto& it : _delta_writers) {
+        if (auto* writer = it.second->delta_writer(); writer != nullptr) {
+            writer->cancel(Status::Cancelled("tablet channel cancelled"));
```
kevincai: check the load channel  `cancel()` interface, it contains cancel reason, the reason can be passed down to TabletChannel and writer, so the message can be as accurate as possible.

kevincai (dismissed):

xiangguangyxg (approved):

kevincai (approved):

wyb (approved):

satanson (approved):

PR #68901, opened 2026-02-04
[BugFix] Fix Azure ABFS/WASB FileSystem cache key to include container

banmoy:
## Why I'm doing:

`FILES()` may access the wrong Azure ADLS Gen2 / Blob container when multiple containers under the same storage account are used. FE caches Hadoop `FileSystem` instances by a cache key that previously only included `scheme://host`, which drops the container information encoded in the URI authority (`container@account...`) for ABFS(S)/WASB(S). This can cause different containers to reuse the same `AzureBlobFileSystem` instance and lead to container/path mismatch.

## What I'm doing:

- Build the `FileSystem` cache key using `scheme://authority` for Azure-related schemes (abfs/abfss/wasb/wasbs) so the container is part of the identity.
- Add/extend FE unit tests to verify Azure `FileSystem` caching is container-aware (different containers do not share the same cached `FileSystem`).

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
+++ b/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
@@ -17,9 +17,10 @@
 
 package com.starrocks.fs.hdfs;
 
+import java.util.Objects;
+
 public class HdfsFsIdentity {
 
-    // A URI identity key, which may be built from either host or authority.
     private final String uriIdentity;
     private final String ugiInfo;
 
@@ -30,12 +31,7 @@ public class HdfsFsIdentity {
 
     @Override
     public int hashCode() {
-        final int prime = 31;
-        int result = 1;
-        result = prime * result
-                + ((uriIdentity == null) ? 0 : uriIdentity.hashCode());
-        result = prime * result + ((ugiInfo == null) ? 0 : ugiInfo.hashCode());
-        return result;
+        return Objects.hash(uriIdentity, ugiInfo);
     }
 
     @Override
--- a/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
+++ b/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
@@ -19,7 +19,7 @@ package com.starrocks.fs.hdfs;
 
 public class HdfsFsIdentity {
 
-    // A URI identity key, which may be built from either host or authority,
+    // A URI identity key, which may be built from either host or authority.
     private final String uriIdentity;
     private final String ugiInfo;
 
--- a/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
+++ b/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
@@ -19,11 +19,12 @@ package com.starrocks.fs.hdfs;
 
 public class HdfsFsIdentity {
 
-    private final String hostName;
+    // A URI identity key, which may be built from either host or authority,
+    private final String uriIdentity;
     private final String ugiInfo;
 
-    public HdfsFsIdentity(String hostName, String ugiInfo) {
-        this.hostName = hostName;
+    public HdfsFsIdentity(String uriIdentity, String ugiInfo) {
+        this.uriIdentity = uriIdentity;
         this.ugiInfo = ugiInfo;
     }
 
@@ -32,7 +33,7 @@ public class HdfsFsIdentity {
         final int prime = 31;
         int result = 1;
         result = prime * result
-                + ((hostName == null) ? 0 : hostName.hashCode());
+                + ((uriIdentity == null) ? 0 : uriIdentity.hashCode());
         result = prime * result + ((ugiInfo == null) ? 0 : ugiInfo.hashCode());
         return result;
     }
@@ -49,11 +50,11 @@ public class HdfsFsIdentity {
             return false;
         }
         HdfsFsIdentity other = (HdfsFsIdentity) obj;
-        if (hostName == null) {
-            if (other.hostName != null) {
+        if (uriIdentity == null) {
+            if (other.uriIdentity != null) {
                 return false;
             }
-        } else if (!hostName.equals(other.hostName)) {
+        } else if (!uriIdentity.equals(other.uriIdentity)) {
             return false;
         }
         if (ugiInfo == null) {
--- a/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
+++ b/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
@@ -52,7 +52,6 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InterruptedIOException;
-import java.net.URI;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.Enumeration;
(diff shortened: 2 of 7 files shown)


fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
```diff
@@ -704,6 +705,32 @@ public HdfsFs getGoogleFileSystem(String path, Map<String, String> loadPropertie
...
+     * @param uri the parsed URI of the file path
+     * @param useAuthorityForAllSchemes whether to use authority for non-Azure schemes
+     * @return the identity string in the format "scheme://authority" or "scheme://host"
+     */
+    private String getUriIdentity(URI uri, boolean useAuthorityForAllSchemes) {
+        String scheme = uri.getScheme();
+        boolean useAuthority = useAuthorityForAllSchemes || ABFS_SCHEME.equalsIgnoreCase(scheme)
+                || ABFSS_SCHEME.equalsIgnoreCase(scheme) || WASB_SCHEME.equalsIgnoreCase(scheme)
+                || WASBS_SCHEME.equalsIgnoreCase(scheme);
+        String identity = useAuthority ? uri.getAuthority() : uri.getHost();
```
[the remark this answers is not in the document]
banmoy: just follow the original behaviour

fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
```diff
@@ -32,7 +33,7 @@ public int hashCode() {
         final int prime = 31;
         int result = 1;
         result = prime * result
-                + ((hostName == null) ? 0 : hostName.hashCode());
+                + ((uriIdentity == null) ? 0 : uriIdentity.hashCode());
         result = prime * result + ((ugiInfo == null) ? 0 : ugiInfo.hashCode());
         return result;
```
xiangguangyxg: use Objects.hash(uriIdentity, ugiInfo) is simpler

fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
```diff
@@ -704,6 +705,34 @@ public HdfsFs getGoogleFileSystem(String path, Map<String, String> loadPropertie
...
+     * @return the identity string in the format "scheme://authority" or "scheme://host"
+     */
+    private String getUriIdentity(URI uri, boolean useAuthorityForAllSchemes) {
+        String scheme = uri.getScheme();
+        Preconditions.checkArgument(!Strings.isNullOrEmpty(scheme),
+                "URI scheme must not be null or empty: %s", uri);
+        boolean useAuthority = useAuthorityForAllSchemes || ABFS_SCHEME.equalsIgnoreCase(scheme)
+                || ABFSS_SCHEME.equalsIgnoreCase(scheme) || WASB_SCHEME.equalsIgnoreCase(scheme)
+                || WASBS_SCHEME.equalsIgnoreCase(scheme);
+        String identity = useAuthority ? uri.getAuthority() : uri.getHost();
```
xiangguangyxg: Is it feasible to use authority uniformly?

kevincai: shall promote to use authority as always, which should be flexible than original behavior.

So the new configure var is not necessary.

fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsIdentity.java
banmoy: fixed

fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
banmoy: fixed

fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
```diff
@@ -704,6 +705,27 @@ public HdfsFs getGoogleFileSystem(String path, Map<String, String> loadPropertie
...
+     * @param uri the parsed URI of the file path
+     * @return the identity string in the format "scheme://authority"
+     */
+    private String getUriIdentity(URI uri) {
+        String scheme = uri.getScheme();
+        Preconditions.checkArgument(!Strings.isNullOrEmpty(scheme),
+                "URI scheme must not be null or empty: %s", uri);
+        String identity = uri.getAuthority();
+        if (identity == null) {
+            identity = uri.getHost();
```
kevincai: if authority is null, likely the host is null too.

fe/fe-core/src/test/java/com/starrocks/fs/HdfsFsManagerTest.java
```diff
@@ -198,4 +199,80 @@ public void testAdls2InvalidEndpointThrows() {
...
+        String abfssPath2 = "abfss://container1@account.dfs.core.windows.net/path2/file.parquet";
+        String abfssPath3 = "abfss://container2@account.dfs.core.windows.net/path1/file.parquet";
+        String abfssPath4 = "abfss://container3@account.dfs.core.windows.net/path1/file.parquet";
+        testFileSystemCache(adlsProperties, Pair.create(abfssPath1, abfssPath2), Pair.create(abfssPath3, abfssPath4));
+
+        // Test abfs:// scheme (Azure Data Lake Storage Gen2 without SSL)
+        String abfsPath1 = "abfs://container1@account.dfs.core.windows.net/path1/file.parquet";
+        String abfsPath2 = "abfs://container1@account.dfs.core.windows.net/path2/file.parquet";
+        String abfsPath3 = "abfs://container2@account.dfs.core.windows.net/path1/file.parquet";
+        String abfsPath4 = "abfs://container3@account.dfs.core.windows.net/path1/file.parquet";
```
kevincai, line 236: add test to cover the case that same host with different port, cache to different fs instance.

e.g. 
```
abfs://container3@account.dfs.core.windows.net:8000
abfs://container3@account.dfs.core.windows.net:8080
```

banmoy: seems abfs uri does not support port. Do I miss something?
https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri#uri-syntax

kevincai: or take a different one. `hdfs://root@10.0.0.1:8030/path/...`, it is a standard URI following the BCNF definition in the rfc2396: https://www.ietf.org/rfc/rfc2396.txt 

```
A. Collected BNF for URI

      URI-reference = [ absoluteURI | relativeURI ] [ "#" fragment ]
      absoluteURI   = scheme ":" ( hier_part | opaque_part )
      relativeURI   = ( net_path | abs_path | rel_path ) [ "?" query ]

      hier_part     = ( net_path | abs_path ) [ "?" query ]
      opaque_part   = uric_no_slash *uric

      uric_no_slash = unreserved | escaped | ";" | "?" | ":" | "@" |
                      "&" | "=" | "+" | "$" | ","

      net_path      = "//" authority [ abs_path ]
      abs_path      = "/"  path_segments
      rel_path      = rel_segment [ abs_path ]

      rel_segment   = 1*( unreserved | escaped |
                          ";" | "@" | "&" | "=" | "+" | "$" | "," )

      scheme        = alpha *( alpha | digit | "+" | "-" | "." )

      authority     = server | reg_name

      reg_name      = 1*( unreserved | escaped | "$" | "," |
                          ";" | ":" | "@" | "&" | "=" | "+" )

      server        = [ [ userinfo "@" ] hostport ]
      userinfo      = *( unreserved | escaped |
                         ";" | ":" | "&" | "=" | "+" | "$" | "," )

      hostport      = host [ ":" port ]
      host          = hostname | IPv4address
      hostname      = *( domainlabel "." ) toplabel [ "." ]
      domainlabel   = alphanum | alphanum *( alphanum | "-" ) alphanum
      toplabel      = alpha | alpha *( alphanum | "-" ) alphanum
      IPv4address   = 1*digit "." 1*digit "." 1*digit "." 1*digit
      port          = *digit
```

banmoy: hdfs does not use `getFileSystemByCloudConfiguration` to get file system
image

image

banmoy: at least for now, those schemes using `getFileSystemByCloudConfiguration` seem not support port

kevincai: ok

banmoy: `getDistributedFileSystem` already uses authority
image

kevincai (approved):

fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
banmoy: fixed

xiangguangyxg (approved):

PR #68967, opened 2026-02-05
[BugFix] Fix struct field names not being escaped

kyle-goodale-klaviyo:
## Why I'm doing:
Struct field names are not escaped in some cases even if the user escapes them on the input. This leads to errors.

## What I'm doing:

Fixes #68965

I updated StructField to always backtick field names in the toSQL() function. 
Back ticks are stripped on parsing when we assign the `name` field here so we don't have to worry about double backticks, that same functionality is also the source of this bug. 

I added a test case that reproduced the issue, confirmed it failed, then implemented my fix and confirmed the test case now passes.

## What type of PR is this:

- [X] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/ParseUtil.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/ParseUtil.java
@@ -151,6 +151,6 @@ public class ParseUtil {
         if (StringUtils.isEmpty(str)) {
             return str;
         }
-        return "`" + str + "`";
+        return "`" + str.replace("`", "``") + "`";
     }
 }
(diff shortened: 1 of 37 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

stdpain (approved):

kevincai:
remove the backport to branch-4.0, a lot of code conflict.

PR #69027, opened 2026-02-08
[BugFix] Fix JSON path parsing for quoted field names with dots and brackets

### LINKED ISSUE #69021: [FlatJSON] JSON path not calculated correctly for quoted path terms, leading to performance regression
https://github.com/StarRocks/starrocks/blob/main/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/prunesubfield/SubfieldAccessPathNormalizer.java#L322

This code path incorrectly handles "simple" paths, leading to significant performance regressions for what should be a simple json path expression. The parsing does not retain if the value was quoted or not - which is used to indicate a string literal or a path expression

Example:
- foo."bar[0]".baz
- foo["bar"].baz

These fail the [array pattern check](https://github.com/StarRocks/starrocks/blob/40ed169a12d0197d1c387385f096412b296009a9/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/prunesubfield/SubfieldAccessPathNormalizer.java#L48), even though it is quoted/direct property indexed and should be treated as a literal

Another:
- foo."my.inner.term".baz

This correctly extracts `my.inner.term` - but then gets [incorrectly matched later](https://github.com/StarRocks/starrocks/blob/main/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/JsonPathRewriteRule.java#L512) as an "invalid json path" during query execution

creatstar:
Quoted JSON field names like "bar[0]" and "my.inner.term" were incorrectly parsed: brackets were treated as array indexing and dots as path separators, causing path truncation and query errors. Replace StrTokenizer with a custom tokenizer that respects double-quote quoting, and update BE's split_path to handle quoted segments so the flat path tree is built correctly.

Fixes #69021

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [X] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [X] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

- [ ] 4.1
- [ ] 4.0
- [ ] 3.5
- [ ] 3.4

### DIFF
--- a/be/src/util/json_flattener.cpp
+++ b/be/src/util/json_flattener.cpp
@@ -293,18 +293,36 @@ double estimate_filter_fpp(uint64_t element_nums) {
 }
 
 std::pair<std::string_view, std::string_view> JsonFlatPath::split_path(const std::string_view& path) {
-    size_t pos = 0;
-    pos = path.find('.', pos);
-    std::string_view key;
-    std::string_view next;
-    if (pos == std::string::npos) {
-        key = path;
-    } else {
-        key = path.substr(0, pos);
-        next = path.substr(pos + 1);
+    if (path.empty()) {
+        return {path, {}};
     }
 
-    return {key, next};
+    if (path[0] == '"') {
+        // Quoted key: find the matching closing quote.
+        // The entire quoted segment (including quotes) is the key.
+        // e.g. "my.inner.term".baz -> key="my.inner.term", next=baz
+        size_t end_quote = path.find('"', 1);
+        if (end_quote == std::string_view::npos) {
+            // Unmatched quote, treat the whole thing as the key
+            return {path, {}};
+        }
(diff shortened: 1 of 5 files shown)


eshishki:
@creatstar is this still alive? The bug is real — #69021 is still open and `SubfieldAccessPathNormalizer` on main still uses `StrTokenizer(path, '.', '"')` (lines 274, 351), so quoted paths like `foo."bar[0]".baz` remain mis-parsed. To move it forward it needs the DCO sign-off and a green FE UT. Are you still working on it?

PR #69095, opened 2026-02-10
[Enhancement] fix cloud scan dop and data skew compared with olap scan

before-Sunrise:
## Why I'm doing:
1. if don't use tablet internal parallel scan,cloud native table still adjust scan dop which can be less then pipeline_dop, which cause Not enough parallelism
2. if The number of morsels is small, for example:3,  currently only at most 3 drivers will get morsel, ohter drivers found morsel queue is empty and closed directly, however, these 3 morsels can be split into 30 morsels, so only three drivers have input, which cause data skew.

## What I'm doing:
1.  if don't use tablet internal parallel scan, we choose the number of tablets as the dop of scan node which is the same as olap scan
2. for morsels less then dop*io tasks per scan and trigger tablet internal parallel scan, use individual morsel queue, and append the splitted mosel into next morsel queue in round robin way. And scan operator's has_output can be false only when there won't be any new morsel from splitting

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/pipeline/scan/connector_scan_operator.cpp
+++ b/be/src/exec/pipeline/scan/connector_scan_operator.cpp
@@ -204,6 +204,10 @@ void ConnectorScanOperatorFactory::detach_shared_input(int32_t operator_seq, int
     }
 }
 
+void ConnectorScanOperatorFactory::mark_split_source_morsel_finished() {
+    morsel_queue_factory()->mark_split_source_morsel_finished();
+}
+
 // ===============================================================
 struct ConnectorScanOperatorAdaptiveProcessor {
     // ----------------------
@@ -580,8 +584,27 @@ Status ConnectorScanOperator::append_morsels(std::vector<MorselPtr>&& morsels) {
             }
         }
     }
+
+    auto* morsel_queue_factory = _source_factory()->morsel_queue_factory();
+    if (morsel_queue_factory != nullptr && morsel_queue_factory->size() > 1 &&
+        morsel_queue_factory->enable_random_append_split_morsel()) {
+        auto notify = defer_notify([&]() { return true; });
+        for (auto& morsel : morsels) {
+            Morsels one;
+            one.emplace_back(std::move(morsel));
+            ASSIGN_OR_RETURN(int driver_seq, morsel_queue_factory->next_driver_seq());
+            RETURN_IF_ERROR(morsel_queue_factory->append_morsels(driver_seq, std::move(one)));
+        }
+        return Status::OK();
+    }
+
     RETURN_IF_ERROR(_morsel_queue->append_morsels(std::move(morsels)));
     return Status::OK();
+}
+
+void ConnectorScanOperator::mark_split_source_morsel_finished() {
+    auto* factory = down_cast<ConnectorScanOperatorFactory*>(_factory);
+    factory->mark_split_source_morsel_finished();
 }
 
 int64_t ConnectorScanOperator::get_scan_table_id() const {
@@ -620,6 +643,11 @@ ConnectorChunkSource::ConnectorChunkSource(ScanOperator* op, RuntimeProfile* run
     auto* scan_morsel = (ScanMorsel*)_morsel.get();
     TScanRange* scan_range = scan_morsel->get_scan_range();
     ScanSplitContext* split_context = scan_morsel->get_split_context();
+    // A split source morsel means this morsel can potentially produce split tasks.
+    // `split_context == nullptr` identifies root morsels, and `has_more_from_split()`
+    // indicates split mode is enabled for this scan node.
+    _is_split_source_morsel =
+            (split_context == nullptr) && (op->morsel_queue() != nullptr) && op->morsel_queue()->has_more_from_split();
 
     _data_source = scan_node->data_source_provider()->create_data_source(*scan_range);
     _data_source->set_driver_sequence(op->get_driver_sequence());
@@ -653,6 +681,16 @@ const std::string ConnectorChunkSource::get_custom_coredump_msg() const {
 ConnectorScanOperatorIOTasksMemLimiter* ConnectorChunkSource::_get_io_tasks_mem_limiter() const {
     auto* f = down_cast<ConnectorScanOperatorFactory*>(_scan_op->get_factory());
     return f->_io_tasks_mem_limiter;
+}
+
+void ConnectorChunkSource::_report_split_source_morsel_finished_once() {
+    if (!_is_split_source_morsel || _split_source_morsel_reported) {
+        return;
+    }
+
+    auto* scan_op = down_cast<ConnectorScanOperator*>(_scan_op);
+    scan_op->mark_split_source_morsel_finished();
+    _split_source_morsel_reported = true;
 }
 
 void ConnectorChunkSource::close(RuntimeState* state) {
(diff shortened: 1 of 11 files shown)


satanson (dismissed):

dirtysalt (dismissed):

wyb (dismissed):

kevincai (dismissed):

be/src/exec/scan_node.cpp
```diff
@@ -132,7 +134,18 @@ static StatusOr<std::map<int, pipeline::MorselQueuePtr>> uniform_distribute_mors
     } else {
         for (auto& [operator_seq, morsels] : morsels_per_driver) {
             queue_per_driver.emplace(operator_seq, std::make_unique<pipeline::DynamicMorselQueue>(
-                                                           std::move(morsels), morsel_queue->has_more()));
+                                                           std::move(morsels), has_more_scan_ranges));
+        }
+
+        if (has_more_from_split) {
```
kevincai, line 142: add comments explain why do this in this `uniform_distribute_morsels()`?

be/src/exec/scan_node.cpp
```diff
@@ -132,7 +134,18 @@ static StatusOr<std::map<int, pipeline::MorselQueuePtr>> uniform_distribute_mors
...
+            for (; driver_seq < dop; driver_seq++) {
+                queue_per_driver.emplace(driver_seq, std::make_unique<pipeline::DynamicMorselQueue>(
```
kevincai, line 144: although the emplace() will deal with duplicate driver_seq if the key already exists in the map. However, this is ambiguous on why needs do this.
Maybe more explicit with the following implementation with both explicit intention and remove the side effect of `driver_seq` in case its value modified somewhere in between.

```
    for (driver_seq = queue_per_driver.size(); driver_seq < dop; driver_seq++) {
                queue_per_driver.emplace(driver_seq, std::make_unique<pipeline::DynamicMorselQueue>(Expand commentComment on line R142
                                                             pipeline::Morsels(), has_more_scan_ranges));
            }
```

be/src/exec/scan_node.cpp
```diff
@@ -157,12 +170,19 @@ StatusOr<pipeline::MorselQueueFactoryPtr> ScanNode::convert_scan_range_to_morsel
                                                  morsel_queue_type == pipeline::MorselQueue::Type::DYNAMIC);
 
         // If not so much morsels, try to assign morsel uniformly among operators to avoid data skew
-        if (!always_shared_scan() && scan_dop > 1 && is_fixed_or_dynamic_morsel_queue &&
+        // for DLA, always use SharedMorselQueueFactory
+        // for cloud, if enable_shared_scan, then always use SharedMorselQueueFactory
+        // else it will consider the morsel numbers and io_parallelism to choose individual or shared morsel queue factory
+        if (!always_shared_scan() && !enable_shared_scan && scan_dop > 1 && is_fixed_or_dynamic_morsel_queue &&
```
kevincai, line 178: so there are two settings controlling whether the shared_scan turned on at the runtime. better have a matrix to explain the combination.
shared_scan is turned on either by the data source settings or by the runtime settings, combining with the default to `false` for LakeDataSource() which means, before the change, LakeDataSource always enables shared_scan while now, it defaults to turn off and can be turned on by runtime `enable_shared_scan`, this is a behavior change that should be highlighted.

kevincai (commented):
TBD

be/src/exec/scan_node.cpp
```diff
@@ -157,12 +170,19 @@ StatusOr<pipeline::MorselQueueFactoryPtr> ScanNode::convert_scan_range_to_morsel
...
             morsel_queue->num_original_morsels() <= io_parallelism) {
+            bool enable_random_append_split_morsel = morsel_queue->has_more_from_split();
+            DCHECK(!enable_random_append_split_morsel || morsel_queue_type == pipeline::MorselQueue::Type::DYNAMIC);
             ASSIGN_OR_RETURN(auto morsel_queue_map, uniform_distribute_morsels(std::move(morsel_queue), scan_dop));
             return std::make_unique<pipeline::IndividualMorselQueueFactory>(std::move(morsel_queue_map),
-                                                                            /*could_local_shuffle*/ true);
+                                                                            /*could_local_shuffle*/ true,
+                                                                            enable_random_append_split_morsel);
         } else {
+            morsel_queue->set_has_more_from_split(false);
```
kevincai, line 187: only see explicit set to `false` here, but no explicit set to `true` elsewhere, and its default to `false` in the morsel_queue implementation, so what's the reason for this explicit false setting without explicit `true`?

be/src/exec/pipeline/scan/connector_scan_operator.cpp
```diff
@@ -580,10 +584,29 @@ Status ConnectorScanOperator::append_morsels(std::vector<MorselPtr>&& morsels) {
...
+
+    auto* morsel_queue_factory = _source_factory()->morsel_queue_factory();
+    if (morsel_queue_factory != nullptr && morsel_queue_factory->size() > 1 &&
+        morsel_queue_factory->enable_random_append_split_morsel()) {
+        auto notify = defer_notify([&]() { return true; });
+        for (auto& morsel : morsels) {
+            Morsels one;
+            one.emplace_back(std::move(morsel));
+            ASSIGN_OR_RETURN(int driver_seq, morsel_queue_factory->next_driver_seq());
+            RETURN_IF_ERROR(morsel_queue_factory->append_morsels(driver_seq, std::move(one)));
```
kevincai, line 596: if all the info are maintained by morselqueueFactory, why not let morselQueueFactory to determine which morsel queue should be appended?

And here from codebase, it will be branched either the MorselQueue to append the morsel or delegate to MorselQueueFactory to append the morsel, will this be unified from interface and inside the morsel queue, it will determine whether itself do the append or delegate to its Factory?

be/src/exec/scan_node.cpp
before-Sunrise: > only see explicit set to `false` here, but no explicit set to `true` elsewhere, and its default to `false` in the morsel_queue implementation, so what's the reason for this explicit false setting without explicit `true`?

convert_scan_range_to_morsel_queue-> DataSourceProvider::convert_scan_range_to_morsel_queue can   set it to true explicitly. But only one case need really use this feature, so set it to false in other cases(like shared scan/colocate_exec_group etc)

be/src/exec/scan_node.cpp
before-Sunrise: > so there are two settings controlling whether the shared_scan turned on at the runtime. better have a matrix to explain the combination. shared_scan is turned on either by the data source settings or by the runtime settings, combining with the default to `false` for LakeDataSource() which means, before the change, LakeDataSource always enables shared_scan while now, it defaults to turn off and can be turned on by runtime `enable_shared_scan`, this is a behavior change that should be highlighted.

nope, you totally misunderstand this. enable_shared_scan is from session variable, and if set it to ture, then it will turn on shared scan when  (scan_ranges_per_driver_seq.empty() && !in_colocate_exec_group).
what i changed is when enable_shared_scan is false, before this pr, lake datasource will always create shared morsel queue factory, right now it may create individual morsel queue

be/src/exec/scan_node.cpp
kevincai: So it means the morselqueue says "YES, I have more to split, and I will produce more split morsels", and here scan node overrule it by "No, you don't have". "whether to have more split morsels" is a factor that determined  by the morselqueue itself. "Whether to wait for more split morsels" can be controlled by scan node. Maybe this can make things clear?

be/src/exec/scan_node.cpp
kevincai: ok

be/src/exec/scan_node.cpp
before-Sunrise: > So it means the morselqueue says "YES, I have more to split, and I will produce more split morsels", and here scan node overrule it by "No, you don't have". "whether to have more split morsels" is a factor that determined by the morselqueue itself. "Whether to wait for more split morsels" can be controlled by scan node. Maybe this can make things clear?

yes, you totally get it!

PR #69101, opened 2026-02-10
[Enhancement] adaptive choose whether to batch chunks when have topn rf

before-Sunrise:
## Why I'm doing:
TopN runtime filters can prune data much earlier when the scan side updates them quickly, especially if the filter is applied to a table’s sort key. The current chunk batching behavior may delay those updates especially when predicates have good selectivity, which reduces the benefit of TopN runtime filtering. This PR makes scan batching adaptive so the engine can expose useful TopN filter information earlier and improve pruning efficiency.

## What I'm doing:
This PR adds an adaptive chunk-batching mechanism for scan operators when a TopN runtime filter is present on sort keys. It  introduces FE-to-BE plumbing to mark whether a TopN filter targets sort keys, adds a new session variable  topn_runtime_filter_update_mode with modes like off, auto, and force, and passes that setting through thrift. On the BE side, it  adds a shared update context that tracks filtering effectiveness and uses it to decide whether to bypass chunk accumulation or use  smaller output thresholds.

if mode is OFF: the logic is same as before. accumulate chunks
if mode is FORCE: don't  accumulate at all
if mode is AUTO: don't accumulate for the first ten chunks, so they can update topn rf, then depends on topn rf's selectivity, if topn rf works well then don't accumulate, otherwise accumulate.

This  update context is per scan operator.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/pipeline/scan/connector_scan_operator.cpp
+++ b/be/src/exec/pipeline/scan/connector_scan_operator.cpp
@@ -849,7 +849,8 @@ Status ConnectorChunkSource::_read_chunk(RuntimeState* state, ChunkPtr* chunk) {
     ConnectorScanOperatorAdaptiveProcessor& P = *(scan_op->adaptive_processor());
 
     DeferOp defer_op([&]() { P.last_chunk_souce_finish_timestamp = GetCurrentTimeMicros(); });
-    bool skip_chunk_accumulate = _should_skip_chunk_accumulate();
+    // Direct output is only safe when the accumulator is empty; otherwise we would use accumulator
+    bool skip_chunk_accumulate = _should_skip_chunk_accumulate() && _ck_acc.empty();
     auto is_terminal_status = [](const Status& st) {
         return st.is_end_of_file() || st.is_cancelled() || (!st.ok() && !st.is_time_out() && !st.is_eagain());
     };
@@ -858,6 +859,7 @@ Status ConnectorChunkSource::_read_chunk(RuntimeState* state, ChunkPtr* chunk) {
         int64_t total_time_ns = 0;
         int64_t delta_io_time_ns = 0;
         int64_t delta_scan_bytes = 0;
+        ChunkPtr direct_chunk;
         {
             SCOPED_RAW_TIMER(&total_time_ns);
             int64_t prev_io_time_ns = get_io_time_spent();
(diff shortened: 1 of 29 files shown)


CelerData-Reviewer:
@codex review

satanson (approved):

be/src/storage/rowset/segment_iterator.cpp
```diff
@@ -2324,7 +2332,8 @@ StatusOr<size_t> SegmentIterator::_predicate_evaluate_late_materialize(vector<ro
 
 StatusOr<size_t> SegmentIterator::_predicate_evaluate_without_late_materialize(vector<rowid_t>* rowid) {
     const uint32_t chunk_capacity = _reserve_chunk_size;
-    const uint32_t return_chunk_threshold = std::max<uint32_t>(chunk_capacity - chunk_capacity / 4, 1);
+    const uint32_t return_chunk_threshold =
+            _use_small_chunk_threshold() ? 1 : std::max<uint32_t>(chunk_capacity - chunk_capacity / 4, 1);
```
stdpain, line 2336: This may not be a good strategy. Reserves are necessary.

be/src/exec/pipeline/scan/connector_scan_operator.cpp
```diff
@@ -854,6 +886,10 @@ Status ConnectorChunkSource::_read_chunk(RuntimeState* state, ChunkPtr* chunk) {
                 _status = _data_source->get_next(state, &tmp);
                 if (_status.ok()) {
                     if (tmp->num_rows() == 0) continue;
+                    if (skip_chunk_accumulate) {
+                        direct_chunk = std::move(tmp);
+                        break;
+                    }
                     _ck_acc.push(tmp);
```
stdpain, line 893:
```suggestion
                    _ck_acc.push(tmp);
                    if (skip_chunk_accumulate) {
                        break;
                    }
```

before-Sunrise: > Suggested change
>                     if (skip_chunk_accumulate) {
>                         direct_chunk = std::move(tmp);
>                         break;
>                     }
>                     _ck_acc.push(tmp);
>                     _ck_acc.push(tmp);
>                     if (skip_chunk_accumulate) {
>                         break;
>                     }

my first version is like above, but we can't call _ck_acc's pull() when _ck_acc's has_output is false.

be/src/storage/rowset/segment_iterator.cpp
before-Sunrise: > This may not be a good strategy. Reserves are necessary.

actually it only output first ten chunks without accumulating, so i guess this is ok？
wihtout this pr, we will accumulate when predicates are not column-expr predicates, but if it's column-expr predicates, then we won't accumulate, this make no sense in logic

PR #69196, opened 2026-02-11
[Enhancement] support materialized view priority when schedule refresh

kaijianding:
## Why I'm doing:

Currently, the task priority of materialized view's async refresh task is `0`.
If all `task_runs_concurrency` are occupied, the new created materialized view's refresh task is always pending.
If all `task_runs_concurrency` are occupied, the new scheduled materialized view's refresh task(like new partition's refresh task, which usually requires faster scheduling) is always pending.

## What I'm doing:

Add `task_priority` property both in materialized views' properties and in `Task`'s properties.
Set default priority from 0 to 90 if mv is async, so it's behavior change.
User can alter materialized view's `task_priority` to higher priority so that the new refresh task can be scheduled faster.

We can reuse this ability to support priority in `Submit Task` later.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4


LiShuMing (dismissed):

LiShuMing (approved):

satanson (approved):

HangyuanLiu (changes requested):
Thank you for your contribution.This problem does exist, but the current implementation is too simplistic. We will address this issue in the future based on warehouse and resource group principles.

PR #69256, opened 2026-02-13
[Enhancement] Upgrade HikariCP and add connection lifecycle management for JDBC catalog

igor-raits:
## Why I'm doing:

JDBC catalog connections can become stale when external databases, load balancers, or firewalls silently close idle TCP connections. When StarRocks tries to reuse such a stale connection from the HikariCP pool, queries fail with "connection reset" or similar errors. Additionally, HikariCP 3.4.5 lacks the `keepaliveTime` feature (introduced in 4.x) that proactively validates idle connections before they go stale.

## What I'm doing:

  Upgrade HikariCP from 3.4.5 to 7.0.2 and expose three new connection pool lifecycle configuration parameters:

- `jdbc_connection_max_lifetime_ms` (default 300000): Maximum lifetime of a pooled connection. Connections are recycled before this timeout to prevent stale connections reaching queries. Should be shorter than the external database's connection timeout.
- `jdbc_connection_keepalive_time_ms` (default 30000): Idle connections are proactively tested at this interval to detect and evict stale connections before they are handed to a query.
- `jdbc_connection_leak_detection_threshold_ms` (default 0, disabled): Debugging aid that logs a warning if a connection is held longer than this threshold.

These configs are applied to both FE metadata connections and BE scanner connections (except leak detection, which is FE-only since BE connections are short-lived per-query).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

**Notes on my choices**:
- Parameter changes is checked because we're adding new config parameters that change pool behavior (connections now have a max lifetime and keepalive probing, whereas before they could live indefinitely).
- Miscellaneous is checked because upgrading HikariCP from 3.4.5 to 7.0.2 is a major version bump that could have upgrade/downgrade compatibility implications (e.g., different logging, different internal behavior).

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


be/src/exec/jdbc_scanner.cpp
```diff
@@ -144,9 +144,22 @@ Status JDBCScanner::_init_jdbc_scan_context(RuntimeState* state) {
...
+    // maximum lifetime of a connection in the pool - minimum 30 seconds
+    int64_t max_lifetime_ms = config::jdbc_connection_max_lifetime_ms;
+    if (UNLIKELY(max_lifetime_ms < 30000)) {
+        max_lifetime_ms = 300000; // default 5 minutes
+    }
+
+    // keepalive frequency for idle connections - must be less than maxLifetime
+    int64_t keepalive_time_ms = config::jdbc_connection_keepalive_time_ms;
+    if (UNLIKELY(keepalive_time_ms <= 0 || keepalive_time_ms >= max_lifetime_ms)) {
+        keepalive_time_ms = 30000; // default 30 seconds
```
GavinMar: This line prevents users from explicitly disabling the keepalive check (by setting keepalive_time_ms=0), is it expected?

igor-raits: I think it is good idea to have keepalive always enabled but I can imagine that users will potentially want to disable it. what is your preference / reccomendation?

be/src/exec/jdbc_scanner.cpp
```diff
@@ -144,9 +144,22 @@ Status JDBCScanner::_init_jdbc_scan_context(RuntimeState* state) {
     // some driver (like sqlserver) needs connection timeout less than 65536
     connection_timeout_ms = std::max(connection_timeout_ms, 30 * 1000);
     connection_timeout_ms = std::min(connection_timeout_ms, 65535 * 1000);
+
+    // maximum lifetime of a connection in the pool - minimum 30 seconds
+    int64_t max_lifetime_ms = config::jdbc_connection_max_lifetime_ms;
+    if (UNLIKELY(max_lifetime_ms < 30000)) {
```
Wenjun7J: Why `30000` in the judgement, and `300000` as default? I think it is a little werid. And using static variable here may be better. they look like too same.

igor-raits: I was thinkgin that it could be confusing / non-intuitive to users. default is 5m and if you specify something lower than 30s then hikaricp will simply use its own default (30s) and not tell anything. so  Iwas thinking if user specifies invalid value we could default to SR defaults. happy to change it though

DanRoscigno:
@igor-raits I reverted the large changes caused by the doc translation and added just the minimal change.

be/src/exec/jdbc_scanner.cpp
GavinMar: > I think it is good idea to have keepalive always enabled but I can imagine that users will potentially want to disable it. what is your preference / reccomendation?

OK, I prefer to align with HikariCP parameters and allow users to explicitly disable keep-alive, thus avoiding some confusion.

fe/fe-core/src/main/java/com/starrocks/connector/jdbc/JDBCMetadata.java
```diff
@@ -195,9 +195,34 @@ private HikariDataSource createHikariDataSource() {
...
+        if (maxLifetime < 30000) {
+            maxLifetime = 300000;
+        }
+        config.setMaxLifetime(maxLifetime);
+
+        // keepaliveTime: must be > 0 and < maxLifetime, default to 30000ms
+        long keepaliveTime = Config.jdbc_connection_keepalive_time_ms;
+        if (keepaliveTime <= 0 || keepaliveTime >= maxLifetime) {
+            keepaliveTime = 30000;
+        }
```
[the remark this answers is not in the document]
igor-raits: Fixed in c32ca7ef25. The fallback now goes to `0` (disabled) instead of `30000`, which is always safe regardless of `maxLifetime`. Both FE and BE now use named constants (`MINIMUM_MAX_LIFETIME_MS`, `KEEPALIVE_DISABLED`, etc.) and log a `WARN` when overriding user values.

When `maxLifetime=30000`, there is no valid enabled keepalive value (HikariCP requires `keepaliveTime >= 30000 && < maxLifetime`), so the only correct fallback is `0` (disabled).

be/src/exec/jdbc_scanner.cpp
```diff
@@ -144,9 +144,22 @@ Status JDBCScanner::_init_jdbc_scan_context(RuntimeState* state) {
...
+        max_lifetime_ms = 300000; // default 5 minutes
+    }
+
+    // keepalive frequency for idle connections - must be less than maxLifetime
+    int64_t keepalive_time_ms = config::jdbc_connection_keepalive_time_ms;
+    if (UNLIKELY(keepalive_time_ms <= 0 || keepalive_time_ms >= max_lifetime_ms)) {
+        keepalive_time_ms = 30000; // default 30 seconds
+    }
```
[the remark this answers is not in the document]
igor-raits: Fixed in c32ca7ef25 — same approach as FE. The BE validation now uses named constants, falls back to `0` (disabled) for invalid values, and logs `LOG(WARNING)` when overriding. Also removed the `UNLIKELY` macros since this code runs once per scan, not in a hot path.

be/src/exec/jdbc_scanner.cpp
igor-raits: Fixed in c32ca7ef25 — `keepalive_time_ms=0` is now respected as "disabled" (aligned with HikariCP semantics per your recommendation). The validation logic is now:

- `0` → pass through (disabled, no keepalive probing)
- `> 0 but < 30000` → invalid, reset to `0` with `LOG(WARNING)`
- `>= maxLifetime` → invalid, reset to `0` with `LOG(WARNING)`
- otherwise → valid, pass through

Every override is logged so misconfiguration is visible in production logs.

be/src/exec/jdbc_scanner.cpp
igor-raits: Fixed in c32ca7ef25. Introduced named constants to eliminate the confusing `30000` vs `300000` magic numbers:

```cpp
static constexpr int64_t MINIMUM_MAX_LIFETIME_MS = 30000;
static constexpr int64_t DEFAULT_MAX_LIFETIME_MS = 300000;
static constexpr int64_t MINIMUM_KEEPALIVE_TIME_MS = 30000;
static constexpr int64_t KEEPALIVE_DISABLED = 0;
```

Same constants defined in both BE (`jdbc_scanner.h`) and FE (`JDBCMetadata.java`). The `30000` is HikariCP's enforced minimum for both maxLifetime and enabled keepaliveTime — they're the same number because HikariCP defines it that way, but now they're referenced by distinct named constants so the intent is clear.

GavinMar (approved):

Wenjun7J (approved):

kevincai (dismissed):

kevincai (approved):

trueeyu (approved):

dirtysalt (approved):

PR #69271, opened 2026-02-13
[Enhancement] Support reading Iceberg Parquet shredded VARIANT with typed-path reconstruction

dirtysalt:
## Why I'm doing:
  Iceberg format v3 introduces VARIANT shredding (metadata/value/typed_value).
  Current Parquet VARIANT reading path mainly focuses on legacy non-shredded layout, which can cause:
  - missing typed-path materialization opportunities,
  - inconsistent behavior between shredded and non-shredded files (especially nested/array paths),
  - weak diagnostics when type conversion/building typed readers fails.

  This PR adds full shredded VARIANT read support in BE and verifies result consistency with non-shredded data.

  ## What I'm doing:
  ### Functional summary
  - Extend Parquet VARIANT reader to support shredded layout with mixed sources:
    - typed_value columns for typed fast path,
    - value columns as binary fallback,
    - row-level overlay reconstruction for complete VARIANT semantics.
  - Add path-level type hints (`preferred_type`) and optional strict compatibility check when building typed readers.
  - Keep output shape stable for requested paths even when a path is not shredded in a specific file/row-group.
  - Improve unsupported-conversion error message with field debug info.
  - Add unit/integration tests to validate shredded vs non-shredded consistency (including nested fields and arrays).

  ### Class/struct responsibilities
  - `VariantPathTypeHint`:
    - Declares one shredded path and its optional preferred read type.
  - `VariantShreddedReadHints`:
    - Aggregates path hints and `strict_preferred_type` policy.
  - `ColumnReaderFactory`:
    - Parses/normalizes shredded read hints.
    - Infers typed descriptors from Parquet schema/logical annotations.
    - Builds VARIANT readers for metadata/value/typed_value and recursive shredded field tree.
  - `ShreddedFieldNode`:
    - Represents one node in shredded typed_value tree.
    - Holds fallback reader, typed reader, typed read type, child nodes, and batch-local columns.
  - `VariantColumnReader`:
    - Prepares/reads metadata + fallback + typed shredded columns.
    - Rebuilds overlays (including array element overlays) and materializes final VARIANT output.
    - Produces top-level typed bindings (SCALAR or VARIANT) with fallback-safe behavior.
  - `ParquetUtils` (new helper methods):
    - `get_non_null_data_column_and_row`: unwrap const/nullable columns safely.
    - `has_non_null_value` / `has_non_null_binary_value`: fast checks for binding materialization decisions.
  - Internal helper `_TopBinding`:
    - Maps requested/auto-discovered path to materialization kind and target node.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


be/src/column/variant_merger.cpp
```diff
@@ -0,0 +1,534 @@
...
+#include <unordered_map>
+
+#include "column/column_builder.h"
+#include "column/column_helper.h"
+#include "column/variant_column.h"
+#include "gutil/casts.h"
+#include "gutil/strings/substitute.h"
+#include "types/datum.h"
+#include "types/logical_type.h"
+#include "util/variant_encoder.h"
```
alvin-phoenix-ai: please refer to the deps requirement, ColumnCore is not allowed to depends on util module now

be/src/column/variant_merger.h
```diff
@@ -0,0 +1,65 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "column/column.h"
+#include "common/status.h"
+#include "common/statusor.h"
```
alvin-phoenix-ai: better to use base/status, common is only a wrapper

be/src/column/variant_encoder.cpp
```diff
@@ -32,136 +32,62 @@
...
-#include "column/map_column.h"
 #include "column/nullable_column.h"
-#include "column/struct_column.h"
 #include "common/status.h"
 #include "storage/types.h"
 #include "types/time_types.h"
 #include "types/type_descriptor.h"
-#include "types/variant.h"
+#include "types/variant_base.h"
 #include "velocypack/vpack.h"
```
alvin-phoenix-ai: We don't allow circular dependency anymore, as this file depends on column and column depends on util.

be/test/CMakeLists.txt
```diff
@@ -323,6 +324,7 @@ set(EXEC_FILES
         ./serde/protobuf_serde_test.cpp
         ./types/type_checker_test.cpp
         ./types/type_checker_xml_loader_test.cpp 
+        ./types/variant_builder_test.cpp
```
alvin-phoenix-ai: make it in types_test and compile types_test will be fast
use the following to run UT.
STARROCKS_LINKER=lld STARROCKS_THIRDPARTY=/var/local/thirdparty ./run-be-ut.sh --clean --with-dynamic --build-target types_test --without-java-ext

be/test/CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ set(EXEC_FILES
         ./column/decimalv3_column_test.cpp
         ./column/fixed_length_column_extended_test.cpp
         ./column/json_column_test.cpp
-        ./column/variant_column_extended_test.cpp
+        ./column/variant_column_test.cpp
+        ./column/variant_merger_test.cpp
```
alvin-phoenix-ai: move them to column_test

alvin-phoenix-ai (changes requested):

be/src/exprs/variant_path_reader.cpp
```diff
@@ -0,0 +1,295 @@
...
+        if (!seg.is_object()) break;
+        object_keys.emplace_back(seg.get_key());
+    }
+    if (object_keys.empty()) return;
+
+    // Build the full dotted prefix once — reused for the _has_typed_child check below.
+    std::string full_prefix;
+    for (size_t i = 0; i < object_keys.size(); ++i) {
+        if (i > 0) full_prefix.push_back('.');
+        full_prefix.append(object_keys[i]);
```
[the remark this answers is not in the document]
dirtysalt: Fixed. `VariantPathReader::_try_match_typed` now matches typed paths via canonical shredded-path encoding (`VariantPath::to_shredded_path`) instead of dotted string concatenation.

  I also updated the typed-child detection path to use parsed shredded paths (segment-wise parent check), so quoted keys like `$['a.b']` are handled correctly and we no longer rely on fragile string-prefix matching.

be/src/column/variant_column.cpp
```diff
@@ -33,28 +261,67 @@ void VariantColumn::serialize_batch(uint8_t* dst, Buffer<uint32_t>& slice_sizes,
...
+    }
+    // Rebuild parsed-path cache in sync.
+    _parsed_shredded_paths.clear();
+    _parsed_shredded_paths.reserve(_shredded_paths.size());
+    for (const auto& path : _shredded_paths) {
+        auto parsed = VariantPathParser::parse_shredded_path(std::string_view(path));
+        // Invalid shredded paths are stored as empty VariantPath so the cache
+        // is always parallel with _shredded_paths. collect_typed_overlays will
+        // detect the mismatch when it tries to use the cached entry.
+        _parsed_shredded_paths.emplace_back(parsed.ok() ? std::move(parsed).value() : VariantPath{});
```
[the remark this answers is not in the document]
dirtysalt: Fixed. We now reject malformed shredded paths during schema validation (`VariantColumn::validate_shredded_schema`) instead of caching an empty `VariantPath`.

  `_rebuild_path_index` no longer inserts empty paths on parse failure. This prevents invalid metadata from being interpreted as an empty overlay path (root replacement semantics).

  Also added a regression test:
  `VariantColumnTest.test_shredded_schema_reject_invalid_paths`.

be/src/exprs/array_functions.cpp
```diff
@@ -60,7 +60,7 @@ StatusOr<ColumnPtr> ArrayFunctions::array_length([[maybe_unused]] FunctionContex
         if (arg0->has_null()) {
             // Copy null flags.
             return NullableColumn::create(std::move(col_result),
-                                          down_cast<const NullableColumn*>(arg0)->null_column()->clone());
+                                          std::move(*(down_cast<const NullableColumn*>(arg0)->null_column())).mutate());
```
dirtysalt, line 63: fix back by https://github.com/StarRocks/starrocks/pull/69811

PR #69272, opened 2026-02-13
[Feature] Introduce virtual statistics and implement `unnest` stats

### LINKED ISSUE #69318: Calculate unnest statistics and propagate them
## Feature request

When unnesting an array column, we do not have sophisticated statistics on the array elements.
For example, if you unnest an array column containing arrays which have a lot of NULL elements (i.e. the resulting unnest is NULL-skewed), then the optimizer can not apply any optimization (such as a join skew optimization) since we do not have statistics which refer to the elements in the arrays.
In general, this particular issue is transferable to a lot more complex types, e.g. think of any operators executed on top of arrays or structs.

An elegant solution would be to introduce virtual column statistics, which are not calculated directly on physical columns but expressions.

Alternatively, one could add a new statistic type (duplicating all of `nullsFraction`, `histogram`, etc.). This is not scalable since the issue is not only with `unnest` but also other operators operating on non-primitive data.

**Additional context**

OliLay:
## Why I'm doing:

When `unnest`ing an array column, we do not have sophisticated statistics on the array elements. 
For example, if you `unnest` an array column containing arrays which have a lot of NULL elements (i.e. the resulting `unnest` is NULL-skewed), then the optimizer can not apply any optimization (such as a join skew optimization) since we do not have statistics which refer to the elements in the arrays. 
In general, this particular issue is transferable to a lot more complex types, e.g. think of any operators executed on top of arrays or structs.

## What I'm doing:
Instead of adding a new statistics type for every complex data structure, I took the approach of introducing _virtual statistics_. Virtual statistics are statistics that are not directly computed over a physical column, but over an expression.
An example for this are array columns and `unnest`. In this case, the virtual unnest statistic computes the column statistics over `unnest(column)` instead of `column`. 

### Generation of virtual statistics
I implemented a generic framework for virtual statistics and one concrete implementation for the aforementioned array columns + `unnest` use-case.

- Virtual statistics are persisted just as normal column statistics are, just with the exception that their column name is encoded as `VIRTUAL_STATISTICS_{column_name}_{concrete_virtual_statistic_name}. 
    - For example, if virtual `unnest`statistics are computed for a column called `col1`, the corresponding entry in the column statistics table would be `VIRTUAL_STATISTICS_col1_UNNEST`.
- Virtual statistics creation is controlled over ANALYZE properties. It is disabled per default.
    - For example, the unnest stats creation can be triggered by using the following properties:
    ```sql
     ANALYZE FULL TABLE `TABLE1` PROPERTIES("unnest_virtual_statistics" = "true");
    ```
- Virtual statistics are implemented for legacy `FullStatisticsCollectJob` and `SampleStatisticsCollectJob` as well as the more recent `HyperStatisticsCollectJob` (both SAMPLE and FULL).
- The framework allows to easily add more virtual statistics by subclassing the `VirtualStatistic` class. `UnnestStatistic` does exactly this.

### Usage of virtual statistics
For every concrete virtual statistic being implemented, the optimizer has to be adjusted to take the virtual statistic into account. For the concrete `unnest` statistic, the optimizer was adapted so that when encountering an unnest expression, virtual statistics are looked up. 

- The unnest statistics usage can be controlled over a new session variable `enable_unnest_virtual_statistics`. It is disabled per default.
- If enabled, the optimizer will search for the virtual statistic, and if existent, will use the stats of the virtual statistic to propagate it beyond the unnest.

### Example
For example, assume we have an array column which contains arrays that have a NULL skew, e.g.
```sql
select ARRAY_TEST from TABLE1 limit 5;
+----------------------------+
| ARRAY_TEST                 |
+----------------------------+
| [null,null,null,null,null] |
| [null,null,"HBTAmhMXhJR9g5Gwpq",null,null] |
| [null,null,null,null,null] |
| [null,null,null,null,null] |
| [null,null,null,null,null] |
+----------------------------+
```

The current statistics do not capture the array element skew, hence when we unnest the column the optimizer knows nothing about the statistics of the elements:
```sql
analyze full table `TABLE1`;

explain costs
select
    unnest
from
    TABLE1 t1,
    unnest(ARRAY_TEST) AS t(unnest);

+---------------------------------------------------------------------------------+
| Explain String                                                                  |
+---------------------------------------------------------------------------------+
| PLAN COST                                                                       |
|   CPU: 8.0E7                                                                    |
|   Memory: 0.0                                                                   |
|                                                                                 |
| PLAN FRAGMENT 0(F01)                                                            |
|   Output Exprs:9: unnest                                                        |
|   Input Partition: UNPARTITIONED                                                |
|   RESULT SINK                                                                   |
|                                                                                 |
|   2:EXCHANGE                                                                    |
|      cardinality: 5000000                                                       |
|                                                                                 |
| PLAN FRAGMENT 1(F00)                                                            |
|                                                                                 |
|   Input Partition: RANDOM                                                       |
|   OutPut Partition: UNPARTITIONED                                               |
|   OutPut Exchange Id: 02                                                        |
|                                                                                 |
|   1:TableValueFunction                                                          |
|   |  tableFunctionName: unnest                                                  |
|   |  columns: [unnest]                                                          |
|   |  returnTypes: [VARCHAR(65000)]                                              |
|   |  cardinality: 5000000                                                       |
|   |  column statistics:                                                         |
|   |  * unnest-->[-Infinity, Infinity, 0.0, 1.0, 1.0] UNKNOWN                    |
|   |                                                                             |
|   0:OlapScanNode                                                                |
|      table: TABLE1, rollup: TABLE1                                              |
|      preAggregation: on                                                         |
|      partitionsRatio=1/1, tabletsRatio=20/20                                    |
|      tabletList=10176,10178,10180,10182,10184,10186,10188,10190,10192,10194 ... |
|      actualRows=5000000, avgRowSize=16.0                                        |
|      Pruned type: 8 <-> [ARRAY<VARCHAR(65000)>]                                 |
|      cardinality: 5000000                                                       |
|      column statistics:                                                         |
|      * ARRAY_TEST-->[-Infinity, Infinity, 0.0, 16.0, 1.0] ESTIMATE              |
+---------------------------------------------------------------------------------+
```

When computing the virtual unnest stats and enabling the session variable, the optimizer knows the stats after unnest and correctly identifies that the unnest has >0.97 nulls fraction. 
```sql
analyze full table `TABLE1` PROPERTIES("unnest_virtual_statistics" = "true");

set
    enable_unnest_virtual_statistics = true;

explain costs
    -> select
    ->     unnest
    -> from
    ->     TABLE1 t1,
    ->     unnest(ARRAY_TEST) AS t(unnest);
+---------------------------------------------------------------------------------+
| Explain String                                                                  |
+---------------------------------------------------------------------------------+
| PLAN COST                                                                       |
|   CPU: 8.0E7                                                                    |
|   Memory: 0.0                                                                   |
|                                                                                 |
| PLAN FRAGMENT 0(F01)                                                            |
|   Output Exprs:9: unnest                                                        |
|   Input Partition: UNPARTITIONED                                                |
|   RESULT SINK                                                                   |
|                                                                                 |
|   2:EXCHANGE                                                                    |
|      cardinality: 5000000                                                       |
|                                                                                 |
| PLAN FRAGMENT 1(F00)                                                            |
|                                                                                 |
|   Input Partition: RANDOM                                                       |
|   OutPut Partition: UNPARTITIONED                                               |
|   OutPut Exchange Id: 02                                                        |
|                                                                                 |
|   1:TableValueFunction                                                          |
|   |  tableFunctionName: unnest                                                  |
|   |  columns: [unnest]                                                          |
|   |  returnTypes: [VARCHAR(65000)]                                              |
|   |  cardinality: 5000000                                                       |
|   |  column statistics:                                                         |
|   |  * unnest-->[-Infinity, Infinity, 0.97995, 14.950885, 4089.0] ESTIMATE      |
|   |                                                                             |
|   0:OlapScanNode                                                                |
|      table: TABLE1, rollup: TABLE1                                              |
|      preAggregation: on                                                         |
|      partitionsRatio=1/1, tabletsRatio=20/20                                    |
|      tabletList=10176,10178,10180,10182,10184,10186,10188,10190,10192,10194 ... |
|      actualRows=5000000, avgRowSize=16.0                                        |
|      Pruned type: 8 <-> [ARRAY<VARCHAR(65000)>]                                 |
|      cardinality: 5000000                                                       |
|      column statistics:                                                         |
|      * ARRAY_TEST-->[-Infinity, Infinity, 0.0, 16.0, 1.0] ESTIMATE              |
+---------------------------------------------------------------------------------+
```

Based on this, the optimizer can utilize the stats to rewrite the plan after unnest, e.g. if it is used in a join.

Fixes #69318

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
```diff
@@ -1640,17 +1649,92 @@ private Void computeRepeatNode(ExpressionContext context, List<ColumnRefOperator
...
+            table = ((PhysicalScanOperator) scanOperator).getTable();
+        }
+
+        if (table == null) {
+            LOG.warn("Unnest statistics: Failed to load unnest statistics as no scan operator could be found.");
+            return Optional.empty();
+        }
+
+        try {
+            String virtualColumnName = VirtualStatistic.UNNEST.getVirtualColumnName(columnRef.getName());
```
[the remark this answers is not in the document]
OliLay: This actually works, example:
```sql
explain costs
select
    t.unnest
from
    (select ARRAY_TEST as foo from TABLE1) t1,
    unnest(t1.foo) AS t(unnest);
```
propagates unnest stats correctly:
```
+----------------------------------------------------------------------------------+
| Explain String                                                                   |
+----------------------------------------------------------------------------------+
| PLAN COST                                                                        |
|   CPU: 4.0E8                                                                     |
|   Memory: 0.0                                                                    |
|                                                                                  |
| PLAN FRAGMENT 0(F01)                                                             |
|   Output Exprs:9: unnest                                                         |
|   Input Partition: UNPARTITIONED                                                 |
|   RESULT SINK                                                                    |
|                                                                                  |
|   2:EXCHANGE                                                                     |
|      cardinality: 25000000                                                       |
|                                                                                  |
| PLAN FRAGMENT 1(F00)                                                             |
|                                                                                  |
|   Input Partition: RANDOM                                                        |
|   OutPut Partition: UNPARTITIONED                                                |
|   OutPut Exchange Id: 02                                                         |
|                                                                                  |
|   1:TableValueFunction                                                           |
|   |  tableFunctionName: unnest                                                   |
|   |  columns: [unnest]                                                           |
|   |  returnTypes: [VARCHAR(65000)]                                               |
|   |  cardinality: 25000000                                                       |
|   |  column statistics:                                                          |
|   |  * unnest-->[-Infinity, Infinity, 0.98001832, 0.29975068, 502373.0] ESTIMATE |
|   |                                                                              |
|   0:OlapScanNode                                                                 |
|      table: TABLE1, rollup: TABLE1                                               |
|      preAggregation: on                                                          |
|      partitionsRatio=1/1, tabletsRatio=20/20                                     |
|      tabletList=10176,10178,10180,10182,10184,10186,10188,10190,10192,10194 ...  |
|      actualRows=5000000, avgRowSize=80.0                                         |
|      Pruned type: 8 <-> [ARRAY<VARCHAR(65000)>]                                  |
|      cardinality: 5000000                                                        |
|      column statistics:                                                          |
|      * ARRAY_TEST-->[-Infinity, Infinity, 0.0, 80.0, 1.0] COS: 5.0 ESTIMATE      |
+----------------------------------------------------------------------------------+
```
Note in this case the optimizer pruned the project but we also propagate through project operators to find the column so this is fine.

fe/fe-core/src/main/java/com/starrocks/statistic/sample/SampleInfo.java
```diff
@@ -169,69 +184,113 @@ public String generatePrimitiveTypeColumnTask(long tableId, long dbId, String ta
...
+
     private String generateQueryDataSql(String tableName, String dbName,
-                                       List<ColumnStats> primitiveTypeStats,
-                                       TabletSampleManager manager) {
+                                        List<ColumnStats> primitiveTypeStats,
+                                        TabletSampleManager manager) {
         StringBuilder sql = new StringBuilder();
         String fullQualifiedName = "`" + dbName + "`.`" + tableName + "`";
         StringJoiner joiner = new StringJoiner(", ");
+        String laterals = buildLaterals(primitiveTypeStats);
```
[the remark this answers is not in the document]
OliLay: [3e225d7](https://github.com/StarRocks/starrocks/pull/69272/commits/3e225d7109df4bacb227f366ca08ec5af21a20e9)

fe/fe-core/src/main/java/com/starrocks/statistic/sample/SampleInfo.java
```diff
@@ -153,85 +168,110 @@ public String generatePrimitiveTypeColumnTask(long tableId, long dbId, String ta
...
-        builder.append(queryDataSql).append(") ");
-
-        int idx = 0;
-        int size = primitiveTypeStats.size();
-        for (ColumnStats columnStats : primitiveTypeStats) {
-            idx++;
-            builder.append(generateQueryColumnSql(tableId, dbId, tableName, dbName, columnStats, "col_" + idx));
-            if (idx != size) {
+
+        // CTEs
```
kangkaisen, line 181: did you consider the regular columns?

The new per-column CTE generation scans the table once per column even when no virtual/lateral stats are involved. This is a significant performance regression for sampled stats on wide tables (sampling work now scales with column count instead of being amortized in one scan).

OliLay: That is a good finding Kaisen! 
I fixed it now by only using separate CTEs for stats that require a lateral join. Regular stats are sampled in the same CTE, so they should always only require one scan now.
[cc449e1](https://github.com/StarRocks/starrocks/pull/69272/commits/cc449e1cb88f4aa42f95c201e5658f6166566d1e)

kangkaisen (dismissed):

fe/fe-core/src/main/java/com/starrocks/statistic/sample/SampleInfo.java
```diff
@@ -153,85 +168,155 @@ public String generatePrimitiveTypeColumnTask(long tableId, long dbId, String ta
...
+        builder.append("WITH ");
+        // Create a single shared CTE for all regular stats (if any)
+        if (!regularStats.isEmpty()) {
+            builder.append("cte_shared as (");
+            String queryDataSql = generateQueryDataSqlForMultipleColumns(tableName, dbName, regularStats, manager);
+            builder.append(queryDataSql).append(")");
+        }
+
+        // Create separate CTEs for each stat with lateral join (if any)
+        for (int i = 0; i < lateralJoinStats.size(); i++) {
```
LiShuMing, line 191: may we reserve the original sample query without CTE if there is no `lateralJoinStats` just like before.

docs/zh/using_starrocks/Cost_based_optimizer.md
```diff
@@ -449,68 +473,47 @@ CREATE ANALYZE FULL ALL db_name PROPERTIES (
...
-
-| **PROPERTIES**                        | **类型** | **默认值** | **说明**                                                                                                                           |
-|-------------------------------------------|--------|---------|---------------------------------------------------------------------------------------------------------------------------- |
-| enable_statistic_collect_on_first_load    | BOOLEAN  | TRUE     | 执行 INSERT INTO/OVERWRITE 后是否触发统计信息采集任务。该属性既是 FE 全局配置项，也可在表级别通过 CREATE TABLE 或 ALTER TABLE 设置。只有当全局配置和表级配置均为 `TRUE` 时，才会触发统计信息采集。                                                                       |
-| semi_sync_collect_statistic_await_seconds | LONG    | 30     | 返回结果前等待统计信息采集的最长时间。                                                                  |
-| statistic_sample_collect_ratio_threshold_of_first_load | DOUBLE    | 0.1  | OVERWRITE 不触发统计信息搜集任务的数据变动比例。                                                               |
-| statistic_sample_collect_rows             | LONG | 200000    | DML 导入总数据量超过该值时，使用抽样采集统计信息。                                                                               |
-
```
LiShuMing, line 474: Can you split the PR into `Feature` and `Doc` two parts? and this pr seems to have changed some unrelated docs, which needs to be confirmed.

LiShuMing (commented):
https://www.postgresql.org/docs/current/sql-createstatistics.html

Postgres supports `CREATE STATISTICS` with expressions. Can we extend this feature to support common expressions later?

fe/fe-core/src/main/java/com/starrocks/statistic/sample/SampleInfo.java
OliLay: Did I understand you correctly that if there are no lateral joins you don't want to have `cte_shared` but no CTE at all (everything inline)?
I think that would be possible but it makes the code more complex, and I am not sure it brings any benefits.

docs/zh/using_starrocks/Cost_based_optimizer.md
OliLay: Ok, I'll remove all docs in this PR and open a separate one for the docs.

OliLay: [5074c08](https://github.com/StarRocks/starrocks/pull/69272/commits/5074c080fe85ea2fd8f00716f6a7e79951e01d6d)

stephen-shelby:
Thank you for your contribution. I understand that this is a targeted enhancement for table function support. As for expression statistics, the approach is not generic enough and is not easy to extend to other expressions. We haven’t defined yet how to expose expression statistics  interface to users. You can think about a more general solution. Thanks.

stephen-shelby:
Thanks for working on this. I think supporting statistics for `unnest(array_col)` is useful, but I’m a bit concerned about the current abstraction.

The current `VirtualStatistic` design seems quite tied to the `unnest` use case. It models the result as a virtual column, with storage names like `VIRTUAL_STATISTIC_<col>_UNNEST`. This works for this specific case, but it may be hard to extend to more general expression statistics, for example:

- `upper(col)`
- `lower(col)`
- `date_trunc('day', ts)`
- `cast(col AS varchar)`
- other scalar or table-function expressions

Conceptually, `unnest(array_col)` is not just a virtual column. It is a row-expanding table function: one input row can produce multiple output rows. This is different from scalar expressions such as `upper(col)`, which preserve the input row count. I think it would be cleaner to model this as expression statistics rather than virtual column statistics.

A possible direction is:

1. Introduce an expression statistics abstraction, for example `ExpressionStatistic`, which describes:
   - the normalized expression key
   - applicable input types
   - output type
   - collection SQL
   - whether the expression is scalar or row-expanding
   - how optimizer expressions match this statistic

2. Keep expression statistics metadata separate from normal column statistics metadata. For example, an `ExpressionStatsMeta` could record:
   - normalized expression key, such as `SCALAR(upper,COLUMN_ID(12))` or `TABLE_FUNCTION(unnest,COLUMN_ID(15))`
   - expression kind: scalar vs table function
   - base column ids
   - storage key
   - update time and analyze type

3. Store the actual statistic values in the existing statistics table initially, using an internal storage key, but avoid encoding expression semantics directly in the column name. The storage key can be an implementation detail, while the expression semantics live in metadata.

4. In the optimizer, use a centralized lookup path for expression statistics:
   - scalar expressions such as `upper(col)` can be handled from `ExpressionStatisticCalculator`
   - table functions such as `unnest(array_col)` can be handled from table-function statistics derivation
   - both should resolve through the same expression-statistics metadata and storage lookup

This would make `unnest` one specific expression-statistics target instead of making `unnest` drive the shape of the whole framework. It should also leave room for supporting scalar expression statistics in the future without adding more function-specific logic throughout the statistics collection and optimizer code.

OliLay:
Stephen, thanks for taking another look, but I don't really understand your concerns. Have you looked at the design doc https://gist.github.com/OliLay/f0fac48a18ff85b1a38f088ecbf54052?

> The current VirtualStatistic design seems quite tied to the unnest use case. It models the result as a virtual column, with storage names like VIRTUAL_STATISTIC_<col>_UNNEST. This works for this specific case, but it may be hard to extend to more general expression statistics, for example:

I currently only implemented it for `unnest`. But it is super easy to extend it to any other expression statistics. As a PoC, I quickly built expression/virtual stats for `UPPER` based on the framework in this PR: https://github.com/OliLay/starrocks/pull/1/changes. 

**Note that based on the linked PR implementing it for e.g. `LOWER` or any other scalar function is just instantiating `DynamicExpressionStatistic` with other parameters.** This could well be done by some other component in FE (e.g. some component managing virtual stats which receives SQL DDL and creates/deletes `DynamicExpressionStatistic` based on this). As you see the PR has just a little bit over 100 LoC, and it implements virtual stats for potentially any scalar function without any changes to the framework in this PR. 

> Introduce an expression statistics abstraction, for example ExpressionStatistic, which describes:

This is exactly what the `VirtualStatistic` class is though to be. Have a look at the PoC https://github.com/OliLay/starrocks/pull/1/changes where there is a concrete implementation allowing for dynamic expressions to be registered.

> Keep expression statistics metadata separate from normal column statistics metadata. For example, an ExpressionStatsMeta could record:

Sure, we can do it. But I don't think we need it in an initial version. This is something that can be added when we write the layer to register/delete expressions using SQL DDL in my opinion.

> In the optimizer, use a centralized lookup path for expression statistics:

Yes, we need this for generic (scalar) expression. This is what I have done in the PoC: https://github.com/OliLay/starrocks/pull/1/changes. Note that this PR is not supposed to be a full-fledged implementation, this is the starting point for virtual stats (or expression stats, if you want to call them like that). As shown in the PoC, adding support for scalars is easy based on this framework. It does not have to be changed.

> This would make unnest one specific expression-statistics target instead of making unnest drive the shape of the whole framework. It should also leave room for supporting scalar expression statistics in the future without adding more function-specific logic throughout the statistics collection and optimizer code.

It may be driven by unnest now, but as shown in the PoC you don't need any framework changes to support scalar functions. The stats code was not really touched since `VirtualStatistic` already abstracts all necessary logic. In the optimizer, only the path for the scalars had to be added (but this was a add-only change).

PR #69427, opened 2026-02-24
[Enhancement] Use tryLock with timeout in canTxnFinished to reduce lock contention

kevincai:
## Why I'm doing:

The `canTxnFinished()` method previously acquired the intensive DB read lock unconditionally (blocking), which could be highly contended under heavy publish workloads and stall the publish daemon.

## What I'm doing:

This change mirrors the pattern already applied to `finishTransaction()` (ref: 450477ac7b):

- Add a `lockTimeoutMs` parameter to `DatabaseTransactionMgr.canTxnFinished()` and replace the blocking `lockTablesWithIntensiveDbLock` call with `tryLockTablesWithIntensiveDbLock`. When the tryLock fails, a `LockTimeoutException` is thrown.
- Add an overloaded 4-arg variant to `GlobalTransactionMgr.canTxnFinished()`; the existing 3-arg overload delegates to it with `lockTimeoutMs=0` (blocking) for full backward compatibility.
- Update `PublishVersionDaemon.tryFinishTransaction()` to pass `Config.finish_transaction_default_lock_timeout_ms` (the same config item already used for `finishTransaction`) and handle `LockTimeoutException` by logging a warning and returning early, allowing a retry on the next daemon cycle.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/DatabaseTransactionMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/DatabaseTransactionMgr.java
@@ -59,6 +59,7 @@ import com.starrocks.common.StarRocksException;
 import com.starrocks.common.TraceManager;
 import com.starrocks.common.util.DebugUtil;
 import com.starrocks.common.util.TimeUtils;
+import com.starrocks.common.util.concurrent.lock.LockTimeoutException;
 import com.starrocks.common.util.concurrent.lock.LockType;
 import com.starrocks.common.util.concurrent.lock.Locker;
 import com.starrocks.lake.LakeTableHelper;
@@ -986,7 +987,8 @@ public class DatabaseTransactionMgr {
     // check whether transaction can be finished or not
     // for each tablet of load txn, if most replicas version publish successed
     // the trasaction can be treated as successful and can be finished
-    public boolean canTxnFinished(TransactionState txn, Set<Long> errReplicas, Set<Long> unfinishedBackends) {
+    public boolean canTxnFinished(TransactionState txn, Set<Long> errReplicas, Set<Long> unfinishedBackends,
+                                  long lockTimeoutMs) throws LockTimeoutException {
         Database db = globalStateMgr.getLocalMetastore().getDb(txn.getDbId());
         if (db == null) {
             return true;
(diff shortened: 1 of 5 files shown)


fe/fe-core/src/main/java/com/starrocks/transaction/DatabaseTransactionMgr.java
```diff
@@ -986,15 +986,25 @@ public List<TransactionStateBatch> getReadyToPublishTxnListBatch() {
...
             return true;
         }
 
         List<Long> tableIdList = txn.getTableIdList();
         Locker locker = new Locker();
-        locker.lockTablesWithIntensiveDbLock(db.getId(), tableIdList, LockType.READ);
+        if (lockTimeoutMs > 0) {
+            if (!locker.tryLockTablesWithIntensiveDbLock(db.getId(), tableIdList, LockType.READ, lockTimeoutMs,
+                    TimeUnit.MILLISECONDS)) {
+                throw new StarRocksException(ERR_LOCK_ERROR,
```
banmoy: use `LockTimeoutException`?

kevincai: shall keep consistent with others, use ErrorReport to convert it to StarRocksException. `LockTimeoutException` is an internal exception that should not be leaking to the user level, throw LockTimeoutException exception here causes all the callers to handle the timeout and convert to StarRocksException if the caller place is a user-face function.

kevincai: After checking the code again, `ErrorReportException` will set error code to connection context which indicates it should be used specifically for end-user request. `StarRocksException` is proper here.

gengjun-git (dismissed):

fe/fe-core/src/main/java/com/starrocks/transaction/DatabaseTransactionMgr.java
wyb: If `LockTimeoutException` is caught in `PublishVersionDaemon.tryFinishTransaction`, it is not leaked to the user.

kevincai: done

wyb (approved):

banmoy (approved):

gengjun-git (approved):

PR #69496, opened 2026-02-25
[BugFix] Allow incremental ingest with replace snapshots in iceberg

kyle-goodale-klaviyo:
## Why I'm doing:
Replace snapshots do not change table data and are very common operations that are the result of iceberg maintenance procedures. Skipping incremental ingest because these snapshots exist will result in large tables which require frequent maintenance runs never being able to incrementally refresh even if they are append only. These are also the tables that would benefit the most from this feature.

## What I'm doing:
Allow incremental ingest when the snapshots are append || replace only. Overwrite snapshots will still avoid the incremental path and trigger a normal partition based refresh.

Fixes #69493

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


kyle-goodale-klaviyo:
I think a better solution may be to just skip over replace operations as no data is changed in a replace operation. I'll update this PR to do that instead

PR #69525, opened 2026-02-26
[Feature] Support Iceberg v3 default value feature

### LINKED ISSUE #69709: Supports the default values feature in Iceberg v3
- [x] https://github.com/StarRocks/starrocks/pull/69525
<!-- (At least include the following, feel free to add if you have more content) -->

## Feature request
<!-- A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] -->

Iceberg v3 introduced default value support for columns.
There can be two defaults with a field:

* `initial-default` is used to populate the field's value for all records that were written before the field was added to the schema.
* `write-default` is used to populate the field's value for any records written after the field was added to the schema, if the writer does not supply the field's value.

However, starrocks has not support this feature yet.

<!-- A clear and concise description of what you want to happen. -->

We plan to support the default values as follows:
* In scan path,   when scanning data files with missing fields, return the initial-default from Iceberg schema instead of NULL.
* In insert path,  when INSERT does not provide a column value, write the write-default.
* Support specifying defaults for columns via `CREATE` and `ALTER` syntax.
```
   Example usage:                                                                                                                    
                                                                                                                                       
     -- Create table with default values                                                                                                
     CREATE TABLE user_info (                                                                                                          
         id INT,                                                                                                                       
         name STRING,                                                                                                                   
         age INT DEFAULT "18",                                                                                                          
         score DOUBLE DEFAULT "100.0",                                                                                                 
         status STRING DEFAULT 'active'                                                                                                
     ) PROPERTIES ("format-version" = "3");                                                                                            
                                                                                                                                        
     -- Insert without specifying all columns (uses defaults)                                                                           
     INSERT INTO user_info (id, name) VALUES (1, 'alice');                                                                              
                                                                                                                                        
     -- Add column with default value                                                                                                   
     ALTER TABLE user_info ADD COLUMN bonus DOUBLE DEFAULT "50.5";                                                                     
                                                                                                                                        
     -- Modify default value                                                                                                            
     ALTER TABLE user_info MODIFY COLUMN status STRING DEFAULT "inactive";
```
* Support displaying the default values via `DESC` and `SHOW CREATE TABLE` syntax.

**Additional context**
<!-- Add any other context or screenshots about the feature request here. -->
Benefits:
* Simplified INSERT statements
* Seamless schema evolution
* Better compatibility with Iceberg v3 specification
* Consistent experience with native StarRocks tables

GavinMar:
## Why I'm doing:
Iceberg format version 3 introduces column default values (`initial-default` and `write-default`), which enable automatic value filling during `INSERT` operations and schema evolution. StarRocks users need this feature to simplify data insertion and handle schema changes seamlessly without manual default value management.

## What I'm doing:
 This PR implements Iceberg v3 default value support in StarRocks, including:
 - FE side: Added default value conversion in `IcebergApiConverter`, updated `IcebergAlterTableExecutor` to support `ADD/MODIFY COLUMN` with defaults, and enhanced `InsertPlanner` and `InsertAnalyzer` to apply write-default values during INSERT operations
 - BE side: Updated `HdfsScanner` to apply initial-default values when reading historical data files that lack newly added columns
 - Supported operations: `CREATE TABLE` with default values, `ALTER TABLE ADD COLUMN` with defaults, `ALTER TABLE MODIFY COLUMN` defaults, and `INSERT` statements with missing columns
 - Data types: Supports INT, BIGINT, FLOAT, DOUBLE, BOOLEAN, STRING, DATE, and TIMESTAMP types
 - Added comprehensive SQL integration tests and documentation
 - 
Fixes [#issue](https://github.com/StarRocks/starrocks/issues/69709)

### Notes
Currently, we do not support new syntax for setting Iceberg's write-default and initial-default values ​​separately. When a user sets a default value for a field, that value will be used as the value for both write-default  and initial-default values. This is generally consistent with user scenarios. We may support new syntax to set these two types of default values ​​separately in the future, based on actual needs.

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
+++ b/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
@@ -45,9 +45,8 @@ static Status fill_default_value_for_not_existed_slot(SlotDescriptor* slot_desc,
         return Status::InternalError(fmt::format("failed to get type info for slot {}", slot_desc->col_name()));
     }
     if (default_value.empty() && !slot_desc->type().is_string_type()) {
-        return Status::InvalidArgument(
-                fmt::format("empty default value is only supported for string-like columns, col_name={}",
-                            slot_desc->col_name()));
+        return Status::InvalidArgument(fmt::format(
+                "empty default value is only supported for string-like columns, col_name={}", slot_desc->col_name()));
     }
 
     // Parse default value into Datum using TypeInfo::from_string
--- a/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
+++ b/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
@@ -16,6 +16,7 @@
 
 #include "cache/data_cache_hit_rate_counter.hpp"
 #include "column/column_helper.h"
+#include "column/datum_convert.h"
 #include "column/type_traits.h"
 #include "common/config_cache_fwd.h"
 #include "common/config_scan_io_fwd.h"
@@ -29,6 +30,7 @@
 #include "storage/predicate_parser.h"
 #include "storage/rowset/default_value_column_iterator.h"
 #include "storage/runtime_range_pruner.hpp"
+#include "storage/type_info_allocator_adapter.h"
 #include "storage/types.h"
 #include "util/compression/compression_utils.h"
 #include "util/compression/stream_decompressor.h"
@@ -42,11 +44,33 @@ static Status fill_default_value_for_not_existed_slot(SlotDescriptor* slot_desc,
     if (type_info == nullptr) {
         return Status::InternalError(fmt::format("failed to get type info for slot {}", slot_desc->col_name()));
     }
-    std::unique_ptr<DefaultValueColumnIterator> default_value_iter = std::make_unique<DefaultValueColumnIterator>(
-            true, default_value, slot_desc->is_nullable(), type_info, slot_desc->type().len, row_count);
-    ColumnIteratorOptions iter_opts;
-    RETURN_IF_ERROR(default_value_iter->init(iter_opts));
-    return default_value_iter->fetch_values_by_rowid(nullptr, row_count, column);
+
+    // For Iceberg: treat the default value string literally
+    // - Empty string means SQL NULL (when IcebergApiConverter returns null in Java)
+    // - "NULL" is a literal string, not SQL null
+    if (default_value.empty()) {
+        if (!slot_desc->is_nullable()) {
+            return Status::InternalError(
+                    fmt::format("non-nullable column {} has empty default value", slot_desc->col_name()));
+        }
+        [[maybe_unused]] bool ok = column->append_nulls(row_count);
+        DCHECK(ok) << "cannot append null to non-nullable column";
+        return Status::OK();
+    }
+
+    // Parse default value into Datum using TypeInfo::from_string
+    // This handles all basic types: INT, BIGINT, FLOAT, DOUBLE, BOOLEAN, STRING, DATE, TIMESTAMP
+    MemPool mem_pool;
+    TypeInfoAllocator type_info_allocator = make_type_info_allocator(&mem_pool);
+    Datum datum;
+    RETURN_IF_ERROR(datum_from_string(type_info.get(), &datum, default_value, &type_info_allocator));
+
+    // Fill column with the default value
+    for (size_t i = 0; i < row_count; ++i) {
+        column->append_datum(datum);
+    }
+
+    return Status::OK();
 }
 
 class CountedSeekableInputStream final : public io::SeekableInputStreamWrapper {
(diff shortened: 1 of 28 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/plan/PlanFragmentBuilder.java
```diff
@@ -1136,7 +1136,9 @@ private void prepareContextSlots(PhysicalScanOperator node, ExecPlan context, Tu
                 SlotDescriptor slotDescriptor =
                         context.getDescTbl().addSlotDescriptor(tupleDescriptor, new SlotId(entry.getKey().getId()));
                 slotDescriptor.setColumn(entry.getValue());
-                slotDescriptor.setIsNullable(entry.getValue().isAllowNull());
+                // External table scan slots are always nullable because the source data
+                // is not governed by StarRocks NOT NULL constraints and may contain nulls.
+                slotDescriptor.setIsNullable(true);
```
Youngwb: Are you sure this method is only used in external table?

GavinMar: > Are you sure this method is only used in external table?

This function mainly be used to scanning data stored in external system, such as external catalogs, and table functions.  The olap table scanning will not use it.  However, just as you worried, because the function name itself isn't restricted to use only by external tables, it could still be misused by olap tables later. To avoid this, I've fixed it and removed the changes from this function and only modify the iceberg path as changed in this PR.

GavinMar: The reason for this change is that we force set the external table columns to nullable during schema transformation before. This caused a bug during writes: if a table was not null and had no default value, we would still write a null value, which was unexpected. Therefore, we keep the original table's null attribute during schema transformation for validation during writes. For reads, we also set it to nullable in the read path to ensure consistency with the original read behavior.

test/sql/test_iceberg/T/test_iceberg_v3_default_value
```diff
@@ -0,0 +1,225 @@
+-- name: testIcebergV3DefaultValueCreateTable
+-- Test case 1: Create iceberg table with default value for a column
+
+shell: ossutil64 mkdir oss://${oss_bucket}/test_iceberg_v3_default_value/${uuid0} >/dev/null || echo "exit 0" >/dev/null
+
+create external catalog `ice_def_hadoop${uuid0}`
```
stephen-shelby, line 6: Did your test cover all types of default values, such as decimal?

be/src/exec/hdfs_scanner/hdfs_scanner.cpp
```diff
@@ -27,13 +28,51 @@
...
+                                                      size_t row_count, Column* column) {
+    auto type_info = get_type_info(slot_desc->type());
+    if (type_info == nullptr) {
+        return Status::InternalError(fmt::format("failed to get type info for slot {}", slot_desc->col_name()));
+    }
+
+    // For Iceberg: treat the default value string literally
+    // - Empty string means SQL NULL (when IcebergApiConverter returns null in Java)
+    // - "NULL" is a literal string, not SQL null
+    if (default_value.empty()) {
```
stephen-shelby: This makes empty string defaults indistinguishable from NULL on the read path. `toDefaultValueString()` can return `""` for a valid Iceberg string default, but `fill_default_value_for_not_existed_slot()` treats `default_value.empty()` as SQL NULL. As a result, `DEFAULT ''` will be read back as NULL (or even fail for non-nullable columns).

test/sql/test_iceberg/T/test_iceberg_v3_default_value
stephen-shelby: You can refer to the test cases under this [directory](https://github.com/StarRocks/starrocks/tree/main/test/sql/test_default_value/R)

GavinMar: Good advice, added.

be/src/exec/hdfs_scanner/hdfs_scanner.cpp
GavinMar: fixed.

stephen-shelby (approved):

Youngwb (approved):

wyb (approved):

trueeyu (approved):

PR #69529, opened 2026-02-26
[BugFix] Fix sort key not including newly added key columns after schema change on aggregate/unique tables

banmoy:
## Why I'm doing:

After performing schema change (`ALTER TABLE ADD COLUMN` for key columns) on aggregate or unique key tables (both shared-nothing and shared-data), two problems occur:

1. **SHOW CREATE TABLE** output for the `ORDER BY` clause does not include the newly added key column.
2. **Data import fails** with error: `E: (5609, '127.0.0.1: tablet type: 2 sort key columns is different with key columns')`.

For AGG_KEYS/UNIQUE_KEYS tables, the sort key must include all key columns. The bug causes FE metadata and BE expectations to diverge, leading to both display and import failures.

**Reproduction example:**
```
CREATE TABLE t (k0 INT, k1 INT, k2 INT, v INT SUM)
AGGREGATE KEY(k0, k1, k2) ORDER BY(k2, k1, k0)
DISTRIBUTED BY HASH(k0) BUCKETS 1;

ALTER TABLE t ADD COLUMN k3 INT KEY DEFAULT '0';

SHOW CREATE TABLE t;  -- ORDER BY still shows (k2, k1, k0), missing k3
INSERT INTO t VALUES (1,2,3,0,4);  -- Fails: sort key columns is different with key columns
```

### Root Cause Analysis

**1. `calculateShortKey()` (SchemaChangeHandler)**

- During schema change, this method computes `sortKeyIdxes` and `sortKeyUniqueIds` for the new schema.
- It only **remapped** existing sort key columns to their new positions in the altered schema.
- It **did not append** newly added key columns to the sort key lists.
- For AGG_KEYS/UNIQUE_KEYS, sort key must cover all key columns, so the new key column must be added to sort key.

**2. `applyFastSchemaEvolutionMetaChange()` (SchemaChangeHandler)**

- Used for cloud-native tables with `cloud_native_fast_schema_evolution_v2` (both live update and edit log replay).
- It reconstructed `newSortKeyIdxes` and `newSortKeyUniqueIds` from the **old** `currentIndexMeta.getSortKeyUniqueIds()` instead of using the corrected sort key from `SchemaChangeData`.
- It only updated `sortKeyIdxes` in `MaterializedIndexMeta`, not `sortKeyUniqueIds`, causing inconsistency.
- Even if `calculateShortKey()` were fixed, this path would still overwrite with stale metadata.

**3. Why data import fails**

The BE enforces a constraint in `MemTable::_sort_column_inc()` (`be/src/storage/memtable.cpp`):

- For AGG_KEYS/UNIQUE_KEYS, the sort key columns (as a set) must equal the set of all key columns.
- The tablet schema (including `sort_key_idxes`) is derived from FE metadata.
- If FE sends `sort_key_idxes` that omits the new key column, the BE check fails and returns: `tablet type: 2 sort key columns is different with key columns`.

So the import error is a direct consequence of incorrect sort key metadata propagated from FE to BE.

### Fix

**Summary**: In both `calculateShortKey()` and `applyFastSchemaEvolutionMetaChange()`, for AGG_KEYS/UNIQUE_KEYS tables during schema change, append newly added key columns to the end of the sort key lists, and keep `sortKeyIdxes` and `sortKeyUniqueIds` in sync, so FE metadata matches BE constraints.

1. **`appendNewKeyColumnsToSortKey()` helper**: For AGG_KEYS/UNIQUE_KEYS, append any key columns that are not yet in the sort key to the end of `sortKeyIdxes` and `sortKeyUniqueIds`. New key columns are always appended at the end, regardless of their position in the schema (e.g. AFTER / FIRST).

2. **`calculateShortKey()`**: After remapping existing sort key columns, call `appendNewKeyColumnsToSortKey()` so newly added key columns are included.

3. **`applyFastSchemaEvolutionMetaChange()`**: Use the same logic: remap existing sort key by unique ID, then call `appendNewKeyColumnsToSortKey()`. Update both `sortKeyIdxes` and `sortKeyUniqueIds` in `MaterializedIndexMeta`.

4. **`MaterializedIndexMeta.setSortKeyUniqueIds()`**: Add setter so `sortKeyUniqueIds` can be updated programmatically.

### Scope

- **Shared-nothing**: Fixed via `calculateShortKey()` → `SchemaChangeJobV2`.
- **Shared-data non-fast path**: Fixed via `calculateShortKey()` → `LakeTableSchemaChangeJob`.
- **Shared-data fast path (cloud_native_fast_schema_evolution_v2)**: Fixed via both `calculateShortKey()` and `applyFastSchemaEvolutionMetaChange()`.

### Tests

- `LakeFastSchemaChangeTestBase`: shared-data fast schema evolution (V2 and non-V2).
- `SchemaChangeHandlerTest`: shared-nothing.
- `LakeTableSchemaChangeJobTest`: shared-data non-fast schema change.
- SQL integration tests in `test/sql/test_sort_key/`.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [X] 4.1
  - [X] 4.0
  - [X] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -1708,6 +1708,9 @@ public class SchemaChangeHandler extends AlterHandler {
                     sortKeyUniqueIds.add(alterSchema.get(sortKeyIdx).getUniqueId());
                 }
             }
+
+            appendNewKeyColumnsToSortKey(olapTable.getKeysType(), alterSchema,
+                    sortKeyIdxes, useSortKeyUniqueId ? sortKeyUniqueIds : null);
         }
 
         if (!sortKeyIdxes.isEmpty()) {
@@ -1744,6 +1747,31 @@ public class SchemaChangeHandler extends AlterHandler {
             boolean isShortKeyChanged = isShortKeyChanged(originShortKeyColumns, newShortKeyColumns);
             dataBuilder.withNewIndexMetaIdToShortKeyCount(alterIndexMetaId,
                     newShortKeyCount, isShortKeyChanged).withNewIndexMetaIdToSchema(alterIndexMetaId, alterSchema);
+        }
+    }
+
+    /**
+     * For AGG_KEYS/UNIQUE_KEYS tables, sort key must include all key columns.
+     * This method appends newly added key columns to the end of the sort key lists,
+     * regardless of where the column was inserted in the schema (e.g. AFTER / FIRST).
+     *
+     * @param sortKeyUniqueIds if non-null, newly appended columns' unique IDs are also added to this list
+     */
+    private static void appendNewKeyColumnsToSortKey(
+            KeysType keysType, List<Column> schema,
+            List<Integer> sortKeyIdxes, @Nullable List<Integer> sortKeyUniqueIds) {
+        if (keysType != KeysType.AGG_KEYS && keysType != KeysType.UNIQUE_KEYS) {
+            return;
+        }
+        Set<Integer> existing = new HashSet<>(sortKeyIdxes);
+        for (int i = 0; i < schema.size(); i++) {
+            Column col = schema.get(i);
+            if (col.isKey() && !existing.contains(i)) {
+                sortKeyIdxes.add(i);
+                if (sortKeyUniqueIds != null) {
+                    sortKeyUniqueIds.add(col.getUniqueId());
+                }
+            }
         }
     }
 
@@ -3030,6 +3058,7 @@ public class SchemaChangeHandler extends AlterHandler {
 
                 List<Integer> sortKeyUniqueIds = currentIndexMeta.getSortKeyUniqueIds();
                 List<Integer> newSortKeyIdxes = new ArrayList<>();
+                List<Integer> newSortKeyUniqueIds = new ArrayList<>();
                 if (sortKeyUniqueIds != null) {
                     for (Integer uniqueId : sortKeyUniqueIds) {
                         Optional<Column> col = indexSchema.stream().filter(c -> c.getUniqueId() == uniqueId).findFirst();
@@ -3038,12 +3067,17 @@ public class SchemaChangeHandler extends AlterHandler {
                         }
                         int sortKeyIdx = indexSchema.indexOf(col.get());
                         newSortKeyIdxes.add(sortKeyIdx);
+                        newSortKeyUniqueIds.add(uniqueId);
                     }
+
+                    appendNewKeyColumnsToSortKey(olapTable.getKeysType(), indexSchema,
+                            newSortKeyIdxes, newSortKeyUniqueIds);
                 }
 
                 currentIndexMeta.setSchema(indexSchema);
                 if (!newSortKeyIdxes.isEmpty()) {
                     currentIndexMeta.setSortKeyIdxes(newSortKeyIdxes);
+                    currentIndexMeta.setSortKeyUniqueIds(newSortKeyUniqueIds);
                 }
 
                 int currentSchemaVersion = currentIndexMeta.getSchemaVersion();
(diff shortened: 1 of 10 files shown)


banmoy:
@cursor review

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -3038,12 +3067,17 @@ public void applyFastSchemaEvolutionMetaChange(Database db, OlapTable olapTable,
                         }
                         int sortKeyIdx = indexSchema.indexOf(col.get());
                         newSortKeyIdxes.add(sortKeyIdx);
+                        newSortKeyUniqueIds.add(uniqueId);
                     }
+
+                    appendNewKeyColumnsToSortKey(olapTable.getKeysType(), indexSchema,
+                            newSortKeyIdxes, newSortKeyUniqueIds);
```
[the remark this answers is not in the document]
banmoy: In this code path (applyFastSchemaEvolutionMetaChange), we only handle tables that support fast schema change. For such tables:

1. Shared-data (cloud-native): Tables that support fast schema change are created with sortKeyUniqueIds populated in OlapTableFactory (lines 307–314).
2. Shared-nothing: Columns always have unique IDs, and sortKeyUniqueIds is set when the table is created with useFastSchemaEvolution=true.

So in practice, sortKeyUniqueIds should not be empty for tables that reach this path. The scenario you described (upgraded tables that only persisted sortKeyIdxes) is unlikely to occur here.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -1747,6 +1750,31 @@ private void calculateShortKey(OlapTable olapTable, long alterIndexMetaId, List<
...
+            KeysType keysType, List<Column> schema,
+            List<Integer> sortKeyIdxes, @Nullable List<Integer> sortKeyUniqueIds) {
+        if (keysType != KeysType.AGG_KEYS && keysType != KeysType.UNIQUE_KEYS) {
+            return;
+        }
+        Set<Integer> existing = new HashSet<>(sortKeyIdxes);
+        for (int i = 0; i < schema.size(); i++) {
+            Column col = schema.get(i);
+            if (col.isKey() && !existing.contains(i)) {
+                sortKeyIdxes.add(i);
```
sevev, line 1770: The columns order in `sortKeyIdxes` should match the sort key order

banmoy: > New key columns are always appended at the end, regardless of their position in the schema (e.g. AFTER / FIRST).

This PR has this assumption

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -1747,6 +1750,31 @@ private void calculateShortKey(OlapTable olapTable, long alterIndexMetaId, List<
         }
     }
 
+    /**
+     * For AGG_KEYS/UNIQUE_KEYS tables, sort key must include all key columns.
+     * This method appends newly added key columns to the end of the sort key lists,
+     * regardless of where the column was inserted in the schema (e.g. AFTER / FIRST).
```
wyb, line 1756: How should we handle the case that the user wants to change the sort key order at the same time?

If the user explicitly specifies the sort key when creating the table, the column order of the aggregate key doesn't really matter. In that case, using AFTER FIRST should probably modify the sort key order？

banmoy: This PR does not consider this requirement. We can support it later if it's a must. The urgent requirement currently is to put the new column at the last of agg keys, so it make sense to also put it at the last of sort keys.

wyb (approved):

sevev (approved):

gengjun-git (approved):

PR #69594, opened 2026-02-27
[Enhancement] remove high-cost log location patterns to fix FE latency 

yohengyang:
## Why I'm doing this

FE production monitoring observed HTTP/JDBC latency spikes, sometimes reaching 10s to 14s.

In one dial-test case, Arthas showed that most of the latency in the connection path was spent in Log4j2 caller-location lookup rather than log file I/O. The affected FE nodes were already using NVMe disks, and the trace showed the time under `StackLocator.calcLocation()` / `StackWalker.walk()`:

```text
---[5354.485427ms] com.starrocks.mysql.nio.AcceptListener:handleEvent()
    +---[3103.027647ms] org.apache.logging.log4j.Logger:info()
        ---[2643.006046ms] org.apache.logging.log4j.util.StackLocator:calcLocation()
            ---[2622.302338ms] java.lang.StackWalker:walk()
```

When the log layout contains `%C`, `%M`, `%L`, or JSON `source` resolvers, Log4j2 needs to walk the stack to resolve caller class, method, file, and line number for each emitted log event. This cost is usually unnecessary in production FE logs.

## Scope of the issue

This is likely a general production risk rather than a problem tied to one specific log statement.

The mechanism is common: when `%C`, `%M`, `%L`, or JSON `source` resolvers are configured, caller-location lookup introduces stack walking into the logging path for emitted FE log events.

The visible impact is workload-dependent. A single `StackWalker.walk()` may be cheap in some clusters or observation windows. However, production FE workloads contain many hot log paths, including connection handling, query registration, transaction replay, scheduler tasks, and schema-change replay. If caller-location lookup happens on a hot path or while metadata locks are held, the cost can be amplified into visible latency.

We observed this in at least two production cases:

- a dial-test connection path, where `AcceptListener.handleEvent()` spent most of its time in `Logger.info()` / `StackWalker.walk()`;
- a schema-change replay case on an older 3.1-based branch, where the table had about 20k partitions and replay/finish emitted one `visualise the shadow index` INFO log per partition while metadata DB write locks were held.

In the schema-change case, the DB lock information is important. The slow-lock dump showed the `replayer` thread holding a metadata DB write lock for about 30s, while several `starrocks-mysql-nio-pool-*` threads were waiting for the same lock. The `replayer` stack was inside:

```text
StackWalker.walk
  -> StackLocator.calcLocation
  -> AbstractLogger.info
  -> PhysicalPartition.visualiseShadowIndex
  -> SchemaChangeJobV2.replayFinished
  -> EditLog.loadJournal
  -> GlobalStateMgr.replayJournalInner
```

So this was not only high-frequency logging. It was high-frequency caller-location lookup while a metadata DB write lock was held, which amplified the impact on FE latency.

This PR does not claim that every FE latency issue is caused by logging. It removes an avoidable stack-walking cost from production FE logging layouts.

## Latency improvement after the change

After applying this change in affected production clusters, FE latency dropped significantly.

Observed improvements include:

- FE HTTP request latency dropped from second-level spikes to millisecond-level in affected clusters.
- Dial-test query latency became stable after removing caller-location lookup.
- In the schema-change replay case, replay/finish became much faster because high-frequency INFO logs no longer triggered stack walking while metadata locks were held.

### FE HTTP latency improvement
image
image

### Dial-test latency improvement

image

## What I'm doing

Optimize FE logging layouts by removing caller-location lookup from production log patterns:

- Plaintext layout: replace `%C{1}.%M():%L` with logger name `%c{1}`.
- JSON layout: remove `source` resolvers for `line`, `file`, and `method`, and set `locationInfoEnabled="false"`.
- Keep useful class identification through the logger name, which does not require stack walking.

## Tests

Added regression tests to verify that:

- plaintext layout no longer contains `%C`, `%M`, or `%L`;
- plaintext layout still contains logger name `%c{1}`;
- JSON layout disables location info and does not contain `source` resolvers;
- JSON log output remains valid and still includes useful fields such as `level`, `logger`, `thread`, and `message`.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Log4jConfig.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Log4jConfig.java
@@ -325,13 +325,11 @@ public class Log4jConfig extends XmlConfiguration {
                             "'level':{'$resolver':'level','field':'name'}," +
                             "'thread.name':{'$resolver':'thread','field':'name'}," +
                             "'thread.id':{'$resolver':'thread','field':'id'}," +
-                            "'line':{'$resolver':'source','field':'lineNumber'}," +
-                            "'file':{'$resolver':'source','field':'fileName'}," +
-                            "'method':{'$resolver':'source','field':'methodName'}," +
+                            "'logger':{'$resolver':'logger','field':'name'}," +
                             "'message':{'$resolver':'message','stringfield':'true'}," +
                             "'exception':{'$resolver':'exception','field':'stackTrace','stackTrace':{'stringified':true,'full':true}}}";
             jsonConfig = jsonConfig.replace("'", "\"");
-            String jsonLayoutFormatter = "<JsonTemplateLayout maxStringLength=\"%d\" locationInfoEnabled=\"true\">\n" +
+            String jsonLayoutFormatter = "<JsonTemplateLayout maxStringLength=\"%d\" locationInfoEnabled=\"false\">\n" +
                     "<EventTemplate><![CDATA[" + jsonConfig + "]]></EventTemplate>\n" + "</JsonTemplateLayout>";
             String jsonLayoutDefault = String.format(jsonLayoutFormatter, Config.sys_log_json_max_string_length);
             String jsonLayoutProfile = String.format(jsonLayoutFormatter, Config.sys_log_json_profile_max_string_length);
@@ -344,9 +342,9 @@ public class Log4jConfig extends XmlConfiguration {
         } else {
             // fallback to plaintext logging
             properties.put("syslog_default_layout",
-                    "<PatternLayout charset=\"UTF-8\" pattern=\"%d{yyyy-MM-dd HH:mm:ss.SSSXXX} %p (%t|%tid) [%C{1}.%M():%L] %m%n\"/>");
+                    "<PatternLayout charset=\"UTF-8\" pattern=\"%d{yyyy-MM-dd HH:mm:ss.SSSXXX} %p (%t|%tid) [%c{1}] %m%n\"/>");
             properties.put("syslog_warning_layout",
-                    "<PatternLayout charset=\"UTF-8\" pattern=\"%d{yyyy-MM-dd HH:mm:ss.SSSXXX} %p (%t|%tid) [%C{1}.%M():%L] %m%n %ex\"/>");
+                    "<PatternLayout charset=\"UTF-8\" pattern=\"%d{yyyy-MM-dd HH:mm:ss.SSSXXX} %p (%t|%tid) [%c{1}] %m%n%ex\"/>");
             properties.put("syslog_audit_layout",
                     "<PatternLayout charset=\"UTF-8\" pattern=\"%d{yyyy-MM-dd HH:mm:ss.SSSXXX} [%c{1}] %m%n\"/>");
             properties.put("syslog_dump_layout",
--- a/fe/fe-core/src/main/java/com/starrocks/common/Log4jConfig.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Log4jConfig.java
@@ -358,13 +358,11 @@ public class Log4jConfig extends XmlConfiguration {
                             "'level':{'$resolver':'level','field':'name'}," +
                             "'thread.name':{'$resolver':'thread','field':'name'}," +
                             "'thread.id':{'$resolver':'thread','field':'id'}," +
-                            "'line':{'$resolver':'source','field':'lineNumber'}," +
-                            "'file':{'$resolver':'source','field':'fileName'}," +
-                            "'method':{'$resolver':'source','field':'methodName'}," +
+                            "'logger':{'$resolver':'logger','field':'name'}," +
                             "'message':{'$resolver':'message','stringfield':'true'}," +
                             "'exception':{'$resolver':'exception','field':'stackTrace','stackTrace':{'stringified':true,'full':true}}}";
             jsonConfig = jsonConfig.replace("'", "\"");
-            String jsonLayoutFormatter = "<JsonTemplateLayout maxStringLength=\"%d\" locationInfoEnabled=\"true\">\n" +
+            String jsonLayoutFormatter = "<JsonTemplateLayout maxStringLength=\"%d\" locationInfoEnabled=\"false\">\n" +
                     "<EventTemplate><![CDATA[" + jsonConfig + "]]></EventTemplate>\n" + "</JsonTemplateLayout>";
             String jsonLayoutDefault = String.format(jsonLayoutFormatter, Config.sys_log_json_max_string_length);
             String jsonLayoutProfile = String.format(jsonLayoutFormatter, Config.sys_log_json_profile_max_string_length);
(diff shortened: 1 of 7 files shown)


gengjun-git:
@codex review

gengjun-git:
@yohengyang Could I take a look at your test cases? Why does logging account for most of the latency?

yohengyang:
> @yohengyang Could I take a look at your test cases? Why does logging account for most of the latency?

@gengjun-git Thanks for the question. I added regression tests for this change:

1. plaintext layout no longer contains `%C`/`%M`/`%L` and uses logger name instead;
2. JSON layout disables `locationInfoEnabled` and does not use `source` resolvers;
3. JSON log output is still valid and contains `level`/`logger`/`message`/`thread` fields.

For the latency question, the expensive part was not log file I/O. The affected FE machines were already using NVMe disks, and Arthas showed the time under caller-location lookup:

`AbstractLogger.getLocation() -> StackLocator.calcLocation() -> StackWalker.walk()`

In one dial-test case, `AcceptListener.handleEvent()` took 5354ms, and `Logger.info()` took 3103ms. We also observed a schema-change replay case on an older 3.1-based branch with about 20k partitions, where one `visualise the shadow index` INFO log was emitted per partition while metadata locks were held. After applying this patch, schema-change replay became much faster and FE HTTP latency dropped significantly.

This patch does not claim that every FE latency issue is caused by logging. It removes an avoidable stack-walking cost from FE production log layouts, especially for hot paths and lock-held paths.

gengjun-git:
However, without this information, locating the problem would be very difficult.

yohengyang:
@gengjun-git I performed an additional production-safe investigation and now have a more concrete answer for why logging can account for most of the latency.

### Production evidence

The affected FE uses Zulu JDK 11 with ZGC and the plaintext pattern `%C{1}.%M():%L`.

During one recurring burst, we observed **407 INFO events in 221 ms**, involving **122 distinct logging threads**. The high-rate batch came from `QeProcessorImpl.reportExecStatus()`. Around these bursts, FE slow-HTTP warnings reported 5–9 second latency.

A wall-clock Arthas profile captured this synchronous path:

```text
QeProcessorImpl.reportExecStatus()
  -> AbstractLogger.info()
  -> AbstractLogger.getLocation()
  -> StackLocator.calcLocation()
  -> StackWalker.walk()
  -> ResolvedMethodTable::lookup/add
  -> ResolvedMethodTable_lock
  -> pthread_cond_wait
```

In the corresponding CPU profile, `ResolvedMethodTable::lookup` accounted for **21.16%** of sampled CPU, while `fdatasync` accounted for only **0.01%**. This indicates that caller-location resolution, rather than writing the log file, was the important cost in that observation window.

The JDK 11 implementation uses a fixed 1007-bucket `ResolvedMethodTable`, and insertion is protected by a global lock. This is the same mechanism described by OpenJDK [JDK-8221393](https://bugs.openjdk.org/browse/JDK-8221393), “ResolvedMethodTable too small for StackWalking applications”.

### Controlled local replay

I replayed the production-shaped schedule with a real synchronous Log4j2 `RollingFileAppender`.

Common configuration:

- Linux/amd64 container, 8 processors;
- Azul JDK 11.0.26 with ZGC, `-Xms2g -Xmx2g`;
- 407 messages using the observed 221 ms millisecond-bucket schedule;
- 122 workers;
- one warm-up pair and six measured pairs;
- old/new execution order alternated, with GC before each burst.

#### Aggregate results

| Precondition | `%C{1}.%M():%L` logger-call p99 | `%c{1}` logger-call p99 | Paired p50 delta (old - new) | Paired p50 ratio | End-to-end p99, old / new |
|---|---:|---:|---:|---:|---:|
| Fresh JVM (`0 x 0`) | 1088.425 ms | 864.540 ms | **+237.952 ms** | **1.321x** | 1574.647 / 1247.895 ms |
| Synthetic retained frames (`1000 x 100 = 100k`) | 1314.119 ms | 846.677 ms | **+227.608 ms** | **1.306x** | 2126.414 / 921.305 ms |

The two logger-call p99 columns are computed across all 2,442 calls for each pattern. The paired p50 columns are computed from the six per-burst p99 differences using the reproducer's nearest-rank percentile.

#### Raw six paired bursts for the 100k synthetic precondition

| Pair | Execution order | Old logger-call p99 | New logger-call p99 | Delta (old - new) | Ratio |
|---:|---|---:|---:|---:|---:|
| 0 | new first | 1034.532 ms | 729.176 ms | **+305.356 ms** | 1.419x |
| 1 | old first | 1318.652 ms | 848.449 ms | **+470.203 ms** | 1.554x |
| 2 | new first | 971.230 ms | 743.623 ms | **+227.608 ms** | 1.306x |
| 3 | old first | 926.961 ms | 629.909 ms | **+297.052 ms** | 1.472x |
| 4 | new first | 747.596 ms | 783.642 ms | -36.046 ms | 0.954x |
| 5 | old first | 762.970 ms | 731.245 ms | +31.725 ms | 1.043x |

The old caller-location pattern was slower in four of six pairs; two pairs were close/noisy. The fresh-JVM run also showed a comparable paired p50 penalty, so the result does not depend solely on the synthetic 100k condition.

These absolute latencies are **not** intended as a prediction of production latency: 122 workers are deliberately contending on an 8-processor test environment. The relevant signal is the paired difference under the same appender, message schedule, JVM and worker count. The production evidence remains the captured synchronous call chain and profiler hotspot.

Also, not every log burst produced a slow HTTP request, so concurrency alone is not a sufficient condition. However, when the slow request was captured, it spent about 2.62 seconds directly in `StackWalker.walk()`, which explains how logging dominated that request's latency.

The individual `LoadAction` redirect INFO message was not itself a 5-second log call. The issue is the aggregate synchronous caller-location work under concurrent logging.

### Reproducer code

I pushed the manual performance reproducer in commit [`776dab5`](https://github.com/StarRocks/starrocks/commit/776dab52518):

- [`Log4jRollingFileBurstReplay`](https://github.com/yohengyang/starrocks/blob/log4j-boost/fe/fe-core/src/test/java/com/starrocks/benchmark/Log4jRollingFileBurstReplay.java) replays the 407-event/221-ms schedule, alternates the patterns, and saves each sample plus a summary.
- [`Log4jStackWalkerPreconditioner`](https://github.com/yohengyang/starrocks/blob/log4j-boost/fe/fe-core/src/test/java/com/starrocks/benchmark/Log4jStackWalkerPreconditioner.java) creates the bounded synthetic retained-frame condition.
- The two JMH classes provide isolated latency/throughput diagnostics. These measurements are intentionally not CI timing assertions.

Example JDK 11 run after compiling FE test classes:

```bash
MAVEN_OPTS="-Xms2g -Xmx2g -XX:+UnlockExperimentalVMOptions -XX:+UseZGC" \
mvn -f fe/pom.xml -pl fe-core -DskipTests \
  -Dexec.classpathScope=test \
  -Dexec.mainClass=com.starrocks.benchmark.Log4jRollingFileBurstReplay \
  -Dexec.args="production 122 1 1000 100 1 6 1000 /tmp/log4j-burst.csv old-first gc-before-each-burst rolling-file" \
  exec:java
```

This PR keeps component identification through `%c{1}` (the logger name) and removes only the expensive caller method/file/line lookup. The existing regression tests verify the resulting plaintext and JSON layouts; the performance code makes the supporting diagnosis reproducible without adding a flaky timing assertion to CI.

Issue #68875, opened 2026-02-03
Data loss - DROP VIEW over another MV cleared all data from underlaying MV


apatrida:
### Steps to reproduce the behavior (Required)
I have an MV `recsys.watch_time_segment_normalized` with 500 million rows in it. Took all night to fill.

I created a view on it.  

```
CREATE OR REPLACE VIEW watch_time_segment_normalized_for_export AS
SELECT * EXCEPT (event_value_json, content_attribution_json),
           CAST(event_value_json AS VARCHAR) AS event_value_json,
           CAST(content_attribution_json AS VARCHAR) AS content_attribution_json
    FROM recsys.watch_time_segment_normalized;
```

I export some data, then drop my export view...

```
...export running and wrote a few 1G0ish parquet files
[2026-02-03 09:21:30] Cancelling…
[2026-02-03 09:21:33] completed in 40 s 679 ms
[2026-02-03 09:21:33] Rumble AI - Starrocks: process exited
[2026-02-03 09:31:10] Connected
[2026-02-03 09:31:10] recsys> use recsys
[2026-02-03 09:31:10] completed in 55 ms
[2026-02-03 09:31:10] recsys> drop   VIEW watch_time_segment_normalized_for_export
[2026-02-03 09:31:10] completed in 172 ms
[2026-02-03 09:32:22] recsys> SELECT count(*) AS count, min(created_at), max(created_at) FROM recsys.watch_time_segment_normalized
[2026-02-03 09:32:25] 1 row retrieved starting from 1 in 3 s 260 ms (execution: 2 s 906 ms, fetching: 354 ms)
```

I check the original MV view and it is now empty.

EMPTY, missed deadline, other developers out of work for the day.

Don't think this is intended no?

I can see my query execution log, there is no other statements.

### Expected behavior (Required)

drop the view I dropepd, NOT THE DATA IN UNDERLAYING TABLE.

### Real behavior (Required)

deleted all data from the MV referenced from the dropped view.

### StarRocks version (Required)
4.0.4

PR #69690, opened 2026-03-02
[BugFix] Preserve MV version map on DROP VIEW to prevent unnecessary full re-refresh

osscm:
Fix issue #68875: DROP VIEW over a base view was unconditionally clearing the version map of dependent materialized views, causing all MV partitions to be treated as fully stale on the next refresh cycle and triggering an unnecessary INSERT OVERWRITE over all partitions.

Root cause: `inactiveRelatedMaterializedViewsRecursive`() always passed clearVersionMap=true. A VIEW is a logical object — its MV physical partition data remains valid on disk after the view is dropped. Only schema changes and DROP TABLE warrant clearing the version map.

Fix:
- Add `inactiveRelatedMaterializedViewsRecursive`(Table, String, boolean) overload in `AlterMVJobExecutor` so callers can control whether the version map is cleared.
- Override `View.onDrop()` to call with `clearVersionMap=false`, preserving the version map so re-activation and refresh can be incremental.
- DROP TABLE continues to pass `clearVersionMap=true` since the base table is physically gone and a full refresh is correct if the table is re-created.
- Add `DropViewOrTableClearsVersionMapBugTest` regression test covering DROP TABLE, DROP VIEW, and the cascading scenario.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] This is a backport pr

- [ ] 4.1
- [ ] 4.0
- [ ] 3.5
- [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/AlterMVJobExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/AlterMVJobExecutor.java
@@ -1242,6 +1242,18 @@ public class AlterMVJobExecutor extends AlterJobExecutor {
      *  has broken from mv existed refreshed data.
      */
     public static void inactiveRelatedMaterializedViewsRecursive(Table olapTable, String reason) {
+        inactiveRelatedMaterializedViewsRecursive(olapTable, reason, true);
+    }
+
+    /**
+     * Inactive related materialized views because of base table/view is changed or dropped in the leader background.
+     *
+     * @param clearVersionMap true for schema-change / DROP TABLE events (version map must be cleared so MV is fully
+     *                        refreshed); false for DROP VIEW events (version map must be preserved — existing MV
+     *                        partition data is still valid).
+     */
+    public static void inactiveRelatedMaterializedViewsRecursive(Table olapTable, String reason,
+                                                                 boolean clearVersionMap) {
         if (olapTable == null) {
             return;
         }
@@ -1263,7 +1275,7 @@ public class AlterMVJobExecutor extends AlterJobExecutor {
                 LOG.info("Ignore materialized view {} does not exists", mvId);
                 continue;
             }
-            doInactiveMaterializedViewRecursive(mv, reason, true, inactiveMVIds);
+            doInactiveMaterializedViewRecursive(mv, reason, clearVersionMap, inactiveMVIds);
         }
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/View.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/View.java
@@ -37,6 +37,8 @@ package com.starrocks.catalog;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.gson.annotations.SerializedName;
+import com.starrocks.alter.AlterMVJobExecutor;
+import com.starrocks.common.MaterializedViewExceptions;
 import com.starrocks.common.StarRocksException;
 import com.starrocks.common.util.SqlCredentialRedactor;
 import com.starrocks.server.GlobalStateMgr;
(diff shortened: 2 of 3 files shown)


HangyuanLiu (changes requested):
I verified the original issue locally, and this patch does fix the "drop + recreate equivalent view" case. However, preserving the version map for every `DROP VIEW` introduces a correctness regression when the same view name is recreated with a different but schema-compatible definition.

Concrete repro on the patched code:
1. Create `base_tbl` with 2 rows.
2. Create `bug_view AS SELECT k1, k2, v1 FROM base_tbl` and `bug_mv` on top of it, then refresh `bug_mv` so it contains 2 rows.
3. `DROP VIEW bug_view`.
4. Recreate `bug_view` as `SELECT k1, k2, v1 FROM base_tbl WHERE k2 = 1`.
5. `ALTER MATERIALIZED VIEW bug_mv ACTIVE`.
6. `REFRESH MATERIALIZED VIEW bug_mv WITH SYNC MODE`.

Observed result:
- the recreated view returns 1 row
- the MV still returns 2 rows
- `last_refresh_state = SKIPPED`
- `last_refresh_base_refresh_partitions = {}`

So `View.onDrop(..., clearVersionMap = false)` is too broad. `ACTIVE` can succeed after the view is recreated, but it does not prove semantic equivalence with the old view definition. Reusing the old version map in that case lets refresh skip and leaves stale MV data behind.

I think we need an additional guard here: either preserve the version map only when we can prove the recreated view is semantically equivalent, or clear/invalidate it once the analyzed view definition differs. Without that, this change fixes the unnecessary full refresh case but can silently return incorrect MV results.

osscm:
thanks @HangyuanLiu 

>I think we need an additional guard here: either preserve the version map only when we can prove the recreated view is semantically equivalent, or clear/invalidate it once the analyzed view definition differs. Without that, this change fixes the unnecessary full refresh case but can silently return incorrect MV results.

what more guardrail we can add to avoid the incorrectness issue? if you have any suggestions?

HangyuanLiu:
Thanks @osscm. My suggestion is to put the guardrail on the MV re-activation / rebind path, not only in `View.onDrop()`.

The unsafe case is:

1. `DROP VIEW` preserves the old MV version map.
2. The same view name is created again.
3. `ALTER MATERIALIZED VIEW ... ACTIVE` re-analyzes the MV successfully because the new view is schema-compatible.
4. Refresh then reuses the old version map and may decide `SKIPPED`, even though the view semantics changed.

So `ACTIVE` should not only prove that the MV definition can be analyzed and that the output schema is compatible. It also needs to prove that the preserved version map is still valid for the current view dependency.

One conservative guardrail could be:

- When preserving the version map because of `DROP VIEW`, also remember a fingerprint of the view dependency that made the version map valid. The fingerprint can include catalog/db/view name, sql mode, normalized inline view definition, output schema, and the expanded underlying base table dependencies.
- During `ALTER MATERIALIZED VIEW ... ACTIVE`, after re-analyzing the MV and collecting current view dependencies, recompute the fingerprint.
- Preserve the version map only if the fingerprint still matches. If the view is missing, was rebound to a different definition, has different output schema, or its expanded base dependencies changed, clear/invalidate the MV `AsyncRefreshContext` visible version map before setting the MV active.

If we want a smaller first patch, the safest fallback is to clear the version map whenever `ACTIVE` detects that a preserved view dependency was rebound to a different table id. That avoids incorrect results, but it loses the optimization for the equivalent drop/recreate case. The fingerprint approach keeps the optimization only when we can prove it is safe.

I think the tests should cover both directions:

- drop and recreate the same view definition: version map can be preserved and refresh can remain incremental/skipped;
- drop and recreate the same view name with a schema-compatible but semantically different definition, e.g. adding `WHERE k2 = 1`: version map must be cleared and refresh must not be skipped.

PR #69698, opened 2026-03-02
[Enhancement] Trigger early flush in LakePersistentIndex when rebuild row count exceeds threshold

luohaha:
## Why I'm doing:

When recovering a Lake Primary Key index, StarRocks replays all rowset segments and del files that have not yet been checkpointed into SSTables. The existing early-flush strategy in `LakePersistentIndex::commit` only considered the **number of files** to rebuild (`too_many_rebuild_files`). A tablet with a small number of large rowsets could still accumulate millions of rows pending rebuild without triggering a flush, leading to high recovery cost (large amounts of data to read from remote storage and re-insert into the index).

## What I'm doing:

Add a complementary early-flush strategy based on **row count**: when the estimated number of rows that would need to be rebuilt exceeds a configurable threshold, `commit()` triggers a flush just like the existing file-count strategy.

**Key changes:**

- **`need_rebuild_counts()` (new, replaces two separate functions)**: Merges the old `need_rebuild_file_cnt` and `need_rebuild_row_cnt` into a single function returning `std::pair<size_t, int64_t>` so the rowset list is traversed only once. An inner helper `rebuild_segment_counts()` also performs a single segment loop to accumulate both file count and row count simultaneously.
  - Row count uses exact per-segment `num_rows` from `RowsetMetadataPB::segment_metas` when available; falls back to a proportional estimate (`rowset.num_rows * rebuild_seg_cnt / total_seg_cnt`) for older data without per-segment metadata.
  - Del files are counted in `file_cnt` only (their rows are not tracked in metadata).

- **`too_many_rebuild_rows()` (new private method)**: Returns true when `_need_rebuild_row_cnt >= cloud_native_pk_index_rebuild_rows_threshold` and the threshold is non-zero (0 disables the strategy).

- **`too_many_rebuild_files()` (updated)**: Added `> 0` guard to match `too_many_rebuild_rows()` behavior, so setting threshold to 0 correctly disables the strategy.

- **`commit()` updated**: Now checks `too_many_rebuild_files() || too_many_rebuild_rows()` before deciding to flush.

- **New config `cloud_native_pk_index_rebuild_rows_threshold` (mutable, default 10,000,000)**: Controls the row-count threshold. Setting to 0 disables the strategy.

- **Tests**:
  - `LakePersistentIndexTest::test_need_rebuild_counts`: unit test directly exercising `need_rebuild_counts` across 5 scenarios (empty, all-rebuild with segment_metas, partial SST coverage, proportional fallback, del-file handling).
  - `LakePrimaryKeyPublishTest::test_cloud_native_index_minor_compact_because_rows`: integration test that sets a low row threshold and verifies SST files are generated after the 3rd publish triggers the early flush.

## What type of PR is this:
- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:
- [ ] This is a backport PR

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1314,6 +1314,10 @@ CONF_mBool(lake_clear_corrupted_cache_data, "false");
 // The maximum number of files which need to rebuilt in cloud native pk index.
 // If files which need to rebuilt larger than this, we will flush memtable immediately.
 CONF_mInt32(cloud_native_pk_index_rebuild_files_threshold, "50");
+// The maximum number of rows which need to be rebuilt in cloud native pk index.
+// If rows which need to be rebuilt exceed this threshold, we will flush memtable immediately
+// to reduce index rebuild cost. 0 means disabled.
+CONF_mInt64(cloud_native_pk_index_rebuild_rows_threshold, "10000000");
 // if set to true, CACHE SELECT will only read file, save CPU time
 // if set to false, CACHE SELECT will behave like SELECT
 CONF_mBool(lake_cache_select_in_physical_way, "true");
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -116,7 +116,8 @@ bool LakePersistentIndex::is_memtable_full() const {
 }
 
 bool LakePersistentIndex::too_many_rebuild_files() const {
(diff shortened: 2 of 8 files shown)


CelerData-Reviewer:
@codex review

wyb (approved):

kevincai (approved):

PR #69747, opened 2026-03-03
[Enhancement] Split txn publish latency and add publish daemon loop metric

kevincai:
* add publish_version_daemon_loop_total to count publish daemon runs
* add readyToFinishTime and txn_publish_can_finish_latency_ms to split publish latency stages
* change publish ack latency to measure from ready-to-finish to visible
* improve txn debug output with can-finish and ack timing details

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4


kevincai:
@codex review

banmoy (commented):
> From my cursor agent

## Review Summary

The concept is sound and well-implemented for the shared-nothing path. A few issues to address:

---

### Issue 1 (P0): Shared-data (lake) publish path loses `publishAckLatencyMs`

The `publishAckLatencyMs` metric now measures `readyToFinishTime → finishTime`, but in the shared-data (lake) code path, `readyToFinishTime` is **never set**.

In `PublishVersionDaemon.publishLakeTransactionAsync()`:
```java
return publishFuture.thenAccept(success -> {
    if (success) {
        try {
            txnState.updatePublishTaskFinishTime();
            globalTransactionMgr.finishTransaction(dbId, txnId, null);
        } catch (StarRocksException e) {
            throw new RuntimeException(e);
        }
    }
})
```

There is no `setReadyToFinishTimeIfUnset()` call here. Since `readyToFinishTime` stays at `-1`, `recordLatencyIfValid` silently skips both `publishCanFinishLatencyMs` and `publishAckLatencyMs` for **all** lake transactions.

**Before this PR**, `publishAckLatencyMs` was `publishVersionFinishTime → finishTime` and worked for lake transactions. **After this PR**, that metric is lost for shared-data mode — this is a regression.

**Suggestion**: Add `txnState.setReadyToFinishTimeIfUnset()` right after `updatePublishTaskFinishTime()` in `publishLakeTransactionAsync` (and in `publishLakeTransactionBatchAsync` if applicable). For lake transactions there's no quorum check, so `readyToFinishTime ≈ publishVersionFinishTime`, preserving existing ack metric behavior.

---

### Issue 2 (P1): Minor timing inaccuracy in the `else` branch of `tryFinishTransaction`

When all publish tasks are already finished (the `else` branch at line 377), `setReadyToFinishTimeIfUnset()` calls `System.currentTimeMillis()` — capturing the **current daemon loop iteration time**, not the time when tasks actually finished. If this runs in a later daemon loop iteration than when `publishVersionFinishTime` was set, `publishCanFinishLatencyMs` will be inflated by the daemon loop interval.

Consider using `publishVersionFinishTime` as the timestamp in this specific case, since the transaction was ready to finish as soon as all publish tasks completed.

---

### Issue 3 (P2): Thread safety of `readyToFinishTime`

The field is a plain `long` and `setReadyToFinishTimeIfUnset()` uses a check-then-act pattern without synchronization. In practice this is likely safe because all callers run on the publish daemon thread. However, making the field `volatile` would make the single-writer intent clear and cost essentially nothing.

---

### Issue 4 (P3): `TransactionMetrics` javadoc is stale

The `TransactionMetrics` class javadoc says:
```
publishLatencyMs = publishScheduleLatencyMs + publishExecuteLatencyMs + publishAckLatencyMs
```
This should now be updated to include `publishCanFinishLatencyMs`:
```
publishLatencyMs = schedule + execute + can_finish + ack
```

---

### Issue 5 (P3): `publish_version_daemon_loop_total` docs placement

This counter is placed inside the "Transaction Latency Metrics" section, but it's a counter measuring daemon loop iterations, not a latency/summary metric. Consider a separate subsection or a note clarifying it's a cumulative counter distinct from the surrounding summary metrics.

---

### Issue 6 (P3): PR template sections are empty

The "Why I'm doing" and "What I'm doing" sections are blank. Filling these in (e.g., what operational problem motivated splitting publish latency) would help reviewers understand the context.

---

Overall the implementation is clean, tests are solid, and docs are thorough across all three languages. **Issue 1 should be fixed before merge** as it silently drops the ack latency metric for all lake/shared-data transactions.

trueeyu (approved):

banmoy (approved):

HangyuanLiu (approved):

PR #69751, opened 2026-03-03
[BugFix] Fix  possible PartitionColumnMinMaxRewriteRule bugs caused by Partition.hasStorageData

LiShuMing:
## Why I'm doing:

Currently, PartitionColumnMinMaxRewriteRule and other some components (e.g., Materialized View refresh logic) rely on the assumption that if a partition's visibleVersion is greater than PARTITION_INIT_VERSION, it implies the partition has data (storageRowCount() > 0). However, this assumption is flawed and can lead to incorrect behaviors, such as MV refresh incorrectly skipping partitions that are actually empty.

The divergence between visibleVersion and actual row count happens in two main scenarios:

- Schema Changes: Operations like adding a column, creating a rollup, or altering metadata can advance the visibleVersion without writing any new data rows. A new shadow index is published and becomes visible with empty or copied metadata, incrementing the version while the row count remains at zero.

- Shared-data (Lake) Mode: In Lake mode, this mismatch is even more highly probable. Alter or rollup jobs update the version directly, but tablet statistics are updated asynchronously. This creates a wider time window where visibleVersion > PARTITION_INIT_VERSION is true, but storageRowCount() == 0.

Fix bugs cuased by: https://github.com/StarRocks/starrocks/pull/49942/

## What I'm doing:
- Add `dataVersion` to check.

TODO:
- Support more common optimizations for `min/max` partitions to compute.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
@@ -642,7 +642,7 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
         // The fe unit test need to check the selected index id without any data.
         // So if set FeConstants.runningUnitTest, we can ensure that the number of partitions is not empty,
         // And the test case can continue to execute the logic of 'select best roll up'
-        return ((visibleVersion != PARTITION_INIT_VERSION)
+        return ((dataVersion != PARTITION_INIT_VERSION)
                 || FeConstants.runningUnitTest);
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
@@ -642,7 +642,7 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
         // The fe unit test need to check the selected index id without any data.
         // So if set FeConstants.runningUnitTest, we can ensure that the number of partitions is not empty,
         // And the test case can continue to execute the logic of 'select best roll up'
-        return ((dataVersion != PARTITION_INIT_VERSION)
+        return ((dataVersion != PARTITION_INIT_VERSION && visibleVersion != PARTITION_INIT_VERSION)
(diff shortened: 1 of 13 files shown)


fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
```diff
@@ -642,7 +642,7 @@ public boolean hasStorageData() {
         // The fe unit test need to check the selected index id without any data.
         // So if set FeConstants.runningUnitTest, we can ensure that the number of partitions is not empty,
         // And the test case can continue to execute the logic of 'select best roll up'
-        return ((visibleVersion != PARTITION_INIT_VERSION)
+        return ((dataVersion != PARTITION_INIT_VERSION && visibleVersion != PARTITION_INIT_VERSION)
```
kevincai, line 645: It's better to add comments to this interface to explain the consequence of `false positive` and `false negative`, will it be absolutely accurate or can be somehow inaccurate?

kevincai: also for duplicate key table, what if a partition in inserted with data and then get cleared with `delete from table {table}`, the table is logically empty, but the data version and visible version will be never back to INIT VERSION any more.

LiShuMing: > also for duplicate key table, what if a partition in inserted with data and then get cleared with delete from table {table}, the table is logically empty, but the data version and visible version will be never back to INIT VERSION any more. 

This case is already handled in `PartitionColumnMinMaxRewriteRule` .

xiangguangyxg (approved):

kevincai (approved):

wyb (approved):

PR #69817, opened 2026-03-04
[BugFix] Fix reflection for nested column with comments

KarolGongolaCledar:
## Why I'm doing:

DataHub ingestion through SQLAlchemy with `platform: starrocks` fails to properly reflect nested StarRocks column types when `COLUMN_TYPE` contains inline field comments (for example `COMMENT ''` inside `STRUCT/ARRAY/MAP`).  
This causes parse warnings and prevents downstream nested schema extraction from working correctly.
I  got this error when tried to ingest delta table metadata from starrocs to datahub via datahub cli ingestion which uses sqlalchemy.

## What I'm doing:

- Normalize reflected StarRocks `COLUMN_TYPE` strings in reflection by removing inline `COMMENT '...'` fragments before datatype parsing.
- Keep datatype parser strict (no fallback normalization inside `parse_data_type`).
- Add/adjust unit test coverage in reflection path to verify nested complex types with inline comments are parsed correctly.

Fixes #<issue-number>

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/contrib/starrocks-python-client/test/unit/test_parser.py
+++ b/contrib/starrocks-python-client/test/unit/test_parser.py
@@ -144,6 +144,33 @@ class TestDataTypeParser:
     def test_nested_complex_types(self, type_str, expected_type):
         assert repr(parse_data_type(type_str)) == repr(expected_type)
 
+    def test_nested_complex_types_with_inline_comments(self):
+        type_str = (
+            "struct<"
+            "headers array<struct<key varchar(65533) COMMENT '', value varchar(65533) COMMENT ''>> COMMENT '', "
+            "payload struct<event_id bigint COMMENT '', labels map<string, string> COMMENT ''> COMMENT ''"
+            ">"
+        )
+        expected_type = datatype.STRUCT(
+            (
+                "headers",
+                datatype.ARRAY(
+                    datatype.STRUCT(
+                        ("key", datatype.VARCHAR(65533)),
+                        ("value", datatype.VARCHAR(65533)),
+                    )
+                ),
+            ),
+            (
+                "payload",
+                datatype.STRUCT(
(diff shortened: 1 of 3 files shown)


contrib/starrocks-python-client/test/unit/test_parser.py
```diff
@@ -144,6 +144,33 @@ def test_complex_types(self, type_str, expected_type):
     def test_nested_complex_types(self, type_str, expected_type):
         assert repr(parse_data_type(type_str)) == repr(expected_type)
 
+    def test_nested_complex_types_with_inline_comments(self):
+        type_str = (
+            "struct<"
+            "headers array<struct<key varchar(65533) COMMENT '', value varchar(65533) COMMENT ''>> COMMENT '', "
```
jaogoy: Will there be some characters in the quoted comment?  If so we need to test it.

jaogoy (commented):
It's a easy and workable way.
And for long term and elegancy, we need to use lark grammar to parse it.  And it will be appreciate if you can try to submit another pr to use the lark grammar.

contrib/starrocks-python-client/test/unit/test_parser.py
KarolGongolaCledar: Tests added

jaogoy (dismissed):

jaogoy (approved):

nshangyimingCD (approved):

PR #66522, opened 2025-12-09
[Enhancement] Add range-split, parallel compaction and size-tiered compaction strategy for cloud native pk index (part1)

luohaha:
## Why I'm doing:
This PR introduces significant enhancements to the primary key (PK) index compaction and parallelism mechanisms in shared-data mode for StarRocks. The changes implement a size-tiered compaction strategy, range-splitting functionality for parallel compaction, and related infrastructure including new thread pools and configuration parameters. The implementation adds three major components: size-tiered compaction strategy for selecting candidates, parallel compaction manager for executing compaction tasks concurrently, and fileset abstraction for managing non-overlapping SSTable ranges.
More details : #66457

* size-tiered compaction strategy. (`lake/lake_persistent_index_size_tiered_compaction_strategy.cpp`)
* range-split. (`lake/persistent_index_sstable_fileset.cpp`)
* parallel pk index compact. (`lake/lake_persistent_index_parallel_compact_mgr.cpp`)

## What I'm doing:
This pull request introduces significant enhancements to the primary key (PK) index compaction and parallelism mechanisms in shared-data mode, focusing on configurability, scalability, and maintainability. The changes include new configuration options for PK index parallel compaction, the addition of thread pools for parallel operations, refactoring of the PK index compaction logic to support multiple output files, and integration of a manager for parallel compaction tasks.

**Configuration and Parallelism Enhancements:**

* Added multiple new configuration options in `config.h` for fine-tuning PK index parallel compaction, file sizes, thread pool limits, and size-tiered compaction strategies. This enables more granular control over PK index performance and resource usage.

[...]

PR #69865, opened 2026-03-05
[BugFix] Fix SST file deleted by vacuum due to parallel compaction full-contain optimization

luohaha:
## Why I'm doing:

When parallel compaction's "full contain" optimization reuses input SST files as output (only changing `fileset_id`), `remove_compacted_sst()` blindly adds **all** input SSTs to `orphan_files`. This causes the same file to appear in both `sstable_meta` (as output) and `orphan_files` (as input). When vacuum runs `collect_garbage_files`, it deletes files in `orphan_files` without checking `sstable_meta`, resulting in:

```
load primary index failed: Not found: starlet err Object ... .sst does not exist
```

The "full contain" optimization was introduced in PR #66522 as part of the large tablet support feature (#66457), so this issue only affects 4.1 and does not exist in 4.0.

### Root Cause

In `LakePersistentIndexParallelCompactTask::do_run()` (`lake_persistent_index_parallel_compact_mgr.cpp:164-179`), when a single fileset's SSTs are fully contained within the seek range, the input SST PBs are directly copied as output (only `fileset_id` is changed):

```cpp
if (full_contain) {
    for (const auto& sst : sstables) {
        _output_sstables.push_back(sst->sstable_pb());
        _output_sstables.back().mutable_fileset_id()->CopyFrom(_output_fileset_id.to_proto());
    }
    return Status::OK();
}
```

Later in `MetaFileBuilder::remove_compacted_sst()`, all input SSTs are added to `orphan_files` for garbage collection — including the ones that were reused as output. Vacuum then deletes these files, breaking the persistent index.

## What I'm doing:

In `remove_compacted_sst()`, collect output SST filenames first, and skip any input SST whose filename also appears in the output. This prevents reused SST files from being added to `orphan_files`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport PR

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/meta_file.cpp
+++ b/be/src/storage/lake/meta_file.cpp
@@ -320,8 +320,22 @@ void trim_partial_compaction_last_input_rowset(const MutableTabletMetadataPtr& m
 }
 
 void MetaFileBuilder::remove_compacted_sst(const TxnLogPB_OpCompaction& op_compaction) {
-    // remove compacted sst
+    // Collect output SST filenames. Parallel compaction's "full contain" optimization
+    // may reuse input SST files as output (only changing fileset_id). Skip those files
+    // to avoid the same file appearing in both sstable_meta and orphan_files, which
+    // would cause vacuum to delete a still-referenced SST file.
+    std::unordered_set<std::string> output_sst_filenames;
+    if (op_compaction.has_output_sstable()) {
+        output_sst_filenames.insert(op_compaction.output_sstable().filename());
+    }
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

sevev (approved):

PR #69949, opened 2026-03-06
[Enhancement] introduce a new column buffer

silverbullet233:
## Why I'm doing:

part work of #69128

  The current Column implementations  use `std::vector` with `ColumnAllocator` as their underlying storage. While
  `std::vector` is a well-tested general-purpose container, it has inherent limitations in a columnar database context:

  1. **Allocator-bound lifecycle for precise memory statistics**: Per the plan in #69128, we want finer-grained memory accounting (e.g. Query → Fragment → Operator). Column Buffer memory must be bound to an explicit `Allocator` so that every allocation can be attributed to the right context. Although `std::vector` can take a stateful allocator, implementing more complex semantics (e.g. **transferring memory ownership** across operators or into cache) is difficult with its current API.
  2. **No `realloc` support**: `std::vector` always follows an alloc-copy-free pattern when growing, even when the allocator (jemalloc) could extend the allocation
  in-place via `xallocx`/`realloc`. For large column buffers (common in analytical workloads), this means unnecessary memcpy of megabytes of data.
  3. **Unnecessary zero-initialization**: `std::vector::resize(n)` value-initializes new elements (zeroing for POD types), which is wasteful when the caller will
  immediately overwrite them (e.g., `resize` followed by bulk write).
  4. **No tail padding support**: SIMD code paths (e.g. AVX2/AVX-512) require the data region to be padded so that the allocated size aligns with the SIMD register width (e.g. 32 or 64 bytes). This avoids unaligned loads/stores and over-read at the end of the buffer. `std::vector` has no built-in mechanism for such tail padding.

  For **3** and **4**, callers can today work around the limitations using interfaces in `raw_container`, but they must be careful (e.g. avoid default `resize` semantics or manually manage padding); otherwise they pay unnecessary overhead. **2** (no `realloc` support) cannot be addressed with `std::vector` at all.

  `Buffer` is designed as a drop-in replacement that addresses these issues while maintaining a familiar vector-like API.

 ## What I'm doing:

  Introduce a two-layer column buffer implementation:

  ### Architecture

  - **`RawBuffer<T, padding>`**: Low-level, allocator-explicit storage. All mutating operations take `memory::Allocator*` explicitly. The destructor does NOT free memory —
   callers must call `release(allocator)`. This enables flexible ownership transfer and external memory management.
  - **`Buffer<T, padding>`**: RAII wrapper that binds a `memory::Allocator*` at construction time and forwards all operations. Frees memory in the destructor. This is the
  primary API for Column implementations.

  ### Key semantic differences from `std::vector`

  The design philosophy of `Buffer` is **pay only for what you use**: no implicit zeroing, no hidden allocator, and growth uses `realloc` when the allocator supports it.

  | Behavior | `std::vector` | `Buffer` |
  |----------|--------------|----------|
  | `resize(n)` (grow, POD types) | Value-initializes (zeros) new elements | **No initialization** — new elements have indeterminate values; call `resize(n, value)` when initialization is required |
  | Growth reallocation | Always alloc + memcpy + free | Uses `realloc` when appropriate to avoid extra memory copy |
  | Tail padding | Not supported | Template parameter `padding` reserves extra bytes after the data region |
  | Copy semantics | Copyable | Move-only (copy disabled) |
  | Self-referencing safety | Undefined behavior | Debug-mode assertions (`check_not_intersects`, `check_range_not_intersects`) detect aliased references |
  | Insert / append | `insert` at arbitrary position | **Append-only** — no `insert`; suited to columnar append workloads |

  ### Performance: leveraging `realloc` / `xallocx`

  The key performance advantage comes from how `Buffer` handles growth for relocatable types (trivially copyable, or types tagged with `is_relocatable`):

  - **Small allocations (< 4096 bytes)**: Uses alloc + `memcpy_inlined` + free, same as `std::vector`. At this size the copy cost is negligible, and `memcpy_inlined` can
  be fully inlined by the compiler.
  - **Large allocations (>= 4096 bytes)**: Calls `allocator->realloc()`, which maps to jemalloc's `xallocx` + fallback. When jemalloc can extend the allocation in-place
  (common for large extents), **zero bytes are copied** — the buffer simply grows. This is where the biggest wins occur.

  Benchmark highlights (int64_t, single machine). 
For detailed test data and setup, see the end of `be/src/bench/buffer_bench.cpp`.

  | Operation | Size | Vector | Buffer | Speedup |
  |-----------|------|--------|--------|---------|
  | `single_grow` (reserve 2n) | 64K | 18,661 ns | 810 ns | **23x** |
  | `resize_value_grow` | 4M | 15,646 ns | 7,045 ns | **2.2x** |
  | `push_back_no_reserve` | 64M | 567,914 ns | 389,144 ns | **1.5x** |
  | `push_back_no_reserve` | 128M | 1,127,921 ns | 960,280 ns | **1.2x** |

  **`resize_value_grow`** starts from size `n/2` (e.g. 2M elements) and grows to `n` (e.g. 4M) via `resize(n, value)`, filling the new elements with the given value. Buffer is faster because it can use `realloc` when growing and only fills when the caller explicitly requests it, whereas `std::vector::resize(n, value)` follows the same alloc-copy-free growth path.

  **`single_grow`** starts with a buffer of `n` elements (e.g. 64K) and then calls `reserve(2n)` once, so it measures the cost of a single reallocation that doubles capacity. At 64K elements (512KB), jemalloc's `xallocx` extends the allocation in-place, giving Buffer a **23x** speedup over the vector's alloc-copy-free path.

  For operations where growth doesn't dominate (e.g., `reserve` + sequential `push_back`, `assign`, `append` with pre-reserved capacity), Buffer and vector perform
  equivalently, confirming no regression from the abstraction.

  ### Other details

  - **Relocatable type trait**: `is_relocatable<T>` defaults to `std::is_trivially_copyable<T>` but can be specialized via a `T::is_relocatable` tag for types that are
  safe to `memcpy`/`realloc` but not trivially copyable.
  - **Overflow checking**: `element_bytes()` and `element_bytes_with_padding()` use `__builtin_mul_overflow` / `__builtin_add_overflow` to detect size computation
  overflows.
  - **Exception safety**: For non-relocatable types, growth uses `uninitialized_move` (or `uninitialized_copy` if move can throw) with cleanup on exception.

  ### Future work

  - Gradually refactor all interfaces that create or own Columns so that they can explicitly pass an `Allocator` into the underlying Buffer.
  - After that migration is complete, rename the current `ColumnBuffer` type alias to `Buffer` so that Column implementations use the new buffer by default.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


be/src/column/raw_buffer.h
```diff
@@ -0,0 +1,561 @@
...
+#include <memory>
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+
+#include "common/logging.h"
+#include "fmt/format.h"
+#include "gutil/strings/fastmem.h"
+#include "runtime/memory/memory_allocator.h"
+#include "util/stack_util.h"
```
alvin-phoenix-ai: These header includes will break the modular boundary of ColumnCore, you can check Agnets.md.

be/src/column/raw_buffer.h
```diff
@@ -0,0 +1,561 @@
...
+        std::uninitialized_fill_n(begin(), count, value);
+    }
+    _end = _start + count * kElementSize;
+}
+
+template <class T, size_t padding>
+template <class InputIt, std::enable_if_t<std::is_base_of_v<std::forward_iterator_tag,
+                                                            typename std::iterator_traits<InputIt>::iterator_category>,
+                                          bool>>
+void RawBuffer<T, padding>::assign(memory::Allocator* allocator, InputIt first, InputIt last) {
```
alvin-phoenix-ai, line 429: Having this template function in header files will slow down the compilation speed. Please try to move it into cpp.

alvin-phoenix-ai (dismissed):

be/src/column/raw_buffer.h
silverbullet233: > Having this template function in header files will slow down the compilation speed. Please try to move it into cpp.

Thanks for the suggestion. However, since RawBuffer is a class template, moving implementations to a .cpp file would require
  explicit instantiation for every <T, padding> combination used across the codebase. More importantly, it would prevent the
  compiler from inlining at call sites in other translation units (without LTO), which is a significant concern for a
  performance-critical, low-level buffer class.

  Hot-path methods like push_back, emplace_back, operator[], size(), and begin()/end() rely heavily on inlining for performance —
  this is the same reason std::vector keeps all its implementations in the header.

be/src/column/raw_buffer.h
silverbullet233: fixed

silverbullet233:
@codex review

satanson (approved):

trueeyu (approved):

dirtysalt (approved):

be/src/column/raw_buffer.h
```diff
@@ -0,0 +1,600 @@
...
+#include <cstring>
+#include <initializer_list>
+#include <iterator>
+#include <memory>
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+
+#include "base/memory/memory_allocator.h"
+#include "common/logging.h"
```
alvin-phoenix-ai, line 29: Use "base/logging.h" instead, because common/logging.h is a wrapper.

be/src/column/raw_buffer.h
```diff
@@ -0,0 +1,600 @@
...
+#include "fmt/format.h"
```
alvin-phoenix-ai, line 30:
```suggestion
#include <fmt/format.h>
```

be/src/column/raw_buffer.h
```diff
@@ -0,0 +1,600 @@
...
+}
+
+template <class T, size_t padding>
+void RawBuffer<T, padding>::push_back(memory::Allocator* allocator, const T& value) {
+    check_not_intersects(&value);
+    if (UNLIKELY(_end == _end_of_storage)) {
+        grow(allocator);
+    }
+    if constexpr (std::is_trivially_copyable_v<T>) {
+        *reinterpret_cast<T*>(_end) = value;
```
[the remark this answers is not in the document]
silverbullet233: this is by design, DEBUG mode will check it but release mode won't because of performance

be/src/column/raw_buffer.h
```diff
@@ -0,0 +1,600 @@
...
+void RawBuffer<T, padding>::grow(memory::Allocator* allocator) {
+    if (UNLIKELY(_end_of_storage == _start)) {
+        // First allocation: call allocator directly (avoids relocate function call overhead).
+        size_t new_bytes = required_bytes(1);
+        _start = reinterpret_cast<uint8_t*>(allocator->alloc(new_bytes));
+        DCHECK(_start != nullptr);
+        _end = _start;
+        _end_of_storage = _start + new_bytes - kPadding;
+    } else {
+        relocate(allocator, allocated_bytes() * 2);
```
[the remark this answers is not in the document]
silverbullet233: The main goal is to improve performance and simplify calculations as much as possible during expansion. Adding a little extra memory here won't have a significant impact.

PR #69952, opened 2026-03-06
[Enhancement] Improve lake vacuum partial-failure recovery in version-chain mode

sevev:
## Why I'm doing:
Track per-version cumulative data-file counts so that when a batch delete fails mid-way, we can determine exactly which versions had all their data files confirmed deleted and safely delete only those versions' metadata files, advancing min_version accordingly.

## What I'm doing:
Recycle garbage files by version. For each tablet version, submit a task to delete the meta file immediately after deleting the dat files. Do not delete all dat files for the tablet first and then delete all meta files in bulk.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1660,7 +1660,16 @@ CONF_mBool(enable_drop_tablet_if_unfinished_txn, "true");
 // 0 means no limit
 CONF_Int32(lake_service_max_concurrency, "0");
 
+// Minimum batch size for lake vacuum delete-file tasks.
 CONF_mInt64(lake_vacuum_min_batch_delete_size, "100");
+
+// Enable the lake vacuum implementation that walks tablet version chains
+// and deletes data files + metadata in a safer, failure-tolerant way.
+CONF_mBool(lake_vacuum_enable_version_chain_mode, "false");
+
+// Per-tablet time budget (in ms) for the version-chain-based lake vacuum path.
+// 0 means no timeout.
+CONF_mInt64(lake_vacuum_version_chain_timeout_ms, "600000");
 
 // TOPN RuntimeFilter parameters
 CONF_mInt32(desc_hint_split_range, "10");
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1667,11 +1667,6 @@ CONF_mInt64(lake_vacuum_min_batch_delete_size, "100");
 // and deletes data files + metadata in a safer, failure-tolerant way.
 CONF_mBool(lake_vacuum_enable_version_chain_mode, "false");
 
-// Per-tablet time budget (in ms) for the version-chain-based lake vacuum path.
-// 0 means no timeout.
-// default: 50 minutes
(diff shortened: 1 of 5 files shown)


be/src/storage/lake/vacuum.cpp
```diff
@@ -483,23 +663,33 @@ static Status vacuum_tablet_metadata(TabletManager* tablet_mgr, std::string_view
...
-                                                vacuum_version_range.get(), &datafile_deleter, &metafile_deleter,
-                                                &shared_file_deleter, vacuumed_file_size, &tablet_vacuumed_version,
-                                                extra_file_size, tablet_retain_info));
-        RETURN_IF_ERROR(datafile_deleter.finish());
-        (*vacuumed_files) += datafile_deleter.delete_count();
-        if (!enable_file_bundling) {
-            RETURN_IF_ERROR(metafile_deleter.finish());
-            (*vacuumed_files) += metafile_deleter.delete_count();
+
+        if (config::lake_vacuum_enable_version_chain_mode && !enable_file_bundling) {
```
xiangguangyxg, line 673: why can only enabled when not enable file bunding?

be/src/common/config.h
```diff
@@ -1660,8 +1660,13 @@ CONF_mBool(enable_drop_tablet_if_unfinished_txn, "true");
 // 0 means no limit
 CONF_Int32(lake_service_max_concurrency, "0");
 
+// Minimum batch size for lake vacuum delete-file tasks.
 CONF_mInt64(lake_vacuum_min_batch_delete_size, "100");
 
+// Enable the lake vacuum implementation that walks tablet version chains
+// and deletes data files + metadata in a safer, failure-tolerant way.
+CONF_mBool(lake_vacuum_enable_version_chain_mode, "false");
```
xiangguangyxg, line 1668: why default not enabled

be/src/storage/lake/vacuum.cpp
sevev: > why can only enabled when not enable file bunding?

In file_bundling mode, it is necessary to first check all tablets under the partition and finally delete the meta. Since the current implementation processes each tablet individually, it does not support execution in file_bundling mode.

be/src/common/config.h
sevev: > why default not enabled

When the deletion volume is low, most vacuum deletion requests can succeed. The vacuum method based on the tablet chain requires traversing the tablet metadata twice and is more suitable for scenarios with extremely high deletion QPS.

be/src/storage/lake/vacuum.cpp
xiangguangyxg: So the optimization is limited since file_bunding is enabled by default in 4.0

PR #70004, opened 2026-03-09
[Enhancement] Improve sql_mode to handle Division by zero and Fail to parse date for str_to_date/str2date

satanson:
## Why I'm doing:

ERROR_FOR_DIVISION_BY_ZERO and ALLOW_THROW_EXCEPTION are expected to raise user-facing errors consistently across FE constant-folding and BE runtime evaluation. Previously, some cases were silently folded to NULL or executed in BE due to constant-folding exception swallowing and JIT paths that didn’t enforce division-by-zero checks. Also, NULL/0 behavior was inconsistent between FE and BE.

## What I'm doing:
NOT_NULL_VALUE/0 report "Division by zero" error if sql_mode ERROR_FOR_DIVISION_BY_ZERO is on, NULL/0 still returns null;
str2date/str_to_date("invalid_date_str",  str_format) report "Fail to parse date" error if sql_mode ALLOW_THROW_EXCEPTION is on, str2date/str_to_date(NULL,  str_format) stills return NULL.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/arithmetic_expr.cpp
+++ b/be/src/exprs/arithmetic_expr.cpp
@@ -16,6 +16,9 @@
 
 #include <optional>
 
+#include "column/column_helper.h"
+#include "column/const_column.h"
+#include "column/nullable_column.h"
 #include "column/type_traits.h"
 #include "common/object_pool.h"
 #include "common/status.h"
@@ -26,7 +29,9 @@
 #include "exprs/decimal_cast_expr.h"
 #include "exprs/overflow.h"
 #include "exprs/unary_function.h"
+#include "gutil/casts.h"
 #include "runtime/runtime_state.h"
+#include "types/decimalv2_value.h"
 #include "types/logical_type.h"
 
 #ifdef STARROCKS_JIT_ENABLE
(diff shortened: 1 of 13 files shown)


be/src/exprs/arithmetic_expr.cpp
```diff
@@ -45,6 +50,39 @@ namespace starrocks {
...
+        }
+        // Skip rows where dividend is NULL (NULL / 0 -> NULL, no error).
+        if (dividend->is_null(i)) {
+            continue;
+        }
+        // For const columns, logical row i maps to physical row 0.
+        const size_t data_row = divisor->is_constant() ? 0 : i;
+        if (divisor_values[data_row] == zero) {
+            return true;
+        }
```
stdpain, line 81: how about using column viewer? It will be more clearly

stdpain (approved):

be/src/exprs/arithmetic_expr.cpp
satanson: I will fix in next pr.

stephen-shelby (approved):

be/src/exprs/arithmetic_expr.cpp
```diff
@@ -45,6 +50,39 @@ namespace starrocks {
...
+    const Column* divisor_data_col = ColumnHelper::get_data_column(divisor.get());
+    const auto* divisor_data = ColumnHelper::template cast_to_raw<Type>(divisor_data_col);
+    const auto* divisor_values = divisor_data->get_data().data();
+
+    const auto num_rows = divisor->size();
+
+    constexpr auto zero = RunTimeCppType<Type>{};
+    for (size_t i = 0; i < num_rows; i++) {
+        // Skip rows where divisor is NULL.
+        if (divisor->is_null(i)) {
```
ZiheLiu, line 70: Do we need a vectorized optimization here? Otherwise, division could be slower. In other words, should we extract the data column and the null column, and then handle the check in a branch-free `for` loop?

gengjun-git (approved):

ZiheLiu (approved):

wangruin (approved):

be/src/exprs/arithmetic_expr.cpp
satanson: good suggest, we can handle it in the next pr.

Issue #70024, opened 2026-03-09
Starrocks connect S3 with vip always connect the same S3 backend IP


GuoPhilipse:
### Steps to reproduce the behavior (Required)
When starrocks connect  S3 by the VIP connection , starrocks will always connect  permanent S3 backend IPs, it will give S3 cluster pressure, when we connect S3 with DNS, it behaviors normal ,  starrocks connects with multiple backend S3 IPs

### Expected behavior (Required)
we can make vip and dns connection behavior the same 
### Real behavior (Required)
starrocks will always connect  permanent S3 backend IPs

### StarRocks version (Required)
-3.5

PR #70025, opened 2026-03-09
[BugFix] Starrocks connect S3 with vip always connect the same S3 backend IP

GuoPhilipse:
## Why I'm doing:

When starrocks connect  S3 by the VIP connection , starrocks will always connect  permanent S3 backend IPs, it will give S3 cluster pressure, when we connect S3 with DNS, it behaviors normal ,  starrocks connects with multiple backend S3 IPs, we can make vip and dns connection behavior the same 

## What I'm doing:
fix code 

Fixes #issue
#70024
## What type of PR is this:

- [x] BugFix

## Checklist:

- [x] This pr needs auto generate documentation
- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/fs/s3/poco_common.cpp
+++ b/be/src/fs/s3/poco_common.cpp
@@ -23,6 +23,7 @@
 #include <sstream>
 #include <unordered_map>
 
+#include "base/network/network_util.h"
 #include "runtime/current_thread.h"
 
 namespace starrocks::poco {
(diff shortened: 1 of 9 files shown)


be/src/fs/s3/poco_common.cpp
```diff
@@ -101,21 +104,69 @@ HTTPSessionPools& HTTPSessionPools::instance() {
     return instance;
 }
 
+// Resolve hostname to IP with TTL cache to avoid blocking DNS on every getSession() and to limit
+// _endpoint_pools growth when DNS results change. Caller must not hold _mutex.
+std::string HTTPSessionPools::resolveHostWithCache(const std::string& host) {
```
kevincai, line 109: I still don't get it, how this DNS resolve can help with the case that the host is a VIP? VIP is not a DNS name, right?

GuoPhilipse: I think DNS will resolve the backend ip everytime, while if we configure the vip, the connection will cached the resolved ip,the following traffic will always find the resolved ip,it will case hot point.

kevincai: VIP is an IP address, how can it solves to different IPs by DNS?

PR #70029, opened 2026-03-09
[Enhancement] Force Drop Sync MV recovery feature added

srihithg:
When base table is not in NORMAL state (e.g. stuck sync MV build), DROP MATERIALIZED VIEW ... FORCE cancels related alter jobs, sets table to NORMAL, then drops the MV. Async MVs unchanged.

## Why I'm doing:

Users cannot drop a synchronous materialized view when its base table is not in NORMAL state (e.g. a sync MV build job is stuck). The only workaround today is manual intervention. This adds a FORCE option so they can cancel the stuck jobs, restore the table to NORMAL, and drop the MV in one step.

## What I'm doing:

- Add optional `FORCE` to `DROP MATERIALIZED VIEW` syntax (grammar, parser, AST).
- When FORCE is used and the base table is not NORMAL: cancel related RollupJobV2/LakeRollupJob for that MV, set table state to NORMAL, then perform the drop.
- When FORCE is used, skip strict table-state and partition checks in the handler and record force-drop in the edit log (DropInfo).
- Skip MV dependency check in the analyzer when FORCE is specified.
- Async MVs are unchanged (still dropped via existing path with no FORCE handling).
- Document FORCE in DROP_MATERIALIZED_VIEW (en/zh/ja) and add parsing tests.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobMgr.java
@@ -188,7 +188,11 @@ public class AlterJobMgr {
                     boolean foundMvRollupJob = materializedViewHandler.cancelRollupJobsForForceDrop(
                             targetTable.getId(), stmt.getMvName(), "force drop materialized view");
                     if (!foundMvRollupJob) {
-                        throw InvalidOlapTableStateException.of(targetTable.getState(), targetTable.getName());
+                        // Re-check state: it may have transitioned to NORMAL by another thread
+                        // (race with rollup job completion) before we could cancel.
+                        if (targetTable.getState() != OlapTableState.NORMAL) {
+                            throw InvalidOlapTableStateException.of(targetTable.getState(), targetTable.getName());
+                        }
                     }
                 } else {
                     throw InvalidOlapTableStateException.of(targetTable.getState(), targetTable.getName());
--- a/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobMgr.java
@@ -185,21 +185,11 @@ public class AlterJobMgr {
             // check table state
             if (targetTable.getState() != OlapTableState.NORMAL) {
                 if (stmt.isForceDrop()) {
-                    List<AlterJobV2> jobs =
-                            materializedViewHandler.getUnfinishedAlterJobV2ByTableId(targetTable.getId());
-                    String mvName = stmt.getMvName();
-                    for (AlterJobV2 job : jobs) {
-                        if (job instanceof RollupJobV2
-                                && ((RollupJobV2) job).getRollupIndexName().equals(mvName)) {
-                            job.cancel("force drop materialized view");
-                        } else if (job instanceof LakeRollupJob
-                                && ((LakeRollupJob) job).getRollupIndexName().equals(mvName)) {
-                            job.cancel("force drop materialized view");
-                        }
+                    boolean foundMvRollupJob = materializedViewHandler.cancelRollupJobsForForceDrop(
+                            targetTable.getId(), stmt.getMvName(), "force drop materialized view");
+                    if (!foundMvRollupJob) {
+                        throw InvalidOlapTableStateException.of(targetTable.getState(), targetTable.getName());
                     }
-                    targetTable.setState(OlapTableState.NORMAL);
-                    LOG.info("Force drop MV: cancelled stuck jobs and set table {} to NORMAL for MV {}",
-                            targetTable.getName(), mvName);
                 } else {
                     throw InvalidOlapTableStateException.of(targetTable.getState(), targetTable.getName());
                 }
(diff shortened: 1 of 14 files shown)


fe/fe-core/src/test/java/com/starrocks/planner/MaterializedViewManualTest.java
```diff
@@ -357,35 +357,14 @@ public void testDateTruncPartitionColumnExpr3() throws Exception {
...
-        {
-            sql("SELECT \n" +
-                    "count(DISTINCT `order_id`) AS `order_num`, \n" +
-                    "date_trunc('day', `dt`) AS ds \n" +
-                    "FROM `test_partition_expr_tbl1`\n" +
-                    "WHERE `dt` BETWEEN '2023-04-11' AND '2023-04-12'\n" +
-                    "group by ds")
-                    .match("test_partition_expr_mv1");
-        }
+        // Query uses date_trunc('minute') so it must not match MV defined with date_trunc('hour').
```
kevincai: reason for touching this case?

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/MaterializedViewAnalyzer.java
```diff
@@ -1678,8 +1678,8 @@ public Void visitDropMaterializedViewStatement(DropMaterializedViewStmt stmt, Co
             stmt.setTableRef(tableRef);
             Table mvTable = GlobalStateMgr.getCurrentState().getMetadataMgr().getTable(context, stmt.getCatalogName(),
                     stmt.getDbName(), stmt.getMvName());
-            // Check mv dependency
-            if (context.getSessionVariable().isEnableDropTableCheckMvDependency() && mvTable != null) {
+            // Check mv dependency (skip when FORCE drop - user explicitly bypasses safety checks)
+            if (context.getSessionVariable().isEnableDropTableCheckMvDependency() && mvTable != null && !stmt.isForceDrop()) {
```
kevincai: is this the right place to handle the error we met? the mv can be dropped without FORCE by tuning the session variable.

fe/fe-core/src/test/java/com/starrocks/planner/MaterializedViewManualTest.java
srihithg: I think I simplified the original case since it was failing due to a bug and I made AI change and simplify the test case. 

Just now, after I reverted this change the test still passes meaning the bug is no longer there, I will be making another commit with the new change.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/MaterializedViewAnalyzer.java
srihithg: I see, I will remove this change and revert it to the original. 
To make sure I understand this more is it because bypassing the dependency check is already possible via the session variable. If users set `enable_drop_table_check_mv_dependency=false`, users can drop an MV even when others depend on it. So if we use FORCE to skip the dependency check that usually checks this, would it duplicate it and mix two separate paths which are to 
- force drop a stuck sync MV
- ignore MV dependencies

So, are we keeping FORCE only for the stuck sync MV case and leaving the dependency bypass to the session variable?

kevincai: keep `enable_drop_table_check_mv_dependency` session variable as is. `FORCE` will deal with the case that even `enable_drop_table_check_mv_dependency=false` the MV still can't be dropped.

kevincai: The behavior can be tuned later if `enable_drop_table_check_mv_dependency` can be completely replaced by `FORCE` option.

fe/fe-core/src/main/java/com/starrocks/alter/AlterJobMgr.java
```diff
@@ -181,10 +181,18 @@ public void processDropMaterializedView(DropMaterializedViewStmt stmt) throws Dd
                 throw new DdlException(
                         "Do not support non-OLAP table [" + table.getName() + "] when drop materialized view");
             }
-            // check table state
             targetTable = (OlapTable) table;
+            // check table state
             if (targetTable.getState() != OlapTableState.NORMAL) {
-                throw InvalidOlapTableStateException.of(targetTable.getState(), targetTable.getName());
+                if (stmt.isForceDrop()) {
+                    boolean foundMvRollupJob = materializedViewHandler.cancelRollupJobsForForceDrop(
```
kevincai, line 188: Since the check is under no lock. it is possible that in line 186, the tablet is still non NORMAL state, but when it runs to line 188, the job is done and the state may changed to NORMAL state. This is a race condition need to be considered.

kevincai (approved):

gengjun-git (approved):

stephen-shelby (approved):

PR #70049, opened 2026-03-10
[Enhancement] Avoid object storage LIST in random bucketing import path

### LINKED ISSUE #68783: Remove list object storage operation when loading data into random bucket tables
## Enhancement

In shared-data mode, when loading data into random bucketing tables, the BE would trigger an expensive list operation on object storage to get
tablet schema in certain fallback scenarios. 

We can fix  the issue by properly utilizing the tablet_meta parameter in `_fallback_load_to_schema_file()` to extract schema directly from the
provided metadata, avoiding the need to call `get_tablet()` which would trigger `list_tablet_metadata()` and `iterate_dir()` on object storage.

wxl24life:
## Why I'm doing:

When importing data into random bucketing tables in shared-data mode, add_in_writing_data_size() was calling get_tablet_data_size(tablet_id, nullptr), which triggers list_tablet_metadata() and thus an expensive object storage LIST operation for every tablet on first access.

## What I'm doing:

Fixes #68783
Replace the LIST-based fallback with a cache lookup via get_latest_cached_tablet_metadata():
* If cached metadata exists → compute the base data size from it
* If no cache hit → use 0 as the base size

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/tablet_manager.cpp
+++ b/be/src/storage/lake/tablet_manager.cpp
@@ -1252,7 +1252,18 @@ int64_t TabletManager::add_in_writing_data_size(int64_t tablet_id, int64_t size)
         }
     }
 
-    int64_t base_size = get_tablet_data_size(tablet_id, nullptr).value_or(0);
+    // Immutable check for random bucketing is best-effort.
+    // On cache hit, compute base_size directly from the cached metadata to avoid any remote I/O.
+    // On cache miss, whether to fallback LIST metadata files is controlled by
+    // `allow_list_object_for_random_bucketing_on_cache_miss`.
+    int64_t base_size = 0;
+    if (auto metadata = get_latest_cached_tablet_metadata(tablet_id); metadata != nullptr) {
+        for (const auto& rowset : metadata->rowsets()) {
+            base_size += rowset.data_size();
+        }
+    } else if (config::allow_list_object_for_random_bucketing_on_cache_miss) {
+        base_size = get_tablet_data_size(tablet_id, nullptr).value_or(0);
+    }
 
     std::unique_lock wrlock(_meta_lock);
     const auto& it = _tablet_in_writing_size.find(tablet_id);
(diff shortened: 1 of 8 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -1252,7 +1252,19 @@ int64_t TabletManager::add_in_writing_data_size(int64_t tablet_id, int64_t size)
...
+    // `lake_in_writing_data_size_fallback_to_list_on_cache_miss`.
+    int64_t base_size = 0;
+    int64_t version_hint = 0;
+    if (auto metadata = get_latest_cached_tablet_metadata(tablet_id); metadata != nullptr) {
+        version_hint = metadata->version();
+    }
+    if (version_hint > 0) {
+        base_size = get_tablet_data_size(tablet_id, &version_hint).value_or(0);
+    } else if (config::lake_in_writing_data_size_fallback_to_list_on_cache_miss) {
+        base_size = get_tablet_data_size(tablet_id, nullptr).value_or(0);
```
kevincai, line 1265: this configuration exists for saving list object api with the cost of inaccurate size estimation of the tablet. shall give the config item a better name.

random_bucketing_list_optimization_proposal.md
kevincai, line 1: This doc should be moved to the issue description.

be/src/storage/lake/tablet_manager.cpp
wxl24life: have replaced with `allow_list_object_for_random_bucketing_on_cache_miss`

kevincai (approved):

meegoo (approved):

wyb:
4.1 https://github.com/StarRocks/starrocks/pull/70157

PR #70051, opened 2026-03-10
[Enhancement] Batch tablet deletion to reduce write lock contention

luohaha:
## Why I'm Doing This

When deleting multiple tablets (e.g., during table truncation, partition erasure, or cleanup of failed operations), the current code acquires and releases the `TabletInvertedIndex` write lock once per tablet in a tight loop. This creates a **convoy effect**: readers continuously find a writer queued and can never proceed, causing sustained read blockade even though each individual write lock hold is short. This impacts any concurrent read operations on the tablet inverted index during bulk tablet deletions.

## What I'm Doing

- Added batch `deleteTablets(Collection<Long>)` method to `TabletInvertedIndex` that acquires the write lock **once** for an entire batch of tablet deletions
- Added `markTabletsForceDelete(Collection<Long>)` for batch force-delete marking
- Extracted `deleteTabletUnlocked()` private method to share logic between single and batch delete paths
- Updated **all 8 loop-based call sites** in `LocalMetastore` to use the batch API:
  - `cleanExistPartitionNameSet()` - partition cleanup on add
  - `cleanTabletIdSetForAll()` - general tablet cleanup
  - `deleteUselessTablets()` - failed partition creation cleanup
  - `truncateTableInternal()` - table truncation
  - `onEraseTable()` - table drop
  - `onErasePartition()` - partition drop
  - `createTable()` error path - failed table creation cleanup
  - `createMaterializedView()` error path - failed MV creation cleanup
  - `createTempPartitionsFromPartitions()` error path

**Technical Details:**
- Lock ordering remains unchanged: `DB/Table WriteLock → TabletInvertedIndex writeLock`, no deadlock risk
- Single-tablet `deleteTablet()` API is preserved for callers that only delete one tablet
- Total lock hold time is similar, but **one contiguous hold** eliminates the convoy effect and gives readers a clean window to proceed after release

**Risks:**
- The write lock is held slightly longer in a single stretch, but this is a net win because it eliminates the repeated lock/unlock overhead and the reader starvation caused by always having a writer queued

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
@@ -61,6 +61,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


luohaha:
Recreating PR based on new-main2 branch with correct template.

PR #70053, opened 2026-03-10
[Enhancement] Validate storage volume accessibility in FE DDL

### LINKED ISSUE #68791: Validating storage volume accessibility during CREATE/ALTER operations
## Feature request

When creating or modifying a Storage Volume with incorrect configuration (e.g., wrong credentials), the error is only discovered during table creation when the system actually tries to access the storage. This leads to a poor user experience.

Design a solution for validating storage volume accessibility during CREATE/ALTER operations.

Add an optional validation step during CREATE/ALTER STORAGE VOLUME DDL operations that:
* Attempts to access the target storage location to verify connectivity and permissions
* This validation is controlled by a session variable to allow users to skip it if needed
* The validation must be efficient as SV DDL operations are synchronous

Key Design Decisions
1. Validation Method: 
    * Use a lightweight check like listing the bucket/path to validate access
    * This validates both connectivity and read permissions
2. Configuration:
    * Add a session variable validate_storage_volume_on_ddl (default: true)
    * Add a config storage_volume_validation_timeout_ms to control timeout
3. Property Control:
    * Users can override via a DDL property "skip_validation" = "true" in the PROPERTIES clause
4. Implementation Location:
    * Add validation logic in StorageVolumeMgr.createStorageVolume() and updateStorageVolume()
    * Create a new utility class StorageVolumeValidator to handle the validation logic

**Additional context**

wxl24life:
## What I'm doing:

Fixes #68791 

Add FE-side accessibility checks for CREATE/ALTER STORAGE VOLUME in shared-data mode to fail early before metadata persistence when location credentials or endpoints are invalid.

E2E Results: Storage Volume Accessibility Precheck

    ┌──────┬─────────────────────────┬─────────────────────────────┬──────────────────────────────────────────────────┬────────┐
    │ Test │ Scenario                │ Expected                    │ Actual                                           │ Status │
    ├──────┼─────────────────────────┼─────────────────────────────┼──────────────────────────────────────────────────┼────────┤
    │ T1   │ Valid credentials       │ Success                     │ ✅ Query OK                                      │ PASS   │
    │      │ CREATE                  │                             │                                                  │        │
    ├──────┼─────────────────────────┼─────────────────────────────┼──────────────────────────────────────────────────┼────────┤
    │ T2   │ Invalid AK/SK CREATE    │ Fail with "accessibility    │ ✅ Storage volume accessibility check failed...  │ PASS   │
    │      │                         │ check failed"               │ Forbidden (Status Code: 403)                     │        │
    ├──────┼─────────────────────────┼─────────────────────────────┼──────────────────────────────────────────────────┼────────┤
    │ T3   │ flag=false, invalid     │ Success (check skipped)     │ ✅ Query OK                                      │ PASS   │
    │      │ creds CREATE            │                             │                                                  │        │
    ├──────┼─────────────────────────┼─────────────────────────────┼──────────────────────────────────────────────────┼────────┤
    │ T4   │ Invalid creds ALTER     │ Fail, old config unchanged  │ ✅ ALTER failed with check error, SV properties  │ PASS   │
    │      │                         │ (rollback)                  │ preserved                                        │        │
    ├──────┼─────────────────────────┼─────────────────────────────┼──────────────────────────────────────────────────┼────────┤
    │ T5   │ enabled-only ALTER      │ Success, no check triggered │ ✅ Query OK                                      │ PASS   │
    └──────┴─────────────────────────┴─────────────────────────────┴──────────────────────────────────────────────────┴────────┘

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/docs/en/faq/shared_data_faq.md
+++ b/docs/en/faq/shared_data_faq.md
@@ -12,6 +12,7 @@ Check the BE log (`be.INFO`) to identify the exact cause. Common causes include:
 
 - Misconfigured object storage settings (for example, `aws_s3_path`, `endpoint`, `authentication`).
 - Object storage service instability or exceptions.
+- For `CREATE STORAGE VOLUME` or `ALTER STORAGE VOLUME`, StarRocks validates storage accessibility in shared-data mode when the BE configuration `enable_storage_volume_access_check` is enabled. If this check is disabled, validation is skipped. If validation fails, fix credentials, endpoint, or network access first, and then retry.
 
 Other errors:
 
--- a/docs/en/faq/shared_data_faq.md
+++ b/docs/en/faq/shared_data_faq.md
@@ -12,7 +12,7 @@ Check the BE log (`be.INFO`) to identify the exact cause. Common causes include:
 
 - Misconfigured object storage settings (for example, `aws_s3_path`, `endpoint`, `authentication`).
 - Object storage service instability or exceptions.
-- For `CREATE STORAGE VOLUME` or `ALTER STORAGE VOLUME`, StarRocks validates storage accessibility in shared-data mode when the BE configuration `enable_storage_volume_access_check` is enabled. If this check is disabled, validation is skipped. If validation fails, fix credentials, endpoint, or network access first, and then retry.
+- For `CREATE STORAGE VOLUME` or `ALTER STORAGE VOLUME`, StarRocks validates storage accessibility in shared-data mode when the FE configuration `enable_storage_volume_access_check` is enabled (enabled by default). If this check is disabled, validation is skipped. If validation fails, fix credentials, endpoint, or network access first, and then retry.
 
 Other errors:
 
--- a/fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeAccessChecker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeAccessChecker.java
@@ -1,0 +1,47 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.server;
+
+import com.starrocks.common.DdlException;
+import com.starrocks.common.StarRocksException;
+import com.starrocks.fs.HdfsUtil;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class StorageVolumeAccessChecker {
+    private StorageVolumeAccessChecker() {
+    }
+
+    public static void check(String svName, String svType, List<String> locations, Map<String, String> params)
+            throws DdlException {
+        for (String location : locations) {
+            try {
+                // checkPathExist() will trigger a real access attempt to validate credential/network/path reachability.
+                HdfsUtil.checkPathExist(location, new HashMap<>(params));
+            } catch (StarRocksException e) {
+                Throwable cause = e;
+                while (cause.getCause() != null) {
+                    cause = cause.getCause();
+                }
+                String message = cause.getMessage();
+                throw new DdlException(String.format(
+                        "Storage volume accessibility check failed. storage volume: '%s', type: '%s', location: '%s', error: %s",
+                        svName, svType, location, message == null ? cause.toString() : message));
+            }
+        }
+    }
+}
--- a/fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeAccessChecker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeAccessChecker.java
@@ -14,6 +14,7 @@
 
 package com.starrocks.server;
 
+import com.google.common.base.Strings;
 import com.starrocks.common.DdlException;
 import com.starrocks.common.StarRocksException;
 import com.starrocks.fs.HdfsUtil;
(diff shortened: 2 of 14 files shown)


docs/en/faq/shared_data_faq.md
```diff
@@ -12,6 +12,7 @@ Check the BE log (`be.INFO`) to identify the exact cause. Common causes include:
 
 - Misconfigured object storage settings (for example, `aws_s3_path`, `endpoint`, `authentication`).
 - Object storage service instability or exceptions.
+- For `CREATE STORAGE VOLUME` or `ALTER STORAGE VOLUME`, StarRocks validates storage accessibility in shared-data mode when the BE configuration `enable_storage_volume_access_check` is enabled. If this check is disabled, validation is skipped. If validation fails, fix credentials, endpoint, or network access first, and then retry.
```
kevincai: it appears to be a FE configuration, not BE configuration

fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeMgr.java
```diff
@@ -126,6 +127,7 @@ public String createStorageVolume(String name, String svType, List<String> locat
             if (exists(name)) {
                 throw new AlreadyExistsException(String.format("Storage volume '%s' already exists", name));
             }
+            checkStorageVolumeAccessIfNeeded(name, svType, locations, params);
```
kevincai: move it outside the lock and do the check, and then lock again with existence check and do the creation. Don't try to do the check under write lock that may block entire system.

fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeAccessChecker.java
```diff
@@ -0,0 +1,47 @@
...
+public final class StorageVolumeAccessChecker {
+    private StorageVolumeAccessChecker() {
+    }
+
+    public static void check(String svName, String svType, List<String> locations, Map<String, String> params)
+            throws DdlException {
+        for (String location : locations) {
+            try {
+                // checkPathExist() will trigger a real access attempt to validate credential/network/path reachability.
+                HdfsUtil.checkPathExist(location, new HashMap<>(params));
```
kevincai: This result is not reliable, e.g. the path doesn't exist but the access creds are correct. It will be totally fine since the path can be created later when new file needs to be written to the location.

fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeAccessChecker.java
```diff
@@ -0,0 +1,62 @@
...
+        for (String location : locations) {
+            // Write an empty temp file and then delete it to verify that the credentials have
+            // actual write access to the location. checkPathExist() only verifies path existence,
+            // not write permission, and will silently succeed when the path does not yet exist.
+            String normalizedLoc = location.endsWith("/") ? location : location + "/";
+            String tempPath = normalizedLoc + ".starrocks_sv_access_check_" + UUID.randomUUID();
+            try {
+                HdfsUtil.writeFile(new byte[0], tempPath, new HashMap<>(params));
+            } catch (StarRocksException e) {
+                Throwable cause = e;
```
kevincai, line 51: It could be possible that the path doesn't exist, (in case of HDFS), direct writting a file whose parent directory doesn't exist, what's the expected result?

fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeMgr.java
```diff
@@ -180,40 +184,79 @@ public void updateStorageVolume(String name, String svType, List<String> locatio
                 throw new DdlException(String.format("Storage volume property '%s' is immutable!", param));
             }
         }
+
+        // Only check storage volume accessibility when connectivity-affecting properties are changed
+        // (type, locations, or cloud credentials). Skip the check for metadata-only changes
+        // (enabled, comment) so that operators can still disable a volume when storage is unreachable.
+        boolean connectivityChanged = !Strings.isNullOrEmpty(svType) || locations != null || !params.isEmpty();
```
kevincai: type and locations can't be updated once created.

fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeMgr.java
```diff
@@ -180,40 +184,79 @@ public void updateStorageVolume(String name, String svType, List<String> locatio
...
+            }
+        }
+
+        // Step 2: build a tentative copy and run the access check OUTSIDE any lock so that a slow
+        // or unreachable endpoint does not block concurrent storage-volume operations.
+        if (connectivityChanged) {
+            StorageVolume candidate = new StorageVolume(svSnapshot);
+            validateParams(candidate.getType(), params);
+            applyChangesToVolume(candidate, svSnapshot.getId(), svType, locations, comment, params, enabled);
+            checkStorageVolumeAccessIfNeeded(candidate.getName(), candidate.getType(),
```
kevincai, line 233: This is too late to check if the sv parameters shall be checked. Shall be fastpath as lightweight as possible if the check is not turned on.

fe/fe-core/src/main/java/com/starrocks/server/StorageVolumeAccessChecker.java
wxl24life: For HDFS, writeFile (via Hadoop's FileSystem.create()) automatically creates parent directories if they don't exist, so a non-existent path with valid credentials will pass the check successfully. This is the desired behavior — we're verifying that the credentials
have write access to the storage location, not that the path already exists. The path will be created on demand when data is actually written.

kevincai:
@mergifyio update

kevincai:
@mergifyio rebase main

kevincai (dismissed):

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

PR #70068, opened 2026-03-10
[BugFix] Fix str_to_date losing microsecond precision in BE runtime e…

earlbread:
## Why I'm doing:

`str_to_date` with `%f` silently drops microsecond precision when evaluated at BE runtime (e.g., Routine Load, Stream Load column mappings). The wrapper overload of `TimestampValue::from_uncommon_format_str` hardcodes microseconds to 0 when calling `timestamp::from_datetime`, discarding the value correctly parsed by the inner function into `content._microsecond`.

This only affects BE runtime evaluation paths. FE constant-folded paths (e.g., `INSERT INTO with literal values`) work correctly because they use a different code path.

## What I'm doing:

Pass `content._microsecond` instead of hardcoded 0 to `timestamp::from_datetime` in the wrapper overload of `TimestampValue::from_uncommon_format_str`.
Added regression tests at both layers:
  - `TimestampValueTest.from_uncommon_format_str_microsecond`: unit test for the parsing function directly
  - `TimeFunctionsTest.str_to_date_microsecond`: expression-layer test with const fmt + non-const input, matching the Routine Load/Stream Load column mapping scenario

Fixes #70056

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/types/timestamp_value.cpp
+++ b/be/src/types/timestamp_value.cpp
@@ -402,7 +402,7 @@ bool TimestampValue::from_uncommon_format_str(const char* format, int format_len
     bool result = from_uncommon_format_str(format, format_len, value, value_len, &content, nullptr);
     if (result) {
         _timestamp = timestamp::from_datetime(content._year, content._month, content._day, content._hour,
-                                              content._minute, content._second, 0);
+                                              content._minute, content._second, content._microsecond);
     }
     return result;
 }
--- a/be/test/exprs/time_functions_test.cpp
+++ b/be/test/exprs/time_functions_test.cpp
@@ -1591,6 +1591,37 @@ TEST_F(TimeFunctionsTest, str_to_date_of_datetimeformat) {
     }
 }
 
+TEST_F(TimeFunctionsTest, str_to_date_microsecond) {
+    FunctionContext* ctx = FunctionContext::create_test_context();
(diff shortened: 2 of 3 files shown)


kevincai (dismissed):

kevincai:
```
  [==========] Running 1 test from 1 test suite.
  [----------] Global test environment set-up.
  [----------] 1 test from TimeFunctionsTest
  [ RUN      ] TimeFunctionsTest.str_to_date_microsecond
  /root/starrocks/be/test/exprs/time_functions_test.cpp:1613: Failure
  Value of: result->is_nullable()
    Actual: false
  Expected: true
  [  FAILED  ] TimeFunctionsTest.str_to_date_microsecond (6 ms)
  [----------] 1 test from TimeFunctionsTest (6 ms total)
```

@earlbread UT fails.

Seaven (approved):

kevincai (approved):

trueeyu:
https://github.com/Mergifyio backport branch-3.5-cc

PR #70069, opened 2026-03-10
[Doc] Use correct name in the doc for the metric

behcetm:
## Why I'm doing:

There is a metric with name `resource_group_mem_allocated_bytes` in the documentation since version 3.1
but it is exposed with the internal name `resource_group_mem_inuse_bytes` instead of the name given in the documentation. This causes confusion when trying to find and use this metric
See https://docs.starrocks.io/docs/3.1/administration/management/monitoring/metrics/#starrocks_be_resource_group_mem_allocated_bytes
## What I'm doing:
Changing the name in the doc to match with internal name

Fixes #issue

## What type of PR is this:

- [] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [X] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [X] 4.1
  - [X] 4.0
  - [X] 3.5
  - [X] 3.4


kevincai (commented):
need fix the usage in `docs/*/administration/resource_group.md` as well

be/src/exec/workgroup/work_group.cpp
```diff
@@ -354,7 +354,7 @@ void WorkGroupManager::add_metrics_unlocked(const WorkGroupPtr& wg, UniqueLockTy
         // mem use bytes.
         auto resource_group_mem_allocated_bytes = std::make_unique<IntGauge>(MetricUnit::BYTES);
         bool mem_inuse_registered = GlobalMetricsRegistry::instance()->metrics()->register_metric(
-                "resource_group_mem_inuse_bytes", MetricLabels().add("name", wg->name()),
+                "resource_group_mem_allocated_bytes", MetricLabels().add("name", wg->name()),
```
kevincai: I think the AI comment is valid, since this is already in use, we'd better fix the doc, turn to `starrocks_be_resource_group_mem_inuse_bytes` in doc, rather than change the code to follow the doc.

behcetm: Ok, makes sense. I changed the doc now.

kevincai (dismissed):

DanRoscigno (dismissed):

DanRoscigno (approved):

EsoragotoSpirit (approved):

PR #70121, opened 2026-03-11
[BugFix] Fix connector sink mem tracker release on poller threads

GavinMar:
## Why I'm doing:
A connector sink query can keep query pool memory accounted after async flush polling or victim flushing is triggered from the blocked-driver poller thread. That causes query pool consumption to keep growing until it exceeds the process tracker and eventually hits the query memory limit incorrectly.

## What I'm doing:
Bind the fragment instance mem tracker while `ConnectorSinkOperator::need_input()` and `ConnectorSinkOperator::is_finished()` execute, so poller-thread cleanup and victim flushing release memory back to the correct query and query-pool tracker chain.

Add regression tests covering both the `need_input()` victim-flush path and the `is_finished()` async-stream cleanup path.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/pipeline/sink/connector_sink_operator.cpp
+++ b/be/src/exec/pipeline/sink/connector_sink_operator.cpp
@@ -21,6 +21,7 @@
 #include "formats/utils.h"
 #include "glog/logging.h"
 #include "runtime/exec_env.h"
+#include "runtime/current_thread.h"
 
 namespace starrocks::pipeline {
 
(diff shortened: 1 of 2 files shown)


be/src/exec/pipeline/sink/connector_sink_operator.cpp
```diff
@@ -61,32 +62,46 @@ bool ConnectorSinkOperator::need_input() const {
...
+    Status status;
+    bool can_accept_more_input;
+    {
+        SCOPED_THREAD_LOCAL_MEM_TRACKER_SETTER(runtime_state->instance_mem_tracker());
+        std::tie(status, std::ignore) = _io_poller->poll();
+        if (status.ok()) {
+            status = _connector_chunk_sink->status();
+        }
+        can_accept_more_input = _sink_mem_mgr->can_accept_more_input(_op_mem_mgr);
     }
```
stdpain, line 75: Have you considered making these function calls lighter-weight to facilitate future support for an event scheduler?

GavinMar: > Have you considered making these function calls lighter-weight to facilitate future support for an event scheduler?

You mean to make the poller lighter-weight ? Yes, I plan to refactor the original related code soon, moving the memory detection and flush logic from the `need_input` to the main I/O path for writing, aligning it with the internal table processing. This fix is ​​primarily to quickly resolve the memory issue inherited from previous versions.

trueeyu (approved):

Youngwb (approved):

PR #70144, opened 2026-03-11
[BugFix] Preserve null fractions for outer join columns

OliLay:
## Why I'm doing:

Consider the following query:
```sql
explain costs
select
    *
from
    TABLE1 t1
    LEFT JOIN TABLE2 t2 ON t1.NULLABLE = t2.NULLABLE
    LEFT JOIN TABLE3 t3 ON t1.NULLABLE = t3.NULLABLE;
```

At the moment, even though `t1.NULLABLE` has proper null fractions (e.g. 0.8), the second join assumes that `t1.NULLABLE` has 0.0 null fractions. This is because for equality predicates, we [reset the null fractions to 0](https://github.com/StarRocks/starrocks/blob/623504c363b0be714cc04c8835f233b89822f9fc/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/BinaryPredicateStatisticCalculator.java#L409) since NULLs are not preserved. In the context of the predicate (and inner joins) this is right, but outer joins should not reset null fractions. Subsequent joins should re-use the original column null fractions instead.

See the following [EXPLAIN](https://gist.github.com/OliLay/cc33c1669a2a4d761d4cb5f375f09dc6) where the in the topmost join (7) 4: NULLABLE has 0 null fractions, while for the joins prior to that join the correct null fractions are shown. The fixed [EXPLAIN](https://gist.github.com/OliLay/e1c14bf67e7eefecdeed1e770e72dd92) contains the proper null fractions for 4: NULLABLE

Wrong null fractions lead to the join not being rewritten, e.g. by the join skew rule.

## What I'm doing:

For outer joins, do not reset the null fractions of columns used in join predicates.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
@@ -1296,13 +1296,13 @@ public class StatisticsCalculator extends OperatorVisitor<Void, ExpressionContex
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
                 joinStatsBuilder.setOutputRowCount(max(innerRowCount, leftRowCount));
                 computeNullFractionForOuterJoin(leftRowCount, innerRowCount, leftStatistics, rightStatistics,
-                        joinStatsBuilder);
+                        eqOnPredicates, joinStatsBuilder);
                 break;
             case ASOF_LEFT_OUTER_JOIN:
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
                 joinStatsBuilder.setOutputRowCount(leftRowCount);
                 computeNullFractionForOuterJoin(leftRowCount, innerRowCount, leftStatistics, rightStatistics,
-                        joinStatsBuilder);
+                        eqOnPredicates, joinStatsBuilder);
                 break;
             case LEFT_SEMI_JOIN:
                 joinStatsBuilder = Statistics.buildFrom(StatisticsEstimateUtils.adjustStatisticsByRowCount(
@@ -1323,7 +1323,7 @@ public class StatisticsCalculator extends OperatorVisitor<Void, ExpressionContex
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
                 joinStatsBuilder.setOutputRowCount(max(innerRowCount, rightRowCount));
                 computeNullFractionForOuterJoin(rightRowCount, innerRowCount, rightStatistics, leftStatistics,
-                        joinStatsBuilder);
+                        eqOnPredicates, joinStatsBuilder);
                 break;
             case RIGHT_ANTI_JOIN:
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
@@ -1335,9 +1335,9 @@ public class StatisticsCalculator extends OperatorVisitor<Void, ExpressionContex
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
                 joinStatsBuilder.setOutputRowCount(max(1, leftRowCount + rightRowCount - innerRowCount));
                 computeNullFractionForOuterJoin(leftRowCount + rightRowCount, innerRowCount, leftStatistics,
-                        leftStatistics, joinStatsBuilder);
+                        leftStatistics, eqOnPredicates, joinStatsBuilder);
                 computeNullFractionForOuterJoin(leftRowCount + rightRowCount, innerRowCount, rightStatistics,
-                        rightStatistics, joinStatsBuilder);
+                        rightStatistics, eqOnPredicates, joinStatsBuilder);
                 break;
             default:
                 throw new StarRocksPlannerException("Not support join type : " + joinType,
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
```diff
@@ -1419,11 +1421,29 @@ private Statistics buildStatisticsForUKFKJoin(JoinOperator joinType, UKFKConstra
...
+                                                 Statistics outerSideStatistics, Statistics innerSideStatistics,
+                                                 Statistics.Builder builder) {
+        // Restore the original null fractions for the outer (preserved) side's columns
+        for (Map.Entry<ColumnRefOperator, ColumnStatistic> entry : outerSideStatistics.getColumnStatistics().entrySet()) {
+            ColumnStatistic originalStat = entry.getValue();
+            ColumnStatistic currentStat = builder.getColumnStatistics(entry.getKey());
+            if (currentStat != null) {
+                builder.addColumnStatistic(entry.getKey(), buildFrom(currentStat) //
+                        .setNullsFraction(originalStat.getNullsFraction()) //
+                        .build());
```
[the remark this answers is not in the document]
OliLay: [adc6dd2](https://github.com/StarRocks/starrocks/pull/70144/commits/adc6dd2e9b7e45b1d910cd85dc4c19b45ba8329e)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
```diff
@@ -1419,11 +1422,34 @@ private Statistics buildStatisticsForUKFKJoin(JoinOperator joinType, UKFKConstra
...
+    // This method first restores the original null fractions for the outer side's columns, then computes
+    // the new null fractions for the inner (nullable) side's columns to account for additional null rows.
     private void computeNullFractionForOuterJoin(double outerTableRowCount, double innerJoinRowCount,
-                                                 Statistics statistics, Statistics.Builder builder) {
+                                                 Statistics outerSideStatistics, Statistics innerSideStatistics,
+                                                 Statistics.Builder builder) {
+        // Restore the original null fractions for the outer (preserved) side's columns.
+        // When outerTableRowCount < innerJoinRowCount (e.g. one-to-many matches), the output
+        // has more rows than the original outer table.
+        for (Map.Entry<ColumnRefOperator, ColumnStatistic> entry : outerSideStatistics.getColumnStatistics().entrySet()) {
```
stephen-shelby: The scaling makes sense for preserved-side join keys that were previously reset to 0.

One follow-up question: should this restoration apply to all preserved-side columns, or only to the eq-join columns? From `BinaryPredicateStatisticCalculator`, it looks like only the equality join columns are nulled out during join estimation.

To keep the fix scoped to the confirmed bug and minimize behavior changes, would it make sense to restore only the preserved-side join keys first? Applying the same scaling to non-join columns in one-to-many outer joins may underestimate their null fraction.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
```diff
@@ -1332,9 +1335,9 @@ private Void computeJoinNode(ExpressionContext context, JoinOperator joinType, S
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
                 joinStatsBuilder.setOutputRowCount(max(1, leftRowCount + rightRowCount - innerRowCount));
                 computeNullFractionForOuterJoin(leftRowCount + rightRowCount, innerRowCount, leftStatistics,
```
stephen-shelby, line 1337: nit: for `FULL OUTER JOIN`, both calls still use the same row-count parameters, so it seems hard for this helper to distinguish null-extension on the left side vs right side. This may already be a pre-existing limitation rather than something introduced by this PR, so probably fine to handle separately if you prefer.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
OliLay: I think that is a sensible change. [f08b7d9](https://github.com/StarRocks/starrocks/pull/70144/commits/f08b7d9b8b3ff4fcf3cede93595e75e112a26135)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
OliLay: I agree, but I also think it is not in the scope of the PR

stephen-shelby (approved):

trueeyu (approved):

PR #70220, opened 2026-03-12
[Enhancement] Parallelize tablet file copy in shared-data cluster replication

wxl24life:
## Why I'm doing:

During shared-data (lake-to-lake) cross-cluster replication, a single tablet may contain many segment and non-segment files. The previous implementation copies all files sequentially, which wastes available bandwidth and amplifies end-to-end replication latency.

Also, the file copy loop in `replicate_lake_remote_storage` had grown complex with inline error handling, retry logic, and mixed concerns, making it hard to maintain and test.

### 1. Parallel file copy via shared thread pool

Refactored the file copy loop into a **task-based architecture**: each file copy is wrapped as a `ReplicationTask` (lambda → `Status`), and all tasks are either run sequentially or submitted to the existing `REPLICATE_SNAPSHOT` thread pool for parallel execution.

- **Parallel mode** uses `CountDownLatch` to wait for all submitted tasks, with atomic error flag + mutex-protected first-error capture.
- **Backlog guard**: parallel copy is automatically disabled when the thread pool queue depth exceeds `num_threads × 8`, preventing further pressure on an already saturated pool — falls back to sequential copy gracefully.
- **Fast cancel**: each task checks `min_active_txn_id` before starting the copy, so aborted transactions don't waste I/O.

### 2. Introduced new dynamic BE configs

| Config | Default | Description |
|---|---|---|
| `lake_replication_parallel_copy_min_file_count` | 2 | Min files to enable parallel copy; 0 = force disable |
| `lake_replication_max_file_copy_retry` | 3 | Max retry for non-segment file copy on failure or size mismatch |

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [x] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/lake_replication_txn_manager.h
+++ b/be/src/storage/lake/lake_replication_txn_manager.h
@@ -14,6 +14,10 @@
 
 #pragma once
 
+#include <atomic>
+#include <functional>
+#include <vector>
+
 #include "common/status.h"
 #include "fs/encryption.h"
 #include "gen_cpp/AgentService_types.h"
@@ -26,6 +30,7 @@ struct FileEncryptionPair;
 struct WritableFileOptions;
 class TabletMetadataPB;
 class FileSystem;
+class ThreadPool;
 } // namespace starrocks
 
 using starrocks::FileConverterCreatorFunc;
@@ -35,6 +40,8 @@ class TabletManager;
 
 class LakeReplicationTxnManager {
 public:
+    using ReplicationTask = std::function<Status()>;
+
     explicit LakeReplicationTxnManager(lake::TabletManager* tablet_manager)
             : _tablet_manager(tablet_manager)
 #ifdef USE_STAROS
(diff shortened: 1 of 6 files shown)


be/src/storage/lake/lake_replication_txn_manager.h
```diff
@@ -98,8 +105,21 @@ class LakeReplicationTxnManager {
                                                              const std::string& target_file_location,
                                                              const WritableFileOptions& opts, int max_retry);
 
+    // Run tasks in order and stop on first error.
+    static Status run_tasks_sequential(const std::vector<ReplicationTask>& tasks);
```
xiangguangyxg: Seems these are common functions, move to a separete helper class?

wxl24life: fixed

xiangguangyxg (approved):

luohaha (approved):

trueeyu (approved):

kevincai (approved):

PR #70221, opened 2026-03-12
[Enhancement] Calculate null fractions for CASE WHEN expressions

OliLay:
## Why I'm doing:
Assume you have a query with a CASE WHEN where the WHENs likely do not match, e.g. like this:
```sql
SELECT
    CASE
        WHEN LIKE(`STRING`, "%pat1%") THEN `ID`
        WHEN LIKE(`STRING`, "%pat2%") THEN `ID2`
    END
FROM
    TABLE1
```
There is an implicit ELSE that returns NULL. The SR stats of such an expressions are the following:
```
* ID1-->[0.0, 9999999.0, 0.0, 8.0, 909299.0] ESTIMATE
* ID2-->[0.0, 9999999.0, 0.0, 8.0, 909299.0] ESTIMATE
* case-->[-Infinity, Infinity, 0.0, 8.0, 1832770.0] ESTIMATE
```
Note the 0.0 null fractions. Since it is hard to evaluate the predicates, in a first step it makes sense to assume equal probability of each WHEN and the ELSE branches to get an good approximation of null fractions instead of defaulting them to 0.0. Later we can also look at predicate probabilities and use these to weigh the branches.

## What I'm doing:
Instead of defaulting null fractions of CASE WHENs to 0.0, I am weighting all branches in the CASE WHEN equally and compute the average over all the null fractions and use this as stat output.
In our case, if we assume ID1 and ID2 have 0.0 null fractions, the CASE WHEN would propagate 0.33 null fractions since `(0.0 + 0.0 + 1.0) / 3`.
```
* ID1-->[0.0, 9999999.0, 0.0, 8.0, 909299.0] ESTIMATE
* ID2-->[0.0, 9999999.0, 0.0, 8.0, 909299.0] ESTIMATE
* case-->[-Infinity, Infinity, 0.3333333333333333, 8.0, 1818599.0] ESTIMATE
```

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

Seaven (dismissed):

trueeyu (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
```diff
@@ -115,14 +115,23 @@ public ColumnStatistic visitCaseWhenOperator(CaseWhenOperator caseWhenOperator,
...
+                childrenColumnStatistics.add(ColumnStatistic.builder() //
+                        .setNullsFraction(1.0) //
+                        .setDistinctValuesCount(0) //
+                        .build());
             }
             // 2. use sum of then clause and else clause's distinct values as column distinctValues
             double distinctValues = childrenColumnStatistics.stream().mapToDouble(
                     ColumnStatistic::getDistinctValuesCount).sum();
+            // 3. Use the average null fraction of all branches.
+            double nullFractions = childrenColumnStatistics.stream().mapToDouble(ColumnStatistic::getNullsFraction).sum()
```
stephen-shelby, line 133: The overall `nullsFraction` for a `CASE WHEN` probably should not be computed as a simple average across branches. The result nullability depends on how likely each `WHEN` branch is to be taken, so a weighted estimate based on branch selectivity would be more appropriate here. For example, in `CASE WHEN k = 1 THEN 'A' ELSE NULL END`, if `k = 1` is very rare, the final `nullsFraction` should be close to 1.0, while the current averaging approach would still produce 0.5. Could we estimate each branch probability using predicate selectivity and then compute the final `nullsFraction` as a weighted sum instead?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
```diff
@@ -115,14 +115,23 @@ public ColumnStatistic visitCaseWhenOperator(CaseWhenOperator caseWhenOperator,
             }
             if (caseWhenOperator.hasElse()) {
                 childrenColumnStatistics.add(caseWhenOperator.getElseClause().accept(this, context));
+            } else {
+                // A missing ELSE is an implicit ELSE that returns NULL.
+                childrenColumnStatistics.add(ColumnStatistic.builder() //
```
stephen-shelby, line 122: There is still an inconsistency here. A missing `ELSE` is modeled as an implicit NULL branch, but an explicit `ELSE NULL` still goes through `visitConstant(NULL)` and keeps the old statistics. This means semantically equivalent expressions can still produce different estimates, so `visitConstant(NULL)` likely needs a follow-up fix.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
OliLay: Yes, that would be the best way, I also mentioned this in the PR description: "Later we can also look at predicate probabilities and use these to weigh the branches."
Is it fine for you if I do this in a follow up? I think a simple null fraction calculation is better than always returning 0.0 for now.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
OliLay: That is a good point Stephen! Addressed in [602afb4](https://github.com/StarRocks/starrocks/pull/70221/commits/602afb4fc6ee83bf07b3a12c75891955ad2a1303)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
```diff
@@ -115,14 +115,23 @@ public ColumnStatistic visitCaseWhenOperator(CaseWhenOperator caseWhenOperator,
...
                     ColumnStatistic::getDistinctValuesCount).sum();
+            // 3. Use the average null fraction of all branches.
+            double nullFractions = childrenColumnStatistics.stream().mapToDouble(ColumnStatistic::getNullsFraction).sum()
+                    / childrenColumnStatistics.size();
             return ColumnStatistic.builder()
                     .setMinValue(Double.NEGATIVE_INFINITY)
                     .setMaxValue(Double.POSITIVE_INFINITY)
-                    .setNullsFraction(0)
+                    .setNullsFraction(nullFractions)
                     .setAverageRowSize(caseWhenOperator.getType().getTypeSize())
```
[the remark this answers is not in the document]
OliLay: [602afb4](https://github.com/StarRocks/starrocks/pull/70221/commits/602afb4fc6ee83bf07b3a12c75891955ad2a1303)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
```diff
@@ -115,14 +115,23 @@ public ColumnStatistic visitCaseWhenOperator(CaseWhenOperator caseWhenOperator,
...
             double distinctValues = childrenColumnStatistics.stream().mapToDouble(
                     ColumnStatistic::getDistinctValuesCount).sum();
+            // 3. Use the average null fraction of all branches.
+            double nullFractions = childrenColumnStatistics.stream().mapToDouble(ColumnStatistic::getNullsFraction).sum()
+                    / childrenColumnStatistics.size();
             return ColumnStatistic.builder()
                     .setMinValue(Double.NEGATIVE_INFINITY)
                     .setMaxValue(Double.POSITIVE_INFINITY)
-                    .setNullsFraction(0)
+                    .setNullsFraction(nullFractions)
```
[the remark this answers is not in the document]
OliLay: Also addressed with [602afb4](https://github.com/StarRocks/starrocks/pull/70221/commits/602afb4fc6ee83bf07b3a12c75891955ad2a1303)

trueeyu (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
stephen-shelby: sure. thanks for your contribution!

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
```diff
@@ -91,7 +91,8 @@ public ColumnStatistic visitVariableReference(ColumnRefOperator operator, Void c
         @Override
         public ColumnStatistic visitConstant(ConstantOperator operator, Void context) {
             if (operator.isNull()) {
-                return new ColumnStatistic(Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY, 0, 1, 1);
+                // NULL has no distinct non-null values and is always null.
+                return new ColumnStatistic(Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY, 1.0, 0, 0);
```
stephen-shelby: Setting `nullsFraction=1.0` and `distinctValuesCount=0` for a NULL literal is correct. However, setting `averageRowSize=0` introduces a potential division-by-zero risk.

In `visitBinaryPredicateOperator`, the `LEFT`/`RIGHT` string function case uses `left.getAverageRowSize()` as a divisor:

```java
distinctValues = left.getDistinctValuesCount() * averageRowSize / left.getAverageRowSize();
```

If the left operand happens to be a NULL literal (e.g. `LEFT(NULL, 3)`), `averageRowSize=0` would cause a divide-by-zero here.

Suggest keeping `averageRowSize` as `operator.getType().getTypeSize()` (consistent with how non-null constants are sized) rather than 0:

```java
if (operator.isNull()) {
    return new ColumnStatistic(Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY,
            1.0, operator.getType().getTypeSize(), 0);
}
```

OliLay: Ah, good catch: [9930488](https://github.com/StarRocks/starrocks/pull/70221/commits/99304884bf92007e4b34b779d9e3e6a24e3fdc86)

stephen-shelby (dismissed):

stephen-shelby (approved):

trueeyu (approved):

PR #70242, opened 2026-03-13
[Enhancement] Fail-fast for unsupported Iceberg V3 features

GavinMar:
## Why I'm doing:

StarRocks does not yet support several Iceberg V3 features: Deletion Vectors (DVs), table encryption, and new extended types (geometry, geography, timestamp_nano). Without explicit guards, queries against tables using these features can produce silent data correctness issues (e.g., reading deleted data when DVs are cached incorrectly) or confusing late-stage errors from the BE.

## What I'm doing:

Add fail-fast checks for unsupported Iceberg V3 features at the appropriate stage of query processing:

1. **Deletion Vectors**: Detect DV files (Puffin-format position deletes) during scan range building and throw a clear error. Also fix `DeleteFileWrapper` missing `referencedDataFile()`, `contentOffset()`, and `contentSizeInBytes()` proxies, which caused DVs to silently disappear from the delete file cache after the first query.
2. **Table encryption**: Check `TableProperties.ENCRYPTION_TABLE_KEY` and `TableMetadata.encryptionKeys()` in `triggerIcebergPlanFilesIfNeeded()` — early enough to avoid wasted planning work, but not in `getTable()` so DDL operations like DROP TABLE are not blocked.
3. **Unsupported types**: Unsupported V3 types (geometry, geography, timestamp_nano) gracefully degrade to `UNKNOWN_TYPE`. The existing `Scope.resolveField` / `SelectAnalyzer` guards reject queries on those columns at analysis time, while queries on other columns of the same table continue to work.
4. **Error message transparency**: Add `catch (StarRocksConnectorException e) { throw e; }` before the generic `catch (Exception e)` in `toScanRanges()` so that DV/encryption errors propagate with their original message instead of being wrapped as "build scan range failed".

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

Queries against Iceberg tables with Deletion Vectors or encryption now return a clear error message instead of silently returning incorrect results or failing with an obscure BE error.

## Checklist:

- [ ] This is a backport PR

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/DeleteFileWrapper.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/DeleteFileWrapper.java
@@ -147,6 +147,21 @@ public class DeleteFileWrapper implements DeleteFile {
     }
 
     @Override
+    public String referencedDataFile() {
+        return deleteFile.referencedDataFile();
+    }
+
+    @Override
+    public Long contentOffset() {
+        return deleteFile.contentOffset();
+    }
+
+    @Override
+    public Long contentSizeInBytes() {
+        return deleteFile.contentSizeInBytes();
+    }
+
+    @Override
     public DeleteFile copy() {
         return deleteFile.copy();
     }
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -132,6 +132,7 @@ import org.apache.iceberg.SnapshotSummary;
 import org.apache.iceberg.SnapshotUpdate;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.StarRocksIcebergTableScan;
+import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableScan;
 import org.apache.iceberg.Transaction;
 import org.apache.iceberg.catalog.TableIdentifier;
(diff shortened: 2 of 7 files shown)


fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
```diff
@@ -657,6 +659,8 @@ public Table getTable(ConnectContext context, String dbName, String tblName) {
                 icebergTable = icebergCatalog.getTable(context, dbName, tblName);
             }
 
+            checkUnsupportedEncryption(icebergTable);
```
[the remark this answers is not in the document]
GavinMar: fixed

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/DeleteFileWrapper.java
```diff
@@ -146,6 +146,21 @@ public List<Integer> equalityFieldIds() {
         return deleteFile.equalityFieldIds();
     }
 
+    @Override
+    public String referencedDataFile() {
+        return deleteFile.referencedDataFile();
+    }
+
+    @Override
+    public Long contentOffset() {
```
Youngwb, line 155: Why need add these methods

GavinMar: Now we reject a query if found DVs, so we need to check whether there are related DVs for the data files.  For the first query, we can get the delete files based on the original `DeleteFile` object in `planFiles`.
In the later queries,  if a data file hits cache, we will try to find the related DV from `deleteFileIndex`. However, the default `referencedDataFile()` for `DeleteFileWrapper` returns null, which makes the `findDV` cannot return the correct DVs, and the fail-fast check will be skipped. Therefore, the actual test results are as follows: when querying a table containing a DV exception, the first query will result in an error, but from the second query onwards, the query will succeed (but return incorrect data). As for `contentOffset()`, this function is actually not used in this PR, but in order to avoid some bugs caused by it when we support DV later, we also fix it here together with `referencedDataFile()`.

Youngwb (approved):

dirtysalt (approved):

PR #70269, opened 2026-03-13
[BugFix] preserve varchar length after reduce cast with global variables

jaogoy:
## Why I'm doing

For statements such as `CAST(col AS VARCHAR(N))`, FE may return different result-set metadata between prepare and execute. This is visible to MSDASQL/ODBC clients and can cause client-side metadata drift.

The root cause is that `ReduceCastRule` removes same-type `VARCHAR -> VARCHAR` casts during optimization, but the target `VARCHAR(N)` length metadata can be lost after the cast is reduced. In the planner path, reused `Expr` and `SlotRef` objects may still keep the original column varchar length, so execute can expose a different output type than prepare.

## What I'm doing

- Add a global switch `enable_reduce_cast_varchar_length_inheritance`. (default `false`)
  - Preserve the target varchar length when `ReduceCastRule` eliminates a same-type `VARCHAR -> VARCHAR` cast.
- Add a global switch `enable_reduce_cast_varchar_expr_sync_type`. (default `false`)
  - Keep it as a switch when planner `Expr` metadata also needs to be synchronized when `inheritance` is `true`.

Others:
- Add planner-level regression tests for final output slot metadata.
- Add sql-test by using `function: xxx` to call a python function, which will using row mysql protocol to check the varchar length from `prepare` and `execute`. (Other commonly used python mysql drivers can get metadata for `prepare` interface)
- Document the new global variables and their introduced versions.

Fixes #issue

## Usage

Enable the fix with:

```sql
SET GLOBAL enable_reduce_cast_varchar_length_inheritance = true;
SET GLOBAL enable_reduce_cast_varchar_expr_sync_type = true;
```

## Impact

This fixes prepare/execute varchar length metadata drift for clients that rely on result-set metadata, especially MSDASQL/ODBC. Default behavior stays unchanged unless the inheritance switch is enabled, and the change is scoped to same-type `VARCHAR -> VARCHAR` cast reduction.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/ReduceCastRule.java
```diff
@@ -169,6 +170,30 @@ public boolean checkCastTypeReduceAble(Type parent, Type child, Type grandChild)
         return childCompatibleType != InvalidType.INVALID && parentCompatibleType != InvalidType.INVALID;
     }
 
+    private ScalarOperator inheritVarcharLengthAfterReduceCast(CastOperator operator) {
+        ScalarOperator child = operator.getChild(0);
+        if (!shouldInheritVarcharLength(operator.getType(), child.getType())) {
+            return child;
+        }
+
+        if (GlobalVariable.isEnableReduceCastVarcharTypeLengthMutation()) {
```
stephen-shelby:
```
        ScalarOperator rewrittenChild = child.clone();
        if (GlobalVariable.isEnableReduceCastVarcharTypeLengthMutation()) {
            Type rewrittenType = child.getType().clone();
            ((ScalarType) rewrittenType).setLength(((ScalarType) operator.getType()).getLength());
            rewrittenChild.setType(rewrittenType);
        } else {
            rewrittenChild.setType(operator.getType().clone());
        }
        return rewrittenChild;
```

stephen-shelby (dismissed):

stephen-shelby (approved):

HangyuanLiu (changes requested):
## PR Review

### Rule: behavior-change-label ❌ FAIL

**Finding:** The PR introduces two new global variables (`enable_reduce_cast_varchar_length_inheritance` and `enable_reduce_cast_varchar_expr_sync_type`) in `GlobalVariable.java`. When enabled, these variables alter the evaluation behavior of same-type `VARCHAR -> VARCHAR` casts — specifically, the output varchar length metadata returned during `PREPARE`/`EXECUTE` differs from the default path. The PR's own `## Impact` section acknowledges this behavioral effect on MSDASQL/ODBC clients.

The behavior-change checkbox is currently set to:
```
```

While the *default* behavior is unchanged (both switches default to `false`), this PR does introduce new opt-in behavior that alters SQL expression evaluation results. Adding user-visible, behavior-altering global variables falls within the spirit of the behavior-change declaration requirement.

**Request:** Please update the checkbox to:
```
```
And select the applicable change type (e.g., *Parameter changes: default values, similar parameters but with different default values*). Consider adding the `behavior-change` label as well.

---

### Rule: meaningful-description ✅ PASS

Both the **Why** and **What** sections contain substantive content that clearly explains the root cause (prepare/execute metadata drift due to `ReduceCastRule` discarding varchar length) and the fix approach. No action needed.

---

> Minor nit: `Fixes #issue` appears to be an unfilled placeholder — please replace with the actual issue number or remove if there is no linked issue.

HangyuanLiu (approved):

Issue #64324, opened 2025-10-21
Translate Trino not working for Array Type, cardinality and array_intersect function


arundh93:
### Steps to reproduce the behavior (Required)
1. Use translate trino functionality on array types.
2. Use array functions like cardinality and array_intersect

### Expected behavior (Required)
We get the translated starrocks query

### Real behavior (Required)
We get an exception
ERROR 1064 (HY000): Unknown error

### StarRocks version (Required)
3.4.5 (but seems to be missing in main as well)

<img width="968" height="146" alt="Image" src="https://github.com/user-attachments/assets/558987f7-efe0-49d4-9ece-13b512fb1768" />

PR #70338, opened 2026-03-16
[BugFix] Fix array types support and cardinality and array_intersect functions in TRANSLATE TRINO

arundh93:
## Why I'm doing:
Functionality with array types and some array functions is not working

## What I'm doing:
Registering `cardinality` and `array_intersect` function in Trino to Starrocks Transformer.
Additionally we are also enhancing the ArrayType creation in the `ASTBuilder`

Fixes #issue
https://github.com/StarRocks/starrocks/issues/64324

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4
  - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/parser/trino/AstBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/parser/trino/AstBuilder.java
@@ -759,7 +759,28 @@ public class AstBuilder extends AstVisitor<ParseNode, ParseTreeContext> {
         } else {
             exprs = Collections.emptyList();
         }
-        return new ArrayExpr(null, exprs);
+
+        Type arrayType = null;
+        if (!exprs.isEmpty()) {
+            try {
+                Expr firstNonNullExpr = null;
+                for (Expr expr : exprs) {
+                    if (!(expr instanceof NullLiteral)) {
+                        firstNonNullExpr = expr;
+                        break;
+                    }
+                }
+
+                if (firstNonNullExpr != null) {
+                    if (firstNonNullExpr.getType() != null) {
+                        arrayType = new ArrayType(firstNonNullExpr.getType());
+                    }
+                }
+            } catch (Exception e) {
+            }
+        }
+
+        return new ArrayExpr(arrayType, exprs);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/connector/parser/trino/AstBuilder.java
```diff
@@ -759,7 +759,28 @@ protected ParseNode visitArrayConstructor(ArrayConstructor node, ParseTreeContex
...
+                for (Expr expr : exprs) {
+                    if (!(expr instanceof NullLiteral)) {
+                        firstNonNullExpr = expr;
+                        break;
+                    }
+                }
+
+                if (firstNonNullExpr != null) {
+                    if (firstNonNullExpr.getType() != null) {
+                        arrayType = new ArrayType(firstNonNullExpr.getType());
```
stephen-shelby, line 776: One concern with this change is the new array type inference in Trino AstBuilder.

`translate trino` does not go through the normal `ExpressionAnalyzer` path, so I agree we likely need to assign an `ArrayExpr` type during translation if we want `AstToSQLBuilder` to print `ARRAY<T>[...]`. However, using the first non-null element type here looks unsafe.

It can incorrectly narrow mixed-type arrays such as `array[1, 1000]`, `array[1, 1.5]`, or `array[null, 1, 1000]`. In the normal SR path, array element typing is based on `TypeManager.getCommonSuperType(...)`, not the first non-null child. So this change introduces a different typing rule in the translation path.

I think the safer fix here would be to infer the array item type from all element types via `TypeManager.getCommonSuperType(...)` instead of `firstNonNullExpr.getType()`.

Could we also add coverage for mixed-type cases like:
- `translate trino select array[1, 1000]`
- `translate trino select array[1, 1.5]`
- `translate trino select array[null, 1, 1000]`

That would help ensure we do not introduce a type-narrowing regression.

arundh93: that makes complete sense, will make those changes

eshishki:
@arundh93 this supersedes #64326 (same change, just with the commit email fixed) — could you close #64326 so we don't carry the duplicate?

Still outstanding on your side is the **DCO check**: your commit needs a `Signed-off-by` line:
```bash
git commit --amend --signoff
git push --force-with-lease
```
The FE UT red is an unrelated flaky batch (lake/tablet/StarOS tests, not the Trino change) and should clear on a re-run after the push.

@stdpain could you take a look? This registers `cardinality` / `array_intersect` in the Trino dialect transformer (issue #64324).

stdpain:
https://github.com/Mergifyio update main

PR #70362, opened 2026-03-17
[Enhancement] WIP: Reuse common sub-expressions during query optimization

lxynov:
# 🚧 WORK IN PROGRESS 🚧
---
# Overview
Existing CSE (common sub-expression elimination) is applied at the final physical rewrite phase. Before that, and throughout the query optimization phase, common sub-expressions are inlined.

For queries containing nested case-when expressions, this inlining leads to an multiplying effect on the number of expression nodes, which is inefficient both time-wise and memory-wise. We came across relevant failures (`Expression too complex` and `Expression CaseWhen too complex`) for some of our workloads, similar to what's mentioned in https://github.com/StarRocks/starrocks/issues/66325 and https://github.com/StarRocks/starrocks/pull/66324. In an extreme case, we had to increase `Config.max_scalar_operator_flat_children` to 2million so that one query can pass, and its planning took 38s.

This change aims to support reusing common sue-expressions throughout the query optimization phase.

# Core Change
The core change is to introduce a common sub-expression map to projections. This includes both `LogicalProjectNode` and `Projection`. Note that `Projection` already has an existing common sub-expression map but today it's always empty before the physical rewrite phase.

### Merge
```
top output column map
top common column map
                          ---->  merged output column map
bottom output column map         merged common column map       
bottom common column map
```
Used by `MergeTwoProjectRule`, `JoinAssociateBaseRule`.

In some cases, some bottom output columns may be "promoted" into the merged output column map.

### Split
**Vertical split**
```
                          top output column map
                          top common column map
output column map  ---->
common column map
                          bottom output column map
                          bottom common column map 
```
Used by `HoistHeavyCostExprsUponTopnRule`

**Horizontal split**
```
output column map  ----> first output column map   +   second output column map
common column map        first common column map       second common column map
```

### Traverse
Happens in stats calculation.

### Calculate Inputs
Happens when the input columns are needed.

# Required Changes
## Project Merging and Physical CSE Rewriting
Key files:
- `MergeTwoProjectRule`
- `MergeProjectWithChildRule`
- `ScalarOperatorsReuse`
## TODO: How to make sure everything is covered?

# Alternative Solutions
- Make the scalar expression a DAG rather than a tree?
- Today, before CBO, adjacent `LogicalProjectOperator`s are merged and incorporated to its child as a `Projection`, is it okay to retain `LogicalProjectOperator`s after RBO?

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/common/DebugOperatorTracer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/common/DebugOperatorTracer.java
@@ -93,8 +93,18 @@ import java.util.stream.Collectors;
 public class DebugOperatorTracer extends OperatorVisitor<String, Void> {
     private String appendProjectionAndPredicate(Operator op) {
         StringBuilder sb = new StringBuilder();
-        if (op.getProjection() != null && !op.getProjection().getColumnRefMap().isEmpty()) {
-            sb.append(appendProject(op.getProjection().getColumnRefMap()));
+        if (op.getProjection() != null) {
+            if (!op.getProjection().getColumnRefMap().isEmpty()) {
+                sb.append(appendProject(op.getProjection().getColumnRefMap()));
+                if (!op.getProjection().getCommonSubOperatorMap().isEmpty()) {
+                    sb.append(", ");
+                    String subExprStr = op.getProjection().getCommonSubOperatorMap().entrySet()
+                            .stream()
+                            .map(e -> e.getKey() + "->" + e.getValue())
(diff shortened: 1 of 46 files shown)


Seaven (commented):
@stephen-shelby you know this changes ? This is a very high-risk PR, and the implementation isn't very elegant

lxynov:
The `Expression too complex` and `Expression CaseWhen too complex` issues we encountered are fixed by https://github.com/StarRocks/starrocks/pull/70802.

Closing this PR because its scope is too huge and it's hard to find a clean&robust solution due to the nature of the problem.

PR #70389, opened 2026-03-17
[Enhancement] Auto warmup SST files for cloud-native PK index during CACHE SELECT

luohaha:
## Why I'm doing:
During CACHE SELECT for cloud-native primary key tables, only segment data files are warmed up into local DataCache. However, PK ingestion workloads also heavily depend on persistent index SST files. Without pre-caching SST files, the first ingestion after warmup still suffers from cold-start latency when reading SST files from remote storage.

## What I'm doing:
Automatically warm up persistent index SST files when CACHE SELECT includes PK columns on a cloud-native PK table. No additional PROPERTIES needed — SST warmup is triggered transparently when PK columns are detected in the selected columns.

**How it works:**
- `_has_pk_columns_selected()` checks if the tablet schema is PRIMARY_KEYS and any selected slot maps to a key column
- If true (and `cache_file_only` mode is active), `_warmup_pk_index_sst_files()` iterates over SST files from tablet metadata and warms each via `touch_cache()`

**SST warmup conditions (all must be met):**
- CACHE SELECT in physical mode (`cache_file_only = true`)
- Primary key table with selected PK columns
- Cloud-native persistent index (`persistent_index_type = CLOUD_NATIVE`)
- Tablet metadata contains SST files (`sstable_meta` is non-empty)

## What type of PR is this:
- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:
- [ ] This is a backport PR

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -16,9 +16,8 @@
 
 #include <vector>
 
-#include "base/testutil/sync_point.h"
-
 #include "base/string/string_parser.hpp"
+#include "base/testutil/sync_point.h"
 #include "column/column_access_path.h"
 #include "common/config_lake_fwd.h"
 #include "common/config_scan_io_fwd.h"
@@ -755,8 +754,7 @@ Status LakeDataSource::prune_schema_by_access_paths(Schema* schema) {
     return Status::OK();
 }
 
-bool has_all_pk_columns_selected(const TabletSchema* tablet_schema,
-                                 const std::vector<SlotDescriptor*>& slots) {
+bool has_all_pk_columns_selected(const TabletSchema* tablet_schema, const std::vector<SlotDescriptor*>& slots) {
     if (tablet_schema == nullptr) {
         return false;
     }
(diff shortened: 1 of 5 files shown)


gensrc/thrift/InternalService.thrift
```diff
@@ -367,6 +367,8 @@ struct TQueryOptions {
   210: optional bool enable_global_late_materialization;
   211: optional bool enable_schedule_log;
 
+  212: optional bool warmup_cache_select_sst;
```
kevincai: should this be turned on as always in cache select case?

be/src/connector/lake_connector.cpp
```diff
@@ -127,6 +130,14 @@ Status LakeDataSource::open(RuntimeState* state) {
     RETURN_IF_ERROR(build_scan_range(_runtime_state));
 
     RETURN_IF_ERROR(init_tablet_reader(_runtime_state));
+
+    // Warmup SST files for PK table with cloud-native persistent index during CACHE SELECT
+    if (_params.lake_io_opts.cache_file_only && _runtime_state->query_options().__isset.warmup_cache_select_sst &&
```
starrocks-xupeng: i think you can do it only when PK column is selected in cache select, and you don't have to introdce new property, and put warmup in `LakeDataSource::open` might not be a good idea

luohaha: Good idea, will do

luohaha: done

be/src/connector/lake_connector.cpp
```diff
@@ -127,6 +129,15 @@ Status LakeDataSource::open(RuntimeState* state) {
     RETURN_IF_ERROR(build_scan_range(_runtime_state));
 
     RETURN_IF_ERROR(init_tablet_reader(_runtime_state));
+
+    // Automatically warmup SST files when CACHE SELECT includes PK columns
+    // on a cloud-native PK table with persistent index
+    if (_params.lake_io_opts.cache_file_only && _slots != nullptr &&
+        has_all_pk_columns_selected(_tablet_schema.get(), *_slots)) {
+        auto st = warmup_pk_index_sst_files(_tablet.metadata().get(), ExecEnv::GetInstance()->lake_tablet_manager());
```
kevincai: It is not good to do it in DataSource::open(). this interface is not guaranteed to be invoked in IO thread pool.

luohaha: done

kevincai (dismissed):

be/src/connector/lake_connector.cpp
```diff
@@ -738,6 +754,93 @@ Status LakeDataSource::prune_schema_by_access_paths(Schema* schema) {
...
+    }
+    for (size_t i = 0; i < num_key_columns; i++) {
+        if (!slot_names.contains(tablet_schema->column(i).name())) {
+            return false;
+        }
+    }
+    return true;
+}
+
+Status warmup_pk_index_sst_files(const TabletMetadataPB* metadata, lake::TabletManager* tablet_mgr) {
```
starrocks-xupeng, line 781: why not return error directly to user, should be treated like normal query, so we don't have to find the error in log file

luohaha: done

starrocks-xupeng (approved):

kevincai (approved):

wyb (approved):

PR #70392, opened 2026-03-17
[BugFix][Iceberg] Fix OAuth2 token refresh scheduler never started for REST Catalog

bksahu-atlan:
Problem:
When using Iceberg REST Catalog with OAuth2 (e.g. Apache Polaris), queries fail after ~1 hour idle with:
  NotAuthorizedException: Failed to verify the token

Root cause:
RefreshingAuthManager.keepRefreshed(boolean) is never called after delegate.initialize() in IcebergRESTCatalog. The background token refresh ScheduledExecutorService never starts. Token expires after TTL (~3600s for Polaris) with no renewal. Confirmed via bytecode analysis: keepRefreshed does not appear in IcebergRESTCatalog.class constant pool at all. No catalog property can work around this - the scheduler is architecturally never started.

Fix:
After delegate.initialize(), check if authManager is a RefreshingAuthManager and call keepRefreshed(true) to start the background token refresh scheduler. Backward compatible: only activates when auth manager is a RefreshingAuthManager instance (i.e. OAuth2 is in use).

Tested: StarRocks 4.0.6 + Apache Polaris, catalog remains functional after 90+ minutes idle with no token expiry errors.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
@@ -44,6 +44,7 @@ import org.apache.iceberg.exceptions.BadRequestException;
 import org.apache.iceberg.exceptions.RESTException;
 import org.apache.iceberg.rest.RESTSessionCatalog;
 import org.apache.iceberg.rest.auth.OAuth2Properties;
+import org.apache.iceberg.rest.auth.RefreshingAuthManager;
 import org.apache.iceberg.util.PropertyUtil;
 import org.apache.iceberg.view.View;
 import org.apache.iceberg.view.ViewBuilder;
@@ -126,6 +127,9 @@ public class IcebergRESTCatalog implements IcebergCatalog {
             delegate = new RESTSessionCatalog();
             configureHadoopConf(delegate, conf);
             delegate.initialize(name, restCatalogProperties);
+            if (delegate.authManager() instanceof RefreshingAuthManager) {
+                ((RefreshingAuthManager) delegate.authManager()).keepRefreshed(true);
+            }
         } catch (Exception re) {
             LOG.error("Failed to rest load catalog", re);
             throw new StarRocksConnectorException("Failed to load rest catalog",
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
@@ -127,14 +127,22 @@ public class IcebergRESTCatalog implements IcebergCatalog {
             delegate = new RESTSessionCatalog();
             configureHadoopConf(delegate, conf);
             delegate.initialize(name, restCatalogProperties);
-            Object authManager = delegate.authManager();
-            if (authManager instanceof RefreshingAuthManager) {
-                ((RefreshingAuthManager) authManager).keepRefreshed(true);
-            }
         } catch (Exception re) {
             LOG.error("Failed to rest load catalog", re);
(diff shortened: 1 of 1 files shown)


fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
```diff
@@ -126,6 +127,9 @@ public IcebergRESTCatalog(String name, Configuration conf, Map<String, String> p
             delegate = new RESTSessionCatalog();
             configureHadoopConf(delegate, conf);
             delegate.initialize(name, restCatalogProperties);
+            if (delegate.authManager() instanceof RefreshingAuthManager) {
+                ((RefreshingAuthManager) delegate.authManager()).keepRefreshed(true);
```
kevincai: will it need to do any action if this Catalog get deleted from the system?

bksahu-atlan: Great question @kevincai! Addressed in the latest commit — stored 
`RefreshingAuthManager` as a field and added a `shutdown()` method 
that calls `keepRefreshed(false)` when the catalog is deleted/dropped. 
This stops the background token refresh scheduler and prevents thread leaks.

bksahu-atlan:
@kevincai @Wenjun7J — the CI workflow is awaiting maintainer approval  to run. Could you please approve it? Also pushed a fix for the compile 
error — `authManager` is a private field in Iceberg 1.10.0 with no public getter, so updated to access it via reflection in a separate  try-catch block.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
```diff
@@ -131,6 +132,18 @@ public IcebergRESTCatalog(String name, Configuration conf, Map<String, String> p
             throw new StarRocksConnectorException("Failed to load rest catalog",
                     new RuntimeException("Failed to load rest catalog, exception: " + re.getMessage(), re));
         }
+        try {
+            java.lang.reflect.Field authManagerField =
```
kevincai, line 136: it is really hack to use reflect to access the authManager, is there any other accessible interface that starts the auth refreshing?

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
```diff
@@ -131,6 +132,18 @@ public IcebergRESTCatalog(String name, Configuration conf, Map<String, String> p
...
+                    delegate.getClass().getDeclaredField("authManager");
+            authManagerField.setAccessible(true);
+            Object authMgr = authManagerField.get(delegate);
+            if (authMgr instanceof RefreshingAuthManager) {
+                this.refreshingAuthManager = (RefreshingAuthManager) authMgr;
+                this.refreshingAuthManager.keepRefreshed(true);
```
kevincai, line 142: also refer to this code: https://github.com/apache/iceberg/blob/apache-iceberg-1.10.0/core/src/main/java/org/apache/iceberg/rest/auth/RefreshingAuthManager.java#L42, the `keepRefresh` is true by default, is it still needed to explicitly set to `true` again?

feel like the real root cause is still not identified.

kevincai (changes requested):

kevincai:
close this PR as it can't be the root cause and can't make any difference at all.

PR #70476, opened 2026-03-18
[BugFix] Fix delete tablets misdeleting shared files from txn logs after tablet split

xiangguangyxg:
## Why I'm doing:

  After tablet split, `delete_tablets_impl()` processes txn logs before metadata. Old txn logs (written before split) do not mark files as shared, while the latest metadata (updated after split) correctly marks them as shared. Because txn logs were processed first, `is_shared_segment()` returned false for these files, causing them to be deleted. By the time metadata was processed, the shared files were already gone.

  Additionally, `rowset.del_files` marked as shared in metadata were never collected into the protection set, so shared PK del files suffered the same mis-deletion.

  There was also a retry-safety issue: since `AsyncFileDeleter` submits batches eagerly (when batch is full, not waiting for `finish()`), if metadata files were deleted before txn log processing and txn log reading subsequently failed, retry would find metadata gone and be unable to reconstruct the shared file protection set.

## What I'm doing:

  1. Reorder `delete_tablets_impl()` execution: process metadata before txn logs to collect `retained_files` (shared files that must not be deleted), then process txn logs with `retained_files` as a protection set.
  2. Add `retained_files` parameter to `delete_files_under_txnlog()` so it skips files that are identified as shared in metadata (segments, del files, delvec, dcg, sstable).
  3. Handle `rowset.del_files()` in metadata processing — both for deletion of non-shared del files and for collecting shared del files into `retained_files`.
  4. Defer metadata file deletion to after txn log processing, ensuring metadata remains available for retry if txn log processing fails.

Fixes #64986

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/vacuum.cpp
+++ b/be/src/storage/lake/vacuum.cpp
@@ -714,23 +714,33 @@ static StatusOr<BundleTabletMetaState> check_bundle_tablet_meta_state(
 }
 
 static Status delete_files_under_txnlog(const std::string& data_dir, const TxnLogPB& log, bool contains_alive_tablets,
-                                        bool is_combined_log, AsyncFileDeleter& deleter) {
+                                        bool is_combined_log, AsyncFileDeleter& deleter,
+                                        const std::unordered_set<std::string>& retained_files) {
     if (log.has_op_write()) {
         const auto& op = log.op_write();
         // Shared segments can be deleted only when we know all tablets in a combined log are being deleted.
         for (int i = 0; i < op.rowset().segments_size(); ++i) {
+            if (retained_files.count(op.rowset().segments(i))) {
+                continue;
+            }
             if (!is_shared_segment(op.rowset(), i) || (is_combined_log && !contains_alive_tablets)) {
                 RETURN_IF_ERROR(deleter.delete_file(join_path(data_dir, op.rowset().segments(i))));
             }
         }
         // delete del files
         for (const auto& f : op.dels()) {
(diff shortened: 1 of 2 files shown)


be/src/storage/lake/vacuum.cpp
```diff
@@ -948,13 +923,24 @@ Status delete_tablets_impl(TabletManager* tablet_mgr, const std::string& root_di
                     // If the segment file is not shared by other alive tablets, we can delete it directly.
                     if (!is_shared_segment(rowset, i) || allow_delete_shared_files) {
                         RETURN_IF_ERROR(deleter.delete_file(join_path(data_dir, rowset.segments(i))));
+                    } else {
+                        retained_files.insert(rowset.segments(i));
```
srlch, line 927: Is that means the latest tablet meta of the deleted tablet must have the share file flag here? So we can collected such files and protected them when delete the logs

xiangguangyxg: Yes, after tablet splitting, data files is set shared in old tablet metadata.

srlch (approved):

wyb (approved):

PR #70499, opened 2026-03-19
[BugFix] fix truncate bug

Wenjun7J:
## Why I'm doing:
PREPARE_COLUMN_WITH_CONST_AND_NULL_FOR_ICEBERG_FUNC unwraps ConstColumn Via
get_data_column_of_const , So cO->size（） becomes 1 when the caller passes a const
column （typical for literals/constant-folded exprs）. Using const int size = c0->size（）；
after this macro can therefore produce a 1-row result and/or a size mismatch with
null_flags （which was sized using the original co->size（））， potentially leading to
incorrect results or crashes when const inputs occur. Consider capturing the original input
row count before unwrapping const, and if columns ［e］ is const, either broadcast the
computed value to that row count or return a ConstColumn with the original size （and
ensure the indexing uses row O for const inputs）
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/function_helper.h
+++ b/be/src/exprs/function_helper.h
@@ -122,24 +122,25 @@ inline void FunctionHelper::get_data_of_column<BinaryColumn, Slice>(const Column
         }                                    \
     } while (false)
 
-#define PREPARE_COLUMN_WITH_CONST_AND_NULL_FOR_ICEBERG_FUNC(c0, c1)     \
-    do {                                                                \
-        if (c0->only_null() || c1->only_null()) {                       \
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #70503, opened 2026-03-19
[BugFix] Track Arrow memory pool allocations in Parquet reader via MemTracker

xiangguangyxg:
##  Why I'm doing:
                                                                                                                                                                                                                                                                             
  Arrow IO prebuffering in the legacy Parquet reader (ParquetReaderWrap) accounts for a significant
  portion of memory consumption, but is completely invisible to StarRocks' MemTracker system. The root                                                                                                                                                                       
  cause is that ParquetReaderWrap uses arrow::default_memory_pool(), and Arrow's internal IO threads                                                                                                                                                                         
  allocate memory via posix_memalign without any MemTracker integration.

  Additionally, replacing the global arrow::default_memory_pool() with a per-instance pool introduces
  a lifetime concern: Arrow's future.Cancel() is cooperative — already-running IO tasks won't stop
  immediately. If ParquetReaderWrap is destroyed while an IO thread still holds buffers allocated from
  the pool, those buffers would call Free() on a destroyed pool (UAF).

##  What I'm doing:

  Introduce ArrowMemoryPool with MemTracker integration, safe pool lifetime management, and correct
  interaction with the jemalloc allocator hooks.

  1. ArrowMemoryPool with MemTracker tracking

  ArrowMemoryPool gains an optional MemTracker* constructor. In production builds, memory tracking
  is done via SCOPED_THREAD_LOCAL_MEM_TRACKER_SETTER(_mem_tracker) which temporarily redirects the
  jemalloc allocator hook to attribute posix_memalign/std::free to the target tracker — this is the
  single tracking path, avoiding double-counting between the hook and explicit consume()/release().
  In test builds (BE_TEST, ASAN) where jemalloc hooks are not active, explicit consume()/release()
  is used instead.

  2. Pool lifetime via shared_ptr + custom buffer deleter

  Arrow's PoolBuffer stores a raw MemoryPool* pointer internally. Buffers returned by
  ParquetChunkFile::Read(nbytes) use a custom deleter that captures shared_ptr<ArrowMemoryPool>,
  keeping the pool alive until all buffers are freed — even those held by Arrow IO threads after
  ParquetReaderWrap is destroyed.

  3. Global query_pool_mem_tracker — no UAF risk

  The pool is initialized with ExecEnv::GetInstance()->query_pool_mem_tracker(), a global singleton
  that is never destroyed. This eliminates the parent-chain UAF problem entirely: no need for
  shared_ptr<MemTracker>, no need for detach_mem_tracker(), no need for atomic.

  4. Member declaration order

  _arrow_pool is declared before _rb_batch, _batch, _reader in ParquetReaderWrap, so C++
  destruction order guarantees the pool outlives internal Arrow objects.

  5. Defensive initialization

  ParquetChunkFile initializes _pool with a default ArrowMemoryPool (no tracker) in its
  constructor, preventing null dereference if set_memory_pool() is not called.

  6. Write path unaffected

  ParquetFileWriter continues using the default no-arg ArrowMemoryPool.



Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/file_scanner/parquet_scanner.cpp
+++ b/be/src/exec/file_scanner/parquet_scanner.cpp
@@ -472,7 +472,8 @@ Status ParquetScanner::open_next_reader() {
         _conv_ctx.current_file = file->filename();
         auto parquet_file = std::make_shared<ParquetChunkFile>(file, 0, _counter);
         auto parquet_reader = std::make_shared<ParquetReaderWrap>(std::move(parquet_file), _num_of_columns_from_file,
-                                                                  range_desc.start_offset, range_desc.size);
+                                                                  range_desc.start_offset, range_desc.size,
+                                                                  ExecEnv::GetInstance()->query_pool_mem_tracker());
         if (_scan_range.params.__isset.flexible_column_mapping && _scan_range.params.flexible_column_mapping) {
             parquet_reader->set_invalid_as_null(true);
         }
@@ -510,7 +511,8 @@ Status ParquetScanner::get_schema(std::vector<SlotDescriptor>* schema) {
     _conv_ctx.current_file = file->filename();
     auto parquet_file = std::make_shared<ParquetChunkFile>(file, 0, _counter);
     auto parquet_reader = std::make_shared<ParquetReaderWrap>(std::move(parquet_file), _num_of_columns_from_file,
-                                                              range_desc.start_offset, range_desc.size);
+                                                              range_desc.start_offset, range_desc.size,
+                                                              ExecEnv::GetInstance()->query_pool_mem_tracker());
     return parquet_reader->get_schema(schema);
 }
 
--- a/be/src/exec/parquet_reader.cpp
+++ b/be/src/exec/parquet_reader.cpp
@@ -29,7 +29,6 @@
 #include "parquet/schema.h"
 #include "parquet_schema_builder.h"
 #include "runtime/descriptors.h"
-#include "util/byte_buffer.h"
 
 namespace starrocks {
 // ====================================================================================================================
@@ -39,15 +38,19 @@ ParquetReaderWrap::~ParquetReaderWrap() {
 }
 
 ParquetReaderWrap::ParquetReaderWrap(std::shared_ptr<arrow::io::RandomAccessFile>&& parquet_file,
-                                     int32_t num_of_columns_from_file, int64_t read_offset, int64_t read_size)
+                                     int32_t num_of_columns_from_file, int64_t read_offset, int64_t read_size,
+                                     MemTracker* mem_tracker)
 
         : _num_of_columns_from_file(num_of_columns_from_file),
 
+          _arrow_pool(std::make_shared<ArrowMemoryPool>(mem_tracker)),
           _read_offset(read_offset),
           _read_size(read_size) {
     _parquet = std::move(parquet_file);
     _properties = ::parquet::ReaderProperties();
-    _filename = (reinterpret_cast<ParquetChunkFile*>(_parquet.get()))->filename();
+    auto* chunk_file = reinterpret_cast<ParquetChunkFile*>(_parquet.get());
+    _filename = chunk_file->filename();
+    chunk_file->set_memory_pool(_arrow_pool);
 }
 
 Status ParquetReaderWrap::next_selected_row_group() {
@@ -107,7 +110,7 @@ Status ParquetReaderWrap::_init_parquet_reader() {
         arrow_reader_properties.set_cache_options(cache_options);
 
         // new file reader for parquet file
-        auto st = ::parquet::arrow::FileReader::Make(arrow::default_memory_pool(),
+        auto st = ::parquet::arrow::FileReader::Make(_arrow_pool.get(),
                                                      ::parquet::ParquetFileReader::Open(_parquet, _properties),
                                                      arrow_reader_properties, &_reader);
         if (!st.ok()) {
@@ -375,7 +378,7 @@ using ArrowStatus = ::arrow::Status;
 
 ParquetChunkFile::ParquetChunkFile(std::shared_ptr<starrocks::RandomAccessFile> file, uint64_t pos,
                                    ScannerCounter* counter)
-        : _file(std::move(file)), _pos(pos), _counter(counter) {}
+        : _file(std::move(file)), _pos(pos), _counter(counter), _pool(std::make_shared<ArrowMemoryPool>()) {}
 
 ParquetChunkFile::~ParquetChunkFile() {
     [[maybe_unused]] auto st = Close();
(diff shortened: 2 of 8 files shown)


CelerData-Reviewer:
@codex review

be/src/exec/parquet_reader.h
```diff
@@ -111,6 +115,8 @@ class ParquetReaderWrap {
     std::string _filename;
 
     bool _invalid_as_null{false};
+
+    ArrowMemoryPool _arrow_pool;
```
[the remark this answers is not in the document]
xiangguangyxg: fixed

be/src/exec/file_scanner/parquet_scanner.cpp
```diff
@@ -472,7 +472,8 @@ Status ParquetScanner::open_next_reader() {
         _conv_ctx.current_file = file->filename();
         auto parquet_file = std::make_shared<ParquetChunkFile>(file, 0, _counter);
         auto parquet_reader = std::make_shared<ParquetReaderWrap>(std::move(parquet_file), _num_of_columns_from_file,
-                                                                  range_desc.start_offset, range_desc.size);
+                                                                  range_desc.start_offset, range_desc.size,
```
stdpain, line 475: Can we ensure that the instance's lifecycle is valid?

xiangguangyxg: Yes, the lifecycle is safe. Here's the ownership chain:                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                             
```
  FragmentContext (shared_ptr<RuntimeState>)                                                   
    └─ RuntimeState (shared_ptr<MemTracker> _instance_mem_tracker)
         └─ [outlives]
              ParquetScanner (raw RuntimeState* _state)
                └─ _curr_file_reader (shared_ptr<ParquetChunkReader>)
                      └─ _parquet_reader (shared_ptr<ParquetReaderWrap>)
                            └─ ArrowMemoryPool (raw MemTracker* _mem_tracker)
```

  Destruction order:
  1. ParquetScanner::close() calls _curr_file_reader.reset(), which destroys ParquetReaderWrap (and its ArrowMemoryPool)
  2. ParquetScanner itself is destroyed when the scan node closes
  3. RuntimeState (and its _instance_mem_tracker) is still alive, held by FragmentContext via shared_ptr

  So _instance_mem_tracker always outlives ParquetReaderWrap. This is consistent with other places in the codebase that pass _state->instance_mem_tracker() as a raw pointer (e.g. RETURN_IF_LIMIT_EXCEEDED macro in runtime_state.h).

stdpain: ParquetChunkFile::Read is executed in a separate thread. How is the destruction order guaranteed?

xiangguangyxg: Good question. The destruction order guarantees safety through two mechanisms:

  1. Member destruction order (reverse of declaration):
```
  _reader   destroyed first  → FileReader + internal ReadRangeCache cleaned up
  _batch    destroyed second → Arrow Buffers call pool->Free()
  _rb_batch destroyed third  → same
  _arrow_pool destroyed fourth → pool is now gone (but no one references it)
  _parquet  destroyed last   → ParquetChunkFile closed
```

  _reader (parquet::arrow::FileReader) is destroyed before _arrow_pool. Arrow's FileReader destructor tears down its internal ReadRangeCache, which waits for / cancels all pending async IO futures before returning. After _reader is destroyed, no IO thread will call ParquetChunkFile::Read() anymore.

  2. The old code had the same async-IO concern but was actually worse: the previous implementation called CurrentThread::mem_tracker() inside ParquetChunkFile::Read() — on Arrow IO threads where tls_mem_tracker is nullptr, this would return an error. The new approach eliminates the thread-local dependency entirely: _pool is a stable pointer set once in the constructor and valid for the pool's entire lifetime.

stdpain: The current reasoning assumes that FileReader destruction guarantees that
all async IO has fully completed, but Arrow only provides best-effort
cancellation and does not strictly guarantee that already running tasks
have finished.

This is important because buffers allocated via `arrow::AllocateBuffer(..., _pool)`
do not hold ownership of the MemoryPool. If any async IO or decode task
still holds such buffers after `_arrow_pool` is destroyed, their destruction
would call `pool->Free()` on a freed pool, leading to a potential UAF.

Given this, the lifetime of `_arrow_pool` does not seem strictly bounded by
the lifetime of `_reader` or `_batch`.

stdpain:
```
Time →
────────────────────────────────────────────────────────────

Thread A (destructor)              Thread B (IO thread / pool)

FileReader::~FileReader()
    │
    │ destroy ReadRangeCache
    │
    ├── future.Cancel()  ───────────────┐
    │                                  │
    │                                  │  (cancellation is cooperative)
    │                                  │
    │                          [task already running]
    │                                  │
    │                                  ├── ParquetChunkFile::Read()
    │                                  │       │
    │                                  │       ├── AllocateBuffer(..., _pool)
    │                                  │       │
    │                                  │       └── use buffer
    │                                  │
    │
    ├── ~_batch()  → Buffer ref--       │
    │                                  │
    ├── ~_arrow_pool()  ❗              │
    │                                  │
    │                                  ├── task continues running
    │                                  │
    │                                  ├── buffer destructs
    │                                  │       │
    │                                  │       └── pool->Free()  UAF
    │                                  │
────────────────────────────────────────────────────────────
```

xiangguangyxg: Cancellation is indeed a problem. let me fix it.

xiangguangyxg: Fixed by using shared_ptr

stdpain: If `mem_tracker` is an instance mem tracker, the same issue will occur. Please test thoroughly and consider all implications before submitting a PR.

Memtracker is a very important and sensitive module. Please be sure to perform thorough testing and validation.

stdpain (changes requested):
Please thoroughly test the changes to the mem tracker !

be/src/exec/file_scanner/parquet_scanner.cpp
xiangguangyxg: I extend MemTracker with shared-parent ownership while keeping the existing raw-parent API. For trackers that are already managed by shared_ptr, keep the parent chain alive through _parent_holder. And use the shared-parent path where appropriate, so the instance -> query -> workgroup/   query_pool -> process tracker chain stays alive during late Arrow buffer free.
Furthermore, a delayed release scenario was specifically constructed for testing.

stdpain: This is a risky change.
I suggest you add a separate “record-only” memory tracker to track the arrow's memory usage. Similar to the passthrough memory tracker.

xiangguangyxg: A separate “record-only” memory tracker is simple, but its capabilities are limited. Since Arrow might consume the vast majority of the process memory, I think it's necessary to limit its memory usage. 
BTW, although this change is risky, it also resolves the memory tracker's parent dependency issue, which has been encountered in other modules, so I think the risk is worthwhile.

stdpain: Making direct changes to the memory tracker requires careful consideration of many edge cases. I will not approve this PR until we have sufficient proof.

xiangguangyxg: Can you give some examples to explain what is sufficient proof so I can assess it

xiangguangyxg: Updated to use the global `query_pool_mem_tracker`, not make any change to the memory tracker, please review.

be/src/exec/parquet_reader.cpp
```diff
@@ -421,22 +424,23 @@ arrow::Result<int64_t> ParquetChunkFile::Tell() const {
...
     int64_t bytes_read_res = 0;
     ARROW_RETURN_NOT_OK(ReadAt(_pos, nbytes, read_buf->mutable_data()).Value(&bytes_read_res));
-    // If bytes_read is equal with read_buf's capacity, we just assign
+    // Custom deleter captures shared_ptr<ArrowMemoryPool>, keeping pool alive
+    // until the buffer is freed (even from Arrow IO threads after
+    // ParquetReaderWrap is destroyed).
+    auto deleter = read_buf.get_deleter();
+    auto shared_buf = std::shared_ptr<arrow::Buffer>(
+            read_buf.release(), [pool = _pool, deleter = std::move(deleter)](arrow::Buffer* buf) mutable {
+                (void)pool; // prevent pool from being released before buffer is freed
```
silverbullet233, line 436: Will the lifespan of the arrow memory pool always be longer than that of the memory tracker?

xiangguangyxg: memory tracker is a global memory tracker

be/src/exec/file_scanner/parquet_scanner.cpp
stdpain: Using the query pool memtracker is essentially no different from using the process memtracker.

wyb (dismissed):

be/src/exec/file_scanner/parquet_scanner.cpp
xiangguangyxg: Recording this memory usage in the query pool helps identify which module within the process is consuming it. Ensuring that the total memory usage of submodules matches the process memory usage.

be/src/formats/parquet/arrow_memory_pool.cpp
```diff
@@ -65,6 +70,9 @@ void ArrowMemoryPool::Free(uint8_t* buffer, int64_t size, int64_t /*alignment*/)
     }
     std::free(buffer);
     _stats.DidFreeBytes(size);
+    if (_mem_tracker != nullptr) {
+        _mem_tracker->release(size);
+    }
```
stdpain, line 90: Please thoroughly test this code.

stdpain (changes requested):

be/src/formats/parquet/arrow_memory_pool.cpp
stdpain: This code causes the process tracker to be released twice. `std::free` calls the process tracker release, and then the pool tracker is released once.

xiangguangyxg: Changed to use SCOPED_THREAD_LOCAL_MEM_TRACKER_SETTER

PR #70382, opened 2026-03-17
[BugFix] Make Iceberg MV refresh tolerate non-monotonic snapshot timestamps

LiShuMing:
## Why I'm doing:

Issue 1 — Unit mismatch in last_updated_at
When fetching the last_updated_at of an Iceberg partition from the Iceberg metatable fails, the code falls back to currentSnapshot.timestampMillis(). However, the unit of this fallback value is not aligned with the last_updated_at used in the normal code path — the expected unit is microseconds (Micro), not milliseconds (Millis).

Issue 2 — timestampMillis() is not strictly monotonically increasing
Log analysis reveals that currentSnapshot.timestampMillis() itself is not strictly increasing, which makes it unreliable as a change-detection signal. (See attached logs for reference.)

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

PR #70523, opened 2026-03-19
[BugFix] [Followup] Fix mv refresh bugs with expired snapshot iceberg partitions to avoid repeat refresh

LiShuMing:
## Why I'm doing:
This is a followup of 
https://github.com/StarRocks/starrocks/pull/70382

 When Iceberg snapshot expires, last_updated_snapshot_id becomes null. We cannot decide whetehr to refresh the partition or not.

before we only maintain one variable for each iceberg partition:
-  modified time

after:
- modified time
- fingerprint version which is calcuated by partition's stats

when modified time is not -1 which means it can be calculated from iceberg snapshot, we use it to comprare; otherwise we use the fingerprint version.

This can ensure iceberg mv will not refresh more or less even for expired snpashot partitions.
 

## What I'm doing:
    This commit fixes the issue where Iceberg materialized view refresh fails
    when the base table's snapshot has been expired.

    Problem:
    - When Iceberg snapshot expires, last_updated_snapshot_id becomes null
    - The previous code relied on snapshot ID/sequence number as partition version
    - This caused MV refresh to fail or produce incorrect results

    Solution:
    1. Use partition stats (record_count, file_count, total_data_file_size) to
       compute a fingerprint as partition version when snapshot is expired
    2. Fix PCTTableSnapshotInfo to properly handle external table partition info
    3. Enhance MetaFunctions to include MV name and to-refresh partitions in
       refresh info metadata

    Changes:
    - IcebergCatalog: Add computePartitionStatsFingerprint() method
    - PCTTableSnapshotInfo: Use BasePartitionInfo.fromExternalTable()
    - MetaFunctions: Add mvName and mvToRefreshPartitions to MVRefreshInfoMeta

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergCatalog.java
@@ -57,6 +57,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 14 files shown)


fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergCatalog.java
```diff
@@ -415,38 +430,37 @@ private long getPartitionLastUpdatedTime(IcebergTable icebergTable, StructLike r
...
-            Snapshot updateSnapshot = nativeTable.snapshot(lastUpdatedSnapshotId);
-            if (updateSnapshot != null) {
-                return updateSnapshot.sequenceNumber();
-            }
-            // The snapshot may already be expired from table metadata. Keep using the snapshot id as an opaque
-            // version token so refresh detection still observes snapshot changes even if the wall-clock timestamp
-            // is null or non-monotonic.
-            return lastUpdatedSnapshotId;
+        try {
+            long rc = getStatsColumnAsLong(row, recordCountIndex);
```
Youngwb, line 483: These indo is not enough, these only data files info, need info about delete file

Youngwb (dismissed):

Youngwb (dismissed):

stephen-shelby (approved):

PR #70609, opened 2026-03-20
[Enhancement] Add per-target-BE MIN/MAX for RPC metrics in load profile

luohaha:
## Why I'm doing:

In the load profile, `RpcClientSideTime`, `RpcServerSideTime`, and `RpcServerWaitFlushTime` are accumulated across all target BEs via `update()`/`fetch_add` for each RPC call. In multi-machine scenarios, the total can far exceed the wall clock time (e.g., 26 target BEs accumulate to 51s while the actual `SendDataTime` is only 8.5s), making it impossible to identify which receiver BE is the bottleneck.

## What I'm doing:

During the close phase of `TabletSinkSender` and `TabletSinkColocateSender`, iterate over each target BE's `AddBatchCounter` to compute per-target-BE MIN/MAX, and set them on the corresponding profile counters via `set_min()`/`set_max()`. The FE-side profile merge framework automatically displays these as `__MIN_OF_` / `__MAX_OF_` child counters.

**Changed files:**
- `tablet_sink_index_channel.h`: Fix `AddBatchCounter` client RPC time field naming (`client_prc_time_us` → `client_rpc_time_ns`) and accumulate it in `operator+=`
- `tablet_sink_index_channel.cpp`: Record each RPC latency into `AddBatchCounter` in `NodeChannel::_wait_request`
- `tablet_sink_sender.cpp`: Compute per-target-BE MIN/MAX for all three metrics during close phase
- `tablet_sink_colocate_sender.cpp`: Same changes for the colocate sender path

**Expected profile output after this change:**
```
- RpcClientSideTime: 51s434ms
  - __MIN_OF_RpcClientSideTime: 1s200ms
  - __MAX_OF_RpcClientSideTime: 3s800ms
- RpcServerSideTime: 48s17ms
  - __MIN_OF_RpcServerSideTime: 800ms
  - __MAX_OF_RpcServerSideTime: 3s500ms
```

**Risks:**
- MIN/MAX reflects the accumulated value per target BE (sum of all batches), not single RPC latency

## What type of PR is this:
- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:
- [ ] This is a backport PR

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/tablet_sink_index_channel.h
+++ b/be/src/exec/tablet_sink_index_channel.h
@@ -58,8 +58,8 @@ struct AddBatchCounter {
     int64_t add_batch_wait_lock_time_us = 0;
     // number of add_batch call
     int64_t add_batch_num = 0;
-    // total time of client rpc
-    int64_t client_prc_time_us = 0;
+    // total time of client rpc (in nanoseconds)
+    int64_t client_rpc_time_ns = 0;
     // total time of wait memtable flush
     int64_t add_batch_wait_memtable_flush_time_us = 0;
 
@@ -68,6 +68,7 @@ struct AddBatchCounter {
         add_batch_wait_lock_time_us += rhs.add_batch_wait_lock_time_us;
         add_batch_num += rhs.add_batch_num;
         add_batch_wait_memtable_flush_time_us += rhs.add_batch_wait_memtable_flush_time_us;
+        client_rpc_time_ns += rhs.client_rpc_time_ns;
         return *this;
     }
     friend AddBatchCounter operator+(const AddBatchCounter& lhs, const AddBatchCounter& rhs) {
--- a/be/src/exec/tablet_sink_index_channel.h
+++ b/be/src/exec/tablet_sink_index_channel.h
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include <limits>
 #include <memory>
 #include <queue>
 #include <set>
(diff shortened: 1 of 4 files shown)


banmoy (commented):
## Code Review

Nice observability improvement! The per-target-BE MIN/MAX for RPC metrics will make it much easier to identify bottleneck receivers. A few comments:

### 1. Bug: Zero-valued entries skew MIN to 0

Nodes that never actually send an `AddBatch` (no matching rows, early cancellation) still appear in `node_add_batch_counter_map` with zeroed counters. This makes `min_*` values report `0` instead of the true minimum among participating BEs.

**Suggestion:** Filter entries with `add_batch_num == 0`:

```cpp
for (auto const& pair : node_add_batch_counter_map) {
    // ... existing accumulation ...
    if (pair.second.add_batch_num > 0) {
        // min/max tracking here
    }
}
```

And guard the `set_min`/`set_max` calls on whether any entry actually participated.

### 2. Code duplication between sender and colocate_sender

The min/max computation block (~30 lines) is copy-pasted identically between `tablet_sink_sender.cpp` and `tablet_sink_colocate_sender.cpp`. If this logic needs to change (e.g., to fix issue #1), it must be updated in two places.

**Suggestion:** Extract a helper function, e.g.:

```cpp
void update_rpc_min_max_profile(const std::unordered_map<int64_t, AddBatchCounter>& counter_map,
                                 TabletSinkProfile* ts_profile);
```

### 3. Unused `#include <utility>`

Both sender files add `#include <utility>` but nothing new in the diff uses symbols from it. This appears unnecessary.

### 4. Minor: unit conversion comment

The struct now mixes `_us` and `_ns` suffixed fields (`add_batch_execution_time_us` vs `client_rpc_time_ns`). The `* 1000` conversions in the loop are correct, but a brief inline comment would help prevent future confusion.

---

Overall the PR is a useful improvement. The main actionable issue is **#1** — without filtering zero-entry nodes, the MIN counter will misleadingly show 0 in common multi-BE scenarios.

be/src/exec/tablet_sink_index_channel.h
```diff
@@ -100,6 +102,47 @@ struct TabletSinkProfile {
...
+        int64_t server_wait_flush_ns = pair.second.add_batch_wait_memtable_flush_time_us * 1000;
+        min_server_wait_flush_ns = std::min(min_server_wait_flush_ns, server_wait_flush_ns);
+        max_server_wait_flush_ns = std::max(max_server_wait_flush_ns, server_wait_flush_ns);
+
+        min_client_rpc_ns = std::min(min_client_rpc_ns, pair.second.client_rpc_time_ns);
+        max_client_rpc_ns = std::max(max_client_rpc_ns, pair.second.client_rpc_time_ns);
+    }
+
+    if (has_participating_node) {
+        ts_profile->server_rpc_timer->set_min(min_server_rpc_ns);
```
banmoy, line 137: AI-generated comment (GPT-5.4 / Cursor)

One thing I'm not fully convinced about is the metric semantics after profile merge.

`RpcClientSideTime`, `RpcServerSideTime`, and `RpcServerWaitFlushTime` are total per-sink counters, but the new `set_min()` / `set_max()` values are derived from per-target-BE accumulated totals.

For a single sink instance this is already a bit overloaded, but after FE-side profile merge it becomes more confusing:
- the main counter value represents the merged total across sink instances
- the MIN/MAX represent extrema of per-BE accumulated values inside each sink instance

So the counter value and its MIN/MAX no longer describe the same population.

Example:
- sink profile 1: per-BE client RPC totals = {2s, 8s}, total = 10s
- sink profile 2: per-BE client RPC totals = {1s, 9s}, total = 10s

After merge, the profile becomes:
- `RpcClientSideTime = 20s`
- `MIN = 1s`
- `MAX = 9s`

But `1s` / `9s` are not the min/max of the merged counter itself, they are min/max of a different per-BE distribution.

Would it be cleaner to expose these as dedicated child counters rather than attaching them as the native MIN/MAX of the total counter?

be/src/exec/tablet_sink_index_channel.h
```diff
@@ -100,6 +102,47 @@ struct TabletSinkProfile {
...
+    int64_t min_server_rpc_ns = std::numeric_limits<int64_t>::max();
+    int64_t max_server_rpc_ns = std::numeric_limits<int64_t>::min();
+    int64_t min_server_wait_flush_ns = std::numeric_limits<int64_t>::max();
+    int64_t max_server_wait_flush_ns = std::numeric_limits<int64_t>::min();
+    int64_t min_client_rpc_ns = std::numeric_limits<int64_t>::max();
+    int64_t max_client_rpc_ns = std::numeric_limits<int64_t>::min();
+    bool has_participating_node = false;
+
+    for (const auto& pair : counter_map) {
+        if (pair.second.add_batch_num == 0) {
```
banmoy, line 118: AI-generated comment (GPT-5.4 / Cursor)

I think there is still a correctness gap for `RpcClientSideTime` MIN/MAX in failure cases.

In `NodeChannel::_wait_request()`, the client RPC latency is recorded before checking `closure->cntl.Failed()` / result status, so a failed backend still contributes to the total `RpcClientSideTime`.

However, `update_rpc_min_max_profile()` only includes nodes with `add_batch_num > 0`. That means if a BE fails on its first `AddBatch`, its client RPC latency is counted in the total, but the same BE is excluded from the MIN/MAX calculation.

Example:
- BE-A succeeds, client RPC total = 100ms
- BE-B succeeds, client RPC total = 120ms
- BE-C spends 8s on the first RPC and then fails

The profile would show:
- `RpcClientSideTime = 8.22s`
- `__MIN_OF_RpcClientSideTime = 100ms`
- `__MAX_OF_RpcClientSideTime = 120ms`

So the slowest / failing backend disappears from the new diagnostic signal.

Could we either:
1. include failed nodes in the client-RPC MIN/MAX calculation, or
2. use a separate participation condition for client-side latency instead of `add_batch_num > 0`?

PR #70747, opened 2026-03-25
[Enhancement] Support schema evoluation for widening varchar length with key/non-key column

srlch:
## What I'm doing:
1. Support widening varchar length for key / sort key / distribution key / partition key in share nothing and share data for all table model.
3. Support regular and fast path widening varchar length.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -132,6 +132,7 @@ import com.starrocks.thrift.TTabletMetaType;
 import com.starrocks.thrift.TTaskType;
 import com.starrocks.thrift.TWriteQuorumType;
 import com.starrocks.type.ArrayType;
+import com.starrocks.type.PrimitiveType;
 import com.starrocks.type.StructField;
 import com.starrocks.type.StructType;
 import com.starrocks.type.Type;
@@ -147,6 +148,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -672,27 +674,43 @@ public class SchemaChangeHandler extends AlterHandler {
 
     // User can modify column type and column position
     private boolean processModifyColumn(ModifyColumnClause alterClause, OlapTable olapTable,
-                                        Map<Long, LinkedList<Column>> indexMetaIdToSchema) throws DdlException {
+                                        Map<Long, LinkedList<Column>> indexMetaIdToSchema,
+                                        Map<Long, Set<String>> alterIndexMetaIdToIncrVarcharLenColNames)
+            throws DdlException {
         // The fast schema evolution mechanism is only supported for modified columns in shared data mode.
+        // But fast schema evolution for widening varchar length can be performed in both shared data and share nothing.
         boolean fastSchemaEvolution = RunMode.isSharedDataMode() && olapTable.getUseFastSchemaEvolution();
         Column modColumn = buildColumnForModify(alterClause.getColumnDef(), olapTable);
         if (KeysType.PRIMARY_KEYS == olapTable.getKeysType()) {
-            if (olapTable.getBaseColumn(modColumn.getName()) != null && olapTable.getBaseColumn(modColumn.getName()).isKey()) {
-                throw new DdlException("Can not modify key column: " + modColumn.getName() + " for primary key table");
+            Column baseColumn = olapTable.getBaseColumn(modColumn.getName());
+            if (baseColumn != null) {
+                modColumn.setAggregationType(baseColumn.getAggregationType(), baseColumn.isAggregationTypeImplicit());
+                if (baseColumn.isKey()) {
+                    // backward compatibility: if user does not specify key attribute in modify column clause for pk table,
+                    // we will keep it as a key column which is the same as before.
+                    modColumn.setIsKey(baseColumn.isKey());
+                    if (!isVarcharLengthIncrease(baseColumn, modColumn)) {
+                        throw new DdlException(
+                            "Can not modify key column: " + modColumn.getName() +
+                                " for primary key table except for increasing varchar length");
+                    }
+                    if (modColumn.isAllowNull()) {
+                        throw new DdlException("primary key column[" + modColumn.getName() + "] cannot be nullable");
+                    }
+                }
             }
+
             MaterializedIndexMeta indexMeta = olapTable.getIndexMetaByMetaId(olapTable.getBaseIndexMetaId());
             if (indexMeta.getSortKeyIdxes() != null) {
                 for (Integer sortKeyIdx : indexMeta.getSortKeyIdxes()) {
                     if (indexMeta.getSchema().get(sortKeyIdx).getName().equalsIgnoreCase(modColumn.getName())) {
-                        throw new DdlException("Can not modify sort column in primary data model table");
+                        if (!isVarcharLengthIncrease(indexMeta.getSchema().get(sortKeyIdx), modColumn)) {
+                            throw new DdlException(
+                                "Can not modify sort column in primary data model table except for increasing varchar length");
+                        }
                     }
                 }
             }
-            if (modColumn.getAggregationType() != null && modColumn.getAggregationType() != AggregateType.REPLACE) {
-                throw new DdlException("Can not assign aggregation method on column in Primary data model table: " +
-                        modColumn.getName());
-            }
-            modColumn.setAggregationType(AggregateType.REPLACE, true);
         } else if (KeysType.AGG_KEYS == olapTable.getKeysType()) {
             if (modColumn.isKey() && null != modColumn.getAggregationType()) {
                 throw new DdlException("Can not assign aggregation method on key column: " + modColumn.getName());
@@ -846,12 +864,12 @@ public class SchemaChangeHandler extends AlterHandler {
 
         List<Column> otherIndexModifiedColumn = new ArrayList<>();
         // check if column being mod
+        List<Long> otherIndexMetaIds = new ArrayList<>();
         if (!modColumn.equals(oriColumn)) {
             // column is mod. we have to mod this column in all indices
 
             // handle other indices
             // 1 find other indices which contain this column
-            List<Long> otherIndexMetaIds = new ArrayList<>();
             for (Map.Entry<Long, List<Column>> entry : olapTable.getIndexMetaIdToSchema().entrySet()) {
                 if (entry.getKey() == indexMetaIdForFindingColumn) {
                     // skip the index we used to find column. it has been handled before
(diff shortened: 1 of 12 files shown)


gengjun-git:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java
```diff
@@ -992,7 +987,21 @@ public Void visitModifyColumnClause(ModifyColumnClause clause, ConnectContext co
         }
         try {
             if (table.isOlapOrCloudNativeTable() && ((OlapTable) table).getKeysType() == KeysType.PRIMARY_KEYS) {
-                columnDef.setAggregateType(AggregateType.REPLACE);
+                Column baseColumn = ((OlapTable) table).getBaseColumn(columnDef.getName());
+                if (baseColumn != null) {
```
wyb, line 996: When is `baseColumn` null? throw exception if `baseColumn` is null?

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -672,27 +674,40 @@ private void processDropField(DropFieldClause alterClause, OlapTable olapTable,
...
-                throw new DdlException("Can not modify key column: " + modColumn.getName() + " for primary key table");
+            Column baseColumn = olapTable.getBaseColumn(modColumn.getName());
+            if (baseColumn != null) {
+                modColumn.setAggregationType(baseColumn.getAggregationType(), baseColumn.isAggregationTypeImplicit());
+                if (baseColumn.isKey()) {
+                    // backward compatibility: if user does not specify key attribute in modify column clause for pk table,
+                    // we will keep it as a key column which is the same as before.
+                    modColumn.setIsKey(baseColumn.isKey());
+                    if (!isVarcharLengthIncrease(baseColumn, modColumn)) {
+                        throw new DdlException("Can not modify key column: " + modColumn.getName() + " for primary key table");
```
wyb: Key column k1 can only be modified the VARCHAR length for primary key table？

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -672,27 +674,40 @@ private void processDropField(DropFieldClause alterClause, OlapTable olapTable,
...
+                }
             }
+
             MaterializedIndexMeta indexMeta = olapTable.getIndexMetaByMetaId(olapTable.getBaseIndexMetaId());
             if (indexMeta.getSortKeyIdxes() != null) {
                 for (Integer sortKeyIdx : indexMeta.getSortKeyIdxes()) {
                     if (indexMeta.getSchema().get(sortKeyIdx).getName().equalsIgnoreCase(modColumn.getName())) {
-                        throw new DdlException("Can not modify sort column in primary data model table");
+                        if (!isVarcharLengthIncrease(indexMeta.getSchema().get(sortKeyIdx), modColumn)) {
+                            throw new DdlException("Can not modify sort column in primary data model table");
```
wyb: Same as above

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
srlch: Yes it is. Before this pr, if the ALTER TABLE MODIFY COLUMN touch the key or sort key of primary key table, it will always throw an exception without considering the type of the column modification. This pr only enable widening VARCHAR length but not other.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
srlch: Same as above

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java
srlch: `SchemaChangeHandler` will throw exception for `throw new DdlException("Column[" + newColName + "] does not exists")`, seems no need to throw exception here?

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
wyb: I think the error message should be clearer.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java
wyb: ok

fe/fe-core/src/test/java/com/starrocks/alter/SchemaChangeHandlerVarcharWidenTest.java
```diff
@@ -0,0 +1,436 @@
...
+            Assertions.assertTrue(exception.getMessage().contains("for colocate table"),
+                    exception.getMessage());
+            assertColumnType(tableName, tableName, "k1", PrimitiveType.VARCHAR, 10);
+            assertIndexColumns(tableName, tableName, "k1", "k2", "v1");
+        } finally {
+            GlobalStateMgr.getCurrentState().setColocateTableIndex(originalColocateIndex);
+        }
+    }
+
+    @Test
```
banmoy, line 219: AI-generated comment (GPT-5.4 / Cursor)

This new test block only covers reject paths for primary-key tables. Since this PR explicitly opens a positive path for widening `VARCHAR` length on PK key / sort-key columns, could we also add at least one passing test for a PK key column widen, and ideally one for a PK sort column widen? Right now the suite demonstrates what must fail, but not the newly allowed behavior.

test/sql/test_fast_schema_evolution/T/test_varchar_widen_native
```diff
@@ -0,0 +1,270 @@
...
+    v1 INT
+) ENGINE=OLAP
+DUPLICATE KEY(k1, k2)
+DISTRIBUTED BY HASH(k2) BUCKETS 1
+PROPERTIES (
+    "replication_num" = "1",
+    "fast_schema_evolution" = "true"
+);
+insert into t_short_slow values ("1", 1, 21);
+ALTER TABLE t_short_slow MODIFY COLUMN k1 INT KEY NOT NULL;
```
banmoy, line 298: AI-generated comment (GPT-5.4 / Cursor)

The PR description says this supports both the regular path and the fast path for `VARCHAR` widening, but the only slow-path case added here is `VARCHAR -> INT`. Could we add one explicit slow-path `VARCHAR(n) -> VARCHAR(m)` case as well? That would verify the new behavior itself still works when fast schema evolution is unavailable or disallowed.

test/sql/test_fast_schema_evolution/T/test_varchar_widen_native
```diff
@@ -0,0 +1,270 @@
...
+    k2 INT NOT NULL,
+    v1 INT
+) ENGINE=OLAP
+DUPLICATE KEY(k1, k2)
+DISTRIBUTED BY HASH(k2) BUCKETS 1
+PROPERTIES (
+    "replication_num" = "1",
+    "fast_schema_evolution" = "true"
+);
+function: assert_query_error_contains("ALTER TABLE t_multi_same MODIFY COLUMN k1 VARCHAR(30) KEY NOT NULL, MODIFY COLUMN k1 INT KEY NOT NULL", "Column[k1] does not exists")
```
banmoy, line 335: AI-generated comment (GPT-5.4 / Cursor)

We only cover failing multi-clause ALTERs in this section. Since the implementation now carries widen-specific state across `finalAnalyze()`, it would be good to add one successful mixed ALTER as well, for example a `VARCHAR` widen together with another non-conflicting column modification, so we can prove the exception still behaves correctly in a batch statement.

test/sql/test_fast_schema_evolution/T/test_varchar_widen_native
```diff
@@ -0,0 +1,270 @@
...
+DUPLICATE KEY(k1, k2)
+DISTRIBUTED BY HASH(k1) BUCKETS 1
+ROLLUP (
+    r1(k2, k1)
+)
+PROPERTIES (
+    "replication_num" = "1",
+    "fast_schema_evolution" = "true"
+);
+function: assert_query_error_contains("ALTER TABLE t_multi_cross MODIFY COLUMN v1 BIGINT, MODIFY COLUMN k1 INT KEY NOT NULL FIRST FROM r1", "modify distribution column[k1]")
```
banmoy, line 353: AI-generated comment (GPT-5.4 / Cursor)

A number of the negative SQL cases stop at `assert_query_error_contains(...)`, which verifies the error message but not the post-condition. Especially for these multi-clause cases, it would be safer to also assert that no schema change job was created and that the schema stayed unchanged after the statement fails.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -926,6 +941,52 @@ private boolean processModifyColumn(ModifyColumnClause alterClause, OlapTable ol
...
+            List<String> oldKeyColumns =
+                    olapTable.getBaseSchema().stream().
+                        filter(Column::isKey).map(Column::getName).collect(Collectors.toList());
+            List<String> newKeyColumns =
+                    schemaForFinding.stream().filter(Column::isKey).
+                        map(c -> Column.removeNamePrefix(c.getName())).collect(Collectors.toList());
+            keyOrderChanged = !oldKeyColumns.equals(newKeyColumns);
+        } // sort key need not be considered here, because modify column can not change the order of sort key
+
+        alterIndexMetaIdToIncrVarcharLenColNames.putIfAbsent(indexMetaIdForFindingColumn, Sets.newHashSet());
```
banmoy, line 961: AI-generated comment (GPT-5.4 / Cursor)

Non-blocking, but this widening special case is getting a bit hard to reason about because the exception is represented as mutable cross-phase state that is built here and then consumed later in `finalAnalyze()`. The behavior may be correct, but it raises the maintenance cost since the partition/distribution/short-key relaxations are no longer localized to one decision point.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
banmoy: AI-generated comment (GPT-5.4 / Cursor)

Since this PR now allows one special case instead of rejecting all PK key / sort-key modifications uniformly, the error message here could be more explicit. Something like "only widening VARCHAR length is allowed for primary key columns" would be easier for users and future maintainers to understand than the current blanket "Can not modify ..." wording.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
srlch: Updated

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
srlch: Updated

srlch: Updated

fe/fe-core/src/test/java/com/starrocks/alter/SchemaChangeHandlerVarcharWidenTest.java
srlch: updated

test/sql/test_fast_schema_evolution/T/test_varchar_widen_native
srlch: updated

test/sql/test_fast_schema_evolution/T/test_varchar_widen_native
srlch: updated

test/sql/test_fast_schema_evolution/T/test_varchar_widen_native
srlch: updated

wyb (approved):

stephen-shelby (approved):

kangkaisen (approved):

gengjun-git (approved):

PR #70770, opened 2026-03-25
[Enhancement] Add sql_digest_exclude_db session variable to support c…

yangrong688:
…ross-database SQL digest aggregation

By default, SQL Digest includes the database name in its computation, causing the same SQL pattern executed in different databases to produce different digests. This makes it difficult to aggregate and analyze SQL patterns across databases.

This PR adds a new session variable `sql_digest_exclude_db` (default false) that when enabled, excludes the database name from the digest computation, allowing the same SQL pattern to produce identical digests regardless of which database it runs in. External catalog names are always preserved.

Changes:
- Add `sql_digest_exclude_db` session variable in SessionVariable
- Add `toSqlWithoutDb()` method in TableName
- Add `excludeDbFromDigest` option in FormatOptions
- Update AST2SQLVisitor to conditionally exclude db names
- Add overloaded methods in AstToSQLBuilder and ConnectProcessor
- Add unit tests for cross-database digest consistency
- Update en/zh documentation for sql_digest

Generated with [Claude Code](https://claude.ai/code) via [Happy](https://happy.engineering)

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/docs/en/administration/sql_digest.md
+++ b/docs/en/administration/sql_digest.md
@@ -98,3 +98,41 @@ WHERE digest IN (SELECT digest FROM top_sql);
 - Constant values in SQL will be normalized. For example, similar SQL statements with `WHERE a = 1` and `WHERE a = 2` will have the same Digest.
 - IN predicates will be normalized. For example, similar SQL statements with `IN (1,2,3)` and `IN (1,2)` will have the same Digest.
 - `LIMIT N` clauses will be normalized. For example, similar SQL statements with `LIMIT 10` and `LIMIT 30` will have the same Digest.
+
+## Cross-database Digest
+
+By default, SQL Digest includes the database name in the computation. This means the same SQL pattern executed in different databases will produce different Digests. For example:
+
+```SQL
+USE db1;
+SELECT * FROM t WHERE id = 1;
+-- Digest: aaa...
+
+USE db2;
+SELECT * FROM t WHERE id = 1;
+-- Digest: bbb... (different from above)
+```
+
+If you want the same SQL pattern to produce the same Digest regardless of which database it runs in, you can set the session variable `sql_digest_exclude_db` to `true`:
+
+```SQL
+SET sql_digest_exclude_db = true;
+```
+
+After enabling this option, the database name will be excluded from the Digest computation, allowing you to aggregate SQL patterns across different databases.
+
+```SQL
+USE db1;
+SELECT * FROM t WHERE id = 1;
+-- Digest: ccc...
+
+USE db2;
+SELECT * FROM t WHERE id = 1;
+-- Digest: ccc... (same as above)
+```
+
+:::note
+
+External catalog names are always included in the Digest regardless of this setting.
+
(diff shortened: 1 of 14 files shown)


docs/en/administration/sql_digest.md
```diff
@@ -98,3 +98,37 @@ WHERE digest IN (SELECT digest FROM top_sql);
...
+USE db1;
+SELECT * FROM t WHERE id = 1;
+-- Digest: ccc...
+
+USE db2;
+SELECT * FROM t WHERE id = 1;
+-- Digest: ccc... (same as above)
+```
+
+> **Note:** External catalog names are always included in the Digest regardless of this setting.
```
DanRoscigno:
```suggestion
:::note

External catalog names are always included in the Digest regardless of this setting.

:::
```

docs/zh/administration/sql_digest.md
```diff
@@ -94,3 +94,37 @@ WHERE digest IN (SELECT digest FROM top_sql);
...
+USE db1;
+SELECT * FROM t WHERE id = 1;
+-- Digest: ccc...
+
+USE db2;
+SELECT * FROM t WHERE id = 1;
+-- Digest: ccc...（与上面相同）
+```
+
+> **注意：** 无论是否启用此选项，外部 Catalog 名称始终会包含在 Digest 中。
```
DanRoscigno:
```suggestion
:::note

无论是否启用此选项，外部 Catalog 名称始终会包含在 Digest 中。

:::
```

eshishki:
@yangrong688 the use case (aggregating a SQL pattern across databases) is reasonable, but two things about the shape would be worth settling before this goes further:

1. **Make it a global config, not a session variable.** The existing `enable_sql_digest` / `isEnableSQLDigest` toggle only controls *whether* a digest is computed — two sessions never produce *different* digests for the same query. This PR changes the digest *content*, so with a session var two sessions would compute divergent digests for an identical query, which breaks the cross-session aggregation the feature is meant to enable. A normalization rule like this belongs in global FE config (mirroring `enable_sql_digest`).

2. **Normalize the qualifier as a whole rather than db-only.** `TableName.toSql()` only emits the catalog for external catalogs, so "exclude db, keep catalog" isn't really a deliberate choice — it just drops one level and leaves the catalog in for external tables. A cleaner model would be a single "qualifier scope" for the digest (e.g. table-only vs db.table vs catalog.db.table) so internal and external tables behave consistently, instead of a one-off `exclude_db`.

Worth a maintainer weighing in on whether cross-db digest aggregation is wanted at all before reworking, but if it is, those two changes would make it robust.

PR #70802, opened 2026-03-25
[Enhancement] Support materialization hints for CTEs

lxynov:
## Why I'm doing:

Implements https://github.com/StarRocks/starrocks/issues/66230. (cc @LiShuMing )

An additional benefit apart from those mentioned in https://github.com/StarRocks/starrocks/issues/66230:
- It allows the users to prevent inlining an CTE when the inlining will incur a too complex scalar expression. For complex deeply-nested CTEs like the below one, inlining will create a scalar expression that has an excessive number of nodes. 
```sql
WITH cte1 AS (
  SELECT
    ...,
    (CASE
      WHEN col < val1 THEN ..
      WHEN col >= val1 AND col < val2 THEN ..
      ...
      WHEN col >= valN-1 AND col >= valN THEN ..
      ELSE NULL
      END
    ) AS case_when_result_col1
  FROM tbl1
),
cte2 AS (
  SELECT
    ...,
    (CASE
      WHEN case_when_result_col = 'aa' AND col = 'xx' THEN ..
      WHEN case_when_result_col = 'bb' AND col = 'xx' THEN ..
      ...
      ELSE NULL
      END
    ) AS case_when_result_col2
  FROM cte1
),
...
SELECT ... FROM cteN
```

## What I'm doing:
Implements https://github.com/StarRocks/starrocks/issues/66230.

Allow users to control CTE materialization strategy via SQL syntax hints:
```sql
WITH cte AS (SELECT * FROM t0) [materialized]
SELECT v1 FROM cte UNION ALL SELECT v2 FROM cte;

WITH cte AS (SELECT rand() AS r, v1 FROM t0) [not_materialized]
SELECT r FROM cte UNION ALL SELECT r FROM cte;
```
- `[materialized]` forces the CTE to be materialized regardless of reference count and cost evaluation.
- `[not_materialized]` forces inlining regardless of whether it contains non-deterministic functions, and cost evaluation.
- Without a hint, existing behaviors are retained.

Please note that, the hint takes precedence over existing behaviors. For example, `WITH cte (SELECT rand() AS r, v1 FROM t0) [not_materialized]` will make the CTE inlined. The motivation is to allow users have full control over whether a CTE is materialized or inlined. 

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
@@ -6010,11 +6010,12 @@ public class AstBuilder extends com.starrocks.sql.parser.StarRocksBaseVisitor<Pa
         QueryRelation queryRelation = (QueryRelation) visit(context.queryRelation());
 
         CTERelation.CTEMaterializationHint hint = CTERelation.CTEMaterializationHint.NONE;
-        if (context.MATERIALIZED() != null) {
-            if (context.NOT() != null) {
-                hint = CTERelation.CTEMaterializationHint.NOT_MATERIALIZED;
-            } else {
+        if (context.bracketHint() != null) {
+            String hintText = ((Identifier) visit(context.bracketHint().identifier(0))).getValue().toLowerCase();
+            if (hintText.equals("materialized")) {
                 hint = CTERelation.CTEMaterializationHint.MATERIALIZED;
+            } else if (hintText.equals("not_materialized")) {
+                hint = CTERelation.CTEMaterializationHint.NOT_MATERIALIZED;
             }
         }
 
--- a/fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
@@ -6011,11 +6011,21 @@ public class AstBuilder extends com.starrocks.sql.parser.StarRocksBaseVisitor<Pa
 
         CTERelation.CTEMaterializationHint hint = CTERelation.CTEMaterializationHint.NONE;
         if (context.bracketHint() != null) {
-            String hintText = ((Identifier) visit(context.bracketHint().identifier(0))).getValue().toLowerCase();
+            List<Identifier> hintIdentifiers = visit(context.bracketHint().identifier(), Identifier.class);
+            if (hintIdentifiers.size() != 1) {
+                throw new ParsingException(
+                        "CTE hint must be a single [materialized] or [not_materialized]",
+                        createPos(context.bracketHint()));
+            }
+            String hintText = hintIdentifiers.get(0).getValue().toLowerCase();
             if (hintText.equals("materialized")) {
                 hint = CTERelation.CTEMaterializationHint.MATERIALIZED;
             } else if (hintText.equals("not_materialized")) {
                 hint = CTERelation.CTEMaterializationHint.NOT_MATERIALIZED;
+            } else {
+                throw new ParsingException(
+                        "Unknown CTE hint [" + hintText + "]. Use [materialized] or [not_materialized]",
+                        createPos(context.bracketHint()));
             }
         }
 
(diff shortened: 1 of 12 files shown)


fe/fe-grammar/src/main/antlr/com/starrocks/grammar/StarRocks.g4
```diff
@@ -2464,7 +2464,9 @@ groupingSet
     ;
 
 commonTableExpression
-    : name=identifier (columnAliases)? AS '(' queryRelation ')'
+    : name=identifier (columnAliases)? AS MATERIALIZED '(' queryRelation ')'
```
stephen-shelby: I wonder whether this would fit better as a CTE-scoped optimizer hint rather than new grammar.

For example, something along the lines of the existing hint style may feel more natural:

```sql
WITH cte AS (SELECT * FROM t0) [materialized]
SELECT v1 FROM cte UNION ALL SELECT v2 FROM cte;
```

or

```sql
WITH cte AS (SELECT rand() AS r, v1 FROM t0) [not_materialized]
SELECT r FROM cte UNION ALL SELECT r FROM cte;
```

The key reason is that `MATERIALIZED` / `NOT MATERIALIZED` here are really overriding the optimizer's inline/materialize decision, and can also bypass some existing protection rules, for example non-deterministic CTEs or `LIMIT`-without-`ORDER BY` cases. That makes it feel more like an optimizer directive than part of the SQL semantics itself.

Using a hint may also give us more flexibility for future evolution without turning this into a long-term syntax contract too early.

lxynov: Makes sense. Just created https://github.com/StarRocks/starrocks/pull/70802/commits/76e83562cb696d0c65a40156cc9288b83424ba9a. (todo: squash before merging)

Could you take another look?

Let me know whether it should override `ForceCTEReuseRule` or throw an error for `(SELECT rand() AS r, v1 FROM t0) [not_materialized]`. I thought about that and decided to override so that users have total control. But it also makes sense to throw an error.

fe/fe-core/src/main/java/com/starrocks/sql/parser/AstBuilder.java
```diff
@@ -6008,15 +6008,26 @@ public ParseNode visitQueryRelation(com.starrocks.sql.parser.StarRocksParser.Que
...
         QueryRelation queryRelation = (QueryRelation) visit(context.queryRelation());
-        // Regenerate cteID when generating plan
+
+        CTERelation.CTEMaterializationHint hint = CTERelation.CTEMaterializationHint.NONE;
+        if (context.bracketHint() != null) {
+            String hintText = ((Identifier) visit(context.bracketHint().identifier(0))).getValue().toLowerCase();
+            if (hintText.equals("materialized")) {
+                hint = CTERelation.CTEMaterializationHint.MATERIALIZED;
+            } else if (hintText.equals("not_materialized")) {
+                hint = CTERelation.CTEMaterializationHint.NOT_MATERIALIZED;
```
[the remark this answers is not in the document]
lxynov: https://github.com/StarRocks/starrocks/pull/70802/commits/ec3b8a9d40bc9817e01e8c2a14d36197819bdfc7

fe/fe-grammar/src/main/antlr/com/starrocks/grammar/StarRocks.g4
stephen-shelby: yes,  it should throw an error.

stephen-shelby (approved):

trueeyu (approved):

trueeyu:

Issue #70810, opened 2026-03-26
`getQueryStatisticsColumnType` throws SemanticException for columns with dots in their names


smithaps89:
### Steps to reproduce the behavior (Required)

1. Create a table with a dot-containing column name:

```sql
CREATE TABLE test_db.dot_column_test (
    id INT,
    `customer.is_verified_email` BOOLEAN
)
ENGINE=OLAP
DUPLICATE KEY(id)
DISTRIBUTED BY HASH(id) BUCKETS 1
PROPERTIES ("replication_num" = "1");
```

2. Insert data:

```sql
INSERT INTO test_db.dot_column_test VALUES (1, true), (2, false), (3, NULL);
```

3. Verify the column is queryable:

```sql
SELECT `customer.is_verified_email` FROM test_db.dot_column_test LIMIT 10;
-- Works correctly, returns results
```

4. Wait for automatic statistics collection to trigger (or manually trigger with `ANALYZE TABLE test_db.dot_column_test;`), then observe FE logs:

```
com.starrocks.sql.analyzer.SemanticException: Getting analyzing error. Detail message: Unknown column 'customer.is_verified_email' in 'dot_column_test'.
```

The error repeats continuously as the `ColumnBasicStatsCacheLoader` retries on every cache refresh cycle.

### Expected behavior (Required)

Automatic statistics collection should succeed for columns with dots in their names. The method `StatisticUtils.getQueryStatisticsColumnType()` should resolve `customer.is_verified_email` as a literal column name and return its type (`BOOLEAN`), the same way the sibling method `StatisticUtils.quoting(Table, String)` already does.

### Real behavior (Required)

`getQueryStatisticsColumnType()` splits the column name on `.` and looks up `parts[0]` (`"customer"`) as a column name, assuming struct field access. Since no column named `customer` exists, it throws:

```
SemanticException: Getting analyzing error. Detail message: Unknown column 'customer.is_verified_email' in 'dot_column_test'.
```

This error is thrown from the `ColumnBasicStatsCacheLoader` → `StatisticExecutor.queryStatisticSync()` → `StatisticUtils.getQueryStatisticsColumnType()` path. The cache loader catches the `RuntimeException` and retries on the next refresh cycle, producing a continuous error storm. In production we observed **533,000+ errors in 24 hours** from a single table with 8 dot-containing columns (`customer.*`, `loyaltyProgram.*`, `attributes.*`, `zoominfo.*`).

[...]

PR #70811, opened 2026-03-26
[BugFix] Attempt to get the column name for statistics before trying to split by dot (backport #71301)

smithaps89:
## Why I'm doing:

`StatisticUtils.getQueryStatisticsColumnType()` fails with `SemanticException: Unknown column` when processing columns whose names contain literal dots (e.g., `` `customer.is_verified_email` ``). The method immediately splits the column name on `.` and treats the first segment as a column name (assuming struct field access), but never checks if the full dot-containing string is itself a column name. This causes a continuous error storm from the statistics cache loader (`ColumnBasicStatsCacheLoader`), as every cache refresh attempt for such columns throws and retries indefinitely. In production we observed **533,000+ errors in 24 hours** from a single table with 8 dot-containing columns.

## What I'm doing:

Added a direct column lookup using the full column name **before** the dot-split logic in `getQueryStatisticsColumnType()`. If the full name (dots included) matches an existing column, return its type immediately. Otherwise, fall through to the existing struct field access resolution — no behavioral change for struct columns. This is consistent with the sibling method `quoting(Table, String)` in the same file which already performs this check.

Fixes https://github.com/StarRocks/starrocks/issues/70810

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [x] 3.5
  - [x] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/statistic/StatisticUtils.java
+++ b/fe/fe-core/src/main/java/com/starrocks/statistic/StatisticUtils.java
@@ -608,6 +608,10 @@ public class StatisticUtils {
     }
 
     public static Type getQueryStatisticsColumnType(Table table, String column) {
+        Column directMatch = table.getColumn(column);
+        if (directMatch != null) {
+            return directMatch.getType();
+        }
         String[] parts = column.split("\\.");
         Preconditions.checkState(parts.length >= 1);
         Column base = table.getColumn(parts[0]);


eshishki:
@smithaps89 this targets the same bug as #71301 (`getQueryStatisticsColumnType` splitting on `.` and failing for columns with dotted names). #71301 covers it more fully — it handles the nested-struct case where the base column name itself contains dots, and it's in active review with @stephen-shelby. Since this one is the narrower fix and would conflict, suggest closing it in favor of #71301.

PR #70929, opened 2026-03-29
[BugFix] Fix stale project output columns after GLM split projection

stephen-shelby:
## Why I'm doing:
Fix a planner failure caused by stale output columns after GLM splits an embedded projection into a standalone `PhysicalProjectOperator`.

In some cost-based GLM cases, the new project node inherited the child logical property, so its `outputColumns` no longer matched the projection keys. This could make a valid eq-join condition invisible to `PlanFragmentBuilder` and trigger `must be eq-join`.

## What I'm doing:
Update `SplitProjectionRewriter.splitProjection()` to rebuild logical properties separately for the child and the new project node.

The child now keeps the output columns of the operator after removing its embedded projection, while the new `PhysicalProjectOperator` keeps the output columns of the original projected row output. 

fix: https://github.com/StarRocks/StarRocksTest/issues/11178

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lazymaterialize/GlobalLateMaterializationRewriter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lazymaterialize/GlobalLateMaterializationRewriter.java
@@ -166,19 +166,22 @@ public class GlobalLateMaterializationRewriter {
                 // remove projection from the original operator and create a new one
                 PhysicalProjectOperator projectOperator = new PhysicalProjectOperator(
                         projection.getColumnRefMap(), projection.getCommonSubOperatorMap());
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lazymaterialize/GlobalLateMaterializationRewriter.java
stdpain, line 98: We may need to move the short-circuit check before the split projection.

silverbullet233 (approved):

kangkaisen (approved):

PR #70958, opened 2026-03-30
[Enhancement] support inline agg state

silverbullet233:
## Why I'm doing:

  In AggHashMap, each GROUP BY key maps to an AggDataPtr pointing to aggregate state memory allocated by HashTableKeyAllocator. Every state
  update during the build phase requires dereferencing this pointer to reach the actual state which is an extra memory indirection access.

  For simple aggregates like COUNT, the state fits entirely within the space of an AggDataPtr (8 bytes). We can store the state directly in the value slot itself, making updates inline and eliminating the pointer dereference.

## What I'm doing:
  - Added an inline aggregate-state path, so a single POD aggregate state no larger
  than 8 bytes can be stored directly in the hash map value slot instead of allocating state memory from
  `HashTableKeyAllocator`.
  - Extended `AggHashMap` build paths (`build_hash_map`, selection/limit variants, and result iteration) to support inline
  state, including correct initialization for zero-initialized aggregates such as `COUNT`/`SUM` and non-zero-initialized
  aggregates such as `MIN`/`MAX`.
  - Wired the optimization into `Aggregator`, fusing hash-map insertion/probe with per-row aggregate update when inline
  mode is enabled, and falling back to the existing path for ineligible cases.
  - Restricted inline state to safe hash map variants and execution paths. In particular, unsupported layouts such as
  `SmallFixedSizeHashMap`, stream aggregation, two-level maps, and spill/output iterator paths continue to use the normal
  non-inline state flow or receive dedicated compatibility handling.
  - Added the session variable `enable_inline_agg_state` to control the optimization.

## Test
env: 1 FE + 1 BE(16c, 32g)
data set: ssb_100g

```sql
select lo_custkey,count() from lineorder group by lo_custkey order by count() limit 10
```
  | Metric | Baseline | After Optimization |
  | --- | ---: | ---: |
  | Query Time | 2.3s | 1.7s |
  | AggComputeTime | 1.948s | 1.393s |
  | AggFuncComputeTime | 557.024ms | 0ns |
  | HashTableMemoryUsage | 791.875 MB | 544.000 MB |

```sql
select lo_custkey,max(lo_orderkey) from lineorder group by lo_custkey order by max(lo_orderkey) limit 10;
```

  | Metric | Baseline | After Optimization |
  | --- | ---: | ---: |
  | Query Time | 2.6s | 2.09s |
  | AggComputeTime | 2.157s | 1.642s |
  | AggFuncComputeTime | 684.988ms | 0ns |
  | HashTableMemoryUsage | 667.938 MB | 544.000 MB |

The AggFuncComputeTime drops to 0 because, in the inline agg state path, aggregate update/merge is no longer executed in the separate agg-function compute stage. Instead, it is fused directly into the hash map build/probe path, so the work is accounted for inside AggComputeTime rather than AggFuncComputeTime.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


dirtysalt (dismissed):

silverbullet233:
@mergify rebase

PR #69434, opened 2026-02-24
[Enhancement] Bypass caching in CachingIcebergCatalog when vended credentials are enabled

Youngwb:
## What I'm doing:
This pull request introduces improvements to the caching logic for Iceberg REST catalogs, specifically addressing scenarios where vended credentials are used. The main goal is to ensure that table caching is bypassed when vended credentials are enabled, preventing issues with expired credentials. The changes include updates to caching logic, interface extensions, and new tests to verify correct behavior.

Caching logic enhancements:

* Updated `CachingIcebergCatalog.getTable` to bypass caching when vended credentials are enabled, ensuring that credentials do not expire before cache TTL.

Iceberg catalog interface updates:

* Added the `isVendedCredentialsEnabled` method to the `IcebergCatalog` interface, allowing catalogs to indicate whether vended credentials are used.
* Implemented `isVendedCredentialsEnabled` in `IcebergRESTCatalog`, with proper initialization and handling in constructors. 

Testing improvements:

* Added unit tests in `CachingIcebergCatalogTest` to verify that caching is bypassed when vended credentials are enabled and that caching works normally when vended credentials are disabled.
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

PR #70988, opened 2026-03-30
[BugFix] Fix CachingIcebergCatalog reload() to refresh vended credentials without bypassing cache

bohuslav-broza:
## Why I'm doing:

PR #69434 introduced a cache bypass for Iceberg REST catalogs with vended credentials to prevent stale STS tokens from being served. While it solved the credential expiration issue, it caused **all `getTable()` calls to hit the REST catalog and Lake Formation directly**, bypassing the Caffeine cache entirely. Under real workloads this generates hundreds of Lake Formation `GetDataAccess` requests per minute, triggering AWS rate limiting (`Rate exceeded`) and causing query failures.

The root cause is a bug in `CachingIcebergCatalog.reload()`: it compares `metadataFileLocation` between the old cached Table and the freshly loaded one, and when they match (the common case), it **discards the fresh Table and returns the old one**. Since vended credentials (STS tokens with ~1h TTL) are embedded in the Table's FileIO, the cache keeps serving expired tokens.

## What I'm doing:

1. **Fix `reload()`**: Always return the freshly loaded Table from `delegate.getTable()`, ensuring vended credentials in the cache are always current. The `metadataFileLocation` comparison optimization is removed — it is unsafe for REST catalogs where the Table object carries short-lived credentials even when metadata hasn't changed.

2. **Revert cache bypass from PR #69434**: Remove the `!delegate.isVendedCredentialsEnabled()` condition from `getTable()`, restoring the Caffeine cache for vended-credential catalogs. The cache is essential for absorbing the request load from background processes (statistics manager, stats-cache-refresher, catalog refresh) and concurrent queries.

3. **Remove `isVendedCredentialsEnabled()`** from `IcebergCatalog` interface and `IcebergRESTCatalog` — no longer needed since the cache is not bypassed.

**Recommended configuration** for REST catalogs with vended credentials:
- `iceberg_meta_cache_ttl_sec` < credential TTL (e.g. `3400` for AWS LF 1-hour tokens)
- `iceberg_table_cache_refresh_interval_sec` = `600` (reduces reload frequency while keeping credentials fresh)

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

**Behavior change**: Iceberg REST catalogs with vended credentials will now use the Caffeine table cache (as they did before PR #69434). The cache `reload()` method now always returns fresh Table objects, ensuring credentials are refreshed on every reload cycle. This replaces the full cache bypass introduced by PR #69434.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
@@ -125,13 +125,8 @@ public class CachingIcebergCatalog implements IcebergCatalog {
                     @Override
                     public Table reload(IcebergTableName key, Table oldValue) {
                         try {
-                            BaseTable updateTable = 
+                            BaseTable updateTable =
                                     (BaseTable) delegate.getTable(new ConnectContext(), key.dbName, key.tableName);
-                            TableOperations newOps = updateTable.operations();
-                            TableOperations oldOps = ((BaseTable) oldValue).operations();
-                            if (oldOps.current().metadataFileLocation().equals(newOps.current().metadataFileLocation())) {
-                                return oldValue;
-                            }
                             return updateTable;
                         } catch (Exception e) {
                             LOG.warn("refresh table {}.{} failed", key.dbName, key.tableName, e);
@@ -234,17 +229,15 @@ public class CachingIcebergCatalog implements IcebergCatalog {
     public Table getTable(ConnectContext connectContext, String dbName, String tableName) throws StarRocksConnectorException {
         IcebergTableName icebergTableName = new IcebergTableName(dbName, tableName);
 
-        // do not cache if jwt or oauth2 is used AND it is a REST Catalog.
-        // do not cache if vended credentials are enabled, because credentials may expire before cache TTL.
-        boolean cacheAllowed = icebergProperties.isEnableIcebergTableCache() &&
-                (Strings.isNullOrEmpty(connectContext.getAuthToken()) || !(delegate instanceof IcebergRESTCatalog)) &&
-                !delegate.isVendedCredentialsEnabled();
-        if (!cacheAllowed) {
-            return delegate.getTable(connectContext, dbName, tableName);
-        }
-
         if (ConnectContext.get() == null || ConnectContext.get().getCommand() == MysqlCommand.COM_QUERY) {
             tableLatestAccessTime.put(icebergTableName, System.currentTimeMillis());
+        }
+
+        // do not cache if jwt or oauth2 is used AND it is a REST Catalog.
+        boolean cacheAllowed = icebergProperties.isEnableIcebergTableCache() &&
+                (Strings.isNullOrEmpty(connectContext.getAuthToken()) || !(delegate instanceof IcebergRESTCatalog));
+        if (!cacheAllowed) {
+            return delegate.getTable(connectContext, dbName, tableName);
         }
         try {
             if (shouldOnlyReadCache(connectContext)) {
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergCatalog.java
@@ -278,17 +278,6 @@ public interface IcebergCatalog extends MemoryTrackable {
         return new HashMap<>();
     }
 
-    /**
-     * Check if this catalog uses vended credentials for table access.
-     * When vended credentials are used, caching tables may cause issues
-     * because credentials expire before the cache TTL.
-     *
-     * @return true if vended credentials are enabled
-     */
-    default boolean isVendedCredentialsEnabled() {
-        return false;
-    }
-
     default String defaultTableLocation(ConnectContext context, Namespace ns, String tableName) {
         Map<String, String> properties = loadNamespaceMetadata(context, ns);
         String databaseLocation = properties.get(LOCATION_PROPERTY);
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/rest/IcebergRESTCatalog.java
@@ -87,7 +87,6 @@ public class IcebergRESTCatalog implements IcebergCatalog {
     private Map<String, String> restCatalogProperties;
     private final boolean nestedNamespaceEnabled;
     private final boolean viewEndpointsEnabled;
-    private final boolean enableVendedCredentials;
 
 
     public IcebergRESTCatalog(String name, Configuration conf, Map<String, String> properties) {
@@ -105,7 +104,7 @@ public class IcebergRESTCatalog implements IcebergCatalog {
         restCatalogProperties.put(CatalogProperties.FILE_IO_IMPL, IcebergCachingFileIO.class.getName());
         restCatalogProperties.put(CatalogProperties.METRICS_REPORTER_IMPL, IcebergMetricsReporter.class.getName());
 
-        this.enableVendedCredentials =
+        boolean enableVendedCredentials =
                 Boolean.parseBoolean(restCatalogProperties.getOrDefault(KEY_VENDED_CREDENTIALS_ENABLED, "true"));
         if (enableVendedCredentials) {
             restCatalogProperties.put("header.X-Iceberg-Access-Delegation", "vended-credentials");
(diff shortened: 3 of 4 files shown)


cstaud:
anything i could do to move forward with this?

willhoney7:
We're also seeing this issue and getting "Rate exceeded" errors from AWS and it's causing tons of query failures. This is a pretty big problem for us. Can we get some eyes on this? Anything we can do to help?

alvin-phoenix-ai:
@mergify update

Youngwb:
Thanks for tracking this down — the `reload()` fix plus reverting the cache bypass is the right direction, and folks are clearly hitting the LF rate-limit in production. Two areas of feedback to start with: the tests, and one small thing in `getTable()`.

### Tests

1. **`testReloadAlwaysReturnsFreshTable` is timing-based and will likely flake.** It drives the async reload with `Thread.sleep(1100)` + `Thread.sleep(300)`. That assumption tends to break under CI load, and this PR is heading to three backport branches where flakes multiply. Caffeine's `LoadingCache.refresh(key)` returns a `CompletableFuture`, so `cache.refresh(key).get()` (or wiring the cache with a same-thread / `directExecutor`) makes the reload deterministic and removes the sleeps entirely.

2. **The `metadataFileLocation()` stubs are now dead.** After the fix, `reload()` no longer reads `metadataFileLocation()`, so the two `Mockito.when(...metadataFileLocation()).thenReturn(sharedMetadataLocation)` lines don't exercise anything. Worth dropping them (or leaving a one-line comment that they only document the "same metadata location" scenario), otherwise they imply a code path that no longer exists.

3. **Coverage gap on the user-facing half of the fix.** Removing `testGetTableBypassCacheWhenVendedCredentialsEnabled` was correct (it asserted the now-reverted behavior), but nothing now asserts that a vended-credential `IcebergRESTCatalog` actually goes *through* the cache again — which is the part that fixes the rate-limiting. The new test uses a generic `@Mocked IcebergCatalog delegate`, not a REST catalog. A small test that a vended REST delegate is loaded only once across two `getTable()` calls would lock that behavior in.

### `tableLatestAccessTime.put` placement

The `.put` was moved above the `if (!cacheAllowed) return;` early-return, so it now records access times on the bypass path too (JWT/OAuth2 REST catalogs, and the case where the table cache is globally disabled) — i.e. for tables that never enter the `tables` cache. Since `refreshCatalog()` only iterates `tables.asMap().keySet()`, those entries are never read and only cleared on drop/rename, so `tableLatestAccessTime` grows unbounded for bypass-path catalogs. Keeping the `.put` after the early-return (where it sat before #69434) avoids the orphan entries. Minor, but easy to fix.

willhoney7:
We applied/tested the patch from PR `#70988` in production for an Iceberg REST catalog backed by AWS Glue/S3 Tables + Lake Formation vended credentials.

So far:
- The original Lake Formation `Rate exceeded` errors appear resolved.
- We are no longer seeing the request rate-limiting behavior after rolling the patched FE image.
- our current catalog settings are:
  - `iceberg_meta_cache_ttl_sec = 600`
  - `iceberg_table_cache_refresh_interval_sec = 60`

However, we're still seeing some S3 `ExpiredToken` failures.
- Queries are short, generally 1-3 seconds, so this does not look like credentials expiring during long planning or scan execution.
- Failures are tied to specific tables, suggesting stale credential-bearing metadata is being reused for those table entries.

By manually setting iceberg_meta_cache_ttl_sec to 0, then requerying the table, I was able to get the queries to work again. (then I set it back to 600)

---

I know very little about starrocks code itself, but this is what my buddy claude suggested as a potential root issue:

- `CachingIcebergCatalog` now refreshes cached `Table` objects correctly, but `IcebergMetadata` also has its own `tables` cache:

```java
private final Map<TableIdentifier, org.apache.iceberg.Table> tables = new ConcurrentHashMap<>();
```

- `IcebergMetadata.getTable()` reuses entries from that map before calling `icebergCatalog.getTable(...)`.
- For REST catalogs with vended credentials, those cached native Iceberg `Table` objects can retain expired FileIO credentials outside the TTL/refresh behavior fixed in `CachingIcebergCatalog`.
- This may explain why only certain tables fail with `ExpiredToken` even though the queries are fast and the catalog cache TTL is low.


Could maintainers confirm whether `IcebergMetadata.tables` should be disabled, TTL-bound, or invalidated for REST catalogs with vended credentials? It seems like PR `#70988` fixes the Lake Formation rate-limit issue, but this outer `IcebergMetadata` native table cache may still preserve expired vended credentials.

Youngwb:
@willhoney7 thanks again for testing the patch and the detailed report — that was really helpful.

I've opened #75431, which supersedes this PR with a provider-agnostic fix. Beyond what #70988 did, it also fixes the **background-refresh path**, which is the part most likely relevant to the `ExpiredToken` you still saw: `refreshTableUnderLock()` was loading a fresh table every cycle but **discarding it whenever the snapshot was unchanged**, so the cache kept serving the old (expiring) vended credentials. #75431 swaps the reloaded table in so the cached credentials are actually renewed. It keeps caching (so the Lake Formation rate-limit fix stays) and deliberately does not try to parse per-cloud token expiry.

Could you test #75431 against your AWS Glue / S3 Tables + Lake Formation setup? Your config looks right — please keep `iceberg_meta_cache_ttl_sec` below the token lifetime (your `600` is good) and `iceberg_table_cache_refresh_interval_sec` at `60`/`600`.

If you still hit `ExpiredToken` on specific tables afterward, it'd help a lot to know:
- which tables, and whether they're queried frequently or only occasionally;
- whether the FE log shows `refresh table <db>.<tbl> failed` for those tables around the failure.

That would tell us whether anything else is holding onto an expired credential (e.g. a longer-lived `Table` reference outside this cache). Happy to keep iterating — let's continue on #75431.

willhoney7:
@Youngwb I appreciate your work on this! Thank you!

I took a deeper look at #75431. It looks like it should fix the original Lake Formation rate limiting, and it also fixes stale credentials inside `CachingIcebergCatalog`.

One thing I’m still concerned about is the outer native table cache in `IcebergMetadata`:

```java
private final Map<TableIdentifier, org.apache.iceberg.Table> tables = new ConcurrentHashMap<>();
```

From what I can tell, `IcebergMetadata.getTable()` can return a cached native Iceberg `Table` from that map before it calls into `icebergCatalog.getTable(...)`. For REST catalogs with Lake Formation vended credentials, that native `Table` can carry a `FileIO` with temporary credentials.

So even if #75431 keeps `CachingIcebergCatalog` fresh, this outer `IcebergMetadata.tables` map may still hold an expired credential-bearing `Table` object indefinitely and bypass the fixed inner cache path.

We applied our own follow-up patch yesterday trying to fix the ExpiredToken issue. The patch bypasses the outer `IcebergMetadata.tables` native table cache for REST catalogs with vended credentials, so queries always go through the TTL/refresh-aware `CachingIcebergCatalog` instead of reusing stale credential-bearing `Table` objects. Since then we haven't seen any expired token errors. 

Could you verify whether #75431 also needs to handle `IcebergMetadata.tables`? My current understanding is:

- #75431 fixes rate limiting.
- #75431 fixes stale credentials inside `CachingIcebergCatalog`.
- But it may not fix stale native `Table` objects retained by `IcebergMetadata.tables`.

I'll work to get your PR up on our dev cluster, but I'm reluctant to mess with our current production setup since we've been error free since our patch.

Youngwb:
@willhoney7 to explain what that map is: `IcebergMetadata.tables` is a small in-memory cache (a `ConcurrentHashMap`) that de-duplicates native Iceberg `Table` lookups **within a single statement**. During planning the same table can be resolved several times; this map lets the first lookup hit the catalog and the rest reuse the result. For a REST catalog, that cached native `Table` carries the vended-credential `FileIO`.

Its lifetime is per-statement, though: each statement gets its own `IcebergMetadata`, and the map is cleared right after planning (`StatementPlanner`'s `finally` calls `removeQueryMetadata()`); the next statement gets a fresh instance with an empty map. So from my reading of current `main`, it doesn't retain a credential-bearing `Table` across short queries, and two queries don't share it — the scan credentials come from the native `Table` resolved during that statement's own planning.

That's why I think #75431 should cover your case: it fixes the inner `CachingIcebergCatalog` so the native `Table` resolved at planning is fresh. Your bypass was tested without #75431, so it was working around the still-stale inner cache; #75431 fixes that at the source.

Could you give #75431 a try (without your own bypass patch) and see whether the `ExpiredToken` errors come back? That's the cleanest way to confirm whether the inner-cache fix is enough on its own. Totally fine to do it in a non-prod / low-risk window given you're stable right now.

PR #71034, opened 2026-03-31
[BugFix] Fix PK tablet rowset meta loss caused by GC race during disk re-migration (A→B→A) (backport #70727)

banmoy:
## Why I'm doing:

Backport of #70727 to branch-3.5.

During rapid PK tablet re-migration (disk A→B→A), a race between GC (`start_trash_sweep`) and migration can cause `_add_shutdown_tablet_unlocked` to eagerly remove the **new** tablet's rowset metadata from RocksDB via `_remove_tablet_meta()`, because a stale shutdown entry still references the old tablet on the same path. This leads to permanent rowset meta loss and tablet corruption after BE restart.

## What I'm doing:

Before calling `_remove_tablet_meta()` in `_add_shutdown_tablet_unlocked`, re-read the on-disk tablet meta and verify that the `tablet_uid` and `tablet_state` still match the stale entry. If they don't match (ownership has moved to a newer tablet instance), skip the destructive meta cleanup and let the normal shutdown queue handle it safely.

Also adds a regression test (`test_pk_migration_gc_race_preserves_new_tablet_meta`) that reproduces the exact race window using `TEST_SYNC_POINT`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [x] This is a backport PR

- [ ] I have checked the version labels which the PR will be auto-backported to the target branch
  - [ ] 3.5

### DIFF
--- a/be/src/storage/tablet_manager.cpp
+++ b/be/src/storage/tablet_manager.cpp
@@ -42,6 +42,7 @@
 #include <memory>
 
 #include "common/config.h"
+#include "testutil/sync_point.h"
 #include "exec/schema_scanner/schema_be_tablets_scanner.h"
 #include "fs/fs.h"
 #include "fs/fs_util.h"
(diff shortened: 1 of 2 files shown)


luohaha (approved):

trueeyu:

PR #71024, opened 2026-03-31
[Performance] Reduce lake_pk_compaction_max_input_rowsets from 500 to 200

luohaha:
## Why I'm doing:

The default value of `lake_pk_compaction_max_input_rowsets=500` causes excessive delvec remote IO during compaction publish in shared-data mode, leading to high write P99 latency.

## What I'm doing:

Reduce the default value from 500 to 200.

## Bottleneck Analysis

**Benchmark setup:** realtime-benchmark on shared-data cluster (1 FE 8c32g + 3 CN 16c64g), commit `87707ea` (main branch).

**Baseline metrics (steady state):**
| Metric | Value | Target |
|--------|-------|--------|
| Write P50 | 2713ms | < 1000ms |
| Write P99 | 8438ms | < 5000ms |
| Write throughput | ~26K rows/sec | > 100K rows/sec |

**Root cause:** During compaction publish, the `LakePrimaryKeyCompactionConflictResolver` loads delvecs for each unique segment from remote storage (S3/OSS). With 500 input rowsets, this causes 3-10 seconds of remote IO:

| input_rowsets_size | compaction_delvec_loader_latency | Total publish cost |
|-------------------|----------------------------------|-------------------|
| 500 | 4.8s - 9.6s | 4.0s - 6.0s |
| 64-84 | 54ms - 182ms | < 500ms |

**Code path:** `primary_key_compaction_conflict_resolver.cpp:86-88` → `lake_delvec_loader.cpp:43-58` performs sequential remote IO for each unique rssid.

## Expected Impact

- Write P99 should decrease by ~40-60% (capping worst-case publish time)
- Compaction runs more frequently with smaller batches — same throughput, lower latency
- No regression expected for read performance

## What type of PR is this:

- [x] Performance Improvement
- [ ] Bug Fix
- [ ] Feature
- [ ] Code Clean Up

PR #71038, opened 2026-03-31
[Performance] Parallelize delvec loading during compaction conflict resolution

luohaha:
## Summary

Parallelize delvec (delete vector) loading during compaction conflict resolution in the publish path for shared-data (lake) mode. This addresses the root cause of high write P99 latency without needing to reduce `lake_pk_compaction_max_input_rowsets`.

## Problem

During compaction publish, `PrimaryKeyCompactionConflictResolver` loads delvecs **sequentially** for each unique segment from remote storage (S3/OSS). With large compactions (500 input rowsets), this causes hundreds of sequential remote IO calls, leading to publish latencies of **3-10 seconds** that block the write path.

Evidence from realtime-benchmark CN publish traces:
| input_rowsets_size | compaction_delvec_loader_latency | Total publish cost |
|---|---|---|
| 500 | 9.6s | >6s |
| 500 | 4.8s | 6.0s |
| 64-84 | 54-182ms | <1s |

## Solution

- **Batch-collect** all unique rssids from the rows mapper data before processing
- **Parallel-load** delvecs using `std::async` when batch size exceeds 4
- Small batches (≤4) still load sequentially to avoid thread overhead
- Both `execute()` and `execute_without_update_index()` paths are optimized

## Why not just reduce max input rowsets?

PR #71024 tried reducing `lake_pk_compaction_max_input_rowsets` from 500 to 200. While this improved write P99 by 30%, it caused an **83% query QPS regression** because more frequent compaction cycles increased resource contention. This PR fixes the root cause (sequential remote IO) without changing compaction behavior.

## Expected Impact

- Write P99: ~8.4s → ~2-4s (parallel IO reduces wall-clock time by 3-5x for large batches)
- Write throughput: should improve proportionally
- Query performance: should NOT regress (compaction frequency unchanged)
- No config changes needed

## Test Plan

- [ ] Build and deploy to shared-data cluster
- [ ] Run realtime-benchmark with same parameters as baseline
- [ ] Verify `compaction_delvec_loader_latency_us` decreases in CN publish traces
- [ ] Verify query QPS does not regress vs baseline

### DIFF
--- a/be/src/storage/primary_key_compaction_conflict_resolver.cpp
+++ b/be/src/storage/primary_key_compaction_conflict_resolver.cpp
@@ -14,6 +14,8 @@
 
 #include "storage/primary_key_compaction_conflict_resolver.h"
 
+#include <future>
+
 #include "base/debug/trace.h"
 #include "common/config_exec_fwd.h"
 #include "runtime/current_thread.h"
@@ -25,6 +27,71 @@
 #include "storage/tablet_schema.h"
 
 namespace starrocks {
+
+// Batch-load delvecs for a set of rssids in parallel.
+// This avoids sequential remote IO (S3/OSS) calls when many delvecs need to be loaded
+// during compaction conflict resolution.
+static Status batch_load_delvecs(const CompactConflictResolveParams& params,
+                                 const std::vector<uint32_t>& rssids_to_load,
+                                 std::map<uint32_t, DelVectorPtr>* rssid_to_delvec) {
+    if (rssids_to_load.empty()) {
+        return Status::OK();
+    }
+
+    // For small batches, load sequentially to avoid thread overhead
+    static constexpr size_t kParallelThreshold = 4;
+    if (rssids_to_load.size() <= kParallelThreshold) {
+        for (uint32_t rssid : rssids_to_load) {
+            DelVectorPtr delvec_ptr;
+            RETURN_IF_ERROR(params.delvec_loader->load({params.tablet_id, rssid}, params.base_version, &delvec_ptr));
+            (*rssid_to_delvec)[rssid] = std::move(delvec_ptr);
+        }
+        return Status::OK();
+    }
+
+    // Load delvecs in parallel using async tasks
+    struct LoadResult {
+        uint32_t rssid;
+        DelVectorPtr delvec;
+        Status status;
+    };
+
+    std::vector<std::future<LoadResult>> futures;
+    futures.reserve(rssids_to_load.size());
+
+    for (uint32_t rssid : rssids_to_load) {
+        futures.push_back(std::async(std::launch::async, [&params, rssid]() -> LoadResult {
+            LoadResult result;
+            result.rssid = rssid;
+            result.status = params.delvec_loader->load({params.tablet_id, rssid}, params.base_version, &result.delvec);
+            return result;
+        }));
+    }
+
+    // Collect results
+    for (auto& future : futures) {
+        auto result = future.get();
+        if (!result.status.ok()) {
+            return result.status;
+        }
+        (*rssid_to_delvec)[result.rssid] = std::move(result.delvec);
+    }
+
+    return Status::OK();
+}
+
+// Extract unique rssids from rssid_rowids that are not already in the cache.
+static std::vector<uint32_t> collect_missing_rssids(const std::vector<uint64_t>& rssid_rowids,
+                                                     const std::map<uint32_t, DelVectorPtr>& rssid_to_delvec) {
+    std::set<uint32_t> unique_rssids;
+    for (const auto& rssid_rowid : rssid_rowids) {
+        uint32_t rssid = rssid_rowid >> 32;
+        if (rssid_to_delvec.count(rssid) == 0) {
+            unique_rssids.insert(rssid);
+        }
+    }
+    return {unique_rssids.begin(), unique_rssids.end()};
+}
 
 Status PrimaryKeyCompactionConflictResolver::execute() {
     Schema pkey_schema = generate_pkey_schema();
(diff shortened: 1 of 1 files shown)


luohaha (commented):
## Review (Auto-Perf-Opt Reviewer Agent)

**Verdict: REQUEST CHANGES**

The optimization idea is sound — parallelizing sequential delvec remote IO during compaction conflict resolution directly addresses the diagnosed bottleneck. However, the implementation has two significant issues that need to be fixed before merging.

---

### BLOCKING ISSUE 1: Unbounded thread creation

`batch_load_delvecs()` uses `std::async(std::launch::async)` which creates **one OS thread per delvec load**. With 500 input rowsets, the first chunk/segment can easily have hundreds of unique rssids, spawning hundreds of threads simultaneously.

**Problems:**
- OS thread creation overhead (stack allocation, kernel scheduling)
- No backpressure — can exhaust OS thread limits under heavy compaction load
- Multiple tablets compacting concurrently multiplies the problem

**File:** `primary_key_compaction_conflict_resolver.cpp`, lines ~60-68

**Suggested fix:** Either:
1. Use an existing StarRocks thread pool (e.g., `ExecEnv` compaction pool), or
2. Add a concurrency cap — batch the loads in groups of e.g. 16-32:
```cpp
static constexpr size_t kMaxParallel = 16;
for (size_t start = 0; start < rssids_to_load.size(); start += kMaxParallel) {
    size_t end = std::min(start + kMaxParallel, rssids_to_load.size());
    std::vector<std::future<LoadResult>> futures;
    futures.reserve(end - start);
    for (size_t i = start; i < end; i++) {
        futures.push_back(std::async(std::launch::async, [&params, rssid = rssids_to_load[i]]() -> LoadResult {
            // ... same as current
        }));
    }
    for (auto& future : futures) {
        auto result = future.get();
        RETURN_IF_ERROR(result.status);
        (*rssid_to_delvec)[result.rssid] = std::move(result.delvec);
    }
}
```

Option 2 is simpler and still gets most of the parallelism benefit (16 concurrent IO calls is ~16x speedup vs sequential).

---

### BLOCKING ISSUE 2: Missing memory tracker on async threads

StarRocks uses thread-local `CurrentThread` / `MemTracker` for memory accounting. The `std::async` threads will have **no memory tracker attached**, meaning:
- Memory allocations inside `load()` (metadata cache lookups, delvec deserialization) won't be tracked
- Compaction memory limits (`MemTrackerType::COMPACTION_TASK`) are bypassed
- Could cause OOM without triggering memory limit protection

**Suggested fix:** Capture the parent thread's `MemTracker` and attach it in the async lambda:
```cpp
auto* mem_tracker = tls_thread_status.mem_tracker();
futures.push_back(std::async(std::launch::async, [&params, rssid, mem_tracker]() -> LoadResult {
    SCOPED_THREAD_LOCAL_MEM_TRACKER_SETTER(mem_tracker);
    // ... rest of load logic
}));
```
Or use StarRocks' `ThreadPool::submit_func()` which handles this automatically.

---

### Non-blocking observations

1. **`collect_missing_rssids` is clean** — local `std::set`, returns by value, no concurrency concern. ✅

2. **`LakeDelvecLoader::load()` is thread-safe** — no mutable state after construction, `TabletManager` supports concurrent access, `_pk_builder` accessed via const pointer/method. ✅

3. **Tracing scope placement is correct** — wraps the full batch (collect + load), correctly measures wall-clock delvec loading time. ✅

4. **Both code paths (`execute()` and `execute_without_update_index()`) consistently updated.** ✅

5. **Sequential fallback for ≤4 rssids** avoids thread overhead for small batches. ✅

---

### Summary

The approach is correct and will significantly reduce publish latency for large compactions. Please fix the two blocking issues (bounded parallelism + memory tracking) and this is ready to merge.

luohaha:
Thanks for the detailed review! Both blocking issues are now fixed in the latest push:

**1. Bounded parallelism (BLOCKING ISSUE 1):**
- Added `kMaxParallel = 16` cap
- Delvec loads are now batched in groups of 16 — each batch completes before the next starts
- This provides ~16x speedup vs sequential while avoiding thread exhaustion

**2. Memory tracker (BLOCKING ISSUE 2):**
- Captured parent thread's `MemTracker` via `tls_thread_status.mem_tracker()`
- Attached it in each async lambda using `SCOPED_THREAD_LOCAL_MEM_TRACKER_SETTER(mem_tracker)`
- All allocations inside `load()` are now properly tracked against compaction memory limits

luohaha (commented):
## Review (Auto-Perf-Opt Reviewer Agent) — Follow-up

**Verdict: APPROVE** (submitted as comment — cannot self-approve)

Both blocking issues from the previous review are now properly fixed:

### Fix 1: Bounded parallelism ✅
- `kMaxParallel = 16` caps concurrent threads
- Batched loop processes groups of 16, collecting results before launching the next batch
- This provides ~16x speedup vs sequential while preventing thread exhaustion under heavy compaction

### Fix 2: Memory tracker propagation ✅
- `tls_thread_status.mem_tracker()` is captured before the parallel loop
- `SCOPED_THREAD_LOCAL_MEM_TRACKER_SETTER(mem_tracker)` is correctly applied inside each async lambda
- This macro is well-established in the StarRocks codebase (used in `async_flush_output_stream.cpp`, `connector_sink_executor.cpp`, etc.)
- Memory allocations in `load()` (metadata cache, delvec deserialization) are now properly tracked against compaction memory limits

### Code quality checks ✅
- `current_thread.h` is already included (line 19), providing `tls_thread_status` and the `SCOPED_THREAD_LOCAL_MEM_TRACKER_SETTER` macro
- `<future>` header correctly added for `std::async` / `std::future`
- Both `execute()` and `execute_without_update_index()` paths are consistently updated
- Sequential fallback for ≤4 rssids (`kParallelThreshold`) avoids thread overhead for small batches
- Error handling is correct: first failed status propagates immediately, remaining futures are cleaned up by destructors

No further issues. LGTM — ready to build and benchmark.

PR #71068, opened 2026-03-31
[Enhancement] Support create tablet retry for lake table in shared-data mode

starrocks-xupeng:
## What I'm doing:

When a CN is down or unreachable during CREATE TABLE, the RPC to create tablets on that node fails immediately. Previously this caused the entire CREATE TABLE to fail. Now, failed tasks are retried on an alternative alive CN.

Key changes:
- Add `lake_create_tablet_max_retries` config (default 1) to control retry
- In sendCreateReplicaTasks, send failures are handled per-node: mark tasks as failed and count down individually instead of countDownToZero, so other nodes' tasks can continue
- Add retryFailedTasks() to reassign failed tasks to random alive nodes (excluding previously failed nodes) and resend
- In buildCreateReplicaTasks, fallback to a random alive node when the assigned node is unavailable (retry enabled only)
- Fix waitForFinished to throw DdlException on InterruptedException instead of silently swallowing it
- Ensure AgentTaskQueue cleanup via finally blocks in all send+wait paths

Only send-phase failures (node down, RPC error) are retried. CN-reported errors still fast-fail via countDownToZero. Timeouts are not retried. When retry is disabled (config=0), behavior is identical to before.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/task/TabletTaskExecutor.java
```diff
@@ -304,45 +315,173 @@ private static List<CreateReplicaTask> buildCreateReplicaTasks(long dbId,
...
+        for (int retry = 0; retry < maxRetries && !failedTasks.isEmpty(); retry++) {
+            for (CreateReplicaTask t : failedTasks) {
+                excludeNodes.add(t.getBackendId());
+            }
+            LOG.info("retry #{} for {} failed create tablet tasks, excludeNodes={}",
+                    retry + 1, failedTasks.size(), excludeNodes);
+
+            final WarehouseManager warehouseManager = GlobalStateMgr.getCurrentState().getWarehouseMgr();
+            List<ComputeNode> aliveNodes = warehouseManager.getAliveComputeNodes(computeResource);
+            List<ComputeNode> candidates = aliveNodes.stream()
```
banmoy, line 401: if `excludeNodes` are in `aliveNodes`, maybe these nodes have already recovered, no need to exclude them?

fe/fe-core/src/main/java/com/starrocks/task/TabletTaskExecutor.java
```diff
@@ -304,45 +315,173 @@ private static List<CreateReplicaTask> buildCreateReplicaTasks(long dbId,
...
+     */
+    // Visible for testing
+    static long getNodeIdForTablet(WarehouseManager warehouseManager, ComputeResource computeResource, long tabletId) {
+        try {
+            return warehouseManager.getComputeNodeAssignedToTablet(computeResource, tabletId).getId();
+        } catch (Exception e) {
+            // When retry is disabled, propagate the exception immediately.
+            // When retry is enabled, fallback to a random alive node so that the task can be
+            // created and sent. If this node also fails, the retry mechanism will reassign it.
+            if (Config.lake_create_tablet_max_retries <= 0) {
```
banmoy, line 330: `Config.lake_create_tablet_max_retries` can be changed on the fly, and lead to inconsistent behavior between replica tasks

banmoy:
Have you verified it in a real scenario, such as  rolling upgrade?

fe/fe-core/src/main/java/com/starrocks/task/TabletTaskExecutor.java
starrocks-xupeng: the alive test might be delayed, because alive is tested every 10 seconds

starrocks-xupeng:
yes, tested with actual cluster

fe/fe-core/src/main/java/com/starrocks/task/TabletTaskExecutor.java
starrocks-xupeng: it's taken befor tablet creation for an entire table, so even if config changed, does not affect behavior for that table

banmoy: > it's taken befor tablet creation for an entire table

`getNodeIdForTablet` is called for each tablet, so each call may see different values? Seems not critical, but it let me think about the behavior if it happens

starrocks-xupeng: for this case, it's fine, because when we reach here, we haven't started the send create tablet task, so even if there is a config change, ddl error will be sent that short time period

starrocks-xupeng: this is not the place to get node id for sending task, there are 2 places

banmoy (approved):

kevincai (approved):

gengjun-git (approved):

PR #71083, opened 2026-03-31
[BugFix] Fix use-after-free in parallel segment/rowset loading on error path

xiangguangyxg:
## Why I'm doing:
```
branch-4.1.0-f-0330 RELEASE (build e1112f9 distro ubuntu arch aarch64)
query_id:00000000-0000-0000-0000-000000000000, fragment_instance:00000000-0000-0000-0000-000000000000, plan_node_id:-1
*** Aborted at 1774947540 (unix time) try "date -d @1774947540" if you are using GNU date ***
PC: @          0xe62ad40 starrocks::TabletReaderParams::~TabletReaderParams()
*** SIGSEGV (@0x19) received by PID 331484 (TID 0xf664ebdafac0) LWP(332276) from PID 25; stack trace: ***
    @     0xf665745053dc (/usr/lib/aarch64-linux-gnu/libc.so.6+0x853db)
    @         0x1366a984 google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @     0xf66575bab8f8 ([vdso]+0x8f7)
    @          0xe62ad40 starrocks::TabletReaderParams::~TabletReaderParams()
    @          0xfa62300 starrocks::lake::TabletWarmupManager::do_warmup_tablet(std::shared_ptr<starrocks::lake::TabletWarmupManager::WarmupContext> const&)
    @         0x1014a764 starrocks::ThreadPool::dispatch_thread()
    @         0x10140d6c starrocks::Thread::supervise_thread(void*)
    @     0xf66574500398 (/usr/lib/aarch64-linux-gnu/libc.so.6+0x80397)
    @     0xf66574569e9c (/usr/lib/aarch64-linux-gnu/libc.so.6+0xe9e9b)

branch-4.1.0-f-0330 RELEASE (build e1112f9 distro ubuntu arch aarch64)
query_id:00000000-0000-0000-0000-000000000000, fragment_instance:00000000-0000-0000-0000-000000000000, plan_node_id:-1
*** Aborted at 1774947699 (unix time) try "date -d @1774947699" if you are using GNU date ***
PC: @          0xf9f1720 starrocks::lake::Rowset::read(starrocks::Schema const&, starrocks::RowsetReadOptions const&)
*** SIGSEGV (@0xa3) received by PID 332870 (TID 0xf9c9f71efac0) LWP(334250) from PID 163; stack trace: ***
    @     0xf9caee9853dc (/usr/lib/aarch64-linux-gnu/libc.so.6+0x853db)
    @         0x1366a984 google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @          0xf9f1720 starrocks::lake::Rowset::read(starrocks::Schema const&, starrocks::RowsetReadOptions const&) ::_Result<starrocks::StatusOr<std::vector<std::shared_ptr<starrocks:<A0><C3>^^<F7><C9><F9>
    @          0xcee9dac std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*)
    @     0xf9caee9853dc (/usr/lib/aarch64-linux-gnu/libc.so.6+0x853db) ator<std::shared_ptr<starrocks::ChunkIterator> > >*)::{lambda()#2}>:<A0><C3>^^<F7><C9><F9>
    @         0x10149f84 starrocks::ThreadPool::dispatch_thread()
    @         0x10140d6c starrocks::Thread::supervise_thread(void*)
    @     0xf9caee980398 (/usr/lib/aarch64-linux-gnu/libc.so.6+0x80397)
    @     0xf9caee9e9e9c (/usr/lib/aarch64-linux-gnu/libc.so.6+0xe9e9b)
```

  CN crashes (SIGSEGV) in shared-data mode during warmup when `enable_load_segment_parallel=true`.
  The crash stacks point to `TabletReaderParams::~TabletReaderParams()` and `Rowset::read()`,
  both are typical null+offset dereferences caused by use-after-free.

  Root cause: when one parallel task returns an error, the function returns early **without waiting
  for remaining futures to complete**. Tasks still running in the thread pool then access dangling
  references to already-destroyed stack locals (`rs_opts`, `this`).

  Two defect sites:
  1. `TabletReader::get_segment_iterators` — parallel lambda captures `rs_opts` and `this` by reference via `[&, rowset]`. Early return on future error leaves pending tasks holding dangling references.
  2. `Rowset::load_segments` — parallel lambda captures `this` (Rowset*) via `[=]`. Early return lets `TabletReader::close()` destroy the Rowset while tasks are still running.

## What I'm doing:

  Add a `DeferOp` right after the futures container declaration in both functions. The `DeferOp`
  ensures all submitted futures are waited on before the function returns through **any** path
  (including `ASSIGN_OR_RETURN` serial-fallback failures and future error early-returns).

  This is the minimal fix — no change to lambda capture semantics, no change to control flow,
  just a scope guard that calls `future.wait()` on all valid futures at function exit.

  Also adds `BE_TEST`-only sync point hooks in the parallel lambdas and corresponding unit tests
  that inject errors to verify the DeferOp correctly waits for all tasks.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/tablet_reader.cpp
+++ b/be/src/storage/lake/tablet_reader.cpp
@@ -17,6 +17,8 @@
 #include <future>
 #include <utility>
 
+#include "base/testutil/sync_point.h"
+#include "base/utility/defer_op.h"
 #include "column/datum_convert.h"
 #include "common/config_ingest_fwd.h"
 #include "common/config_json_flat_fwd.h"
@@ -384,6 +386,16 @@ Status TabletReader::get_segment_iterators(const TabletReaderParams& params, std
     SCOPED_RAW_TIMER(&_stats.create_segment_iter_ns);
 
     std::vector<std::future<StatusOr<std::vector<ChunkIteratorPtr>>>> futures;
+    // The parallel tasks capture local variables and |this| by reference.
+    // Must wait for all tasks to complete before returning, otherwise
+    // dangling references to |rs_opts| and |this| cause use-after-free.
+    DeferOp wait_futures([&futures]() {
+        for (auto& future : futures) {
+            if (future.valid()) {
+                future.wait();
+            }
+        }
+    });
     for (auto& rowset : _rowsets) {
         if (params.rowid_range_option != nullptr && !params.rowid_range_option->contains_rowset(rowset.get())) {
             continue;
@@ -391,7 +403,17 @@ Status TabletReader::get_segment_iterators(const TabletReaderParams& params, std
 
         if (config::enable_load_segment_parallel) {
             auto task = std::make_shared<std::packaged_task<StatusOr<std::vector<ChunkIteratorPtr>>()>>(
-                    [&, rowset]() { return enhance_error_prompt(rowset->read(schema(), rs_opts)); });
+                    [&, rowset]() {
+#ifdef BE_TEST
+                        Status injected_st;
+                        TEST_SYNC_POINT_CALLBACK("TabletReader::get_segment_iterators::parallel_read",
+                                                 &injected_st);
+                        if (!injected_st.ok()) {
+                            return StatusOr<std::vector<ChunkIteratorPtr>>(injected_st);
+                        }
+#endif
+                        return enhance_error_prompt(rowset->read(schema(), rs_opts));
+                    });
 
             auto packaged_func = [task]() { (*task)(); };
             if (auto st = ExecEnv::GetInstance()->load_rowset_thread_pool()->submit_func(std::move(packaged_func));
--- a/be/src/storage/lake/tablet_reader.cpp
+++ b/be/src/storage/lake/tablet_reader.cpp
@@ -402,18 +402,16 @@ Status TabletReader::get_segment_iterators(const TabletReaderParams& params, std
         }
 
         if (config::enable_load_segment_parallel) {
-            auto task = std::make_shared<std::packaged_task<StatusOr<std::vector<ChunkIteratorPtr>>()>>(
-                    [&, rowset]() {
+            auto task = std::make_shared<std::packaged_task<StatusOr<std::vector<ChunkIteratorPtr>>()>>([&, rowset]() {
 #ifdef BE_TEST
-                        Status injected_st;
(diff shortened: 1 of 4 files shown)


be/src/storage/lake/tablet_reader.cpp
```diff
@@ -384,14 +386,32 @@ Status TabletReader::get_segment_iterators(const TabletReaderParams& params, std
     SCOPED_RAW_TIMER(&_stats.create_segment_iter_ns);
 
     std::vector<std::future<StatusOr<std::vector<ChunkIteratorPtr>>>> futures;
+    // The parallel tasks capture local variables and |this| by reference.
+    // Must wait for all tasks to complete before returning, otherwise
+    // dangling references to |rs_opts| and |this| cause use-after-free.
+    DeferOp wait_futures([&futures]() {
+        for (auto& future : futures) {
+            if (future.valid()) {
```
kevincai, line 394: this is not caused by this PR change, but may introduce infinite hang due to the thread pool shutdown behavior, the task can be submitted to the threadpool but never get executed because of shutdown. So the future is valid but the result can never be coming.

kevincai: may think of threadpool token to handle this cleanup of the tasks.

xiangguangyxg: I checked the `ThreadPool::shutdown()` implementation, it explicitly destroys all pending tasks via `_pop_and_cancel_tasks_in_queue()`. When a `std::packaged_task` is destroyed without execution, C++ standard guarantees a broken_promise is set on the shared state, making `future.wait()` return immediately (no hang).

kevincai (approved):

luohaha (approved):

PR #71098, opened 2026-03-31
[Enhancement] DPP file-skip pruning for Iceberg scans using runtime filter min/max vs manifest stats

osscm:
## Why I'm doing:

  When StarRocks executes a query with a Dynamic Partition Pruning (DPP) runtime filter, it already prunes partitions at the partition level. However, within a surviving partition,
  individual Iceberg data files may still be entirely outside the runtime filter's min/max range. Without this change, those files are opened and scanned unnecessarily, wasting IO and CPU.

  Iceberg manifests already carry per-file column-level min/max statistics (THdfsScanRange.min_max_values). This change leverages those stats to skip entire files before they are opened,
  purely at the scan operator level — no planner changes required.

## What I'm doing:

Fixes #71005

  Adds a file-skip optimization for Iceberg scans driven by DPP runtime filter min/max bounds:

  - Adds should_prune_by_rf_min_max_stats() virtual method to DataSource (default: no pruning).
  - Implements it in HiveDataSource: after parse_runtime_filters(), iterates over all probe-slot runtime filters that have a min/max sub-filter, looks up the corresponding file-level column
  stats from the scan range, and returns true if the file's [min, max] range is disjoint from the RF's [min, max] range. Handles integral, date, and floating-point types via
  type_dispatch_filter + RfMinMaxFileStatsChecker.
  - In ConnectorChunkSource::prepare(), evaluates should_prune_by_rf_min_max_stats() once and caches the result in _skip_by_rf_stats. If set, _read_chunk() returns EndOfFile immediately
  without opening the file.
  - Exposes min_raw(), max_raw(), and has_min_max() accessors on MinMaxRuntimeFilter for use by the checker.
  - Adds a RfMinMaxPrunedFiles runtime profile counter to track how many files are skipped.
  - Adds unit tests in be/test/connector/hive_connector_dpp_test.cpp covering integer, float, date, string (no-prune), all-null (no-prune), and always-true RF (no-prune) cases.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

- [x] 4.1
- [x] 4.0
- [ ] 3.5
- [ ] 3.4

### DIFF
--- a/be/src/connector/hive_connector.h
+++ b/be/src/connector/hive_connector.h
@@ -93,6 +93,7 @@ public:
 
     void get_split_tasks(std::vector<pipeline::ScanSplitContextPtr>* split_tasks) override;
     Status _init_chunk_if_needed(ChunkPtr* chunk, size_t n) override;
+    bool should_prune_by_rf_min_max_stats() const override;
 
 private:
     const HiveDataSourceProvider* _provider;
--- a/be/src/connector/hive_connector.h
+++ b/be/src/connector/hive_connector.h
@@ -21,7 +21,9 @@
 #include "connector/connector.h"
 #include "exec/connector_scan_node.h"
 #include "exec/hdfs_scanner/hdfs_scanner.h"
+#include "exprs/runtime_filter.h"
 #include "hive_chunk_sink.h"
+#include "types/logical_type_infra.h"
 
 namespace starrocks {
 class HiveTableDescriptor;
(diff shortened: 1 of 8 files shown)


be/src/connector/hive_connector.h
```diff
@@ -198,4 +201,54 @@ class HiveDataSource final : public DataSource {
     HdfsScanProfile _profile;
 };
 
+// Functor used with type_dispatch_filter to check whether a file's column range
+// [file_min, file_max] (from THdfsScanRange.min_max_values) overlaps with the
+// runtime filter's range [rf_min, rf_max].  Returns true if the file can be pruned
+// (i.e. there is definitely NO overlap).
+struct RfMinMaxFileStatsChecker {
```
dirtysalt, line 208: duplicated to the code in hive_connector.cpp?

dirtysalt (changes requested):

PR #71146, opened 2026-04-01
[Perf] Batch-preload delvecs during compaction conflict resolution

luohaha:
## Why I'm doing:

During publish of compaction results in shared-data mode, the compaction conflict resolver loads delete vectors (delvecs) **one-by-one** for each input rowset segment. Under high-throughput stream load workloads, a single compaction can have 400-500 input rowsets, causing 400-500 **sequential** remote storage reads (S3/OSS) at ~2-3ms each, totaling **1.0-1.4 seconds per publish**.

### Benchmark Evidence (30GB single-table, 3x CN 16c64g shared-data)

From CN publish traces (`lake_service.cpp:514`):

| input_rowsets_size | compaction_delvec_loader_us | Total publish cost |
|---|---|---|
| 88 | 56 ms | 1,716 ms |
| 147 | 180 ms | 1,410 ms |
| 263 | 518 ms | 2,545 ms |
| 407 | 1,095 ms | 2,592 ms |
| 481 | 1,341 ms | 3,699 ms |

**Clear linear scaling**: ~2.8 ms per input rowset, dominating publish time as rowset count grows. This directly causes high ingestion latency (P99: 76s vs target: <5s) because stream loads must wait for publish to complete.

## What I changed:

Added `preload_delvecs()` to `LakeDelvecLoader` that batch-reads all delvecs for a tablet version **before** the conflict resolution loop.

**How it works:**
1. Load the tablet metadata at `base_version` (already cached by tablet_manager)
2. Group all delvec entries by their file (`version_to_file` mapping)
3. For each unique delvec file, read the **entire file** into memory in a single IO
4. Extract all needed delvec pages from the in-memory buffer
5. The main loop then finds all delvecs already preloaded — zero on-demand remote reads

This turns **N sequential file reads** into **F file reads** where F = number of unique delvec files (typically 1-10), reducing delvec loading time by ~10x.

### Files changed:
- `be/src/storage/lake/lake_delvec_loader.h` — Add `preload_delvecs()` method and internal map
- `be/src/storage/lake/lake_delvec_loader.cpp` — Implement batch preloading; check preloaded map in `load()`
- `be/src/storage/lake/lake_primary_key_compaction_conflict_resolver.cpp` — Call `preload_delvecs()` before conflict resolution in both code paths

## Expected impact:
- Reduce `compaction_delvec_loader_latency_us` from ~1s to ~50-200ms for tablets with 400+ input rowsets
- Reduce overall publish time, directly improving ingestion latency
- Minimal memory overhead (delvec files are typically <1MB)

## Risks:
- Preloads ALL delvecs for the tablet at the base version, including those not needed by this compaction. For very large tablets with many segments, this could use extra memory temporarily. The delvec data is small (roaring bitmaps) so this is unlikely to be an issue in practice.

### DIFF
--- a/be/src/storage/lake/lake_delvec_loader.cpp
+++ b/be/src/storage/lake/lake_delvec_loader.cpp
@@ -70,8 +70,8 @@ Status LakeDelvecLoader::preload_delvecs(int64_t tablet_id, int64_t version) {
         const std::string filepath = _lake_io_opts.location_provider->tablet_metadata_location(tablet_id, version);
         ASSIGN_OR_RETURN(metadata, _tablet_manager->get_tablet_metadata(filepath, _fill_cache, 0, _lake_io_opts.fs));
     } else {
-        ASSIGN_OR_RETURN(metadata, _tablet_manager->get_tablet_metadata(tablet_id, version, _fill_cache, 0,
-                                                                        _lake_io_opts.fs));
+        ASSIGN_OR_RETURN(metadata,
+                         _tablet_manager->get_tablet_metadata(tablet_id, version, _fill_cache, 0, _lake_io_opts.fs));
     }
 
     const auto& delvec_meta = metadata->delvec_meta();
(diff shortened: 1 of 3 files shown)


luohaha:
preload 方案收益有限 — 当 compaction 合并的 segment 多时（400-500），即使按 file 分组读，总远程 IO 量没有本质减少。

建议换一个方向：**把小 delvec inline 到 tablet metadata 中**。

具体思路：在 `DelvecMetadataPB` 中，对于小于某个阈值（比如 4KB）的 delvec，直接把数据内联存储（而不是只存 file offset + size）。这样 tablet metadata 加载时小 delvec 就已经在内存里了，compaction conflict resolution 时完全不需要额外的远程 IO。

下一轮迭代请按这个方向分析可行性并实现。

PR #71167, opened 2026-04-01
[BugFix] Clone nullable result null bitmaps in array/map functions (backport #71153)

GavinMar:
## Why I'm doing:

Several array/map expression functions (`array_length`, `array_append`, `array_remove`, `array_flatten`, `map_size`) return a NullableColumn whose `null_column` shares the same pointer as the input's `null_column` (via `mutate()` or direct reference). When the result and input columns coexist in the same Chunk (e.g., via ProjectOperator), `Chunk::filter()` modifies the shared null bitmap through one column, corrupting the other and causing a `DCHECK` crash:

```
Check failed: _data_column->size() == _null_column->size() (3 vs. 2)
```

## What I'm doing:

- Changed `null_column().mutate()` back to `null_column()->clone()` in `array_length` to ensure exclusive ownership of the null bitmap
- Changed direct `null_column()` references to `null_column()->clone()` in `array_append`, `array_remove`, `array_flatten`, and `map_size`
- Added regression tests verifying that result NullableColumns have independent null bitmaps that are not corrupted by in-place filtering of the input

Fixes https://github.com/StarRocks/StarRocksTest/issues/11200

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/array_functions.cpp
+++ b/be/src/exprs/array_functions.cpp
@@ -60,7 +60,7 @@ StatusOr<ColumnPtr> ArrayFunctions::array_length([[maybe_unused]] FunctionContex
         if (arg0->has_null()) {
             // Copy null flags.
             return NullableColumn::create(std::move(col_result),
-                                          std::move(*(down_cast<const NullableColumn*>(arg0)->null_column())).mutate());
+                                          down_cast<const NullableColumn*>(arg0)->null_column()->clone());
         } else {
             return col_result;
         }
@@ -136,7 +136,7 @@ StatusOr<ColumnPtr> ArrayFunctions::array_append([[maybe_unused]] FunctionContex
     RETURN_IF_ERROR(result);
 
     if (nullable_array != nullptr) {
-        return NullableColumn::create(result.value(), nullable_array->null_column());
+        return NullableColumn::create(std::move(result.value()), nullable_array->null_column()->clone());
     }
     return result;
 }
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

be/src/exprs/array_functions.cpp
```diff
@@ -60,7 +60,7 @@ StatusOr<ColumnPtr> ArrayFunctions::array_length([[maybe_unused]] FunctionContex
         if (arg0->has_null()) {
             // Copy null flags.
             return NullableColumn::create(std::move(col_result),
-                                          std::move(*(down_cast<const NullableColumn*>(arg0)->null_column())).mutate());
+                                          down_cast<const NullableColumn*>(arg0)->null_column()->clone());
```
satanson: use Column::mutate() instead of clone, this function has the same functionality; as follows.
```cpp
        auto null_column = NullColumn::static_pointer_cast(Column::mutate(nullable_array->null_column()));
        return NullableColumn::create(std::move(result.value()), std::move(null_column));
```

GavinMar: Good Point, fixed.

satanson (approved):

PR #71168, opened 2026-04-01
[BugFix] Fix batch publish blocked by incorrect transaction graph

smartlxh:
## Why I'm doing:
`batch publish` relies on `TransactionGraph` to decide dependency-safe publish order. If the graph is maintained incorrectly after node removal, publish ordering can become invalid and publish can be blocked.

The issue is most likely to surface when switching from normal (single) publish to batch publish. In single publish, readiness is judged by partition version, so a transaction removed there may not be a zero in-degree node (a node with no incoming edges) in `TransactionGraph`. If `remove()` does not correctly preserve graph dependencies in this path, batch publish can later make wrong scheduling decisions.

### Root cause in `remove()`
`TransactionGraph.remove()` had two correctness gaps when removing a non-root node:

1. **`lastTableWriter` is cleared too aggressively**
   - Old logic: if removed node is current `lastTableWriter[tableId]`, it directly removes the mapping.
   - Problem: there may still be an earlier in-graph writer for the same table (a predecessor), but mapping is lost.
   - Impact: next `add()` for that table sees no previous writer and can become a zero in-degree node incorrectly.

2. **Predecessor/successor edges are not re-linked**
   - Old logic: for each successor, only `next.ins.remove(node)` is done.
   - Problem: happens-before relationship is dropped when removing a middle node (`prev -> node -> next`), because `prev -> next` is never restored.
   - Impact: successor can become zero in-degree too early and be considered publishable before predecessor chain is actually respected.

These two problems directly affect the dependency graph consumed by batch publish.

## What I'm doing:
- Keep `lastTableWriter` by falling back to the nearest predecessor that still writes the same table when removing a node.
- Re-link predecessors to successors when removing an intermediate node so dependency chains remain consistent.
- Add regression tests for tail-node removal and middle-node removal, and adjust existing assertions.

Fixes #N/A

## What type of PR is this:
- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:
- [ ] This is a backport PR

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionGraph.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionGraph.java
@@ -131,10 +131,26 @@ public class TransactionGraph {
         }
         nodes.remove(txnId);
         nodesWithoutIns.remove(node);
+        // When removing the tail node for a table, update lastTableWriter to the nearest
+        // predecessor that also writes that table, instead of clearing it entirely.
+        // This ensures subsequent transactions for the same table can still find their dependency.
         for (long tableId : node.writeTableIds) {
             Node holder = lastTableWriter.get(tableId);
             if (holder == node) {
-                lastTableWriter.remove(tableId);
+                Node prevWriter = null;
+                if (node.ins != null) {
+                    for (Node dep : node.ins) {
+                        if (dep.writeTableIds.contains(tableId)) {
+                            prevWriter = dep;
+                            break;
+                        }
+                    }
+                }
+                if (prevWriter != null) {
+                    lastTableWriter.put(tableId, prevWriter);
+                } else {
+                    lastTableWriter.remove(tableId);
+                }
             }
         }
         if (node.outs == null) {
@@ -142,6 +158,14 @@ public class TransactionGraph {
         }
         for (Node next : node.outs) {
             next.ins.remove(node);
+            // Re-link predecessors directly to successors so that ordering is preserved
+            // even when an intermediate node is removed out of order.
+            if (node.ins != null && !node.ins.isEmpty()) {
+                for (Node prev : node.ins) {
+                    next.addIns(prev);
+                    prev.addOuts(next);
+                }
+            }
             if (next.ins.isEmpty()) {
                 nodesWithoutIns.add(next);
             }
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionGraph.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionGraph.java
@@ -29,6 +29,14 @@ import java.util.stream.Collectors;
 
 /**
  * store transactions' dependency relationships
+ *
+ * The dependency graph is modeled as the overlay of one doubly-linked writer
+ * chain per table: for every table T a node writes, the node stores its
+ * immediate predecessor (tablePrev[T]) and successor (tableNext[T]) on T's
+ * chain. Keeping the per-table chain explicit makes add/remove deterministic
+ * linked-list operations and avoids reverse-engineering chain membership from
+ * an unlabeled dependency set.
+ *
  * this class is used in DatabaseTransactionMgr and all methods are protected by mgr's lock
  * so this class does not require additional synchronization
  */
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/transaction/TransactionGraph.java
```diff
@@ -131,17 +131,41 @@ public void remove(long txnId) {
...
         for (long tableId : node.writeTableIds) {
             Node holder = lastTableWriter.get(tableId);
             if (holder == node) {
-                lastTableWriter.remove(tableId);
+                Node prevWriter = null;
+                if (node.ins != null) {
+                    for (Node dep : node.ins) {
+                        if (dep.writeTableIds.contains(tableId)) {
+                            prevWriter = dep;
+                            break;
```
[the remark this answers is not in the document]
smartlxh: This scenario cannot occur by construction. In the graph, if two nodes both write the
  same table, the later one must have the earlier one as a predecessor (via lastTableWriter).
  They can never appear as two parallel, independent predecessors of the same node for the
  same table. For any table T, node.ins contains at most one node that is the direct
  "responsible" predecessor for T. The HashSet ordering concern therefore does not apply

xiangguangyxg (dismissed):

kevincai:
@mergifyio rebase main

xiangguangyxg:
@Mergify rebase

CelerData-Reviewer:
@codex review

fe/fe-core/src/test/java/com/starrocks/transaction/TransactionGraphTest.java
```diff
@@ -193,15 +199,147 @@ public void testGetTxnsWithTxnDependencyBatch() {
...
+        graph2.remove(3);
 
         txnIds = graph2.getTxnsWithoutDependency();
         assertEquals(txnIds.size(), 1);
         batchTxnIds = graph2.getTxnsWithTxnDependencyBatch(1, 5, 4);
         assertEquals(batchTxnIds.size(), 1);
     }
 
+    @Test
+    public void testRemoveTailBreaksLastTableWriter() {
```
kevincai, line 212: Misleading name: this test asserts the *fixed* (correct) behavior -- after removing the tail, `lastTableWriter` is preserved so the later `add` still depends on the remaining writer. The name says "Breaks", and the inline `// Actual (BUG)` comment reads as if the test expects the bug. A future reader will think this is a known-failing test. Suggest renaming to `testRemoveTailPreservesLastTableWriter` and reframing the comment as "regression guard: would have been {1,4} before the fix".

fe/fe-core/src/test/java/com/starrocks/transaction/TransactionGraphTest.java
```diff
@@ -193,15 +199,147 @@ public void testGetTxnsWithTxnDependencyBatch() {
...
+
+        // If correct: roots should be {1} only (txn4 depends on txn2, which depends on txn1)
+        // If buggy:   roots will be {1, 4} because txn4 has no dependency
+        assertEquals(Lists.newArrayList(1L), roots,
+                "txn4 should depend on txn2 since both write table1, " +
+                        "but lastTableWriter was incorrectly cleared when txn3 was removed");
+    }
+
+    @Test
+    public void testRemoveMiddleNodeBreaksDependencyChain() {
```
kevincai, line 250: Same naming issue as the tail test: the assertion checks the corrected behavior (txn3 still depends on txn1 after the middle node is spliced out), but the name says "Breaks..." and the comment says "doesn't re-link". Suggest `testRemoveMiddleNodePreservesDependencyChain` and adjust the comment accordingly.

fe/fe-core/src/test/java/com/starrocks/transaction/TransactionGraphTest.java
```diff
@@ -106,6 +111,7 @@ public void testLargeGraph() {
                     if (writeTableIds.contains(tableId)) {
                         continue;
                     }
+                    writeTableIds.add(tableId);
```
kevincai, line 114: Good catch -- this fixes a pre-existing latent bug in the test, not just a tweak. Without this `add`, the inner loop never populated `writeTableIds`, so every node had an empty table list and was an independent root; `testLargeGraph` exercised essentially no graph structure before. Worth calling out in the PR description since it materially changes what this test covers.

fe/fe-core/src/test/java/com/starrocks/transaction/TransactionGraphTest.java
```diff
@@ -193,15 +199,147 @@ public void testGetTxnsWithTxnDependencyBatch() {
         assertEquals(txnIds.size(), 1);
         batchTxnIds = graph2.getTxnsWithTxnDependencyBatch(1, 5, 2);
         assertEquals(batchTxnIds.size(), 2);
-        graph.remove(2);
-        graph.remove(3);
+        graph2.remove(2);
```
kevincai, line 202: This corrects a copy-paste bug (`graph` -> `graph2`); the old code was mutating the wrong graph so the subsequent assertions on `graph2` weren't testing the intended state. I re-traced the resulting chains and the new assertions (root `{4}`, batch-from-4 size 1) hold. Also worth a mention in the PR body as an incidental test fix.

kevincai (commented):
Reviewed the `remove()` rewrite branch-by-branch (head/middle/tail/only, single- and multi-table) against the new per-table chain model and the `DatabaseTransactionMgr` caller -- the core fix is correct and the chain-walk contract still holds. A few non-blocking test-quality notes inline.

kevincai:
deprecated by https://github.com/StarRocks/starrocks/pull/78077

PR #71097, opened 2026-03-31
[BugFix][Iceberg] Use manifest metadata to answer COUNT(*)/COUNT(1) queries

super-annie:
Avoid full file planning for COUNT(*) and COUNT(1) by reading row counts directly from Iceberg manifest metadata.

Two optimization paths:

1. FE manifest short-circuit (no GROUP BY): Computes count directly from manifest addedRowsCount + existingRowsCount in the FE O(manifests), zero file I/O. Returns a constant via LogicalValuesOperator. Applies only when no delete manifests exist and all row counts are available.

2. BE scan-range optimization (with or without GROUP BY): Rewrites COUNT(*) to SUM(___count___) and sends scan ranges to BE. BE reads record_count from scan-range metadata without opening files. aggregateTasksByPartition groups FileScanTasks by partition and builds synthetic SyntheticCountFileScanTask entries with aggregated record counts.

Other changes:
- isAllPartitionColumnsAlwaysIdentity checks all historical partition specs to guard against non-identity transforms.
- scannedTables uses ConcurrentHashMap.newKeySet() for thread safety.
- makePartitionKey uses length-prefixed encoding to prevent key collisions.
- Handle unpartitioned tables correctly in tryAggregateFromManifestRowCounts.

## Why I'm doing:
When running count agg queries like below, StarRocks shows very bad performance (~5 minutes) while Trino uses around > 50 seconds to finish.

```
mysql> SELECT date_trunc('day', ds) AS day, count(*) AS row_count  FROM     icebergDB.tableA GROUP BY     date_trunc('day', ds) ORDER BY     row_count DESC LIMIT 20;
```
## What I'm doing:

Avoid full file planning for COUNT(*) and COUNT(1) by reading row counts directly from Iceberg manifest metadata.

Two optimization paths:

1. FE manifest short-circuit (no GROUP BY): Computes count directly from manifest addedRowsCount + existingRowsCount in the FE O(manifests), zero file I/O. Returns a constant via LogicalValuesOperator. Applies only when no delete manifests exist and all row counts are available.

2. BE scan-range optimization (with or without GROUP BY): Rewrites COUNT(*) to SUM(___count___) and sends scan ranges to BE. BE reads record_count from scan-range metadata without opening files. aggregateTasksByPartition groups FileScanTasks by partition and builds synthetic SyntheticCountFileScanTask entries with aggregated record counts.

Other changes:
- isAllPartitionColumnsAlwaysIdentity checks all historical partition specs to guard against non-identity transforms.
[...]

PR #71176, opened 2026-04-02
[Enhancement] Early-terminate scan planning for LIMIT queries

super-annie:
## Why I'm doing:
Improve performance for select statement with limit clause.

StarRocks behaves slowly when select with limit from an iceberg table, e.g.
select * from iceberg_catalog.db.table limit 10;
costs more than a few minutes to finish when the table is super large (PB level)

## What I'm doing:
When a query has a LIMIT and the predicate is fully enforced by Iceberg partition pruning (identity columns across all historical specs), stop iterating FileScanTasks once the accumulated record_count from delete-free files reaches the limit.

canTerminateEarly requires limit >= 0 and isPredicatePartitionEnforced(). Non-partition predicates yield inclusive record counts (upper bounds), so early termination is unsafe for them.
Files with delete entries are excluded from the row-count accumulation since their effective row count may be lower than record_count.
Results from limited scans are not cached in scannedTables, preventing truncated file lists from being served to subsequent unlimited queries.
getPrunedPartitions forces limit=-1 so partition enumeration is never short-circuited by a query LIMIT.
Non-LIMIT queries skip incremental scan range delivery, eliminating per-round ACK overhead in multi-fragment plans.

Extra Note: this PR probably carries part of the change from https://github.com/StarRocks/starrocks/pull/71097, would recommend review and merge https://github.com/StarRocks/starrocks/pull/71097 first and then this one.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -90,6 +90,7 @@ import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DataFiles;
 import org.apache.iceberg.DeleteFile;
 import org.apache.iceberg.FileContent;
+import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.HistoryEntry;
 import org.apache.iceberg.ManifestFile;
@@ -141,10 +142,12 @@ import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


dirtysalt:
@super-annie starrocks support incremental scan ranges delivery since 3.5, and it will parse data files from manifest files in batches in one thread.  And during in each batch, some data files will be sent down to BE.  And when FE reads enough(limit) rows in another thread, it will end the parsing process.

What's the improvement with your patch on your scenario? 

```
FE (Parsing Thread - single-threaded)        FE (Limit Thread)
─────────────────────────────────────        ──────────────────
  │                                            │
  ▼                                            ▼
┌─────────────────────┐                ┌──────────────────┐
│  Read manifest file │                │ Monitor rows read │
└──────────┬──────────┘                │     from BE       │
           │                           └────────┬─────────┘
           ▼                                    │
┌─────────────────────┐                         ▼
│ Parse batch of      │◄──── yes ────  ┌──────────────────┐
│ data files          │                │  Reached LIMIT?  │
└──────────┬──────────┘                └────────┬─────────┘
           │                                    │
           ▼                                    │ yes
┌─────────────────────┐                         │
│ Send scan ranges    │        signal stop ──────┘
│     ──► BE          │◄──────────────────────────
└──────────┬──────────┘
           │
           ▼
┌─────────────────────┐
│   More batches?     │
└──────────┬──────────┘
      no   │    yes
           └──────────► (loop back to parse next batch)
           │
           ▼
┌─────────────────────┐
│  Parsing complete   │
└─────────────────────┘
           │
           ▼
  ┌────────────────┐
  │ BE: scan files │
  └────────────────┘
```

PR #71194, opened 2026-04-02
[Enhancement] support per stmt audit log for multi stmts

before-Sunrise:
## What I'm doing:
This PR changes multi-statement handling from "audit once per command" to "audit once per executed statement", while keeping parsing unified for the original
  SQL text.

  Main changes:
  - Parse the original SQL text once up front in `handleQuery`.
  - If parsing fails, emit exactly one failed audit record with the original SQL text.
  - If parsing succeeds, execute statements one by one and audit each executed statement independently.
  - For multi-statement requests:
    - successful statements before a failure each produce one audit record
    - the failed statement also produces one audit record
    - statements after the failure are not executed and are not audited
  - Add the session variable `audit_stmt_before_execute`:
    - when disabled, behavior remains "after execution audit only"
    - when enabled, each executed statement emits one `BEFORE_QUERY` audit event before execution and one `AFTER_QUERY` audit event after execution
  - Keep exception handling consistent with the previous behavior by auditing first and then rethrowing execution exceptions.
  - Support `LargeInPredicateException` retry at statement granularity:
    - the request is still parsed as a whole first
    - if one statement throws `LargeInPredicateException` during execution, temporarily disable `enable_large_in_predicate`
    - reparse the original SQL text, rebuild the statement at the same index, and retry only that statement
    - other statements are not retried
  - Extend `handleExecute()` so prepared statement execution also covers the failure-audit path and the before/after audit hooks.
  - Upgrade `QueryDetail` to record multi-statement execution per statement instead of only recording the last statement.
  - Rename the context flag from `isSingleStmt` to `isMultiStmt` so the code matches the actual semantics more clearly: single-statement is the default, and only
  true multi-statement paths opt in.

  Behavior after this change:
  - parse failure: one failed audit record containing the full original SQL
  - `stmt1; stmt2; stmt3`, where `stmt2` fails:
    - `stmt1` is audited as success
    - `stmt2` is audited as failure
    - `stmt3` is not executed and is not audited
  - `LargeInPredicateException` retry only retries the current statement, not the whole request
  - QueryDetail is generated per executed statement

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/docs/en/administration/management/FE_parameters/log_server_meta.md
+++ b/docs/en/administration/management/FE_parameters/log_server_meta.md
@@ -119,6 +119,15 @@ This topic introduces the following types of FE configurations:
 - Description: The maximum number of audit log files that can be retained within each retention period specified by the `audit_log_roll_interval` parameter.
 - Introduced in: -
 
+### `audit_stmt_before_execute`
+
+- Default: false
+- Type: Boolean
+- Unit: -
+- Is mutable: Yes
+- Description: Controls whether FE emits a `BEFORE_QUERY` audit event before each statement executes. When enabled, ConnectProcessor writes one audit record before statement execution and still writes the normal `AFTER_QUERY` audit record after the statement finishes. For multi-statement requests, this happens per executed statement: statements executed before a failure each emit a before/after pair, the failed statement also emits both records, and statements after the failure emit nothing because they are not executed. Parse failures are unchanged and still only produce the existing failed after-audit for the original SQL text. This is an FE-wide switch, so changing it affects all sessions on the FE.
+- Introduced in: v4.1
+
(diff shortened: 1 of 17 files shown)


CelerData-Reviewer:
@codex review

satanson (approved):

gengjun-git (approved):

stephen-shelby (approved):

PR #71242, opened 2026-04-02
[BugFix] Fix UpdateTabletSchemaTask signature collision across alter jobs

wxl24life:
## Why I'm doing:

When multiple ALTER TABLE jobs operate on the same table/partition in quick succession, `UpdateTabletSchemaTask` produces identical `AgentTaskQueue` signatures because the signature is computed as `tablets.hashCode()`. Since different alter jobs on the same table operate on the same tablet set, they get the same signature, leading to a race condition:

1. **Task rejection**: AgentTaskQueue rejects the new task while the old task with the same signature is still queued
2. **Task mismatch**: After the old task is removed from FE queue (job timeout/cancel), the new task enters the queue with the same signature. When BE finishes executing the old task and reports completion, FE matches it to the new job's task by `(backendId, taskType, signature)` — matching the wrong job
3. **Publish failure**: The new job proceeds to publish with a txnId whose txn log was never written by BE, resulting in a 404 error on object storage

This was observed in production where two consecutive `ALTER TABLE ADD COLUMN` statements on the same table caused the second job to get stuck in `FINISHED_REWRITING` state, continuously failing to publish with "404 Not Found" on txn log.

## What I'm doing:

Update `setTxnId()` in the base class `TabletMetadataUpdateAgentTask` to recompute the signature as `Objects.hash(baseSignature, txnId)`, where `baseSignature` is the original signature saved at construction time. This ensures:

- Different alter jobs produce different signatures even when operating on the same tablet set
- All `TabletMetadataUpdateAgentTask` subclasses benefit defensively (not just `UpdateTabletSchemaTask`)
- `setTxnId()` is idempotent — calling it multiple times with the same txnId always produces the same signature
- No backward compatibility concern — signatures are purely in-memory (`AgentTaskQueue`), never persisted

This is safe because in the `LakeTableAlterMetaJobBase.sendAgentTask()` call chain, `setTxnId(watershedTxnId)` is always called before `AgentTaskQueue.addBatchTask()`, so the signature includes the txnId before the task enters the queue.

**Changed files:**
- `TabletMetadataUpdateAgentTask.java`: Added `baseSignature` field and updated `setTxnId()` to compute `Objects.hash(baseSignature, txnId)`
- `TabletMetadataUpdateAgentTaskFactory.java`: Removed redundant `setTxnId()` override from `UpdateTabletSchemaTask` (now handled by base class)
- `AgentTaskQueueSignatureCollisionTest.java`: 4 test cases covering signature divergence, queue coexistence, race condition prevention, and dedup preservation

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
+++ b/fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
@@ -358,6 +358,17 @@ public class TabletMetadataUpdateAgentTaskFactory {
         }
 
         @Override
+        public void setTxnId(long txnId) {
+            super.setTxnId(txnId);
+            // Include txnId in signature to prevent collision between different alter jobs
+            // operating on the same tablet set. Without this, two jobs altering the same
+            // table/partition produce identical signatures (tablets.hashCode()), causing
+            // AgentTaskQueue to reject the second task or FE to match BE's old task
+            // completion report to the wrong job.
+            this.signature = Objects.hash(tablets, txnId);
+        }
+
+        @Override
         public Set<Long> getTablets() {
             return new HashSet<>(tablets);
         }
--- a/fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
+++ b/fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
@@ -358,17 +358,6 @@ public class TabletMetadataUpdateAgentTaskFactory {
         }
 
         @Override
-        public void setTxnId(long txnId) {
-            super.setTxnId(txnId);
-            // Include txnId in signature to prevent collision between different alter jobs
-            // operating on the same tablet set. Without this, two jobs altering the same
-            // table/partition produce identical signatures (tablets.hashCode()), causing
-            // AgentTaskQueue to reject the second task or FE to match BE's old task
-            // completion report to the wrong job.
-            this.signature = Objects.hash(tablets, txnId);
-        }
-
-        @Override
         public Set<Long> getTablets() {
             return new HashSet<>(tablets);
         }
--- a/fe/fe-core/src/test/java/com/starrocks/task/AgentTaskQueueSignatureCollisionTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/task/AgentTaskQueueSignatureCollisionTest.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Set;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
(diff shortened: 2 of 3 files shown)


xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
```diff
@@ -357,6 +357,17 @@ private UpdateTabletSchemaTask(long backendId, List<Long> tablets, TTabletSchema
             this.createSchemaFile = createSchemaFile;
         }
 
+        @Override
+        public void setTxnId(long txnId) {
```
kevincai: why only override `UpdateTabletSchemaTask`, how about other tasks that can also set transaction id?

fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
```diff
@@ -357,6 +357,17 @@ private UpdateTabletSchemaTask(long backendId, List<Long> tablets, TTabletSchema
...
+            super.setTxnId(txnId);
+            // Include txnId in signature to prevent collision between different alter jobs
+            // operating on the same tablet set. Without this, two jobs altering the same
+            // table/partition produce identical signatures (tablets.hashCode()), causing
+            // AgentTaskQueue to reject the second task or FE to match BE's old task
+            // completion report to the wrong job.
+            this.signature = Objects.hash(tablets, txnId);
```
kevincai: the other question is the backwards compatibility. Will an upgrade cause all agentTasks for this type invalid due to signature changed.

fe/fe-core/src/test/java/com/starrocks/task/AgentTaskQueueSignatureCollisionTest.java
```diff
@@ -0,0 +1,191 @@
...
+        task2.setTxnId(1000L);
+
+        assertEquals(task1.getSignature(), task2.getSignature(),
+                "Same tablets + same txnId should produce same signature (dedup still works)");
+
+        // AgentTaskQueue correctly deduplicates retries
+        assertTrue(AgentTaskQueue.addTask(task1));
+        // task2 has the same signature, so it should be rejected (correct dedup)
+        // This is intentional: if the same job retries, the old task must be removed first
+    }
```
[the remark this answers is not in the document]
wxl24life: Good catch! Fixed in e856256 — added explicit `assertFalse(AgentTaskQueue.addTask(task2))` and `assertSame(task1, ...)` to verify dedup rejection.

fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
wxl24life: Great point. Moved the fix to the base class `TabletMetadataUpdateAgentTask.setTxnId()` in e856256 so all subclasses benefit defensively. The signature is now updated as `Objects.hash(this.signature, txnId)` whenever `setTxnId()` is called, regardless of which concrete task type.

fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
wxl24life: No backward compatibility concern. The signature is purely in-memory — it's only used for `AgentTaskQueue` lookup matching within a single FE session. It is never persisted (not in EditLog, not in image). On FE restart, AgentTaskQueue is cleared and all tasks are re-created with fresh signatures. So an upgrade simply starts with a clean queue.

wxl24life: no, the stat only existed in memory, it will be reset after fe/be restart

fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
wxl24life: great catch, have refactored to make all  tablet meta update tasks's signature include txn_id

kevincai (dismissed):

kevincai (approved):

xiangguangyxg (approved):

PR #71251, opened 2026-04-03
[Refactor] Add ExecExpr typed expression hierarchy for physical plan layer

HangyuanLiu:
## Why I'm doing:

The current FE uses AST `Expr` classes for both parsing/analysis and physical plan execution. This tight coupling prevents moving expression classes (e.g., `FunctionCallExpr`, `SlotRef`) to the `fe-parser` module and makes the planner depend on mutable AST state. A separate, typed, immutable expression hierarchy for the physical plan layer is needed.

## What I'm doing:

Introduce the `ExecExpr` hierarchy — a typed, immutable expression representation for the physical plan layer, fully decoupled from AST `Expr`. **This is a pure addition: 33 new files, zero modifications to existing code.**

New components:
- **`ExecExpr`** base class + **28 concrete types** (`ExecSlotRef`, `ExecFunctionCall`, `ExecLiteral`, `ExecCast`, `ExecBinaryPredicate`, `ExecCompoundPredicate`, etc.)
- **`ExecExprVisitor`** — type-safe visitor for traversal
- **`ExecExprSerializer`** — serializes `ExecExpr` → Thrift `TExpr`
- **`ExecExprExplain`** — EXPLAIN formatting for `ExecExpr`
- **`ScalarOperatorToExecExpr`** — converts optimizer `ScalarOperator` → `ExecExpr`
- **`ThriftEnumConverter`** — centralized keys-type and opcode Thrift conversions
- **`ExprOpcodeRegistry`** — maps AST expression metadata to `TExprOpcode`

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/expression/ExecSlotRef.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/expression/ExecSlotRef.java
@@ -1,0 +1,135 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.planner.expression;
+
+import com.starrocks.planner.SlotDescriptor;
+import com.starrocks.planner.SlotId;
+import com.starrocks.planner.TupleId;
+import com.starrocks.thrift.TExprNode;
+import com.starrocks.thrift.TExprNodeType;
+import com.starrocks.thrift.TSlotRef;
+import com.starrocks.type.VarcharType;
+
+/**
+ * Execution-plan slot reference. Holds a resolved {@link SlotDescriptor}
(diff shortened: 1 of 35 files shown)


fe/fe-core/src/main/java/com/starrocks/planner/expression/ExecSlotRef.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public <R, C> R accept(ExecExprVisitor<R, C> visitor, C context) {
+        return visitor.visitExecSlotRef(this, context);
+    }
+
+    @Override
+    public ExecSlotRef clone() {
+        return new ExecSlotRef(this);
```
stephen-shelby, line 133: `ExecSlotRef.clone()` currently keeps the same `SlotDescriptor` instance (`this.desc = other.desc`), while `setNullable()` mutates `desc.setIsNullable(nullable)`. That means changing the nullable property on a cloned `ExecSlotRef` also changes the original one through the shared descriptor.

If `ExecExpr` clones are expected to be independently mutable, this can lead to surprising cross-tree side effects. Is that intentional here?

HangyuanLiu: The current implementation is consistent with the behavior of SlotRef. After the later migration, if there is no longer a need for setting values, these setter methods can actually be removed.

stephen-shelby (approved):

Astralidea (approved):

PR #71269, opened 2026-04-03
[Enhancement] Handle NULL variant as type-minimum for non-nullable PK columns in SstSeekRange

xiangguangyxg:
## Why I'm doing:

  Range distribution tables in shared-data mode support tablet split/merge with composite sort-key ranges. FE uses MIN sentinels to represent negative-infinity bounds in per-column range boundaries, mapped to NULL_VALUE before sending to BE. However, BE's `create_sst_seek_range_from()` rejects NULL on non-nullable PK columns, and `convert_t_range_to_pb_range()` rejects any TVariant without a value field. This blocks range distribution with multi-column sort keys.

## What I'm doing:

  Two source-level fixes in `tablet_range_helper.cpp`, plus explicit rejection of raw MINIMUM/MAXIMUM variants that FE should never send:

  1. **`create_sst_seek_range_from`**: For non-nullable PK columns, a NULL datum can only represent a MIN sentinel (PK columns cannot hold real NULLs). Instead of rejecting it, fill with the per-type minimum value via `TypeInfo::set_to_min()` for correct PK encoding.

  2. **`convert_t_range_to_pb_range`**: Accept missing `value` field for non-NORMAL_VALUE variants (NULL_VALUE, MINIMUM, MAXIMUM). FE's `NullVariant.toThrift()` omits the value field, which previously caused an unconditional rejection.

  3. **Explicit MINIMUM/MAXIMUM rejection**: Add early validation in `convert_t_range_to_pb_range` and `RangeRouter::parse_bound` to reject raw MINIMUM/MAXIMUM variants with clear error messages, preventing silent data corruption if FE sends them without mapping to NULL_VALUE first.

Fixes #64986

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/tablet_range_helper.cpp
+++ b/be/src/storage/lake/tablet_range_helper.cpp
@@ -26,9 +26,56 @@
 #include "storage/datum_variant.h"
 #include "storage/primary_key_encoder.h"
 #include "storage/types.h"
+#include "types/storage_type_traits.h"
 #include "types/type_descriptor.h"
 
 namespace starrocks::lake {
+
+// Produce a Datum holding the minimum value for the given logical type.
+// Uses TypeInfo::set_to_min() which calls std::numeric_limits<CppType>::lowest().
+// Covers all PK-supported types (APPLY_FOR_ALL_PK_SUPPORT_TYPE in logical_type_infra.h).
+template <LogicalType TYPE>
+static Datum datum_from_type_min_impl() {
+    using CppType = StorageCppType<TYPE>;
+    CppType value{};
+    get_type_info(TYPE)->set_to_min(&value);
+    Datum d;
+    d.set(value);
+    return d;
+}
+
+static StatusOr<Datum> datum_from_type_min(LogicalType type) {
+    switch (type) {
+    case TYPE_BOOLEAN: {
+        bool v;
+        get_type_info(TYPE_BOOLEAN)->set_to_min(&v);
+        Datum d;
+        d.set_int8(v);
+        return d;
+    }
+    case TYPE_TINYINT:
+        return datum_from_type_min_impl<TYPE_TINYINT>();
+    case TYPE_SMALLINT:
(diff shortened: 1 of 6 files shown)


srlch (dismissed):

srlch (approved):

be/src/storage/lake/tablet_range_helper.cpp
```diff
@@ -146,11 +193,18 @@ StatusOr<SstSeekRange> TabletRangeHelper::create_sst_seek_range_from(const Table
...
+                // can only represent a MIN sentinel (FE maps MIN → NULL_VALUE).
+                // Fill with the per-type minimum value for correct PK encoding.
+                ASSIGN_OR_RETURN(datum, datum_from_type_min(type_desc.type));
+                auto column = ColumnHelper::create_column(type_desc, false);
+                column->append_datum(datum);
+                chunk->append_column(std::move(column), (SlotId)idx);
+            } else {
+                auto column = ColumnHelper::create_column(type_desc, is_nullable);
+                column->append_datum(datum);
+                chunk->append_column(std::move(column), (SlotId)idx);
```
wyb, line 206:
```
Column column;
if (!is_nullable && datum.is_null()) {
    ASSIGN_OR_RETURN(datum, datum_from_type_min(type_desc.type));
    column = ColumnHelper::create_column(type_desc, false);
} else {
    column = ColumnHelper::create_column(type_desc, is_nullable);
}
column->append_datum(datum);
chunk->append_column(std::move(column), (SlotId)idx);
```

wyb (approved):

satanson (approved):

PR #71382, opened 2026-04-07
[Enhancement] Add `force_group_by_skew_eliminate_when_skewed` to force skew rule application

OliLay:
## Why I'm doing:
When `enableDistinctColumnBucketization` is set to `true`, we may rewrite the plan (from 3-stage agg + shuffle to 4-stage agg + 2 shuffles) if there is skew because else the count distinct is executed on a single pipeline thread which leads to high query runtime. We use a factor to reduce the penalty of the additional aggregation & shuffle in case there is skew. In particular cases, reducing the factor is not enough (e.g. if the extra shuffle induces high network cost), so that even in case of a skew the skew rewrite is not selected. In our use-cases, omitting the skew rewrite leads to significantly slower queries.

## What I'm doing:
I am adding a feature flag `force_group_by_skew_eliminate_when_skewed` which, when enabled, always forces the skew rule to be applied if we have skew detected. The default is `false`.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
```diff
@@ -584,19 +593,29 @@ public CostEstimate visitLogicalIntersect(LogicalIntersectOperator node, Express
...
                     factor = 0.1;
                 }
             } else if (childOp instanceof PhysicalHashAggregateOperator) {
                 PhysicalHashAggregateOperator childAggOp = childOp.cast();
                 DataSkewInfo skewInfo = childAggOp.getDistinctColumnDataSkew();
-                if (skewInfo != null && skewInfo.getStage() == 3) {
-                    factor = skewInfo.getPenaltyFactor();
+                if (skewInfo != null) {
+                    if (enableForceGroupBySkewEliminateWhenSkewed && skewInfo.isGroupBySkewDetected()) {
+                        factor = 0.0;
```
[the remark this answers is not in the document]
OliLay: Fixed with [5148b25](https://github.com/StarRocks/starrocks/pull/71382/commits/5148b25ef6865cd08db02b6511cf38c9d88965e9)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
```diff
@@ -584,18 +593,26 @@ public CostEstimate visitLogicalIntersect(LogicalIntersectOperator node, Express
         // is relatively slow when exchange a large amount of data.
         private double setExchangeCostFactor(Operator childOp) {
             double factor = 1.0;
+            final var enableForceGroupBySkewEliminateWhenSkewed = ConnectContext.get().getSessionVariable()
+                    .isEnableForceGroupBySkewEliminateWhenSkewed();
             if (childOp instanceof LogicalAggregationOperator) {
                 LogicalAggregationOperator childAggOp = childOp.cast();
                 DataSkewInfo skewInfo = childAggOp.getDistinctColumnDataSkew();
-                if (skewInfo != null && skewInfo.getStage() == 3) {
+                if (skewInfo != null && enableForceGroupBySkewEliminateWhenSkewed
```
stephen-shelby, line 601: I think the current implementation still has a correctness / maintainability issue.

With `enable_force_group_by_skew_eliminate_when_skewed=true`, the follow-up fix constrained the forced `0.0` exchange factor to `stage < 4`, but `computeDataSkewPenaltyOfGroupByCountDistinct()` still returns `0.0` for all non-stage-1 aggregate nodes once stage 1 has detected group-by skew. So this is no longer just “do not penalize the extra skew-elimination shuffle”; it also makes the later aggregate operators themselves effectively free, which seems to under-cost the whole 4-stage plan more aggressively than intended.

More generally, this design feels quite tricky because the behavior is now split across:
1. `GroupByCountDistinctDataSkewEliminateRule` stage numbering,
2. shared mutable state in `DataSkewInfo.SharedSkewState`,
3. aggregate costing in `computeDataSkewPenaltyOfGroupByCountDistinct()`, and
4. exchange costing in `setExchangeCostFactor()`.

That makes it easy to fix one path but miss another one, which is what seems to have happened here.

One possible alternative (not necessarily the only solution) would be to keep the force behavior local to a single place instead of propagating cross-stage mutable state. For example, either:
- only make the stage-1 skew decision more aggressive and leave later stages on their existing fixed penalty factors, or
- only discount the additional internal skew-elimination exchange(s), without also zeroing the later aggregate nodes themselves.

I think either of those would be simpler to reason about and less fragile than the current shared-state + multi-site stage checks.

OliLay: > only make the stage-1 skew decision more aggressive and leave later stages on their existing fixed penalty factors, or

This would have the issue that it does not really force the rule to be applied. We had cases where even if the stage-1 was set to zero cost, the rule would not be triggered because later nodes induce too much cost. 

> only discount the additional internal skew-elimination exchange(s), without also zeroing the later aggregate nodes themselves.

I think in this case we can still end up with the 3-stage plan being cheaper and therefore the rule not being forced although we have skew and the 3-stage plan will take way longer to execute than the 4-stage plan.

I agree that the logic is not very robust, but I think we need some way to force activation of the rule when there is skew. E.g. for the joins or window skew we also do not make any cost assumption, because we assume that when there is skew the original plan will be very suboptimal. I think it is the very same case here. 

Maybe there should be another rule that does the same transformation after cost-based optimization? (Similar to how it is done for the join)
Then there first would be the cost consideration, and as a final step we could have a `TreeRewriteRule` that is always applied if there is considerable amountof skew. So basically my proposal to handle this better would be to add another rule. If we do not want that I think a change like in this PR is the only other option that comes to my mind.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
```diff
@@ -377,6 +382,10 @@ private double computeDataSkewPenaltyOfGroupByCountDistinct(PhysicalHashAggregat
             // Basic skew detection on the group by columns. We are weighting this heavily if we know
             // that group by columns are skewed since this optimization has significant speed up then.
             if (isGroupBySkewed(groupByStatistics, inputStatistics)) {
+                skewInfo.setGroupBySkewDetected(true);
```
stephen-shelby: Consider setting groupBySkewDetected only when enableForceGroupBySkewEliminateWhenSkewed is true—same gating as all readers—default-off queries still return 0.2 here.

OliLay: [08a73b5](https://github.com/StarRocks/starrocks/pull/71382/commits/08a73b56ea89e2696d34cb794cfb5a97587467ab)
Done

stephen-shelby (approved):

trueeyu (approved):

OliLay:
https://github.com/Mergifyio backport branch-3.5-cc

PR #71386, opened 2026-04-07
[Feature] Support aggregatefunction pushdown for ClickHouse

### LINKED ISSUE #53950: Support Query ClickHouse AggregatingMergeTree Engine Table.
### StarRocks as Query Engine
Our team is trying to use StarRocks as a unified query engine layer. We are encountering issues when querying materialized views in ClickHouse.

e.g.

### ClickHouse
```sql
CREATE DATABASE db0;
CREATE TABLE db0.simple (id UInt64, val SimpleAggregateFunction(sum, UInt64)) ENGINE=AggregatingMergeTree() ORDER BY id;
INSERT INTO db0.simple(id, val) VALUES (1, 10),(1, 11),(2, 12);
```

### Expected Result
```text
ClickHouse :) select id, sum(val) from db0.simple GROUP BY id;

SELECT
    id,
    sum(val)
FROM db0.simple
GROUP BY id

Query id: da27ce92-1f98-4790-9a3d-00f5e9d72b03

   ┌─id─┬─sum(val)─┐
1. │  2 │       12 │
2. │  1 │       21 │
   └────┴──────────┘

2 rows in set. Elapsed: 0.004 sec.
```

### StarRocks
```sql
CREATE EXTERNAL CATALOG clickhouse
COMMENT 'example'
PROPERTIES
(
    "type"="jdbc", 
    "user"="default",
    "password"="",
    "jdbc_uri"="jdbc:clickhouse:[your ClickHouse address]:8123",
    "driver_url"="file:///[url]/clickhouse-jdbc-0.7.1-all.jar",
    "driver_class"="com.clickhouse.jdbc.ClickHouseDriver"
);
```

### Query ClickHouse table from StarRocks
```sql
SELECT id, sum(val) FROM clickhouse.db0.simple GROUP BY id;
```

Got Error:
```log
ERROR 1064 (HY000): Getting analyzing error. Detail message: Datatype of external table column [val] is not supported!.
```

### Features Required

1. Support AggregateFuntion types
2. Support Aggregation pushdown

limc5462:
## Why I'm doing:
ClickHouse uses specialized column types such as AggregateFunction and SimpleAggregateFunction to store intermediate aggregation states in its AggregatingMergeTree/SummingMergeTree engines. To obtain the correct final result, these intermediate states must be processed using corresponding ClickHouse aggregation functions (e.g., sumMerge). Pushing down the aggregation is not only a performance optimization but also a requirement to correctly resolve ClickHouse's aggregate models, as StarRocks cannot natively process these remote intermediate binary states.
## What I'm doing:

Fixes [#issue](https://github.com/StarRocks/starrocks/issues/53950)

1.Identifier Mapping: Enhanced ClickhouseSchemaResolver to identify ClickHouse-specific aggregate types and map them to StarRocks internal AGG_STATE_UNION metadata.
2.Optimizer Rule: Implemented PushDownAggToJDBCScanRule to automatically transform logical aggregation plans into remote-executable SQL, specifically targeting these aggregate models.
3.Thrift & Plan Enhancement: Added group_by_columns to TJDBCScanNode to ensure the grouping context is preserved during pushdown.
4.Test Verification: Added JDBCAggregationPushdownPlanTest to verify that the generated remote SQL correctly uses *Merge functions for AggregateFunction columns and standard functions for SimpleAggregateFunction columns.
## What type of PR is this:

- [x] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
image
image


metegenez (commented):
Thanks for working on this. It's a highly correct requirement on Clickhouse and probably every engine available in JDBC and future connections types, like ODBC or ADBC. That said, I have some architectural concerns about the approach. **- My concerns are not enforcing since i don't hold any position in StarRocks-**

**Core issue: this solves a general problem with a ClickHouse-specific solution**
JDBC catalog is not ClickHouse-only. Aggregate pushdown, GROUP BY pushdown, and dialect-aware SQL generation are problems that will recur for every source. Today the JDBC pushdown path constructs remote SQL by concatenating raw strings in two separate places: FE side in `JDBCScanNode.getJDBCQueryStr()` and BE side in `jdbc_connector.cpp::get_jdbc_sql()`, with ad-hoc special cases for Oracle, SQL Server, etc. Every new pushdown capability extends this concatenation, and each one introduces new quoting hacks. In this PR, `colName.contains("(")` is used to skip identifier quoting for aggregate expressions like `sumMerge(col)`. That's fragile.

**What I think is missing is a fragment-level SQL unparser with dialect support**, like Calcite's `RelToSqlConverter` or DataFusion's `plan_to_sql`. The optimizer rule marks that an aggregation should be pushed to the remote source, the unparser takes the pushed fragment and produces a complete, dialect-correct SQL string on the FE side, and the BE receives the final query rather than reassembling it from decomposed parts.

Adding `group_by_columns` to `TJDBCScanNode` is a design smell: we already have `columns` and `filters` as string lists in Thrift, now `group_by_columns`, next it'll be `having_clauses` and `order_by_columns`. At some point what we're transmitting is a decomposed SQL statement.

**Suggestion:** I'd suggest scoping this PR to the schema resolver changes (recognizing the ClickHouse aggregate types and mapping them correctly) and opening a separate design discussion for the pushdown mechanism. The pushdown part deserves a more general solution that benefits all JDBC sources, not just ClickHouse. A fragment unparser with dialect support would be the right foundation.

Issue #67447, opened 2026-01-05
information table query does not handle escape character correctly


dirtysalt:
### Steps to reproduce the behavior (Required)

Suppose there is a table called `a_a`.  if you query information schema table using predicate like `tablename like "a\\_a", the starrocks does not not return any table. 

```
mysql> select table_name from information_schema.tables where table_name like 'a\\_a';
Empty set (0.04 sec)

mysql> select table_name from information_schema.tables where table_name like 'a_a';
+------------+
| table_name |
+------------+
| a_a        |
+------------+
1 row in set (0.03 sec)
```

### StarRocks version (Required)

Probably introduced by this pr: https://github.com/StarRocks/starrocks/pull/66200

PR #71407, opened 2026-04-08
[BugFix] fix information_schema.tables not escaping special characters in equality predicates (backport #71273)

dontknow9179:
## Why I'm doing:

When querying `information_schema.tables` with an equality predicate like:

```sql
SELECT * FROM information_schema.tables
WHERE table_schema = 'my_db' AND table_name = 'a_a';
```

The value `a_a` was passed directly to `PatternMatcher.createMysqlPattern()` as a LIKE pattern, where `_` is treated as a single-character wildcard. This caused the query to return unrelated tables like `aba`, `a1a`, etc., instead of only the exact table `a_a`.

Similarly, table names containing backslashes (e.g. `a\_a`) were not handled correctly — the backslash and underscore interacted with LIKE pattern escaping rules in unexpected ways.

Fixes https://github.com/StarRocks/starrocks/issues/67447

## What I'm doing

Fix `information_schema.tables` queries with equality predicates (`=`) on `table_name` or `table_schema` incorrectly treating LIKE-special characters (`_`, `%`, `\`) as wildcards instead of literal characters.

### Changes

1. **`PatternMatcher.convertMysqlPattern`**: Rewrite the two-pass conversion to a single-pass approach that correctly escapes regex metacharacters (e.g. `(`, `)`, `+`, `[`, `]`) in table/database names, and properly handles a trailing backslash as a literal character.

2. **`PatternMatcher.escapeLikeValue`**: Add a new utility method that escapes a literal string so it can be safely used as a MySQL LIKE pattern for exact matching (prefixing `\`, `%`, and `_` with a backslash).

3. **`TablesSystemTable.evaluate`**: Call `escapeLikeValue()` on the value extracted from `=` predicates before passing it to the Thrift request. This is the **primary fix** — the `SchemaTableEvaluateRule` optimizer rule short-circuits simple equality queries on `information_schema.tables` and evaluates them directly in the FE without going through `PlanFragmentBuilder.visitPhysicalSchemaScan`.

4. **`PlanFragmentBuilder.visitPhysicalSchemaScan`**: Also call `escapeLikeValue()` for `=` predicates on `TABLE_NAME`/`TABLE_SCHEMA`/`DATABASE_NAME` in the SchemaScan path, which is used when `SchemaTableEvaluateRule` is disabled or when the predicate is not a simple equality.

5. **`InformationSchemaDataSource.generateTablesInfoResponse`**: Remove the fallback OR-logic in `matchPattern` and use `matcher.match()` directly, preventing false-positive matches when the pattern matcher rejects a name but the raw string comparison accidentally matches.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4


gengjun-git (approved):

trueeyu:

PR #71448, opened 2026-04-08
[Enhancement] Allow creating CTE when merging complex projects

lxynov:
## Why I'm doing:

When `MergeTwoProjectRule` inlines the lower project's expressions into the upper project, the result can exceed the `max_scalar_operator_flat_children` limit and fail with "Expression too complex". A new config `merge_project_cte_rewrite_max_flat_children` (default 0 / disabled) controls a softer threshold: when enabled, the lower project is materialized as a forced CTE instead of being inlined only when
- a merged expression's flat children count exceeds this threshold,
- actually increases compared to before the merge, and
- exceeds the maximum complexity of the lower project;

otherwise, existing behavior is preserved.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/MergeTwoProjectRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/MergeTwoProjectRule.java
@@ -17,9 +17,14 @@ package com.starrocks.sql.optimizer.rule.transformation;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.starrocks.catalog.FunctionSet;
+import com.starrocks.common.Config;
 import com.starrocks.sql.optimizer.OptExpression;
 import com.starrocks.sql.optimizer.OptimizerContext;
+import com.starrocks.sql.optimizer.base.ColumnRefFactory;
 import com.starrocks.sql.optimizer.operator.OperatorType;
+import com.starrocks.sql.optimizer.operator.logical.LogicalCTEAnchorOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalCTEConsumeOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalCTEProduceOperator;
 import com.starrocks.sql.optimizer.operator.logical.LogicalProjectOperator;
 import com.starrocks.sql.optimizer.operator.pattern.Pattern;
 import com.starrocks.sql.optimizer.operator.scalar.CallOperator;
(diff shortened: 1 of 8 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/MergeTwoProjectRule.java
```diff
@@ -44,15 +49,43 @@ public List<OptExpression> transform(OptExpression input, OptimizerContext conte
         LogicalProjectOperator firstProject = (LogicalProjectOperator) input.getOp();
         LogicalProjectOperator secondProject = (LogicalProjectOperator) input.getInputs().get(0).getOp();
 
+        int maxMergeFlatChildren = Config.merge_project_cte_rewrite_flat_children;
+        int maxBottomFlatChildren = 0;
+        if (maxMergeFlatChildren > 0) {
+            for (ScalarOperator expr : secondProject.getColumnRefMap().values()) {
+                maxBottomFlatChildren = Math.max(maxBottomFlatChildren, expr.getNumFlatChildren());
+            }
+        }
```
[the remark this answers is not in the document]
lxynov: > A more accurate guard would compute the bottom-side max only for column refs actually used by the current top expression

This would complicate the logic and incur additional compute overhead. I'd prefer not to implement this.

lxynov:
Addressed most of the comments.

Also added more code comments for clarity.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/MergeTwoProjectRule.java
```diff
@@ -44,11 +49,42 @@ public List<OptExpression> transform(OptExpression input, OptimizerContext conte
         LogicalProjectOperator firstProject = (LogicalProjectOperator) input.getOp();
         LogicalProjectOperator secondProject = (LogicalProjectOperator) input.getInputs().get(0).getOp();
 
+        // cteRewriteMaxFlatChildren > 0 means to enable CTE rewrite if the merged project has an expression whose flat children
+        // count is above cteRewriteMaxFlatChildren.
+        int cteRewriteMaxFlatChildren = Config.merge_project_cte_rewrite_max_flat_children;
```
stephen-shelby, line 54: can't [70802](https://github.com/StarRocks/starrocks/pull/70802) solve this problem?

lxynov: it may. (prod rollout blocked by the issue mentioned on slack)

This PR might be a better solution. With https://github.com/StarRocks/starrocks/pull/70802, we have to design materialization rules in pql2sql translation, which can be hard.

I also think this PR is generally helpful because it makes queries which previously fail with "Expression too complex" errors successful now.

But I also understand it adds complexity to `MergeTwoProjectRule`.

Will let you guys decide whether to merge it. For our business we could try https://github.com/StarRocks/starrocks/pull/70802 first.

PR #71461, opened 2026-04-09
[BugFix] preserve varchar length after reduce cast with global variables (backport #70269)

jaogoy:
## Why I'm doing

For statements such as `CAST(col AS VARCHAR(N))`, FE may return different result-set metadata between prepare and execute. This is visible to MSDASQL/ODBC clients and can cause client-side metadata drift.

The root cause is that `ReduceCastRule` removes same-type `VARCHAR -> VARCHAR` casts during optimization, but the target `VARCHAR(N)` length metadata can be lost after the cast is reduced. In the planner path, reused `Expr` and `SlotRef` objects may still keep the original column varchar length, so execute can expose a different output type than prepare.

## What I'm doing

- Add a global switch `enable_reduce_cast_varchar_length_inheritance`. (default `false`)
  - Preserve the target varchar length when `ReduceCastRule` eliminates a same-type `VARCHAR -> VARCHAR` cast.
- Add a global switch `enable_reduce_cast_varchar_expr_sync_type`. (default `false`)
  - Keep it as a switch when planner `Expr` metadata also needs to be synchronized when `inheritance` is `true`.

Others:
- Add planner-level regression tests for final output slot metadata.
- Add sql-test by using `function: xxx` to call a python function, which will using row mysql protocol to check the varchar length from `prepare` and `execute`. (Other commonly used python mysql drivers can get metadata for `prepare` interface)
- Document the new global variables and their introduced versions.

Fixes #issue

## Usage

Enable the fix with:

```sql
SET GLOBAL enable_reduce_cast_varchar_length_inheritance = true;
SET GLOBAL enable_reduce_cast_varchar_expr_sync_type = true;
```

## Impact

This fixes prepare/execute varchar length metadata drift for clients that rely on result-set metadata, especially MSDASQL/ODBC. Default behavior stays unchanged unless the inheritance switch is enabled, and the change is scoped to same-type `VARCHAR -> VARCHAR` cast reduction.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


stephen-shelby (approved):

trueeyu:

PR #71483, opened 2026-04-09
[BugFix] Fix batch publish deadlock caused by partition version gap

xiangguangyxg:
## Why I'm doing:

  Several operations (tablet split, schema change, alter job, etc.) reserve phantom versions for
  partitions by advancing `nextVersion` without a corresponding publish, creating a version gap
  (e.g., `visibleVersion=8` but the next committed version is `10` instead of `9`).

  If a transaction touching a gap-affected partition is included in a batch publish,
  `publishPartitionBatch()` fails the version check (`visibleVersion + 1 != versions.get(0)`),
  and the ALL-or-nothing batch semantics (`allMatch`) cause the **entire batch** to fail — blocking
  all transactions in the batch, including those not affected by the gap. This creates a permanent
  deadlock because the job that reserved the phantom version is waiting for other partitions'
  visibleVersions to advance, while those partitions' transactions are blocked by the failed batch.

## What I'm doing:

  Add `trimBatchAtVersionGap()` in `PublishVersionDaemon` to detect version gaps before publishing.
  For each partition touched by the batch, the method checks whether `visibleVersion + 1` matches
  the first committed version. If a gap is detected, the batch is truncated before the first
  transaction that touches the gap-affected partition, so only the publishable prefix is sent.

  Key properties:
  - A single trimmed `TransactionStateBatch` object flows through publish → finish → deleteLog,
    avoiding inconsistency between original and trimmed batch references.
  - If trimmed to size 1, the batch naturally enters the single-txn `publishLakeTransactionAsync()`
    path with full `versionTime` backoff support.
  - Only publishable txns are marked `hasSendTask`.
  - If the head txn touches a gap partition (entire batch unpublishable), returns `null` and the
    batch is skipped with `continue`, no publish attempted.
  - REPLICATION txns are excluded from the gap check (non-consecutive versions by design).

  Once the publishable prefix is finished, the job that reserved the phantom version becomes
  unblocked and publishes it, filling the gap. Subsequent publish cycles then pick up the
  remaining transactions normally.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
@@ -79,6 +79,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -87,6 +88,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
 
 public class PublishVersionDaemon extends FrontendDaemon {
@@ -468,6 +470,17 @@ public class PublishVersionDaemon extends FrontendDaemon {
         Set<Long> publishingLakeTransactionsBatchTableId = getPublishingLakeTransactionsBatchTableId();
         Set<Long> publishingTransactions = getPublishingTransactions();
         for (TransactionStateBatch txnStateBatch : readyTransactionStatesBatch) {
+            // Trim batch at version gap boundary to avoid deadlock with SplitTabletJob.
+            // See trimBatchAtVersionGap() javadoc for details.
+            try {
+                txnStateBatch = trimBatchAtVersionGap(txnStateBatch);
+            } catch (Exception e) {
+                LOG.warn("Failed to trim batch at version gap, proceeding with original batch", e);
+            }
+            if (txnStateBatch == null) {
+                continue; // entire batch unpublishable (head txn has gap), skip this cycle
+            }
+
             if (txnStateBatch.size() == 1) {
                 // there are two situations:
                 // 1. the transactionState in txnStateBatch is with multi-tables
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -773,6 +786,103 @@ private void submitDeleteTxnLogJob(TransactionStateBatch txnStateBatch) {
...
+     * caused by SplitTabletJob reserving a phantom version), the batch is truncated before
+     * the first transaction that touches the gap-affected partition.
+     *
+     * @return the original batch if no gap is found, or a new trimmed batch containing
+     *         only the publishable prefix. Returns null if the entire batch is unpublishable
+     *         (first transaction already touches a gap partition).
+     */
+    @VisibleForTesting
+    @Nullable
+    TransactionStateBatch trimBatchAtVersionGap(TransactionStateBatch txnStateBatch) {
```
smartlxh, line 803: May move the check to TransactionGraph#getReadyToPublishTxnListBatch to determine is better？

xiangguangyxg: I initially intended to add the check to `getReadyToPublishTxnListBatch`, but that might cause a deadlock because it involves adding a table lock within the txn lock.

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -468,6 +470,17 @@ void publishVersionForLakeTableBatch(List<TransactionStateBatch> readyTransactio
         Set<Long> publishingLakeTransactionsBatchTableId = getPublishingLakeTransactionsBatchTableId();
         Set<Long> publishingTransactions = getPublishingTransactions();
         for (TransactionStateBatch txnStateBatch : readyTransactionStatesBatch) {
+            // Trim batch at version gap boundary to avoid deadlock with SplitTabletJob.
+            // See trimBatchAtVersionGap() javadoc for details.
+            try {
+                txnStateBatch = trimBatchAtVersionGap(txnStateBatch);
```
smartlxh, line 476: Consider this example, a batch contains two transactions, txn1 and txn2, involving partitions p1 and p2 respectively. A SplitTabletJob also involves partitions p1 and p2. If SplitTabletJob p1 is committed before txn1, but p2 is committed after txn2, wouldn't a deadlock still occur?

The  solution to this problem seems to be to batch the transactions according to the partition granularity. If there is no partition dependency between two transactions, then the transactions should be split into batches not the same batch.

smartlxh (approved):

sevev (approved):

PR #71485, opened 2026-04-09
[Performance] Reduce lake PK compaction max input rowsets from 500 to 100

luohaha:
## Why I'm doing:

In shared-data (lake) mode with primary key tables, compaction publishes with 400-500 input rowsets take **1.3-3.5 seconds** due to delvec loading and persistent index updates. The compaction publish holds a **per-tablet lock** that blocks stream load publishes for the same tablet. During bulk loads (which trigger large compactions), this causes ingestion P99 to spike by **+1111%** (from 6.2s to 75.3s in a 30GB benchmark with 500M rows, 3 CN nodes).

### Trace evidence from CN publish logs:

| Input Rowsets | delvec_loader Latency | Total Publish Cost |
|---------------|----------------------|-------------------|
| 81 | 24ms | 1,252ms |
| 257 | 403ms | 1,915ms |
| 327 | 410ms | 3,011ms |
| 469 | 1,166ms | 3,419ms |
| 482 | 1,312ms | 3,444ms |

The delvec_loader latency scales super-linearly with input rowsets above ~300, because more remote delvec file reads exceed the cache capacity.

### Root cause chain:
1. Bulk load → many small rowsets accumulated on tablet
2. Compaction picks up to 500 rowsets (current default of `lake_pk_compaction_max_input_rowsets`)
3. Compaction publish loads delvec for all input rowsets: **1.3s** at 500 rowsets
4. Per-tablet lock blocks stream load publishes during compaction publish
5. Stream load publishes queue up → **P99 spike**

## What I'm doing:

Reduce `lake_pk_compaction_max_input_rowsets` default from **500** to **100**.

With 100 input rowsets, delvec_loader stays under **~100-200ms** per compaction publish, and total publish time drops from **3.5s to ~1-1.5s**. This allows stream load publishes to interleave between compactions instead of queueing behind long-running compaction publishes.

**Trade-off**: More frequent but shorter compactions. Each compaction processes fewer rowsets, so more compaction rounds are needed. But per-compaction overhead is low, and the latency tail is dramatically reduced. The config is mutable (`CONF_mInt64`), so users can tune it at runtime if needed.

## What type of PR is this:

- [x] Improvement

- [x] Yes

- [x] Default configuration value change (lake_pk_compaction_max_input_rowsets: 500 → 100)

## Checklist:

## Expected Impact:

Reduce compaction publish lock hold time by ~60-70%, reducing stream load ingestion P99 during bulk loads. Expected to reduce P99 spike during bulk loads from +1111% to ~+200-400%.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1326,7 +1326,7 @@ CONF_mInt64(lake_vacuum_retry_min_delay_ms, "100");
 CONF_mInt64(lake_max_garbage_version_distance, "100");
 CONF_mBool(enable_primary_key_recover, "false");
 CONF_mBool(lake_enable_compaction_async_write, "false");
-CONF_mInt64(lake_pk_compaction_max_input_rowsets, "500");
+CONF_mInt64(lake_pk_compaction_max_input_rowsets, "100");
 CONF_mInt64(lake_pk_compaction_min_input_segments, "5");
 // Enable cleanup of orphan delvec entries during compaction.
 // Orphan delvecs are leaked metadata entries from a historical bug that reference


luohaha:
## Benchmark Result: SEVERE REGRESSION — Closing This PR

Tested on a clean cluster (1 FE 8c32g, 3 CN 16c64g, shared-data mode) with the 30GB single-table benchmark (500M rows, 64 query threads, 100K rows/sec ingest, 30-min steady state).

### Results vs Baseline

| Metric | Baseline | This PR | Change |
|--------|----------|---------|--------|
| Query QPS | 26.2 | **1.5** | **-94%** |
| Query P50 | 202ms | **1,506ms** | +646% |
| Ingest P50 | 6,407ms | **19,156ms** | +199% |
| Ingest P99 | 76,116ms | **105,813ms** | +39% |
| Query Error Rate | 0.016% | **6.9%** | SEVERE |

### Root Cause

Reducing compaction max input rowsets from 500 to 100 causes **5x more compaction operations**. The resulting IO contention overwhelms the system:

- delvec_loader at 100 rowsets: predicted 100-200ms, **actual P50 = 3,000ms, P90 = 21,000ms, max = 184,657ms**
- Compaction IO starvation causes queries to time out at 300s and stream loads to queue
- Total delvec IO volume is the same (100 rowsets × 5x compactions ≈ 500 rowsets × 1x compaction), but spreading it across more concurrent operations causes worse cache thrashing

### Conclusion

Reducing the compaction rowset limit is counterproductive for this workload. The delvec bottleneck needs to be addressed through caching improvements or IO scheduling, not by changing compaction granularity.

Closing this PR.

PR #71500, opened 2026-04-09
[Performance] Pre-fetch compaction delvecs before per-tablet lock to reduce stream load P99

luohaha:
## Why I'm doing:

In shared-data (lake) mode with primary key tables, compaction publishes with 500 input rowsets hold the **per-tablet lock** for **1-8 seconds** while loading delete vectors (delvecs) from remote storage. During this time, stream load publishes for the same tablet return `ResourceBusy`, causing the FE to retry in the next publish cycle. This cascading retry pattern causes `wait_for_publish` to reach **57 seconds at P99**.

### Benchmark evidence (30GB single-table, 500M rows, 3 CN nodes):

**Stream load latency breakdown (FE-side, 51,372 transactions):**

| Component | P50 | P90 | P99 | Max |
|-----------|-----|-----|-----|-----|
| total cost | 1,231ms | 9,187ms | 67,856ms | 558,805ms |
| write cost | 254ms | 446ms | 7,418ms | 417,841ms |
| **wait_for_publish** | **381ms** | **5,683ms** | **57,688ms** | **161,922ms** |
| publish_rpc | 573ms | 2,549ms | 7,679ms | 59,092ms |

**Compaction delvec_loader latency (per tablet publish):**

| CN | P50 | P90 | Max |
|----|-----|-----|-----|
| CN1 | 9ms | 3,314ms | 74,418ms |
| CN2 | 16ms | 1,950ms | 21,824ms |
| CN3 | 5ms | 637ms | 65,644ms |

### Root cause chain:
1. Compaction publish loads delvec for each input rowset segment sequentially (500 reads × 10-300ms each)
2. This happens under the per-tablet lock (`tablet_txns` in transactions.cpp)
3. Stream load publish for the same tablet returns `ResourceBusy`
4. FE retries in next publish cycle → cascading `wait_for_publish` delays up to 57s

## What I'm doing:

Pre-fetch delvec data into cache **before** acquiring the per-tablet lock. Since delvec data is immutable and versioned, pre-loading into cache is safe and doesn't affect correctness.

**Before:** `acquire_lock → load_delvecs(1-8s) → update_index → release_lock`
**After:** `prefetch_delvecs(1-8s, NO LOCK) → acquire_lock → update_index(cache hit, <100ms) → release_lock`

The implementation:
1. Before `add_tablet()`, load the txn log and detect compaction operations
2. For compaction txns, load the base metadata and pre-fetch all delvecs for input rowset segments into metacache
3. When the actual publish runs under the lock, all delvec reads hit cache

**Expected impact:**
- Tablet lock hold time: 1-8s → <200ms
- Stream load `wait_for_publish` P99: 57s → <5s
- Ingest P99: 125s → estimated <10s

**Graceful degradation:** If pre-fetch fails (e.g., metadata not found), publish proceeds normally with the same behavior as before.

## What type of PR is this:

- [x] Improvement

- [x] Yes

- [x] Performance optimization (delvec loading moved outside per-tablet lock)

## Checklist:

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1772,6 +1772,9 @@ CONF_mInt32(olap_string_max_length, "1048576");
 // Skip get from pk index when light pk compaction publish is enabled
 CONF_mBool(enable_light_pk_compaction_publish, "true");
 
+// Pre-fetch delvec data for compaction publishes before acquiring the per-tablet lock.
+CONF_mBool(lake_enable_compaction_delvec_prefetch, "true");
+
 // jit LRU object cache size for total 32 shards, it will be an auto value if it < 0
 // mem_limit = system memory or process memory limit if set.
 // if mem_limit < 16 GB, disable JIT.
--- a/be/src/common/config_compaction_fwd.h
+++ b/be/src/common/config_compaction_fwd.h
@@ -162,6 +162,11 @@ CONF_mInt64(lake_pk_compaction_min_input_segments, "5");
 // Skip get from pk index when light pk compaction publish is enabled
 CONF_mBool(enable_light_pk_compaction_publish, "true");
 
+// Pre-fetch delvec data for compaction publishes before acquiring the per-tablet lock.
+// This reduces lock hold time from seconds to milliseconds, preventing compaction publishes
+// from blocking stream load publishes on the same tablet.
+CONF_mBool(lake_enable_compaction_delvec_prefetch, "true");
+
 // if turned on, each compaction will use at most `max_cumulative_compaction_num_singleton_deltas` segments,
 // for now, only support non-pk LAKE compaction in size tierd compaction.
 CONF_mBool(enable_lake_compaction_use_partial_segments, "false");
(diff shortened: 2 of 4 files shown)


CelerData-Reviewer:
@codex review

luohaha:
Closing: regression confirmed in 2 independent benchmark runs (iter 020: -57% QPS, iter 021: -29% QPS). Pre-fetch creates massive IO contention that overwhelms the cache benefit. Delvec_loader latency under lock increased 72-484x vs baseline. Thread pool exhaustion on 2/3 CNs.

The wait_for_publish P99 did improve (-44%), validating that compaction lock hold time IS the bottleneck for stream load P99. But the implementation approach (pre-loading into cache before lock) doesn't work under concurrent workload pressure — cache eviction and IO bandwidth saturation negate the benefit.

Alternative approaches to explore: inline delvec (PR #71154), batched parallel delvec reads with controlled concurrency inside lock, or separating compaction/stream-load publish paths.

PR #71511, opened 2026-04-10
[Performance] Replace non-blocking tablet publish lock with timed-wait to reduce stream load P99

luohaha:
## Why I'm doing:

In shared-data (lake) mode with primary key tables, compaction publishes hold the **per-tablet lock** (`tablet_txns` in `transactions.cpp`) for **1-8 seconds** while loading delete vectors (delvecs) from remote storage. During this time, stream load publishes for the same tablet receive `ResourceBusy` and bounce back to the FE PublishVersionDaemon, which retries on the next cycle (~1s interval).

This creates a **cascading retry pattern**: each retry incurs ~1s of FE daemon cycle latency plus RPC overhead. With compaction holding the lock for 5s, a stream load may retry 5+ times, accumulating delays. When multiple stream loads queue behind the same compaction, the cascade compounds — `wait_for_publish` reaches **57 seconds at P99**.

### Benchmark evidence (30GB single-table, 500M rows, 3 CN nodes):

**Stream load latency breakdown (FE-side, 51,372 transactions):**

| Component | P50 | P90 | P99 | Max |
|-----------|-----|-----|-----|-----|
| total cost | 1,231ms | 9,187ms | 67,856ms | 558,805ms |
| write cost | 254ms | 446ms | 7,418ms | 417,841ms |
| **wait_for_publish** | **381ms** | **5,683ms** | **57,688ms** | **161,922ms** |
| publish_rpc | 573ms | 2,549ms | 7,679ms | 59,092ms |

Stream load publish itself takes only **100-335ms** when not blocked by compaction. The 57s P99 is entirely caused by the FE retry cascade.

### Previous approaches that failed:

| Approach | PR | Result | Why It Failed |
|----------|-----|--------|--------------|
| Pre-fetch delvecs outside lock | #71500 | -29% to -57% QPS | Added IO contention that saturated remote storage |
| Concurrent batch delvec loading | #71184 | -21% to -41% QPS | Same IO contention issue |
| Reduce max input rowsets 500→100 | #71485 | -94% QPS | 5x more compactions caused worse IO contention |

All previous approaches tried to speed up delvec loading (either by parallelizing or moving it outside the lock). They all failed because adding more IO operations — even read-only — saturates remote storage bandwidth under this workload.

## What I'm doing:

**Key insight**: The problem is not that delvec loading is slow — it's that the FE retry mechanism amplifies 1-8s of lock contention into 30-57s of end-to-end latency. The fix should target the retry cascade, not the IO.

This change replaces the non-blocking `ParallelSet<int64_t>` tablet lock with a `TabletPublishLock` that supports **timed waiting via condition variables**:

**Before (current):**
```
Stream load publish → tablet busy → return ResourceBusy immediately
→ FE waits ~1s → retry → still busy → return ResourceBusy
→ FE waits ~1s → retry → ... (5-50+ retries) → finally succeeds
Total: 30-57s at P99
```

**After (this PR):**
```
Stream load publish → tablet busy → wait on condition variable (up to 10s)
→ compaction finishes (1-8s) → condition notified → proceed immediately
Total: 1-8s (bounded by actual compaction duration)
```

### Implementation details:

- `TabletPublishLock`: A sharded lock class (16 shards) using `std::condition_variable` per shard. When `try_lock()` finds a tablet already locked, it blocks with `wait_until()` instead of returning false immediately.
- New config: `lake_publish_version_tablet_wait_timeout_ms` (default: 10000ms). Set to 0 for original behavior.
- `notify_all()` on unlock wakes all waiting publishers for that shard.

### Why this works where prefetch failed:

- **Zero additional IO**: No prefetch reads, no cache warming, no bandwidth impact
- **Precise waiting**: Stream load waits exactly until compaction releases the lock — no 1s daemon cycle overhead per retry
- **Bounded latency**: Worst case is the timeout value (10s), not unbounded retry cascading
- **Graceful fallback**: If timeout expires, returns ResourceBusy as before (FE retries normally)

### Expected impact:

- **wait_for_publish P99**: 57s → bounded by compaction lock hold time (1-8s)
- **Stream load P99**: 67-125s → estimated <15s
- **QPS**: Should improve due to reduced publish queueing

### Risks:

- BE threads block while waiting (instead of returning immediately). With 3 tablets per CN and typical compaction rates, at most 2-3 threads would be blocked at any time — well within thread pool capacity on 16c64g nodes.
- If compaction publish takes >10s (rare edge case), behavior degrades to original ResourceBusy retry.

## What type of PR is this:

- [x] Improvement

- [x] Yes

- [x] Performance optimization (timed wait replaces immediate ResourceBusy for tablet publish conflicts)

## Checklist:

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1319,6 +1319,10 @@ CONF_mInt64(experimental_lake_wait_per_get_ms, "0");
 CONF_mInt64(experimental_lake_wait_per_delete_ms, "0");
 CONF_mBool(experimental_lake_ignore_pk_consistency_check, "false");
 CONF_mInt64(lake_publish_version_slow_log_ms, "1000");
+// When a tablet is already being published, wait up to this duration (ms) for it to finish
+// before returning ResourceBusy. This avoids FE retry cascading that causes high P99 latency.
+// Set to 0 to disable waiting (original behavior).
+CONF_mInt64(lake_publish_version_tablet_wait_timeout_ms, "10000");
 CONF_mBool(lake_enable_publish_version_trace_log, "false");
 CONF_mString(lake_vacuum_retry_pattern, "*request rate*");
 CONF_mInt64(lake_vacuum_retry_max_attempts, "5");
--- a/be/src/storage/lake/transactions.cpp
+++ b/be/src/storage/lake/transactions.cpp
@@ -14,8 +14,15 @@
 
 #include "storage/lake/transactions.h"
 
+#include <array>
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
+#include <unordered_set>
+
 #include "base/container/lru_cache.h"
 #include "base/utility/defer_op.h"
+#include "common/config.h"
 #include "common/config_lake_fwd.h"
 #include "fs/fs_factory.h"
 #include "fs/fs_util.h"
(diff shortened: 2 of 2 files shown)


luohaha:
Closing: Benchmark shows -61% QPS regression due to thread pool starvation. Timed-wait blocks publish threads for up to 10s, saturating the thread pool. All 5 per-tablet lock optimization approaches have now been tested and all cause regression. See iteration 022 diagnosis for details.

PR #71529, opened 2026-04-10
[BugFix] Validate Iceberg incremental snapshot lineage before IVM planning

Youngwb:
## Why I'm doing:
When Iceberg snapshots between the MV's last-refresh checkpoint and the current table version have been expired (e.g., by Iceberg's snapshot expiration mechanism), the incremental snapshot lineage becomes broken. In this case, IVM (Incremental View Maintenance) cannot correctly compute the delta between the two snapshots, potentially leading to incorrect refresh results or errors. Additionally, when AUTO refresh mode has no checkpoint yet (first refresh), it should not attempt IVM planning since there is no valid starting snapshot to compute deltas from.

## What I'm doing:
Add snapshot lineage validation in IcebergMetadata.listTableDeltaTraits: Before walking the snapshot chain, validate that the starting snapshot (exclusive) is a parent ancestor of the ending snapshot using SnapshotUtil.isParentAncestorOf. If the lineage is broken (e.g., intermediate snapshots were expired), throw a StarRocksConnectorException to signal that incremental delta computation is not possible. Also fix the fromSnapshotIdExclusive type from long to Long to correctly handle the empty-snapshot case (first refresh with no checkpoint).

Add checkpoint guard in MVIVMBasedRefreshProcessor: When the MV is in AUTO refresh mode and the base table has no checkpoint (maxTvrDelta.start() is empty), throw a SemanticException to prevent IVM planning. This ensures the first refresh in AUTO mode falls back to PCT (Partition Change Tracking) refresh, which will establish the initial checkpoint for subsequent IVM refreshes.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -767,12 +767,20 @@ public class IcebergMetadata implements ConnectorMetadata {
             return Collections.emptyList();
         }
         // fromSnapshotExclusive can be empty, but toSnapshotInclusive must have a valid snapshot ID
-        final long fromSnapshotIdExclusive = fromSnapshotExclusive.getSnapshotId();
+        final Long fromSnapshotIdExclusive = fromSnapshotExclusive.isEmpty() ? null : fromSnapshotExclusive.getSnapshotId();
         final long toSnapshotIdInclusive =
                 toSnapshotInclusive.end().orElseThrow(() -> new StarRocksConnectorException(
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

GavinMar (approved):

dirtysalt (approved):

Astralidea (approved):

PR #71559, opened 2026-04-11
[Enhancement] Support LDAP direct bind authentication via DN pattern

jaogoy:
## Why I'm doing

In the existing LDAP authentication flow, when a user is created without an
explicit DN (e.g. `CREATE USER tom IDENTIFIED WITH authentication_ldap_simple`),
StarRocks must perform a two-step "search-and-bind": first bind as an admin to
search for the user's DN, then bind again with the user's credentials. This
requires configuring an admin account (bind_root_dn/bind_root_pwd) and granting
it search privileges, which adds operational overhead and may not be desirable
in environments where the user DN structure is predictable.

More importantly, many users integrate with third-party managed LDAP services
(e.g. cloud-hosted LDAP or partner-managed directories) where they do not have
admin privileges to configure a root DN or search account at all. The
search-and-bind mode is simply not viable for these users. Direct bind via DN
pattern provides a lightweight alternative that requires no admin credentials.

## What I'm doing

Add a new configuration parameter `authentication_ldap_simple_bind_dn_pattern`
that enables direct bind authentication. When set, the system constructs the
user DN from the pattern (using `${USER}` as a placeholder) and binds directly,
skipping the admin search step entirely.

Key changes:
- Add `authentication_ldap_simple_bind_dn_pattern` to FE Config (mutable)
- Implement `authenticateByPattern()` in LDAPAuthProvider with multi-pattern
  support (patterns separated by semicolon `;`, tried in order)
- Wire the new parameter through both AuthPlugin (legacy path) and
  SimpleLDAPSecurityIntegration (security integration path)
- Authentication priority: per-user DN > DN pattern > search-and-bind
- Add LDAP injection protection via `escapeLdapValue()` for pattern mode
- Add unit tests for direct bind, multi-pattern fallback, and edge cases
- Update documentation (EN/ZH/JA) for ldap_authentication, security_integration,
  and FE_configuration reference

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

HangyuanLiu (changes requested):
Requesting changes before merge for three issues.

1. DN correctness in the direct-bind path.
`authenticateByPattern()` returns the substituted bind principal verbatim, and `authenticate()` stores that value as `distinguishedName`. That is only correct when the pattern produces a real DN. For the newly supported UPN-style pattern such as `${USER}@abc.com`, the stored value is not a DN, which can break downstream LDAP group lookup and `permitted_groups` checks when the DN-based path is used.

2. DN and UPN semantics are conflated in one field.
This implementation currently treats the bind principal and the canonical distinguished name as the same thing. They are not always equivalent. A DN-shaped pattern is one case; a UPN-shaped pattern is another. Please either restrict this feature to DN-only patterns, or resolve/store the canonical DN separately after a successful bind.

3. Documentation is missing.
The PR body says documentation was added and the checklist marks user documentation as required, but the diff does not contain any `docs/` changes. Please add the promised docs, or update the PR description/checklist so it matches the actual change set.

test/sql/test_security_integration/T/test_security_integration
```diff
@@ -3,4 +3,11 @@ create security integration oidc properties("type"="authentication_jwt", "jwks_u
 show create security integration oidc;
 alter security integration oidc set ("principal_field"="preferred_name");
 show create security integration oidc;
-drop security integration oidc;
\ No newline at end of file
+drop security integration oidc;
+
+-- name: test_ldap_security_integration_with_bind_dn_pattern
+create security integration ldap_pattern properties("type"="authentication_ldap_simple", "authentication_ldap_simple_server_host"="localhost", "authentication_ldap_simple_server_port"="389", "authentication_ldap_simple_bind_dn_pattern"="uid=${USER},ou=People,dc=example,dc=com");
+[UC]show create security integration ldap_pattern;
```
HangyuanLiu, line 10: so, what are you checking in this case ?

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -2183,6 +2183,17 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true)
     public static String authentication_ldap_simple_bind_root_pwd = "";
 
+    /**
+     * The DN pattern for direct bind authentication for authentication_ldap_simple.
+     * Use ${USER} as a placeholder for the username.
+     * e.g. "uid=${USER},ou=People,dc=example,dc=com"
+     * Multiple patterns can be separated by colon, e.g. "uid=${USER},ou=A,dc=com:uid=${USER},ou=B,dc=com"
```
HangyuanLiu: use ";"

test/sql/test_security_integration/T/test_security_integration
jaogoy: Because Map will print non-determinate item list, and UT has cover this part. So, no detailed check here.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
jaogoy: OK

HangyuanLiu (approved):

nshangyimingCD (approved):

PR #71575, opened 2026-04-13
[BugFix] Add mem_tracker to AsyncDeltaWriter ExecutionQueue callbacks

luohaha:
## Why I'm doing:

During investigation of an OOM in INSERT INTO SELECT with `pipeline_dop=16`, we found that the `AsyncDeltaWriter`'s `ExecutionQueue` callback (`execute()` / `_execute()`) runs without any default mem_tracker set on the thread.

While individual operations like `DeltaWriter::write()` / `flush()` / `finish()` set `SCOPED_THREAD_LOCAL_MEM_SETTER` internally, memory allocations happening **between** these calls — task iteration, `shared_ptr` operations, error handling, task stat updates — escape tracking entirely because the `ExecutionQueue` thread has no default mem_tracker.

This is a defensive fix to ensure complete memory tracking coverage for the entire `ExecutionQueue` callback duration, consistent with the pattern used in `SinkIOBuffer::_process_chunk()`.

## What I'm doing:

Add `SCOPED_THREAD_LOCAL_MEM_SETTER(delta_writer->mem_tracker(), false)` at the top of:
- `be/src/storage/lake/async_delta_writer.cpp` `AsyncDeltaWriterImpl::execute()` (cloud-native path)
- `be/src/storage/async_delta_writer.cpp` `AsyncDeltaWriter::_execute()` (local path)

The `false` parameter disables memory limit checking, matching the behavior inside `DeltaWriter::write()` where the SCOPED setter is already used with check=false.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/async_delta_writer.cpp
+++ b/be/src/storage/async_delta_writer.cpp
@@ -33,6 +33,10 @@ int AsyncDeltaWriter::_execute(void* meta, bthread::TaskIterator<AsyncDeltaWrite
         return 0;
     }
     auto writer = static_cast<DeltaWriter*>(meta);
+    // Track all memory allocated in this ExecutionQueue callback under the DeltaWriter's
+    // mem_tracker. Without this, allocations between write/flush/commit calls escape tracking
+    // because the ExecutionQueue thread has no default mem_tracker set.
+    SCOPED_THREAD_LOCAL_MEM_SETTER(writer->mem_tracker(), false);
     bool flush_after_write = false;
     int num_tasks = 0;
     int64_t pending_time_ns = 0;
--- a/be/src/storage/lake/async_delta_writer.cpp
+++ b/be/src/storage/lake/async_delta_writer.cpp
@@ -23,6 +23,7 @@
 #include "base/testutil/sync_point.h"
 #include "common/compiler_util.h"
 #include "common/util/stack_trace_mutex.h"
+#include "runtime/current_thread.h"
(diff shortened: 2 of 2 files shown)


luohaha:
Closing this PR. After further investigation using the real-time /mem_tracker HTTP endpoint, I confirmed that mem_tracker is actually tracking memory correctly — query_pool and load trackers both show accurate peak values (5G and 12G respectively) during INSERT INTO SELECT.

The original OOM investigation was misled by daemon.cpp log samples that happened AFTER query cancellation, showing query_pool=0 which was just reflecting the post-cleanup state, not a tracking gap.

The real root cause of the OOM was a configuration issue: `datacache_mem_size = 20G` on a 29GB machine was consuming 77% of BE's memory budget, leaving insufficient space for query/load execution. Reducing it to 4G (or using the default 20%) resolves the OOM without any code change.

The defensive fix to add SCOPED_THREAD_LOCAL_MEM_SETTER in ExecutionQueue callbacks is marginal at best — all significant allocations already happen inside DeltaWriter::write() which has its own SCOPED setter.

PR #71576, opened 2026-04-13
[Enhancement] Support IN predicate for HMS partition name filtering in external tables

konas:
## Why I'm doing:

When querying Hive external tables with multi-level partitions, using multi-value `IN` predicates on partition columns causes a severe planning performance regression.

**Root cause:** `getEffectivePartitionPredicate()` in `OptExternalPartitionPruner` only recognizes simple EQ predicates (`column = constant`) for HMS-side partition filtering. A single-value `IN('a')` gets rewritten to `= 'a'` by `SimplifiedPredicateRule` and triggers the fast path (`listPartitionNamesByValue()`), but multi-value `IN('a', 'b')` stays as `InPredicateOperator` and falls back to `listPartitionNames()`, which fetches **all** partitions from HMS.

**Impact:** For tables with tens of thousands of partitions, planning time degrades from ~200ms to 10+ seconds.

**Example - slow query (10+ seconds planning):**
```sql
WHERE "$part_event" IN ('dws_users_metrics_df', 'purchase')
  AND "$part_date" BETWEEN '2025-05-30' AND '2025-06-12'
```

**Example - fast query (~200ms planning):**
```sql
WHERE "$part_event" IN ('dws_users_metrics_df')
  AND "$part_date" BETWEEN '2025-05-30' AND '2025-06-12'
```

## What I'm doing:

Added an IN predicate recognition path in `OptExternalPartitionPruner.initPartitionInfo()` between the existing EQ path and the full-fetch fallback:

1. **`getColumnINConstantValues()`** — Extracts `column IN (v1, v2, ...)` constant values from top-level conjuncts.
2. **`getEffectiveInPartitionValues()`** — For each STRING-type partition column, checks if an IN predicate exists and returns its values.
3. **`listPartitionNamesForInValues()`** — Generates cartesian product of IN values across partition columns (capped at 64 combinations to prevent excessive HMS RPCs), calls `listPartitionNamesByValue()` for each combination, and merges results via Set union.

The partition fetching flow becomes:
```
EQ effective              → filtered fetch via listPartitionNamesByValue() (unchanged)
IN effective (≤64 combos) → multi-call filtered fetch + union (NEW)
else                      → full fetch via listPartitionNames() (unchanged)
```

This change only affects the HMS partition name fetching stage. The subsequent in-memory partition pruning via `ListPartitionPruner` remains unchanged and continues to handle all predicate types correctly.

## What type of PR is this:

- [x] Improvement

The optimization is transparent — queries that previously fell back to full partition fetch will now use filtered fetch when IN predicates are present, producing the same final result with significantly reduced planning time.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/OptExternalPartitionPruner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/OptExternalPartitionPruner.java
@@ -71,6 +71,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


stephen-shelby:
please resolve conflict and add some sql tests

PR #71589, opened 2026-04-13
[Enhancement] Support csv.enclose and csv.escape in INSERT INTO FILES CSV export

xiangguangyxg:
## Why I'm doing:                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                             
  When exporting data via `INSERT INTO FILES ... format = "csv"`, string/JSON columns                                                                                                                                                                                        
  containing commas produce malformed CSV — the comma inside the field value is                                                                                                                                                                                              
  indistinguishable from the column delimiter, causing column mismatch when
  downstream tools (Redshift, Excel, pandas, etc.) read the file.                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                             
  StarRocks already supports `csv.enclose`/`csv.escape` on the **LOAD** (read) path,                                                                                                                                                                                         
  but silently ignores them on the **UNLOAD** (export/write) path. User-provided                                                                                                                                                                                             
  properties are parsed but never serialized to BE, and the writer emits raw, unescaped bytes.                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                             
  Example of the current broken behavior:

  ```sql
  INSERT INTO FILES (
      "path" = "s3://bucket/csv/",
      "format" = "csv",
      "csv.column_separator" = ",",
      "csv.enclose" = '"',
      "csv.escape"  = '"'
  ) SELECT id, name FROM t;

  -- Row: (1, "hello,world")
  -- Current output (broken): 1,hello,world        <- parsed as 3 columns
  -- Expected output:         1,"hello,world"      <- comma preserved
  ```

  ## What I'm doing:

  End-to-end support for csv.enclose and csv.escape in the CSV unload path.

  ### Behavior

  - When csv.enclose is set, every non-NULL field value is wrapped with the
  enclose character.
  - Occurrences of the enclose character and the escape character itself within
  field content are prefixed with the escape character. Escaping the escape
  character is required for roundtrip correctness because the CSV reader's
  ESCAPE state (be/src/formats/csv/csv_reader.cpp:282-294) strips escape
  characters during parsing.
  - Special case: when escape == enclose, a single occurrence is emitted (RFC 4180 doubling).
  - NULL values are output as \N without enclosing (matches Redshift / Postgres COPY).
  - csv.enclose / csv.escape accept a single-byte ASCII character only; multi-character or multi-byte (non-ASCII) values are rejected with a semantic error (tightened from the load-path contract, intentionally scoped to this new unload surface).
  - When csv.enclose is not set, behavior is unchanged.

  StarRocks reader roundtrip caveat

  RFC 4180 output (escape == enclose == '"') cannot be re-imported with
  csv.escape='"' because the reader's START state checks escape before
  enclose (see csv_reader.cpp:217-234). When re-importing, set only
  csv.enclose='"' without csv.escape; the reader's ENCLOSE state natively
  handles doubled quotes. Documented in all three languages (en/zh/ja).

  ### Implementation

  - gensrc/thrift/Descriptors.thrift: add csv_enclose and csv_escape to
  TTableFunctionTable (fields 13/14, optional i8 to match existing
  enclose/escape fields in TBrokerScanRangeParams).
  - fe/.../TableFunctionTable.java: parse in parsePropertiesForUnload() with
  single-byte ASCII validation via getBytes(StandardCharsets.UTF_8);
  serialize in toTTableFunctionTable().
  - be/src/runtime/table_function_table_sink.cpp: propagate thrift fields into
  sink context options.
  - be/src/formats/csv/csv_file_writer.{h,cpp}:
    - Add enclose/escape to CSVWriterOptions and factory option parsing.
    - _write_enclosed_string() writes a string wrapped with the enclose char,
  escaping both enclose and escape occurrences (with escape == enclose
  handled once).
    - _write_enclosed_field() captures converter output into a
  FormattedOutputStreamString in-memory buffer, then calls the escape
  routine. This keeps enclose/escape logic out of every converter and covers
  all data types (string, JSON, numeric, etc.) uniformly.
    - write() calls ColumnHelper::unpack_and_duplicate_const_column() on
  every evaluated column before per-row iteration. This removes a latent
  crash path where ConstColumn inputs (e.g. from SELECT NULL ... or
  SELECT 'literal' ...) would hit down_cast<NullableColumn*> /
  down_cast<BinaryColumn*> assertions inside the per-type converters.
  The fix matches the Chunk::unpack_and_duplicate_const_columns() pattern
  already used elsewhere in the codebase.
    - write() uses the polymorphic Column::is_null(r) (not a down_cast) to
  branch between enclosed and \N emission.
    - _write_header() applies the same enclose/escape rules to header names
  when enabled.

  Note: the ConstColumn crash is a pre-existing issue in the CSV writer (not
  introduced by this PR), but is fixed here because the new enclose path made it
  easier to reproduce and because the fix is a 3-line addition.

  ### Tests

  - FE (TableFunctionTableTest.testCSVEncloseEscapeForUnload): covers
  default, RFC 4180, backslash-escape, only-enclose cases, plus empty /
  multi-character / multi-byte (Chinese) rejection for both enclose and escape.
  Tests run: 13, Failures: 0, Errors: 0.
  - BE (CSVFileWriterTest): new tests cover
    - TestEncloseEscapeRFC4180 — enclose==escape doubling
    - TestEncloseEscapeBackslash — enclose!=escape including trailing \,  mixed \"
    - TestEncloseWithoutEscape, TestEncloseWithHeader, TestEncloseDisabled
    - TestEncloseConstNullableColumn — regression for SELECT NULL crash
    - TestEncloseConstStringColumn — regression for SELECT 'literal' crash
    - TestConstColumnWithoutEnclose — pre-existing const-column crash on the  enclose-disabled path
    - TestFactoryWithEncloseEscapeOptions 9/9 PASSED.
  - SQL integration (test/sql/test_sink/T/test_csv_enclose_escape +
  matching R file): three OSS-based roundtrip cases following the
  test_csv_file_sink pattern (RFC 4180, backslash escape, include-header).

  ### Docs

  Updated docs/en, docs/zh, docs/ja for:

  - sql-reference/sql-functions/table-functions/files.md — unload_data_param
  table (added csv.column_separator, csv.row_delimiter,
  csv.include_header, csv.enclose, csv.escape) with explicit
  single-byte ASCII contract and re-import caveat.
  - loading/loading_introduction/feature-support-loading-and-unloading.md —
  unloading CSV feature table (added enclose, escape, include_header).
  - unloading/unload_using_insert_into_files.md — new CSV export example with
  csv.enclose/csv.escape and a note about the re-import caveat.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/formats/csv/csv_file_writer.cpp
+++ b/be/src/formats/csv/csv_file_writer.cpp
@@ -18,12 +18,14 @@
 #include <utility>
 
 #include "base/utility/defer_op.h"
+#include "column/column_helper.h"
 #include "common/http/content_type.h"
 #include "exec/hdfs_scanner/hdfs_scanner_text.h"
 #include "formats/column_evaluator.h"
 #include "formats/csv/csv_escape.h"
 #include "formats/utils.h"
 #include "io/formatted_output_stream_file.h"
+#include "io/formatted_output_stream_string.h"
 #include "runtime/current_thread.h"
 #include "util/compression/compression_utils.h"
 
@@ -83,11 +85,17 @@ Status CSVFileWriter::_write_header() {
             LOG(WARNING)
                     << "include_header is enabled but column_names is empty, this may indicate an upstream logic issue";
         } else {
+            const bool use_enclose = (_writer_options->enclose != 0);
             for (size_t i = 0; i < _column_names.size(); i++) {
-                // Escape column names that contain special characters (delimiter, quotes, newlines)
-                std::string escaped_name =
-                        csv::escape_csv_field(_column_names[i], _writer_options->column_terminated_by);
-                RETURN_IF_ERROR(_output_stream->write(escaped_name));
+                if (use_enclose) {
+                    // Enclose the column name with the same escape semantics as data fields
+                    RETURN_IF_ERROR(_write_enclosed_string(_column_names[i]));
+                } else {
+                    // Escape column names that contain special characters (delimiter, quotes, newlines)
+                    std::string escaped_name =
+                            csv::escape_csv_field(_column_names[i], _writer_options->column_terminated_by);
+                    RETURN_IF_ERROR(_output_stream->write(escaped_name));
+                }
                 if (i + 1 != _column_names.size()) {
                     RETURN_IF_ERROR(_output_stream->write(_writer_options->column_terminated_by));
                 }
@@ -125,6 +133,16 @@ Status CSVFileWriter::write(Chunk* chunk) {
         columns.push_back(std::move(column));
     }
 
+    // Expression evaluation can produce ConstColumn (e.g. `SELECT NULL AS c, ...`),
+    // whose `is_nullable()` may be true but whose dynamic type is not NullableColumn.
+    // Unpack those here so downstream converters can safely rely on concrete column
+    // types without every per-type `down_cast` having to handle ConstColumn.
+    for (auto& column : columns) {
+        column = ColumnHelper::unpack_and_duplicate_const_column(chunk->num_rows(), column);
+    }
+
+    const bool use_enclose = (_writer_options->enclose != 0);
+
     for (size_t r = 0; r < chunk->num_rows(); r++) {
         for (size_t c = 0; c < columns.size(); c++) {
             csv::Converter* converter;
@@ -133,7 +151,25 @@ Status CSVFileWriter::write(Chunk* chunk) {
             } else {
                 converter = _column_converters[c].second.get();
             }
-            RETURN_IF_ERROR(converter->write_string(_output_stream.get(), *columns[c], r, *_converter_options));
+
+            if (use_enclose) {
+                // Use the polymorphic Column::is_null() to decide on enclosure. After
+                // unpack_and_duplicate_const_column above, this safely handles both
+                // regular and previously-const nullable columns.
+                if (columns[c]->is_null(r)) {
+                    // NULL values go through the NullableConverter, which emits "\N"
+                    // without enclosing (matches Redshift / Postgres COPY behavior).
+                    RETURN_IF_ERROR(converter->write_string(_output_stream.get(), *columns[c], r,
+                                                            *_converter_options));
+                } else {
+                    // Non-NULL: wrap with enclose char and escape internal enclose/escape chars.
+                    RETURN_IF_ERROR(_write_enclosed_field(converter, *columns[c], r, *_converter_options));
+                }
+            } else {
+                RETURN_IF_ERROR(converter->write_string(_output_stream.get(), *columns[c], r,
+                                                        *_converter_options));
+            }
+
             if (c + 1 != columns.size()) {
                 RETURN_IF_ERROR(_output_stream->write(_writer_options->column_terminated_by));
             }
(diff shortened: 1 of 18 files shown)


CelerData-Reviewer:
@codex review

be/src/formats/csv/csv_file_writer.cpp
```diff
@@ -133,7 +143,25 @@ Status CSVFileWriter::write(Chunk* chunk) {
...
                 converter = _column_converters[c].second.get();
             }
-            RETURN_IF_ERROR(converter->write_string(_output_stream.get(), *columns[c], r, *_converter_options));
+
+            if (use_enclose) {
+                // Determine if this cell is NULL; NULL values are written as \N without enclosing.
+                bool is_null = false;
+                if (columns[c]->is_nullable()) {
+                    auto* nullable = down_cast<const NullableColumn*>(columns[c].get());
+                    is_null = (nullable->immutable_null_column_data()[r] != 0);
```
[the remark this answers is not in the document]
xiangguangyxg: fixed

be/src/formats/csv/csv_file_writer.cpp
```diff
@@ -83,11 +85,17 @@ Status CSVFileWriter::_write_header() {
...
         } else {
+            const bool use_enclose = (_writer_options->enclose != 0);
             for (size_t i = 0; i < _column_names.size(); i++) {
-                // Escape column names that contain special characters (delimiter, quotes, newlines)
-                std::string escaped_name =
-                        csv::escape_csv_field(_column_names[i], _writer_options->column_terminated_by);
-                RETURN_IF_ERROR(_output_stream->write(escaped_name));
+                if (use_enclose) {
+                    // Enclose the column name with the same escape semantics as data fields
+                    RETURN_IF_ERROR(_write_enclosed_string(_column_names[i]));
```
kevincai, line 92: will the column name still need to be escaped, e.g., the enclose is `"`, and the column name contains `"`.

be/src/formats/csv/csv_file_writer.cpp
```diff
@@ -133,7 +151,23 @@ Status CSVFileWriter::write(Chunk* chunk) {
...
                 converter = _column_converters[c].second.get();
             }
-            RETURN_IF_ERROR(converter->write_string(_output_stream.get(), *columns[c], r, *_converter_options));
+
+            if (use_enclose) {
+                // Use the polymorphic Column::is_null() to decide on enclosure. After
+                // unpack_and_duplicate_const_column above, this safely handles both
+                // regular and previously-const nullable columns.
+                if (columns[c]->is_null(r)) {
+                    // NULL values go through the NullableConverter, which emits "\N"
```
kevincai: are we confirmed this behavior?

be/src/formats/csv/csv_file_writer.cpp
xiangguangyxg: Yes, column names containing the enclose character are escaped by _write_enclosed_string(), which is shared by both header and data field paths — the same escape logic handles both.                                                                      
  This is covered by TestEncloseWithHeader in csv_file_writer_test.cpp:1467, which uses the column name my"col with enclose='"', escape='"' and asserts the header output is "my""col" (internal " correctly doubled).

kevincai (approved):

wyb (approved):

satanson (approved):

gengjun-git (approved):

PR #71596, opened 2026-04-13
[Enhancement] Support audit queried relations in audit log

before-Sunrise:
## What I'm doing:
record query's direct queried tables and views

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/docs/en/administration/management/audit_loader.md
+++ b/docs/en/administration/management/audit_loader.md
@@ -53,7 +53,7 @@ CREATE TABLE starrocks_audit_db__.starrocks_audit_tbl__ (
   `pendingTimeMs` BIGINT COMMENT "Time the query waited in the queue (milliseconds)",
   `candidateMVs` VARCHAR(65533) NULL COMMENT "List of candidate materialized views",
   `hitMvs` VARCHAR(65533) NULL COMMENT "List of matched materialized views",
-  `relations` ARRAY<VARCHAR(65533)> NULL COMMENT "List of directly referenced tables and views",
+  `QueriedRelations` ARRAY<VARCHAR(65533)> NULL COMMENT "List of directly referenced tables and views",
   `warehouse` VARCHAR(32) NULL COMMENT "Warehouse name"
 ) ENGINE = OLAP
 DUPLICATE KEY (`queryId`, `timestamp`, `queryType`)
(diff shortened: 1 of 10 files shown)


huailiu1122:
@mergify rebase

CelerData-Reviewer:
@codex review

huailiu1122:
@mergify rebase

HangyuanLiu:
I think the main issue in this implementation is not the result itself, but that the behavior is encoded in a very implicit way.

`showInternalCatalog` is introduced as mutable state on `TableName`, then set inside `AnalyzerUtils.TableAndViewRelationsCollector`, and finally only becomes observable much later through `TableName.toString()`. That makes the audit-log behavior hard to reason about because the setter and the actual effect are not explicit at the call site.

In other words, this is an audit-formatting concern, but it is now carried by a generic semantic object (`TableName`). That feels confusing and easy to misuse later.

Would it be cleaner to keep `TableName` unchanged and make the audit path explicit instead, for example with a dedicated formatter / helper that renders relation names with internal catalog when needed?

CelerData-Reviewer:
@codex review

before-Sunrise:
> I think the main issue in this implementation is not the result itself, but that the behavior is encoded in a very implicit way.
> 
> `showInternalCatalog` is introduced as mutable state on `TableName`, then set inside `AnalyzerUtils.TableAndViewRelationsCollector`, and finally only becomes observable much later through `TableName.toString()`. That makes the audit-log behavior hard to reason about because the setter and the actual effect are not explicit at the call site.
> 
> In other words, this is an audit-formatting concern, but it is now carried by a generic semantic object (`TableName`). That feels confusing and easy to misuse later.
> 
> Would it be cleaner to keep `TableName` unchanged and make the audit path explicit instead, for example with a dedicated formatter / helper that renders relation names with internal catalog when needed?

use formatTableNameForAudit to do this, and keep TableAndViewRelationsCollector unchanged

HangyuanLiu (approved):

stephen-shelby (approved):

before-Sunrise:
https://github.com/Mergifyio backport branch-4.0

PR #71614, opened 2026-04-13
[BugFix] Avoid holding DB read lock during shard-group id collection

kevincai:
StarMgrMetaSyncer.getAllPartitionShardGroupId held the database READ lock for the full iteration of every cloud-native table and physical partition in each database. On large catalogs this stalls any FE thread waiting on the DB INTENTION_SHARE/WRITE lock.

Snapshot the cloud-native tables without any DB lock - getTablesIncludeRecycleBin is backed by a ConcurrentHashMap view plus a synchronized recycle-bin accessor and returns a fresh list - then wrap each per-table traversal with lockTableWithIntensiveDbLock (DB INTENTION_SHARED + table READ). The collected set is already treated as best-effort by deleteUnusedShardAndShardGroup.

Also emit a LOG.warn when the whole collection exceeds 30s so future regressions are observable.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/StarMgrMetaSyncer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/StarMgrMetaSyncer.java
@@ -78,6 +78,9 @@ public class StarMgrMetaSyncer extends FrontendDaemon {
             "starmgr_meta_sync_process_time_total", Metric.MetricUnit.SECONDS,
             "The total number of seconds spent on meta sync by StarMgrMetaSyncer");
 
+    // Log a warning when getAllPartitionShardGroupId takes longer than this.
+    private static final long SLOW_COLLECTION_WARN_THRESHOLD_MS = 30_000L;
+
     // make sure the metrics are registered only once
     private static final AtomicBoolean IS_METRIC_REGISTERED = new AtomicBoolean(false);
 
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

kevincai:
@codex review

alvin-celerdata (approved):

fe/fe-core/src/main/java/com/starrocks/lake/StarMgrMetaSyncer.java
```diff
@@ -99,32 +102,55 @@ public synchronized void start() {
...
-                    if (table.isCloudNativeTableOrMaterializedView()) {
-                        GlobalStateMgr.getCurrentState().getLocalMetastore()
-                                .getAllPartitionsIncludeRecycleBin((OlapTable) table)
-                                .stream()
-                                .map(Partition::getSubPartitions)
-                                .flatMap(p -> p.stream().map(PhysicalPartition::getShardGroupIds))
-                                .forEach(groupIds::addAll);
-                    }
-                }
+                tables = GlobalStateMgr.getCurrentState().getLocalMetastore().getTablesIncludeRecycleBin(db);
```
xiangguangyxg, line 124: `getTablesIncludeRecycleBin` this may not need to lock db because the `idToTable` member in the database class is a concurrent map.

starrocks-xupeng (approved):

fe/fe-core/src/main/java/com/starrocks/lake/StarMgrMetaSyncer.java
kevincai: It contains two steps in the `getTablesIncludeRecycleBin`, and the two ops are not atomic, that's why the db IS lock is necessary.

xiangguangyxg (approved):

PR #71622, opened 2026-04-13
[BugFix] Calculate statistics in SplitWindowSkewToUnionRule

martinr0x:
## Why I'm doing:
When `OptExpressionDuplicator` duplicates a plan subtree (used by `SplitWindowSkewToUnionRule`), Statistics are not propagated to the duplicated OptExpression nodes. This causes downstream instances of the same rule to operate on duplicated branches that have no statistics, leading to incorrect plan choices.
## What I'm doing:
Recalculate statistics for the duplicated branch on each invocation of `SplitWindowSkewToUnionRule`. 

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/operator/OpRuleBit.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/operator/OpRuleBit.java
@@ -34,4 +34,6 @@ public class OpRuleBit {
     public static final int OP_MV_AGG_PRUNE_COLUMNS = 4;
     // Operator has been push down topn below agg or not, if push down topn, no need to push down again.
     public static final int OP_PUSH_DOWN_TOPN_AGG = 5;
+    // Window operator has been split by the window skew rewrite, no need to split again.
+    public static final int OP_SPLIT_WINDOW_SKEW = 6;
 }
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitWindowSkewToUnionRule.java
@@ -20,6 +20,7 @@ import com.starrocks.sql.analyzer.SemanticException;
 import com.starrocks.sql.ast.expression.BinaryType;
 import com.starrocks.sql.optimizer.OptExpression;
 import com.starrocks.sql.optimizer.OptimizerContext;
+import com.starrocks.sql.optimizer.Utils;
 import com.starrocks.sql.optimizer.base.ColumnRefFactory;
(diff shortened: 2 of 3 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/materialization/OptExpressionDuplicator.java
```diff
@@ -164,29 +179,86 @@ class OptExpressionDuplicatorVisitor extends OptExpressionVisitor<OptExpression,
...
+                OptExpression duplicated = child.getOp().accept(this, child, null);
+                if (copyStatistics) {
+                    copyStatisticsIfPresent(child, duplicated);
+                }
+                inputs.add(duplicated);
             }
             return inputs;
         }
 
+        private void copyStatisticsIfPresent(OptExpression source, OptExpression target) {
```
stephen-shelby: I suggest not copying statistics inside the duplicator. Duplicating the plan structure and maintaining statistics are separate responsibilities, and directly copying stats can easily introduce consistency issues around column remapping, multi-column stats, and similar metadata.

A cleaner approach is: keep the duplicator responsible only for copying the expression tree; after SplitWindowSkewToUnionRule builds the new UNION subtree, recalculate statistics for that subtree. Meanwhile, mark the newly created branch windows with an OpRuleBit so RewriteTreeTask will not apply the same rule again to the windows that were just split, avoiding recursive splitting.

This bit is scoped to the current window node only. It does not stop traversal into the node’s children, so nested windows can still be rewritten normally.

martinr0x: @stephen-shelby thanks for the review. I was thinking about this approach as well but it seemed more verbose. But i agree, its the cleaner choice. 
I will make the changes.

martinr0x: done: [319b880](https://github.com/StarRocks/starrocks/pull/71622/commits/319b880da5a1edc646154526c673e62d98e32e13)

stephen-shelby (approved):

trueeyu (approved):

kevincai (approved):

martinr0x:
@mergifyio backport branch-3.5-cc

Issue #71579, opened 2026-04-13
CN POD restart frequently (shared-data)


guangtao:
### Expected behavior (Required)
CN pod restart every day

### Real behavior (Required)
cn.out log:

```
3.3.22 RELEASE (build 753696f distro ubuntu arch x86_64)
query_id:00000000-0000-0000-0000-000000000000, fragment_instance:00000000-0000-0000-0000-000000000000
tracker:process consumption: 16957347272
tracker:jemalloc_metadata consumption: 1459934592
tracker:query_pool consumption: 0
tracker:query_pool/connector_scan consumption: 0
tracker:load consumption: 0
tracker:metadata consumption: 1366883016
tracker:tablet_metadata consumption: 589220
tracker:rowset_metadata consumption: 0
tracker:segment_metadata consumption: 143096362
tracker:column_metadata consumption: 1223197434
tracker:tablet_schema consumption: 589220
tracker:segment_zonemap consumption: 121870149
tracker:short_key_index consumption: 1864073
tracker:column_zonemap_index consumption: 228540570
tracker:ordinal_index consumption: 236064640
tracker:bitmap_index consumption: 0
tracker:bloom_filter_index consumption: 0
tracker:compaction consumption: 0
tracker:schema_change consumption: 0
tracker:column_pool consumption: 0
tracker:page_cache consumption: 11467068800
tracker:jit_cache consumption: 0
tracker:update consumption: 215995992
tracker:chunk_allocator consumption: 0
tracker:passthrough consumption: 0
tracker:clone consumption: 0
tracker:consistency consumption: 0
tracker:datacache consumption: 354912931
tracker:replication consumption: 0
*** Aborted at 1775954870 (unix time) try "date -d @1775954870" if you are using GNU date ***
PC: @          0xa7bc7fd google::protobuf::Reflection::SpaceUsedLong(google::protobuf::Message const&) const
*** SIGSEGV (@0x88) received by PID 27 (TID 0x7f40055ff640) from PID 136; stack trace: ***
    @     0x7f64387f5ee8 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x99ee7)
    @          0xa2ec729 google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @     0x7f643879e520 (/usr/lib/x86_64-linux-gnu/libc.so.6+0x4251f)
[...]

PR #71630, opened 2026-04-13
[BugFix] Cache SpaceUsedLong() to prevent SIGSEGV in ColumnReader destructor

tracymacding:
## Summary
- Cache protobuf `SpaceUsedLong()` values at `ColumnReader::_init()` time
- Use cached values in `~ColumnReader()` instead of calling the virtual method
- Prevents SIGSEGV when protobuf heap memory is corrupted by an external bug

## Background
Multiple customers (Demandbase 3.5.15, Sands 3.3.11, #71283 4.0.8, #71579 3.3.22) report SIGSEGV in `ColumnReader::~ColumnReader()` at `SpaceUsedLong()`. The crash occurs because the protobuf object's vtable pointer is zeroed (freed/corrupted heap memory), causing dereference at offsets 0x88/0x98.

This is a **defensive fix** — the root cause (heap memory corruption) still needs ASAN investigation. But this eliminates the crash in the destructor path.

## Test plan
- [ ] Verify memory tracking values are consistent (cached == runtime)
- [ ] Stress test with concurrent PK writes + table drops + garbage sweep
- [ ] Run ASAN build to verify no new issues introduced

Fixes: POST-1334
Related: #71283, #71579

### DIFF
--- a/be/src/storage/rowset/column_reader.cpp
+++ b/be/src/storage/rowset/column_reader.cpp
@@ -90,29 +90,33 @@ ColumnReader::ColumnReader(const private_type&, Segment* segment) : _segment(seg
 }
 
 ColumnReader::~ColumnReader() {
+    // Use cached SpaceUsedLong() values instead of calling the virtual method on the
+    // protobuf objects. If heap memory is corrupted by an external bug, the protobuf
+    // vtable may be zeroed/invalid, causing SIGSEGV in SpaceUsedLong().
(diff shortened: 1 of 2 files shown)


guangtao:
Hi @tracymacding, thanks for this fix! We're hitting the same crash on 3.3.22 
(see #71579). Will this be backported to branch-3.3?

PR #71642, opened 2026-04-14
[Enhancement] Expose MV lastRefreshTime in information_schema.materialized_views and SHOW MATERIALIZED VIEWS

neerav-sharma:
## Why I'm doing:

The internal lastRefreshTime from MvRefreshScheme is the timestamp used to determine staleness of the MV compared to the base table, and is the value used in the staleness formula for mv_rewrite_staleness_second. This value is not currently exposed externally.

The existing LAST_REFRESH_FINISHED_TIME comes from TaskRunStatus and represents when the refresh job finished executing -- an operational metric, not a data-freshness metric.

## What I'm doing:

Added a new LAST_REFRESH_TIME column to information_schema.materialized_views and SHOW MATERIALIZED VIEWS that reads from MvRefreshScheme.getLastRefreshTime() on the MaterializedView object. This enables external monitoring and alerting on MV data freshness using the same value that the query optimizer uses internally for staleness decisions.

Note: the column name LAST_REFRESH_TIME was chosen as a direct mapping to the internal variable name. Open to renaming if a more descriptive name is preferred (e.g. LAST_REFRESH_DATA_TIME).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

### DIFF
--- a/be/src/exec/schema_scanner/schema_materialized_views_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_materialized_views_scanner.cpp
@@ -53,6 +53,7 @@ SchemaScanner::ColumnDesc SchemaMaterializedViewsScanner::_s_tbls_columns[] = {
         {"CREATOR", TypeDescriptor::create_varchar_type(sizeof(Slice)), sizeof(Slice), true},
         {"LAST_REFRESH_PROCESS_TIME", TypeDescriptor::from_logical_type(TYPE_DATETIME), sizeof(DateTimeValue), true},
         {"LAST_REFRESH_JOB_ID", TypeDescriptor::create_varchar_type(sizeof(Slice)), sizeof(Slice), true},
+        {"LAST_REFRESH_TIME", TypeDescriptor::from_logical_type(TYPE_DATETIME), sizeof(DateTimeValue), true},
 };
 
 SchemaMaterializedViewsScanner::SchemaMaterializedViewsScanner()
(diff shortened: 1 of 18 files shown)


HangyuanLiu (changes requested):
Thanks for putting this together. I think this needs another round before it can be merged.

There are a few integration gaps around the new `LAST_REFRESH_TIME` column:

1. `SHOW MATERIALIZED VIEWS` metadata is still the old 27-column schema. `ShowMaterializedViewStatus.toResultSet()` now appends the new value, but `ShowResultMetaFactory.visitShowMaterializedViewStatement()` still advertises 27 columns. That means the plain SHOW path can send 28 row values under a 27-column header.

2. `SHOW MATERIALIZED VIEWS ... WHERE ...` still uses the old projection list. This path is rewritten through `ShowStmtToSelectStmtConverter.toSelectStmt()`, and `MATERIALIZED_VIEW_META_COLUMNS` still stops at `last_refresh_job_id`, so the filtered SHOW path will silently omit `last_refresh_time`.

3. `information_schema.materialized_views` also needs the BE schema-scan path to be updated. This table is backed by FE thrift + BE `SchemaMaterializedViewsScanner`, so adding a new field in `TMaterializedViewStatus` is not enough by itself. The scanner column list / slot mapping and the related BE tests need to be updated as well.

Please fix the SHOW metadata/projection paths and the BE scanner side together, then add the corresponding tests for both plain/filtered SHOW and `information_schema.materialized_views`.

CelerData-Reviewer:
@codex review

HangyuanLiu (approved):

stephen-shelby (approved):

satanson (approved):

PR #71651, opened 2026-04-14
[BugFix] Fix disk cache overflow when Iceberg metadata entries are pinned

osscm:
The Caffeine weigher in TwoLevelContentCache assigned weight=0 to pinned DiskCacheEntry objects (useCount > 0). This caused Caffeine to believe those bytes were free, allowing new entries to fill the freed space. Once the reader unpinned, Caffeine re-weighed the entry back to its real size, but the extra files were already on disk -- causing physical disk usage to exceed DISK_CACHE_CAPACITY by the size of all concurrently-pinned files.

Fix:
- Weigher now always uses the actual file size regardless of pin state, so Caffeine capacity accounting stays accurate under concurrent reads.
- Eviction listener skips file deletion when useCount > 0 (logs a warning); the orphaned file is cleaned up on next FE restart via loadMetadataDiskCache.
- useCount changed from int to AtomicInteger for thread safety.
- Added debug logging on pin/unpin with cache stats.
- Added two unit tests: one reproducing the bug with the old weigher, one verifying the fix maintains capacity invariant under pinning.

## Why I'm doing:

  TwoLevelContentCache uses a Caffeine weigher that assigns weight=0 to any DiskCacheEntry with useCount > 0 (i.e., currently being read). This causes Caffeine to treat pinned entries as
  "free" capacity. While a file is being read:

  1. Caffeine believes the cache has more room than it actually does.
  2. New files are admitted to fill the phantom free space.
  3. After the reader closes and the entry is unpinned, Caffeine re-weighs it back to its real size — but the extra files are already on disk.
  
  Result: physical disk usage = DISK_CACHE_CAPACITY + Σ(sizes of all concurrently-pinned files). Under sustained concurrent reads, disk usage grows unboundedly.

## What I'm doing:

  - Weigher: always use actual file size regardless of useCount. Caffeine's capacity accounting is now accurate even while entries are pinned.
  - Eviction listener: if an entry is evicted while still pinned (useCount > 0), skip the file deletion and log a warning. The orphaned file will be picked up and evicted on the next FE
  restart via loadMetadataDiskCache.
  - Thread safety: useCount changed from int to AtomicInteger — pin()/unpin() are called from Caffeine's computeIfPresent which can run on different threads.
  - Observability: added DEBUG logging on every pin/unpin with current useCount, file size, and Caffeine stats.

Tests

  - testDiskCacheWeigherAllowsOverflowWhenPinned — reproduces the bug using the old weigher; asserts that actual bytes exceed capacity.
  - testFixedWeigherRespectsCapacity — verifies that with the fixed weigher, bytes in cache never exceed DISK_CACHE_CAPACITY even when entries are pinned.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4


dirtysalt (approved):

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
```diff
@@ -437,11 +437,20 @@ private TwoLevelContentCache() {
...
                     // use sync evictionListener to avoid delete file newly generated by another thread
                     .evictionListener((key, value, cause) -> {
-                        LOG.debug("{} to be eliminated from disk, reason: {}", key, cause);
+                        if (value.useCount.get() > 0) {
+                            // Entry evicted by Caffeine while still in use. Skip delete—file remains as a temporary orphan.
+                            // It will be reloaded and properly evicted on next FE restart.
+                            LOG.warn("diskCache eviction skipped for pinned entry: key={}, useCount={}, " +
+                                            "size={}MB, cause={}", key, value.useCount.get(),
+                                    value.length >> 20, cause);
+                            return;
```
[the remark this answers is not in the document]
dirtysalt: The suggestion is technically correct. When Caffeine evicts a pinned entry (useCount > 0), the entry is already removed from `diskCache` before the eviction listener fires — the listener cannot prevent removal. As a result, the subsequent `computeIfPresent` in `DiskCacheSeekableInputStream.close()` finds no key and `unpin()` is never called, leaving an orphaned file that accumulates disk space until FE restart.

There is also a subtle regression introduced by the weigher fix itself: with the old weigher (weight=0 for pinned entries), Caffeine had no incentive to evict them. With the fixed weigher (weight=actual size), Caffeine now sees pinned entries as occupying real capacity and may actively evict them under pressure — making this orphan scenario more likely than before.

The correct fix is to handle deferred deletion in `DiskCacheSeekableInputStream.close()`:

```java
@Override
public void close() throws IOException {
    try {
        stream.close();
    } finally {
        boolean stillInCache = diskCache.asMap().computeIfPresent(key, (k, v) -> {
            v.unpin();
            return v;
        }) != null;

        if (!stillInCache) {
            // Entry was evicted while pinned; delete the orphaned file now
            IOUtil.deleteLocalFileWithRemotePath(key);
        }
    }
}
```

The eviction listener can keep the warning log but no longer needs special handling since cleanup is guaranteed on `close()`.

dirtysalt (changes requested):

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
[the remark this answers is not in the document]
osscm: @dirtysalt this make sense, please check, I have updated PR based on your suggestion.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
```diff
@@ -579,24 +579,31 @@ public static class DiskCacheSeekableInputStream extends SeekableInputStream {
...
-                diskCache.asMap().computeIfPresent(key, (k, v) -> {
-                    v.unpin();
-                    LOG.debug("diskCache unpin: key={}, useCount={}, size={}MB, stats=[{}]",
-                            k, v.useCount.get(), v.length >> 20, diskCache.stats());
-                    return v;
-                });
+                int remaining = entry.useCount.decrementAndGet();
+                LOG.debug("diskCache unpin: key={}, useCount={}, size={}MB, stats=[{}]",
+                        key, remaining, entry.length >> 20, diskCache.stats());
+                if (remaining == 0 && diskCache.asMap().get(key) != entry) {
```
dirtysalt: @osscm I don't quite get the condition. It's bit complicated.

> diskCache.asMap().get(key) != entry

If the value of key is not the `entry`, then this new value is probbaly being used and we should not remove it at all. 

if the value of the key is the entry, then ref count should not be zero,  so it probably will be evicted later and hanlded by the listener.  So we don't need to care about.

I think maybe the right way to handle is like the following

```
if (remaining == 0) {
    // Only delete if *our specific entry* is no longer in the cache.
    // diskCache.asMap().get(key) == null  → evicted, no replacement → safe to delete
    // diskCache.asMap().get(key) == entry → still live → skip, Caffeine handles it
    // diskCache.asMap().get(key) is a NEW object → evicted + replaced → DON'T delete new file
    DiskCacheEntry current = diskCache.asMap().get(key);
    if (current == null) {
        IOUtil.deleteLocalFileWithRemotePath(key);
        LOG.warn("diskCache cleaning up orphaned file for evicted-while-pinned entry: key={}", key);
    }
    // if current != null && current != entry: new entry loaded, skip deletion
}
```

dirtysalt (changes requested):

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
osscm: @dirtysalt updated

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
```diff
@@ -540,12 +549,14 @@ public boolean isExistOnDiskCache(String key) {
...
+                        k, v.useCount.get(), v.length >> 20, diskCache.stats());
                 return v;
             });
             if (diskCacheEntry != null) {
                 try {
                     SeekableInputStream stream = diskCacheEntry.toSeekableInputStream();
-                    return new DiskCacheSeekableInputStream(stream, diskCache, key);
+                    return new DiskCacheSeekableInputStream(stream, diskCache, key, diskCacheEntry);
                 } catch (Exception e) {
                     diskCache.asMap().computeIfPresent(key, (k, v) -> {
```
stephen-shelby: nit: Could we factor the unpin + orphan cleanup logic in `DiskCacheSeekableInputStream.close()` into a small helper, and call it here as well?

This catch path currently releases by looking up the key in `diskCache` via `computeIfPresent`. Since `diskCacheEntry` has already been captured and pinned, releasing the captured entry directly would be more consistent with `close()`. It also covers the edge case where the entry is evicted after `pin()` but before `toSeekableInputStream()` throws, in which case `computeIfPresent` would not run.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
```diff
@@ -568,22 +579,36 @@ public static class DiskCacheSeekableInputStream extends SeekableInputStream {
...
         @Override
         public void close() throws IOException {
             try {
                 stream.close();
             } finally {
-                diskCache.asMap().computeIfPresent(key, (k, v) -> {
-                    v.unpin();
-                    return v;
-                });
+                int remaining = entry.useCount.decrementAndGet();
```
Wenjun7J: let pin and unpin return useCount, and call unpin here.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
```diff
@@ -540,12 +549,14 @@ public boolean isExistOnDiskCache(String key) {
...
                 return v;
             });
             if (diskCacheEntry != null) {
                 try {
                     SeekableInputStream stream = diskCacheEntry.toSeekableInputStream();
-                    return new DiskCacheSeekableInputStream(stream, diskCache, key);
+                    return new DiskCacheSeekableInputStream(stream, diskCache, key, diskCacheEntry);
                 } catch (Exception e) {
                     diskCache.asMap().computeIfPresent(key, (k, v) -> {
                         v.unpin();
```
Wenjun7J: just call `diskCacheEntry.unpin()` here. between the computeIfpresent calls, the key value object may be changed.

Wenjun7J (changes requested):

osscm:
@stephen-shelby have added the helper method...

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
osscm: updated @Wenjun7J plz check

dirtysalt (approved):

stephen-shelby (approved):

osscm:
Thanks @dirtysalt @stephen-shelby @Wenjun7J for the guidance and help!

Wenjun7J (approved):

CelerData-Reviewer:
@codex review

kangkaisen (approved):

HangyuanLiu (approved):

PR #71660, opened 2026-04-14
[BugFix] Paimon primary key columns incorrectly marked as non-nullable when querying external catalog

### LINKED ISSUE #71657: [BUG] Paimon primary key columns incorrectly marked as non-nullable when querying external catalog
Primary key columns in Paimon tables cannot be queried through external catalogs because StarRocks marks them as non-nullable, which violates the external table constraint that all columns must be nullable.

Image

mysql> select * from customer_pk1;
ERROR 1064 (HY000): All columns must be nullable for external table. Column 'c_custkey' is not nullable, You can rebuild theexternal table and We strongly recommend that you use catalog to access external data.: BE:10001

This is a bug in paimon connector. 
In Paimon, primary key columns are NOT nullable by default at the DataType level. When you create a table with a primary key, Paimon marks those columns as non-nullable in their DataType definition, even if you don't explicitly set NOT NULL.
starRocks relies on type.isNullable() to set the Column's nullable flag, but for Paimon primary keys, this always returns false.
The fix should be in ColumnTypeConverter.fromPaimonSchemas() 
public static List<Column> fromPaimonSchemas(List<DataField> fields) {
        ....
        //Force all columns to be nullable for external Paimon tables
        //Column column = new Column(fieldName, fieldType, type.isNullable(), field.description());
        Column column = new Column(fieldName, fieldType, true, field.description());
        ....
}

paimon pk table query should be success

### Real behavior (Required)
query fails with error

mysql> select current_version();
+-------------------+
| current_version() |
+-------------------+
| main-4520c04      |
+-------------------+
1 row in set (0.03 sec)

rohitrs1983:
## What I'm doing:
In Paimon, primary key columns are NOT nullable by default at the DataType level. When you create a table with a primary key, Paimon marks those columns as non-nullable in their DataType definition, even if you don't explicitly set NOT NULL.
starRocks relies on type.isNullable() to set the Column's nullable flag, but for Paimon primary keys, this always returns false. the fix removes the type.isNullable(), and sets the parameter to true.

Fixes #71657
image

## What type of PR is this:

- [X] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/ColumnTypeConverter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/ColumnTypeConverter.java
@@ -600,7 +600,7 @@ public class ColumnTypeConverter {
             String fieldName = field.name();
             org.apache.paimon.types.DataType type = field.type();
             Type fieldType = ColumnTypeConverter.fromPaimonType(type);
-            Column column = new Column(fieldName, fieldType, type.isNullable(), field.description());
+            Column column = new Column(fieldName, fieldType, true, field.description());
             columns.add(column);
         }
         return columns;
--- a/fe/fe-core/src/main/java/com/starrocks/connector/ColumnTypeConverter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/ColumnTypeConverter.java
@@ -600,6 +600,7 @@ public class ColumnTypeConverter {
             String fieldName = field.name();
             org.apache.paimon.types.DataType type = field.type();
             Type fieldType = ColumnTypeConverter.fromPaimonType(type);
+            // Force all Paimon columns to be nullable (true) regardless of their DataType's nullable property.
             Column column = new Column(fieldName, fieldType, true, field.description());
             columns.add(column);
         }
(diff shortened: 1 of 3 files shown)


fe/fe-core/src/main/java/com/starrocks/connector/ColumnTypeConverter.java
```diff
@@ -600,7 +600,7 @@ public static List<Column> fromPaimonSchemas(List<DataField> fields) {
             String fieldName = field.name();
             org.apache.paimon.types.DataType type = field.type();
             Type fieldType = ColumnTypeConverter.fromPaimonType(type);
-            Column column = new Column(fieldName, fieldType, type.isNullable(), field.description());
+            Column column = new Column(fieldName, fieldType, true, field.description());
```
HangyuanLiu, line 604: Please add comment describe why flag is always true

CelerData-Reviewer:
@codex review

HangyuanLiu (approved):

miomiocat (approved):

stephen-shelby (approved):

miomiocat:
@mergify backport branch-4.1

miomiocat:
@mergify backport branch-4.0

miomiocat:
@mergify backport branch-3.5

PR #71688, opened 2026-04-14
[Doc] Fix Cache Observability description in 4.1 release notes

celerdata2308:
## Summary
- Removed per-table hit rates from Cache Observability description in 4.1 release notes (en/zh/ja)
- Table/partition/tablet-level cache metrics are enterprise-only; community edition only exposes query-level and global-level hit ratios
- Updated wording to clarify "query-level" cache hit ratio and "latency diagnosis" (instead of "latency predictability")

## Test plan
- [x] Verify en/zh/ja release notes have consistent descriptions
- [ ] Visual check on docs site after merge


DanRoscigno (approved):
Thanks!

EsoragotoSpirit (approved):

PR #71742, opened 2026-04-15
[BugFix] Fix SIGFPE in lake compaction when rowset num_rows is zero

xiangguangyxg:
## Why I'm doing:

  In shared-data mode, after an online range-distribution tablet split, a sub-tablet can inherit rowsets whose metadata `num_rows` was rewritten to `0`, while the shared physical segments still contain rows that fall inside the sub-tablet's range. When compaction subsequently runs on such a sub-tablet, both compaction paths use `total_num_rows` as a divisor when computing progress:

  - `horizontal_compaction_task.cpp:134`: `100 * reader.stats().raw_rows_read / total_num_rows`
  - `vertical_compaction_task.cpp:272-273`: `(100 * column_group_index + 100 * reader.stats().raw_rows_read / _total_num_rows) / column_group_size`

  When the denominator is `0`, this triggers a BE SIGFPE (integer divide-by-zero) and crashes the process.

  ## What I'm doing:

  A minimal, backport-safe hotfix that guards the two progress-update divisions. When the denominator is zero, we skip the incremental update and preserve the previous progress value. The final unconditional `_context->progress.update(100)` at `execute()` exit still guarantees the terminal state, so observers relying on `progress.value() == 100` at completion remain unaffected.

  The root-cause fix — eliminating the splitter-vs-reader overlap disagreement that produces the inconsistent metadata — will be addressed in a separate follow-up PR against `main` only. This PR ships L1 first as a narrow hotfix to unblock affected users on release branches.

  Changes:
  - `be/src/storage/lake/horizontal_compaction_task.cpp`: guard the divide with `if (total_num_rows > 0)`.
  - `be/src/storage/lake/vertical_compaction_task.cpp`: guard the divide with `if (_total_num_rows > 0 && column_group_size > 0)`.
  - `be/test/storage/lake/compaction_task_test.cpp`: add `test_zero_num_rows_no_crash` (parameterized across HORIZONTAL / VERTICAL × size-tiered on/off — 4 variants) that reproduces the exact post-split scenario: write data, zero out per-rowset `num_rows` / `data_size` in metadata, then run compaction. Before the fix this crashes; after the fix it completes and reaches `progress.value() == 100`.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/horizontal_compaction_task.cpp
+++ b/be/src/storage/lake/horizontal_compaction_task.cpp
@@ -131,7 +131,9 @@ Status HorizontalCompactionTask::execute(CancelFunc cancel_func, ThreadPool* flu
         chunk->reset();
         rssid_rowids.clear();
 
-        _context->progress.update(100 * reader.stats().raw_rows_read / total_num_rows);
+        if (total_num_rows > 0) {
+            _context->progress.update(100 * reader.stats().raw_rows_read / total_num_rows);
+        }
         _context->stats->collect(reader.stats());
     }
 
--- a/be/src/storage/lake/vertical_compaction_task.cpp
+++ b/be/src/storage/lake/vertical_compaction_task.cpp
@@ -269,8 +269,11 @@ Status VerticalCompactionTask::compact_column_group(bool is_key, int column_grou
             source_masks->clear();
         }
 
-        _context->progress.update((100 * column_group_index + 100 * reader.stats().raw_rows_read / _total_num_rows) /
-                                  column_group_size);
+        if (_total_num_rows > 0 && column_group_size > 0) {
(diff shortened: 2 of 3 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #71809, opened 2026-04-16
[BugFix] Set transaction compute resource to session warehouse in explicit transaction

xiangguangyxg:
## Why I'm doing:
                                               
  In shared-data mode, when a DML inside an explicit transaction (BEGIN/COMMIT) uses a
  different warehouse than the one captured at BEGIN time — via `SET_VAR(warehouse=...)` hint
  or `SET warehouse` — the publish version RPCs are still routed to the BEGIN-time warehouse.

  This happens because `txnState.computeResource` is set once in
  `TransactionStmtExecutor.beginStmt()` and never updated. Although the Coordinator correctly
  snapshots the session's current warehouse for DML execution (via `JobSpec.fromQuerySpec()`),
  the publish path in `PublishVersionDaemon` reads the stale `txnState.computeResource`,
  causing execution and publish to diverge to different warehouses.

## What I'm doing:

  Sync `txnState.computeResource` to the session's effective warehouse in both `loadData()`
  overloads before Coordinator creation. This ensures publish version RPCs are routed to the
  same warehouse that executes the DML.

  - For `SET_VAR` DMLs, `processQueryScopeHint()` has already applied the hint to the session
    before `loadData()` is called.
  - For non-hint DMLs, the session carries the user's current warehouse.
  - Different DMLs within the same transaction may use different warehouses — the transaction's
    compute resource reflects the last DML's warehouse (last-writer-wins).

  This is safe in shared-data mode because data resides in shared storage and publish can be
  handled by any active warehouse's CN.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
@@ -173,12 +173,12 @@ public class TransactionStmtExecutor {
                                 DmlStmt dmlStmt,
                                 OriginStatement originStmt,
                                 ConnectContext context) {
-        Coordinator coordinator = new DefaultCoordinator.Factory().createInsertScheduler(
-                context, execPlan.getFragments(), execPlan.getScanNodes(), execPlan.getDescTbl().toThrift(), execPlan);
-
         GlobalTransactionMgr globalTransactionMgr = GlobalStateMgr.getCurrentState().getGlobalTransactionMgr();
         ExplicitTxnState explicitTxnState = globalTransactionMgr.getExplicitTxnState(context.getTxnId());
         TransactionState transactionState = explicitTxnState.getTransactionState();
+
+        Coordinator coordinator = new DefaultCoordinator.Factory().createInsertScheduler(
+                context, execPlan.getFragments(), execPlan.getScanNodes(), execPlan.getDescTbl().toThrift(), execPlan);
 
         try {
             if (transactionState.getDbId() == 0) {
@@ -221,6 +221,15 @@ public class TransactionStmtExecutor {
                     load(database, targetTable, execPlan, dmlStmt, originStmt, context, coordinator);
             explicitTxnState.addTransactionItem(item);
 
+            // Sync the transaction's compute resource to the warehouse that actually
+            // executed this DML, so publish version RPCs are routed to the same warehouse.
+            // Use NoAcquire because Coordinator creation above has already populated the
+            // cache via JobSpec.fromQuerySpec() — this returns exactly the resource the
+            // DML ran on, without side effects.
+            // Placed AFTER addTransactionItem so a failed DML does not overwrite the
+            // publish target set by an earlier successful DML.
+            transactionState.setComputeResource(context.getCurrentComputeResourceNoAcquire());
+
             context.getState().setOk(item.getLoadedRows(), Ints.saturatedCast(item.getFilteredRows()),
                     buildMessage(transactionState.getLabel(), TransactionStatus.PREPARE,
                             transactionState.getTransactionId(), database.getId()));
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
```diff
@@ -173,13 +173,19 @@ public static void loadData(Database database,
...
-
         GlobalTransactionMgr globalTransactionMgr = GlobalStateMgr.getCurrentState().getGlobalTransactionMgr();
         ExplicitTxnState explicitTxnState = globalTransactionMgr.getExplicitTxnState(context.getTxnId());
         TransactionState transactionState = explicitTxnState.getTransactionState();
 
+        // Sync the transaction's compute resource to the session's effective warehouse so that
+        // publish version RPCs are routed to the same warehouse that executes the DML.
+        // For SET_VAR DMLs, processQueryScopeHint() has already applied the hint to the session.
+        // For non-hint DMLs, the session carries the user's current warehouse.
+        transactionState.setComputeResource(context.getCurrentComputeResource());
```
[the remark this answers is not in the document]
xiangguangyxg: done

fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
```diff
@@ -249,6 +256,9 @@ public static void loadData(long dbId, long tableId, ExplicitTxnState.ExplicitTx
 
         explicitTxnState.addTransactionItem(item);
 
+        // Sync the transaction's compute resource to the session's effective warehouse.
```
meegoo: `SET warehouse='A'; BEGIN; INSERT /*+ SET_VAR(warehouse='B') */ INTO t1; INSERT /*+ SET_VAR(warehouse='C') */ INTO t2; COMMIT;` — t1 executes on B's CNs, t2 on C's, but publish routes all tablets through C (last-wins). t1's tablets get published on the wrong warehouse.

xiangguangyxg: Yes, if users use it this way, then the current behavior will be like this.

xiangguangyxg: Any suggestions?

kevincai: thought it should be on warehouse `A` in such case, otherwise the statement behavior is hard to understand.

xiangguangyxg: Even without this PR, it already behaves this way, the hint is ignored, which may confuse users.

fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
```diff
@@ -249,6 +258,10 @@ public static void loadData(long dbId, long tableId, ExplicitTxnState.ExplicitTx
 
         explicitTxnState.addTransactionItem(item);
 
+        // Sync the transaction's compute resource to the warehouse that actually executed
+        // this DML. Use NoAcquire to avoid side effects after the DML has succeeded.
+        transactionState.setComputeResource(context.getCurrentComputeResourceNoAcquire());
```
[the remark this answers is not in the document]
xiangguangyxg: compute resource is set before here, so it should not return null.

fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
kevincai: so we are good? need the user to adjust the sqls in the transaction?

The hint should only take effect for the individual sql it is embedded, once the sql ends, the hint deactivated, reasonable?

xiangguangyxg: ok

PR #71811, opened 2026-04-16
[BugFix] Register loaded indexes for compaction/replication/bypass-write transactions

xiangguangyxg:
## Why I'm doing:                                                                                                                                                                                                                                                          
                            
  During tablet split on a shared-data table, compaction transactions fail to commit with
  `"table 'xxx' has unfinished tablets: [yyy]"`.
                                                                 
  `LakeTableTxnStateListener.preCommit()` calls `getPartitionLoadedIndexesWithoutLock()` to
  determine which MaterializedIndex to validate. When no loaded indexes are registered, it
  falls back to `getLatestMaterializedIndices(ALL)`, which after tablet split returns I_new
  (with new tablet IDs) instead of I_old that the transaction actually wrote to.

  Full audit of all 12 `LoadJobSourceType` values found 3 types that never call
  `addPartitionLoadedIndexes()`: `LAKE_COMPACTION`, `REPLICATION`, `BYPASS_WRITE`.

## What I'm doing:

  Add `addPartitionLoadedIndexes()` calls for the three affected transaction source types:

  - **CompactionScheduler**: Add `PhysicalPartition` parameter to `beginTransaction()` and
    register current visible indexes inside it, under the same DB read lock scope as
    `collectPartitionTablets()`.
  - **ReplicationJob**: Register indexes from `partitionInfos` at the end of
    `beginTransaction()`.
  - **BypassWriteTransactionHandler**: Resolve committed tablets to their partition/index
    via `TabletInvertedIndex` and register before calling `prepareTransaction()`.

  Add two tests in `LakeTableTxnStateListenerTest`:
  - Positive: register loaded indexes before tablet split, verify `preCommit()` passes.
  - Negative: skip registration, verify `preCommit()` fails with "unfinished tablets".

Fixes #64986

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/compaction/CompactionScheduler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/compaction/CompactionScheduler.java
@@ -337,7 +337,7 @@ public class CompactionScheduler extends Daemon {
 
             // Note: call `beginTransaction()` in the scope of database reader lock to make sure no shadow index will
             // be added to this table(i.e., no schema change) before calling `beginTransaction()`.
-            txnId = beginTransaction(partitionIdentifier, info.computeResource);
+            txnId = beginTransaction(partitionIdentifier, partition, info.computeResource);
 
             partition.setMinRetainVersion(currentVersion);
 
@@ -516,8 +516,9 @@ public class CompactionScheduler extends Daemon {
         return beToTablets;
     }
 
-    // REQUIRE: has acquired the exclusive lock of Database.
-    protected long beginTransaction(PartitionIdentifier partition, ComputeResource computeResource)
+    // REQUIRE: has acquired the read lock of Database.
+    protected long beginTransaction(PartitionIdentifier partition, PhysicalPartition physicalPartition,
+            ComputeResource computeResource)
             throws RunningTxnExceedException, AnalysisException, LabelAlreadyUsedException, DuplicatedRequestException {
         long dbId = partition.getDbId();
         long tableId = partition.getTableId();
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

wyb (dismissed):

fe/fe-core/src/main/java/com/starrocks/lake/compaction/CompactionScheduler.java
```diff
@@ -516,8 +516,9 @@ protected Map<Long, List<Long>> collectPartitionTablets(PhysicalPartition partit
         return beToTablets;
     }
 
-    // REQUIRE: has acquired the exclusive lock of Database.
-    protected long beginTransaction(PartitionIdentifier partition, ComputeResource computeResource)
+    // REQUIRE: has acquired the read lock of Database.
```
kevincai, line 519: is this comment still accurate?

kevincai (approved):

fe/fe-core/src/main/java/com/starrocks/lake/compaction/CompactionScheduler.java
xiangguangyxg: yes, the original comment is wrong, now it is accurate

srlch (approved):

PR #71829, opened 2026-04-17
[BugFix] Fix EOF errors for SQL comments

tonywasson:
## Why I'm doing:

Comment-only SQL such as `-- SELECT 1;` or `/* comment */` can fail with an EOF syntax error even though StarRocks already lexes these as comments. This is especially annoying in MySQL client sessions, and it also makes `--` end-of-line comments feel inconsistent.

## What I'm doing:

- treat comment-only SQL as an empty statement in the StarRocks parser
- add focused FE parser coverage for:
  - `-- SELECT 1;`
  - `/* comment */`
  - inline end-of-line comments such as:
    `SELECT 1 -- or whatever`
    `, 2 + 2; -- trailing comment`
- add a short release note documenting that `--` comments no longer trigger EOF parse errors

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Validation:

- `mvn -pl fe-core -am -DskipITs -Dcheckstyle.skip -Dsurefire.failIfNoSpecifiedTests=false -Dmaven.clean.skip=true -Dtest=ParserTest test`

### DIFF
--- a/docs/en/release_notes/release-4.1.md
+++ b/docs/en/release_notes/release-4.1.md
@@ -155,6 +155,7 @@ Release Date: April 13, 2026
   - Supports multiple compression formats (GZIP/SNAPPY/ZSTD/LZ4/DEFLATE/ZLIB/BZIP2) for CSV file exports. [#68054](https://github.com/StarRocks/starrocks/pull/68054)
   - Supports `STRUCT_CAST_BY_NAME` SQL mode for name-based struct field matching. [#69845](https://github.com/StarRocks/starrocks/pull/69845)
   - Supports `last_query_id()` in `ANALYZE PROFILE` for easy query profile analysis. [#64557](https://github.com/StarRocks/starrocks/pull/64557)
+  - Supports `--` comments without EOF parse errors, including comment-only input and end-of-line comments in SQL statements.
 
 ### Management & Observability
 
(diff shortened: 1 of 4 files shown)


eshishki:
Thanks for the fix — the EOF error on comment-only input is a real bug (the grammar's `singleStatement+ EOF` requires at least one statement, and a pure comment lexes entirely onto the hidden channel, so there's nothing to match).

One suggestion on the approach: `hasNoVisibleTokens` runs a full separate lexer pass (`tokenStream.fill()`) on every statement before the real parse, which then lexes the input a second time — so it doubles lexing on the hot path for the sake of a rare comment-only case. Since the parser already throws on this input, it'd be cheaper to handle it in the error path instead: catch the EOF parse failure and, if there were no visible tokens, return `EmptyStmt`. That keeps normal queries on a single lex pass and only does the extra check when parsing actually fails.

A couple of process things too: the commit isn't signed off so DCO is red — `git commit --amend --signoff && git push --force-with-lease` fixes it — and the PR is now CONFLICTING and needs a rebase.

PR #71845, opened 2026-04-17
[BugFix] Fix tablet stats after shared-data range split

xiangguangyxg:
## Why I'm doing:

Range-distribution split on shared-data PK tables copies the parent's segment
and delvec files to each child tablet and only scales `num_rows` / `data_size`
by the range-overlap ratio. The rowset's `num_dels` stays at the parent's full
cardinality, and the shared delvec still reports the full delete set for every
child, so `get_tablet_stats` computes `live_rows = child.num_rows - parent.num_dels`,
clamps the negative result to 0, and collapses the partition row count after
every split. Downstream consumers (`SHOW PARTITIONS`, compaction picker,
`data_size_after_deletion`) then see incorrect values.

## What I'm doing:

- `be/src/storage/lake/tablet_splitter.cpp`: read each rowset's total `num_dels`
  (falling back to `UpdateManager::get_rowset_num_deletes` when the field is
  absent) and allocate it across split groups proportional to each group's row
  share, using largest-remainder (Hare-Niemeyer). Use a 128-bit intermediate so
  the product `num_dels * num_rows` never overflows for large PK rowsets. Write
  the per-child `num_dels` alongside `num_rows` and `data_size` in `split_tablet`,
  clamped to the child's `num_rows`.
- `be/src/storage/lake/tablet_reshard_helper.cpp`: in `update_rowset_data_stats`
  scale `num_dels` the same way we already scale `num_rows` / `data_size`, so the
  cross-publish path (op_write / op_compaction / op_schema_change / op_replication
  / op_parallel_compaction) behaves consistently.
- `be/src/service/service_be/lake_service.cpp`: in `get_tablet_stats`, route
  rowsets with `has_range()` (the reliable "this rowset went through a reshard"
  signal) to the pre-scaled `num_dels` so reshard children never re-read the
  shared delvec and over-subtract. Non-reshard rowsets retain the existing
  accurate / approximate behavior.
- `be/src/storage/lake/tablet_merger.cpp`: `update_canonical` sums `num_dels`
  alongside `num_rows` and `data_size` so merge-back is a true inverse of split.
  DCHECK the `num_dels <= num_rows` invariant upheld by Step 1a's write-back
  clamp; no release-build clamp needed because tablet merge is greenfield (no
  deployed tablets carry the pre-fix legacy shape).
- `be/src/storage/lake/rowset.cpp`: clamp live rows to 0 in
  `Rowset::data_size_after_deletion` so callers that still observe an inflated
  `num_dels` (legacy reshard data, other paths fixed in follow-ups) do not see
  negative byte counts.
- Tests: new unit tests cover split `num_dels` conservation and fallback,
  cross-publish scaling across every op branch, merge-back accumulation, the
  `data_size_after_deletion` clamp, and the end-to-end PK split stats path in
  `lake_service_test`.

### Known limitations (follow-up PRs)

- Shared delvec file size is counted once per child in `data_size`, so
  partition-level `Σ data_size` overcounts by `(N-1) * shared_delvec_size`.
  Delvec files are typically small compared to segments; left as documented
  approximation.
- `replication_txn_manager.cpp:444` sets `num_dels = get_num_delete_files()`
  (file count, not delete cardinality). If such a rowset later gets a range
  via split cross-publish, the new `has_range()` route in `get_tablet_stats`
  will trust that stale count. Fixing the replication semantic is out of scope
  here and tracked separately.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/service/service_be/lake_service.cpp
+++ b/be/src/service/service_be/lake_service.cpp
@@ -1164,23 +1164,26 @@ void LakeServiceImpl::get_tablet_stats(::google::protobuf::RpcController* contro
                     for (const auto& rowset : (*tablet_metadata)->rowsets()) {
                         int64_t num_deletes = 0;
                         if (is_pk_tablet) {
-                            if (accurate_mode) {
+                            if (rowset.has_range()) {
+                                // Rowset went through a reshard (split or merge-back). Shared delvec
+                                // cardinality reflects the parent's full delete set, not this child's
+                                // share, so reading delvec would over-subtract. Use the pre-scaled
+                                // num_dels written by tablet_splitter / tablet_merger / tablet_reshard_helper.
+                                num_deletes = rowset.has_num_dels() ? rowset.num_dels() : 0;
+                            } else if (accurate_mode) {
                                 // Accurate mode (default): fetch delete vectors from object storage.
                                 // Reuses the already-loaded tablet metadata to avoid repeated loads per segment.
                                 num_deletes = static_cast<int64_t>(
                                         _tablet_mgr->update_mgr()->get_rowset_num_deletes(**tablet_metadata, rowset));
-                            } else {
+                            } else if (rowset.has_num_dels()) {
                                 // Approximate mode: prefer the pre-stored num_dels field in rowset
                                 // metadata. This avoids additional I/O while still providing a reasonable estimate.
                                 // Rows deleted but not yet compacted may be slightly overcounted.
-                                if (rowset.has_num_dels()) {
-                                    num_deletes = rowset.num_dels();
-                                } else {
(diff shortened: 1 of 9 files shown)


CelerData-Reviewer:
@codex review

wyb (approved):

srlch (approved):

PR #71858, opened 2026-04-17
[Enhancement] Allow random distribution on agg tables via opt-in flag

rickxz:
## Why I'm doing:

PR #60702 forbade `DISTRIBUTED BY RANDOM` on aggregate-key tables at both `CREATE TABLE` and `ALTER TABLE` time because non-aggregation queries against such tables could return inconsistent results when tablets are split or merged. That blanket rejection is correct for the general case, but it breaks a legitimate deployment pattern: tables that are created and altered dynamically by a UI layer where end-users cannot reliably pick meaningful hash columns. Operators in that situation currently patch #60702 out of every StarRocks release they consume, which blocks them from tracking vanilla releases.

This was raised upstream by Yves Burlot on the original PR: https://github.com/StarRocks/starrocks/pull/60702#issuecomment-3401832610. His suggestion was an opt-in flag defaulting to `false`, which is exactly what this PR implements.

## What I'm doing:

Adding a new FE config `enable_random_distribution_for_agg_table`, mutable, default `false`. When `false` (the default) the check added by #60702 applies unchanged, no behavior change versus today. When an operator sets it to `true`, `DISTRIBUTED BY RANDOM` is accepted on aggregate-key tables that do NOT declare `REPLACE` / `REPLACE_IF_NOT_NULL` columns, both at CREATE TABLE and ALTER TABLE (`OPTIMIZE`) time. REPLACE-family aggregate columns remain rejected even when the flag is on, because random distribution would break REPLACE semantics outright. This matches the pre-#60702 behavior, which also rejected `AGG_KEYS with replace`. The change is intentionally conservative: it restores exactly the set of cases that were legal before #60702, gated behind a single boolean, so users who accept the consistency trade-off can opt in and everyone else sees zero change.

The check sites live in two files, mirroring the pre-#60702 structure:

- `fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CreateTableAnalyzer.java` around line 737: the `CREATE TABLE` path, keyed on `stmt.isHasReplace()`.
- `fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java` around line 676: the `ALTER TABLE ... OPTIMIZE` path, keyed on the `hasReplace` local already computed from the new column defs at line 572.

Test coverage:

- `AnalyzeCreateTableTest.testRandomDistributionForAggKey` (unchanged) keeps proving the flag-off / current behavior.
- New `AnalyzeCreateTableTest.testRandomDistributionForAggKeyWithOptInFlag` flips the config on and asserts: `sum` aggregate succeeds, `replace` and `replace_if_not_null` still fail, `primary key` / `unique key` still fail (regression guard), `duplicate key` still succeeds.
- New `AnalyzeAlterTableStatementTest.testOptimizeRandomDistributionForAggTable` repeats the matrix against real OLAP tables via `ALTER TABLE ... DISTRIBUTED BY RANDOM`.
- Each test saves and restores `Config.enable_random_distribution_for_agg_table` in a `try`/`finally` so it does not leak into sibling tests.

Documentation:

- Added the new config entry in `docs/en/administration/management/FE_parameters/stats_storage.md` and its Chinese mirror, alphabetically placed next to `enable_auto_tablet_distribution`, documenting the consistency trade-off and the REPLACE caveat.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/docs/en/administration/management/FE_parameters/stats_storage.md
+++ b/docs/en/administration/management/FE_parameters/stats_storage.md
@@ -199,6 +199,15 @@ This topic introduces the following types of FE configurations:
   - If this parameter is set to `FALSE`, you need to manually specify the number of buckets when you create a table or add a partition. If you do not specify the bucket count when adding a new partition to a table, the new partition inherits the bucket count set at the creation of the table. However, you can also manually specify the number of buckets for the new partition.
 - Introduced in: v2.5.7
 
+### `enable_random_distribution_for_agg_table`
+
+- Default: false
+- Type: Boolean
+- Unit: -
+- Is mutable: Yes
+- Description: Whether to allow `DISTRIBUTED BY RANDOM` on aggregate-key tables at CREATE TABLE and ALTER TABLE time. The default `false` matches the behavior introduced by the analyzer check that rejects random distribution on aggregate tables, because non-aggregation queries over such tables may return inconsistent results when tablets are split or merged. Enable this only if you accept that trade-off — for example, on tables that are created and altered dynamically by a UI layer where end-users cannot reliably choose hash columns. Tables declaring `REPLACE` / `REPLACE_IF_NOT_NULL` aggregate columns remain rejected even when this flag is enabled, because random distribution would break `REPLACE` semantics.
+- Introduced in: -
+
 ### `enable_experimental_rowstore`
 
 - Default: false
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

PR #71860, opened 2026-04-17
[Enhancement] Add range-colocate join planner support

xiangguangyxg:
## Why I'm doing:

Range-distributed lake tables can already be placed into a colocate group
via P0/P1 (PACK shard groups), but the optimizer still emits a shuffle
exchange on every join because it has no spec to represent scan-local
range distribution. This PR (P2) teaches the planner about that locality
so qualifying joins skip the shuffle.

## What I'm doing:

Introduces optimizer-side support for range-colocate join, staged as a
single coherent change after ~14 rounds of design review.

**Core pieces**

- New `DistributionType.RANGE` (isolated from `SHUFFLE` to avoid the many
  `(HashDistributionSpec) shuffleSpec` casts in `PlanFragmentBuilder`).
  `toThrift()` on RANGE throws — this spec is scan-local only.
- `RangeDistributionSpec`: carries colocate columns + `EquivalentDescriptor`.
  `isSatisfy` returns true for `ANY`; for another range spec via `canColocate`
  (same group, both stable, non-empty partitions, equal colocate column
  count); for a `HashDistributionSpec(SHUFFLE_JOIN)` when every colocate
  column is covered (set-based equivalence, permutation-safe) by some
  required shuffle column. `SHUFFLE_AGG` / `SHUFFLE_ENFORCE` / `LOCAL`
  hash specs are rejected, scoping cross-type satisfaction to hash-join
  contexts.
- Spec production: `RelationTransformer` and `MvRewritePreprocessor`
  build a skeleton spec via `DistributionSpecHelper`;
  `OutputPropertyDeriver.visitPhysicalOlapScan` rebuilds with selected
  partitions (mirrors hash-local). `OptExpressionDuplicator` gets a
  `processRangeDistributionSpec` helper for MV-rewrite column rewiring.
- Framework invariants: `DistributionProperty.appendEnforcers` asserts
  range is never enforced; `getNullStrictProperty` is a no-op for range
  (columns are always `nullStrict=true`); `CostModel` case `RANGE` throws.
- `ChildOutputPropertyGuarantor.visitPhysicalJoin`: range dispatch runs
  before hash/hash. Both range + `canRangeColocateJoin` → skip exchange;
  otherwise convert range side(s) via `convertRangeToHashShuffle` and
  fall through. Gate order in `canRangeColocateJoin`: join-type allowlist
  (`INNER` / `LEFT OUTER` / `LEFT SEMI` / `LEFT ANTI`), `disable_colocate_join`,
  `canColocate` (structural), position-preserving pairing (same `k` on
  both sides — blocks swapped-column joins).
- `convertRangeToHashShuffle` emits `HashDistributionSpec(SHUFFLE_JOIN)`
  (not `SHUFFLE_ENFORCE`, which `HashDistributionDesc.isShuffle()` rejects).
- `visitPhysicalSetOperation`: top-of-method range-normalization pass
  converts every range child to hash `SHUFFLE_JOIN` before any
  `isShuffle()` precondition or `HashDistributionSpec` cast. Post-
  normalization: Union All → round-robin; Union Distinct → round-robin
  fallback; Intersect/Except → hash shuffle. All pre-P2 plan shapes.
- `OutputPropertyDeriver.visitPhysicalJoin`: emits left-dominated range
  output for range-range colocate join. `updateEquivalentDescriptor`
  type-dispatches on hash vs range. `visitPhysicalRepeat` preserves range
  only when every colocate column is in the repeat intersection (direct
  column-id intersection — equivalence through Repeat-nulled columns is
  unsafe). `updatePropertyWithProjection` extended with range branch.
- `PlanFragmentBuilder.isColocateJoin` accepts `RangeDistributionSpec`.
  `SPMPlan2SQLBuilder.getJoinDistributionHints` detects range colocate
  before the existing `HashDistributionSpec` cast.

**Rollout**

No new feature flag. Per-query kill uses the existing
`disable_colocate_join` session variable. Correctness of the emitted
colocate plan depends on the runtime dispatching scan ranges by PACK
shard group — this must land with the coordinator piece for the feature
to be usable at runtime. This PR is the FE planner side; the BE/coordinator
piece is tracked separately.

**Tests**

- `RangeDistributionSpecTest` (17): `isSatisfy` cross-type semantics;
  `canColocate` group / stability / partition / schema; `toThrift`
  throws for RANGE; permuted / partial shuffle coverage; `equals`/
  `hashCode` excludes mutable `EquivalentDescriptor`.
- `DistributionPropertyRangeInvariantTest` (2): `appendEnforcers` rejects
  range; `getNullStrictProperty` is identity for range.
- `ChildOutputPropertyGuarantorRangeTest` (8): join-type allowlist rejects
  right/full outer; `disable_colocate_join` kill; identity pairing
  accepts inner/left-semi; swapped-column join rejected; partial shuffle
  coverage rejected; shuffle-arity mismatch rejected.
- `ColocateTableIndexTest` +2: `isRangeColocateGroup` for known and
  unknown groups.
- Regression: `ColocateJoinTest` (44) and `JoinTest` (121) pass unchanged
  — hash colocate path is untouched.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
@@ -20,6 +20,7 @@ import com.google.common.collect.Lists;
 import com.starrocks.common.Pair;
 import com.starrocks.qe.ConnectContext;
 import com.starrocks.sql.ast.HintNode;
+import com.starrocks.sql.ast.JoinOperator;
 import com.starrocks.sql.optimizer.base.DistributionCol;
 import com.starrocks.sql.optimizer.base.DistributionProperty;
 import com.starrocks.sql.optimizer.base.DistributionSpec;
@@ -27,6 +28,7 @@ import com.starrocks.sql.optimizer.base.EquivalentDescriptor;
 import com.starrocks.sql.optimizer.base.HashDistributionDesc;
 import com.starrocks.sql.optimizer.base.HashDistributionSpec;
 import com.starrocks.sql.optimizer.base.PhysicalPropertySet;
+import com.starrocks.sql.optimizer.base.RangeDistributionSpec;
 import com.starrocks.sql.optimizer.base.RoundRobinDistributionSpec;
 import com.starrocks.sql.optimizer.cost.CostModel;
 import com.starrocks.sql.optimizer.operator.Operator;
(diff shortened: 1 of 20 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
```diff
@@ -424,6 +532,32 @@ public Void visitPhysicalJoin(PhysicalJoinOperator node, ExpressionContext conte
...
+        DistributionSpec leftRawSpec = leftChildOutputProperty.getDistributionProperty().getSpec();
+        DistributionSpec rightRawSpec = rightChildOutputProperty.getDistributionProperty().getSpec();
+        boolean leftIsRange = leftRawSpec instanceof RangeDistributionSpec;
+        boolean rightIsRange = rightRawSpec instanceof RangeDistributionSpec;
+        if (leftIsRange && rightIsRange) {
+            if (canRangeColocateJoin(node.getJoinType(),
+                    (RangeDistributionSpec) leftRawSpec,
+                    (RangeDistributionSpec) rightRawSpec,
+                    leftShuffleColumns, rightShuffleColumns)) {
+                return visitOperator(node, context);
```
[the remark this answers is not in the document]
xiangguangyxg: done

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/base/DistributionSpec.java
```diff
@@ -59,6 +59,7 @@ public enum DistributionType {
         SHUFFLE,
         GATHER,
         ROUND_ROBIN,
+        RANGE,
```
stephen-shelby: RANGE feels too broad here and is easy to read as a regular executable distribution / exchange type.
This spec is really scan-local and only used to recognize the range colocate fast path, so I’d strongly suggest renaming it to RANGE_LOCAL (or another similarly explicit name) to better reflect its semantics.

stephen-shelby:
This planner fast path seems to depend on the coordinator/runtime side for correctness, especially once tablet split / merge enters the picture. If that dependency is not guaranteed to land together, do we need a stronger guard here (feature gate / required e2e test / merge dependency) to avoid exposing a planner-only fast path too early?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/base/DistributionSpec.java
xiangguangyxg: changed to RANGE_LOCAL

stephen-shelby (approved):

HangyuanLiu (approved):

Issue #67158, opened 2025-12-23
Queries over Iceberg don't use Iceberg metadata


mixermt:
We have a large table (over 250TB) partitioned by hour (ts column). 
The ts column contains only hours and has full iceberg statistics over it, which have been verified in Iceberg metadata.

However, when we perform a simple query to retrieve the maximum and minimum values of the ts column, I observe that the query profile shows a full scan of the ts column instead of checking the Iceberg metadata for the statistics.

We have tried to apply following parameters but still full scan performed
```
enable_rewrite_simple_agg_to_hdfs_scan=true,
enable_rewrite_simple_agg_to_meta_scan=true
```

### Steps to reproduce the behavior (Required)
``` 
SELECT
    /*+ SET_VAR(
    enable_rewrite_simple_agg_to_hdfs_scan=true,
    enable_rewrite_simple_agg_to_meta_scan = true)
      */
    min(ts), max(ts)
FROM our_schema.iceberg_table;
```

### Expected behavior (Required)
Queries over Iceberg should use Iceberg metadata to avoid costly scan of the data

### Real behavior (Required)
Result in 1TB scan and execution time of ~4m on our, ~640 cores cluster.

### StarRocks version (Required)
4.0.2-1f1aa9c

Query Profile attached:
```
...
on-default:
                 - size: 567
                 - columns: {..., ts=full, ...}
...
- HdfsIOMetrics: 
                 - TotalBytesRead: 967.851 GB
...
```

[max_min_iceberg_query_ profile.txt](https://github.com/user-attachments/files/24316873/max_min_iceberg_query_.profile.txt)

eshishki:
1. enable_rewrite_simple_agg_to_meta_scan  Only for OlapTable

  File: RewriteSimpleAggToMetaScanRule.java

  When it applies:
  - Only for DUPLICATE_KEY StarRocks tables
  - No GROUP BY
  - No HAVING or filters
  - No LIMIT
  - Aggregations: MIN, MAX, COUNT(*), COUNT(not-null column)
  - Columns must have zonemap index (not string/complex types for MIN/MAX)
  - No delete operations on the table

  How it works: Uses LogicalMetaScanOperator — reads min/max/count from metadata rowsets (zonemap index) instead of data files.

 ---
 2. enable_rewrite_simple_agg_to_hdfs_scan (for external tables)

  File: RewriteSimpleAggToHDFSScanRule.java

  When it applies:
  - For Hive, Iceberg, FileScan tables
[...]

PR #71913, opened 2026-04-20
[BugFix] Preserve Iceberg column stats in manifest data file cache

dirtysalt:
## Why

When `iceberg_manifest_cache_with_column_statistics` is enabled, StarRocks expects cached Iceberg `DataFile` objects to preserve column statistics.

However, if the first query against a table does not request column stats, for example `SELECT * FROM t`, the manifest reader may fill the data file cache with `DataFile` objects that do not contain stats. Later queries such as `SELECT min(id) FROM t` can reuse that cache entry and miss the column stats.

## What

This change makes the manifest reader include stats columns when filling the data file cache with metrics enabled.

It also adds a unit test to verify that the data file cache keeps lower bounds, upper bounds, value counts, and null value counts even when the scan projection itself does not request stats.

## Behavior

```text
Iceberg manifest
    |
    v
ManifestReader
    |
    |-- iceberg_manifest_cache_with_column_statistics=true
    |      -> read stats columns while filling dataFileCache
    |
    |-- iceberg_manifest_cache_with_column_statistics=false
           -> cache DataFile without stats

dataFileCache
    |
    v
Query planning / optimization
    |
    |-- enable_iceberg_column_statistics=true
    |      -> optimizer may use cached Iceberg column stats
    |
    |-- enable_iceberg_column_statistics=false
           -> stats may exist in cache, but optimizer does not use them

Catalog property:
  iceberg_manifest_cache_with_column_statistics
    Controls what is stored in the manifest data file cache.

Session variable:
  enable_iceberg_column_statistics
    Controls whether query planning uses Iceberg column stats.

Query behavior:
  SELECT * FROM t
    Can fill cache, but does not need stats itself.

  SELECT min(id) FROM t
    May use column stats if enabled and available.

```

Fixes #67158

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/iceberg/ManifestReader.java
+++ b/fe/fe-core/src/main/java/org/apache/iceberg/ManifestReader.java
@@ -251,6 +251,8 @@ public class ManifestReader<F extends ContentFile<F>> extends CloseableGroup
     }
 
     private CloseableIterable<ManifestEntry<F>> entries(boolean onlyLive) {
+        boolean shouldCacheDataFiles = shouldCacheDataFiles();
+        boolean shouldCacheDeleteFiles = shouldCacheDeleteFiles();
         if (hasRowFilter() || hasPartitionFilter() || partitionSet != null) {
             Evaluator evaluator = evaluator();
             InclusiveMetricsEvaluator metricsEvaluator = metricsEvaluator();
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

Youngwb (approved):

stephen-shelby (approved):

PR #71919, opened 2026-04-20
[Enhancement] Pin Iceberg snapshot across Hybrid PCT multi-batch refresh

Youngwb:
## Why I'm doing:

Iceberg materialized views refreshing in Hybrid mode (IVM attempt + PCT fallback) could mix data from different Iceberg snapshots within a single refresh job when the PCT path splits work into multiple batches:

- Each batch independently called `refreshExternalTable()` and `syncPartitions()`. Background `CachingIcebergCatalog.refreshCatalog()`, concurrent MV refreshes, or cache TTL expiration could advance the cached Iceberg table between batches.
- Different batches would enumerate different partitions, scan different data, and produce an MV whose partitions reflect inconsistent Iceberg snapshots.
- The persisted TVR checkpoint became divorced from what was actually scanned, causing subsequent change detection to miss deltas.

## What I'm doing:

Introduce a **pinned PCT mode** that freezes one target Iceberg snapshot per base table for the whole refresh job. All partition enumeration, scan, and partition-info persistence operate against exactly that frozen snapshot — regardless of background catalog refreshes or newer refresh jobs.

### Infrastructure

- `MvTaskRunContext.MVRefreshRuntimeState.pinnedTvrMap`: runtime `Map<String, TvrVersionRange>` keyed by `BaseTableInfo.getTableIdentifier()` (stable for external tables, unlike `tableId`).
- `PCTTableSnapshotInfo.pinnedRange`: per-table frozen snapshot for partition-info persistence.
- `BaseMVRefreshProcessor.isPinnedMode()` / `setupPinnedRangesIfNeeded()`: state predicate + hydration.
- `BaseMVRefreshProcessor.getStartTaskRunId()`: shared null-safe accessor.

### Hook-based freeze timing

- `MVPCTBasedRefreshProcessor.afterSyncHook`: one-shot callback invoked right after `syncAndCheckPCTPartitions` and before plan building, with clear-before-run semantics so a throwing hook cannot leave a stale reference.
- `MVHybridBasedRefreshProcessor.switchToPCTRefresh` installs the freeze hook on `pctProcessor` instead of running after the delegate call — pinned TVR is visible to the plan builder for batch 1 too, whose `TableRelation.getTable()` may re-fetch a fresher instance than `snapshotBaseTables`.
- `MVPCTBasedRefreshProcessor.getProcessExecPlan` early-SKIPs stale batches whose pinning owner has been overwritten by a newer job.

### Consumers

- `syncAndCheckPCTPartitions` skips `refreshExternalTable` entirely in pinned mode.
- `PCTTableSnapshotInfo.updatePCTExternalPartitionInfos` reads partition info at the frozen snapshot when pinned.
- `PCTTableSnapshotInfo.checkBaseTableChanged` short-circuits in pinned mode — partition DDL at newer snapshots is irrelevant to what we scan.
- `MVPCTRefreshPlanBuilder.buildInsertPlan` injects frozen TVR into every `TableRelation`; the lookup is self-gating through `pinnedTvrMap` (only IVM-supported connectors populate it).
- `PartitionDiffer` + `RangePartitionDiffer` + `ListPartitionDiffer` accept pinned ranges via `setPinnedRanges`; partitioners feed them from runtime state before `computePartitionDiff`.

### Merge isolation + safety guards

- `TaskRun.PINNED_REFRESH_JOB_ID` set only on pinned batches; added to `MV_COMPARABLE_PROPERTIES` so pinned batches from different jobs do not merge in TaskRunManager queue dedup, and to `MV_UNCOPYABLE_PROPERTIES` so it cannot leak onto unrelated runs. **Pure PCT merge semantics are unchanged.**
- `MVPCTBasedRefreshProcessor.updateVersionMeta` has owner-match guards on both TVR promote and `clearTempBaseTableInfoTvrDeltaState`, re-reading `AsyncRefreshContext` after `updatePCTMeta`'s copy-on-write swap. A stale batch that somehow passed the early-SKIP cannot overwrite or clear a newer job's pending state.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/MvTaskRunContext.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/MvTaskRunContext.java
@@ -19,6 +19,7 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.starrocks.catalog.Table;
 import com.starrocks.catalog.TableProperty;
+import com.starrocks.common.tvr.TvrVersionRange;
 import com.starrocks.scheduler.mv.BaseTableSnapshotInfo;
 import com.starrocks.scheduler.mv.pct.PCTPartitionTopology;
 import com.starrocks.scheduler.mv.pct.PCTRefreshScope;
@@ -31,6 +32,10 @@ public class MvTaskRunContext extends TaskRunContext {
     public static class MVRefreshRuntimeState {
         private final Map<Long, BaseTableSnapshotInfo> snapshotBaseTables = Maps.newHashMap();
 
+        // Pinned TvrVersionRange per base table, keyed by Table.getTableIdentifier() (stable across
+        // connector getTable() calls, unlike tableId for external tables).
+        private final Map<String, TvrVersionRange> pinnedTvrMap = Maps.newHashMap();
+
         public Map<Long, BaseTableSnapshotInfo> getSnapshotBaseTables() {
             return snapshotBaseTables;
         }
@@ -40,8 +45,13 @@ public class MvTaskRunContext extends TaskRunContext {
             this.snapshotBaseTables.putAll(snapshotBaseTables);
         }
 
+        public Map<String, TvrVersionRange> getPinnedTvrMap() {
+            return pinnedTvrMap;
+        }
+
         public void reset() {
             snapshotBaseTables.clear();
+            pinnedTvrMap.clear();
         }
     }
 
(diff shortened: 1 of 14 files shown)


CelerData-Reviewer:
@codex review

dirtysalt (approved):

stephen-shelby (approved):

PR #71922, opened 2026-04-20
[Enhancement] Prune mv whose group by or predicate columns not cover columns from query

kaijianding:
## Why I'm doing:

It's slow to calculate best MVs when there are lots of.  
It's a follow up of  #51044

In my scenario, a sql may have more than 50 joins.
The most large table has 74 **SPG** MVs, then aggregate pushdown rule can't choose a best MV for one sub-query in a limited time window.  

There are lots of MV like `select bitmap_agg(spu_id) from t group by pt` while part of a very complex sql is like

```sql
select t.spu_id, pt, sum(t.pv), ...
from t
where t.spu_id in (select spu_id from sub_table where type=xx group spu_id) 
and t.pt between ... and ...
group by t.spu_id, t.pt
```

but there are much less mv like `select ... group by spu_id,pt`

Currently, almost all MV contains column spu_id (no matter spu_id appears in bitmap(spu_id) or group by spu_id), the best MV can't be calculated within the timeout window `optimizer_materialized_view_timelimit=2000` and `cbo_materialized_view_rewrite_candidate_limit=100`.

After this patch is applied, the plan time is reduced by 2~3s in my production environment, which is a lot for plan phase.

## What I'm doing:

For agg pushdow rule, prune MVs that don't cover group by+predicate columns from query. So that a mv's spu_id appears in `bitmap_agg(spu_id)` rather than `group by spu_id` can be pruned early.

And this methodology can also be applied where query and MV are both SPG(**the no join case, which is not covered by aggregate pushdown rule**), and also prune MV by comparing predicate columns of query and MV.

## How it's done

If MV is not SPG, prune is not happened.
A query or sub-query is SPG, continue, otherwise do not prune.

The query expression and mv expression are both SPG.
A SPG MV is sustained only when MV's group by columns ⊇ query group-by columns, which is obviously true.   

For predicate comparison, it is little more complex.
Because of https://docs.starrocks.io/docs/using_starrocks/async_mv/use_cases/query_rewrite_with_materialized_views/#predicate-union-rewrite  , it can't simply prune MV when query's predicate columns doesn't contain MV's predicate columns.

1. MV has no predicate and has only group by columns: MV's group by columns should contain all query's group by columns + predicate columns 
2. MV has predicate columns and group by columns:
2.1 query has same predicate columns, MV's predicate columns + group by columns ⊇ query's predicate columns + group-by columns, it is obviously true.
2.2 query has no predicate columns, MV's predicate columns + group by columns ⊇ query's group-by columns, in this case, this MV maybe valid or maybe not, due to https://docs.starrocks.io/docs/using_starrocks/async_mv/use_cases/query_rewrite_with_materialized_views/#predicate-union-rewrite , so **DO NOT** prune this MV, let later materialize view optimizer rules decide.

For these 2 cases: MV's predicate columns + group by columns ⊇ query's predicate columns + group-by columns

Why there are 2 caches?
1. There is lazy cache for group by/predicate columns in SPG MV to avoid duplicated group by/predicate columns extract because `doPrune()` will be called multi-times.
2. A query column guava cached is introduced, because a **SPG** query/sub-query will be compared to multiple MVs.  This cache's size should be set to QPS * number of sub queries, I think default 1000(20 qps * 50 joins) is good enough.
Why `mv_aggregate_join_push_down_query_column_cache_max_size` is `mutable=false`? Because guava cache's size is fixed when cache is constructed, the cache size can't be modified at runtime. And the size of this cache only impact performance not correctness, and also, actually this part is not a bottleneck, this cache is a NICE TO HAVE.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

kaijianding:
@codex review

HangyuanLiu:
Hi @kaijianding, the PR description is quite brief. Could you expand it so reviewers can follow the design more easily?

1. **Problem scenario**: What workload/customer triggered this? How many MVs are typically present, what does the query shape look like, and which phase of the optimizer is the bottleneck? A before/after timing would help a lot.

2. **Approach**: Why pick "column-name coverage pruning" as the direction? Is the key insight that it can replace O(MV-rewrite attempts) with O(num-columns) checks without introducing false prunes?

3. **Solution**: Please break down the changes:
   - The pruning condition itself (MV grouping cols ⊇ query group-by; MV grouping ∪ predicate cols ⊇ query predicate).
   - Where it applies. The patch touches two call sites: `MaterializationContext.prune` (affects **all** single-table SPG MV rewrite rules) and `AggregateJoinPushDownRule.validMv` (multi-table join → single-table MV). The PR title only mentions agg-push-down, but the former has a broader blast radius — please confirm this is intentional.
   - What repeated work each cache eliminates: MV-side lazy cache (`groupingColumns`/`predicateColumns`) vs. query-side cache (`queryColumnCache`).
   - Why is `mv_aggregate_join_push_down_query_column_cache_max_size` declared `mutable = false`?

4. **Conservativeness claim**: Could you state in the PR description that "this pruning only removes MVs that fail column-name coverage and never drops an MV that downstream rewrite could have used"? That spares reviewers from re-deriving the correctness argument.

kaijianding:
> Hi @kaijianding, the PR description is quite brief. Could you expand it so reviewers can follow the design more easily?
> 
> 1. **Problem scenario**: What workload/customer triggered this? How many MVs are typically present, what does the query shape look like, and which phase of the optimizer is the bottleneck? A before/after timing would help a lot.
> 2. **Approach**: Why pick "column-name coverage pruning" as the direction? Is the key insight that it can replace O(MV-rewrite attempts) with O(num-columns) checks without introducing false prunes?
> 3. **Solution**: Please break down the changes:
>    
>    * The pruning condition itself (MV grouping cols ⊇ query group-by; MV grouping ∪ predicate cols ⊇ query predicate).
>    * Where it applies. The patch touches two call sites: `MaterializationContext.prune` (affects **all** single-table SPG MV rewrite rules) and `AggregateJoinPushDownRule.validMv` (multi-table join → single-table MV). The PR title only mentions agg-push-down, but the former has a broader blast radius — please confirm this is intentional.
>    * What repeated work each cache eliminates: MV-side lazy cache (`groupingColumns`/`predicateColumns`) vs. query-side cache (`queryColumnCache`).
>    * Why is `mv_aggregate_join_push_down_query_column_cache_max_size` declared `mutable = false`?
> 4. **Conservativeness claim**: Could you state in the PR description that "this pruning only removes MVs that fail column-name coverage and never drops an MV that downstream rewrite could have used"? That spares reviewers from re-deriving the correctness argument.

@HangyuanLiu pr description is updated, please check.

CelerData-Reviewer:
@codex review

PR #71928, opened 2026-04-21
[BugFix] Fix pipeline event scheduler dispatch tautology

Hyper-FF:
- observer.cpp: _do_update() used `state != INPUT_EMPTY || state != OUTPUT_FULL` which is always true, making the event-specific dispatch branches (source/sink/all) unreachable and forcing every wakeup through try_schedule. Flip to `&&` so the dispatch matches the intent described in the comment: take the fast path only when the driver is blocked for a reason other than INPUT_EMPTY / OUTPUT_FULL.

- pipeline_timer: PipelineTimerTask used release/acquire on _has_consumer paired with a seq_cst store on _finished, which is not enough for the Dekker-style handshake between doRun() and waitUtilFinished(). With acquire-release only, a producer whose load of _has_consumer is reordered before the consumer's release store can miss the consumer, while the consumer's acquire load of _finished in the while-loop need not observe the producer's store, causing a lost wakeup and a hang. Promote both _has_consumer accesses to seq_cst so at least one side always observes the other's write.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/pipeline/schedule/observer.cpp
+++ b/be/src/exec/pipeline/schedule/observer.cpp
@@ -59,9 +59,12 @@ void PipelineObserver::_do_update(int event) {
     }
 
     if (driver->is_in_blocked()) {
-        // In PRECONDITION state, has_output need_input may return false. In this case,
-        // we need to schedule the driver to INPUT_EMPTY/OUTPUT_FULL state.
-        bool pipeline_block = driver->driver_state() != DriverState::INPUT_EMPTY ||
+        // When the driver is blocked for a reason other than INPUT_EMPTY / OUTPUT_FULL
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

trueeyu (approved):

stdpain (approved):

PR #71932, opened 2026-04-21
[Enhancement] Stop leader task safely part2: Stop leader-only daemons gracefully on demotion

gengjun-git:
## Why I'm doing:

  Part 2 of safe leader demotion (#63357). Part 1 (#70911, merged) landed the
  primitives — `LeaderLease`, admission fence, `JournalWriter` sealing, and
  failure-visible `EditLog` APIs. Demotion still cannot avoid `System.exit(-1)`
  because leader-only daemons have no way to stop cleanly:

  - They extend `FrontendDaemon` / `Daemon`, which wrap `Thread` directly, so
    once stopped they cannot be restarted when this FE is re-elected.
  - They own leader-session state (pending queues, executors, in-flight maps)
    that must be released *immediately* on demotion, not on the next
    activation — otherwise memory is retained on followers and stale leader
    state can leak into replay paths.
  - They have no built-in lease check, so each subclass would have to
    re-implement fence logic to notice demotion.

  ## What I'm doing:

  ### New framework: `LeaderDaemon`

  A composition-based base class (holds a `Thread`, doesn't extend it) that:

  - Is **restartable**: `start()` after `stopGracefully()` spins up a fresh
    worker, so the same singleton Mgr instance works across leader sessions.
  - **Self-stops on lease invalidation**: each iteration captures and
    revalidates a `LeaderLease`; once invalid the loop exits without needing
    subclass cooperation.
  - Exposes two cleanup hooks:
    - `onBeforeStop()` — runs *before* the worker is interrupted; subclasses
      clear leader-session state here.
    - `onAfterStop()` — runs after the worker has actually exited (or
      timed out).
  - `stopGracefully(timeoutMs)` is the coordinated shutdown: hook → interrupt
    → join → re-interrupt on timeout → after-hook. Idempotent.

  ### Representative migrations (4 daemons)

  Each subclass owns its cleanup logic; the framework only provides the seam.

  - `TabletScheduler` — clears `pendingTablets`, `allTabletIds`,
    `runningTablets`, `schedHistory`, `backendsWorkingSlots`, `loadStatistic`,
    and timing watermarks.
  - `PublishVersionDaemon` — shuts down `taskExecutor` and
    `deleteTxnLogExecutor`, clears `publishingTransactionIds`.
  - `HeartbeatMgr` — shuts down its executor and nulls the reference so the
    next activation can rebuild it.
  - `ReportHandler` (outer + nested `ResourceReportDaemon`) — clears
    `reportQueue`, `resourceReportQueue`, `pendingTaskMap`, and the static
    `TABLET_TO_DROP_TIME` map. Also:
    - Adds an inbound fence at `handleReport` via
      `captureLeaderLeaseOrThrow()`.
    - Converts the blocking `queue.take()` in the consumer loop to
      `queue.poll(1s)` so lease re-checks happen regularly.
    - Fixes a pre-existing bug where `InterruptedException` was swallowed by
      a broad `catch (Exception)` and the consumer looped forever after
      interrupt.
  - `LeaderImpl.report` translates the new `IllegalStateException` from the
    inbound fence into the existing `INTERNAL_ERROR` / "current fe is not
    master" response, so BEs see the same error they already handle.

  ### Wiring

  - `GlobalStateMgr.stopLeaderOnlyDaemonThreads()` calls `stopGracefully` on
    the migrated daemons in reverse-start order.
  - Invoked from `transferToNonLeader` when leaving `LEADER`.
  - New config `leader_demotion_drain_timeout_sec` (default `30`) bounds the
    per-daemon join timeout.

  ### Scope note

  44 other leader-only daemons remain on `FrontendDaemon`. They are left for
  a follow-up PR to keep this review focused on the framework and the
  wiring. The migrated four cover the patterns the rest will reuse
  (executor-owning, queue-draining, nested daemons, inbound RPC fence).

  Related to #63357

  ## What type of PR is this:

  - [ ] BugFix
  - [ ] Feature
  - [x] Enhancement
  - [ ] Refactor
  - [ ] UT
  - [ ] Doc
  - [ ] Tool

  - [ ] Feature removed

  Leader demotion now stops leader-only daemons cleanly instead of relying on
  process exit. Inbound `report` RPC from BE may return "current fe is not
  master" during the demotion window (the same response it already returns
  from non-leader FEs).

  ## Checklist:

  - [ ] This is a backport pr

    - [ ] 4.1
    - [ ] 4.0
    - [ ] 3.5
    - [ ] 3.4

### DIFF
--- a/docs/en/administration/management/FE_parameters/log_server_meta.md
+++ b/docs/en/administration/management/FE_parameters/log_server_meta.md
@@ -1378,6 +1378,15 @@ This topic introduces the following types of FE configurations:
 - Description: When StarRocks deserializes TaskRun history rows for `information_schema.task_runs`, a corrupted or invalid JSON row will normally cause deserialization to log a warning and throw a RuntimeException. If this item is set to `true`, the system will catch deserialization errors, skip the malformed record, and continue processing remaining rows instead of failing the query. This will make `information_schema.task_runs` queries tolerant of bad entries in the `_statistics_.task_run_history` table. Note that enabling it will silently drop corrupted history records (potential data loss) instead of surfacing an explicit error.
 - Introduced in: v3.3.3, v3.4.0, v3.5.0
 
+### `leader_demotion_drain_timeout_sec`
+
+- Default: 180
+- Type: Int
+- Unit: Seconds
+- Is mutable: Yes
+- Description: Per-daemon timeout used during leader demotion to wait for each leader-only daemon worker thread to exit after being interrupted. If a worker is still alive when the timeout elapses, the FE process is terminated, because a stuck worker combined with a later re-election would run two workers against the same singleton state - strictly worse than a process restart. Increase this if you routinely see heartbeat/report/publish/tablet-scheduler daemons needing more than the default to drain.
+- Introduced in: v4.1
+
 ### `lock_checker_interval_second`
 
 - Default: 30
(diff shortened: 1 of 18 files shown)


CelerData-Reviewer:
@codex review

gengjun-git:
@codex review

alvin-phoenix-ai:
@codex review

gengjun-git:
@codex review

gengjun-git:
@codex review

gengjun-git:
@codex review

gengjun-git:
@codex review

srlch (approved):

HangyuanLiu (approved):

PR #71961, opened 2026-04-21
[Enhancement] Support condition update for lake column-mode partial update

luohaha:
## Why I'm doing:

Lake (shared-data) primary-key tables previously rejected the combination of column-mode partial column update (PCU) and `merge_condition` at the backend with a `NotSupported` status, even though both features were already supported in isolation. Users who want conditional upsert semantics (e.g. "only overwrite when the new version column is larger") on wide tables had to fall back to row-mode PCU, which rewrites whole rows and wastes I/O.

## What I'm doing:

Wire condition evaluation into the lake column-mode PCU handler and lift the hard block:

- `be/src/storage/lake/delta_writer.cpp`: replace the `NotSupported` guard with a validation — when `merge_condition` is set together with `COLUMN_UPDATE_MODE` / `COLUMN_UPSERT_MODE`, the condition column must be part of the partial update column set; otherwise the handler has no new value to compare against.
- `be/src/storage/lake/column_mode_partial_update_handler.{h,cpp}`:
  - `execute()` parses `txn_meta.merge_condition()` and, when set, forces a single column batch so the condition column and the rest of the partial columns share one `partial_schema` and one DCG, avoiding multi-DCG ambiguity.
  - `_update_source_chunk_by_upt()` takes the condition column index inside the partial schema; for each `(source_rowid, upt_rowid)` pair it does `compare_at` between the already-read source chunk and the upt chunk and drops losing pairs (`old > new`), so source rows that lose keep their previous values instead of being partially overwritten. Equal values let the new row win, matching `UpdateManager::_process_single_chunk_update_with_condition`.
  - Two static helpers (`_resolve_condition_cid`, `_locate_condition_idx_in_partial_schema`) keep `execute()` short and surface `Status` errors loudly instead of silently falling back to the no-condition path in release builds.

The change reuses the source/upt chunks that column-mode PCU already reads — no extra I/O is introduced for the condition check.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### BE unit tests (`be/test/storage/lake/partial_update_test.cpp`)
- `test_partial_update_with_condition` extended to run under `COLUMN_UPDATE_MODE` — covers `new<old` reject, `new==old` (values identical so visible state unchanged), `new>old` apply.
- New `test_column_mode_condition_missing_column_rejected` — covers the validation error when the merge_condition column is not in the partial set.

### SQL regression (`test/sql/test_partial_update_column_mode/{T,R}/test_partial_update`)
End-to-end on a shared-data TSP cluster (build 1186, FE 172.26.80.2):

- `test_lake_column_partial_update_with_condition_update` — happy path + missing-condition-column error path via stream load.
- `test_lake_column_pcu_condition_after_drop_column` — schema evolution: drop a non-PK, non-condition column (which shifts the merge_condition column's column-id), then run PCU + merge_condition. Pins that `_resolve_condition_cid` (by name) and `_locate_condition_idx_in_partial_schema` (within the same publish-time cid space) stay correct under cid shifts.
- `test_lake_column_pcu_condition_after_add_column` — schema evolution: append a nullable column with a non-NULL default, then run PCU + merge_condition over a mix of rejected, applied, and newly-inserted (UPSERT) rows. Pins that the appended column gets its declared default for both pre-existing rows and new UPSERT rows.
- `test_lake_column_pcu_condition_with_order_by` — pins the pre-existing BE rule that column-mode UPSERT (the path used by stream-load PCU) on a primary-key table with a non-PK sort key is rejected upstream of the new condition-update logic.

### DIFF
--- a/test/sql/test_partial_update_column_mode/R/test_partial_update
+++ b/test/sql/test_partial_update_column_mode/R/test_partial_update
@@ -275,3 +275,242 @@ select * from site_access;
 drop database test_column_partial_update_with_condition_update;
 -- result:
 -- !result
+
+
+
+
+-- name: test_lake_column_partial_update_with_condition_update @sequential @cloud
+create database test_lake_column_partial_update_with_condition_update;
+-- result:
+-- !result
+use test_lake_column_partial_update_with_condition_update;
+-- result:
+-- !result
+CREATE TABLE tab1 (
+  k INT NOT NULL,
+  ver INT NOT NULL,
+  v1 INT NULL,
+  v2 INT NULL
+) ENGINE=OLAP
+PRIMARY KEY(k)
+DISTRIBUTED BY HASH(k) BUCKETS 1
+PROPERTIES ("replication_num" = "1");
+-- result:
+-- !result
+insert into tab1 values (1, 10, 100, 1000), (2, 20, 200, 2000);
+-- result:
+-- !result
+select * from tab1 order by k;
+-- result:
+1	10	100	1000
+2	20	200	2000
+-- !result
+shell: curl --location-trusted -u root: -d "1,5,999|2,30,888" -XPUT -H label:lake_pcu_cond_ok -H column_separator:, -H row_delimiter:| -H partial_update:true -H partial_update_mode:column -H merge_condition:ver -H columns:k,ver,v1 ${url}/api/test_lake_column_partial_update_with_condition_update/tab1/_stream_load
+-- result:
+0
+{
+    "Status": "Success",
+    "Message": "OK"
+}
+-- !result
+sync;
+-- result:
+-- !result
+select * from tab1 order by k;
+-- result:
+1	10	100	1000
+2	30	888	2000
+-- !result
+shell: curl --location-trusted -u root: -d "1,5,999|2,30,888" -XPUT -H label:lake_pcu_cond_bad -H column_separator:, -H row_delimiter:| -H partial_update:true -H partial_update_mode:column -H merge_condition:v2 -H columns:k,ver,v1 ${url}/api/test_lake_column_partial_update_with_condition_update/tab1/_stream_load
+-- result:
+0
+{
+    "Status": "Fail",
+    "Message": "Merge condition column v2 does not exist. If you are doing partial update with condition update, please check condition column is in the given update columns. Otherwise please check condition column is in table tab1"
+}
+-- !result
+sync;
+-- result:
+-- !result
+select * from tab1 order by k;
+-- result:
+1	10	100	1000
+2	30	888	2000
+-- !result
+drop database test_lake_column_partial_update_with_condition_update;
+-- result:
+-- !result
+
+
+
+
+
+-- name: test_lake_column_pcu_condition_after_drop_column @sequential @cloud
+create database test_lake_column_pcu_condition_after_drop_column;
+-- result:
+-- !result
+use test_lake_column_pcu_condition_after_drop_column;
+-- result:
+-- !result
+CREATE TABLE tab1 (
+  k INT NOT NULL,
+  v1 INT NULL,
+  v2 INT NULL,
+  v3 INT NULL,
+  ver INT NOT NULL,
+  v4 INT NULL
+) ENGINE=OLAP
+PRIMARY KEY(k)
+DISTRIBUTED BY HASH(k) BUCKETS 1
+PROPERTIES ("replication_num" = "1");
+-- result:
+-- !result
+insert into tab1 values (1, 1, 1, 1, 10, 1), (2, 2, 2, 2, 20, 2);
+-- result:
+-- !result
+select * from tab1 order by k;
+-- result:
+1	1	1	1	10	1
+2	2	2	2	20	2
+-- !result
+alter table tab1 drop column v2;
+-- result:
+-- !result
+function: wait_alter_table_finish()
+-- result:
+None
+-- !result
+select * from tab1 order by k;
+-- result:
+1	1	1	10	1
+2	2	2	20	2
+-- !result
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

luohaha:
Pushed fixes for the two Codex review comments + BE UT compile error + DCO:

- **P1 (compare semantics)**: Changed the condition filter from `cmp < 0` to `cmp <= 0` in `_update_source_chunk_by_upt`, so equal condition values let the new row win, matching `UpdateManager::_process_single_chunk_update_with_condition`. Comments updated accordingly.
- **BE UT failure**: `finish_with_txnlog()` returns `StatusOr<TxnLogPtr>`, not `Status` — switched the assertion to `.status().is_not_supported()`.
- **DCO**: Ran clang-format locally so no unsigned fixup commit is needed; amended my original commit with the formatting + fixes and re-pushed with `-s`.
- **P2 (missing R file)**: Regenerating against a fresh shared-data TSP cluster now; will push the R entry as a follow-up commit once the run completes.

luohaha:
P2 fixed: generated the R-case for `test_lake_column_partial_update_with_condition_update` against a fresh shared-data cluster (FE 172.26.80.114, build 1186) and appended it to `test/sql/test_partial_update_column_mode/R/test_partial_update`. Validated locally with `run.py -v -C cloud -a sequential`.

luohaha:
Addressed the latest Copilot comments (SHA 762ae45643a):

- **DCHECK silent in release** — Changed `_locate_condition_idx_in_partial_schema` to return `StatusOr<int32_t>` and propagated via `ASSIGN_OR_RETURN` at the call site, so a broken single-batch invariant fails loudly instead of silently disabling condition filtering.
- **InternalError → InvalidArgument** — `_resolve_condition_cid` now returns `InvalidArgument` when the user-supplied `merge_condition` column is missing from the schema.
- **SQL test comment inaccuracy** — Updated the T-file comment to reflect that the error path is rejected by FE-side `Load.checkMergeCondition`, not BE.

Not addressed (left for a follow-up):
- **Case-insensitive column name match** in `_resolve_condition_cid` / `delta_writer.cpp` — the existing upsert path (`UpdateManager::_get_condition_column` at `be/src/storage/lake/update_manager.cpp:1856`) and `TabletSchema::field_index` are already case-sensitive. Fixing only my new code would diverge from the rest of BE; the casing-drift concern predates this PR and should be addressed consistently across the BE (including normalizing `merge_condition` in FE before sending to BE) in its own PR.

luohaha:
Added two more SQL regression cases under `test/sql/test_partial_update_column_mode` (SHA a23522a0dd4):

- **test_lake_column_pcu_condition_after_drop_column** — covers the schema-evolution concern: drop a non-PK, non-condition column (which shifts the column-id of the merge_condition column) and then run a column-mode PCU stream load with `merge_condition`. Confirms the handler resolves the condition column by NAME against the publish-time schema, so post-drop column-id shifts don't corrupt either the merge-condition lookup (`_resolve_condition_cid`) or the per-batch index lookup (`_locate_condition_idx_in_partial_schema`). Verified end-to-end: rejected row keeps prior values; winning row applies the partial update with the post-drop layout.
- **test_lake_column_pcu_condition_with_order_by** — covers the sort-key (ORDER BY) path. Pins the existing BE rule that column-mode UPSERT (the path used by stream-load PCU) on a primary-key table with a non-PK sort key is rejected upstream of the new condition-update logic. The condition-update plumbing therefore never runs for this case; the test pins the rejection so a future change to either the sort-key restriction or the upsert-mode partial logic surfaces here.

Both cases validated against a fresh shared-data TSP cluster (build 1186, FE 172.26.80.2). The order-by case's output is grep-filtered through curl to strip the BE/CN IP prefix that varies between runs.

luohaha:
Added `test_lake_column_pcu_condition_after_add_column` (SHA 85ebaeda943) covering ALTER TABLE ADD COLUMN: append a non-PK nullable column with a non-NULL default after some baseline rows, then run column-mode PCU + merge_condition. Pins three properties:

- Existing rows whose condition value loses keep all prior values (including the newly-defaulted appended column).
- Existing rows that win get their partial columns overwritten while the appended column stays at the column's declared default.
- New PKs taking the UPSERT path get the appended column from its declared default — confirms the `_handle_column_upsert_mode` path is unaffected by the schema-evolution-after-baseline pattern.

Validated end-to-end against the same shared-data TSP cluster.

sevev (dismissed):

test/sql/test_partial_update_column_mode/R/test_partial_update
```diff
@@ -274,4 +274,243 @@ select * from site_access;
...
+-- !result
+insert into tab1 values (1, 10, 100, 1000), (2, 20, 200, 2000);
+-- result:
+-- !result
+select * from tab1 order by k;
+-- result:
+1	10	100	1000
+2	20	200	2000
+-- !result
+shell: curl --location-trusted -u root: -T ${root_path}/lib/../common/data/stream_load/sr_lake_pcu_condition_1.csv -XPUT -H label:lake_pcu_cond_ok -H column_separator:, -H partial_update:true -H partial_update_mode:column -H merge_condition:ver -H columns:k,ver,v1 ${url}/api/test_lake_column_partial_update_with_condition_update/tab1/_stream_load
```
kevincai: if those .csv files are only used for this curl command, it can be replaced with `--data-binary $'1,2,3\n4,5,6\n'`

luohaha: Done in e893949d551 — replaced all 4 inline CSV files with `-d "row1|row2|..." -XPUT` plus `-H row_delimiter:|`, matching the pattern already used in `test/sql/test_condition_update/T/test_condition_update`. The CSVs were only consumed by these tests, so removing them keeps the test data co-located with the case.

luohaha: fixed

kevincai (approved):

sevev (approved):

PR #71966, opened 2026-04-21
[Enhancement] IVM commit owner guard + snapshot observability

Youngwb:
### IVM commit owner guard

IVMInsertLoadTxnCallback previously called clearTempBaseTableInfoTvrDeltaState unconditionally in afterCommitted. If the pinning owner on AsyncRefreshContext was overwritten by a newer refresh job between beforeCommitted and afterCommitted (e.g. leader fail-over + schedule of a different job), the unconditional clear would wipe the newer job's pending TVR state.

Scheduler-level serialization of TaskRuns on the same MV (TaskRunScheduler.canTaskRunBeScheduled) makes this unlikely in normal runtime, but the guard is required for leader fail-over / editlog replay paths and matches the pattern already applied in
MVPCTBasedRefreshProcessor.updateVersionMeta.

* capturedOwner field captures the observed owner in beforeCommitted.
* afterCommitted re-reads the live owner; only clears when it still matches (or no owner remains). Mismatch logs WARN with both owners.

Test: testIVMCallbackDoesNotClearNewerJobOwner manually stages the takeover scenario — owner=A installed, callback captures via beforeCommitted, owner overwritten to B, callback runs afterCommitted. Asserts B's owner and pending map survive.

### Pinned snapshot observability

Pinned PCT batches previously left no trace of which frozen snapshot they scanned — diagnosis required grepping refresh logs. Now the per-base-table frozen snapshot id is exposed on task run extra message:

* MVTaskRunExtraMessage.pinnedSnapshotIdMap: Map<String, Long> keyed by BaseTableInfo.getTableIdentifier(), value is the frozen snapshot id.
* BaseMVRefreshProcessor.setupPinnedRangesIfNeeded populates it alongside the runtime pinnedTvrMap hydration.

Visible via:
  SELECT extra_message FROM information_schema.task_runs WHERE task_name = 'mv-<id>' ORDER BY create_time DESC LIMIT 10;

The JSON will contain e.g.
  "pinnedSnapshotIdMap": {"iceberg0.db.t1:uuid-xyz": 2}

Pure PCT / non-pinned runs leave it empty, matching the existing refBasePartitionsToRefreshMap pattern.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/loadv2/IVMInsertLoadTxnCallback.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/loadv2/IVMInsertLoadTxnCallback.java
@@ -43,6 +43,8 @@ public class IVMInsertLoadTxnCallback implements InsertLoadTxnCallback {
     private long dbId;
     private MaterializedView mv;
     private final Map<BaseTableInfo, TvrVersionRange> baseTableInfoTvrDeltaMap = Maps.newConcurrentMap();
+    // Owner captured in beforeCommitted; afterCommitted refuses to clear on mismatch.
+    private String capturedOwner;
 
     public IVMInsertLoadTxnCallback(long dbId, long tableId) {
         this.dbId = dbId;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

Astralidea (approved):

dirtysalt (approved):

PR #71992, opened 2026-04-22
[BugFix] Fix combined txn log missing entries on incremental-only lake channel

luohaha:
## Why I'm doing:

In shared-data combined txn log mode, `LakeTabletsChannel` hard-codes **"sender_id == 0 collects all txn logs"** as the single aggregation rule. This rule silently drops txn logs whenever sender 0 is absent on a given channel, which happens routinely on incremental-only channels (channels opened via `_incremental_open_node_channel` at runtime rather than via the initial distribution).

### Symptom

After a successful `commit`, `publish_version` fails forever with:

```
com.starrocks.rpc.RpcException: Fail to publish version for tablets:[[...]],
error msg: txn log list does not contain txn log of tablet X, host: a.b.c.d
    at com.starrocks.lake.Utils.sendAggregatePublishVersionRequest(...)
```

The `combined_txn_log` file on S3 is missing exactly the tablets hosted by the CN whose channel had no sender 0 participant. Because the file is immutable, the txn can never be published and must be aborted.

### When it triggers

- **Any** stream load sink starts with an empty `_vectorized_partition` cache. The first row it sees for a given partition value triggers `_incremental_open_node_channel` — whether that partition is brand new or an existing 3-week-old partition. So the bug is not restricted to "auto-partition creating new partitions"; **late-arriving-data / backfill loads are high-risk** because they commonly touch multiple partitions whose owner CNs are not in the load's initial distribution.
- On the affected incremental-only channel, if the sink driver with `_sender_id == 0` happens not to route any row through it, sender 0's NodeChannel is never opened and its eos never arrives → the `sender_id == 0` collection branch never fires → the `_txn_log_collector` contents are lost.

### Why the current design is fragile

`sender_id == 0` is a static coordinator assignment. It works only when sender 0 is guaranteed to be present on every participating channel — true for the initial distribution path, violated as soon as `_incremental_open_node_channel` brings in a new CN.

## What I'm doing:

**Per-partition coordinator election, gated by an FE-dispatched transaction-level switch.**

Each partition elects its own coordinator (the smallest `sender_id` among those whose (incremental_)open tablet list covered the partition). The elected coordinator, by construction, has NodeChannels to every owner CN of its partition, so it can aggregate the full log set.

### Design

1. **Proto** (`gensrc/proto/internal_service.proto`): add `optional bool enable_per_partition_coordinator = 4;` to `LakeTabletParams`. No payload bloat — the claim set is derived from the existing `PTabletWithPartition.partition_id` already in every OpenRequest.
2. **Thrift** (`gensrc/thrift/DataSinks.thrift`): new `TOlapTableSink.enable_lake_per_partition_coordinator_txn_log`, carrying the flag FE → BE sink → BE CN.
3. **FE Config**: `lake_enable_per_partition_coordinator_txn_log` (mutable, default **true**). Lets operators disable at runtime if needed.
4. **FE `OlapTableSink`**: populate the Thrift flag from Config on every sink plan.
5. **BE sink** (`tablet_sink.{h,cpp}`, `tablet_sink_index_channel.cpp`): read Thrift into `_enable_lake_per_partition_coordinator_txn_log`, forward to proto `LakeTabletParams.enable_per_partition_coordinator` on every `NodeChannel::_open` (initial and incremental).
6. **BE CN** (`lake_tablets_channel.cpp`):
   - `_record_coordinator_claims`: latch the flag with defensive AND across opens; when enabled, build `partition_id -> min(sender_id)` by iterating `params.tablets()`.
   - Close-path dispatch:
     - flag true → each sender returns logs only for partitions it coordinates.
     - flag false/unset → legacy "sender 0 collects all".

### Why this is safe

- **Different partitions → different shard groups → different S3 paths** for `combined_txn_log`. Different senders writing different partitions never collide.
- **Same partition always has exactly one coordinator** — every CN that owns tablets of the partition sees the same claimer set (the sink's open/_incremental_open_node_channel broadcasts the same partition set to each owner CN), so every CN elects the same min-sender_id coordinator → exactly one writer downstream.
- For initial-distribution partitions, every sender claims, min == 0 → behaviorally equivalent to legacy for that partition. For incremental-only partitions, only the auto-partition trigger claims, so the correct non-zero sender becomes coordinator.

### Upgrade / downgrade compatibility

Upgrade order is **BE first, then FE**. The FE flag is the single source of truth per transaction:

| Scenario | Flag on the wire | Behavior |
|---|---|---|
| Old FE, any BE mix | unset | Every CN channel stays in legacy — **identical to pre-fix** |
| New FE + all BEs upgraded (guaranteed when FE is last) | true | Per-partition coordinator mode |
| New FE, want to disable at runtime | Config → false | Flip back to legacy without rolling FE |

A single transaction's `OpenRequest`s all originate from the same FE plan, so **the flag is uniform across every NodeChannel of the transaction**. This closes a gap the earlier `coordinated_partition_ids` design had, where a partially-rolled BE fleet could produce a partially-populated coordinator map and silently drop logs for unclaimed partitions. CN-side AND semantics further hardens against any pathological inconsistency: if any open disagrees, the whole channel falls back to legacy.

Proto/Thrift use new optional field ordinals — no reuse, no required. Conforms to repo rules. `combined_txn_log` file format is unchanged. Publish / vacuum / read paths are unchanged.

Note: the fix does not retroactively repair `combined_txn_log` files already written with missing entries (immutable on object storage). Clusters should abort and re-import any stuck txns before relying on this fix.

### Observability

Two new operation counters on every `LakeTabletsChannel` eos that enters the collection path, so operators and tests can confirm which mode is live:

| Metric | Unit | Meaning |
|---|---|---|
| `lake_txn_log_collect_legacy_total` | OPERATIONS | Legacy "sender 0 collects all" fallback fired (old FE, or AND'd-down channel). |
| `lake_txn_log_collect_per_partition_total` | OPERATIONS | Per-partition coordinator path fired (new FE + BE upgraded). |

A healthy cluster after full upgrade and Config = true should see only `per_partition` ticking; seeing `legacy` after upgrade points at a stale FE.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Test plan

**Unit tests (this PR)**

BE `lake_tablets_channel_test.cpp`:
- `test_flag_unset_bumps_legacy_metric` — old-FE RPC shape, legacy metric ticks.
- `test_flag_false_bumps_legacy_metric` — new FE with Config flipped off.
- `test_flag_true_bumps_per_partition_metric` — happy path, per-partition metric ticks.
- `test_flag_disagreement_falls_back_to_legacy` — AND semantics, channel reverts to legacy on any mismatch.

FE `OlapTableSinkTest.java`:
- `testPerPartitionCoordinatorFlagFromConfig` — verifies `Config.lake_enable_per_partition_coordinator_txn_log` → `TOlapTableSink.enable_lake_per_partition_coordinator_txn_log` for both true and false, and that the field is always set on the wire.

**Manual verification**

- Shared-data cluster with auto-partition + parallel streaming loads, reproducing the original loss scenario, verifying both counters reflect the expected mode across a Config flip.


CelerData-Reviewer:
@codex review

be/src/runtime/lake_tablets_channel.cpp
```diff
@@ -922,6 +1016,35 @@ Status LakeTabletsChannel::log_and_error_tablet_not_found(int64_t tablet_id, con
...
+    }
+
+    // Claim set = distinct partition_ids in this (incremental_)open's tablet list.
+    // `PTabletWithPartition.partition_id` is a required field on the wire, so no
+    // has_partition_id() guard is needed.
+    for (const auto& t : params.tablets()) {
+        int64_t pid = t.partition_id();
+        auto it = _partition_coordinator.find(pid);
+        // Smallest sender_id wins (deterministic election across concurrent claimers).
+        if (it == _partition_coordinator.end() || sender_id < it->second) {
```
kevincai, line 1047: Is it possible that for a certain partition, some of the senders run much faster than others, which its eos reaches to this TabletChannel before other senders open()/incremental_open() been processed?

luohaha: No, it must open first, otherwise the follow request will fail

kevincai: I mean 
* sender-A: open() -> add_chunk() + eos()
* sender-B: open() -> add_chunk() + eos()
Right now it assume that all senders already done the open() before any add_chunk()+eos().
What if send-A completes its eos() before sender-B's open() is still one the way?

When the channel processing A's eos(), it doesn't see the sender-B's open(), and think sender-A is the min sender and collects the txn log. Then what will happen to sender-B's open() and eos() request?

be/src/runtime/lake_tablets_channel.cpp
```diff
@@ -922,6 +1016,35 @@ Status LakeTabletsChannel::log_and_error_tablet_not_found(int64_t tablet_id, con
...
+    _enable_per_partition_coordinator = _enable_per_partition_coordinator && flag_enabled;
+    if (!_enable_per_partition_coordinator) {
+        // Legacy path: coordinator map unused.
+        return;
+    }
+
+    // Claim set = distinct partition_ids in this (incremental_)open's tablet list.
+    // `PTabletWithPartition.partition_id` is a required field on the wire, so no
+    // has_partition_id() guard is needed.
+    for (const auto& t : params.tablets()) {
```
wyb:
```
std::unordered_set<int64_t> unique_pids;
for (const auto& t : params.tablets()) {
    unique_pids.insert(t.partition_id());
}

std::lock_guard l(_partition_coordinator_mtx);
for (int64_t pid : unique_pids) {
    // select sender
}
```

be/src/runtime/lake_tablets_channel.cpp
luohaha: Good catch — I dug into this with the same worry, but I think the existing two-phase close in `TabletSinkSender::try_close` already prevents that ordering. Walking through it:

**Sink side (`be/src/exec/tablet_sink_sender.cpp:200-252`)**: when an index has any incremental node channel, `try_close` runs in two phases:

1. **Phase 1** — `try_finish()` on every *initial* node channel. This sends a request with `finished=true` (and `wait_all_sender_close=true`), but **not** eos.
2. Block in `try_close` until *all* initial channels report `is_finished()` (their finished-response has come back from the CN).
3. **Phase 3** — only then does it call `try_close()` on every node channel (initial + incremental), which is what actually sends eos.

**CN side (`be/src/runtime/tablets_channel.cpp:22-46`, `lake_tablets_channel.cpp:653-662`)**: the initial-founded channel records `_num_initial_senders = num_senders` at open. Each sink's "finished" RPC decrements it. `drain_senders` then blocks until `_num_initial_senders <= 0` — i.e. until **every sink** has called `try_finish` on the initial channels.

So the global timeline is:

```
all sinks process data in parallel
  └─ during data processing: _automatic_create_partition → _incremental_open_node_channel(P)
                              → incremental_open RPC sent to every owner CN of P
each sink, when its data exhausts, enters try_close:
  ├─ Phase 1: send finished=true on its initial node channels
  ├─ Phase 2: blocks until drain_senders on every initial CN releases
  │           ← BARRIER: all sinks have stopped data processing,
  │              therefore all _incremental_open_node_channel calls
  │              have been issued and the RPCs are already in flight / arrived
  └─ Phase 3: only now send eos on initial + incremental
```

The key invariant the barrier gives us: **no sender's eos can fire before all senders have finished sending data**, and `_incremental_open_node_channel` only happens *during* data processing. So by the time the first eos lands at a CN, every incremental_open from every sink is guaranteed to have already arrived.

In your two-sender example:

- `sender-A: open → add_chunk → eos`
- `sender-B: open → add_chunk → eos`

For A.eos to actually be sent on the wire, A must be past Phase 2, which requires B to have hit Phase 1 (`try_finish` on initial channels). That requires B to have *already exhausted its data* — including any `_incremental_open_node_channel` calls B was going to make. So A.eos arriving before B.open is not reachable in normal operation.

The remaining edge case is `drain_senders` timing out (60s) — but that's a degenerate scenario where the load is already failing, and per-partition coordinator semantics aren't the right thing to reason about there.

Net: I don't think we need to add post-wait re-snapshotting on top of this for correctness. Happy to add it as defense-in-depth if you'd prefer, though.

be/src/runtime/lake_tablets_channel.cpp
luohaha: Done in [edfa9981](https://github.com/StarRocks/starrocks/pull/71992/commits/edfa9981ace) — built the unique-pid set outside the lock, then a single map lookup per partition inside. Thanks for the suggestion.

kevincai (approved):

wyb (approved):

stdpain (approved):

PR #72006, opened 2026-04-22
[BugFix] Unregister HTTP connection even when isRegistered() is false

Hyper-FF:
## Why I'm doing:

HttpServerHandler.channelInactive gated unregisterConnection on HttpConnectContext.isRegistered(). But that flag is set AFTER ConnectScheduler.registerConnection puts the context into connectionMap, and ExecuteSqlAction runs catalog/db changes and SQL parsing before calling registerContextOnce at all. If the channel goes inactive in either window, isRegistered() is still false and the context never gets removed from connectionMap.

Symptom: after the TCP socket is already gone (not visible in netstat), ConnectScheduler's timer keeps firing ConnectContext.checkTimeout on the stranded context once per second, which hits the COM_SLEEP wait_timeout branch and repeatedly calls HttpConnectContext.kill, flooding fe.log with "kill wait timeout connection" / "kill query, ..., kill connection: true" for the same query id.

## What I'm doing:

drop the isRegistered() gate. unregisterConnection is already idempotent (connectionMap.remove on a missing id is a no-op and the counter decrement is guarded by `if (removed)`), so calling it on a never-registered context is safe.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/http/HttpServerHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/http/HttpServerHandler.java
@@ -174,14 +174,16 @@ public class HttpServerHandler extends ChannelInboundHandlerAdapter {
         HttpServerHandlerMetrics.getInstance().httpConnectionsNum.increase(-1L);
         HttpConnectContext context = ctx.channel().attr(HTTP_SQL_CONNECT_CONTEXT_ATTRIBUTE_KEY).get();
         if (context != null) {
-            // unregisterConnection is idempotent (map.remove on a missing id is a no-op),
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

before-Sunrise (approved):

satanson (approved):

kevincai (approved):

PR #72015, opened 2026-04-22
[BugFix] Honor partition_refresh_number in REFRESH MV WITH SYNC MODE

### LINKED ISSUE #71974: `partition_refresh_number` is ignored by `REFRESH MATERIALIZED VIEW ... WITH SYNC MODE`
## Summary

For partitioned materialized views, `partition_refresh_number` does not limit the number of partitions refreshed when the command uses `WITH SYNC MODE`.

The first sync refresh processes all stale partitions in one run, and subsequent sync refreshes become no-ops.

I initially hit this on an Iceberg-backed MV, but I also reproduced the same behavior on an OLAP-backed MV. So this is not Iceberg-specific; it appears to be a generic sync refresh behavior in the current implementation.

## Reproduction (OLAP-backed MV)

```sql
create database bug_sync_refresh;
use bug_sync_refresh;

CREATE TABLE t1 (
  id int,
  dt date,
  val int
)
DUPLICATE KEY(id, dt)
PARTITION BY RANGE(dt) (
  PARTITION p20240101 VALUES [('2024-01-01'), ('2024-01-02')),
  PARTITION p20240102 VALUES [('2024-01-02'), ('2024-01-03')),
  PARTITION p20240103 VALUES [('2024-01-03'), ('2024-01-04')),
  PARTITION p20240104 VALUES [('2024-01-04'), ('2024-01-05')),
  PARTITION p20240105 VALUES [('2024-01-05'), ('2024-01-06'))
)
PROPERTIES('replication_num'='1');

INSERT INTO t1 VALUES
  (1, '2024-01-01', 10),
  (2, '2024-01-02', 20),
  (3, '2024-01-03', 30),
  (4, '2024-01-04', 40),
  (5, '2024-01-05', 50);

CREATE MATERIALIZED VIEW test_mv1
PARTITION BY dt
REFRESH DEFERRED MANUAL
PROPERTIES (
  'replication_num' = '1',
  'partition_refresh_number' = '2'
)
AS
SELECT dt, sum(val) as sv
FROM t1
GROUP BY dt;

-- Round 1
REFRESH MATERIALIZED VIEW test_mv1 WITH SYNC MODE;
SELECT count(*) FROM test_mv1;
-- expected: 2
-- actual:   5

-- Round 2
REFRESH MATERIALIZED VIEW test_mv1 WITH SYNC MODE;
SELECT count(*) FROM test_mv1;
-- expected: 4
-- actual:   5

-- Round 3
REFRESH MATERIALIZED VIEW test_mv1 WITH SYNC MODE;
SELECT count(*) FROM test_mv1;
-- expected: 5
-- actual:   5
```

## Also reproduced on Iceberg-backed MV

I reproduced the same `5 / 5 / 5` behavior on an Iceberg-backed MV with the same shape:

- base table partitioned by `dt`
- MV partitioned by `dt`
- `partition_refresh_number = 2`
- `REFRESH MATERIALIZED VIEW ... WITH SYNC MODE`

So the issue does not seem to be connector-specific.

## Expected behavior

Either:

1. `WITH SYNC MODE` should still honor `partition_refresh_number` and process at most `N` stale partitions per refresh task/run, or
2. if sync refresh is intentionally defined to refresh all stale partitions at once, that behavior should be documented clearly because it is surprising and easy to misread.

## Actual behavior

`WITH SYNC MODE` refreshes all stale partitions in the first refresh even when `partition_refresh_number` is set to a smaller value.

## Implementation notes

The current behavior seems to come from the sync refresh path disabling next-batch generation entirely:

- `LocalMetastore.refreshMaterializedView(...)` passes the sync flag into `ExecuteOption`
- `TaskManager.executeTask()` routes sync requests to `executeTaskSync()`
- `MVPCTRefreshPartitioner.isGenerateNextTaskRun()` returns `false` for sync refresh
- `MVPCTRefreshPartitioner.filterMVToRefreshPartitions()` returns early when next-task generation is disabled, so the `partition_refresh_number` truncation logic is skipped

Relevant code paths in the current tree:

- `fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java`
- `fe/fe-core/src/main/java/com/starrocks/scheduler/TaskManager.java`
- `fe/fe-core/src/main/java/com/starrocks/scheduler/mv/pct/MVPCTRefreshPartitioner.java`

## Notes

I verified this on a local checkout at `189283f334c0` on April 21, 2026.

HangyuanLiu:
## Why I'm doing:

Before this change, `REFRESH MATERIALIZED VIEW ... WITH SYNC MODE` silently
ignored the `partition_refresh_number` table property and refreshed every
stale partition in a single task run. The property was intended as an
internal knob for resource/concurrency control (cap per-task-run partition
count), but the sync path dropped that cap entirely. On MVs with many
stale partitions this led to unbounded single-transaction refreshes that
could exhaust memory, blow past query timeouts, or overrun the
transaction-visible wait timeout.

Reproducer on OLAP-backed MV (also reproduced on Iceberg-backed MV):

```sql
CREATE TABLE t1 (id int, dt date, val int) DUPLICATE KEY(id, dt)
PARTITION BY RANGE(dt) (
  PARTITION p20240101 VALUES [('2024-01-01'), ('2024-01-02')),
  PARTITION p20240102 VALUES [('2024-01-02'), ('2024-01-03')),
  PARTITION p20240103 VALUES [('2024-01-03'), ('2024-01-04')),
  PARTITION p20240104 VALUES [('2024-01-04'), ('2024-01-05')),
  PARTITION p20240105 VALUES [('2024-01-05'), ('2024-01-06'))
) DISTRIBUTED BY HASH(id) BUCKETS 1;

CREATE MATERIALIZED VIEW test_mv1 PARTITION BY dt REFRESH DEFERRED MANUAL
PROPERTIES ('partition_refresh_number' = '2')
AS SELECT dt, sum(val) FROM t1 GROUP BY dt;

REFRESH MATERIALIZED VIEW test_mv1 WITH SYNC MODE;
-- Before: one task run refreshed all 5 partitions, ignoring N=2.
-- After : one SQL call still refreshes all 5 partitions (external contract
--         preserved), internally fanning out into 3 task runs (2 + 2 + 1).
```

## What I'm doing:

Make sync refresh honor `partition_refresh_number` **internally** while
preserving the external contract that a single `WITH SYNC MODE` call returns
only after all stale partitions have been refreshed.

- **`MVPCTRefreshPartitioner.isGenerateNextTaskRun`**: drop the sync
  short-circuit so `filterPartitionByRefreshNumber` applies to sync mode as
  well. Each internal batch is now bounded by the property as intended.
- **`MVPCTBasedRefreshProcessor.generateNextTaskRunIfNeeded`**: always
  publish the next-batch cursor to `MVTaskRunExtraMessage`. For sync mode,
  defer dispatch to the caller loop (avoid racing it); for async mode,
  dispatch the follow-up task run as before.
- **`TaskManager.executeTaskSync`**: loop on the persisted cursor and
  submit follow-up batches one at a time until the cursor is empty. The
  SQL thread holds only the current batch's future at any moment, so the
  task-runs thread pool cannot be starved by deep sync nesting.
- **`MVPCTBasedRefreshProcessor.buildNextBatchOption(TaskRun)`**: single
  shared static builder used by both the sync loop and the async follow-up
  spawn. Centralizing the option construction ensures the two continuation
  paths can never drift apart on property handling
  (`MV_UNCOPYABLE_PROPERTIES` filtering, partition cursor installation,
  `START_TASK_RUN_ID` propagation, priority bump, sync/manual flags).

External behavior verification (5 partitions, `partition_refresh_number=2`):

| Config | `count(*)` from MV | Internal task runs |
| --- | --- | --- |
| sync, `partition_refresh_number=2` | 5 (one call)  | 3 (2 + 2 + 1) |
| sync, `partition_refresh_number=-1` | 5 (one call)  | 1 (no batching) |
| async, `partition_refresh_number=1` | 3 (eventually) | 3 (chained) |

Fixes #71974

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

PR #72033, opened 2026-04-23
[Feature] Add "schema" parameter to FILES() for explicit read schema

### LINKED ISSUE #72032: [Feature] Add "schema" parameter to FILES() for explicit read schema
- [x] https://github.com/StarRocks/starrocks/pull/72033
## Feature request

`FILES()` today relies on schema inference across sampled files. With schema drift — especially in nested types (`STRUCT` / `ARRAY` / `MAP`) — inferred results are unstable and difficult to control. Common pain points:

- A STRUCT whose subfields differ between files may be demoted to `VARCHAR`, breaking nested field access.
- Columns only present in unsampled files disappear from the inferred schema.
- CSV integer columns are inferred as `BIGINT`; downstream `TINYINT` targets get silent `NULL` on out-of-range values because the read-time type is too wide.
- Users have no way to say "read exactly these columns with exactly these types" — the only control knobs are `auto_detect_sample_files` / `auto_detect_sample_rows` / `auto_detect_types`, which influence but don't determine the read schema.

Add a new property `schema` to `FILES()` so users can declare the read schema explicitly and bypass inference.

### Syntax

```sql
FROM FILES(
  "path" = "...",
  "format" = "parquet",
  "schema" = "col_name TYPE[, col_name TYPE ...]"
)
```

Only `name type` pairs are supported. `NULL` / `NOT NULL` / `DEFAULT` / `COMMENT` / `KEY` / `AUTO_INCREMENT` / generated columns are rejected at grammar level. Complex types (`ARRAY`, `MAP`, `STRUCT`, partial-STRUCT) are supported.

### Behavior contract

- Read columns are determined by `schema`.
- Extra columns in files not declared in `schema` are ignored.
- Missing declared columns follow existing `fill_mismatch_column_with` semantics (`none`: fail, `null`: fill NULL).
- Parquet / ORC / Avro: match by name, **case-sensitive**.
- CSV: match by position; schema column names are aliases.
- Partial STRUCT declarations work — undeclared subfields are ignored.

### Scope

In-scope read paths: `SELECT ... FROM FILES(...)`, `INSERT ... SELECT ... FROM FILES(...)`, `CTAS AS SELECT ... FROM FILES(...)`. Read formats: parquet, orc, avro, csv.

Out of scope: `DESC FILES(..., schema=...)` (rejected), `INSERT INTO FILES(..., schema=...)` unload path (rejected).

### Interaction rules

- Mutually exclusive with `auto_detect_sample_files` / `auto_detect_sample_rows` / `auto_detect_types`.
- Name conflict with `columns_from_path` fails directly.
- `fill_mismatch_column_with` governs missing-column handling under schema.
- `list_files_only=true` silently ignores `schema`.
- INSERT `files_enable_insert_push_down_column_type` config-level push-down is silently skipped when `schema` is set.
- INSERT property `enable_push_down_schema=true` combined with `schema` fails validation (explicit conflict).

### Examples

```sql
-- Parquet by name
SELECT user_id, event_time FROM FILES(
  "path" = "s3://bucket/path/*.parquet",
  "format" = "parquet",
  "schema" = "user_id BIGINT, event_time DATETIME"
);

-- CSV by position
SELECT * FROM FILES(
  "path" = "s3://bucket/path/*.csv",
  "format" = "csv",
  "csv.column_separator" = ",",
  "schema" = "a BIGINT, b VARCHAR(64)"
);

-- Partial nested declaration
SELECT request_data.device_data.platform, request_data.now FROM FILES(
  "path" = "s3://bucket/path/*.parquet",
  "format" = "parquet",
  "schema" = "request_data STRUCT<device_data STRUCT<platform VARCHAR(64)>, now BIGINT>"
);
```

banmoy:
## Why I'm doing:

`FILES()` today relies on schema inference across sampled files. With schema drift — especially in nested types (`STRUCT` / `ARRAY` / `MAP`) — inferred results are unstable and difficult to control:

- A STRUCT whose subfields differ between files may be demoted to `VARCHAR`, breaking nested field access.
- Columns only present in unsampled files disappear from the inferred schema.
- CSV integer columns are inferred as `BIGINT`; downstream `TINYINT` targets get silent `NULL` on out-of-range values because the read-time type is too wide.
- Users have no way to say "read exactly these columns with exactly these types" — the only control knobs are `auto_detect_sample_files` / `auto_detect_sample_rows` / `auto_detect_types`, which influence but don't determine the read schema.

## What I'm doing:

Add a new `schema` property to `FILES()` so users can declare the read schema explicitly and bypass inference. Behavior contract:

- Read columns are determined by `schema`; extra file columns are ignored.
- Missing declared columns follow the existing `fill_mismatch_column_with` semantics (`none`: fail, `null`: fill NULL).
- Parquet / ORC / Avro match by name, **case-sensitive**; CSV matches by position (schema names are aliases).
- Partial STRUCT declarations are supported (undeclared subfields ignored).
- Only `name type` pairs are allowed; `NULL` / `NOT NULL` / `DEFAULT` / `COMMENT` / `KEY` / `AUTO_INCREMENT` / generated columns are rejected at grammar level.
- Mutually exclusive with `auto_detect_sample_files` / `auto_detect_sample_rows` / `auto_detect_types`.
- Name conflict with `columns_from_path` fails directly.
- `list_files_only=true` silently ignores `schema`.
- `DESC FILES(..., schema=...)` and `INSERT INTO FILES(..., schema=...)` (unload) are explicitly rejected.
- On INSERT: config-level `files_enable_insert_push_down_column_type` push-down is silently skipped when `schema` is set; `PROPERTIES("enable_push_down_schema"="true")` combined with `schema` fails validation (explicit conflict).

Fixes #72032

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [x] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/InsertAnalyzer.java
```diff
@@ -467,6 +467,15 @@ private static void pushDownTargetTableSchemaToFiles(InsertStmt insertStmt, Conn
         }
 
         Consumer<TableFunctionTable> pushDownSchemaFunc = (fileTable) -> {
+            if (fileTable.hasExplicitSchema()) {
+                if (insertStmt.isEnablePushDownSchema()) {
+                    throw new SemanticException(
+                            "'enable_push_down_schema' on INSERT cannot be combined with " +
+                            "'schema' property on FILES(); the explicit schema already fixes " +
```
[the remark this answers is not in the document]
banmoy: Fixed in `3a7ed0f4346`. The conflict check now lives at the top of `pushDownTargetTableSchemaToFiles` and walks every `FileTableFunctionRelation` under the InsertStmt via `AnalyzerUtils.collectFileTableFunctionRelation`, so the contradiction is rejected regardless of whether `FILES()` sits at the top level, inside a subquery, in a CTE, or behind a join. Two regression tests were added to `InsertAnalyzerFilesSchemaPushDownTest`: `testInsertPushDownSchemaPropertyConflictsThroughSubquery` and `testInsertPushDownSchemaPropertyConflictsThroughCte` — both verify the documented `SemanticException` fires for the wrapped shapes.

docs/en/sql-reference/sql-functions/table-functions/files.md
```diff
@@ -642,6 +642,105 @@ From v3.2 onwards, StarRocks can extract the value of a key/value pair from the
 
 Suppose the data file **file1** is stored under a path in the format of `/geo/country=US/city=LA/`. You can specify the `columns_from_path` parameter as `"columns_from_path" = "country, city"` to extract the geographic information in the file path as the value of columns that are returned. For further instructions, see Example 4.
 
+#### `schema`
+
+From v4.0 onwards, `FILES()` supports an explicit `schema` parameter that lets you declare exactly which columns to read and their StarRocks types, bypassing BE-side schema inference.
```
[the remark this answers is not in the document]
banmoy: Thanks. Keeping the "No" intentionally: this PR introduces a new opt-in property without changing any default behavior. A `FILES()` invocation that does not set `schema` is byte-for-byte identical to the prior release — same inference path, same column types, same error surface. Users only see new behavior when they explicitly set `"schema" = "..."`, in which case the new behavior is precisely what they asked for.

Reading the template literally — "Does this PR entail a change in behavior?" — the answer for upgrading users who do not adopt the new property is no. That matches how other purely additive `FILES()` properties (e.g. `fill_mismatch_column_with`, `list_files_only`) were previously merged.

If the project would prefer "Yes" for any *new opt-in syntax*, please flag and I will flip the box.

wyb (dismissed):

HangyuanLiu (dismissed):

EsoragotoSpirit (dismissed):

HangyuanLiu (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/parser/SqlParser.java
```diff
@@ -256,6 +263,31 @@ public static ImportColumnsStmt parseImportColumns(String expressionSql, long sq
                 .visit(importColumnsContext);
     }
 
+    public static List<Column> parseFilesSchema(String schemaStr) {
```
stephen-shelby, line 274: Could we add semantic validation for types parsed from the FILES `schema` property?

Right now `parseFilesSchema()` appears to call `TypeParser.getType()` and directly build `Column`s. That may accept types normal DDL paths would reject or normalize, e.g. wildcard VARCHAR/CHAR, invalid length/decimal precision, or StarRocks-only metric types like HLL/BITMAP/PERCENTILE. Since the docs say all StarRocks data types are supported, it would be good to either validate with `TypeDefAnalyzer` plus a FILES-readable-type allowlist, or narrow the documented type support. Some negative tests would help too.

stephen-shelby: it's best to add some units to check these invalid cases

banmoy: Good catch, validation added in 2adf42eea63. `parseFilesSchema` now (a) forwards the parsed type through `TypeDefAnalyzer` (same path as `ColumnDefAnalyzer`) so wildcard / out-of-range `CHAR`/`VARCHAR`/`DECIMAL` are rejected the same way as `CREATE TABLE`, and (b) walks the type tree to reject `HLL`/`BITMAP`/`PERCENTILE` at any depth (top level, ARRAY element, MAP key/value, STRUCT field) — these have no representation in Parquet/ORC/Avro/CSV. Bare `VARCHAR` / `CHAR` follow `ColumnDefAnalyzer` and silently default to length 1.

`SqlParserFilesSchemaTest` now covers the negative cases in a table-driven `testRejectsInvalidTypes`: `VARCHAR(100000000)`, `DECIMAL(80, 5)`, `DECIMAL(10, 11)`, top-level `HLL`/`BITMAP`/`PERCENTILE`, `ARRAY<HLL>`, `STRUCT<a INT, b BITMAP>`, `MAP<INT, PERCENTILE>`. Docs (EN/ZH/JA) updated to call out the metric-type rejection.

stephen-shelby (approved):

wyb (approved):

HangyuanLiu (approved):

PR #72055, opened 2026-04-23
[Enhancement] Protect IVM rewrite fallback from plan mutation

Youngwb:
## Why I'm doing:
IVM rewrite currently wraps and rewrites the original query plan directly. When rewrite fails to converge or throws during rule execution, the root may be restored, but parts of the original plan can already have been mutated through shared OptExpression / Operator references. This can leave the fallback path with a partially modified plan instead of the true pre-IVM plan.

## What I'm doing:
Run IVM rewrite on a cloned trialPlan instead of the original plan, and commit the rewritten result only after all IVM markers are eliminated. If rewrite does not converge or throws, restore the original plan and required column sets in finally. Also add a lightweight structure digest guard to detect unexpected mutation leakage into the original plan, plus unit coverage for successful rewrite, non-converged fallback, and exception fallback paths.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

dirtysalt (approved):

PR #72057, opened 2026-04-23
[auto-perf-opt] Parallelize del-file OSS reads in LakePersistentIndex::load_dels

luohaha:
## What type of PR is this?

- [x] improvement
- [ ] feature
- [ ] bug
- [ ] refactor
- [ ] doc

## Which StarRocks version(s) does this PR apply to?

- main

## Summary

On cold-start PK-index publishes on shared-data tables, `LakePersistentIndex::load_dels` is the dominant latency contributor: each rowset with del_files reads + deserializes them strictly serially, so a tablet with N rowsets × M del_files waits `N*M × OSS_read_latency` before any index work can begin.

This PR splits `load_dels` into two phases:

1. **Phase 1 (parallel)** — open, `read_all()` and `ColumnArraySerde::deserialize()` one pk column per del file on the `pk_index_execution` thread pool, using the same `enable_pk_index_parallel_execution` gate and `CONCURRENT` `ThreadPoolToken` pattern already used by `_open_sstables_parallel`. Nothing in this phase touches the pk index.
2. **Phase 2 (sequential)** — iterate the per-file pk columns **in the original order** and run `get()` / `generate_filter` / `replay_erase()` against the index. `PersistentIndexMemtable` keeps its single-writer contract and replay order is preserved exactly.

The parallel phase is skipped when `ndel <= 1` or when the flag is off, so the common single-del-file fast path has no overhead. Submission failures fall back to inline execution, so no new failure mode is introduced.

## Evidence this is the right target

Running the `999_1gb_1_1tb` realtime-benchmark template against a 6-CN shared-data cluster on top of PR #72016, iter-005's publish-version trace on the slowest tablet during the first-write burst showed:

```
txn=2309387  tablet=13991  cost=10.9s
    primary_index_load_latency_us      10.4s   ← whole publish-side load
    pindex_load_from_lake_tablet_us    10.1s   ← load_from_lake_tablet scope
    rebuild_index_del_cost_us          10.0s   ← cumulative load_dels scope
```

The entire 10 s cold-start tail is OSS read + deserialize in `load_dels`, not index mutation. Pool `cloud_native_pk_index_execution` simultaneously hit **~1848 active threads**, so 1848 tablets were all burning their budget on serialized per-rowset del-file OSS reads at the same moment. Across all 6 CNs, `segment_io_remote` + local-disk IO was the binder, not CPU (sysstat showed `usr%` in the 70s, not saturation).

## Expected impact

On cold-start tablets with multiple del files per rowset, `load_from_lake_tablet` latency falls toward `max(OSS_read_latency)` across del files instead of `sum(OSS_read_latency)`. Steady-state tablets that already had resident pk indexes are unaffected (no load_dels call).

Concretely we expect iter-005's whole-run upsert Max (32 786 ms, dominated by cold-start CommitAndPublishTimeMs) to fall, while steady-state P99 (1627 ms) should be within run-to-run noise.

Result will be validated in the next iteration of the optimization run against baseline iter-005.

## Thread safety

Phase 1 only touches disjoint slots in `pkcs[]` / `per_task_status[]` and calls `clone()`, `new_random_access_file()`, `read_all()`, and `ColumnArraySerde::deserialize()` — all safe to invoke concurrently. `pk_column` is a locally-constructed empty template column, and `clone()` does not mutate it. Phase 2 runs on the calling thread, preserving every existing invariant around `get()` / `replay_erase()`.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -480,7 +480,17 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_max_threads, "0");
 // The queue size for pk index memtable flush threadpool in shared-data mode.
 CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // The maximum number of memtables for pk index in shared-data mode.
-CONF_mInt32(pk_index_memtable_max_count, "2");
+// Controls the async-flush pipeline depth in LakePersistentIndex::flush_memtable:
+// while (inactive_memtables.size() + 1 < pk_index_memtable_max_count) the new
+// flush is submitted to the pk_index_memtable_flush pool; otherwise the write
+// path falls back to a synchronous flush that blocks the caller. With the prior
+// value of 2, only one inactive memtable could be in flight, so any hot PK
+// tablet that filled two memtables within a flush cycle (~100-200 ms of SST
+// IO) paid a full sync flush on its write path — a primary contributor to the
+// upsert P99 / Max tail on shared-data. 4 allows 3 pending async flushes,
+// giving bursty writers headroom without unbounded memory growth (each memtable
+// is bounded by l0_max_mem_usage / l0_min_mem_usage and the update memtracker).
+CONF_mInt32(pk_index_memtable_max_count, "4");
 // The maximum wait flush timeout for pk index memtable in shared-data mode, in milliseconds.
 CONF_mInt64(pk_index_memtable_max_wait_flush_timeout_ms, "30000");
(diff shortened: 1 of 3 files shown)


luohaha:
Closing — exploratory PR from an earlier auto-perf-opt run (`my-rt`). Equivalent to my-rt-2's PR #72574 (`Parallelise delete-file reads in LakePersistentIndex::load_dels`), which is the cleaner, validated version on top of the my-rt-2 stack. This older PR is no longer needed.

PR #72068, opened 2026-04-23
[auto-perf-opt] Prefetch SST tail to warm starlet cache before Table::Open

luohaha:
## Context (measured on 999_1gb_1_1tb realtime upsert benchmark, 1 TB shared-data, 6 CN)

`LakePersistentIndex::init` → `PersistentIndexSstable::new_sstable` → `sstable::Table::Open` issues **4 scattered OSS reads** per SST on cold-load:

1. footer (48 bytes, end of file)
2. index block
3. metaindex block
4. filter block (bloom)

All four live in the tail of the file per the LevelDB table layout. They run on the `pk_index_execution` pool (iter-004..008 sizes this at 1024 threads) but the bottleneck is OSS egress, not pool slots: at cold-start on the 1 TB fixture, 300+ tablets init their PK indexes nearly simultaneously, so aggregate per-CN OSS concurrency saturates.

Observed per-publish cost (cost ≥ 5 s publish traces, iter-008 / build 1224 = main + PR #72057):

| Pattern | Scope | iter-008 count (of 44) | Slowest in that pattern |
|---|---|---:|---|
| A | `rebuild_index_segment_cost_us` (segment-scan / local-disk)          | 1  | ~7 s  |
| B | `rebuild_index_del_cost_us` (post-PR #72057: del-file parallel IO)   | 21 | 6.8 s |
| **C** | **`pindex_init_sst_open_us` (SST metadata OSS reads)**         | **20** | **~12 s** |

Pattern B was the old top lever; PR #72057 took it from 13 s max down to 6.8 s max. **Pattern C (12 s) is now the longest remaining cold-start per-publish bar.**

## What this PR does

Before `sstable::Table::Open`, pre-warm the **starlet local disk cache** for the tail of the SST with a single `touch_cache(filesize - tail, tail)`. Default tail = 1 MiB, which comfortably covers a PK-index SST filter of ~1 M keys at 10 bits/key plus the index, metaindex, and footer blocks.

Result: the 4 subsequent scattered reads in `Table::Open` become cache hits instead of 4 separate OSS GETs per SST open. At N tablets × M SSTs/tablet concurrent cold-loads, aggregate OSS requests drop ~4×.

Only `new_sstable` is modified — every call site (`_open_sstables_parallel`, compaction, etc.) inherits the behavior.

## Gating

- `enable_pk_index_sst_tail_prefetch` (default `true`, mutable) — kill switch.
- `pk_index_sst_tail_prefetch_bytes` (default `1048576`, mutable) — tail size.
- No-op on filesystems that do not implement `touch_cache` (fs_s3, fs_posix). Shared-nothing mode therefore sees no change.

## Expected impact

- Cold-start `pindex_init_sst_open_us` p99 should drop from ~12 s toward single-digit seconds.
- Cold-start Upsert Max (which has been CommitAndPublish-dominated) should drop proportionally.
- Upsert P99 already well under the 3 000 ms target (1 651 ms in iter-008) — that P99 may tighten further but the primary target is cold-start Max.

## What this PR does NOT fix

- Steady-state `WriteDataTimeMs` P99 / Max (~5-10 s on iter-008). Different code path.
- Cold-start CN hot-spotting (one CN serialising multiple per-publish waits) — orthogonal, driven by workload placement, not PK-index code.

## Test plan

- [ ] TSP build + deploy to `benchmark_luoyixin_1tb_6cn_2disk2` (in-place over build 1224 which already has PR #72057 merged into iter-006 branch).
- [ ] Re-run 999_1gb_1_1tb upsert workload (`--skip-setup --skip-prepare` to match iter-008 conditions) and compare:
  - [ ] Per-publish `pindex_init_sst_open_us` distribution on cost ≥ 5 s cold-start traces (target: max dropped ≥ 2× vs iter-008's 11-12 s)
  - [ ] Whole-run Upsert P99 / Max
  - [ ] Steady-state P99 / Max (should be unchanged within noise)
  - [ ] Error rate (must remain 0 %)
- [ ] Confirm compile + unit tests green in TSP build.

### DIFF
--- a/be/src/storage/lake/persistent_index_sstable.cpp
+++ b/be/src/storage/lake/persistent_index_sstable.cpp
@@ -259,6 +259,18 @@ StatusOr<PersistentIndexSstableUniquePtr> PersistentIndexSstable::new_sstable(
         opts.encryption_info = std::move(info);
     }
     ASSIGN_OR_RETURN(auto rf, fs::new_random_access_file(opts, location));
+    // sstable::Table::Open will issue 4 scattered reads for the SST tail
+    // (footer / index / metaindex / filter). On starlet, turn them into
+    // cache hits by pre-warming the tail with a single touch_cache. No-op on
+    // local filesystems.
+    if (config::enable_pk_index_sst_tail_prefetch) {
+        const int64_t filesize = sstable_pb.filesize();
+        const int64_t tail_size =
+                std::min<int64_t>(filesize, std::max<int64_t>(0, config::pk_index_sst_tail_prefetch_bytes));
+        if (tail_size > 0) {
+            (void)rf->touch_cache(filesize - tail_size, tail_size);
+        }
+    }
     RETURN_IF_ERROR(sstable->init(std::move(rf), sstable_pb, cache, need_filter, delvec, metadata, tablet_mgr));
     return std::move(sstable);
 }
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -493,6 +493,21 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 CONF_mInt32(pk_index_memtable_max_count, "4");
 // The maximum wait flush timeout for pk index memtable in shared-data mode, in milliseconds.
 CONF_mInt64(pk_index_memtable_max_wait_flush_timeout_ms, "30000");
+// When opening a persistent-index SST for read (LakePersistentIndex::init ->
+// PersistentIndexSstable::new_sstable -> sstable::Table::Open), Table::Open
+// issues 4 small, scattered reads for the tail (footer, index block,
+// metaindex block, filter block). On shared-data clusters at cold-start the
+// per-CN fan-out of those reads saturates OSS egress and drives the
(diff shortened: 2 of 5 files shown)


be/src/storage/lake/persistent_index_sstable.cpp
```diff
@@ -259,6 +259,18 @@ StatusOr<PersistentIndexSstableUniquePtr> PersistentIndexSstable::new_sstable(
...
+    // sstable::Table::Open will issue 4 scattered reads for the SST tail
+    // (footer / index / metaindex / filter). On starlet, turn them into
+    // cache hits by pre-warming the tail with a single touch_cache. No-op on
+    // local filesystems.
+    if (config::enable_pk_index_sst_tail_prefetch) {
+        const int64_t filesize = sstable_pb.filesize();
+        const int64_t tail_size =
+                std::min<int64_t>(filesize, std::max<int64_t>(0, config::pk_index_sst_tail_prefetch_bytes));
+        if (tail_size > 0) {
+            (void)rf->touch_cache(filesize - tail_size, tail_size);
```
kevincai, line 271: block cache will align the read to its block size, read range across the block boundary will be split into two consequent remote IO calls.

PR #72077, opened 2026-04-23
[BugFix] Restore macOS build: jemalloc macros, abseil ABI, shim signatures

vectorsss:
## Why I'm doing:

The macOS BE build broke on current main due to a combination of upstream
refactors and latent Homebrew dependency quirks:

1. **Upstream source moves & signature changes** since the macOS build was
   last verified (util→base refactor, `LookUpNode` now extends `PipelineNode`,
   `HiveTableSink::close` takes a `const Status&`, `LookUpDispatcherMgr`
   dropped `RuntimeState` and gained `lookup_close`, `Connector` grew
   `BENCHMARK` / `CACHE_STATS` enums, `glm_manager` requires
   `lake::Rowset` RTTI, thrift 0.22 ships `TUuid` natively).
2. **Homebrew quirks** that produce cryptic link/runtime errors:
   jemalloc's empty-prefix macros cause bare `je_free` to expand to
   `free` and bind to `JemallocAllocator::free()` (infinite recursion);
   abseil ABI drift between re2's pinned bottle and the currently-linked
   abseil keg causes undefined `_AbslInternalSpinLockWake_lts_<date>`
   at link time.
3. `./build_be.sh` sources `env_macos.sh` before calling
   `build_thirdparty.sh`; exported `CMAKE_OSX_ARCHITECTURES=arm64` leaks
   into nested thirdparty cmake invocations and breaks velocypack's
   `TargetArch.cmake` with "Invalid OS X arch name: arm64".

## What I'm doing:

Restore the macOS BE build: fix jemalloc macro recursion, pin abseil to
the keg re2 was bottled against, stop `build_be.sh` leaking CMake vars
into thirdparty, and re-sync shims with upstream signature changes.

- **jemalloc**: add `JEMALLOC_NO_DEMANGLE` and qualify `::je_free`.
- **abseil**: detect re2's required ABI from `libre2.dylib` and locate
  the matching keg; fatal-error instead of silently linking mismatched
  `lts_*` symbols.
- **build_thirdparty.sh**: unset leaked `CMAKE_OSX_ARCHITECTURES` (and
  friends) at entry so `build_be.sh` matches standalone invocation.
- **shims**: track upstream signature changes (`LookUpNode → PipelineNode`,
  `HiveTableSink::close`, `lookup_close`, `Connector` enums); add `base/`
  and `udf/` source globs (util→base refactor); drop `thrift_uuid_stub`
  now that thrift 0.22 ships `TUuid` natively.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Verification

Verified locally on Apple Silicon via `./build-mac/build_thirdparty.sh &&
./build-mac/build_be.sh`:
- MacBook Pro (M1 Max, macOS 26.4.1)
- MacBook Pro (M4, macOS 26.4.1)

## Checklist:

- [ ] This is a backport pr

> N/A — build-only change for macOS (not a CI-gated platform).
> No unit-testable runtime behavior, no user-facing feature, not a backport.

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

> N/A — macOS build only exists on main, and several fixes (util→base
> refactor, LookUpNode/PipelineNode split) don't apply to release branches.

### DIFF
--- a/be/src/exec/file_scanner/json_scanner.cpp
+++ b/be/src/exec/file_scanner/json_scanner.cpp
@@ -589,7 +589,7 @@ Status JsonReader::_construct_row_without_jsonpath(simdjson::ondemand::object* r
                 DCHECK(_envelope_type != TEnvelopeType::DEBEZIUM || _cdc_op != 0xFF);
                 uint8_t op_val = _envelope_type != TEnvelopeType::NONE ? _cdc_op : 0;
                 if (column->is_binary()) {
-                    column->append_strings(&kCdcOpSlices[op_val], 1);
+                    (void)column->append_strings(&kCdcOpSlices[op_val], 1);
                 } else {
                     column->append_datum(Datum(op_val));
                 }
@@ -619,7 +619,7 @@ Status JsonReader::_construct_row_with_jsonpath(simdjson::ondemand::object* row,
                 DCHECK(_envelope_type != TEnvelopeType::DEBEZIUM || _cdc_op != 0xFF);
                 uint8_t op_val = _envelope_type != TEnvelopeType::NONE ? _cdc_op : 0;
                 if (column->is_binary()) {
-                    column->append_strings(&kCdcOpSlices[op_val], 1);
+                    (void)column->append_strings(&kCdcOpSlices[op_val], 1);
                 } else {
                     column->append_datum(Datum(op_val));
                 }
@@ -654,7 +654,7 @@ Status JsonReader::_construct_row_with_jsonpath(simdjson::ondemand::object* row,
                     DCHECK(_envelope_type != TEnvelopeType::DEBEZIUM || _cdc_op != 0xFF);
                     uint8_t op_val = _envelope_type != TEnvelopeType::NONE ? _cdc_op : 0;
                     if (column->is_binary()) {
-                        column->append_strings(&kCdcOpSlices[op_val], 1);
+                        (void)column->append_strings(&kCdcOpSlices[op_val], 1);
                     } else {
                         column->append_datum(Datum(op_val));
                     }
--- a/be/src/formats/parquet/complex_column_reader.cpp
+++ b/be/src/formats/parquet/complex_column_reader.cpp
@@ -1304,18 +1304,6 @@ StatusOr<std::optional<VariantRowValue>> VariantColumnReader::build_variant_bind
     return std::optional<VariantRowValue>(std::move(built));
 }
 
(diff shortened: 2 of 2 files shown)


alvin-phoenix-ai:
@vectorsss 
Thanks for the contribution.
`./build-mac` is the method that we want to deprecate; the recommended way is to build BE from the root directory `./build.sh --be`. 
Could you try this way and to see if it works?

vectorsss:
> @vectorsss Thanks for the contribution. `./build-mac` is the method that we want to deprecate; the recommended way is to build BE from the root directory `./build.sh --be`. Could you try this way and to see if it works?

it not work yet, need to fix. I am trying to fix it right now.

vectorsss:
> @vectorsss Thanks for the contribution. `./build-mac` is the method that we want to deprecate; the recommended way is to build BE from the root directory `./build.sh --be`. Could you try this way and to see if it works?

Hi, I’ve fixed the issue with `./build.sh --be`, and everything is working fine now. The `./build_be.sh` script in the `build-mac` directory also works fine.

alvin-phoenix-ai:
@vectorsss 
This is great that everything works fine.
But it is hard to review to have all the stuff in one single PR. 
Could you try to split this PR?
1. I see that you changed the build-third-party.It will influence the compilation on other platforms like Ubuntu. This should be submitted separately.
2. You should separate the changes for `./build.sh --be` from this PR and submit them in an separate PR
3. And the change for build-mac is not necessary, because we want to deprecate it someday in the future.

vectorsss:
> @vectorsss This is great that everything works fine. But it is hard to review to have all the stuff in one single PR. Could you try to split this PR?
> 
> 1. I see that you changed the build-third-party.It will influence the compilation on other platforms like Ubuntu. This should be submitted separately.
> 2. You should separate the changes for `./build.sh --be` from this PR and submit them in an separate PR
> 3. And the change for build-mac is not necessary, because we want to deprecate it someday in the future.

Sure, I will do it soon

vectorsss:
Split as suggested:
- #72157 — thrift `--without-cl` (cross-platform)
- #72158 — `./build.sh --be` macOS support

This PR keeps the original scope.

vectorsss:
Re-opened as #72182 (this PR auto-closed after I accidentally synced upstream to my fork's main).

PR #72108, opened 2026-04-24
[BugFix] Relax StarMgrMetaSyncer DB locks that stall DDL over StarOS RPCs

kevincai:
syncTableMetaInternal and syncTableColocationInfo ran periodically while holding a DB-wide READ or WRITE lock across external StarOS RPCs (listShard, updateMetaGroup). For the duration of each RPC the whole database was effectively frozen: concurrent CREATE / DROP / ALTER / RENAME on any other table in the same DB blocked waiting for the syncer, and any RPC stall extended that window arbitrarily.

The critical sections only need table-scoped consistency on a single known table: syncTableMetaInternal walks one table's partitions and flips shardGroupChanged; syncTableColocationInfo reads one table's shard-group IDs and forwards them to StarOS under ColocateTableIndex's own internal writeLock. Drop / rename of this DB or table still takes DB WRITE, which conflicts with IS, so the manual getDb / getTable == null re-check still catches drops that committed before the lock was acquired.

Switch all three sites to the intensive-lock path already used by getAllPartitionShardGroupId in the same file:

  syncTableMetaInternal   : DB READ  -> IS + table READ
  syncTableColocationInfo : DB WRITE -> IS + table READ

syncTableColocationInfo drops to table READ rather than table WRITE because the call is semantically a read of the OlapTable's shard groups plus a mutation of external StarOS state and ColocateTableIndex's own maps (the latter guarded by its internal writeLock), not a mutation of the Java Table object.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/StarMgrMetaSyncer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/StarMgrMetaSyncer.java
@@ -542,7 +542,12 @@ public class StarMgrMetaSyncer extends FrontendDaemon {
         HashMap<Long, Set<Long>> redundantGroupToShards = new HashMap<>();
         List<PhysicalPartition> physicalPartitions = new ArrayList<>();
         Locker locker = new Locker();
-        locker.lockDatabase(db.getId(), LockType.READ);
+        // Intensive path: IS on DB + READ on this table. We only need table-scoped
+        // consistency here (partition walk on one known table plus a flag flip);
+        // DROP TABLE / DROP DATABASE both take DB WRITE which still conflict with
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

starrocks-xupeng (approved):

xiangguangyxg (approved):

gengjun-git (approved):

PR #72129, opened 2026-04-24
[BugFix] Distinguish bundled segment slices in page/segment caches

xiangguangyxg:
## Why I'm doing:

Multiple caches in shared-data key on `(filename, offset)` or `filename`
without a slice disambiguator. When a physical `.dat` file is written in
**bundle format** (a single file holding per-tablet slices at different
offsets), two slices of the same file can therefore collide on the same
cache key, and a lookup for slice B returns slice A's cached bytes —
slice A silently duplicates in query results, slice B vanishes.

### The data-loss path: storage page cache

`be/src/storage/rowset/page_io.cpp::read_and_decompress_page` keyed the
page cache on `encode_cache_key(read_file->filename(), page_pointer.offset)`.
For a `BundleSeekableInputStream`, `page_pointer.offset` is
**stream-relative** (0-based within the slice), so two slices of the same
bundled file produce identical cache keys at equal intra-slice positions.

Reproducer on a shared-data primary-key range-distributed table:

```
INSERT  0..20000     salt=0
UPSERT  5000..15000  salt=1
ALTER TABLE ... SPLIT TABLETS (...)                          -- four children
UPSERT  10000..25000 salt=2     -- two batches straddle child boundaries
                                -- → written as bundled files
ALTER TABLE ... MERGE TABLETS (...)                          -- one tablet again
```

After the merge, `SELECT COUNT(DISTINCT id)` returns 22,998 instead of
25,000. 1,998 ids appear twice (slice-A rows read twice), 2,002 ids
disappear entirely (slice-B rows replaced by slice-A's cached pages). The
missing-id ranges are exactly the offset-1 slice contents of the two
bundled segments.

Runtime verification: setting `disable_storage_page_cache=true` on all CNs
and repeating the same flow returns 25000/25000/25000 — confirming the
page cache is the sole source of the corruption.

### Two adjacent defense-in-depth fixes (same mechanism)

1. **Lake Segment metacache.** `TabletManager::load_segment` keyed on
   `segment_info.path` alone. Two rowsets of a bundled file in one tablet
   would share a `Segment` object whose `_segment_file_info` was bound to
   whichever slice loaded first. In isolation this does not cause the
   symptom above — the page cache collision is what corrupts reads — but
   it is the same fragility, and any future read path that bypasses the
   page cache would surface the bug.
2. **Segment cache-size refresh.** `Segment::update_cache_size` and
   `turn_off_batch_update_cache_size` still passed `file_name()` (path
   only) to `update_segment_cache_size` after the Segment-cache key was
   forked. For segments with non-zero `bundle_file_offset` the lookup in
   `cache_segment_if_present` would miss, leaving the cache entry's
   memory cost unrefreshed.

### Audit of other `(filename, offset)`-style caches

Every adjacent cache was reviewed for the same class of bug; none are
affected:

| Cache | File | Why it is safe |
|---|---|---|
| Parquet page cache | `formats/parquet/page_reader.cpp` | External-table only; parquet writers do not use `BundleWritableFile`. |
| HDFS/Iceberg `CacheInputStream` block cache | `io/cache_input_stream.cpp` | External-table only; lake-native never wraps a bundle stream in this layer; key already includes `modification_time`. |
| SST block cache | `storage/sstable/table.cpp` | `(cache_id, handle.offset)` with per-Table `cache_id` from `block_cache->new_id()`; SST files are not written via `BundleWritableFile`. |
| Delvec cache | `lake/meta_file.cpp::delvec_cache_key` | `(tablet_id, DelvecPagePB)`; delvec files are not bundled; `DelvecPagePB` embeds version + offset + size. |
| Legacy rowset metadata cache | `rowset/metadata_cache.cpp` | Keyed by `rowset_id_str`; no file offset. |
| Tablet metadata cache | `lake/tablet_manager.cpp` | Keyed by `tablet_metadata_location(tablet_id, version)`. |

The only writer that produces bundles is `BundleWritableFileContext`, wired
only into `DeltaWriter` / `GeneralTabletWriter` / `PkTabletWriter` (data
segments); `PkTabletSSTWriter` uses `fs->new_writable_file` directly, so
sstables are never bundled.

## What I'm doing:

Encapsulate the slice disambiguator on the types that carry it, instead
of leaving each cache call site to know about BundleFile. Three caches
share one principle: **the cache key must name the page/segment's
position in the physical file, not in the slice.**

### 1. Page cache — `be/src/storage/rowset/page_io.cpp`

- New virtual `io::SeekableInputStream::page_cache_key(int64_t stream_offset)`.
  Default encodes `(filename, stream_offset)` (the existing binary format,
  moved out of `page_io.cpp`).
- `BundleSeekableInputStream` overrides it to fold its slice base offset
  into the encoding, returning a key unique to the page's absolute byte
  position in the physical file.
- `SeekableInputStreamWrapper` forwards to `_impl`; all wrappers
  (`CacheInputStream`, `CountedSeekableInputStream`,
  `ThrottledSeekableInputStream`, `EncryptSeekableInputStream`) inherit
  it automatically.
- `read_and_decompress_page` calls
  `read_file->page_cache_key(opts.page_pointer.offset)` and never names
  `bundle_file_offset`.

### 2. Lake Segment metacache — `be/src/fs/fs.h`, `be/src/storage/lake/tablet_manager.cpp`

- New `FileInfo::cache_key()` returns `"<path>#<offset>"` for bundled
  slices (`bundle_file_offset > 0`) and the bare `path` otherwise.
  Non-bundled layout unchanged.
- `TabletManager::load_segment` keys both `lookup_segment` and
  `cache_segment_if_absent` on `segment_info.cache_key()`.

### 3. Segment cache-size refresh — `be/src/storage/rowset/segment.cpp`

- `Segment::update_cache_size` and `turn_off_batch_update_cache_size`
  route through `_segment_file_info.cache_key()`, so the insert key (set
  by `load_segment`) and the update-size key stay identical for bundled
  slices by construction.

The previous public `SeekableInputStream::bundle_file_offset()` accessor
and the static `TabletManager::segment_cache_key()` helper are removed;
the BundleFile detail only exists inside
`BundleSeekableInputStream::page_cache_key` and `FileInfo::cache_key`.
Future caches keyed on stream/file identity get correct bundled-slice
handling without any new conditionals.

## Tests

New unit tests:

- `PageIOTest.test_bundle_slices_do_not_collide_in_page_cache` writes two
  distinct pages back-to-back into one physical file, wraps each with its
  own `BundleSeekableInputStream` at offsets 0 and `slice_a_size`, warms
  the cache on slice A, then reads slice B at the same stream-relative
  offset 0 and asserts (1) cache miss and (2) the decoded bytes are slice
  B's, not A's.
- `LakeTabletManagerTest.segment_cache_key_distinguishes_bundle_slices`
  covers the `FileInfo::cache_key()` rule directly.
- `LakeTabletManagerTest.load_segment_returns_distinct_slices_for_same_path`
  seeds the metacache at one offset and confirms the lookup at another
  does not collide.

Existing `PageIOTest` and `LakeTabletManagerTest` suites remain green
(6/6 and 26/26 respectively).

End-to-end verification on a shared-data test cluster with the reproducer
above: the post-merge `SELECT COUNT(DISTINCT id)` returns 25,000 with all
expected per-id values, where the same recipe previously returned 22,998
with mixed slice-A / slice-B corruption.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/tablet_manager.cpp
+++ b/be/src/storage/lake/tablet_manager.cpp
@@ -1348,7 +1348,15 @@ StatusOr<SegmentPtr> TabletManager::load_segment(const FileInfo& segment_info, i
     //       for example, in tablet X, segment `a` has segment id 10, if partial compaction happens,
     //                    in tablet X+1, segment `a` might still exists, but its actual id will not be 10.
     //       but in meta cache, segment `a` still has segment id 10, it is not changed.
-    auto segment = metacache()->lookup_segment(segment_info.path);
+    //
+    // A bundled data file can be referenced by multiple rowsets at different byte ranges
+    // (distinct bundle_file_offsets). The ranges contain different rows, so each must produce
+    // its own Segment bound to its own slice. Keying the cache on path alone would let the
+    // first-loaded slice be returned for every other offset, silently replacing later slices'
+    // data with the first one. FileInfo::cache_key() folds bundle_file_offset into the key
+    // for bundled slices and falls back to path for non-bundled segments.
+    const std::string cache_key = segment_info.cache_key();
+    auto segment = metacache()->lookup_segment(cache_key);
     if (segment == nullptr) {
         std::shared_ptr<FileSystem> fs;
         if (segment_info.fs) {
@@ -1360,7 +1368,7 @@ StatusOr<SegmentPtr> TabletManager::load_segment(const FileInfo& segment_info, i
         if (fill_meta_cache) {
             // NOTE: the returned segment may be not the same as the parameter passed in
             // Use the one in cache if the same key already exists
-            if (auto cached_segment = _metacache->cache_segment_if_absent(segment_info.path, segment);
+            if (auto cached_segment = _metacache->cache_segment_if_absent(cache_key, segment);
                 cached_segment != nullptr) {
                 segment = cached_segment;
             }
--- a/be/src/storage/rowset/page_io.cpp
+++ b/be/src/storage/rowset/page_io.cpp
@@ -140,20 +140,6 @@ Status PageIO::write_page(WritableFile* wfile, const std::vector<Slice>& body, c
     return Status::OK();
 }
 
-// The unique key identifying entries in the page cache.
-// Each cached page corresponds to a specific offset within
-// a file.
-//
-// TODO(zc): Now we use file name(std::string) as a part of key,
-//  which is not efficient. We should make it better later
-std::string encode_cache_key(const std::string& fname, int64_t offset) {
-    std::string str;
-    str.reserve(fname.size() + sizeof(offset));
-    str.append(fname);
-    str.append((char*)&offset, sizeof(offset));
-    return str;
-}
-
 #if defined(USE_STAROS) && !defined(BUILD_FORMAT_LIB)
 Status drop_local_cache_data(const std::string& fname) {
     if (!config::lake_clear_corrupted_cache_data) {
@@ -326,7 +312,11 @@ static Status read_and_decompress_page_internal(const PageReadOptions& opts, Pag
     bool page_cache_available = (cache != nullptr) && cache->available();
 #endif
     PageCacheHandle cache_handle;
-    std::string cache_key = encode_cache_key(opts.read_file->filename(), opts.page_pointer.offset);
+    // Let the stream produce its own page-cache key for opts.page_pointer.offset. For non-bundled
+    // streams this is just (filename, offset); for bundle slices the override folds in the slice
+    // base offset so two slices of the same physical file never collide at equal stream-relative
+    // offsets. Callers do not need to know whether the stream is a slice.
+    std::string cache_key = opts.read_file->page_cache_key(opts.page_pointer.offset);
     if (opts.use_page_cache && page_cache_available && cache->lookup(cache_key, &cache_handle)) {
         return parse_page_from_cache(handle, body, footer, &cache_handle, opts);
     }
--- a/be/src/fs/bundle_file.cpp
+++ b/be/src/fs/bundle_file.cpp
@@ -149,4 +149,12 @@ StatusOr<int64_t> BundleSeekableInputStream::read(void* data, int64_t count) {
     return _stream->read(data, count);
 }
 
+std::string BundleSeekableInputStream::page_cache_key(int64_t stream_offset) const {
+    // Fold the slice base offset into the offset and delegate to the base's default impl. The
+    // default uses virtual filename(), which dispatches back to this slice's filename()
+    // (which forwards to _stream->filename()) — so the key is encoded against the underlying
+    // physical file's name, not against _stream's possibly-empty _filename member.
+    return SeekableInputStream::page_cache_key(_offset + stream_offset);
+}
+
 } // namespace starrocks
--- a/be/src/fs/bundle_file.h
+++ b/be/src/fs/bundle_file.h
@@ -112,6 +112,11 @@ public:
     Status touch_cache(int64_t offset, size_t length) override;
     StatusOr<std::unique_ptr<io::NumericStatistics>> get_numeric_statistics() override;
 
+    // Override so two slices of the same physical file produce distinct keys at the same
+    // stream-relative offset: we fold the slice base offset into the key, yielding a key
+    // that names the page's absolute position in the physical file.
+    std::string page_cache_key(int64_t stream_offset) const override;
+
 private:
     std::shared_ptr<io::SeekableInputStream> _stream;
     int64_t _offset = 0;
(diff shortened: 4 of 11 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
Addressed in 8ef5cdb: routed `Segment::update_cache_size` and `turn_off_batch_update_cache_size` through `TabletManager::segment_cache_key(_segment_file_info)` so the insert key and the update-size key stay identical for bundled slices. LakeTabletManagerTest suite (26/26) still passes locally.

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -1348,7 +1355,14 @@ StatusOr<SegmentPtr> TabletManager::load_segment(const FileInfo& segment_info, i
...
     //       but in meta cache, segment `a` still has segment id 10, it is not changed.
-    auto segment = metacache()->lookup_segment(segment_info.path);
+    //
+    // A bundled data file can be referenced by multiple rowsets at different byte ranges
+    // (distinct bundle_file_offsets). The ranges contain different rows, so each must produce
+    // its own Segment bound to its own slice. Keying the cache on path alone would let the
+    // first-loaded slice be returned for every other offset, silently replacing later slices'
+    // data with the first one. Include bundle_file_offset in the key to keep slices distinct.
+    const std::string cache_key = segment_cache_key(segment_info);
+    auto segment = metacache()->lookup_segment(cache_key);
```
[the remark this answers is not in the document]
xiangguangyxg: fixed

be/src/storage/rowset/page_io.cpp
```diff
@@ -323,7 +323,13 @@ static Status read_and_decompress_page_internal(const PageReadOptions& opts, Pag
     bool page_cache_available = (cache != nullptr) && cache->available();
 #endif
     PageCacheHandle cache_handle;
-    std::string cache_key = encode_cache_key(opts.read_file->filename(), opts.page_pointer.offset);
+    // For bundled-format segments, opts.page_pointer.offset is stream-relative (0-based within
+    // the slice). Two slices of the same physical file therefore produce identical
+    // (filename, page_pointer.offset) keys even though the pages live at different absolute
+    // byte ranges. Mix in the stream's bundle_file_offset so the key names the page's unique
+    // position in the physical file; non-bundled streams return 0 and the key is unchanged.
+    std::string cache_key = encode_cache_key(opts.read_file->filename(),
```
sevev: Is it possible to have a scenario where bundle_offset = a, page_offset = b and bundle_offset = b, page_offset = a,resulting in the same sum of bundle_offset + page_offset for two different slices?

xiangguangyxg: page_pointer.offset is stream-relative (0-based within the slice), so bundle_file_offset + page_pointer.offset is exactly the page's absolute byte position in the physical file. The key is therefore (filename, absolute_offset_in_file), not an opaque sum of two independent integers.

kevincai: Although this implementation solves the issue, but it exposes implementation details of bundleFile and pollute the code everywhere, and still can't prevent future break if someone doesn't aware or forget about the awesome of BundleFile.

The fundamental problem is that the filePath() can't not be unique any more due to the BundleFile implementation, so a new UniqueId() shall be provided instead of the FileName or FilePath, use this uniqueId to do cache key or whatever calculation. Each File/Stream itself will be responsible for producing an unique id that can identify itself uniquely.

xiangguangyxg: > Each File/Stream itself will be responsible for producing an unique id that can identify itself uniquely.

Agreed — pushed in 855beea4e. The "+offset" rule no longer lives in any cache call site; instead each owner answers for its own identity:

- **`io::SeekableInputStream::page_cache_key(stream_offset)`** — virtual on the base class, default encodes `(filename, stream_offset)`. `BundleSeekableInputStream` overrides it to fold in the slice base offset, returning a key unique to the page's absolute position in the physical file. `page_io.cpp` just calls `read_file->page_cache_key(opts.page_pointer.offset)` and never names `bundle_file_offset`.
- **`FileInfo::cache_key()`** — returns `"path#offset"` for bundled slices, bare `path` otherwise. `lake::TabletManager::load_segment` and `Segment::update_cache_size` both go through it, so the insert key and the update-size key share one expression by construction.

The previous static `TabletManager::segment_cache_key()` helper and the public `SeekableInputStream::bundle_file_offset()` accessor are both removed. Future caches keyed on stream/file identity get correct bundled-slice handling without any new conditionals — the BundleFile detail only exists inside `BundleSeekableInputStream::page_cache_key` and `FileInfo::cache_key`.

8/8 targeted UTs pass locally (`PageIOTest.*`, `LakeTabletManagerTest.segment_cache_key_distinguishes_bundle_slices`, `LakeTabletManagerTest.load_segment_returns_distinct_slices_for_same_path`).

kevincai (dismissed):

wyb (dismissed):

kevincai (approved):

meegoo (approved):

PR #72141, opened 2026-04-24
[Enhancement] Protect pipe-internal tasks from DROP TASK

EdwardArchive:
## Why I'm doing

Addressing part of issue #71294: `DROP TASK` has no safeguard against deleting Pipe-internal tasks (named `pipe-<pipe>-task-<id>-<seq>`). Any authenticated user can delete them today, silently breaking ingestion while the Pipe keeps reporting `RUNNING`. Materialized-view internal tasks already have an analogous guard in `DDLStmtExecutor.visitDropTaskStmt`; pipe tasks were never retrofitted after `TaskSource.PIPE` was introduced.

## What I'm doing

Extending the existing source guard in `DDLStmtExecutor.visitDropTaskStmt` so `DROP TASK` is rejected for `TaskSource.PIPE` unless `FORCE` is specified — exactly the same pattern used for `TaskSource.MV`.

Before:
```
mysql> DROP TASK `pipe-p1-task-42-0`;
Query OK  -- silently breaks the pipe's ingestion
```

After:
```
mysql> DROP TASK `pipe-p1-task-42-0`;
ERROR 1064 (HY000): Can not drop task generated by pipe. Use DROP PIPE to remove the pipe, or DROP TASK ... FORCE to delete this task directly.

mysql> DROP TASK `pipe-p1-task-42-0` FORCE;
Query OK  -- escape hatch preserved for orphan cleanup (same as MV)
```

Internal cleanup paths (`PipeTaskDesc.interrupt`, `Pipe.suspend` → `TaskManager.dropTasks`) call the drop API directly and bypass this guard, so they keep working unchanged.

### Scope

This PR covers only the "delete protection" sliver of #71294. The remaining items — `AuthorizerStmtVisitor.visitDropTaskStmt` for actual privilege enforcement, making `TASK` a grantable privilege object, and filtering `information_schema.tasks` by owner — are out of scope and will be split into follow-up PRs. Until the privilege PR lands, `DROP TASK ... FORCE` is still callable by any authenticated user (matches existing MV behavior).

`DROP TASK <pipe-task>` without `FORCE` previously succeeded; it now throws an error. `DROP TASK ... FORCE` and all internal cleanup paths are unchanged.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [x] Interface/UI changes: new error message / rejection for `DROP TASK` on pipe-internal tasks
- [ ] Parameter changes
- [ ] Policy changes
- [ ] Feature removed
- [ ] Miscellaneous

## Checklist:

  - `SubmitTaskStmtTest#testDropPipeTaskProtected` covers: regular drop blocked, `IF EXISTS` still blocked when the task is pipe-sourced, `FORCE` succeeds.
- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Test plan

- [x] `mvn checkstyle:check -pl fe-core` — clean
- [x] Added `SubmitTaskStmtTest#testDropPipeTaskProtected`
- [x] Manual smoke against a running cluster: `CREATE PIPE`, attempt `DROP TASK` on the `pipe-...` task, confirm rejection, verify `DROP TASK ... FORCE` and `DROP PIPE` both work
image

Refs StarRocks/starrocks#71294.


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/qe/DDLStmtExecutor.java
```diff
@@ -1141,17 +1141,24 @@ public ShowResultSet visitSubmitTaskStatement(SubmitTaskStmt stmt, ConnectContex
...
-                throw new RuntimeException("Can not drop task generated by materialized view. You can use " +
-                        "DROP MATERIALIZED VIEW to drop task, when the materialized view is deleted, " +
-                        "the related task will be deleted automatically.");
+            if (!dropTaskStmt.isForce()) {
+                if (task.getSource() == Constants.TaskSource.MV) {
+                    throw new RuntimeException("Can not drop task generated by materialized view. You can use " +
+                            "DROP MATERIALIZED VIEW to drop task, when the materialized view is deleted, " +
+                            "the related task will be deleted automatically.");
+                }
+                if (task.getSource() == Constants.TaskSource.PIPE) {
```
kevincai, line 1157: can the Pipe be recovered automatically detecting the gone of the Task and recreate it for next round? If not, this task shall not be dropped even with `FORCE` option.

EdwardArchive: Hi @kevincai 
From what I could trace, the pipe does seem to auto-recover via future.completeExceptionally → recordTaskError → retry with a new task name (Pipe.java:516-553), but only up to FAILED_TASK_THRESHOLD = 5 

(PipeTaskDesc.java:106) — beyond that the pipe itself moves to State.ERROR and needs a manual ALTER PIPE RETRY/RESUME, if I'm reading it right.

Data correctness looks fine since StmtExecutor aborts the INSERT txn on kill (StmtExecutor.java:3598-3631), so no partial commit or duplicate as far as I can tell.

Given that FORCE would still go through the same kill path and consume the same retry budget, would it make sense to gate FORCE on pipe-internal tasks to users with privilege on the parent pipe rather than treat it as a generic escape hatch? 

Happy to be corrected if I'm missing something.

kevincai: If that is the case, pipe tasks can be dropped by `DROP TASK` with `FORCE` option.

kevincai (approved):

stdpain (approved):

fe/fe-core/src/main/java/com/starrocks/qe/DDLStmtExecutor.java
EdwardArchive: Thanks @kevincai so I plain to working on https://github.com/StarRocks/starrocks/pull/72421 this PR for the only the task maker and root can be use DROP FORCE operation

Issue #1278, opened 2021-11-15
some compound predicate could be normalized


stdpain:
## Feature request
example:
if expr was root expr and expr was a predicate, then
```
    OR
  /   \
NULL a = val
```
could rewrite to 
```
a = val
```

PR #72167, opened 2026-04-25
[auto-perf-opt] PR-2/4: pk-index snapshot proto + on-disk format + I/O helpers

luohaha:
## Summary

PR-2 of a 4-PR series introducing a local-disk snapshot of `LakePersistentIndex` so a cloud-native PK index can skip the full cold rebuild on every BE restart, memory-pressure eviction, and tablet rebalance.

This PR ships only the wire format and I/O plumbing — **behaviour-neutral by construction, no caller invokes the new helpers**. PR-3 will implement the stubs introduced in PR-1 against this format.

**Stacked on top of PR #72166 (PR-1/4)**. Until #72166 merges, the diff below shows 2 commits; after #72166 merges this PR rebases to a single commit.

## Why

Tracing on PL2 hardware in our shared-data benchmark (1 TB / 999-table workload) shows `LakePersistentIndex::load_from_lake_tablet` dominates upsert tail latency for any cold cache scenario:

- `pindex_init_sst_open_us` ~50% of cold rebuild
- `rebuild_index_segment_cost_us` ~30%
- `rebuild_index_del_cost_us` ~20%

All three sub-phases are fundamentally fetch-and-replay-from-OSS. A locally-persisted snapshot of the post-rebuild semantic state (memtable key→value map + SST fileset references) lets us short-circuit the entire path with a single local-disk read. PR-1 (#72166) added the API contract and the integration point; PR-2 adds the format on which that API will operate.

### 1. Proto messages — `gensrc/proto/persistent_index.proto`

Three new top-level messages, additive only — existing readers / writers are unaffected:

- `LakePersistentIndexSnapshotEntryPB` — one memtable `(key, packed-rssid:rowid value, version)`
- `LakePersistentIndexSnapshotFilesetRefPB` — SST fileset reference `(version, sst_filenames, max_rss_rowid)` — re-opened from OSS at restore time
- `LakePersistentIndexSnapshotMetaPB` — top-level: `format_version`, `tablet_id`, `captured_version`, `schema_id`, `captured_at_unix_sec`, `memtable_entries`, `filesets`

### 2. On-disk framing — `be/src/storage/lake/lake_persistent_index_snapshot.h`

Small fixed-size header followed by the proto blob followed by a CRC:

```
offset  size  field
------  ----  ------------------------------------------------------------
     0    8   magic = "SRPISNP1"
     8    4   format_version (current = 1)
    12    4   meta_pb_size
    16    N   serialised LakePersistentIndexSnapshotMetaPB
16 + N    4   crc32c over bytes [0, 16 + N)
```

Readers must verify magic, `format_version <= kSnapshotFormatVersion`, declared size, and the CRC before parsing the proto. Any mismatch yields a `Corruption` status; callers fall back to the cold-rebuild path on any error.

### 3. I/O helpers — `lake_persistent_index_snapshot.cpp`

- `Status write_lake_persistent_index_snapshot(path, meta)` — serialise → CRC → 3-iov `appendv` → `sync` → `close`. Truncates an existing file. Caps payload at 4 GiB (proto serialised size).
- `Status read_lake_persistent_index_snapshot(path, meta*)` — open → header read → magic / version / size check → payload read → CRC verify → `ParseFromString`.

Uses the existing project utilities — `base/coding.h` for fixed32 LE encoding/decoding, `base/hash/crc32c.h` for the checksum, `fs/fs.h` for the filesystem abstractions.

### 4. CMakeLists registration

One line in `be/src/storage/CMakeLists.txt` to compile the new translation unit.

## What this PR does NOT do

- **No call sites**. The PR-1 stubs (`try_serialize_to_local_snapshot` / `try_restore_from_local_snapshot` in `LakePersistentIndex`) are untouched.
- **No `_memtable` / `_sstable_filesets` serialisation logic** — that's PR-3.
- **No validity rules** (tablet-version check, schema-hash check) — PR-3.
- **No unit tests for round-trip equivalence of the in-memory state** — PR-3 (when there is real serialisation logic to round-trip).
- **No eviction / shutdown capture hooks** — PR-4.
- **No GC of stale snapshots on disk** — PR-4.

A round-trip test of the framing alone (write → read of a pre-built proto) would be cheap to add; deferred to PR-3 to keep this PR strictly format-only.

## Risk

Behaviour-neutral by construction — no caller invokes the helpers. The proto additions are append-only and use new top-level message names, so existing serialisations of `persistent_index.proto` are unaffected. The on-disk format reserves a `format_version` field so future incompatible changes (e.g. streaming entries instead of a single proto blob) can fail gracefully on old readers.

## Test plan

- [x] Compiles on TSP CI (verified via TSP build #1278 against the PR-1 branch tip; PR-2 stacks on top).
- [ ] PR-3 will exercise the helpers with real `_memtable` / `_sstable_filesets` content; round-trip equivalence test lands there.
- [ ] PR-4 will exercise the end-to-end flow under benchmark cold-restart on the PL2 cluster.


luohaha:

PR #72169, opened 2026-04-25
[auto-perf-opt] PR-3/4: pk-index snapshot save/restore + validity rules + UTs

luohaha:
## Summary

Stacked on #72167 (PR-2). Wires up the stub `try_serialize_to_local_snapshot` and `try_restore_from_local_snapshot` declared in #72166 (PR-1) with real implementations that round-trip the in-memory `LakePersistentIndex` state through the on-disk framing introduced in PR-2. Still gated by the default-off `config::enable_pk_index_snapshot_persistence`, so runtime behaviour is unchanged unless an operator explicitly opts in.

## Why

`LakePersistentIndex::load_from_lake_tablet` is the dominant cold-rebuild path on shared-data PK-tables. On this run's 999_1gb_1_1tb workload it costs ~1.5 s of OSS SST opens (`pindex_init_sst_open`) + ~1 s of OSS del-file load (`rebuild_index_del`) + up to ~7 s of local-disk segment scan (`rebuild_index_segment`) per cold tablet load. Persisting a local-disk snapshot of the post-rebuild state lets BE-restart, memory-pressure eviction, and tablet-rebalance scenarios skip the rowset-scan + load_dels stage entirely — those are the parts that scale with rowset count, not with tablet metadata.

This is PR-3 of the 4-PR snapshot-persistence series. PR-4 will plumb in the eviction-time / BE-shutdown capture hook, GC stale snapshots, and flip the config-flag default if validation lands on a real iteration.

### Helpers in `be/src/storage/lake/lake_persistent_index_snapshot.{h,cpp}`
- `get_lake_persistent_index_snapshot_path(tablet_id, version, *path)` — derives `<root>/lake_pk_snapshot/<tablet_id>/v<version>.snapshot`. Honours `config::pk_index_snapshot_local_dir` when set, otherwise the first `config::storage_root_path` entry.
- `validate_lake_persistent_index_snapshot(meta, expected_*, now_unix_sec, max_age_sec)` — pure helper deciding if a captured snapshot is safe to use here. Returns NotFound on any mismatch so callers cleanly fall back to cold rebuild.

### Memtable accessors in `PersistentIndexMemtable`
- `for_each_entry(callback)` — iterates `_map` for snapshot capture. Template, no overhead in non-snapshot paths.
- `bulk_insert_from_snapshot(entries)` — rebuilds the map from captured `(key, version, value)` triples, preserving per-entry versions. Returns InternalError on duplicate keys.
- `size()` — used by the restore path to assert the memtable was empty before bulk-insert.

### Real impls in `LakePersistentIndex`
**`try_serialize_to_local_snapshot`** — builds a `LakePersistentIndexSnapshotMetaPB` (`tablet_id`, `captured_version=metadata->version()`, `schema_id`, `captured_at_unix_sec`, walks `_memtable->for_each_entry` for `memtable_entries`, walks `metadata->sstable_meta()` for fileset SST refs), creates the per-tablet directory, calls `write_lake_persistent_index_snapshot`.

**`try_restore_from_local_snapshot`**:
1. Reads the file; any read error (NotFound / Corruption / IOError) is normalised to NotFound for clean fallback.
2. Calls `validate_lake_persistent_index_snapshot` against `_tablet_id`, `base_version`, `metadata->schema().id()`, `config::pk_index_snapshot_max_age_sec`.
3. SST presence pre-flight: every filename in the snapshot must still appear in `metadata->sstable_meta()`. Any missing SST → snapshot stale → NotFound.
4. Bulk-inserts captured `memtable_entries` into the empty memtable produced by the caller's prior `init(metadata)`. Sets `_key_size` and zeros `_need_rebuild_*` to match a cold-rebuilt index.
5. On a post-flush snapshot (common case) `memtable_entries` is empty and bulk-insert is a no-op — but the rowset-scan + `load_dels` work in `load_from_lake_tablet` is still skipped. That is the actual win.

### Unit tests `be/test/storage/lake/lake_persistent_index_snapshot_test.cpp`
- File-layer round-trip via the PR-2 read/write helpers; verifies all proto fields match.
- Magic-byte corruption → Corruption. Payload-byte flip → CRC mismatch → Corruption. Missing file → NotFound / IOError.
- `validate_lake_persistent_index_snapshot` happy path + each mismatch (tablet_id / version / schema_id / format_version) + expiry + age-check-disabled (`max_age_sec=0`).
- `get_lake_persistent_index_snapshot_path` with override and storage_root_path fallback.

## What this PR does NOT do
- No call to `try_serialize_to_local_snapshot` exists yet — the eviction-time and BE-shutdown capture hooks land in PR-4. Snapshots are produced only if a future caller invokes the serialise method.
- No GC of stale snapshots — PR-4.
- No interaction with `merge_base_level=true` compactions; those rewrite SSTs and the validity check correctly returns NotFound in that case, a missed restore (not a correctness bug).

## Risk
Default off via `config::enable_pk_index_snapshot_persistence`. With the flag off, both methods short-circuit and never touch disk. With it on, the only externally observable change is two extra `pindex_snapshot_restore_*` TRACE counters in the publish-version trace; the write path still requires PR-4's caller. Validity rules are conservative — always prefer falling back to cold rebuild over restoring questionable state.

## Test plan
- [ ] `be/test/storage/lake/lake_persistent_index_snapshot_test.cpp` passes via TSP CI build.
- [ ] PR-2 snapshot framing tests (already in #72167) still pass.
- [ ] Manual review that the config-flag-off short-circuit is preserved (no observable behaviour change in default builds).
- [ ] Validation rule coverage (each mismatch returns NotFound, OK only on full match) is exercised by the new tests.


luohaha:

PR #72170, opened 2026-04-25
[auto-perf-opt] PR-4/4: pk-index snapshot capture + GC hooks (stacked on #72169)

luohaha:
## Summary

Final PR in the lake `_index_cache` snapshot persistence series. Stacked on PR-3 (#72169). Wires the capture path that PR-3 left as a public method into the three places that own the lifecycle of cached PK indexes, plus periodic GC of stale snapshot files.

**Stack**: PR-1 #72166 (scaffold) → PR-2 #72167 (proto + framing) → PR-3 #72169 (real save/restore + UTs) → **PR-4 (this PR)** (capture hooks + GC).

### Capture hooks (3 sites)

1. **`~UpdateManager()`** — captures all currently-cached entries before the destructor clears them. Lets a future BE start skip cold rebuild for whatever was resident at shutdown time. Gated by `enable_pk_index_snapshot_persistence && pk_index_snapshot_capture_on_shutdown` (default on once persistence is enabled).
2. **`UpdateManager::expire_cache()`** — pre-walk via `DynamicCache::get_all_entries()` before `_index_cache.clear_expired()`. The pre-walk is the operator-locked Option B from iter-021 — bumps each entry's ref so we can serialise outside the cache lock; restore-time version-check catches any TOCTOU. Gated by `pk_index_snapshot_capture_on_eviction` (default **false**, opt-in until validated separately).
3. **`UpdateManager::evict_cache()`** — same pre-walk before memory-pressure `try_evict`. Same gate.

### Capture helper

`UpdateManager::_capture_index_cache_snapshots(const char* reason)` walks `_index_cache.get_all_entries()`, calls `entry->value().try_snapshot_to_local(_tablet_mgr)` on each, then releases. Per-entry failures throttled at 100x to keep logs clean.

### LakePrimaryIndex::try_snapshot_to_local(TabletManager*)

New public method on `LakePrimaryIndex`. Fetches metadata at `data_version()`, takes `try_fetch_guard()` (non-blocking — a torn snapshot during an active upsert would be unsafe and the next round will retry), delegates to `LakePersistentIndex::try_serialize_to_local_snapshot` from PR-3. Skips silently when:
- Snapshot persistence disabled
- `tablet_mgr == nullptr` (shutdown ordering edge case)
- `_persistent_index` is not a `LakePersistentIndex` (LOCAL variant or no persistent index)
- Index is in active use (try_lock fails)
- `data_version` not yet set
- Metadata for that version no longer fetchable

The restore-side validity rules from PR-3 (tablet_id / version / schema_id / max_age / SST-presence pre-flight) are the final correctness backstop.

### GC of stale snapshots

- `gc_stale_lake_persistent_index_snapshots(root, max_age_sec, *removed_count)` — pure helper in the snapshot module. Walks `<root>/lake_pk_snapshot/<tablet>/v*.snapshot`, deletes files whose mtime is older than `max_age_sec`. Best-effort: per-file failures logged but do not abort. `max_age_sec <= 0` disables GC.
- `UpdateManager::gc_stale_pk_index_snapshots()` — self-gated by `pk_index_snapshot_gc_interval_sec` (default 1800s). Called from the existing cache-expire background thread (`olap_server.cpp` `_update_cache_expire_thread_callback`, USE_STAROS branch).
- Refactor: extracted `get_lake_persistent_index_snapshot_root()` from the existing `get_lake_persistent_index_snapshot_path` so GC and path-derivation share the same root rules.

### New BE configs

- `pk_index_snapshot_capture_on_shutdown` (mBool, default `true`)
- `pk_index_snapshot_capture_on_eviction` (mBool, default `false`)
- `pk_index_snapshot_gc_interval_sec` (mInt64, default `1800`)

### Tests

5 new UT cases in `lake_persistent_index_snapshot_test.cpp`:
- `LakePersistentIndexSnapshotRoot.derives_with_trailing_slash`
- `LakePersistentIndexSnapshotGcTest.removes_only_stale_files`
- `LakePersistentIndexSnapshotGcTest.max_age_zero_is_no_op`
- `LakePersistentIndexSnapshotGcTest.missing_root_is_ok`
- `LakePersistentIndexSnapshotGcTest.non_snapshot_files_ignored`

## Behaviour summary

| `enable_pk_index_snapshot_persistence` | `capture_on_shutdown` | `capture_on_eviction` | Effect |
|---|---|---|---|
| `false` (default) | * | * | Entire stack is a no-op. |
| `true` | `true` (default) | `false` (default) | BE-restart picks up shutdown snapshots; TTL/memory eviction does NOT capture. |
| `true` | `true` | `true` | All eviction paths capture. |

## Expected impact

Per iter-022 PL2 instrumentation, snapshot **restore** on a hit replaces ~1.16 s of pattern-A (rebuild_index_segment) + pattern-B (rebuild_index_del / load_dels) IO with a single local PL2 disk read + protobuf parse. Pattern C (SST-open) work is unavoidable infrastructure cost — the snapshot does not bypass it because PR-3's restore path runs **after** `init(metadata)` has already opened the SSTs.

The headline benefit is on **BE restart** — every tablet that was resident at shutdown skips its cold rebuild on first access after restart. Memory-pressure and TTL-eviction paths benefit too once `pk_index_snapshot_capture_on_eviction` is flipped on.

## Test plan

- [ ] CI green on the new UTs (5 GC test cases)
- [ ] Compile-only validation via TSP build for the iter-032 commit
- [ ] Future iteration: enable `enable_pk_index_snapshot_persistence=true` on a test cluster, restart all CNs, verify cold-rebuild time drops on first upsert per tablet vs baseline

## Refs

- iter-021 design (Option B)
- iter-028 feasibility study
- PR-1 #72166 (scaffold)
- PR-2 #72167 (proto + on-disk framing)
- PR-3 #72169 (save/restore + validity rules + UTs)


luohaha:

PR #72172, opened 2026-04-25
[BugFix] Cover projected sstable max_rss_rowid in update_next_rowset_id

xiangguangyxg:
## Why I'm doing:

Repeated SPLIT/MERGE cycles on a shared-data range-distribution PK table get stuck because a background COMPACTION publish on a SPLIT child fails with "sstables are not ordered" — and the next reshard job parks in `PREPARING` waiting for the publish to finish.

## What I'm doing:

`merge_sstables` (be/src/storage/lake/tablet_merger.cpp:615-618) projects each child's sstable `max_rss_rowid` by adding `ctx.rssid_offset` to the high word. The legacy non-`shared_rssid` branch can produce a projected high that exceeds every `rowset.id` in the merged metadata — for example, a delete-only sstable whose memtable max_rss_rowid was set to `(rowset_id<<32|UINT32_MAX)` (be/src/storage/lake/persistent_index_memtable.cpp:110, 131) contributes a high rssid that has no surviving rowset to pin it.

`update_next_rowset_id` walked only `metadata->rowsets()`, so the merged tablet's `next_rowset_id` was set from `rowset.id` alone. Any subsequent write on this tablet — and on every SPLIT child that inherits this metadata via `std::make_shared<TabletMetadataPB>(*old_tablet_metadata)` in `split_tablet` (be/src/storage/lake/tablet_splitter.cpp:486) — would assign a new rssid less than the projected sstable highs, producing a sstable whose `max_rss_rowid` falls below an existing entry in `sstable_meta`. That breaks the ascending-order invariant `LakePersistentIndex::commit()` enforces (be/src/storage/lake/lake_persistent_index.cpp:881):

```
sstables are not ordered, last_max_rss_rowid=A : max_rss_rowid=B
```

The COMPACTION transaction whose publish triggered the check parks in `COMMITTED`. The next reshard job on the same tablet allocates its `commitVersion` via `getNextTransactionId()` and waits in `PREPARING` for `visibleVersion` to catch up (`MergeTabletJob.runPreparingJob`), but the stuck compaction blocks the publish chain — so `PREPARING` hangs indefinitely and the table is parked in `TABLET_RESHARD`.

### Reproducer (shared-data PK range distribution, 3-CN cluster)

1. Seed N rows on a range-distributed PK table.
2. Run repeated SPLIT/MERGE cycles, each with overlapping UPSERTs:
   `SPLIT 1->8, UPSERT, MERGE 8->1, SPLIT 1->8, UPSERT, MERGE 8->1, ...`
3. After the second MERGE, a background COMPACTION on a SPLIT child fails to publish:
   ```
   Fail to publish version for tablets:[[27673, 27666, 27672, 27667, 27668, 27670, 27671, 27669]],
   error msg: sstables are not ordered,
   last_max_rss_rowid=1090921695183 : max_rss_rowid=601295447439
   ```
   (high words 254 vs 140; the new compaction output sstable's projected high is below an inherited shared sstable's projected high.)
4. The next `MERGE_TABLET` job sits in `JOB_STATE=PREPARING`; `SHOW PROC "/transactions/<db>/running"` shows the `COMPACTION_…` txn parked in `COMMITTED`, and the table state stays in `TABLET_RESHARD`.

### Fix

In `update_next_rowset_id` also walk `sstable_meta.sstables(*)` and bound `next_rowset_id` by `(max projected high) + 1`, clamping at `UINT32_MAX` to match the rssid encoding. Future writes — including those on any SPLIT child inheriting this metadata — are guaranteed to assign rssids strictly greater than every projected sstable high, so new sstables append to `sstable_meta` in ascending `max_rss_rowid` order.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/tablet_merger.cpp
+++ b/be/src/storage/lake/tablet_merger.cpp
@@ -1331,6 +1331,31 @@ void update_next_rowset_id(TabletMetadataPB* metadata) {
     for (const auto& rowset : metadata->rowsets()) {
         max_end = std::max(max_end, rowset.id() + get_rowset_id_step(rowset));
     }
+    // Also consider sstable_meta projected max_rss_rowid. merge_sstables advances
+    // each shared sstable's high word by ctx.rssid_offset (tablet_merger.cpp ~615),
+    // and the legacy non-shared_rssid branch can produce projected high words
+    // larger than any surviving rowset.id (e.g. when a delete-only sstable from a
+    // child contributes a high rssid that has no corresponding rowset in the merged
+    // metadata). If next_rowset_id is set from rowset.id alone, future writes on
+    // this tablet — and on SPLIT children that inherit this metadata — will assign
+    // rssids smaller than the projected sstable highs, producing sstables whose
+    // max_rss_rowid is LESS than existing entries and violating the ascending-order
+    // invariant that LakePersistentIndex::commit() (lake_persistent_index.cpp:881)
+    // enforces. The downstream symptom is the compaction publish failing with
+    // "sstables are not ordered, last_max_rss_rowid=A : max_rss_rowid=B" and the
+    // next reshard job parking in PREPARING because visibleVersion never catches
+    // up. Bound next_rowset_id by (max projected high word) + 1 so any new rssid
+    // is strictly greater than every projected sstable rssid.
+    if (metadata->has_sstable_meta()) {
+        for (const auto& sst : metadata->sstable_meta().sstables()) {
+            uint64_t projected_high = sst.max_rss_rowid() >> 32;
+            // Clamp the projected high to uint32 range; rssids are uint32 elsewhere
+            // and any saturated high word would already break the rssid encoding.
+            if (projected_high < std::numeric_limits<uint32_t>::max()) {
+                max_end = std::max(max_end, static_cast<uint32_t>(projected_high) + 1);
+            }
+        }
+    }
     metadata->set_next_rowset_id(max_end);
 }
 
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
## Cluster validation update

After deploying this PR's fix to a test cluster (`combined-dcg-pk-flush`, build 1283 / `main-ea46c47`) and rerunning the v17 S172 5x SPLIT/MERGE-cycle reproducer:

**The fix is a clear win**:
- Cycles 0-2 now run with **real** `1->8 SPLIT` + `8->1 MERGE` in ~8s each (previously 65s with cycles 0/1 silently falling back to `1->1`).
- The "sstables are not ordered" error is gone — confirming the `update_next_rowset_id` fix correctly bounds new rssids above the projected sstable highs.

**A separate bug (Bug B) is now exposed** that was previously masked by the Bug D ordering check:

```
Fail to publish partition 10173 error Fail to publish version for tablets:[[10378]],
error msg: unexpected segment id: 373 tablet id: 10378
```

Source: `MetaFileBuilder::update_num_del_stat` (be/src/storage/lake/meta_file.cpp:608) — the PK index returned a `rssid` that no `rowset.id + segment_idx` in the merged metadata covers. Likely root cause: post-MERGE `sstable_meta` retains projected sstable entries whose source rowset was dropped during merge consolidation.

This is a separate, deeper issue and should be tracked as a follow-up. **This PR (#72172) is correct and self-contained**; merging it removes one whole class of reshard hangs and unblocks investigation of the remaining one.

CelerData-Reviewer:
@codex review

luohaha (dismissed):

CelerData-Reviewer:
@codex review

luohaha (approved):

srlch (approved):

PR #72171, opened 2026-04-25
[auto-perf-opt] PR-fix: register pk-index snapshot configs in fwd-header manifest

luohaha:
## Summary
The `_index_cache` snapshot persistence 4-PR series (#72166 → #72167 → #72169 → #72170) added 6 new BE configs to `be/src/common/config.h`, but missed adding them to `be/src/common/config_fwd_headers_manifest.json`. Because the auto-generated `config_primary_key_fwd.h` is what `lake_persistent_index.cpp` actually includes (not the full `config.h`), the symbols were not visible at compile time.

**TSP build 1280** for PR #72169 (PR-3 of the stack) failed with:
```
error: 'enable_pk_index_snapshot_persistence' is not a member of 'starrocks::config'
error: 'pk_index_snapshot_max_age_sec' is not a member of 'starrocks::config'
```

PR #72170 (PR-4 of the stack) inherits the same compile error.

## What this PR does
- Adds the 6 missing configs to the `config_primary_key_fwd.h` manifest entry:
  - `enable_pk_index_snapshot_persistence`
  - `pk_index_snapshot_local_dir`
  - `pk_index_snapshot_max_age_sec`
  - `pk_index_snapshot_capture_on_shutdown`
  - `pk_index_snapshot_capture_on_eviction`
  - `pk_index_snapshot_gc_interval_sec`
- Regenerates `config_primary_key_fwd.h` via `python3 build-support/gen_config_fwd_headers.py`.

## Impact
No behaviour change. All 6 configs still default off / empty / unchanged. This PR only makes the snapshot-persistence stack actually compile so the end-to-end validation experiment in iter-033 can proceed.

## PR series context
This is a fixup that should logically have been part of PR-1 (#72166). It is opened separately as iter-033 to preserve the iteration/PR audit trail. After merge, the stack `#72166 → #72167 → #72169 → #72170 → THIS` compiles cleanly.

## Test plan
- [ ] TSP build of this PR's branch succeeds (the 4 stacked PRs + this fixup)
- [ ] BE compiles with no `'… is not a member of starrocks::config'` errors
- [ ] Default-off behaviour preserved (no functional change)

PR #72173, opened 2026-04-25
[auto-perf-opt] PR-fix: use RandomAccessFile::get_size() in pk-index snapshot reader

luohaha:
## Bottleneck context

This is a build-fix follow-up in the `_index_cache` snapshot persistence stack:

- PR #72166 — PR-1: scaffold + 6 BE configs
- PR #72167 — PR-2: proto + on-disk framing (introduced the bug)
- PR #72169 — PR-3: real save/restore + UTs
- PR #72170 — PR-4: capture hooks + GC
- PR #72171 — manifest/fwd-header fixup so PR-1's configs are actually visible
- **this PR** — `RandomAccessFile::size()` → `get_size()`

## What was wrong

`be/src/storage/lake/lake_persistent_index_snapshot.cpp:75` called
`rf->size()`. `RandomAccessFile` inherits from `io::SeekableInputStreamWrapper`
and exposes `StatusOr<int64_t> get_size()` — there is no `size()` method.

TSP build 1282 (PR #72171, full stack) failed at 96 % compilation with:

```
lake_persistent_index_snapshot.cpp:75:36: error:
'class starrocks::RandomAccessFile' has no member named 'size'
```

## Fix

Switch to `ASSIGN_OR_RETURN(const int64_t file_size, rf->get_size())` and cast
the `size_t` comparison RHS to `int64_t` to silence `-Wsign-compare`.
Behaviour-neutral.

## Expected impact

Once green, the 5-PR snapshot persistence stack finally compiles and the
end-to-end validation experiment (flip flag → restart CN → verify snapshot
files written + `pindex_snapshot_restore_hit > 0` on second start) becomes
runnable for the first time.


luohaha:

PR #72174, opened 2026-04-25
[auto-perf-opt] PR-fix: capture pk-index snapshots before BE threadpool teardown

luohaha:
## What
Adds `UpdateManager::pre_shutdown_hook()` invoked from `starrocks_be.cpp` immediately after `exec_env->wait_for_finish()` so the pk-index snapshot capture path runs while threadpools / tablet manager are still alive. Stacked on **#72173**.

## Why (root cause from iter-035 E2E validation)
The destructor-based shutdown hook in `~UpdateManager` is **unreachable** on a real shared-data BE.

Iter-035 deployed the snapshot persistence stack (#72166 → #72167 → #72169 → #72170 → #72171 → #72173) and ran the long-deferred end-to-end validation: enabled `enable_pk_index_snapshot_persistence=true` on one CN via `/api/update_config`, populated `_index_cache` via a 2-min realtime-benchmark Test, then ran `stop_be.sh -g` (graceful, SIGTERM).

Result: BE entered shutdown cleanly (`got signal: Terminated` ... `exit step 1: wait exec engine tasks finish` ... `exit step 4: daemon threads exit successfully`), then **FATAL'd ~50 s later in `exec_env->stop()`**:

```
F20260426 03:40:38.219921 ... threadpool.cpp:270] Check failed: 1 == _tokens.size() (1 vs. 2) Threadpool cloud_native_pk_index_compact destroyed with 2 allocated tokens
```

The CHECK abort fires before storage-engine teardown reaches `~UpdateManager`, so the destructor never runs and `_capture_index_cache_snapshots("shutdown")` never fires. Filesystem confirmed: `/home/disk1/sr/be/storage/lake_pk_snapshot/` was never created and 0 files written.

The threadpool token leak on shutdown is a separate latent bug (orthogonal to this PR), but it makes the snapshot persistence stack measurably useless on this cluster: every BE shutdown abort()s before capture.

## What this PR does
1. **`UpdateManager::pre_shutdown_hook()`** (new public method): config-gated, idempotent (`_pre_shutdown_capture_done` atomic), calls `_capture_index_cache_snapshots("pre-shutdown")`.
2. **`starrocks_be.cpp`**: invokes `update_mgr->pre_shutdown_hook()` immediately after `exec_env->wait_for_finish()` and adds an `exit step` log line so operators can confirm the capture phase ran.
3. **`~UpdateManager`**: kept as a best-effort fallback but guarded on `_pre_shutdown_capture_done` so a clean teardown does not double-walk the cache.
[...]

PR #72175, opened 2026-04-25
[auto-perf-opt] Capture stub snapshots for evicted PK tablets at shutdown

luohaha:
## Summary

Extends the BE-shutdown PK-index snapshot capture path (PR #72174) to ALSO cover tablets that were loaded on this CN but were evicted from `_index_cache` before the graceful stop (TTL or memory-pressure). For each such tablet, a "stub" snapshot is written: empty `memtable_entries` + the SST filenames from the most recent tracked metadata + same framing as a full snapshot.

Restoring from a stub at the same version yields an empty memtable while keeping `_sstable_filesets` (populated by `init(metadata)`) intact — the correct post-rebuild state IFF the tablet has no rowsets pending merge into the SSTs. The capture enforces that invariant via `LakePersistentIndex::need_rebuild_counts == {0, 0}`, the same predicate `load_from_lake_tablet` uses to decide whether to skip the rebuild loop. Tablets failing the gate are skipped silently and fall back to the existing cold-rebuild path on next load.

## Why

iter-038's 6-CN A/B for the snapshot persistence stack landed on a 55.2% restore HIT rate (1,159 HIT / 940 MISS / 2,099 total restore attempts post-restart). The 940 MISSes were tablets first-touched in Variant B that had no `_index_cache` entry at the prior shutdown — so no snapshot existed for them. Per-event speedup was 10–13× for HITs (HIT max 213.9 ms vs MISS max 2,685.2 ms on .125), but the workload-level upsert Max stayed at 7.75 s because the worst MISS events still ran a full cold rebuild.

Closing the MISS-share is therefore the highest-leverage next lever for this optimization run, exactly as the iter-038 diagnosis (option 4) and the iter-021 reframing suggestion both flagged. This change is the smallest correct realisation of "snapshot beyond `_index_cache`" — no new threads, no boot-time scanning, no metadata RPCs. Just a tracking map updated from the existing publish hot path and a second walk at the existing shutdown hook.

## Mechanism

- `UpdateManager::_last_seen_pk_metadata` (mutex-protected `unordered_map<int64_t, TabletMetadataPtr>`) records the latest metadata for every CLOUD_NATIVE PK tablet that has gone through `prepare_primary_index`.
- `pre_shutdown_hook()` runs the existing `_capture_index_cache_snapshots` first (full-memtable capture for in-cache entries), then the new `_capture_evicted_tablet_snapshots` (stub capture for the residual). Same destructor fallback ordering.
- Stub writer (`write_stub_lake_persistent_index_snapshot` in `lake_persistent_index_snapshot.{h,cpp}`) is a pure helper that builds the proto, writes via the existing `write_lake_persistent_index_snapshot` (preserving framing + CRC), and reuses `get_lake_persistent_index_snapshot_path` for the on-disk layout.
- The walk pre-checks file existence at the target path so a stub never clobbers a richer (full-memtable) snapshot already written by the in-cache walk for the same `(tablet_id, version)` pair.

## Memory cost

`_last_seen_pk_metadata` holds shared_ptr<TabletMetadataPB> per distinct PK tablet ever served on the CN. For the realtime-benchmark `999_1gb_1_1tb` template with ~10 K tablets, this is a few MB of pointers + a single shared metadata payload that's already cached on the publish path. Bounded by the CN's tablet assignment.

## Default behaviour

Gated by the existing `enable_pk_index_snapshot_persistence` master flag and `pk_index_snapshot_capture_on_shutdown`. No new config is added — when persistence is disabled the new tracking is a no-op and the new walk early-exits.

## Test plan

- [x] Unit test `LakePersistentIndexSnapshotStubTest.write_then_read_roundtrip` — covers the helper end-to-end (write → read → validate, both matching and version-mismatched lookups).
- [x] Existing `LakePersistentIndexSnapshotTest` suite continues to cover format framing, CRC, and validity-rule semantics — unchanged.
- [ ] iter-040 cluster-wide A/B on `benchmark_luoyixin_1tb_6cn_2disk2` (PL2): same 6-CN cold-restart workload as iter-038, comparing restore HIT rate and upsert Max with vs without this patch on top of PR #72174.
- [ ] Verify the new `LOG(INFO) "pk-index stub-snapshot capture (pre-shutdown-evicted) walked=N ok=K skipped=S failed=F"` line shows up at graceful stop and that `<root>/lake_pk_snapshot/<tablet_id>/v<ver>.snapshot` files exist for tablets NOT in `_index_cache` at shutdown.
- [ ] Confirm pre-existing snapshot UTs in `lake_persistent_index_snapshot_test.cpp` still pass (no behaviour change to the read / validate / GC paths).

## Stacked on

This PR builds on PR #72174 (and the earlier #72166–#72173 stack). The stub-snapshot reader path is the existing `try_restore_from_local_snapshot`, which already handles the empty-memtable case correctly (the `memtable_entries_size() > 0` gate skips `bulk_insert_from_snapshot`).

## Limitations & follow-ups

- A stub snapshot's correctness depends on `need_rebuild_counts == {0, 0}` at capture time. Tablets with active writes between last publish and shutdown won't be stub-captured (skipped). For those, full capture via the in-cache walk is the only path — same as today.
- This PR does NOT add a boot-time pre-warm of `_index_cache`. That's iter-038 priority 1 and is independent; it shifts lazy-restore latency to boot but doesn't change hit rate. Best to land separately once stub-capture's measured gain is in.
- The pre-existing `cloud_native_pk_index_compact` threadpool token-leak FATAL (sidestepped by PR #72174's pre-shutdown hook ordering) is still latent. Fix in `LakePersistentIndexParallelCompactMgr` token-release path is a separate follow-up PR.


luohaha:

PR #72176, opened 2026-04-26
[auto-perf-opt] Pre-warm pk-index cache from local snapshots at BE boot

luohaha:
## Summary

Symmetric counterpart to the pre-shutdown capture path landed in PR #72174 / #72175. After exec_env init and before BRPC accepts requests, the BE walks `<root>/lake_pk_snapshot/`, picks the highest-versioned `v<N>.snapshot` per tablet directory, and asynchronously restores each tablet into `_index_cache` so that the first publish on each tablet hits a warm cache rather than paying the lazy-restore-on-publish latency.

## Why

iter-038 cluster-wide A/B (PR #72174 alone, 6-CN graceful-restart) measured:

- restore HIT rate **55.2%** (1159/2099) — capped by `|_index_cache|` at shutdown
- HIT-path latency: p50 13.6 ms, p99 170.8 ms, max 213.9 ms
- MISS-path latency (cold rebuild): p50 141.2 ms, p99 2540.3 ms, max 2685.2 ms

For the 1159 HIT tablets, the post-restart publish path still pays the snapshot-read + parse + bulk_insert window the first time it touches each tablet. Boot-time pre-warm restores those tablets *before* the first publish so the publish hits a fully-warm cache.

## What changes

- New BE config `enable_pk_index_snapshot_prewarm_on_boot` (default `true`) and `pk_index_snapshot_prewarm_threads` (default `8`). Registered in `config.h`, `config_primary_key_fwd.h`, and the fwd-headers manifest.
- New `LakePrimaryIndex::try_lake_load_from_snapshot()` — restore-only variant of `lake_load` that returns `NotFound` on snapshot miss instead of falling through to the cold-rebuild loop. Pre-warm should never do the heavyweight work that a publish-time load would do anyway.
- New `UpdateManager::boot_prewarm_hook()` — walks the snapshot root, picks the highest-versioned snapshot per tablet directory, and fans out per-tablet pre-warm tasks across `worker_threads` `std::thread`s on a detached background thread. Best-effort: per-tablet failures are logged but do not abort the walk or block subsequent publishes. Logs a final tally `pk-index snapshot prewarm: complete root=<root> walked=N ok=K miss=S failed=F`.
- Hook call site in `starrocks_be.cpp` after `start_bg_threads()`, mirroring the placement of `pre_shutdown_hook` on the shutdown side.
- Shared `parse_snapshot_version_from_filename(string_view)` helper exposed in `lake_persistent_index_snapshot.h` so the boot walk reuses the same on-disk schema parser.
- Unit tests for the parser (well-formed, malformed prefixes/suffixes, non-digit characters, overflow).

## Concurrency notes

- Race with publish path is benign: both `_index_cache.get_or_create` + `fetch_guard` call sites take the same mutex, so whichever wins first does the load and the other returns the cached result early via the `_loaded && !need_rebuild()` short-circuit.
- Avoids the recurring `cloud_native_pk_index_compact` ThreadPool-token-leak FATAL pattern (iter-035 / iter-040) by using bare `std::thread` workers without ThreadPool tokens.

## Expected impact

- Post-graceful-restart upsert P99/Max for HIT tablets drops further (the 213 ms HIT max in iter-038 becomes a cache-hit on the publish path).
- Cluster-wide impact gated on publish-side timing — to be measured in iter-042.
- No effect on non-PK tablets, on tablets with no snapshot file, or when snapshot persistence is disabled.

## Test plan

- [x] Unit tests for `parse_snapshot_version_from_filename` (well-formed, malformed, overflow inputs)
- [x] `python3 build-support/gen_config_fwd_headers.py --check` passes
- [ ] TSP build of this commit succeeds
- [ ] iter-042 cluster-wide A/B (boot pre-warm OFF vs ON) on `benchmark_luoyixin_1tb_6cn_2disk2` post-graceful-restart, comparing upsert P99/Max and the new `pk-index snapshot prewarm: complete` log line tally vs iter-038's 55.2% HIT rate


luohaha:

PR #72177, opened 2026-04-26
[BugFix] Remap deduped source rowset's sstable via shared_rssid in MERGE

xiangguangyxg:
## Why I'm doing:

`merge_sstables`'s non-shared_rssid branch (`be/src/storage/lake/tablet_merger.cpp:610`) applied `ctx.rssid_offset()` blindly to every legacy memtable-flush sstable. When the sstable's source rowset (high word of `max_rss_rowid`) was a duplicate of another child's canonical rowset (`shared_rssid_map` hit, populated in `merge_rowsets` at `tablet_merger.cpp:247`), the blanket offset translated the sstable's keys to the duplicate's would-be slot in the merged id space — a position with no rowset, since `merge_rowsets` skipped the duplicate.

Downstream, an INSERT publish whose PK index lookup returns that ghost rssid hits the consistency check at `MetaFileBuilder::update_num_del_stat` (`be/src/storage/lake/meta_file.cpp:608`) with:

```
unexpected segment id: <ghost> tablet id: <merged_tablet>
```

The publish parks in `COMMITTED`. The next reshard job allocates its `commitVersion` via `getNextTransactionId()` and waits in `PREPARING` for `visibleVersion` to catch up — but the stuck publish blocks the chain, so `PREPARING` hangs indefinitely and the table is parked in `TABLET_RESHARD`.

This bug had been masking (and was masked by) a related ordering issue fixed in the companion PR for `update_next_rowset_id` (#72172); with the sstable ordering invariant restored, the dedup'd-rowset sstable remap surfaces.

### Reproducer (shared-data PK range distribution, 3-CN cluster)

1. Seed N rows on a range-distributed PK table.
2. Run repeated `SPLIT 1->8 / UPSERT / MERGE 8->1` cycles. Inherited shared rowsets get deduped on each MERGE; their memtable-flush sstables are legacy (no `shared_rssid`) and carry source rssids that match the deduped id.
3. After 2-3 cycles, an INSERT publish on a SPLIT child fails with:
   ```
   Fail to publish version for tablets:[[12094]],
   error msg: unexpected segment id: 90 tablet id: 12094
   ```
4. The next `MERGE_TABLET` job sits in `JOB_STATE=PREPARING` forever.

## What I'm doing:

In `merge_sstables`'s non-shared_rssid branch, derive the source rssid from the high word of the sstable's `max_rss_rowid`. If `shared_rssid_map` has an explicit entry for it, switch the output sstable to the shared_rssid projection:
- stamp `shared_rssid` with the canonical rssid
- set `shared_version` to the merged tablet's version (>0 so the read path takes the remap branch at `persistent_index_sstable.cpp:207`)
- clear `rssid_offset` to avoid double-transform
- patch the high word of `max_rss_rowid`

The read path then replaces every key's rssid with the canonical, matching the surviving `rowset.id+segment_idx` in the merged metadata. When `shared_rssid_map` has no entry, the existing offset path is preserved unchanged.

## Test

`test_tablet_merging_sstable_remap_for_deduped_source_rowset` constructs two children where both carry an identical shared rowset (id=5) plus a unique local rowset, with a legacy memtable-flush sstable encoding source rssid=5 and no `shared_rssid`. After MERGE, the test asserts every sstable's read-path effective rssid lands on a `rowset.id+segment_idx` that exists in the merged metadata — the exact invariant `meta_file.cpp:608` enforces.

All 66 `LakeTabletReshardTest` tests pass locally (65 existing + 1 new).

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Test plan

- [x] BE UT: 66/66 `LakeTabletReshardTest` PASS locally including new test
- [ ] CI: BUILD + BE UT + SQL-Tester green on this branch
- [ ] Integration: validate on the test cluster that v17 S172 5x SPLIT/MERGE/UPSERT cycles complete without the "unexpected segment id" hang

### DIFF
--- a/be/src/storage/lake/tablet_merger.cpp
+++ b/be/src/storage/lake/tablet_merger.cpp
@@ -62,6 +62,17 @@ public:
 
     bool has_shared_rssid_mapping() const { return !_shared_rssid_map.empty(); }
 
+    // Returns the canonical rssid if |rssid| was deduped, else std::nullopt.
+    // Lets callers distinguish "explicit dedup remap" from "default offset map"
+    // without falling through map_rssid's offset path.
+    std::optional<uint32_t> shared_rssid_map_lookup(uint32_t rssid) const {
+        auto it = _shared_rssid_map.find(rssid);
+        if (it == _shared_rssid_map.end()) {
+            return std::nullopt;
+        }
+        return it->second;
+    }
+
     // Fills shared_rssid_map so that all rssids occupied by |rowset| map to
     // the corresponding rssid in |canonical_rowset|.
     void update_shared_rssid_map(const RowsetMetadataPB& rowset, const RowsetMetadataPB& canonical_rowset) {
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
## Closing — fix is incorrect

Validated against the v17 S172 reproducer on TSP build 1288: **this fix made things worse** (cycle 2 hangs instead of cycle 3 with Bug D fix alone).

**Root cause of the regression**: my fix sets `shared_version` on memtable-flush sstables when the source rowset is in `shared_rssid_map`. The read-path remap branch at `persistent_index_sstable.cpp:207-212` then replaces every key's version *and* rssid with the stamped values — but memtable-flush sstables can hold keys from multiple writes/versions, so stamping a single `shared_version` is incorrect.

**Why the hypothesis was off**: I assumed `max_rss_rowid >> 32` identifies a single source rowset for the whole sstable. That's true for compaction-output sstables (where ingest_sst legitimately sets shared_rssid) but **not** for memtable-flush sstables, which can span multiple rowset rssids and versions.

The `unexpected segment id` symptom is real (Bug B), but a correct fix needs to either:
- preserve per-key rssid translation while still avoiding ghost mappings (e.g. carry shared_rssid_map state through to read time), or
- ensure `merge_rowsets` doesn't drop rowsets whose sstables still contribute keys at that rssid.

Reverted from the test branch (`combined_dcg_sampling`). Bug D fix in #72172 stands on its own — it correctly fixes the ordering invariant violation.

Bug B reopened as a follow-up; needs deeper investigation than this loop iteration affords.

PR #72181, opened 2026-04-26
[WIP] Investigate Bug B duplicate-row issue (no fix yet)

xiangguangyxg:
## Status
Reverted to empty diff after F.1 (skip merge_sstables) failed cluster validation.

## What we tried (all failed)
1. **Drop-on-NotFound** (build 1307): cycles 0-2 stable, cycle 3 drift to 270000, cycle 4 to 288000.
2. **Sentinel-rssid write** (build 1308): cycles 0-1 stable, cycle 2 drift to 270000, cycle 3-4 saturate at 283393.
3. **F.1: skip merge_sstables, rely on load-time rebuild** (build 1315): cycles 0-2 stable, cycle 3 drift to 306928, cycle 4 hangs.

## Real bug shape (build 1315 forensics)
- Merged tablet 10396 (single tablet after MERGE 8→1): COUNT(*)=306928, COUNT(DISTINCT id)=250000, MAX(id)=249999.
- 56928 duplicate physical rows for PKs in seed range — persistent index isn't filtering them at read time.
- `GROUP BY id HAVING c>1` returns NO rows but COUNT(*) is 56k bigger than DISTINCT — bizarre contradiction.
- SHOW TABLET reports RowCount=315336 (~8k delvec'd).

## Conclusion
Bug B is NOT in `merge_sstables` / sstable_meta inheritance / per-key rssid remap.
Bug B is in **delvec/rowset merging producing duplicate physical rows post-MERGE** that persistent index doesn't mask at read time. Drop/sentinel/F.1 only attacked the persistent-index symptom; the underlying duplicate physical rows in rowsets persist.

Closing this PR until we have a real fix. Tracking the investigation as Bug B Plan B Phase 5.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/ivm/IvmRewriter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/ivm/IvmRewriter.java
@@ -41,6 +41,7 @@ import com.starrocks.sql.optimizer.operator.scalar.ConstantOperator;
 import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
 import com.starrocks.sql.optimizer.rule.RuleSet;
 import com.starrocks.sql.optimizer.rule.ivm.common.IvmRuleUtils;
+import com.starrocks.sql.optimizer.rule.transformation.materialization.MvUtils;
 import com.starrocks.sql.optimizer.task.TaskContext;
 import com.starrocks.sql.optimizer.task.TaskScheduler;
 import com.starrocks.type.IntegerType;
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
## Phase 2 implementation update

I've implemented Phase 2 (read+remap+rewrite of flagged sstables) on top of this Phase 1 detection. **Cluster validation revealed the trigger conditions are still incomplete**:

- Phase 2 widens detection to scan the entire \`[0, source_high]\` stored-rssid range (not just \`max_rss_rowid >> 32\`)
- Despite that, v17 S172 cycle 2 still hangs with \`unexpected segment id: 45 tablet id: 10355\` while the rebuild metric reads 0 across all CNs
- The hang is on a SPLIT child (10355) that inherited metadata from cycle 1's MERGE — meaning the bad state was already in the merged metadata when cycle 1 committed, and my detection didn't fire then either

Bug B has additional flavors that don't go through the non-shared_rssid offset path I'm targeting. Likely candidates:
- Shared_rssid sstables whose mapped rssid via \`ctx.map_rssid\` lands in a metadata gap from a prior cycle
- SPLIT-side propagation: \`tablet_splitter\` simply copies parent metadata, so any bad state from MERGE persists through SPLIT
- Multi-cycle accumulation: bad state gets amplified across SPLIT/MERGE generations even when each individual operation looks safe

Phase 1 detection (this PR) still ships as-is — it's a useful observability signal for the case it catches, and the metric is now production-ready. Phase 2 work is paused pending a more comprehensive design pass on the architectural issue rather than another iteration of "narrow detection criteria + retest".

xiangguangyxg:
Force-pushed and updated PR title/body to reflect expanded scope:

- **Title** changed from "[Enhancement] Detect MERGE sstable rssid translation that lands outside merged metadata" to "[BugFix] Rebuild MERGE sstable for ghost rssid + preserve dedup consistency"
- **Body** rewritten to cover all three pieces (detection / rebuild / dedup consistency)
- Behavior-change checkbox flipped to **yes** (rebuild rewrites sstables on detection)
- Added `Fixes #64986` umbrella ref
- Ticked **4.1** auto-backport

The cluster validation with the dedup-consistency fix is in flight on TSP build 1297.

xiangguangyxg:
Correction: behavior-change checkbox flipped back to **No**. Internal correctness fixes (rebuild offending sstable to avoid hang) don't qualify as user-visible behavior change in this template's sense — only SQL/config/policy/feature-removal changes do.

xiangguangyxg:
Force-pushed and updated PR title/body. Scope expanded to four pieces (added compaction-log no-op fix at `lake_persistent_index.cpp:843`). Build 1297 cluster validation surfaced the post-MERGE compaction-log path; build 1298 with the no-op fix is in flight.

xiangguangyxg:
## Path B investigation — structural hole in rebuild

Cluster validation on build 1298 (commit `49eacfd`) still hangs at v17 S172 cycle 2 with `unexpected segment id: 52 tablet id: 10357` at `update_num_del_stat` (`meta_file.cpp:609`).

The ghost rssid flows from a structural hole in `rebuild_sstable_with_per_key_remap`, not from a missing downstream patch.

**Analysis** (full writeup in commit description on next push):

After SPLIT 1→8 + per-child compaction + MERGE 8→1: compacting children's R' rowsets get added to merged metadata as separate instances (`shared_segments=false` correctly defeats dedup), `shared_rssid_map` stays empty for those rowsets. `rebuild_sstable_with_per_key_remap` then falls through to `stored_rssid + offset` — which lands on positions that no rowset occupies (gaps where the 5 deduped children's R was collapsed). The rebuilt sstable encodes those ghost rssids, which propagate into the persistent index and surface as `unexpected segment id` on the next UPSERT.

The 4 patches in this PR each fixed a downstream consumer of ghost rssids (read path → dedup consistency → compaction log) but the rebuild keeps producing them. **Path B** is needed: when offset translation lands outside `covered_rssids` AND `shared_rssid_map` has no entry, the rebuild must `find_covering_rowset` by key range in the merged metadata. Open questions on overlapping ranges and tombstones need review before shipping.

**Pausing autonomous fixes** here. Reverting to Phase 4 (compaction-log no-op) leaves the PR at 1 build's worth of validation; merge readiness depends on the Path B design call.

xiangguangyxg:
## Path B v5 — implementation submitted

Codex review approved v5 (5 rounds, gpt-5.5). Fix branch `fix_bug_b_rebuild_sstable_at_merge` force-pushed at `81d681b68f7`.

**v5 implementation summary**:
- Source-lineage remap: per-stored-rssid lookup via `source_rssid_to_rowset` map (built from `get_rssid()`, honors sparse `segment_metas.segment_idx`); maps source rowset → merged via shared_rssid_map or add_rowset offset
- Tombstones (`rssid==UINT32_MAX && rowid==UINT32_MAX`) passed through verbatim
- max_rss_rowid: 3 cases (real-only / mixed / tombstone-only), uses `std::max_element` for tombstone floor when source max is itself null
- Empty output → `std::nullopt` → `dest->RemoveLast()` to avoid empty-PB residue
- shared_rssid path validates `ctx.map_rssid()` result against covered_rssids
- int32 overflow guarded before `set_rssid_offset` cast
- Rebuild failure propagates Corruption (no offset-fallback masking)

New bvar `tablet_reshard_merge_sstable_rebuild_failed_total` for monitoring.

**Validation**: TSP build 1303 in flight. v17 S172 + full v8-v17 regression to follow.

**BE UTs** (10 cases per design): scheduled as follow-up commit before merge — current commit ships the runtime fix; UT additions need access to existing tablet_reshard_test.cpp fixtures and will be added in a subsequent push.

PR #72199, opened 2026-04-27
[BugFix] Restrict Ranger root bypass to StarRocks-internal controller

### LINKED ISSUE #72198: StarRocks root user bypasses Ranger authorization on external catalogs (Hive/Iceberg/Hudi/...)
 1. Configure a Hive external catalog with Ranger authorization enabled
     (`RangerHiveAccessController`).
  2. In Ranger Admin, create a Hive policy that explicitly denies `root` (or
     simply does not grant `root` any access) on `db.table`.
  3. Connect to StarRocks as the built-in `root` user.
  4. Run `SELECT * FROM hive_catalog.db.table;`

 Query is rejected by Ranger because the configured Hive policy does not
  grant `root` access to that table.

### Real behavior (Required)
 Query succeeds. StarRocks `root` is treated as a super-user against the
  external Hive catalog and bypasses Ranger entirely.

### Root cause
`RangerAccessController.hasPermission()` (the abstract base class) short-
  circuits when `UserIdentity.ROOT.equals(user)`. This bypass is inherited
  by every Ranger controller — including `RangerHiveAccessController`, etc. —
  which means the StarRocks built-in `root` becomes a de-facto super-user
  for *all* external Ranger-protected data sources, breaking the data-lake
  authorization boundary.

  The bypass is only semantically correct for the StarRocks-internal
  service controller (`RangerStarRocksAccessController`), where `root` is
  the StarRocks super-user. It is incorrect for any external service whose
  Ranger policies are owned by that external system.

 - 4.1.0

liuzhao-lz:
## Why I'm doing:
`RangerAccessController.hasPermission()` short-circuits authorization when
  the current user is the built-in `root`. Because the check lives in the
  abstract base class, every subclass inherits it — including
  `RangerHiveAccessController`, etc.

  As a result, the StarRocks `root` user implicitly becomes a super-user on
  **all** external Ranger-protected catalogs and bypasses the policies
  owned by the external system (Hive Ranger, Iceberg Ranger, ...). This
  breaks the data-lake authorization boundary.

## What I'm doing:
 1. **`RangerAccessController` (base):** removed the `UserIdentity.ROOT`
     short-circuit from `hasPermission`. External Ranger controllers now
     always evaluate Ranger policies, even for the StarRocks `root` user.
  2. **`RangerStarRocksAccessController`:** override `hasPermission` and
     keep the `root` bypass there. The bypass is semantically correct only
     for the StarRocks-internal service, where `root` is the built-in
     super-user. Annotated with `@VisibleForTesting` since the override
     widens visibility to allow direct unit-test invocation.

Fixes #72198

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1.0
  - [x] 4.0.9
  - [x] 4.0.8

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/authorization/ranger/RangerAccessController.java
+++ b/fe/fe-core/src/main/java/com/starrocks/authorization/ranger/RangerAccessController.java
@@ -113,10 +113,6 @@ public abstract class RangerAccessController extends ExternalAccessController im
     protected void hasPermission(RangerAccessResourceImpl resource, UserIdentity user, Set<String> groups,
                                  PrivilegeType privilegeType)
             throws AccessDeniedException {
-        // root user bypasses Ranger authorization entirely
-        if (UserIdentity.ROOT.equals(user)) {
-            return;
-        }
         String accessType;
         if (privilegeType.equals(PrivilegeType.ANY)) {
             accessType = RangerPolicyEngine.ANY_ACCESS;
--- a/fe/fe-core/src/main/java/com/starrocks/authorization/ranger/starrocks/RangerStarRocksAccessController.java
+++ b/fe/fe-core/src/main/java/com/starrocks/authorization/ranger/starrocks/RangerStarRocksAccessController.java
@@ -14,6 +14,7 @@
 
 package com.starrocks.authorization.ranger.starrocks;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
 import com.starrocks.authorization.AccessDeniedException;
 import com.starrocks.authorization.PrivilegeType;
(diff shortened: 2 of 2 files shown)


HangyuanLiu:
I understand your concerns, and I think they are reasonable. However, different users may have different needs. You suggested adding a configuration to control this behavior, or keeping the default root privileges fully granted, and then specifically disabling root privileges.

jaogoy:
Agree with @HangyuanLiu 
Even for `RangerHiveAccessController`, `root` is also a special user. And, some security-care users will also restrict granting any access ability to tables and columns for `root` user.
So, maybe a better way is to provide two mutable fe configration items, such as `authorization_ranger_bypass_root_privilege_check_for_hive_type` and `authorization_ranger_bypass_root_privilege_check_for_sr_type` with default values `true` (for compatibility). And with your code to control them in `RangerHiveAccessController` and `RangerStarRocksAccessController` separately.

jaogoy:
@liuzhao-lz Any progress about it?

PR #72200, opened 2026-04-27
[auto-perf-opt] Admission control on PK-index cold rebuilds

luohaha:
## Summary

Adds a per-CN cap on concurrent `LakePersistentIndex::load_from_lake_tablet`
cold-rebuild executions, gated by new BE config
`pk_index_cold_rebuild_max_concurrent` (default 8; `<=0` disables).

## Motivation

In cold-burst scenarios (BE restart, scale-out, first publish after a long
idle window) the publish-version pool can dispatch tens of cold rebuilds in
the same ~100 ms window. Each rebuild's wall time is dominated by
`multi_get_us` + `rebuild_index_del_cost_us` — both OSS-RTT- and CPU-bound.
Running 30+ concurrently contends on local CPU and OSS bandwidth, inflating
the per-task wall time and producing a visible 7-11 s upsert Max tail in
publish-version traces (observed on shared-data clusters with PK-index
tables, 6 CN × ~30 tablets/burst).

## Mechanism

A process-wide `std::mutex` + `condition_variable` gates entry to the
function. Critically this is a plain CV wait — **NOT** a threadpool token
wait — so it is safe to call from any thread, including:

- Publish-version pool workers (typical caller path).
- `cloud_native_pk_index_execution` workers (inner pool, reachable via
  `_apply_thread_pool` replay during BE startup).
- BE startup `_apply_thread_pool` replay itself.

A token-based gate would risk pool self-submit `LOG(FATAL)` via
`ThreadPool::check_not_pool_thread_unlocked`; the plain CV is unaffected.

On acquire, wait time is recorded in TRACE counter
`cold_rebuild_admission_wait_us` so post-run analysis can quantify
admission contention. If the gate never fires under cold burst, the counter
stays 0 and admission control is effectively a no-op.

## Default tuning

`pk_index_cold_rebuild_max_concurrent = 8`. Each rebuild is naturally
CPU-bound at ~1-3 active cores per task; on a 16 vCPU CN, 8 concurrent
rebuilds leaves headroom while strictly limiting the OSS bandwidth
contention that drives the tail. Operators can tune via
`set config pk_index_cold_rebuild_max_concurrent` at runtime; set `<= 0`
to revert to legacy behavior.

## Files

- `be/src/common/config.h` — new `pk_index_cold_rebuild_max_concurrent` config.
- `be/src/common/config_fwd_headers_manifest.json` — register config.
- `be/src/common/config_primary_key_fwd.h` — regenerated by
  `build-support/gen_config_fwd_headers.py`.
- `be/src/storage/lake/lake_persistent_index.cpp` — admission gate on
  `load_from_lake_tablet` entry, anonymous-namespace `ColdRebuildAdmission`
  singleton.

## Test plan

- [ ] Build BE on TSP.
- [ ] Deploy on the realtime-benchmark 6 CN shared-data test cluster.
- [ ] Trigger a cold burst (BE restart, then immediately run upsert
      workload). Compare cold-burst upsert P99 / Max distribution against a
      pre-PR baseline.
- [ ] Inspect `cold_rebuild_admission_wait_us` in publish-version traces
      to confirm the gate fires when expected.
- [ ] Verify `set config pk_index_cold_rebuild_max_concurrent = 0` reverts
      to legacy behavior at runtime (no BE restart needed).

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -481,6 +481,19 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // The maximum number of memtables for pk index in shared-data mode.
 CONF_mInt32(pk_index_memtable_max_count, "2");
+// Per-CN admission cap on concurrent cold rebuilds in `LakePersistentIndex::load_from_lake_tablet`.
+// During cold-burst scenarios (BE restart, scale-out, first publish after a long idle window) the
+// publish-version pool can dispatch tens of cold rebuilds in the same ~100 ms window. Each rebuild
+// is OSS-RTT- and CPU-bound (multi_get + del_cost dominate the per-task wall time); when they all
+// run together they contend on local CPU and OSS bandwidth, which inflates the per-task time and
+// produces a tail of 7-11 s upsert Max events. Capping concurrency to a small N keeps each admitted
+// task with enough dedicated CPU/IO bandwidth to finish quickly, smoothing the burst from a
+// high-peak/short-duration shape into a lower-peak/longer-duration one. Set <= 0 to disable
+// admission control entirely (legacy behavior). Backed by a process-wide `std::mutex` +
+// `condition_variable` (NOT a threadpool token), so callers from `cloud_native_pk_index_execution`
+// or other internal pools cannot deadlock. Wait time is reported via TRACE counter
+// `cold_rebuild_admission_wait_us`.
+CONF_mInt32(pk_index_cold_rebuild_max_concurrent, "8");
 // The maximum wait flush timeout for pk index memtable in shared-data mode, in milliseconds.
 CONF_mInt64(pk_index_memtable_max_wait_flush_timeout_ms, "30000");
 // The parameters for pk index size-tiered compaction strategy.
--- a/be/src/common/config_fwd_headers_manifest.json
+++ b/be/src/common/config_fwd_headers_manifest.json
@@ -713,6 +713,7 @@
         "pk_index_parallel_execution_min_rows",
         "pk_index_memtable_max_count",
         "pk_index_memtable_max_wait_flush_timeout_ms",
+        "pk_index_cold_rebuild_max_concurrent",
         "pk_index_size_tiered_min_level_size",
         "pk_index_size_tiered_level_multiplier",
         "pk_index_size_tiered_max_level",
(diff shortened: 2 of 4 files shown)


luohaha:
## iter-069 validation

Deployed PR-A on clean upstream/main + cleared snapshot dir; ran the standard 999_1gb_1_1tb upsert workload with full Prepare (MAX_CS=9 settled in 120s) + 20-min Test.

**Gate trigger metrics** (publish-trace, 13:02:30→13:23:00, 6 CNs, 577K traces):
- 338 cold-rebuild events (0.1% of all events) — all carry the gate field.
- **76.9% of cold rebuilds had nonzero `cold_rebuild_admission_wait_us`** (260/338).
- Wait p50 = 1.1s, p99 = 3.3s, max = 3.4s.

So the gate fires on every cold rebuild and queues 77% of them past cap=8 — exactly the herd it's designed to absorb.

**Headline metrics**:
- upsert P99 = 1567 ms (vs iter-064 1504, +4%, comparable)
- upsert Max = 14047 ms (vs iter-064 11187, +25%; within ±50% variance band)
- 0 errors

**Why Max didn't drop**: per-rebuild work itself is the new binder. Top slow event tablet 10661, cost 10.7s = aw 2.6s + multi_get 3.4s + rebuild_del 3.2s + init_sst 2.7s. Each rebuild's serial OSS-RTT chain is 5–8s, so cap=8 + 6s/rebuild = ~30s burst window for the 338-event burst — exactly what we observe (all 48 events ≥5s landed in 13:02:59–13:03:27).

vdb util peak 63.7% (.129) during cold-burst minute, await max 11.4ms (.128) — below the 70% guardrail. CPU+OSS RTT is the binder, not local IO.

**Conclusion**: PR-A's admission gate is functioning correctly and provides real backpressure (without it, 30+ tablets would all hit OSS in the same 100ms window). However, PR-A alone cannot reach the 3s Max target on this workload. Per the operator's iter-068 decision tree, next iteration pivots to PR-B (within-fileset parallel multi_get re-author with separate thread pool).

Recommend keeping this PR open as-is for the backpressure value and merging once paired with PR-B's within-rebuild speedup.

luohaha:
Closing — exploratory PR from an earlier auto-perf-opt run (`my-rt`). Admission-control direction was not pursued in the follow-up my-rt-2 run, which instead pursued allocator + thread-pool restructuring (validated in #72434, #72460, #72503, #72554, #72654, #72674). This older PR is no longer needed.

PR #72203, opened 2026-04-27
[auto-perf-opt] Parallelize within-fileset multi_get on cold rebuild

luohaha:
## Why

`PersistentIndexSstableFileset::multi_get` is called from
`LakePersistentIndex::get_from_sstables` during cold rebuild and steady-state
upserts. When a fileset contains N independent sstables (non-overlapping key
ranges), the per-sstable lookups currently run serially. Each
`PersistentIndexSstable::multi_get` is an OSS read whose RTT cannot be hidden
by a single thread.

On the rt-999_1gb_1_1tb workload's cold-restart burst (PL2 ESSD), prior
iterations measured `multi_get_us` ~2–3 s of serial OSS RTT per cold rebuild;
338 cold rebuilds in a 30 s burst produce 5–8 s of per-tablet rebuild work
that dominates upsert tail (Max ≥ 10 s). PR #72200 (admission control) gates
concurrency but cannot help once a tablet is admitted — the per-rebuild work
itself is the binder. Parallelizing across the N sstables in a fileset
attacks that work directly.

## What this PR does

Fans out the inner per-sstable loop in `PersistentIndexSstableFileset::multi_get`
across a **NEW dedicated thread pool** `cloud_native_pk_index_sst_fanout`,
distinct from the existing `cloud_native_pk_index_execution` pool.

### Why a separate pool

The parent caller frequently runs on the existing
\`cloud_native_pk_index_execution\` pool (e.g. via
\`ParallelPublishContext::token\` in \`lake_persistent_index.cpp\`).
Submitting fanout work back into that pool and then \`token->wait()\`-ing
on it would trigger \`ThreadPool::check_not_pool_thread_unlocked\` LOG(FATAL)
— the same reentrance trap that took down BEs in earlier (now-abandoned)
parallelization attempts. The new pool's workers never re-enter the parent
pool, and the parent waits via a plain \`mutex+condition_variable\` counter
rather than a token, which sidesteps the pool-self-submit check entirely.

### New BE configs

| Config | Default | Purpose |
|---|---|---|
| \`enable_pk_index_sstable_fanout\` | \`true\` | Master toggle |
| \`pk_index_sstable_fanout_threadpool_max_threads\` | \`0\` (= CPU/2) | Pool size |
| \`pk_index_sstable_fanout_min_ssts\` | \`2\` | Don't fanout 1-sstable filesets |

### New trace counters

| Counter | Meaning |
|---|---|
| \`fileset_get_fanout_us\` | Wall-clock spent in parallel section |
| \`fileset_get_fanout_tasks\` | Number of sstables fanned out per call |
| \`fileset_get_fanout_inline_us\` | Fallback time when pool submit failed |

## Correctness

- Per-sstable key index sets are **disjoint by construction** in step 1
  (each key falls in exactly one sstable's range via std::map upper_bound),
  so each task writes to a disjoint slice of \`values[]\` — no synchronization
  needed for the output buffer.
- The shared \`found_key_indexes\` (std::set) is accumulated locally per task
  and merged under a mutex.
- Status uses first-error-wins. All in-flight tasks are awaited before
  returning, so the lambda captures of stack references remain valid.
- If \`submit_func\` fails (pool full / shutdown), the task runs inline on
  the caller thread and \`remaining\` is still decremented, so the wait
  loop terminates correctly.

## Expected impact

- Cold-burst \`multi_get_us\` ~2–3 s → < 1 s when fileset has ≥ 2 sstables.
- Cold-burst Max should drop from ~14 s (iter-069 baseline on PR #72200) to
  the 3–5 s range. Independent of #72200; the two PRs are complementary
  (admission control caps concurrency; this PR speeds up each admitted
  rebuild).

## Standalone

No dependency on the abandoned LakePersistentIndex snapshot stack
(#72166 / #72167 / #72169 / #72170 / #72171 / #72173 / #72174 / #72175 /
#72176, all closed without merge). Branched off clean \`upstream/main\`,
single-motivation.


luohaha:

PR #72215, opened 2026-04-27
[Enhancement] Introduce a cache for avro schema in shadowed partitionData on partition load

GavinMar:
## Why I'm doing:

Profiling iceberg partition load (`IcebergCatalog#getPartitions` on tables with many partitions) showed ~63% of allocations in `AvroSchemaUtil.convert` — Jackson `ObjectMapper` plus `JsonNode` tree — invoked from `PartitionData(Types.StructType)`. `PartitionsTable.partitions()` constructs one `PartitionData` per partition row, so the chain runs N times and pinpoints schema construction as the dominant allocator.

The key observation is that `partitionDataSchema(Types.StructType)` is a **pure function of its `StructType` argument** — it does not depend on the the partition values, or the row count. Inside a single `PartitionsTable` scan, all N `PartitionData` instances are constructed with the *same* `StructType` (Iceberg derives it from the table's unified `Partitioning.partitionType(table)`), so, we currently rebuild the identical Avro schema once per row, this is a complete waste of CPU and memory. And that is the redundancy the cache eliminates: schema construction becomes O(1) per partition struct shape per FE lifetime, instead of O(N) per refresh.

We also considered switching the per-partition construction to a streaming model (process partitions one-by-one instead of materializing the whole `Map<String, Partition>` in memory before returning), but did not pursue it for two reasons:

- **Engineering cost is high.** The current path delegates to Iceberg's own `PartitionsTable#newScan().planFiles()`, which produces a `CloseableIterable<FileScanTask>`. Turning that into a true streaming pipeline means re-implementing partition aggregation and rewiring all downstream consumers (split assignment, stats collection, the shape of `partitionCache`'s value type, MV refresh paths, etc.). The blast radius is far larger than this localized fix.
- **Streaming does not actually fix the dominant cost.** Even in a streaming layout we would still call `new PartitionData(StructType)` per partition row and so still pay the full `AvroSchemaUtil.convert` cost N times. Caching the schema removes the per-row cost regardless of whether the outer loop materializes or streams

Also, it is currently hard to tell which (table, snapshot) is responsible when partition load is slow, because no log line records the partition count or related stats per refresh.

## What I'm doing:

1. In the project's shadowed `org.apache.iceberg.PartitionData`, memoize `(Avro Schema, schema.toString())` by partition `Types.StructType` using a Caffeine cache (`maximumSize=1024`). The Avro schema is a pure function of the partition struct, so sharing across instances is safe; the cache is bounded so unfamiliar struct shapes cannot pin schemas indefinitely.  A new JUnit5 test file (`PartitionDataTest`) covers cache hits across equal struct types, schema/stringSchema consistency, copy-constructor sharing, Avro-reflection path, and Java-serialization round-trip.

   The cache itself is small. Each entry holds an Avro `Schema.RecordSchema` tree plus its stringified JSON: ~0.6–0.8 KB for a single-field BIGINT spec, ~3–5 KB for a 5-field mixed spec (date / string / decimal / int / bigint), ~15–20 KB for very wide specs. With `maximumSize=1024` the worst-case ceiling is in the low-MB range; in practice an FE holds at most a few hundred distinct `Types.StructType`s, so the actual footprint stays in the few-hundred-KB range.

2. In `CachingIcebergCatalog`'s partition `CacheLoader.load(...)`, emit a single INFO line on every (table, snapshot) cache miss with: catalog, db.table, snapshotId, partitions count, dataFiles / deleteFiles (from snapshot summary), spec count, partition field count. Frequency is bounded by Caffeine's per-key loader contract (at most one invocation per key per absence period), so this cannot flood logs. Refresh paths (background, manual, query cold-start, MV-rewrite warm-up) all converge here, so this single log point covers them.

## Test results
During my local test for a large number of partitions, the cache can help to avoid the full gc issue and decrease the refresh time from 100s+ to ~7s.

Before:
image

After:
image

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
@@ -41,6 +41,8 @@ import org.apache.iceberg.ManifestFile;
 import org.apache.iceberg.PartitionData;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SnapshotSummary;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.StarRocksIcebergTableScan;
 import org.apache.iceberg.StructLike;
@@ -54,6 +56,7 @@ import org.apache.logging.log4j.Logger;
 import org.apache.parquet.Strings;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -153,7 +156,22 @@ public class CachingIcebergCatalog implements IcebergCatalog {
                                             .setSrTableName(key.tableName)
                                             .setCatalogTableName(key.tableName)
                                             .setNativeTable(nativeTable).build();
-                            return delegate.getPartitions(icebergTable, key.snapshotId, null);
+                            Snapshot snapshot = nativeTable.snapshot(key.snapshotId);
+                            Map<String, String> summary =
+                                    (snapshot != null && snapshot.summary() != null)
+                                            ? snapshot.summary() : Collections.emptyMap();
+                            Map<String, Partition> partitions =
+                                    delegate.getPartitions(icebergTable, key.snapshotId, null);
+                            LOG.info("Loaded iceberg partitions: catalog={}, table={}.{}, snapshot={}, "
+                                            + "partitions={}, dataFiles={}, deleteFiles={}, specs={}, "
+                                            + "partitionFields={}",
+                                    catalogName, key.dbName, key.tableName, key.snapshotId,
+                                    partitions.size(),
+                                    summary.getOrDefault(SnapshotSummary.TOTAL_DATA_FILES_PROP, "?"),
+                                    summary.getOrDefault(SnapshotSummary.TOTAL_DELETE_FILES_PROP, "?"),
+                                    nativeTable.specs().size(),
+                                    nativeTable.spec().fields().size());
+                            return partitions;
                         }
                     });
         long dataFileCacheSize = Math.round(Runtime.getRuntime().maxMemory() *
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
@@ -74,6 +74,9 @@ public class CachingIcebergCatalog implements IcebergCatalog {
     private static final Logger LOG = LogManager.getLogger(CachingIcebergCatalog.class);
     public static final long NEVER_CACHE = 0;
     public static final long DEFAULT_CACHE_NUM = 100000;
+    // Only emit the partition-load INFO line when a refresh exceeds this many partitions, so the log
+    // remains useful for diagnosing slow loads on large tables without flooding for normal-sized ones.
+    private static final int PARTITION_LOAD_LOG_THRESHOLD = 10000;
     private static final ThreadLocal<ConnectContext> TABLE_LOAD_CONTEXT = new ThreadLocal<>();
     private final String catalogName;
     private final IcebergCatalog delegate;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

Youngwb (dismissed):

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
```diff
@@ -153,7 +156,22 @@ public Map<String, Partition> load(IcebergTableName key) throws Exception {
...
                                             .setCatalogTableName(key.tableName)
                                             .setNativeTable(nativeTable).build();
-                            return delegate.getPartitions(icebergTable, key.snapshotId, null);
+                            Snapshot snapshot = nativeTable.snapshot(key.snapshotId);
+                            Map<String, String> summary =
+                                    (snapshot != null && snapshot.summary() != null)
+                                            ? snapshot.summary() : Collections.emptyMap();
+                            Map<String, Partition> partitions =
+                                    delegate.getPartitions(icebergTable, key.snapshotId, null);
+                            LOG.info("Loaded iceberg partitions: catalog={}, table={}.{}, snapshot={}, "
```
trueeyu: Will this produce so many logs?

trueeyu (dismissed):

fe/fe-core/src/main/java/org/apache/iceberg/PartitionData.java
```diff
@@ -40,8 +42,28 @@
 public class PartitionData
         implements IndexedRecord, StructLike, SpecificData.SchemaConstructable, Serializable {
 
+    // Avro schema is a pure function of the partition struct, so it is safe to share across
+    // instances. Bounded so an unfamiliar set of struct shapes cannot pin schemas forever; an
+    // evicted entry just rebuilds on next use.
+    private static final Cache<Types.StructType, CachedSchema> SCHEMA_CACHE =
```
stephen-shelby, line 48: can we use a lighter key, such as table id + spec id?

stephen-shelby (dismissed):

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
```diff
@@ -153,7 +156,22 @@ public Map<String, Partition> load(IcebergTableName key) throws Exception {
                                             .setSrTableName(key.tableName)
                                             .setCatalogTableName(key.tableName)
                                             .setNativeTable(nativeTable).build();
-                            return delegate.getPartitions(icebergTable, key.snapshotId, null);
+                            Snapshot snapshot = nativeTable.snapshot(key.snapshotId);
+                            Map<String, String> summary =
+                                    (snapshot != null && snapshot.summary() != null)
+                                            ? snapshot.summary() : Collections.emptyMap();
```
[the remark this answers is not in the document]
GavinMar: Good catch. Fixed in bf5ec0da60: when `key.snapshotId == -1` we now fall back to `nativeTable.currentSnapshot()` so both the logged snapshot id and the summary reflect the snapshot actually scanned. The cache key is left as-is (`key.snapshotId` keeps the caller's intent).

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
GavinMar: Bounded by Caffeine's per-key loader contract (at most one invocation per (table, snapshotId) per absence period), so it cannot loop. To keep the line targeted at large/slow loads only, I gated it behind a partition-count threshold in bf5ec0da60: the INFO is only emitted when `partitions.size() > 10000`. Smaller refreshes skip both the log and the snapshot/summary lookup.

fe/fe-core/src/main/java/org/apache/iceberg/PartitionData.java
GavinMar: Considered, but kept the `StructType` key. Two reasons:

1. The dominant call site is `PartitionsTable.partitions()`, which uses `Partitioning.partitionType(table)` — a unified type across all historical specs, so there isn't a single `spec_id` that maps to it. https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/Partitioning.java#L240
2. We implement the cache in the underly iceberg ParitionData module, which has no reference to table identity (it's also instantiated by Avro reflection via `SpecificData.SchemaConstructable`), so threading in `(table_id, spec_id)` would require invasive constructor-signature changes across non-partitions-table call sites too.

`StructType.equals/hashCode` is O(F) on partition fields (typically 1–5), which is already negligible vs. the `AvroSchemaUtil.convert` cost the cache eliminates.

Wenjun7J:
I think a more elegant method is to mock the PartitionsTable.java from iceberg. and maintain a temporary cache map in it. and maybe we can issue a pr for iceberg :)

we can just change the code below, `new PartitionData(keyType)` is now using type paramters to construct the PartitionData, and we can change it with schema constructor, like `new PartitionData(getSchema(keyType))`.
```
private PartitionData toPartitionData(StructLike key, Types.StructType keyType) {
      PartitionData data = new PartitionData(keyType);
...
}
```

this pr i think is all right though.

Wenjun7J (approved):

GavinMar:
@Wenjun7J Agree an upstream Iceberg PR would be a nice follow-up — that would let us drop the shadow class entirely. However, two reasons we landed on caching inside `PartitionData(StructType)`:

1. **`new PartitionData(Schema)` doesn't actually bypass the Avro conversion.** Looking at the existing constructor:
   ```java
   PartitionData(Schema schema) {
       this.partitionType = AvroSchemaUtil.convert(schema).asNestedType().asStructType();
       this.stringSchema = schema.toString();
       ...
   }
   ```
   You still pay `AvroSchemaUtil.convert(Schema)` (Schema→Type, lighter direction than Type→Schema but still O(F) per row) and `schema.toString()` (Jackson serialization) on every row. With the current cache, both are 0 per row — the constructor just does field assignments.

2. **Wider coverage.** Caching inside `PartitionData(StructType)` benefits every call site: manifest reading, `BaseFile.partition()`, Avro reflection via `SpecificData.SchemaConstructable`, etc. A cache localized to `PartitionsTable.toPartitionData` would only catch the partitions-table scan path.

Wenjun7J:
@GavinMar Yep, you are right.  for 1, we still need to add api for PartitionData. for 2 this will introduce a cache size parameter, but it's ok

stephen-shelby (approved):

Youngwb (approved):

GavinMar:
@mergify backport branch-3.5

PR #72218, opened 2026-04-27
[BugFix] ConsistencyChecker: relax DB READ to per-table READ in periodic scans

kevincai:
ConsistencyChecker's two periodic paths held a DB-wide READ lock across long scans, stalling concurrent DDL on every unrelated table in the DB for the whole duration of each pass.

1. checkTabletMetaConsistency iterates every tablet reported by every backend and, per tablet, looks up exactly one (db, table, partition, index, tablet) chain. Swap the DB READ for IS + table READ on the one table being validated.

2. chooseTablets walks db -> table -> partition -> index -> tablet to pick up to MAX_JOB_NUM tablets for consistency checking. The original code held DB READ for the entire hierarchical walk of a database, which on large DBs can pin DDL for a long time. Refactor to: - Snapshot the table list lock-free. Database.idToTable is a ConcurrentHashMap, so getTables() returns a thread-safe weakly-consistent ArrayList. Holding a DB-level lock just for this call would be theatre: it would be released before the per-table walk and so the snapshot is stale immediately afterwards regardless. The per-table re-fetch below is what actually guards against drops. - Pre-filter OLAP + NORMAL-state tables. - For each surviving table, take IS + table READ, then re-fetch the table from the metastore and re-check state under the lock. The pre-filter / snapshot reads happen without a lock, so in the gap between snapshot and per-table lock acquisition the table may have been dropped (DROP TABLE / DROP DATABASE take DB WRITE) or may have transitioned out of NORMAL state. Skipping dropped or non-NORMAL tables here avoids walking a concurrently-mutated structure and avoids enqueueing consistency jobs for tablets already being torn down.

The per-table walk that runs inside the per-table READ lock is extracted into a private helper chooseTabletsFromTable so the outer chooseTablets stays compact. The helper takes a Database, the already-locked OlapTable, and the chosenTablets accumulator, and returns true iff MAX_JOB_NUM has been reached so the caller can stop. No behavior difference from the extraction; it is a structural cleanup bundled here because both changes touch the same method body.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/consistency/ConsistencyChecker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/consistency/ConsistencyChecker.java
@@ -180,6 +180,15 @@ public class ConsistencyChecker extends FrontendDaemon {
             List<Long> tabletIds = tabletInvertedIndex.getTabletIdsByBackendId(backendId);
             backendTabletNumReport.put(backendId, new Pair<>(0L, 0L));
 
+            // TODO: group tablets by (dbId, tableId) before this loop and process each
+            // group under a single lockTableWithIntensiveDbLock acquisition. Today we
+            // acquire/release the per-table lock once per tablet, plus one getDb +
+            // getTable lookup per tablet. With grouping, that drops to once per
+            // (dbId, tableId) group: O(tables) instead of O(tablets). For backends
+            // reporting tens of thousands of tablets clustered on far fewer tables --
+            // the typical case -- this is a 10x-100x reduction in LockManager calls.
+            // Even in the pathological 1-tablet-per-table case the cost is essentially
+            // the same, so this is strictly beneficial in realistic deployments.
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

starrocks-xupeng (approved):

wyb (approved):

gengjun-git:
@codex review

xiangguangyxg (approved):

gengjun-git (approved):

PR #72270, opened 2026-04-28
[Enhancement] Add getTabletMetadata RPC for cn-free tablet creation

starrocks-xupeng:
Introduce the FE-side RPC that an upcoming CN-Free Tablet Creation feature will call to fetch tablet metadata when DDL skips writing version 1 metadata to object storage.

This change is FE infrastructure only and changes no existing behavior:
  * No caller of the new RPC ships in this PR. The CN-side consumer and fallback wire-up land in a follow-up.
  * The user-visible switch (table property cn_free_tablet_creation and DDL skip in TabletTaskExecutor) lands in a later PR.

The request includes a version field so callers can ask for any tablet metadata version, but only version 1 (the initial empty metadata produced by tablet creation) is implemented. Other versions return NOT_IMPLEMENTED_ERROR; they will be supported when a future change extends the response with the fields needed (rowsets, historical schemas, etc.).

For version 1 the handler resolves the tablet via TabletInvertedIndex and returns the schema and configuration needed to construct the initial TabletMetadataPB on CN: compression type/level, persistent index settings, flat-json config, compaction strategy, GTID, and per-tablet ranges.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/service/FrontendServiceImpl.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/FrontendServiceImpl.java
@@ -53,10 +53,12 @@ import com.starrocks.catalog.CatalogUtils;
 import com.starrocks.catalog.Column;
 import com.starrocks.catalog.Database;
 import com.starrocks.catalog.ExpressionRangePartitionInfo;
+import com.starrocks.catalog.FlatJsonConfig;
 import com.starrocks.catalog.InternalCatalog;
 import com.starrocks.catalog.ListPartitionInfo;
 import com.starrocks.catalog.LocalTablet;
 import com.starrocks.catalog.MaterializedIndex;
+import com.starrocks.catalog.MaterializedIndexMeta;
 import com.starrocks.catalog.MaterializedView;
 import com.starrocks.catalog.OlapTable;
 import com.starrocks.catalog.Partition;
@@ -65,6 +67,7 @@ import com.starrocks.catalog.PartitionKey;
 import com.starrocks.catalog.PhysicalPartition;
 import com.starrocks.catalog.RangePartitionInfo;
 import com.starrocks.catalog.Replica;
+import com.starrocks.catalog.SchemaInfo;
 import com.starrocks.catalog.Table;
 import com.starrocks.catalog.TableName;
 import com.starrocks.catalog.Tablet;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

kevincai (dismissed):

wangruin (dismissed):

gensrc/thrift/FrontendService.thrift
```diff
@@ -2295,6 +2295,45 @@ struct TBatchGetTableSchemaResponse {
...
+    1: optional Status.TStatus status;
+    2: optional AgentService.TTabletSchema schema;
+    3: optional bool enable_persistent_index;
+    4: optional AgentService.TPersistentIndexType persistent_index_type;
+    5: optional AgentService.TCompactionStrategy compaction_strategy;
+    6: optional AgentService.TFlatJsonConfig flat_json_config;
+    7: optional map<i64, Types.TTabletRange> tablet_ranges;
+    8: optional i64 gtid;
+    9: optional Types.TCompressionType compression_type;
+    10: optional i32 compression_level;
```
banmoy, line 2326: These fields are same with those in TCreateTabletReq, which means we need to maintain them in two places. Could we define a struct like TTabletMetadata to hold these fields, use it both in TGetTabletMetadataResponse and TCreateTabletReq, and fill it in a common function? I think this way we will not forget to modify both

fe/fe-core/src/main/java/com/starrocks/service/FrontendServiceImpl.java
```diff
@@ -3587,6 +3595,135 @@ public TBatchGetTableSchemaResponse getTableSchema(TBatchGetTableSchemaRequest b
...
+            // 5. All tablet ranges for this index (only for range distribution)
+            if (olapTable.isRangeDistribution()) {
+                Map<Long, TTabletRange> tabletRanges = new java.util.HashMap<>();
+                for (Tablet tablet : index.getTablets()) {
+                    if (tablet.getRange() != null) {
+                        tabletRanges.put(tablet.getId(), tablet.getRange().toThrift());
+                    }
+                }
+                if (!tabletRanges.isEmpty()) {
+                    response.setTablet_ranges(tabletRanges);
```
banmoy: rpc requests are aggregated by index id on CN?

starrocks-xupeng: yeah

banmoy (approved):

wangruin (dismissed):

banmoy (approved):

wangruin (approved):

fe/fe-core/src/main/java/com/starrocks/service/FrontendServiceImpl.java
```diff
@@ -3587,6 +3596,133 @@ public TBatchGetTableSchemaResponse getTableSchema(TBatchGetTableSchemaRequest b
...
+        batchResponse.setStatus(new TStatus(OK));
+        if (batchRequest.isSetRequests()) {
+            for (TGetTabletMetadataRequest request : batchRequest.getRequests()) {
+                batchResponse.addToResponses(handleGetTabletMetadata(request));
+            }
+        }
+        return batchResponse;
+    }
+
+    private TGetTabletMetadataResponse handleGetTabletMetadata(TGetTabletMetadataRequest request) {
```
kevincai, line 3612: try to group the requests by table, so the tablets belongs to the same table can be processed under the same lock, which the meta consistency can be assured.

kevincai (approved):
leave the TODO fix  (group by table) in separate PR.

PR #72282, opened 2026-04-28
[Enhancement] Introduce a cache for avro schema in shadowed partitionData on partition load (backport #72215)

GavinMar:
## Why I'm doing:

Profiling iceberg partition load (`IcebergCatalog#getPartitions` on tables with many partitions) showed ~63% of allocations in `AvroSchemaUtil.convert` — Jackson `ObjectMapper` plus `JsonNode` tree — invoked from `PartitionData(Types.StructType)`. `PartitionsTable.partitions()` constructs one `PartitionData` per partition row, so the chain runs N times and pinpoints schema construction as the dominant allocator.

The key observation is that `partitionDataSchema(Types.StructType)` is a **pure function of its `StructType` argument** — it does not depend on the the partition values, or the row count. Inside a single `PartitionsTable` scan, all N `PartitionData` instances are constructed with the *same* `StructType` (Iceberg derives it from the table's unified `Partitioning.partitionType(table)`), so, we currently rebuild the identical Avro schema once per row, this is a complete waste of CPU and memory. And that is the redundancy the cache eliminates: schema construction becomes O(1) per partition struct shape per FE lifetime, instead of O(N) per refresh.

We also considered switching the per-partition construction to a streaming model (process partitions one-by-one instead of materializing the whole `Map<String, Partition>` in memory before returning), but did not pursue it for two reasons:

- **Engineering cost is high.** The current path delegates to Iceberg's own `PartitionsTable#newScan().planFiles()`, which produces a `CloseableIterable<FileScanTask>`. Turning that into a true streaming pipeline means re-implementing partition aggregation and rewiring all downstream consumers (split assignment, stats collection, the shape of `partitionCache`'s value type, MV refresh paths, etc.). The blast radius is far larger than this localized fix.
- **Streaming does not actually fix the dominant cost.** Even in a streaming layout we would still call `new PartitionData(StructType)` per partition row and so still pay the full `AvroSchemaUtil.convert` cost N times. Caching the schema removes the per-row cost regardless of whether the outer loop materializes or streams

Also, it is currently hard to tell which (table, snapshot) is responsible when partition load is slow, because no log line records the partition count or related stats per refresh.

## What I'm doing:

1. In the project's shadowed `org.apache.iceberg.PartitionData`, memoize `(Avro Schema, schema.toString())` by partition `Types.StructType` using a Caffeine cache (`maximumSize=1024`). The Avro schema is a pure function of the partition struct, so sharing across instances is safe; the cache is bounded so unfamiliar struct shapes cannot pin schemas indefinitely.  A new JUnit5 test file (`PartitionDataTest`) covers cache hits across equal struct types, schema/stringSchema consistency, copy-constructor sharing, Avro-reflection path, and Java-serialization round-trip.

   The cache itself is small. Each entry holds an Avro `Schema.RecordSchema` tree plus its stringified JSON: ~0.6–0.8 KB for a single-field BIGINT spec, ~3–5 KB for a 5-field mixed spec (date / string / decimal / int / bigint), ~15–20 KB for very wide specs. With `maximumSize=1024` the worst-case ceiling is in the low-MB range; in practice an FE holds at most a few hundred distinct `Types.StructType`s, so the actual footprint stays in the few-hundred-KB range.

2. In `CachingIcebergCatalog`'s partition `CacheLoader.load(...)`, emit a single INFO line on every (table, snapshot) cache miss with: catalog, db.table, snapshotId, partitions count, dataFiles / deleteFiles (from snapshot summary), spec count, partition field count. Frequency is bounded by Caffeine's per-key loader contract (at most one invocation per key per absence period), so this cannot flood logs. Refresh paths (background, manual, query cold-start, MV-rewrite warm-up) all converge here, so this single log point covers them.

## Test results
During my local test for a large number of partitions, the cache can help to avoid the full gc issue and decrease the refresh time from 100s+ to ~7s.

Before:
image

After:
image

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
@@ -38,6 +38,8 @@ import org.apache.iceberg.ManifestFile;
 import org.apache.iceberg.PartitionData;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SnapshotSummary;
 import org.apache.iceberg.StarRocksIcebergTableScan;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.TableMetadata;
@@ -50,6 +52,7 @@ import org.apache.spark.util.SizeEstimator;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


wangsimo0 (approved):

wangsimo0 (approved):

trueeyu:

PR #72330, opened 2026-04-29
[Enhancement] Support PostgreSQL-style `::` cast shorthand

### LINKED ISSUE #72301: Support shorthand cast syntax in StarRocks
## Background & Motivation

Modern analytical engines increasingly provide concise type-casting syntax to improve developer ergonomics and reduce boilerplate in SQL queries.

PostgreSQL's `::` operator is a widely adopted shorthand for type casting that is also supported by DuckDB, ClickHouse, and several other engines. It allows users to write `expr::TYPE` instead of the more verbose `CAST(expr AS TYPE)`.

For StarRocks, supporting the `::` shorthand cast would bring two key benefits:
- Cleaner and more readable SQL
  - Analytical queries frequently require explicit casts when comparing across types, normalizing decimals, or projecting JSON-extracted values. Long expressions wrapped in `CAST(... AS ...)` quickly become noisy and hard to scan.
  - The `::` operator keeps the cast visually adjacent to the expression and avoids deep nesting.
- Improved compatibility and developer experience
  - Aligning with PostgreSQL, DuckDB, and ClickHouse enhances cross-engine portability, simplifies migration of existing queries, and lowers friction for developers writing portable analytical SQL.

## Proposed Behavior

`expr::TYPE` should be interpreted as fully equivalent to `CAST(expr AS TYPE)`, with identical semantics for type coercion, overflow, and error handling.

Example:
```
SELECT '123'::INT;
SELECT col::VARCHAR FROM t;
SELECT (a + b)::DECIMAL(10, 2) FROM t;
SELECT json_value::JSON FROM t;
```
Equivalent to:
```
SELECT CAST('123' AS INT);
SELECT CAST(col AS VARCHAR) FROM t;
SELECT CAST(a + b AS DECIMAL(10, 2)) FROM t;
SELECT CAST(json_value AS JSON) FROM t;
```

The operator should:
- Bind tighter than most binary operators, matching PostgreSQL precedence, so that `a + b::INT` parses as `a + (b::INT)`.
- Support all types accepted by `CAST`, including parameterized types such as `DECIMAL(p, s)` and `VARCHAR(n)`.
- Be chainable: `expr::T1::T2` should be equivalent to `CAST(CAST(expr AS T1) AS T2)`.

## Use Cases
- Migrating PostgreSQL/DuckDB/ClickHouse queries to StarRocks with minimal rewriting
- Concise inline casts in BI/ad-hoc queries and dashboards
- JSON/string normalization pipelines where casts dominate the projection list
- Auto-generated SQL from ORMs and query builders that already emit `::` syntax

petrikoro:
## Why I'm doing:

`CAST(expr AS type)` is verbose, and users coming from PostgreSQL, DuckDB, ClickHouse, and Trino routinely reach for the `expr::type` shorthand. Tools and notebooks that target multiple engines (dbt, BI exports, generated SQL) often emit `::` casts unconditionally; today those queries fail at parse time on StarRocks, forcing a manual rewrite. Adding the shorthand removes that friction without changing semantics — the resulting plan is identical to the explicit `CAST` form.

## What I'm doing:

Add PostgreSQL-style `::` cast shorthand. `expr::type` is parsed as `CAST(expr AS type)` and produces the same AST and the same optimized plan.

Fixes https://github.com/StarRocks/starrocks/issues/72301

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


petrikoro:
@codex review

petrikoro:
@codex review

petrikoro:
@codex review

stdpain:
@wangsimo0

wangsimo0 (approved):

wangsimo0 (changes requested):
Nice direction! Thank you for contributing! A few design concerns worth closing before merge:

1. View / MV persistence & downgrade compatibility (most important). The PR description states EXPLAIN / SHOW CREATE VIEW / audit logs always emit canonical CAST(... AS ...). But what gets persisted to the catalog for CREATE VIEW v AS SELECT x::int FROM t? If the original SQL is stored, downgrades and mixed-version clusters will fail to parse view/MV bodies. Could you confirm the persisted originalDef/inlineViewDef is canonicalized, and add an end-to-end test (CREATE VIEW → SHOW CREATE VIEW) that asserts the CAST form?

2. Lexer regression check. DOUBLE_COLON is added next to AT. Could we get an explicit test that SET @x := 1 and SELECT @x := col FROM t (MySQL user-variable assignment) still parse correctly? The current negative test only covers spaced : :.

3. Hint pre-pass. The existing test only checks that select /*+ SET_VAR(query_timeout=10) */ x::int from t doesn't throw. It would be safer to assert query_timeout was actually applied — token-offset shifts from a new lexer token can silently break hint attribution.

4. Precedence corner cases not covered. Worth adding tests for:

x::int < 5 and a < b::int — to confirm < isn't greedy-eaten by the type rule.
x::array<int>[1] — element access after cast (ClickHouse needed a follow-up PR for this).
foo::int.col — dereference after cast; expect either correct parse or clean error.
-2147483648::int — the PG footgun (-(2147483648::int) overflow). Either match PG or document the divergence.
5. "PostgreSQL-style" ≠ "PostgreSQL-compatible". select 1.5::int returns 1 here (truncation) but 2 in PostgreSQL (rounding). Users copying PG SQL will get silent semantic differences. I'd recommend the docs frame this as syntax compatibility only and call out the rounding/type-alias differences explicitly. PG type aliases like int4/int8/text/bool/timestamptz will also fail with a fairly cryptic ANTLR error — a friendly hint or an explicit note in the doc would help.

6. Test surface is mostly SELECT. Sanity tests for INSERT … SELECT x::int, CREATE TABLE AS SELECT x::int, WHERE x::int > 0, JOIN ON, and subqueries would confirm the primaryExpression hook reaches all contexts.

Smaller things:

ShorthandCastPlanTest compares fragment plans but doesn't assert the textual canonicalization promise made in the docs — adding one SHOW CREATE VIEW assertion would close that loop.
Worth checking whether query cache / plan cache / audit-log dedup keys are based on AST-rebuilt SQL or raw text; if raw text, x::int and cast(x as int) will look like distinct queries.

HangyuanLiu (dismissed):

petrikoro:
@codex review

petrikoro:
> Nice direction! Thank you for contributing! A few design concerns worth closing before merge:
> 
> 1. View / MV persistence & downgrade compatibility (most important). The PR description states EXPLAIN / SHOW CREATE VIEW / audit logs always emit canonical CAST(... AS ...). But what gets persisted to the catalog for CREATE VIEW v AS SELECT x::int FROM t? If the original SQL is stored, downgrades and mixed-version clusters will fail to parse view/MV bodies. Could you confirm the persisted originalDef/inlineViewDef is canonicalized, and add an end-to-end test (CREATE VIEW → SHOW CREATE VIEW) that asserts the CAST form?
> 2. Lexer regression check. DOUBLE_COLON is added next to AT. Could we get an explicit test that SET @x := 1 and SELECT @x := col FROM t (MySQL user-variable assignment) still parse correctly? The current negative test only covers spaced : :.
> 3. Hint pre-pass. The existing test only checks that select /*+ SET_VAR(query_timeout=10) */ x::int from t doesn't throw. It would be safer to assert query_timeout was actually applied — token-offset shifts from a new lexer token can silently break hint attribution.
> 4. Precedence corner cases not covered. Worth adding tests for:
> 
> x::int < 5 and a < b::int — to confirm < isn't greedy-eaten by the type rule. x::array[1] — element access after cast (ClickHouse needed a follow-up PR for this). foo::int.col — dereference after cast; expect either correct parse or clean error. -2147483648::int — the PG footgun (-(2147483648::int) overflow). Either match PG or document the divergence. 5. "PostgreSQL-style" ≠ "PostgreSQL-compatible". select 1.5::int returns 1 here (truncation) but 2 in PostgreSQL (rounding). Users copying PG SQL will get silent semantic differences. I'd recommend the docs frame this as syntax compatibility only and call out the rounding/type-alias differences explicitly. PG type aliases like int4/int8/text/bool/timestamptz will also fail with a fairly cryptic ANTLR error — a friendly hint or an explicit note in the doc would help.
> 
> 6. Test surface is mostly SELECT. Sanity tests for INSERT … SELECT x::int, CREATE TABLE AS SELECT x::int, WHERE x::int > 0, JOIN ON, and subqueries would confirm the primaryExpression hook reaches all contexts.
> 
> Smaller things:
> 
> ShorthandCastPlanTest compares fragment plans but doesn't assert the textual canonicalization promise made in the docs — adding one SHOW CREATE VIEW assertion would close that loop. Worth checking whether query cache / plan cache / audit-log dedup keys are based on AST-rebuilt SQL or raw text; if raw text, x::int and cast(x as int) will look like distinct queries.

@wangsimo0 thanks for the detailed review. I went through the branch again and the latest commits cover the main concerns:

1. View/MV persistence: by default `CREATE VIEW`, `ALTER VIEW ... AS ...`, and `CREATE MATERIALIZED VIEW` now persist the canonical definition SQL via `enable_persist_canonical_view_sql=true`. The SQL regression adds end-to-end `SHOW CREATE VIEW` and `SHOW CREATE MATERIALIZED VIEW` checks that assert `CAST(...)`; setting the variable to `false` is kept as an explicit opt-out and the docs call out the downgrade risk.
2. Hint pre-pass: there is parser coverage that the `SET_VAR(query_timeout=10)` hint survives `x::int`, plus a planner/executor test that applies the hint and asserts the session `query_timeout` becomes `10`.
3. Precedence/corner cases: `::` now binds tighter than `||` under `PIPES_AS_CONCAT`, and tests cover `x::int < 5`, `a < b::int`, `x::array<int>[1]`, scalar dereference after cast as a clean semantic error, and `-2147483648::int` as `-CAST(2147483648 AS INT)`.
4. Broader statement contexts: the SQL regression now exercises `INSERT ... SELECT`, CTAS, `WHERE`, `JOIN ON`, and subquery usage, in addition to basic `SELECT` cases.
5. Docs: the cast docs frame `::` as shorthand for StarRocks `CAST(input AS type)` and document the view/MV canonical persistence behavior.

Two clarifications from checking the existing code paths:

- StarRocks user variables use `SET @var = ...` and `SELECT @var`; MySQL-style `:=` assignment is not currently part of the grammar, so I did not add a passing `:=` regression. `DOUBLE_COLON` is a distinct lexer token and the existing `@` user-variable paths remain separate.
- Audit output still preserves the original single-statement SQL text by design, but the audit digest is AST-based (`AstToSQLBuilder.toDigest`) and query cache keys are based on normalized plan fragments, so `x::int` and `CAST(x AS INT)` do not require a separate raw-text cache-key change here.

petrikoro:
@codex review

wangsimo0:
Thanks for the very thorough follow-up! Really appreciate it, and thank you for adding those tests. Most of the parser/test concerns look addressed.

One remaining product concern: enable_persist_canonical_view_sql=true now changes the persisted original definition for all CREATE VIEW / ALTER VIEW / CREATE MATERIALIZED VIEW statements, not only definitions that use ::. That changes SHOW CREATE output for existing-style SQL as well, including comments/formatting/raw text preservation, and seems broader than the shorthand-cast feature requires.

After thinking more about the downgrade story, I do not think we need to over-optimize for a perfect downgrade path here. Persisting canonical SQL avoids old-version parser failures, but it also changes SHOW CREATE fidelity and drops comments/formatting. That tradeoff is acceptable only for definitions that use the new `::` syntax, not for all view/MV definitions. And a lot of them rely on the hint to give them the business meaning. So maybe we just need a very detailed downgrade warning. 

Also please make the docs explicit that this is syntax compatibility with StarRocks CAST semantics, not PostgreSQL cast semantic compatibility,(since we don't support those data types), and sync the JA docs for the new variable/persistence behavior.

petrikoro:
@codex review

petrikoro:
> Thanks for the very thorough follow-up! Really appreciate it, and thank you for adding those tests. Most of the parser/test concerns look addressed.
> 
> One remaining product concern: enable_persist_canonical_view_sql=true now changes the persisted original definition for all CREATE VIEW / ALTER VIEW / CREATE MATERIALIZED VIEW statements, not only definitions that use ::. That changes SHOW CREATE output for existing-style SQL as well, including comments/formatting/raw text preservation, and seems broader than the shorthand-cast feature requires.
> 
> After thinking more about the downgrade story, I do not think we need to over-optimize for a perfect downgrade path here. Persisting canonical SQL avoids old-version parser failures, but it also changes SHOW CREATE fidelity and drops comments/formatting. That tradeoff is acceptable only for definitions that use the new `::` syntax, not for all view/MV definitions. And a lot of them rely on the hint to give them the business meaning. So maybe we just need a very detailed downgrade warning.
> 
> Also please make the docs explicit that this is syntax compatibility with StarRocks CAST semantics, not PostgreSQL cast semantic compatibility,(since we don't support those data types), and sync the JA docs for the new variable/persistence behavior.

@wangsimo0 agreed, thanks for calling out the product tradeoff. I went through the branch again and narrowed the latest commits around that feedback:

1. Removed `enable_persist_canonical_view_sql` and the related system-variable docs. `CREATE VIEW`, `ALTER VIEW ... AS ...`, and `CREATE MATERIALIZED VIEW` now persist the original query text again, so existing-style SQL keeps its `SHOW CREATE` fidelity, including comments and formatting.
2. Updated the cast docs to say `::` is equivalent to StarRocks `CAST(input AS type)` and uses StarRocks cast semantics, so the documented promise is syntax compatibility rather than PostgreSQL semantic compatibility.
3. Added the downgrade warning in the EN, ZH, and JA cast docs: if a view or materialized view DDL uses `::`, the saved original definition can contain `::`, and that definition is not backward compatible with older StarRocks versions or tools that cannot parse the shorthand syntax.

wangsimo0:
LGTM, thanks a lot!!

wangsimo0 (approved):

HangyuanLiu (approved):

PR #72331, opened 2026-04-29
[BugFix] Preserve CompactionTaskStats in be_cloud_native_compactions PROFILE for parallel subtasks

meegoo:
## Why I'm doing:
For running parallel compaction subtasks, the PROFILE column was overwritten with only subtask metadata (subtask_id, input_rowsets, input_bytes, is_parallel_subtask), discarding the CompactionTaskStats counters that to_json_stats() produced. Add a new CompactionTaskStats::to_json_stats_with_subtask_metadata() that emits the same JSON layout as to_json_stats() with the subtask metadata fields appended, and use it from TabletParallelCompactionManager::list_tasks() so both pieces are visible in information_schema.be_cloud_native_compactions.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/compaction_task_context.cpp
+++ b/be/src/storage/lake/compaction_task_context.cpp
@@ -118,15 +118,13 @@ std::string CompactionTaskStats::to_json_stats() const {
     return serialize(root);
 }
 
-std::string CompactionTaskStats::to_json_stats_with_subtask_metadata(int32_t subtask_id, size_t input_rowsets,
-                                                                     int64_t input_bytes) const {
+std::string CompactionTaskStats::to_json_stats_with_subtask_metadata(int32_t subtask_id, size_t input_rowsets) const {
     rapidjson::Document root;
     root.SetObject();
     fill_stats_fields(root, *this);
     auto& allocator = root.GetAllocator();
     root.AddMember("subtask_id", rapidjson::Value(subtask_id), allocator);
     root.AddMember("input_rowsets", rapidjson::Value(static_cast<int64_t>(input_rowsets)), allocator);
-    root.AddMember("input_bytes", rapidjson::Value(input_bytes), allocator);
     root.AddMember("is_parallel_subtask", rapidjson::Value(true), allocator);
     return serialize(root);
 }
--- a/be/src/storage/lake/compaction_task_context.cpp
+++ b/be/src/storage/lake/compaction_task_context.cpp
@@ -84,31 +84,50 @@ CompactionTaskStats CompactionTaskStats::operator-(const CompactionTaskStats& th
     return diff;
 }
 
-std::string CompactionTaskStats::to_json_stats() {
-    rapidjson::Document root;
-    root.SetObject();
+static void fill_stats_fields(rapidjson::Document& root, const CompactionTaskStats& s) {
     auto& allocator = root.GetAllocator();
-    // add stats
-    root.AddMember("read_local_sec", rapidjson::Value(io_ns_read_local_disk / TIME_UNIT_NS_PER_SECOND), allocator);
-    root.AddMember("read_local_mb", rapidjson::Value(io_bytes_read_local_disk / BYTES_UNIT_MB), allocator);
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

luohaha (dismissed):

be/src/storage/lake/compaction_task_context.cpp
```diff
@@ -84,31 +84,50 @@ CompactionTaskStats CompactionTaskStats::operator-(const CompactionTaskStats& th
...
+
+std::string CompactionTaskStats::to_json_stats_with_subtask_metadata(int32_t subtask_id, size_t input_rowsets,
+                                                                     int64_t input_bytes) const {
+    rapidjson::Document root;
+    root.SetObject();
+    fill_stats_fields(root, *this);
+    auto& allocator = root.GetAllocator();
+    root.AddMember("subtask_id", rapidjson::Value(subtask_id), allocator);
+    root.AddMember("input_rowsets", rapidjson::Value(static_cast<int64_t>(input_rowsets)), allocator);
+    root.AddMember("input_bytes", rapidjson::Value(input_bytes), allocator);
```
starrocks-xupeng: what's the difference between input bytes and input files size

starrocks-xupeng: i think we only need 1?

be/src/storage/lake/compaction_task_context.cpp
```diff
@@ -84,31 +84,50 @@ CompactionTaskStats CompactionTaskStats::operator-(const CompactionTaskStats& th
...
+    root.AddMember("is_parallel_subtask", rapidjson::Value(true), allocator);
```
starrocks-xupeng, line 128: if it's subtask, it's always parallel, so we can remove is_parallel_subtask?

be/src/storage/lake/compaction_task_context.cpp
meegoo: Good question — these two fields come from different sources and represent different concepts:

- `input_file_size` is part of `CompactionTaskStats` and is measured **at runtime** during execution (actual bytes read from storage).
- `input_bytes` comes from `SubtaskInfo` and is the **planned input size** set at scheduling time, before execution starts.

Keeping both lets operators compare planned vs. actual I/O, which is useful for detecting plan deviations (e.g., if actual reads significantly exceed the estimate, it may indicate rowset layout changes or scheduling inaccuracies).

---
_Generated by [Claude Code](https://claude.ai/code)_

be/src/storage/lake/compaction_task_context.cpp
meegoo: That's a fair point — a non-negative `subtask_id` already implies parallel execution. The reason `is_parallel_subtask` is kept is purely for **observability ergonomics**: it lets users filter PROFILE entries with a simple boolean check (e.g. `json_extract(profile, '$.is_parallel_subtask') = true`) without having to inspect `subtask_id` presence or value. It also serves as a forward-compatibility guard: if a parallel compaction framework ever produces a single subtask (`subtask_id = 0`, parallelism = 1), the flag still unambiguously marks it as belonging to the parallel path.

---
_Generated by [Claude Code](https://claude.ai/code)_

be/src/storage/lake/compaction_task_context.cpp
starrocks-xupeng: we already have read_local_mb and read_remote_mb

starrocks-xupeng (approved):

sevev (approved):

luohaha (approved):

PR #64760, opened 2025-10-30
[Enhancement] support split topn agg

Seaven:
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

PR #72332, opened 2026-04-29
[Enhancement] Enable TopN runtime filter pushdown for Iceberg table aggregation

dirtysalt:
## Why I'm doing:

Single-table Iceberg aggregation queries with `GROUP BY key ORDER BY same_key LIMIT N` could miss TopN runtime filter pushdown because the optimizer rule only matched plans that already contained a partial TopN above the global aggregate.

## What I'm doing:

- extend `PushDownTopNToPreAggRule` to also handle an unsplit final TopN
- synthesize `final TopN -> partial TopN` when needed before pushing TopN below global agg
- preserve the existing behavior for already-split partial TopN plans

Take following sql for example

```
explain verbose SELECT l_orderkey, SUM(l_extendedprice * (1 - l_discount)) AS revenue
  FROM iceberg.zz_iceberg_tpch_sf100_iceberg_parquet_lz4.lineitem
  GROUP BY l_orderkey
  ORDER BY l_orderkey DESC
  LIMIT 10;
```

**With this pr, runtime filter resides on scan node. And the execution time reduces from 12s down to 1s**

```
PLAN FRAGMENT 2(F00)
  Fragment Cost: 6.800000000000001

  Input Partition: RANDOM
  OutPut Partition: HASH_PARTITIONED: 1: l_orderkey
  OutPut Exchange Id: 04

  3:TOP-N
  |  order by: [1, BIGINT, true] DESC
  |  offset: 0
  |  limit: 10
  |  cardinality: 1
  |  
  2:AGGREGATE (update serialize)
  |  STREAMING
  |  aggregate: sum[([19: expr, DECIMAL128(31,4), true]); args: DECIMAL128; result: DECIMAL128(38,4); args nullable: true; result nullable: true]
  |  group by: [1: l_orderkey, BIGINT, true]
  |  streaming preaggregation mode: force_preaggregation
  |  build runtime filters:
  |  - filter_id = 0, build_expr = (<slot 1> 1: l_orderkey), remote = false  // <--- HERE
  |  cardinality: 1
  |  
  1:Project
  |  output columns:
  |  1 <-> [1: l_orderkey, BIGINT, true]
  |  19 <-> cast([6: l_extendedprice, DECIMAL64(15,2), true] as DECIMAL128(15,2)) * cast(1 - [7: l_discount, DECIMAL64(15,2), true] as DECIMAL128(16,2))
  |  cardinality: 1
  |  
  0:IcebergScanNode
     TABLE: zz_iceberg_tpch_sf100_iceberg_parquet_lz4.lineitem
     TABLE VERSION: Snapshot@(8826450500175032053)
     cardinality=1
     avgRowSize=4.0
     dataCacheOptions={populate: true}
     partitions=2526/2526
     cardinality: 1
     probe runtime filters:
     - filter_id = 0, probe_expr = (1: l_orderkey)   // <----- HERE
```

### Root cause

The issue was not that `IcebergScanNode` lacked TopN runtime filter support. The missing piece was in the FE optimizer rule chain.

Before the fix, `PushDownTopNToPreAggRule` only matched plans where the TopN above the aggregate was already a `PARTIAL TOP-N`. In other words, it expected the plan to already look like this:

```text
PARTIAL TOP-N
  GLOBAL AGG
    LOCAL AGG
```

But the Iceberg single-table query often reached that rule in this shape instead:

```text
FINAL TOP-N
  GLOBAL AGG
    LOCAL AGG
```

Since the old rule only accepted `PARTIAL TOP-N`, it did not fire. As a result:

- no local TopN was inserted above `LOCAL AGG`
- `LOCAL AGG` was not marked as `topNLocalAgg`
- no aggregation runtime filter was built
- `IcebergScanNode` had no runtime filter to probe

The fix was to make `PushDownTopNToPreAggRule` also handle an unsplit `FINAL TOP-N`. If it sees that shape, it now synthesizes the missing split itself and continues the pushdown.

### Before

```mermaid
flowchart TD
    A["FINAL TOP-N"] --> B["GLOBAL AGG"]
    B --> C["LOCAL AGG"]
    C --> D["IcebergScanNode"]

    E["PushDownTopNToPreAggRule"] -. "does not match\n(only accepted PARTIAL TOP-N)" .-> A
    F["Result"] --> G["No local TopN above LOCAL AGG"]
    F --> H["No build runtime filters on AGG"]
    F --> I["No probe runtime filters on IcebergScanNode"]
```

### After

```mermaid
flowchart TD
    A["FINAL TOP-N (split)"] --> B["PARTIAL TOP-N"]
    B --> C["GLOBAL AGG"]
    C --> D["LOCAL TOP-N (per-pipeline)"]
    D --> E["LOCAL AGG (topNLocalAgg=true)"]
    E --> F["IcebergScanNode"]

    G["PushDownTopNToPreAggRule"] -. "matches FINAL TOP-N\nand synthesizes PARTIAL TOP-N" .-> A
    H["Result"] --> I["AGG builds TopN runtime filter"]
    H --> J["IcebergScanNode probes runtime filter"]
```

#### What `SplitTopNRule` Actually Does

`SplitTopNRule` has one job: **split a `FINAL TopN` into two layers**.

```
BEFORE:              AFTER:
FINAL TopN           FINAL TopN
  |                    |
Scan            PARTIAL TopN
                       |
                     Scan
```

It only concerns the TopN operator itself. It has **no knowledge** of aggregation operators below it.

#### What `PushDownTopNToPreAggRule` Does

This rule is responsible for **pushing TopN below the Global Aggregation** to create a local TopN that can build a runtime filter:

```
BEFORE:                           AFTER:
FINAL TopN                        FINAL TopN
  |                                  |
Global Agg                        Global Agg
  |                                  |
Local Agg            =>           EXCHANGE
  |                                  |
IcebergScan                  PARTIAL TopN
                                     |
                                Local Agg  ← builds runtime filter
                                     |
                                IcebergScan  ← receives runtime filter
```

**The runtime filter is built at the `Local Agg` node, not at the TopN node.**

#### Why Modifying `SplitTopNRule` Is the Wrong Place

| Aspect | `SplitTopNRule` | `PushDownTopNToPreAggRule` |
|--------|----------------|---------------------------|
| **Pattern** | `TopN -> AnyChild` | `TopN -> Agg -> Agg` |
| **Concern** | Split TopN phases | Push TopN across aggregation boundary |
| **Runtime Filter** | ❌ Doesn't build it | ✅ Builds it via Local Agg |

1. **SplitTopNRule doesn't touch aggregations**: Its pattern is `TopN -> leaf`. It never sees the Agg operators below.

2. **The root cause is not "TopN isn't split"**: Even after SplitTopNRule runs, `PushDownTopNToPreAggRule` still needs to match and push the TopN down. The issue was that `PushDownTopNToPreAggRule` only accepted `PARTIAL` TopN, missing cases where `FINAL` TopN needed to be pushed for external tables.

3. **Changing SplitTopNRule would be over-broad**: It runs for ALL queries (OLAP and external). Adding special cases for external tables there would affect every query shape, not just `GROUP BY key ORDER BY key`.

#### Our Fix

We keep `SplitTopNRule` unchanged and instead modify **`PushDownTopNToPreAggRule`** to:

- Accept `FINAL TopN` **only for external table scans** (Iceberg, Hive, Hudi, etc.)
- Continue requiring `PARTIAL TopN` for native OLAP tables (preserving existing behavior)

This is the minimal, targeted fix that solves the runtime filter issue without risking regressions on native tables.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/PushDownTopNToPreAggRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/PushDownTopNToPreAggRule.java
@@ -19,9 +19,12 @@ import com.starrocks.sql.optimizer.OptExpression;
 import com.starrocks.sql.optimizer.OptimizerContext;
 import com.starrocks.sql.optimizer.base.Ordering;
 import com.starrocks.sql.optimizer.operator.AggType;
+import com.starrocks.sql.optimizer.operator.Operator;
 import com.starrocks.sql.optimizer.operator.OperatorType;
 import com.starrocks.sql.optimizer.operator.SortPhase;
 import com.starrocks.sql.optimizer.operator.logical.LogicalAggregationOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalOlapScanOperator;
+import com.starrocks.sql.optimizer.operator.logical.LogicalScanOperator;
 import com.starrocks.sql.optimizer.operator.logical.LogicalTopNOperator;
 import com.starrocks.sql.optimizer.operator.pattern.Pattern;
 import com.starrocks.sql.optimizer.operator.scalar.ColumnRefOperator;
@@ -89,7 +92,7 @@ public class PushDownTopNToPreAggRule extends TransformationRule {
             return false;
         }
 
-        if (topn.getSortPhase() != SortPhase.PARTIAL || topn.hasOffset() || topn.getPredicate() != null) {
+        if (topn.hasOffset() || topn.getPredicate() != null) {
             return false;
         }
 
@@ -112,6 +115,13 @@ public class PushDownTopNToPreAggRule extends TransformationRule {
             return false;
         }
 
+        // Check if the scan under local agg is an external table scan.
+        boolean isExternalScan = isExternalScanOperator(aggGlobalChild);
+
+        if (!isSupportedTopN(topn, isExternalScan)) {
+            return false;
+        }
+
         // verify aggregation result columns are not used in the order by columns of topN.
         List<Ordering> orderByElements = topn.getOrderByElements();
         List<ColumnRefOperator> groupingKeys = aggGlobal.getGroupingKeys();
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

HangyuanLiu:
Could you clarify two points about the optimizer shape and the rule boundary?

1. The PR description says: "But the Iceberg single-table query often reached that rule in this shape instead: FINAL TOP-N -> GLOBAL AGG -> LOCAL AGG." What is the Iceberg-specific reason this plan reaches `PushDownTopNToPreAggRule` as an unsplit `FINAL TOP-N`? Is it due to external-scan distribution/costing, statistics, the timing/conditions of `SplitTopNRule`, or another rule interaction? It would be helpful to include the before/after logical plan or optimizer trace showing why the `PARTIAL TOP-N -> GLOBAL AGG -> LOCAL AGG` candidate is not the one matched here.

2. From the rule change itself, the new path seems potentially useful for internal tables as well when a `FINAL TOP-N` above two-stage aggregation satisfies the existing safety checks: limit, no offset/predicate, and `ORDER BY` columns are grouping keys. Why is `FINAL TOP-N` support restricted to `isExternalScan`? Is this only a regression-containment choice for native OLAP tables, or is there a correctness/performance reason internal tables should not use this path?

This context would make the optimization boundary clearer.

dirtysalt:
The reason is that `SplitTopNAggregateRule` can generate partial top-n for OLAP tables, and then, in conjunction with `PushDownTopNToPreAggRule`, generate a top-n runtime filter.

However, many tricky parts of `SplitTopNAggregateRule` only apply to OLAP table interfaces and are not easily extrapolated to Iceberg scans. https://github.com/StarRocks/starrocks/pull/64760

this rule only works for a broader case like described in pr

```
/*
 * e.g.
 *    select a, b, count(c) as cc, sum(d)
 *    from t
 *    order by cc limit 10
 *
 * -> select t.a, t.b, cc, sum(d)
 *    from t join (select a, b, count(c) as cc from t group by a, b order by cc limit 10) t1 on t.a = t1.a and t.b = t1.b
 *    group by t.a, t.b, cc
 *    order by cc limit 10
 */
```


But this rule relies on many interfaces or features that are only available for olap table.

---------

I propose another PR https://github.com/StarRocks/starrocks/pull/72521, which only works for  the cases:

> group by x order by x

fe/fe-core/src/test/java/com/starrocks/sql/plan/IcebergTopNRuntimeFilterTest.java
```diff
@@ -0,0 +1,126 @@
...
+                    .setNewPlanerAggStage(SessionVariableConstants.AggregationStage.AUTO.ordinal());
+            String autoPlan = getVerboseExplain(sql);
+            System.out.println("=== ICEBERG VERBOSE PLAN (AUTO) ===");
+            System.out.println(autoPlan);
+
+            // Also get the non-verbose plan to see TopN phase
+            connectContext.getSessionVariable()
+                    .setNewPlanerAggStage(SessionVariableConstants.AggregationStage.TWO_STAGE.ordinal());
+            String plan = getFragmentPlan(sql);
+            System.out.println("=== ICEBERG FRAGMENT PLAN (TWO_STAGE) ===");
```
stephen-shelby: delete it

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/PushDownTopNToPreAggRule.java
```diff
@@ -112,6 +115,13 @@ public boolean check(final OptExpression input, OptimizerContext context) {
             return false;
         }
 
+        // Check if the scan under local agg is an external table scan.
+        boolean isExternalScan = isExternalScanOperator(aggGlobalChild);
+
+        if (!isSupportedTopN(topn, isExternalScan)) {
```
stephen-shelby, line 121: I think we can move forward with this PR first.

Compared with the other two approaches, this one has the smallest impact surface: it only handles the missing `FINAL TopN -> GLOBAL AGG -> LOCAL AGG` case for external scans, while keeping the existing OLAP path unchanged.

#72521 looks like a cleaner long-term direction, but I’d prefer to do that after this fix is stable, with more explicit tests around the OLAP/non-OLAP boundary.

dirtysalt: OK

CelerData-Reviewer:
@codex review

stephen-shelby (approved):

Youngwb (approved):

dirtysalt:
@mergify backport branch-4.1

PR #72335, opened 2026-04-29
[auto-perf-opt] Reuse IndexValuesWithVerPB in KeyValueMerger merge/flush loop

luohaha:
## Bottleneck (from iter-001 baseline + out-of-band perf record on `branch-4.1-51e81ff`)

Single dominant CPU stack on the hottest CN at 99 % busy, ~270 async-delta-writer tasks/s:

```
29.83 %  cloud_native_pk_* → LakePersistentIndexParallelCompactTask::run
 └ 27.67 %  ::do_run
   └ 20.99 %  starrocks::lake::KeyValueMerger::merge
     ├ 10.59 %  ::flush  (3.86 % new + 3.72 % free)
     ├  4.29 %  protobuf::MessageLite::ParseFromString
     │           └ IndexValuesWithVerPB::_InternalParse
     └  1.83 %  ~IndexValuesWithVerPB → free
```

Self time: `my_malloc` 15.9 % + `my_free` 15.5 % = ~31 % of all real CPU work spent in the allocator. The PK-index merge/flush loop is the dominant caller. Source: `be/src/storage/lake/lake_persistent_index_key_value_merger.cpp:43, 148`.

## What this PR changes

`KeyValueMerger::merge` constructs a fresh `IndexValuesWithVerPB` on every input key and destroys it at scope exit; same in `flush()` per output key. With millions of keys flowing through PK-index parallel compaction during ingest, that's a heap alloc + free of the message and its `RepeatedPtrField` storage per key.

Hoist both to `KeyValueMerger` members (`_scratch_value_pb`, `_scratch_flush_pb`):

- `merge()` relies on `ParseFromString` resetting the message before populating.
- `flush()` calls `Clear()` explicitly because it builds the message via `add_values()`.

Protobuf's `RepeatedPtrField::Clear()` retains the underlying element storage, so the second and subsequent calls reuse the same slot allocations — eliminating the per-key new/free pair.

No behavioural change. The message lifetime in both functions is local to a single call, the message is fully consumed before the next call starts, and there is no concurrent access (each `KeyValueMerger` instance runs on a single compaction worker thread).

## Expected impact

Removes the dominant allocator/parse cost in the merge loop; the merge-side ~21 % inclusive CPU share should drop substantially. Direct effect on publish/apply tail latency depends on how much the freed CPU unblocks the contended PK-index lock — to be measured in iter-002.

Reversible by reverting the diff.

## Iter chain (auto-perf-opt run my-rt)

This branch stacks `bd356ebfb8 "Defer PK-index base-level merges"` (already deployed as `branch-4.1-51e81ff`) plus this iter's new commit `b538db5cbc`. The previous title of this PR ("Defer PK-index base-level merges") referred to `bd356ebfb8`; that commit is unchanged. Reviewers of the *new* optimization should focus on `b538db5cbc`.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -485,6 +485,12 @@ CONF_mInt64(pk_index_memtable_max_wait_flush_timeout_ms, "30000");
 CONF_mInt64(pk_index_size_tiered_min_level_size, "131072");
 CONF_mInt64(pk_index_size_tiered_level_multiplier, "10");
 CONF_mInt64(pk_index_size_tiered_max_level, "5");
+// Defer merge_base_level=true compactions when the selected size-tiered level has
+// at least this many non-base filesets — picks just the non-base filesets instead,
+// forcing merge_base_level=false. Avoids 100-300 s base rewrites on hot/oversized
+// PK-index tablets that block publish_version. Set to 0 to disable. Default 2 means
+// any selected level with >=2 non-base filesets will skip the base.
+CONF_mInt64(pk_index_size_tiered_defer_base_merge_min_non_base_filesets, "2");
 // Used to control the sampling interval size for SSTable files.
 CONF_mInt64(pk_index_sstable_sample_interval_bytes, "16777216");
 // Controls merge condition evaluation behavior within the same transaction for primary key tables.
--- a/be/src/storage/lake/lake_persistent_index_key_value_merger.cpp
+++ b/be/src/storage/lake/lake_persistent_index_key_value_merger.cpp
@@ -40,7 +40,10 @@ Status KeyValueMerger::merge(const sstable::Iterator* iter_ptr) {
     const std::string& value = iter_ptr->value().to_string();
     uint64_t max_rss_rowid = iter_ptr->max_rss_rowid();
 
-    IndexValuesWithVerPB index_value_ver;
+    // Reuse the per-merger protobuf instance: ParseFromString resets it before
+    // populating, and RepeatedPtrField retains its element backing across calls,
+    // so we avoid a heap alloc + free of the message and its values list per key.
+    IndexValuesWithVerPB& index_value_ver = _scratch_value_pb;
     if (!index_value_ver.ParseFromString(value)) {
         return Status::InternalError("Failed to parse index value ver");
     }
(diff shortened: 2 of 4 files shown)


luohaha:
## Iteration 001 follow-up commit pushed: a767311

Added a second commit that raises `lake_pk_index_block_cache_limit_percent` default from 10 → 30. This targets the dominant Pattern B bottleneck observed in iter-001's measurement — `multiget_t3_us` (SST data-block reads inside `Table::MultiGet`) accounts for ~99 % of cold-publish time on the hottest CN, with `read_block_miss_cache_cnt:10 / hit:1` typical.

### iter-001 measurement summary (cluster running this PR's prior commit)
- Upsert P50 / P99 / **Max** = 165 / 2 411 / **21 302** ms (target Max ≤ 3 000 ms)
- 194 SLOW_LOAD events ≥ 10 s, **all** inside the 6-min cold-start window (10:11–10:18); steady state has 0
- Slow-load step split:
  - 999 small tables: `WriteDataTimeMs` dominates (avg 1 808 ms) ← Pattern B
  - tbl=0 (1 TB / 60 buckets): `CommitAndPublishTimeMs` dominates (avg 3 791 ms) ← still some merge_base_level lock stalls even with the deferral patch
- `cloud_native_pk_index_compact` avg execute = 5 178 ms cluster-wide (.128 = 10 005 ms peak) — base-level deferral helped but did not eliminate
- Publish-trace (cost ≥ 1 s, six CNs, cold-start window): 1 007 events; on hot CN .129, 32 % classify as Pattern B (multi_get/multiget_t3 dominated)

### Why this follow-up commit
The block cache holds SST data blocks. With ~10 GB update-mem per CN and the old 10 % limit, each CN had ~1 GB of cache to serve all 1 000 tables' PK SST blocks — barely enough, so post-cold-start churn (compaction rewrites, new SST flushes) forced re-reads. 30 % gives ~3 GB per CN, comfortably covering the working set.

This is the smallest possible follow-up — one-line default change, no algorithm change, easy to revert if it regresses. Targets the post-cold-start "transition" window's tail, not the cold-start spike itself. The cold-start fix (e.g. parallel block prefetch in MultiGet, or pre-warming PK index at CN startup) is design-grade and deferred to a future iteration.

Full diagnosis at `.workspace/my-rt/state/runs/run_20260430_100647/iterations/001/diagnosis.md` in the auto-perf-opt repo.

luohaha:

PR #72364, opened 2026-04-30
[Enhancement] Add CN-side fallback for cn-free tablet metadata

starrocks-xupeng:
Introduce the CN-side consumer of getTabletInitialMetadata so that when version 1 tablet metadata is missing from object storage, CN constructs it on demand by calling the FE RPC. Required infrastructure for the upcoming CN-Free Tablet Creation feature, where DDL will skip writing v1 metadata.

This change does not alter existing behavior on its own:
  * FE still sends CreateReplicaTask for every DDL, so v1 metadata is always present and the fallback is dormant in the normal path.
  * The user-visible switch (table property cn_free_tablet_creation and DDL skip in TabletTaskExecutor) lands in a later PR.

Changes:
  * TableSchemaService::get_tablet_initial_metadata calls the RPC with SingleFlight deduplication keyed by (table_id, index_id) so concurrent requests for sibling tablets share one round trip, plus retry and latency/error metrics.
  * TabletManager grows three helpers: parse_shard_properties extracts table/partition/index ids from the StarOS shard, build_initial_metadata converts the RPC response into a TabletMetadataPB, and construct_initial_metadata orchestrates the two with USE_STAROS guards and a #ifdef BE_TEST early return so unit tests do not require a live starlet worker.
  * TabletManager::get_tablet_metadata(path, ...) falls back to construct_initial_metadata when the v1 metadata file is missing, tablet_id is non-zero, version == kInitialVersion, and fs == nullptr. The fs == nullptr guard excludes cross-cluster replication reads, which pass the source cluster's filesystem.
  * MetadataIterator routes the initial-metadata path to the (tablet_id, version) overload so the fallback is reachable from listing as well.
  * vacuum.cpp's drop_tablet_cache loop skips version 1 (the empty initial metadata) so the cn-free fallback's cache is not repeatedly evicted.
  * AgentService.thrift gains a NOTE on TCreateTabletReq reminding future contributors to propagate any new field through TGetTabletMetadataResponse to keep the two paths in sync.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/tablet_manager.cpp
+++ b/be/src/storage/lake/tablet_manager.cpp
@@ -208,6 +208,8 @@ UpdateManager* TabletManager::update_mgr() {
     return _update_mgr;
 }
 
+// NOTE: if you add a new field here, also update construct_initial_metadata() and
+// TCloudTabletMeta in FrontendService.thrift to keep them in sync.
 Status TabletManager::create_tablet(const TCreateTabletReq& req) {
     // generate tablet metadata pb
     auto tablet_metadata_pb = std::make_shared<TabletMetadataPB>();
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -275,6 +277,145 @@ Status TabletManager::create_tablet(const TCreateTabletReq& req) {
...
+StatusOr<TabletMetadataPtr> TabletManager::build_initial_metadata(int64_t tablet_id,
+                                                                  const TGetTabletMetadataResponse& resp) {
+    if (!resp.__isset.meta) {
+        return Status::InternalError(strings::Substitute("missing tablet meta in response, tabletId:$0", tablet_id));
+    }
+    const auto& meta = resp.meta;
+    if (meta.__isset.tablet_id && meta.tablet_id != tablet_id) {
+        return Status::InternalError(
+                strings::Substitute("tablet_id mismatch: requested $0, response $1", tablet_id, meta.tablet_id));
+    }
```
banmoy, line 353: move these check to table schema service, and here only need TCloudTabletMeta not TGetTabletMetadataResponse?

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -577,6 +718,20 @@ StatusOr<TabletMetadataPtr> TabletManager::get_tablet_metadata(const string& pat
...
+    // writing version 1 metadata to object storage. When a consumer first needs version 1
+    // metadata, we fetch the tablet's initial configuration from FE and construct the
+    // TabletMetadataPB on demand.
+    //
+    // The fs == nullptr check excludes cross-cluster replication reads, where |fs| points to
+    // the source cluster's object storage. Without this check, a source tablet_id that
+    // coincidentally matches a local tablet_id would cause us to fetch the wrong metadata
+    // from the local FE. Other local callers that pass non-null fs (e.g. LakeDelvecLoader)
+    // do not request version 1 in practice, since version 1 has no rowsets or delvecs.
+    if (metadata_or.status().is_not_found() && tablet_id != 0 && version == kInitialVersion && fs == nullptr) {
```
banmoy, line 737: should always fetch the initial metadata from FE before reading from s3 ?  only need to read from s3 for compatability?

banmoy (commented):
Two correctness concerns from review.

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -275,6 +277,145 @@ Status TabletManager::create_tablet(const TCreateTabletReq& req) {
...
+
+// NOTE: if you add a new field to create_tablet(), also update this method and
+// TCloudTabletMeta in FrontendService.thrift to keep them in sync.
+StatusOr<TabletMetadataPtr> TabletManager::build_initial_metadata(int64_t tablet_id,
+                                                                  const TGetTabletMetadataResponse& resp) {
+    if (!resp.__isset.meta) {
+        return Status::InternalError(strings::Substitute("missing tablet meta in response, tabletId:$0", tablet_id));
+    }
+    const auto& meta = resp.meta;
+    if (meta.__isset.tablet_id && meta.tablet_id != tablet_id) {
```
banmoy: `meta.tablet_id != tablet_id` is reasonable if rpc requests from multiple tablets are grouped?

be/src/storage/lake/tablet_manager.cpp
starrocks-xupeng: i think putting it here is fine, tablet id check is removed, right now only meta set is checked

be/src/storage/lake/tablet_manager.cpp
starrocks-xupeng: we better not reverse the order now, we can do it later when we enable cn free by default, and right now a cache miss for the version 1 tablet is acceptable

be/src/storage/lake/tablet_manager.cpp
starrocks-xupeng: done

be/src/storage/lake/tablet_manager.cpp
banmoy: is cn-free not default even for new tables?

starrocks-xupeng: not now, we can reverse the order later, when all PRs are merged

starrocks-xupeng: FE pr for property is not submitted

banmoy: ok

banmoy (approved):

kevincai (approved):

xiangguangyxg (approved):

PR #72376, opened 2026-04-30
[BugFix] Reject multi-request batches in getTabletMetadata

starrocks-xupeng:
Multiple TGetTabletMetadataRequests for sibling tablets under the same index each acquired and released the table read lock independently. A schema change (or other DDL) committing between two iterations could leave sibling responses in the same batch with mismatched schemas, leading CN to assemble TabletMetadataPB with inconsistent fields.

A correct multi-request implementation would have to take a read lock on every distinct (db, table) covered by the batch and hold all of them for the full duration so that sibling tablets observe one consistent snapshot. The grouping, lock-ordering (to avoid deadlocks against other DDL), and partial-failure bookkeeping needed to do that correctly are not worth the complexity given that the current CN caller packs exactly one request. Reject batches with more than one request at the batch level with NOT_IMPLEMENTED_ERROR. The thrift wire shape is preserved so a future change can lift the restriction without breaking compatibility.

Add testGetTabletMetadataBatchSizeRejected to verify the rejection path.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/service/FrontendServiceImpl.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/FrontendServiceImpl.java
@@ -3601,11 +3601,28 @@ public class FrontendServiceImpl implements FrontendService.Iface {
             TBatchGetTabletMetadataRequest batchRequest) {
         TBatchGetTabletMetadataResponse batchResponse = new TBatchGetTabletMetadataResponse();
         batchResponse.setStatus(new TStatus(OK));
-        if (batchRequest.isSetRequests()) {
-            for (TGetTabletMetadataRequest request : batchRequest.getRequests()) {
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

kevincai (dismissed):

kevincai (approved):

xiangguangyxg (approved):

PR #72413, opened 2026-04-30
[auto-perf-opt] Reuse PB + SerializeToString buffer in PK index SST emit

luohaha:
## Bottleneck

iter-005 in the auto-perf-opt run (`my-rt`, template `999_1gb_1_1tb`, 6-CN shared-data, 1 hr warm) deployed PR #72399 and shows the previously-dominant `KeyValueMerger::merge` stack has dropped to noise (1.48 % inclusive vs 21 % pre-#72335). But on the hottest CN (172.26.80.129, 85.7 % non-idle) the allocator still owns ~10 % of CPU samples in the same `cloud_native_pk_*` thread family:

```
self  symbol
5.10 % my_memalign     (mostly std::string heap allocs)
4.95 % my_valloc       (mostly RepeatedPtrField growth)
```

Two per-key allocation sites remain in the SST emit path that PR #72335 / #72399 did not touch:

1. **`KeyValueMerger::flush()` in `lake_persistent_index_key_value_merger.cpp`** — already reuses the message via `_scratch_flush_pb` (added in #72335), but still calls `SerializeAsString()` per output key, which returns a brand new `std::string` (heap alloc + memcpy + dtor) every call.
2. **`PersistentIndexSstable::build_sstable()` in `persistent_index_sstable.cpp`** — for every entry of the in-memory PK-index batch, allocates a fresh `IndexValuesWithVerPB`, calls `add_values()` (slot alloc), then `SerializeAsString()` (string alloc). Three heap allocations per key, on a loop driven by every key in the flushed map.

iter-004's plan explicitly flagged the second site as the next escalation target.

## Change

- `KeyValueMerger`: add a `std::string _scratch_serialized` member. In `flush()`, replace `Slice(index_value_pb.SerializeAsString())` with `index_value_pb.SerializeToString(&_scratch_serialized); ... Slice(_scratch_serialized)`. The buffer's capacity is reused across calls; the `Slice` only has to outlive the `TableBuilder::Add()` call (which copies into its block builder before returning).
- `build_sstable()`: hoist the `IndexValuesWithVerPB` and a `std::string` out of the per-key loop. Inside the loop, call `Clear()` on the message (`RepeatedPtrField::Clear()` retains the values slot's backing allocation) and `SerializeToString` into the reused buffer.

No behavioural change. In both functions the message lifetime and the serialized bytes lifetime are scoped to one loop iteration, with no concurrent access — each `KeyValueMerger` / `build_sstable` call runs on a single compaction worker thread.

## Expected impact

Expect the combined `my_memalign` + `my_valloc` self-time on the hottest CN to drop from 10 % to ~3-5 %, freeing CPU on the PK-index parallel-compaction threads where `tbl=0`'s CommitAndPublish work concentrates. Direct effect on upsert max depends on how much the freed CPU unblocks the rest of the publish path — to be measured in the next iteration.

## Cumulative

This branch stacks on PR #72335 (`44260ec`) and PR #72399 (`445e8cf`, `872a0f3`) so the deploy validates them together against the iter-001 baseline.

## Risk

Pure memory-reuse refactor, mirrors PR #72335's pattern. Reversible by reverting the diff.

### DIFF
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -641,8 +641,7 @@ StatusOr<std::vector<KeyValueMerger::KeyValueMergerOutput>> LakePersistentIndex:
         // Compare via Slice directly; .to_string() would allocate a fresh
         // std::string per key on a loop driven by every input row.
         const Slice cur_key = iter_ptr->key();
-        if (!seek_range.stop_key.empty() &&
-            options.comparator->Compare(cur_key, Slice(seek_range.stop_key)) >= 0) {
+        if (!seek_range.stop_key.empty() && options.comparator->Compare(cur_key, Slice(seek_range.stop_key)) >= 0) {
             // meet the scan range boundary, quit.
             break;
         }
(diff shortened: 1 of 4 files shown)


luohaha:

PR #72421, opened 2026-05-01
[Enhancement] Require OPERATE on SYSTEM (or task ownership) for DROP TASK (#71294)

EdwardArchive:
Based on this PR: https://github.com/StarRocks/starrocks/pull/72141

## Why I'm doing

Addressing the core privilege hole in issue StarRocks/starrocks#71294: `DROP TASK <name>` has no authorization check at all. `AuthorizerStmtVisitor` has no `visitDropTaskStmt` override, so the default no-op visit is used, and `DDLStmtExecutor.visitDropTaskStmt` only checks task existence and the MV-source guard. Any authenticated user — even a SELECT-only account — can delete any task on the cluster, including tasks owned by other users and tasks generated by materialized views (with `FORCE`).

The Pipe-task delete-protection sliver and the broader "make TASK a grantable privilege" approach are tracked as separate PRs against the same issue.

## What I'm doing

Two coordinated changes (one commit each):

### Commit 1 — `697f85a` close the privilege hole

Add `AuthorizerStmtVisitor.visitDropTaskStmt` using the same primitives `KILL` already uses (`StmtExecutor.handleKill`):

- if `task.getUserIdentity().equals(ctx.getCurrentUserIdentity())` → allow (self-ownership)
- otherwise require `Authorizer.checkSystemAction(ctx, OPERATE)` on `SYSTEM`

```text
DROP TASK <name>
        │
        ▼
AuthorizerStmtVisitor.visitDropTaskStmt   ← NEW
        │
        ├── task = TaskManager.getTask(name)
        │     └── null  → defer to executor (IF EXISTS / SemanticException)
        │
        ├── caller.equals(task.userIdentity) ?  ──► allow
        │
        └── else: Authorizer.checkSystemAction(ctx, OPERATE)
                    └── deny → AccessDeniedException(OPERATE, SYSTEM)
        │
        ▼
DDLStmtExecutor.visitDropTaskStmt          ← unchanged
```

No new `ObjectType`, no new `PrivilegeType`, no GRANT surface, no docs/grammar changes. The not-found / `IF EXISTS` path is left to `DDLStmtExecutor.visitDropTaskStmt` so existence is not leaked at authorization time.

### Commit 2 — `3d4a393` legacy serialisation compatibility (W1) and tests (W2)

The visitor's strict `equals()` ownership check requires `Task.userIdentity` to be non-null. That field was only persisted from #47561 (June 2024) onwards. Pre-#47561 metadata images deserialise with `userIdentity == null` while the deprecated `createUser` field still carries the user-name portion of the original creator. Without a migration, those original owners would lose self-ownership semantics on upgrade and need `OPERATE on SYSTEM` to drop their own tasks.

`Task.gsonPostProcess` now back-fills `userIdentity` from `createUser` via `UserIdentity.createAnalyzedUserIdentWithIp(createUser, "%")` when only the legacy field is present:

```java
if (userIdentity == null && createUser != null) {
    this.userIdentity = UserIdentity.createAnalyzedUserIdentWithIp(createUser, "%");
}
```

- Host-restricted creators (`'alice'@'10.0.0.5'`) collapse to `'alice'@'%'` — **fail-closed** and matches the common `@%` SUBMIT TASK case
- Idempotent — tasks that already carry `userIdentity` are not overwritten
- Follows the StarRocks gsonPostProcess migration convention (`PhysicalPartition.baseIndex` → `baseIndexMetaId` in #68334, `MaterializedView.partitionRefTableExprs` migration, `StorageVolumeMgr` legacy-key cleanup in #30491, etc.)

Tests added in `PrivilegeCheckerTest`:
- `testDropTaskStmtLegacyTaskNoUserIdentity` exercises the visitor's `creator == null` defensive branch (deny without OPERATE, allow after GRANT)
- `testTaskGsonPostProcessMigratesLegacyCreateUser` calls `gsonPostProcess()` directly and asserts the migrated identity is `'<user>'@'%'`, equal to the live `ConnectContext` user, and that the migration is idempotent

## Behaviour change for upgraders

Before:
```
mysql> -- as low_priv_user (SELECT-only)
mysql> DROP TASK someone_elses_task;
Query OK  -- silently deletes another user's task
```

After:
```
mysql> -- as low_priv_user
mysql> DROP TASK someone_elses_task;
ERROR 1064 (HY000): Access denied; you need (at least one of) the OPERATE privilege(s) on SYSTEM for this operation

mysql> -- task creator drops their own task (post-#47561 task, or legacy task with @% identity)
mysql> DROP TASK my_own_task;
Query OK

mysql> -- after GRANT OPERATE ON SYSTEM TO low_priv_user
mysql> DROP TASK someone_elses_task;
Query OK
```

**Migration impact for pre-#47561 host-restricted creators**

A creator originally registered as `'alice'@'10.0.0.5'` whose task pre-dates #47561 will, after upgrade, see their task carry the migrated identity `'alice'@'%'`. When that user reconnects from `10.0.0.5`, the strict `equals()` check fails (`'alice'@'10.0.0.5' != 'alice'@'%'`) and they need `OPERATE on SYSTEM` to drop their own task. The deprecated `createUser` only stored the user-name portion, so the original host is unrecoverable — `@'%'` is the safest reconstruction and matches the dominant `SUBMIT TASK ... @%` case in deployments.

Affected operators can recover via:
```sql
GRANT OPERATE ON SYSTEM TO 'alice'@'10.0.0.5';
-- or, after dropping/recreating, the new task carries the full identity
```

## Design discussion: Option A (this PR) vs Option B (alternative)

This PR implements **Option A**: strict `UserIdentity.equals()` self-ownership + gsonPostProcess migration. Picked because StarRocks's existing privilege model is uniformly host-aware and Option A is the only choice consistent with it.

**Option B** considered and not taken: skip the gsonPostProcess migration and have the visitor compare by user-name only (`task.getUserIdentity().getUser()` with `task.getCreateUser()` fallback). Trade-offs:

|                                  | A (this PR)                                  | B (user-name match)                                |
|----------------------------------|----------------------------------------------|----------------------------------------------------|
| Visitor comparison               | `UserIdentity.equals()` (host-aware)         | user-name only                                     |
| Migration                        | gsonPostProcess back-fill, `@'%'` collapse   | none                                               |
| StarRocks convention             | matches existing pattern                     | departs (no `gsonPostProcess` precedent for skip)  |
| Pre-#47561 host-restricted owner | needs `OPERATE on SYSTEM` to self-drop       | retains self-drop                                  |
| Cross-host alias semantics       | `'alice'@'A'` cannot drop `'alice'@'B'` task | `'alice'@'A'` *can* drop `'alice'@'B'` task        |
| GRANT model alignment            | aligned (GRANT is per-`user@host`)           | weakened (ignores host that GRANT itself respects) |

### Why Option A — rationale and supporting evidence

The user identifier in StarRocks (and MySQL) is `(user, host)`, not `user`. `'alice'@'%'` and `'alice'@'127.0.0.1'` are functionally separate principals — same kind of separation as `'alice_a'` and `'alice_b'`. Option A keeps task self-ownership consistent with that identifier; Option B would be the *only* place in the privilege model that ignores host.

**Code-level evidence the rest of the model is host-aware:**

| Layer | Evidence |
|---|---|
| Authorization store | `AuthorizationMgr.userToPrivilegeCollection` is `Map<UserIdentity, UserPrivilegeCollectionV2>` — keyed by full identity, not user-name (`fe-core/.../privilege/AuthorizationMgr.java:87`) |
| Authentication | `AuthenticationHandler` calls `context.setCurrentUserIdentity(authenticationResult.authenticatedUser)` at login and stores the full `UserIdentity` for the session lifetime (`fe-core/.../authentication/AuthenticationHandler.java:184`) |
| Session | `ConnectContext.getCurrentUserIdentity()` returns the same object on every privilege check; user-name is never extracted as the canonical identity (`fe-core/.../qe/ConnectContext.java:472`) |
| Equality | `UserIdentity.equals()` returns true only when `user`, `host`, and `isDomain` all match (`fe-spi/.../catalog/UserIdentity.java:121-126`) |
| Login matching | `AuthenticationMgr.getBestMatchedUserIdentity(user, remoteHost)` selects the most-specific match; the chosen identity is what the session keeps. There is no "fall back to less-specific" |

**Empirical verification on a live cluster (FE running this branch, port shifted to `:19030`):**

1. `'alice'@'%'` and `'alice'@'127.0.0.1'` were created with **different passwords** and **different GRANTs** (one had `SELECT` on `t1`, the other on `t2`).
2. Connecting as `alice` from 127.0.0.1 with the password belonging to `'alice'@'127.0.0.1'` → `SELECT current_user()` returned `'alice'@'127.0.0.1'`. The same identity persisted unchanged across **multiple statements** in one session (verified by querying `current_user()` at four points).
3. From that session, `SELECT FROM t2` succeeded but `SELECT FROM t1` was denied — privileges granted to `'alice'@'%'` were **not** seen, even mid-session, even though both identities share the user-name `alice`.
4. `SHOW GRANTS FOR CURRENT_USER()` showed only the grants attached to `'alice'@'127.0.0.1'`. The `'alice'@'%'` grants were invisible to this session.
5. Direct test of the visitor: `'alice'@'%'` created `alice_host_task`; from a session bound to `'alice'@'127.0.0.1'`, `DROP TASK alice_host_task` returned `ERROR 5203 Access denied; you need OPERATE on SYSTEM`. Audit log:
   ```
   SUBMIT — User=alice  AuthorizedUser='alice'@'%'         State=EOF
   DROP   — User=alice  AuthorizedUser='alice'@'127.0.0.1' State=ERR INTERNAL_ERR
   DROP   — User=root   AuthorizedUser='root'@'%'          State=OK     (control)
   ```

The audit `User=` column shows the connection user-name (both `alice`); the `AuthorizedUser=` column shows the actual `UserIdentity` — and the privilege decision tracks the latter. This is the same model GRANT, SELECT, INSERT, and every other privileged operation already uses.

**What Option B would mean given that model**

Option B would make `DROP TASK` self-ownership the single privilege check in StarRocks that compares only the user-name component while ignoring the host. A user holding both `'alice'@'A'` and `'alice'@'B'` (intentionally separated for security policy reasons — common pattern for differentiating internal-network vs VPN access, or app-server vs admin-host) would see one identity able to delete the other's tasks, while every other privileged action between those identities remains correctly isolated. That asymmetry is what tipped the choice toward A.

### Migration impact (the cost of choosing A)

The cost is on pre-#47561 *host-restricted* task creators, who lose self-drop after the upgrade because the deprecated `createUser` field only stored the user-name portion (host is unrecoverable). The `gsonPostProcess` back-fill chooses `@'%'` as the safest reconstruction because `@'%'` is the dominant `SUBMIT TASK` pattern in real deployments — a `'alice'@'%'` original creator is unaffected; a `'alice'@'10.0.0.5'` original creator gets a one-time `GRANT OPERATE ON SYSTEM` from an admin and is back to working. New tasks created post-upgrade carry the full identity from `SUBMIT TASK` onward and are unaffected.

I'm proposing A with this rationale and supporting evidence on the table. If reviewers feel the migration cost on legacy host-restricted creators outweighs model consistency, B is a coherent fallback — but it should be a deliberate decision to introduce a host-blind exception in an otherwise host-aware model, not a default. The central privilege hole is closed under either option.

## Scope

This PR covers only the privilege-enforcement slice of #71294. Out of scope:
- `ALTER TASK` has the same hole and will be a follow-up PR
- The "Pipe-task delete-protection" sliver (rejecting `DROP TASK` on `TaskSource.PIPE` without `FORCE`) — already merged separately (#72141)
- Promoting `TASK` to a first-class grantable `ObjectType` with `CREATE/ALTER/DROP` actions (a future privilege-model PR will need to migrate existing task creators into the new grant surface — the gsonPostProcess back-fill in this PR makes that migration straightforward since every task carries a non-null `userIdentity`)
- Filtering `information_schema.tasks` by owner

`DROP TASK <name>` previously succeeded for any authenticated user; it now requires that the caller be the task creator or hold `OPERATE on SYSTEM`. The not-found / `IF EXISTS` paths and all internal cleanup paths (`TaskManager.dropTasks` calls from MV/Pipe internals) are unchanged. Pre-#47561 host-restricted task owners may need a one-time `OPERATE on SYSTEM` grant — see *Migration impact* above.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [x] Interface/UI changes: `DROP TASK` now returns `Access denied` for unauthorized callers
- [ ] Parameter changes
- [x] Policy changes: authorization required for `DROP TASK` (owner or `OPERATE on SYSTEM`)
- [ ] Feature removed
- [ ] Miscellaneous

## Checklist:

  - `PrivilegeCheckerTest#testDropTaskStmt` covers: cross-user denial without OPERATE, allow after `GRANT OPERATE ON SYSTEM`, self-ownership allow, and `DROP TASK IF EXISTS <missing>` deferred to the executor
  - `PrivilegeCheckerTest#testDropTaskStmtLegacyTaskNoUserIdentity` covers the visitor's `creator == null` defensive branch
  - `PrivilegeCheckerTest#testTaskGsonPostProcessMigratesLegacyCreateUser` covers the legacy serialisation back-fill (correct `@%` form, equality with live `ConnectContext` user, idempotency)
- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Test plan

- [x] `mvn checkstyle:check -pl fe-core` — clean
- [x] `PrivilegeCheckerTest#testDropTaskStmt` — passes locally
- [x] `PrivilegeCheckerTest#testDropTaskStmtLegacyTaskNoUserIdentity` — passes locally
- [x] `PrivilegeCheckerTest#testTaskGsonPostProcessMigratesLegacyCreateUser` — passes locally
- [x] Manual smoke against a running cluster (FE rebuilt from this branch, BE pre-built):
  - As `root`: `SUBMIT TASK foo AS INSERT INTO db.t SELECT ...` — succeeds
  - As `low_priv_user`: `DROP TASK foo` → `Access denied` (`OPERATE` on `SYSTEM`)
  - `GRANT OPERATE ON SYSTEM TO low_priv_user`
  - As `low_priv_user`: `DROP TASK foo` → succeeds
  - Self-drop by task owner — succeeds
  - `DROP TASK IF EXISTS missing` — no error, deferred to executor
image

- [x] Full `com.starrocks.sql.analyzer.PrivilegeCheckerTest` class — **74 tests, 0 failures, 0 errors, 1 skipped** (skipped test is pre-existing `@Disabled testTableSelectDeleteInsertUpdateColumn`, unrelated)
- [x] Full `com.starrocks.qe.RedirectStatusTest` class — **268 tests, 0 failures, 0 errors, 0 skipped**; the existing `testDropTaskStmt` there is confirmed unaffected (no signature changes)

Refs StarRocks/starrocks#71294.


CelerData-Reviewer:
@codex review

PR #72423, opened 2026-05-01
[BugFix] Locker: rollback partial intensive-lock acquisition

kevincai:
lockTableWithIntensiveDbLock and lockTablesWithIntensiveDbLock acquire a DB intention lock first and then one or more table locks. If any of the table-lock acquisitions threw (deadlock victim selection, interrupt, timeout, etc.), the helpers caught LockException and rethrew it as ErrorReportException without releasing the DB intention lock that was already held -- and, in the multi-table variant, without releasing any table locks acquired earlier in the loop. The result was a stranded IS / IX lock on the database that blocked subsequent DB-WRITE operations (DROP DATABASE / DROP TABLE / quota change / rename) indefinitely until the FE process restarted.

Callers could not work around this from the call site:
  - Lock-acquisition outside try/finally: catch is never reached, the finally never runs, so the partial DB lock leaks.
  - Lock-acquisition inside try/finally: the finally calls the matching unLock... helper, which first releases the DB intention lock (good) and then iterates the table-id list calling release on each -- but table locks past the failure point were never acquired, so their release throws and masks the original LockException with a less informative "lock not held" error. The DB lock is released in this case, but the diagnostics are wrong.

Fix the helpers themselves. On any LockException during acquisition, call a new private rollbackPartialIntensiveLock that releases each table lock that did get acquired (in acquisition order, like the existing tryLockTablesWithIntensiveDbLock pattern at lines 338-348), then releases the DB intention lock if it was acquired. Each release is wrapped in a defensive try/catch that logs and swallows any secondary error, so the original LockException surfaces to the caller instead of being masked by a release-side failure.

After this change, the lock-acquisition-outside-try/finally idiom at the call sites is correct: a partial-acquire failure leaves no held locks behind, so there is nothing for a finally block to release.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/concurrent/lock/Locker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/concurrent/lock/Locker.java
@@ -288,23 +288,55 @@ public class Locker {
         Preconditions.checkState(lockType.equals(LockType.READ) || lockType.equals(LockType.WRITE));
         List<Long> tableListClone = new ArrayList<>(tableList);
         if (Config.lock_manager_enabled) {
+            LockType intentionType = (lockType == LockType.WRITE)
+                    ? LockType.INTENTION_EXCLUSIVE
+                    : LockType.INTENTION_SHARED;
+            boolean dbLockHeld = false;
+            List<Long> ridLockedList = new ArrayList<>();
             try {
-                if (lockType == LockType.WRITE) {
-                    this.lock(dbId, LockType.INTENTION_EXCLUSIVE, 0);
-                } else {
-                    this.lock(dbId, LockType.INTENTION_SHARED, 0);
-                }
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

alvin-phoenix-ai (approved):

alvin-phoenix-ai:
@codex review

kevincai:
give up backport to branch-3.4

PR #72496, opened 2026-05-03
[Enhancement] Generate partial TopN for TopN over aggregate group keys

dirtysalt:
## Why I'm Doing

Some TopN-over-aggregate queries on external tables cannot produce TopN runtime filters because the plan may not contain a partial TopN before `PushDownTopNToPreAggRule` runs. This affects queries like `GROUP BY key ORDER BY key LIMIT n`, where pushing a partial TopN below the aggregate can enable scan-side runtime filtering.

## What I'm Doing

This change lets the existing TopN split path handle `TopN -> Project -> Aggregate` when the order-by columns can be mapped through the project to aggregate group keys. The optimizer now re-runs the existing split/topn-preagg rule chain for generic TopN-over-group-key aggregates, not just HDFS-family scans. It also adds plan tests for both native and Iceberg tables and leaves `PushDownTopNToPreAggRule` unchanged.

## Tradeoff

`SplitTopNAggregateRule` is an OLAP-specific and more aggressive optimization. It targets queries where `ORDER BY` depends on aggregate outputs, and rewrites the plan into a duplicated scan plus TopN subplan joined back to the original aggregation path. https://github.com/StarRocks/starrocks/pull/64760

This change targets a different pattern: `TopN` over aggregate group keys on non-OLAP scans. For this case, we only need to produce a partial TopN so that the existing `PushDownTopNToPreAggRule` can push it through the two-phase aggregate and enable scan-side runtime filters.

The new rewrite chain is therefore gated to non-OLAP scans. Native OLAP tables keep using their existing TopN/Aggregate optimization path, which avoids changing established OLAP plan baselines while still filling the missing partial-TopN path for external tables.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/QueryOptimizer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/QueryOptimizer.java
@@ -29,6 +29,7 @@ import com.starrocks.sql.Explain;
 import com.starrocks.sql.ast.JoinOperator;
 import com.starrocks.sql.optimizer.base.ColumnRefFactory;
 import com.starrocks.sql.optimizer.base.ColumnRefSet;
+import com.starrocks.sql.optimizer.base.Ordering;
 import com.starrocks.sql.optimizer.base.PhysicalPropertySet;
 import com.starrocks.sql.optimizer.cost.CostEstimate;
 import com.starrocks.sql.optimizer.cost.feature.PlanFeatures;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

dirtysalt:
@codex review

PR #72554, opened 2026-05-03
[auto-perf-opt] Eliminate HTTP-server accept thundering-herd via SO_REUSEPORT

luohaha:
## Summary

The BE HTTP server (\`EvHttpServer\`) spawns N libevent worker threads (\`be_http_num_workers\`, default 1000), all calling \`evhttp_accept_socket\` on the **same** \`_server_fd\`. With no \`EPOLLEXCLUSIVE\` flag, every incoming SYN wakes all N threads and the kernel serializes them on the listening socket spinlock. This is a textbook accept thundering herd.

This PR adds a \`enable_http_server_so_reuseport\` config (default true) so each worker creates its own \`SO_REUSEPORT\` listener socket bound to the same address. The kernel hash-balances accepts across listeners in-kernel, eliminating the spinlock contention. Falls back to legacy shared-fd behaviour on any setsockopt/bind failure.

## Empirical impact

Validated on the \`999_1gb_1_1tb\` realtime-benchmark template (1 FE + 6 BE, branch-4.1, shared-data, S3 backend) using \`auto-perf-opt\`:

| Metric | Before (iter-018, no SO_REUSEPORT) | After (iter-021, SO_REUSEPORT on) | Δ |
|---|---:|---:|---:|
| \`native_queued_spin_lock_slowpath\` (perf self %) | 7.50 % | < 0.5 % | -94 % |
| Upsert max latency | 12 200 ms | **2 851 ms** | -77 % |
| Upsert P99 | 1 171 ms | 629 ms | -46 % |
| Throughput | 760 K rows/s | 917 K rows/s | +21 % |
| Ingest error rate | 2.24 % | 0.00 % | guardrail restored |

The released ~7.5 % CPU/CN now goes to legitimate compact + publish work. Cluster also became resilient to FE phantom-retry storms (storm continues at ~100/s/CN but no longer breaks ingest).

## CPU profile (cn 130, 60 s, 95 K samples)

Before:
\`\`\`
event_base_loop → listener_read_cb → accept4 → inet_csk_accept
  → lock_sock_nested → native_queued_spin_lock_slowpath   7.50 % self / 8.57 % incl
\`\`\`

After: entire stack drops below the 0.5 % sampling threshold.

## Test plan

- [x] perf A/B confirms spinlock contention removed (data above)
- [x] No steady-state regression in P99 / throughput / error rate (data above)
- [x] Falls back cleanly when \`SO_REUSEPORT\` setsockopt fails (compile-time check + runtime warning preserved)
- [ ] CI smoke test (let CI run)

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -519,6 +519,13 @@ CONF_Int32(be_http_port, "8040");
 CONF_Alias(be_http_port, webserver_port);
 // Number of http workers in BE
 CONF_Int32(be_http_num_workers, "48");
+// When true, each HTTP worker creates its own listening socket with
+// SO_REUSEPORT so the kernel load-balances incoming connections in-kernel,
+// instead of all workers epoll-waiting on a single shared listen fd and
+// thundering-herd-racing on accept(). Eliminates the kernel
+// inet_csk_accept / native_queued_spin_lock_slowpath contention that
+// dominates HTTP-server CPU when be_http_num_workers is large.
+CONF_Bool(enable_http_server_so_reuseport, "true");
 // Period to update rate counters and sampling counters in ms.
 CONF_mInt32(periodic_counter_update_period_ms, "500");
 
--- a/be/src/http/ev_http_server.cpp
+++ b/be/src/http/ev_http_server.cpp
@@ -49,6 +49,13 @@
 #include <sstream>
 #include <utility>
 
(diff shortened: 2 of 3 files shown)


luohaha:
Superseded by upstream/main #72956. Closing this branch-4.1 draft.

PR #72557, opened 2026-05-03
[auto-perf-opt] Log per-phase breakdown of slow lake publishPartition

luohaha:
## Summary

Adds a single WARN-level log line in `PublishVersionDaemon.publishPartition` that
fires only when the total wall-clock from publish-task submission to BRPC return
exceeds 500ms, breaking it into four sub-phases. Steady-state publishes
(<500ms) log nothing — log volume on the hot path is unchanged.

```
Slow publishPartition txn=… partition=… table=… total_ms=…
        executor_acquire_ms=… lock_wait_ms=… fe_prep_ms=… brpc_ms=…
```

| field                     | what it measures                                                              |
|---------------------------|-------------------------------------------------------------------------------|
| `executor_acquire_ms`     | `CompletableFuture.supplyAsync(..., publish-task pool)` queue wait            |
| `lock_wait_ms`            | `lockTablesWithIntensiveDbLock` (DB intensive READ lock acquire)              |
| `fe_prep_ms`              | table/partition/tablet lookup under the lock                                  |
| `brpc_ms`                 | `Utils.publishVersion(...)` round-trip (BRPC + BE handler + receipt)          |

These four sum to the publishPartition contribution to the existing
\`publish rpc cost\` emitted on the VISIBLE log line by
\`TransactionState.toString()\`.

## Motivation

Across four consecutive optimization iterations on a 6-CN shared-data
cluster we have observed a stable 1-3 s gap between the FE-measured
\`publish rpc cost\` and the maximum BE-measured handler cost
(\`Published txns=… cost=…us\`) on slow upsert events. The previous
iteration (PR #72555 — log BRPC server-side queue time) ruled out
BRPC server-side queueing (slow-handler \`brpc_queue_us\` ≤ 1 ms across
1.14 M events). The gap therefore lives on the FE side, but the
existing \`publish rpc cost\` is one opaque window covering executor
pool queueing, DB lock acquire, tablet lookup, and the BRPC fan-out.
This 15-LOC log-only patch surfaces which of the four dominates the
next time a slow event reproduces, so the actual fix can be targeted
rather than guessed.

## What it's NOT

- Not a perf optimization itself — instrumentation only.
- Not a config change — pool sizes / lock policy / thread counts are unchanged.
- Not a new metric / not exposed to /metrics — single existing log file.

## Performance impact

- Five \`System.currentTimeMillis()\` calls per partition publish (~µs each).
- One \`LOG.warn\` invocation only on ≥ 500 ms tails. On a 20-minute realtime
  benchmark of ~1.08 M txns the steady-state publish_rpc p99 is 334 ms; only
  141 events crossed the 500 ms threshold (≈ 0.013 % of all VISIBLE lines).
- No change to the BRPC call site, thread-pool config, or the existing
  \`publish rpc cost\` field.

## Test plan
- [x] Compiles in TSP build (will be exercised by next auto-perf-opt iteration).
- [x] Steady-state log volume unchanged on a 20-min benchmark — verified by the
  500 ms threshold being far above the observed p99 (334 ms).
- [x] Slow events (≥ 500 ms total) emit exactly one WARN line per partition,
  with all four sub-phase fields summing to (within ~µs of) total_ms.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
@@ -1000,8 +1000,11 @@ public class PublishVersionDaemon extends FrontendDaemon {
             return CompletableFuture.completedFuture(false);
         }
 
+        long submitTimeMs = System.currentTimeMillis();
         return CompletableFuture.supplyAsync(() -> {
-            boolean success = publishPartition(db, tableCommitInfo, partitionCommitInfo, txnState);
+            long lambdaEntryMs = System.currentTimeMillis();
+            boolean success = publishPartition(db, tableCommitInfo, partitionCommitInfo, txnState,
+                    submitTimeMs, lambdaEntryMs);
             partitionCommitInfo.setVersionTime(success ? System.currentTimeMillis() : -System.currentTimeMillis());
             return success;
         }, getTaskExecutor()).exceptionally(ex -> {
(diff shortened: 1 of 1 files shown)


luohaha:

PR #72572, opened 2026-05-03
[auto-perf-opt] Log per-phase breakdown of slow Utils.publishVersionBatch

luohaha:
## Summary
Add a 4-LOC WARN log line in `com.starrocks.lake.Utils.publishVersionBatch` that fires only when total wall-clock ≥ 500 ms, breaking the call into three sub-phases:
- `process_tablets_ms` — `processTablets(...)`: tablet-to-node grouping under `WarehouseManager` / `GlobalStateMgr` lookups.
- `rpc_submit_ms` — the per-node submit loop (`BrpcProxy.getLakeService` + async `lakeService.publishVersion`).
- `rpc_wait_ms` — `responseList.get(i).get()` loop = `max(BE response time)` including client-side response receipt.

Steady-state publishes (< 500 ms) log nothing — log volume on the hot path is unchanged.

## Motivation
PR #72557 (iter-025) added per-partition phase logging in `PublishVersionDaemon.publishPartition` and proved **100 % of slow publishPartition wall time lives in `brpc_ms`** (= `Utils.publishVersion(...)`) — none in executor-acquire, lock-wait, or fe-prep. Across iter-026's 20-min benchmark this fired on **10,448 in-test events** (p99 = 10.9 s, max = 25.3 s), with 75 % concentrated in the first 3 min after a fresh deploy.

Combined with iter-024's evidence (BRPC server-side queueing < 0.1 % of slow latency; max BE handler cost on slow events = 200-300 ms while FE wall = 1.5-25 s), the bottleneck localizes inside `publishVersionBatch` itself but the existing measurement window cannot tell us **which of its three sub-phases dominates**.

This patch surfaces the dominant sub-phase the next time the cold-start storm reproduces, so the actual fix can be targeted (per the program's "trace over guess" rule).

## Performance
- 4 `System.currentTimeMillis()` calls per `publishVersionBatch` invocation (+ one threshold branch).
- One `LOG.warn` only on ≥ 500 ms tails. No change to BRPC, no change to thread-pool config, no change to existing log fields.

## Test plan
- [ ] Build via TSP and deploy to my-rt-2 cluster.
- [ ] Re-run the 20-min `999_1gb_1_1tb` benchmark; trigger a fresh-deploy cold-start storm.
- [ ] `grep "Slow publishVersionBatch" fe.log*`; bin events by which sub-phase carries ≥ 50 % of `total_ms`.
- [ ] Cross-correlate with iter-025's `Slow publishPartition` lines (matching by timestamp + node count) to confirm the same events surface in both logs.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/Utils.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/Utils.java
@@ -135,6 +135,7 @@ public class Utils {
                                            ComputeResource computeResource,
                                            Map<Long, Long> tabletRowNum)
             throws NoAliveBackendException, RpcException {
+        long entryMs = System.currentTimeMillis();
         WarehouseManager warehouseManager = GlobalStateMgr.getCurrentState().getWarehouseMgr();
         if (!warehouseManager.isResourceAvailable(computeResource)) {
             LOG.warn("publish version operation should be successful even if the warehouse is not exist, " +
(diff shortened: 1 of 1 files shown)


luohaha:
Closing: this PR instruments `com.starrocks.lake.Utils.publishVersionBatch`, but auto-perf-opt my-rt-2 iter-027 verified the cluster runs with `enable_file_bundling=true` (and all benchmark tables have `file_bundling=true` property), so FE actually calls `Utils.aggregatePublishVersion` (Utils.java:236) instead of `Utils.publishVersionBatch` (Utils.java:233). The instrumentation produced 0 'Slow publishVersionBatch' events in a 20-min benchmark across fe.log + fe.warn.log + 4 rotated fe.log.20260504-{1..4} files. Not useful as merged code; superseded by PR #72574 which targets the actual cold-start bottleneck (`rebuild_index_del_cost_us` in `LakePersistentIndex::load_dels`).

PR #72577, opened 2026-05-03
[auto-perf-opt] Instrument Table::MultiGet t3 sub-phases (filter / block-read / search)

luohaha:
## Why

`my-rt-2` 999_1gb_1_1tb realtime-benchmark cold-start publish-version traces at iter-028 (TSP build 1416, `branch-4.1-b0a9851`, this branch's parent + iter-028's parallel `LakePersistentIndex::load_dels`) showed:

- **upsert max 25 s** (8.3× over the 3 s primary goal); cold-start storm in first ~30 s after fresh deploy.
- Among 1882 BE Published events with cost ≥ 1 s, **`multi_get` ≥ 50 % of cost in 71.8 %** of them. `pidx_load ≥ 70 %` only 33.5 %. `multi_get` is the dominant cold-start cost.
- Top-10 slowest events: `multi_get_us` 4-7 s out of 17-18 s total handler cost.

`Table::MultiGet` already emits coarse `multiget_t1_us / t2_us / t3_us`. `t3_us` is the section that does the bloom-filter check, `BlockReader()` (cache-hit OR OSS-read), and `search_in_block()`. We can't tell from existing counters whether the cold tail is OSS-RTT-bound, bandwidth-bound, bloom-filter-CPU-bound, or in-memory-search-bound.

The iter-015 `Table::Open` tail-prefetch attempt (#72476 v4) **regressed** P99 by +78 % because the same-shape OSS path turned out to be bandwidth-bound, not RTT-bound. We must NOT propose another `MultiGet` parallelisation without first measuring which subcomponent dominates.

## What this PR does

Splits `multiget_t3_us` into three phases inside the existing per-key loop in `Table::MultiGet`:

| New counter                          | Measures                                                                 |
|--------------------------------------|--------------------------------------------------------------------------|
| `multiget_t3_filter_us`              | bloom-filter `KeyMayMatch()` time                                        |
| `multiget_t3_block_read_us`          | total `BlockReader()` wall-time (cache hit + miss)                       |
| `multiget_t3_block_read_miss_us`     | `BlockReader()` wall-time only on cache miss (= actual OSS read)         |
| `multiget_t3_search_us`              | `search_in_block()` time on the freshly-loaded block                     |
| `multiget_block_read_miss_cnt`       | number of cache-miss BlockReader calls                                   |

Cache-miss detection uses the existing `ReadIOStat::block_cnt_from_file` counter delta around the `BlockReader` call — no new mutexes / atomics.

[...]

PR #72579, opened 2026-05-04
[auto-perf-opt] Parallelize get_from_sstables across filesets

luohaha:
## Bottleneck (iter-030 of auto-perf-opt my-rt-2 run)

Cold-start cache-miss block reads from OSS dominate the publish-version critical path on shared-data PK tables.

### Evidence (1,046,417 Published events parsed across 6 BEs over 20-min fresh-deploy benchmark)

- Slow >= 1s events: **3,578 (0.34%)**
- Of those, 77.1% are multi_get-dominated (2,759 events)
- On multi_get-dominated slow events:
  - `multiget_t3_us / multi_get_us` p50 = **99.8%** (t3 IS multi_get)
  - `multiget_t3_block_read_miss_us / multiget_t3_us` p50 = **99.9%** (block-read on cache miss IS t3)
  - `multiget_t3_filter_us / multiget_t3_us` p50 = 0.0% (bloom filter eval is negligible)
  - `multiget_t3_search_us / multiget_t3_us` p50 = 0.0% (in-block search is negligible)
- Dominant t3 sub-component: `block_read_miss_oss` in **100%** of multi_get-dominated slow events
- Per-miss latency: p50 = **99 ms** / p95 = 307 ms / max = 710 ms (high — OSS first-byte service latency)
- Per-event miss count: p50 = 20, p95 = 37, max = 435

The new t3_* counters were introduced by #72577 specifically to disambiguate this matrix. They show that the cost is purely OSS service latency on cache-miss block reads, not filter CPU, in-memory search, or bandwidth saturation.

### Code path

`LakePersistentIndex::get_from_sstables` walks `_sstable_filesets` SEQUENTIALLY (rbegin -> rend). Each fileset's `multi_get` does its own sequential cache-miss block reads. With ~5-20 filesets per tablet on a fresh deploy and ~99 ms per miss, total publish wall time stacks to 5-25s.

## Change

When `config::enable_pk_index_parallel_execution` is true (existing flag, default true) and there are multiple filesets, dispatch each fileset's `multi_get` on the existing `pk_index_execution_thread_pool` and merge results so the most-recent fileset's value wins (preserves the original rbegin -> rend ordering semantics).

The `set_difference` shortcut is dropped inside the parallel path. On cold-start most candidate keys are rejected cheaply by per-fileset bloom filters, so the extra work is small compared with the wall-time win from overlapping OSS first-byte latency.

## Risk

Risk-managed against the iter-015 trap (large-tail prefetch parallelised on `Table::Open` turning out to be bandwidth-bound on a 2 MB Bloom filter): individual block reads here are small (~32 KB; per-event total ~640 KB for p50 miss-count of 20), so concurrency overlaps OSS service latency rather than saturating bandwidth. Behind the same flag as the earlier parallel-load_dels work (#72574), so a single toggle reverts if regression appears.

## Expected impact (999_1gb_1_1tb fresh-deploy)

- `multi_get_us` p50 on slow events: ~1.84 s -> < 0.5 s
- BE Published cost p99: 7.81 s -> < 2.5 s
- Upsert max latency: 25.9 s -> aim < 10 s
- Steady-state P99 / throughput: unchanged (slow events are 0.34% of all publishes; steady state already hits cache)

A/B will be measured in iter-031 with this PR built and deployed.

### DIFF
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -36,6 +36,7 @@
 #include "storage/sstable/merger.h"
 #include "storage/sstable/options.h"
 #include "storage/sstable/table_builder.h"
+#include "util/thread.h"
 #include "util/trace.h"
 
 namespace starrocks::lake {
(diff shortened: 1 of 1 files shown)


luohaha:
Closing — superseded by #72605, which is the same parallel `get_from_sstables` logic but on a dedicated `pk_index_inner_io` pool. Merging this PR standalone would reproduce the iter-031 same-pool re-entrance crash that #72605 was designed to avoid. The validated parallel-fileset implementation lives in #72605's stack.

PR #72460, opened 2026-05-02
[auto-perf-opt] Drop Slice::to_string() heap allocs in PK-index compaction merge loop

luohaha:
## Why

`999_1gb_1_1tb` realtime-benchmark iter-005 on `branch-4.1 + #72434` (PR that hoists `IndexValuesWithVerPB` scratch buffers to KeyValueMerger members) shows the cluster has lost most of #72434's gains under sustained ingest:

| Metric | iter-003 | iter-005 (same build) |
|---|---:|---:|
| Upsert P99 | 1297 ms | 1531 ms (+18 %) |
| Throughput | 788 K rows/s | 586 K rows/s (-26 %) |
| Hottest CN steady-state CPU | 47-67 % | 91-99 % |

A 60 s `perf -F 99 -e cpu-clock` profile mid-test on the hottest CN (172.26.80.125) attributes 24 % of total CPU self-time to allocator work (`my_malloc 12.14 % + my_free 11.91 %`). Inclusive view, `cloud_native_pk_index_compact` thread family:

```
31.15 % start_thread
23.30 %  LakePersistentIndexParallelCompactTask::run
 19.90 %   do_run
   8.81 %    KeyValueMerger::merge
     4.70 %   KeyValueMerger::flush
   6.55 %    operator new (do_run direct)  -> cur_key.to_string()
   3.12 %    my_free      (do_run direct)  -> ~std::string
```

The `IndexValuesWithVerPB` lifecycle that #72434 fixed is *no longer* the dominant allocator consumer. Three remaining `Slice::to_string()` calls in the PK-index merge hot path now own most of the residual allocator self-time:

- `KeyValueMerger::merge` &mdash; converts the iterator-owned key/value slices into fresh `std::string`s before calling `ParseFromString` and the equality check.
- `LakePersistentIndexParallelCompactTask::do_run` &mdash; converts `merging_iter->key()` to a `std::string` only to feed it back into a `Slice`-based comparator.
- `LakePersistentIndex::merge_sstables` &mdash; same pattern.

Each call is a per-input-row heap allocation pair against an iterator whose slice storage is stable until the next `->Next()`.

## What this changes

Keep the iterator-owned `Slice` as-is in all three sites. The protobuf parse goes through `ParseFromArray(value.data, value.size)` to drop the per-call value copy; equality against the cached `_key` becomes `Slice(_key) == key`; the only remaining string write is the rare `_key.assign(key.data, key.size)` on the branch where the key actually changes (once per distinct output key, not once per input row).

```
[...]

PR #72434, opened 2026-05-02
[auto-perf-opt] Reuse protobuf scratch buffers in KeyValueMerger to cut PK-index compaction CPU

luohaha:
## Summary

Cloud-native PK-index compaction (`cloud_native_pk_index_compact`) burns ~38% of all BE/CN CPU in `my_free` (jemalloc) under sustained upsert workloads. A 60-second `perf` capture on the hottest CN of an `auto-perf-opt my-rt-2` 999_1gb_1_1tb run (6×32-core CNs, all at 99-100% non-idle CPU) traces it back to per-key allocator churn inside `starrocks::lake::KeyValueMerger`:

```
78.91% LakePersistentIndexParallelCompactTask::run
 55.26%  KeyValueMerger::merge
  26.59%  KeyValueMerger::flush
   9.63% my_free                                 # dtor of local index_value_pb
   9.62% operator new                            # RepeatedField growth
   4.92% ~IndexValuesWithVerPB
   1.76% sstable::TableBuilder::Add
  11.62% my_free                                 # dtor of local index_value_ver
  11.18% MessageLite::ParseFromString  (10.72% in operator new)
   5.45% ~IndexValuesWithVerPB
```

Both `merge()` and `flush()` allocate a fresh `IndexValuesWithVerPB` on the stack on every call and let the destructor free all of the protobuf's internal `RepeatedField` storage at scope exit. The merger is constructed once per compact task and called once per key from the merging iterator — millions of malloc/free pairs per task. At the per-task exec time of 46.6 s observed in `auto-perf-opt my-rt-2` iteration 001, allocator churn alone accounts for the majority of compaction CPU.

## Change

Hoist the two `IndexValuesWithVerPB` instances and the `SerializeAsString()` output buffer onto `KeyValueMerger` as members, and `Clear()` / `SerializeToString()` into them per call. Clear() preserves `RepeatedField` capacity, so subsequent `ParseFromString` reuses already-grown internal arrays instead of going back to the allocator. The merger is single-threaded within a task, so no synchronization is needed.

Files: `be/src/storage/lake/lake_persistent_index_key_value_merger.{h,cpp}` — 2 files, +15 / -4 lines.

## Expected impact

- **CPU**: removing in-loop allocator churn should shave 30-50% off `cloud_native_pk_index_compact` per-task exec time.
[...]

PR #72591, opened 2026-05-04
[auto-perf-opt] iter-031 stack-build: PR #72579 deadlock fix

luohaha:
## Build slot for iter-031 with deadlock fix on PR #72579

iter-030's TSP build 1419 (= post-iter-021 superset + #72557 + #72574 + #72577 + #72579) crashed all 6 BEs ~30 s into iter-031's benchmark with a same-pool re-entrance abort triggered by ThreadPool::check_not_pool_thread_unlocked().

This branch is a clean rebase of iter-030-stack with one extra commit that adds a defensive check to PR #72579's `get_from_sstables` parallel path, avoiding the recursive submit-and-wait when the caller is itself a `cloud_native_pk_index_execution` worker.

### Root cause

LakePersistentIndex::upsert(...) submits a lambda to `ParallelPublishContext::token`, which is built on `pk_index_execution_thread_pool` (lake_primary_index.cpp:424,605 and update_manager.cpp:931,1069). That lambda calls `get_from_sstables`. Under PR #72579, `get_from_sstables` opens a token on the SAME pool and waits — `ThreadPool::check_not_pool_thread_unlocked()` aborts because a worker can't safely wait on its own pool.

### Fix

Detect `Thread::current_thread()->name() == "cloud_native_pk_index_execution"` and fall back to the sequential per-fileset loop for the on-pool case. Sync caller paths (`LakePersistentIndex::get` / `erase` / non-ctx `upsert`) still get the parallel optimization.

### Trade-off

Cold-start hot path is the parallel-publish upsert path, which now stays sequential within `get_from_sstables`. A follow-up will move inner parallelism to a separate pool (e.g. `lake_metadata_fetch_thread_pool`) so the async-upsert path can also benefit.

### Stack contents

- post-iter-021 validated stack (KeyValueMerger #72434+#72460, _index_cache TTL #72503, SO_REUSEPORT #72554)
- BRPC queue logging #72555
- FE publishPartition phases #72557
- Parallel load_dels #72574
- Table::MultiGet t3 sub-phase counters #72577
- Parallel get_from_sstables #72579 + iter-031 deadlock fix (this commit)

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -526,6 +526,13 @@ CONF_Int32(be_http_port, "8040");
 CONF_Alias(be_http_port, webserver_port);
 // Number of http workers in BE
 CONF_Int32(be_http_num_workers, "48");
+// When true, each HTTP worker creates its own listening socket with
+// SO_REUSEPORT so the kernel load-balances incoming connections in-kernel,
+// instead of all workers epoll-waiting on a single shared listen fd and
+// thundering-herd-racing on accept(). Eliminates the kernel
+// inet_csk_accept / native_queued_spin_lock_slowpath contention that
+// dominates HTTP-server CPU when be_http_num_workers is large.
+CONF_Bool(enable_http_server_so_reuseport, "true");
 // Period to update rate counters and sampling counters in ms.
 CONF_mInt32(periodic_counter_update_period_ms, "500");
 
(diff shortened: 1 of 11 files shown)


luohaha:
Closing — superseded by #72605, which fixes the same-pool re-entrance with a structural separate-pool design instead of a defensive Thread::current_thread() check. Cleanup of auto-perf-opt my-rt-2 run.

PR #72626, opened 2026-05-04
[BugFix] Drain runtime_filter worker without joining forwarded RPCs

### LINKED ISSUE #71986: [Enhancement] CN bottleneck in runtime_filter during high concurrency
## Enhancement

Under high concurrency we observe the runtime filter worker queue backing up considerably — we've seen runtime_filter_event_queue_len exceed 1.1M events and runtime_filter_bytes_in_queue exceed 60 GiB on a single CN, growing unboundedly until we intervene.

The drain is single-threaded (RuntimeFilterWorker::execute, one std::thread) and the hot event type RECEIVE_TOTAL_RF is dominated by synchronous RPC fan-out. _receive_total_runtime_filter (be/src/runtime/runtime_filter_worker.cpp:957) issues forward RPCs for the broadcast tree, then BatchClosuresJoinAndClean's destructor (line 533) joins every closure via bthread_id_join before the function returns — so the drain thread cannot pull the next event until every forward completes. Stack sample from a stalled worker:

```
#1  bthread::wait_pthread
#2  bthread::butex_wait
#3  bthread_id_join
#4  BatchClosuresJoinAndClean::~BatchClosuresJoinAndClean
#5  RuntimeFilterWorker::_receive_total_runtime_filter
#6  RuntimeFilterWorker::execute()
```
The thread was 85% in S state with 23.9M voluntary context switches while 60+ GiB of events sat queued. transmit_runtime_filter_concurrency was 0 during the stall — outbound RPCs complete quickly (p99 = 2.6 ms) but the worker spends its time parked in bthread_id_join between events rather than draining the queue.

There is no lever to parallelize drain or to make the forward step asynchronous. runtime_filter_queue_limit caps queue size but does not address throughput. Our only effective mitigation has been set global enable_global_runtime_filter = false, which we'd like to avoid.

Hyper-FF:
## Why I'm doing:

The single RuntimeFilterWorker thread previously blocked on bthread_id_join for every RPC it dispatched (BatchClosuresJoinAndClean / SingleClosureJoinAndClean destructors), which limited drain throughput to one full RPC round-trip per event. Under high concurrency this caused the worker queue to back up by >1.1M events / 60+ GiB on a single CN.

Convert the four worker-thread RPC paths
(_send_total_runtime_filter, _receive_total_runtime_filter forwarding, _deliver_broadcast_runtime_filter_relay, _deliver_part_runtime_filter) to fire-and-forget via a new AsyncRuntimeFilterRpcClosure that logs RPC errors and self-deletes when brpc invokes Run(). The worker thread can now drain queued events without waiting on RPC completion.

The passthrough broadcast path keeps its sync batching because it uses deliver_broadcast_rf_passthrough_inflight_num to bound in-memory copies of large broadcast payloads.

Detach the dispatch helper from any query mem tracker the caller may have set: brpc frees the closure and its serialization buffers from a bthread with no thread-local tracker, so alloc and free now both land on the process default tracker instead of drifting the query's accounting.

## What I'm doing:

Fixes #71986

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/runtime/runtime_filter_worker.cpp
+++ b/be/src/runtime/runtime_filter_worker.cpp
@@ -550,9 +550,9 @@ struct AsyncRuntimeFilterRpcClosure : public RuntimeFilterRpcClosure {
     }
 };
 
-static void submit_async_runtime_filter_rpc(const RpcServices* rpc_services, const TNetworkAddress& dest, int timeout_ms,
-                                            int64_t http_min_size, const PTransmitRuntimeFilterParams& request,
-                                            const char* debug_info) {
+static void submit_async_runtime_filter_rpc(const RpcServices* rpc_services, const TNetworkAddress& dest,
+                                            int timeout_ms, int64_t http_min_size,
+                                            const PTransmitRuntimeFilterParams& request, const char* debug_info) {
     // The closure and brpc's internal serialization buffers are released by a
     // brpc bthread after this function returns, where no thread-local mem
     // tracker is set. Callers usually run under a query mem tracker
--- a/be/src/runtime/runtime_filter_worker.cpp
+++ b/be/src/runtime/runtime_filter_worker.cpp
@@ -537,21 +537,43 @@ private:
     DISALLOW_COPY_AND_MOVE(BatchClosuresJoinAndClean);
 };
 
-struct SingleClosureJoinAndClean {
-public:
-    SingleClosureJoinAndClean(RuntimeFilterRpcClosure* closure) : _closure(closure) {}
-    ~SingleClosureJoinAndClean() {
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

trueeyu (approved):

stdpain (approved):

PR #72667, opened 2026-05-04
[auto-perf-opt] Skip TabletMetadataPB deep copy in LakeDelvecLoader

luohaha:
## Summary

`TabletManager::get_tablet_metadata(tablet_id, version, ...)` always performs
`std::make_shared<TabletMetadata>(*ptr)` on its result and then `set_id(tablet_id)`,
which deep-copies every nested `RowsetMetadataPB` / `SegmentMetadataPB` / `TuplePB`
/ `VariantPB` through protobuf `MergeFrom`.

Cold-start `cpu-profile` (iter-040, host `172.26.80.130`, 25 s, branch-4.1-1d30cd1)
shows that copy + the matching destructor accounting for **~6.5 % inclusive CPU**
on the `cloud_native_co` (vertical compaction) thread family:

```
6.32 %  VerticalCompactionTask::compact_column_group
  5.13 %  TabletReader::open -> init_collector -> get_segment_iterators
    5.10 %  Segment::new_iterator -> SegmentIterator
      5.08 %  LakeDelvecLoader::load -> load_from_file
        2.96 %  TabletManager::get_tablet_metadata (tablet_id overload)
          2.94 %  TabletMetadataPB(const&) constructor
            2.83 %  MergeFromInnerLoop<RowsetMetadataPB>
              2.80 %  RowsetMetadataPB::MergeFrom
                2.63 %  MergeFromInnerLoop<SegmentMetadataPB>
                  2.60 %  SegmentMetadataPB::MergeFrom
                    1.86 %  MergeFromInnerLoop<TuplePB>
                      1.80 %  TuplePB::MergeFrom -> 1.51 %  VariantPB::MergeFrom
        2.08 %  ~RepeatedPtrFieldBase / Destroy<RepeatedPtrField<...>>
```

The path-based overload `TabletManager::get_tablet_metadata(path, ...)`
returns the cached shared `TabletMetadataPtr` without any copy. Both overloads
run the same aggregation-vs-single-tablet logic internally, so this PR computes
the path ourselves in `LakeDelvecLoader::load_from_file` and routes through the
path-based call.

The use is read-only — `metadata` flows into `lake::get_del_vec(_tablet_manager,
*metadata, ...)` which takes `const TabletMetadata&`. No mutation, so sharing
the cached `shared_ptr` is safe.

This change is **localised to LakeDelvecLoader** — other callers of the
tablet_id-based overload (transactions.cpp, update_manager.cpp, etc.) still
[...]

PR #72654, opened 2026-05-04
[auto-perf-opt] Replace std::list with std::optional in KeyValueMerger

luohaha:
## Why

`999_1gb_1_1tb` iter-039 fresh-deploy on TSP build 1427 (= post-iter-038 stack: PR #72639 + #72622 + #72619 + #72606 + #72591 + post-iter-021) reproduced the cold-start storm with cluster CPU saturated **98.4–98.8 % peak** and vdb util **84–93 %** on every BE.

A 25 s `perf -F 99 -e cpu-clock` profile on the hottest CN (172.26.80.128) inside the 21:05–21:10 cold-start window shows the parallel PK-index compaction pool (`cloud_native_pk_index_compact`) holding **42.23 %** of inclusive CPU, with allocator self-time at **31 %** across `my_malloc` + `my_free`. Children view, top stack:

```
42.23 %  LakePersistentIndexParallelCompactTask::run
  42.09 %  do_run
    23.05 %  KeyValueMerger::merge
      19.93 %  KeyValueMerger::flush
        11.38 %  my_free                    ← list node free + scratch
         6.40 %  TableBuilder::Add
    13.77 %  operator new (do_run direct)
      13.52 %  my_malloc                    ← list node alloc + per-row temps
```

Both prior optimisations on this file (#72434 reused the protobuf scratch buffers; #72460 dropped `Slice::to_string()`) are in this build, but the remaining allocation pair was hidden inside the `std::list<IndexValueWithVer> _index_value_vers` member.

## Root cause

Every `emplace_front` on the list heap-allocates one node, and every `flush()` ends with `_index_value_vers.clear()` which heap-frees it — **once per input row** of the merge. On a 1 TB tablet that is millions of paired allocations per compaction task on a thread pool sized `num_cores / 2`.

Tracing every mutation site in `merge()` and `flush()` shows the list **never** holds more than one entry — it is either empty or contains the single "winning" `(version, IndexValue)` for the current key:

```
flush() ends with _index_value_vers.clear()                -> empty
merge() empty-branch:        emplace_front(...)            -> 1 elem
merge() same-key replace:    std::list t; t.emplace_front; swap(t)  -> 1 elem
merge() key-changed:         flush(); emplace_front(...)   -> 1 elem
```

## What this changes

- `_index_value_vers` (`std::list`) → `_current_value` (`std::optional<IndexValueWithVer>`).
- `emplace_front` on the list → `emplace` on the optional. The swap-with-temp pattern collapses to a direct `emplace`.
[...]

PR #72639, opened 2026-05-04
[auto-perf-opt] Move SST opens to a dedicated pk_index_sst_open pool

luohaha:
## Why

Stacked on top of #72619 (chunk-parallel `multi_get`) + #72622 (`pk_index_inner_io` size 128/CN).

Cold-start fresh-deploy publish-trace (iter-036, 529 slow events ≥ 1 s) shows the dominant remaining wall-clock component on slow events is the **parallel SST-open phase**, not the multi-get block reads:

| Counter | p50 | p99 | max |
|---|---:|---:|---:|
| `primary_index_load_latency_us` | wall-clock | **12.96 s** | 16.1 s |
| `pindex_init_sst_open_us` | (wall-clock for `_open_sstables_parallel`) | **8.88 s** | **11.9 s** |
| `multiget_t3_block_read_miss_us` | (sum across chunks) | 12.98 s | 34 s |

The slowest 16 events all spend 8-12 s in `pindex_init_sst_open_us` and only 0-3 s in chunk-summed `multiget_t3_block_read_miss_us`. The wall-clock contribution of the multi-get path is much smaller than the SST-open phase, because PR #72619+#72622 made multi-get fan out across 128 chunk_io threads — but **SST opens still queue head-of-chain through a small CPU-leaning pool**.

`LakePersistentIndex::_open_sstables_parallel` dispatches one task per SST on `pk_index_execution_thread_pool`. Each task is `PersistentIndexSstable::new_sstable → Table::Open`, which issues 4 sequential OSS round-trips per SST: footer (53 B) → index block → metaindex → filter block.

`pk_index_execution_thread_pool` is sized `num_cores()/2 = 16/CN`. It also serves parallel-publish workers (`parallel_upsert`, `parallel_get`, `condition_merge_func`, `batch_parallel_get_rss_rowids`). Under cold-start storm:

- ~150 SST-open tasks queue at once across all coordinating tablets (≈ 27 SSTs × 5+ tablets × multi-CN parallel publish).
- Each task takes 4 × ~92 ms ≈ 400 ms best-case; under heavy concurrent OSS load, individual RTs balloon to 500+ ms each, so per-task is closer to 2 s.
- The 16-thread pool drains 150 tasks in ~150/16 × 2 s ≈ 19 s **and** competes with parallel-publish workers running on the same pool.

The observed 11.9 s wall-clock on the affected hot publish-coord CN (172.26.80.128 in iter-036) matches that picture.

## What

[...]

PR #72622, opened 2026-05-04
[auto-perf-opt] Bump pk_index_inner_io pool default to num_cores * 4

luohaha:
## Context

This PR is the iter-035 follow-up on top of the iter-034 PR #72619 (chunk-parallel `multi_get` into `pk_index_chunk_io` pool).

iter-035 fresh-deploy of build 1423 (PR #72619 alone, on top of #72605/#72591/post-iter-021) on the `999_1gb_1_1tb` cluster validated the chunk-parallel optimization at the publish-trace level:

| signal | iter-033 (build 1422) | iter-035 (build 1423, +#72619) |
|---|---|---|
| BE Published slow ≥1s count   | 815                  | **159 (−81%)** |
| FE Slow publishPartition count | 23,289 (iter-032)   | **1,951 (−92%)** |
| FE total_ms max               | 35,517 ms (iter-032) | **3,305 ms (−91%)** |
| multiget_block_read_miss_cnt p99 | 153              | **32 (−79%)** |

## Why bump `pk_index_inner_io` next

Prometheus on the iter-035 cold-start window (`16:04–16:08`):

| metric (per CN)                                         | 172.26.80.126 | 172.26.80.130 | 172.26.80.128 |
|--------------------------------------------------------|---------------|---------------|---------------|
| `cloud_native_pk_index_inner_io_active_threads` max    | 32 / 32 ✘     | 32 / 32 ✘     | 29 / 32       |
| `cloud_native_pk_index_inner_io_queue_count` max       | **245**       | **234**       | 2             |
| `cloud_native_pk_index_inner_io_pending_time` avg/task | 12 ms         | **252 ms**    | 108 ms        |
| `cloud_native_pk_index_inner_io_execute_time` avg/task | 11 ms         | 60 ms         | 51 ms         |
| `cloud_native_pk_index_chunk_io_active_threads` max    | 51 / 128      | 23 / 128      | 61 / 128      |
| `cloud_native_pk_index_chunk_io_queue_count` max       | 2             | 2             | 6             |

The `chunk_io` pool below has plenty of headroom; the new ceiling is the `inner_io` pool above. Two of six CNs hit pool saturation with 234–245 tasks queued during the cold-start storm.

## Why oversubscribe is correct here

[...]

PR #72671, opened 2026-05-04
[auto-perf-opt] Add pindex_sstables_queried_cnt trace counter

luohaha:
## Bottleneck

Cold-start publish traces on the `999_1gb_1_1tb` real-time benchmark
template (1 large 1 TB upsert table + 999 small 1 GB tables, 6-CN
shared-data cluster) show that even after the cumulative parallel-publish
and PK-index wins (`#72434`, `#72460`, `#72503`, `#72554`, `#72591`,
`#72606`, `#72619`, `#72622`, `#72639`, `#72654`, `#72667`),
`pindex_init_sst_open_us` p99 stays around 9.3 s on every cold-start
event:

```
pindex_init_sst_cnt        p50=25  p99=39   max=85
pindex_init_sst_open_us    p99=9.3 s  max=10.5 s
primary_index_load_latency_us p99=15.0 s  max=34.0 s
```

i.e. `LakePersistentIndex::init` eagerly calls
`_open_sstables_parallel` for **every** SST in the tablet's
`sstable_meta` before any query arrives, and that batch dominates the
critical path.

## Why this PR is instrumentation, not a fix

A natural structural fix is **lazy SST open**: defer each SST's
footer / index / metaindex / filter loading until the first `multi_get`
actually walks into that SST. Whether the win is large or small
depends on the ratio of SSTs **actually queried** to SSTs **opened**:

- If a typical cold-start publish only queries 1-3 SSTs out of 25-39
  opened, lazy open is a big win.
- If queries spread evenly across all opened SSTs, the win is much
  smaller.

There is currently no way to tell from existing trace data which
regime we are in. This PR adds the missing counter so the next
iteration can decide quantitatively, instead of guessing.

## Change

In `PersistentIndexSstableFileset::multi_get` (the inner per-fileset
function) add:

```cpp
TRACE_COUNTER_INCREMENT("pindex_sstables_queried_cnt", N);
```

where `N` is:

- **1** when the fileset is a single standalone SST (every multi_get
  on it touches that one SST), or
- **`sstable_key_indexes_map.size()`** for a sharded fileset — i.e.
  only the SSTs whose key range actually covers ≥ 1 of the query
  keys, after the existing per-fileset divide step. SSTs that the
  divide step rules out are not counted, because they are not read.

The counter is per-publish (TRACE_COUNTER on the current trace),
matching the level at which `pindex_init_sst_cnt` is already
recorded. Comparing the two in publish-trace output gives the queried
/ opened ratio.

## Behavior

Pure instrumentation:

- No control-flow change, no allocation in the hot path beyond the
  existing trace counter machinery.
- Zero overhead when no trace is active (`TRACE_COUNTER_INCREMENT`
  short-circuits if `Trace::CurrentTrace()` is null).
- Output adds one new field `pindex_sstables_queried_cnt` to slow
  publish traces; existing fields are unchanged.
- On-disk format unchanged. Public API unchanged. No new config flag.

## Stack

Iteration 042 of the `my-rt-2` auto-perf-opt run. Stacked on the
iter-041 cumulative branch (`b1a09f9b7d`), which carries:

- `#72667` LakeDelvecLoader skip TabletMetadataPB deep copy
- `#72654` KeyValueMerger `std::list` → `std::optional`
- `#72639` dedicated `pk_index_sst_open` ThreadPool (128 / CN)
- `#72622` `pk_index_inner_io` pool 32 → 128 / CN
- `#72619` chunk-parallel multi_get on PK index
- `#72606` separate `pk_index_inner_io` pool
- `#72591` deadlock fix in `get_from_sstables`
- post-iter-021 baseline (#72434 / #72460 / #72503 / #72554)

## Validation plan

After merge, iter-043 will deploy this build, run a fresh-deploy
benchmark, and compare distributions of
`pindex_sstables_queried_cnt` vs `pindex_init_sst_cnt` over the
cold-start window:

- Ratio < 0.3 → schedule lazy SST open as iter-044's structural fix.
- Ratio ~ 1 → reconsider; pivot to compaction-side coalescing
  (fewer-but-larger SSTs).

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
+// Minimum number of keys in one PersistentIndexSstable::multi_get for the chunk-parallel
+// path to engage. Below this, run the legacy sequential path so we don't pay open-file
+// + dispatch overhead for tiny lookups.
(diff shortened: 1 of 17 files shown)


luohaha:

PR #72672, opened 2026-05-04
[auto-perf-opt] Add pindex_unique_sstables_touched_cnt trace counter

luohaha:
## Why

PR #72671 added `pindex_sstables_queried_cnt` to determine whether a lazy-SST-open refactor would skip work. Iter-043 cold-start data on 849 slow events shows:

- `pindex_init_sst_cnt` p50 = 25, p99 = 39 (sstables eagerly opened per cold-start event)
- `pindex_sstables_queried_cnt` p50 = 80, p99 = 265
- ratio queried/init p50 = **2.96** mean = **3.72** max = **10.21**

The ratio is >> 1 because `pindex_sstables_queried_cnt` accumulates across `multi_get(fileset, chunk)` calls and double-counts the same sstable when revisited from different chunks. It cannot be read directly as "fraction of opens ever touched" — that was the iter-044 decision lever per the human operator's [2026-05-04 23:45] guidance.

## What

Add `pindex_unique_sstables_touched_cnt`: the size of an `unordered_set<const PersistentIndexSstable*>` populated as `LakePersistentIndex::get_from_sstables` walks filesets. Each `PersistentIndexSstableFileset::multi_get` call inserts the sstables it actually walks into. The set is emitted as a single trace counter once per cold-start event, after all filesets have been processed.

Direct interpretation, ratio = `pindex_unique_sstables_touched_cnt / pindex_init_sst_cnt`:

- close to 1.0 → all opened sstables get touched → lazy-open does **not** skip work; pivot to compaction-side coalescing or data-path fix.
- close to 0.0 → most opened sstables are dead weight → lazy-open is the structural fix to ship.

The implementation:

- `persistent_index_sstable_fileset.h` — adds an optional `std::unordered_set<const PersistentIndexSstable*>* touched_sstables = nullptr` out-parameter to `multi_get`. Default nullptr keeps existing call sites unaffected.
- `persistent_index_sstable_fileset.cpp` — both standalone and sharded paths insert touched sstables when the out-parameter is non-null.
- `lake_persistent_index.cpp` — `get_from_sstables` owns the set, threads it through both the sequential path and the `pk_index_inner_io_thread_pool` parallel path (per-fileset local sets merged into master after `token->wait()`), then emits `TRACE_COUNTER_INCREMENT("pindex_unique_sstables_touched_cnt", touched.size())` exactly once per call.

## Risk

Pure instrumentation; no semantic change. Set insertions are bounded by `pindex_init_sst_cnt` (typical 25-39 per event) — negligible cost vs the OSS reads on the same path. No overhead when no trace context is active (the `TRACE_COUNTER_INCREMENT` macro is a no-op then).

## Acceptance for iter-044

Iter-044 deploys this and reads `pindex_unique_sstables_touched_cnt / pindex_init_sst_cnt` distribution on cold-start events. Pre-registered direction:

- p50 unique/init < 0.30 → ship lazy-SST-open as iter-045 (target: `pindex_init_sst_open_us` p99 9.3 s → ≤ 0.3 s).
- p50 unique/init > 0.70 → confirmed pivot: bump `pk_index_parallel_compaction_task_split_threshold_bytes` 32 MB → 1 GB (fewer-but-larger SSTs; mind the iter-015 prefetch trap), or memory-only PK-index footer/filter blocks (skip-disk-cache for hot small blocks given vdb-bw is at ESSD ceiling).
- in between → consider both, weighted by mean.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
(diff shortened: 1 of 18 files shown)


luohaha:

PR #72619, opened 2026-05-04
[auto-perf-opt] Parallelise PK-index multi_get into key chunks on a dedicated pool

luohaha:
## Summary

Cold-start PK-index slow publishes after PR #72606 are dominated by **inner-level
sequential cache-miss block reads inside a single `Table::MultiGet`** call:

| metric (iter-033, TSP build 1422) | value |
|---|---|
| slow ≥ 1 s BE Published events (cold-start window) | 815 |
| `multi_get ≥ 50 %` cost share of slow events | 65.4 % |
| `multiget_t3_block_read_miss / multiget_t3` (p50) | 99.5 % |
| per-Table::MultiGet `block_read_miss_cnt` (p99) | 153 |
| per-miss latency (p50) — OSS first-byte | ~99 ms |
| per-fileset `multi_get_us` (p99) | ~3,207 ms |
| upsert max latency vs 3 s goal | **33,284 ms (11.1×)** |

`Table::MultiGet` walks keys in one sequential loop, calling `BlockReader` per
cache miss. Each miss issues one synchronous OSS GET (~99 ms first-byte, ~5 KB
payload). With 153 misses at p99 the upper bound is ~15 s; continuation logic
amortizes some, leaving the per-fileset wall time at ~3.2 s.

The path is **concurrency-bound on OSS service latency** (NOT bandwidth, NOT
RTT, NOT CPU), so concurrency directly translates to wall time without
amplifying bytes.

## Change

Split each `PersistentIndexSstable::multi_get` into N contiguous balanced
chunks of `key_indexes`. Dispatch each chunk's `Table::MultiGet` on a
**dedicated `cloud_native_pk_index_chunk_io` ThreadPool** so the per-fileset
sequential cache-miss loop fans out across worker threads.

Each chunk task gets its own `RandomAccessFile` — the underlying
`SeekableInputStream` is not safe for concurrent use. The AWS S3Client is
shared (and thread-safe), so opening N file wrappers is cheap.

The new pool is **a third dedicated pool**, distinct from
`pk_index_inner_io_thread_pool` (which dispatches per-fileset `multi_get`s in
`LakePersistentIndex::get_from_sstables` after PR #72605) and from
`pk_index_execution_thread_pool` (parallel publish). Re-entering `inner_io`
from one of its workers would trigger
`ThreadPool::check_not_pool_thread_unlocked()` and abort — same trap iter-031
(PR #72579) fell into for `pk_index_execution_thread_pool`. The new pool
sidesteps it structurally.

[...]

PR #72605, opened 2026-05-04
[auto-perf-opt] Move parallel get_from_sstables to dedicated inner-io pool

luohaha:
## Why

`LakePersistentIndex::get_from_sstables` walks `_sstable_filesets` `rbegin -> rend`, and each fileset's `multi_get` is dominated by sequential cache-miss block reads from OSS (~99 ms p50 first-byte service latency). On a fresh-deploy cold-start, this serial walk × ~5–6 filesets pushes per-tablet upsert wall time to 5–25 s and per-publish-RPC max to 32 s, well above the 3 s primary goal of the auto-perf-opt run.

PR #72579 attempted to parallelise this fan-out using `pk_index_execution_thread_pool` and crashed all 6 BEs ~30 s into the iter-031 benchmark with `ThreadPool::check_not_pool_thread_unlocked`: on the parallel-publish hot path, `LakePersistentIndex::upsert`'s lambda is dispatched on `pk_index_execution_thread_pool` (`ParallelPublishContext::token`, see `lake_primary_index.cpp:424,605` and `update_manager.cpp:931,1069`); calling `new_token()->wait()` on the SAME pool from one of its workers triggers the assert.

PR #72591 added a defensive `Thread::current_thread()->name() == "cloud_native_pk_index_execution"` check that fell back to sequential when on the same pool. iter-032 fresh-deploy benchmark on PR #72591 (TSP build 1420) confirmed:
- ✅ no crash (err 0 %)
- ❌ cold-start storm unchanged: per-tablet upsert max 32 885 ms; multi_get-dominated 95 % of slow ≥ 1 s events; `block_read_miss / multi_get` = 99.5 % (p50). The parallel optimisation is now gated OFF on the cold-start hot path.

## What

Add a dedicated `pk_index_inner_io_thread_pool` (BE thread pool name `cloud_native_pk_index_inner_io`) and dispatch the per-fileset `multi_get` tasks on it. Because this pool is separate from `pk_index_execution_thread_pool`, a `pk_index_execution_thread_pool` worker can submit + wait on the inner pool without re-entrance. The defensive `Thread::current_thread()` check from PR #72591 is removed; `#include "util/thread.h"` is dropped.

Pool sizing:
- `pk_index_inner_io_threadpool_max_threads` (default `0` = `CpuInfo::num_cores()` — each task is OSS-IO-bound, threads can sleep without burning CPU)
[...]

PR #72674, opened 2026-05-04
[auto-perf-opt] Skip local disk cache for PK-index sstable open reads

luohaha:
## Bottleneck

Realtime-benchmark cold-start storm (`tsp deploy apply` of the same build wipes the in-memory `_index_cache` and forces ~17 PK-index SSTs/tablet to be reopened across ~1000 tablets per CN) saturates the local SSD bandwidth on every CN. Per iter-040 / iter-042 / iter-044 profiling on the `999_1gb_1_1tb` template:

- vdb peak bw 593-656 MB/s on every CN ≈ ESSD bandwidth ceiling.
- vdb IOPS 9.9-11.5 K, well below the 25-50 K ESSD ceiling — the bottleneck is bandwidth, not IOPS.
- `pindex_init_sst_open_us` p99 stays at 7.2-9.3 s across iter-027..044 even after the parallelism / pool-sizing wins (PR #72605 / #72619 / #72622 / #72639); per-task wall-clock is OSS-RTT-bound and pool sizing can't reduce it further.
- iter-044's diagnostic counter (PR #72672) confirmed that lazy-SST-open is **contraindicated**: each cold-start `get_from_sstables` walks ~all opened sstables, so deferring opens would not save work. The remaining lever is the data-movement path itself.

## Change

`PersistentIndexSstable::init` previously opened each sstable through the long-lived `RandomAccessFile`, which fills the local disk cache with the footer, index block, metaindex block, and bloom-filter block. These small blocks are kept resident inside `Table::Rep` for the lifetime of the sstable, so caching them on disk only consumes vdb bandwidth without serving any later read. The 4 small init reads (~150 KB total per SST) × ~17 SSTs × ~1000 tablets across 6 CN account for a meaningful share of cold-start vdb writes.

This PR opens the table through a temporary `RandomAccessFile` created with `skip_fill_local_cache=true` and `skip_disk_cache=true`, then swaps the long-lived `_rf` into `Table::Rep::file` via a new `Table::set_file()` hook. Runtime data-block reads (compaction iterators, and the per-call file built in `PersistentIndexSstable::multi_get` under `enable_pk_index_parallel_execution=true`) keep using the long-lived `_rf` and still populate the disk cache for repeated queries.

## Expected Impact

Pre-registered acceptance criteria (per iter-044's recommendation):

- Cold-start vdb peak bw drops ≥ 30 % (656 → ≤ 460 MB/s).
- `pindex_init_sst_open_us` p99 drops ≥ 30 % (8.0 → ≤ 5.6 s).
- Upsert max ≤ 12 s.
- Steady-state envelope intact (P99 ≤ 800 ms, throughput ≥ 800 K).

Hard guardrails: err = 0 %, no `be.out` aborts, no compaction throughput regression.

## Files

- `be/src/storage/sstable/table.h` — declare `Table::set_file()`.
- `be/src/storage/sstable/table.cpp` — define `Table::set_file()`.
- `be/src/storage/lake/persistent_index_sstable.cpp` — open via skip-disk-cache `init_rf`, swap to runtime `_rf` via `set_file()`. Corruption retry path symmetrically uses `init_rf` for the retry open.

3 files, +42/-4 LOC.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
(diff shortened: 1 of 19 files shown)


luohaha:
Not promoted to upstream/main per project decision (validated locally but not selected for upstream submission). Closing this branch-4.1 draft.

PR #72676, opened 2026-05-04
[auto-perf-opt] Bump pk_index_parallel_compaction_task_split_threshold_bytes 32MB -> 64MB

luohaha:
## Summary

Bump `pk_index_parallel_compaction_task_split_threshold_bytes` default from 32 MB → 64 MB so each PK-index compaction emits fewer-but-larger SSTs, reducing the count of SST opens at fresh-deploy cold-start.

Stacked on **PR #72674** (validated in iter-046 of the auto-perf-opt run): footer / index / metaindex / filter blocks bypass the on-disk block cache, so the historical iter-015 prefetch trap (per-SST filter-block bandwidth blow-up with larger SSTs) is mitigated.

## Cold-start evidence (iter-046 fresh-deploy, build 1433 = #72674 + cumulative stack)

| Metric                              | Value | Note                                         |
|-------------------------------------|-------|----------------------------------------------|
| `pindex_init_sst_cnt` p50            | 26    | per cold-start tablet                       |
| `pindex_init_sst_cnt` max            | 53    |                                              |
| `pindex_init_sst_open_us` p99        | 4.26s | already cut from 8.0s by #72674              |
| Cold-start vdb peak bw (per CN)      | 545 MB/s mean | down 13% vs iter-045's 627 MB/s; still BW-pressured |
| Upsert max (fresh-deploy)            | 12.97s | target ≤ 3s                                  |

Each SST init issues 4 sequential OSS RTs for footer / index / metaindex / filter blocks. The dedicated `pk_index_sst_open` pool (#72639, 128 threads/CN) parallelises across SSTs but cannot reduce wall-clock per cold-start tablet — that is bounded by the slowest open in the parallel batch. Cutting the SST count per cold-start tablet is therefore the structural lever.

## Why 32 → 64 (single-step), not 32 → 1024

- iter-015 (Table::Open tail-prefetch) regressed tput −13% / P99 +78% because filter blocks ~100 KB were saturating local-disk bandwidth. With #72674 those filter blocks no longer hit vdb at all.
- A single-step bump keeps per-SST filter-block footprint contained; iteration 047 measures the actual impact and we can bump further (64 → 128 → 256) if the steady-state envelope holds.

## Expected impact (iter-047 will measure)

- `pindex_init_sst_cnt` p50 26 → ~13 per cold tablet
- `pindex_init_sst_open_us` p99 4.26s → ~2.5s (parallelism unchanged, fewer ssts/wait)
- Upsert max ≤ 8s in fresh-deploy
- Steady envelope intact (P99 ≤ 800ms, tput ≥ 800K rows/s, err 0%)
- No compaction-throughput regression

## Test plan

- [x] Existing tests in `be/test/storage/lake/lake_persistent_index_parallel_compact_mgr_test.cpp` parameterise this config — they save/restore via `auto old_threshold = ...; ... = old_threshold;` and so are unaffected by the default bump.
- [ ] iter-047 deploys this PR via TSP and runs the 999_1gb_1_1tb fresh-deploy 20-min benchmark, captures cold-start publish traces and Prometheus vdb metrics, and reports the actual impact on the metrics above.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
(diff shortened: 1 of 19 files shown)


luohaha:
Closing — iter-047 measured a regression vs iter-046 baseline: upsert P99 +30 % (694 → 905 ms), upsert max +16 % (12.97 → 14.99 s). Single-step bump of split threshold 32 MB → 64 MB shifted compaction work in a way that outweighed the SST-count reduction; per-SST filter-block growth ate the gain. Cleanup of auto-perf-opt my-rt-2 run.

PR #72606, opened 2026-05-04
[auto-perf-opt] iter-033 stack-build: PR #72605 + missing metric def fix

luohaha:
Iter-033 stack-build slot for the runner agent.

Stack contents (all on top of pinned base 7af3ba7):
- BRPC server-side queue-time logging (#72555)
- Slow lake publishPartition / Utils.publishVersionBatch breakdown logging
- LakePersistentIndex::load_dels parallelisation (#72574)
- Table::MultiGet t3 sub-phase counters (#72577)
- get_from_sstables parallel-by-fileset (#72579 origin) + iter-031 defensive guard (#72591) + iter-032 dedicated `cloud_native_pk_index_inner_io` thread pool (#72605)
- **NEW** in iter-033: `METRICS_DEFINE_THREAD_POOL(cloud_native_pk_index_inner_io)` in `starrocks_metrics.h` — fixes the build failure in TSP build 1421 (PR #72605 referenced these metric members but never declared them).

Purpose: build slot only — do not merge. The optimisation lives in #72605 (whose head branch has been updated with the same fix commit so #72605 itself now builds).

PR #72703, opened 2026-05-05
[auto-perf-opt] Add multiget miss-run + byte-contig pairs trace counters

luohaha:
## What

Adds two diagnostic trace counters to `Table::MultiGet` in
`be/src/storage/sstable/table.cpp` so the publish trace exposes the
distribution of contiguous cache-miss runs:

- `multiget_miss_run_cnt` — number of maximal runs of back-to-back
  cache-miss `BlockReader` calls within a single `MultiGet` invocation.
  A run ends on a cache hit; `filter_skip` and `continue_block_read_cnt`
  paths do not split a run because no read is attempted.
- `multiget_byte_contig_miss_pairs_cnt` — count of consecutive (i, i+1)
  miss pairs whose i+1's `BlockHandle.offset` equals i's
  `BlockHandle.offset + size + kBlockTrailerSize`, i.e. directly
  byte-contiguous on disk.

## Why (auto-perf-opt iter-048)

Cumulative-stack diagnosis after iter-047 / build 1434 shows
`multiget_t3_block_read_miss_us` p99 = 11.7 s with miss_cnt p99 = 297 —
the dominant single contributor on the cold-start critical path,
~2.3× larger than `pindex_init_sst_open_us` p99 = 5.1 s. Cold publish
queries fan out across many filesets and each cache miss is one
~5 KB OSS round-trip (~29 ms first-byte), serialised in the per-key
loop. The natural fix is to coalesce contiguous misses into a single
`read_at`, but the right horizon depends on (a) how many consecutive
keys actually miss in a row (run length) and (b) what fraction of
consecutive-miss pairs are truly byte-adjacent on disk (so a single
range read serves them without pulling in unnecessary bytes — the
iter-015 prefetch-trap that turned a similar refactor into a 78 %
P99 regression).

Two counters let analysis derive both:

- mean miss-run length = `multiget_block_read_miss_cnt /
  multiget_miss_run_cnt`
- byte-contig pair fraction = `multiget_byte_contig_miss_pairs_cnt /
  max(0, multiget_block_read_miss_cnt - multiget_miss_run_cnt)`

If mean run length ≥ ~3 and byte-contig fraction ≥ ~0.7, the
coalescing patch wins big and the horizon should be ≈ mean run length.
If either is low, the actual structural fix is different (e.g. larger
default `block_size`).

## Cost / risk

Pure instrumentation, no behavior change. Per-key cost on the
cache-miss path adds:

- one `BlockHandle::DecodeFrom` call only when `filter == nullptr`
  (existing code already decodes `handle` inside the filter branch
  when `filter != nullptr` — we re-use it).
- two int64 comparisons + two int64 increments.

Cache-hit and `filter_skip` paths add at most a single int64 compare
(`stat != nullptr`) and one int64 reset. No allocation. No new RPCs.

## Validation plan

- iter-049 will deploy this build, re-run the same fresh-deploy
  benchmark, and read the new counters from publish traces. Decision
  follows directly:
  - mean run length ≥ 3 ∧ contig fraction ≥ 0.7 → ship the
    coalescing patch in iter-050 with horizon ≈ mean run length.
  - mean run length < 2 → coalescing won't help; pivot to alternative
    cold-start lever (data-block size, prefetcher).
  - byte-contig fraction < 0.5 with high run length → keys are hitting
    non-adjacent blocks; need a smarter coalescer (sort-then-merge)
    rather than a naive run-of-misses batch.

## Stack

Stacked on validated iter-046 cumulative: #72674 + #72672 + #72671 +
#72667 + #72654 + #72639 + #72622 + #72619 + #72606 + #72591 + #72676
+ post-iter-021. Branch `apo/my-rt-2/iter-048` rebased onto
`apo/my-rt-2/iter-046` tip.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
+// Minimum number of keys in one PersistentIndexSstable::multi_get for the chunk-parallel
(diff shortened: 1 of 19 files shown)


luohaha:

PR #72476, opened 2026-05-02
[auto-perf-opt] [ABANDONED] Prefetch SSTable tail in Table::Open — invalidated by iter-015 data

luohaha:
## Status: Abandoned (iter-015)

The prefetch hypothesis — that `Table::Open` cold-start cost is dominated by 4 sequential OSS RTTs — is **invalidated** by iter-015 measurements. Per-SST `Table::Open` p99 is statistically identical with or without the prefetch (348 ms v4 vs 351 ms baseline), and the larger 1 MB tail read in v4 introduced a measurable steady-state regression (throughput -13 %, upsert P99 +78 % vs iter-013 v3). The PR is now reset to the validated PR #72460 + #72434 stack (commit `ab2361c5`) and retained only as a historical record of the experiment. Closing or merging is at the maintainer's discretion — those wins live as their own PRs (#72460, #72434).

## What was tried

| Variant | Code | Per-SST `Table::Open` p99 | Steady-state |
|--|--|--|--|
| baseline (no prefetch, 4 reads) | branch-4.1 | 351 ms (iter-012) | 925 K rows/s, upsert P99 778 ms |
| v3 — 64 KB tail (`817dcec`) | small tail captures footer only | 352 ms (iter-013) | 907 K rows/s, upsert P99 647 ms |
| v4 — 1 MB tail + extension + filter stage-2 (`f6c9f8d3`) | `Table::Open` issues 2-3 reads instead of 4 | **348 ms (iter-015)** | **791 K rows/s (-13 %), upsert P99 1149 ms (+78 %)** |

## Why it didn't work

Per-SST `Table::Open` cost is **bandwidth-bound**, not RTT-bound, for these PK-index SSTs.

- 30 MB SSTs with `block_size = 4 KB` and `NewBloomFilterPolicy(10)` produce filter blocks of ~2 MB each.
- A read for the filter block must transfer those bytes regardless of how few RTTs we issue.
- v4's 1 MB stage-1 tail captures footer + metaindex + index in one RTT (the cost saved), but then a second RTT must fetch the multi-MB filter, so per-SST cost stays at "1 RTT + filter download" ≈ same as baseline's "4 RTTs of small reads + filter download".
- Worse, the 1 MB stage-1 read pollutes the page cache (and consumes OSS bandwidth) for data that's almost never used (last MB of the file is mostly unrelated data blocks). At 6 CN × hundreds of concurrent SST opens, this measurably starves user traffic.

## Direction for future work

[...]

PR #72710, opened 2026-05-05
[auto-perf-opt] Skip disk cache fill for PK-index chunked fan-out reads

luohaha:
## Summary

Extends PR #72674's skip-disk-cache mechanism (which targeted
`PersistentIndexSstable::init` for footer / filter / metaindex
blocks) to the **runtime data-block reads** in
`PersistentIndexSstable::multi_get`'s chunked-parallel path.

1 file, +19/-0 LOC (just adds `opts2.skip_fill_local_cache = true` plus a comment).

## Cold-start bottleneck identified (iter-050 trace, build 1435)

Iter-050's redeploy of `branch-4.1-d3b8e6f` (PR #72703 instrumentation
in cumulative stack) reproduced cold-start storm. 2792 cold-start
records in the first 5 min (slow ≥ 1 s, all 6 BE):

| Counter | p50 | p99 | p99.9 |
|---------|-----|-----|-------|
| `multiget_t3_block_read_miss_us` (μs) | 1,722,228 | **8,786,165** | 19,799,915 |
| `multiget_block_read_miss_cnt` | 17 | 296 | 895 |
| `pindex_init_sst_open_us` (μs) | 1,940,399 | 5,416,072 | 37,907,831 |
| `multi_get_us` (μs) | 1,608,712 | 6,865,360 | 12,143,494 |

`multiget_t3_block_read_miss_us` p99 ≈ **8.79 s** — the dominant
single contributor on the cold-start critical path, ~1.6× larger
than `pindex_init_sst_open_us` p99 5.42 s. ~297 cache misses per
cold publish event × ~5 KB blocks × ~29 ms OSS first-byte
≈ **9 s of serial RTT per cold publish event**.

## Why coalescing is not the right lever (validated this iter)

PR #72703's diagnostic counters tell us:

- AGG mean miss-run length = **1.434** (< 2 threshold from iter-049 protocol)
- AGG byte-contig pair fraction = **0.275** (< 0.5 threshold)
- per-record mean run length p99 = 3.89

A `Table::MultiGet` coalescer would deliver at most a **30 % OSS-RT
reduction** (103,777 misses → 72,351 runs), well below the 50 % bar
that justifies a hot-path refactor. With contig fraction 0.275, naive
merge would over-fetch and risk re-triggering iter-015 / PR #72476's
bandwidth-blowup trap (-13 % tput, +78 % P99).

iter-049 saw the same band (1.518 / 0.313). Two iterations agree.

## Mechanism

`PersistentIndexSstable::multi_get` chunks large multi_get requests
across `pk_index_chunk_io` workers, with each chunk creating its own
`RandomAccessFile` (line 233) — currently with default
`RandomAccessFileOptions{}`, so every cache-miss block read writes
the fetched bytes into the local on-disk cache.

vdb (local SSD block cache) is **bandwidth-bound** at cold-start:
peak 562-667 MB/s ≈ ESSD ceiling band; IOPS 9-12 K ≪ 25-50 K ceiling.
Each cache-fill write contends with the BE process for the same
disk bandwidth that the cold-start fan-out OSS reads ultimately
depend on (write completions block read returns through the
fs_starlet codepath).

This PR sets `opts2.skip_fill_local_cache = true` on the per-chunk
files. Effects:

- **Cold-start**: chunked fan-out OSS reads are still served, but
  their result is NOT written into the on-disk cache. vdb bandwidth
  is freed up for other writers.
- **Steady-state**: the chunked path is gated by
  `key_indexes.size() >= pk_index_parallel_chunk_min_keys=32` — the
  cold-start fan-out signature. Single-key upserts and small batches
  (steady-state) flow through the sequential / non-chunked fallback
  on line ~293 with the long-lived `_rf` (default cache-fill enabled).
- **Memory cache untouched**: `sstable::Options::block_cache` is a
  separate first-tier inside `Table::BlockReader` (line 215-237 of
  `be/src/storage/sstable/table.cpp`). Hot-key re-reads inside the
  same multi_get / subsequent multi_gets still hit RAM.
- **Existing cached blocks**: `skip_fill_local_cache` only blocks
  WRITES; existing cached blocks are still served from disk.

## Pre-registered acceptance for next iteration (iter-051)

- **Primary**: cold-start `multiget_t3_block_read_miss_us` p99
  8.79 s → ≤ **5 s** (≥ 40 % drop).
- **Secondary**: cold-start vdb peak BW 660 MB/s → ≤ **460 MB/s**
  (≥ 30 % drop) via PromQL.
- **Headline**: upsert max ≤ **8 s** in fresh-deploy.
- **Steady envelope intact**: per-batch P99 ≤ 800 ms, tput ≥ 800 K.
- **Anti-regression**: `multiget_block_read_miss_cnt` p99 must not
  balloon > ~600 (≈ 2× iter-050 baseline). If it does, hot blocks
  are being re-fetched from OSS because memory cache cannot hold the
  cold-start working set — revert.
- **Guardrails**: err 0 %, no `be.out` aborts, no compaction-throughput
  regression, no vdb-bw peak > 700 MB/s sustained.

Stacked on iter-046 / iter-048 cumulative validated stack
(#72674 + #72672 + #72667 + #72654 + #72639 + #72622 + #72619 + #72606 +
#72591 + #72703 + #72676 + post-iter-021).

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
+// Minimum number of keys in one PersistentIndexSstable::multi_get for the chunk-parallel
+// path to engage. Below this, run the legacy sequential path so we don't pay open-file
+// + dispatch overhead for tiny lookups.
+CONF_mInt32(pk_index_parallel_chunk_min_keys, "32");
+// Target keys per chunk. Effective chunk count = clamp(ceil(N/target), 1, max_chunks).
+CONF_mInt32(pk_index_parallel_chunk_target_keys, "64");
+// Hard upper bound on chunks per multi_get to protect the chunk_io pool's queue.
+CONF_mInt32(pk_index_parallel_chunk_max_chunks, "16");
 // The maximum number of memtables for pk index in shared-data mode.
 CONF_mInt32(pk_index_memtable_max_count, "2");
 // The maximum wait flush timeout for pk index memtable in shared-data mode, in milliseconds.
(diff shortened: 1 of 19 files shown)


luohaha:
Closing — empirically falsified by iter-051 measurements. Skip-disk-cache on chunked PK-index fan-out **regressed** `multiget_t3_block_read_miss_us` p99 8.79 → 15.27 s (+74 %, vs target -43 %); vdb peak BW barely moved (-1 to -2 %, vs target -27 %). Hypothesis that disk-cache writes were the dominant per-miss cost was wrong — the cost is OSS first-byte service latency, which this PR does not address. Cleanup of auto-perf-opt my-rt-2 run.

PR #72724, opened 2026-05-05
[auto-perf-opt] PK-index multi_get chunk parallelism: target_keys 64->16, max_chunks 16->64

luohaha:
## Bottleneck

Cold-start storm post-iter-051 (#72710 abandoned) is dominated by `multiget_t3_block_read_miss_us` p99 ≈ 11.9 s per cold publish event, with `multiget_block_read_miss_cnt` p99 ≈ 296-359 across 6-22 multi_gets. Per-miss wall ≈ 40 ms (OSS first-byte for ~5 KB PK-index data blocks), so the bottleneck is the **sequential block fetches inside each chunk's `Table::MultiGet`**.

The existing chunked-parallel branch in `PersistentIndexSstable::multi_get` (introduced in #72619) caps chunks at 16 and targets 64 keys/chunk. For big cold-start multi_gets (hundreds of keys) this means 16 chunks of ~50-60 keys each — within a chunk, ~18 sequential OSS misses × 40 ms ≈ 720 ms per-chunk wall. That sets the per-multi_get wall.

## Change

Lower default `pk_index_parallel_chunk_target_keys` 64 → 16 AND raise `pk_index_parallel_chunk_max_chunks` 16 → 64 so a 1024-key multi_get gets up to 64 chunks of 16 keys each — within each chunk now ~5 sequential misses × 40 ms ≈ 200 ms per-chunk wall, ~3.5× faster.

Pool ceiling: `pk_index_chunk_io_threadpool_max_threads = num_cores * 4 = 128` on a 32-core CN. With several concurrent multi_gets (≤ 5 filesets × 64 chunks = 320 work items), 192 chunks queue, so effective per-multi_get parallelism ≈ 128/5 = ~25 chunks. Even so, per-multi_get wall drops from ~120 ms to ~78 ms (50 misses / 25.6 active threads × 40 ms). Across 6-22 multi_gets per cold publish event the cumulative `multiget_t3_block_read_miss_us` p99 should drop ~35 % (11.9 s → ~7.7 s).

1 file, +16/-2 LOC: `be/src/common/config.h`.

## Why not the iter-051 alternatives

- **Lever B (bump `lake_pk_index_block_cache_limit_percent` 10 → 30)**: cache is 3.24 GB per BE while estimated cold-start working set per BE is ~200 MB. NOT capacity-bound; the 30 % cold-start hit ratio is a cold-cache filling effect, not LRU eviction.
- **Lever A as iter-051 framed it (lower `pk_index_parallel_chunk_min_keys` 32 → 16)**: that knob is the THRESHOLD for entering the chunked branch, not the chunk-count knob; cold-start multi_gets are already well above the threshold, so lowering it doesn't change parallelism for the affected workload.
- **Lever C (within-chunk speculative prefetch)**: requires walking the SST index upfront for all keys before issuing block fetches — significantly more code than the iter-052 budget. The lever-A reframing here gives a similar effect (more chunks ⇒ shorter per-chunk serial wall) at a config-only cost.

## Pre-registered iter-053 acceptance

Primary (≥ 1 of 2 must hold):
- `multiget_t3_block_read_miss_us` p99 11.9 s → ≤ 8 s (-33 %).
- `multi_get_chunk_count` p99 distribution shifts up (≥ 32 chunks per multi_get in p50 vs current ≤ 16) — instrumentation already in place via `TRACE_COUNTER_INCREMENT("multi_get_chunk_count", chunks)`.

Secondary:
- Cold-start upsert max ≤ 8 s (within fresh-deploy 11-42 s noise band).

Steady envelope:
- tput ≥ 800 K, P99 ≤ 800 ms, err 0 %, no `be.out` aborts.

Anti-regression:
- `pk_index_chunk_io_pending_time_ns_total` rate ≤ 3× iter-048 cumulative baseline (queue starvation guard).
- `pindex_init_sst_open_us` p99 unchanged (this lever doesn't touch init).

## Stacking

Stacked on iter-048 cumulative tip `d3b8e6f` (#72703 + #72676 + iter-046 cumulative). #72710 is **abandoned** per iter-051 and is NOT in this stack.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
+// Minimum number of keys in one PersistentIndexSstable::multi_get for the chunk-parallel
+// path to engage. Below this, run the legacy sequential path so we don't pay open-file
+// + dispatch overhead for tiny lookups.
+CONF_mInt32(pk_index_parallel_chunk_min_keys, "32");
+// Target keys per chunk. Effective chunk count = clamp(ceil(N/target), 1, max_chunks).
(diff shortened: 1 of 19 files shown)


luohaha:
Closing — empirically falsified by iter-053 measurements. Bumping per-multi_get `max_chunks 16 → 64` produced `multi_get_chunk_count` p99 = 1,398 (~4× lift) but effective intra-multi_get parallelism only 1.76×; `chunk_io` pool active peak 26-88 / 128 (NOT saturated). Wall-clock ceiling lives elsewhere (per-sstable serial walk in `Fileset::multi_get`, addressed by #72735). Bumping the chunk knob alone has no effect on cold-start max.

PR #72735, opened 2026-05-05
[auto-perf-opt] Parallelize per-sstable walk inside Fileset::multi_get on a dedicated pool

luohaha:
## Bottleneck (from iter-053)

`multi_get_us` p99 = 7.72 s on cold-start, but `multi_get_chunk_count` p99 = 1,398
(iter-053 PR #72724 raised the per-multi_get max-chunks ceiling 16 → 64). Effective
intra-multi_get parallelism only **1.76×** despite chunk_count ~4× lift. Trace evidence
showed `cloud_native_pk_index_chunk_io` pool active peak only **26-88 / 128** —
**NOT saturated**.

The wall-clock ceiling lives in `PersistentIndexSstableFileset::multi_get`
(`be/src/storage/lake/persistent_index_sstable_fileset.cpp:152-157`):

```cpp
for (const auto& [sstable, sstable_key_indexes] : sstable_key_indexes_map) {
    if (touched_sstables != nullptr) { touched_sstables->insert(sstable); }
    RETURN_IF_ERROR(sstable->multi_get(keys, sstable_key_indexes, version, values, found_key_indexes));
}
```

`SST::multi_get` already chunks onto `chunk_io` and calls `chunk_io_token->wait()`,
so at any instant only ONE sstable's chunks are in flight. With ~22 sstables touched
per multi_get on cold-start (iter-048 trace `pindex_init_sst_cnt` p50), wall stacks to
~5-7 s.

## Change

Behind `enable_pk_index_parallel_sst_walk` (default true), dispatch each per-sstable
`multi_get` on a NEW dedicated `cloud_native_pk_index_sst_walk` ThreadPool (default
`num_cores * 4`). After `token->wait()`, merge per-task `found_key_indexes` sets;
`values[key_index]` writes are disjoint by construction (sstables in a fileset have
non-overlapping ranges, enforced at `init`/`append`).

After: chunks from multiple sstables in flight on `chunk_io` simultaneously. With 22
sstables × ~16 chunks/sst = ~352 chunks queued at once, chunk_io 128-thread ceiling
becomes the limit and saturates. Estimated wall: 352 / 128 × ~250 ms = **~0.7 s** vs
current ~5-7 s.

## Pool isolation (iter-031 deadlock trap)

Each layer waits on a STRICTLY DEEPER pool's token. No re-entrance.

```
parallel_publish (pk_index_execution)
   ↓ submits per-tablet upsert
(execution worker) → LakePersistentIndex::upsert
   ↓ inline call
[...]

PR #72762, opened 2026-05-05
[auto-perf-opt] Lower pk_index_chunk_hedge_after_ms default 1500 -> 500 ms

luohaha:
## Summary

Lower the default `pk_index_chunk_hedge_after_ms` from **1500 ms → 500 ms** so the chunk-hedge mechanism (introduced in #72758) actually engages on the chunk-level OSS slowdowns it was designed for.

Stacked on iter-048 cumulative tip + #72735 (parallel SST walk, in-flight) + #72758 (chunk-hedge mechanism, in-flight). 1-line config-default change, no code-path change.

## Motivation — iter-057 measurement

`#72758` shipped with `hedge_after_ms = 1500`. iter-057 fresh-deploy measurement (run_20260502_103938 / iterations / 057):

- Hedge fired on **only 0.11 %** of all cold-start publishes.
- Hedge fired on **only 3.4 %** of slow publishes (`cost_ms ≥ 1 s`); the other 96 % of slow tails were never hedged.
- Reason: per-chunk wall is rarely > 1.5 s. The slow-publish tail is **aggregated wall** across ~22 parallel sstables × ~7 chunks each, so the *outer* `multi_get` p99 is large (15-22 s in the 1 % tail) but no individual *chunk task* exceeds 1.5 s.

At 500 ms, expected dispatch rate is roughly 5-10 × higher and should cover a much broader tail. The hedge remains capped per multi_get at 16 (`pk_index_chunk_hedge_max_per_multi_get`), so worst-case extra OSS RPS per cold-start publish stays bounded.

## Change

```c
-CONF_mInt32(pk_index_chunk_hedge_after_ms, "1500");
+CONF_mInt32(pk_index_chunk_hedge_after_ms, "500");
```

`CONF_m*` ⇒ runtime mutable; can be tuned without restart for falsification.

## Acceptance criteria for iter-059 (≥ 2 of 3, fresh-deploy A/B vs iter-057)

1. cold-start upsert max ≤ **8 s** (from 15.68 s; STRICT — primary bar)
2. cold-start `multi_get_hedged_chunks` summed across cold-start window ≥ **200 on ≥ 1 BE** (10× iter-057's BE-130 = 20)
3. cold-start `multi_get_us` p99 ≤ **3.5 s** (from 2.55 s, modest tightening)

## Hard guardrails

- `err_rate = 0 %`, no `be.out` aborts
- vdb peak BW ≤ 700 MB/s sustained (currently 599-670 MB/s on iter-055; chunk_io pool peak 30-92/128, hedge cap 16/multi_get keeps the slack contained)
- Steady envelope: tput ≥ 800 K, per-batch P99 ≤ 800 ms (post-i021 envelope: 597-987 ms)
- chunk_io active peak ≤ 110 / 128
- `pindex_init_sst_open_us` p99 unchanged

## Falsification

- Criterion 3 PASS but criterion 1 FAIL ⇒ tail is OSS-side correlated across servers; move to phased fan-out / OSS connection-isolation.
- Criterion 2 FAIL with default 500 ms ⇒ chunk wall is consistently <500 ms in slow tails; aggregated SST wall is the binder, not chunk wall — back to "fix data path".
- Hard guardrail breach ⇒ revert to 1500 ms (or `enable_pk_index_chunk_hedge=false`).

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
+// Minimum number of keys in one PersistentIndexSstable::multi_get for the chunk-parallel
+// path to engage. Below this, run the legacy sequential path so we don't pay open-file
+// + dispatch overhead for tiny lookups.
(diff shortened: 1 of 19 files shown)


luohaha:
Closing — iter-059 measurement showed hard guardrail breach. Lowering hedge_after_ms 1500 → 500 ms succeeded at engaging the hedge mechanism (rate × 16, slow-rate 8% → 85%) and dropped `multi_get_us` p99 to 1.82s, but: (a) per-batch P99 went 658 → 937 ms (+42%, vs ≤ 800 ms guardrail); (b) chunk_io pool saturated at 128/128 + queue 604 on the storming BE (vs ≤ 110 / 128 guardrail). The 500 ms threshold is too aggressive; hedging fires faster than chunk_io can absorb. The mechanism (#72758) stays valid; just the default needs to be in the 750-1500 ms range, which #72758 already covers (default 1500 ms).

PR #72780, opened 2026-05-05
[auto-perf-opt] Bump pk_index_sst_open pool default to num_cores * 8

luohaha:
## Bottleneck

Iter-061/062 measured the cold-start `cloud_native_pk_index_sst_open` thread pool saturating at **128 / 128 on 5 of 6 BEs**, with queue depth up to **875** on the worst BE during the cold-start storm under the realtime-benchmark `999_1gb_1_1tb` template.

The pool gates SST footer / index / metaindex / filter loads in `Table::Open`. Each task is **4 sequential OSS round-trips** of pure I/O wait, no CPU work and no local disk traffic (the path bypasses the disk cache since #72674).

Since #72674 closed the per-byte vdb pressure on this path, the only remaining lever is **more concurrent OSS HTTP**. Throughput is determined by `pool_size × OSS_RPS_per_thread` — currently capped at 128 threads.

## Path A (lazy SST open) was ruled out in iter-062

Diagnostic counter `pindex_unique_sstables_touched_cnt` (added in cum-tip from iter-043) ran across 834 tablets:

- ratio `touched_max / init_max`: p50 **2.88**, p99 **6.95**, max **11.50**
- fraction ≤ 0.5: **0.000**
- fraction ≤ 0.8: **0.000**
- fraction ≥ 1.0: **0.971**

Mechanism: keys distribute uniformly across each fileset's `_sstable_map.upper_bound` divide, so almost every SST sub-range gets queried in a typical cold-start batch. Plus post-init opvec applies grow the in-memory SST set further. Lazy open is dead under the iter-061 ≥ 80 % bail rule.

## Path B (this change)

Lift the `cloud_native_pk_index_sst_open` fallback ceiling in `be/src/runtime/exec_env.cpp` from `CpuInfo::num_cores() * 4` (= 128 on the 32-core CN) to `CpuInfo::num_cores() * 8` (= 256). 1 int literal + comment refresh. Existing operator override `pk_index_sst_open_threadpool_max_threads` is unchanged.

## Why this is vdb-neutral

#72674 made `Table::Open` reads use `RAFOptions{ skip_disk_cache=true, skip_fill_local_cache=true }`. SST footer / index / metaindex / filter blocks are read directly from OSS into the per-task buffer without going through the local block cache. Cold-start vdb peak BW (625-693 MB/s, 90 % of 700 MB/s ESSD guardrail) is observed on the **chunk_io** block-cache-write path, not sst_open.

Doubling the sst_open pool ceiling adds OSS HTTP concurrency only — the local disk path is untouched. Steady-state behaviour is unchanged because the pool is empty when no PK-index init is in flight.

## Acceptance for iter-063 (>= 2 of 3, fresh-deploy A/B vs iter-058)

1. cold-start `pindex_init_sst_open_us` p99 ≤ **1.8 s** (from 3.87 s) — PRIMARY for Path B.
2. cold-start `multi_get_us` p99 ≤ **2.0 s** (from 2.33 s) — modest.
3. cold-start upsert max ≤ **8 s** (from 33.26 s).

## Hard guardrails

- err 0 %; no `be.out` aborts
- vdb peak BW ≤ **700 MB/s** sustained
- tput ≥ 800 K
- per-batch P99 ≤ 800 ms (steady-envelope)
- new sst_open pool peak < 256 / 256 (if 256 also saturates, OSS server-side caps below 256 — falsifies Path B)

## Falsification rules

- (1) PASS + (3) PASS ⇒ promote PR.
- (1) PASS + (3) FAIL ⇒ ceiling lifted but cold-start max lives elsewhere; keep merged, pivot iter-064 to Path C / compaction-reduce-N.
- (1) FAIL ⇒ pool wasn't the binder under doubled ceiling; OSS server-side caps below 256 — close PR, pivot iter-064 to Path C (separate S3 connection pool).

## Stack

Stacked on cum-tip `747c9b6` (= iter-048 + #72735 + #72758). +7 / -5 LOC, single file.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
+// Minimum number of keys in one PersistentIndexSstable::multi_get for the chunk-parallel
+// path to engage. Below this, run the legacy sequential path so we don't pay open-file
+// + dispatch overhead for tiny lookups.
+CONF_mInt32(pk_index_parallel_chunk_min_keys, "32");
+// Target keys per chunk. Effective chunk count = clamp(ceil(N/target), 1, max_chunks).
+CONF_mInt32(pk_index_parallel_chunk_target_keys, "64");
+// Hard upper bound on chunks per multi_get to protect the chunk_io pool's queue.
+CONF_mInt32(pk_index_parallel_chunk_max_chunks, "16");
 // The maximum number of memtables for pk index in shared-data mode.
 CONF_mInt32(pk_index_memtable_max_count, "2");
 // The maximum wait flush timeout for pk index memtable in shared-data mode, in milliseconds.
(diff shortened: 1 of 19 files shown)


luohaha:
Closing — acceptance fail. iter-063 measurement: 1/3 PASS.

- (1) cold-start `pindex_init_sst_open_us` p99 ≤ 1.8 s — **FAIL** (got 4.74 s nonzero, REGRESSION 22 % vs iter-062's 3.87 s; max exploded to 60.02 s on BE-127)
- (2) cold-start `multi_get_us` p99 ≤ 2.0 s — PASS (got 1.02 s, best ever)
- (3) cold-start upsert max ≤ 8 s — **FAIL** (got 42.6 s; insert max 62.8 s)

Resource posture confirms the pool was NOT the binder under doubled ceiling:
- BE-127's `sst_open` active = 111 / 256 (NOT saturated), yet its `pindex_init_sst_open_us` max = 60 s. A single OSS round-trip stalled ~60 s with no queue ahead of it.
- 5/6 BEs had pool relief (active 71-165/256, no queue) yet sst_open p99 still regressed.
- BE-130 still saturates at the new 256 ceiling (queue 843), proving the storm sink will saturate any reasonable pool size.

Falsification rule (pre-registered iter-062) fires: OSS server-side latency caps below 256 connections per BE. Lifting pool concurrency exposes correlated OSS server-side stalls. Pivoting iter-064 to Path C (separate S3 client pool for sst_open reads, or per-request OSS timeout + retry).

Positives that survive (iter-063 vs iter-062): throughput 889 K → 920 K, multi_get_us p99 2.33 → 1.02 s, slow-rate 2.29 % → 0.81 %, vdb peak BW 632 → 503-632 MB/s. These don't change the headline because the dominant tail is owned by single-RT OSS stalls, not by pool wait.

Validated stack remains at cum-tip 747c9b6 (= iter-048 + #72735 + #72758).

PR #72783, opened 2026-05-05
[auto-perf-opt] Isolate PK-index SST open S3 client + cap stuck OSS tails

luohaha:
## Why

Cold-start storms on shared-data clusters are bottlenecked by the OSS tail of `pindex_init_sst_open_us`. Investigation across iters 061-063 (workload `999_1gb_1_1tb`, 1 FE + 6 BE, S3-backed):

- iter-061 (no-op verification, current cum-tip): cold-start upsert max **33 s**; `pindex_init_sst_open_us` p99 stuck at **4.7 s**, max **60 s**.
- iter-062 (Path A diagnostic, lazy-open candidate): unique-touched / opened SST ratio p50 = 2.88, fraction ≤ 0.5 = 0.000. Lazy-open dead — every opened SST is queried.
- iter-063 (Path B, doubled `cloud_native_pk_index_sst_open` thread pool 128 → 256, [#72780](https://github.com/StarRocks/starrocks/pull/72780)): pool relief on 5 / 6 BEs (active 71-165 / 256), but BE-127 single OSS GetObject stalled **60 s** with `active = 111 / 256, queue = 0`. **Pool was not the binder**; an OSS-server-side tail on a shared HTTP connection was correlated across all in-flight reads on the cluster-wide S3 client. PR closed.

Falsification verdict: `pivot to Path C — separate connection pool / per-request timeout for PK-index sst_open reads`.

## What

Mirror the existing `S3ClientFactory::OperationType::RENAME_FILE` pattern.

- Add `S3ClientFactory::OperationType::PK_INDEX_SST_OPEN`.
- Plumb it through `RandomAccessFileOptions::s3_operation_type` (a plain int that mirrors the enum so storage code stays free of AWS SDK includes).
- In `S3ClientFactory::new_client(...)` and the file-scope `new_s3client(...)`, a new requestTimeoutMs override is applied when the operation type is `PK_INDEX_SST_OPEN`. The different `requestTimeoutMs` value yields a different `ClientConfiguration` cache key, so the PK-index path gets its own cached `Aws::S3::S3Client` with a separate HTTP connection pool, isolated from cluster-wide S3 traffic.
- All five PK-index SST file open paths route through the new operation type: `PersistentIndexSstable::init` (footer / filter open + corruption-retry), `PersistentIndexSstable::multi_get` (parallel + chunk-parallel branches), `PersistentIndexSstable::new_sstable` (long-lived `rf` for compaction iterators), and `LakePersistentIndex::ingest_sst`.

A `static_assert` block in `fs_s3.cpp` keeps the integer constants in `fs.h` synced with the enum.

## Knob

`object_storage_pk_index_sst_open_request_timeout_ms` (`CONF_mInt64`, default **8000** ms = 8 s).

- Above measured `pindex_init_sst_open_us` p99 (~4.7 s); well below observed 60 s outliers.
- `requestTimeoutMs` has CURL low-speed-time semantics — a slow-but-progressing read is not aborted, only a stuck connection is. Compaction body reads should not be affected.
- Set `-1` to disable the override entirely (falls back to default `S3Client`, no isolation).

## Expected impact

- Stuck OSS GetObject on a single connection of the main client no longer blocks PK-index reads (independent connection pool).
- Within the PK-index pool, a stuck request fails after 8 s; AWS SDK retries on a fresh HTTP connection. Worst-case wall ≈ `8 s + retry × (connect + small_read)`.
- Cold-start `pindex_init_sst_open_us` max should drop from ~60 s to ≤ 10 s; cold-start upsert max should drop from 33-43 s to ≤ 12 s.

## Acceptance (≥ 2 of 3, fresh-deploy A/B vs iter-058 baseline)

1. Cold-start upsert max ≤ **8 s** (STRICT — primary).
2. Cold-start `pindex_init_sst_open_us` max ≤ **10 s** (cap the worst single stall).
3. Cold-start `multi_get_us` p99 ≤ **1.5 s** (re-hold iter-063's 1.02 s + margin).

Hard guardrails: err 0 %; no aborts; vdb peak BW ≤ 700 MB/s; tput ≥ 800 K; per-batch P99 ≤ 800 ms.

## Diff size

6 files, +57 / −2 LoC. Localized; gated by config.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -490,6 +490,27 @@ CONF_mInt32(pk_index_memtable_flush_threadpool_size, "2048");
 // without same-pool re-entrance.
 CONF_mInt32(pk_index_inner_io_threadpool_max_threads, "0");
 CONF_mInt32(pk_index_inner_io_threadpool_size, "1048576");
+// Threadpool for splitting a single PersistentIndexSstable::multi_get into key-chunk tasks
+// so the per-Table::MultiGet sequential cache-miss block-read loop runs on multiple threads.
+// MUST be a different pool from pk_index_inner_io_thread_pool: get_from_sstables already
+// dispatches per-fileset multi_get on inner_io, and re-entering inner_io from inside
+// PersistentIndexSstable::multi_get would deadlock the worker via ThreadPool::wait().
+// Default size is num_cores * 4 because each chunk task is dominated by sleeping on OSS
+// HTTP responses; sizing by cores alone leaves available HTTP concurrency unused.
+CONF_mInt32(pk_index_chunk_io_threadpool_max_threads, "0");
+CONF_mInt32(pk_index_chunk_io_threadpool_size, "1048576");
+// Master kill-switch for the per-multi_get key-chunk parallelism. Tunable so the
+// optimization can be disabled at runtime if it turns pathological under load (e.g.
+// pool starvation, pathological cache contention).
+CONF_mBool(enable_pk_index_parallel_chunk_multi_get, "true");
+// Minimum number of keys in one PersistentIndexSstable::multi_get for the chunk-parallel
+// path to engage. Below this, run the legacy sequential path so we don't pay open-file
+// + dispatch overhead for tiny lookups.
+CONF_mInt32(pk_index_parallel_chunk_min_keys, "32");
+// Target keys per chunk. Effective chunk count = clamp(ceil(N/target), 1, max_chunks).
+CONF_mInt32(pk_index_parallel_chunk_target_keys, "64");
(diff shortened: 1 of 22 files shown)


luohaha:
Closing — never validated. Shipped in iter-064 as the iter-065 experiment payload; iter-065's pipeline aborted before TSP build 1446 deploy + benchmark could measure the acceptance criteria (cold-start upsert max ≤ 8s, pindex_init_sst_open_us max ≤ 10s, multi_get_us p99 ≤ 1.5s). With the run closed, no data supports the change. If the timeout-cap-OSS-tail direction is worth pursuing, re-open with fresh measurements in a future run.

PR #72828, opened 2026-05-06
[BugFix] Fix duplicate format-version key when creating Iceberg table

GavinMar:
## Why I'm doing:

When a user explicitly specifies `"format-version"` in `CREATE TABLE ... PROPERTIES (...)` for an Iceberg table, the statement fails with:

```
ERROR 1064 (HY000): Multiple entries with same key: format-version=2 and format-version=2
```

Reproduction (from StarRocks/StarRocksTest#11276):

```sql
CREATE TABLE iceberg_catalog.db.t (
    granularity string,
    business    string,
    city_id     int,
    date_id     int,
    online_seconds bigint,
    driver_id   bigint
) PARTITION BY (date_id) PROPERTIES ("format-version" = "2");
```

Root cause: `IcebergApiConverter.rebuildCreateTableProperties` first copies all user properties into an `ImmutableMap.Builder` and then unconditionally puts `TableProperties.FORMAT_VERSION` ("format-version") again. Guava's `ImmutableMap.Builder.build()` rejects duplicate keys, hence the IAE. Note that the adjacent `DEFAULT_FILE_FORMAT` put is already correctly guarded with a `containsKey` check — `FORMAT_VERSION` was simply missing the same guard.

## What I'm doing:

- Guard the `FORMAT_VERSION` put with `!createProperties.containsKey(TableProperties.FORMAT_VERSION)` in `IcebergApiConverter.rebuildCreateTableProperties`, mirroring the existing `DEFAULT_FILE_FORMAT` guard.
- When the user supplies `format-version`, their value is preserved (works for both `"1"` and `"2"`); when absent, the previous default of `"2"` is applied — so existing behavior is unchanged.
- Add unit-test coverage in `IcebergApiConverterTest.testRebuildCreateTableProperties` for the three cases: user-supplied `"2"`, user-supplied `"1"`, and absent (default `"2"`).

Because the main and branch-4.1 has fixed this issue, so we only fix the branch-4.0 by this PR.

Fixes StarRocks/StarRocksTest#11276

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergApiConverter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergApiConverter.java
@@ -468,7 +468,9 @@ public class IcebergApiConverter {
         } else if (compressionCodec != null) {
             setCompressionProperties(tableProperties, fileFormat, compressionCodec);
         }
-        tableProperties.put(TableProperties.FORMAT_VERSION, "2");
+        if (!createProperties.containsKey(TableProperties.FORMAT_VERSION)) {
+            tableProperties.put(TableProperties.FORMAT_VERSION, "2");
+        }
 
         return tableProperties.build();
     }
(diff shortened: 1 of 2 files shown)


HangyuanLiu (approved):

stephen-shelby (approved):

GavinMar:
@mergify backport branch-4.0.10

PR #72832, opened 2026-05-06
[BugFix] Clear LambdaArgument.transformedOp before INSERT OVERWRITE re-plan

### LINKED ISSUE #72831: [Bug] INSERT OVERWRITE + UNION ALL + Lambda expressions causes "expr_type does not match slot_type" error
### Summary

When executing `INSERT OVERWRITE` with a query that combines `UNION ALL` and lambda expressions (e.g., `array_map`, `map_from_arrays`), the query fails with a type mismatch error on the BE side. The root cause is that `LambdaArgument.transformedOp` caches a stale `ColumnRefOperator` from the first plan phase, which gets incorrectly reused during the re-plan phase where a new `ColumnRefFactory` assigns the same IDs to different columns.

1. Create tables:

```sql
CREATE DATABASE IF NOT EXISTS repro_bug_db;
USE repro_bug_db;

CREATE TABLE IF NOT EXISTS repro_bug_db.target_table (
    dt DATE,
    id INT,
    name VARCHAR(100),
    kv_map MAP<VARCHAR(100), BIGINT>
) ENGINE=OLAP
DUPLICATE KEY(dt, id)
PARTITION BY RANGE(dt) (
    PARTITION p20260101 VALUES [('2026-01-01'), ('2026-01-02')),
    PARTITION p20260102 VALUES [('2026-01-02'), ('2026-01-03'))
)
PROPERTIES("replication_num" = "1");

CREATE TABLE IF NOT EXISTS repro_bug_db.source_table_1 (
    dt DATE,
    id INT,
    name VARCHAR(100),
    json_col VARCHAR(65533)
) ENGINE=OLAP
DUPLICATE KEY(dt, id)
PROPERTIES("replication_num" = "1");

CREATE TABLE IF NOT EXISTS repro_bug_db.source_table_2 (
    dt DATE,
    id INT,
    name VARCHAR(100),
    json_col VARCHAR(65533)
) ENGINE=OLAP
DUPLICATE KEY(dt, id)
PROPERTIES("replication_num" = "1");
```

2. Insert test data:

```sql
INSERT INTO repro_bug_db.source_table_1 VALUES 
    ('2026-01-01', 1, 'alice', '{"a":10,"b":20}'),
    ('2026-01-01', 2, 'bob', '{"c":30,"d":40}');

INSERT INTO repro_bug_db.source_table_2 VALUES 
    ('2026-01-01', 3, 'charlie', '{"e":50,"f":60}'),
    ('2026-01-01', 4, 'dave', '{"g":70,"h":80}');
```

3. Execute INSERT OVERWRITE with UNION ALL + Lambda:

```sql
INSERT OVERWRITE repro_bug_db.target_table PARTITION(p20260101)
SELECT dt, id, name,
    map_from_arrays(
        cast(json_keys(json_col) as array<string>),
        array_map(x -> cast(get_json_string(json_col, concat('$.', x)) as bigint),
                  cast(json_keys(json_col) as array<string>))
    ) as kv_map
FROM repro_bug_db.source_table_1
WHERE dt = '2026-01-01'
UNION ALL
SELECT dt, id, name,
    map_from_arrays(
        cast(json_keys(json_col) as array<string>),
        array_map(x -> cast(get_json_string(json_col, concat('$.', x)) as bigint),
                  cast(json_keys(json_col) as array<string>))
    ) as kv_map
FROM repro_bug_db.source_table_2
WHERE dt = '2026-01-01';
```

**Note:** All three conditions must be present simultaneously to trigger the bug:
- `INSERT OVERWRITE` (not `INSERT INTO`)
- `UNION ALL`
- Lambda expressions (e.g., `array_map(x -> ...)`)

The `INSERT OVERWRITE` statement should execute successfully, inserting 4 rows (2 from each source table) into the target table's `p20260101` partition.

### Real behavior (Required)

The query fails with the following error:

```
ERROR 1064 (HY000): type of exprs is not match slot's, 
  expr_type=17, slot_type=50, slot_name=dt 
  backend [id=10003] [host=xxx.xxx.xxx.xxx]
```

Where `expr_type=17` corresponds to DATETIME and `slot_type=50` corresponds to DATE.

The FE generates an incorrect execution plan where the `dt` column's type is inferred as DATETIME instead of DATE due to a stale cached `ColumnRefOperator` from the first plan phase.

- Reproduced on: StarRocks 3.3.22
- Also affects: StarRocks `main` branch (no fix present)

You can get the StarRocks version by executing SQL: `select current_version()`

love-star:
Fixes #72831

### Root Cause Analysis

`INSERT OVERWRITE` executes a two-phase plan:
1. First `plan()` — determines partitions, creates `ColumnRefFactory #1`, caches `ColumnRefOperator` in `LambdaArgument.transformedOp`
2. `prepareInsert()` — modifies target partition names to temp partitions
3. Second `plan()` (re-plan) — creates `ColumnRefFactory #2` (IDs start from 1 again), but the AST nodes (including `LambdaArgument`) are **reused**

Since `LambdaArgument.transformedOp` still holds the `ColumnRefOperator` from Factory #1, the re-plan returns stale references. The same ID in Factory #2 may be assigned to a different column with a different type, causing the type mismatch error on BE.

**Key code path** — `SqlToScalarOperatorTranslator.java`:

```java
@Override
public ScalarOperator visitLambdaArguments(LambdaArgument node, Context context) {
    if (node.getTransformed() == null) {   // ← Not null during re-plan! Stale cache hit!
        node.setTransformed(columnRefFactory.create(node.getName(), node.getType(), node.isNullable(), true));
    }
    return node.getTransformed();  // ← Returns stale ColumnRefOperator from old Factory!
}
```

### Why all three conditions are required

| Condition | Explanation |
|-----------|-------------|
| **INSERT OVERWRITE** | Only OVERWRITE triggers a second `plan()` call; INSERT INTO plans only once |
| **UNION ALL** | Increases AST complexity via `processSetOperation()`, creating more ColumnRefOperators and raising the probability of ID collision |
| **Lambda expressions** | Only `LambdaArgument` has the `transformedOp` cache field; other AST nodes don't cache ColumnRefOperators |

### What I do

Clear all `LambdaArgument.transformedOp` caches before re-planning by leveraging the existing `AstTraverser`:

```java
// In InsertOverwriteJobRunner.executeInsert(), before StatementPlanner.plan():
private void clearLambdaArgumentTransformedOps(InsertStmt stmt) {
    new AstTraverser<Void, Void>() {
        @Override
        public Void visitLambdaArguments(LambdaArgument node, Void context) {
            node.setTransformed(null);
            return null;
        }
    }.visit(stmt);
}
```

This approach:
- Uses the official `AstTraverser` which correctly covers **all** AST paths (CTE, ORDER BY, Aggregate, ViewRelation, JoinRelation.onPredicate, Expr-level Subquery nodes, etc.)
- Is only 8 lines of core code
- Has zero performance impact (simple AST traversal, no computation)
- Is future-compatible (new AST node types are automatically handled by the framework)

### Additional context

- The bug is **deterministic in trigger conditions** — it always fails when all three conditions (INSERT OVERWRITE + UNION ALL + Lambda) are met.
- CTE (`WITH` clause) containing UNION ALL + Lambda also triggers the same bug.
- The more MAP/Lambda columns in the query, the higher the probability of ID collision, but even a single Lambda column can trigger it.
- `INSERT INTO` (non-OVERWRITE) is **not** affected because it only plans once.
- Pure `SELECT` queries are **not** affected because they only plan once.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/InsertOverwriteJobRunner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/InsertOverwriteJobRunner.java
@@ -418,7 +418,7 @@ public class InsertOverwriteJobRunner {
      * CTE relations, ORDER BY, aggregate, ViewRelation, JoinRelation.onPredicate,
      * and Expr-level Subquery nodes.
      */
-    private void clearLambdaArgumentTransformedOps(InsertStmt stmt) {
+    public void clearLambdaArgumentTransformedOps(InsertStmt stmt) {
         new AstTraverser<Void, Void>() {
             @Override
             public Void visitLambdaArguments(LambdaArgument node, Void context) {
--- a/fe/fe-core/src/main/java/com/starrocks/load/InsertOverwriteJobRunner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/InsertOverwriteJobRunner.java
@@ -45,12 +45,14 @@ import com.starrocks.sql.analyzer.AlterTableClauseAnalyzer;
 import com.starrocks.sql.analyzer.AnalyzerUtils;
 import com.starrocks.sql.analyzer.SemanticException;
 import com.starrocks.sql.ast.AddPartitionClause;
+import com.starrocks.sql.ast.AstTraverser;
 import com.starrocks.sql.ast.InsertStmt;
 import com.starrocks.sql.ast.ListPartitionDesc;
 import com.starrocks.sql.ast.PartitionDesc;
 import com.starrocks.sql.ast.PartitionRef;
 import com.starrocks.sql.ast.RangePartitionDesc;
 import com.starrocks.sql.ast.expression.Expr;
+import com.starrocks.sql.ast.expression.LambdaArgument;
 import com.starrocks.sql.ast.expression.LiteralExpr;
 import com.starrocks.sql.common.DmlException;
 import com.starrocks.sql.common.MetaUtils;
@@ -370,6 +372,14 @@ public class InsertOverwriteJobRunner {
     private void executeInsert() throws Exception {
         long insertStartTimestamp = System.currentTimeMillis();
         // should replan here because prepareInsert has changed the targetPartitionNames of insertStmt
+        // Before re-plan, clear cached LambdaArgument.transformedOp from the first plan.
+        // The first plan() creates ColumnRefOperators via ColumnRefFactory and caches them in
+        // LambdaArgument.transformedOp. The re-plan creates a NEW ColumnRefFactory, but the AST
+        // nodes (including LambdaArgument) are reused. If transformedOp is not cleared, the stale
+        // ColumnRefOperator (from the old ColumnRefFactory) will be returned by
+        // SqlToScalarOperatorTranslator.visitLambdaArguments(), causing ID collisions and
+        // "expr_type does not match slot_type" errors, especially with UNION ALL + lambda expressions.
+        clearLambdaArgumentTransformedOps(insertStmt);
         try (var guard = context.bindScope()) {
             // For dynamic overwrite, set the txnId to insertStmt so that StatementPlanner.beginTransaction()
             // will skip creating a new transaction and reuse the one we created in prepare().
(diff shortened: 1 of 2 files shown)


stdpain (dismissed):
It's ok to me for the hotfix. We should refactor lambda argument later.

stdpain:
@love-star Could add a UT ?

love-star:
> @love-star Could add a UT ?
yes, this is what I did:
- Added unit tests for the clearLambdaArgumentTransformedOps method to verify it correctly clears all cached LambdaArgument.transformedOp in the AST tree during INSERT OVERWRITE re-plan.
- Changed method visibility of clearLambdaArgumentTransformedOps from private to public to make it directly testable.

stdpain (approved):

CelerData-Reviewer:
@codex review

love-star:
@gengjun-git Could you take a moment to review this PR when you are free?

meegoo (approved):

love-star:
@meegoo It seems that the CI pipeline lacks sufficient resources to deploy StarRocks.

PR #71652, opened 2026-04-14
[Enhancement] Parallelize row-mode partial update publish for lake PK tables

luohaha:
## Why I'm doing:

For shared-data (lake) primary key tables, the row-mode partial update publish processes segments sequentially within a tablet: `load_segment → rewrite_segment → _do_update → release_segment`. The `load_segment` (PK index lookup + reading unmodified columns) and `rewrite_segment` (merging and writing complete segment files) are I/O-heavy operations. For large tablets (e.g., 50GB with many segments), this serial execution becomes a bottleneck.

## What I'm doing:

Parallelize the I/O-heavy `load_segment + rewrite_segment` phase across segments using a dedicated thread pool, while keeping the PK index update phase (`_do_update`) sequential.

**Architecture:**
```
Phase 1 (parallel, per segment):
    load_segment + rewrite_segment  (I/O heavy, parallelized)

Phase 2 (sequential, unchanged):
    _do_update + SST ingestion      (PK index modification, must be serial)
```

**Key changes:**
- Add dedicated `lake_partial_update_thread_pool` (max_threads controls concurrency and memory)
- Add `RowsetUpdateState::prepare_for_parallel()` to pre-initialize shared state before parallel execution
- Add `RowsetUpdateState::release_segment_partial_state()` to release write_columns while keeping upserts for Phase 2
- Make `RowsetUpdateState::_memory_usage` atomic for concurrent updates
- Condition merge (`merge_condition`) falls back to serial path in this version

**Correctness guarantee:** Within a single transaction, all segments read unmodified columns from the base_version snapshot. The unmodified column values are identical regardless of segment processing order, so parallelization is semantically equivalent to serial execution.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
[...]

PR #72838, opened 2026-05-06
[Refactor] Remove skip_pk_preload config and PK update-state preload path

luohaha:
## Why I'm doing:

The PK partial-update preload path (`config::skip_pk_preload` + `UpdateManager::preload_update_state` + the matching call site in `DeltaWriter`) duplicates work that publish already does and is gated off by default (`skip_pk_preload="true"`). It is a recurring source of bugs whenever `RowsetUpdateState`'s lifecycle is refactored — most recently the parallel row-mode partial-update PR (#71652) added a required `prepare()` call on the publish path but missed the preload path, which broke BE UT in DEBUG builds. With preload disabled by default in production, the path is effectively dead weight that imposes ongoing maintenance cost on every refactor in this area.

## What I'm doing:

Removing `skip_pk_preload` and the PK update-state preload entry points end-to-end:

- **Config**: drop `skip_pk_preload` from `config.h`, `config_primary_key_fwd.h`, and `config_fwd_headers_manifest.json`.
- **Lake update manager**: remove `UpdateManager::preload_update_state` (declaration in `update_manager.h`, definition in `update_manager.cpp`) and one stale comment that referenced it.
- **Delta writer call sites**: remove the `skip_pk_preload`-gated preload block in `be/src/storage/delta_writer.cpp` and the corresponding block in `be/src/storage/lake/delta_writer.cpp`.
- **Tests**: drop `LakePrimaryKeyPublishTest.test_abort_txn` (relied on the `UpdateManager::preload_update_state:return` SyncPoint, no longer reachable) and `test_preload_update_state_slow_log`; remove the `skip_pk_preload` parameter from `RowsetColumnPartialUpdateTest`; clean up the temporary `config::skip_pk_preload` tweaks in `test_publish_with_oom`.
- **Docs**: drop the `skip_pk_preload = true` recommendation from `docs/{en,zh,ja}/best_practices/primarykey_table.md`. The 3.4 release notes that historically described the default flip are intentionally left untouched.

### Deliberately preserved
- `preload_compaction_state` (separate compaction-publish preload, not gated by `skip_pk_preload`).
- `UpdateManager::on_rowset_finished` (the shared-nothing API; still driven directly by several existing UTs).
- `pk_preload_*` metric / stat fields (kept reporting 0) for downstream dashboard compatibility.
- `load_pk_preload` failpoint definition (still has its own UT in `load_fail_point_test.cpp`).

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

`skip_pk_preload` was a mutable BE config; removing it means BEs that explicitly set it in `be.conf` will fail to start until the line is removed. Users who relied on flipping it back to `false` lose the preload acceleration; correctness is unaffected because the publish path always loads update state.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1824,7 +1824,6 @@ CONF_Bool(report_python_worker_error, "true");
 CONF_Bool(python_worker_reuse, "true");
 CONF_Int32(python_worker_expire_time_sec, "300");
 CONF_mBool(enable_pk_strict_memcheck, "true");
-CONF_mBool(skip_pk_preload, "true");
 // Reduce core file size by not dumping jemalloc retain pages
 CONF_mBool(enable_core_file_size_optimization, "true");
 // Current supported modules:
--- a/be/src/common/config_fwd_headers_manifest.json
+++ b/be/src/common/config_fwd_headers_manifest.json
@@ -768,7 +768,6 @@
         "primary_key_batch_get_index_memory_limit",
         "pk_dump_interval_seconds",
         "enable_pk_strict_memcheck",
-        "skip_pk_preload",
         "apply_version_slow_log_sec",
         "pk_column_lazy_load_threshold_bytes",
         "column_mode_partial_update_insert_batch_size"
--- a/be/src/common/config_primary_key_fwd.h
+++ b/be/src/common/config_primary_key_fwd.h
@@ -225,8 +225,6 @@ CONF_mInt64(pk_dump_interval_seconds, "3600"); // 1 hour
 
 CONF_mBool(enable_pk_strict_memcheck, "true");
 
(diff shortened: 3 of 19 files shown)


CelerData-Reviewer:
@codex review

sevev (approved):

wyb (approved):

dirtysalt (approved):

luohaha:

PR #72884, opened 2026-05-07
[Feature] Support Iceberg UPDATE analyzer and physical plan on FE

### LINKED ISSUE #72046: Support UPDATE statement on Iceberg V2 tables using Merge-On-Read
- [x] https://github.com/StarRocks/starrocks/pull/73203
- [x] https://github.com/StarRocks/starrocks/pull/72884
- [x] https://github.com/StarRocks/starrocks/pull/72064
## Feature request

StarRocks can query Iceberg tables and already supports `INSERT` and `DELETE` on Iceberg V2 tables, but `UPDATE` is not available. Today users who need to correct or refresh a subset of rows in an Iceberg warehouse have to issue a full-table rewrite (`INSERT OVERWRITE`), or fall back to a manual `DELETE` followed by `INSERT`, which is non-atomic — readers can observe the table in an inconsistent state between the two statements — and loses row-level change semantics for downstream CDC consumers. This gap forces users to keep an external engine around solely for UPDATE, which complicates their lakehouse pipelines.

Support `UPDATE ... SET ... WHERE` on Iceberg format-V2 tables. The write must be atomic (readers see either the pre- or post-UPDATE state, never an intermediate one) and follow the Iceberg V2 Merge-On-Read model so the result is interoperable with Spark and other Iceberg-aware engines.

Initial scope:
- `WHERE` clause required.
- Partition column updates not supported.
- Only Iceberg format V2.

Non-goals for this phase: `MERGE INTO`, partition column updates, UPDATE with correlated subqueries, Copy-On-Write mode.

- Copy-On-Write: rewrite every affected data file. Cheap to read but expensive for selective updates on large partitions.
- Client-side DELETE + INSERT: two statements, not atomic; readers can observe an inconsistent intermediate state.
- Keep UPDATE in an external Iceberg-aware engine: forces users to operate two engines against the same table.

Merge-On-Read is preferred because it matches the Iceberg V2 spec, is cheap for selective updates, and preserves ecosystem interoperability.

**Additional context**
Applies to Iceberg V2 tables only; V1 tables continue to reject UPDATE at analyze time. Possible follow-ups: `MERGE INTO`, partition column updates, UPDATE with `FROM` clause, equality deletes.

GavinMar:
## Why I'm doing:

Add FE-side support for Iceberg `UPDATE` so users can perform row-level updates on Iceberg V2 tables directly from StarRocks. This is step 2 of the Iceberg UPDATE work and builds on the BE `RowDelta` sink landed in the previous commit; the actual commit logic will land in the next PR.

## What I'm doing:

Front-end pieces for Iceberg UPDATE: analyzer, planner, and the `IcebergRowDeltaSink` planner node that serializes to the BE `RowDelta` sink. Execution path wiring and metric plumbing are included.

Analyzer (`UpdateAnalyzer`):
- Rewrites `UPDATE ... SET ... WHERE` on an Iceberg V2 table into `SELECT _file, _pos, <data-cols>, op_code FROM table WHERE condition`.
- Enforces Phase 1 restrictions: `WHERE` clause required, partition column updates rejected (case-insensitive), format V2 only, hidden metadata columns (`_file`, `_pos`) not writable, no CTE / FROM clause.

Planner:
- `UpdatePlanner` builds the physical plan with `IcebergRowDeltaSink`, partition shuffle for partitioned tables, conflict-detection filter for serializable isolation, and proper session-variable save/restore.
- `IcebergPlannerUtils` centralizes shuffle-property construction and scan-node-to-Iceberg-expression translation shared with `DeletePlanner` (`DeletePlanner` is slimmed down to call the shared helpers).
- `IcebergRowDeltaSink` (planner node) carries the per-branch output tuple, `op_code` column index, and `IcebergSinkExtra` for commit-time conflict detection.

Execution wiring:
- `StmtExecutor` recognizes the new sink, fills rewrite files into `IcebergSinkExtra`, and reports UPDATE success/failure through `ConnectorMetricsMgr`.
- `ConnectorMetricsMgr` gains `iceberg_update_*` counters with a `file_type` label (`data`, `position_delete`) matching the BE file-type taxonomy.

Tests:
- `UpdateAnalyzerIcebergTest` covers hidden metadata column rejection, `DEFAULT` normalization, expression / `NULL` assignments, `op_code` literal wiring, property population, output-column-name contract, and case-insensitive partition-column check.
- `UpdatePlanTest` covers partition shuffle on multi-column partitioned update, session-variable restore in `finally`, Iceberg sink flag, complex `WHERE` predicate, pipeline DOP, and output-expr-count invariant.
- `MockIcebergMetadata` exposes V2 mock tables (`t0_v2`, `t1_v2`) used by the new tests.

Fixes: https://github.com/StarRocks/starrocks/issues/72046
StarrocksTest: https://github.com/StarRocks/StarRocksTest/pull/11258

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/metric/ConnectorMetricsMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/metric/ConnectorMetricsMgr.java
@@ -50,6 +50,13 @@ public class ConnectorMetricsMgr {
         final ConcurrentHashMap<String, LongCounterMetric> deleteRows = new ConcurrentHashMap<>();
         final ConcurrentHashMap<String, LongCounterMetric> deleteBytes = new ConcurrentHashMap<>();
 
+        // Update metrics (UPDATE/MERGE with file_type label: data, position_delete)
+        final ConcurrentHashMap<String, LongCounterMetric> updateTotal = new ConcurrentHashMap<>();
+        final ConcurrentHashMap<String, LongCounterMetric> updateDurationMs = new ConcurrentHashMap<>();
+        final ConcurrentHashMap<String, LongCounterMetric> updateRows = new ConcurrentHashMap<>();
+        final ConcurrentHashMap<String, LongCounterMetric> updateBytes = new ConcurrentHashMap<>();
+        final ConcurrentHashMap<String, LongCounterMetric> updateFiles = new ConcurrentHashMap<>();
+
         // Compaction metrics
         final ConcurrentHashMap<String, LongCounterMetric> compactionTotal = new ConcurrentHashMap<>();
         final ConcurrentHashMap<String, LongCounterMetric> compactionDurationMs = new ConcurrentHashMap<>();
@@ -75,6 +82,9 @@ public class ConnectorMetricsMgr {
 
     public static final String DELETE_TYPE_POSITION = "position";
     public static final String DELETE_TYPE_METADATA = "metadata";
+
+    public static final String FILE_TYPE_DATA = "data";
+    public static final String FILE_TYPE_POSITION_DELETE = "position_delete";
 
     public static final String COMPACTION_TYPE_MANUAL = "manual";
     public static final String COMPACTION_TYPE_AUTO = "auto";
@@ -238,6 +248,83 @@ public class ConnectorMetricsMgr {
             MetricRepo.addMetric(metric);
             return metric;
         }).increase(bytes);
+    }
+
+    // ======================= Update Metrics =======================
+
+    public static void increaseUpdateTotal(String connectorType, String status, String reason) {
+        String normalizedStatus = normalizeStatus(status);
+        String normalizedReason = normalizeReason(reason);
+
+        ConnectorMetrics m = getOrCreate(connectorType);
+        String key = normalizedStatus + "|" + normalizedReason;
+        m.updateTotal.computeIfAbsent(key, k -> {
+            LongCounterMetric metric = new LongCounterMetric(connectorType + "_update_total",
+                    Metric.MetricUnit.REQUESTS,
+                    "total " + connectorType + " update tasks by status and reason");
+            metric.addLabel(new MetricLabel("status", normalizedStatus));
+            metric.addLabel(new MetricLabel("reason", normalizedReason));
+            MetricRepo.addMetric(metric);
+            return metric;
+        }).increase(1L);
+    }
+
+    public static void increaseUpdateTotalSuccess(String connectorType) {
+        increaseUpdateTotal(connectorType, STATUS_SUCCESS, REASON_NONE);
+    }
+
+    public static void increaseUpdateTotalFail(String connectorType, Throwable throwable) {
+        increaseUpdateTotal(connectorType, STATUS_FAILED, classifyFailReason(throwable));
+    }
+
+    public static void increaseUpdateTotalFail(String connectorType, String errorMessage) {
+        increaseUpdateTotal(connectorType, STATUS_FAILED, classifyFailReason(errorMessage));
+    }
+
+    public static void increaseUpdateDurationMs(String connectorType, long durationMs) {
+        ConnectorMetrics m = getOrCreate(connectorType);
+        m.updateDurationMs.computeIfAbsent("default", k -> {
+            LongCounterMetric metric = new LongCounterMetric(connectorType + "_update_duration_ms_total",
+                    Metric.MetricUnit.MILLISECONDS,
+                    "total duration in milliseconds of " + connectorType + " update operations");
+            MetricRepo.addMetric(metric);
+            return metric;
+        }).increase(durationMs);
+    }
+
+    public static void increaseUpdateRows(String connectorType, long rows) {
+        ConnectorMetrics m = getOrCreate(connectorType);
+        m.updateRows.computeIfAbsent("default", k -> {
+            LongCounterMetric metric = new LongCounterMetric(connectorType + "_update_rows",
+                    Metric.MetricUnit.ROWS,
+                    "total rows affected by " + connectorType + " update operations");
+            MetricRepo.addMetric(metric);
+            return metric;
+        }).increase(rows);
+    }
+
+    public static void increaseUpdateBytes(String connectorType, long bytes, String fileType) {
+        ConnectorMetrics m = getOrCreate(connectorType);
+        m.updateBytes.computeIfAbsent(fileType, k -> {
+            LongCounterMetric metric = new LongCounterMetric(connectorType + "_update_bytes",
+                    Metric.MetricUnit.BYTES,
+                    "total bytes written by " + connectorType + " update operations by file type");
+            metric.addLabel(new MetricLabel("file_type", fileType));
+            MetricRepo.addMetric(metric);
+            return metric;
+        }).increase(bytes);
+    }
+
+    public static void increaseUpdateFiles(String connectorType, long files, String fileType) {
+        ConnectorMetrics m = getOrCreate(connectorType);
+        m.updateFiles.computeIfAbsent(fileType, k -> {
+            LongCounterMetric metric = new LongCounterMetric(connectorType + "_update_files",
+                    Metric.MetricUnit.NOUNIT,
+                    "total files written by " + connectorType + " update operations by file type");
+            metric.addLabel(new MetricLabel("file_type", fileType));
+            MetricRepo.addMetric(metric);
+            return metric;
+        }).increase(files);
     }
 
     // ======================= Compaction Metrics =======================
--- a/fe/fe-core/src/main/java/com/starrocks/planner/IcebergDeleteSink.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/IcebergDeleteSink.java
@@ -137,8 +137,10 @@ public class IcebergDeleteSink extends DataSink {
         tIcebergTableSink.setData_location(dataLocation);
         tIcebergTableSink.setFile_format("parquet"); // Delete files are always parquet
         tIcebergTableSink.setIs_static_partition_sink(false);
+        // DeleteSink only emits position-delete files; the codec belongs in the
+        // delete-file slot. `compression_type` is reserved for data files now.
         TCompressionType compression = PARQUET_COMPRESSION_TYPE_MAP.get(compressionType);
-        tIcebergTableSink.setCompression_type(compression);
+        tIcebergTableSink.setDelete_compression_type(compression);
         tIcebergTableSink.setTarget_max_file_size(targetMaxFileSize);
         com.starrocks.thrift.TCloudConfiguration tCloudConfiguration = new com.starrocks.thrift.TCloudConfiguration();
         cloudConfiguration.toThrift(tCloudConfiguration);
--- a/fe/fe-core/src/main/java/com/starrocks/planner/IcebergDeleteSink.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/IcebergDeleteSink.java
@@ -70,8 +70,7 @@ public class IcebergDeleteSink extends DataSink {
         this.compressionType = nativeTable.properties().getOrDefault(DELETE_PARQUET_COMPRESSION,
                 nativeTable.properties().getOrDefault(PARQUET_COMPRESSION,
                         sessionVariable.getConnectorSinkCompressionCodec()));
-        this.targetMaxFileSize = sessionVariable.getConnectorSinkTargetMaxFileSize() > 0 ?
-                sessionVariable.getConnectorSinkTargetMaxFileSize() : 1024L * 1024 * 1024;
+        this.targetMaxFileSize = IcebergUtil.resolveTargetMaxFileSize(nativeTable, sessionVariable);
     }
 
     public void init() {
--- a/fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
@@ -19,7 +19,6 @@ import com.starrocks.connector.exception.StarRocksConnectorException;
 import com.starrocks.connector.iceberg.IcebergUtil;
 import com.starrocks.credential.CloudConfiguration;
 import com.starrocks.qe.SessionVariable;
-import com.starrocks.thrift.TCompressionType;
 import com.starrocks.thrift.TDataSink;
 import com.starrocks.thrift.TDataSinkType;
 import com.starrocks.thrift.TExplainLevel;
@@ -74,7 +73,8 @@ public class IcebergRowDeltaSink extends DataSink {
     private final String fileFormat;
     private final String tableLocation;
     private final String dataLocation;
-    private final String compressionType;
+    private final String dataCompressionType;
+    private final String deleteCompressionType;
     private final long targetMaxFileSize;
     private final String tableIdentifier;
     private CloudConfiguration cloudConfiguration;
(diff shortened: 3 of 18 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
```diff
@@ -3482,7 +3483,15 @@ public void handleDMLStmt(ExecPlan execPlan, DmlStmt stmt) throws Exception {
...
                 DataSink dataSink = execPlan.getFragments().get(0).getSink();
-                if (dataSink instanceof IcebergDeleteSink deleteSink) {
+                if (dataSink instanceof IcebergRowDeltaSink rowDeltaSink) {
+                    // This is an UPDATE/MERGE operation, use IcebergRowDeltaSink.
+                    // NOTE: the row-delta commit path in MetadataMgr.finishSink (mixing new DATA files
+                    // with POSITION_DELETES into a single Iceberg RowDelta) is added in a follow-up PR;
+                    // this PR only wires the FE planner / executor entry points.
+                    IcebergMetadata.IcebergSinkExtra extra = rowDeltaSink.getSinkExtraInfo();
+                    context.getGlobalStateMgr().getMetadataMgr().finishSink(
+                            catalogName, dbName, tableName, commitInfos, null, extra, context);
```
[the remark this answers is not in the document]
GavinMar: Acknowledged — this is a real issue and is intentionally out of scope for this PR.

This is step 2 of a multi-PR series for Iceberg UPDATE on the FE side. The row-delta commit path in `MetadataMgr.finishSink` (correctly routing the mixed `data files + POSITION_DELETES` batch to an Iceberg `RowDelta` transaction instead of `commitDeleteOperation`) is implemented in the next PR (step 3).

See the inline note added at the `IcebergRowDeltaSink` branch in `StmtExecutor` calling this out, so end-to-end Iceberg UPDATE is not safe to run until the step-3 PR lands. The analyzer / planner / sink wiring in this PR is gated on that follow-up.

fe/fe-core/src/main/java/com/starrocks/sql/ast/UpdateStmt.java
```diff
@@ -36,6 +36,20 @@ public class UpdateStmt extends DmlStmt {
 
     private boolean usePartialUpdate;
 
+    // Column names for Iceberg RowDelta output tuple, set by UpdateAnalyzer.
+    // Needed because getColumnOutputNames() may not match the SELECT list
+    // (optimizer can expand it), and ColumnRefOperator.getName() returns
+    // "expr" for SET expressions instead of the target column name.
+    private List<String> icebergColumnOutputNames;
```
HangyuanLiu: I think this field should not live on `UpdateStmt`.

`icebergColumnOutputNames` is a temporary Iceberg RowDelta planning detail: `UpdateAnalyzer` derives it from the generated SELECT list aliases, then `UpdatePlanner` reads it only in the Iceberg branch. `UpdateStmt` is a shared AST node, so adding a format-specific ferry field here makes a derived intermediate value part of the generic UPDATE statement contract.

Can we keep this local to the Iceberg planning boundary instead? The planner can derive these names from the existing generated `QueryStatement` / `SelectRelation` select-list aliases, or wrap them in an Iceberg-specific sink payload/helper when building the RowDelta sink.

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
```diff
@@ -0,0 +1,223 @@
...
+        OpCode(int value) {
+            this.value = value;
+        }
+
+        public int value() {
+            return value;
+        }
+    }
+
+    protected final TupleDescriptor desc;
```
HangyuanLiu, line 70: Can we avoid introducing `TupleDescriptor` into this new RowDelta sink API?

`TupleDescriptor` is already marked `@Deprecated` on the FE side, and this sink only seems to need the RowDelta output schema/layout: `_file`, `_pos`, data columns, `op_code`, plus the corresponding names/types/nullability and layout indices. Those are already derivable from `colNames`, `outputExprs`, and the fixed RowDelta layout, or can be passed as an Iceberg-specific sink payload.

Using a deprecated generic descriptor here makes the new Iceberg path depend on an abstraction we are trying to move away from. It would be cleaner to pass the required RowDelta schema/layout explicitly and keep `TupleDescriptor` out of this new sink.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/UpdateAnalyzer.java
```diff
@@ -70,6 +75,151 @@ private static void analyzeProperties(UpdateStmt updateStmt, ConnectContext sess
...
+                selectList.addItem(new SelectListItem(assignExpr, col.getName()));
+            } else {
+                selectList.addItem(new SelectListItem(new SlotRef(tableName, col.getName()), col.getName()));
+            }
+        }
+
+        // Add op_code = OP_UPDATE as the last column
+        try {
+            selectList.addItem(new SelectListItem(
+                    new IntLiteral(IcebergRowDeltaSink.OpCode.UPDATE.value(), IntegerType.TINYINT), "op_code"));
```
HangyuanLiu, line 175: I am not sure this `op_code` needs to be part of the generated SELECT output for the current UPDATE path.

For a pure Iceberg UPDATE, every output row has the same operation semantics: delete the old row by `_file` / `_pos`, then write the updated data row. Since this plan already uses an Iceberg RowDelta-specific sink, could the sink type / write mode imply that all incoming rows are UPDATE rows, similar to how `IcebergDeleteSink` does not require the analyzer to append a per-row DELETE marker?

Putting `IcebergRowDeltaSink.OpCode.UPDATE` into the analyzer-generated SELECT makes a sink-internal routing protocol part of the query output layout. It also creates an implicit contract that `op_code` is the last output column and that BE should route by that column. If row-level mixed operations are needed later for MERGE, could that be modeled as a RowDelta/MERGE-specific routing expression or sink payload instead of adding a synthetic column to the generic SELECT output?

fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
```diff
@@ -3482,7 +3483,15 @@ public void handleDMLStmt(ExecPlan execPlan, DmlStmt stmt) throws Exception {
                 List<TSinkCommitInfo> commitInfos = coord.getSinkCommitInfos();
 
                 DataSink dataSink = execPlan.getFragments().get(0).getSink();
-                if (dataSink instanceof IcebergDeleteSink deleteSink) {
+                if (dataSink instanceof IcebergRowDeltaSink rowDeltaSink) {
+                    // This is an UPDATE/MERGE operation, use IcebergRowDeltaSink.
+                    // NOTE: the row-delta commit path in MetadataMgr.finishSink (mixing new DATA files
+                    // with POSITION_DELETES into a single Iceberg RowDelta) is added in a follow-up PR;
+                    // this PR only wires the FE planner / executor entry points.
+                    IcebergMetadata.IcebergSinkExtra extra = rowDeltaSink.getSinkExtraInfo();
```
HangyuanLiu, line 3491: The extras section doesn't seem to have any settings for adding or deleting data. Am I missing it, or is the functionality not yet complete?

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
```diff
@@ -0,0 +1,223 @@
...
+        this.icebergTable = icebergTable;
+        Table nativeTable = icebergTable.getNativeTable();
+        this.desc = desc;
+        this.tableLocation = nativeTable.location();
+        this.dataLocation = IcebergUtil.tableDataLocation(nativeTable);
+        this.targetTableId = icebergTable.getId();
+        this.tableIdentifier = icebergTable.getUUID();
+        this.fileFormat = nativeTable.properties().getOrDefault(DEFAULT_FILE_FORMAT, DEFAULT_FILE_FORMAT_DEFAULT)
+                .toLowerCase();
+        // Priority: write.delete.parquet.compression-codec > write.parquet.compression-codec > session variable
```
Youngwb: I think we need two compression-codec, one for delete, one for data file

fe/fe-core/src/main/java/com/starrocks/metric/ConnectorMetricsMgr.java
```diff
@@ -50,6 +50,13 @@ private static class ConnectorMetrics {
         final ConcurrentHashMap<String, LongCounterMetric> deleteRows = new ConcurrentHashMap<>();
         final ConcurrentHashMap<String, LongCounterMetric> deleteBytes = new ConcurrentHashMap<>();
 
+        // Update metrics (UPDATE/MERGE with file_type label: data, position_delete)
+        final ConcurrentHashMap<String, LongCounterMetric> updateTotal = new ConcurrentHashMap<>();
```
Youngwb, line 54: remember to update doc

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
```diff
@@ -0,0 +1,223 @@
...
+    private com.starrocks.connector.iceberg.IcebergMetadata.IcebergSinkExtra sinkExtraInfo;
+
+    /**
+     * Constructor for IcebergRowDeltaSink
+     *
+     * @param icebergTable    The target Iceberg table
+     * @param desc            Tuple descriptor containing operation columns
+     * @param sessionVariable Session variables for configuration
+     */
+    public IcebergRowDeltaSink(IcebergTable icebergTable, TupleDescriptor desc,
```
Youngwb, line 90: We could abstract a base class for icebergTableSink/icebergDeleteSink/IcebergRowDeltaSink, you can refactor these later

fe/fe-core/src/main/java/com/starrocks/sql/ast/UpdateStmt.java
GavinMar: Good point, will fix it.

fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
GavinMar: The absence of an add/delete file list in the extras is expected; RowDelta's add/delete files come from the commitInfos returned by BE, not IcebergSinkExtra. As the comment here states, the current PR's commit routing is not yet complete, and mixed commit handling will be added in the next step.

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
```diff
@@ -0,0 +1,223 @@
...
+        this.dataLocation = IcebergUtil.tableDataLocation(nativeTable);
+        this.targetTableId = icebergTable.getId();
+        this.tableIdentifier = icebergTable.getUUID();
+        this.fileFormat = nativeTable.properties().getOrDefault(DEFAULT_FILE_FORMAT, DEFAULT_FILE_FORMAT_DEFAULT)
+                .toLowerCase();
+        // Priority: write.delete.parquet.compression-codec > write.parquet.compression-codec > session variable
+        this.compressionType = nativeTable.properties().getOrDefault(DELETE_PARQUET_COMPRESSION,
+                nativeTable.properties().getOrDefault(PARQUET_COMPRESSION,
+                        sessionVariable.getConnectorSinkCompressionCodec()));
+        this.targetMaxFileSize = sessionVariable.getConnectorSinkTargetMaxFileSize() > 0 ?
```
Youngwb: we need to fellow the table property "write.target-file-size-bytes"

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
GavinMar: Makes sense. Considering that other sinks in Iceberg also rely on this `TupleDescriptor`, I will submit a separate PR later to make unified modifications instead of handling this issue in the current feature.

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/UpdateAnalyzer.java
GavinMar: `op_code` is not the output of the user's `SELECT` statement, but rather the routing column of the internal `RowDelta` payload. This design aims to reuse the same set of `BE` row routing for `UPDATE` and `MERGE`. From the overall design perspective, we also consider `UPDATE` as a special case of `MERGE`. Therefore, this is to simplify and unify the `BE`'s processing logic for both. Thus, the `BE` itself is unaware whether the request is triggered by an `UPDATE` or a `MERGE`; it only has the concept of row delta sink. In addition to `op_code`, `_file` and `_pos` are also added to the `SELECT` statement in both `DELETE` and `UPDATE`, and their positions within the `SELECT` statement are predetermined. Of course, you have a point; `routing_expr` is more decoupled, but it introduces additional sink-side expression binding complexity. Therefore, weighing the pros and cons, I prefer to use the existing model for now.

fe/fe-core/src/main/java/com/starrocks/sql/ast/UpdateStmt.java
GavinMar: fixed.

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
GavinMar: fixed in https://github.com/StarRocks/starrocks/pull/72884/changes/b1014d9efe08b40aacd08bd0a39b7b0db491eb0c

fe/fe-core/src/main/java/com/starrocks/metric/ConnectorMetricsMgr.java
GavinMar: OK, a later pr will update the related doc.

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
GavinMar: OK, will refactor it i another pr.

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
GavinMar: fixed in https://github.com/StarRocks/starrocks/pull/72884/changes/1ad6d6cdae04325233263a5165ac3520158aed00

stephen-shelby (approved):

Youngwb (approved):

HangyuanLiu (approved):

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/UpdateAnalyzer.java
HangyuanLiu: modify in next pr

trueeyu (approved):

fe/fe-core/src/main/java/com/starrocks/planner/IcebergDeleteSink.java
```diff
@@ -137,8 +136,10 @@ protected TDataSink toThrift() {
         tIcebergTableSink.setData_location(dataLocation);
         tIcebergTableSink.setFile_format("parquet"); // Delete files are always parquet
         tIcebergTableSink.setIs_static_partition_sink(false);
+        // DeleteSink only emits position-delete files; the codec belongs in the
+        // delete-file slot. `compression_type` is reserved for data files now.
         TCompressionType compression = PARQUET_COMPRESSION_TYPE_MAP.get(compressionType);
-        tIcebergTableSink.setCompression_type(compression);
+        tIcebergTableSink.setDelete_compression_type(compression);
         tIcebergTableSink.setTarget_max_file_size(targetMaxFileSize);
```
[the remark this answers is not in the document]
GavinMar: Will fix in step3.

fe/fe-core/src/main/java/com/starrocks/planner/IcebergRowDeltaSink.java
```diff
@@ -0,0 +1,226 @@
...
+        tIcebergTableSink.setTarget_table_id(targetTableId);
+        tIcebergTableSink.setTuple_id(desc.getId().asInt());
+        tIcebergTableSink.setLocation(tableLocation);
+        tIcebergTableSink.setData_location(dataLocation);
+        tIcebergTableSink.setFile_format(fileFormat);
+        tIcebergTableSink.setIs_static_partition_sink(false);
+        tIcebergTableSink.setWrite_mode(TIcebergWriteMode.ROW_DELTA);
+        tIcebergTableSink.setCompression_type(PARQUET_COMPRESSION_TYPE_MAP.get(dataCompressionType));
+        tIcebergTableSink.setDelete_compression_type(PARQUET_COMPRESSION_TYPE_MAP.get(deleteCompressionType));
+        tIcebergTableSink.setTarget_max_file_size(targetMaxFileSize);
```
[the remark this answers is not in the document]
GavinMar: ditto

PR #72926, opened 2026-05-07
[Enhancement] Compaction-aware async vector index build scheduling

sevev:
## Why I'm doing:

Today the async vector-index build scheduler (`VectorIndexBuildScheduler`) is single-version-aware: it only knows the newest pending version per tablet. This causes two problems on shared-data tables under realistic load:

1. **Wasted builds.** A compaction publishes shortly after a load, but the scheduler may have already dispatched a build at the load version. The just-finished VI is then immediately invalidated by the compaction's new rowset layout, so we rebuild and the first build is wasted.
2. **Missed coalescing.** Conversely, when a compaction lands first, we have no way to wait briefly for an in-flight load tail and merge them — we either dispatch immediately (and rebuild on the load tail) or rely on a uniform delay that hurts the compaction case.

We need scheduling that distinguishes "version produced by compaction" from "version produced by load," so that compaction products dispatch immediately while load tails wait briefly for a coalescing compaction.

## What I'm doing:

Introduce a two-version frontier on each pending tablet:

- `latestVersion` — newest pending version (load or compaction).
- `latestCompactionVersion` — newest pending compaction-produced version.

Two-phase scheduling in `scheduleFromPending`:

- **Phase 1**: `built < latestCompactionVersion` → dispatch immediately to `latestCompactionVersion`. Compaction products are not at risk of being swept by an imminent compaction, so we don't wait.
- **Phase 2**: `built >= latestCompactionVersion && built < latestVersion` → load-only tail. Wait `lake_vi_build_load_tail_delay_ms` (default 5min); if a new compaction arrives within that window it bumps `latestCompactionVersion` and we go back to phase 1, merging the tail into the combined build instead of rebuilding twice.

API changes:

- `addPendingTablet(tabletId, version, fromCompaction)`
- `onPublishComplete(infos, fromCompaction)`

`PublishVersionDaemon` derives `fromCompaction` from `TransactionState.LoadJobSourceType.LAKE_COMPACTION` (per-batch and per-partition publish paths). All non-publish callers (`LakeUtils`, `LakeRollupJob`, `LakeTableSchemaChangeJob`, `LakeTableAlterMetaJobBase`, `SplitTabletJob`, `MergeTabletJob`) pass `fromCompaction=false` since they are not compaction-driven publishes.

Recovery scan after leader switch is conservative: the in-memory compaction-frontier is lost, so we set `latestCompactionVersion = visibleVersion` to re-dispatch the backlog in phase 1 (no extra delay).

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/vector/VectorIndexBuildScheduler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/vector/VectorIndexBuildScheduler.java
@@ -11,6 +11,7 @@ import com.starrocks.catalog.Table;
 import com.starrocks.catalog.Tablet;
 import com.starrocks.catalog.TabletInvertedIndex;
 import com.starrocks.catalog.TabletMeta;
+import com.starrocks.common.Config;
 import com.starrocks.common.util.FrontendDaemon;
 import com.starrocks.lake.LakeTablet;
 import com.starrocks.proto.BuildVectorIndexResponse;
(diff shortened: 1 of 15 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/lake/vector/VectorIndexBuildScheduler.java
```diff
@@ -279,16 +366,30 @@ void cleanupStaleEntries() {
 
     // ========== Schedule from pending queue ==========
 
+    /**
+     * Two-phase scheduling. For each pending tablet:
+     * <ul>
+     *   <li>Phase 1: {@code built < latestCompactionVersion} → dispatch immediately with
```
srlch, line 372: Is it possible that: `built < latestCompactionVersion < visible version`? In Phase 1, the index will be built immediately but if `latestCompactionVersion < visible version`, it means this is still possible that it will become wasted?

srlch:
@codex review

fe/fe-core/src/main/java/com/starrocks/lake/vector/VectorIndexBuildScheduler.java
sevev: > latestCompactionVersion

Yes，the assumption is that the output rowset generated by compaction will not be compacted again immediately.If this output rowset does get compacted right away, it will indeed cause redundant waste again.

srlch (dismissed):

srlch (approved):

gengjun-git (approved):

PR #72945, opened 2026-05-08
[Enhancement] per-catalog Kerberos credential isolation for Iceberg/Hive catalogs sharing one HMS

kwontaeheon:
## Why I'm doing:
Override `hive-site.xml` with catalog properties same as hive type catalog.

## What I'm doing:

Apply catalog properties to Hadoop Configuration so that per-catalog overrides 
(e.g. hive.metastore.kerberos.principal, hive.metastore.uris, hadoop.kerberos.principal/keytab) take precedence over the globally baked hive-site.xml / hdfs-site.xml. 

This mirrors what HiveMetaClient.createHiveMetaClient does for the Hive connector
(HiveMetaClient.java: `properties.forEach(conf::set);`)

https://github.com/StarRocks/starrocks/blob/main/fe/fe-core/src/main/java/com/starrocks/connector/hive/HiveMetaClient.java#L82

 and is required when a single FE serves multiple HMS endpoints with different service principals.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/hive/HiveMetaClient.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/hive/HiveMetaClient.java
@@ -20,6 +20,7 @@ import com.starrocks.common.profile.Timer;
 import com.starrocks.common.profile.Tracers;
 import com.starrocks.connector.HdfsEnvironment;
 import com.starrocks.connector.exception.StarRocksConnectorException;
+import com.starrocks.connector.hadoop.HadoopExt;
 import com.starrocks.connector.hive.events.MetastoreNotificationFetchException;
 import com.starrocks.connector.hive.glue.AWSCatalogMetastoreClient;
 import org.apache.commons.lang.exception.ExceptionUtils;
@@ -45,12 +46,14 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import static com.starrocks.common.profile.Tracers.Module.EXTERNAL;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_CONNECTION_POOL_SIZE;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_TIMEOUT;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_TYPE;
 import static com.starrocks.connector.hive.HiveConnector.HIVE_METASTORE_URIS;
+import static org.apache.hadoop.hive.common.FileUtils.unescapePathName;
 
 public class HiveMetaClient {
     private static final Logger LOG = LogManager.getLogger(HiveMetaClient.class);
@@ -67,12 +70,20 @@ public class HiveMetaClient {
     private final Object clientPoolLock = new Object();
 
     private final HiveConf conf;
+    // Catalog-level hadoop.kerberos.principal/keytab. If either is empty, fall back to the
+    // system UGI (original behavior). HadoopExt.doAsWithSwap briefly swaps the process
+    // loginUser to the catalog keytab around each RPC so that the HMS thrift SASL handshake
+    // authenticates with the catalog credential.
+    private final String catalogPrincipal;
+    private final String catalogKeytab;
 
     // Required for creating an instance of RetryingMetaStoreClient.
     private static final HiveMetaHookLoader DUMMY_HOOK_LOADER = tbl -> null;
 
     public HiveMetaClient(HiveConf conf) {
         this.conf = conf;
+        this.catalogPrincipal = conf.get(HadoopExt.HADOOP_KERBEROS_PRINCIPAL);
+        this.catalogKeytab = conf.get(HadoopExt.HADOOP_KERBEROS_KEYTAB);
         this.maxPoolSize = conf.getInt(HIVE_METASTORE_CONNECTION_POOL_SIZE, MAX_HMS_CONNECTION_POOL_SIZE_DEFAULT);
     }
 
@@ -110,18 +121,14 @@ public class HiveMetaClient {
         // When the number of currently used clients is less than maxPoolSize,
         // the client will be recycled and reused. If it does, we close the client.
         public void finish() {
-            boolean shouldClose = false;
             synchronized (clientPoolLock) {
                 if (clientPool.size() >= maxPoolSize) {
-                    shouldClose = true;
+                    LOG.warn("There are more than {} connections currently accessing the metastore",
+                            maxPoolSize);
+                    close();
                 } else {
                     clientPool.offer(this);
                 }
-            }
-            if (shouldClose) {
-                LOG.warn("There are more than {} connections currently accessing the metastore",
-                        maxPoolSize);
-                close();
             }
         }
 
@@ -148,12 +155,17 @@ public class HiveMetaClient {
             Thread.currentThread().setContextClassLoader(ClassLoader.getSystemClassLoader());
         }
 
-        RecyclableClient client = null;
         synchronized (clientPoolLock) {
-            client = clientPool.poll();
+            RecyclableClient client = clientPool.poll();
+            // The pool was empty so create a new client and return that.
+            // Serialize client creation to defend against possible race conditions accessing
+            // local Kerberos state
+            if (client == null) {
+                return new RecyclableClient(conf);
+            } else {
+                return client;
+            }
         }
-
-        return client != null ? client : new RecyclableClient(conf);
     }
 
     public <T> T callRPC(String methodName, String messageIfError, Object... args) {
@@ -161,12 +173,17 @@ public class HiveMetaClient {
     }
 
     public <T> T callRPC(String methodName, String messageIfError, Class<?>[] argClasses, Object... args) {
+        final Class<?>[] finalArgClasses = argClasses == null ? ClassUtils.getCompatibleParamClasses(args) : argClasses;
+        return HadoopExt.getInstance().doAsWithSwap(catalogPrincipal, catalogKeytab,
+                () -> callRPCInternal(methodName, messageIfError, finalArgClasses, args));
+    }
+
+    private <T> T callRPCInternal(String methodName, String messageIfError, Class<?>[] argClasses, Object... args) {
         RecyclableClient client = null;
         StarRocksConnectorException connectionException = null;
 
         try {
             client = getClient();
-            argClasses = argClasses == null ? ClassUtils.getCompatibleParamClasses(args) : argClasses;
             Method method = client.hiveClient.getClass().getDeclaredMethod(methodName, argClasses);
             return (T) method.invoke(client.hiveClient, args);
         } catch (Throwable e) {
@@ -314,8 +331,10 @@ public class HiveMetaClient {
             RecyclableClient client = null;
             StarRocksConnectorException connectionException = null;
             try {
+                List<String> decodedPartitionNames = partitionNames.stream().
+                        map(name -> unescapePathName(name)).collect(Collectors.toList());
                 client = getClient();
-                partitions = client.hiveClient.getPartitionsByNames(dbName, tblName, partitionNames);
+                partitions = client.hiveClient.getPartitionsByNames(dbName, tblName, decodedPartitionNames);
                 if (partitions.size() != partitionNames.size()) {
                     LOG.warn("Expect to fetch {} partition on [{}.{}], but actually fetched {} partition",
                             partitionNames.size(), dbName, tblName, partitions.size());
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergConnector.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergConnector.java
@@ -92,6 +92,16 @@ public class IcebergConnector implements Connector {
         IcebergCatalogType nativeCatalogType = icebergCatalogProperties.getCatalogType();
         Configuration conf = hdfsEnvironment.getConfiguration();
 
+        // Apply catalog properties to Hadoop Configuration so that per-catalog
+        // overrides (e.g. hive.metastore.kerberos.principal, hive.metastore.uris,
+        // hadoop.kerberos.principal/keytab) take precedence over the globally
+        // baked hive-site.xml / hdfs-site.xml. This mirrors what
+        // HiveMetaClient.createHiveMetaClient does for the Hive connector
+        // (HiveMetaClient.java: `properties.forEach(conf::set);`) and is required
+        // when a single FE serves multiple HMS endpoints with different
+        // service principals.
+        properties.forEach(conf::set);
+
         if (Config.enable_iceberg_custom_worker_thread) {
             LOG.info("Default iceberg worker thread number changed " + Config.iceberg_worker_num_threads);
             Properties props = System.getProperties();
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/hive/IcebergHiveCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/hive/IcebergHiveCatalog.java
@@ -24,6 +24,8 @@ import com.starrocks.common.Config;
 import com.starrocks.common.MetaNotFoundException;
 import com.starrocks.common.util.Util;
 import com.starrocks.connector.exception.StarRocksConnectorException;
+import com.starrocks.connector.hadoop.GenericExceptionAction;
+import com.starrocks.connector.hadoop.HadoopExt;
 import com.starrocks.connector.iceberg.IcebergCatalog;
 import com.starrocks.connector.iceberg.IcebergCatalogType;
 import com.starrocks.connector.iceberg.cost.IcebergMetricsReporter;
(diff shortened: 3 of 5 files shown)


kwontaeheon:
@wangsimo0 This change makes starrocks to connect multiple HMS using different kerberos credentials. If I checked options incorrectly, please let me know.

CelerData-Reviewer:
@codex review

kwontaeheon:
## Extension: per-catalog Kerberos keytab isolation (4 files added on top of the existing IcebergConnector change)

The original commit on this PR (`feat: override catalog properties`) propagates catalog properties into the Hadoop `Configuration` so a single FE can serve multiple HMS endpoints with different service principals. This was necessary but not sufficient — even with the right SPN, the **client-side Kerberos credential** is still a single process-wide UGI (the system keytab loaded from `KRB5*` env), so two catalogs in the same FE cannot authenticate with two different client keytabs.

This new commit extends the same idea: in addition to overriding the SPN/conf, **the client UGI itself is overridden per catalog**.

### Mechanism

`HadoopExt.doAsWithSwap(principal, keytab, action)`, under a single process-wide lock:
1. `loginUserFromKeytab(catalog_principal, catalog_keytab)` — swaps process loginUser to the catalog keytab
2. runs `action` inside that UGI's `doAs` — HMS thrift SASL handshake / HDFS RPC pick up the catalog credential
3. `finally`: restores the system default loginUser from `KRB5*` env (skip if env is unset — next swap will set it again)

All entry points route through this helper:
- `IcebergHiveCatalog.withCatalogUgi(action) → doAsWithSwap`
- `HiveMetaClient.callRPC(...) → doAsWithSwap`

### Why thread-local `doAs` alone is not enough

We initially tried the standard `UserGroupInformation.doAs(catalogUgi, action)` pattern, but Hadoop's HMS thrift client / SASL handshake reaches process-wide state in several places (logged-in user inside `HiveMetaStoreClient.open`, JGSS credential cache when `useSubjectCredsOnly=false`, cached `HiveClientPool` instances). In a multi-catalog scenario the first call's UGI gets entrenched on these caches and subsequent calls from other catalogs reuse it.

Tested alternatives that did **not** work for full isolation:
- `loginUserFromKeytabAndReturnUGI` + plain `doAs` (thread-local only, doesn't propagate to SASL)
- `-Djavax.security.auth.useSubjectCredsOnly=true` (Hadoop still reads process loginUser in a few paths)
- Disabling Iceberg `CachedClientPool` (`CLIENT_POOL_CACHE_EVICTION_INTERVAL_MS=1` + `CLIENT_POOL_SIZE=1`) — SASL credential source is below the cache layer

The `loginUserFromKeytab` swap actually mutates the process loginUser, which is the only thing that reliably steers Hadoop's deep SASL/JGSS paths. The single global lock keeps it safe under concurrency, and the critical section covers only metadata fetch (data scan is outside the lock).

### Additional pieces of the change

- `IcebergHiveCatalog`: `CLIENT_POOL_CACHE_KEYS=ugi` so Iceberg's `CachedClientPool` keys by current UGI (i.e. by catalog after the swap), giving each catalog its own `HiveClientPool`. After `loadCatalog` we also re-apply catalog properties to `delegate.getConf()` because `new HiveConf(parent, HiveCatalog.class)` re-reads the bundled `hive-site.xml` and drops the catalog-level `hive.metastore.kerberos.principal` override (resulting in `Server not found in Kerberos database` on the wrong SPN).
- `HiveMetaClient.callRPC`: same `doAsWithSwap` wrap so the Hive sub-connector (used by Unified catalog's hive side and direct `Hive` catalogs) is isolated identically.
- `HDFSCloudCredential.toThrift`: was previously empty; now emits `hadoop.security.authentication` / `hadoop.kerberos.principal` / `hadoop.kerberos.keytab` plus `fs.hdfs.impl` + `fs.hdfs.impl.disable.cache` into `cloud_properties`. The BE picks these up via `hdfsBuilderConfSetStr` and the patched `fs.hdfs.impl` makes the BE's libhdfs JVM authenticate with the catalog keytab too (the actual `CatalogUgiDistributedFileSystem` class is dropped into the BE's `hadoop/common` classpath separately — kept out of this PR's scope as it lives in deployment glue).

### Test setup

Two Iceberg/Unified catalogs against the same Kerberized HMS but with **different client keytabs** (so any leak of the client UGI shows up as an HDFS permission error against the table's warehouse path):

```sql
CREATE EXTERNAL CATALOG catalog_a PROPERTIES (
    "type" = "unified", "unified.metastore.type" = "hive",
    "hive.metastore.uris" = "thrift://hms.example:9083",
    "hive.metastore.kerberos.principal" = "hive/_HOST@EXAMPLE.COM",
    "hadoop.security.authentication" = "kerberos",
    "hadoop.kerberos.principal" = "user_a@EXAMPLE.COM",
    "hadoop.kerberos.keytab"    = "/path/to/user_a.keytab"
);
CREATE EXTERNAL CATALOG catalog_b PROPERTIES (
    -- same uris/SPN, different client keytab
    "hadoop.kerberos.principal" = "user_b@EXAMPLE.COM",
    "hadoop.kerberos.keytab"    = "/path/to/user_b.keytab"
);
```

`user_a` has read access on some tables that `user_b` does not, and vice versa, so a permission-denied error tells us exactly which client UGI authenticated.

### Verification (fresh cluster, `KRB5*` env unset)

| scenario                                                                       | before this commit                                              | after this commit                                                |
|--------------------------------------------------------------------------------|-----------------------------------------------------------------|------------------------------------------------------------------|
| `catalog_a` SELECT on a table only `user_a` can read                           | `Permission denied: user=user_b` (process used system UGI)      | ✅ rows returned                                                  |
| `catalog_b` SELECT on a table only `user_b` can read                           | ✅ (system UGI happened to match)                                | ✅                                                                |
| Sequential `a → b → a → b`, distinct tables (forces metadata + HDFS cache miss)| second call reuses first call's UGI → permission denied         | ✅ all 4 correct                                                  |
| Parallel 3 × catalog_a + 3 × catalog_b + 3 × catalog_c                         | first request's UGI wins for all                                | ✅ all 9 correct, each with its own keytab                        |
| BE HDFS data scan on a *new* table (cache-miss path on the BE side)            | BE reused system UGI, failed                                    | ✅ catalog UGI                                                    |
| Cluster restart + `KRB5PRINCIPAL`/`KRB5KEYTAB` env unset                       | breaks: process has no default UGI                              | ✅ works as long as every catalog declares its own keytab         |

### What I deliberately did **not** add

- No new BE C++ code. The BE side is solved by a small `fs.hdfs.impl` hook class loaded into the libhdfs JVM (`CatalogUgiDistributedFileSystem`) — that's deployment glue, not a code change in this repo. Open to moving it in if reviewers prefer.
- No per-catalog thread pool. The single `LOGIN_USER_LOCK` is intentional: the lock-protected critical section is only the metadata fetch (`loadTable` / single thrift RPC); BE data scan happens entirely outside it. Measured impact in our setup: a multi-second metadata-fetch query holds the lock for a few hundred ms; a light query starting right after completes essentially at its solo baseline.
- No fallback to the previous (incorrect) behavior when the catalog has no keytab — `doAsWithSwap` simply runs the action with the current loginUser, identical to the original code path. Fully backward compatible.

### Ablation summary

Each of the added lines was checked by removing it and observing which scenario breaks:

| removed                                              | breaks                                                              |
|------------------------------------------------------|---------------------------------------------------------------------|
| `getHMSUGI/getHDFSUGI` real impl (stub → null)        | the patched `HiveMetaStoreClient.open` / `FileSystem.createFileSystemInternal` call this; null means no UGI to doAs with → permission denied |
| `HDFSCloudCredential.toThrift` puts                   | BE on cache-miss path uses system UGI → permission denied           |
| `delegate.getConf` re-apply                           | `new HiveConf` re-reads `hive-site.xml`, drops catalog SPN → `Server not found in Kerberos database` |
| `CLIENT_POOL_CACHE_KEYS=ugi`                          | first call's `HiveClientPool` is reused across all catalogs         |
| any of the 6 read-method wraps                        | the corresponding `SHOW`/`SELECT` path bypasses the swap            |
| `doAsWithSwap` + `LOGIN_USER_LOCK`                    | the whole isolation collapses                                       |

kwontaeheon:
## Background: error logs that motivated this change

### 1) `Server not found in Kerberos database` — bundled hive-site.xml overrides the catalog SPN

```
javax.security.sasl.SaslException: GSS initiate failed
  at com.sun.security.sasl.gsskerb.GssKrb5Client.evaluateChallenge(...)
Caused by: org.ietf.jgss.GSSException: No valid credentials provided
  (Mechanism level: Server not found in Kerberos database (7) - LOOKING_UP_SERVER)
Caused by: sun.security.krb5.KrbException: Server not found in Kerberos database (7)
```

Root cause: the catalog declared a non-default `hive.metastore.kerberos.principal` (e.g. a dedicated service SPN for that HMS), but `org.apache.iceberg.hive.HiveCatalog` internally builds its own conf via `new HiveConf(parent, HiveCatalog.class)`, which re-reads the classpath `hive-site.xml` and **silently overwrites the catalog-level SPN** with the bundled default. The client then asks the KDC for a service ticket using the wrong SPN.

→ Fixed in `IcebergHiveCatalog` by re-applying catalog properties on `delegate.getConf()` after `loadCatalog`.

### 2) `Permission denied: user=<system_user>` — BE uses the process-wide system UGI, ignoring the catalog keytab

```
ERROR 1064 (HY000): hdfsOpenFile failed, backend=<be_host>,
  file=hdfs://<ns>/warehouse/<db>/<table>/data/<partition>/00001-...-snappy.parquet.
  err_msg: error=Permission denied,
  root_cause=RemoteException: Permission denied: user=<system_user>, access=EXECUTE,
  inode="<warehouse_path>":<owner>:<group>:drwx------
    at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.check(...)
    at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkPermission(...)
```

Root cause: the catalog requires `user_a` (its own keytab), but the BE's libhdfs JVM logged in once at startup from `KRB5*` env (`system_user`) and the catalog's keytab info never reached BE. HDFS namenode authenticates the client as `system_user`, which has no permission on the warehouse path.

→ Fixed by populating `HDFSCloudCredential.toThrift` to emit catalog kerberos creds + `fs.hdfs.impl=CatalogUgiDistributedFileSystem` into `cloud_properties`. BE picks these up via `hdfsBuilderConfSetStr` and the hook class wraps `super.initialize` in the catalog UGI's `doAs`.

### 3) Multi-catalog race — the second catalog's call authenticates with the **first** catalog's keytab

```
ERROR 1064: Failed to get iceberg table catalog_b.<db>.<table>.
  java.security.AccessControlException: Permission denied: user=user_a, access=EXECUTE,
  inode="<warehouse_path>":<owner>:<group>:drwx------
```

(Here `catalog_b` is declared with `user_b`'s keytab, but the call ended up authenticated as `user_a` — `user_a` being whichever catalog was queried first in this FE process.)

We added diagnostic logging inside the wrapper to make sure the thread-local doAs context was right:

```
[IcebergHiveCatalog.withCatalogUgi inside doAs]
  currentUser=user_b@EXAMPLE.COM (auth:KERBEROS)
```

The thread-local Subject was indeed `user_b`, yet the HMS-side error reported `user_a`. That's the smoking gun: Hadoop's HMS thrift client / SASL handshake doesn't read from the current thread's Subject — it reaches process-wide state (process loginUser, JGSS native credential cache, cached `HiveClientPool`). The first catalog's call wins the race and entrenches itself there.

→ Fixed by `HadoopExt.doAsWithSwap`, which swaps the **process loginUser** itself via `loginUserFromKeytab` under a single global lock, then `doAs` inside that swapped UGI.

---

## Verification (fresh cluster, `KRB5*` env unset)

### Single catalog

```
[1] catalog_a (user_a keytab):
    SELECT count(*) FROM catalog_a.<db>.<table_x> WHERE <partition_filter>;
    → <small number of rows>

[2] catalog_b (user_b keytab):
    SELECT count(*) FROM catalog_b.<db>.<table_y> LIMIT 1;
    → <large number of rows>

[3] catalog_c (user_c keytab, declared explicitly so no system default needed):
    SHOW DATABASES FROM catalog_c;
    → <list of dbs>

[4] catalog_a, a different table than [1] (forces BE data-cache miss path):
    SELECT count(*) FROM catalog_a.<db>.<table_z> WHERE <partition_filter>;
    → <small number of rows>
```

### Sequential multi-catalog (forces UGI flip across distinct keytabs)

```
SELECT count(*) FROM catalog_a.<...> WHERE <...>;
→ ✅

SELECT count(*) FROM catalog_b.<...> LIMIT 1;
→ ✅

SELECT count(*) FROM catalog_a.<...> WHERE <...>;
→ ✅

SELECT count(*) FROM catalog_b.<...> LIMIT 1;
→ ✅
```

All four calls authenticate with their declared keytab. Before this change, the second call's user always matched whichever catalog was hit first.

### Parallel 3×3 (catalog_a + catalog_b + catalog_c interleaved)

9 mysql client processes started concurrently, 3 against each catalog. After the change: every result is correct (each catalog's call hits HDFS as its own user); before, all 9 collapse to the first-arriving catalog's UGI.

### Concurrency impact — does a long query block other catalogs?

A 3-second metadata fetch against `catalog_b` was started; 0.5s later a light query against `catalog_a` was launched:

```
T=0.00s : HEAVY catalog_b starts (baseline ~3.7s)
T=0.64s : LIGHT catalog_a starts
T=1.71s : LIGHT done (1.07s — essentially solo baseline of 1.32s)
T=2.66s : HEAVY done
```

The light query finishes before the heavy one. The `LOGIN_USER_LOCK` is only held during metadata fetch (a couple hundred ms in practice); BE data scan happens entirely outside the lock.

### Build & restart reproducibility

```
$ docker compose down -v && docker compose up -d --build
$ # all patches re-applied automatically on image build
$ # process env no longer needs KRB5PRINCIPAL/KRB5KEYTAB at all
$ docker exec <fe> cat /proc/<fe_pid>/environ | tr '\0' '\n' | grep KRB5
$ # (no KRB5* in FE env — works because every catalog declares its own keytab)
```

This last bullet is the cleanest invariant of the change: with all catalogs declaring their own keytab, the FE/BE process no longer needs a system-wide Kerberos identity at all — each metadata call brings its own credentials.

kwontaeheon:
## Rebased onto current `main` — `toThrift` conflict resolved

Pushed `b3171727af` (force-with-lease). Only `HDFSCloudCredential.toThrift` had a real conflict — `main` independently introduced a generic pass-through version:

```java
@Override
public void toThrift(Map<String, String> properties) {
    if (hadoopConfiguration != null) {
        for (Map.Entry<String, String> entry : hadoopConfiguration.entrySet()) {
            if (!isReservedHadoopExtProperty(entry.getKey())) {
                properties.put(entry.getKey(), entry.getValue());
            }
        }
    }
}
private boolean isReservedHadoopExtProperty(String key) {
    return HadoopExt.HADOOP_CLOUD_CONFIGURATION_STRING.equals(key);
}
```

It looks like it might subsume this PR's explicit puts — but it cannot, because `hadoopConfiguration` on this code path does not contain the keys we need:

`HDFSCloudConfigurationProvider.preprocessProperties` explicitly removes the kerberos-related keys before constructing `hadoopConfiguration`:

```java
String[] keys = {
    HDFS_AUTHENTICATION,
    HDFS_USERNAME_DEPRECATED, HDFS_USERNAME,
    HDFS_PASSWORD_DEPRECATED, HDFS_PASSWORD,
    HDFS_KERBEROS_PRINCIPAL_DEPRECATED, HDFS_KERBEROS_PRINCIPAL,
    HDFS_KERBEROS_KEYTAB_DEPRECATED, HADOOP_KERBEROS_KEYTAB,
    HDFS_KERBEROS_KEYTAB_CONTENT_DEPRECATED, HADOOP_KERBEROS_KEYTAB_CONTENT
};
for (String k : keys) prop.remove(k);
```

After this, kerberos credentials live as typed fields on `HDFSCloudCredential` (`authentication` / `krbPrincipal` / `krbKeyTabFile`), not inside `hadoopConfiguration`. `main`'s loop iterates over a map that has already lost those keys — and the BE never sees them. Similarly, `fs.hdfs.impl=CatalogUgiDistributedFileSystem` + `fs.hdfs.impl.disable.cache=true` are never declared in user-facing `CREATE EXTERNAL CATALOG PROPERTIES`, so they aren't in `hadoopConfiguration` either.

### Scopes are orthogonal

| concern | `main`'s new code | this PR's adds |
|---|---|---|
| Catalog-level overrides for generic Hadoop conf (`dfs.*`, `fs.s3a.*`, …) reaching BE | ✅ handled | not its responsibility |
| Catalog-level kerberos creds (`hadoop.security.authentication`, principal, keytab) reaching BE | ❌ stripped in preprocess | ✅ explicitly put |
| Directing BE's libhdfs JVM to the catalog-aware FS hook (`fs.hdfs.impl` + `disable.cache`) | ❌ not in any input map | ✅ explicitly put |

### Final resolved version (kept both, `main`'s loop first then this PR's explicit puts)

```java
@Override
public void toThrift(Map<String, String> properties) {
    // Pass through generic hadoop configuration overrides from the catalog (from main).
    if (hadoopConfiguration != null) {
        for (Map.Entry<String, String> entry : hadoopConfiguration.entrySet()) {
            if (!isReservedHadoopExtProperty(entry.getKey())) {
                properties.put(entry.getKey(), entry.getValue());
            }
        }
    }
    // Per-catalog keytab isolation: kerberos credentials are stripped from
    // hadoopConfiguration by HDFSCloudConfigurationProvider.preprocessProperties and the
    // BE libhdfs FS hook is never declared in catalog PROPERTIES, so both must be put
    // here explicitly. Backward compatible: non-kerberos catalogs early-return.
    if (!KERBEROS_AUTH.equals(authentication)
            || krbPrincipal == null || krbPrincipal.isEmpty()
            || krbKeyTabFile == null || krbKeyTabFile.isEmpty()) {
        return;
    }
    properties.put(CloudConfigurationConstants.HDFS_AUTHENTICATION, authentication);
    properties.put(CloudConfigurationConstants.HDFS_KERBEROS_PRINCIPAL, krbPrincipal);
    properties.put(CloudConfigurationConstants.HADOOP_KERBEROS_KEYTAB, krbKeyTabFile);
    properties.put("fs.hdfs.impl", "com.starrocks.connector.hadoop.CatalogUgiDistributedFileSystem");
    properties.put("fs.hdfs.impl.disable.cache", "true");
}
```

This preserves both behaviors and is fully backward compatible with `main` (non-kerberos / no-keytab catalogs early-return, exactly the original code path).

The other three patched files (`HadoopExt.java`, `IcebergHiveCatalog.java`, `HiveMetaClient.java`) rebased cleanly with no conflict.

PR #72951, opened 2026-05-08
[Enhancement] Add diagnostic trace counters and slow-publish phase logs for PK-index publish path

luohaha:
## Why I'm doing:

The realtime-benchmark / PK-index publish path needs finer-grained observability than `cost=N us` alone. During the `auto-perf-opt my-rt-2` run, every actionable bottleneck localization required adding a trace counter or a phase split before the fix could be designed (and several wrong directions were prevented by counters that showed the assumption was false).

This PR consolidates the six instrumentation patches that the run produced. They are pure observability — log lines and `TRACE_COUNTER_INCREMENT` calls — with no behavior change on the steady-state path.

## What I'm doing:

Six small instrumentation commits, each independently scoped:

1. **`#72555` BRPC server queue time on slow publish_version** (`be/src/service/service_be/lake_service.cpp`). Records `cntl->latency_us()` (BRPC server-side queue time before handler dispatch) at handler entry, appends `brpc_queue_us=<n>` to the existing slow-publish log line. Used to discriminate FE-vs-BE source of slow `publish_rpc` cost.

2. **`#72557` Per-phase breakdown of slow `lake publishPartition`** (`fe/fe-core/.../PublishVersionDaemon.java`). Splits the slow-path log line into `executor_acquire_ms / lock_wait_ms / fe_prep_ms / brpc_ms`, only emitted when `total_ms ≥ 500 ms`. Used to attribute slow FE publishPartition events to a specific sub-phase.

3. **`#72577` `Table::MultiGet` t3 sub-phase counters** (`be/src/storage/sstable/table.cpp`). Splits `multiget_t3_us` into `t3_filter_us / t3_block_read_us / t3_block_read_miss_us / t3_search_us` plus a cache-miss count (via `ReadIOStat::block_cnt_from_file`). Used to classify cold-start `multi_get` cost as RTT-bound vs bandwidth-bound vs filter-CPU-bound vs in-memory-search-bound — the iter-29/30 instrumentation that prevented an iter-15-style prefetch trap.

4. **`#72671` `pindex_sstables_queried_cnt` trace counter** (`be/src/storage/lake/persistent_index_sstable_fileset.cpp`). Counts how many sstables a `multi_get` actually queried (1 for standalone fileset, `sstable_key_indexes_map.size()` for sharded). Compared against `pindex_init_sst_cnt` (eager opens) to assess whether a lazy-SST-open refactor would skip real work.

5. **`#72672` `pindex_unique_sstables_touched_cnt` trace counter** (`be/src/storage/lake/lake_persistent_index.cpp` + `persistent_index_sstable_fileset.{h,cpp}`). Dedup variant of (4): tracks the *unique* set of sstables touched across all filesets in one `get_from_sstables` call. Settled the lazy-open question — the data showed lazy-open had zero benefit because cold-start queries hit nearly every SST, saving the team from shipping a 100-line refactor that would not have helped.

6. **`#72703` `multi_get` miss-run + byte-contig pairs counters** (`be/src/storage/sstable/table.cpp`). Tracks consecutive cache-miss block runs and byte-contiguous miss pairs in `Table::MultiGet`. Used to evaluate whether merging adjacent OSS reads into one larger read would help (it would, in some cold-start subcases).

Net diff: 6 files, +116 / -9 LOC.

Validated standalone in earlier PRs #72555 / #72557 / #72577 / #72671 / #72672 / #72703 (all against `branch-4.1`). This PR rebases the lineage to `main` for unified review. The conflict on `lake_service.cpp` (upstream removed `lake_enable_publish_version_trace_log`) and on `lake_persistent_index.cpp` (the parallel-walk optimization that originally landed alongside `pindex_unique_sstables_touched_cnt` is not part of this instrumentation-only PR) was resolved keeping only the diagnostic-counter portion.

### Where these counters surface

In trace data parseable from rotated `be.INFO.log.*` files via the `publish-trace` skill (or directly grepping for the counter names), and in the existing slow-publish log lines that already exist in the codebase.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Pure observability addition. The new log content only fires inside paths that are already classified slow (≥ thresholds that already exist in main). The new `TRACE_COUNTER_INCREMENT` calls are no-ops unless trace collection is enabled by the existing publish-trace mechanism.

## Checklist:

- [ ] This is a backport pr

No new tests added: instrumentation has no functional behavior to assert. Manual validation: each counter / log was used in at least one `auto-perf-opt my-rt-2` iteration to drive (or rule out) a code change, with the trace data preserved in the run's iteration directories.

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
@@ -1016,8 +1016,11 @@ public class PublishVersionDaemon extends LeaderDaemon {
             return CompletableFuture.completedFuture(false);
         }
 
+        long submitTimeMs = System.currentTimeMillis();
         return CompletableFuture.supplyAsync(() -> {
-            boolean success = publishPartition(db, tableCommitInfo, partitionCommitInfo, txnState);
+            long lambdaEntryMs = System.currentTimeMillis();
+            boolean success = publishPartition(db, tableCommitInfo, partitionCommitInfo, txnState,
+                    submitTimeMs, lambdaEntryMs);
             partitionCommitInfo.setVersionTime(success ? System.currentTimeMillis() : -System.currentTimeMillis());
             return success;
         }, getTaskExecutor()).exceptionally(ex -> {
@@ -1029,7 +1032,8 @@ public class PublishVersionDaemon extends LeaderDaemon {
 
     private boolean publishPartition(@NotNull Database db, @NotNull TableCommitInfo tableCommitInfo,
                                      @NotNull PartitionCommitInfo partitionCommitInfo,
-                                     @NotNull TransactionState txnState) {
+                                     @NotNull TransactionState txnState,
+                                     long submitTimeMs, long lambdaEntryMs) {
         long tableId = tableCommitInfo.getTableId();
         long baseVersion = 0;
         long txnVersion = partitionCommitInfo.getVersion();
@@ -1042,6 +1046,7 @@ public class PublishVersionDaemon extends LeaderDaemon {
 
         Locker locker = new Locker();
         locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(tableId), LockType.READ);
+        long lockAcquiredMs = System.currentTimeMillis();
         boolean useAggregatePublish = Config.enable_file_bundling;
         try {
             OlapTable table =
@@ -1082,6 +1087,7 @@ public class PublishVersionDaemon extends LeaderDaemon {
         }
 
         TxnInfoPB txnInfo = TxnInfoHelper.fromTransactionState(txnState);
+        long brpcStartMs = System.currentTimeMillis();
         try {
             if (CollectionUtils.isNotEmpty(shadowTablets)) {
                 Utils.publishLogVersion(shadowTablets, txnInfo, txnVersion, computeResource);
@@ -1114,6 +1120,16 @@ public class PublishVersionDaemon extends LeaderDaemon {
             LOG.error("Fail to publish partition {} of txn {}: {}", partitionCommitInfo.getPhysicalPartitionId(),
                     txnId, e.getMessage());
             return false;
+        } finally {
+            long brpcEndMs = System.currentTimeMillis();
+            long totalMs = brpcEndMs - submitTimeMs;
+            if (totalMs >= 500) {
+                LOG.warn("Slow publishPartition txn={} partition={} table={} total_ms={} executor_acquire_ms={} " +
+                                "lock_wait_ms={} fe_prep_ms={} brpc_ms={}",
+                        txnId, partitionCommitInfo.getPhysicalPartitionId(), tableId,
+                        totalMs, lambdaEntryMs - submitTimeMs,
+                        lockAcquiredMs - lambdaEntryMs, brpcStartMs - lockAcquiredMs, brpcEndMs - brpcStartMs);
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
```diff
@@ -1114,6 +1120,16 @@ private boolean publishPartition(@NotNull Database db, @NotNull TableCommitInfo
             LOG.error("Fail to publish partition {} of txn {}: {}", partitionCommitInfo.getPhysicalPartitionId(),
                     txnId, e.getMessage());
             return false;
+        } finally {
+            long rpcEndMs = System.currentTimeMillis();
+            long totalMs = rpcEndMs - submitTimeMs;
+            if (totalMs >= 3000) {
+                LOG.warn("Slow publishPartition txn={} partition={} table={} total_ms={} executor_acquire_ms={} " +
```
starrocks-xupeng: if i don't read code, just by reading the log, i dont' understand `executor_acquire_ms` and `lock_wait_ms` and `fe_prep_ms` and `rpc_ms`

starrocks-xupeng: and for each transaction, in `finishTransaction`, we will print profile about publish phase, what's the relationship between them

luohaha: Thanks — addressed in `a36caae`.

**On the field names** — reworked into an additive form so the log shows the decomposition itself:

```
Slow publishPartition: txn=X partition=Y table=Z, total=Tms
  (executor_queue=Ams + db_lock_wait=Bms + fe_prep=Cms + rpc=Dms)
```

Where:
- `executor_queue` — `CompletableFuture.supplyAsync` submit → lambda actually entering the executor thread (publish-executor scheduling delay / queue depth).
- `db_lock_wait` — lambda entry → `lockTablesWithIntensiveDbLock` acquired (db read-lock contention).
- `fe_prep` — lock acquired → first BRPC issued. Pure FE work: `OlapTable` lookup, walking partitions to collect normal/shadow tablets, `TxnInfoPB` construction.
- `rpc` — wall time of `Utils.publishLogVersion` + `Utils.publishLakePartition` (BRPC round-trip; on the BE side this further includes the server-side queue captured by `brpc_queue_us` added in commit `70b0b22`).

By construction `total = executor_queue + db_lock_wait + fe_prep + rpc`. Same explanation is also inlined as a code comment for future readers.

**On the relationship to the existing `finishTransaction` summary** (the `prepare time / commit time / wait for publish cost / publish rpc cost / publish total cost / total cost` line in `TransactionState.toString`):

- The txn-level line is **wall-clock across all of a txn's `publishPartition` invocations running in parallel**, so its `publish rpc cost` field can't tell you *which* partition was slow, or *which phase* of `publishPartition` dominated.
- This new line is **per-partition** and **only fires on outliers** (`total_ms ≥ 3000`). It exists specifically for the case where the txn-level `publish rpc cost` shows a big number and we need to attribute it.

So the two are complementary: txn-level summary always prints; this fires only when at least one partition crosses the slow threshold.

xiangguangyxg (approved):

wyb (approved):

PR #72955, opened 2026-05-08
[Enhancement] Promote variant virtual columns to typed-value readers for dict-filter participation

dirtysalt:
## Why I'm doing:

Variant shredded columns read significantly slower than struct columns for the same dataset. Profiling showed a key asymmetry: struct subfield predicates (`commit.collection IN (...)`) are evaluated as dict-code integer comparisons in Phase 2, eliminating rows before string materialization. For variant virtual columns, all conjuncts were unconditionally deferred to Phase 4 after full VariantColumn reconstruction, and even simple projections went through the expensive Phase 3 `_read_range_skip_base_payload` → Phase 4 `project_variant_leaf_column` path.

## What I'm doing:

**Variant Virtual Column Promotion (Phase 2 bypass)**

When a variant's hidden source has `_skip_base_payload = true` (all requested shredded paths are scalar `typed_value` leaves), AND each path satisfies two safety conditions (see below), promote ALL virtual slots to `VariantTypedValueProxy` readers in `_column_readers`. This lets them participate in Phase 2 like normal physical columns — including dict filter — and skips Phase 3 hidden-source reads entirely.

A `VariantTypedValueProxy` is a thin non-owning wrapper around the leaf `ScalarColumnReader` from `ShreddedFieldNode::typed_value_reader`. It delegates `read_range`, `fill_dst_column`, `try_to_use_dict_filter`, `rewrite_conjunct_ctxs_to_predicate`, and `filter_dict_column` directly to the underlying leaf reader, enabling full dict-filter participation without materializing the VARIANT column first.

**Two safety conditions for promotion per path:**

1. **Fallback must be all-null for this row group** (`fallback_values_all_null_in_row_group_for_path`). If the `value` (base payload) column has any non-null entries, `_read_range_skip_base_payload`'s runtime detection would fall back to base-payload reconstruction — which promotion bypasses entirely, producing all-NULLs. Example: `department.typed_value` is all-null in the test file (data lives in `department.value`).

2. **Type must be directly write-compatible** between the typed_value leaf's `read_type` and the target slot type. Variable-length types (VARCHAR/VARBINARY/CHAR/BINARY) share the same `BinaryColumn` layout and are mutually safe. Fixed-size types must match exactly — a mismatch (e.g. INT32 `typed_value` → BIGINT slot) causes `PlainDecoder<int>` to write 4-byte elements into an 8-byte column, crashing with `length % sizeof(ValueType) == 0` assertion failure.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/formats/parquet/complex_column_reader.cpp
+++ b/be/src/formats/parquet/complex_column_reader.cpp
@@ -876,6 +876,9 @@ static bool _requested_scalar_fallback_values_all_null(const std::vector<Shredde
                                                        const std::vector<VariantPath>& requested_paths) {
     for (const auto& path : requested_paths) {
         const ShreddedFieldNode* node = find_shredded_field_node_for_path(nodes, path);
+        if (node == nullptr) {
+            return false;
+        }
         if (node->value_reader != nullptr && !_column_chunk_all_null(node->value_reader.get())) {
             return false;
         }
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

Youngwb (approved):

stephen-shelby (approved):

trueeyu (approved):

PR #72956, opened 2026-05-08
[Enhancement] Eliminate HTTP-server accept thundering-herd via SO_REUSEPORT

luohaha:
## Why I'm doing:

The BE HTTP server (`EvHttpServer`) starts N libevent worker threads (`be_http_num_workers`, default 1000) that all call `evhttp_accept_socket` on the **same** `_server_fd`. Without `EPOLLEXCLUSIVE`, every incoming SYN wakes all N threads and the kernel serializes them on the listening socket spinlock. This is a textbook accept thundering herd.

`perf` sampling on a busy CN (60 s, 95 K samples) showed `[kernel] native_queued_spin_lock_slowpath` at **7.5 % self-time**, reached entirely from `event_base_loop → listener_read_cb → accept4 → inet_csk_accept → lock_sock_nested`. Every concurrent stream-load — typical for realtime ingestion on a multi-table workload — pays the spinlock cost.

## What I'm doing:

Adds an `enable_http_server_so_reuseport` config (default `true`) so each libevent worker creates its own `SO_REUSEPORT` listener bound to the same address. The kernel then hash-balances accepts across listeners in-kernel, eliminating the spinlock contention.

- Worker 0 keeps `_server_fd` (the legacy single-listener path) so the existing bind-and-port-resolution flow is unchanged.
- Workers 1..N-1 each call a new `listen_with_reuseport(point)` helper that creates a TCP listener, sets `SO_REUSEADDR + SO_REUSEPORT`, binds to the same address, and listens. Each listener is non-blocking and tracked in `_worker_fds` for shutdown.
- Any `setsockopt` / `bind` failure logs a warning and falls back to the legacy shared-fd path for that worker — no behavior change in environments without SO_REUSEPORT (e.g. macOS dev builds).
- Net diff: 3 files, +134 / -2 LOC, contained to `be/src/http/ev_http_server.{cpp,h}` and one config flag in `be/src/common/config.h`.

Validated standalone in PR #72554 (against `branch-4.1`); this PR rebases the same change to `main` for review against the latest tree.

### Empirical impact

Validated on the `999_1gb_1_1tb` realtime-benchmark template (1 FE + 6 BE shared-data, S3 backend) via `auto-perf-opt`:

| Metric | Before (no SO_REUSEPORT) | After (SO_REUSEPORT on) | Δ |
|---|---:|---:|---:|
| `native_queued_spin_lock_slowpath` self-time | 7.50 % | < 0.5 % | -94 % |
| Upsert max latency | 12 200 ms | **2 851 ms** | -77 % |
| Upsert P99 | 1 171 ms | 629 ms | -46 % |
| Throughput | 760 K rows/s | 917 K rows/s | +21 % |
| Ingest error rate | 2.24 % | 0.00 % | guardrail restored |

The released ~7.5 % CPU/CN goes back to legitimate compaction + publish work. As a side effect, the cluster also became resilient to FE phantom-retry storms (storm continues at ~100/s/CN but no longer breaks ingest, because workers no longer block on the listening-socket spinlock).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

The new config defaults to `true`. Each worker thread gets its own kernel listener; the workload-level effect is "fewer thundering-herd wakeups" — externally indistinguishable from the legacy path except via `perf`. Any environment that does not support `SO_REUSEPORT` (macOS dev builds, kernels lacking the feature) falls back automatically.

## Checklist:

- [ ] This is a backport pr

No new tests added: the change is in the BE HTTP server bootstrap path, observable only by kernel-side spinlock metrics under sustained concurrent connect load. Existing HTTP integration tests cover correctness on each worker (a connection accepted via the SO_REUSEPORT listener is functionally identical to one accepted via the shared fd).

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/http/ev_http_server.cpp
+++ b/be/src/http/ev_http_server.cpp
@@ -54,7 +54,6 @@
 
 #include "base/brpc/brpc.h"
 #include "base/system/errno.h"
-#include "common/config.h"
 #include "common/config_ingest_fwd.h"
 #include "common/logging.h"
 #include "common/system/backend_options.h"
--- a/be/src/http/ev_http_server.cpp
+++ b/be/src/http/ev_http_server.cpp
@@ -201,17 +201,18 @@ Status EvHttpServer::start() {
 
     // Resolve the bind endpoint once, on the main thread, for per-worker
     // SO_REUSEPORT listeners. _bind() has already populated _real_port (incl.
-    // the port==0 auto-assign case).
+    // the port==0 auto-assign case). On platforms without SO_REUSEPORT or
+    // when endpoint resolution fails, every worker silently falls back to
+    // the shared-fd accept path (no functional change vs the legacy
+    // behaviour).
     butil::EndPoint listen_point;
     bool reuseport_enabled = false;
 #ifdef SO_REUSEPORT
-    if (config::enable_http_server_so_reuseport) {
-        if (butil::str2endpoint(_host.c_str(), _real_port, &listen_point) == 0) {
-            reuseport_enabled = true;
-        } else {
-            LOG(WARNING) << "Failed to resolve listen endpoint " << _host << ":" << _real_port
-                         << " for SO_REUSEPORT mode; falling back to shared-fd accept";
-        }
+    if (butil::str2endpoint(_host.c_str(), _real_port, &listen_point) == 0) {
+        reuseport_enabled = true;
+    } else {
+        LOG(WARNING) << "Failed to resolve listen endpoint " << _host << ":" << _real_port
+                     << " for SO_REUSEPORT mode; falling back to shared-fd accept";
     }
 #endif
 
@@ -295,7 +296,17 @@ void EvHttpServer::stop() {
     // Per-worker SO_REUSEPORT listeners need their own shutdown, otherwise
     // their event_base_dispatch will not unblock from epoll_wait and join()
     // will hang.
-    for (int fd : _worker_fds) {
+    //
+    // Snapshot _worker_fds under the lock so we don't race with workers that
+    // are still finishing setup in start() and appending their fd. Any fd
+    // published after this snapshot belongs to a worker whose event_base has
+    // not yet entered dispatch — its own dispatch loop will see the loopbreak
+    // set above and exit immediately, so a missed shutdown there is harmless.
+    std::vector<int> worker_fds_snapshot;
+    pthread_rwlock_rdlock(&_rw_lock);
+    worker_fds_snapshot = _worker_fds;
+    pthread_rwlock_unlock(&_rw_lock);
+    for (int fd : worker_fds_snapshot) {
         ::shutdown(fd, SHUT_RDWR);
     }
 }
@@ -355,8 +366,10 @@ Status EvHttpServer::_bind() {
     // listener to the same {addr, real_port}. The kernel only permits multiple
     // listening sockets on the same address when *all* of them carry
     // SO_REUSEPORT — without setting it on the primary fd, every worker's
-    // bind() would fail with EADDRINUSE.
-    if (config::enable_http_server_so_reuseport) {
+    // bind() would fail with EADDRINUSE. setsockopt failure here is benign:
+    // workers will detect the mismatch and individually fall back to sharing
+    // _server_fd.
+    {
         int yes = 1;
         if (::setsockopt(_server_fd, SOL_SOCKET, SO_REUSEPORT, &yes, sizeof(yes)) != 0) {
             LOG(WARNING) << "Failed to set SO_REUSEPORT on primary listener; per-worker listen sockets "
--- a/be/src/http/ev_http_server.cpp
+++ b/be/src/http/ev_http_server.cpp
@@ -127,36 +127,50 @@ static int on_connection(struct evhttp_request* req, void* param) {
     return 0;
 }
 
-// Create a TCP listening socket with SO_REUSEPORT, bound to `point`. Returns
-// -1 on any failure (and `errno` is set). When SO_REUSEPORT is set on every
-// listener for the same {addr, port}, the kernel load-balances new
-// connections across them — workers no longer thundering-herd on a single
-// shared listen fd's accept queue lock.
+// Open a TCP listener bound to `point`, optionally placed in the kernel's
+// SO_REUSEPORT group. SO_REUSEPORT must be set on every participating fd
+// *before* bind() — Linux's man socket(7) is explicit, and the kernel only
+// inserts the fd into the reuseport hash via inet_reuseport_add_sock() at
+// bind() time. setsockopt() after bind() merely flips sk_reuseport without
+// joining the group, so the next bind() to the same port from another fd
+// gets EADDRINUSE.
 //
-// Delegates the bind+listen to `butil::tcp_listen()` (the same helper used
-// by `_bind()` for the primary `_server_fd`); we only have to set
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

alvin-phoenix-ai:
@luohaha 
Do you think it is necessary to have enable_http_server_so_reuseport? 
It should be true all the time

be/src/http/ev_http_server.cpp
```diff
@@ -45,12 +45,18 @@
...
+#include <unistd.h>
+
+#include <cstring>
 #include <memory>
 #include <sstream>
 #include <utility>
 
 #include "base/brpc/brpc.h"
 #include "base/system/errno.h"
+#include "common/config.h"
```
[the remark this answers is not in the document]
alvin-celerdata: It will cause this file to be recompiled every time common/config.h is changed, which will cause compilation to slow down.

kevincai (commented):
how about a dedicate thread doing the accept and then dispatch the task in shared queue so every workers can consume the task, following the single-producer-multiple-consumer model. This is the similar pattern netty NioEventLoopGroup used as BossGroup and WorkerGroup.

luohaha:
@kevincai thanks — the boss/worker (Netty NioEventLoopGroup) model is a valid alternative and we considered it. A quick note on the trade-off we made:

**Both approaches eliminate the same root cause.** The thundering-herd contention in `inet_csk_accept` exists because all N libevent threads share one listen fd and one accept queue lock. Boss/worker fixes this by having only one thread (the boss) call `accept()`; SO_REUSEPORT fixes it by giving each thread its own listen fd, and the kernel hash-distributes incoming SYNs across the per-fd accept queues with no shared lock.

**Comparison:**

| Aspect | SO_REUSEPORT (this PR) | Boss/Worker |
|---|---|---|
| Code change | +134 / -2 LOC, contained to `ev_http_server.{h,cpp}` | 1000+ LOC — new accept loop, lock-free MPMC queue, worker pull logic, lifecycle/shutdown rewrite |
| Throughput at high connection rate | Kernel-side hash, scales linearly with workers | Single boss thread can become a bottleneck; needs SO_REUSEPORT itself or `EPOLLEXCLUSIVE` to scale further |
| Latency variance | Low — accept happens on the worker that will service the request, no cross-thread handoff | Higher — every connection takes one extra thread hop (boss → queue → worker) |
| Failure isolation | A worker death loses 1/N capacity until restart | Boss thread death stops all accept |
| Linux kernel support | Required (3.9+); falls back to shared-fd path on older kernels and on macOS dev builds | None |

The minimal-change path was attractive given the validated 7.5 % → < 0.5 % spinlock self-time delta and the fact that StarRocks already targets recent Linux kernels in production. If a future workload hits a SO_REUSEPORT-specific corner case (e.g. uneven kernel hashing under load skew), the Boss/Worker direction stays open as a follow-up — happy to pick it up if/when that signal appears.

luohaha:
> @luohaha Do you think it is necessary to have enable_http_server_so_reuseport? It should be true all the time

@alvin-celerdata Done.

be/src/http/ev_http_server.cpp
```diff
@@ -123,6 +129,54 @@ static int on_connection(struct evhttp_request* req, void* param) {
     return 0;
 }
 
+// Create a TCP listening socket with SO_REUSEADDR + SO_REUSEPORT, bound to
+// `point`. Returns -1 on any failure (and `errno` is set). When SO_REUSEPORT
+// is set on every listener for the same {addr, port}, the kernel load-balances
+// new connections across them — workers no longer thundering-herd on a single
+// shared listen fd's accept queue lock.
+static int listen_with_reuseport(const butil::EndPoint& point) {
+#ifndef SO_REUSEPORT
```
xiangguangyxg: Why not use a config

be/src/http/ev_http_server.cpp
```diff
@@ -123,6 +129,54 @@ static int on_connection(struct evhttp_request* req, void* param) {
...
+    errno = ENOTSUP;
+    return -1;
+#else
+    int fd = ::socket(AF_INET, SOCK_STREAM, 0);
```
kevincai: so this detailed implemention can be delegated to `butil::tcp_listen()` with either `FLAGS_reuse_port =  true` or set reuse sockopts explicitly to the returned fd.

luohaha: Good catch, addressed in [10c7b003d5](https://github.com/StarRocks/starrocks/pull/72956/commits/10c7b003d5) — `listen_with_reuseport` now delegates to `butil::tcp_listen()` and only sets `SO_REUSEPORT` on the returned fd. -28 lines vs the manual socket/setsockopt/bind/listen sequence.

be/src/http/ev_http_server.cpp
luohaha: We don't need a config, and `SO_REUSEPORT` is a macro in <sys/socket.h>, if `SO_REUSEPORT` no exist, that means we can't enable this feature

be/src/http/ev_http_server.cpp
luohaha: done

kevincai (dismissed):

be/src/http/ev_http_server.cpp
luohaha: Thanks for the suggestions — I tried both before landing on the explicit helper. Sharing what I hit:

**(1) Set `SO_REUSEPORT` on the fd returned by `butil::tcp_listen()`.** This is exactly what the previous revision of this PR did (commit 9c19620). It compiles and looks right, but the BE UT (`EvHttpServerTest.port_auto_assignment_distinct`) timed out, with the listener spamming `[warn] Error from accept() call: Invalid argument` and a `Worker 1: SO_REUSEPORT listen failed: Address already in use` right before the flood. Root cause: Linux man `socket(7)` requires `SO_REUSEPORT` to be set on every participating fd *before* `bind()` — the kernel only inserts the fd into the reuseport hash via `inet_reuseport_add_sock()` at bind() time. setsockopt() after bind() merely flips `sk_reuseport` locally without joining the group, so the secondary worker's `bind(real_port)` still returns `EADDRINUSE`, every worker silently falls back to the shared fd, and the optimisation is effectively disabled. That's the failure I'm trying to fix in this revision.

**(2) `FLAGS_reuse_port = true` + `butil::tcp_listen()`.** I went to wire it up and stopped because `FLAGS_reuse_port` is a *process-global* gflag — flipping it makes every subsequent `butil::tcp_listen()` call in the BE opt into SO_REUSEPORT, including brpc's service listener and any other helper that ends up in this codepath. We only want fan-out on the be_http_port listener; rewiring brpc's port silently is a much bigger blast radius than this PR's intent. Toggling the flag around just the http call doesn't help either since other threads can be calling `butil::tcp_listen()` concurrently.

So I ended up writing the small `open_listen_socket(point, reuse_port)` helper — it sets the option *before* `bind()` (giving us the kernel-side group membership we need), stays scoped to this single listener, and mirrors `butil::tcp_listen()`'s `SO_REUSEADDR` + `listen(65535)` defaults so the primary-listener behaviour is unchanged. Open to alternatives if there's a way to delegate without taking either of those two hits.

kevincai (approved):

xiangguangyxg (approved):

PR #72986, opened 2026-05-08
[Enhancement] Parallelize delete-file reads in LakePersistentIndex::load_dels

luohaha:
## Why I'm doing:

Cold-start `primary_index_load_latency_us` was measured at 14-18 s on slow tablets in the `auto-perf-opt my-rt-2` run (1 FE + 6 BE shared-data, S3 backend, `999_1gb_1_1tb` template). BE Published trace decomposition:

- `pindex_init_sst_open_us`: 8.26 s — already parallel via `_open_sstables_parallel`.
- `rebuild_index_del_cost_us`: 9.81 s — **the only remaining sequential I/O hot path** in `primary_index_load`. Each tablet's del files (typically 1-8 of them, mean ~3) were being read from OSS one at a time on a single worker thread.

This PR parallelizes the OSS reads + protobuf deserialization across `pk_index_execution_thread_pool`, while keeping `get / filter / replay_erase` on the original sequential path so existing erase-ordering semantics are preserved.

## What I'm doing:

Two-phase `load_dels` in `be/src/storage/lake/lake_persistent_index.cpp`:

1. **Phase 1 (parallel):** dispatch one task per del file to `pk_index_execution_thread_pool`. Each task fetches the `DelDataPB` from OSS and parses it. A barrier collects all parsed protos into an in-order vector.
2. **Phase 2 (sequential, in original order):** the existing `index.get(...)` / filter / `replay_erase` loop runs over the prepared protos. Erase ordering is unchanged.

Gated by the existing `enable_pk_index_parallel_execution` config (default `true`) **and** by the new `pk_index_parallel_load_dels_mem_ratio` config (default `50`, range 0-100): when the update mem tracker is already past that percent of its limit, the function falls back to a true single-pass loop that only holds one decoded del-file column at a time. Avoids the peak-memory regression that the two-phase prefetch would otherwise introduce when the optimization isn't worth its memory cost. Set the ratio to `100` to disable the memory gate entirely.

Net diff: 4 files (1 code + 1 BE config + 2 docs), ≈ +150 / -50 LOC across the three commits.

Validated standalone in PR #72574 (against `branch-4.1`); this PR rebases the same change to `main`.

### Empirical impact

Measured on the `999_1gb_1_1tb` realtime-benchmark template, fresh-deploy A/B (cold-start storm scenario):

| Metric | iter-027 baseline | iter-028 (with this PR) |
|---|---:|---:|
| Max `rebuild_index_del_cost_us` | 9.81 s | 9.04 s (-8 %) |
| Slow ≥ 1 s `rebuild_del` p99 | n/a | 4.99 s |
| Slow ≥ 1 s `rebuild_del` p95 | n/a | 3.35 s |
| Slow ≥ 1 s event count | 1,891 | 1,882 (≈ unchanged) |
| Steady-state P99 | unchanged | unchanged |
| Error rate | 0 % | 0 % |

**Honest assessment:** the gain is partial (8 %) rather than the 4-6× we hoped for. Two reasons surfaced after deploy:

1. **174 / 675** slow `rebuild_del > 0` events have only **1 del file** — parallelism cannot help these.
2. Most of the rest have 2-5 files; wall-clock is bounded by the *slowest* file's tail latency (~1.5-2 s OSS first-byte), not by total bytes — so going from "8 files serial" to "8 files parallel" still pays one tail.

Plus, `pindex_init_sst_open_us` (~8-11 s) runs sequentially **before** `rebuild_del` inside `primary_index_load`, so even an instant `rebuild_del` couldn't push primary_index_load below 8 s.

The PR is still a no-regression net win on multi-del-file events and unblocks subsequent cold-start chain optimizations (the iter-29 sub-counter instrumentation that came next was triggered by the residual `rebuild_del` distribution this PR exposed).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Pure performance optimization. The two-phase split preserves the original erase-ordering semantics: Phase 2 iterates over the parsed protos in input order, exactly as the legacy single loop did.

## Checklist:

- [ ] This is a backport pr

Added `LakePersistentIndexTest.test_load_dels_parallel_propagates_io_error` — failure-injection guard for the parallel path: with N=3 del files where every read fails (files do not exist on disk), the parallel phase must propagate the error via `shared_status` and return cleanly instead of crashing or proceeding to Phase 2 with default-constructed `pkc` slots. This is the regression surface added by the parallel refactor that existing publish-version tests do not cover. Existing `LakePersistentIndex` / publish-version tests continue to cover the per-shape correctness (0 / 1 / many del files).

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

Validation was done on `branch-4.1`. Backport to 4.0 / 3.5 / 3.4 is plausible (the call site predates 4.0) but the perf delta has not been re-measured on those branches.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -482,6 +482,11 @@ CONF_mInt64(pk_index_parallel_execution_min_rows, "16384");
 CONF_mInt32(pk_index_parallel_execution_threadpool_max_threads, "0");
 // The queue size for pk index parallel get threadpool in shared-data mode.
 CONF_mInt32(pk_index_parallel_execution_threadpool_size, "1048576");
+// Skip the parallel two-phase prefetch in LakePersistentIndex::load_dels when the update
+// mem tracker is already past this percent (0-100) of its limit. In that regime the function
+// falls back to a single-pass loop that holds only one decoded del-file column at a time,
+// trading the cold-start latency win for bounded peak memory.
+CONF_mInt32(pk_index_parallel_load_dels_mem_ratio, "50");
 // Memtable flush threadpool max thread num for pk index in shared-data mode.
 CONF_mInt32(pk_index_memtable_flush_threadpool_max_threads, "0");
 // The queue size for pk index memtable flush threadpool in shared-data mode.
--- a/be/src/common/config_fwd_headers_manifest.json
+++ b/be/src/common/config_fwd_headers_manifest.json
@@ -720,6 +720,7 @@
         "pk_index_early_sst_compaction_threshold",
         "enable_pk_index_parallel_execution",
         "pk_index_parallel_execution_min_rows",
+        "pk_index_parallel_load_dels_mem_ratio",
         "pk_index_memtable_max_count",
         "pk_index_memtable_max_wait_flush_timeout_ms",
         "pk_index_size_tiered_min_level_size",
--- a/be/src/common/config_primary_key_fwd.h
+++ b/be/src/common/config_primary_key_fwd.h
@@ -74,6 +74,12 @@ CONF_mBool(enable_pk_index_parallel_execution, "true");
 // The minimum rows threshold to enable parallel get for primary key index in shared-data mode.
 CONF_mInt64(pk_index_parallel_execution_min_rows, "16384");
 
+// Skip the parallel two-phase prefetch in LakePersistentIndex::load_dels when the update
+// mem tracker is already past this percent (0-100) of its limit. In that regime the function
+// falls back to a single-pass loop that holds only one decoded del-file column at a time,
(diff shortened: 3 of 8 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg (dismissed):

kevincai (approved):

xiangguangyxg (approved):

PR #72993, opened 2026-05-08
[Tool] disable auto-update of unstable_case.json in weekly review

huailiu1122:
## What type of PR is this?

> Tool

## Why are we doing this? Any context or related work?

The weekly review run found issues that make the auto-removal of cases unreliable:

1. All 3 FE-UT runs reported `BUILD FAILURE` — XML output may be incomplete
2. 7 class-level entries in `unstable_case.json` (no `#method`) are silently classified as `not_found` and skipped, but were incorrectly removed by the auto-commit
3. Report does not yet show per-case run counts, making results hard to verify

Disabling the auto-commit step (`if: false`) until the review logic is confirmed correct. The report and XML upload steps are unaffected — results will still be generated and visible.

Re-enable by restoring `if: steps.gen.outputs.JSON_UPDATED == 'true'`.

## What changes were made?

- `.github/workflows/weekly-unstable-case-review.yml`: set `Commit updated unstable_case.json` step to `if: false`

## Checklist

- [x] I have self-reviewed this PR

### DIFF
--- a/.github/workflows/weekly-unstable-case-review.yml
+++ b/.github/workflows/weekly-unstable-case-review.yml
@@ -93,7 +93,7 @@ jobs:
             --linuxdistro ubuntu
 
       - name: Clean ECI
-        if: always()
+        if: always() && steps.run_ut.outputs.ECI_ID != ''
         run: eci rm ${{ steps.run_ut.outputs.ECI_ID }} || true
 
       - name: Clean ENV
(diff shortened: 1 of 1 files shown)


huailiu1122:
Closing: branch contained unrelated commit (e2cf635b). Replaced by clean PR.

PR #72999, opened 2026-05-08
[Enhancement] Add adaptive ef_search for HNSW vector index queries

sevev:
## Why I'm doing:

In shared-data (lake) mode, compaction merges segments more aggressively than shared-nothing, producing fewer/larger segments. Empirically, recall on HNSW vector index drops noticeably at the same query parameters as segment row count grows — HNSW search quality scales as roughly `O(ef × log N)`, so a fixed `ef_search` cannot cover enough of the graph as `N` grows.

Benchmark also confirms only `ef_search` (not `k_factor` / candidate pool size) is the effective lever — increasing `k_factor` without growing `ef` returns identical recall, because faiss returns candidates ordered by distance and the global top-k merge discards the "filler" candidates outside the beam.

We need a per-segment, per-query effective `ef_search` that scales with the segment's row count, so recall is preserved when compaction enlarges segments without forcing every cluster operator to retune `ef_search` manually.

## What I'm doing:

Introduce an adaptive effective ef computed per segment at query time:

```
ef_effective = clamp(
    max(user_ef, query_k) * (1 + α * log2(rows / baseline_rows)),
    max(user_ef, query_k),                       // floor
    max(user_ef, query_k) * cap                  // ceiling
)
```

Key design points:

- Base uses `max(user_ef, query_k)` to align with faiss's internal `ef = max(efSearch, k)` rule. Otherwise when the user sets `efSearch < k`, faiss silently bumps `ef` to `k` and the user's `efSearch` becomes a no-op.
- Below `baseline_rows` (default 300K) no scaling is applied — small segments behave exactly as before.
- The per-segment row count is read at query time, so each segment in a tablet can use a different effective ef.
- The whole feature is gated behind a master switch and four mutable BE configs, so users can opt out or retune online.
- If the user supplies `efSearch` explicitly via query hint / session var, adaptive scaling is disabled for that query so user intent is honored.

New BE configs (all mutable):

| Config | Default | Purpose |
|---|---|---|
| `enable_vector_adaptive_search` | `true` | Master switch |
| `vector_adaptive_ef_alpha` | `1.0` | Growth slope |
| `vector_adaptive_ef_baseline_rows` | `300000` | Below this rows, no scaling |
| `vector_adaptive_ef_cap` | `8.0` | Multiplier ceiling |

Implementation surface:

- `tenann_index_utils.{h,cpp}`: new `compute_adaptive_ef_search(rows, user_ef, query_k)` helper applying the formula and clamps, plus `apply_adaptive_ef_search(meta, ...)` that mutates `IndexMeta::search_params["efSearch"]` only when enabled and the user did not set ef explicitly.
- `tenann_index_reader.{h,cpp}` and `vector_index_reader.h`: thread the segment row count and `user_set_ef` flag through `init_searcher`.
- `segment_iterator.cpp`: derive `user_set_ef` from `query_params` and pass the segment row count to `init_searcher`.
- `config.h` + generated `config_vector_index_fwd.h`: register the four new mutable configs (consumed via the forward header per `be/src/common/AGENTS.md`).
- `vector_index_test.cpp` (+121 lines): unit tests covering clamp behavior, the baseline rule, and the faiss `max(ef, k)` interaction.
- `docs/{en,zh}/.../BE_parameters/query_loading.md`: user-facing documentation for the four new configs.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [x] Parameter changes: four new BE configs are added; default `enable_vector_adaptive_search=true` enables adaptive scaling out of the box
- [x] Policy changes: large-segment HNSW queries now use a higher effective ef_search by default to preserve recall
- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1862,6 +1862,17 @@ CONF_mInt32(config_vector_index_default_build_threshold, "10000");
 // value may exceed nproc * this value.
 CONF_mDouble(vector_index_build_max_cpu_ratio, "0.5");
 
+// Per-segment adaptive ef_search for HNSW vector queries.
+// Compensates for recall degradation on larger segments (e.g. after compaction
+// merges many small segments into one large segment).
+//
+// Formula:
+//   ef_effective = max(user_ef, query_k) * min(1 + alpha * log2(rows/baseline), cap)
+CONF_mBool(enable_vector_adaptive_search, "true");
+CONF_mDouble(vector_adaptive_ef_alpha, "1.0");
+CONF_mDouble(vector_adaptive_ef_cap, "8.0");
+CONF_mInt64(vector_adaptive_ef_baseline_rows, "300000");
+
 // When upgrade thrift to 0.20.0, the MaxMessageSize member defines the maximum size of a (received) message, in bytes.
 // The default value is represented by a constant named DEFAULT_MAX_MESSAGE_SIZE, whose value is 100 * 1024 * 1024 bytes.
 // This will cause FE to fail during deserialization when the returned result set is larger than 100M. Therefore,
--- a/be/src/common/config_fwd_headers_manifest.json
+++ b/be/src/common/config_fwd_headers_manifest.json
@@ -339,7 +339,11 @@
         "enable_vector_index_block_cache",
         "config_vector_index_build_concurrency",
         "config_vector_index_default_build_threshold",
-        "vector_index_build_max_cpu_ratio"
+        "vector_index_build_max_cpu_ratio",
+        "enable_vector_adaptive_search",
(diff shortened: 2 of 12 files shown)


CelerData-Reviewer:
@codex review

srlch:
@codex review

srlch (approved):

luohaha (approved):

kevincai (approved):

PR #50425, opened 2024-08-29
[Feature] (Part2) Support _state/_union/_merge aggregate funciton combinator

LiShuMing:
## Why I'm doing:

1. To support common aggregate functions transforms, add _state/_union/_merge aggregate funciton combinator.

- `_state` combinator
```
/**
 * DESC: immediate_type {agg_func}_state(arg_types)
 *  input type  : aggregate function's argument types
 *  return type : aggregate function's immediate type
 */
```

- `_union` combinator
```

/**
 * Union combinator for aggregate function to union the agg state to return the immediate result of aggregate function.
 * DESC: immediate_type {agg_func}_union(immediate_type)
 *  input type          : aggregate function's immediate_type
 *  intermediate type   : aggregate function's immediate_type
 *  return type         : aggregate function's immediate_type
 */
```

- `_merge` combinator
```
/**
 * Merge combinator for aggregate function to merge the agg state to return the final result of aggregate function.
 * DESC: return_type {agg_func}_merge(immediate_type)
 *  input type          : aggregate function's immediate_type
 *  intermediate type   : aggregate function's immediate_type
 *  return type         : aggregate function's return type
 */
```
2. Support common aggregate functions in aggregate key models, so we can store aggregate's state into storage and can be used for merging later. 

## What I'm doing:
- Add _state/_union/_merge aggregate funciton combinator
- Support common aggregate functions in aggregate key models
- Fix some little bugs in some aggregate functions.
- Add creating/merging/querying all functions tests in sql-tester.

Fixes https://github.com/StarRocks/starrocks/issues/49000

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

[...]

Issue #49000, opened 2024-07-26
Support Generic Aggregate Function State


LiShuMing:
- [x] https://github.com/StarRocks/starrocks/pull/50069
- [x] https://github.com/StarRocks/starrocks/pull/50425
- [x] https://github.com/StarRocks/starrocks/pull/51510

## Feature request

**Is your feature request related to a problem? Please describe.**

When the detail data is very large, the aggregation model is very useful for querying after summarizing by dimensions; moreover, the aggregation model is also the necessary means for incremental computation of aggregate functions in StarRocks at present. 

However, it currently only supports the following aggregate functions, which are very limited; even in terms of extensibility, if you want to support intermediate states such as HllSketch or Avg functions, the current development support logic is also quite complicated, with a lot of repetitive work.

**Describe the solution you'd like**

Since all aggregate functions have already defined the types of their intermediate states and methods for serialization/deserialization, all aggregate functions defined in the Query Engine can be used as columns in the aggregation model, and are no longer limited to the aggregate functions currently supported by the aggregation model.

#### Support Generic Aggregate Function in Aggregate Model
- https://github.com/StarRocks/starrocks/pull/50069

```
ColName aggregateFunctionName(paramType1, paramType2, ....)
```

- ColName: The column name when the column is stored.
- aggregateFunctionName: Defines the aggregate function corresponding to this column, used for compaction and re-aggregation.
- The type of this column is the state type of the aggregate function, which will be automatically inferred.
Since obtaining an aggregate function requires the function's input parameters + return type + whether it is nullable, the input parameters of the aggregate function need to be input here to uniquely determine an aggregate function.

#### Support Aggregate Function Combinator
- https://github.com/StarRocks/starrocks/pull/50425

- Use combinator to support common aggregate functions transforms:

- `_state` combinator
```
/**
 * DESC: immediate_type {agg_func}_state(arg_types)
 *  input type  : aggregate function's argument types
 *  return type : aggregate function's immediate type
 */
```

- `_union` combinator
```

/**
 * Union combinator for aggregate function to union the agg state to return the immediate result of aggregate function.
 * DESC: immediate_type {agg_func}_union(immediate_type)
 *  input type          : aggregate function's immediate_type
 *  intermediate type   : aggregate function's immediate_type
 *  return type         : aggregate function's immediate_type
 */
```

- `_merge` combinator
```
/**
 * Merge combinator for aggregate function to merge the agg state to return the final result of aggregate function.
[...]

PR #73000, opened 2026-05-08
[BugFix] Allow 0-arg count combinator for IVM count(*) support

Youngwb:
## Why I'm doing:

`COUNT(*)` cannot currently be used in an INCREMENTAL materialized view because creating such an MV fails at analysis time:

```sql
CREATE MATERIALIZED VIEW mv ... PROPERTIES ('refresh_mode'='incremental')
AS SELECT k, COUNT(*) FROM iceberg_t GROUP BY k;
-- ERROR 1064: Getting analyzing error. Detail message: No matching function with signature: count_combine().
```

This is a real usability gap: `COUNT(*)` is one of the most common aggregate forms, and the equivalent `COUNT(1)` works fine — so users see two semantically-identical SQL forms behave very differently in IVM.

**Root cause** is in `AggStateUtils.isSupportedAggStateFunction`:

```java
// count only support count(col)
if (FunctionSet.COUNT.equalsIgnoreCase(fnName) && aggFunc.getArgs().length == 0 && !isAggIf) {
    return false;
}
```

This filter prevents the 0-arg form of `count` from getting any `_combine` / `_state_merge` / `_state_union` combinator registered, so the IVM rewrite path (which constructs `count_combine()` based on the user's aggregate) cannot find a matching function.

The original rationale (introduced in #50425, parent issue #49000) was that **AGG_STATE column DDL** needs explicit argument types to uniquely identify the aggregate function:

```sql
CREATE TABLE t (cnt count(int) ...);  -- requires explicit arg type
```

So the filter was meant to keep `count()` (no args) out of the AGG_STATE column DDL syntax.

**However, the SQL parser already blocks `count()` in the column type position** with a syntax error (`Unexpected input ')'`) — verified empirically. The filter at `AggStateUtils:87` is therefore redundant: it protects a path that the grammar has already closed off.

Removing the filter:
- ✅ Allows `COUNT(*)` to work in IVM (both first refresh and incremental refresh).
- ✅ Does not affect the AGG_STATE column DDL — that path is still blocked at the parser layer.
- ✅ Does not introduce any new ambiguity: `count_combine()` and `count_combine(1)` resolve correctly and produce equivalent results.
- ✅ Does not regress regular `SELECT COUNT(*)` queries.

## What I'm doing:

- Remove the 4-line filter in `AggStateUtils.isSupportedAggStateFunction` that excluded the 0-arg form of `count` from combinator registration.
- Add a unit test (`IVMAnalyzerTest#testAggregateQueryWithCountStar`) pinning the contract that `COUNT(*)` aggregate MV queries are accepted by the IVM analyzer.

## Verification

End-to-end tested on a dev cluster running the IVM POC build:

| Test | Before fix | After fix |
|---|---|---|
| `CREATE MV ... AS SELECT k, COUNT(*) FROM iceberg_t GROUP BY k` | ❌ `No matching function: count_combine()` | ✅ Created |
| First REFRESH (3 rows in 2 groups) | n/a | ✅ Correct: A=2, B=1 |
| Incremental REFRESH after delta INSERT | n/a | ✅ Correct: A=3, B=2, C=1 (matches `SELECT k, COUNT(*) FROM iceberg_t GROUP BY k`) |
| Regular `SELECT COUNT(*)` (no IVM) | ✅ Works | ✅ Still works |
| `CREATE TABLE t (cnt count() ...)` (AGG_STATE column DDL) | ❌ Parser error | ❌ Parser error (unchanged) |

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/combinator/AggStateUtils.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/combinator/AggStateUtils.java
@@ -83,10 +83,6 @@ public class AggStateUtils {
             return false;
         }
         String fnName = aggFunc.functionName();
-        // count only support count(col)
-        if (FunctionSet.COUNT.equalsIgnoreCase(fnName) && aggFunc.getArgs().length == 0 && !isAggIf) {
-            return false;
-        }
         if (ONLY_NUMERIC_ARGUMENT_FUNCTIONS_L1.contains(fnName) &&
                 Stream.of(aggFunc.getArgs()).anyMatch(t -> !t.canApplyToNumeric())) {
             return false;
--- a/fe/fe-core/src/test/java/com/starrocks/sql/analyzer/mv/IVMAnalyzerTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/sql/analyzer/mv/IVMAnalyzerTest.java
@@ -121,6 +121,36 @@ public class IVMAnalyzerTest extends MVIVMIcebergTestBase {
     }
 
     /**
+     * {@code COUNT(*)} in an aggregate MV query must be accepted by IVMAnalyzer.
+     *
+     * <p>Regression: until the fix that allows 0-arg count combinators, {@code count(*)}
+     * caused IVM rewrite to fail with {@code No matching function with signature: count_combine()}
+     * because {@code AggStateUtils.isSupportedAggStateFunction} excluded the 0-arg count form
+     * to protect AGG_STATE column DDL — but that DDL path is already blocked by the parser,
+     * so the exclusion was redundant and only blocked IVM.
+     */
+    @Test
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

Astralidea (approved):

HangyuanLiu (approved):

gengjun-git (approved):

PR #72843, opened 2026-05-06
[Enhancement] Eliminate per-row allocator hot spots in PK-index compaction KeyValueMerger

luohaha:
## Why I'm doing:

PK-index compaction's `KeyValueMerger` was the dominant CPU consumer on the realtime upsert path. `perf` sampling on the `999_1gb_1_1tb` workload (StarRocks shared-data, branch-4.1, 1 FE + 6 BE) showed ~77 % of CPU in heap alloc/free, with ~38 % of *self time* in `my_free` alone, all reachable from the merge-and-flush inner loop. This PR closes three allocator hot spots in that one class.

## What I'm doing:

Three allocator-cleanup commits in one pass over `KeyValueMerger`:

1. **Reuse protobuf scratch buffers.** Hoist `IndexValuesWithVerPB` and the serialize-output buffer to `KeyValueMerger` members; reuse them via `Clear()` / `SerializeToString(&buf)` instead of stack-construct-and-destroy per merged key.
2. **Drop `Slice::to_string()` heap allocs.** Three call sites in the merge hot path (`KeyValueMerger::merge` lines 39-40, `do_run` line 200, `merge_sstables` line 641) used `to_string()` to copy iterator-owned `Slice` data into temporary `std::string`s. The iterator's slice storage is stable until the next `Next()` call, so those copies were unnecessary. Replace with `ParseFromArray` + `Slice` comparison.
3. **Replace `std::list` with `std::optional`.** The `std::list<IndexValueWithVer> _index_value_vers` member never holds more than 0 or 1 element under any code path (`flush()` ends with `clear()`; `merge()`'s three branches each either `emplace_front` a single element or swap in a one-element list). `std::optional` stores `T` inline and is allocation-free; semantics are preserved by the structural invariant.

Net diff: 4 files, +51 / -35 LOC, all in `be/src/storage/lake/`.

Validated standalone in earlier PRs #72434, #72460, #72654 (all against `branch-4.1`); this PR rebases the lineage to `main` for unified review.

### Empirical impact

| Metric | Baseline | After all 3 | Δ |
|---|---:|---:|---:|
| \`my_free\` self-time | 38.6 % | < 3 % | -90 % |
| \`my_malloc + my_free\` self-time | ~77 % | ~6 % | -92 % |
| Compact per-task wall-clock | 46.6 s | 0.69 s | **67×** |
| Cluster CPU peak | 98.4 % | 71-88 % | -10 to -27 pts |
[...]

PR #73001, opened 2026-05-08
[Enhancement] Establish PK-index multi_get pool descent (inner_io / chunk_io / sst_open)

luohaha:
## Why I'm doing:

Cold-start cache rebuilds on shared-data PK tables produced large `upsert max` outliers. CPU and trace profiling on the realtime-benchmark `999_1gb_1_1tb` workload pinpointed two coupled causes inside `LakePersistentIndex::get_from_sstables`:

1. **Same-pool re-entrance / starvation**: parallel-publish workers run on `pk_index_execution_thread_pool`. They call `get_from_sstables` which itself wanted to fan out per-fileset multi_get tasks. Submitting that fan-out back to `pk_index_execution_thread_pool` either deadlocked (when the fan-out thread waited for tasks queued behind its own caller) or starved publish work behind cache-miss block reads.
2. **Single-stage I/O pool**: even when the publish/IO descent was split, all SST chunk-reads (the OSS-bound innermost stage) shared one pool with the orchestrator threads. Active-thread profiling showed the pool sitting under-utilized (CPU/IO mix masked the real bottleneck) while individual chunk reads piled up against a few-thread default.

The right fix is to give each layer of the descent its own pool so a worker can wait on a downstream stage without re-entering its own queue, and to size the I/O-bound layers to oversubscribe cores (since each task is dominated by OSS HTTP wait time).

## What I'm doing:

Establish the pool descent for the lake PK-index multi_get path:

```
publish workers (pk_index_execution_thread_pool)
    → Fileset::multi_get fan-out (pk_index_inner_io_thread_pool)
        → SST::multi_get chunk fan-out (pk_index_chunk_io_thread_pool)
            → SST::Open (pk_index_sst_open_thread_pool)
                → OSS HTTP
```

Each stage now lives on its own `ThreadPool`, registered for runtime metrics. New `BE_CONFIG` knobs gate the sizing; defaults are I/O-bound (`num_cores() * 4`) for `inner_io` / `chunk_io` / `sst_open`. Source draft PRs consolidated into this one:

- **#72605** — Move parallel `get_from_sstables` to dedicated `pk_index_inner_io` pool. Includes the `pk_index_execution_thread_pool` re-entrance guard so callers already running on that pool fall back to inline work instead of re-queueing.
- **#72619** — Parallelise `Table::MultiGet` into key chunks on `pk_index_chunk_io` pool. Splits the OSS-bound inner read away from the orchestration pool.
- **#72622** — Bump `pk_index_inner_io` pool default to `num_cores() * 4` after iter-035 confirmed the pool is I/O-wait-bound, not CPU-bound.
- **#72639** — Move SST `Table::Open` (4 sequential OSS round-trips per file) onto a dedicated `pk_index_sst_open` pool so cold-start opens stop queuing behind orchestration work.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

New BE config knobs (all `CONF_mInt32`, all default to `0` = derive from `num_cores`):
`pk_index_inner_io_threadpool_max_threads`, `pk_index_inner_io_threadpool_size`,
`pk_index_chunk_io_threadpool_max_threads`, `pk_index_chunk_io_threadpool_size`,
`pk_index_sst_open_threadpool_max_threads`, `pk_index_sst_open_threadpool_size`,
plus runtime gates `enable_pk_index_parallel_execution` (already present, wired into the new pools).

## Checklist:

- [ ] This is a backport pr

The new pools are reachable only from the existing PK-index code path; behavior is functionally identical to the prior single-pool path under default knobs (just non-blocking by construction). Existing `LakePersistentIndex` / `PersistentIndexSstable` tests cover correctness; the change is in pool plumbing, not data semantics.

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Validation (auto-perf-opt run my-rt-2)

- Run: 999_1gb_1_1tb realtime template, branch-4.1 baseline.
- Iterations 032 → 037 (validated stack): cold-start `upsert max` cliff produced from compounding the four pool changes; combined with the KVMerger work landed in #72843, post-iter-021 steady-state `upsert P99` settled in the 597–987 ms band and `tput` rose from baseline 293 K rows/s to ≥ 920 K rows/s.
- Source draft PRs (validated on branch-4.1): #72605, #72619, #72622, #72639.


CelerData-Reviewer:
@codex review

luohaha:
Closing based on A/B benchmark on `benchmark_luoyixin_1tb_6cn_2disk2` (1 FE + 6×32-core CN, branch-4.1 binary, 999_1gb_1_1tb template). 6 runs total (2 each baseline / tuned chunk-ON / inner-only).

**Findings**:

| Layer | Net effect (n=2 each) | Notes |
|---|---|---|
| `inner_io` (fileset fan-out) | **negative**: tput −6%, P50 +31%, Query P50 +62% | Filesets average ~5/tablet; 5-way fan-out far below 64-thread pool. Outer `pk_index_execution` already saturated (template sets 1024). Inner adds thread budget without filling. |
| `chunk_io` (within-fileset chunking) | mixed: tput +8%, P99 −20%, but **Query P50 +52%** | Real ingest gain, but `chunk_target_keys=64` is too small (warm-path dispatch overhead ~30%; loses internal MultiGet block-sharing). |
| `sst_open` (cold-start Table::Open) | config-dependent: helpful when `pk_index_execution=16` (default), neutral-or-negative when `=1024` (template). Real cold-start work (~110K opens / run, 50–60s storm) but dedicated pool size <  execution pool gives less parallelism. |

**PR-quality bugs hit during validation** (each required a fix push to make the build work on main):

1. `runtime_metrics.h` missing 3 `METRICS_DEFINE_THREAD_POOL` declarations
2. `config_fwd_headers_manifest.json` missing 6 new config entries (`config_exec_env_fwd.h` regen)
3. `util/trace.h` / `util/threadpool.h` include paths stale on main (renamed to `base/debug/trace.h` / `common/thread/threadpool.h`)
4. `config_primary_key_fwd.h` missing 4 chunk-multi-get config entries

These would have been caught by a single main rebase before submit.

**Recommendation for a follow-up PR**: keep only the `chunk_io` layer with tuned defaults (`target_keys=256+`, `max_chunks=8`, `max_threads=num_cores*2`, queue=16K + admission control). Drop `inner_io`. Reconsider `sst_open` as a config-adaptive size or reuse `pk_index_execution_thread_pool`.

Full data + analysis: [internal benchmark thread]

PR #73004, opened 2026-05-08
[BugFix] IvmRewriter throw on convergence failure instead of silent fallback

Youngwb:
## Why I'm doing:

When `IvmRewriter` cannot fully resolve every `LogicalDelta` / `LogicalVersion` marker after running the IVM rewrite rules, it currently falls back by restoring the original plan and returning normally:

```java
// Phase 2: Convergence check — if any markers remain, fall back to original plan.
if (IvmRuleUtils.containsLogicalDelta(tree.inputAt(0))
        || IvmRuleUtils.containsLogicalVersion(tree.inputAt(0))) {
    return;     // ← silent fallback
}
```

The "original plan" here is *not* the user's original SELECT — it's the plan after `IVMAnalyzer` has already mutated the AST (introducing `__ROW_ID__`, `__AGG_STATE_*` columns, and the `<func>_combine` / `<func>_state_merge` substitutions). The plan still carries `TvrVersionRange` on the table scan. So running it does:

1. Scan only the delta rows (TVR range still active).
2. Compute `<func>_combine(args)` over those delta rows ⇒ delta-only state.
3. INSERT into MV.

…**without** the LEFT OUTER JOIN against the existing MV state and the explicit `<func>_state_union(delta_state, mv_state)` projection that `IvmDeltaAggregateRule` would normally produce. Whether the result is correct depends entirely on the MV's storage-layer state-merge behaviour, which is not the contract `IvmRewriter` is documented against. In practice, for any aggregate that isn't trivially additive (or for any future scenario where a marker isn't consumed), the MV ends up with wrong data and no error signal.

This silent fallback is currently unreachable on normal flows — `IVMAnalyzer` rejects unsupported shapes earlier and the rules consume every marker for the patterns that get through — but it is a footgun: any future addition of a reject path in `IvmDeltaAggregateRule.check`, or any new operator that doesn't have a corresponding `IvmDelta*Rule`, would silently corrupt MV data.

## What I'm doing:

- In `IvmRewriter.rewrite`, change the convergence check so unresolved markers throw `SemanticException` instead of silently falling back. The existing `finally` block still restores the original plan on the `tree` object, so the optimizer pipeline sees a clean tree after the throw.
- Update the class javadoc to describe the new behaviour and explain why fallback is unsafe.
- Update `IvmRewriterTest#testRewriteFallsBackForUnsupportedOperator` (renamed `testRewriteThrowsForUnsupportedOperator`) to assert the new throw + that the original plan is still restored on the tree.

## Verification

```
$ mvn test -pl fe-core -Dtest='IVMAnalyzerTest,IvmRewriterTest,IvmIcebergRuleTest,IvmAggregateRuleTest,IvmJoinRuleTest,IvmUnionRuleTest'
[INFO] Tests run: 49, Failures: 0, Errors: 0, Skipped: 0
$ mvn checkstyle:check -pl fe-core
[INFO] You have 0 Checkstyle violations.
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
@@ -446,44 +446,53 @@ public final class MVIVMRefreshProcessor extends MVRefreshProcessor {
                 .setQuerySource(QueryDetail.QuerySource.MV.name())
                 .setCNGroup(ctx.getCurrentComputeResourceName());
 
-        // set tvr target mvid
+        // Scope the IVM session variables to this method only. The same ConnectContext
+        // is reused by MVHybridRefreshProcessor.switchToPCTRefresh on IVM failure, so
+        // leaking enable_ivm_refresh=true into the PCT retry would make IvmRewriter run
+        // on a non-IVM plan and fail at convergence — defeating the AUTO fallback.
+        boolean prevIvmEnabled = ctx.getSessionVariable().isEnableIVMRefresh();
+        String prevTvrTargetMvId = ctx.getSessionVariable().getTvrTargetMvId();
         ctx.getSessionVariable().setEnableIVMRefresh(true);
         ctx.getSessionVariable().setTvrTargetMvid(GsonUtils.GSON.toJson(mv.getMvId()));
+        try {
+            final Set<Table> baseTables = snapshotBaseTables.values()
+                    .stream()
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

Astralidea (approved):

PR #73014, opened 2026-05-09
[BugFix] Restrict IVM aggregate functions to a verified-safe whitelist

Youngwb:
## Why I'm doing:

`IVMAnalyzer` currently accepts any aggregate function for which the combinator naming convention happens to resolve at lookup time. That leaves a permissive surface in which several (function, argument-type) combinations either crash at refresh or silently produce wrong data:

| Combination | Failure mode |
|---|---|
| `AVG(DECIMAL)` | Incremental refresh nulls out the (sum, count) state and every subsequent `SELECT` returns 0 (silent wrong data) |
| `MIN(VARCHAR)` / `MAX(VARCHAR)` | `IvmOpUtils.buildStateUnionScalarOperator` throws `IllegalArgumentException` because the combinator's `intermediateType` widens to `varchar(MAX)` and trips the type-equality precondition |
| `ARRAY_AGG(*)` | Refresh fails serializing `AnyStructType` to thrift |
| `HLL_UNION(*)` / `GROUP_CONCAT(*)` | Combinators are not registered; surfaces as an unhelpful `No matching function` error |

End-to-end repros for all of these were previously confirmed against an Iceberg IVM dev cluster.

## What I'm doing:

Replace the implicit accept-by-naming flow in `IVMAnalyzer.checkAggregate` with an explicit `IVM_SUPPORTED_AGG_FUNCTIONS` whitelist of `(function name, predicate over argument types)` pairs. Combinations that aren't on the list throw `SemanticException` at CREATE time with a clear message naming the function and argument types.

The first version of the whitelist (deliberately narrow):

```java
COUNT                  → args.length <= 1                              // count(*) and count(col)
SUM                    → integer / float / DECIMAL
AVG                    → integer / float                               // exclude DECIMAL pending BE state_union fix
MIN, MAX               → integer / float / DATE / DATETIME             // exclude VARCHAR pending combinator fix
BOOL_OR                → boolean
APPROX_COUNT_DISTINCT  → any single arg
NDV                    → any single arg                                // alias of approx_count_distinct
```

[...]

PR #73027, opened 2026-05-09
[Doc] Document IVM aggregate function support and limitations

Youngwb:
## Why I'm doing:

`refresh_mode = INCREMENTAL` accepts a deliberately narrow set of aggregate functions: only the (function, argument-type) combinations that have been validated end-to-end. Users today have no way to know which combinations are supported short of trying `CREATE MATERIALIZED VIEW` and reading the error message — and the error message, while clearer with #73014, only tells them about the specific combination they tried.

Add a discoverable reference: a new **Supported Aggregate Functions** subsection to the IVM section of `CREATE_MATERIALIZED_VIEW.md`. The page lists what is supported, what is rejected, and **why** each rejected form is rejected (so it is clear what the path to enabling it is).

## What I'm doing:

Add a new subsection right after **Supported Incremental Operators** in `docs/en/sql-reference/sql-statements/materialized_view/CREATE_MATERIALIZED_VIEW.md` and its `docs/zh/...` mirror.

The new subsection covers:

- **Supported functions table** — `COUNT(*)` / `COUNT(col)`, `SUM`, `AVG`, `MIN/MAX`, `BOOL_OR`, `APPROX_COUNT_DISTINCT` / `NDV`, with the type predicate for each.
- **Explicitly rejected** — distinct aggregates, `AVG(DECIMAL)`, `MIN/MAX(VARCHAR)`, `ARRAY_AGG`, `HLL_UNION`/`BITMAP_UNION`/`GROUP_CONCAT`, window functions, `PERCENTILE_*` / `STDDEV_*` etc. — each with the underlying reason and a note that the whitelist will widen as the underlying issues are fixed.

The content mirrors `IVMAnalyzer.IVM_SUPPORTED_AGG_FUNCTIONS` introduced in #73014.

## Verification

- `grep -nE "Supported Aggregate Functions|支持的聚合函数"` returns one match per file confirming both versions have the new subsection at the right place.
- `git diff --stat` shows only the two CREATE_MATERIALIZED_VIEW.md files changed (`+25 / +25`).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [x] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


Youngwb:
Closing — better to land the doc after the underlying combinator/BE fixes (PR 7–10 in the IVM agg cleanup sequence) so it can describe the final supported set rather than the current narrow whitelist that we expect to widen. The whitelist itself is committed in #73014 and the rejection messages name what's unsupported, so users get actionable feedback during this transition.

PR #73064, opened 2026-05-10
[Enhancement] Flat-layout Lake load_spill with txn-id-based vacuum

zhangheihei:
## Why I'm doing:
Lake write path's load spill cleanup currently relies on
`LoadSpillBlockManager`'s destructor issuing per-block DeleteObject calls.
On object storage this is both:

1. **Slow on the write hot path** – every aborted/finished load pays N
   round-trip DeleteObject calls before the writer goes away.
2. **Non-resilient to BE crash / OOM-kill** – if the process dies before
   the destructor finishes, the partial spill tree under
   `<root>/load_spill/<load_id>/` is leaked forever; nothing else looks
   at it.

We need a vacuum-driven, idempotent reclaimer that can clean up after
those crashes and amortize bulk deletes off the hot path.
## What I'm doing:

Fixes #issue
### TL;DR
Lake write spill files move to a **flat, single-level** directory rooted
at `<root>/load_spill_txns/`, with the txn_id baked into each filename.
A new vacuum entry point reclaims expired files in **one** `iterate_dir`
pass keyed off the filename's hex prefix, instead of walking a nested
per-txn / per-load directory tree.

### Layout change (Lake write path only): flat, single level
`LoadSpillBlockManager` now writes Lake-write spill files as **siblings**
under a single directory:

    <root>/load_spill_txns/<txn_id_hex>_<load_id>_<frag_id>_<seq>

Why flat:

- **O(N_files) vacuum, single `iterate_dir`.** No outer-loop over txn
  directories, no inner-loop over load_id directories, no recursive
  delete of subtrees. One listing, parse prefix, batch-delete.
- **No empty-directory residue.** Because we never create per-txn or
  per-load subdirs, there is nothing to leave behind when the last file
  of a txn is reclaimed.
- **Concurrency-safe naming.** `<txn_id>_<load_id>_<frag_id>_<seq>` is
  unique by construction across concurrent fragments and re-spills of
  the same load, without coordinating on a shared parent directory.

Connector spill paths (Hive / Iceberg / Files Sink, and the
`SpillPartitionChunkWriter` path) keep the original nested
`<root>/load_spill/<load_id>/...` layout and are unchanged. Selection
between the two layouts is made by the `enable_vacuum_cleanup` ctor flag
on `LoadSpillBlockManager`; only the Lake write path opts in.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

be/src/common/config.h
```diff
@@ -1334,6 +1334,11 @@ CONF_mInt64(lake_publish_version_slow_log_ms, "1000");
 CONF_mString(lake_vacuum_retry_pattern, "*request rate*");
 CONF_mInt64(lake_vacuum_retry_max_attempts, "5");
 CONF_mInt64(lake_vacuum_retry_min_delay_ms, "100");
+// Whether vacuum should delete on-disk load_spill directories whose encoded txn_id is no longer
+// active. Keep disabled during rolling upgrade so that newly-upgraded BEs do not remove
+// legacy-format spill directories (directories without a txn_id layer) that are still being
+// written by BEs running the old code path. Enable only after the whole cluster is upgraded.
+CONF_mBool(lake_enable_vacuum_load_spill, "false");
```
luohaha: Default to true, we don't need to introduce new config item.

be/src/exec/spill/block_manager.h
```diff
@@ -127,6 +127,19 @@ struct AcquireBlockOptions {
     // in its destructor. This is used by LoadSpillBlockManager which defers parent path cleanup
     // to its own destructor via clear_parent_path(), ensuring all spill files are removed first.
     bool skip_parent_path_deletion = false;
```
luohaha, line 139: You can remove `skip_parent_path_deletion` after you introduce `skip_file_deletion`

be/src/storage/lake/vacuum.cpp
```diff
@@ -608,6 +611,118 @@ Status vacuum_txn_log(std::string_view root_location, int64_t min_active_txn_id,
...
+//     which proves the cluster has progressed past any load that could still
+//     reference them.
+Status vacuum_load_spill(std::string_view root_location, int64_t min_active_txn_id, int64_t* deleted_dirs) {
+    ASSIGN_OR_RETURN(auto fs, FileSystemFactory::CreateSharedFromString(root_location));
+    auto t0 = butil::gettimeofday_s();
+    auto ret = Status::OK();
+    int64_t local_deleted = 0;
+    auto load_spill_dir = join_path(root_location, kLoadSpillDirectoryName);
+
+    auto is_pre_upgrade_load_id = [](std::string_view name) {
```
luohaha: How about introduce a new load spill path `load_spill_v2`, so that we don't need to parse filename and path which is kind of risky

be/src/storage/load_spill_block_manager.h
```diff
@@ -86,11 +87,27 @@ class LoadSpillBlockContainer {
...
+    // offline vacuum can reclaim expired spill files by matching the <txn_id_hex> directory
+    // against the cluster-wide min_active_txn_id.
+    // For non-Lake callers (txn_id = 0), the legacy <remote_spill_path>/load_spill/<load_id>/
+    // layout is preserved to keep behaviour backward compatible.
     LoadSpillBlockManager(const TUniqueId& load_id, const TUniqueId& fragment_instance_id,
-                          const std::string& remote_spill_path, std::shared_ptr<FileSystem> fs)
-            : _load_id(load_id), _fragment_instance_id(fragment_instance_id), _fs(std::move(fs)) {
-        _remote_spill_path = remote_spill_path + "/load_spill";
+                          const std::string& remote_spill_path, std::shared_ptr<FileSystem> fs,
+                          bool enable_vacuum_cleanup = false, int64_t txn_id = 0)
```
luohaha:
```suggestion
                          bool enable_vacuum_cleanup = false, int64_t txn_id)
```

be/src/storage/load_spill_block_manager.h
```diff
@@ -86,11 +87,27 @@ class LoadSpillBlockContainer {
...
+            : _load_id(load_id),
+              _fragment_instance_id(fragment_instance_id),
+              _fs(std::move(fs)),
+              _enable_vacuum_cleanup(enable_vacuum_cleanup),
+              _txn_id(txn_id) {
+        if (txn_id != 0) {
```
luohaha: DCHECK(txn_id != 0)
txnid == 0 should not happen

be/src/storage/lake/location_provider.h
```diff
@@ -28,6 +28,7 @@ namespace starrocks::lake {
 static const char* const kMetadataDirectoryName = "meta";
 static const char* const kTxnLogDirectoryName = "log";
 static const char* const kSegmentDirectoryName = "data";
+static const char* const kLoadSpillDirectoryName = "load_spill";
```
luohaha, line 34:
```suggestion
static const char* const kLoadSpillDirectoryName = "load_spill";
static const char* const kLoadSpillDirectoryV2Name = "load_spill_v2";
```

be/src/storage/lake/vacuum.cpp
```diff
@@ -608,6 +611,118 @@ Status vacuum_txn_log(std::string_view root_location, int64_t min_active_txn_id,
...
+                    return false;
+                }
+            }
+            return true;
+        };
+
+        StringParser::ParseResult parse_res = StringParser::PARSE_FAILURE;
+        int64_t txn_id = 0;
+        if (is_hex_txn_id(name)) {
+            txn_id = StringParser::string_to_int<int64_t>(name.data(), name.size(), 16, &parse_res);
```
luohaha: Consider `extract_txn_id_prefix(...)`

be/src/storage/load_spill_block_manager.h
```diff
@@ -131,6 +149,8 @@ class LoadSpillBlockManager {
     std::unique_ptr<spill::BlockManager> _block_manager;       // Manager for blocks.
     std::unique_ptr<LoadSpillBlockContainer> _block_container; // Container for blocks.
     bool _initialized = false;                                 // Whether the manager is initialized.
+    bool _enable_vacuum_cleanup = false;                       // Lake-write vacuum cleanup mode.
```
luohaha: It should be `_enable_flat_layout` ?

be/src/storage/load_spill_block_manager.h
```diff
@@ -86,11 +87,28 @@ class LoadSpillBlockContainer {
...
+                          const std::string& remote_spill_path, std::shared_ptr<FileSystem> fs,
+                          bool enable_vacuum_cleanup = false, int64_t txn_id = 0)
+            : _load_id(load_id),
+              _fragment_instance_id(fragment_instance_id),
+              _fs(std::move(fs)),
+              _enable_vacuum_cleanup(enable_vacuum_cleanup),
+              _txn_id(txn_id) {
+        if (enable_vacuum_cleanup) {
+            DCHECK(txn_id != 0) << "vacuum cleanup mode requires non-zero txn_id";
+            _remote_spill_path = remote_spill_path + "/load_spill_txns/" + fmt::format("{:016x}", txn_id);
```
luohaha:
```suggestion
            _remote_spill_path = remote_spill_path + "/load_spill_txns";
```

be/src/storage/lake/vacuum.cpp
```diff
@@ -608,6 +612,133 @@ Status vacuum_txn_log(std::string_view root_location, int64_t min_active_txn_id,
...
+    }));
+    flush_batch();
+    ret.update(txns_iter_st);
+
+    // ---- (2) Legacy layout: <root>/load_spill/<load_id>/ ----
+    auto legacy_dir = join_path(root_location, kLoadSpillDirectoryName);
+    if (!cleanup_legacy_load_spill) {
+        LOG_EVERY_N(INFO, 1000) << "Skip legacy load_spill tree (caller did not opt in): " << legacy_dir;
+    } else {
+        auto legacy_iter_st = ignore_not_found(fs->iterate_dir2(legacy_dir, [&](DirEntry entry) {
```
luohaha: Use `delete_dir_recursive` delete legacy path` ?

be/src/storage/lake/vacuum.cpp
```diff
@@ -1392,8 +1527,23 @@ static StatusOr<std::pair<int64_t, int64_t>> path_datafile_gc(std::string_view r
...
+                        // the well-known Lake subdir names below match on both POSIX and S3.
+                        std::string_view name = entry.name;
+                        if (!name.empty() && name.back() == '/') {
+                            name.remove_suffix(1);
+                        }
+
+                        // load_spill/ is reclaimed by vacuum_load_spill (which knows about
+                        // active txn_ids); the data-file GC path would mistake live spill
+                        // files for orphans, so skip the whole subtree.
+                        if (name == kLoadSpillDirectoryName) {
```
luohaha: And `kLoadSpillTxnsDirectoryName` ?

be/src/common/config_fwd_headers_manifest.json
```diff
@@ -947,4 +941,4 @@
       ]
     }
   ]
-}
+}
```
kevincai: remove this unnecessary code change.

gensrc/proto/lake_service.proto
```diff
@@ -358,6 +358,19 @@ message VacuumRequest {
...
+    // (i.e. <root>/load_spill/<load_id_uuid>/). This layout was written by older
+    // BE versions before the txn_id-scoped layout was introduced; new code never
+    // creates it, but old data may still exist on storage.
+    //
+    // When false (default), the vacuum worker only reclaims the active layout
+    // (<root>/load_spill_txns/<txn_id>/<load_id>/) gated by min_active_txn_id;
+    // legacy directories are skipped with a throttled warning. When true, legacy
+    // directories are additionally reclaimed unconditionally. The caller (FE) should
+    // set this to true only after it can guarantee no in-flight load is still writing
+    // to the legacy layout (i.e. all BEs have been upgraded past the layout change).
```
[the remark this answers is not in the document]
kevincai: need change comment, make it up to date.

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/FullVacuumDaemon.java
```diff
@@ -152,7 +154,8 @@ private void vacuumPartitionImpl(Database db, OlapTable table, PhysicalPartition
         Locker locker = new Locker();
         locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(table.getId()), LockType.READ);
         try {
-            for (MaterializedIndex index : partition.getLatestMaterializedIndices(MaterializedIndex.IndexExtState.ALL)) {
+            for (MaterializedIndex index : partition
```
kevincai: looks like using a different style configuration, these reformat is not necessary for the PR, which introduces review noise.

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/FullVacuumDaemon.java
kevincai, line 1: A lot of reformat change in this file, try to update your auto format/auto-style configuration.

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
kevincai, line 1: similar things in this file too.

be/src/storage/lake/vacuum.cpp
```diff
@@ -608,6 +612,135 @@ Status vacuum_txn_log(std::string_view root_location, int64_t min_active_txn_id,
...
+    };
+
+    // Batch deletion buffer. S3 DeleteObjects API allows up to 1000 keys per request and
+    // staros::S3FileSystem::delete_files batches by FLAGS_delete_files_max_key_in_batch
+    // (default 1000). See design-doc §10.3.
+    constexpr size_t kBatchSize = 1000;
+    std::vector<std::string> batch;
+    batch.reserve(kBatchSize);
+
+    auto flush_batch = [&]() {
```
luohaha: Use `delete_files_async`

be/src/storage/lake/vacuum_full.cpp
```diff
@@ -214,6 +215,12 @@ Status vacuum_full_impl(TabletManager* tablet_mgr, const VacuumFullRequest& requ
     // 3. deleted txn log which have txn id < min_active_txn_id
     RETURN_IF_ERROR(vacuum_txn_log(root_loc, min_active_txn_id, &vacuumed_files, &vacuumed_file_size));
 
+    // 4. Backstop for AutoVacuum: piggyback load_spill cleanup so partitions
```
luohaha:
```suggestion
    // 4. Backstop for FullVacuum: piggyback load_spill cleanup so partitions
```

luohaha (dismissed):

be/src/storage/lake/delta_writer.cpp
```diff
@@ -1026,11 +1032,10 @@ void DeltaWriterImpl::close() {
...
     _mem_table_sink.reset();
-    if (_load_spill_block_mgr != nullptr) {
-        // ignore the return status of clear_parent_path,
-        // because the spill blocks will be cleared by GC later.
-        (void)_load_spill_block_mgr->clear_parent_path();
-    }
+    // Lake-write load_spill files are reclaimed asynchronously by the vacuum job, which
+    // scans <load_spill_txns>/<txn_id_hex> entries and deletes those whose txn_id is no
+    // longer active. The synchronous clear_parent_path() call is intentionally omitted to
+    // avoid heavy per-load DELETE traffic against object storage.
```
kevincai: The code is removed, so the comment here is explaining for nothing. Better, mentioned that previously the spill_block_mgr clears parent path here, and now it is removed because of this, this and this...

be/src/storage/load_spill_block_manager.h
```diff
@@ -86,11 +87,27 @@ class LoadSpillBlockContainer {
...
+                          const std::string& remote_spill_path, std::shared_ptr<FileSystem> fs,
+                          bool enable_flat_layout = false, int64_t txn_id = 0)
+            : _load_id(load_id),
+              _fragment_instance_id(fragment_instance_id),
+              _fs(std::move(fs)),
+              _enable_flat_layout(enable_flat_layout),
+              _txn_id(txn_id) {
+        if (enable_flat_layout) {
+            DCHECK(txn_id != 0) << "flat layout mode requires non-zero txn_id";
+            _remote_spill_path = remote_spill_path + "/load_spill_txns";
```
kevincai: `load_spill_txns` here is hardcoded, how to connect it to the one used in location_provider.h?

silverbullet233 (dismissed):

wyb (dismissed):

kevincai (approved):

luohaha (approved):

silverbullet233 (approved):

luohaha (approved):

stdpain (approved):

meegoo (approved):

PR #73067, opened 2026-05-11
[BugFix] Relax DB lock to intensive path in qe/ read-only paths

kevincai:
Switch four single-table read paths in qe/ from full DB READ to IS+table-READ via lockTableWithIntensiveDbLock so they no longer block concurrent DDL on other tables in the same database.

- ShowExecutor.showCreateInternalCatalogTable: move lookup outside the lock; revalidate the resolved Table reference once the intensive lock is held so concurrent DROP/RENAME between lookup and lock acquisition is reported as ERR_BAD_TABLE_ERROR rather than serving DDL from a stale Table.
- ShowExecutor.visitShowTabletStatement (both branches): use the known/resolved tableId for table-READ; the table-by-name branch also revalidates after locking. The single-tablet branch already re-fetches under the lock and needs no extra check.
- ConnectProcessor.handleFieldList: move lookup outside the lock; collapse the nested lock-try inside the connector-exception try into a single try-catch-finally; revalidate after locking for internal-catalog DBs (external catalog tables are not tracked by LocalMetastore).

The revalidation closes a TOCTOU window: under the old DB READ, DROP/RENAME could not interleave between lookup and use; under IS+table-READ the lookup is unprotected so the resolved Table can be removed from the DB before the lock is taken. Re-fetching by id after locking and comparing references catches that window and reports the table as not-found, matching the old semantics.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/ConnectProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/ConnectProcessor.java
@@ -775,17 +775,37 @@ public class ConnectProcessor {
             ctx.getState().setError("Unknown database(" + ctx.getDatabase() + ")");
             return;
         }
-        Locker locker = new Locker();
-        locker.lockDatabase(db.getId(), LockType.READ);
+        // Internal catalog: lookup resolves to Database.nameToTable (a ConcurrentHashMap,
+        // lock-free safe). External catalog: lookup routes through ConnectorMetadata, whose
+        // concurrency is connector-managed. No FE write path takes the LockManager lock on
+        // external db ids, so the intensive lock acquired below is harmless overhead on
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

wyb (approved):

gengjun-git (approved):

PR #73003, opened 2026-05-08
[BugFix] Reject distinct aggregate functions in IVM at CREATE time

Youngwb:
## Why I'm doing:

`COUNT(DISTINCT)` / `SUM(DISTINCT)` etc. inside an INCREMENTAL materialized view currently produces silently wrong data:

```sql
CREATE TABLE t (k VARCHAR, n INT) ...;
INSERT INTO t VALUES ('A', 10), ('A', 20), ('A', 30);    -- 3 distinct n values

CREATE MATERIALIZED VIEW mv ... PROPERTIES ('refresh_mode'='incremental')
AS SELECT k, COUNT(DISTINCT n) AS cd FROM t GROUP BY k;
REFRESH MATERIALIZED VIEW mv WITH SYNC MODE;
SELECT k, cd FROM mv;
-- A | 3                                                  -- correct so far

INSERT INTO t VALUES ('A', 10), ('A', 40);                -- 1 duplicate + 1 new ⇒ 4 distinct
REFRESH MATERIALIZED VIEW mv WITH SYNC MODE;
SELECT k, cd FROM mv;
-- A | 5                                                  -- WRONG, should be 4
```

The MV silently shows 5 (= 3 + 2 simple addition) instead of 4 (= |{10,20,30,40}|). No error, `task_runs.STATE = SUCCESS`.

**Root cause** is in `IVMAnalyzer.checkAggregate` (and its helpers): the rewrite copies the user's aggregate's children into a new `<func>_combine(args)` combinator call without preserving the `isDistinct` flag. The optimizer's `MultiDistinctByMultiFuncRewriter` runs after this rewrite and sees a non-distinct combinator, so it does nothing. The defensive `isDistinct()` check in `IvmDeltaAggregateRule.check` (line 88) is therefore dead code — by the time it runs, the AST rewrite has already stripped the distinct flag.

The result: any IVM materialized view containing a distinct aggregate that the analyzer's AST rewrite touches (most notably `COUNT(DISTINCT)` / `SUM(DISTINCT)` / `ARRAY_AGG(DISTINCT)`) produces wrong data after every incremental refresh, with no signal to the user.

## What I'm doing:

- In `IVMAnalyzer.checkAggregate`, reject any aggregate with `isDistinct()` at CREATE time, with a clear error message naming the offending aggregate.
[...]

PR #73075, opened 2026-05-11
[Enhancement] CREATE-time trial rewrite for IVM materialized views

Youngwb:
## Why I'm doing:

The IVM rewriter has a category of bugs that the existing analyzer-level checks (including the aggregate whitelist from #73014) cannot catch: **drift between the supported surface advertised by the whitelist/analyzer and what the rewriter can actually compile.** Examples:

- A new logical operator added without a matching `IvmDelta*Rule` → analyzer still accepts the shape → `IvmRewriter.rewrite` convergence fails at the user's first refresh, not at CREATE.
- A combinator's intermediate-type metadata changes in a way that breaks `IvmOpUtils.buildStateUnionScalarOperator` → analyzer/whitelist still accepts the function → refresh throws `IllegalArgumentException`.
- The rewrite pipeline gains new operators/rules over time; the whitelist will increasingly lag the actual supported surface.

Without a CREATE-time check, the user only finds out at refresh — by which point they've already defined the MV, possibly in production.

## What I'm doing:

After `IVMAnalyzer.rewriteImpl` mutates the AST (introducing `__ROW_ID__` and `__AGG_STATE_*` columns), run a **trial compilation** of the IVM refresh plan with a mocked target MV and synthetic TVR deltas, exercising the same rewrite pipeline production refresh uses. If the trial fails, surface a CREATE-time `SemanticException` so the user gets actionable feedback at CREATE time.

**New class**: `IvmTrialRewriter` (in `fe/fe-core/src/main/java/com/starrocks/sql/analyzer/mv/`).

**Plumbing change**: `OptimizerContext.overrideTargetMv` + matching `IvmRewriter.loadTargetMv` override path. The trial passes a mock MV through this override instead of registering it in the catalog.

**Scope guarantees**:
- Optimizer rewrite phase only — no fragment build, no thrift serialize.
- Uses `OptimizerOptions.newRuleBaseOpt()` to skip Memo / cost-based phase, so the trial doesn't depend on the mock MV having statistics or partitions.
- INCREMENTAL refresh mode only; PCT and (internal) AUTO are unaffected.
- Mock MV is local to the trial — never registered in the catalog. Uses fixed sentinel ids (`-1`) since it lives only inside `runTrial`.
- IVM session vars (`enable_ivm_refresh`, `tvr_target_mv_id`) are scoped via try/finally, mirroring the production pattern in `MVIVMRefreshProcessor.prepareRefreshPlan` (PR #73004).
- Latency: ~50-100ms added to CREATE for a typical incremental MV.

**Error attribution**: trial failures are wrapped as `"Failed to generate IVM refresh plan at CREATE time: <underlying>"` so the user can distinguish trial-rewrite failures from runtime refresh failures.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [x] Policy changes: CREATE MV with `refresh_mode='incremental'` now runs an additional ~50-100ms trial compilation; some MV definitions that previously succeeded at CREATE but failed at first refresh will now fail at CREATE with a clear error message.
- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Relationship to the IVM cleanup series

This is PR 6 in a sequence of IVM aggregate-function cleanup PRs:

- #73000 [BugFix] Allow 0-arg count combinator for IVM count(*) support — **merged**
- #73003 [BugFix] Reject distinct aggregate functions in IVM at CREATE time — **merged**
- #73004 [BugFix] IvmRewriter throw on convergence failure instead of silent fallback — **merged**
- #73014 [BugFix] Restrict IVM aggregate functions to a verified-safe whitelist — **merged**

Trial-rewrite is **complementary** to the whitelist (#73014):
- The whitelist defines the supported surface for known-good (function, argument-type) combinations.
- Trial-rewrite verifies the rewrite pipeline can actually compile a plan for the user's specific MV definition.

Trial-rewrite catches what the whitelist alone cannot: drift between the advertised surface and the actual rewriter implementation. Together they form a stronger CREATE-time gate than either provides alone.

## Verification

- `mvn checkstyle:check -pl fe-core` — 0 violations
- `mvn test -pl fe-core -Dtest='IVMAnalyzerTest,IvmRewriterTest,IvmIcebergRuleTest,IvmAggregateRuleTest,IvmJoinRuleTest,IvmUnionRuleTest,IVMBasedMvRefreshProcessorIcebergTest,MaterializedViewAnalyzerTest'` — **131 tests, 0 failures** (50 iceberg refresh + 24 MV analyzer + 15 IVM analyzer + 42 rule-level)

### New tests in `IVMAnalyzerTest`

**Positive**:
- `testTrialRewriteAcceptsWhitelistedAggregates` — SUM/AVG/MIN/MAX/COUNT over numeric pass the trial.
- `testTrialRewriteAcceptsNonAggregateScan` — AUTO_INCREMENT-strategy path also passes the trial.
- `testTrialRewriteNotInvokedForPCT` — PCT mode short-circuits before the trial runs.

**Negative (value proof)**:
- `testTrialRewriteCatchesRewriterFailure` — uses `MockUp` to make `IvmOpUtils.buildStateUnionScalarOperator` throw `IllegalArgumentException` (simulating combinator drift). Verifies the trial:
  - actually invokes the rewriter (otherwise the mock wouldn't trigger),
  - catches the underlying `IllegalArgumentException`,
  - wraps it with the `"Failed to generate IVM refresh plan at CREATE time"` attribution prefix,
  - preserves the underlying error message in the wrapped exception so the user can diagnose.

This is the value proof: when the rewriter changes and a previously-working combinator stops working, the trial catches it at CREATE time with a clear error pointing the user (and us) at the underlying failure.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/mv/IVMAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/mv/IVMAnalyzer.java
@@ -164,26 +164,31 @@ public class IVMAnalyzer {
 
         try {
             QueryRelation queryRelation = queryStatement.getQueryRelation();
-            // rewriteImpl returns whether the query is "retractable" (currently equivalent
-            // to "has aggregate" after dead-code cleanup). A retractable query produces its
-            // own __ROW_ID__ expression (e.g. encode(group_by_keys)); a non-retractable
-            // query — today an append-only Iceberg scan — relies on storage AUTO_INCREMENT.
+            // Retractable queries produce their own __ROW_ID__ (encode(group_by_keys)); non-retractable
+            // append-only scans rely on storage AUTO_INCREMENT.
             boolean isRetractable = rewriteImpl(queryRelation);
             RowIdStrategy strategy = isRetractable
                     ? RowIdStrategy.QUERY_COMPUTED
                     : RowIdStrategy.AUTO_INCREMENT;
+            // Trial-rewrite catches drift the analyzer-level checks can't: e.g. a new logical
+            // operator without a matching IvmDelta*Rule, or a combinator's metadata that no
+            // longer matches the BE state-union path. INCREMENTAL only.
+            if (refreshMode.isIncremental()) {
+                IvmTrialRewriter.runTrial(connectContext, statement, queryStatement);
+            }
             IVMAnalyzeResult result = IVMAnalyzeResult.of(queryStatement, strategy, refreshMode);
             return Optional.of(result);
+        } catch (SemanticException e) {
+            // Already has a self-describing message (rewriteImpl or IvmTrialRewriter). Don't re-wrap.
+            if (refreshMode.isIncremental()) {
+                throw e;
+            }
+            return Optional.empty();
         } catch (Exception e) {
             if (refreshMode.isIncremental()) {
                 throw new SemanticException("Failed to rewrite the query for IVM: %s", e.getMessage());
-            } else {
-                // If the refresh mode is not strictly incremental, we can fallback to full refresh.
-                // NOTE: pre-existing AST-mutation leak — if rewriteImpl partially mutates the
-                // queryStatement then throws, the caller will see a partially-rewritten AST
-                // when falling back to PCT mode. Tracked as a separate follow-up issue.
-                return Optional.empty();
             }
+            return Optional.empty();
         }
     }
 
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

Youngwb (commented):
Addressed the codex review feedback (PK load-op path):

The trial now uses a synthetic InsertStmt (instead of the originating CreateMaterializedViewStatement) as the OptimizerContext.statement so `IvmRewriter.isPrimaryKeyTargetMv()` returns true and `appendPkLoadOpColumn` runs. Mirrors production refresh, which always goes through an InsertStmt.

Existing positive tests (mock MV is PRIMARY_KEYS) implicitly verify the PK load-op path is exercised — if `appendPkLoadOpColumn` had a bug on the mock plan structure, the tests would fail. They pass.

Force-pushed: c9131b92ac → c4f68be469

Astralidea (dismissed):

stephen-shelby (approved):

Astralidea (approved):

PR #73087, opened 2026-05-11
[Enhancement] Split publish-trace SST miss counters by local-cache vs remote

luohaha:
## Why I'm doing:

In shared-data publish traces today we have `read_block_miss_cache_cnt` on the persistent-index sstable `MultiGet` path. It tells us the in-memory sstable block cache missed, but **not** whether the resulting file read was served by the local data cache or went out to the remote object store (S3/OSS). That distinction is what we actually need to diagnose slow PK publishes — a miss that hits local disk is dramatically different from one that hits remote.

## What I'm doing:

In `PersistentIndexSstable::multi_get`, snapshot the underlying `RandomAccessFile`'s `NumericStatistics` (`bytes_read_local_disk`, `bytes_read_remote`, `io_count_local_disk`, `io_count_remote`) before and after `Table::MultiGet`, and emit the delta as four new trace counters alongside the existing miss counter:

- `sstable_io_local_disk_bytes`
- `sstable_io_remote_bytes`
- `sstable_io_count_local_disk`
- `sstable_io_count_remote`

Streams without `NumericStatistics` (e.g. plain POSIX in shared-nothing UTs) report all-zero, so the wiring is safe across build flavors.

UT: a `FakeStatsInputStream` wraps the on-disk stream and attributes every successful `read_at_fully` to either the local-disk bucket or the remote bucket; three new cases (`test_multi_get_io_breakdown_local_disk`, `test_multi_get_io_breakdown_remote`, `test_multi_get_io_breakdown_no_stats`) drive the breakdown end-to-end and assert the published trace metrics — including the nullptr-stats branch.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/fs/bundle_file.cpp
+++ b/be/src/fs/bundle_file.cpp
@@ -141,6 +141,10 @@ StatusOr<std::unique_ptr<io::NumericStatistics>> BundleSeekableInputStream::get_
     return _stream->get_numeric_statistics();
 }
 
+io::IoStatsSnapshot BundleSeekableInputStream::get_io_stats_snapshot() const {
+    return _stream->get_io_stats_snapshot();
+}
+
 const std::string& BundleSeekableInputStream::filename() const {
     return _stream->filename();
 }
--- a/be/src/fs/bundle_file.h
+++ b/be/src/fs/bundle_file.h
@@ -111,6 +111,7 @@ public:
     StatusOr<int64_t> read(void* data, int64_t count) override;
     Status touch_cache(int64_t offset, size_t length) override;
     StatusOr<std::unique_ptr<io::NumericStatistics>> get_numeric_statistics() override;
+    io::IoStatsSnapshot get_io_stats_snapshot() const override;
 
     // Override so two slices of the same physical file produce distinct keys at the same
     // stream-relative offset: we fold the slice base offset into the key, yielding a key
--- a/be/src/fs/fs_starlet.cpp
+++ b/be/src/fs/fs_starlet.cpp
@@ -222,6 +222,20 @@ public:
         return std::move(stats);
     }
 
+    io::IoStatsSnapshot get_io_stats_snapshot() const override {
+        auto stream_st = _file_ptr->stream();
+        if (!stream_st.ok()) {
+            return {};
+        }
+        const auto& s = (*stream_st)->get_io_stats();
+        return {
+                s.bytes_read_local_disk,  s.bytes_read_remote,  s.bytes_write_local_disk, s.bytes_write_remote,
+                s.io_count_local_disk,    s.io_count_remote,    s.io_ns_read_local_disk,  s.io_ns_read_remote,
+                s.io_ns_write_local_disk, s.io_ns_write_remote, s.prefetch_hit_count,     s.prefetch_wait_finish_ns,
+                s.prefetch_pending_ns,
+        };
+    }
+
 private:
     ReadOnlyFilePtr _file_ptr;
 };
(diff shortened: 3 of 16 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

luohaha:
## Publish-path overhead A/B

Re-checked the cost of the new SST IO breakdown counters on a real shared-data cluster before and after switching from the `get_numeric_statistics()` adapter to a lightweight `IoStatsSnapshot` struct.

### Setup

- 2 identical shared-data clusters (1 FE 4c16g + 2 CN 4c16g, aliyun, same main tip `4ee8f9c`).
- Workload: `INSERT ... SELECT generate_series(1, 20000)` upserts into a single-bucket PK table seeded with 200k rows. `enable_sync_publish=true`, so per-statement wall clock includes the publish phase.
- 5 rounds × 10 iters per cluster, alternating PR ↔ baseline; round mean excludes iter-1 cold-cache spike. Reported below as the per-INSERT mean across rounds.

### Original version (snapshot via `get_numeric_statistics()` adapter)

| | base | PR | diff |
|---|------|-----|------|
| 20k-row PK upsert | 134.09 ms | 141.95 ms | **+7.86 ms (PR ~5.9% slower)** |
| 20k-row DUP control (no `multi_get`) | 97 ms | 96 ms | 0 (clusters equivalent) |

Per-call cost was concentrated in starlet's `NumericStatistics` adapter (`fs_starlet.cpp:206-216`): 1× `make_unique<NumericStatistics>` + 2× vector alloc (`reserve(11)`) + 11× `std::string` constructor in `append()`. Per `multi_get` snapshot ran twice. The publish path always has an active `Trace`, so the cost was always paid.

### Light version (this commit — `IoStatsSnapshot` struct)

`InputStream` now exposes a virtual `get_io_stats_snapshot() const` returning a 13×int64 struct. Starlet override copies directly from `IOStats` — no heap alloc, no strings, no vector.

Two runs (5 rounds each):

| Run | base | PR | diff |
|-----|------|-----|------|
| 1 | 172.95 ms | 167.68 ms | -5.27 ms |
| 2 | 165.94 ms | 151.99 ms | -13.95 ms |
| **aggregate (10 rounds)** | **169.45 ms** | **159.84 ms** | **-9.61 ms (PR ~5.7% _faster_)** |

The original ~6% slowdown is **gone**. The "PR is now faster" reading is biased by ordering (each round measures base first, then PR, so PR benefits from warmer caches); the honest read is that the lightweight snapshot drops the overhead **into the cluster-level noise**, well below resolution at this workload size.

### Bonus: easier to extend

Adding a new counter to the publish trace in the future is now:

1. Add the field to `io::IoStatsSnapshot` in `be/src/io/core/input_stream.h`.
2. Fill it in the starlet override in `be/src/fs/fs_starlet.cpp`.

No changes to the wrapper chain (`SeekableInputStreamWrapper`, `SharedBufferedInputStream`, `CompressedInputStream`, `CompressedSeekableInputStream`, `BundleSeekableInputStream`), no changes to other callers, no changes to the UT fake.

starrocks-xupeng (approved):

xiangguangyxg (dismissed):

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg (approved):

wyb (approved):

PR #73103, opened 2026-05-11
[Enhancement] Recover unstable range-colocate groups via alignment checker

xiangguangyxg:
## Why I'm doing:

After a Level-1 colocate-boundary split splices a new `ColocateRange` and PACK shard group, every other partition / table / DB in the same colocate group is left mid-aligned: `RangeColocateScanDispatch` fails closed on the misalignment, and `SplitTabletJobFactory` blocks new user-driven splits while any peer GroupId is unstable. Without an auto-recovery mechanism, a single Level-1 split puts the colocate group in a permanently broken state.

## What I'm doing:

This PR bundles two layers, structured as two commits:

**Commit 1 — External-boundaries path for tablet split.** Substrate for splitting a tablet by FE-supplied boundaries instead of BE-computed segment-distribution boundaries. FE pre-computes K-1 sort-key boundaries and passes them on the publish RPC; BE honors the boundaries verbatim while computing per-rowset stats internally. Symmetric identical-fallback on either path's validation failure.

**Commit 2 — Leader-only `ColocateChecker`.** Drives every unstable range-colocate `colocateGroupId` × table × visible index toward range alignment with `ColocateRangeMgr`'s expected ranges. Uses the external-boundaries substrate from commit 1: per-tablet canonical boundaries are converted to per-new-tablet `TabletRange` lists and submitted via `SplittingTablet.newTabletRanges`, so BE dispatches to `compute_split_ranges_from_external_boundaries` on publish.

### External-boundaries layer (commit 1)

- **Proto** (`gensrc/proto/lake_types.proto`): `SplittingTabletInfoPB.new_tablet_ranges` (repeated `TabletRangePB`) on the publish RPC.
- **BE** (`tablet_splitter.cpp`): `compute_split_ranges_from_external_boundaries` produces `vector<TabletRangeInfo>` from `external_ranges`, reusing `distribute_segment_to_ranges` / `build_rowset_anchor` / `apply_rowset_anchor` so per-rowset stat math is identical to the data-driven path. Distinct bvar counters distinguish external-boundaries vs data-driven fallback sources.
- **BE** (`tablet_range_helper.cpp`): `TabletRangeHelper::validate_new_tablet_ranges` — schema-free structural validator: closed-open per range, no byte-equal zero-width child, first.lower / last.upper match parent, adjacent ranges tile exactly via `tuple_pb_equal` (`MessageDifferencer::Equals` wrapper).
- **BE schema-aware bound validation**: reads `TabletSchemaPB` directly (no `TabletSchema::create` allocation); precomputes sort-key column metadata once per call; rejects type mismatches and decimal precision/scale mismatches with readable error messages.
- **Refactors**: `make_identical_new_tablet_metadata` helper + `build_new_tablets_from_split_ranges` helper + `build_segments_from_rowsets` shared between data-driven and external-boundaries paths.
- **FE**: `SplittingTablet.newTabletRanges` field + Gson journaling; `SplitTabletJobFactory.forExternalBoundaries` entry accepts a caller-supplied range list.

### Alignment-checker layer (commit 2)

- **`ColocateChecker`** — stateless POJO held by `TabletReshardJobMgr`. No thread of its own, no Config, no `isLeader` check, no interrupt machinery: `runOneCycle()` invoked from the manager's `runAfterCatalogReady` tick and self-gates on shared-data mode + empty unstable-groups before doing any real work via `ColocateTableIndex.hasUnstableGroups()`. The manager is started only via `startLeaderOnlyDaemonThreads`, so leader-only-ness is inherited.
- Cycle iterates every unstable range-colocate `colocateGroupId`; for each peer `GroupId`, every `NORMAL`-state table, every visible `MaterializedIndex`; per misaligned tablet, computes canonical boundaries strictly inside its range, expands them to per-new-tablet `TabletRange`s via `RangeColocateScanDispatch.computeAlignedTabletRanges`, and batches all misaligned tuples in a table into **one** `SplitTabletJob` (one job per table per cycle). When every peer is range-aligned, `markAllGroupsWithSameColocateGroupIdStable` fires across peers in lock-step.
- **`SplitTabletJobFactory.forColocateAlignment`** — checker-only entry. Bypasses the unstable-guard (checker IS the resolver). Per-new-tablet PACK assignment via `StarOSAgent.createShardsForSplitPerNewShard` (additive new overload) so each alignment-split child joins its target `ColocateRange`'s PACK group. Re-validates leader admission immediately before each StarOS shard-creation RPC.
- **`ColocateTableIndex.markAllGroupsWithSameColocateGroupIdStable` + `getAllGroupIdsWithSameColocateGroupId` + `hasUnstableGroups`** — symmetric to the existing `markAllGroupsWithSameColocateGroupIdUnstable` and friends; share a private `peersOfColocateGroup` helper.
- **`RangeColocateScanDispatch.isTabletRangesAligned` + `computeForcedAlignmentBoundaries` + `computeAlignedTabletRanges`** — static helpers shared with the checker. The last converts the canonical-boundary list into the per-new-tablet `TabletRange` list the BE consumes. Rejects non-inclusive lower bounds via `Preconditions.checkState` (would otherwise silently livelock the alignment loop).
- **`RuntimeTypeAdapterTypes`** — register `NullVariant` / `MinVariant` / `MaxVariant` so canonical `(k, NULL, ..., NULL)` tuples on `SplittingTablet.newTabletRanges` survive Gson journaling.

**Stable gate semantic**: every tablet's range is contained in exactly one `ColocateRange` and every `ColocateRange` has at least one covering tablet. PACK-shard-group physical placement of the originating-partition split-children is a tracked follow-up gated on a new staros shard-group-reassignment API.

**Known follow-ups (intentionally out of scope):**
- Production leader-demotion path needs to call `beginLeaderDemotion()` before `stopLeaderOnlyDaemonThreads()`; the checker's `isLeaderWorkAdmissionOpen()` check is forward-compatible and will start fencing as soon as that lands.
- Multi-RPC torn state during demotion has the same pattern as the user-driven split path. Needs either batched StarOS calls or a rollback API.
- Shard-creation ordering: `forColocateAlignment` creates StarOS shards before `addTabletReshardJob` writes the journal record and runs its cap check, so a cap-rejection or duplicate-id throw can leak shards. Same shape as the user-driven split path. Architecturally the fix is to move StarOS shard creation from job-creation time into the job's PENDING → PREPARING transition so it sits under the journaled state machine and inherits its retry / cleanup semantics. Tracked as a separate PR.
- MV / rollup index with reduced sort-key arity falls back to identical-tablet on strict schema validation, leaving the group permanently unstable. Per-index arity handling in `computeAlignedTabletRanges` is the fix; tracked separately.

### Testing

- **BE external-boundaries**: ~30 tests in `tablet_splitter_test.cpp` covering structural / schema / semantic validation paths + parity vs data-driven path + envelope-clipping happy path; ~10 direct `validate_new_tablet_ranges` tests in `tablet_range_helper_test.cpp` covering every rejection branch in the structural validator.
- **FE external-boundaries**: `SplitTabletJobTest` covers `forExternalBoundaries` end-to-end with a mocked `MockLakeService` that echoes ranges back; `SplittingTabletTest` covers Gson round-trip of `newTabletRanges`.
- **FE alignment-checker**: 4 `ColocateCheckerTest` cases (aligned-unstable → stable, misaligned → alignment job, table in non-NORMAL state skipped, no-unstable → no-op).
- **SQL integration**: 3 T-files in `test/sql/test_colocate_range/` (`@cloud @sequential` tagged; R recordings deferred to TSP follow-up).
- **End-to-end validation on TSP shared-data cluster**: 9 scenarios validated — basic Level-1 + recovery, sparse-partition forced-boundary, cross-DB peer alignment, unstable-guard rejection, multi-boundary Level-1 (3 prefixes), Level-2 stays stable, concurrent unstable groups, DROP table during unstable, TRUNCATE during unstable. All scenarios converged within 1-2 manager ticks.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [ ] Parameter changes
- [ ] Policy changes
- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/tablet_range_helper.cpp
+++ b/be/src/storage/lake/tablet_range_helper.cpp
@@ -286,19 +286,26 @@ StatusOr<TabletRangePB> TabletRangeHelper::convert_t_range_to_pb_range(const TTa
     return pb_range;
 }
 
-// Helper for validate_new_tablet_ranges: compare two range bounds (TuplePB)
-// for byte-for-byte equality, treating "neither has bound" as equal.
+// Structural (proto byte-for-byte) equality of two TuplePB messages. Used by
+// validate_new_tablet_ranges to detect zero-width and adjacency-tile mismatches.
 namespace {
+bool tuple_pb_equal(const TuplePB& lhs, const TuplePB& rhs) {
+    return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
+}
+
+// Same as tuple_pb_equal but treats "neither side has the bound" as equal —
+// used at the first.lower / last.upper endpoint checks where the parent range
+// may be unbounded.
 bool tuple_bound_equal(bool lhs_has, const TuplePB& lhs, bool rhs_has, const TuplePB& rhs) {
     if (lhs_has != rhs_has) return false;
     if (!lhs_has) return true;
-    return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
+    return tuple_pb_equal(lhs, rhs);
 }
 } // namespace
 
 Status TabletRangeHelper::validate_new_tablet_ranges(
         const TabletRangePB& old_tablet_range,
-        const ::google::protobuf::RepeatedPtrField<TabletRangePB>& new_tablet_ranges) {
+        const google::protobuf::RepeatedPtrField<TabletRangePB>& new_tablet_ranges) {
     if (new_tablet_ranges.empty()) {
         return Status::InvalidArgument("validate_new_tablet_ranges: new_tablet_ranges is empty");
     }
@@ -316,8 +323,7 @@ Status TabletRangeHelper::validate_new_tablet_ranges(
     //    the tablet schema).
     for (const auto& r : new_tablet_ranges) {
         RETURN_IF_ERROR(validate_tablet_range(r));
-        if (r.has_lower_bound() && r.has_upper_bound() &&
-            google::protobuf::util::MessageDifferencer::Equals(r.lower_bound(), r.upper_bound())) {
+        if (r.has_lower_bound() && r.has_upper_bound() && tuple_pb_equal(r.lower_bound(), r.upper_bound())) {
             return Status::InvalidArgument(
                     "validate_new_tablet_ranges: range with lower_bound == upper_bound (zero-width)");
         }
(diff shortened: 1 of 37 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

HangyuanLiu (approved):

luohaha (approved):

PR #73125, opened 2026-05-12
[BugFix] Fix MV rewrite type/signature mismatch for sync MV rollup

### LINKED ISSUE #72799: Wrong plan in sync MV rewrite
```
CREATE TABLE t1 (
    k1 DATE NULL,
    k2 INT  NULL,
    k3 SMALLINT NULL
) DUPLICATE KEY(k1)
DISTRIBUTED BY HASH(k1) BUCKETS 1
PROPERTIES("replication_num" = "1");

CREATE MATERIALIZED VIEW test_mv1 AS
  SELECT k1, k2, sum(k3) AS sum1 FROM t1 GROUP BY k1, k2;

function: wait_materialized_view_finish()

INSERT INTO t1 VALUES
    ('2024-01-01', 0, 30000),
    ('2024-01-01', 0, 30000),
    ('2024-01-01', 0, 30000),
    ('2024-01-01', 0, 10000),
    ('2024-01-01', 1, 5);

SELECT k1,
       sum(k3) AS sum1,
       sum(case when k2=0 then k3 else 0 end) AS sum2
FROM t1 GROUP BY k1;

SELECT k1,
       sum(if(k2=0, k3, 0)) AS sum_if
FROM t1 GROUP BY k1;
```
### Real behavior (Required)

ASAN crash

 - 33ccc4ae

HangyuanLiu:
## Why I'm doing:

Sync MV rewrite was producing type-incoherent ScalarOperator trees:
`ReplaceColumnRefRewriter` substitutes the original column (e.g. `k3 SMALLINT`)
with the MV's wider column (e.g. `mv_sum_k3 BIGINT`) but never re-derives parent
operator types or rebinds `CallOperator` `Function` signatures. As a result,
queries like

```sql
SELECT k1, sum(if(k2 = 0, k3, 0)) FROM t1 GROUP BY k1;
SELECT k1, sum(k3), sum(case when k2=0 then k3 else 0 end) FROM t1 GROUP BY k1;
```

over a sync MV `SELECT k1, k2, sum(k3) FROM t1 GROUP BY k1, k2` either crashed
BE (`VectorizedIfExpr` `down_cast` assert) or failed FE `PlanValidator` with
"Invalid plan". See #72799.

## What I'm doing:

Introduces three layered components and rewires the sync MV rewrite path:

* **`ScalarOperatorTypeReDeriver`** (generic, MV-agnostic) — bottom-up visitor
  that re-derives ScalarOperator types and rebinds `CallOperator` `Function`
  references after a leaf type change. Specialized for IF / COALESCE / CaseWhen
  / Cast (preserves user-written casts) / BinaryPredicate / InPredicate /
  SUM / COUNT. Conservative default fallback throws `TypeReDeriveException`
  on unhandled subclasses with changed children.

* **`MvColumnRefSubstitutor`** (MV package) — composes `ReplaceColumnRefRewriter`
  with `ScalarOperatorTypeReDeriver`, returning `Optional<ScalarOperator>`.

* **`MvRewriteOutputValidator`** (MV package) — observability-only output
  validator. Walks the candidate `OptExpression` and checks `CallOperator`
  signature, `CaseWhen` branch unification, and aggregate child compatibility.
  Tolerates known false-positive patterns: null `Function` pointers in `cast`
  pseudo-calls, variadic functions (`coalesce`, `concat`, ...), the
  `*_merge` / `*_union` / `*_state` agg-state function family (VARCHAR vs
  VARBINARY divergence), and nullability-mismatch produced by count→sum
  rollups. The validator NEVER gates a candidate in production — callers
  `LOG.error` on failure and let the candidate proceed; if the validator was
  wrong (false positive) the rewrite still ships, if it was right the
  downstream `PlanValidator` / BE will surface the real problem. fe-ut sets
  `FeConstants.strictMvRewriteValidator = true` (internal toggle,
  intentionally NOT a user-tunable Config) to escalate validator failures
  to `IllegalStateException` so any ReDeriver coverage gap fails CI rather
  than reaching production silently.

`MaterializedViewRule.transform` gains a post-pass between `BestIndexRewriter`
and `MaterializedViewRewriter` that collects `colRef-id → new-type` changes
from the rewritten scan's `colRefToColumnMetaMap`, walks the entire plan
tree patching every matching `ColumnRef` in place, and uses
`ScalarOperatorTypeReDeriver` to re-derive parent expressions whose used
columns intersect the widened set.

`MaterializedViewRewriter`'s three visitor methods (`visitLogicalProject` /
`visitLogicalTableScan` / `visitLogicalAggregate`) route their column
substitutions through `MvColumnRefSubstitutor` and track a
`substitutionFailed` flag. On flag set or validator failure the rule
`LOG.error`s loudly but keeps emitting the rewritten plan. Two stray `break`
statements in `visitLogicalAggregate` that prevented multi-aggregate queries
from being fully rewritten are also fixed.

`isMVMatchAggFunctions`'s eligibility check is generalized from
"if/case-when shape enumeration + `IsNoCallChildrenValidator`" to a
column-coverage criterion: any expression whose leaf `ColumnRef`s are all
present in the MV's used base column id set is eligible, with a pre-screen
that rejects subquery / lambda / non-deterministic shapes
(`containsForbiddenShape`). This restores eligibility for shapes like
`sum(coalesce(k3, 0))`, `sum(k3 * 2)`, `sum(nullif(k3, 0))`. The dead
`IsNoCallChildrenValidator` class is removed.

Also folds in an unrelated upstream Gradle build fix: `fe/build.gradle.kts`
held `staros.version = 4.0.0` while `fe/pom.xml` was already `4.1-rc4`
(commit 7d3c2007c1b updated only the Maven side), breaking the Gradle
build path's compilation of `StarMgrServer.java`. Happy to split into a
separate PR if preferred.

**Tests:**
- `ScalarOperatorTypeReDeriverTest` — 14 unit tests covering every visitor
  specialization
- `MvColumnRefSubstitutorTest` — 3 integration tests
- `MvRewriteOutputValidatorTest` — 6 unit tests including strict-mode
  escalation and null-fn `CallOperator` tolerance
- `SyncMvRewriteTypeConsistencyTest` — 1 regression anchor +
  parameterized 13-case shape matrix
- `AsyncMvRewriteTypeConsistencyTest` — 13-case async matrix documenting
  expression-rollup limitations of the async path
- Full `MaterializedViewTest` (219 tests) passes under
  `FeConstants.strictMvRewriteValidator = true` in fe-ut

Fixes #72799

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

Youngwb (approved):

stephen-shelby (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/ScalarOperatorTypeReDeriver.java
```diff
@@ -0,0 +1,334 @@
...
+ * unrecoverable type mismatch, throws {@link TypeReDeriveException}.
+ *
+ * <p>This visitor has NO MV-specific knowledge. It only knows ScalarOperator
+ * type rules. MV-specific concerns (rollup-fn family mapping, candidate
+ * rejection plumbing) live in MvColumnRefSubstitutor (added in a later phase).
+ *
+ * <p>This is the leaves-only skeleton. CallOperator, IF/CaseWhen, Cast, and
+ * predicate support are added in subsequent phases.
+ */
+public final class ScalarOperatorTypeReDeriver
```
stephen-shelby, line 55: I think the generic `visitCall` path should apply the existing implicit-cast semantics after rebinding the `Function`.

Right now this method may resolve a wider signature via `IS_NONSTRICT_SUPERTYPE_OF` / `IS_SUPERTYPE_OF`, but then constructs the new `CallOperator` with the original `newChildren`. For example, after MV substitution we can get `ADD(BIGINT, SMALLINT)`, resolve it to `ADD(BIGINT, BIGINT) -> BIGINT`, and still return children `[BIGINT, SMALLINT]`. That leaves `fn.getArgs()[1]` inconsistent with `child[1].getType()`, which is the same class of type/signature incoherence this PR is trying to fix.

Can we reuse the existing `ImplicitCastRule` here instead of hand-rolling a partial cast rule? For example, after constructing the rebound `CallOperator`, run `ScalarOperatorRewriter.DEFAULT_TYPE_CAST_RULE` so the children are aligned to `fn.getArgs()`:

`new ScalarOperatorRewriter().rewrite(newCall, ScalarOperatorRewriter.DEFAULT_TYPE_CAST_RULE)`

The existing rule already handles normal call arguments, decimal arithmetic behavior, and varargs consistently with the rest of the optimizer. It would also be good to extend `callRebindsFunctionWhenChildTypeWidens` to assert `newCall.getChild(1).getType() == BIGINT` and that the validator considers the rebuilt call coherent.

PR #73134, opened 2026-05-12
[BugFix] Persist IVM TVR position across multiple incremental refreshes

Youngwb:
## Why I'm doing:

After #73004 landed (scoping `enable_ivm_refresh` away from the Hybrid→PCT fallback), incremental MV refresh on an unpartitioned IVM MV started silently accumulating duplicate rows from the second refresh onward. The persistent TVR position (`baseTableInfoTvrVersionRangeMap`) stayed pinned at the baseline snapshot, so every later refresh recomputed delta from baseline → current and re-applied all post-baseline rows.

### Reproducer

```sql
-- iceberg base
CREATE TABLE iceberg_t (k INT, v BIGINT);
INSERT INTO iceberg_t VALUES (1,10),(2,20);

CREATE MATERIALIZED VIEW mv
  REFRESH DEFERRED MANUAL
  PROPERTIES ('refresh_mode' = 'incremental')
  AS SELECT k, v FROM iceberg_t;
REFRESH MATERIALIZED VIEW mv WITH SYNC MODE;  -- mv = 2 ✓

INSERT INTO iceberg_t VALUES (3, 30);
REFRESH MATERIALIZED VIEW mv WITH SYNC MODE;  -- mv = 3 ✓
INSERT INTO iceberg_t VALUES (4, 40);
REFRESH MATERIALIZED VIEW mv WITH SYNC MODE;  -- mv = 5  ❌ (should be 4)
INSERT INTO iceberg_t VALUES (5, 50);
REFRESH MATERIALIZED VIEW mv WITH SYNC MODE;  -- mv = 8  ❌ (should be 5)
INSERT INTO iceberg_t VALUES (6, 60), (7, 70);
REFRESH MATERIALIZED VIEW mv WITH SYNC MODE;  -- mv = 12 ❌ (should be 7)
```

### Root cause

`enable_ivm_refresh` is consulted at two distinct call sites:

1. `IvmRewriter.rewrite()` — plan-build (current behaviour: gated correctly)
2. `InsertLoadTxnCallbackFactory.of()` — INSERT execution, decides whether to register `IVMInsertLoadTxnCallback` (the only writer of `baseTableInfoTvrVersionRangeMap.put` via `afterCommitted()`)

`prepareRefreshPlan()`'s `try-finally` (added by #73004) reset the session variable as soon as the plan was built, before `execProcessExecPlan()` invoked `executor.executePlan()`. By the time the factory ran, the variable was `false`, the callback was null, `afterCommitted()` never fired, and the persistent TVR map stayed at baseline.

The only refresh that *appeared* correct was the first one after baseline — its delta = (baseline, S1) happens to be the right delta even when the persistent map stayed at baseline.

## What I'm doing:

Extend the IVM session-variable scope across both phases without adding a second set/reset. **One set, one reset per refresh.**

- `prepareRefreshPlan()`: `try-finally` → `try-catch`. The catch resets only on prepare failure (the path Hybrid takes to fall back to PCT — this preserves #73004's design intent that PCT planning must not see `enable_ivm_refresh=true`).
- `execProcessExecPlan()`: a new top-level `try-finally` resets the scope after `executor.executePlan()` returns (whether the INSERT succeeded or threw).
- A small private `closeIvmScope(ctx)` helper, idempotent on re-entry, used by both paths.

No other call sites changed: `InsertLoadTxnCallbackFactory.of()` still gates on `enable_ivm_refresh`; the Hybrid fallback path still sees a clean session variable.

Fixes the regression introduced by #73004.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Test plan

- Added `test/sql/test_materialized_view_ivm/T/test_ivm_with_iceberg_multi_refresh_tvr_advance` pinning the post-fix mv row-count sequence (2 → 3 → 4 → 5 → 7); the pre-fix would produce (2 → 3 → 5 → 8 → 12).
- Verified locally against a dev cluster running FE built from this branch — 3/3 in cdsql flaky-test mode.
- Manually reproduced the pre-fix behaviour on the same cluster before applying the fix (matches the sequence in the reproducer above).

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
@@ -87,12 +87,9 @@ public final class MVIVMRefreshProcessor extends MVRefreshProcessor {
     // whether the next task run is needed
     private boolean hasNextTaskRun = false;
 
-    // IVM session-variable scope state. The scope spans from prepareRefreshPlan() through
-    // executor.executePlan() in execProcessExecPlan(), because enable_ivm_refresh gates
-    // two distinct call sites: IvmRewriter.rewrite() at plan-build, and
-    // InsertLoadTxnCallbackFactory.of() at INSERT execution. Scoping it to plan-build
-    // only would leave the TVR callback unregistered, so baseTableInfoTvrVersionRangeMap
-    // would never advance past baseline (see closeIvmScope javadoc for the reset paths).
+    // IVM scope: enable_ivm_refresh is read both at plan-build (IvmRewriter) and at INSERT
+    // execution (InsertLoadTxnCallbackFactory), so the scope must span prepareRefreshPlan
+    // through executor.executePlan() in execProcessExecPlan. Reset paths: closeIvmScope.
     private boolean savedEnableIvmRefresh;
     private String savedTvrTargetMvId;
     private boolean ivmScopeOpen = false;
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

Astralidea (approved):

gengjun-git (approved):

PR #73142, opened 2026-05-12
[BugFix] Resolve self-referencing UniqueConstraint without external lookup

Youngwb:
## Why I'm doing:

Any `GROUP BY` / `DISTINCT` on an IVM materialized view (`refresh_mode = incremental`) fails at analyzer time with `Table null.null.<mv> is not found`. Baseline `SELECT *` and `SELECT count(*)` work; disabling `enable_materialized_view_rewrite` does not help because the throw fires during semantic analysis (via `EliminateAggRule` → `UKFKConstraintsCollector`), not during MV rewrite.

Reproducer (against any cluster with an iceberg catalog):

```sql
SET CATALOG iceberg1;
USE cdsql_pr8_demo;
CREATE TABLE bug_base (k INT, v BIGINT) PROPERTIES ('format-version'='2');
INSERT INTO bug_base VALUES (1,10),(2,20),(3,30),(1,11);
SET CATALOG default_catalog;
CREATE DATABASE IF NOT EXISTS cdsql_grpby_bug;
USE cdsql_grpby_bug;
CREATE MATERIALIZED VIEW mv
  REFRESH DEFERRED MANUAL
  PROPERTIES ('refresh_mode' = 'incremental')
  AS SELECT k, v FROM iceberg1.cdsql_pr8_demo.bug_base;
SELECT count(*) FROM mv;                  -- OK
SELECT k, count(*) FROM mv GROUP BY k;    -- ERROR 1064: Table null.null.mv is not found
```

## What I'm doing:

Two-bug interaction, both on `main`:

1. `OlapTable.getUniqueConstraints` auto-generates a self-referencing `UniqueConstraint` (`catalog=null`, `db=null`, `tableName=getName()`) for every `PRIMARY_KEYS` / `UNIQUE_KEYS` table.
2. `UniqueConstraint.getUniqueColumnNames` assumes `selfTable.isMaterializedView()` implies an external-table reference, so it unconditionally calls `MetadataMgr.getTable(ctx, catalogName, dbName, tableName)`. With `null` catalog/db that returns `null`, and the throw fires.

IVM materialized views internally use `KeysType.PRIMARY_KEYS` for the `__ROW_ID__` column (`MaterializedView.java:1057`). So as soon as IVM started using PRIMARY_KEYS, every IVM MV started carrying a self-reference constraint, and every analyzer pass that needs unique-key info (`EliminateAggRule`, `JoinOrder`, `StatisticsCalculator`) hits the throw.

The fix: skip the external lookup when `catalogName` and `dbName` are both null (the self-reference shape produced by `OlapTable.getUniqueConstraints`) and fall back to `selfTable`, mirroring the non-MV branch. `UniqueConstraint.isMatch` already has the same nullness guard, so this aligns the two code paths.

The narrower null-guard fix was chosen over special-casing the MV path because the existing `else` branch already resolves the constraint correctly against the table that owns it — only the conditional was wrong.

### Tests

- New unit test `testMvAutoGeneratedUniqueConstraintResolvesAgainstSelf` in `ShowCreateMaterializedViewStmtTest`. Verified locally with TDD red→green: without the fix the test errors with `Table null.null.mv_self_ref is not found` at `UniqueConstraint.java:63`; with the fix it passes.
- New SQL T/R `test_ivm_with_iceberg_group_by` under `test/sql/test_materialized_view_ivm/` covering GROUP BY, DISTINCT, HAVING, plus the baseline `count(*)` / `SELECT *` shapes. R recorded against the patched FE deployed on a shared dev cluster.
- Existing `testMvUniqueConstraintsTableNotExist` (which covers the external-reference branch the original guard was meant for) still passes — the fix is strictly more permissive.
- `mvn checkstyle:check -pl fe-core`: 0 violations.

`ForeignKeyConstraint` was inspected for the same pattern; FK constraints are user-declared via table properties only (no `OlapTable` auto-generation), so there is no analogous self-reference path there.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

Astralidea (approved):

gengjun-git (approved):

PR #73163, opened 2026-05-12
[Internal/TSP-Validation] Test PR — flip enable_load_segment_parallel default to true (base)

luohaha:
Internal TSP validation PR for PR #72827. Do not merge.

This branch is upstream/main + a single commit that flips `enable_load_segment_parallel` default to `true`, so a TSP shared-data cluster picks up the parallel-load path on the **base** (no parallel iterator construction) for an apples-to-apples comparison vs #72827.

Will be closed once measurement is complete.


luohaha:
TSP build-only PR. Validation complete, see #72827 follow-up #73200.

PR #73203, opened 2026-05-13
[Feature] Implement Iceberg UPDATE commit path and routing decoupling

### LINKED ISSUE #72046: Support UPDATE statement on Iceberg V2 tables using Merge-On-Read

GavinMar:
## Why I'm doing:

This is step 3 of the Iceberg UPDATE rollout. It brings the feature to a usable state by implementing the commit path with end-to-end coverage, fixing two issues found during integration, and responding to review feedback by decoupling the row-delta routing protocol from the analyzer's SELECT output.

## What I'm doing:

We covered the following aspects in this PR:

-  Implements the RowDelta atomic snapshot (position-delete files + new data files in a single commit), wires up partition / exchange handling for partitioned tables, and adds SQL integration coverage.

- Use the actually-validated snapshot id rather than the latest at commit time, preventing false serialization-isolation failures under concurrent writers.

- Reject unsupported codecs up front rather than during write; clean up an error message produced by the UPDATE analyzer.

- Decouple Iceberg row-delta routing from analyzer SELECT output. We split `TIcebergWriteMode` into `ROW_DELTA_UPDATE` / `ROW_DELTA_MIXED`. UPDATE no longer carries the `op_code` literal in its SELECT output; the BE row-delta sink defaults to "all rows are UPDATE" when no routing column is configured. MERGE-style per-row routing is preserved for the next step (mixed mode requires a trailing TINYINT `op_code` column). See the appendix below for the rationale and compatibility analysis.

Notes：
1. We decouple Iceberg row-delta routing from analyzer SELECT output

The earlier iteration appended `2 AS op_code` to the analyzer's SELECT output so the BE row-delta sink could read each row's operation from the trailing column. Review feedback raised that this **mixes a sink-internal routing protocol into the query output layout**:

- It contradicts the precedent set by `IcebergDeleteSink`, which derives the operation from the sink type itself and needs no per-row marker.
- The column carries no data the user asked for; it is a side channel for routing that happens to ride on the SELECT output.
- If not careful, it might be pruned out by some new optimization rules later on.

The new design moves the routing declaration into the sink's write-mode:

- **Pure UPDATE** (every row is delete-old + write-new): the analyzer emits a clean `[_file, _pos, data_cols...]` tuple and the planner sets `write_mode = ROW_DELTA_UPDATE`. BE forwards every chunk to both sub-sinks without scanning any routing column.
- **Mixed routing** (reserved for MERGE): the planner injects an `op_code` column as a **sink-private physical projection** and sets `write_mode = ROW_DELTA_MIXED`. Because the planner registers the column as part of the sink's input tuple (so column pruning cannot remove it) and Thrift write_mode = ROW_DELTA_MIXED declares the layout convention that BE then validates (trailing TINYINT slot required, otherwise Status::InternalError).

2. The Thrift change is safe for compatibility

(1). **The Iceberg UPDATE feature has not been released**, so there are no historical query plans, no persisted state, and no production deployments speaking the old wire enum.
(2). **Even if a stale plan reached a new BE**, no silent miswriting is possible. `create_row_delta_sink_context` explicitly validates the tuple layout against `write_mode`: `ROW_DELTA_UPDATE` requires no `op_code` column, `ROW_DELTA_MIXED` requires a trailing TINYINT `op_code` column, and any mismatch returns `Status::InternalError` before any rows are written.

Fixes: https://github.com/StarRocks/starrocks/issues/72046
StarrocksTest: https://github.com/StarRocks/StarRocksTest/pull/11258

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -1762,7 +1762,7 @@ public class IcebergMetadata implements ConnectorMetadata {
                            ConnectContext context) {
         // Skip the commit entirely when there is nothing to write — a zero-row INSERT/UPDATE/DELETE
         // would otherwise produce an empty `append` snapshot that pollutes snapshot history and
-        // confuses downstream CDC consumers. Aligns with Trino's fix in trinodb/trino#12412.
+        // confuses downstream CDC consumers.
         if (commitInfos.isEmpty()) {
             LOG.info("Skipping empty Iceberg commit for {}.{} (no data or delete files produced)", dbName, tableName);
             return;
@@ -2102,8 +2102,16 @@ public class IcebergMetadata implements ConnectorMetadata {
                 getOrDefault(UPDATE_ISOLATION_LEVEL, UPDATE_ISOLATION_LEVEL_DEFAULT));
         if (isolationLevel == IsolationLevel.SERIALIZABLE) {
             rowDelta.validateNoConflictingDataFiles();
-            rowDelta.validateNoConflictingDeleteFiles();
         }
+        // Per Iceberg's RowDelta contract, validateNoConflictingDeleteFiles must be
+        // called for UPDATE/MERGE "independently of the isolation level" — a concurrent
+        // writer adding position-delete files that overlap rows being updated would
+        // otherwise commit silently and resurrect rows that were concurrently deleted.
+        // Not required for pure DELETE (deleting an already-deleted record is
+        // idempotent), which is why the sibling DELETE-commit path above does not call
+        // this method.
+        // Reference: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/RowDelta.java#L149-L163
+        rowDelta.validateNoConflictingDeleteFiles();
 
         if (context != null) {
             updateCommitInfo(rowDelta, context);
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -201,6 +201,8 @@ import static org.apache.iceberg.TableProperties.DEFAULT_WRITE_METRICS_MODE_DEFA
 import static org.apache.iceberg.TableProperties.DELETE_ISOLATION_LEVEL;
 import static org.apache.iceberg.TableProperties.DELETE_ISOLATION_LEVEL_DEFAULT;
 import static org.apache.iceberg.TableProperties.ENCRYPTION_TABLE_KEY;
+import static org.apache.iceberg.TableProperties.UPDATE_ISOLATION_LEVEL;
+import static org.apache.iceberg.TableProperties.UPDATE_ISOLATION_LEVEL_DEFAULT;
 
 public class IcebergMetadata implements ConnectorMetadata {
 
@@ -1758,6 +1760,14 @@ public class IcebergMetadata implements ConnectorMetadata {
     @Override
     public void finishSink(String dbName, String tableName, List<TSinkCommitInfo> commitInfos, String branch, Object extra,
                            ConnectContext context) {
+        // Skip the commit entirely when there is nothing to write — a zero-row INSERT/UPDATE/DELETE
+        // would otherwise produce an empty `append` snapshot that pollutes snapshot history and
+        // confuses downstream CDC consumers. Aligns with Trino's fix in trinodb/trino#12412.
+        if (commitInfos.isEmpty()) {
+            LOG.info("Skipping empty Iceberg commit for {}.{} (no data or delete files produced)", dbName, tableName);
+            return;
+        }
+
         // Normalize db name and table name to lower case for commit queue key
         // because some catalogs are case-insensitive (e.g., Hive, Glue)
         //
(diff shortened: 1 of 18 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
```diff
@@ -1974,6 +2036,110 @@ private void commitDeleteOperation(Transaction transaction, org.apache.iceberg.T
...
+                rowDelta.conflictDetectionFilter(conflictDetectionFilterObj);
+            }
+        }
+
+        IsolationLevel isolationLevel = IsolationLevel.fromName(nativeTbl.properties().
+                getOrDefault(UPDATE_ISOLATION_LEVEL, UPDATE_ISOLATION_LEVEL_DEFAULT));
+        if (isolationLevel == IsolationLevel.SERIALIZABLE) {
+            rowDelta.validateNoConflictingDataFiles();
+            rowDelta.validateNoConflictingDeleteFiles();
+        }
```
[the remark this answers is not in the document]
GavinMar: fixed

fe/fe-core/src/main/java/com/starrocks/sql/IcebergPlannerUtils.java
```diff
@@ -66,6 +66,24 @@ public static PhysicalPropertySet createShuffleProperty(IcebergTable icebergTabl
         return new PhysicalPropertySet(distributionProperty);
     }
 
+    /**
+     * Extracts the base snapshot id frozen at plan time from the first IcebergScanNode in the plan.
+     * Returns null when there is no Iceberg scan or the table had no snapshot when planned.
+     * The returned value must be propagated to commit so that RowDelta.validateFromSnapshot
+     * checks the same window the scan was based on, not a fresher snapshot read at commit time.
+     */
+    public static Long extractBaseSnapshotId(ExecPlan execPlan) {
```
Youngwb: The base snapshot id may not in the target iceberg table, like this 
`UPDATE t SET ... WHERE id IN (SELECT id FROM other_iceberg_table))`

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
```diff
@@ -1797,15 +1807,32 @@ public void finishSink(String dbName, String tableName, List<TSinkCommitInfo> co
...
+                        dbName, tableName, extra, context);
+            } else if (hasPositionDeletes) {
+                // Pure DELETE (unchanged)
+                commitDeleteOperation(transaction, nativeTbl, dataFiles, branch,
+                        dbName, tableName, extra, context);
             } else {
-                commitDataOperation(transaction, nativeTbl, dataFiles, branch, isOverwrite, isRewrite, extra,
-                        dbName, tableName, context);
+                // Pure INSERT/OVERWRITE (unchanged)
+                commitDataOperation(transaction, nativeTbl, dataFiles, branch, isOverwrite,
```
Youngwb, line 1834: now insert/delete/update operations use commitQueueManager to commit, right?

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
```diff
@@ -1882,40 +1958,26 @@ private void commitDeleteOperation(Transaction transaction, org.apache.iceberg.T
...
         }
 
-        // Validate from snapshot if table has current snapshot
-        Snapshot currentSnapshot = nativeTbl.currentSnapshot();
-        if (currentSnapshot != null) {
-            rowDelta.validateFromSnapshot(currentSnapshot.snapshotId());
+        // Use the base snapshot id frozen at plan time so conflict detection covers
+        // every commit landed between scan and commit. Falling back to currentSnapshot
+        // here would silently skip that window and defeat SERIALIZABLE isolation.
+        Long baseSnapshotId = extra instanceof IcebergSinkExtra
```
Youngwb, line 1971: I think this is a bugFix, you better submit another PR, so we can backport this to branch-4.1 which already release iceberg delete operation

fe/fe-core/src/main/java/com/starrocks/sql/IcebergPlannerUtils.java
GavinMar: Good catch — and the same defect also affected `buildIcebergFilterExpr`, which picked the first `IcebergScanNode` regardless of which table it scanned. Both fixes have been carved out into #73354 (now merged as `8290f959619` and auto-backported to `branch-4.1`). This PR has been rebased on top.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
GavinMar: Yes. All three branches (INSERT/OVERWRITE, pure DELETE, RowDelta UPDATE) build the same `CommitTask` and route to `commitQueueManager` (if enabled). The Delete/RowDelta/Data branching only decides what goes into the transaction.

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
GavinMar: Done — carved out as #73354, merged into `main` (`8290f959619`) and auto-backported to `branch-4.1`. This PR is rebased on top. Thanks for the suggestion.

stephen-shelby (approved):

Youngwb (approved):

HangyuanLiu (approved):

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/UpdateAnalyzer.java
```diff
@@ -81,7 +78,7 @@ private static void analyzeProperties(UpdateStmt updateStmt, ConnectContext sess
      *   UPDATE table SET col1=expr1, col2=expr2 WHERE condition
      * to:
      *   INSERT INTO iceberg_row_delta_sink
-     *   SELECT _file, _pos, partition_col1, ..., new_col1, new_col2, ..., 2 AS op_code
+     *   SELECT _file, _pos, partition_col1, ..., new_col1, new_col2, ...
```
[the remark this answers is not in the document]
GavinMar: will upate later

trueeyu (approved):

PR #73210, opened 2026-05-13
[BugFix] Push down table_name predicate for tables_config

JasonQu1:
## Why I'm doing:

`SELECT ... FROM information_schema.tables_config WHERE table_schema='...' AND table_name='...'`
previously iterated every table in every authorized database before any
filtering at the table level, even when the query only asked for a single
table. The `table_name` predicate was not pushed down to the FE, so all
the per-table work (DB READ lock + per-table processing) ran regardless of
which row the caller actually wanted.

Note: the `table_schema` predicate was **already** pushed down — the BE
schema scanner forwards `_param->db` into `TAuthInfo.pattern`, and
`InformationSchemaDataSource.getAuthDbRequestResult` filters
`authorizedDbs` by that pattern. So irrelevant databases were already
being skipped before this PR. This PR closes the symmetric gap for
`table_name`.

## What I'm doing:

Push the `table_name` predicate from BE down to FE so non-matching tables
are skipped before per-table work runs.

### Changes

- **Thrift:** add optional `table_name` field to
  `TGetTablesConfigRequest`.
- **BE:** `SchemaTablesConfigScanner` forwards the optimizer-resolved
  WHERE predicate (`_param->table`) into the request.
- **FE:** `InformationSchemaDataSource.generateTablesConfigResponse`
  builds a `PatternMatcher` before the per-DB loop and short-circuits
  non-matching tables before any per-table work runs. An inline comment
  records that `table_schema` filtering is handled earlier via
  `auth_info.pattern` and intentionally not duplicated here.

## Unit tests

`LakeInformationSchemaDataSourceTest` covers the new predicate path:

- exact match `table_name` filter returns only the matching table
- MySQL LIKE pattern (`order%`) returns matching tables, excludes others
- non-existent table name returns an empty response

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [ ] Parameter changes
- [ ] Policy changes
- [ ] Feature removed
- [ ] Miscellaneous

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/gensrc/thrift/FrontendService.thrift
+++ b/gensrc/thrift/FrontendService.thrift
@@ -1502,6 +1502,7 @@ struct TAuthInfo {
 
 struct TGetTablesConfigRequest {
     1: optional TAuthInfo auth_info
+    2: optional string table_name
 }
 
 struct TTableConfigInfo {
--- a/be/src/exec/schema_scanner/schema_tables_config_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_tables_config_scanner.cpp
@@ -62,6 +62,9 @@ Status SchemaTablesConfigScanner::start(RuntimeState* state) {
     }
     TGetTablesConfigRequest tables_config_req;
     tables_config_req.__set_auth_info(auth_info);
+    if (nullptr != _param->table) {
+        tables_config_req.__set_table_name(*(_param->table));
+    }
 
     // init schema scanner state
     RETURN_IF_ERROR(SchemaScanner::init_schema_scanner_state(state));
--- a/fe/fe-core/src/main/java/com/starrocks/service/InformationSchemaDataSource.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/InformationSchemaDataSource.java
@@ -241,6 +241,19 @@ public class InformationSchemaDataSource {
 
         AuthDbRequestResult result = getAuthDbRequestResult(request.getAuth_info());
 
+        // WHERE table_schema='...' predicate is already pushed down via auth_info.pattern
+        // (set by the BE schema scanner from _param->db) and applied in getAuthDbRequestResult,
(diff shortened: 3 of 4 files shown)


fe/fe-core/src/main/java/com/starrocks/server/SharedDataStorageVolumeMgr.java
```diff
@@ -66,6 +66,10 @@ public class SharedDataStorageVolumeMgr extends StorageVolumeMgr {
 
     private final Map<String, Long> storageVolumeIdToVTabletGroupId = new ConcurrentHashMap<>();
 
+    // Local cache keyed by storage-volume id to avoid per-call gRPC round-trips to StarManager.
+    // Entries are invalidated whenever a storage volume is updated, replaced, or removed.
+    private final ConcurrentHashMap<String, StorageVolume> svIdCache = new ConcurrentHashMap<>();
```
wyb: Cache is not necessary, we can change the `gRPC` to `internal function call` in `StarClient` like `getShardInfo` later.

gensrc/thrift/FrontendService.thrift
```diff
@@ -1502,6 +1502,7 @@ struct TAuthInfo {
 
 struct TGetTablesConfigRequest {
     1: optional TAuthInfo auth_info
+    2: optional string table_name
```
wyb, line 1505: better to add `db_name` as well.

`select * from information_schema.tables_config where table_schema = "db" and table_name = "tbl";`

wyb (dismissed):

JasonQu1:
rerun

wyb (approved):

satanson (approved):

PR #73216, opened 2026-05-13
[Doc] Document FE database/table lock rubric in AGENTS.md

kevincai:
Add a "Database and Table Locks" subsection under Key Design Principles covering:

- The three lock-acquisition styles (plain `lockDatabase` for WRITE / READ / IS, vs the intensive path `lockTable[s]WithIntensiveDbLock`), with the recommendation to default to the intensive path for table-scoped work. Plain `lockDatabase` is `@Deprecated` ("prefer intensive path unless you have a specific reason"), not banned.

- The decision rubric mapping critical-section behavior to the right lock API: mutate DB-level fields -> plain WRITE; unbounded `getTables()` + per-table state walk -> plain READ; specific known tables -> intensive path; bounded table-list snapshot then per-table work -> outer plain IS + inner per-table intensive READ (the `TabletScheduler.java:2114` split-design pattern).

- The conflict-matrix facts that matter for the rubric: IS blocks DB-WRITE (so CREATE/DROP TABLE wait) but does NOT block IX (so concurrent ALTER on other tables can proceed). The choice between IS and READ for iteration is about whether you also need to freeze per-table internal state, not about whether create/drop is prevented.

- The lock-inside-try anti-pattern: if `lockX()` throws, the `finally` tries to release a never-acquired lock, which itself throws `"lock not held"`; that secondary exception suppresses the original cause, making the failure hard to diagnose. Always acquire before the `try`.

- The "trace the actual mutation" lesson from the StarMgrMetaSyncer relaxation: a site that looked like NECESSARY (WRITE) on first read turned out to only need IS+table-READ because the real mutation went through ColocateTableIndex's own writeLock. "The comment says we need WRITE" is not the rubric.

Also add a PR-checklist line flagging any new metadata-lock acquisition for explicit justification against the rubric.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [x] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/AGENTS.md
+++ b/fe/AGENTS.md
@@ -252,6 +252,84 @@ import com.starrocks.catalog.OlapTable;  // Internal!
 - No business logic or FE internals
 - Can be used by any module
 
+### Database and Table Locks
+
+FE metadata access goes through `com.starrocks.common.util.concurrent.lock.Locker`. There are two lock-acquisition styles; **new code should default to the intensive path**.
+
+```java
+// AVOID as the default choice: full DB-scope lock. WRITE blocks every op
+// on every table in the DB; READ blocks every DDL in the DB and every
+// concurrent ALTER on any table. Still the right answer for a few cases
+// (see rubric below); the @Deprecated marker on Locker.java reflects
+// "prefer the intensive path unless you have a specific reason".
+locker.lockDatabase(dbId, LockType.READ /* or WRITE */);
+
+// PREFERRED for table-scoped work: IS/IX on the DB + READ/WRITE on
+// listed tables. Unrelated tables in the same DB proceed without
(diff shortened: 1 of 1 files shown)


HangyuanLiu (approved):

alvin-phoenix-ai (approved):

kevincai:
ignore backport to branch-4.1, branch doesn't have any agent related instruction files.

PR #73226, opened 2026-05-13
[Enhancement] Add parquet_read_rows TVF to rehydrate rejected parquet rows

meegoo:
## Why I'm doing:

The rejected-records pipeline writes Parquet load failures as a short diagnostic fragment (the offending column value) plus a source anchor (`file` / `row_in_file` / `file_size` / `file_mtime_ms`) in `_statistics_.rejected_records.source_info`. That is enough to diagnose what went wrong, but the row's other columns are not preserved, so users cannot inspect the full original row or replay the load after a fix.

This blocks the most natural fix loop: read a rejected row, modify it, and re-insert.

## What I'm doing:

Ships `parquet_read_rows(source_info)`, a lateral table-valued function that re-reads the source Parquet file by anchor and returns the full row as a JSON object keyed by Parquet column name. The intended use is a single-SQL fix loop joined directly with `_statistics_.rejected_records`:

```sql
INSERT INTO db.t
SELECT cast(json_extract_string(p.raw_record, '$.id')  AS INT),
       cast(case when json_extract_string(p.raw_record, '$.val') = 'bad_val'
                 then '0'
                 else json_extract_string(p.raw_record, '$.val')
            end AS INT),
       json_extract_string(p.raw_record, '$.name')
FROM _statistics_.rejected_records r,
     TABLE(parquet_read_rows(r.source_info)) p
WHERE r.target_table = 't' AND r.format = 'parquet';
```

### Implementation notes

**FE** registers the function via `TableFunction.initBuiltins` with a fixed output schema `(file VARCHAR, row_in_file BIGINT, raw_record JSON)`. The true Parquet schema is only known at runtime, so a polymorphic return type is not possible; users decompose the JSON envelope with `json_extract_string` / `json_extract` and cast.

**BE** implements it in `be/src/exprs/table_function/parquet_read_rows.{h,cpp}` as a row-level `TableFunction`. The `.cpp`:

- Opens the file via `FileSystemFactory::CreateSharedFromString`.
- Fail-closes on size / mtime mismatch when the anchor provides them. Gracefully degrades to "no check" when the underlying backend (S3 / OSS / Starlet) does not expose `get_file_size` / `get_file_modified_time`.
- Uses `parquet::arrow::FileReader::ReadRowGroup` so only the row group containing the anchor is fetched, not the whole file.
- Serializes the row to JSON via rapidjson with type-specific paths for primitives and a `slice->ToString()` fallback for the rest (decimal / timestamp / list / struct / map / union).
- Carries its own minimal `arrow::io::RandomAccessFile` adapter over StarRocks `RandomAccessFile`, so it does not need a `FileScanNode`'s `ScannerCounter`.

A BE config `parquet_read_rows_max_anchors` (default 10000) caps anchors per invocation chunk.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

PR #73239, opened 2026-05-13
[BugFix] Reuse HttpClient instances to fix FE FD leak

stephen-shelby:
## Why I'm doing:

A production FE has been running out of file descriptors and getting restarted regularly. Investigation found `java.net.http.HttpClient.newHttpClient()` is invoked per-request in two sites and the instances are never released:

- **`OAuth2Action.getToken()`** — one HttpClient per OAuth2 token exchange.
- **`StreamLoader.loadBatch()`** — one HttpClient per query-history flush (default every 15 min when `enable_query_history` is on).

In JDK 17, `java.net.http.HttpClient` has no `close()` method (only added in JDK 21). The only release path is GC — but the SelectorManager daemon thread keeps the implementation reachable, and under sustained periodic load GC simply cannot keep up. Each leaked instance holds:

- a SelectorManager thread
- an `anon_inode:[eventpoll]` FD
- an `anon_inode:[eventfd]` FD
- a pipe pair
- any still-open keep-alive sockets (left in `CLOSE_WAIT`)

Evidence from one affected FE (uptime ~65 h):

| Signal | Count |
|---|---|
| Live `HttpClient-*-SelectorManager` threads | 370 |
| `anon_inode:[eventpoll]` FDs | 370 |
| `anon_inode:[eventfd]` FDs | 370 |
| TCP sockets in `CLOSE_WAIT` | 781 |
| HttpClient instances created across JVM lifetime (peak observed) | 19,119+ |
| Total FDs at the time | ~2,700 (≈ 3× normal baseline) |

The FE ultimately exhausts its FD limit and stops accepting connections.

The same anti-pattern is used in three places on `main`; one (`QueryStatisticsInfo`) has already been refactored away to not need HTTP at all, leaving the two sites this PR fixes. Other FE code uses HTTP correctly: `EsRestClient` keeps a `private static final OkHttpClient` singleton, `CostPredictor` keeps an instance-level `CloseableHttpClient`. This PR aligns the two regressing sites with that existing pattern.

## What I'm doing:

Convert both call sites to a `private static final HttpClient HTTP_CLIENT` singleton, configured with:

- `connectTimeout(Duration.ofSeconds(5))` — avoid hanging indefinitely on an unreachable endpoint.
- `version(HttpClient.Version.HTTP_1_1)` — match the existing implicit behavior and avoid HTTP/2-specific stream FD usage.

`java.net.http.HttpClient` is documented as thread-safe and intended to be reused; sharing a single instance is the recommended pattern.

No behavioral change for callers — same request, same response handling — only the underlying client is now shared instead of recreated.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/http/rest/OAuth2Action.java
+++ b/fe/fe-core/src/main/java/com/starrocks/http/rest/OAuth2Action.java
@@ -43,10 +43,21 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 public class OAuth2Action extends RestBaseAction {
+    // Reuse a single HttpClient instance across all OAuth2 token exchanges.
+    // java.net.http.HttpClient is thread-safe and designed to be used as a singleton.
+    // Creating a new instance per request leaks FDs (selector epoll/eventfd/pipe + idle
+    // sockets) because the JDK 17 HttpClient has no close() method and relies on GC,
+    // which fails to keep up under sustained load. See: https://bugs.openjdk.org/browse/JDK-8277459
+    private static final HttpClient HTTP_CLIENT = HttpClient.newBuilder()
+            .connectTimeout(Duration.ofSeconds(5))
+            .version(HttpClient.Version.HTTP_1_1)
+            .build();
+
     public OAuth2Action(ActionController controller) {
         super(controller);
     }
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

gengjun-git (approved):

satanson (approved):

fe/fe-core/src/main/java/com/starrocks/http/rest/OAuth2Action.java
```diff
@@ -43,10 +43,24 @@
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 public class OAuth2Action extends RestBaseAction {
+    private static final int CONNECT_TIMEOUT_SECOND = 5;
+    private static final int REQUEST_TIMEOUT_SECOND = 10;
```
kevincai, line 52: eventually, this hardcoded timeout will be requested configurable. lol.

kevincai (approved):

PR #73253, opened 2026-05-13
[auto-perf-opt] PK index size-tiered compaction: pick multiple levels per cycle

luohaha:
## What this changes

`LakePersistentIndexSizeTieredCompactionStrategy::pick_compaction_candidates` now appends filesets from multiple priority-sorted levels into a single `result.candidate_filesets`, up to the existing `lake_pk_index_sst_max_compaction_versions` cap (default 100). Previously it returned only the single best level.

Gated by `pk_index_size_tiered_pick_multi_levels` (default `true`; flip to `false` for the prior behaviour).

## Why

In the auto-perf-opt run `pk-sortkey-opt` (1×100 GB sortkey, 6 CN shared-data), iter-004 measured the post-#73245/#73250 residual upsert tail:

- ~40 traces of 140 k in a 20-min window with `cost ≥ 300 ms`, dominated by `parallel_upsert_wait_us = 90–412 ms` bounded by `pindex_sstables_queried_cnt = 400–410` and `parallel_upsert_cnt = 10`.
- `cloud_native_pk_index_compact` thread pool: **max queue 0, max util 0.26 %** across the test window (PromQL `max_over_time` over 20 min).

The compact pool has tons of headroom; the bottleneck is purely scheduling. `compact_mgr->async_compact` already parallelises across the filesets handed to it. By feeding it more filesets per cycle, level compactions overlap instead of draining round-by-round.

## Risk / blast radius

- Memory: each fileset's sstable list is held in `result.candidate_filesets` until compaction completes. Already capped by `lake_pk_index_sst_max_compaction_versions = 100`. The cap is unchanged.
- Correctness: each fileset still becomes an independent compaction task (mergers run per-fileset). Mixing filesets across levels does not cross-pollinate per-fileset merges — it just submits more independent tasks to the pool.
- `merge_base_level`: becomes true if ANY picked level contains the base-level fileset_id. Same condition as before, just evaluated across the broader set of picked levels.
- Failure mode for memory-tight clusters: set `pk_index_size_tiered_pick_multi_levels=false` to restore the previous behaviour.

## Acceptance criteria (to be validated by the next auto-perf-opt iteration)

- `pindex_sstables_queried_cnt` P99 for upsert publishes on rt_bench tablets drops from ~410 (iter-004) to ≤ 250.
- `parallel_upsert_wait_us` P99 for slow upserts drops from ~300 ms to ≤ 150 ms.
- No regression on:
[...]

PR #73254, opened 2026-05-14
[auto-perf-opt] Parallelise PK-index cold-load segment iteration

luohaha:
## Why

`LakePersistentIndex::load_from_lake_tablet` (PK-index cold-load on the first publish per (CN, tablet) after a binary restart) is currently a single-threaded loop over (rowset, segment) pairs. Each iteration serially reads a segment from OSS, encodes its PKs, then inserts a chunk-at-a-time into the in-memory memtable. Even at idle host CPU / disk, this can dominate freshness P99 right after a deploy.

Observed on the `1_100gb_sortkey` benchmark (commit `ee2e3da`, PR #73253 build, deploy-induced cold-load):

| Trace counter (one cold-load on tablet 109668089, 2 segments, ~1 M rebuild rows) | Value |
| --- | ---: |
| `primary_index_load_latency_us` | 4 218 845 us |
| `pindex_load_from_lake_tablet_us` | 4 203 657 us |
| `rebuild_index_segment_cost_us` (sum across segments) | 4 194 404 us |
| `rebuild_get_next_cost_us` (OSS read) | 255 555 us |
| `rebuild_pk_encode_cost_us` | 63 068 us |
| `lake_persistent_index_insert_us` | 3 864 266 us |
| `pindex_memtable_insert_us` | 2 528 943 us |

So per cold-load tablet:
- ~17 % is in OSS get_next + PK encode (parallelisable, no shared state)
- ~60 % is in `_memtable->insert` (serial btree, NOT parallelisable in this patch)
- ~30 % is in the `LakePersistentIndex::insert` wrapper around the memtable (per-call vector construction, trace overhead — serial today)

The publish-version path runs each tablet's cold-load on its own worker, so within one tablet the 4.2 s is wall-clock; freshness P99 of the deploy-window publishes inherits this.

## What

Reshape the inner segment loop so segments WITHIN one rowset run on `pk_index_execution_thread_pool` concurrently. The outer rowset loop stays sequential because different rowsets may carry updates of the same PK and require monotonic-rss_rowid ordering; segments within one rowset (one transaction) cannot share a PK, so they are independent and a per-rowset mutex around the memtable insert is sufficient.

Each worker:
- Drains its assigned segment iterator (OSS read + chunk decode happen outside the lock).
- PK-encodes locally into its own `pk_column`.
- Builds the `IndexValue` vector locally.
- Takes `insert_mutex` only for the `insert(...)` call (which internally does `_memtable->insert` + `flush_memtable()`).

The orchestrator collects per-iteration stats from atomics and emits a single `TRACE_COUNTER_INCREMENT` per counter after `token->wait()` — trace context is thread-local so worker-thread increments would be silently dropped (same pattern as `load_dels`).

A new mutable config `enable_pk_index_rebuild_parallel_segment` (default `true`) gates the parallel path; it's AND'd with the existing `enable_pk_index_parallel_execution` master switch. Single-segment rowsets fall through to the original inline path.

## Correctness

- Memtable mutations are fully serialised by `insert_mutex`. The btree `_map` and `flush_memtable()`'s replacement of `_memtable` both run only inside the lock.
- The outer rowset loop remains sequential, preserving "later rss_rowid wins" semantics across rowsets.
- On `submit_func()` failure the segment runs inline, so we never drop a row from the rebuild.
- `OlapReaderStatistics` is still shared by all per-segment iterators (the iterators receive the pointer once at creation). Counter races there are observability-only (lost increments, never invalid values).
- Per-segment local buffers (`local_chunk`, `local_rowids`, `local_pk_column`) live on the worker stack — they cannot be hoisted out because workers run concurrently. Memory cost per concurrent segment ≈ a 4 K-row chunk + encoded PK column, well under 10 MB.

## Expected impact (pre-registered)

On the `1_100gb_sortkey` benchmark, restart-induced cold-load window:

| Metric | Target |
| --- | --- |
| Per-tablet `rebuild_index_segment_cost_us` (wall clock) | ≤ `sum(per_segment_cost) * 0.65` (i.e. ≥ 35 % reduction when ≥ 2 segments rebuild) |
| `pindex_load_from_lake_tablet_us` P99 in the first 60 s of test | ≤ 3 000 us (down from ~4 200 ms today) |
| Steady-state `pindex_load_from_lake_tablet_us` after cold-load | unchanged (the serial path is identical) |
| Upsert P99 / throughput | within noise vs PR #73253 |

The fix is intentionally bounded: it only collapses the ~17 % parallelisable portion of cold-load wall-clock. The dominant `_memtable->insert` cost (~60 %) is single-threaded btree work that needs a different attack (bulk-sort + sstable build) and is left for a follow-up.

## Verification

The next auto-perf-opt iteration will deploy this PR and measure all of the above on the same cluster.

This PR is opened as **draft**. It will be promoted to ready-for-review only after a subsequent iteration's benchmark confirms the pre-registered acceptance criteria; otherwise it is closed.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -489,6 +489,11 @@ CONF_mInt32(lake_partial_update_thread_pool_max_threads, "0");
 CONF_mInt32(lake_partial_update_thread_pool_queue_size, "2048");
 // The maximum number of memtables for pk index in shared-data mode.
 CONF_mInt32(pk_index_memtable_max_count, "2");
+// Parallelise the segment-iteration loop in LakePersistentIndex::load_from_lake_tablet
+// (PK-index cold-load on first publish per CN/tablet). Reads and decodes segments
+// of the same rowset concurrently on `pk_index_execution_thread_pool`; the memtable
+// insert itself remains serialised by a per-rowset mutex.
+CONF_mBool(enable_pk_index_rebuild_parallel_segment, "true");
 // The maximum wait flush timeout for pk index memtable in shared-data mode, in milliseconds.
 CONF_mInt64(pk_index_memtable_max_wait_flush_timeout_ms, "30000");
 // The parameters for pk index size-tiered compaction strategy.
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -1140,8 +1140,8 @@ Status LakePersistentIndex::load_from_lake_tablet(TabletManager* tablet_mgr, con
     // segment, so segments are independent and can run concurrently; the only shared
     // state is the memtable, guarded by `insert_mutex`. OSS reads + decode + PK
     // encoding all run outside the lock, which is the cold-load wall-clock win.
-    const bool parallel_segments = config::enable_pk_index_rebuild_parallel_segment &&
-                                   config::enable_pk_index_parallel_execution;
+    const bool parallel_segments =
+            config::enable_pk_index_rebuild_parallel_segment && config::enable_pk_index_parallel_execution;
     for (auto& rowset : rowsets) {
         TRACE_COUNTER_INCREMENT("total_segment_cnt", rowset->num_segments());
         TRACE_COUNTER_INCREMENT("total_num_rows", rowset->num_rows());
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -1124,22 +1124,24 @@ Status LakePersistentIndex::load_from_lake_tablet(TabletManager* tablet_mgr, con
     const uint64_t rebuild_rss_rowid_point = sstables.empty() ? 0 : sstables.rbegin()->max_rss_rowid();
     const uint32_t rebuild_rss_id = rebuild_rss_rowid_point >> 32;
     OlapReaderStatistics stats;
-    MutableColumnPtr pk_column;
-    if (pk_columns.size() > 1 || pk_encoding_type == PrimaryKeyEncodingType::PK_ENCODING_TYPE_V2) {
-        // more than one key column or big endian encoding
-        if (!PrimaryKeyEncoder::create_column(pkey_schema, &pk_column, pk_encoding_type).ok()) {
-            CHECK(false) << "create column for primary key encoder failed";
-        }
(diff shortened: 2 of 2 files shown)


luohaha:
Closing — iter-007 measurement: this PR delivers a measurable but below-target cold-load reduction; 3/5 acceptance criteria fail.

Setup: TSP build 1586 (`--pr 73254 --scope be`, commit `0a92188`), deployed alone (not stacked) on `benchmark_luoyixin_1tb_6cn_2disk2` (6 BEs, branch-4.1 baseline `8813309`). 20-min `realtime_benchmark --template 1_100gb_sortkey` workload.

| Acceptance criterion | Target | Measured (3 cold-load tablets) | Verdict |
|---|---:|---:|---|
| `rebuild_index_segment_cost_us` ≤ 0.65 × sequential sum | ≤ 0.65 | **~0.85** (3 660-3 940 ms vs iter-006 4 200-4 800 ms) | ❌ |
| `pindex_load_from_lake_tablet_us` P99 (first 60 s) | ≤ 3 000 ms | **3 665-3 945 ms** | ❌ |
| Steady-state real-upsert P99 | unchanged | 55 ms (vs iter-005's 87 ms — better) | ✅ |
| Throughput within ±10 % | ±10 % | 42 282 r/s (+0.2 % vs iter-005 42 201 r/s) | ✅ |
| Freshness P99 | ≤ 18 000 ms | **22 247 ms** | ❌ |

Why the small speedup: the design serialises `_memtable->insert()` (and the surrounding wrapper) via `insert_mutex`, so only the OSS `get_next` + PK encode work runs in parallel. That portion is ~10 % of cold-load (~200 ms across 2 segments); the dominant ~30 % (`LakePersistentIndex::insert` wrapper, ~1 070 ms over ~982 calls) and ~60 % (`_memtable->insert`, ~2 400 ms) stay serial. The wall-clock improvement is real (~600-900 ms per cold-load) but caps out far below the target.

iter-008 takes a different angle: **batch the wrapper calls** in the rebuild loop so we make ~60 large `insert()` calls instead of ~982 small ones. That directly attacks the ~1 070 ms wrapper tax that this PR cannot touch. See PR #73256.

Trace + numbers in `/auto-perf-opt/.workspace/pk-sortkey-opt/state/runs/run_20260513_162006/iterations/007/`.

PR #73256, opened 2026-05-14
[auto-perf-opt] Batch PK-index cold-load rebuild inserts

luohaha:
## What

`LakePersistentIndex::load_from_lake_tablet` calls the public `insert(n, keys, values, version)` helper once per `ChunkIterator` chunk (~2 K rows). On a 2 M-row cold-load rebuild that is ~982 wrapper round trips. Each round trip does:

1. Slice/vector construction (the `else` branch builds a `std::vector<Slice>` from `pk_column->continuous_data()` per chunk).
2. `_memtable->insert(n, keys, values, version)` — the actual btree work, counted in `pindex_memtable_insert_us`.
3. `flush_memtable()` — cap check; on this workload the cap is not hit (`flush_times = 1` across the whole cold-load), so this is a tight no-op.

iter-007 trace numbers on rt_bench 1_100gb_sortkey (PR #73254 deployed):

| trace counter | per cold-load tablet |
|---|---:|
| `pindex_load_from_lake_tablet_us` | 3 665 – 3 945 ms |
| `lake_persistent_index_insert_us` | 3 438 – 3 613 ms |
| `pindex_memtable_insert_us` | 2 368 – 2 480 ms |
| **gap** (wrapper, ~982 calls) | **~1 070 ms** |

The ~1 070 ms gap is the wrapper tax that #73254's segment-level parallelism cannot touch because the wrapper sits inside the per-call critical path and is serialised by `insert_mutex` anyway. Get-next + PK-encode (the parallelisable portion) is < 200 ms total — small compared to the wrapper.

## How

Accumulate keys + values across chunks WITHIN one rowset into a stack-local batch buffer (`batch_key_buf` deep-copies bytes so `pk_column->reset_column()` on the next chunk can't invalidate them). When the buffer reaches `lake_pk_index_rebuild_batch_rows` rows (default **32 768**, ~16 chunks), flush with a single `insert()` call. Mandatory drain at the end of each rowset because every rowset carries its own rebuild version and `insert()` persists that version into the memtable.

- New mutable config: `lake_pk_index_rebuild_batch_rows` (default 32768). Setting it to `0` falls back to the exact per-chunk path — byte-identical to pre-patch behaviour, used as a kill switch.
- New trace counters: `rebuild_batch_insert_cnt`, `rebuild_batch_total_rows` (lets us verify the batching took effect in trace logs without grepping BE.INFO).
- Memory cost: `threshold × (avg_key_size + sizeof(IndexValue))` per rebuild, ~3-5 MB at default — bounded and lives on the publish thread's stack.

## Why not just remove the per-call `flush_memtable()` check?

`flush_memtable()` is the only cap-enforcement point (see `LakePersistentIndex::{insert,replace,upsert}()` — every public mutator funnels through it). Removing it would let the memtable grow unbounded mid-rebuild. The batch approach keeps `flush_memtable()` invariants intact while amortising the per-call overhead.

## Why batch only within a rowset?

`insert(n, keys, values, version)` writes `version` into the memtable for each key. Different rowsets carry different rebuild versions (`rowset_version = rowset->version() != 0 ? rowset->version() : base_version`), so mixing them in one batch would corrupt the version stamp. Within one rowset the version is fixed, and different segments of the same rowset cannot share a PK (one txn → one segment per PK), so it is also safe to batch across segments.

## Expected impact

Per-cold-load tablet should drop from ~3 800 ms (iter-007 with #73254) / ~4 500 ms (iter-006 baseline, no parallel rebuild) to **≤ 2 800 ms**, by collapsing ~922 wrapper calls into ~60 batched ones (~1 ms saved per dropped call).

Steady-state path: untouched. The publish-version path uses `replace()` / `upsert()` and never goes through `load_from_lake_tablet`. The compaction path uses the conflict-resolver from #73250, also untouched.

## Test plan

- [ ] `realtime_benchmark --template 1_100gb_sortkey --duration-minutes 20` on a 6-CN cluster after deploy; verify `pindex_load_from_lake_tablet_us` P99 in first 60 s ≤ 2 800 ms.
- [ ] Verify `rebuild_batch_insert_cnt` ≈ ceil(`rebuild_index_num_rows` / 32768) per cold-load trace.
- [ ] No regression on steady-state real-upsert publish (P99 ≤ 100 ms server-side).
- [ ] Throughput within ±10 % of baseline.
- [ ] `lake_pk_index_rebuild_batch_rows = 0` falls back to exact pre-patch behaviour.

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -491,6 +491,13 @@ CONF_mInt32(lake_partial_update_thread_pool_queue_size, "2048");
 CONF_mInt32(pk_index_memtable_max_count, "2");
 // The maximum wait flush timeout for pk index memtable in shared-data mode, in milliseconds.
 CONF_mInt64(pk_index_memtable_max_wait_flush_timeout_ms, "30000");
+// Cold-load rebuild of LakePersistentIndex used to call LakePersistentIndex::insert() once per
+// segment-iterator chunk (~2 K rows). That keeps the per-call wrapper overhead (vector
+// construction, flush_memtable cap check) on the critical path for ~hundreds of calls per
+// tablet. Accumulate up to this many rows across chunks within one rowset before invoking
+// insert(); flush at rowset boundary to preserve the rowset->version contract. 0 disables
+// the optimisation (falls back to per-chunk insert, pre-patch behaviour).
+CONF_mInt32(lake_pk_index_rebuild_batch_rows, "32768");
 // The parameters for pk index size-tiered compaction strategy.
 CONF_mInt64(pk_index_size_tiered_min_level_size, "131072");
 CONF_mInt64(pk_index_size_tiered_level_multiplier, "10");
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -1286,15 +1286,13 @@ Status LakePersistentIndex::load_from_lake_tablet(TabletManager* tablet_mgr, con
                                 keys.emplace_back(fkeys, _key_size);
                                 fkeys += _key_size;
                             }
-                            RETURN_IF_ERROR(insert(pkc->size(),
-                                                   reinterpret_cast<const Slice*>(keys.data()),
+                            RETURN_IF_ERROR(insert(pkc->size(), reinterpret_cast<const Slice*>(keys.data()),
                                                    values.data(), rowset_version));
                         }
                     } else {
                         // Append into the rowset-scoped batch buffer.
                         if (pkc->is_binary()) {
-                            append_batch(reinterpret_cast<const Slice*>(pkc->raw_data()), pkc->size(),
-                                         values.data());
+                            append_batch(reinterpret_cast<const Slice*>(pkc->raw_data()), pkc->size(), values.data());
                         } else {
                             // Synthesize fixed-size Slices on the stack to feed append_batch,
                             // which immediately copies the bytes into batch_key_buf.
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -1139,6 +1139,48 @@ Status LakePersistentIndex::load_from_lake_tablet(TabletManager* tablet_mgr, con
     int64_t get_next_cost_us = 0;
     int64_t pk_encode_cost_us = 0;
     int64_t build_values_cost_us = 0;
+    int64_t batch_insert_cnt = 0;
+    int64_t batch_total_rows = 0;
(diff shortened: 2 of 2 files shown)


luohaha:
## iter-008 verification — PROMOTED TO READY

Built TSP 1592 (commit `e6e82ff`), deployed to `benchmark_luoyixin_1tb_6cn_2disk2`, ran the
`1_100gb_sortkey` template (20-min test phase, `--skip-setup`).

### Pre-registered acceptance criteria

| Criterion | Target | Measured | |
|---|---|---|---|
| `rebuild_batch_insert_cnt` per cold-load ≈ ceil(rows / 32 768) | ~60 for 2 M rows | **57** for 1 842 525 rows | ✅ |
| `pindex_load_from_lake_tablet_us` P99 first 60 s | ≤ 2 800 ms | **1 937–1 960 ms** | ✅ |
| Gap `lake_persistent_index_insert_us` − `pindex_memtable_insert_us` (wrapper tax) | ≤ 150 ms | **141–146 ms** | ✅ |
| Steady-state real-upsert P99 (server-side) | ≤ 100 ms | **83 ms** | ✅ |
| Throughput | ≥ 41 K r/s | **42 382 r/s** | ✅ |
| Compaction publish max | ≤ 350 ms | `compaction_replace_us` max 401 ms (P99 160 ms); 3 of 309 traces > 500 ms cost | ⚠️ soft miss, unrelated to this patch |

### Headline numbers

- Cold-load `pindex_load_from_lake_tablet_us`: **1 937–1 960 ms** vs iter-007 baseline 3 665–3 945 ms — **~2× faster, ~1.85 s saved per cold-load tablet**.
- Wrapper tax: ~1 070 ms → **141–146 ms** = **7.4× reduction**. Per-call overhead is now amortised across 57 batched calls instead of paid per ~2 048-row chunk.
- Steady-state real-upsert P99: **83 ms** (no regression vs iter-005's 87 ms).
- Throughput: 42 382 r/s (no regression).
- Compaction publish path (which this PR does not touch): `compaction_replace_us` max 401 ms / P99 160 ms — slightly above the iter-005 measurement of 317 ms max, attributable to table growth (~2× rowset size) not to this change.

### Notes on the compaction caveat

The 350 ms compaction-max criterion was a noise-guardrail, not a target this patch was meant to move.
Only 3 of 309 compaction publishes exceeded 500 ms; the worst (775 ms) had `compaction_replace_us` of
only 68 ms — most of its cost is non-replace work (input read / sort / write), unaffected by this PR.

### Operational observation

Two of six CNs received zero publish-version traces during the test window (recurring partial-deploy
FE-routing stickiness seen since iter-004). Surviving CNs were healthy with max 775 ms per publish.
Belongs to the deploy / `upgrade-cn` skill, not this PR.

Promoting to ready-for-review.

luohaha:

PR #73258, opened 2026-05-14
[Enhancement] Add opt-in wide-string column isolation for statistics collection

JasonQu1:
## Why I'm doing:

Statistics collection batches multiple columns into a single SQL. When several wide string columns are batched together, the exchanged/buffered row becomes wide and inflates the per-query memory peak. On a large MV with billions of rows, a sampled statistics SQL that collected wide columns such as `market_text` and `car_model` together showed this amplification:

```sql
WITH base_cte_table AS (
    SELECT * FROM mv TABLET(...) SAMPLE('percent'='1')
    UNION ALL
    SELECT * FROM mv TABLET(...) SAMPLE('percent'='10')
)
SELECT ... hll_raw(market_text), max(left(market_text, 200)), ...
FROM base_cte_table
UNION ALL
SELECT ... hll_raw(car_model), max(left(car_model, 200)), ...
FROM base_cte_table;
```

Under memory pressure (concurrent queries, data cache, etc.), batching several wide string columns into the same statistics SQL raises the per-query peak and makes the statistics SQL a likely victim. This PR provides an **opt-in** knob to relieve that, without changing default behavior.

> Note: this is intentionally a narrow, opt-in hardening. Operators can enable it according to the actual SQL shape and choose an appropriate threshold.

## What I'm doing:

When `statistics_large_string_column_merge_threshold` is set to a **positive** value, `SampleQueryJob.buildQuerySQL` and `FullQueryJob.buildQuerySQL` split columns into two groups:

- **Wide string columns**: each column/partition collection unit is emitted as a dedicated single-column SQL, instead of being merged with other columns by `UNION ALL`.
- **Other columns**: batched as before according to `statistic_collect_parallel`.

A column is considered wide when its type is `CHAR` / `VARCHAR` family and its declared length is greater than the threshold. The check uses `ScalarType.getLength()` instead of `PrimitiveType.getTypeSize()`, because `getTypeSize()` returns a fixed estimate for char-family types and cannot distinguish narrow from wide strings.

## Config (opt-in, disabled by default):

```text
statistics_large_string_column_merge_threshold = 0   # default: disabled
```

- `0` (default): isolation disabled; all columns go through the original merged-batch path. **Upgrading clusters see no behavior change.**
- `> 0`: char-family columns whose declared length exceeds the threshold are isolated one-per-SQL.

> **Caveat on `STRING`**: In StarRocks, `STRING` is internally a maximum-length `VARCHAR` (`varchar(65533)`). So enabling this option with a positive threshold may also isolate `STRING` columns. This is called out in the docs so operators enabling the knob know what to expect.

## Empirical impact (with the knob enabled):

**Environment**: 3-BE dev cluster, single machine.

**Data**: `bench.t_witness`, 10M rows. Three wide varchar columns (`market_text varchar(2048)`, `car_model varchar(1024)`, `long_text varchar(4096)`), plus two narrow varchar columns and two bigint columns.

**Command**: `ANALYZE SAMPLE TABLE bench.t_witness(region, short_attr, market_text, car_model, long_text) WITH SYNC MODE`, 5 runs per cell.

### Unlimited memory

| Metric (5-run mean) | baseline | knob enabled |
|---|---:|---:|
| Wall time | 123.8 s | 121.6 s |
| Outer ANALYZE `MemCostBytes` | 1.44 GB | **1.11 GB (-23%)** |

### Limited memory (`query_mem_limit = 800 MB`, `enable_spill = false`)

| | success rate | failure signal |
|---|---:|---|
| baseline | **0 / 5** | `Used: 839-846 MB, Limit: 800 MB` |
| knob enabled | **5 / 5** | n/a |

Same ANALYZE, same data, same per-query memory budget. The baseline's single 5-column batched inner SQL exceeds 800 MB on every run; with the knob enabled it is split into multiple inner SQLs whose individual peaks fit under the budget.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

> The new config `statistics_large_string_column_merge_threshold` defaults to `0` (disabled), so default behavior is unchanged. The isolation only takes effect when an operator sets a positive value.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4


JasonQu1:
### Empirical impact

**Environment**: 3-BE dev cluster, single machine.

**Data**: `bench.t_witness`, 10M rows.
3 wide varchar cols — `market_text varchar(2048)`, `car_model varchar(1024)`, `long_text varchar(4096)` — plus 2 narrow varchar cols and 2 bigints.

**Command**: `ANALYZE SAMPLE TABLE bench.t_witness(region, short_attr, market_text, car_model, long_text) WITH SYNC MODE`, 5 runs per cell.

#### Unlimited memory — outer ANALYZE memory drops 23 %

| Metric (5-run mean) | baseline | this PR |
|---|---:|---:|
| Wall time | 123.8 s | 121.6 s |
| Outer ANALYZE `MemCostBytes` | 1.44 GB | **1.11 GB (-23 %)** |

#### Limited memory (`query_mem_limit = 800 MB`, `enable_spill = false`) — baseline OOMs, PR runs

| | success rate | failure signal |
|---|---|---|
| baseline | **0 / 5** | `Used: 839-846 MB, Limit: 800 MB` |
| this PR | **5 / 5** | n/a |

Same ANALYZE, same data, same per-query memory budget. The baseline's single 5-column batched inner SQL exceeds 800 MB on every run; this PR splits it into 4 inner SQLs whose individual peaks fit under the budget.

fe/fe-core/src/main/java/com/starrocks/statistic/hyper/SampleQueryJob.java
```diff
@@ -14,16 +14,22 @@
...
 import com.starrocks.statistic.base.ColumnStats;
 import com.starrocks.statistic.base.PartitionSampler;
 import com.starrocks.statistic.sample.SampleInfo;
+import com.starrocks.type.ScalarType;
+import com.starrocks.type.Type;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class SampleQueryJob extends HyperQueryJob {
```
satanson, line 31: Does FullQueryJob   also merge several column stats collecting sql into one?

fe/fe-core/src/main/java/com/starrocks/statistic/hyper/SampleQueryJob.java
```diff
@@ -40,7 +46,21 @@ protected SampleQueryJob(ConnectContext context, long analyzeId, Database db,
...
+        // Note: PrimitiveType.getTypeSize() returns a fixed 16-byte estimate for char-family,
+        // so we use ScalarType.getLength() (the declared length) which actually reflects the
+        // potential row width.
+        long wideThreshold = Config.statistics_wide_string_column_threshold;
+        List<ColumnStats> wideColumns = wideThreshold > 0
+                ? columnStats.stream().filter(s -> isWideStringColumn(s, wideThreshold)).collect(Collectors.toList())
+                : List.of();
+        List<ColumnStats> normalColumns = wideThreshold > 0
+                ? columnStats.stream().filter(s -> !isWideStringColumn(s, wideThreshold)).collect(Collectors.toList())
+                : columnStats;
```
satanson: verbose code, use Collectors.partitionBy instead?

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -2557,6 +2549,16 @@ public class Config extends ConfigBase {
...
 
+    /**
+     * Char-family columns (VARCHAR / CHAR) whose declared length exceeds this threshold are
+     * collected in a dedicated single-column SQL instead of being batched with other columns.
+     * Isolating wide string columns bounds the per-query Exchange memory footprint and prevents
+     * large VARCHAR / TEXT fields from inflating the agg state of co-batched columns.
+     * Set to 0 to disable the isolation and always batch all columns together.
+     */
+    @ConfField(mutable = true)
+    public static long statistics_wide_string_column_threshold = 512;
```
satanson: how about statistics_large_string_column_merge_threshold?

docs/zh/administration/management/FE_parameters/user_query_loading.md
```diff
@@ -804,6 +804,15 @@ ADMIN SET FRONTEND CONFIG ("key" = "value");
 - 描述: 统计信息缓存的更新间隔。
 - 引入版本: -
 
+### `statistics_large_string_column_merge_threshold`
+
+- 默认值: 512
+- 类型: Long
+- 单位: 字节
+- 是否可变: Yes
+- 描述: 在统计信息收集过程中，声明长度超过该阈值的字符列（`VARCHAR` / `CHAR`）会被独立放入单独的统计 SQL 中收集，不与其他列合并在同一条 SQL 里。该行为同时适用于采样统计信息收集和全量统计信息收集。此优化限制了单条统计 SQL 的 Exchange 内存峰值，避免长 `VARCHAR` / `TEXT` 列在与其他列合并时叠加放大聚合算子的状态。副作用是：含较多宽字符列的表会产生更多统计 SQL（每个宽列一条），属于「用少量额外扫描时间换内存峰值」的取舍。设置为 `0` 可关闭该隔离行为，所有列重新走原合并批次的路径。
```
wyb: Add ja doc.

wyb:
```suggestion
- 描述: 在统计信息收集过程中，声明长度超过该阈值的字符列（`VARCHAR` / `CHAR`）会被独立放入单独的统计 SQL 中收集，不与其他列合并在同一条 SQL 里。该行为同时适用于采样统计信息收集和全量统计信息收集。此优化限制了单条统计 SQL 的 Exchange 内存峰值，避免长 `VARCHAR` / `CHAR` 列在与其他列合并时叠加放大聚合算子的状态。副作用是：含较多宽字符列的表会产生更多统计 SQL（每个宽列一条），属于「用少量额外扫描时间换内存峰值」的取舍。设置为 `0` 可关闭该隔离行为，所有列重新走原合并批次的路径。
```

fe/fe-core/src/main/java/com/starrocks/statistic/hyper/SampleQueryJob.java
```diff
@@ -40,7 +43,13 @@ protected SampleQueryJob(ConnectContext context, long analyzeId, Database db,
     protected List<String> buildQuerySQL() {
         int parts = Math.max(1, context.getSessionVariable().getStatisticCollectParallelism());
 
-        List<List<ColumnStats>> partColumns = Lists.partition(columnStats, parts);
+        // Isolate wide string columns to bound per-query Exchange memory; batch the rest.
+        long threshold = Config.statistics_large_string_column_merge_threshold;
+        Map<Boolean, List<ColumnStats>> partitioned = columnStats.stream()
+                .collect(Collectors.partitioningBy(s -> threshold > 0 && isWideStringColumn(s, threshold)));
+        List<ColumnStats> wideColumns = partitioned.get(true);
+        List<ColumnStats> normalColumns = partitioned.get(false);
```
wyb, line 51:
```suggestion
        List<ColumnStats> normalColumns = partitioned.get(false);
        List<List<>> normalPartColumns = Lists.partition(normalColumns, parts);
        for (Long partitionId : partitionIdList) {
            if (!normalColumns.isEmpty()) {
                for (List<ColumnStats> batch : normalPartColumns) {
                    sampleSQLs.add();
                }
            }
        }
```

wyb (dismissed):

satanson (dismissed):

fe/fe-core/src/main/java/com/starrocks/statistic/hyper/SampleQueryJob.java
```diff
@@ -40,7 +43,14 @@ protected SampleQueryJob(ConnectContext context, long analyzeId, Database db,
     protected List<String> buildQuerySQL() {
         int parts = Math.max(1, context.getSessionVariable().getStatisticCollectParallelism());
 
-        List<List<ColumnStats>> partColumns = Lists.partition(columnStats, parts);
+        // Isolate wide string columns to bound per-query Exchange memory; batch the rest.
+        long threshold = Config.statistics_large_string_column_merge_threshold;
```
stephen-shelby, line 47: Before we add a new default-on behavior plus a config knob, I'd like to align on the root cause — I'm not yet convinced column batching is it.

**1. Background statistics collection is serial, so the "concurrent statistics tasks" premise doesn't hold.**

- Cross-task concurrency is bounded by `statistic_analyze_task_pool_size`, which defaults to **3** (`Config.java`, `AnalyzeMgr.ANALYZE_TASK_THREAD_POOL`) — and in practice it's effectively 1 for auto-collect.
- Within a single table it's fully serial: `HyperStatisticsCollectJob` runs `queryJob.queryStatistics()` in a plain `for` loop, and `HyperQueryJob.queryStatistics()` runs each SQL in a `for (String sql : sqlList)` loop. So at any instant there is **one** statistics SQL in flight per table.

Given that, for a single serial stats SQL to OOM a BE node, one of two things must be true — and neither is fixed by batching:
- (a) its own peak is large enough to topple the process on its own → the real problem is that the stats query has **no per-query memory cap and can't spill** (see point 3); or
- (b) it's a *victim* — the process was already near its limit from business queries + data cache, and the serial stats query was just the allocation that tipped it over → then shaving ~25% off it (1.44 GB → 1.11 GB) is marginal.

Could you attach the **BE memory-tracker breakdown / query profile at OOM time**? That would tell us whether the stats query was actually the dominant consumer (cause) or a victim. Without it, the split is a guess.

**3. The repro itself points to a more bounded fix.**

The limited-memory experiment had to set `query_mem_limit = 800 MB` and `enable_spill = false` to reproduce. But the statistics connect context (`StatisticUtils`, where `EnableProfile`, `ConnectorIoTasksPerScanOperator`, `CboCteReuse`, etc. are configured) sets **no per-query memory limit and does not enable spill** — stats queries just inherit whatever global defaults exist, and many clusters run with spill off.

So the principled fix looks like: set `exec_mem_limit` / `query_mem_limit` and enable spill (or a low-memory mode) on the statistics context. That **bounds** memory — a wide query then spills or fails at the query level instead of OOMing the BE — and it covers *every* wide shape (JSON, large `ARRAY<varchar>`, or just many narrow columns batched together), not only the char-family > threshold subset this PR targets. Column-splitting, by contrast, is an unbounded, data-dependent heuristic: a single very wide column, or a wider table/MV than the one benchmarked, will cross the line again.

**4. The default threshold of 512 mis-fires on the common `STRING` type.**

`isWideStringColumn` classifies by declared length via `ScalarType.getLength()`. But in StarRocks `STRING` is `varchar(65533)` (`TypeParser` maps `"STRING"` → `StringType.DEFAULT_STRING`, `DEFAULT_STRING_LENGTH = 65533`). So with the default threshold of 512, **every `STRING` column is classified wide and isolated into its own SQL** — and `STRING` is extremely common, especially in MVs where expression-derived columns often default to the max length.

The benchmark uses explicit `varchar(100/512/2048)`, where only 3 of 7 columns exceed the threshold, which hides this. On a realistic `STRING`-heavy schema, nearly all text columns get split into dedicated SQLs; under serial execution the wall-time regression will be much larger than the "trade-off" section implies. Also note declared length is a poor proxy for runtime memory, which depends on *actual* string width — a `STRING` column holding 8-byte values would be isolated needlessly.

Net: this is a reasonable hardening to reduce a single query's peak, but I'd like to confirm it's actually the cause (point 1) before making it default-on, and I think bounding stats-query memory + spill (point 3) is the more robust, more general fix.

JasonQu1: Thanks for the detailed review. I agree this feature should not be enabled by default.

My intention is to keep this PR as an opt-in optimization knob, not the only or definitive fix for the original field issue. With the default value 0, behavior is unchanged. When a field case shows a statistics SQL using high memory because multiple wide string columns are merged into one SQL, operators can enable this knob and pick a threshold based on the actual SQL shape to reduce the per-query peak.

To actually address the field issue, I also made a separate change for non-group-by aggregate pushdown, which tackles the same memory problem from another direction. So this PR does not need to carry that responsibility on its own — it can stay as a narrow, opt-in optimization that is disabled by default and enabled case by case.

stephen-shelby (approved):

wyb (approved):

HangyuanLiu (approved):

PR #73262, opened 2026-05-14
[auto-perf-opt] Bulk-load PK index memtable on lake cold-load rebuild

luohaha:
## Summary

Replace the per-chunk `_memtable->insert()` chain in
`LakePersistentIndex::load_from_lake_tablet` with a single
*accumulate → sort → bulk-load* pass. For the cold-load case the
memtable starts empty, the input is in random PK order (sort key ≠ PK),
and every emplace into the `phmap::btree_map` is an O(log N)
cache-miss-y tree walk. With sorted input and `_map.insert(end(), …)`,
phmap's hint path collapses to an O(1)-amortized internal_emplace —
total O(N) instead of O(N log N).

## Bottleneck this addresses

Iter-008 measurement on the `1_100gb_sortkey` workload (StarRocks
shared-data, 6 CN, PK ≠ sort key, ~1.84 M rebuild rows / cold-load
tablet):

| metric | value | share |
|---|---:|---:|
| `pindex_load_from_lake_tablet_us` | 1 960 ms | 100 % |
| `pindex_memtable_insert_us` | 1 672 ms | **85 %** |
| wrapper tax (`lake_persistent_index_insert_us − pindex_memtable_insert_us`) | 146 ms | 7 % |
| OSS read + PK encode | ~150 ms | 7 % |

CPU and disk were idle during this — the bottleneck is single-threaded
btree fan-out, not host resources. Per-iteration history is in
`docs/showcase-*.md` of the auto-perf-opt run; iter-008's PR #73256
already amortized the wrapper tax 7.4× (1 070 → 146 ms), exposing the
btree emplace chain as the next dominant cost.

## How it works

1. **Accumulation.** During the rebuild loop, deep-copy each PK into a
   `std::string` and push `(key, IndexValueWithVer)` onto a stack-local
   `std::vector`. Deep copy is needed because the source chunk is
   reused across iterations.
2. **Sort once.** After the loop, `std::sort` the scratch by key.
   Contiguous-vector access pattern beats btree-walk cache misses
   even though both are O(N log N) in comparisons.
3. **Bulk-load.** Call new `PersistentIndexMemtable::bulk_load_sorted_unique()`:
   walks the sorted input and emits `_map.insert(end(), pair)` for each.
   Phmap's `insert_hint_unique(end(), key, ...)` short-circuits to
   `internal_emplace` when the new key > the rightmost — O(1) amortized
   per element.
4. **Defer load_dels.** Per-rowset `load_dels` calls used to interleave
   with inserts. In bulk-load mode they're deferred to after the
   bulk-load. Final index state is identical because `load_dels`'s own
   filter (`found_rssid > del_rebuild_rssid → skip`) already encodes
   the "delete is too old for this key" rule against whatever index
   state `get()` returns.

## Kill switch

`CONF_mBool(lake_pk_index_rebuild_bulk_load_sort, "true")`. Setting it
to `false` restores the per-chunk insert path bit-for-bit, including
in-loop `load_dels` ordering.

## New trace counters

- `rebuild_bulk_load_pair_count` — total `(key, value)` pairs handed
  to the bulk-load.
- `rebuild_bulk_load_sort_us` — `std::sort` wall-clock.
- `pindex_memtable_bulk_load_insert_us` — `bulk_load_sorted_unique`
  wall-clock (sort excluded).

## Expected impact

Reduces per-tablet cold-load wall-clock from ~1 960 ms toward
~700-900 ms (~50 % reduction). This shortens the deploy-time "freshness
hump" measurable as `freshness P99` on the realtime-benchmark
`1_100gb_sortkey` template. Steady-state writes are unaffected — this
code path runs only at tablet metadata load.

## Test plan

- [ ] Build BE on `branch-4.1` with the patch applied. (TSP build for
      the auto-perf-opt run handles this in the next iteration.)
- [ ] Deploy to a 6-CN cluster with the `1_100gb_sortkey` template.
- [ ] Run `realtime_benchmark --template 1_100gb_sortkey
      --skip-setup --duration-minutes 20`.
- [ ] Confirm new trace counters are emitted on cold-load events
      (`grep pindex_memtable_bulk_load_insert_us` in publish-trace logs).
- [ ] Confirm `pindex_load_from_lake_tablet_us` P99 in the first 60 s
      drops from ~1 960 ms to ≤ 1 100 ms.
- [ ] Confirm steady-state real-upsert P99 unchanged (≤ 100 ms
      server-side).
- [ ] Confirm freshness P99 over 20 min drops by ≥ 5 s vs the
      iter-008 baseline.
- [ ] Verify the kill switch: at least one trace with
      `lake_pk_index_rebuild_bulk_load_sort=false` should reproduce the
      pre-patch per-chunk insert pattern (optional — leave for code
      review).

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -1359,6 +1359,15 @@ CONF_mInt32(cloud_native_pk_index_rebuild_files_threshold, "50");
 // If rows which need to be rebuilt exceed this threshold, we will flush memtable immediately
 // to reduce index rebuild cost. 0 means disabled.
 CONF_mInt64(cloud_native_pk_index_rebuild_rows_threshold, "10000000");
+// If true, the lake PK-index cold-load rebuild path accumulates all
+// (key, IndexValueWithVer) pairs into a scratch vector, sorts them once at the
+// end of the rebuild loop, and bulk-inserts them into the memtable using
+// end-hint inserts (O(1) amortized per element for strictly-sorted input).
+// This replaces the per-chunk `_memtable->insert()` call site, which was the
+// single-threaded `phmap::btree_map::emplace()` chain that dominated cold-load
+// wall-clock (~85% on the `1_100gb_sortkey` workload at iter-008 measurement).
+// Set to false to fall back to the per-chunk insert path bit-for-bit.
+CONF_mBool(lake_pk_index_rebuild_bulk_load_sort, "true");
 // if set to true, CACHE SELECT will only read file, save CPU time
 // if set to false, CACHE SELECT will behave like SELECT
 CONF_mBool(lake_cache_select_in_physical_way, "true");
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -1139,6 +1139,19 @@ Status LakePersistentIndex::load_from_lake_tablet(TabletManager* tablet_mgr, con
     int64_t get_next_cost_us = 0;
     int64_t pk_encode_cost_us = 0;
     int64_t build_values_cost_us = 0;
+    // Scratch buffer for the bulk-load path. The per-chunk `insert()` chain
+    // dominates cold-load wall-clock (~85% on the 1_100gb_sortkey workload)
+    // because every emplace into an initially-empty `phmap::btree_map` is
+    // O(log N) with random-PK keys. We accumulate the keys here, sort once at
+    // the end, and call `_memtable->bulk_load_sorted_unique()`, which uses
+    // end-hint inserts that are O(1) amortized for sorted input.
+    const bool bulk_load_mode = config::lake_pk_index_rebuild_bulk_load_sort;
+    std::vector<std::pair<std::string, IndexValueWithVer>> scratch;
+    if (bulk_load_mode && _need_rebuild_row_cnt > 0) {
+        scratch.reserve(_need_rebuild_row_cnt);
+    }
+    // Defer del-file processing in bulk-load mode (see the call site below).
+    std::vector<std::pair<RowsetPtr, int64_t>> deferred_dels;
     // Rowset whose version is between max_sstable_version and base_version should be recovered.
     for (auto& rowset : rowsets) {
         TRACE_COUNTER_INCREMENT("total_segment_cnt", rowset->num_segments());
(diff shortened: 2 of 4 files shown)


luohaha:
Closing — iter-010 verification missed 3/4 hard acceptance criteria from iter-009 pre-registration.

**Cluster state:** TSP build 1595, commit d517f39, deployed alone on `benchmark_luoyixin_1tb_6cn_2disk2`. All 6 CNs on `branch-4.1-d517f39`. Branched off run baseline `8813309` — NOT stacked on #73256, so this is a strict A/B against baseline.

**Bulk-load path activation:** confirmed working as designed.
- `rebuild_bulk_load_pair_count == rebuild_index_num_rows`: 4/4 cold-load events. ✓
- 4 real cold-load events at 16:09:58 right after CN restart, ~1.74M rows / 1 segment each.

**Per-cold-load measurements (n=4, ~1.74M rows):**
- `pindex_load_from_lake_tablet_us` = 1660-1715 ms (target ≤ 1200 ms → **MISS by 43%**)
- `rebuild_bulk_load_sort_us` = 930-959 ms (target ≤ 500 ms → **MISS by 90%**)
- `pindex_memtable_bulk_load_insert_us` = 454-472 ms (target ≤ 300 ms → **MISS by 55%**)

**Root cause of miss:** `std::sort` on ~1.75M variable-length PK strings is the dominant cost (~550 ns/row, dominated by string comparisons and cache misses). The end-hint insert path itself works — bulk_insert phase is close to target. The design under-budgeted the sort phase.

**Per-row throughput vs already-promoted #73256:**
- iter-008 (#73256 alone, batched per-chunk insert): 1937-1960 ms / 1.84M rows ≈ **1057 ns/row**
- iter-010 (#73262 alone, sort + end-hint bulk_load): 1660-1715 ms / 1.75M rows ≈ **960 ns/row**

That's a ~9 % per-row improvement, well below the predicted ~50 %. Macro freshness P99 (13.7s vs iter-008's 22.3s) suggests the cold-load tail is meaningfully reduced, but does not by itself justify replacing the lighter #73256 patch.

**Next direction (iter-011):** attack the sort phase — either parallel sort (pdqsort / `std::execution::par_unseq`) or skip the in-memory memtable entirely and build the sstable directly from the sorted scratch buffer. The current design's structural ceiling is dominated by sort, not by btree insert.

PR #61410, opened 2025-07-29
[Enhancement] optimize tablet meta copy when enable file bundling 

luohaha:
## Why I'm doing:
When file bundling is enabled, we need to collect tablet meta and assemble them into a bundled meta. To speed up this process, we need to accelerate the copy operations involved.

## What I'm doing:
This pull request optimizes metadata handling in the `LakeServiceImpl::publish_version` method and improves test coverage for the `LakeServiceTest` suite. Key changes include reducing lock contention by moving metadata operations outside of critical sections, enhancing efficiency with `Swap` to avoid unnecessary copies, and adding assertions to validate tablet metadata in tests.

### Metadata handling optimizations:
* [`be/src/service/service_be/lake_service.cpp`](diffhunk://#diff-f7d8833524add42c8ba5cc5ffb982686e7056297f84b684fc52c65c61306cfcaR246-R247): Introduced a preallocated `TabletMetadataPB` pointer to defer metadata copying outside the critical section, reducing lock contention and enabling parallel execution. [[1]](diffhunk://#diff-f7d8833524add42c8ba5cc5ffb982686e7056297f84b684fc52c65c61306cfcaR246-R247) [[2]](diffhunk://#diff-f7d8833524add42c8ba5cc5ffb982686e7056297f84b684fc52c65c61306cfcaL257-R264)
* [`be/src/service/service_be/lake_service.cpp`](diffhunk://#diff-f7d8833524add42c8ba5cc5ffb982686e7056297f84b684fc52c65c61306cfcaL353-R361): Replaced `emplace` with `Swap` in `AggregatePublishContext` to avoid unnecessary metadata copying, improving performance.

### Test enhancements:
* [`be/test/service/lake_service_test.cpp`](diffhunk://#diff-7e7546a285303649b12ec94135003c7c099a4b032ebe2ae95a044ea791386d5fR2618-R2630): Added assertions in `test_aggregate_publish_version` to validate tablet metadata correctness for multiple tablet IDs, improving test coverage and reliability.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
[...]

PR #73274, opened 2026-05-14
[BugFix] Fix race on response->tablet_metas in lake publish_version

xiangguangyxg:
## Why I'm doing:

`LakeServiceImpl::publish_version` crashes the CN under file-bundling
publish (`enable_aggregate_publish=true` → `skip_write_tablet_metadata`)
when two concurrent `PublishTabletInfo` tasks resolve to the same
`metadata->id()`. Crash signature on a production branch-4.1 core
(build a1daea6):

```
RowsetMetadataPB::IsInitialized() this=0x276d1ad1   ← torn pointer
TabletMetadataPB::IsInitialized()
PublishVersionResponse::IsInitialized()
brpc::policy::SendRpcResponse(...)
LakeServiceImpl::publish_version(...)
```

The `this=0x276d1ad1` is a torn 64-bit pointer (high 32 bits zeroed)
inside `tablet_metas[child].rowsets_.rep_->elements[i]`. The same core
shows `rowsets_.current_size_ = 146` vs `total_size_ = 73` for both
children — exactly 2× the original allocation, the unmistakable
signature of two threads concurrently `Add()`-ing into the same
`RepeatedPtrField`.

#### BE — `lake_service.cpp`

`lake_service.cpp:222-267` expands `publish_tablet_infos` from
`request->tablet_ids` and from each `resharding_tablet_info` without
dedup. A SPLITTING op whose `new_tablet_ids` overlap with `tablet_ids`
produces two `PublishTabletInfo` per child (`PUBLISH_NORMAL` plus
`SPLITTING_TABLET` cross-publish):

| # | type | tablet_id_in_metadata |
|---|---|---|
| 1 | PUBLISH_NORMAL | child_A |
| 2 | PUBLISH_NORMAL | child_B |
| 3 | SPLITTING_TABLET(old → child_A) | child_A |
| 4 | SPLITTING_TABLET(old → child_B) | child_B |

The per-tablet defense `tablet_txns` in `transactions.cpp` only covers
the body of `lake::publish_version` — its `DeferOp` releases on return
BEFORE the caller's `prealloc_metadata->CopyFrom(*metadata)` runs
outside `response_mtx`:

```
Task A: lake::publish_version(T) acquires tablet_txns[T]
Task A: produces metadata, DeferOp releases tablet_txns[T]   ←
Task A: under response_mtx: P = &map[T]; releases response_mtx
Task A: P->CopyFrom(*metadata)              [outside both locks]
                                                       Task B: lake::publish_version(T) re-acquires tablet_txns[T]
                                                       Task B: under response_mtx: same P = &map[T]
                                                       Task B: P->CopyFrom(*metadata)  ← CONCURRENT WITH TASK A
```

`CopyFrom = Clear + MergeFrom`. `Clear` walks `rowsets_.elements[]`
while the other thread's `MergeFrom` reallocates that array → torn
pointer → SIGSEGV in `RowsetMetadataPB::IsInitialized` at serialization
time. Introduced by #61410 (`75854adf72`, "optimize tablet meta copy
when enable file bundling"); the resharding branch in the same file
already implements the correct swap-under-lock pattern.

#### FE — `PublishVersionDaemon.publishPartitionBatch`

The crash core's request body:

```
tablet_ids                 = [child_A, child_B]
resharding_tablet_infos[0] = SPLITTING(old=77599, new_tablet_ids=[child_A, child_B])
txn_infos                  = 5 transactions, commit_time span ≈ 102 seconds
base_version=70, new_version=75, enable_aggregate_publish=true
```

`publishPartitionBatch` unions per-txn loaded materialized indexes
across **every** `TransactionState` in the batch. Each txn's
`loadedIndexIds` was snapshotted at OlapTableSink planning time. If
`SplitTabletJob.addNewMaterializedIndexes()` runs in the middle of a
batch's planning window, early txns capture `[oldIndexId]` and late
txns capture `[newIndexId]`. The union then contains:

- `77599` (early txns' OLD tablet, still resolvable until
  `removeOldMaterializedIndexes` in CLEANING)
- children (late txns', from the new index)

`Utils.processTablets` then emits both lists in the same RPC,
producing the dup-id input that triggers the BE race.

## What I'm doing

Two-layer defense; both layers necessary. BE stops the crash regardless
of caller. FE removes the production trigger so the underlying
data-correctness gap (two tasks producing different partial metadata
for the same child) does not occur in the first place.

### BE: swap-under-lock (`be/src/service/service_be/lake_service.cpp`)

Build a thread-local `TabletMetadataPB` outside `response_mtx`, then
`Swap` it into `response->tablet_metas[id]` under the lock. The heavy
`CopyFrom` still runs in parallel; the in-lock `Swap` is fast and
atomic w.r.t. other tasks. Two tasks targeting the same id will Swap
serially under the lock — last writer wins, no torn array. Mirrors the
safe pattern in the resharding branch.

### FE: strict per-partition snapshot invariant (`fe/fe-core/.../transaction/DatabaseTransactionMgr.java`)

In `getReadyToPublishTxnListBatch`, fold the existing
version-must-be-consecutive check and a new
loaded-materialized-index-id-snapshot-must-be-equal check into a single
per-partition `PartitionCommitState` record. Break the batch on either
invariant violation. Any divergence (including schema-change shadow
addition/removal) is a cut point — keeps the invariant simple and
conservative, doesn't depend on knowing whether the diverging txns
came from a split or schema change.

A new accessor
`TransactionState.getPartitionLoadedIndexIdsWithoutLock(tableId,
physicalPartitionId)` exposes the raw recorded id list. Sequential
sub-batches are correct because the BE's
`metacache().lookup_tablet_metadata` makes already-published
`(tablet_id, version)` pairs idempotent.

### Test

`LakePublishBatchTest.testBatchPublishShadowIndex` previously relied on
three out-of-order commits (txn2, txn1, txn3) batching into one publish
whose shadow `PublishLogVersionBatchRequest` contained both txn2 and
txn3. The strict cut now separates txn1 (snapshot lacks the shadow id)
from txn2/txn3, producing one single-txn batch per shadow-loaded txn.
With the class-level `Config.lake_batch_publish_min_version_num = 2`
(test-only stress; production default is 1), `TransactionGraph` refuses
chains of size 1 and `awaitPublish` times out. The test scopes
`min_version_num = 1` to this method and expects two separate
`PublishLogVersionBatchRequest` entries (one per shadow-loaded txn).

### Empirical repro

A standalone test modeling `acquire → produce → release → CopyFrom
outside lock` with 16 threads × 500 ops × 32 IDs (heavy dup-id
pressure): **5/5 SIGSEGV** with the buggy BE pattern, **0/5** with the
swap-under-lock fix.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/service/service_be/lake_service.cpp
+++ b/be/src/service/service_be/lake_service.cpp
@@ -388,7 +388,11 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
                     if (res.ok()) {
                         auto metadata = std::move(res).value();
                         auto score = compaction_score(_tablet_mgr, metadata);
-                        TabletMetadataPB* prealloc_metadata = nullptr;
+                        // Copy metadata out of the lock(response_mtx), to let it execute in parallel.
+                        TabletMetadataPB local_metadata;
+                        if (skip_write_tablet_metadata) {
+                            local_metadata.CopyFrom(*metadata);
+                        }
                         {
                             std::lock_guard l(response_mtx);
                             response->mutable_compaction_scores()->insert({metadata->id(), score});
@@ -400,13 +404,8 @@ void LakeServiceImpl::publish_version(::google::protobuf::RpcController* control
                                 response->mutable_tablet_row_nums()->insert({metadata->id(), row_nums});
                             }
                             if (skip_write_tablet_metadata) {
-                                auto& map = *response->mutable_tablet_metas();
-                                prealloc_metadata = &map[metadata->id()];
+                                (*response->mutable_tablet_metas())[metadata->id()].Swap(&local_metadata);
                             }
-                        }
-                        // Move copy metadata out of the lock(response_mtx), to let it execute in parallel.
-                        if (prealloc_metadata != nullptr) {
-                            prealloc_metadata->CopyFrom(*metadata);
                         }
                         // Report tablet for async VI build only on async-mode tables
                         // whose new rowset(s) have vector_index_ids. Sync-mode tables
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/DatabaseTransactionMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/DatabaseTransactionMgr.java
@@ -934,6 +934,12 @@ public class DatabaseTransactionMgr {
         }
     }
 
+    // Per-partition state carried across TransactionStates while building a
+    // TransactionStateBatch: previous version (must stay consecutive) and the
+    // materialized-index id snapshot (must stay identical across the batch).
+    private record PartitionCommitState(long version, List<Long> loadedIndexIds) {
+    }
+
     public List<TransactionStateBatch> getReadyToPublishTxnListBatch() {
         List<TransactionStateBatch> result = new ArrayList<>();
         readLock();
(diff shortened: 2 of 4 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #73287, opened 2026-05-14
[Enhancement] Add AVX2/NEON SIMD primitives for analytic workloads

### LINKED ISSUE #73294: [Enhancement] AVX2/NEON SIMD coverage for analytic hot paths
## Enhancement

loops.

### Stacked PRs

* #73289 — column operations
* #73292 — storage layer
* #73293 — execution layer

eshishki:
## Why I'm doing:

Foundation helpers for SIMD-vectorising hot paths across parquet decoding,
storage scans, predicate evaluation, aggregates, column operations and the
execution layer. Splits the existing AVX-512-only path with AVX2 (Intel hosts
without AVX-512) and NEON (AArch64) coverage so the follow-up stacked PRs
share a portable set of primitives.

Fixes #73294

## What I'm doing:

* `base/simd/expand.cpp` — AVX2 LUT-based and NEON `expand_load_simd`
  alongside the existing AVX-512VLBW path. Header documents the
  `src_data` over-read precondition at trailing nulls.
* `base/simd/multi_version.h` — `MFV_*` macros switched to variadic so
  callers can pass function bodies that contain top-level commas
  (aggregate-initialiser LUT tables). Adds `MFV_AUTOVEC_STATIC` for
  stand-alone scalar fallbacks: `MFV_DEFAULT`'s `target("default")`
  attribute was overriding the build's `-m` flags and emitting SSE2-only
  code for `simd_*_default` helpers, making the "scalar fallback" slower
  than an undecorated loop on USE_AVX512=OFF. The new macro inherits the
  build's `-m` flags so the body auto-vectorises correctly.
* `base/simd/simd.h` — add `all_zeros`/`all_ones` (pointer + container)
  on top of the existing 8-bit primitives. `all_ones` uses `std::memchr`
  which is SIMD-optimised in modern libcs.
* `base/simd/simd_utils.h` — `SIMDUtils::simd_fill<T>` is a thin
  `std::fill_n` forwarder (compiler emits the same broadcast-store the
  hand-rolled loop produced). `set_data` replaces `reinterpret_cast` of
  `float*` / `double*` to integer pointers with `memcpy` to fix
  strict-aliasing UB.
* `base/simd/rle_simd.h` + `base/simd/rle_simd.cpp` (new) —
  `simd_fill_int16/int32` are inline `std::fill_n` wrappers;
  `simd_minmax_int32`, `simd_dict_gather_int32/int64/float/double`,
  `simd_widen_int8/int16_to_int32` are declared in the header, defined
  in the .cpp so callers don't pull `<immintrin.h>` transitively. The
  `int16` fill correctly targets AVX-512BW (`set1_epi16` requires BW,
  not just F). `simd_dict_gather_float/double` use typed gather
  intrinsics (`_mm256_i32gather_ps/pd`) instead of pointer-punning
  `float*` to `int32_t*`. `simd_minmax_int32_default` reduces into
  locals so the compiler can auto-vectorise (writing through the
  `out_min`/`out_max` reference parameters previously blocked it).
* `base/simd/string_length_filter.h` (new) — `length_eq_mask` and
  `length_in_range_mask` for parquet PLAIN binary length filtering.
  The range variant uses the AVX2 saturate-and-equal idiom
  (`_mm256_max_epu32` / `_mm256_min_epu32`) so it stays correct for
  `min_len == 0` and `max_len == UINT32_MAX`.

`simd_dict_gather_int32/64` restored with AVX2 implementations. The
dispatcher prefers AVX2 over the AVX-512 hand path on x86: bench data
below shows AVX2 `vpgatherdd ymm` beats both auto-vectorised scalar and
AVX-512 `vpgatherdd zmm` on Intel Ice Lake.

These are foundation helpers; observable behaviour lives in the
consuming stacked PRs.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

## Performance

Microbench scalar vs SIMD, 5 reps with `--benchmark_report_aggregates_only=true`,
AWS `m6i.4xlarge` (Intel Ice Lake), via the `simd_common_bench` target.

|---|---|---|
| `BM_AllZeros /4096` | 2056 → 56.8 ns (**36.2x**) | 2051 → 57.1 ns (**35.9x**) |
| `BM_AllOnes /4096` | 2052 → 30.6 ns (**67.1x**) | 2048 → 30.6 ns (**66.9x**) |
| `BM_WidenInt8 /256` | 20.8 → 14.1 ns (**1.48x**) | 10.7 → 10.7 ns (1.0x*) |
| `BM_WidenInt8 /4096` | 333 → 214 ns (**1.56x**) | 205 → 205 ns (1.0x*) |
| `BM_DictGather /256` | 98.9 → 57.5 ns (**1.72x** AVX2) | 96.8 → 57.8 ns (**1.67x** AVX2)  (AVX-512 86.4 — slower) |
| `BM_DictGather /1024` | 505 → 247 ns (**2.04x** AVX2) | 366 → 235 ns (**1.56x** AVX2)  (AVX-512 346 — slower) |
| `BM_DictGather /4096` | 2377 → 1049 ns (**2.27x** AVX2) | 2374 → 1052 ns (**2.26x** AVX2)  (AVX-512 1484 — slower) |
| `BM_LengthEqMask /256` | 91.6 → 21.9 ns (**4.18x**) | 57.6 → 22.2 ns (**2.59x**) |
| `BM_LengthEqMask /1024` | 313 → 92.4 ns (**3.39x**) | 128 → 91.5 ns (**1.40x**) |
| `BM_LengthEqMask /4096` | 1197 → 383 ns (**3.13x**) | 406 → 375 ns (**1.08x**) |

\* `simd_widen_int*_to_int32` is gated on `!__AVX512F__`: on AVX-512
builds the compiler auto-vectorises the scalar fallback to a 16-lane
AVX-512 widen that matches or beats the hand-written 8-lane AVX2 path,
so the dispatcher routes through the scalar default and both bench rows
end up running identical autovec'd code. Hand-AVX2 stays the active
path on AVX2-only builds where this wins 1.48-1.56x.

`simd_minmax_int32` keeps its AVX-512F hand path (16-lane `_mm512_reduce_min/max`)
because the auto-vectorised default underuses the wider lanes; AVX2/NEON were
removed earlier when microbench showed parity with the scalar default.


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/base/simd/expand.cpp
```diff
@@ -42,6 +120,58 @@ MFV_AVX512VLBW(void expand_load_selection_i32(int32_t* dst_data, const int32_t*
...
                                            size_t count) { expand_load_branchless(dst_data, src_data, nulls, count); });
 
+MFV_AVX2(void expand_load_selection_i64(int64_t* dst_data, const int64_t* src_data, const uint8_t* nulls,
+                                        size_t count) {
+    // 4-bit mask -> 8 x int32 permutation indices that treat the 256-bit
+    // register as four 64-bit lanes. NB: the permute is done via
+    // _mm256_permutevar8x32_epi32 because there is no native epi64 permute on AVX2.
+    static const int32_t kPerm64LUT[16][8] = {
+            {0, 1, 0, 1, 0, 1, 0, 1}, // 0b0000
+            {0, 1, 0, 1, 0, 1, 0, 1}, // 0b0001
```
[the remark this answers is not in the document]
eshishki: Not a bug — `expand_load`'s contract is "valid lanes only; null lanes are unspecified scratch". The existing UT (`be/test/base/simd/simd_expand_test.cpp:71-75,101-105,123-127`) explicitly only compares non-null positions:
```cpp
if (nulls[i] == 0) {
    ASSERT_EQ(dsts1[i], dsts2[i]);
}
```

All implementations diverge at null lanes by design:
- scalar: leftover next-valid (`dst[i]=src[cnt]`, cnt unchanged)
- AVX2 i32 (`kExpandPermuteLUT`): also stores `0` at null positions, mask `0b00000001` → `[src0, src0, src0, src0, src0, src0, src0, src0]` (not `[src0, src1, src1, src1, ...]`)
- AVX2 i64 (`kPerm64LUT`): same — `0` at null positions
- AVX-512: `_mm512_maskz_expandloadu_epi32` zeroes null lanes

Only caller is `encoding_dict.h:assign_data_with_nulls` which writes into a `NullableColumn`'s `data_column` — null-lane values are always masked away by the parent `NullableColumn::null_column`.

The one-line comment in `expand.h:24-25` describes the reference branchless implementation, not the contract. I'll leave the code as-is.

be/src/base/simd/rle_simd.h
```diff
@@ -0,0 +1,351 @@
...
+
+namespace starrocks {
+
+// ============================================================================
+// SIMD-optimized fill operation for RLE repeated values (int16_t)
+// Used for Parquet level decoding (def_level, rep_level are int16_t)
+// ============================================================================
+
+#if defined(__AVX512BW__)
+MFV_AVX512BW(void simd_fill_int16_avx512(int16_t* __restrict dst, int16_t value, int32_t count) {
```
stdpain: Put the implemention to cpp files.

be/src/base/simd/rle_simd.h
```diff
@@ -0,0 +1,351 @@
...
+        dst[i] = value;
+    }
+})
+#endif
+
+MFV_DEFAULT(void simd_fill_int16_default(int16_t* __restrict dst, int16_t value, int32_t count) {
+    for (int32_t i = 0; i < count; ++i) {
+        dst[i] = value;
+    }
+})
```
stdpain: prefer using std::fill_n for AVX2 https://godbolt.org/z/xsbb6hbj9

be/src/base/simd/simd_utils.h
```diff
@@ -32,10 +41,92 @@ class SIMDUtils {
...
         }
     }
-};
-} // namespace starrocks
+#endif
+
+    // SIMD-optimized fill: fills array with repeated value
+    // Uses AVX2 when available, falls back to NEON on aarch64, else scalar loop.
+    template <class T>
+    static void simd_fill(T* dest, T value, size_t count) {
```
stdpain, line 58: std::fill_n is enough

eshishki:
@stdpain thanks for review, i am doing benchmarks with avx512 on/off to pin down regressions. will update code and numbers later

be/src/base/simd/rle_simd.h
```diff
@@ -0,0 +1,351 @@
...
+#endif
+
+MFV_DEFAULT(void simd_minmax_int32_default(const int32_t* __restrict data, int32_t count, int32_t& out_min,
+                                           int32_t& out_max) {
+    out_min = std::numeric_limits<int32_t>::max();
+    out_max = std::numeric_limits<int32_t>::min();
+    for (int32_t i = 0; i < count; ++i) {
+        out_min = std::min(out_min, data[i]);
+        out_max = std::max(out_max, data[i]);
+    }
```
stdpain: https://godbolt.org/z/1fdc3djoa use a local variable to auto-vectorized

```suggestion
    int32_t local_min = std::numeric_limits<int32_t>::max();
    int32_t local_max = std::numeric_limits<int32_t>::min();
    for (int32_t i = 0; i < count; ++i) {
        local_min = std::min(out_min, data[i]);
        local_max = std::max(out_max, data[i]);
    }
    out_min = local_min;
    out_max = local_max;
```

be/src/base/simd/rle_simd.h
```diff
@@ -0,0 +1,351 @@
...
+                                                const uint32_t* __restrict indices, int32_t count) {
+    for (int32_t i = 0; i < count; ++i) {
+        dest[i] = dict[indices[i]];
+    }
+})
+
+// AVX2 vpgatherdd was removed: on modern Intel after the post-Spectre
+// microcode update the gather throughput was lowered to the point where
+// the AVX2 path is ~25% slower than the scalar default. AVX-512F gather
+// has its own scheduler and stays a win, so it is retained.
```
stdpain: AVX2 gather works fine on AWS instances. I believe your processor has an additional microcode patch applied to circumvent this vulnerability. I previously encountered this issue on some Aliyun Cloud instances. You can add a configuration to control the behavior of AVX2. https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/technical-documentation/gather-data-sampling.html

be/src/base/simd/rle_simd.h
eshishki: Done. New `be/src/base/simd/rle_simd.cpp` carries all hand-SIMD bodies (MFV_AVX2/AVX512F) and dispatcher definitions. Header keeps only declarations, the `std::fill_n` inline wrappers, and the `MFV_AUTOVEC_STATIC` scalar fallbacks. See commit e1bbd4a27ae.

be/src/base/simd/rle_simd.h
eshishki: Done. `simd_fill_int16` / `simd_fill_int32` are now inline `std::fill_n` wrappers — compiler emits the same AVX2/AVX-512 broadcast-store the hand-rolled loop produced.

be/src/base/simd/simd_utils.h
eshishki: Done. `SIMDUtils::simd_fill<T>` is a thin `std::fill_n` forwarder. The `set_data` layer above it bit-casts `float`/`double` via `memcpy` to avoid strict-aliasing UB (Copilot caught this in #73292).

be/src/base/simd/rle_simd.h
eshishki: Done. `simd_minmax_int32_default` reduces into local `int32_t` and writes through the reference parameters at the end; compiler auto-vectorises the local-var reduction. Kept the `MFV_AVX512F` hand-path — `_mm512_reduce_min/max_epi32` still beats the autovec output at AVX-512 width — and dropped the AVX2/NEON hand variants which microbenched at parity with the autovec scalar.

be/src/base/simd/rle_simd.h
eshishki: Thanks for the pointer. Restored `simd_dict_gather_int32/64_avx2`; the dispatcher prefers AVX2 on x86 (including USE_AVX512=ON builds). Bench on m6i.4xlarge: AVX2 `vpgatherdd ymm` 247 ns @ /1024 vs AVX-512 `vpgatherdd zmm` 346 ns vs autovec scalar 505 ns. Inline comment in `rle_simd.cpp` notes that hosts with the GDS microcode mitigation (some Aliyun Intel SKUs) should flip the dispatch to `_default`; happy to add a runtime config later if you'd like.

stdpain (approved):

kangkaisen (approved):

PR #73288, opened 2026-05-14
[Enhancement] Apply SIMD primitives to parquet encoding/decoding pipeline

### LINKED ISSUE #73294: [Enhancement] AVX2/NEON SIMD coverage for analytic hot paths

eshishki:
## Why I'm doing:

Vectorises hot paths in the parquet reader/writer. AVX2 + NEON coverage;
AVX-512 stays via the existing multi-versioned helpers in `base/simd`.

Fixes #73294

## What I'm doing:

* `base/bit/bit_packing_neon.h` (new) — AArch64 NEON `UnpackValues` for
  bit widths 1..32, parallel to the existing `bit_packing_avx2.h`.
  Routed through `BitPacking::UnpackValues` so RLE_DICTIONARY index
  decoding hits it on ARM. Forward-declares the primary unpack template
  so the explicit specialisations compile cleanly.
* `base/bit/rle_encoding.h` — `RleDecoder` / `RleBatchDecoder` use
  `std::fill_n` directly for repeated-value runs (compiler emits the
  AVX2/AVX-512 broadcast-store, equivalent to the simd_fill_int*
  wrappers but without the strict-aliasing pointer pun when `T` is
  `float`). `IndicesInRange` and `GetBatchWithDict` literal-gather call
  the new `simd_minmax<T>` / `simd_dict_gather<DstT,SrcT,IdxT>` template
  wrappers in `base/simd/rle_simd.h`, so the call sites stay free of
  size dispatch.
* `formats/parquet/encoding_dict.h` — anonymous-namespace
  `indices_out_of_bounds()` collapses four ad-hoc bounds-check loops into
  one SIMD min/max call. Stack-overflow-prone VLAs replaced with member
  `std::vector` buffers. Tail of `assign_data_with_nulls` now guards the
  load so a trailing-nulls batch does not read past the last valid
  source element.
* `formats/parquet/encoding_plain.h` — VLA → member vectors;
  `reserve()+data()` (UB on uninitialised storage) → `resize()`; the
  clang-specific `__builtin_shufflevector` replaced with portable AVX2
  `unpacklo`/`unpackhi` + `permute2x128` to materialise `Slice{ptr,len}`
  pairs.
* `formats/parquet/encoding_bss.h` — same `reserve`→`resize` fix.
* `formats/parquet/parquet_block_split_bloom_filter.cpp` — AVX2 path for
  `add_hash` (one 256-bit OR replaces the 8-iter loop) and `test_hash`
  (`_mm256_testc_si256`: CF set iff every mask lane is also set in the
  block lane).
* `formats/parquet/column_converter.cpp` —
  `Int32ToDateConverter::convert` treats `dst_data` as a flat int32
  stride and adds `UNIX_EPOCH_JULIAN` with AVX2 (8/iter) or NEON
  (4/iter).
* `formats/parquet/level_builder.cpp` — `_write_int_column_chunk` uses
  `simd_widen_int8/16_to_int32` for the common parquet widening cases;
  `_make_null_bitset` materialises the bitset via cmpeq+movemask (AVX2)
  or pairwise-add reduction (NEON).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

## Performance

Microbench scalar vs SIMD, 5 reps with `--benchmark_report_aggregates_only=true`,
AWS `m6i.4xlarge` (Intel Ice Lake), via the `parquet_simd_bench` target.

|---|---|---|
| `BM_BloomAddHash /256` | 740 → 199 ns (**3.72x**) | 755 → 191 ns (**3.95x**) |
| `BM_BloomAddHash /1024` | 3106 → 891 ns (**3.49x**) | 3216 → 889 ns (**3.62x**) |
| `BM_BloomAddHash /8192` | 25408 → 7757 ns (**3.27x**) | 26132 → 8440 ns (**3.10x**) |
| `BM_BloomTestHash /4` (early-exit lane 4) | 1.49 → 0.49 ns (**3.04x**) | 1.50 → 0.49 ns (**3.06x**) |
| `BM_BloomTestHash /7` (early-exit lane 7) | 2.76 → 0.49 ns (**5.63x**) | 2.93 → 0.49 ns (**5.98x**) |
| `BM_BloomTestHash /8` (all-pass, typical) | 2.75 → 0.49 ns (**5.61x**) | 2.81 → 0.49 ns (**5.73x**) |

`BM_BloomTestHash /0` is a degenerate scalar best case (~0.43 ns where the
first AND lane fails and the loop exits immediately). SIMD has no early exit
by design and stays at ~0.49 ns. Real bloom probes typically run all 8 lanes
(either all match or only the last fails), so the SIMD win matches the /4–/8
columns above.


CelerData-Reviewer:
@codex review

be/src/base/simd/rle_simd.h
```diff
@@ -0,0 +1,569 @@
...
+        dst[i] = value;
+    }
+}
+#endif
+
+MFV_DEFAULT(void simd_fill_int16_default(int16_t* __restrict dst, int16_t value, int32_t count) {
+    for (int32_t i = 0; i < count; ++i) {
+        dst[i] = value;
+    }
+})
```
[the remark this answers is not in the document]
eshishki: Fixed in `ea69d1f58e0` (simd-common base). Updated the non-x86 fallback in `base/simd/multi_version.h` to `#define MFV_DEFAULT(...) static __VA_ARGS__`, mirroring the x86 `MFV_IMPL` linkage. Existing in-header users (`delta_decode.h` and now `rle_simd.h`) get internal-linkage definitions on aarch64 / other targets too, so each TU keeps its own copy and the linker stays happy.

CelerData-Reviewer:
@codex review

be/src/formats/parquet/parquet_block_split_bloom_filter.cpp
```diff
@@ -45,12 +57,24 @@ bool ParquetBlockSplitBloomFilter::test_hash(uint64_t hash) const {
...
+
+#ifdef __AVX2__
+    // Vectorised hash test: AND the 8 mask lanes with the 8 block lanes; if any
+    // lane is zero the hash cannot be present. cmpeq+movemask gives us a bitmask
+    // of zero lanes; non-zero result means the hash is missing.
+    __m256i v_block = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(block_offset));
+    __m256i v_masks = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(masks));
+    __m256i v_and = _mm256_and_si256(v_block, v_masks);
+    __m256i v_zero = _mm256_cmpeq_epi32(v_and, _mm256_setzero_si256());
+    return _mm256_movemask_ps(_mm256_castsi256_ps(v_zero)) == 0;
```
stdpain: use test to avoid move mask costs
```
__m256i v_block = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(block_offset));
__m256i v_masks = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(masks));
return _mm256_testc_si256(v_block, v_masks);

File: be/src/formats/parquet/stored_column_reader.cpp [hunk 3]
```diff
@@ -298,9 +304,32 @@ Status StoredColumnReader::create(const ColumnReaderOptions& opts, const Parquet
...
+        uint16x8_t v_cmp = vceqq_s16(v_data, v_max);
+        uint16x8_t v_ones = vshrq_n_u16(v_cmp, 15);
+        count += vaddvq_u16(v_ones);
+    }
+#endif
+
+    for (; i < num_parsed_levels; ++i) {
+        if (def_levels[i] == max_def_level) {
             count++;
         }
```
stdpain: auto vectorized loop in gcc 14 https://godbolt.org/z/o81f9x8je

be/src/base/bit/rle_encoding.h
```diff
@@ -965,14 +966,23 @@ static inline bool IndexInRange(T idx, int32_t dictionary_length) {
...
+        return IndexInRange(static_cast<T>(min_index), dictionary_length) &&
+               IndexInRange(static_cast<T>(max_index), dictionary_length);
+    } else {
+        T min_index = std::numeric_limits<T>::max();
+        T max_index = std::numeric_limits<T>::min();
+        for (int x = 0; x < length; x++) {
+            min_index = std::min(values[x], min_index);
+            max_index = std::max(values[x], max_index);
+        }
+        return IndexInRange(min_index, dictionary_length) && IndexInRange(max_index, dictionary_length);
```
stdpain: We can add a template function in `simd/` to handle the dispatching for us. This ensures the readability of the main logic.

be/src/formats/parquet/parquet_block_split_bloom_filter.cpp
eshishki: Done — `_mm256_testc_si256` now replaces the AND+cmpeq+movemask sequence in `test_hash`. Thanks.

be/src/formats/parquet/stored_column_reader.cpp
eshishki: Per your godbolt note — dropped the hand-rolled SIMD block; the scalar loop relies on gcc 12+ auto-vec. Also removed the `<immintrin.h>` / `<arm_neon.h>` includes. Thanks.

be/src/base/bit/rle_encoding.h
eshishki: Done — added `simd_minmax<T>` and `simd_dict_gather<DstT,SrcT,IdxT>` in `base/simd/rle_simd.h`; the `IndicesInRange` and `GetBatchWithDict` literal-gather sites are now one line each. Thanks.

CelerData-Reviewer:
@codex review

stdpain (approved):

kangkaisen (approved):

PR #73289, opened 2026-05-14
[Enhancement] Add SIMD optimizations for column operations

### LINKED ISSUE #73294: [Enhancement] AVX2/NEON SIMD coverage for analytic hot paths

eshishki:
## Why I'm doing:

Vectorise a couple of column-layer hot paths used by hash-aggregate and hash-join build/probe.

Each candidate was microbenched against the pre-PR scalar form on both USE_AVX512=OFF and USE_AVX512=ON builds (m6i.4xlarge, Ice Lake) via the new `column_ops_bench` target; only paths that beat or match scalar in both modes ship in this PR.

Fixes #73294

## What I'm doing:

* `column/column.cpp` — `empty_null_in_complex_column` packs the byte-wide `null_data` into a 32-byte AVX2 `cmpgt` + `movemask` + `ctz` walk (or NEON `vld1q_u8` + `vmaxvq_u8` + scalar bit walk on aarch64), so the offsets-vs-null check only fires on positions flagged as null. The AVX2 path is gated on `!__AVX512F__`: at high null density on AVX-512 builds gcc auto-vectorises the scalar fallback into a 64-byte zmm sweep that matches or beats this 32-byte ymm loop, so the fallback wins there.
* `column/chunk.cpp` — `Chunk::filter` / `MutableChunk::filter` early-return guard switches from `SIMD::count_zero(selection) == 0` to `SIMD::all_ones(selection)` (short-circuits via `memchr` instead of full scan).
* `column/column_helper.cpp` — `merge_two_anti_filters` delegates into `or_two_filters` / `merge_two_filters`; the bytewise AND/OR/ANDN loops stay scalar because the compiler auto-vectorises them to `vpand`/`vpor`/`vpandn` (microbench was 132 ns vs 132 ns).
* `column/binary_column.cpp` — `remove_first_n_values` is now an in-place rewrite (`memmove` + offset shift, no temp column via `cut()`) with explicit `_resource = {}` on full truncation so zero-copy backings drop their external owner.
* `column/array_column.cpp` / `column/map_column.cpp` — add an aarch64 NEON variant of `filter_range` mirroring the existing AVX2 path.
* `bench/column_ops_bench.cpp` — microbench covering each shipping SIMD path, with USE_AVX512=OFF and USE_AVX512=ON columns.

**Considered but dropped after benching against pre-PR scalar in both USE_AVX512 modes:**

* `column_hash/column_hash.cpp` boundary detection via `SIMD::find_zero` / `find_nonzero` (the "optimize for sparse nulls" TODO): wins 2.5× at d=1% but regresses 2-3.5× at d=10-90% (memchr setup outweighs the inline branch-predicted byte-loop on short runs). At d=1% boundary detection is not the bottleneck — hash mix dominates — so the global win is small while the regression at typical density is real. Tried two safe-gate variants (peek-32-then-memchr, prescan via `count_nonzero`) — both still regress 10-26% on dense at AVX-512=ON. TODO left in place for a future hybrid that uses column-level metadata (e.g. `NullableColumn::null_count`) to pick strategy once per visit instead of per-run.
* `fixed_length_column_base.cpp` `replicate()` via `SIMDUtils::simd_fill`: wash on AVX-512 OFF, 5-15% regression on AVX-512 ON for variable `fill_count` (the realistic workload). gcc auto-vectorises the nested loop at least as well as `std::fill_n` on this input shape.
* `fixed_length_column_base.cpp` `fill_default` pointer hoist: identical perf (gcc already hoists `data()` calls at -O3); not a SIMD change.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

## Performance

Microbench scalar vs SIMD, 10 reps with `--benchmark_report_aggregates_only=true`, AWS `m6i.4xlarge` (Intel Ice Lake), via the `column_ops_bench` target.

### `empty_null_in_complex_column` (FullSweep scenario — nulls have empty offsets, real hot path for `std::optional<varchar>`)

| Density | USE_AVX512=OFF scalar → SIMD | USE_AVX512=ON scalar → SIMD |
|---|---|---|
| d=0%  | 1195 → 85.7 ns (**13.9×**) | 1193 → 84.5 ns (**14.1×**) |
| d=1%  | 1554 → 99.9 ns (**15.5×**) | 1520 → 99.8 ns (**15.2×**) |
| d=10% | 1582 → 327 ns (**4.8×**)   | 1576 → 327 ns (**4.8×**)   |
| d=50% | 3160 → 1342 ns (**2.4×**)  | 3404 → 1348 ns (**2.5×**)  |
| d=90% | 3020 → 3012 ns (1.0×)      | 2513 ns (scalar, gate off) |

### `Chunk::filter` all-ones short-circuit (`count_zero` vs `all_ones`)

| Density | USE_AVX512=OFF scalar → SIMD | USE_AVX512=ON scalar → SIMD |
|---|---|---|
| d=0%  | 100 → 30.5 ns (**3.3×**) | 99.4 → 30.5 ns (**3.3×**) |
| d=1%  | 100 → 3.18 ns (**31×**)  | 99.4 → 2.89 ns (**34×**)  |
| d=10% | 100 → 2.31 ns (**43×**)  | 99.4 → 2.02 ns (**49×**)  |
| d=50% | 100 → 2.31 ns (**43×**)  | 99.4 → 2.03 ns (**49×**)  |


CelerData-Reviewer:
@codex review

be/src/column/binary_column.cpp
```diff
@@ -584,17 +664,42 @@ void BinaryColumnBase<T>::assign(size_t n, size_t idx) {
...
+        return;
+    }
+
     size_t remain_size = _offsets.size() - 1 - count;
+    if (remain_size == 0) {
+        _bytes.clear();
+        _offsets.resize(1);
+        _offsets[0] = 0;
+        invalidate_slice_cache();
+        return;
```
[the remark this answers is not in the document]
eshishki: Fixed in `4802bb6c661`. Added `_resource = {};` to the `remain_size == 0` early-return path so zero-copy `BinaryColumnBase` instances drop the external owner when fully truncated. The previous `cut()`-based implementation cleared `_resource` implicitly; the new direct path needs to do it explicitly.

be/src/column/binary_column.cpp
```diff
@@ -30,6 +32,43 @@
...
+        strings::memcpy_inlined(dst, src, size);
+        return;
+    }
+
+    size_t i = 0;
+#if defined(__AVX2__)
+    for (; i + 32 <= size; i += 32) {
+        __m256i v = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(src + i));
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(dst + i), v);
+    }
```
stdpain: glibc's `memcpy` is sufficient.

be/src/column/binary_column.cpp
```diff
@@ -74,12 +113,50 @@ void BinaryColumnBase<T>::_append_binary_impl(const BinaryColumnBase<SrcOffset>&
...
+            for (; i + 2 <= count; i += 2) {
+                uint64x2_t off = vld1q_u64(reinterpret_cast<const uint64_t*>(new_offsets + i));
+                vst1q_u64(reinterpret_cast<uint64_t*>(new_offsets + i), vaddq_u64(off, delta_vec));
+            }
+            for (; i < count; i++) new_offsets[i] += delta;
+        }
+#else
         for (size_t i = 0; i < count; i++) {
             new_offsets[i] += delta;
         }
```
stdpain, line 79: auto vectorized loop

be/src/column/fixed_length_column_base.cpp
```diff
@@ -124,24 +131,110 @@ StatusOr<MutableColumnPtr> FixedLengthColumnBase<T>::replicate(const Buffer<uint
...
+            uint64x2_t cur = vld1q_u64(reinterpret_cast<const uint64_t*>(data + i));
+            vst1q_u64(reinterpret_cast<uint64_t*>(data + i), vbslq_u64(mask, val_vec, cur));
         }
+        for (; i < size; i++) {
+            if (f[i] == 1) data[i] = val;
+        }
+    } else {
+        for (size_t i = 0; i < size; i++) {
+            if (f[i] == 1) data[i] = val;
+        }
```
stdpain: auto vectorized loop in GCC14 https://godbolt.org/z/f7WPYeEP9

be/src/column/map_column.cpp
```diff
@@ -393,8 +399,12 @@ size_t MapColumn::filter_range(const Filter& filter, size_t from, size_t to) {
                 // ```
                 auto delta = offsets[check_offset] - offsets[result_offset];
                 memmove(offsets + result_offset + 1, offsets + check_offset + 1, kBatchSize * sizeof(offsets[0]));
-                for (int i = 0; i < kBatchSize; i++) {
-                    offsets[result_offset + i + 1] -= delta;
+                // SIMD-optimized offset delta subtraction (kBatchSize = 32, process 8 uint32_t per iteration)
+                const __m256i delta_vec = _mm256_set1_epi32(static_cast<int32_t>(delta));
+                for (int i = 0; i < kBatchSize; i += 8) {
```
stdpain:
```suggestion
                for (int i = 0; i <= kBatchSize; i += 8) {
```

be/src/column/column_hash.h
```diff
@@ -31,15 +32,73 @@
...
+
+// AVX2 version: 32 bytes per iteration with early-exit on first mismatching
+// lane. Good for short strings (<= 128 bytes) where the early exit dominates;
+// for longer strings libc memcmp (ERMS / REP MOVSB on x86_64) is faster than
+// the hand-rolled loop, so we delegate above the threshold.
+// NOTE: This function will access 31 excessive bytes after p1 and p2 for the
+//       SIMD path; callers must guarantee the trailing padding.
+template <typename T>
+typename std::enable_if<sizeof(T) == 1, bool>::type memequal_padded(const T* p1, size_t size1, const T* p2,
+                                                                    size_t size2) {
```
stdpain: I'm not entirely sure how effective this optimization will be, so please consider submitting a separate PR for this to avoid it being merged into the main optimization.

be/src/column/binary_column.cpp
eshishki: Removed the `simd_copy_bytes` helper; `replicate()` is back on `strings::memcpy_inlined`. Thanks.

be/src/column/binary_column.cpp
eshishki: Dropped the hand-rolled SIMD block. Verified on godbolt: gcc 14.1 -O3 -mavx2 auto-vectorises the scalar `new_offsets[i] += delta` loop into the same `vpaddd ymm0, ymm1, [rax]` + `vmovdqu` sequence, and aarch64 gcc 14.2 emits the equivalent `dup v31.4s / add v.4s, v.4s, v31.4s / str q`. Thanks.

be/src/column/fixed_length_column_base.cpp
eshishki: Dropped the hand-rolled `fill_default` SIMD block per your godbolt link; scalar loop only. Thanks.

be/src/column/map_column.cpp
eshishki: The AVX2 sub-block that loop was inside is now gone — godbolt shows gcc 14.1 -O3 -mavx2 unrolls the scalar `for (int i = 0; i < kBatchSize; i++) offsets[result_offset + i + 1] -= delta;` into 4 fully-unrolled `vpsubd` blocks with a broadcast (compile-time `kBatchSize = 32`), no main loop. One FYI on the suggested `i <= kBatchSize` step `+= 8` — that would emit a 5th store on `offsets[result_offset + 33 .. + 40]`, past the 32-element `memmove`d window, so it would be an OOB write. Thanks.

be/src/column/column_hash.h
eshishki: Removed from this PR. Separate PR coming after fresh local bench numbers. Thanks.

eshishki:
@stdpain — circling back after a USE_AVX512=OFF/ON microbench pass on `column_ops_bench` (now in the PR). Changes since your last review:

* `fixed_length_column_base.cpp` `replicate()` — you flagged it as auto-vectorised by GCC14 (https://godbolt.org/z/f7WPYeEP9). Confirmed: `simd_fill` (== `std::fill_n`) is wash vs the nested loop on AVX-OFF and **regresses 5-15% on AVX-ON** for variable `fill_count`. Reverted to the pre-PR nested loop. The fill_default pointer-hoist was also reverted (identical perf in both modes).
* `column_hash/column_hash.cpp` boundary detection — bench-confirmed regression 2-3.5× on null density 10-90% (memchr setup outweighs the inline byte loop on short runs). Tried two safe-gate variants (peek-32-probe-then-memchr, prescan via `count_nonzero`) — both still regress 10-26% on dense at AVX-512=ON. Reverted. The `TODO: optimize performance for sparse nulls` is back in place; a future hybrid will need column-level metadata (`NullableColumn::null_count`) to make the decision once per visit instead of per-run.
* `column.cpp` `empty_null_in_complex_column` AVX2 path is now gated on `!__AVX512F__`: at d=90% on AVX-512 builds GCC auto-vectorises the scalar fallback into a 64-byte zmm sweep that beats the 32-byte ymm path. AVX-OFF builds stay on the hand AVX2 (wins 1.0-15.5× across densities).

be/src/column/column.cpp
```diff
@@ -95,12 +101,63 @@ bool Column::empty_null_in_complex_column(const ImmBuffer<uint8_t>& null_data, c
...
+            // Process each set bit (null position)
+            while (null_mask && !need_empty) {
+                uint32_t bit_pos = __builtin_ctz(null_mask);
+                size_t idx = i + bit_pos;
+                if (offsets[idx + 1] != offsets[idx]) {
+                    need_empty = true;
+                }
+                null_mask &= null_mask - 1; // Clear lowest set bit
+            }
+        }
```
stdpain: replace to avoid inner loop
```
    for (; i + 8 <= size && !need_empty; i += 8) {
       // load offset (int4 x 8)
        __m256i cur = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(offsets + i));
       // load offset + 1 (int4 x 8)
        __m256i nxt = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(offsets + i + 1));
        // cmp offset and offset + 1
        // eq lanes = 0xFFFFFFFF 
        __m256i eq = _mm256_cmpeq_epi32(cur, nxt);

        // load 8 uint8 nulls then extends to int8x4
        __m128i nulls8  = _mm_loadl_epi64(reinterpret_cast<const __m128i*>(null_data + i));
        __m256i nulls32 = _mm256_cvtepu8_epi32(nulls8);
        // null != 0 -> 0xFFFFFFFF
        __m256i nullmask = _mm256_cmpgt_epi32(nulls32, _mm256_setzero_si256());

        // nonempty & null  ==  (~eq) & nullmask
        __m256i found = _mm256_andnot_si256(eq, nullmask);

        need_empty = !_mm256_testz_si256(found, found);
    }
```

be/src/column/map_column.cpp
```diff
@@ -417,6 +423,42 @@ size_t MapColumn::filter_range(const Filter& filter, size_t from, size_t to) {
...
+            memset(element_filter.data() + offsets[check_offset], 1, element_size);
+            if (result_offset != check_offset) {
+                DCHECK_LE(offsets[result_offset], offsets[check_offset]);
+                auto delta = offsets[check_offset] - offsets[result_offset];
+                memmove(offsets + result_offset + 1, offsets + check_offset + 1, kBatchSize * sizeof(offsets[0]));
+                for (size_t i = 0; i < kBatchSize; i++) {
+                    offsets[result_offset + i + 1] -= delta;
+                }
+            }
+            result_offset += kBatchSize;
```
stdpain, line 451: use nibble_mask algorithm. you could reference arm implemention in ColumnHelper::t_filter_range
```
    while (start_offset + kBatchNums < to) {
        const uint8x16_t vfilter = vld1q_u8(filter_data);
        // nibble_mask[i] != 0 ? 0xFF : 0x00
        uint64_t nibble_mask = SIMD::get_nibble_mask(vtstq_u8(vfilter, vfilter));
        if (nibble_mask == 0) {
           // no hits
        } else if (nibble_mask == 0xffff'ffff'ffff'ffffull) {
            // all hits
            result_offset += kBatchNums;
        } else {
            nibble_mask &= 0x8888'8888'8888'8888ull;
            for (; nibble_mask > 0; nibble_mask &= nibble_mask - 1) {
                uint32_t index = __builtin_ctzll(nibble_mask) >> 2;
                // process index
            }
        }
```

be/src/column/array_column.cpp
```diff
@@ -359,6 +365,42 @@ size_t ArrayColumn::filter_range(const Filter& filter, size_t from, size_t to) {
...
+            auto element_size = offsets[check_offset + kBatchSize] - offsets[check_offset];
+            memset(element_filter.data() + offsets[check_offset], 1, element_size);
+            if (result_offset != check_offset) {
+                DCHECK_LE(offsets[result_offset], offsets[check_offset]);
+                auto delta = offsets[check_offset] - offsets[result_offset];
+                memmove(offsets + result_offset + 1, offsets + check_offset + 1, kBatchSize * sizeof(offsets[0]));
+                for (size_t i = 0; i < kBatchSize; i++) {
+                    offsets[result_offset + i + 1] -= delta;
+                }
+            }
```
stdpain, line 392: use nibble_mask algorithm in ColumnHelper::t_filter_range

be/src/column/column.cpp
eshishki: Thanks, fixed. Also rewrote the NEON path of this function the same branchless way (offsets-equality vs null mask, no inner loop).

be/src/column/map_column.cpp
eshishki: Thanks, fixed — switched to the `get_nibble_mask` walk.

be/src/column/array_column.cpp
eshishki: Thanks, fixed — switched to the `get_nibble_mask` walk.

stdpain (approved):

kevincai (approved):

PR #73290, opened 2026-05-14
[Enhancement] Add SIMD optimizations for aggregate functions

### LINKED ISSUE #73294: [Enhancement] AVX2/NEON SIMD coverage for analytic hot paths

eshishki:
## Why I'm doing:

Vectorise aggregate-function hot paths: `count`, `boolor`, and the
generic selective-update / nullable-aggregate plumbing they all share.

Fixes #73294

## What I'm doing:

* `exprs/agg/aggregate.h` — the `*_batch_selectively` paths dispatch
  between the pre-PR scalar branch loop and a `SIMD::find_zero` (memchr)
  skip loop via shared helpers (`agg_selective::probe_filter_sparse` /
  `for_each_selected`):
  - the probe samples **four 64-byte windows** spread across the filter
    (start, n/3, 2n/3, end) instead of a contiguous prefix: filters
    produced over sorted/clustered keys come in runs, and a prefix-only
    probe misreads a chunk that opens with a skipped run — measured
    **31x slower than scalar** when the rest of the chunk is densely
    selected;
  - the sparse loop carries an **escape hatch**: once it keeps more rows
    than the probe promised (`n/32` + slack), the remainder is finished
    with the branch loop, bounding a fooled probe at ~2x scalar instead
    of 31x; the bookkeeping only runs inside the sparse loop, so it is
    free on both happy paths;
  - threshold `1/32` (~3%) is the measured crossover for these flat
    loops: ~6-7 ns per kept row in the skip loop vs ~0.2 ns per row in
    the branch loop.
* `exprs/agg/nullable_aggregate.h` — the null-aware
  `update_batch_selectively` walks the null mask in 32-byte windows; the
  selected-row iteration inside the windows is probed **once per chunk**
  with divisor `8`, and the walk is instantiated separately per side:
  - unconditional `find_zero` (the first revision of this PR) ran
    **3.4x slower** than the scalar walk at dense selections, even with
    the virtual `nested_function->update` per row;
  - divisor 8, not 32: the heavier per-row payload and the <=32-byte
    memchr ranges put the measured crossover at ~12% selected
    (`find_zero` still wins at 10%, loses at 15%);
  - per-side instantiation because a runtime sparse/dense branch inside
    the per-window iteration degrades the dense-loop codegen ~1.5x at
    10-30% selected;
  - the three `NullBatchKind` branches remain deduplicated into a shared
    lambda; `merge_batch_selectively` and the non-nullable update path
    reuse `agg_selective::for_each_selected`.
* `exprs/agg/count.h` — `update_batch` over nullable input replaces the
  scalar null-count loop with `SIMD::count_zero` (~7.7x at 4096 rows:
  GCC auto-vectorises the reduction, but into a byte->qword widening sum
  without `psadbw` — see the `_ScalarMem*` bench variants). The
  selective path reuses the shared dispatch.
* `exprs/agg/boolor.h` — bulk OR via `SIMD::contains_nonzero_bit`;
  early-exit once the accumulator becomes `true` (~36x at sparse-true;
  dense-true unchanged, both short-circuit on the first non-zero block).
* `exprs/agg/sum.h` — nullable accumulate stays scalar (auto-vectorises
  cleanly); only a hoist of `get_data().data()` out of `get_values`.
* `exprs/agg/group_concat.h` — null-merge loop replaced with
  `ColumnHelper::or_two_filters`.
* `bench/agg_simd_bench.cpp` — standalone microbench mirroring the
  shipped dispatch verbatim next to the pre-PR scalar baseline and the
  rejected prefix-probe version: uniform sweeps, the clustered patterns
  that motivated the stratified probe, and the windowed walk with a
  non-devirtualizable per-row payload. m6i results recorded in the file.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

## Performance

`count` / `boolor` (unchanged since the first revision):

|---|---|---|
| `BM_Count_Nullable /any` | 777 → 100 ns (**7.77x**) | 451 → 99.2 ns (**4.55x**) |
| `BM_BoolOr /0` (sparse-true) | 2071 → 56.8 ns (**36.5x**) | 2057 → 57.8 ns (**35.6x**) |
| `BM_BoolOr /1` | 181 → 5.32 ns (**34.0x**) | 187 → 5.43 ns (**34.4x**) |
| `BM_BoolOr /50, /100` | 0.578 = 0.578 (short-circuit) | 0.578 = 0.578 |

Selective dispatch, AVX2 (m6i.4xlarge Ice Lake, g++ 14.3, 4096 rows,
10 reps, CV <= 3%); "prefix probe" is the rejected first dispatch, kept
in the bench for comparison:

| filter pattern | scalar | prefix probe | shipped |
|---|---|---|---|
| uniform, 0% selected | 777 | 37.7 | 37.2 (**20.9x**) |
| uniform, 1% | 776 | 298 | 296 (**2.6x**) |
| uniform, 5-100% (dense fallback) | 777 | 781 | 782 (~1.0x) |
| 256-byte skip run, rest 50% selected | 777 | **12315 (15.9x slower)** | 782 |
| 256-byte skip run, rest 100% selected | 777 | **24321 (31.3x slower)** | 782 |

Nullable 32-byte-window walk, virtual per-row payload; "always
find_zero" is the first revision of this PR:

| % selected | scalar walk | always find_zero | shipped |
|---|---|---|---|
| 1 | 2766 | 539 (**5.1x**) | 553 (**5.0x**) |
| 10 | 2592 | 2304 | 2359 |
| 15 | 2651 | 3443 (0.77x) | 2819 |
| 50 | 4367 | **15011 (0.29x)** | 4174 |
| 100 | 8468 | **27572 (0.31x)** | 8728 (~1.0x) |


CelerData-Reviewer:
@codex review

be/src/exprs/agg/boolor.h
```diff
@@ -135,13 +133,10 @@ class BoolOrAggregateFunction final
...
-
-            for (size_t i = frame_start; i < frame_end; ++i) {
-                bool value = column.immutable_data()[i];
-                if (value) {
-                    this->data(state).result = true;
-                    break;
-                }
+            const auto* data = column.immutable_data().data();
+            // SIMD optimization: use contains_nonzero_bit to quickly check for any true value in frame
+            if (SIMD::contains_nonzero_bit(data + frame_start, frame_end - frame_start)) {
```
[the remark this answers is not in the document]
eshishki: Good catch — real regression. Fixed in `9feb3cbcc91` by adding `frame_start >= frame_end` to the early-return guard at the top of `update_batch_single_state_with_frame`, so both the SIMD non-null path and the scalar nullable path bail out on empty frames before any indexing.

before-Sunrise:
can you give a performance test report which can prove your optimization is effective

eshishki:
@before-Sunrise thanks for the review. I've added a Google-Benchmark microbench covering every hot path the PR touches: `count.h` update_batch over nullable input (`count_zero`), `boolor.h` (`contains_nonzero_bit`), the adaptive `merge_batch_selectively` / `update_batch_selectively` in `aggregate.h` and `nullable_aggregate.h` (`count_zero` + `find_zero`), and the `get_values` broadcast in `count.h` / `sum.h`. Each pair runs the exact pre-PR scalar loop and the exact post-PR SIMD loop on identical payloads and sweeps the relevant data distribution (null-ratio for `count`, true-ratio for `boolor`, filter pass-ratio around the 12.5% adaptive-branch boundary for selective merge/update, and width 8…4096 for broadcast), so reviewers can read off both the speedup and the absence of regression on the scalar fallback.

See `be/src/bench/agg_simd_bench.cpp` in 9d66fda7705. I'll run it on an AVX2 box today and post the numbers here.

eshishki:
@before-Sunrise as promised, here is the performance report from the microbench (`be/src/bench/agg_simd_bench`). Hardware: AMD m6a.4xlarge (AVX2, no AVX-512). Run with `--benchmark_repetitions=5 --benchmark_report_aggregates_only=true`.

| Bench | scalar | SIMD / Adaptive | Speedup / Note |
|---|---|---|---|
| `BM_Count_Nullable /0` (no nulls) | 916 ns | 130 ns | **7x** |
| `BM_Count_Nullable /50` (mixed) | 916 ns | 130 ns | **7x** |
| `BM_Count_Nullable /100` (all nulls) | 916 ns | 130 ns | **7x** |
| `BM_BoolOr /0` (sparse true) | 1163 ns | 43.3 ns | **27x** |
| `BM_BoolOr /1` | 108 ns | 4.4 ns | **25x** |
| `BM_BoolOr /50` (dense true) | 0.56 ns | 0.56 ns | tie (scalar early-exit best case) |
| `BM_MergeSelectively /0` (no rows pass) | 987 ns | 31.6 ns | **31x** |
| `BM_MergeSelectively /1` | 987 ns | 255 ns | **3.9x** |
| `BM_MergeSelectively /5` | 987 ns | 1020 ns | marginal (~3% slower) |
| `BM_MergeSelectively /12` (boundary) | 987 ns | 978 ns | ≈ scalar |
| `BM_MergeSelectively /50` (dense) | 987 ns | 978 ns | ≈ scalar |
| `BM_MergeSelectively /100` (all pass) | 987 ns | 978 ns | ≈ scalar |

A couple of notes on how to read this:

* **No dense-filter regression.** An earlier version of the PR did a full `SIMD::count_zero(filter, n)` up-front to pick between sparse-skip and dense-iterate branches. On a 4096-byte filter that cost ~100 ns, which on the dense `/12..100` rows turned the adaptive path into a 2.9x regression at `/12` (2819 ns vs 987 ns scalar) and a 12% regression at `/50..100` (1101 ns vs 987 ns). The current version samples only the first 256 bytes of the filter (`probe_n`), so dense rows now pay ~5 ns instead of ~100 ns -- adaptive sits at 978 ns, slightly below scalar. The microbench rows above are the proof that the regression is closed.
* **Synthetic callback.** The bench callback is `acc ^= i`, so the dispatch loop dominates the measurement. In production the `update`/`merge` callback in `Aggregator` is a real function call plus hash-state mutation; the probe overhead disappears entirely there.
* **`BoolOr /50` and `/100` tie at 0.56 ns** because the scalar version exits on the first non-zero byte. The SIMD path's `contains_nonzero_bit` is not slower there -- both are dominated by a single byte read.

The other paths in this PR (`count.h::update_batch`, `count.h::get_values`, `boolor.h::update_batch`) follow the same shape: SIMD wins on sparse-null / sparse-true and matches scalar on dense / saturated regimes.

Full sweep and per-Arg stddev/cv numbers are in the commit message of the [UT] commit (a1b1937cadf at the time of writing).

be/src/exprs/agg/count.h
```diff
@@ -204,9 +240,8 @@ class CountNullableAggregateFunction final
             const auto* nullable_column = down_cast<const NullableColumn*>(columns[0]);
             if (nullable_column->has_null()) {
                 const uint8_t* null_data = nullable_column->immutable_null_column_data().data();
-                for (size_t i = 0; i < chunk_size; ++i) {
-                    this->data(state).count += !null_data[i];
-                }
+                // SIMD optimization: count_zero counts positions where null_data[i] == 0 (non-null rows)
+                this->data(state).count += SIMD::count_zero(null_data, chunk_size);
```
stdpain, line 205: It's a auto-vectorized loop

eshishki: Done — also reverted the same `count_zero` reduction in the two other frame methods in this file.

eshishki: Kept the revert as requested — one concern worth flagging, though.

The loop does get vectorized, but "auto-vectorized" isn't speed-equivalent to `count_zero` here. In our microbench (`BM_Count_Nullable` in `agg_simd_bench`) the scalar reduction ran ~7x slower than `SIMD::count_zero` — 777 → 100 ns at 4096 rows. The gap is that GCC doesn't lower `count += (x == 0)` to `psadbw`; it widens byte-by-byte, whereas `count_zero` does `cmpeq` + `movemask` + `popcnt` over 64 bytes at a time. In production the accumulator is also in memory (`this->data(state).count` via `AggDataPtr`), which is even less favorable for the auto-vectorizer than the bench's local accumulator.

So this may be a real ~7x regression on `COUNT(nullable_col)`, not just cosmetics. We'll re-verify the generated asm / perf under the actual BE build flags (GCC `-O3`, `USE_AVX512=OFF`) today and report back here; if it confirms, we'd propose keeping `count_zero` for this specific path.

eshishki: Re-tested on m6i.4xlarge (Ice Lake), g++ 14.3, `-O3 -mavx2 -mpopcnt`, standalone microbench, 4096 rows, 5 reps (CV < 0.3%):

| variant | ns |
|---|---|
| scalar, local accumulator | 779 |
| scalar, `this->data(state).count` via `__restrict` ptr (= prod shape) | 778 |
| scalar, same but non-restrict ptr | 1933 |
| `SIMD::count_zero` | 101 |

You're right that it's auto-vectorized — `__restrict` survives the `reinterpret_cast` in `data(state)`, so the prod loop does vectorize (778 ns; drop the restrict and it falls back to a scalar byte loop at 1933 ns). But "vectorized" isn't speed-equivalent here: GCC lowers `count += (x == 0)` to a byte→qword widening reduction (`vpcmpeqb` + a `vpmovzx*` cascade + ~19 `vpaddq` per 32 B; no `psadbw` anywhere in the binary), whereas `count_zero` is `vpcmpeqb` + `vpmovmskb` + `popcnt` over 64 B — hence the ~7.7x gap.

So this looks like a real ~7.7x regression on `COUNT(nullable_col)`.

eshishki: While we were at it: `StoredColumnReaderImpl::count_not_null` in #73288 has the exact same shape — `count += (def_levels[i] == max_def_level)` over `int16` — and was reverted from a hand-written `vpcmpeqw` + `movemask` + `popcount` to the scalar loop. Same story as here: it does auto-vectorize, but into a widening reduction that's ~7x slower than the SIMD count (it accumulates into a local `size_t`, so the 779 vs 101 ns gap above is the right ballpark — happy to bench that exact int16 case if useful). It's on the parquet skip path (`stored_column_reader.cpp:424,530`), and #73288 is already merged.

How would you like to handle it — restore the SIMD `count_not_null` yourself, or should we send a separate follow-up PR (or fold the one-liner into this PR)?

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

eshishki:
Probing follow-up — all three review findings on the adaptive dispatch
(the two Codex P2s on `aggregate.h`/`nullable_aggregate.h` and the
`count.h` pre-count one) reproduced in microbench and are addressed in
ef3ea5220f0 + b6de8ec2caf:

* **Clustered filters fooling the prefix probe** — confirmed: a chunk
  opening with a 256-byte skipped run and densely selected after that
  made the prefix probe pick the find_zero loop, 24321 ns vs 777 ns
  scalar (**31x**). The probe now samples four 64-byte windows spread
  across the filter (start, n/3, 2n/3, end), and the sparse loop got an
  escape hatch: once it keeps more rows than the probe promised
  (n/32 + slack) it finishes the remainder with the branch loop, so a
  filter clustered enough to fool all four windows is bounded at ~2x
  scalar instead of 31x. Both are free where it matters: the uniform
  sweep is within 1 ns of the prefix probe at every density, and the
  escape bookkeeping only runs inside the sparse loop.

* **Unconditional find_zero in the nullable window walk** — confirmed:
  at dense selections it ran 3.4x slower than the scalar walk
  (15011 vs 4367 ns at 50% selected) even with the virtual
  `nested_function->update` per row. The walk now probes once per chunk
  and instantiates the window iteration separately per side. Divisor 8
  rather than 32: the heavier per-row payload and the <=32-byte memchr
  ranges put the measured crossover at ~12% selected (find_zero still
  wins at 10%, loses at 15%). The per-side instantiation matters too: a
  runtime sparse/dense branch inside the per-window lambda degraded the
  dense-loop codegen ~1.5x at 10-30% selected.

* **count.h up-front full count_zero** — removed; the selective path
  now reuses the shared dispatch (`agg_selective::for_each_selected`
  in aggregate.h), same as nullable merge and the non-nullable update
  path.

The bench (`agg_simd_bench.cpp`) was reworked to mirror the shipped
dispatch verbatim next to the scalar baseline and the rejected
prefix-probe version, including the clustered patterns and the windowed
walk with a non-devirtualizable per-row payload; m6i numbers are
recorded at the end of the file and in the updated PR description.

CelerData-Reviewer:
@codex review

eshishki:
@stdpain will you be able to pick this one too?

stdpain (approved):

trueeyu:
Is this a bottleneck in production?

trueeyu (approved):

eshishki:
@trueeyu
> Is this a bottleneck in production?

Fair question — I measured it on a real cluster rather than argue from the microbench. Setup: shared-data mode on a single m6i.4xlarge host (16 vCPU), two CN processes so the planner produces the production two-phase aggregation plan (streaming preagg → shuffle → merge), 100M-row table, hot data cache, 10 `EXPLAIN ANALYZE` runs per query, comparing this branch against its merge-base (`05428c3`) with binaries swapped on the same cluster/data.

Honest answer: **no, these loops alone are not an end-to-end bottleneck** — in the queries that hit them, scan, exchange and hash-table probing dominate wall time. What the profiles do show, operator-level:

| Scenario | agg operator (path touched by this PR) | base | this PR |
|---|---|---|---|
| `bool_or(flag)` over an almost-all-false column | update, `contains_nonzero_bit` | 14.2 ms | 4.4 ms (**3.2x**) |
| `count(nullable_col)`, ~30% nulls, no GROUP BY | update, `count_zero` | 34.6 ms | 30.4 ms (1.14x) |
| high-card GROUP BY, AUTO streaming preagg at ~50% HT hit-rate | selective update (dense branch) | 433–536 ms | 424–568 ms (within ±10% run-to-run noise) |
| same over nullable input | nullable windowed walk | ~512 ms | ~510 ms (within noise) |
| unique-key GROUP BY (passthrough + plain merge) | merge | ~1278 ms | ~1283 ms (within noise) |

OLAP_SCAN times are identical across binaries in all cases (sanity check).

Two structural observations from tracing the call sites: the selective-update path in AUTO streaming preagg only ever sees 20–90% selected rows (below 20% it switches to passthrough), so production selective updates sit on the dense branch — which is why the dispatch is tuned to keep that branch byte-identical to the old loop, and those rows land within run-to-run noise: no measurable regression. The sparse skip loop's big wins (the 20x in the bench) apply to `GROUP BY ... LIMIT` merge phases, where the operator is dominated by hash-table probing anyway.

So the honest framing: this PR is a set of cheap local wins (`bool_or` being the visible one), a fix for the nullable-count reduction, and hardened dispatch that provably doesn't regress the dense paths that production actually exercises.

PR #73303, opened 2026-05-14
[Enhancement] Parallel coalesced reads for external tables

### LINKED ISSUE #73304: [Enhancement] Parallel coalesced reads for external tables
## Enhancement

External Iceberg/Hive scans against object storage spend the bulk of scan-thread time blocked on per-range S3/Azure/HDFS GETs. `SharedBufferedInputStream` (SBI) already coalesces byte ranges, but it pulls each coalesced range synchronously through a single backend call, so a slow round-trip blocks the parquet/ORC consumer.

### Proposal

Add chunk-based parallel reads on top of SBI, gated per-backend on a new virtual `SeekableInputStream::is_thread_safe_positional_read()`.

- Default `false` keeps every existing backend on today's sync path.
- A backend overrides `true` only when its `read_at_fully` is genuinely positional: no shared cursor, no shared mutation, no unsynchronized stats writes.
- `SeekableInputStreamWrapper`, `BundleSeekableInputStream`, and `CountedSeekableInputStream` forward the gate so the existing wrapper chain in `HdfsScanner` stays transparent.
- Scheduling is bounded by a global `parallel_io_pool` thread pool and a per-file sliding window; in-flight buffer memory is capped by a global budget.

### Scope of the first PR

Roll out the parallel path for:

- `S3InputStream` — `GetObject` with explicit `Range`.
- `AzBlobInputStream` — `BlobClient::DownloadTo` with explicit `HttpRange`.
- `HdfsInputStream` — `hdfsPread` with explicit offset, per-handle `std::shared_mutex` (shared on pread, unique around `close()+reopen` on retry).

Out of scope:

- `posix` — marginal benefit, local fs is rarely the bottleneck for parquet/orc scan.
- `StarletInputStream` — `staros::starlet::fslib::Stream` exposes only stateful `seek/read`; concurrent positional reads need an upstream `Stream::pread`. CLOUD_NATIVE shared-data reads route through Starlet regardless of storage volume type, so they stay on the sync path until that upstream API exists.

### Configs (generic, not S3-specific)

| Name | Default | Mutable | Purpose |
|---|---|---|---|
| `parallel_io_thread_pool_size` | `128` | restart | Global pool size |
| `parallel_io_workers_per_file` | `8` | yes | Per-file sliding window |
| `parallel_io_buffer_limit` | `10%` | yes | Max in-flight buffer mem (`10%`, `1G`, `512M`) |

eshishki:
Fixes #73304

## Why I'm doing:

External Iceberg/Hive scans against object storage spend the bulk of scan-thread time blocked on per-range S3/Azure/HDFS GETs. The shipped `SharedBufferedInputStream` (SBI) coalesces byte ranges but still pulls them synchronously one at a time, so a single slow round-trip blocks the consumer. This PR hides that latency by reading coalesced ranges in parallel through a global thread pool.

## What I'm doing:

Adding chunk-based parallel reads on top of SBI, behind a per-backend gate (`SeekableInputStream::is_thread_safe_positional_read()`). Rolled out for S3, Azure Blob, and HDFS.

### How it was

A parquet/orc/iceberg scan goes through `HdfsScanner` (the connector-side scanner under `be/src/exec/hdfs_scanner/`). For each file the scanner opens a `RandomAccessFile` from the matching `FileSystem`, then wraps the underlying `SeekableInputStream` in a chain:

```
CountedSeekableInputStream     (per-scan stats via __atomic_fetch_add)
  -> BundleSeekableInputStream  (if bundling, slice of a shared physical file)
       -> SeekableInputStreamWrapper  (encrypt / throttle wrappers, transparent)
            -> backend stream  (S3InputStream / AzBlobInputStream / HdfsInputStream / ...)
```

Parquet/ORC readers know which byte ranges they want and hand the list to `SharedBufferedInputStream::set_io_ranges`. SBI coalesces those into larger ranges (`io_coalesce_read_max_distance_size`, `io_coalesce_read_max_buffer_size`), then on every `read_at_fully` the scan thread synchronously pulls one coalesced range from the backend and blocks waiting for the round-trip. On S3/Azure/HDFS that round-trip is the dominant cost.

### What was added

SBI now owns an explicit per-`SharedBuffer` state machine:

```
Cold -> Initializing -> Ready -> Filling -> Filled
                                    |--> Failed     (terminal)
                                    \--> Abandoned  (terminal, on release/dtor)

         alternative branch when parallel is unavailable:
Cold -> SyncOnly -> SyncFilling -> Filled / Failed / Abandoned
```

After coalescing, each large range is split into chunks (one chunk per `max_buffer_size`). The new `_schedule()` enqueues chunks to a global `parallel_io_pool` thread pool. Workers capture `shared_from_this()` on the SBI so it can't be freed mid-read. The consumer thread (parquet reader) calls `_ensure_sb_filled` which either waits on a per-`SharedBuffer` `mutex + cv`, or — if no chunks have been scheduled yet — fills the remaining `NotQueued` chunks on the consumer thread itself (demand fill). Errors are first-error-wins; status is sticky for re-reads.

### The gate

`SharedBufferedInputStream::_schedule()` engages the parallel path only when both conditions hold:

1. `_stream->is_thread_safe_positional_read()` returns true. Default is false. Backends override true only when their `read_at_fully` is genuinely positional (no shared cursor, no shared mutation, no unsynchronized stats writes).
2. `weak_from_this().lock()` succeeds — i.e. the SBI is shared-owned. For unique-owned SBI (currently `segment_iterator.cpp` for CLOUD_NATIVE shared-data segments), parallel is intentionally skipped to keep the lifetime contract simple.

Both `SeekableInputStreamWrapper` and `BundleSeekableInputStream` forward the gate to the inner stream, so the wrapper chain in `HdfsScanner` is transparent. The `CountedSeekableInputStream` stats path uses `__atomic_fetch_add` on `int64_t`, safe under parallel reads.

### Where we plugged in

The plug point is exclusively inside `SharedBufferedInputStream`. No connector / no scanner / no parquet code changed. Backend streams added `read_at_fully` positional overrides + gate true:

- **S3InputStream** — `GetObject` with explicit `Range`. AWS SDK handle is safe for concurrent positional calls. `_size` lazy-init via `std::atomic<int64_t>` + `_size_init_mu` for the double-checked HEAD.
- **AzBlobInputStream** — `BlobClient::DownloadTo` with explicit `HttpRange`. Same `_size` pattern.
- **HdfsInputStream** — `hdfsPread` with explicit offset. Per-handle `std::shared_mutex`: shared lock around `hdfsPread`, unique lock around `close()+ensureOpened()` on retry. Pattern is release-shared → take-unique-for-reopen → retry-under-shared; never upgrade. Safety of concurrent `hdfsPread` on a shared `hdfsFile` is the HDFS `FSDataInputStream` positional-read invariant (libhdfs/libhdfs3 both).

Backends still on the sync path:

- **posix** — marginal benefit, not worth the code surface (local fs is rarely the bottleneck for parquet/orc scan).
- **Starlet** — `staros::starlet::fslib::Stream` exposes only stateful `seek/read`; no positional API visible from BE. Concurrent reads through a single Stream would race on the internal cursor. Unblocker is an upstream `Stream::pread`.

### What this means for CLOUD_NATIVE shared-data

CN reads CLOUD_NATIVE shared-data segments via `StarletFileSystem` (the StaRos integration that brings sharding + the `<storage_root_path>/starlet_cache` client cache). The underlying stream is `StarletInputStream`. Gate false → sync path → no parallel prefetch in this commit, regardless of the storage volume type (S3/Azure/GS/HDFS/ADLS2). External Iceberg/Hive queries through external catalogs go through the direct `S3InputStream` / `AzBlobInputStream` / `HdfsInputStream` and **are** covered.

### Configs

Generic, not S3-specific:

| Name | Default | Mutable | Purpose |
|---|---|---|---|
| `parallel_io_thread_pool_size` | `128` | restart | Global pool size |
| `parallel_io_workers_per_file` | `8` | yes | Per-file sliding window |
| `parallel_io_buffer_limit` | `10%` | yes | Max in-flight buffer mem (`10%`, `1G`, `512M`) |

Declarations live alongside the scan-side IO knobs in `config_scan_io_fwd.h` (manifest moved by `config_fwd_headers_manifest.json`).

### Lifetime and safety

- Worker tasks capture `std::shared_ptr<SBI>` via `weak_from_this().lock()`.
- `~SharedBufferedInputStream` waits on `_active_read_count == 0` before returning, with explicit `Abandoned` transitions on every drained `SharedBuffer` to unblock consumers waiting on `cv`.
- Budget reservation (`s_parallel_read_buffer_used`) is refunded in `~SharedBuffer` (RAII tied to the buffer's actual lifetime, not the worker's lifetime).
- All backend `_size` lazy inits use `std::atomic<int64_t>` + a small init mutex; reads use acquire, init uses release.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [ ] Parameter changes
- [ ] Policy changes
- [ ] Feature removed
- [ ] Miscellaneous

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3


CelerData-Reviewer:
@codex review

eshishki:
rerun

PR #73344, opened 2026-05-15
[BugFix] Use long for ConnectorTableId to avoid int overflow producing negative IDs

stephen-shelby:
## Why I'm doing:

On a long-running follower FE, Iceberg/Hive queries start failing with the misleading BE error:

```
Invalid table type. Only hive/iceberg/hudi/delta lake/file/paimon/kudu table are supported
```

while the same query succeeds on the leader. Restarting the cluster "fixes" it. Observed on 3.3.20; the underlying code path is the same on current `main`.

Root cause: `ConnectorTableId` allocates ids as an `int` counter plus a fixed `100_000_000` offset. After roughly `Integer.MAX_VALUE - 1e8` ≈ 2.04B allocations the int wraps to negative — fe.log shows e.g. `Table [id=-2086875593 type=ICEBERG]`. Then:

1. [`TupleDescriptor.toThrift()`](https://github.com/StarRocks/starrocks/blob/main/fe/fe-core/src/main/java/com/starrocks/planner/TupleDescriptor.java) skips negative ids (`table.getId() >= 0`) and additionally truncates the long table id with a stray `(int)` cast even though `TTableId` is `i64` in [`Types.thrift`](https://github.com/StarRocks/starrocks/blob/main/gensrc/thrift/Types.thrift).
2. BE [`HiveDataSource::open`](https://github.com/StarRocks/starrocks/blob/main/be/src/connector/hive_connector.cpp) gets no `tableId`, the `dynamic_cast<const HiveTableDescriptor*>` fails, and returns the generic "Invalid table type" error.

The counter is per-FE-process in-memory state — each FE has its own — which is why the leader and follower behave differently and a restart resets the symptom.

## What I'm doing:

- Make `ConnectorTableId` long-based, backed by `AtomicLong`. Do **not** widen the shared `Id<>` base class (its `int id` field is structural for `TupleId`/`SlotId`/`PlanNodeId` etc.). Add `asLong()`; remove `asInt()`.
- Migrate all 31 callers from `CONNECTOR_ID_GENERATOR.getNextId().asInt()` to `.asLong()`. Every consumer (`Table` / `Database` / various `Builder.setId`) already takes `long`, so the int narrowing was forced and lossy.
- Drop the `(int)` cast in `TupleDescriptor.toThrift()` since `TTableId` is `i64`.
- Add `ConnectorTableIdTest` covering: offset & monotonicity, `getMaxId`, regression for ids past `Integer.MAX_VALUE`, and concurrent uniqueness.
- Add a class-level comment on `ConnectorTableId` documenting that the id is per-FE-process, non-persistent, and not synchronized across FEs — to deter future code from assuming it can be stored on disk or shared cross-FE.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

Reviewers: please confirm which release branches need the backport (reproduced on 3.3.20, so all currently-supported branches are candidates).

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/KuduTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/KuduTable.java
@@ -52,7 +52,7 @@ public class KuduTable extends Table {
 
     public KuduTable(String masterAddresses, String catalogName, String dbName, String tblName, String kuduTableName,
                      List<Column> schema, List<String> partColNames) {
-        super(CONNECTOR_ID_GENERATOR.getNextId().asInt(), tblName, TableType.KUDU, schema);
+        super(CONNECTOR_ID_GENERATOR.getNextId().asLong(), tblName, TableType.KUDU, schema);
         this.masterAddresses = masterAddresses;
         this.catalogName = catalogName;
         this.databaseName = dbName;
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/OdpsTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/OdpsTable.java
@@ -50,7 +50,7 @@ public class OdpsTable extends Table {
     }
 
     public OdpsTable(String catalogName, com.aliyun.odps.Table odpsTable) {
-        super(CONNECTOR_ID_GENERATOR.getNextId().asInt(), odpsTable.getName(), TableType.ODPS,
+        super(CONNECTOR_ID_GENERATOR.getNextId().asLong(), odpsTable.getName(), TableType.ODPS,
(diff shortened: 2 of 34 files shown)


CelerData-Reviewer:
@codex review

dirtysalt (approved):

Youngwb (approved):

HangyuanLiu (approved):

unsuitable001:
Are we planning to backport this to 3.5 as well?

PR #73348, opened 2026-05-15
[Enhancement] Wait for oldest PK-index memtable flush on publish path instead of falling back to sync OSS write

luohaha:
## Why I'm doing:

On shared-data PK tables, every compaction publish has to flush the
LakePersistentIndex memtable. When the inflight async-flush count hits
`pk_index_memtable_max_count`, the existing code falls back to a
**synchronous** OSS write on the publish-version thread. That serialises
the next publish behind one OSS round-trip, blocking compaction publish
for seconds at a time on busy tablets.

Observed on a 100 GB sortkey PK table (auto-perf-opt run pk-sortkey-opt,
iter-002): single compaction publish cost up to **41.5 s** with
`flush_times=20`, `pindex_memtable_replace_us=21.3 s`,
`flush_memtable_us=2.18 s`. Hardware was idle; the flush thread pool
(`pk_index_memtable_flush`) sat at peak 2 active out of 1024. The
serialisation cap is the per-tablet inflight count, not the pool size.

## What I'm doing:

When `flush_memtable` is about to add one more inflight memtable but the
cap is hit:

1. Wait for the **oldest** inactive memtable's async flush to complete
   via a new `PersistentIndexMemtable::wait_flush_done()` condvar
   primitive (added in `persistent_index_memtable.{h,cpp}`).
2. Drain the completed memtable into the fileset via the existing
   `merge_sstable_into_fileset` path.
3. Submit the current memtable to the same async pool. The publish
   thread never falls back to sync OSS I/O on its own critical path.

The wait blocks for the same wall time as the old sync flush would
have, but the I/O still runs on `pk_index_memtable_flush_thread_pool`,
which preserves cross-tablet concurrency. Only on pool submit failure
(e.g. shutting down) does the old sync fallback fire.

After this change, the same iter-002 workload measured per-CN publish
max = **574 ms** (auto-perf-opt iter-003) — a 72× reduction vs the
baseline. The change is independent of `pk_index_memtable_max_count`
and works with the default value of 2.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

The wait-flush path replaces the sync-flush fallback inside the same
`flush_memtable` API call; no observable behaviour outside performance.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

## Validation (auto-perf-opt run pk-sortkey-opt)

- iter-002 (deploy #73245 alone): per-CN publish max 41,500 → 1,443 ms.
- iter-003 (#73245 verified in isolation against baseline): per-CN publish max → **574 ms** (−98.6%), structurally eliminating the sync-flush stall.
- Tested on `1_100gb_sortkey` template (1 PK table, 500 M rows, 12 buckets, ORDER BY ≠ PK) on a 1 FE + 6 CN shared-data cluster.
- Source draft PR (validated on branch-4.1): #73245.

### DIFF
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -341,9 +341,24 @@ Status LakePersistentIndex::flush_memtable(bool force) {
         if (finish_point >= 0) {
             _inactive_memtables.erase(_inactive_memtables.begin(), _inactive_memtables.begin() + finish_point + 1);
         }
-        // 3. flush current memtable
+        // 3. If we're at the inflight cap, wait for the oldest async flush to
+        //    complete rather than synchronously flushing the current memtable on
+        //    the publish thread. The wall-clock back-pressure is the same, but
+        //    the I/O runs on the dedicated pk_index_memtable_flush thread pool
+        //    (preserving its concurrency across tablets), and the publish thread
+        //    can resume replace work on a new memtable as soon as one slot frees.
+        while (!_inactive_memtables.empty() &&
+               _inactive_memtables.size() + 1 >= config::pk_index_memtable_max_count) {
+            TRACE_COUNTER_SCOPE_LATENCY_US("wait_oldest_pk_flush_us");
+            _inactive_memtables.front()->wait_flush_done();
+            RETURN_IF_ERROR(_inactive_memtables.front()->flush_status());
+            auto sstable = _inactive_memtables.front()->release_sstable();
+            RETURN_IF_ERROR(merge_sstable_into_fileset(sstable));
+            _inactive_memtables.erase(_inactive_memtables.begin());
+        }
+        // 4. flush current memtable
         bool flush_async = false;
-        if (_inactive_memtables.size() + 1 < config::pk_index_memtable_max_count) {
+        if (_inactive_memtables.size() + 1 <= config::pk_index_memtable_max_count) {
             if (ExecEnv::GetInstance()->pk_index_memtable_flush_thread_pool()->submit(_memtable).ok()) {
                 flush_async = true;
             }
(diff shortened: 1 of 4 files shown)


luohaha:
Follow-up benchmark on 1 FE + 3 BE (32c64g BE, shared-data) cluster against upstream/main, two 20-min stress runs (`--ingest-threads 64 --ingest-rate 200000`), ~27K upsert publish samples.

**Server-side publish-trace summary (vs main baseline):**

| Metric | baseline | this PR | delta |
|---|---|---|---|
| Upsert publish P50 | 33 ms | 40 ms | +23% |
| Upsert publish P99 | 412 ms | 434 ms | +5% |
| Upsert publish P99.9 | 884 ms | 750 ms | -15% |
| Upsert publish max | 7 663 ms | 4 570 ms | **-40%** |
| Compaction publish P99 | 1 468 ms | 989 ms | **-33%** |
| Compaction publish max | 8 963 ms | 8 158 ms | -9% |
| `flush_memtable_us` max | 2 762 ms | 1 426 ms | -48% |

New `wait_oldest_pk_flush_us` trace counter is active — 29 events across the two runs, P99 = 1 360 ms (baseline has 0, as expected).

**Read:** the new path consistently trims tail and worst-case publish cost (max -40%, P99.9 -15%, compaction P99 -33%). The trade-off is a small steady-state penalty on the common-case upsert publish (P50 +23%, P99 +5%) — the wait-for-oldest path adds a few ms of synchronization even when the cap isn't actually hit.

Net effect is positive for any workload that cares about tail latency / publish-version stalls; neutral-to-mildly-negative for workloads dominated by p50.

PR #73365, opened 2026-05-15
[BugFix] carry load times as UTC epoch ms across BE/FE thrift

kevincai:
information_schema.loads silently dropped rows on any cluster whose
session timezone differs from Asia/Shanghai, because the FE-BE thrift
boundary exchanged datetimes as naive wall-clock strings:

- BE pushed pushdown literals (NOW() - INTERVAL 1 MINUTE etc.) as a
  session-zone wall-clock with no zone marker. FE parsed it via
  TimeUtils.timeStringToLong, which is locked to UTC+8, shifting the
  filter bound by (+08:00 - session_offset) hours.
- FE serialized job timestamps via TimeUtils.longToTimeString (also
  UTC+8 on main after the java.time refactor). BE re-evaluated the
  same pushdown predicate after materializing the row, comparing a
  UTC+8 wall-clock column against a session-zone literal, and dropped
  the rows the FE-side filter let through.

Fix: add unambiguous UTC epoch-ms siblings to the existing thrift
string fields and prefer them on both sides.

- TGetLoadsParams: +6 i64 *_ms fields (15-20). BE populates them by
  converting the literal's civil_second through the session time_zone
  via cctz::time_zone::lookup + std::min/std::max(pre, post) so DST
  fall-back / spring-forward ambiguities widen the FE prefilter
  rather than narrow it. The TimestampValue::to_unixtime(zone)
  overload is unusable here on two counts: it returns ms (not
  seconds), and TimezoneUtils::to_utc_offset resolves the offset at
  the Unix epoch, so DST-active zones (e.g. America/New_York in May)
  silently get the standard-time offset.
- TLoadInfo: +4 i64 *_ms fields (34-37). FE populates them with
  System.currentTimeMillis() values directly.
- schema_loads_scanner.cpp: fill_chunk materializes DATETIME columns
  via from_unixtime(ms/1000, session_tz) when *_ms is set, so the
  rendered column is in the session zone and matches the literal.
  information_schema.loads remains a second-precision view (same as
  pre-fix); sub-second is not supported on either FE filtering or BE
  materialization.
- LoadsSystemTable.LoadRequestFilter.from: prefers *_ms; falls back
  to the legacy UTC+8 string parse for old BEs during rolling
  upgrades (semantics-preserving, not introducing a third
  interpretation).

The legacy string fields are kept for cross-version compatibility:
old code on either side keeps working unchanged; the full fix only
activates when both sides know the new fields.

Note on precision: the new *_ms wire fields carry the full
System.currentTimeMillis() value, but all current consumers (BE
materialization, FE matchTimeRange) truncate to second precision so
the user-visible behavior of information_schema.loads matches the
pre-fix view. A follow-up PR will flip the truncation sites to
enable ms precision; that change is a ~5-line consumer-side flip
with no wire-format or data migration, will be backported to 4.1
only (not 4.0), and is documented inline at each flip site for
discoverability. LoadsHistorySyncer is unaffected by either today's
behavior or the future ms flip: its loads_history.load_finish_time
is plain DATETIME (second-precision at INSERT) so its watermark
caps at second regardless of upstream precision.

## Why I'm doing:

Introduced in #69472

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/schema_scanner/schema_column_filler.h
+++ b/be/src/exec/schema_scanner/schema_column_filler.h
@@ -36,6 +36,10 @@ void fill_data_column_with_slot(Column* data_column, void* slot) {
         result->append(date_value);
     } else if constexpr (IsTimestamp<ValueType>) {
         auto* date_time_value = (DateTimeValue*)slot;
+        // Microseconds are dropped here. TimestampValue::create has a default
+        // microsecond=0 7th parameter; pass date_time_value->microsecond() to
+        // preserve it. Today's callers all populate DateTimeValue at second
+        // precision, so this is a no-op for them.
         TimestampValue timestamp_value =
                 TimestampValue::create(date_time_value->year(), date_time_value->month(), date_time_value->day(),
                                        date_time_value->hour(), date_time_value->minute(), date_time_value->second());
--- a/be/src/exec/schema_scanner/schema_loads_scanner.cpp
+++ b/be/src/exec/schema_scanner/schema_loads_scanner.cpp
@@ -14,8 +14,11 @@
 
 #include "exec/schema_scanner/schema_loads_scanner.h"
 
+#include <algorithm>
 #include <boost/algorithm/string.hpp>
 
+#include "cctz/civil_time.h"
+#include "cctz/time_zone.h"
 #include "exec/schema_scanner/schema_helper.h"
 #include "exprs/column_ref.h"
 #include "exprs/expr_context.h"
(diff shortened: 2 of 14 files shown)


CelerData-Reviewer:
@codex review

kevincai:
more issues discovered on main branch than branch-4.0 due to refactor code https://github.com/StarRocks/starrocks/pull/66360 on branch-4.1 and main.

starrocks-xupeng (approved):

luohaha (approved):

trueeyu (approved):

gengjun-git (approved):

PR #73368, opened 2026-05-15
[Enhancement] Cost-penalize local streaming agg when group bys do not match actual distribution

OliLay:
## Why I'm doing:
We had a particular case where the optimizer chose a shuffle-join instead of a broadcast join. The shuffle-join destroyed the co-location of group by keys with partitions downstream, leading to a significantly worse query execution performance. Upon debugging, I found that the issue was that the optimizer assumed the local streaming aggregation can reduce the amount of rows (and requires only a small shuffle) since it assumed `global_ndv / # BEs` of output rows per backend. 
When the distribution is misaligned to the group by keys, this is not the case and the aggregation potentially sees all group keys and can barely reduce the amount of rows. As a result, you need to shuffle a lot of data (up to `global_ndv` per BE) afterwards.
This cost is hidden to the optimizer when it chooses the shuffle-join. A broadcast join leads a significantly better plan in this case. (~80s vs ~1s)

## What I'm doing:
I am introducing a network cost penalty for local split aggs when the distribution of data does not match the group by distribution. It adds an approximation of to be shuffled rows as cost. Ideally, the exchange operator downstream would price this correctly (since the agg itself does not really have a network cost), but since we only know the distribution in phase 2 of the optimization and statistics are derived once per group before physical properties (distribution) and are shared across all equivalent expressions in the group, we cannot adjust them per physical alternative. Instead, we model this through the network cost of the agg.

The penalty only fires when the agg estimates significant reduction (output < 50% of input) that won't actually materialize, avoiding false positives for aggs that already expect low reduction.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
@@ -353,8 +353,10 @@ public class CostModel {
             }
 
             // With mismatched distribution, each node sees all distinct groups.
-            final int numBackends = Math.max(1, ConnectContext.get().getAliveBackendNumber());
-            final double outputRows = Math.min(estimatedOutputRows * numBackends, inputRows);
+            // Include compute nodes for shared-data deployments, consistent with HashJoinCostModel.
+            final int numWorkers = Math.max(1, ConnectContext.get().getAliveBackendNumber()
+                    + GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getAliveComputeNodeNumber());
+            final double outputRows = Math.min(estimatedOutputRows * numWorkers, inputRows);
             final double toShuffleRows = Math.max(0, outputRows - estimatedOutputRows);
 
             return toShuffleRows * statistics.getAvgRowSize();
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
@@ -30,6 +30,7 @@ import com.starrocks.sql.optimizer.GroupExpression;
 import com.starrocks.sql.optimizer.JoinHelper;
 import com.starrocks.sql.optimizer.Utils;
 import com.starrocks.sql.optimizer.base.ColumnRefSet;
+import com.starrocks.sql.optimizer.base.DistributionCol;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/cost/CostModel.java
```diff
@@ -284,8 +285,79 @@ public CostEstimate visitPhysicalHashAggregate(PhysicalHashAggregateOperator nod
...
+                        .allMatch(dc -> groupByDistCols.stream()
+                                .anyMatch(groupByCol -> equivDesc.isConnected(groupByCol, dc)));
+
+                if (distributionCoversGroupBy) {
+                    return 0;
+                }
+            }
+
+            // With mismatched distribution, each node sees all distinct groups.
+            final int numBackends = Math.max(1, ConnectContext.get().getAliveBackendNumber());
```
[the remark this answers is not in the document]
OliLay: [e7edd7b](https://github.com/StarRocks/starrocks/pull/73368/commits/e7edd7b096c9bc81e093b38db888fae9baa01249)

fe/fe-core/src/test/java/com/starrocks/sql/plan/DistributedEnvPlanWithCostTest.java
```diff
@@ -1721,4 +1721,24 @@ public void testPartitionRangeGen() throws Exception {
         Assertions.assertEquals(19930101, keyRange.get(0).end_key);
         Assertions.assertEquals(1, scanNodes.getPartitionConjuncts().size());
     }
+
+    @Test
+    public void testLocalAggDistributionMismatchPenalty() throws Exception {
```
stephen-shelby, line 1726: The test's plan is identical with and without this PR's change. could you please provide a query dump of bad plan? we need to identify the root cause of the bad plan

stephen-shelby: we need to replay the bad plan . thx

PR #73371, opened 2026-05-15
[BugFix] Fix concurrent SegmentFlushTask race in DeltaWriter::commit()

kevincai:
Two SegmentFlushTask threads can reach DeltaWriter::commit() concurrently
on the same DeltaWriter when the upstream sends a duplicate
tablet_writer_add_segment(eos=true) RPC (e.g. a brpc retry on an
unstable peer link) and SegmentFlushToken dispatches the duplicates in
parallel. Without serialization the two commit() invocations race
inside RowsetWriter::build(): one moves _rowset_meta_pb into a new
RowsetMeta while the other is still iterating
TabletSchema::to_schema_pb(ts_pb) on the same TabletSchemaPB.
Concurrent calls to RepeatedPtrField::Add() corrupt the column field
(current_size_ > total_size_, observed as 45 vs 32 in the captured
core) and a subsequent add_column() returns a garbage ColumnPB* that
the next mutable_name() dereferences, producing SIGSEGV inside
InternalMetadata::PtrTag.

The previous state guard in commit() reads _state without _state_lock
and the kClosed -> kCommitted transition is only protected at the end
of the function. Two callers that both observe state == kClosed
cleanly fall past the existing switch and into the unsynchronised
body.

Serialise commit() by extracting its body into a private
_do_commit_body() helper and invoking it through std::call_once on a
per-writer std::once_flag. Exactly one caller runs the body;
concurrent duplicates block inside call_once until the body completes
and then return the captured _commit_result. This makes duplicate
commit() calls observably idempotent at the API level: both threads
return OK on success, or both return the same error on failure, with
no spurious "in progress" / AlreadyExist status leaking to callers
such as SegmentFlushTask::run (which would otherwise convert a
successful commit into an aborted one by calling _writer->cancel() on
any non-OK return).

The head-of-function state switch is unchanged and remains the fast
path for any subsequent commit() invocation: once the first call
returns, _state is kCommitted (or kAborted) and the switch returns
directly without entering call_once.

A doc comment is added at LocalTabletsChannel::add_segment describing
the no-dedup contract at the RPC entrypoint and the matching
guarantees the per-writer code is required to provide; the
SegmentFlushTask path relies on those guarantees and must not be
"helped" by adding dedup upstream without revisiting them.

## Why I'm doing:

```
*** Aborted at 1778639067 (unix time) try "date -d @1778639067" if you are using GNU date ***
PC: @          0x7896b1b starrocks::TabletColumn::to_schema_pb(starrocks::ColumnPB*) const
*** SIGSEGV (@0x0) received by PID 2279730 (TID 0x1528895fe700) LWP(2280453) from PID 0; stack trace: ***
    @     0x15291a36e20b __pthread_once_slow
    @          0xbe2b194 google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @     0x15291a377630 (/usr/lib64/libpthread-2.17.so+0xf62f)
    @          0x7896b1b starrocks::TabletColumn::to_schema_pb(starrocks::ColumnPB*) const
    @          0x7897317 starrocks::TabletSchema::to_schema_pb(starrocks::TabletSchemaPB*) const
    @          0x84bf944 starrocks::RowsetWriter::build()
    @          0x84c8ea4 starrocks::HorizontalRowsetWriter::build()
    @          0x7b489f6 starrocks::DeltaWriter::commit()
    @          0x7767cc3 starrocks::SegmentFlushTask::run()
    @          0x45d2a07 starrocks::ThreadPool::dispatch_thread()
    @          0x45c8ef0 starrocks::Thread::supervise_thread(void*)
    @     0x15291a36fea5 start_thread
    @     0x15291922cb0d __clone
[1778639067.410][thread: 23263847638784] je_mallctl execute purge success
[1778639067.410][thread: 23263847638784] je_mallctl execute dontdump success
```

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/delta_writer.cpp
+++ b/be/src/storage/delta_writer.cpp
@@ -14,6 +14,7 @@
 
 #include "storage/delta_writer.h"
 
+#include <mutex>
 #include <utility>
 
 #include "column/raw_data_visitor.h"
@@ -744,6 +745,23 @@ Status DeltaWriter::commit() {
         break;
     }
 
+    // Exactly one thread runs the commit body. Two SegmentFlushTask threads
+    // can reach this point concurrently for the same DeltaWriter when the
+    // upstream sends a duplicate tablet_writer_add_segment(eos=true) (e.g.
+    // a brpc retry on an unstable peer link) and SegmentFlushToken
+    // dispatches the duplicates in parallel. Concurrent duplicate callers
+    // block inside std::call_once until the first invocation completes
+    // and then return the captured _commit_result. This makes duplicate
+    // commit() calls observably idempotent: both threads return OK on
+    // success, or both return the same error on failure, with no transient
+    // "in progress" status leaking to callers such as SegmentFlushTask::run
+    // (which would otherwise cancel the writer on any non-OK return and
+    // convert a successful commit into a reported failure on the primary).
+    std::call_once(_commit_once, [this] { _commit_result = _do_commit_body(); });
+    return _commit_result;
+}
+
+Status DeltaWriter::_do_commit_body() {
     MonotonicStopWatch watch;
     watch.start();
     if (auto st = _flush_token->wait(); UNLIKELY(!st.ok())) {
--- a/be/src/runtime/local_tablets_channel.cpp
+++ b/be/src/runtime/local_tablets_channel.cpp
@@ -199,6 +199,35 @@ Status LocalTabletsChannel::open(const PTabletWriterOpenRequest& params, PTablet
 
 void LocalTabletsChannel::add_segment(brpc::Controller* cntl, const PTabletWriterAddSegmentRequest* request,
                                       PTabletWriterAddSegmentResult* response, google::protobuf::Closure* done) const {
+    // NOTE: This entrypoint does NOT deduplicate duplicate RPCs (unlike
+    // add_chunk below, which dedups by packet_seq sliding window).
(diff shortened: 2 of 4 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/delta_writer.cpp
```diff
@@ -744,6 +752,52 @@ Status DeltaWriter::commit() {
...
+            auto err = _err_status;
+            lk.unlock();
+            if (final_state == kCommitted) {
+                return Status::OK();
+            }
+            return err.ok() ? Status::InternalError(
+                                      fmt::format("Delta writer has been aborted. tablet_id: {}", _opt.tablet_id))
+                            : err;
+        }
+    }
```
xiangguangyxg: Is it possible to simplify the implementation using std::call_once?

xiangguangyxg (approved):

banmoy (approved):

wyb (approved):

PR #73395, opened 2026-05-17
[Refactor] Centralize refresh-other-FE concurrency management in GlobalStateMgr

dirtysalt:
## Why I'm doing:
Refreshing external table metadata on peer FEs previously mixed two concerns:

1. background dispatch from connector write paths
2. per-peer RPC fan-out to other FEs

Those concerns were partially implemented inside individual connectors, and some paths created ad hoc threads for each refresh. This made concurrency harder to reason about, harder to tune globally, and less efficient when multiple catalogs were active.

This change makes the refresh path easier to manage by moving the concurrency control into `GlobalStateMgr`, so FE-wide capacity is bounded and shared consistently.

## What I'm doing:
- Centralize "refresh other FE" execution in `GlobalStateMgr`.
- Split the flow into two explicit stages with two FE-global executors:
  - dispatch executor: runs background refresh jobs asynchronously
  - rpc executor: fans out bounded refresh RPCs to peer FEs
- Keep both sync and async entry points in `GlobalStateMgr`:
  - sync callers wait for peer refresh completion
  - async callers return immediately and reuse the same bounded RPC fan-out path
- Remove per-catalog refresh executor usage from Hive and Iceberg write paths, so connectors no longer manage this concurrency themselves.
- Add FE config knobs and comments for the new global executors.
- Add unit tests covering success, failure aggregation, RPC exception handling, and executor usage.

High-level flow after this change:

```text
async connector path
    |
    +--> GlobalStateMgr.refreshOthersFeTableAsync(...)
             |
             +--> dispatch executor
                      |
                      +--> refreshOthersFeTable(...)
                               |
                               +--> rpc executor
                                        |
                                        +--> refresh peer FE #1
                                        +--> refresh peer FE #2
                                        +--> refresh peer FE #3
```

This keeps behavior the same at a functional level, but makes resource control FE-global instead of per-catalog.

## Config Notes

This change introduces two FE-level dynamic configs for managing "refresh other FE" concurrency:

- `refresh_other_fe_dispatch_executor_thread_num`
  Controls the size of the global dispatch executor used by asynchronous refresh requests from connector write paths.
  These threads only schedule background refresh work and do not directly send peer FE RPCs.

- `refresh_other_fe_rpc_executor_thread_num`
  Controls the size of the global RPC fan-out executor used to send refresh requests to peer FEs.
  This is the actual concurrency cap for cross-FE refresh RPCs, and it is shared by both synchronous and asynchronous refresh flows.

High-level config behavior:

```text
async connector path
    -> dispatch executor
    -> refresh orchestration
    -> rpc executor
    -> peer FE refresh RPCs

sync DDL path
    -> refresh orchestration
    -> rpc executor
    -> peer FE refresh RPCs
```

In short:
- `dispatch_executor_thread_num` limits how many background refresh jobs can be started at once.
- `rpc_executor_thread_num` limits how many peer FE refresh RPCs can run concurrently.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/server/GlobalStateMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/GlobalStateMgr.java
@@ -281,7 +281,6 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
-import java.util.concurrent.FutureTask;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -543,6 +542,8 @@ public class GlobalStateMgr {
     private final DDLStmtExecutor ddlStmtExecutor;
     private final ShowExecutor showExecutor;
     private final ExecutorService queryDeployExecutor;
+    private final ExecutorService refreshOtherFeDispatchExecutor;
+    private final ExecutorService refreshOtherFeRpcExecutor;
     private final WarehouseIdleChecker warehouseIdleChecker;
 
     private final ClusterSnapshotMgr clusterSnapshotMgr;
@@ -893,6 +894,16 @@ public class GlobalStateMgr {
         this.queryDeployExecutor =
                 ThreadPoolManager.newDaemonFixedThreadPool(Config.query_deploy_threadpool_size, Integer.MAX_VALUE,
                         "query-deploy", true);
+        this.refreshOtherFeDispatchExecutor = ThreadPoolManager.newDaemonFixedThreadPool(
+                Config.refresh_other_fe_dispatch_executor_thread_num,
+                Math.max(16, Config.refresh_other_fe_dispatch_executor_thread_num * 4),
+                "refresh-other-fe-dispatch",
+                true);
+        this.refreshOtherFeRpcExecutor = ThreadPoolManager.newDaemonFixedThreadPool(
+                Config.refresh_other_fe_rpc_executor_thread_num,
+                Math.max(16, Config.refresh_other_fe_rpc_executor_thread_num * 4),
+                "refresh-other-fe-rpc",
+                true);
 
         this.warehouseIdleChecker = new WarehouseIdleChecker();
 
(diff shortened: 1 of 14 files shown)


CelerData-Reviewer:
@codex review

Youngwb (approved):

Youngwb:
@dirtysalt These config change need discuss with @wangsimo0

wangsimo0 (approved):

fe/fe-core/src/main/java/com/starrocks/server/GlobalStateMgr.java
```diff
@@ -2777,39 +2817,72 @@ public void refreshOthersFeTable(TableName tableName, List<String> partitions, b
...
     }
 
-    public Future<TStatus> refreshOtherFesTable(TNetworkAddress thriftAddress, TableName tableName,
-                                                List<String> partitions) {
+    /**
+     * Enqueue a background "refresh other FE" job and return immediately to the caller.
+     * The dispatched task eventually reuses {@link #refreshOthersFeTable(TableName, List, boolean)}
+     * so asynchronous and synchronous refreshes share the same peer RPC concurrency limit.
+     */
+    public Future<?> refreshOthersFeTableAsync(TableName tableName, List<String> partitions) {
```
stephen-shelby, line 2825: One subtle point: refreshOthersFeTableAsync is fire-and-forget in normal cases, but it may still block the connector write path when the dispatch executor is saturated. newDaemonFixedThreadPool uses BlockedPolicy, so once the dispatch pool and its bounded queue are full, submit() can wait up to 60s before rejecting. Is that acceptable for the async commit path, or should this dispatch layer fail fast / drop-and-log when saturated?

dirtysalt: I see in IcebergMetadata and HiveCommitter are calling this function but not handling failure.  

We handle this rejection right now and return failed-future.

stephen-shelby (approved):

dirtysalt:
@mergify backport branch-4.1

PR #73397, opened 2026-05-17
[Enhancement] Complete Iceberg timestamptz partition transform support by filling the sink path gap

dirtysalt:
## Why I'm doing:
StarRocks already has read-path handling for Iceberg `timestamptz` / `withZone()` values, including partition value normalization and predicate conversion. However, the sink path still only supported partition transforms correctly for timestamp without time zone.

This left a gap for users with existing Iceberg tables partitioned on `timestamptz` columns: reads were already timezone-aware, but writes could not route records through `year/month/day/hour` transforms with correct `timestamptz` semantics.

The tricky part is that sink partition routing must be based on the absolute instant represented by the value, not on the session-local wall clock.

## What I'm doing:
This PR completes the end-to-end story by filling the missing sink-side support for Iceberg `timestamptz` partition transforms, while keeping the existing read-path behavior unchanged.

At a high level:

```text
Read path
  already supports withZone() handling

Write path
  FE planner detects withZone()
    -> uses __iceberg_transform_timestamptz_*
    -> BE evaluates transform from the UTC instant
    -> sink writes to the correct Iceberg partition
```

Specifically:
- Keep the existing read path as-is and make the PR scope explicit: the missing piece was sink partition routing.
- Add dedicated internal transform functions for Iceberg `timestamptz` partition transforms.
- Update FE planning so `withZone()` columns use the timestamptz transform path instead of the NTZ path.
- Remove the previous FE sink-side rejection for timezone partition columns.
- Implement BE evaluation so `year/month/day/hour` are computed from the normalized UTC instant, not from session-local wall clock values.
- Add a UTC fast path to avoid unnecessary normalization when the session timezone is effectively UTC.
- Add FE and BE unit tests to cover planner selection and timezone-sensitive transform semantics.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/time_functions.cpp
+++ b/be/src/exprs/time_functions.cpp
@@ -178,6 +178,75 @@ Status TimeFunctions::convert_tz_close(FunctionContext* context, FunctionContext
     return Status::OK();
 }
 
+namespace {
+
+cctz::time_zone get_iceberg_timestamptz_session_timezone(FunctionContext* context) {
+    if (context != nullptr && context->state() != nullptr) {
+        return context->state()->timezone_obj();
+    }
+    return cctz::utc_time_zone();
+}
+
+bool should_use_iceberg_timestamptz_utc_fast_path(const cctz::time_zone& timezone) {
+    if (timezone == cctz::utc_time_zone()) {
+        return true;
+    }
+
+    auto epoch = cctz::time_point<cctz::seconds>();
+    cctz::time_zone::civil_transition transition;
+    return timezone.lookup(epoch).offset == 0 && !timezone.next_transition(epoch, &transition);
+}
+
+bool normalize_iceberg_timestamptz_to_utc(FunctionContext* context, const TimestampValue& timestamp,
+                                          TimestampValue* normalized_timestamp) {
+    int year, month, day, hour, minute, second, usec;
+    timestamp.to_timestamp(&year, &month, &day, &hour, &minute, &second, &usec);
+    DateTimeValue datetime(TIME_DATETIME, year, month, day, hour, minute, second, usec);
+
+    int64_t unix_second;
+    if (!datetime.unix_timestamp(&unix_second, get_iceberg_timestamptz_session_timezone(context))) {
+        return false;
+    }
+
+    normalized_timestamp->from_unix_second(unix_second, usec);
+    return true;
+}
+
+template <typename TransformFn>
+StatusOr<ColumnPtr> evaluate_iceberg_timestamptz_transform(FunctionContext* context, const Columns& columns,
+                                                           TransformFn transform_fn) {
+    DCHECK_EQ(columns.size(), 1);
+
+    auto datetime_viewer = ColumnViewer<TYPE_DATETIME>(columns[0]);
+    auto size = columns[0]->size();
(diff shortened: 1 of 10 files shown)


CelerData-Reviewer:
@codex review

Wenjun7J (approved):

be/src/exprs/time_functions.cpp
```diff
@@ -178,6 +178,91 @@ Status TimeFunctions::convert_tz_close(FunctionContext* context, FunctionContext
...
+cctz::time_zone get_iceberg_timestamptz_session_timezone(FunctionContext* context) {
+    if (context != nullptr && context->state() != nullptr) {
+        return context->state()->timezone_obj();
+    }
+    return cctz::utc_time_zone();
+}
+
+bool should_use_iceberg_timestamptz_utc_fast_path(const cctz::time_zone& timezone) {
+    return timezone == cctz::utc_time_zone();
+}
```
GavinMar, line 192: `timezone == cctz::utc_time_zone()` compares the internal `Impl*` pointer. `RuntimeState::timezone_obj()` for `SET time_zone = 'UTC'` typically goes through `cctz::load_time_zone("UTC", &tz)`, whose result is not guaranteed to be pointer-equal to the `utc_time_zone()` singleton, which cause the fast path then silently falls back to the slow per-row path — correctness is fine.

  Suggest comparing by name `(timezone.name() == "UTC" || timezone.name() == "Etc/UTC")`

dirtysalt: source code 

```
#include <cassert>
#include <iostream>
#include <string>

#include "cctz/time_zone.h"
#include "cctz/civil_time.h"

bool should_use_iceberg_timestamptz_utc_fast_path(const cctz::time_zone& timezone) {
    return timezone == cctz::utc_time_zone();
}

void test_timezone(const std::string& tz_name) {
    cctz::time_zone tz;
    bool ok = cctz::load_time_zone(tz_name, &tz);

    assert(ok);

    std::cout << "timezone = " << tz_name
              << ", fast_path = "
              << should_use_iceberg_timestamptz_utc_fast_path(tz)
              << std::endl;
}

int main() {
    // exact UTC
    test_timezone("UTC");

    // aliases
    test_timezone("Etc/UTC");
    test_timezone("Etc/GMT");
    test_timezone("GMT");

    // non-UTC
    test_timezone("Asia/Shanghai");
    test_timezone("America/New_York");

    return 0;
}
```

output

```
timezone = UTC, fast_path = 1
timezone = Etc/UTC, fast_path = 0
timezone = Etc/GMT, fast_path = 0
timezone = GMT, fast_path = 0
timezone = Asia/Shanghai, fast_path = 0
timezone = America/New_York, fast_path = 0
```

GavinMar (approved):

trueeyu (approved):

stephen-shelby (approved):

wangsimo0 (approved):

PR #73400, opened 2026-05-17
[Feature] Iceberg footer prefetch via CACHE SELECT

### LINKED ISSUE #73401: [Feature Request] Support DataCache warmup for new connector data files after metadata refresh
## Feature request

DataCache population for data lake files is currently driven by query reads or by an explicit `CACHE SELECT`. This means DataCache can react to blocks that were read, but it is not told when connector metadata refresh changes the set of remote data files visible for a table.

A connector can observe that table state changed before the first query reaches the newly visible files. For example, an Iceberg metadata refresh can discover a new table snapshot and a new set of data files while the corresponding DataCache blocks are still cold.

Without that signal, a hot table can still pay a cold-read cost after metadata refresh. The first query that reaches newly visible files must fetch the required file regions from remote storage before those regions can be reused from DataCache.

The cache action is not always obvious. Warming more data can reduce later remote reads, but it can also consume cache capacity and evict blocks that are more valuable to keep. A conservative example is warming reader-initialization regions for Parquet or ORC files: those reads are small and are needed before normal scans can proceed, but the broader design should not assume that this is always the best cache action.

Add proactive DataCache maintenance for connector table changes.

This requires two new pieces:

- a connector-to-DataCache signal from metadata refresh when the visible remote data files of a table change;
- cache-side state and policy that can decide whether that signal should cause proactive cache work, and which cache work is worth doing.

The connector-side signal should describe what the connector knows about the refreshed table state. At minimum, a connector may be able to report that the table state changed. Some connectors may also be able to provide more precise file-set information, such as the current visible files or files that were added or removed by the refresh.

The cache-side part is not just a trigger handler. DataCache needs enough state and policy to evaluate proactive work against the cache it is protecting: expected future reuse, remote-read cost avoided, cache capacity consumed, and the cost of evicting other cached data.

DataCache should use the connector signal together with that cache-side state and policy to decide which cache action is useful:

- do nothing when proactive cache work is unlikely to help;
- warm a small required region of newly relevant files when that is a low-cost way to avoid cold first-query latency;
- warm more file data when the expected reuse justifies the cache cost;
- evict data for files that are no longer relevant when the connector provides enough information and eviction is appropriate.

The cache decision should account for both sides of proactive population: the expected benefit of avoiding future remote reads and the expected cost of cache capacity consumed or other cached data evicted.

The design should define the contract between connector metadata refresh and DataCache:

- what table-change or file-set information a connector may provide;
- which parts of that signal are required for a basic integration and which are optional when a connector has more precise file knowledge;
- which cache actions DataCache may derive from that signal.

The design should also define the DataCache-side decision model:

- what cache-side state DataCache needs to track for proactive decisions;
- how it estimates the expected reuse and remote-read benefit of a candidate warmup action;
- how it estimates the cache cost of that action, including capacity consumption and eviction of other cached data;
- how it chooses between doing nothing, warming a small required region, warming more data, or evicting no-longer-relevant data.

`CACHE SELECT` is the existing legacy stopgap: it gives users at least one manual way to populate DataCache when they know what data to warm. A proactive connector-to-DataCache path, with cache-side decisions about whether and how to act on table changes, can eliminate the need for administrators to drive cache population manually.

## Simplified alternative solution

The solution above requires a new connector-to-DataCache signal and a DataCache-side decision model. A smaller alternative is to avoid that policy work in the first version and always warm only reader-initialization data for refreshed files.

The implementation path for most of this simplified alternative already exists in #73400; this section spells out the heuristic and its tradeoffs explicitly.

This is a heuristic rather than a full proactive cache-maintenance model. The reason to start there is that reader-initialization data is small compared with full file data and is required before a normal scan can proceed. For Parquet and ORC files, this covers the file regions needed to initialize the reader, such as footer or file metadata reads.

### 1. Use reader-initialization warmup as the fixed cache action

Instead of estimating the benefit and cache cost of candidate warmup actions, the simplified path uses one fixed action:

- warm reader-initialization data for newly visible files;
- do not warm regular file data;
- do not add targeted DataCache eviction.

This reduces cache pollution risk compared with full-file warmup, but it does not make DataCache proactive in the general sense described above.

### 2. Extend the legacy `CACHE SELECT` path with a scoped mode

Reuse the existing manual warmup path by extending `CACHE SELECT` with a mode that warms only reader-initialization reads:

```sql
CACHE SELECT *
FROM iceberg_catalog.db.table
PROPERTIES("mode" = "reader_init");
```

The current `CACHE SELECT` behavior remains the default mode. The `reader_init` mode asks the warmup scan to populate only file regions read while the scan reader is being initialized, before regular column or row data is fetched.

For the motivating formats:

- Parquet warms reader-initialization reads that load the file footer and file metadata.
- ORC warms reader-initialization reads that load the postscript and file metadata.

The mode is named after the reader-initialization phase rather than `footer` because the required initialization reads are format-dependent.

### 3. Trigger asynchronous `CACHE SELECT` after refresh

Use asynchronous `CACHE SELECT` tasks to make the simplified path usable after metadata refresh.

The warmup may be triggered in two ways:

- manually, when an administrator knows a refreshed table should be warmed;
- from a connector refresh path, for example after Iceberg refresh installs a refreshed table state.

For an Iceberg catalog integration, the connector-triggered path can enqueue the equivalent table-scoped warmup after refresh observes a new table metadata location:

```sql
CACHE SELECT *
FROM iceberg_catalog.db.table
PROPERTIES("mode" = "reader_init");
```

The warmup is a side effect of refresh, not part of refresh correctness. Refresh should enqueue it asynchronously after the refreshed table state is installed, should not wait for warmup completion, and should not fail or roll back when warmup fails.

### 4. Opt in from the catalog

Connector-triggered warmup is not enabled by metadata refresh alone. For the simplified path, enable it explicitly with a catalog property on the connector catalog that should issue warmup after refresh.

This keeps the manual `CACHE SELECT` path unchanged and makes the extra refresh-side work visible at the catalog boundary where it is introduced.

### 5. Choose the warmup scope

The smallest implementation is table-scoped. It reuses the existing `CACHE SELECT` surface and enqueues `reader_init` warmup for the refreshed table:

```sql
CACHE SELECT *
FROM iceberg_catalog.db.table
PROPERTIES("mode" = "reader_init");
```

That scope may repeat work for files that were already visible before refresh, but it requires the fewest changes to the legacy warmup path.

Iceberg can identify newly visible files after refresh and could use that information to reduce the work done by `CACHE SELECT`. This can make the simplified path cheaper, but it is still a half measure: the warmup action remains fixed to reader-initialization data, and DataCache still does not decide whether warming those files is worth the cache cost.

### Observability for the simplified heuristic

The simplified path replaces a DataCache-side cost model with a fixed reader-initialization warmup action. It therefore needs observability for both the expected benefit and the possible cache-pollution cost.

For the expected benefit, scan profiles should show whether the warmed reader-initialization data is actually reused. For the initial Parquet use case, this means footer hit and miss metrics around reader initialization reads. BE-level DataCache hit and miss metrics should also be monitored to verify that proactive warmup improves cache behavior for the workload as a whole.

The same BE-level DataCache hit and miss metrics are needed for the possible cost. They make it possible to detect a regression where proactively warmed initialization regions start displacing more valuable cached data blocks.

Connector-triggered warmup execution should be observable separately: whether a warmup task was enqueued, completed, skipped, or failed. Warmup failure remains separate from metadata refresh correctness.

### Limitations of the simplified alternative

The simplified alternative is useful as a first heuristic, but it does not know which proactive cache work is actually valuable for the workload.

It does not evict cached data for files that are no longer relevant after metadata refresh. Removed or obsolete files can therefore continue to occupy DataCache until normal eviction removes them.

It warms every reader-initialization target selected by the warmup scope because it does not know which newly visible files are likely to be read soon. The table-scoped version is the broadest form of this tradeoff and may revisit files that were already visible before refresh.

For the same reason, it warms only reader-initialization data instead of regular file data. That keeps the amount of proactive population smaller and targets data needed before normal scans can proceed, but it leaves regular data blocks cold until a user query reads them or an administrator warms them manually with `CACHE SELECT`.

Even this narrower warmup can still be harmful when cache capacity is tight or the warmed files are not read soon. Reader-initialization blocks may displace more valuable cached data and reduce DataCache hit rate for the workload.

**Additional context**

The motivating symptom was observed for cold Parquet/ORC reader initialization data after Iceberg metadata refresh.

A previous implementation attempt showed that the problem is real, but review raised design questions before adding new syntax or background warmup behavior. The next step should be a design discussion that aligns the user-facing warmup model, connector trigger model, and resource controls before implementation.

eshishki:
## Why I'm doing it

When CachingIcebergCatalog refreshes a table's metadata, the freshly-known
data files are not yet present in BE/CN block_cache. The first user query
on the table pays a 50–150ms S3 round-trip per file to read the parquet
footer, multiplied across all files in the scan. On wide cold scans this
dominates first-query latency.

## What I'm doing

Fire-and-forget a CACHE SELECT against the just-refreshed table from inside
`CachingIcebergCatalog.refreshTable` (both the cold-start path and the
metadata-change path). The CACHE SELECT runs through the same scan path as
real queries — `DataCacheSelectExecutor` → `StmtExecutor` → `Coordinator`
→ `HDFSBackendSelector` → `CacheSelectScanner` — so file→node placement
matches what real SELECTs from the same session would compute.

Dispatched onto a per-catalog single-thread daemon executor; duplicate
triggers on the same `(catalog, db, table, warehouse)` are coalesced.

Knobs:
- `enable_iceberg_metadata_refresh_footer_prefetch` (session var, default
  `false`, FE-only — not propagated to BE). Set globally to enable for
  background refresh.
- `cache_select_footer_only` (INVISIBLE session var, only set internally
  by the prefetcher). BE branch in `_fetch_parquet`, `_fetch_orc`,
  `_fetch_textfile`, plus the iceberg-delete-files skip in `do_get_next`.

Generic fix (not feature-gated): `HdfsScanner::create_random_access_file`
now wraps `_file` in a regular `CacheInputStream` for cache_select mode
too. Without this, the parquet `reader->init` footer read went straight
to raw storage even during CACHE SELECT — only the column ranges
explicitly fed to `CacheSelectInputStream` via `_write_disk_ranges` were
populating block_cache. Now the footer read also populates.

New thrift field: `TQueryOptions.cache_select_footer_only` at ordinal 220
(optional, default false).

Metrics:
- `parquet_footer_cache_hit_count` / `parquet_footer_cache_miss_count`
  on `QueryScanMetrics`.
- `datacache_block_cache_write_bytes` / `datacache_block_cache_evict_bytes`
  gauges on `DataCacheMetrics` (sampled from starcache `detail_l2`).

## What type of PR is it

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Default-off feature gate; no existing knob's behaviour or default changes.

## Does this PR need documentation update?

- [ ] No.
- [x] Yes.

New session variable `enable_iceberg_metadata_refresh_footer_prefetch` to
document.

## Backport to release branches

- [ ] 4.0
- [ ] 3.5
- [ ] 3.4
- [ ] 3.3
- [ ] 3.2
- [ ] 3.1
- [ ] 3.0

## Tests

- [ ] Unit tests
- [ ] Manual test on a dev cluster
- [ ] Bench

## Refer issue

Fixes #73401

### DIFF
--- a/be/src/cache/datacache_metrics.cpp
+++ b/be/src/cache/datacache_metrics.cpp
@@ -56,6 +56,8 @@ void DataCacheMetrics::install(MetricRegistry* registry) {
     registry->register_metric("datacache_meta_used_bytes", &datacache_meta_used_bytes);
     registry->register_metric("block_cache_hit_bytes", &block_cache_hit_bytes);
     registry->register_metric("block_cache_miss_bytes", &block_cache_miss_bytes);
+    registry->register_metric("datacache_block_cache_write_bytes", &datacache_block_cache_write_bytes);
+    registry->register_metric("datacache_block_cache_evict_bytes", &datacache_block_cache_evict_bytes);
 }
 
 void DataCacheMetrics::enable_update_hook(bool use_same_instance) {
@@ -99,6 +101,18 @@ void DataCacheMetrics::update() {
     datacache_disk_quota_bytes.set_value(disk_metrics.disk_quota_bytes);
     datacache_disk_used_bytes.set_value(disk_metrics.disk_used_bytes);
     datacache_meta_used_bytes.set_value(meta_used_bytes);
+
+    // Starcache detail_l2 carries cumulative populate (write_bytes) and eviction (remove_bytes)
+    // totals. Sample them so dashboards can derive eviction pressure (evict-rate vs write-rate)
+    // without scraping the HTTP /api/datacache/stat endpoint.
+    if (local_disk_cache != nullptr && local_disk_cache->is_initialized()) {
+        auto* starcache = static_cast<StarCacheEngine*>(local_disk_cache);
+        auto detail_metrics = starcache->starcache_metrics(2);
+        if (detail_metrics.detail_l2 != nullptr) {
+            datacache_block_cache_write_bytes.set_value(detail_metrics.detail_l2->write_bytes);
+            datacache_block_cache_evict_bytes.set_value(detail_metrics.detail_l2->remove_bytes);
+        }
+    }
 
     // Update hit rate metrics from DataCacheHitRateCounter
     auto* hit_rate_counter = DataCacheHitRateCounter::instance();
--- a/be/src/cache/datacache_metrics.h
+++ b/be/src/cache/datacache_metrics.h
@@ -44,6 +44,12 @@ public:
     METRIC_DEFINE_INT_GAUGE(datacache_meta_used_bytes, MetricUnit::BYTES);
     METRIC_DEFINE_INT_ATOMIC_COUNTER(block_cache_hit_bytes, MetricUnit::BYTES);
     METRIC_DEFINE_INT_ATOMIC_COUNTER(block_cache_miss_bytes, MetricUnit::BYTES);
+    // Starcache-internal cumulative populate/eviction byte totals (detail_l2). Snapshot semantics:
+    // gauges set to absolute cumulative bytes; Prom rate() over them gives per-second populate
+    // and eviction pressure for datacache health dashboards (alerts when evict-rate approaches
+    // write-rate, indicating cache churn).
+    METRIC_DEFINE_INT_GAUGE(datacache_block_cache_write_bytes, MetricUnit::BYTES);
+    METRIC_DEFINE_INT_GAUGE(datacache_block_cache_evict_bytes, MetricUnit::BYTES);
 
 private:
     MetricRegistry* _registry = nullptr;
--- a/be/src/exec/hdfs_scanner/cache_select_scanner.cpp
+++ b/be/src/exec/hdfs_scanner/cache_select_scanner.cpp
@@ -20,6 +20,7 @@
 #include "formats/parquet/file_reader.h"
 #include "fs/fs.h"
 #include "io/shared_buffered_input_stream.h"
+#include "runtime/runtime_state.h"
 
 namespace starrocks {
 
@@ -68,7 +69,10 @@ Status CacheSelectScanner::do_get_next(RuntimeState* runtime_state, ChunkPtr* ch
     }
 
     // handle iceberg delete files
-    if (!_scanner_params.deletes.empty()) {
+    // Skip delete-file fetching in footer-only warm-up mode: this path is the iceberg metadata
+    // refresh footer prefetch, which only needs the parquet footer in block_cache. Delete files
+    // are read by real query execution as needed.
+    if (!_scanner_params.deletes.empty() && !runtime_state->query_options().cache_select_footer_only) {
         RETURN_IF_ERROR(_fetch_iceberg_delete_files());
     }
 
@@ -87,6 +91,14 @@ Status CacheSelectScanner::_fetch_orc() {
     } catch (std::exception& e) {
         return Status::InternalError(
                 strings::Substitute("CacheSelectScanner::_fetch_orc failed. reason = $0", e.what()));
+    }
+
+    // Footer-only warm-up: createReader above already read the ORC postscript + file footer
+    // through the wrapping CacheInputStream, populating block_cache. Stripe data ranges are not
+    // needed; skip the per-column resolution and stripe IO collection. Symmetric to the
+    // parquet branch.
+    if (_runtime_state->query_options().cache_select_footer_only) {
+        return Status::OK();
     }
 
     // prepare SlotDescriptor
@@ -184,6 +196,15 @@ Status CacheSelectScanner::_fetch_parquet() {
 
     RETURN_IF_ERROR(reader->init(&_scanner_ctx));
 
+    // Footer-only warm-up for the iceberg metadata refresh prefetch path: reader->init above
+    // already read the footer through the CacheInputStream wrapping _file (created in
+    // HdfsScanner::create_random_access_file when enable_cache_select is on), which populates
+    // block_cache with the footer bytes. Column ranges are not needed; skip the per-column
+    // io-range collection and explicit _write_disk_ranges.
+    if (_runtime_state->query_options().cache_select_footer_only) {
+        return Status::OK();
+    }
+
     std::vector<io::SharedBufferedInputStream::IORange> io_ranges{};
     RETURN_IF_ERROR(reader->collect_scan_io_ranges(&io_ranges));
 
@@ -198,6 +219,13 @@ Status CacheSelectScanner::_fetch_parquet() {
 
 // Split text into multiply disk ranges, then fetch it
 Status CacheSelectScanner::_fetch_textfile() {
+    // Footer-only warm-up has no meaning for row-oriented text files (there is no footer to
+    // populate). The flag is set by the iceberg metadata refresh prefetch path, which only ever
+    // targets parquet/orc; skip entirely if it leaks onto a text format.
+    if (_runtime_state->query_options().cache_select_footer_only) {
+        return Status::OK();
+    }
+
     // If it's compressed file, we only handle scan range whose offset == 0.
     if (get_compression_type_from_path(_scanner_params.path) != UNKNOWN_COMPRESSION &&
         _scanner_params.scan_range->offset != 0) {
--- a/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
+++ b/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
@@ -24,6 +24,7 @@
 #include "common/config_scan_io_fwd.h"
 #include "connector/deletion_vector/deletion_vector.h"
 #include "exec/pipeline/fragment_context.h"
+#include "exec/query_scan_metrics.h"
 #include "exprs/chunk_predicate_evaluator.h"
 #include "fs/hdfs/fs_hdfs.h"
 #include "io/cache_select_input_stream.hpp"
@@ -35,6 +36,7 @@
 #include "storage/type_info_allocator_adapter.h"
 #include "storage/types.h"
 #include "types/timestamp_value.h"
+
 namespace starrocks {
 
 static const std::string kCountOptColumnName = "___count___";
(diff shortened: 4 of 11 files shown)


CelerData-Reviewer:
@codex review

stdpain:
@GavinMar

GavinMar:
Thank you for your contribution. The latency problem this targets (cold parquet/orc footers on first-query after a metadata refresh) is real. However, I feel that the current implementation is not elegant from the overall functional design of Starrocks.

**1. Metadata refresh and data refresh are intentionally separate concerns**
In StarRocks today the FE-side metadata refresh path (CachingIcebergCatalog, snapshot/manifest/data-file caches, the periodic ConnectorTableMetadataProcessor) and the BE-side data warm-up path (block_cache, page_cache, CACHE SELECT) are two different mechanisms, addressing different layers. Footers are physically part of the data files — their lifecycle naturally belongs to the data-cache mechanism, not the metadata-cache one. Coupling the two inside CachingIcebergCatalog.refreshTable blurs a boundary that was deliberately drawn.

**2. Metadata refresh has so far been a FE-only event**
Up to now, refreshTable only mutates FE-resident structures. This PR turns it into a trigger that reaches into BE and runs a full scan (Coordinator → Fragments → BE Scanners → block_cache populate). That's a new cross-layer behavior for what used to be a contained FE operation.

**3. The mechanism is heavily specialized for one specific case**
The footer is just one part of a specific data format used in a data lake query, so simply introducing session, executor, or even specific Thrift fields for the footer is too customized and lacks versatility.

**4. This task is better suited to be implemented based on CACHE SELECT**
If the footer cache preheating is indeed necessary, I think it would be better to improve the syntax and functionality of cache select itself, since cache select is designed for data preheating and is more suitable for this task.

eshishki:
Thanks @GavinMar — the layering concern is worth digging into. Let me address the four points in order.

**(4) CACHE SELECT as the right primitive — agreed.**
The PR already runs entirely through CACHE SELECT machinery (`DataCacheSelectExecutor` → `StmtExecutor` → `Coordinator` → `HDFSBackendSelector` → `CacheSelectScanner`); the footer-only mode is the only new BE-side surface. Happy to lift it from an `INVISIBLE` session var into user-visible CACHE SELECT syntax (e.g. `CACHE SELECT … FOOTER ONLY <table>`), so footer prefetch becomes a format-agnostic CACHE SELECT modifier any user or connector can drive — not an iceberg-only knob.

**(1) / (2) Metadata vs data refresh — this is exactly where the gap shows up.**
The whole problem statement of this PR is that FE *does* learn about new manifests / data files (that's what `CachingIcebergCatalog.refreshTable` is for), but the data cache does not. That's the half-finished job: the caching catalog has just told the planner "here is the new set of files" while the layer that has to serve queries against those files still has nothing cached. Asking the user to chase every metadata refresh with a manual CACHE SELECT defeats the purpose of automatic metadata refresh — the whole reason `ConnectorTableMetadataProcessor` exists is to avoid that kind of manual coordination.

I'd draw the line between *mechanism* (CACHE SELECT, lives in the data-cache layer — agreed) and *trigger* (who decides "files just changed"). The connector is the only layer with that knowledge, so the trigger has to originate there in some form. How do you see solving this without a connector-originated trigger? If there's a design where the data-cache layer learns about file-set changes on its own, I'd like to understand the shape you have in mind — that would shape where dispatch lives and what the API between connector and data-cache looks like.

**(3) Over-specialization.**
Scope is narrow because the symptom is narrow — parquet/orc footer reads on cold scan are 50–150 ms × N files of S3 RTT and dominate first-query latency on wide tables. Worth noting one part of the PR isn't iceberg-specific and isn't gated by the new flag: `HdfsScanner::create_random_access_file` now wraps `_file` in `CacheInputStream` for cache_select mode. Without that, regular CACHE SELECT was silently skipping footer population — the parquet `reader->init` footer read went straight to raw storage, only the explicit ranges fed via `_write_disk_ranges` made it into block_cache. That's a CACHE SELECT bug fix independent of this feature and benefits anyone who runs CACHE SELECT today.

**Blast radius.**
Default-off behind `enable_iceberg_metadata_refresh_footer_prefetch`, fire-and-forget on a per-catalog single-thread daemon executor, `(catalog, db, table, warehouse)` coalescing. No behavior change unless an operator opts in.

Concrete next steps I'd propose, pending your view on the trigger question above:
1. Promote `cache_select_footer_only` into user-visible CACHE SELECT syntax (`FOOTER ONLY` clause).
2. Keep the `HdfsScanner` CACHE SELECT fix — it stands on its own.

GavinMar:
1. Yes, I agree to keep the HdfsScanner CACHE SELECT fix, because it's an independent fix and I think it's better to submit to another pr, so it can be backport to other branch if necessary. 
2. For the footer prefetch feature, I think there are still some issues here that we might need to better consider and design. For example, how do we inherit the `footer-only` syntax within the `CACHE SELECT`? Should we modify the grammar or simply add a property? Also, we're currently using the name `footer` throughout the entire process, which I think might have some limitations. It's only part of a specific file format, and it doesn't exist for formats like `textfile`. Furthermore, if we want to reuse the same process to warm up other data parts in the future, this name isn't suitable. More importantly, there's the issue of trigger timing. I think it's fine for the connector to detect file changes and then trigger the footer refresh, but where should this trigger be placed? How should it be triggered? Even when we detect a file change, should we only trigger the footer, or is it necessary to trigger the entire file data as well?
I think it would be best if you could provide a more detailed design document to help us align some specific design details. Because this is, after all, a new feature, and it may involve new syntax changes, we hope to consider common use cases as much as possible before coding to avoid changes to the user interface later. cc[@wangsimo0 ](https://github.com/wangsimo0)

wangsimo0:
Thanks for the explanation. I agree the cold-footer latency problem is real, but I do not think this PR should move forward in its current form.

This is not a small implementation detail. It changes the relationship between Iceberg metadata refresh and StarRocks data-cache warmup, introduces a new warmup mode, and may affect background resource usage, cache eviction, permissions, observability, and the future design of `CACHE SELECT`.

Because of that, I think we need a design document first, before continuing with code review.

Without this design, we may merge a narrowly implemented feature that is hard to generalize, hard to document, and risky to change later. I would suggest pausing this PR, splitting out any independent bug fix if needed, and starting a design discussion for the footer/data-cache warmup feature first.

PR #73414, opened 2026-05-18
[BugFix] Include underlying cause in HdfsFsManager copy error messages

xiangguangyxg:
## Why I'm doing:

`HdfsFsManager.copyFromLocal` / `copyToLocal` wrap the underlying Hadoop exception in a `StarRocksException` but only put the source / destination paths into the new message, dropping the cause's message.

Callers that surface only `StarRocksException#getMessage` to the user can lose all useful diagnostic information. The motivating case is automated cluster snapshot upload to remote object storage:

```
ClusterSnapshotCheckpointScheduler.java
    errMsg = "upload image failed, err msg: " + e.getMessage();
```

When the underlying S3/OSS upload fails (AccessDenied, NoSuchBucket, SocketTimeoutException, SignatureDoesNotMatch, FileAlreadyExistsException, etc.), the user only sees the following in `information_schema.cluster_snapshot_jobs.error_message`:

```
upload image failed, err msg: Failed to copy local /opt/starrocks/fe/meta/image to s3://tenant-bucket/.../meta/image/automated_cluster_snapshot_xxx
```

…with no hint of the real reason. Operators currently have to grep `fe.log` for the `Exception while copy local …` stack trace to find what actually went wrong.

## What I'm doing:

- Append `e.getMessage()` to the `StarRocksException` wrapper message in both `copyFromLocal` and `copyToLocal` (for both `InterruptedIOException` and the generic `Exception` paths), so the real cause is visible at the call site without losing the existing wrapper for log greppability.
- Drive-by: fix the missing space before `to local` in `copyToLocal`'s error string (previously produced `Failed to copy /pathto local /dest`).
- Add unit tests `testCopyToLocalIncludesCauseMessage` and `testCopyFromLocalIncludesCauseMessage` covering both methods.

The underlying cause is already attached to the `StarRocksException` and still gets logged with full stack trace via `LOG.error` — this change only adds the cause's message to the user-visible message string.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
+++ b/fe/fe-core/src/main/java/com/starrocks/fs/hdfs/HdfsFsManager.java
@@ -1219,10 +1219,10 @@ public class HdfsFsManager {
         } catch (InterruptedIOException e) {
             Thread.interrupted(); // clear interrupted flag
             LOG.error("Interrupted while copy {} to local {} ", srcPath, destPath, e);
-            throw new StarRocksException("Failed to copy " + srcPath + "to local " + destPath, e);
+            throw new StarRocksException("Failed to copy " + srcPath + " to local " + destPath + ": " + e.getMessage(), e);
         } catch (Exception e) {
             LOG.error("Exception while copy {} to local {} ", srcPath, destPath, e);
-            throw new StarRocksException("Failed to copy " + srcPath + "to local " + destPath, e);
+            throw new StarRocksException("Failed to copy " + srcPath + " to local " + destPath + ": " + e.getMessage(), e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

srlch (approved):

kevincai (approved):

PR #73422, opened 2026-05-18
[Enhancement] Introduce light_weight_tablet_creation table property

starrocks-xupeng:
## Why I'm doing:

Introduce the table-level switch for the Light-Weight Tablet Creation feature so DDL on cloud-native tables can opt out of dispatching CreateReplicaTask and writing version 1 metadata to object storage at table-creation time. The CN-side fallback that constructs the initial metadata on demand from FE is already on main (see #72270 and #72364); this PR wires the FE switch and the CREATE TABLE skip behind it.

## What I'm doing:

**Property infrastructure**
* `TableProperty.lightWeightTabletCreation` with the standard `buildProperty` / getter pattern. `buildLightWeightTabletCreation()` defaults the field to `false` on deserialization regardless of the current cluster toggle, so reloads of legacy tables (whose properties map does not carry this key) stay backward compatible.
* `OlapTable.isLightWeightTabletCreation()` / `setLightWeightTabletCreation()` exposed, and the property is surfaced in `SHOW CREATE TABLE` for cloud-native tables.
* `PropertyAnalyzer.PROPERTIES_LIGHT_WEIGHT_TABLET_CREATION` registered.
* `Config.lake_enable_light_weight_tablet_creation` defaults to `false`. The cluster toggle is consulted only when creating a new cloud-native table (in `OlapTableFactory`), which then writes the resolved value back into the properties map; the toggle does not retroactively flip existing tables on FE restart.

**CREATE TABLE plumbing**
* `TabletTaskExecutor.buildPartitionsSequentially` / `buildPartitionsConcurrently` early-return for light-weight tables (no CreateReplicaTask sent).
* `LocalMetastore.createLakeTablets` skips both the alive-node check and the `isResourceAvailable` check for light-weight tables so CREATE TABLE works when no CN is live.
* `OlapTableFactory` parses the property (defaulting via the cluster toggle), sets it on the table, and for light-weight tables resolves the compute resource via `ConnectContext.getCurrentComputeResourceNoAcquire()` so the session's already-acquired resource (including the cngroup binding used by multi-cngroup deployments) is reused without going through `getCurrentComputeResource()`'s live-CN availability check. Falls back to `WarehouseManager.DEFAULT_RESOURCE` only when the session has not acquired a resource yet.
* `CatalogUtils.calBucketNumAccordingToBackends` and `calPhysicalPartitionBucketNum` now take a `lightWeight` flag. Light-weight tables size buckets against `getTotalComputeNodeNumber` so auto-bucket works even when CNs are temporarily down; regular tables keep using `getAliveComputeNodeNumber`. Callers in `OlapTable`, `MaterializedView`, and `ColocateTableIndex` pass the table's own property value.

Alter-table support (`ALTER TABLE SET ("light_weight_tablet_creation" = ...)` and the corresponding skip in `LakeTableSchemaChangeJob` / `LakeRollupJob`) lands in a follow-up PR.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [x] Miscellaneous: opt-in property + cluster-level toggle (defaults to off); the user-visible CREATE TABLE behavior only changes when the property is true.

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/TableProperty.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/TableProperty.java
@@ -252,6 +252,9 @@ public class TableProperty implements Writable, GsonPostProcessable {
     // Only meaningful when enablePersistentIndex = true.
     TPersistentIndexType persistentIndexType;
 
+    @SerializedName(value = "lightWeightTabletCreation")
+    private boolean lightWeightTabletCreation = false;
+
     private int primaryIndexCacheExpireSec = 0;
 
     /*
@@ -383,6 +386,7 @@ public class TableProperty implements Writable, GsonPostProcessable {
         this.mvTransparentRewriteMode = other.mvTransparentRewriteMode;
         this.enablePersistentIndex = other.enablePersistentIndex;
         this.persistentIndexType = other.persistentIndexType;
+        this.lightWeightTabletCreation = other.lightWeightTabletCreation;
         this.primaryIndexCacheExpireSec = other.primaryIndexCacheExpireSec;
         this.storageVolume = other.storageVolume;
         this.storageCoolDownTTL = other.storageCoolDownTTL;
@@ -879,6 +883,12 @@ public class TableProperty implements Writable, GsonPostProcessable {
         return this;
     }
 
+    public TableProperty buildLightWeightTabletCreation() {
+        lightWeightTabletCreation = Boolean.parseBoolean(
+                properties.getOrDefault(PropertyAnalyzer.PROPERTIES_LIGHT_WEIGHT_TABLET_CREATION, "false"));
+        return this;
+    }
+
     public TableProperty buildPrimaryIndexCacheExpireSec() {
         primaryIndexCacheExpireSec = Integer.parseInt(properties.getOrDefault(
                 PropertyAnalyzer.PROPERTIES_PRIMARY_INDEX_CACHE_EXPIRE_SEC, "0"));
@@ -1207,6 +1217,10 @@ public class TableProperty implements Writable, GsonPostProcessable {
 
     public boolean enablePersistentIndex() {
         return enablePersistentIndex;
+    }
+
+    public boolean lightWeightTabletCreation() {
+        return lightWeightTabletCreation;
     }
 
     public boolean isFileBundling() {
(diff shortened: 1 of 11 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/common/util/PropertyAnalyzer.java
```diff
@@ -157,6 +157,8 @@ public class PropertyAnalyzer {
 
     public static final String PROPERTIES_ENABLE_PERSISTENT_INDEX = "enable_persistent_index";
 
+    public static final String PROPERTIES_CN_FREE_TABLET_CREATION = "cn_free_tablet_creation";
```
kevincai: may choose a different name that is more proper for end-user as a production feature. `lightweight-table-creation` or `meta-only-creation` or something.

starrocks-xupeng: changed to light weight tablet creation

kevincai (dismissed):

kevincai (dismissed):

kevincai (dismissed):

banmoy (approved):

kevincai (dismissed):

kevincai (dismissed):

xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/catalog/TableProperty.java
```diff
@@ -1431,6 +1444,7 @@ public void gsonPostProcess() throws IOException {
         buildStorageCoolDownTTL();
         buildEnablePersistentIndex();
         buildPersistentIndexType();
+        buildLightWeightTabletCreation();
```
gengjun-git: // NOTE: new properties should not be built here, just add SerializedName to the field.

CelerData-Reviewer:
@codex review

gengjun-git (approved):

kevincai (approved):

PR #73423, opened 2026-05-18
[BugFix] Handle multi-statement transactions in lake publish_log_version

meegoo:
## Why I'm doing:
publish_log_version() only handled the 2-segment and combined_txn_log paths, ignoring TxnInfoPB.load_ids. During Schema Change, an explicit transaction (BEGIN; INSERT ...; COMMIT;) double-writes to shadow tablets via delta_writer with load_id set, so each statement's .log file lands at the 4-segment path (tablet_txn_loadHi_loadLo.log). BE looked at the 2-segment path, returned NotFound, and FE's publishPartition treated it as a hard failure -- the transaction stayed COMMITTED, visibleVersion was frozen, and Schema Change deadlocked at FINISHED_REWRITING (CANCEL ALTER is rejected in that state).

## What I'm doing:

This change adds the missing load_ids branch:

- publish_log_version: read every per-load_id .log via load_txn_log, merge them into one TxnLogPB (protobuf MergeFrom for repeated fields + manual sum for num_rows/data_size/num_dels accumulators), persist as the version-keyed .vlog, and stage the source .log files for deletion. An empty result falls back to the same idempotent .vlog-exists check as the non-load_id branch.

- abort_txn shares the same anti-pattern -- without walking load_ids it leaks the per-statement .log files plus every segment they reference into shared storage. Add the same branch so aborts clean up correctly.

- collect_files_in_log used the 2-segment path to erase the metacache entry even when the TxnLog was loaded from the 4-segment path, leaving a stale cache entry. Use the load_id-aware path when has_load_id() is true.

Adds regression coverage for both publish_log_version and abort_txn with load_ids set (merged .vlog contents, idempotent re-publish, NotFound on fresh version, full segment cleanup on abort).

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/transactions.cpp
+++ b/be/src/storage/lake/transactions.cpp
@@ -657,8 +657,7 @@ void collect_files_in_log(TabletManager* tablet_mgr, const TxnLog& txn_log, std:
     // Multi-statement transactions cache the .log under the 4-segment key; use the
     // same path here so the cache entry actually drops.
     if (txn_log.has_load_id()) {
-        tablet_mgr->metacache()->erase(
-                tablet_mgr->txn_log_location(tablet_id, txn_log.txn_id(), txn_log.load_id()));
+        tablet_mgr->metacache()->erase(tablet_mgr->txn_log_location(tablet_id, txn_log.txn_id(), txn_log.load_id()));
     } else {
         tablet_mgr->metacache()->erase(tablet_mgr->txn_log_location(tablet_id, txn_log.txn_id()));
     }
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

sevev (approved):

Issue #50671, opened 2024-09-04
[Aggregate Function] Support DataSketches in StarRocks


LiShuMing:
- [x] https://github.com/StarRocks/starrocks/pull/56960
## Enhancement

Since StarRocks has already integrated with [DataSketches](https://datasketches.apache.org/) and supported [HLLSketch](https://github.com/StarRocks/starrocks/issues/20831), but we can support more sketches for more user scenes.

- `ds_theta_count_distinct`(Theta Sketch): https://github.com/StarRocks/starrocks/pull/56960
- `ds_quantile_percentile_approx`(Quantile Sketches): 
- `ds_frequency_approx_top_k`(Frequency Sketches )

- ....

chenminghua8:
@LiShuMing Can you assign this task to me? I want to try it.

LiShuMing:
Of course. Thanks for your contributions. You can do this by referring `ds_hll_count_distinct.h`.

yuriolive:
@cresny started working on it here https://github.com/StarRocks/starrocks/pull/73461

PR #53701, opened 2024-12-08
[Feature] Refactor to support DataSketches Quantile、Theta、Frequency Sketches

chenminghua8:
## What I'm doing:
In order to support DataSketches Quantile, Theta, and Frequency Sketches, HllSketchAggregateFunction was refactored as follows:
1. HllSketchAggregateFunction was renamed to DataSketchesAggregateFunction.
2. The file was renamed from ds_hll_count_distinct.h to ds_agg.h.
3. Class templates were used to make DataSketchesAggregateFunction support Hll, Quantile, Theta, and Frequency Sketches.
4. Unit tests for Hll Sketches were added.
5. Code to support Quantile, Theta, and Frequency Sketches was added to fe to avoid repeated modifications to these files when supporting Quantile, Theta, and Frequency Sketches later.

Fixes #50671 

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [x] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 4.1
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1
  - [ ] 3.0

PR #73461, opened 2026-05-18
[Feature] Add ds_theta accumulate/combine/estimate and scalar set ops

yuriolive:
## Why I'm doing this

Closes part of #50671.

Today StarRocks ships a single Theta-sketch entry point, `ds_theta_count_distinct(col) → BIGINT`, which builds, unions, and estimates a sketch in one shot and discards the sketch. This PR extends the Theta family so sketches can be materialized, persisted, rolled up, set-operated, and **round-tripped with externally produced compact-theta sketches** from any conforming Apache DataSketches consumer (Spark, Druid, BigQuery, etc.) over `VARBINARY` columns.

### Aggregate functions (build / merge over rows)
- `ds_theta_accumulate(col) → VARBINARY` — build a serialized compact theta sketch from raw values. Implemented as a `SyntaxSugars` rewrite over the existing `ds_theta_count_distinct_state` combinator.
- `ds_theta_combine(VARBINARY col) → VARBINARY` — **real BE aggregate** that unions any compact-theta sketches across rows (whether built locally or loaded externally) and emits a single compact-theta sketch.

### Scalar functions (per-row, the differentiator vs HLL)
- `ds_theta_estimate(VARBINARY) → BIGINT` — true scalar that deserializes any compact theta sketch and returns the distinct-count estimate. Composes directly with `ds_theta_union/intersect/a_not_b`.
- `ds_theta_union(VARBINARY, VARBINARY) → VARBINARY`
- `ds_theta_intersect(VARBINARY, VARBINARY) → VARBINARY`
- `ds_theta_a_not_b(VARBINARY, VARBINARY) → VARBINARY`

Theta sketches preserve enough state for intersection and set difference (unlike HyperLogLog, which only supports union). The pairwise scalar set ops use `datasketches::theta_union_alloc`, `datasketches::theta_intersection_alloc`, and `datasketches::theta_a_not_b_alloc`, wired through the existing `STLCountingAllocator` for memory accounting.

### Wire format
All functions read/write the standard Apache DataSketches C++ compact theta serialization with the default hash seed. Sketches interoperate freely between `ds_theta_accumulate`, `ds_theta_combine`, the scalar set ops, the existing `ds_theta_count_distinct` infrastructure, and any external Apache DataSketches consumer.

### Design choice (vs HLL precedent)

The existing `ds_hll_*` family ships as `SyntaxSugars` over the agg-state combinator. That works when sketches are built by `ds_hll_accumulate` and never leave the FE-typed `state<varbinary>` channel — but it breaks when a sketch comes from a scalar function output, or from an external Parquet/Iceberg VARBINARY column. Local cluster smoke testing of an earlier version of this PR confirmed the failure:

```
ERROR 1064: AggState's AggFunc should have AggStateDesc:
ds_theta_count_distinct_merge(varbinary).
```

For Theta this PR ships the cleaner design: `ds_theta_estimate` is a true scalar and `ds_theta_combine` is a true aggregate, both operating on plain `VARBINARY`. The agg-state combinator (`ds_theta_count_distinct_state` / `_state_union` / `_state_merge`) is left in place for power users who want zero-copy state passing. The HLL surface is intentionally untouched; a follow-up PR can add `ds_hll_estimate_compact` etc. without breaking existing behavior.

## Files

- **BE**:
  - new `be/src/exprs/ds_theta_functions.{h,cpp}` — scalar `ds_theta_estimate`, `ds_theta_union`, `ds_theta_intersect`, `ds_theta_a_not_b`
  - new `be/src/exprs/agg/ds_theta_combine.h` — `ThetaSketchCombineAggregateFunction` (real BE aggregate over VARBINARY)
  - factory & resolver wiring in `aggregate_factory.hpp` and `aggregate_resolver_approx.cpp`
  - gtest extension: disjoint-set merge guard + compact-wire round-trip
- **FE**: `FunctionSet.java` (5 constants, 2 agg builtins — `ds_theta_count_distinct` plus `ds_theta_combine`; `ds_theta_estimate` is no longer registered as an aggregate), `SyntaxSugars.java` (kept only `ds_theta_accumulate` rewrite)
- **Codegen**: `gensrc/script/functions.py` (4 scalar IDs 91100–91103), `gensrc/script/gen_functions.py` (`DsThetaFunctions` module dispatch)
- **Tests**: BE gtest cases for disjoint-set merge and compact-wire round-trip; SQL functional test for accumulate/combine/estimate, all three scalar set ops, raw-VARBINARY composition through combine, and NULL propagation
- **Docs**: en/zh/ja for all 6 functions

## Local verification

- BE unit tests (`./run-be-ut.sh --test DataSketchsThetaTest`): **3/3 pass** including `TestMergeDisjointSets` and `TestCompactWireRoundTrip`.
- FE unit tests (`./run-fe-ut.sh --test com.starrocks.sql.analyzer.AnalyzeFunctionTest`): **24/24 pass**.
- Full BE+FE build packaged into an `allin1-ubuntu` image and smoke-tested in a local cluster:
  - `SELECT ds_theta_estimate(ds_theta_accumulate(id)) FROM ...` returns the exact count for small inputs and matches `ds_theta_count_distinct` within theta error for large inputs.
  - `SELECT ds_theta_estimate(ds_theta_union(a, b))` composes correctly — the bug that motivated this redesign.
  - `SELECT ds_theta_estimate(ds_theta_combine(set_op_output))` composes correctly — confirms the raw-VARBINARY round-trip.

## Test plan

- [ ] `./run-be-ut.sh --test DataSketchsThetaTest` — gtest for DataSketchesTheta passes including the new disjoint-set merge and compact-wire round-trip.
- [ ] `./run-fe-ut.sh` — FE registrations parse cleanly.
- [ ] `test/sql/test_agg_function/T/test_ds_theta_sketch_ops.sql` — SQL functional test covers:
  - `accumulate` produces non-empty VARBINARY per group
  - scalar `estimate(accumulate)` matches `ds_theta_count_distinct` within theta error
  - aggregate `combine` rolls daily sketches up
  - `union` / `intersect` / `a_not_b` estimates fall in expected ranges and compose directly with scalar `estimate`
  - `combine` over a stack mixing accumulate output and scalar set op output (the raw-VARBINARY round-trip)
  - NULL propagation in all set ops and in `estimate`
- [ ] Existing `test_ds_theta_count_distinct.sql` still passes (no regression)

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

## Backwards compatibility

`ds_theta_count_distinct` is unchanged. `ds_theta_accumulate` is unchanged. `ds_theta_combine` and `ds_theta_estimate` are new in this PR — no prior callers — so promoting them from sugar to first-class scalar/aggregate is not a breaking change. The existing agg-state combinator surface (`ds_theta_count_distinct_state` etc.) is untouched.

## Out of scope — planned follow-up PRs

The goal here is to land the core surface area first (build / combine / estimate / union / intersect / a_not_b) with a design that round-trips with external Apache DataSketches producers. The following are deliberately deferred to keep this PR reviewable, and are planned as separate follow-up PRs:

### 1. `lg_k` parameter on `ds_theta_accumulate` (and on `ds_theta_count_distinct`)
- Add `ds_theta_accumulate(col, lg_k)` and `ds_theta_count_distinct(col, lg_k)` overloads.
- Apache DataSketches exposes `lg_k` (default 12) to trade off precision and memory; right now we always use the library default. The Apache spec already supports unioning/intersecting sketches built with different `lg_k` (result is downsampled to the minimum k), so adding this is purely additive.

### 2. Additional scalar functions to match the wider Apache DataSketches Theta surface
- `ds_theta_jaccard(VARBINARY, VARBINARY) → DOUBLE`
- `ds_theta_estimate_and_bounds(VARBINARY, INT) → STRUCT<estimate, lower, upper>`
- `ds_theta_to_string(VARBINARY) → VARCHAR`, `ds_theta_get_num_retained(VARBINARY) → INT`, `ds_theta_get_theta(VARBINARY) → DOUBLE`, `ds_theta_is_empty(VARBINARY) → BOOLEAN`

### 3. Variadic / multi-input set ops
- `ds_theta_union_arr(ARRAY<VARBINARY>) → VARBINARY` and analogous `intersect_arr` / `a_not_b` reductions.

### 4. Compressed serialization variants
- Apache DataSketches recently added a compressed compact serialization for theta sketches. Once we bump the bundled DataSketches version, add an opt-in `ds_theta_accumulate_compressed` (or a `serialize_format` argument).

### 5. Tombstone / removal semantics
- Document explicitly in the function docs alongside `a_not_b` to forestall confusion with bitmap-style mutation.

### 6. Cross-engine interop docs
- Recipe page for round-tripping serialized theta sketches between StarRocks and other Apache DataSketches consumers (Spark, Hive, Druid, Pinot) via `VARBINARY` columns over Parquet/Iceberg.

### 7. Iceberg/Parquet VARBINARY load path verification
- Integration tests confirming Broker Load / Iceberg / Spark connector preserve `VARBINARY` byte-for-byte (no UTF-8 reinterpretation).

### 8. Apply the same scalar/aggregate split to the existing `ds_hll_*` family
- `ds_hll_estimate_compact(VARBINARY) → BIGINT` and `ds_hll_combine_compact(VARBINARY col) → VARBINARY` so HLL sketches loaded from external Parquet/Iceberg can be queried directly. Kept out of this PR to avoid touching the HLL contract that has already shipped.


hgupta-roblox:
Nice work! The plain-VARBINARY scalar/aggregate split fixes exactly the composition issue we saw with agg-state sugar (ds_theta_count_distinct_merge(varbinary)).

Currently I am benchmarking against druid and using Java sketch UDAFs for that (hex VARCHAR workaround). Planning to migrate to ds_theta_* once this lands. Or I might end up building a custom image if this PR takes time. 

Two gaps for us: conditional intersect (retention/stickiness) and optionally N-way intersect agg vs pairwise scalar only.

Happy to help with SQL tests from our workloads or a follow-up PR for conditional intersect — should that be a separate issue/PR on #50671?

yuriolive:
> Nice work! The plain-VARBINARY scalar/aggregate split fixes exactly the composition issue we saw with agg-state sugar (ds_theta_count_distinct_merge(varbinary)).
> 
> Currently I am benchmarking against druid and using Java sketch UDAFs for that (hex VARCHAR workaround). Planning to migrate to ds_theta_* once this lands. Or I might end up building a custom image if this PR takes time.
> 
> Two gaps for us: conditional intersect (retention/stickiness) and optionally N-way intersect agg vs pairwise scalar only.
> 
> Happy to help with SQL tests from our workloads or a follow-up PR for conditional intersect — should that be a separate issue/PR on #50671?

Thanks! The Druid benchmark context is useful, and real external-producer round-trip tests would be great.

Conditional intersect is composable today with the aggregate FILTER clause, no new function required for the standard retention/stickiness pattern:

```
WITH sketches AS (
  SELECT
    ds_theta_accumulate(user_id) FILTER (WHERE event = 'viewed')    AS sketch_a,
    ds_theta_accumulate(user_id) FILTER (WHERE event = 'purchased') AS sketch_b
  FROM events
  WHERE ...
)
SELECT ds_theta_estimate(ds_theta_intersect(sketch_a, sketch_b))
FROM sketches;
```

If your use case needs something that doesn't reduce to this pattern (e.g. a temporal window join before intersection), would probably need a follow-up PR, would be happy to review.

N-way intersect agg (ds_theta_intersect_arr(ARRAY<VARBINARY>) → VARBINARY) would be a good add, we can add comment there or create a separate issue, I think this is still not correctly tracked, was also mentioned here https://github.com/StarRocks/starrocks/pull/53701.

SQL tests from your workloads would be a great addition the more coverage we have for the VARBINARY round-trip path (especially sketches produced externally by Java/Druid), the better. PRs welcome.

PR #70197, opened 2026-03-12
[Enhancement] enable information_schema show comments for external catalogs

dontknow9179:
## Why I'm doing

In current versions, `information_schema.tables` and related system views only expose very limited local information for tables in external catalogs (such as Hive, Iceberg, JDBC).  
This leads to several issues:
- Users cannot see table comments for external tables via the unified `information_schema` interface, which is inconsistent with internal tables.
- To inspect comments, users have to query each external system or use different commands, which increases operational and troubleshooting overhead.
We want to provide a configurable way to let `information_schema` show comments for external catalog tables, while keeping the default behavior lightweight and performance‑friendly.

## What I'm doing

- Add a new FE configuration item `enable_external_catalog_information_schema_tables_access_full_metadata` (default `false`) to control whether `information_schema.tables` may load full metadata for tables in external catalogs.
- When this config is set to `true`:
  - `information_schema.tables` is allowed to contact external metadata services (Hive / Iceberg / JDBC, etc.) when resolving external tables, so fields such as `TABLE_COMMENT` can be populated from real external metadata.
- When this config is `false` (default):
  - StarRocks still relies only on lightweight local basic information for external catalogs, does not contact external metadata services, and the behavior/performance remain unchanged (external table comments in `information_schema.tables` may be empty).
- Update the English and Chinese `FE_configuration` docs to document this new configuration, its default value, and when to enable it.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [x] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
[...]

PR #73488, opened 2026-05-19
[BugFix] Defer JDBC REMARKS fetch out of getTable() hot path

dontknow9179:
## Why I'm doing:

#70197 added a `schemaResolver.getTableComment()` round-trip inside `JDBCMetadata.getTable()`'s cache-loader lambda. Every cache miss now costs 2 round-trips (`getColumns` + `getTables`-for-REMARKS) instead of 1. When the loader runs under `PlannerMetaLocker` (mixed internal + external queries, with pre-parse disabled or bypassed) the extra IO extends the intensive READ lock hold on the query's internal tables and blocks their concurrent DDL.

## What I'm doing:

Move comment fetch off the hot path:

- `ConnectorMetadata` adds a `default getTableComment(ctx, db, tbl)` that falls back to the comment already carried by `getTable()` — HMS / Iceberg / Hudi pay nothing because their `getTable` populates comment inline.
- `JDBCMetadata` overrides `getTableComment` to issue the dedicated REMARKS query; the `getTable` lambda no longer pre-populates comment.
- `information_schema.tables` (the only caller that needs comment for external catalogs) explicitly invokes the SPI and writes back to the cached `Table` so subsequent calls within the JDBC meta cache TTL skip the extra round-trip.

Also drop the unused `fetchExternalMetadata=true` argument in `FrontendServiceImpl.getTableNames`: the caller only uses `tbl != null` + `getType()` for auth filtering; full metadata fetch was never read.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/JDBCTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/JDBCTable.java
@@ -61,6 +61,12 @@ public class JDBCTable extends Table {
     // Used for Oracle datetime predicate pushdown to determine TO_DATE/TO_TIMESTAMP wrapping.
     private transient Map<String, Integer> originalJdbcColumnTypes;
 
+    // Transient: marker for {@link com.starrocks.connector.jdbc.JDBCMetadata#getTableComment}
+    // dedup. Once REMARKS has been fetched for this cached instance, further calls return the
+    // already-stored comment without another remote round-trip. Reset when the cache entry is
+    // evicted or refreshTable creates a new JDBCTable.
+    private transient boolean commentFetched;
+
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

GavinMar (approved):

stephen-shelby (approved):

gengjun-git (approved):

PR #73489, opened 2026-05-19
[Enhancement] Consider outer side join key null fractions (unmatched keys) in addition to selectivity

OliLay:
## Why I'm doing:

For outer joins, we did not look at the null fractions of the outer join key column when calculating the resulting null fractions of the join. We only determined the resulting null fractions by looking into the estimated inner row count. If the estimated inner row count is inaccurate (for example due to unknown stats of the inner join key column, we assume `innerRowCount = outerRowCount`) we underestimate the resulting null fractions. We should also look at the null fractions of the outer join key column and take these into consideration to avoid an underestimation.

## What I'm doing:

Instead of only relying on the estimated row count of the join (`nullsRows = outerSideRowCount - innerRowCount`), we also take the null fractions of the outer join columns into consideration (`nullsRows = maxOuterKeyNullFraction * outerSideRowCount`) and take the `max` of both to avoid an underestimation.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
@@ -1297,13 +1297,13 @@ public class StatisticsCalculator extends OperatorVisitor<Void, ExpressionContex
                 outputRowCount = max(innerRowCount, leftRowCount);
                 joinStatsBuilder.setOutputRowCount(outputRowCount);
                 computeNullFractionForOuterJoin(leftRowCount, innerRowCount, outputRowCount,
-                        leftStatistics, rightStatistics, eqOnPredicates, joinStatsBuilder);
+                        leftStatistics, rightStatistics, eqOnPredicates, hasUnknownColumnStatistics, joinStatsBuilder);
                 break;
             case ASOF_LEFT_OUTER_JOIN:
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
                 joinStatsBuilder.setOutputRowCount(leftRowCount);
                 computeNullFractionForOuterJoin(leftRowCount, innerRowCount, leftRowCount, leftStatistics, rightStatistics,
-                        eqOnPredicates, joinStatsBuilder);
+                        eqOnPredicates, hasUnknownColumnStatistics, joinStatsBuilder);
                 break;
             case LEFT_SEMI_JOIN:
                 joinStatsBuilder = Statistics.buildFrom(StatisticsEstimateUtils.adjustStatisticsByRowCount(
@@ -1325,7 +1325,7 @@ public class StatisticsCalculator extends OperatorVisitor<Void, ExpressionContex
                 outputRowCount = max(innerRowCount, rightRowCount);
                 joinStatsBuilder.setOutputRowCount(outputRowCount);
                 computeNullFractionForOuterJoin(rightRowCount, innerRowCount, outputRowCount,
-                        rightStatistics, leftStatistics, eqOnPredicates, joinStatsBuilder);
+                        rightStatistics, leftStatistics, eqOnPredicates, hasUnknownColumnStatistics, joinStatsBuilder);
                 break;
             case RIGHT_ANTI_JOIN:
                 joinStatsBuilder = Statistics.buildFrom(innerJoinStats);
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
```diff
@@ -1420,15 +1427,24 @@ private Statistics buildStatisticsForUKFKJoin(JoinOperator joinType, UKFKConstra
...
     // This method first restores the original null fractions for the outer side's eq-join columns
     // (the only columns whose null fractions are zeroed by the inner join estimation), then computes
     // the new null fractions for the inner (nullable) side's columns to account for additional null rows.
-    private void computeNullFractionForOuterJoin(double outerTableRowCount, double innerJoinRowCount,
-                                                 Statistics outerSideStatistics, Statistics innerSideStatistics,
-                                                 List<BinaryPredicateOperator> eqOnPredicates,
+    private void computeNullFractionForOuterJoin(double outerSideRowCount, double innerRowCount,
+                                                 double outputRowCount, Statistics outerSideStatistics,
+                                                 Statistics innerSideStatistics, List<BinaryPredicateOperator> eqOnPredicates,
                                                  Statistics.Builder builder) {
```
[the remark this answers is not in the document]
OliLay: [bd9426b](https://github.com/StarRocks/starrocks/pull/73489/commits/bd9426bbc81cd97e8cfb298850c77754df8333aa)

trueeyu (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
```diff
@@ -1443,30 +1460,53 @@ private void computeNullFractionForOuterJoin(double outerTableRowCount, double i
...
-                double columnNullCount = columnStatistic.getNullsFraction() * innerJoinRowCount;
-                double newNullFraction = (columnNullCount + nullRowCount) / outerTableRowCount;
+        // Compute new null fractions for the inner (nullable) side's columns.
+        // Two sources of NULLs for inner-side columns after an outer join:
+        // 1. Unmatched rows: outerSideRowCount - innerRowCount (row-count based)
+        // 2. Null-key rows: outer-side rows whose join key is NULL can never match,
+        //    so inner-side columns are NULL for those rows regardless of row-count estimates.
+        // We use the maximum of the two as a lower bound on the null row count.
+        // For EQ_FOR_NULL (<=>) predicates, NULL keys can match, so we must exclude those columns from the null-key adjustment.
+        final var eqForNullJoinKeyColumns = getEqForNullJoinKeyColumns(eqOnPredicates);
```
stephen-shelby:
```markdown
I think the general direction of this PR makes sense, but I’m a bit concerned that using the preserved-side join key null fraction as a hard lower bound may over-adjust some cases where the join row-count estimate is already reasonably accurate.

For example, consider a LEFT OUTER JOIN:

```sql
SELECT *
FROM l
LEFT JOIN r ON l.k = r.k
WHERE r.val IS NULL;
```

Assume:

```text
l row count = 1,000,000
r row count = 950,000

actual l.k null fraction = 5%
estimated l.k null fraction = 15%

r.val null fraction = 1%
inner join row count estimate = 950,000
```

This can happen if the join row count is estimated accurately by another signal, for example histogram-based join estimation, while the column-level null fraction is stale or somewhat overestimated.

Under the actual data distribution:

```text
unmatched rows from l.k IS NULL = 50,000
matched rows = 950,000
matched r.val nulls = 950,000 * 1% = 9,500
actual r.val null fraction ~= (50,000 + 9,500) / 1,000,000 = 5.95%
```

The previous row-count based logic would also estimate roughly:

```text
nullRowCount = 1,000,000 - 950,000 = 50,000
matched nulls = 950,000 * 1% = 9,500
estimated null fraction ~= 5.95%
```

With the new hard lower bound:

```text
nullRowsFromNullKey = 15% * 1,000,000 = 150,000
effectiveNullRowCount = max(50,000, 150,000) = 150,000
matched nulls = 850,000 * 1% = 8,500
estimated null fraction ~= (150,000 + 8,500) / 1,000,000 = 15.85%
```

So `r.val IS NULL` selectivity may be raised from about 6% to about 16%, even though the original row-count based estimate was already close to the actual result. This could affect `IS NULL` / `IS NOT NULL` selectivity and downstream plan choices.

I think we should try to keep the conservative parts of this change first. Cardinality-estimation changes can definitely fix some cases, but they can also make other distributions worse. The FULL OUTER JOIN row-count lower bound and using `outputRowCount` as the null-fraction denominator look relatively conservative and semantically safer to me. My main concern is specifically around treating the preserved-side key null fraction as a hard lower bound for inner-side NULL rows, which may be a more aggressive estimate change.

stephen-shelby: Since statistics can be unknown, stale, or inaccurate due to sampling, theoretical estimation errors will always exist for some cases. My point is that we should keep this strategy more conservative relative to the current behavior. Was this strategy change motivated by many cases observed in production? If so, could you share more details about those cases?

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/StatisticsCalculator.java
OliLay: Discussed offline with Stephen: We make this more conservative by only taking the new path as fallback if we know that the selectivity path is unaccurate due to unknown stats. https://github.com/StarRocks/starrocks/pull/73489/commits/61b29faec9e1a5ab4100002dee006ba8edbc5de9

stephen-shelby (approved):

trueeyu (approved):

HangyuanLiu (approved):

martinr0x:
@mergifyio rebase branch-3.5-cc

martinr0x:

PR #73500, opened 2026-05-19
[Enhancement] Propagate stats for `DictMappingOperator`

OliLay:
## Why I'm doing:
Currently, `DictMappingOperator` does not propagate any stats. 

## What I'm doing:
Implementing stats propagation for the `DictMappingOperator`.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
@@ -105,26 +105,48 @@ public class ExpressionStatisticCalculator {
 
         @Override
         public ColumnStatistic visitDictMappingOperator(DictMappingOperator operator, Void context) {
-            final var stringProvideOperator = operator.getStringProvideOperator();
-            final ScalarOperator sourceOperator;
-            if (stringProvideOperator != null) {
-                sourceOperator = stringProvideOperator;
-            } else {
-                sourceOperator = operator.getOriginScalaOperator();
-            }
-
-            if (!hasAllColumnStatisticsFor(sourceOperator)) {
-                // Return unknown stats in case we don't have the non-dict stats anymore.
+            if (inputStatistics == null) {
                 return ColumnStatistic.unknown();
             }
 
-            final var sourceStatistic = sourceOperator.accept(this, context);
-            if (sourceStatistic.isUnknown() || operator.getType().matchesType(sourceOperator.getType())) {
-                return sourceStatistic;
+            final var stringProvideOperator = operator.getStringProvideOperator();
+            final var originOperator = operator.getOriginScalaOperator();
+            final ColumnStatistic statistic;
+            if (stringProvideOperator != null) {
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
```diff
@@ -102,6 +103,48 @@ public ColumnStatistic visitLambdaFunctionOperator(LambdaFunctionOperator operat
             return operator.getChild(0).accept(this, context);
         }
 
+        @Override
+        public ColumnStatistic visitDictMappingOperator(DictMappingOperator operator, Void context) {
+            final var stringProvideOperator = operator.getStringProvideOperator();
+            final ScalarOperator sourceOperator;
+            if (stringProvideOperator != null) {
+                sourceOperator = stringProvideOperator;
```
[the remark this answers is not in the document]
OliLay: [3757557](https://github.com/StarRocks/starrocks/pull/73500/commits/3757557d0f4bfd3a25a7f751f4a617ff84522092)

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/ExpressionStatisticCalculator.java
```diff
@@ -102,6 +103,76 @@ public ColumnStatistic visitLambdaFunctionOperator(LambdaFunctionOperator operat
             return operator.getChild(0).accept(this, context);
         }
 
+        @Override
+        public ColumnStatistic visitDictMappingOperator(DictMappingOperator operator, Void context) {
```
stephen-shelby, line 107: Thanks for the enhancement! I'm curious what specific issue you ran into that prompted this.

DictMappingOperator is introduced late in the pipeline (inside physicalRuleRewrite, after CBO has already finished), so the number of places that still call ExpressionStatisticCalculator on expressions containing it is quite small — from what I can see, the main one is PredicateReorderRule (via DefaultPredicateSelectivityEstimator), which runs after AddDecodeNodeForDictStringRule.
Did you encounter a concrete plan error caused by missing dict stats? If so, could you share the case?

OliLay: I think you are right, I just tried to reproduce the issue again with the SQL that I initially ran into it. Turns out that there were two issues (so another issue besides the "dict issue"), and fixing the other problem actually propagates the stats correctly. So the reason I started to work on this PR isn't actually fixed by this PR. 
Do you think this improvement still makes sense even though it may only help in a limited amount of cases? (e.g. `PredicateReorderRule` as you mentioned)

PR #73501, opened 2026-05-19
[BugFix] Atomically publish proc profile tar.gz to avoid partial-read race

Hyper-FF:
## Why I'm doing:

ProcProfileCollector.compressFile wrote the gzip+tar stream directly into the final cpu-profile-*.html.tar.gz path. While compression was in progress, the partially-written file was visible to both the listing endpoint (/proc_profile) and the download endpoint (/proc_profile/file), which would then fail to parse the incomplete gzip/tar and return an empty 500 response. This caused intermittent failures in test_proc_profile_multi_node where the test scrapes the newest file from the listing and then downloads it.

Write to a sibling .tar.gz.tmp file first and atomically rename it to the final name on success, so consumers never observe a partial file.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/memory/ProcProfileCollector.java
+++ b/fe/fe-core/src/main/java/com/starrocks/memory/ProcProfileCollector.java
@@ -28,6 +28,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.nio.file.AtomicMoveNotSupportedException;
+import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

stdpain (approved):

PR #73508, opened 2026-05-19
[Enhancement] Add SR-owned VectorIndexCache for vector index

sevev:
## Why I'm doing:

The vector-index cache is owned by the third-party `tenann` library today. That has three concrete problems on the SR side:

1. **No memory accounting.** `tenann::IndexCache` is heap-resident but not attached to any `MemTracker`. The HNSW / IVF-PQ working set stays invisible to `process_mem_tracker`, and `SHOW PROC '/be/mem'` underreports BE memory.

2. **Observability sourced from a tenann singleton.** The existing `vector_index_cache_*` Prometheus gauges read from `tenann::IndexCache::GetGlobalInstance()`; once tenann no longer owns the cache, that source goes away.

3. **No SR-side control.** Capacity is a process-static config, not mutable at runtime; eviction policy is fixed; and concurrent cold-miss loads on the same key duplicate work because tenann's cache has no single-flight contract.

Tenann PRs [#6](https://github.com/StarRocks/tenann/pull/6) and [#8](https://github.com/StarRocks/tenann/pull/8) introduced the `IndexCache` abstract base and the `AttachIndexRef` searcher API so callers can plug their own implementation. This PR is the SR side that consumes those interfaces.

## What I'm doing:

Move vector-index cache ownership from tenann to StarRocks. The new cache is an SR-owned implementation of tenann's `IndexCache` interface, built on the same `DynamicCache` primitive that `PrimaryIndex` and `update_manager` already use — so it inherits the LRU + capacity + mem-tracker plumbing the rest of the codebase relies on, and it does not introduce a new caching mechanism.

**Lifecycle.** BE startup constructs the cache and installs it as tenann's global `IndexCache` before any reader/searcher is built. Shutdown tears it down after all queries drain, before the mem-tracker hierarchy goes away. From tenann's perspective the cache is still "the global cache"; from StarRocks's perspective it is a normal owned subsystem.

**Memory accounting.** The cache attaches a dedicated `vector_index` mem tracker as a direct child of the process tracker — sibling of `page_cache` / `datacache` / `jit_cache`. Insert and evict both flow through this tracker via `DynamicCache`'s consume/release, so the HNSW / IVF-PQ working set shows up in `SHOW PROC '/be/mem'` and in the `vector_index_mem_bytes` Prometheus gauge alongside the other major caches.

**Observability.** The 9 existing `vector_index_cache_*` Prometheus gauges (capacity / usage / usage_ratio / lookup / hit / hit_ratio / dynamic_*) keep working, sourced from the SR cache instead of the removed tenann singleton. Hit/lookup counters are bumped only inside the load path (`GetOrCreate`); the cache's `Lookup` entry point that reader factories use as an opportunistic warm-path probe is counter-silent, so absolute counts stay comparable to pre-PR baselines.

**Single-flight loads.** Cold-miss loads on the same key are deduplicated by a per-entry mutex held across the loader callback — concurrent callers either run the loader once and share its `IndexRef`, or all wait and observe the populated entry. Loader failure (exception or null return) drops the entry so retries start fresh, matching the pattern `update_manager` already uses.

**Reader integration.** `TenANNReader::init_searcher` resolves its `IndexRef` through `GetOrCreate` and pins the handle for the reader's lifetime, so an index in active use cannot be evicted; the searcher attaches directly to that ref and skips tenann's internal cache lookup. The factory probes the cache before doing OSS metadata round-trips (`path_exist` + `open` + `get_size`), so warm-cache reads on remote storage skip the network entirely. Loader errors — including `tenann::Error`, which inherits privately from `std::exception` and would otherwise escape — are caught and turned into the appropriate `Status`, so `NotFound` keeps reaching the brute-force fallback.

**Handle lifetime.** Cache handles use the same raw-cache-pointer-in-deleter contract as `PrimaryIndex` and `StoragePageCache`: every handle must be released before BE shutdown resets the cache, and the existing fragment-drain step at shutdown is the boundary that guarantees that.

**Runtime control.** A new `vector_index_cache_limit` config replaces the old static int64 capacity. It is `mString`-typed, defaults to `"20%"` of the BE memory limit, and accepts absolute bytes, unit suffixes (`"4G"`, `"512M"`), or a percentage — same convention as `storage_page_cache_limit`. `/api/update_config` re-resolves the value and resizes the cache live without restarting BE. EN + ZH `BE_parameters/query_loading.md` docs updated.

**Commit layout.** Five focused commits, in order: bump the tenann pin to the first cache-aware release (`v0.5.1-rc1`); implement the SR cache itself with companion file-reader bridge, error mapper, and a UT suite covering the lookup / single-flight / eviction / mem-tracker / capacity-shrink / cross-thread-evict matrix; wire it into `ExecEnv` with the mem tracker and the full Prometheus gauge set; route `TenANNReader` and the reader factory through it; add the runtime-configurable capacity.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/service/service_be/config_update_hooks.cpp
+++ b/be/src/service/service_be/config_update_hooks.cpp
@@ -21,6 +21,7 @@
 
 #include "agent/agent_common.h"
 #include "agent/agent_server.h"
+#include "base/string/parse_util.h"
 #include "cache/datacache.h"
 #include "cache/datacache_utils.h"
 #include "cache/mem_cache/page_cache.h"
@@ -37,6 +38,7 @@
 #include "common/config_staros_worker_fwd.h"
 #include "common/config_storage_fwd.h"
 #include "common/config_update_registry.h"
+#include "common/config_vector_index_fwd.h"
 #include "common/logging.h"
 #include "common/status.h"
 #include "common/system/cpu_info.h"
@@ -49,6 +51,9 @@
 #include "runtime/exec_env.h"
 #include "runtime/load_channel_mgr.h"
 #include "storage/compaction_manager.h"
+#ifdef WITH_TENANN
+#include "storage/index/vector/vector_index_cache.h"
+#endif
 #include "storage/lake/compaction_scheduler.h"
 #include "storage/lake/lake_persistent_index_parallel_compact_mgr.h"
 #include "storage/lake/tablet_manager.h"
@@ -92,6 +97,20 @@ void register_config_update_hooks(ExecEnv* exec_env, const GlobalEnv& global_env
         cache->set_capacity(cache_limit);
         return Status::OK();
     });
+#ifdef WITH_TENANN
+    registry->register_callback("vector_index_cache_limit", [=]() -> Status {
+        if (exec_env == nullptr || exec_env->vector_index_cache() == nullptr) {
+            return Status::InternalError("Vector index cache is not initialized");
+        }
+        const int64_t proc_mem = GlobalEnv::GetInstance()->process_mem_limit();
+        ASSIGN_OR_RETURN(int64_t limit, ParseUtil::parse_mem_spec(config::vector_index_cache_limit, proc_mem));
+        if (limit < 0) limit = 0;
+        exec_env->vector_index_cache()->SetCapacity(static_cast<size_t>(limit));
+        LOG(INFO) << "vector_index_cache_limit updated: " << config::vector_index_cache_limit << " => " << limit
+                  << " bytes";
+        return Status::OK();
+    });
+#endif
 #endif
 #ifndef __APPLE__
     registry->register_callback("disable_storage_page_cache", [=]() -> Status {
--- a/be/src/service/service_be/config_update_hooks.cpp
+++ b/be/src/service/service_be/config_update_hooks.cpp
@@ -95,15 +95,19 @@ void register_config_update_hooks(ExecEnv* exec_env, const GlobalEnv& global_env
         cache->set_capacity(cache_limit);
         return Status::OK();
     });
-    registry->register_callback("vector_index_cache_limit", [=]() -> Status {
+    registry->register_callback("vector_query_cache_capacity", [=]() -> Status {
         if (exec_env == nullptr || exec_env->vector_index_cache() == nullptr) {
             return Status::InternalError("Vector index cache is not initialized");
         }
         const int64_t proc_mem = GlobalEnv::GetInstance()->process_mem_limit();
-        ASSIGN_OR_RETURN(int64_t limit, ParseUtil::parse_mem_spec(config::vector_index_cache_limit, proc_mem));
+        ASSIGN_OR_RETURN(int64_t limit, ParseUtil::parse_mem_spec(config::vector_query_cache_capacity, proc_mem));
         if (limit < 0) limit = 0;
-        exec_env->vector_index_cache()->SetCapacity(static_cast<size_t>(limit));
-        LOG(INFO) << "vector_index_cache_limit updated: " << config::vector_index_cache_limit << " => " << limit
+        auto* cache = exec_env->vector_index_cache();
+        if (static_cast<size_t>(limit) == cache->capacity()) {
+            return Status::OK();
+        }
+        cache->SetCapacity(static_cast<size_t>(limit));
+        LOG(INFO) << "vector_query_cache_capacity updated: " << config::vector_query_cache_capacity << " => " << limit
                   << " bytes";
         return Status::OK();
     });
--- a/be/src/service/service_be/config_update_hooks.cpp
+++ b/be/src/service/service_be/config_update_hooks.cpp
@@ -95,6 +95,7 @@ void register_config_update_hooks(ExecEnv* exec_env, const GlobalEnv& global_env
         cache->set_capacity(cache_limit);
         return Status::OK();
     });
+#ifdef WITH_TENANN
     registry->register_callback("vector_query_cache_capacity", [=]() -> Status {
         if (exec_env == nullptr || exec_env->vector_index_cache() == nullptr) {
             return Status::InternalError("Vector index cache is not initialized");
@@ -111,6 +112,7 @@ void register_config_update_hooks(ExecEnv* exec_env, const GlobalEnv& global_env
                   << " bytes";
         return Status::OK();
     });
+#endif // WITH_TENANN
 #endif
 #ifndef __APPLE__
     registry->register_callback("disable_storage_page_cache", [=]() -> Status {
--- a/be/src/service/service_be/config_update_hooks.cpp
+++ b/be/src/service/service_be/config_update_hooks.cpp
@@ -51,9 +51,7 @@
 #include "runtime/exec_env.h"
 #include "runtime/load_channel_mgr.h"
 #include "storage/compaction_manager.h"
-#ifdef WITH_TENANN
 #include "storage/index/vector/vector_index_cache.h"
-#endif
 #include "storage/lake/compaction_scheduler.h"
 #include "storage/lake/lake_persistent_index_parallel_compact_mgr.h"
 #include "storage/lake/tablet_manager.h"
@@ -97,7 +95,6 @@ void register_config_update_hooks(ExecEnv* exec_env, const GlobalEnv& global_env
         cache->set_capacity(cache_limit);
         return Status::OK();
     });
-#ifdef WITH_TENANN
     registry->register_callback("vector_index_cache_limit", [=]() -> Status {
         if (exec_env == nullptr || exec_env->vector_index_cache() == nullptr) {
             return Status::InternalError("Vector index cache is not initialized");
@@ -110,7 +107,6 @@ void register_config_update_hooks(ExecEnv* exec_env, const GlobalEnv& global_env
                   << " bytes";
         return Status::OK();
     });
-#endif
 #endif
 #ifndef __APPLE__
     registry->register_callback("disable_storage_page_cache", [=]() -> Status {
--- a/be/src/storage/index/vector/tenann_index_reader.cpp
+++ b/be/src/storage/index/vector/tenann_index_reader.cpp
@@ -35,14 +35,22 @@
 #ifdef WITH_TENANN
 #include "tenann_index_reader.h"
 
+#include <stdexcept>
+
 #include "common/config_vector_index_fwd.h"
 #include "common/status.h"
 #include "common/statusor.h"
 #include "fs/fs.h"
+#include "runtime/current_thread.h"
+#include "runtime/exec_env.h"
+#include "runtime/mem_tracker.h"
 #include "storage/index/vector/tenann/tenann_index_utils.h"
 #include "storage/index/vector/vector_index_file_reader.h"
 #include "tenann/common/error.h"
 #include "tenann/common/seq_view.h"
+#include "tenann/factory/index_factory.h"
+#include "tenann/index/index_cache.h"
+#include "tenann/index/index_reader.h"
 #include "tenann/searcher/id_filter.h"
 
 namespace starrocks {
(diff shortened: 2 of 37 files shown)


CelerData-Reviewer:
@codex review

alvin-phoenix-ai:
@trueeyu 
please review this PR to see if it meet the memory management requirement.

CelerData-Reviewer:
@codex review

be/src/storage/index/vector/vector_index_cache.cpp
```diff
@@ -0,0 +1,104 @@
...
+    if (!ran) _hit_count.fetch_add(1, std::memory_order_relaxed);
+    *handle = _wrap(entry, std::move(ref));
+    return true;
+}
+
+// shared_ptr<void> deleter decrements the entry ref when the handle is destroyed.
+tenann::IndexCacheHandle VectorIndexCache::_wrap(Entry* entry, tenann::IndexRef ref) {
+    Cache* cache = &_cache;
+    return tenann::IndexCacheHandle(
+            std::move(ref), std::shared_ptr<void>(entry, [cache](void* p) { cache->release(static_cast<Entry*>(p)); }));
```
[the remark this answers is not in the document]
sevev: Only happened during BE exit and it is acceptable.

be/src/storage/index/vector/tenann_index_reader.cpp
```diff
@@ -65,43 +73,56 @@ void apply_index_reader_cache_options(tenann::IndexMeta* meta_copy) {
...
+        }
+        return reader->ReadIndexFile(index_path);
+    };
 
-        DCHECK(_searcher->is_index_loaded());
-    } catch (tenann::Error& e) {
-        return Status::InternalError(e.what());
+    if (!cache->GetOrCreate(tenann::CacheKey(index_path), loader, &_cache_handle)) {
+        return Status::InternalError("failed to load vector index: " + index_path);
     }
```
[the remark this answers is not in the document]
sevev: Good catch — fixed in 819eae1a0d2. Two issues addressed:

1. **The flagged one (Status loss):** Loader now captures the real `Status` (incl. `NotFound`) into a closure-bound variable and returns `nullptr`; `init_searcher` then returns the captured Status when `GetOrCreate` reports failure. This preserves `NotFound` for `VectorIndexReaderFactory::create_from_file`'s brute-force fallback.

2. **Plus a related bug:** `tenann::Error` inherits **privately** from `std::exception` (see `tenann/common/error.h` line 39 — no `public` keyword). So `VectorIndexCache::GetOrCreate`'s `catch(const std::exception&)` would NOT catch a `tenann::Error` thrown from `reader->ReadIndexFile()` — it'd escape the cache and crash the BE. Added a `catch(const tenann::Error&)` arm before the `std::exception` arm.

CelerData-Reviewer:
@codex review

be/src/storage/index/vector/vector_index_reader_factory.cpp
```diff
@@ -19,12 +19,25 @@
...
 static Status create_from_file_impl(const std::string& index_path,
                                     const std::shared_ptr<tenann::IndexMeta>& /*index_meta*/,
                                     std::shared_ptr<VectorIndexReader>* vector_index_reader, FileSystem* fs) {
+    // Warm path: an entry in the cache means the .vi file exists and is not
+    // an empty-mark placeholder, so we can skip the OSS/S3 HEAD round-trips
+    // (path_exist + new_random_access_file + get_size) that the cold path runs.
+    auto* cache = tenann::GetGlobalIndexCache();
+    if (cache != nullptr) {
+        tenann::IndexCacheHandle probe;
+        if (cache->Lookup(tenann::CacheKey(index_path), &probe)) {
```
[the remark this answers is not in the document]
sevev: Good catch. Fixed in 4d1693ad131 (folded into the metrics-wiring commit, not a new commit) by making `VectorIndexCache::Lookup` counter-silent.

The factory's `Lookup` probe is purely an opportunistic check to skip the OSS HEAD round-trip — it isn't an independent cache request, so counting it would double the absolute `vector_index_cache_lookup_count` / `_hit_count` gauges (and the dynamic-rate derivatives) on every warm-cache reader init. After the fix, `lookup_count` / `hit_count` track `GetOrCreate` calls exclusively, which is what the historical (pre-PR) tenann::IndexCache counters were measuring — so dashboards/alerts comparing against pre-PR baselines stay accurate.

Hit-ratio semantics were already correct (both counters were inflated 2x, so the ratio came out right), but absolute counts and `dynamic_*_count` gauges would have looked ~2x what they used to. Test `LookupAndHitCounters_TrackedAcrossPaths` updated to reflect the new semantics.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/index/vector/vector_index_cache.cpp
```diff
@@ -0,0 +1,118 @@
...
+// Deleter releases the entry ref when the handle is destroyed. The lambda
+// captures _cache as a raw pointer: callers (TenANNReader::_cache_handle,
+// etc.) must release every handle before ExecEnv::destroy() runs reset().
+// _wait_for_fragments_finish() at shutdown serves as the drain point. Same
+// contract as PrimaryIndex/StoragePageCache and the original tenann global
+// cache; see DynamicCache::~DynamicCache notes.
+tenann::IndexCacheHandle VectorIndexCache::_wrap(Entry* entry, tenann::IndexRef ref) {
+    Cache* cache = &_cache;
+    return tenann::IndexCacheHandle(
+            std::move(ref), std::shared_ptr<void>(entry, [cache](void* p) { cache->release(static_cast<Entry*>(p)); }));
```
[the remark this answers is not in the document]
sevev: > **P1 Badge Avoid dangling cache pointer in index-handle deleter**
> 
> The handle deleter stores a raw `Cache*` and calls `cache->release(...)` later, but `ExecEnv::destroy()` now unconditionally does `_vector_index_cache.reset()` (`be/src/runtime/exec_env.cpp`), so any `IndexCacheHandle` that outlives shutdown will dereference freed memory. This is reachable when shutdown does not fully drain all fragments (for example, graceful-exit wait is bounded by `loop_count_wait_fragments_finish`, default 2 loops), because surviving `TenANNReader` instances keep `_cache_handle` and trigger the deleter after the cache object is gone, causing a UAF crash.
> 
> Useful? React with 👍 / 👎.

it only happened during be exit and it is acceptable

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/index/vector/tenann_index_reader.cpp
```diff
@@ -65,43 +73,70 @@ void apply_index_reader_cache_options(tenann::IndexMeta* meta_copy) {
...
+            if (fs != nullptr) {
+                auto opened_or = VectorIndexFileReader::open(fs, index_path);
+                if (!opened_or.ok()) {
+                    load_status = opened_or.status();
+                    return nullptr;
+                }
+                external_file_reader = std::shared_ptr<VectorIndexFileReader>(opened_or.value().release());
+            }
+            auto reader = tenann::IndexFactory::CreateReaderFromMeta(meta_copy);
+            reader->SetIndexCache(cache);
```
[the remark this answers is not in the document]
sevev: Not a bug — this conflates `IndexReader::ReadIndex(path)` with `IndexReader::ReadIndexFile(path)`. The loader calls the latter, which is the leaf reader that does **not** touch the cache.

The three cache call sites in tenann (verified by `grep -rn 'index_cache' tenann/index/`):

1. `IndexReader::ReadIndex` — wraps `ReadIndexFile` and conditionally goes through the cache when `cache_index_file=true`. **Our loader doesn't call this.** We call `reader->ReadIndexFile(index_path)` directly, deliberately, so the outer `cache->GetOrCreate(index_path, ...)` is the single owner of `index_path` population.

2. `IndexWriter::Insert` — write-path; not on the read path.

3. `BlockCacheInvertedLists::get_ptr(list_no)` (IVF-PQ block cache) — called **during search**, not during `ReadIndexFile`. The construction of `BlockCacheInvertedLists` inside `ReadIndexFile` just allocates the lazy-list metadata; no `Lookup`/`Insert` happens. Once the outer `GetOrCreate` returns and the per-entry mutex is released, search-time `get_ptr` does `cache->Lookup(cache_keys[list_no], ...)`. The cache_keys are `hash(filename) + "_" + mtime + "_" + i` (see `index_ivfpq_reader.cc:289-291`) — **different from the file `index_path`**, so they hit a different cache entry with a different per-entry mutex. No re-entry on the same key, no deadlock.

So the "outer mutex held + inner cache lookup on the same key" scenario doesn't exist:
- HNSW path: `ReadIndexFile` does only `faiss::read_index` (`faiss_index_reader.cc:34-46`). Zero cache calls.
- IVF-PQ path: `ReadIndexFile` builds metadata; cache calls happen later at search time on per-list keys.

The `cache_index_file_key=true` setting `apply_index_reader_cache_options` writes is dead-but-harmless in this path — only `IndexReader::ReadIndex` reads it, and we don't call that.

Closing as not-a-bug; happy to remove the dead `cache_index_file_key` assignment as a follow-up cleanup if helpful.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/service/service_be/config_update_hooks.cpp
```diff
@@ -92,6 +95,22 @@ void register_config_update_hooks(ExecEnv* exec_env, const GlobalEnv& global_env
         cache->set_capacity(cache_limit);
         return Status::OK();
     });
+    registry->register_callback("vector_query_cache_capacity", [=]() -> Status {
+        if (exec_env == nullptr || exec_env->vector_index_cache() == nullptr) {
+            return Status::InternalError("Vector index cache is not initialized");
```
kevincai, line 101: or it could be the TENNA macro is not opened during the build phase?

be/src/storage/index/vector/vector_index_cache.cpp
```diff
@@ -0,0 +1,121 @@
...
+    for (auto* entry : entries) {
+        auto g = entry->value().guard();
+        entry->value().set_ref(nullptr);
+    }
+    for (auto* entry : entries) {
+        _cache.release(entry);
+    }
+}
+
+bool VectorIndexCache::Lookup(const tenann::CacheKey& key, tenann::IndexCacheHandle* handle) {
```
kevincai, line 43: the two counters shall be updated inside this method?

be/src/storage/index/vector/vector_index_cache.h
```diff
@@ -0,0 +1,110 @@
...
+
+#ifdef WITH_TENANN
+
+#include <atomic>
+#include <mutex>
+#include <ostream>
+#include <string>
+
+#include "tenann/index/index_cache.h"
+#include "util/dynamic_cache.h"
```
kevincai, line 32: wondering the design decision of choosing this DynamicCache as the base block. Internal StoragePageCache/MetadataCache uses ShardedLRUCache as the base block.

thirdparty/vars-x86_64.sh
```diff
@@ -40,10 +40,10 @@ JINDOSDK_SOURCE="jindosdk-4.6.8-linux"
 JINDOSDK_MD5SUM="5436e4fe39c4dfdc942e41821f1dd8a9"
 
 # tenann
-TENANN_DOWNLOAD="https://github.com/StarRocks/tenann/releases/download/v0.5.1-beta.3/tenann-v0.5.1-beta.3-x86_64.tar.gz"
-TENANN_NAME="tenann-v0.5.1-beta.3-x86_64.tar.gz"
-TENANN_SOURCE="tenann-v0.5.1-beta.3"
-TENANN_MD5SUM="a34fa3c131c204e3e76de4b9d266bd9f"
+TENANN_DOWNLOAD="https://github.com/StarRocks/tenann/releases/download/v0.5.1-rc1/tenann-v0.5.1-rc1-x86_64.tar.gz"
```
kevincai, line 43: separate the thirdparty update with the code change, avoid heavy thirdparty rebuild CI flow.

test/sql/test_vector_index/T/test_vector_index_cache
```diff
@@ -0,0 +1,97 @@
+-- name: test_vector_index_cache @sequential @native
```
kevincai: thought this is built for @cloud specific or both?

be/src/service/service_be/config_update_hooks.cpp
sevev: > or it could be the TENNA macro is not opened during the build phase?

fix it.

be/src/storage/index/vector/vector_index_cache.cpp
sevev: > the two counters shall be updated inside this method?

this is the warm-path probe in `VectorIndexReaderFactory` and counting it would double up with the `GetOrCreate` call right after.

be/src/storage/index/vector/vector_index_cache.h
sevev: > wondering the design decision of choosing this DynamicCache as the base block. Internal StoragePageCache/MetadataCache uses ShardedLRUCache as the base block.

the core reason is 
1. I aim to implement single-flight for the vector index cache. Loading vector indexes is expensive due to heavy data reading and FAISS parsing. Missing single-flight will result in excessive memory usage. I reuse DynamicCache because it already has mature single-flight support.
2. We can only calculate the size of a vector index after loading it — IndexRef::EstimateMemoryUsage() works post-load. This is consistent with DynamicCache. However, ShardedLRUCache expects the entry size to be known when inserting entries.

thirdparty/vars-x86_64.sh
sevev: > separate the thirdparty update with the code change, avoid heavy thirdparty rebuild CI flow.

`v0.5.1-rc1` introduces breaking interface changes and is not fully backward compatible, so I bundled all changes into a single PR.
The reason for these breaking changes is that keeping compatibility with legacy interfaces left a large amount of dead code. Since we will submit follow-up PRs to clean it up later, I consolidated everything into one PR for now.

test/sql/test_vector_index/T/test_vector_index_cache
sevev: > thought this is built for @cloud specific or both?

both, fix it.

be/src/storage/index/vector/vector_index_cache.h
kevincai: can put this info into code via comments.

be/src/storage/index/vector/vector_index_cache.cpp
kevincai: check the LRUCache metrics, it is not only the hit and lookup, some other metrics may worth exposed as well.

kevincai (approved):

luohaha (approved):

meegoo (approved):

PR #61857, opened 2025-08-12
[Refactor] (IVM Part1) Refactor PartitionBasedMvRefreshProcessor

LiShuMing:
## Why I'm doing:

To support `IVMBasedMVRefreshProcessor`, split the original `PartitionBasedMvRefreshProcessor` class into an abstract class for better extension and reuse.

## What I'm doing:
This pull request introduces significant improvements to the materialized view (MV) refresh system in the StarRocks scheduler. The main changes include adding a new MV refresh processor (`MVTaskRunProcessor`), updating how MV refresh tasks are handled and post-processed, and enhancing partition refresh strategy management. These updates improve reliability, extensibility, and maintainability of MV refresh operations.

**Materialized View Refresh System Enhancements**

* Added a new `MVTaskRunProcessor` class to handle MV refresh tasks, replacing the previous `PartitionBasedMvRefreshProcessor`. This new processor supports better error handling, retry logic, and post-processing for MV refreshes. [[1]](diffhunk://#diff-b48dbddc52bbd26c6c4325800e5837ece612e6feaed1ad8f3783feeb8e5650f5R1-R332) [[2]](diffhunk://#diff-b1c29332e81267e7d36229d5757b6d041f2ed93859a68bfae5de8c7143b5edeaL54-R54)
* Updated the MV refresh task builder to use `MVTaskRunProcessor` for MV tasks, ensuring all new MV refreshes use the improved processor logic.

**MV Base Table Snapshot Support**

* Added a new `BaseTableSnapshotInfo` abstract class to represent snapshots of base tables for MVs, supporting consistent refresh and metadata management.

Fixes https://github.com/StarRocks/starrocks/issues/61789

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
[...]

Issue #59948, opened 2025-06-16
Add Schema Name to logs


arielmoraes:
I want to request an enhancement on the logs to add the Schema Name. Our solution uses a schema per tenant approach, where each schema has the same modeling (same MVs etc). The problem is the logs don't show for which schema the message is from, example: `2025-06-16 14:30:22.985Z INFO (starrocks-taskrun-pool-0|234) [PartitionBasedMvRefreshProcessor.getRefTableRefreshPartitions():1302] [dim_classification_tags] ref-base-table dim_classification_tags is not found in `mvRefBaseTableIntersectedPartitions` because of empty update`. The MV `dim_classification_tags` exists in multiple schemas with the same name.

alvin-phoenix-ai:
Because StarRocks is compatible with MySQL, Tables belong to databases, and `Schema` is not used very much. I guess what you mean by `Schema name` is actually `Database name`

arielmoraes:
Yes, I meant database, I used the term Schema because in some places MySql uses that wording.

alvin-phoenix-ai:
@arielmoraes Do you want to try to resolve it?

PR #73521, opened 2026-05-19
[Enhancement] Include database name in MV refresh log prefix

eshishki:
## Why

In multi-tenant setups where the same materialized view name exists in multiple schemas, the MV refresh logger prefix `[mvName]` cannot distinguish them. Original report in #59948 describes a `dim_classification_tags` MV that lives in many per-tenant schemas — log entries are indistinguishable across tenants.

Example log line today:
```
[PartitionBasedMvRefreshProcessor.getRefTableRefreshPartitions():1302] [dim_classification_tags] ref-base-table dim_classification_tags is not found in `mvRefBaseTableIntersectedPartitions` because of empty update
```
No way to tell which `dim_classification_tags` this is.

## What

`MVTraceUtils.getLogPrefix(MaterializedView)` now produces `[dbName.mvName]` instead of `[mvName]`. The prefix is computed once per logger instantiation (once per refresh-processor instance), not per log message, so the extra `getDb(dbId)` lookup is not on any hot path.

If the database lookup returns null (e.g. concurrent drop), the prefix gracefully falls back to `[mvName]`.

## Background

#60096 by @cutiechi proposed the same fix in the old `PartitionBasedMvRefreshProcessor.java`. That file was removed by the IVM refactor in #61857, so #60096 cannot be rebased without a rewrite. This PR applies the same idea to the new central log-prefix builder so it benefits all refresh processors (`MVPCTRefreshProcessor`, `MVIVMRefreshProcessor`, `MVHybridRefreshProcessor`).

cc @LiShuMing (approved #60096, requested rebase 2025-11-10), @cutiechi (original author).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/MVTraceUtils.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/MVTraceUtils.java
@@ -17,6 +17,7 @@ package com.starrocks.scheduler.mv;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.starrocks.catalog.BaseTableInfo;
+import com.starrocks.catalog.Database;
 import com.starrocks.catalog.HiveTable;
 import com.starrocks.catalog.MaterializedView;
 import com.starrocks.catalog.OlapTable;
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

Youngwb (approved):

PR #73549, opened 2026-05-20
[BugFix] Convert explicit-txn COMMIT timeout from seconds to milliseconds

sevev:
## Why I'm doing:

`TransactionStmtExecutor.commitStmt()` reads `query_timeout` in **seconds** (default 300) and passes it directly to `GlobalTransactionMgr.retryCommitOnRateLimitExceeded(..., long timeoutMs)`, which expects **milliseconds** and forwards the value to `Locker.tryLockTablesWithIntensiveDbLock(..., timeoutMs, TimeUnit.MILLISECONDS)`.

Effect: a `COMMIT` inside an explicit transaction (`START TRANSACTION ... ; INSERT ... ; COMMIT;`) only waits up to ~300 ms (not 300 s) for the database write lock. Any concurrent writer briefly holding the lock causes:

```
LockTimeoutException: get database write lock timeout, database=<db>, timeout=300ms
```

Raising `query_timeout` does not help — the unit mismatch caps the wait at `query_timeout` ms regardless. Only the explicit-transaction COMMIT path is affected; the implicit single-`INSERT` path multiplies seconds by 1000 correctly via `jobDeadLineMs` and is fine.

## What I'm doing:

- Multiply the timeout by `1000L` at the call site so the value handed to `retryCommitOnRateLimitExceeded(..., long timeoutMs)` is in milliseconds.
- Add `ExplicitTxnTest.testCommitTimeoutPassedInMilliseconds`, which mocks `retryCommitOnRateLimitExceeded` to capture the `timeoutMs` argument and asserts it equals `query_timeout * 1000`. Verified to fail before the fix (`expected: <300000> but was: <300>`) and pass after.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/TransactionStmtExecutor.java
@@ -312,7 +312,7 @@ public class TransactionStmtExecutor {
                     commitInfos,
                     failInfos,
                     txnCommitAttachment,
-                    timeout);
+                    timeout * 1000L);
 
             // Re-fetch transactionState after commit because the COW pattern in DatabaseTransactionMgr
             // replaces the in-memory state with a deep copy, making the original reference stale.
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

meegoo (approved):

luohaha (approved):

PR #73557, opened 2026-05-20
[Enhancement] Surface global-dict scan application in profile

### LINKED ISSUE #73587: [Enhancement] Surface FE global-dict scan application in BE scan profile
## Enhancement

When the FE optimizer rewrites a string column into a global-dictionary
code, the BE scan is supposed to deliver `INT32` codes directly without
re-materializing strings.  Today there is no profile signal that tells
us whether the dict-encoded scan path actually ran for a given scan
instance — a missing or broken dict application silently falls back
to string materialization and the only visible symptom is "the query
got slower."

### What we want

A runtime profile signal that exposes, per scan instance, whether the
FE global dictionary low-cardinality optimization actually reached the
scan output.  The signal must cover:

- native OLAP (pipeline + legacy non-pipeline)
- shared-data lake
- Hive / Iceberg / Hudi / Paimon Parquet

### Acceptance

- Profile diff is sufficient to answer "did the global-dict opt apply
  on this scan?" without running with extra logging or a custom build.
- Observability only — no behavior change for any query.

eshishki:
Fixes #73587.

## Why I'm doing:

When FE's low-cardinality rewrite turns a string column into a global-dict
code, the scan should deliver Int32 codes directly without re-materializing
the underlying string.  Today there's no profile signal that surfaces whether
the dict-encoded path actually ran for a given scan instance — a missing
or broken dict-application silently falls back to string materialization
and the only visible symptom is degraded performance.

## What I'm doing:

Adds runtime profile signals so a profile diff exposes whether the FE
global dictionary low-cardinality optimization actually reached the scan.
Covers native OLAP, shared-data lake, and Hive/Iceberg/Hudi/Paimon
Parquet workloads.

Native OLAP and Lake scans publish two info-strings after
`_init_global_dicts` has intersected the FE-level query dictionary with the
materialized tablet schema, so the size reflects columns that will actually
flow through the encoded path on this scan instance:

    GlobalDictOptApplied   = true
    GlobalDictAppliedSlots = <n>

Touch points:
- `be/src/exec/pipeline/scan/olap_chunk_source.cpp` — pipeline OLAP
- `be/src/exec/tablet_scanner.cpp` — legacy non-pipeline OLAP
- `be/src/connector/lake_connector.cpp` — shared-data lake

**What the OLAP marker covers — and what it does not:**

Native OLAP stores string columns via a per-segment local dictionary
(codes 0..N local to each segment).  When the FE global-dict opt is
applied, the scan emits Int32 codes from the **global** dictionary
instead of strings.  Inside `SegmentIterator` there are two paths to
produce that encoded output:

1. **Fast path** — local dict ⊆ global dict: a cheap mapping
   `local_code → global_code` per row.  This is the win the optimization
   was designed for.
2. **Force-encode path** — `ColumnDecoder::need_force_encode_to_global_id()`
   returns true when the local dictionary contains entries that are not
   in the global dictionary (e.g. global dict was built from a sampled
   subset and a segment has values the sample missed).  The scan then
   decodes the string from the local code and re-encodes it into a
   global code — still cheaper than returning raw strings, but
   noticeably more work than the fast-path mapping.

The marker only states *"this scan's output schema is FE-global-dict-
encoded for at least one materialized column,"* which is the bit
downstream operators care about (they branch on encoded vs. raw).
It does NOT say *"the fast-path mapping was hit on every row."*  If
someone sees `GlobalDictOptApplied = true` and the query is still slow,
the force-encode path is a candidate root cause but this marker alone
cannot confirm it — a separate per-row counter would be needed for
that, out of scope here.

For Hive/Iceberg/Hudi/Paimon the marker is published at GroupReader level
from inside the parquet column reader factory, so it reports per-column
whether the file's data delivers dict codes directly (count surfaces in
the `Parquet/GlobalDict*` counters on the scan profile).

ORC's local dict-filter machinery is unrelated to the FE global-dict
output encoding signalled here and is left untouched.

Observability only; no behavior change for any query.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -483,6 +483,13 @@ Status LakeDataSource::init_tablet_reader(RuntimeState* runtime_state) {
     RETURN_IF_ERROR(_extend_schema_by_access_paths());
     ASSIGN_OR_RETURN(_tablet_schema, extend_schema_by_virtual_columns(_tablet_schema, *_slots));
     RETURN_IF_ERROR(init_global_dicts(&_params));
+    // init_global_dicts intersects the FE-level dict map with this scan's
+    // materialized tablet schema, so the resulting size counts columns that
+    // will actually flow through the encoded path on this scan instance.
+    if (_params.global_dictmaps != nullptr && !_params.global_dictmaps->empty() && _runtime_profile != nullptr) {
+        _runtime_profile->add_info_string("GlobalDictOptApplied", "true");
+        _runtime_profile->add_info_string("GlobalDictAppliedSlots", std::to_string(_params.global_dictmaps->size()));
+    }
     RETURN_IF_ERROR(init_unused_output_columns(thrift_lake_scan_node.unused_output_column_name));
     RETURN_IF_ERROR(init_reader_params(_scanner_ranges));
 
--- a/be/src/exec/hdfs_scanner/hdfs_scanner.h
+++ b/be/src/exec/hdfs_scanner/hdfs_scanner.h
@@ -124,6 +124,18 @@ struct HdfsScanStats {
     // late materialize round-by-round
     int64_t group_min_round_cost = 0;
 
+    // Parquet global-dict opt application — populated by parquet::GroupReader as
+    // it wraps column readers with dict-aware adapters.  These counters answer
+    // "did the FE global-dict optimization actually reach Parquet column readers
+    // on this scan instance, and through which wrapper path?" for Hive/Iceberg
+    // workloads where the connector-level dict map can otherwise lie about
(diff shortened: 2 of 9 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

dirtysalt (approved):

PR #73558, opened 2026-05-20
[Enhancement] Use array-table for SMALLINT group-by keys

### LINKED ISSUE #73588: [Enhancement] Use array-table for SMALLINT GROUP BY / DISTINCT keys
## Enhancement

`GROUP BY` / `DISTINCT` on a `SMALLINT` column today pays full
`phmap::flat_hash_map<int16_t>` cost — hash, probe chain, key compare —
even though the key space is only 2¹⁶ values and would fit in a
direct-indexed 65 536-cell array.  TINYINT and BOOL already use a
fixed-size array hash map (`SmallFixedSizeHashMap`); SMALLINT was left
on phmap.

### What we want

Switch `Int16AggHashMap` / `Int16AggHashSet` to the direct-indexed array
table that TINYINT and BOOL already use, so an `int16_t` lookup becomes
`array[static_cast<uint16_t>(key)]` — no hashing, no probe, no key
compare.

### Acceptance

- `GROUP BY` and `DISTINCT` on `SMALLINT` (incl. nullable) measurably
  faster than the phmap baseline across density / distribution shapes;
  no regression on any shape.
- Memory accounting (`allocated_memory_usage`, `reserved_memory_usage`,
  `dump_bound`) reports the true byte footprint of the direct-array
  table — not the slot count, not `sizeof(KeyType) * capacity()`.
  The same fix should retroactively cover the pre-existing TINYINT /
  BOOL array path, which had the same accounting bug.

eshishki:
Fixes #73588.

## Why I'm doing:

`GROUP BY` on a `SMALLINT` column today pays the full phmap cost — hash,
probe chain, key compare — even though the key space is only 2¹⁶ values
and would fit in a direct-indexed array.  TINYINT/BOOL already use a
fixed-size array hash map; SMALLINT was left on phmap.  ClickHouse ships
the same direct-array shape as `AggregatedDataWithUInt16Key` (`key16`).

## What I'm doing:

Switch `Int16AggHashMap` / `Int16AggHashSet` from
`phmap::flat_hash_{map,set}<int16_t>` to `SmallFixedSizeHash{Map,Set}<int16_t>`,
giving a direct-indexed 65 536-cell array.  No hashing, no conflict
chain, no key comparison — just `array[static_cast<uint16_t>(key)]`.

The wrapper (`AggHashMapWithOneNumberKey`) and the nullable wrapper
already specialize on `is_no_prefetch_map`, so prefetch and the
consecutive-keys cache are correctly disabled for the new path.
`convert_to_two_level` only fires for slice/string variants, so the
Int16 path is unaffected.

**No session-var gate:** the optimization is unconditionally faster
(no hash, no probe, no key compare) and ClickHouse ships the equivalent
without a switch.

### Memory accounting fix (rolled in)

`SmallFixedSizeHashMap<int16_t, AggDataPtr>` owns a dense ~512 KiB
pointer array per hash map.  The prior accounting paths under-reported
this:
- `allocated_memory_usage` computed `sizeof(KeyType) * capacity()` (~128 KiB).
- `reserved_memory_usage` went through `SmallFixedSizeHash{Map,Set}::dump_bound()`,
  which returned the slot **count** (65 536) instead of bytes.

Both now report `sizeof(ValueType) * (hash_table_size + 1)`:
- `allocated_memory_usage` consults a new `bucket_byte_size()` constant
  via a `requires` expression.
- `dump_bound()` returns the same byte count, matching phmap's
  byte-based dump_bound semantics.

This keeps spill / streaming-memory budgets and serialization paths
honest at the true footprint of this **and the pre-existing TINYINT/BOOL
array path** (which had the same accounting bug).

### Compressed-key opt gate (rolled in)

`_try_to_apply_compressed_key_opt` rewrites the agg hash variant to
the slice_cx1/4/8/16 path when FE supplies range stats via the
`group_by_min_max` channel and `could_apply_bitcompress_opt`
succeeds.  For TINYINT / BOOL / SMALLINT this **overrides** the
direct-array route with `slice_cx1`, which sits on the same
`SmallFixedSizeHashMap<int8_t>` under the hood but layers a per-row
`bitcompress_serialize` step on top.

The observable consequence: SMALLINT `GROUP BY` would run **slower**
when FE has range stats than when it doesn't.  More information at
FE → slower BE.  Pre-existing for TINYINT / BOOL on `main`; newly
exposed for SMALLINT by the direct-array switch in this PR.

Fix: gate the rewrite on the source type.  When the single group-by
column is `TYPE_TINYINT`, `TYPE_BOOLEAN`, or `TYPE_SMALLINT`, return
the input variant unchanged.  Multi-column GROUP BY and INT / BIGINT /
DECIMAL paths are unaffected — `slice_cx1` remains the best
available path for those today.

### Bench

`be/src/bench/agg_hash_map_bench.cpp` instantiates the phmap baseline
and the array treatment of `AggHashMapWithOneNumberKey<TYPE_SMALLINT,
...>` side-by-side and drives them through the production callsite
`build_hash_map` with a production-shaped `HashTableKeyAllocator`
(`sizeof(int64_t)` aggregate state matching `COUNT(*)`).

Sections:
- **A**: density {1, 4, 24, 366, 1000, 10000, 65536} × distribution
  {random, sorted, clustered-run-length-64, Zipf s=1.5} at 100M rows.
  Forces L3 eviction of the column buffer so hash-table cost dominates.
  Reports both hash-only (`BM_Int16BuildOnly`) and hash+COUNT-update
  (`BM_Int16BuildAndCount`).
- **B**: amortization sweep at density=24 sorted, rows {4K, 100K, 1M,
  10M, 100M}.  `BM_Int16BuildOnly` (ctor paused) + `BM_Int16BuildWithCtor`
  (ctor inside timing) side-by-side shows the row count beyond which the
  array's 524 KiB upfront alloc stops dominating.
- **C**: construct-only cost (`BM_Int16Construct`) — makes the 524 KiB
  memset explicit.
- **D**: nullable coverage at null fractions {0%, 10%, 50%}.

#### Results (100M rows, median, x86_64 r6a.xlarge, ASLR on)

Hash-only:

| distinct | dist | phmap | array | speedup |
|----------|------|-------|-------|---------|
| 1 | random | 424 ms | 130 ms | **3.3×** |
| 24 | random | 539 | 130 | **4.1×** |
| 1000 | sorted | 825 | 132 | **6.3×** |
| 10000 | random | 980 | 148 | **6.6×** |
| 65536 | random | 1121 | 209 | **5.4×** |
| 24 | Zipf | 527 | 130 | **4.1×** |

Build+count (hash insert + per-row counter increment):
**1.6–3.9×** depending on density.

#### Memory trade-off

Array path is always ~524 KiB; phmap scales with distinct.  Crossover
at `distinct ≈ 10k`: at 65 536 distinct phmap holds **2.23 MiB** vs.
**524 KiB** for the array — i.e. the array path is also memory-cheaper
on the densest workloads it has to handle.

#### Amortization break-even (Section B, density=24 sorted)

| rows | phmap ctor-included | array ctor-included | array wins from |
|------|---------------------|----------------------|------------------|
| 4 096 | 0.010 ms | 0.015 ms | — |
| 100 000 | 0.214 | 0.075 | ✅ ~2.85× |
| 1M+ | scales linearly | scales linearly | ✅ ~3.16× |

The 524 KiB ctor is a one-time hit; the array path beats phmap from
~50–100 K rows onward.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exec/aggregate/agg_hash_map.h
+++ b/be/src/exec/aggregate/agg_hash_map.h
@@ -60,7 +60,7 @@ concept AllocFunc = HasKeyType<HashMapWithKey>&& requires(T t, const typename Ha
 template <PhmapSeed seed>
 using Int8AggHashMap = SmallFixedSizeHashMap<int8_t, AggDataPtr, seed>;
 template <PhmapSeed seed>
-using Int16AggHashMap = phmap::flat_hash_map<int16_t, AggDataPtr, StdHashWithSeed<int16_t, seed>>;
+using Int16AggHashMap = SmallFixedSizeHashMap<int16_t, AggDataPtr, seed>;
 template <PhmapSeed seed>
 using Int32AggHashMap = phmap::flat_hash_map<int32_t, AggDataPtr, StdHashWithSeed<int32_t, seed>>;
 template <PhmapSeed seed>
--- a/be/src/exec/aggregate/agg_hash_set.h
+++ b/be/src/exec/aggregate/agg_hash_set.h
@@ -56,7 +56,7 @@ static constexpr size_t AGG_HASH_MAP_DEFAULT_PREFETCH_DIST = 16;
 template <PhmapSeed seed>
 using Int8AggHashSet = SmallFixedSizeHashSet<int8_t, seed>;
 template <PhmapSeed seed>
-using Int16AggHashSet = phmap::flat_hash_set<int16_t, StdHashWithSeed<int16_t, seed>>;
+using Int16AggHashSet = SmallFixedSizeHashSet<int16_t, seed>;
 template <PhmapSeed seed>
 using Int32AggHashSet = phmap::flat_hash_set<int32_t, StdHashWithSeed<int32_t, seed>>;
 template <PhmapSeed seed>
--- a/be/src/exec/aggregate/agg_hash_variant.cpp
+++ b/be/src/exec/aggregate/agg_hash_variant.cpp
@@ -350,7 +350,7 @@ bool AggHashMapVariant::need_expand(size_t increasement) const {
     return visit([increasement](const auto& hash_map_with_key) {
         using HashMap = std::remove_reference_t<decltype(hash_map_with_key->hash_map)>;
         const size_t size = hash_map_with_key->hash_map.size() + increasement;
-        if constexpr (requires { HashMap::bucket_byte_size(); }) {
+        if constexpr (is_fixed_hash_map_v<HashMap>) {
             return size > HashMap::hash_table_size;
         }
         const size_t capacity = hash_map_with_key->hash_map.capacity();
@@ -369,7 +369,7 @@ size_t AggHashMapVariant::allocated_memory_usage(const MemPool* pool) const {
     return visit([pool](const auto& hash_map_with_key) {
         using HashMap = std::remove_reference_t<decltype(hash_map_with_key->hash_map)>;
         size_t hash_map_bytes;
-        if constexpr (requires { HashMap::bucket_byte_size(); }) {
+        if constexpr (is_fixed_hash_map_v<HashMap>) {
             // SmallFixedSizeHashMap pre-allocates a dense pointer array; the
             // SMALLINT path is ~512 KiB and must be reported as such, not as
             // sizeof(KeyType) * capacity.
@@ -455,7 +455,7 @@ bool AggHashSetVariant::need_expand(size_t increasement) const {
     return visit([increasement](const auto& hash_set_with_key) {
         using HashSet = std::remove_reference_t<decltype(hash_set_with_key->hash_set)>;
         const size_t size = hash_set_with_key->hash_set.size() + increasement;
-        if constexpr (requires { HashSet::bucket_byte_size(); }) {
+        if constexpr (is_fixed_hash_set_v<HashSet>) {
             return size > HashSet::hash_table_size;
         }
         const size_t capacity = hash_set_with_key->hash_set.capacity();
@@ -474,7 +474,7 @@ size_t AggHashSetVariant::allocated_memory_usage(const MemPool* pool) const {
     return visit([&](auto& hash_set_with_key) {
         using HashSet = std::remove_reference_t<decltype(hash_set_with_key->hash_set)>;
         size_t hash_set_bytes;
-        if constexpr (requires { HashSet::bucket_byte_size(); }) {
+        if constexpr (is_fixed_hash_set_v<HashSet>) {
             hash_set_bytes = HashSet::bucket_byte_size();
         } else {
             hash_set_bytes = sizeof(typename HashSet::key_type) * hash_set_with_key->hash_set.capacity();
--- a/be/src/exec/aggregate/agg_hash_variant.cpp
+++ b/be/src/exec/aggregate/agg_hash_variant.cpp
@@ -342,11 +342,20 @@ size_t AggHashMapVariant::size() const {
 }
 
 bool AggHashMapVariant::need_expand(size_t increasement) const {
-    size_t capacity = this->capacity();
-    // TODO: think about two-level hashmap
-    size_t size = this->size() + increasement;
-    // see detail implement in reset_growth_left
-    return size >= capacity - capacity / 8;
+    // Direct-array hash maps (SmallFixedSizeHashMap for TINYINT/BOOL/SMALLINT
+    // and the low-card-dict uint8 variant) have a hard-coded full keyspace
+    // and never rehash, so streaming/spill paths should only fall back when
+    // the array is actually full -- not at the 87.5% growth-heuristic
+    // threshold used by phmap.
+    return visit([increasement](const auto& hash_map_with_key) {
+        using HashMap = std::remove_reference_t<decltype(hash_map_with_key->hash_map)>;
+        const size_t size = hash_map_with_key->hash_map.size() + increasement;
+        if constexpr (requires { HashMap::bucket_byte_size(); }) {
+            return size > HashMap::hash_table_size;
+        }
+        const size_t capacity = hash_map_with_key->hash_map.capacity();
+        return size >= capacity - capacity / 8;
+    });
 }
 
 size_t AggHashMapVariant::reserved_memory_usage(const MemPool* pool) const {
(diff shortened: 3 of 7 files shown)


CelerData-Reviewer:
@codex review

eshishki:
rerun

be/src/exec/aggregate/agg_hash_variant.cpp
```diff
@@ -342,11 +342,20 @@ size_t AggHashMapVariant::size() const {
...
+    // the array is actually full -- not at the 87.5% growth-heuristic
+    // threshold used by phmap.
+    return visit([increasement](const auto& hash_map_with_key) {
+        using HashMap = std::remove_reference_t<decltype(hash_map_with_key->hash_map)>;
+        const size_t size = hash_map_with_key->hash_map.size() + increasement;
+        if constexpr (requires { HashMap::bucket_byte_size(); }) {
+            return size > HashMap::hash_table_size;
+        }
+        const size_t capacity = hash_map_with_key->hash_map.capacity();
+        return size >= capacity - capacity / 8;
```
stdpain, line 357: We can directly check if it is a fixed hash table.

be/src/exec/aggregator.cpp
```diff
@@ -1538,6 +1538,22 @@ typename HashVariantType::Type Aggregator::_try_to_apply_compressed_key_opt(type
...
+    // direct arrays; the slice_cx1 path sits on the same direct array
+    // under int8 but adds a per-row bitcompress_serialize step, so any
+    // query that supplies range stats via `group_by_min_max` would
+    // otherwise silently regress to the slower slice path.
+    if (_group_by_types.size() == 1) {
+        switch (_group_by_types[0].result_type.type) {
+        case TYPE_TINYINT:
+        case TYPE_BOOLEAN:
+        case TYPE_SMALLINT:
+            return type;
```
stdpain, line 1552: I'm not entirely sure, but for int16, compressed short key to int8 might perform better.

be/src/exec/aggregate/agg_hash_map.h
```diff
@@ -60,7 +60,7 @@ concept AllocFunc = HasKeyType<HashMapWithKey>&& requires(T t, const typename Ha
 template <PhmapSeed seed>
 using Int8AggHashMap = SmallFixedSizeHashMap<int8_t, AggDataPtr, seed>;
 template <PhmapSeed seed>
-using Int16AggHashMap = phmap::flat_hash_map<int16_t, AggDataPtr, StdHashWithSeed<int16_t, seed>>;
+using Int16AggHashMap = SmallFixedSizeHashMap<int16_t, AggDataPtr, seed>;
```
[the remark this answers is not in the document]
eshishki: Fixed: `AggHashMapVariant::need_expand()` now visits the wrapper and, for fixed maps, falls back to a full-capacity check (`size > HashMap::hash_table_size`) instead of the 87.5% phmap heuristic. Discriminator uses the new `is_fixed_hash_map_v<T>` trait. phmap-backed variants are unchanged.

be/src/exec/aggregate/agg_hash_set.h
```diff
@@ -56,7 +56,7 @@ static constexpr size_t AGG_HASH_MAP_DEFAULT_PREFETCH_DIST = 16;
 template <PhmapSeed seed>
 using Int8AggHashSet = SmallFixedSizeHashSet<int8_t, seed>;
 template <PhmapSeed seed>
-using Int16AggHashSet = phmap::flat_hash_set<int16_t, StdHashWithSeed<int16_t, seed>>;
+using Int16AggHashSet = SmallFixedSizeHashSet<int16_t, seed>;
```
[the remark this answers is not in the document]
eshishki: Same fix on the Set side: `AggHashSetVariant::need_expand()` switched to a per-wrapper visit with `is_fixed_hash_set_v<T>` discriminator.

be/src/exec/aggregate/agg_hash_variant.cpp
eshishki: Done in 80db201: introduced `is_fixed_hash_map` / `is_fixed_hash_set` traits in `be/src/base/container/fixed_hash_map.h` (mirrors the existing `no_prefetch_{map,set}` pattern) and switched the four discriminator sites in `agg_hash_variant.cpp` (`need_expand` and `allocated_memory_usage` for both Map and Set) to `is_fixed_hash_{map,set}_v<T>` instead of the `requires { ... bucket_byte_size(); }` SFINAE check.

be/src/exec/aggregator.cpp
eshishki: Benched cx1 vs direct on the same Int16ChunkStream, 100M rows, distinct ∈ {16, 64, 128, 200, 256}, Sorted + Random (`dist=1` / `dist=0`).

Host: 16-core x86, L1d 48 KiB/core, L2 1280 KiB/core, L3 55 MiB shared.

| distinct | dist | ArrayInt16 (524 KiB ht) | SliceCx1 (2 KiB ht) | direct/cx1 |
|---:|:---:|---:|---:|---:|
| 16  | Sorted | 131 ms / 762 M rows/s | 183 ms / 547 M rows/s | 1.40× |
| 16  | Random | 131 ms / 764 M rows/s | 183 ms / 546 M rows/s | 1.40× |
| 64  | Sorted | 131 ms / 763 M rows/s | 183 ms / 547 M rows/s | 1.40× |
| 64  | Random | 131 ms / 765 M rows/s | 186 ms / 539 M rows/s | 1.42× |
| 128 | Sorted | 131 ms / 764 M rows/s | 183 ms / 546 M rows/s | 1.40× |
| 128 | Random | 131 ms / 764 M rows/s | 184 ms / 544 M rows/s | 1.40× |
| 200 | Sorted | 131 ms / 763 M rows/s | 184 ms / 543 M rows/s | 1.40× |
| 200 | Random | 131 ms / 763 M rows/s | 188 ms / 533 M rows/s | 1.43× |
| 256 | Sorted | 131 ms / 763 M rows/s | 185 ms / 542 M rows/s | 1.41× |
| 256 | Random | 131 ms / 763 M rows/s | 187 ms / 534 M rows/s | 1.43× |

Direct stays at ~131 ms / ~763 M rows/s independent of distinct or distribution. SliceCx1 sits at 183–188 ms / ~533–547 M rows/s; the 2 KiB L1-resident table does not recover the per-row `bitcompress_serialize` + slice hash + phmap probe cost. 524 KiB direct array fits in L2 (1280 KiB/core), so cx1's L1 footprint advantage does not pay off either.

Conclusion: the `case TYPE_SMALLINT: return type` guard in `_try_to_apply_compressed_key_opt` stays. Bench lives in `be/src/bench/agg_hash_map_bench.cpp` Section E (added in defe126f), reproduce with:

```
./be/build_Release/src/bench/output/agg_hash_map_bench \
    --benchmark_filter='BM_Int16(Cx1)?BuildOnly.*distinct:(16|64|128|200|256)/dist:'
```

stdpain (approved):

dirtysalt (approved):

PR #73559, opened 2026-05-20
[Enhancement] Use array-table for low-cardinality global-dict group-by keys

### LINKED ISSUE #73589: [Enhancement] Use array-table for low-cardinality global-dict GROUP BY keys
## Enhancement

For single-column INT `GROUP BY` queries where the FE-supplied min/max range fits in ≤16 bits, BE today routes through `slice_cx1` / `slice_cx4` — `phmap<int8_t>` or `phmap<SliceKey4>` with per-row `bitcompress_serialize` (added in #61632). That phmap hop is unnecessary when `(max - min)` fits in 8 or 16 bits: a direct-indexed array hash table backed by `SmallFixedSizeHashMap<uint{8,16}_t>` collapses the per-row work to `(value - min) -> array[key]` and skips the hash + probe + bitcompress.

This is the same shape ClickHouse uses for `low_cardinality_key8`.

### What we want

Add BE hash-map variant `AggHashMapWithCompressibleInt32Key` backed by `SmallFixedSizeHashMap<uint8_t>` (range ≤ 8 bits) or `SmallFixedSizeHashMap<uint16_t>` (range 9-16 bits), and have `_try_to_apply_compressed_key_opt` route single-INT GROUP BY into it before the existing slice_cx* branches.

Dict-encoded varchar slots already arrive at BE as `LowCardDictType == TYPE_INT` with min=0, max=dictSize via `ApplyMinMaxStatisticRule`, so they flow through the same path naturally — no separate dict-only variant needed.

### Acceptance

- Routing fires only for single-column INT GROUP BY with ≤16-bit FE-supplied range. Multi-column GROUP BY and other source types keep the existing slice_cx* paths.
- Negative min values (including \`INT32_MIN\`) route correctly.
- Nullable / only-null shapes produce the same results as the existing hash-map variants.
- No regression on any range × distribution shape in the decision benches.

eshishki:
Fixes #73589.

## Why I'm doing:

After the FE's low-cardinality rewrite, a `GROUP BY` on a global-dict-
encoded string column becomes a `GROUP BY` on an `Int32Column` of dict
codes bounded by `DICT_DECODE_MAX_SIZE` (256).  Today such queries go
through `Int32AggHashMapWithOneNumberKey` (`phmap::flat_hash_map<int32_t>`)
and pay the full hash + probe + key-compare cost even though the key
space is at most 256 values and would fit in a direct-indexed array.
ClickHouse ships the same shape as `low_cardinality_key8`.

## What I'm doing:

Route these queries through a direct 256-cell
`SmallFixedSizeHashMap<uint8_t>` instead.  Two new variant types
(`phase{1,2}_low_card_dict_uint8` plus nullable counterparts) plus a
wrapper `AggHashMapWithLowCardDictKey` that reads `Int32Column`,
narrows each code to `uint8_t` once, and writes back as `int32` on
iteration.

### How the route is picked

The FE marks dict-encoded group-by slots in a new `TAggregationNode`
field `low_card_dict_group_by_slots` (`PlanFragmentBuilder`).  BE routes
into the new variant only when the lone group-by expression is a bare
`SlotRef` matching one of those slots, so a plain `TYPE_INT` `GROUP BY`
on an unrelated integer column is unaffected.

### Why `uint8_t` (not `int8_t`)

`int8_t` would round-trip codes 128..255 through
`reinterpret_cast<KeyType*>(state)` as -128..-1 and sign-extend negative
on iteration.  `static_assert(std::is_unsigned_v<KeyType>)` on the
wrapper locks this.  A UT (`agg_hash_map_test.cpp:LowCardDictAggHashMapTest.RoundTripHighCodes`)
exercises key=255 to lock the invariant.

### Bounds safety

Routing assumes codes fit in `uint8_t`, so the FE additionally requires
`Config.low_cardinality_threshold <= 255` before marking any slot
(`PlanFragmentBuilder.java:2686`).  Without that guard a future config
change that raised the threshold would let dictionaries grow into
codes 256+ that silently narrow to `uint8_t` and collapse distinct
dictionary values into the same bucket.  A comment on
`Config.low_cardinality_threshold` calls out the BE coupling so a
future change to the upper bound is forced to revisit this path.

### Compressed-key skip

`ApplyMinMaxStatisticRule` fills a 0..dictSize range for the same
dict-encoded group-by, so without intervention
`_try_to_apply_compressed_key_opt` observes a tiny bit width and
rewrites the variant from `low_card_dict_uint8` back to `slice_cx1/cx4` —
trading a direct 256-cell lookup for a hash map on a bit-compressed
key.  `_init_agg_hash_variant` detects the dict-uint8 routing and
skips the compressed-key rewrite so the intended path actually runs.

### Session gate

Session-gated by `enable_agg_low_card_dict_array_table` (default true).
Following the same convention as other `enable_agg_*` opts in the BE
(`enable_global_runtime_filter*`, `enable_join_runtime_*`, etc.) so a
regression can be bisected per query without a BE redeploy.

### Profile signal

When the slot list is non-empty, Aggregator emits
`LowCardDictGroupBySlots = <count>` as an info string so a profile diff
makes the routing decision visible.

### Bench

`be/src/bench/agg_low_card_dict_bench.cpp` instantiates the phmap
baseline (`AggHashMapWithOneNumberKey<TYPE_INT, phmap::flat_hash_map<int32_t, AggDataPtr>>`)
and the array treatment (`AggHashMapWithOneLowCardDictKey<SmallFixedSizeHashMap<uint8_t, AggDataPtr>>`)
side-by-side and drives both through the production callsite
`build_hash_map` with a production-shaped `HashTableKeyAllocator`
(`sizeof(int64_t)` state matching `COUNT(*)`).

Sections:
- **A**: density {1, 4, 24, 100, 200, 255} × distribution
  {random, sorted, clustered-run-length-64, Zipf s=1.5} at 100M rows.
  Density caps at 255 because `Config.low_cardinality_threshold = 255`
  is the FE guard ceiling.  Hash-only (`BM_LowCardBuildOnly`) and
  hash + COUNT(*) update (`BM_LowCardBuildAndCount`).
- **B**: construction-vs-amortization break-even at density=24, sorted,
  rows {4K, 100K, 1M, 10M, 100M}.  Array path's ~2 KiB upfront cost is
  trivial so break-even sits at very small row counts.
- **C**: construct-only regression guard (`BM_LowCardConstruct`).
- **D**: nullable coverage at null fractions {0%, 10%, 50%} exercising
  both the no-nulls fall-through and the per-row null-bit-aware loop.

#### Results (100M rows, median, x86_64 r6a.xlarge, ASLR on)

Hash-only:

| distinct | dist | phmap | array | speedup |
|----------|------|-------|-------|---------|
| 1 | random | 411 ms | 148 ms | **2.8×** |
| 24 | random | 539 | 147 | **3.7×** |
| 100 | random | 533 | 147 | **3.6×** |
| 200 | random | 541 | 147 | **3.7×** |
| 255 | random | 538 | 147 | **3.7×** |
| 24 | Zipf | 533 | 149 | **3.6×** |

Array path time is **constant ~148ms** across all densities (direct
array index, no hashing).  phmap scales with distinct count.

Build+count (hash insert + per-row counter increment):
**1.2–3.6×** depending on density.

#### Memory trade-off

Array always 2 KiB regardless of distinct; phmap up to 8.7 KiB at
distinct=255.  Array path is also memory-cheaper.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/bench/CMakeLists.txt
+++ b/be/src/bench/CMakeLists.txt
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 ADD_BE_BENCH(${SRC_DIR}/bench/chunks_sorter_bench)
+ADD_BE_BENCH(${SRC_DIR}/bench/agg_low_card_dict_bench)
 ADD_BE_BENCH(${SRC_DIR}/bench/runtime_filter_bench)
 ADD_BE_BENCH(${SRC_DIR}/bench/csv_reader_bench)
 ADD_BE_BENCH(${SRC_DIR}/bench/shuffle_chunk_bench)
--- a/be/src/bench/agg_low_card_dict_bench.cpp
+++ b/be/src/bench/agg_low_card_dict_bench.cpp
@@ -112,7 +112,10 @@ static std::vector<int> build_zipf_codes(int distinct_count) {
         int lo = 0, hi = distinct_count - 1;
         while (lo < hi) {
             int mid = (lo + hi) >> 1;
-            if (cum[mid] < r) lo = mid + 1; else hi = mid;
+            if (cum[mid] < r)
+                lo = mid + 1;
+            else
+                hi = mid;
         }
         codes[i] = lo;
     }
--- a/be/src/bench/agg_low_card_dict_bench.cpp
+++ b/be/src/bench/agg_low_card_dict_bench.cpp
@@ -22,15 +22,22 @@
 // parallel: low_cardinality_key8.
 //
 // Driver: production callsite build_hash_map with a production-shaped
-// HashTableKeyAllocator (sizeof(int64_t) state matching COUNT(*)).  Density
-// caps at 255 because dict codes only span [0, 255] under
-// Config.low_cardinality_threshold = 255; higher densities are blocked at
-// FE.  Construction is cheap (256-cell array = ~2 KiB) so Section C is a
-// regression guard, not a primary signal.
+// HashTableKeyAllocator (sizeof(int64_t) state matching COUNT(*)).
+//
+// Density and code-range semantics:
+//   * Config.low_cardinality_threshold caps dictionary size at <=255
+//     entries; the FE guard rejects marking a slot when threshold > 255.
+//     With threshold=255 a dictionary holds up to 255 entries giving
+//     codes in [0, 254].  density=255 in the bench produces exactly that
+//     worst-case shape (255 distinct codes 0..254 filling the array).
+//   * density=256 cannot occur in production with the current guard and
+//     is not benched.  If the guard is relaxed in the future, codes 0..255
+//     remain a single read of the array (cell 256 is the sentinel slot
+//     reserved by SmallFixedSizeHashMap, see fixed_hash_map.h).
 
-#include <base/testutil/assert.h>
 #include <benchmark/benchmark.h>
 
+#include <cmath>
 #include <memory>
 #include <random>
 #include <vector>
@@ -71,6 +78,7 @@ enum class Distribution : int {
     Random = 0,
     Sorted = 1,
     Clustered64 = 2,
+    Zipf = 3, // skewed: ~90% of mass on top decile (s=1.5)
 };
 
 // Production-shaped allocator -- same shape as the SMALLINT bench so the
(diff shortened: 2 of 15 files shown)


CelerData-Reviewer:
@codex review

stdpain:
After #61632, we no longer need this optimization. For group by operations with a cardinality of 255, we can use the minmax information to compress it into int8.

PR #73560, opened 2026-05-20
[Enhancement] Add adaptive consecutive-keys cache to GROUP BY and DISTINCT

### LINKED ISSUE #73590: [Enhancement] Cache consecutive identical GROUP BY / DISTINCT keys to skip hash probes
## Enhancement

When successive rows in a chunk share the same `GROUP BY` (or
`DISTINCT`) key — common after sorted scans, ORDER BY-resident input,
sorted-clustered storage, or pipeline pre-aggregation that delivers
in-order — the hash insert phase repeatedly hashes the same key
value and probes the same bucket, each time paying full hash +
compare cost.  The result is the same aggregate state (or the same
set-membership answer) we already resolved on the previous row.

### What we want

A per-chunk cache of the `(last_key, last_resolved_value)` pair in
each `AggHashMap*` / `AggHashSet*` wrapper.  When the next row's key
matches the cached one, replay the cached value and skip the hash
table probe entirely.

### Acceptance

- Workloads with consecutive-key runs are measurably faster than the
  baseline; workloads with random keys see no regression — i.e. the
  cache disables itself on its own once it observes a low hit rate,
  without operator intervention.
- Both the GROUP BY (`AggHashMap*`) and the DISTINCT / streaming
  preaggregation (`AggHashSet*`, including `compute_and_allocate=false`)
  paths are covered.  Probe-only mode must remain correct when the
  cache replays the answer for repeated absent keys.
- Two-level hash conversion (after the single-level spill threshold)
  preserves the cache's adaptive state so the optimization doesn't
  silently re-enable mid-query on a workload that already opted out.
- Cache state is visible in the profile (hits / misses) so a profile
  diff makes the routing decision auditable.
- Safe rollback path exists for ops bisection without a BE redeploy.

eshishki:
Fixes #73590.

## Why I'm doing:

When successive rows in a chunk share the same `GROUP BY` key — common
after sorted scans, ORDER BY-resident input, sorted-clustered storage,
or pipeline pre-aggregation that delivers in-order — the hash insert
phase repeatedly hashes the same key value and probes the same bucket,
each time paying full hash + compare cost.  The result is the same
`AggDataPtr` we already resolved the row before.

## What I'm doing:

Add an adaptive consecutive-keys cache that remembers the
`(last_key, last_state)` pair within a chunk.  If the next row matches
the cached key, return `last_state` directly and skip the hash map probe.
The cache observes its own hit-rate over the first `EVAL_CHUNKS` chunks
and auto-disables itself if the rate stays below `MIN_HIT_RATE` — so
unsorted workloads pay only one probe per chunk, then nothing.

Specializations cover the three production-relevant wrapper families:
- `AggHashMapWithOneNumberKey` (single-int / SMALLINT / TINYINT keys)
- `AggHashMapWithOneStringKey` (single-Slice key)
- `AggHashMapWithSerializedKey` (multi-column GROUP BY)

### Adaptive disable

`ConsecutiveKeyCacheConfig` exposes `EVAL_CHUNKS=3`, `MIN_HIT_RATE=0.3`,
`MIN_SAMPLES=1000`.  After the third chunk we check
`hits / (hits+misses) >= 0.3` over at least 1000 sampled rows; if not,
we set `cache_enabled = false` and stop touching cache state for the
rest of the query.  This is what guarantees zero overhead on random
workloads — see bench results below.

### Two-level hash conversion carry-over

`AggHashMapVariant::convert_to_two_level` builds a **fresh** destination
wrapper, so without carry-over the cache would re-enable itself after
spilling out of the single-level threshold and the profile counters
would reset mid-query.
`_carry_over_post_init_state` mirrors `cache_enabled`, `hits`, `misses`,
`chunks_processed` from the source wrapper onto the destination, so
disabling the session flag still holds after the conversion and the
profile reflects the full query lifetime.

### Session gate

Activated by `enable_agg_consecutive_keys_cache` (default true).
`ConsecutiveKeyCache::force_disable()` is called from
`_init_agg_hash_variant` when the session variable is off; the variant
visitor uses C++20 `requires`-expressions so wrappers that lack the
cache are silently skipped.

The session gate is for **diagnosis only**: the adaptive disable
already handles the workload-level rollback automatically.  The flag
lets ops bisect a regression against the cache without redeploying BE.

### Profile counters

`ConsecutiveKeysCacheHits` / `ConsecutiveKeysCacheMisses` are populated
by `Aggregator::update_hash_map_profile_counters()` at the same chunk
and finalize boundaries that already publish `HashTableSize`, so
sink/source operators don't need to know about the new counters
individually.

### Tests

`be/test/exec/agg_hash_map_test.cpp` adds:
- `ConsecutiveKeyCacheTest.NumericKeyCacheHit` — verifies hit/miss
  semantics on `1,1,2,2` → 2 hits, 2 misses.
- `ConsecutiveKeyCacheTest.DisableForSmallFixedSizeHashMap` — verifies
  that the cache is correctly skipped for direct-array wrappers
  (TINYINT / BOOL / future SMALLINT), where a probe is already O(1)
  array index.

### Bench

`be/src/bench/agg_consec_keys_cache_bench.cpp` drives
`AggHashMapWithOneStringKey<SliceAggHashMap>` and
`AggHashMapWithSerializedKey<SliceAggHashMap>` through `build_hash_map`
with `cache_off` (synthetic baseline — cache `force_disable()`) and
`cache_on` (default) combos.

Shapes:
- `UUID36_rl1` — 100% UUID36, no consecutive runs (worst case)
- `UUID36_rl64` — run-length 64
- `UUID36_rl1024` — run-length 1024
- `UUID36_mix90` — 90% UUID + 10% non-UUID, random codes
- `UUID36_rl64_mix90` — run-length 64 with 10% noise (dirty production)
- `ascii_random` — 20-byte ascii, random (adversarial)
- `UUID36_high` — 1M distinct UUID36 random (grown hash table, 10M rows)

#### Results (median, x86_64 r6a.xlarge, ASLR on)

Single-Slice wrapper, 30M rows, hash insert only:

| shape | cache_off | cache_on | speedup |
|-------|-----------|----------|---------|
| UUID36_rl1 (no runs) | 667 ms | 666 | **0** |
| UUID36_rl64 | 637 | 415 | **+54%** (hit rate 98%) |
| UUID36_rl1024 | 634 | 384 | **+65%** |
| UUID36_rl64_mix90 | 676 | 520 | **+30%** |
| UUID36_mix90 (random) | 730 | 733 | 0 |
| ascii_random | 527 | 528 | 0 |
| UUID36_high (1M random) | 1202 ms | 1211 | 0 |

Serialized 2-col wrapper, 10M rows:

| shape | cache_off | cache_on | speedup |
|-------|-----------|----------|---------|
| both_UUID36_rl64 | 545 ms | 389 | **+40%** |
| mixed_UUID_ascii (random) | 1505 | 1494 | 0 |
| both_ascii_random | 1371 | 1374 | 0 |
| col0_uuid_col1_random | 1489 | 1496 | 0 |

**Cache never regresses.**  On runs ≥64 it wins 30-65%; on random or
high-cardinality the adaptive disable kicks in after the first 3 chunks
and the per-row cost returns to baseline.

### Set-side port (DISTINCT / `count(distinct ...)`)

The same cache is wired into the five `AggHashSet*` wrappers:
`AggHashSetOfOneNumberKey`, `AggHashSetOfOneNullableNumberKey`,
`AggHashSetOfOneStringKey`, `AggHashSetOfOneNullableStringKey`,
`AggHashSetOfSerializedKey`.  Cached value is `bool last_found`
(1 byte) instead of `AggDataPtr`, so the probe-only path
(`compute_and_allocate=false`, used by the streaming preaggregation
first stage) can replay the absent/present answer for repeated
consecutive keys without touching the hash set.

The cache is shared with the Map side via the templated
`ConsecutiveKeyCache<K, V=AggDataPtr>` / `ConsecutiveSerializedKeyCache<V>`
in `be/src/exec/aggregate/agg_consec_keys_cache.h`; the Set
specialisations instantiate with `V=bool`.

Profile counters `ConsecutiveKeysCacheHits` / `ConsecutiveKeysCacheMisses`
are populated by `Aggregator::update_hash_set_profile_counters()` from
the three distinct sink/source operators (blocking sink, streaming
sink, streaming source).  Two-level conversion carry-over
(`_carry_over_post_init_state`) is mirrored from Map → Set in
`CONVERT_TO_TWO_LEVEL_SET`.

#### Tests

`be/test/exec/agg_hash_map_test.cpp` adds:
- `ConsecutiveKeySetCacheTest.NumericKeyCacheHitAllocate` — keys
  `1,1,2,2` over `AggHashSetOfOneNumberKey<TYPE_INT>` → 2 hits, 2 misses.
- `ConsecutiveKeySetCacheTest.ProbeOnlyAbsentKeysReplay` — pre-seed
  `{5}`, probe `[1,1,7,7]`, verify all `not_founds` set even with
  cache hits.  This is the correctness regression the cached-bool
  contract is designed to prevent.
- `ConsecutiveKeySetCacheTest.DisableForSmallFixedSizeHashSet` —
  TINYINT direct-array set bypasses the cache (probe is already O(1)).

#### Set-side bench results (median, x86_64 m6i.4xlarge, ASLR on)

Single-Slice DISTINCT, 30M rows, `build_hash_set`:

| shape | cache_off | cache_on | speedup |
|-------|-----------|----------|---------|
| UUID36_rl1 (no runs) | 672 ms | 671 | **0** |
| UUID36_rl64 | 637 | 405 | **+36%** (hit rate 98%) |
| UUID36_rl1024 | 633 | 378 | **+40%** |
| UUID36_rl64_mix90 | 680 | 474 | **+30%** |
| UUID36_mix90 (random) | 736 | 736 | 0 |
| ascii_random | 524 | 520 | 0 |
| UUID36_high (1M random, 10M rows) | 1106 | 1114 | 0 |

Probe-only (`build_hash_set_with_selection`, set pre-seeded with half
the key universe), 30M rows:

| shape | cache_off | cache_on | speedup |
|-------|-----------|----------|---------|
| UUID36_rl64 | 534 ms | 410 | **+23%** |
| UUID36_rl1024 | 467 | 378 | **+19%** |
| UUID36_rl64_mix90 | 614 | 523 | **+15%** |
| UUID36_rl1 | 564 | 561 | 0 |
| UUID36_mix90 | 1083 | 1085 | 0 |
| ascii_random | 942 | 933 | 0 |
| UUID36_high (1M) | 453 | 437 | 0 |

Serialized 2-col DISTINCT, 10M rows:

| shape | cache_off | cache_on | speedup |
|-------|-----------|----------|---------|
| both_UUID36_rl64 | 525 ms | 350 | **+33%** |
| mixed_UUID_ascii (random) | 1352 | 1330 | 0 |
| both_ascii_random | 1251 | 1252 | 0 |
| col0_uuid_col1_random | 1340 | 1352 | 0 |

Same adaptive-disable behavior as the Map side: zero regression on
random workloads, 30-40% wins on consecutive-runs shapes.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

eshishki:
@stdpain quick docs question before I push the missing system-variable + profile-counter docs (Codex P1 finding). For `Introduced in:` on a new `enable_agg_*` session variable targeted at `main` — should I use `v4.1`, or do you want me to leave the field off until release branching settles?

stdpain:
We can fill this variable during the release phase.

eshishki:
@stdpain re: `enable_agg_consecutive_keys_cache` — quick check before I add the docs.

The bench table in the PR body shows no measurable regression on the adversarial random shape (`UUID36_rl1`: 667 ms cache_off vs 666 ms cache_on) — adaptive-disable handles that path automatically. I added the session gate mainly to A/B in benches during review.

Two options:
1. **Keep the gate** — I'll add `docs/en/` + `docs/zh/` entries for `enable_agg_consecutive_keys_cache`.
2. **Drop the gate** — no flag, no thrift field, no docs for it. Adaptive-disable is the only safety net.

Which way do you want to go?

PR #73567, opened 2026-05-20
[BugFix] Surface AccessDenied reason for sys.fe_memory_usage / sys.fe_locks

### LINKED ISSUE #73547: [BugFix] sys.fe_memory_usage / sys.fe_locks return misleading 'BE:{id}, host: unknown' when caller lacks OPERATE privilege

On any shared-data cluster (FE + CN, no BE), as a user **without** the `OPERATE ON SYSTEM` privilege (i.e. anyone whose active roles don't include `db_admin` / `security_admin` / `root`):

```sql
-- Both fail the same way
SELECT * FROM sys.fe_memory_usage;
SELECT * FROM sys.fe_locks;
```

The client should receive an `Access denied` error that names the missing privilege, the same way every other `OPERATE`-gated statement does (e.g. `ADMIN SHOW CONFIG` returns `Access denied; you need (at least one of) the GRANT privilege(s) on SYSTEM ...`).

### Real behavior (Required)

The client gets a misleading "node lookup" style error that has nothing to do with the actual cause:

```
ERROR 1064 (HY000): FE RPC failure,
  address=TNetworkAddress(hostname=<fe-host>, port=9020),
  reason=Internal error processing listFeMemoryUsage: BE:10026, host: unknown
```

```
ERROR 1064 (HY000): FE RPC failure,
  address=TNetworkAddress(hostname=<fe-host>, port=9020),
  reason=Internal error processing listFeLocks: BE:10030, host: unknown
```

The `BE:<id>` segment looks like a backend-id lookup failure, and `host: unknown` reinforces that misreading. **Neither is true.** `BE:<id>` is just the CN's own id stamped on a failed pipeline-driver status, and `host: unknown` is a hard-coded literal in `DefaultCoordinator`. The actual cause is an empty-message `AccessDeniedException` that gets swallowed by three layers of error wrapping.

### Server-side stack trace (from `fe.log` of the FE that served the call)

```
ERROR  Internal error processing listFeMemoryUsage
org.apache.thrift.TException
    at com.starrocks.catalog.system.sys.SysFeMemoryUsage.listFeMemoryUsage(SysFeMemoryUsage.java:67)
    at com.starrocks.service.FrontendServiceImpl.listFeMemoryUsage(FrontendServiceImpl.java:662)
    at com.starrocks.thrift.FrontendService$Processor$listFeMemoryUsage.getResult(FrontendService.java:7041)
    at com.starrocks.thrift.FrontendService$Processor$listFeMemoryUsage.getResult(FrontendService.java:7018)
    at org.apache.thrift.ProcessFunction.process(ProcessFunction.java:40)
    at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:40)
    at com.starrocks.common.SRTThreadPoolServer$WorkerProcess.run(SRTThreadPoolServer.java:313)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
    at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: com.starrocks.authorization.AccessDeniedException
    at com.starrocks.authorization.NativeAccessController.checkObjectTypeAction(NativeAccessController.java:321)
    at com.starrocks.authorization.NativeAccessController.checkSystemAction(NativeAccessController.java:44)
    at com.starrocks.sql.analyzer.Authorizer.checkSystemAction(Authorizer.java:66)
    at com.starrocks.catalog.system.sys.SysFeMemoryUsage.listFeMemoryUsage(SysFeMemoryUsage.java:65)
    ... 9 more
```

The coordinator wrap on the same FE:

```
com.starrocks.rpc.RpcException: FE RPC failure,
  address=TNetworkAddress(hostname=<fe-host>, port=9020),
  reason=Internal error processing listFeMemoryUsage: BE:10001, host: unknown
    at com.starrocks.qe.DefaultCoordinator.dealStatusToTryRetry(DefaultCoordinator.java:938)
    at com.starrocks.qe.DefaultCoordinator.getNext(DefaultCoordinator.java:1011)
    at com.starrocks.qe.StmtExecutor.handleQueryStmt(StmtExecutor.java:1686)
    at com.starrocks.qe.StmtExecutor.execute(StmtExecutor.java:879)
    ...
```

### Root cause

Three layers each drop information:

1. **`fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java:321`** (and every other throw site in the same file) does `throw new AccessDeniedException();` — no message.
2. **`fe/fe-core/src/main/java/com/starrocks/catalog/system/sys/SysFeMemoryUsage.java:58-60`** catches the empty-message exception and rethrows as `TException(e.getMessage(), e)` — `e.getMessage()` is `null`, so the rethrown TException carries no information either. Same anti-pattern in `SysFeLocks.java:86-88`.
3. Apache Thrift then wraps the empty-message TException as the generic `TApplicationException("Internal error processing listFeMemoryUsage")`. The BE's `thrift_rpc_helper.cpp:86` adds the `FE RPC failure, address=..., reason=` prefix, `pipeline_driver_executor.cpp:193` appends `: BE:{be_id}`, and finally `DefaultCoordinator.java:947` wraps with `RpcException("unknown", status.getErrorMsg())` — that literal `"unknown"` is what produces `, host: unknown` at the very end.

So the user-facing error is essentially `[decorative-be-tag]:[decorative-host-tag]` with the real reason completely gone.

### Why this matters

- `sys.fe_memory_usage` and `sys.fe_locks` are the two main observability tables for FE memory and lock contention. They are exactly the tables you want to query during an incident, when you want clear errors, not a maze of misleading wrappers.
- Operators reading this error reasonably suspect a shared-data CN/BE node-lookup bug, lock-manager corruption, or RPC routing problem. None of those are happening — but ruling them out takes hours of code spelunking.
- The same anti-pattern (`catch (AccessDeniedException e) { throw new TException(e.getMessage(), e); }`) appears in several other system-table handlers (e.g. `SysObjectDependencies`, `FeThreadsSystemTable`), so this same opaque error surfaces from those tables too.

### Suggested direction (for the team to evaluate)

The fix likely belongs in the `catch (AccessDeniedException e)` block in `SysFeMemoryUsage.java:58-60` and `SysFeLocks.java:86-88` — it should surface a meaningful "access denied" message instead of the empty `e.getMessage()`. The deeper fix is to give `new AccessDeniedException()` a default message at every throw site in `NativeAccessController.java`, since the same null-message swallow can leak through any handler that uses the same `catch`/rethrow pattern. Happy to leave the exact shape to the maintainers.

### Workaround (for affected users until a patched release)

Grant `OPERATE ON SYSTEM` — either narrowly to the specific user, or to a role they hold:

```sql
GRANT OPERATE ON SYSTEM TO USER '<user>'@'<host>';
-- or
GRANT db_admin TO ROLE <your-observability-role>;
```

Note: `OPERATE ON SYSTEM` is a broad admin privilege that also allows `ADMIN SET CONFIG`, `ADMIN SET REPLICA STATUS`, `EXECUTE SCRIPT`, etc. There is no fine-grained "read-only system tables" privilege today. Splitting one out (e.g. `INSPECT`) so that `sys.fe_*` can be queried without granting cluster-write powers would be a useful follow-up.

```
4.0.10-5b36164
```

(Shared-data deployment, 3 FE + 30 CN.)

yakirgb:
## Why I'm doing:

Fixes #73547.

Querying `sys.fe_memory_usage` / `sys.fe_locks` as a user without `OPERATE ON SYSTEM` produces a misleading "node lookup" style error:

```
FE RPC failure, address=TNetworkAddress(hostname=<fe-host>, port=9020),
  reason=Internal error processing listFeMemoryUsage: BE:10026, host: unknown
```

The actual cause is an `AccessDeniedException`, but its message is dropped on the way out. Three layers conspire:

1. `NativeAccessController.java:321` throws `new AccessDeniedException()` with no message.
2. `SysFeMemoryUsage.java:58-60` and `SysFeLocks.java:86-88` catch it and rethrow as `new TException(e.getMessage(), e)` — `e.getMessage()` is null.
3. Apache Thrift wraps the empty-message TException, the BE thrift_rpc_helper adds the `FE RPC failure ... reason=` prefix, the CN pipeline driver appends `: BE:{be_id}`, and `DefaultCoordinator.dealStatusToTryRetry` finally wraps with `RpcException("unknown", status.getErrorMsg())` — the literal `"unknown"` is what produces `, host: unknown`.

The user-facing error ends up as `<decorative-be-tag>:<decorative-host-tag>` with the real reason gone.

## What I'm doing:

In both `SysFeMemoryUsage.listFeMemoryUsage` and `SysFeLocks.listLocks`, replace `e.getMessage()` (null) with a properly formatted message produced via `ErrorCode.ERR_ACCESS_DENIED_FOR_EXTERNAL_ACCESS_CONTROLLER.formatErrorMsg(...)`, so the phrasing matches every other privilege check in the codebase:

```
Access denied; you need (at least one of) the OPERATE privilege(s) on SYSTEM for this operation.
```

Two production files changed, ~3 lines each. No semantics change — same exception type, same control flow, only a clearer message. Existing tests still pass; two new tests cover the access-denied path with `MockUp<Authorizer>` (the same mock pattern used in `TaskRunsSystemTableTest`).

The broader anti-pattern (`new AccessDeniedException()` everywhere in `NativeAccessController`) is intentionally left for a follow-up — this PR keeps the diff minimal and reviewable.

Fixes #73547

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [x] Miscellaneous: error message phrasing only (no API/SQL/config surface change)

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/system/sys/SysFeLocks.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/system/sys/SysFeLocks.java
@@ -19,10 +19,12 @@ import com.google.common.base.Joiner;
 import com.google.gson.JsonObject;
 import com.starrocks.authentication.UserIdentityUtils;
 import com.starrocks.authorization.AccessDeniedException;
+import com.starrocks.authorization.ObjectType;
 import com.starrocks.authorization.PrivilegeType;
 import com.starrocks.catalog.Table;
 import com.starrocks.catalog.system.SystemId;
 import com.starrocks.catalog.system.SystemTable;
+import com.starrocks.common.ErrorCode;
 import com.starrocks.common.util.concurrent.lock.LockHolder;
 import com.starrocks.common.util.concurrent.lock.LockInfo;
 import com.starrocks.common.util.concurrent.lock.LockManager;
@@ -75,7 +77,8 @@ public class SysFeLocks {
                 Authorizer.checkSystemAction(context, PrivilegeType.OPERATE);
             }
         } catch (AccessDeniedException e) {
-            throw new TException(e.getMessage(), e);
+            throw new TException(ErrorCode.ERR_ACCESS_DENIED_FOR_EXTERNAL_ACCESS_CONTROLLER.formatErrorMsg(
+                    PrivilegeType.OPERATE.name(), ObjectType.SYSTEM.name(), ""), e);
         }
 
         TFeLocksRes response = new TFeLocksRes();
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/system/sys/SysFeLocks.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/system/sys/SysFeLocks.java
@@ -21,10 +21,10 @@ import com.starrocks.authentication.UserIdentityUtils;
 import com.starrocks.authorization.AccessDeniedException;
 import com.starrocks.authorization.ObjectType;
 import com.starrocks.authorization.PrivilegeType;
+import com.starrocks.catalog.InternalCatalog;
 import com.starrocks.catalog.Table;
 import com.starrocks.catalog.system.SystemId;
 import com.starrocks.catalog.system.SystemTable;
-import com.starrocks.common.ErrorCode;
 import com.starrocks.common.util.concurrent.lock.LockHolder;
 import com.starrocks.common.util.concurrent.lock.LockInfo;
 import com.starrocks.common.util.concurrent.lock.LockManager;
@@ -77,8 +77,9 @@ public class SysFeLocks {
                 Authorizer.checkSystemAction(context, PrivilegeType.OPERATE);
             }
         } catch (AccessDeniedException e) {
-            throw new TException(ErrorCode.ERR_ACCESS_DENIED_FOR_EXTERNAL_ACCESS_CONTROLLER.formatErrorMsg(
-                    PrivilegeType.OPERATE.name(), ObjectType.SYSTEM.name(), ""), e);
+            AccessDeniedException.reportAccessDenied(InternalCatalog.DEFAULT_INTERNAL_CATALOG_NAME,
+                    context.getCurrentUserIdentity(), context.getCurrentRoleIds(),
+                    PrivilegeType.OPERATE.name(), ObjectType.SYSTEM.name(), null);
         }
 
         TFeLocksRes response = new TFeLocksRes();
--- a/fe/fe-core/src/test/java/com/starrocks/catalog/system/sys/SysFeLocksTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/catalog/system/sys/SysFeLocksTest.java
@@ -60,11 +60,11 @@ public class SysFeLocksTest {
             }
         };
 
-        TException ex = assertThrows(TException.class, () -> SysFeLocks.listLocks(req, true));
+        Exception ex = assertThrows(Exception.class, () -> SysFeLocks.listLocks(req, true));
         assertNotNull(ex.getMessage(), "AccessDenied message should not be null");
-        assertTrue(ex.getMessage().contains("Access denied"), "Should match canonical format: " + ex.getMessage());
-        assertTrue(ex.getMessage().contains("OPERATE"), "Should mention OPERATE: " + ex.getMessage());
-        assertTrue(ex.getMessage().contains("SYSTEM"), "Should mention SYSTEM: " + ex.getMessage());
+        assertTrue(ex.getMessage().contains(
+                        "Access denied; you need (at least one of) the OPERATE privilege(s) on SYSTEM for this operation."),
+                "Should match canonical format: " + ex.getMessage());
     }
 
 }
--- a/fe/fe-core/src/test/java/com/starrocks/catalog/system/sys/SysFeLocksTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/catalog/system/sys/SysFeLocksTest.java
@@ -14,12 +14,20 @@
 
 package com.starrocks.catalog.system.sys;
 
+import com.starrocks.authorization.AccessDeniedException;
+import com.starrocks.authorization.PrivilegeType;
+import com.starrocks.qe.ConnectContext;
+import com.starrocks.sql.analyzer.Authorizer;
 import com.starrocks.thrift.TAuthInfo;
 import com.starrocks.thrift.TFeLocksReq;
+import mockit.Mock;
+import mockit.MockUp;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.thrift.TException;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class SysFeLocksTest {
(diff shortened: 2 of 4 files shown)


CelerData-Reviewer:
@codex review

HangyuanLiu (dismissed):

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/catalog/system/sys/SysFeLocks.java
```diff
@@ -75,7 +77,8 @@ public static TFeLocksRes listLocks(TFeLocksReq request, boolean authenticate) t
                 Authorizer.checkSystemAction(context, PrivilegeType.OPERATE);
             }
         } catch (AccessDeniedException e) {
-            throw new TException(e.getMessage(), e);
+            throw new TException(ErrorCode.ERR_ACCESS_DENIED_FOR_EXTERNAL_ACCESS_CONTROLLER.formatErrorMsg(
```
jaogoy: Why use `ERR_ACCESS_DENIED_FOR_EXTERNAL_ACCESS_CONTROLLER` instead of `ERR_ACCESS_DENIED`? Is it related to external access only?

fe/fe-core/src/test/java/com/starrocks/catalog/system/sys/SysFeLocksTest.java
```diff
@@ -36,4 +44,27 @@ public void testListLocks() throws TException {
...
+        new MockUp<Authorizer>() {
+            @Mock
+            public void checkSystemAction(ConnectContext context, PrivilegeType privilegeType)
+                    throws AccessDeniedException {
+                throw new AccessDeniedException();
+            }
+        };
+
+        TException ex = assertThrows(TException.class, () -> SysFeLocks.listLocks(req, true));
+        assertNotNull(ex.getMessage(), "AccessDenied message should not be null");
```
jaogoy, line 64: It could be `Access denied; you need (at least one of) the %s privilege(s)  on %s%s for this operation.` style easily.
We won't chage the error message frequently.

jaogoy (commented):
It's better you can use AI to check more other similar problematic code and fix them.

fe/fe-core/src/main/java/com/starrocks/catalog/system/sys/SysFeLocks.java
yakirgb: You're right — `ERR_ACCESS_DENIED_FOR_EXTERNAL_ACCESS_CONTROLLER` is intended only for external controllers (Ranger/OPA); the default native path should go through `ERR_ACCESS_DENIED` so the role-activation hint is included.

Switched both `SysFeLocks.listLocks` and `SysFeMemoryUsage.listFeMemoryUsage` to call `AccessDeniedException.reportAccessDenied(...)`, matching the precedent in `FeThreadsSystemTable.generateFeThreadsResponse` (`fe/fe-core/src/main/java/com/starrocks/catalog/system/information/FeThreadsSystemTable.java:71-81`). The helper picks the correct variant based on whether the active controller is native or external, so we no longer hard-pick one.

Pushed in 3e59b543b6. Built and ran both test classes in `starrocks/dev-env-ubuntu` (Maven 3.6.3 / JDK 17): 4 tests, 0 failures.

fe/fe-core/src/test/java/com/starrocks/catalog/system/sys/SysFeLocksTest.java
yakirgb: Done — collapsed the three `contains(...)` checks in both `SysFeLocksTest` and `SysFeMemoryUsageTest` into a single assertion against the canonical sentence:

```java
assertTrue(ex.getMessage().contains(
        "Access denied; you need (at least one of) the OPERATE privilege(s) on SYSTEM for this operation."),
        "Should match canonical format: " + ex.getMessage());
```

That substring is the leading sentence shared by both error codes (5203 native and 5204 external), so the assertion stays valid regardless of which branch the routing helper from the other thread picks at runtime. Also widened the caught type to `Exception` since `reportAccessDenied` throws `ErrorReportException` (a `RuntimeException`) rather than the previous `TException` wrap.

Pushed in 3e59b543b6.

kevincai:
@mergifyio rebase

CelerData-Reviewer:
@codex review

kevincai (approved):

jaogoy (approved):

PR #73584, opened 2026-05-20
[Enhancement] Enable global lazy-materialization for HNSW vector queries

sevev:
## Why I'm doing:

HNSW vector queries today force the embedding column to be eagerly materialized at scan time, even when the query never reads it. For a typical `SELECT id, distance ORDER BY approx_l2_distance(v, [...]) LIMIT 100` against a 1M-row × 1536-dim table, that's ~6 GB of unused column data read off storage and through `_dict_chunk` — every query. The 1536-dim Cohere benchmark spends ~50% of scan-side CPU here on warm cache.

The cause is a vector-specific carve-out in the global lazy-materialize support: `OlapScanLazyMaterializationSupport.predicateUsedColumns()` unconditionally unions the vector-index column into the predicate-used set whenever `isEnableUseANN()` is true. That defeats the global lazy-materialize rule, which would otherwise prune the column from the scan or defer it behind the final TopN. The carve-out was put there for a reason — IVFPQ's TopN evaluates `approx_*_distance(v, [...])` row-by-row against the raw embedding, so v must reach the TopN — but it's overly broad: HNSW had its order-by rewritten to reference the BE-produced virtual distance column (`__vector_*`), so v is unused downstream.

## What I'm doing:

Narrow the carve-out in `OlapScanLazyMaterializationSupport.predicateUsedColumns` from `isEnableUseANN()` to `isUseIVFPQ()`. HNSW queries now flow through the existing global lazy-mat rule: if the projection doesn't reference v, it's pruned from the scan output entirely; if it does, it's deferred to a `FetchNode` that materializes only the K survivors after the final TopN. IVFPQ keeps the original eager path because the rewrite doesn't replace its order-by expression. Brute-force fallback (`.vi` missing) still works — BE's `_setup_brute_force_fallback` adds v back into the BE-internal `_schema` so distance can be recomputed from `_dict_chunk`; v is consumed inside the scan and never propagated up.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lazymaterialize/OlapScanLazyMaterializationSupport.java
```diff
@@ -62,7 +62,25 @@ public ColumnRefSet predicateUsedColumns(PhysicalScanOperator scanOperator) {
...
+        // satisfy in line. We let the global lazy-materialize rule decide:
+        //   * SELECT does not reference v  → v is pruned from scan output entirely
+        //     (zero embedding I/O).
+        //   * SELECT references v  → v is deferred and fetched by row id at the
+        //     FetchNode above the final TopN (only K rows hit storage).
+        //
+        // When the .vi file is missing at runtime, BE's _setup_brute_force_fallback
+        // appends v to the BE-internal _schema so distance can be recomputed from
+        // _dict_chunk; v is consumed inside the scan and never propagated up.
+        if (op.getVectorSearchOptions().isUseIVFPQ()) {
```
stdpain, line 83: What's the difference between isEnableUseANN and isUseIVFPQ?

sevev: > What's the difference between isEnableUseANN and isUseIVFPQ?

 HNSW outputs the final distance from the index; nothing downstream needs v, so lazy-mat is safe.
 IVFPQ outputs an approximate distance and a candidate set; So `TopN` re-rank using raw v, so v must stay eager on the scan output.

luohaha (approved):

sevev:
@stephen-shelby PTAL

stdpain (approved):

kevincai (approved):

PR #73585, opened 2026-05-20
[Enhancement] FE: support prepared-statement vector search via CAST(? AS ARRAY<FLOAT>)

sevev:
## Why I'm doing:

Vector search clients that want to reuse a query plan for many different query vectors typically use a prepared statement of the shape:

```sql
PREPARE p FROM 'SELECT id FROM t ORDER BY approx_l2_distance(v, CAST(? AS ARRAY<FLOAT>)) LIMIT 10';
EXECUTE p USING '[0.13, 0.42, ..., 0.98]';
```

Today the optimizer's vector-index rewrite rule doesn't recognize this form. `RewriteToVectorPlanRule.isConstantArrayFloat()` only accepts a literal array (`[1.1, 2.2, ...]`) or a numeric-typed `CastOperator` whose child is an `ArrayOperator`. The prepared-statement path produces a `CastOperator(ConstantOperator(VARCHAR, "[..]"), ARRAY<FLOAT>)` after analysis — a different shape — so the rule bails out and the query falls back to the non-ANN scan path, paying the full predicate scan + sort instead of the HNSW lookup.

On 1M Cohere 1536-dim @ conc=100, the rule-bailout cost shows up as a 5–6% QPS gap between native-literal and prepared-statement clients running the same query semantics.

## What I'm doing:

Teach `RewriteToVectorPlanRule` a third constant-array form: `CAST(StringLiteral AS ARRAY<FLOAT>)`. Two pieces:

- `isConstantArrayFloat()` now accepts `CastOperator → ConstantOperator(VARCHAR/CHAR)` in addition to the existing literal-array forms.
- `extractValuesFromConstantArray()` dispatches a hand-rolled string-to-float-list parser when it sees the cast-string shape — parses `"[f1, f2, ..., fN]"` into N tokens, validates each via `Double.parseDouble`, and appends them to the query-vector list.

The parser is deliberately narrow: it expects `[..]` brackets, comma-separated tokens, optional whitespace around tokens. Malformed input (missing brackets, empty token, non-numeric token, null) surfaces as `SemanticException` with a documented message — same error class as the existing dimension-mismatch path, so callers see consistent failure semantics regardless of which constant-array shape they used.

After this PR, the planner shape for a prepared-statement query and a native-literal query is identical — both go through the same VECTORINDEX rewrite, both emit the same TopN over `__vector_*` distance column. The downstream BE path is unchanged.

Seven new tests on `planner.VectorIndexTest`:

- `testPreparedStatementCastStringArrayHnsw` / `WithWhitespace` / `Ivfpq` — happy paths for HNSW + IVFPQ + whitespace handling.
- `DimMismatch` / `MissingBrackets` / `InvalidFloat` / `EmptyElement` — malformed-input → `SemanticException`.

Verified locally:

```
./run-fe-ut.sh --run --test "*planner.VectorIndexTest"
# 17/17 pass (10 existing + 7 new prepared-statement tests)
```

Fixes none — this is a feature for prepared-statement-driven vector search clients.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

stephen-shelby (approved):

HangyuanLiu (approved):

PR #73146, opened 2026-05-12
[Enhancement] Widen IVM aggregate-function whitelist (MIN/MAX VARCHAR, AVG DECIMAL, ARRAY_AGG)

Youngwb:
## Why I'm doing:

The current IVM aggregate-function whitelist in `IVMAnalyzer.IVM_SUPPORTED_AGG_FUNCTIONS` (added by #73014) is intentionally conservative — it rejects (function, argument-type) combinations whose CREATE / refresh path was not yet end-to-end-verified. Several of those gates have since been unblocked by follow-up bug-fix PRs:

- **`MIN/MAX(VARCHAR)`** was blocked by a strict type-equality precondition in `IvmOpUtils.buildStateUnionScalarOperator` that rejected `VARCHAR(N)` vs `VARCHAR(65533)` (the catalog-normalized MV column type via `MaterializedViewAnalyzer.transformTableColumnType`). #73095 relaxed that precondition to accept compatible string types.
- **`AVG(DECIMAL)`** was producing wrong values because the typed `AggStateDesc` was being dropped during state_union scalar construction. #73012 preserves it.
- **`ARRAY_AGG(single arg)`** already works end-to-end through the combinator path. `ARRAY_AGG(expr ORDER BY key)` must remain rejected because `FunctionAnalyzer.getAdjustedAnalyzedAggregateFunction` inlines order-by columns as extra `FunctionCallExpr` children, the combinator only accepts one `ANY_ELEMENT` parameter, and IVM `state_union` is associative/unordered (preserving `ORDER BY` semantics through incremental refresh is impossible).

`HLL_UNION` and `BITMAP_UNION` are deliberately **NOT** added: their underlying combinator path triggers a `Column.type` singleton mutation bug. `BitmapType.BITMAP` / `HLLType.HLL` are shared JVM singletons (`TypeFactory.PRIMITIVE_TYPE_SCALAR_TYPE_MAP`); creating an `AGG_STATE` column of these types lands in `Column` ctor calling `setAggStateDesc(...)` on the singleton in-place, poisoning every subsequent raw BITMAP/HLL declaration in the JVM until restart. Fix is tracked separately.

## What I'm doing:

In `fe/fe-core/src/main/java/com/starrocks/sql/analyzer/mv/IVMAnalyzer.java`:
- Extend `MIN` / `MAX` predicates with `|| args[0].isStringType()`
- Extend `AVG` predicate with `|| args[0].isDecimalV3()`
[...]

PR #73095, opened 2026-05-11
[BugFix] Allow state_union to accept compatible string types in IVM

Youngwb:
## Why I'm doing:

IVM refresh over an Iceberg MV containing `MIN(varchar_col)` or `MAX(varchar_col)` fails during planning with a `Preconditions.checkArgument` violation in `IvmOpUtils.buildStateUnionScalarOperator`. The delta-side intermediate aggregate carries the original `VARCHAR(N)` type, while the MV-side AGG_STATE column ref carries `VARCHAR(65533)` — `MaterializedViewAnalyzer.genMaterializedViewColumns` runs every persisted column type through `AnalyzerUtils.transformTableColumnType`, which caps VARCHAR at `OlapMaxVarcharLength = 65533`. The strict type-equality precondition then rejects an otherwise compatible pair.

## What I'm doing:

Replace the strict `equals` precondition in `IvmOpUtils.buildStateUnionScalarOperator` with a new `typesCompatibleForStateUnion` helper that additionally accepts two STRING types sharing the same primitive kind. Unblocks MIN/MAX over VARCHAR/CHAR columns without weakening type safety for non-string paths.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

PR #73627, opened 2026-05-21
[BugFix] Recurse into nested types in IvmOpUtils state_union compat check

Youngwb:
## Why I'm doing this:

`IvmOpUtils.typesCompatibleForStateUnion` only handled flat string types. When an IVM materialized view uses `ARRAY_AGG`, its intermediate state is `AGG_STATE<struct<col1 array<varchar(N)>>>`; the delta side carries the function-signature `VARCHAR(1073741824)` while the MV side persists with `VARCHAR(OlapMaxVarcharLength)`. Because both sides nest the string under struct + array, the top-level `isStringType()` branch never matched and CREATE failed:

```
ERROR 1064 (HY000): Getting analyzing error. Detail message: Failed to generate
IVM refresh plan at CREATE time: intermediateAggScalarOp type
struct<`col1` array<varchar(1073741824)>> is incompatible with MV state
column type struct<`col1` array<varchar(1048576)>>.
```

How the regression slipped in:

1. #73095 introduced `typesCompatibleForStateUnion` for flat strings.
2. #73146 widened the IVM whitelist to include ARRAY_AGG and added `test_ivm_whitelist_widening`.
3. **One day later** #73075 made the type-compat check fire at CREATE time via `IvmTrialRewriter`.
4. The test was tagged `@slow`, and CI's SQL-Tester runs with `[attr]: !slow,!system,!celerdata,!cloud` — so the case **silently never ran** on main, and the regression sat undetected.

Verified on the StarRocks dev cluster (`8.142.108.134`): `main` HEAD + branch-4.1.1 HEAD both report the same error; `mv_array_agg` CREATE succeeds after this fix, and the full `test_ivm_whitelist_widening` passes 3/3 flaky runs.

## What I'm doing:

- `IvmOpUtils.typesCompatibleForStateUnion`: recurse into `ARRAY` / `MAP` / `STRUCT` so the existing "same primitive kind, length may differ" rule applies to nested string types.
- `IvmOpUtilsTest`: add five new cases (array of varchar; array element primitive mismatch; struct of array of varchar — the exact ARRAY_AGG shape; struct different field count; map of varchar key+value).
- `test_ivm_whitelist_widening`: drop the `@slow` tag (no reason for this case to be gated — it's small and finishes in ~20s on dev).
- `test_ivm_whitelist_widening` (R file): three negative-case error messages had an extra outer `"Failed to rewrite the query for IVM: "` wrapping from before #73075 changed how the trial-rewrite path surfaces exceptions; update them to match what the FE actually emits now.

## Fixes

Surfaced on branch-4.1.1 first; same root cause applies to `main` and `branch-4.1`. Should be backported to `branch-4.1` / `branch-4.1.1`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

(Behavior change is "ARRAY_AGG MV creation now succeeds where it previously failed at CREATE-time" — a bug being fixed rather than a user-visible new behavior.)

## Checklist:

- [ ] This is a backport PR

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/ivm/common/IvmOpUtils.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/ivm/common/IvmOpUtils.java
@@ -33,7 +33,11 @@ import com.starrocks.sql.optimizer.operator.scalar.ColumnRefOperator;
 import com.starrocks.sql.optimizer.operator.scalar.ConstantOperator;
 import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
 import com.starrocks.type.AggStateDesc;
+import com.starrocks.type.ArrayType;
+import com.starrocks.type.MapType;
 import com.starrocks.type.ScalarType;
+import com.starrocks.type.StructField;
+import com.starrocks.type.StructType;
 import com.starrocks.type.Type;
 import com.starrocks.type.VarbinaryType;
 import com.starrocks.type.VarcharType;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

Astralidea (approved):

Youngwb:
@mergify backport branch-4.1.1

Issue #73528, opened 2026-05-20
[Enhancement] percentile_approx compression policy: default = max, out-of-range silently replaced, storage/runtime mismatch


eshishki:
## Summary

`percentile_approx` / `percentile_approx_weighted` compression policy has three issues that share the same root — the `[2048, 10000]` range and the 10000 default were inherited from Apache Doris in 2019 and never revisited. They're cheaply fixable together; filing as one issue because the proposal addresses all three.

### 1. Default equals the maximum (10000)

Every user that doesn't pass the third argument runs at the most expensive configuration. The Doris benchmark that originally chose 10000 (see history below) actually shows compression=1000 is indistinguishable in accuracy from 10000 at ~17% lower CPU; the extra cost buys precision the user didn't ask for.

Industry T-Digest defaults for context:

| Project | Default compression |
|---|---|
| T-Digest paper (Ted Dunning) | 100 |
| Apache DataSketches | k = 200 |
| Apache Druid | 100 |
| Presto (recommended start) | 500 |
| Apache Doris / StarRocks | **10000** |

### 2. Out-of-range silently replaced by `DEFAULT_COMPRESSION_FACTOR`, not clamped

`be/src/exprs/agg/percentile_approx.h:108-111` and `185-188`:

```cpp
if (compression < MIN_COMPRESSION || compression > MAX_COMPRESSION) {
    LOG(WARNING) << "compression factor out of range, use default: "
                 << DEFAULT_COMPRESSION_FACTOR;
    compression = DEFAULT_COMPRESSION_FACTOR;  // = MAX
}
```

Combined with problem 1, a user trying to pick a *cheaper* profile (`percentile_approx(v, 0.5, 100)`) silently gets the **most expensive** one (10000). The arithmetic-natural behavior would be to clamp to the nearest bound.

### 3. Storage path uses `compression = 1000`, which the runtime rejects

The on-disk `PERCENTILE` type is constructed via the default `TDigest()` ctor, which hardcodes compression = 1000:

- `be/src/types/tdigest.h:127` — `TDigest() : TDigest(1000) {}`
- `be/src/exprs/percentile_functions.cpp:27-45` — `percentile_hash`
- `be/src/storage/convert_helper.cpp:1049-1071` — `PercentileTypeConverter::convert_materialized`

[...]

PR #73634, opened 2026-05-21
[Enhancement] Respect percentile compression in MV rewrite

### LINKED ISSUE #73636: [Bug] percentile_approx MV rewrite silently downgrades precision to TDigest(1000)
## Problem

`percentile_approx(v, q, c)` builds a T-digest sized by `c` (precision vs
memory). A query asking for `compression=10000` that gets rewritten to read
from a percentile MV is silently answered at `compression=1000`.

User-visible consequences:
- Quantiles returned at lower precision than the query requested, with no
  warning, no message, and no trace entry indicating that a precision drop
  happened.
- No way to author an MV that stores percentile digests at a higher precision
  — every MV ends up at `compression=1000` regardless of how the populating
  query was written.
- No way to tell the optimizer "don't use a percentile MV if it would lose
  precision".

Compression is a semantic contract (precision of result), not a physical hint,
so the optimizer should not break it silently.

## Steps to reproduce

1. Create a materialized view on a base table `t` that aggregates percentiles:

   ```sql
   CREATE MATERIALIZED VIEW mv
   DISTRIBUTED BY HASH(k)
   REFRESH MANUAL
   AS SELECT k, percentile_union(percentile_hash(v)) AS p
   FROM t GROUP BY k;
   ```

2. Issue a query that asks for a high-precision compression:

   ```sql
   SELECT k, percentile_approx(v, 0.95, 10000) FROM t GROUP BY k;
   ```

3. The optimizer rewrites the query to read from `mv`. The user's
   `compression=10000` request is silently answered at `compression=1000`.

## Expected behavior

The compression factor should be respected end to end. Either:
  - The MV stores digests at the precision its DDL implies, and rewrite
    verifies the MV's precision covers the query's request before using it; or
  - The optimizer refuses the rewrite (configurable) and falls back to a base
    table scan.

## Real behavior

Rewrite proceeds silently, the answer is computed at `compression=1000`, no
warning, no trace.

## StarRocks version

main

eshishki:
## Why I'm doing:

`percentile_approx(v, q, c)` builds a T-digest sized by `c`, where the
compression controls the precision / memory tradeoff.

Today percentile MV rewrite does not reason about that compression contract.
A query that asks for a higher compression can be rewritten to a percentile MV
whose stored digest was built at a lower compression, and the rewrite proceeds
without any signal that precision was downgraded.

That makes compression requests invisible to the optimizer at the point where
it decides whether a percentile MV is an adequate substitute for the base-table
aggregation.

## What I'm doing:

Teach percentile MV rewrite to track whether the MV-side digest compression
covers the query-side compression request.

`PercentileRewriteEquivalent` reads:

- `queryC` from the query-side `percentile_approx`
- `mvC` from the MV-side percentile digest shape

The rewrite is considered subsume when `mvC >= queryC`. When `mvC < queryC`,
the rewrite is marked as a percentile non-subsume rewrite and that signal is
propagated through the MV rewrite path into candidate selection.

This lets the rewrite path apply one compression-aware policy:

- prefer subsume percentile MV candidates when they are available;
- preserve legacy rewrite behavior by default for a non-subsume candidate;
- reject a non-subsume candidate when
  `enable_mv_percentile_strict_match` is enabled, with the compression mismatch
  logged for tracing.

Tests cover the new session variable and candidate ordering behavior.

### Stack note

This PR is the headline MV rewrite piece and is currently stacked on:

- #73721, which canonicalizes percentile compression literals in
  `FunctionAnalyzer`
- #73723, which adds `percentile_hash(value, compression)` so MV definitions can
  store an explicit percentile compression

Until those PRs are merged, the GitHub diff against `main` includes their base
changes as well.

The independent `percentile_union` compression preservation fix is split into
#73722.

Fixes #73636

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/agg/percentile_approx.h
+++ b/be/src/exprs/agg/percentile_approx.h
@@ -54,10 +54,21 @@ class PercentileApproxAggregateFunctionBase
 protected:
     static constexpr double MIN_COMPRESSION = 2048.0;
     static constexpr double MAX_COMPRESSION = 10000.0;
-    // Used only as a rolling-upgrade fallback when an old FE ships a legacy
-    // arity (without the canonicalized compression literal). For new-FE calls
-    // FunctionAnalyzer is the single source of truth and DEFAULT lives there.
+    // Kept on BE for rolling upgrades from pre-canonicalization FEs. For
+    // new-FE calls FunctionAnalyzer is the single source of truth and DEFAULT
+    // lives there.
     static constexpr double DEFAULT_COMPRESSION_FACTOR = 10000.0;
+
+    static double clamp_compression_factor(double compression) {
+        if (LIKELY(std::isfinite(compression) && compression >= MIN_COMPRESSION && compression <= MAX_COMPRESSION)) {
+            return compression;
+        }
+        // Old FEs can still ship explicit compression literals without the
+        // analyzer-side [MIN, MAX] clamp during a rolling upgrade.
+        LOG(WARNING) << "Compression factor out of range. Using default compression factor: "
+                     << DEFAULT_COMPRESSION_FACTOR;
+        return DEFAULT_COMPRESSION_FACTOR;
+    }
 
 public:
     virtual double get_compression_factor(FunctionContext* ctx) const = 0;
@@ -107,9 +118,9 @@ public:
 class PercentileApproxAggregateFunction : public PercentileApproxAggregateFunctionBase {
 public:
     double get_compression_factor(FunctionContext* ctx) const override {
-        // FE FunctionAnalyzer is the single source of truth: an explicit
-        // compression literal is always injected (default or user-supplied)
-        // and clamped to [MIN, MAX]. The DCHECKs document that contract.
+        // FE FunctionAnalyzer is the single source of truth for new-FE calls:
+        // an explicit compression literal is always injected (default or
+        // user-supplied) and clamped to [MIN, MAX].
         //
         // Rolling-upgrade fallback: during the upgrade window BEs are upgraded
         // before FEs, so an old FE may still ship 2-arg percentile_approx
@@ -121,10 +132,7 @@ public:
             return DEFAULT_COMPRESSION_FACTOR;
         }
         double compression = ColumnHelper::get_const_value<TYPE_DOUBLE>(ctx->get_constant_column(2));
-        DCHECK(std::isfinite(compression));
-        DCHECK_GE(compression, MIN_COMPRESSION);
-        DCHECK_LE(compression, MAX_COMPRESSION);
-        return compression;
+        return clamp_compression_factor(compression);
     }
 
     void update(FunctionContext* ctx, const Column** columns, AggDataPtr state, size_t row_num) const override {
@@ -187,7 +195,8 @@ class PercentileApproxWeightedAggregateFunction : public PercentileApproxAggrega
 public:
     // SplitAggregateRule passes the const args to merge phase aggregator for performance.
     // Compression parameter is always the last constant parameter (if provided).
-    // FE FunctionAnalyzer pre-clamps the literal to [MIN, MAX] or DEFAULT.
+    // FE FunctionAnalyzer pre-clamps the literal to [MIN, MAX] or DEFAULT for
+    // new-FE calls.
     double get_compression_factor(FunctionContext* ctx) const override {
         // FE always injects the compression argument; it is the last const arg.
         // SplitAggregateRule passes const args verbatim into the merge phase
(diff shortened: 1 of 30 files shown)


CelerData-Reviewer:
@codex review

eshishki:
@stdpain — this PR addresses the storage/runtime compression mismatch called out in #73528 (point 3): the union aggregate now adopts the incoming digest's compression instead of materializing at TDigest(1000), the PASS_THROUGH paths in percentile_approx.h pick up the configured compression, and the new 2-arg percentile_hash(v, c) overload lets MV DDL pin a compression at storage time.

We did NOT touch points 1 and 2 from #73528 — the DEFAULT_COMPRESSION_FACTOR value and the out-of-range clamp semantics. If you'd agree on a direction for the default (e.g. lowering it from 10000), we'd be happy to fold that change into this PR so the policy lands in one piece. Otherwise we can keep them separate.

eshishki:
Thanks. Fixed all four in `6cbb80c363c`:

- **Missing import (`PercentileRewriteEquivalent.java`)**: added `import com.starrocks.sql.analyzer.PercentileCompression;`. Real compile error.
- **Singular `get_compression_factor`**: restored the `num_args <= 2` fallback to DEFAULT. BEs upgrade before FEs, so we need to tolerate an old FE that still ships 2-arg `percentile_approx`. Comment now explicitly calls it a rolling-upgrade shim.
- **Weighted `get_compression_factor`**: added a `num_args <= 3` short-circuit to DEFAULT before scanning. Without it the rightmost const at legacy arity is the quantile, which would feed `PercentileValue(0.95)`. Same rolling-upgrade rationale.
- **`percentile_hash_with_compression` DCHECK range**: tightened to `DCHECK_GE(compression, 2048.0)` / `DCHECK_LE(compression, 10000.0)` so BE mirrors the FE-side `PercentileCompression.MIN/MAX` contract.

HangyuanLiu:
Thanks for the thorough work — the contract design and inline comments are excellent.

Before diving into per-change feedback, I'd like to suggest splitting this PR. As written, it bundles **three semantically independent changes** plus one supporting refactor, and I think they'd each be better served as separate PRs:

**1. BugFix: `percentile_union` silently downgrades to `TDigest(1000)`**
The `PercentileUnionState` lazy-init change. This affects *every* `percentile_union` query, regardless of MVs — it's a latent precision bug that has nothing to do with the MV rewrite story. It deserves a standalone `[BugFix]` that can be backported to release branches on its own merits.

**2. Feature: `percentile_hash(value, compression)` overload**
New DDL capability (opcode 130003). Independent of rewrite logic.

**3. Refactor: FE canonicalizes compression literal, BE becomes strict executor**
Required by #4 to reliably read `mvC`/`queryC`, but carries a real behavior change: `percentile_approx(v, q, 0)` used to hard-fail with "compression must be positive", now silently clamps to `DEFAULT=10000`. Reviewers may want to keep the hard-fail and only canonicalize NULL — that conversation shouldn't block #1.

**4. Enhancement: MV rewrite respects compression**
The actual headline change — `PercentileRewriteEquivalent` + `BestMvSelector` + `enable_mv_percentile_strict_match`. Depends on #2 and #3.

Dependency chain is clean: `#1` independent, `#2` independent, `#3 → #4`, `#2 → #4`.

Why split:
- **#1 is backport-worthy** on its own; bundled inside an `[Enhancement]` it's likely to be missed.
- **#3's behavior regression** deserves its own discussion thread — silently turning `compression=0` from error into default-clamp is the kind of change that needs explicit sign-off, not buried under "supporting changes".
- **Rollback granularity**: if #4 regresses in production, we can revert just that piece without losing the union bugfix.
- **Review depth**: +738/-117 across 20 files spanning BE+FE makes it easy to miss cross-cutting issues.

Suggested order:
1. `[BugFix] Fix percentile_union silently downgrading to TDigest(1000)`
2. `[Refactor] Canonicalize percentile compression literal in FunctionAnalyzer`
3. `[Feature] Add percentile_hash(value, compression) overload`
4. `[Enhancement] Respect percentile compression in MV rewrite` (the current PR title)

Happy to re-review each. I'll hold the detailed per-change feedback (release-build DCHECK fallback, `percentile_empty` ordering issue, missing e2e MV rewrite tests, behavior of `compression=0`) until the split lands, so the discussion doesn't fragment.

eshishki:
@HangyuanLiu thanks for the suggestion — I split the original PR into the smaller pieces:

1. BugFix: fix `percentile_union` silently downgrading digest compression
   #73722
2. Refactor: canonicalize percentile compression literals in `FunctionAnalyzer`
   #73721
3. Feature: add `percentile_hash(value, compression)` overload
   #73723
4. Enhancement: keep this PR as the headline MV rewrite piece
   #73634

The stacked PR bodies call out their current dependencies, and the independent
`percentile_union` fix is separate for review/backport purposes.

Would you prefer separate issues for these split PRs as well, or is keeping
them linked to #73636 enough?

Happy to add any extra coverage or adjust the split/details if you want more
tests or anything else before review.

PR #73646, opened 2026-05-21
[BugFix] Add write timeout to MysqlChannel result send path

Youngwb:
## Why I'm doing:

When a MySQL client stops draining its TCP receive buffer (slow consumer or backpressure), the FE worker that's sending result rows blocks **indefinitely** in `Selector.select()` on `OP_WRITE` inside `MysqlChannel.realNetSend`. The query becomes unkillable via `KILL QUERY` because `Coordinator.cancel()` only cancels work on the BE side and does not interrupt the FE-side socket wait. The connection can only be released via `KILL CONNECTION`, which closes the channel and wakes the selector.

We hit this in production on a follower FE running shared-data. A `WITH ... window function` query stayed in `SHOW PROC '/current_queries'` for ~30 minutes; `KILL QUERY` and the internal `Connect-Scheduler-Check-Timer` both ran every second but had no effect. The stack trace captured during the hang:

```
"starrocks-mysql-nio-pool-15167" #1321580 daemon prio=5 ... runnable
   java.lang.Thread.State: RUNNABLE
    at sun.nio.ch.EPoll.wait(Native Method)
    at sun.nio.ch.EPollSelectorImpl.doSelect(EPollSelectorImpl.java:118)
    at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:129)
    at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:146)
    at org.xnio.nio.SelectorUtils.await(SelectorUtils.java:51)
    at org.xnio.nio.NioSocketConduit.awaitWritable(NioSocketConduit.java:233)
    at org.xnio.channels.Channels.writeBlocking(Channels.java:149)
    at com.starrocks.mysql.MysqlChannel.realNetSend(MysqlChannel.java:239)
    at com.starrocks.mysql.MysqlChannel.send(MysqlChannel.java:232)
    at com.starrocks.mysql.MysqlChannel.flush(MysqlChannel.java:256)
    at com.starrocks.mysql.MysqlChannel.writeBuffer(MysqlChannel.java:294)
    at com.starrocks.mysql.MysqlChannel.sendOnePacket(MysqlChannel.java:318)
    at com.starrocks.qe.StmtExecutor.responseRowBatch(StmtExecutor.java:1621)
    at com.starrocks.qe.StmtExecutor.handleQueryStmt(StmtExecutor.java:1598)
    at com.starrocks.qe.StmtExecutor.execute(StmtExecutor.java:812)
    at com.starrocks.qe.ConnectProcessor.handleQuery(ConnectProcessor.java:372)
    ...
```

The thread is parked on epoll waiting for `OP_WRITE` while sending result rows; the client is just not consuming.

## What I'm doing:

Switch `MysqlChannel.realNetSend` to the timed variant of `Channels.writeBlocking`, gated by a new mutable FE config `mysql_send_packet_timeout_ms` (default 60_000 ms, aligned in spirit with MySQL's `net_write_timeout`). When the timeout fires, the underlying `writeBlocking` returns a partial byte count, the existing `bufLen != writeLen` check throws `IOException`, the call stack unwinds, and the connection is cleaned up through the normal error path — `KILL QUERY` and timeout-based kill both work again. Setting the config to `0` preserves the legacy unbounded-wait behavior for compatibility.

The error message now includes `timeoutMs=...` so the failure mode is easy to identify in logs.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [x] Parameter changes: a new FE config `mysql_send_packet_timeout_ms` is introduced. Default is 60_000 ms; set to 0 for legacy behavior.
- [ ] Policy changes
- [ ] Feature removed
- [ ] Miscellaneous

## Checklist:

  - `MysqlChannelTest` covers: timed variant when timeout > 0, untimed variant when timeout = 0, IOException on partial write after timeout.
- [x] This pr needs user documentation
  - [x] I have added documentation for my new feature or new function (EN / ZH / JA `FE_parameters/log_server_meta.md`)
- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/mysql/MysqlChannel.java
+++ b/fe/fe-core/src/main/java/com/starrocks/mysql/MysqlChannel.java
@@ -34,6 +34,7 @@
 
 package com.starrocks.mysql;
 
+import com.starrocks.common.Config;
 import com.starrocks.common.util.NetUtils;
 import com.starrocks.mysql.nio.MySQLReadListener;
 import com.starrocks.mysql.ssl.SSLChannel;
@@ -48,6 +49,7 @@ import org.xnio.channels.Channels;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
+import java.util.concurrent.TimeUnit;
 
 /**
  * This class used to read/write MySQL logical packet.
@@ -244,12 +246,24 @@ public class MysqlChannel {
 
     public void realNetSend(ByteBuffer buffer) throws IOException {
         long bufLen = buffer.remaining();
-        long writeLen = Channels.writeBlocking(conn.getSinkChannel(), buffer);
+        // timeoutMs <= 0 keeps legacy unbounded wait for both phases. With a timeout, both
+        // writeBlocking and flushBlocking are bounded; either phase converts a slow-client
+        // hang into IOException so the connection cleanup path runs.
+        long timeoutMs = Config.mysql_send_packet_timeout_ms;
+        long writeLen = timeoutMs > 0
+                ? Channels.writeBlocking(conn.getSinkChannel(), buffer, timeoutMs, TimeUnit.MILLISECONDS)
+                : Channels.writeBlocking(conn.getSinkChannel(), buffer);
         if (bufLen != writeLen) {
             throw new IOException("Write mysql packet failed.[write=" + writeLen
-                    + ", needToWrite=" + bufLen + "]");
+                    + ", needToWrite=" + bufLen + ", timeoutMs=" + timeoutMs + "]");
         }
-        Channels.flushBlocking(conn.getSinkChannel());
+        if (timeoutMs > 0) {
+            if (!Channels.flushBlocking(conn.getSinkChannel(), timeoutMs, TimeUnit.MILLISECONDS)) {
+                throw new IOException("Flush mysql packet timed out.[timeoutMs=" + timeoutMs + "]");
+            }
+        } else {
+            Channels.flushBlocking(conn.getSinkChannel());
+        }
         isSend = true;
     }
 
--- a/docs/en/administration/management/FE_parameters/log_server_meta.md
+++ b/docs/en/administration/management/FE_parameters/log_server_meta.md
@@ -839,6 +839,15 @@ This topic introduces the following types of FE configurations:
 - Description: The length of the backlog queue held by the MySQL server in the FE node.
 - Introduced in: -
 
+### `mysql_send_packet_timeout_ms`
+
+- Default: 60000
+- Type: Long
+- Unit: Milliseconds
+- Is mutable: Yes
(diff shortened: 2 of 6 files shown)


CelerData-Reviewer:
@codex review

stdpain (dismissed):

gengjun-git (dismissed):

fe/fe-core/src/main/java/com/starrocks/mysql/MysqlChannel.java
```diff
@@ -244,10 +246,14 @@ private void send(ByteBuffer buffer) throws IOException {
...
+        long timeoutMs = Config.mysql_send_packet_timeout_ms;
+        long writeLen = timeoutMs > 0
+                ? Channels.writeBlocking(conn.getSinkChannel(), buffer, timeoutMs, TimeUnit.MILLISECONDS)
+                : Channels.writeBlocking(conn.getSinkChannel(), buffer);
         if (bufLen != writeLen) {
             throw new IOException("Write mysql packet failed.[write=" + writeLen
-                    + ", needToWrite=" + bufLen + "]");
+                    + ", needToWrite=" + bufLen + ", timeoutMs=" + timeoutMs + "]");
         }
         Channels.flushBlocking(conn.getSinkChannel());
```
HangyuanLiu: This leaves a second unbounded wait in the same send path. XNIO's timed writeBlocking only bounds writing bytes from the ByteBuffer; flushBlocking is the separate phase that drains the channel/send queue, and the untimed variant can also wait on awaitWritable() indefinitely under the same slow-client/TCP backpressure conditions. Please switch this to the timed flushBlocking variant as well and treat a false return as a send timeout IOException, or add evidence that this sink channel cannot block in flush after writeBlocking succeeds.

HangyuanLiu (changes requested):
Requesting changes because the current patch only bounds the write phase. The send path still calls the untimed flushBlocking immediately afterward, so the original slow-consumer/backpressure problem can still leave the FE worker waiting without the new timeout applying.

Youngwb:
Good catch — pushed `cb1db6e` to address this.

You're correct that the untimed `flushBlocking(channel)` calls `awaitWritable()` without bound. Today's `NioSocketConduit.flush()` returns `true` unconditionally so the raw-socket path doesn't actually reach the `awaitWritable` call, but that's an implicit assumption on the conduit chain — any future conduit (SSL framing, batching layer, etc.) whose `flush()` returns `false` would reintroduce the same 30-minute hang.

Switched flush to the timed `Channels.flushBlocking(channel, time, unit)` variant in the same code path. A `false` return is treated as a flush timeout and raises `IOException` so the cleanup path runs, mirroring the partial-write handling above it. The legacy `timeoutMs == 0` branch keeps the untimed flush.

Added `testRealNetSendThrowsOnFlushTimeout` to cover the new branch (4 tests total, all passing).

CelerData-Reviewer:
@codex review

HangyuanLiu (approved):

Astralidea (approved):

PR #69939, opened 2026-03-06
[Enhancement] Default to PK_ENCODING_TYPE_V2 for new cloud-native PK tables

luohaha:
## Why I'm doing:
Currently, newly created cloud-native (shared-data) primary key tables with hash distribution default to `PK_ENCODING_TYPE_V1`, while only range-distributed tables use `PK_ENCODING_TYPE_V2`. V2 encoding provides better support and should be the default for all new cloud-native PK tables.

## What I'm doing:
- Persist `primaryKeyEncodingType` as an immutable field in `OlapTable` (set once at creation, never altered)
- New cloud-native PK tables default to `PK_ENCODING_TYPE_V2` regardless of distribution type
- New non-cloud-native PK tables explicitly persist `PK_ENCODING_TYPE_V1`
- Existing tables (created before this change) have `primaryKeyEncodingType = null` and fall back to the legacy computed logic for full backward compatibility:
  - Cloud-native + hash → V1
  - Cloud-native + range → V2
  - Non-cloud-native → V1

## What type of PR is this:
- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Does this PR entail a change in behavior?
- [x] Yes, this PR will result in a change in behavior.
- [ ] No, this PR will not result in a change in behavior.

## If yes, please specify the type of change:
- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [x] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:
- [x] I have added test cases for my bug fix or my new feature
- [ ] This PR needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport PR

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the PR will be auto-backported to the target branch
  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

PR #73649, opened 2026-05-21
[BugFix] Transcode PK .del files on V1->V2 cross-cluster replication

### LINKED ISSUE #73577: [BugFix] Cross-cluster replication from 3.5 (shared-nothing) to 4.1 (shared-data) PK table fails to deserialize .del file due to PK encoding type V1/V2 mismatch

1. Source cluster: StarRocks **3.5.14** (shared-nothing). Create a primary key table with a **single fixed-length PK column** (e.g. `INT` / `BIGINT` / `DATE`) and produce some delete operations so that `.del` files are generated under the rowset:
   ```sql
   CREATE TABLE t (id INT NOT NULL, v INT) PRIMARY KEY(id)
       DISTRIBUTED BY HASH(id) BUCKETS 3
       PROPERTIES("replication_num"="1");
   INSERT INTO t VALUES (1,1),(2,2),(3,3);
   DELETE FROM t WHERE id = 2;          -- generates a .del file in the rowset
   ```
2. Target cluster: StarRocks **4.1 latest** (shared-data, cloud-native). Use the cross-cluster data migration tool to sync the table from the 3.5.14 source to the 4.1 target.
3. On the target BE, watch `be.WARNING` while the replication txn publishes.

Replication txn `TXN_REPLICATION` publishes successfully and the target table converges to the source's data.

### Real behavior (Required)

Publishing the replication txn fails on the target BE with a buffer-underrun in primary-key column deserialization:

```
W20260520 18:05:18.757639 ... transactions.cpp:437] Fail to apply txn log :
Internal error: Expected remains size 4, but get 0
  be/src/serde/column_array_serde.cpp:66  check_remaining_size(buff, end, sizeof(uint32_t))
  be/src/serde/column_array_serde.cpp:323 value_or_err_L323
  be/src/serde/column_array_serde.cpp:776 value_or_err_L776
  be/src/serde/column_array_serde.cpp:837 column->accept_mutable(&visitor)
  be/src/storage/lake/rowset_update_state.cpp:908 Serd::deserialize(begin, end, col.get())
  be/src/storage/lake/update_manager.cpp:506 state.load_delete(del_id, params)
  be/src/storage/lake/txn_log_applier.cpp:676 apply_write_log(op_write, txn_id)
  be/src/storage/lake/txn_log_applier.cpp:328 apply_replication_log(log.op_replication(), log.txn_id())
  tablet_info={tablet_id: 14913}
  txn=txn_id: 1100 commit_time: 1779271045 combined_txn_log: false
       txn_type: TXN_REPLICATION force_publish: false
       gtid: 422438231569596416
W20260520 18:05:18.760122 ... lake_service.cpp:381] Fail to publish version: <same stack> version=1074429
```

The replication txn keeps retrying / never converges; the target tablet stays behind the source.

### Root cause analysis

The `.del` file is the serialized PK column for deleted rows. `flush_chunk_with_deletes` on the source uses `serde::ColumnArraySerde::serialize(deletes, ...)` with whatever column shape `PrimaryKeyEncoder::create_column` produced. For **V1 + single PK column** that shape is the **raw fixed-length column** (e.g. `Int32Column`); for **V2 (or composite PK)** it is `BinaryColumn`. See `be/src/storage/primary_key_encoder.cpp` around the `key_idxes.size() == 1 && encoding_type == PK_ENCODING_TYPE_V1` branch.

Cross-cluster replication transfers the `.del` file **byte-for-byte** — `ReplicationTxnManager::convert_rowset_meta` only renames the file and rewrites encryption metas (`be/src/storage/lake/replication_txn_manager.cpp` around `// Convert dels`). It does **not** re-encode the column.

But on apply, `RowsetUpdateState::load_delete` (`be/src/storage/lake/rowset_update_state.cpp`) builds the deserialization column from the **target** schema's encoding type:

```cpp
ASSIGN_OR_RETURN(auto pk_encoding_type,
                 params.tablet_schema->primary_key_encoding_type_or_error());
RETURN_IF_ERROR(PrimaryKeyEncoder::create_column(pkey_schema, &pk_column, pk_encoding_type));
...
RETURN_IF_ERROR(Serd::deserialize(begin, end, col.get()));
```

In 4.1, since [[#69939](https://github.com/StarRocks/starrocks/pull/69939)](https://github.com/StarRocks/starrocks/pull/69939) (commit `9964962fd6`, *"Default to PK_ENCODING_TYPE_V2 for new cloud-native PK tables"*), every newly created cloud-native PK table is V2:

```java
// fe/fe-core/.../OlapTableFactory.java
if (table.isCloudNativeTableOrMaterializedView()) {
    table.setPrimaryKeyEncodingType(TPrimaryKeyEncodingType.PK_ENCODING_TYPE_V2);
} else {
    table.setPrimaryKeyEncodingType(TPrimaryKeyEncodingType.PK_ENCODING_TYPE_V1);
}
```

So when the source is 3.5.x shared-nothing (V1, single fixed-length PK → `Int32Column`-shaped bytes on disk) and the target is 4.1 shared-data (V2 → `BinaryColumn`-shaped expected), `BinaryColumnSerde::deserialize` reads the V1 file as `[bytes_size: u32][bytes][offset_bytes_size: u32][offsets]`, consumes the entire file in the first two reads, and the third `read_little_endian_32` for `offset_bytes_size` finds 0 bytes left — exactly the observed `Expected remains size 4, but get 0`.

**Trigger condition**: PK is a single fixed-length column (INT/BIGINT/DATE/…). Composite PK or single VARCHAR PK both use `BinaryColumn` on either side and are unaffected.

### Suggested fix

Two viable directions, the first is preferred:

1. In `RowsetUpdateState::load_delete`, when the rowset's `.del` came from a replication snapshot, deserialize using **the source's** PK encoding type. `op_replication.source_schema()` is already carried in the txn log (`replication_txn_manager.cpp` line ~319/366/371), and `apply_replication_log` already copies it onto `_metadata->mutable_source_schema()`. Plumb that into `RowsetUpdateStateParams` and prefer it over `tablet_schema->primary_key_encoding_type_or_error()` when present.
2. Alternatively, re-encode the `.del` payload during `ReplicationTxnManager::convert_rowset_meta` so on-disk bytes always match the target schema's encoding.

Either way, an integration test crossing V1 source × V2 target with a single-INT PK should be added under `be/test/storage/lake/`.

### Workaround for users hitting this today

Recreate the target table with V1 explicitly. There is currently no `primary_key_encoding` table property exposed; the cleanest workaround is to patch the FE to set V1 in `OlapTableFactory` for the migrated table (revert of #69939 for that table), recreate, and re-run the migration.

- Source: 3.5.14 (shared-nothing)
- Target: 4.1 latest branch (shared-data, cloud-native)

xiangguangyxg:
## Why I'm doing:

When replicating a primary-key table with a **single non-string fixed-length PK column** (INT/BIGINT/DATE/DATETIME/...) from a V1-encoded source cluster (e.g. shared-nothing 3.5.x) to a V2-encoded target (4.1 cloud-native, default since #69939), `.del` files were byte-copied during snapshot intake. The two encodings produce incompatible on-disk shapes:
- V1 + single non-string fixed-length PK → typed column (e.g. `Int32Column`)
- V2 → `BinaryColumn` with order-preserving big-endian slices

The target reader hit `Internal error: Expected remains size 4, but get 0` in `serde::ColumnArraySerde::deserialize` during publish, and the replication txn never converged.

Composite PKs and single VARCHAR PK both materialize as `BinaryColumn` with identical layout in V1 and V2, so they were unaffected.

## What I'm doing:

Introduce `DelFileStreamConverter` (subclass of `FileStreamConverter`, alongside the existing `SegmentStreamConverter`) that re-encodes `.del` payload bytes once at replication intake. After this converter writes the output file, every downstream consumer (`load_delete`, `load_dels`, future compaction, future readers) sees uniform target-encoded bytes — no per-read awareness needed.

Wired into `ReplicationTxnManager::build_file_converters`: when `requires_v1_to_v2_del_transcode(source_enc, target_enc, pkey_schema)` returns true and the file is `.del`, the converter is used; otherwise the plain `FileStreamConverter` byte-copy is preserved.

Other fixes folded in:
- Gate the `source_schema` fallback on `tablet_metadata->has_source_schema()` (the prior nullptr check was a no-op since protobuf returns a default instance).
- Validate cross-cluster PK column count and per-column logical type match; return `NotSupported` on mismatch.
- Reject V2→V1 cross-cluster replication explicitly on the byte-incompatible PK shape (no V2→V1 transcoder).
- Lake-to-lake `NotSupported` guard for the same byte-incompatible shape (lake-to-lake raw-copies `.del` files via `copy_non_segment_file_with_retry`; routing through a transcoding converter there is a follow-up).
- Use `primary_key_encoding_type_or_error()` instead of silently treating `PK_ENCODING_TYPE_NONE` as V1.

Fixes #73577

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/lake_replication_txn_manager.cpp
+++ b/be/src/storage/lake/lake_replication_txn_manager.cpp
@@ -253,35 +253,9 @@ Status LakeReplicationTxnManager::replicate_lake_remote_storage(const TReplicate
     std::vector<std::string> files_to_delete;
     CancelableDefer clean_files([&files_to_delete]() { lake::delete_files_async(std::move(files_to_delete)); });
 
-    // Guard: lake-to-lake replication routes .del files through copy_non_segment_file_with_retry
-    // (a raw byte copy), not through file_converters. So if source/target use different PK
-    // encodings in a way that produces non-byte-compatible .del bytes (single non-string
-    // fixed-length PK + V1->V2), the byte copy would silently corrupt the target. Reject
-    // explicitly until lake-to-lake routes .del through a transcoding converter as well.
-    //
-    // V1 source vs V2 target shows up when the source tablet is from a pre-#69939 cloud-native
-    // cluster (V1 default) and the target is post-#69939 (V2 default for new cloud-native tables).
-    if (is_primary_key(*target_tablet_meta)) {
-        // src_tablet_meta->has_schema() is guaranteed above (early Corruption otherwise).
-        TabletSchema source_schema(src_tablet_meta->schema());
-        TabletSchema target_schema(target_tablet_meta->schema());
-        ASSIGN_OR_RETURN(auto source_encoding, source_schema.primary_key_encoding_type_or_error());
-        ASSIGN_OR_RETURN(auto target_encoding, target_schema.primary_key_encoding_type_or_error());
-        // Both directions of V1<->V2 are byte-incompatible for single non-string fixed-length PK,
-        // and the lake-to-lake path raw-copies .del files. Reject either direction explicitly until
-        // .del is routed through a transcoding converter here as well.
-        if (source_encoding != target_encoding && is_single_fixed_length_non_string_primary_key(target_schema)) {
-            LOG(WARNING) << "Lake-to-lake replication with PK encoding mismatch is not supported, src_tablet: "
-                         << src_tablet_id << ", target_tablet: " << target_tablet_id
-                         << ", source_encoding: " << static_cast<int>(source_encoding)
-                         << ", target_encoding: " << static_cast<int>(target_encoding)
-                         << ", pk_logical_type: " << logical_type_to_string(target_schema.column(0).type());
-            return Status::NotSupported(
-                    "Lake-to-lake replication with V1<->V2 primary-key encoding change on a single "
-                    "non-string fixed-length PK column is not supported; .del files in this path "
-                    "are byte-copied without re-encoding and would fail to apply on the target.");
-        }
-    }
+    // src_tablet_meta->has_schema() is guaranteed by the early Corruption return above.
+    RETURN_IF_ERROR(check_pk_encoding_compat_for_lake_to_lake(src_tablet_id, target_tablet_id, *src_tablet_meta,
+                                                              *target_tablet_meta));
 
     // PK transcoding is gated above; pass nullptr/NONE so the converter never enters the .del
     // transcode branch in this code path.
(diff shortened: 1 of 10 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

kevincai (approved):

be/src/storage/lake/lake_replication_txn_manager.cpp
```diff
@@ -453,6 +464,38 @@ bool LakeReplicationTxnManager::should_use_parallel_copy(size_t file_count, cons
...
 }
 
+Status LakeReplicationTxnManager::check_pk_encoding_compat(int64_t src_tablet_id, int64_t target_tablet_id,
+                                                           const TabletMetadata& src_tablet_meta,
+                                                           const TabletMetadata& target_tablet_meta) {
+    if (!is_primary_key(target_tablet_meta)) {
+        return Status::OK();
+    }
+    // Lake-to-lake replication routes .del files through copy_non_segment_file_with_retry (a raw
+    // byte copy), not through file_converters. So if source/target use different PK encodings in
```
srlch, line 474: Is that means lake to lake replication has no way to perform file conversion, it must be aborted if the type is mismatched?

xiangguangyxg: In this PR, yes. Will fix in another PR

srlch (approved):

PR #73661, opened 2026-05-21
[BugFix] Deduplicate replicas in TabletInvertedIndex.addReplica

gengjun-git:
## Why I'm doing:

`TabletInvertedIndex.addReplica` appends to a per-tablet `CopyOnWriteArrayList<Replica>` without checking for an existing entry on the same backend. Any caller path that replaces a replica without an explicit prior `invertedIndex.deleteReplica` accumulates stale references. `getReplica` returns the **first** matching entry in iteration order, so the stale reference can be returned even after the live one has been added.

The most visible production trigger is the `VERSION_INCOMPLETE` recovery branch of `ReportHandler.processTabletReport` (around line 2152): `tablet.addReplica` is invoked while a version-behind replica still exists on the same backend; `LocalTablet.deleteRedundantReplica` removes the old replica from the tablet's local list but never calls `invertedIndex.deleteReplica`. The inverted index keeps the stale Replica, and subsequent `invertedIndex.getReplica(tabletId, backendId)` calls return that orphan.

**Caused by #66288** (`[Enhancement] Remove TabletinvertedIndex lock to improve metadata performance`). That refactor replaced the prior `Map<Long, Map<Long, Replica>>` (outer tabletId → inner backendId → Replica) with a flat `ConcurrentLong2ObjectHashMap<CopyOnWriteArrayList<Replica>>`. The lock removal is preserved; what's lost is the implicit put-overwrite semantics of the inner Map, which previously masked the missing `deleteReplica` cleanup downstream in `LocalTablet.deleteRedundantReplica`.

## What I'm doing:

In `addReplica`, look up the first existing entry for the same `backendId` and use `CopyOnWriteArrayList.set(index, replica)` to swap it in place. `set` is atomic from a concurrent reader's perspective, so `getReplica` never observes a transient absence — equivalent to the prior `Map.put` semantics.

Falls back to `add` when no prior entry exists. (An earlier revision also swept additional duplicates after the swap, but per @kevincai that path is dead code: FE journal replay on restart rebuilds the inverted index through the fixed `addReplica`, and any upgrade rolling-restarts the FE — so no in-memory state can carry pre-fix duplicates into the post-fix world.)

This avoids:

- restructuring the data layout again (preserves #66288's lock-free design and existing `getReplicaMetaTable` consumers);
- requiring every direct `invertedIndex.addReplica` caller (e.g., `RestoreJob`) to also know to call `deleteReplica` first;
- leaving an audit dependency on `LocalTablet.deleteRedundantReplica` keeping the index in sync.

Unit tests cover three states: fresh add, re-add for the same backendId, and pre-seeded legacy duplicates that the sweep self-heals.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
@@ -224,13 +224,39 @@ public class TabletInvertedIndex implements MemoryTrackable {
         try {
             CopyOnWriteArrayList<Replica> replicas = tabletToReplicaList.computeIfAbsent(tabletId,
                     k -> new CopyOnWriteArrayList<>());
-            replicas.add(replica);
+            long backendId = replica.getBackendId();
 
-            Set<Long> tabletIds = backendToTabletIdList.computeIfAbsent(replica.getBackendId(),
+            // Restore the per-(tabletId, backendId) uniqueness invariant that the
+            // prior Map<backendId, Replica> inner container provided in branch-3.5
+            // before #66288. Replace the first existing entry in place so concurrent
+            // readers see an atomic swap (never a transient absence). Sweep any
+            // remaining duplicates left behind by historical addReplica leaks (e.g.
+            // LocalTablet.deleteRedundantReplica drops a replica from the tablet
+            // without notifying this index) so legacy state self-heals on the next
+            // add for that backend.
+            int firstIdx = -1;
+            for (int i = 0; i < replicas.size(); i++) {
+                if (replicas.get(i).getBackendId() == backendId) {
+                    firstIdx = i;
+                    break;
+                }
+            }
+            if (firstIdx >= 0) {
+                replicas.set(firstIdx, replica);
+                for (int i = replicas.size() - 1; i > firstIdx; i--) {
+                    if (replicas.get(i).getBackendId() == backendId) {
+                        replicas.remove(i);
+                    }
+                }
+            } else {
+                replicas.add(replica);
+            }
+
+            Set<Long> tabletIds = backendToTabletIdList.computeIfAbsent(backendId,
                     k -> ConcurrentHashMap.newKeySet());
             tabletIds.add(tabletId);
 
-            LOG.debug("add replica {} of tablet {} in backend {}", replica.getId(), tabletId, replica.getBackendId());
+            LOG.debug("add replica {} of tablet {} in backend {}", replica.getId(), tabletId, backendId);
         } finally {
             mutationLock.unlock();
         }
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -224,13 +224,39 @@ public void addReplica(long tabletId, Replica replica) {
...
+            int firstIdx = -1;
+            for (int i = 0; i < replicas.size(); i++) {
+                if (replicas.get(i).getBackendId() == backendId) {
+                    firstIdx = i;
+                    break;
+                }
+            }
+            if (firstIdx >= 0) {
+                replicas.set(firstIdx, replica);
+                for (int i = replicas.size() - 1; i > firstIdx; i--) {
```
kevincai: is it duplicate check still necessary? InvertedIndex will be rebuilt once the FE process restarted. With the correct addReplica implementation, there will no duplication once rebuild, and no incremental duplication with the fix in this PR.

gengjun-git: Good catch — agreed, removed in df58bb6818f. The sweep was dead code: journal replay on FE restart rebuilds the inverted index through this same `addReplica`, and any upgrade rolling-restarts the FE, so there's no in-memory state on which the sweep could ever find a legacy duplicate to clean up. The set-or-add alone enforces the per-`(tabletId, backendId)` uniqueness invariant for all future calls. The corresponding `testAddReplicaSelfHealsLegacyDuplicates` unit test was dropped along with the sweep.

fe/fe-core/src/main/java/com/starrocks/catalog/TabletInvertedIndex.java
```diff
@@ -224,13 +224,39 @@ public void addReplica(long tabletId, Replica replica) {
...
+                    if (replicas.get(i).getBackendId() == backendId) {
+                        replicas.remove(i);
+                    }
+                }
```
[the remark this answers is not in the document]
gengjun-git: Addressed by dropping the sweep entirely in df58bb6818f per @kevincai's observation that it is dead code after the fix lands (journal replay on FE restart rebuilds through the fixed `addReplica`). With no sweep loop the O(k*n) concern is moot.

CelerData-Reviewer:
@codex review

kevincai (approved):

HangyuanLiu (approved):

PR #73666, opened 2026-05-21
[BugFix] Fix shared-data lake replication file-copy crashes

wxl24life:
## Why I'm doing:

The `REPLICATE_SNAPSHOT` agent task and the per-file copy sub-tasks were sharing the same thread pool. When the parallel-copy branch calls `ThreadPoolToken::wait()` from inside a worker of that pool, `ThreadPool::check_not_pool_thread_unlocked()` `LOG(FATAL)`s and `SIGABRT`s the CN. With the default `lake_replication_parallel_copy_min_file_count=2`, every replicate_snapshot task touching ≥ 2 files crashes the CN process.

## What I'm doing:

1. Forward `get_file_size` to the shard filesystem's `stat()` and return `Stat.size`, mirroring the existing `get_file_modified_time` path. The size comes directly from object-storage HEAD (S3 `ContentLength` / Azure `BlobSize` / HDFS `mSize`), so there is no dependency on the source cluster version or process state.

2. Add a dedicated `replicate_file` thread pool on `AgentServer`, sized by a new BE config `lake_replication_file_copy_threads`

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/agent/agent_server.cpp
```diff
@@ -370,6 +381,7 @@ void AgentServer::Impl::stop() {
         _thread_pool_clone->shutdown();
         _thread_pool_remote_snapshot->shutdown();
         _thread_pool_replicate_snapshot->shutdown();
+        _thread_pool_replicate_file->shutdown();
```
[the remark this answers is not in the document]
wxl24life: The unconditional shutdown matches the established pattern in this block — `_thread_pool_clone`, `_thread_pool_remote_snapshot`, and `_thread_pool_replicate_snapshot` immediately above are also dereferenced without a null guard (lines 380–382). Adding a guard only for `_thread_pool_replicate_file` would be inconsistent with its neighbors; guarding all of them is scope creep for this PR. If `init()` fails before the file-copy pool is built, the existing pools above would have already SIGABRT'd on the same code path, so the additional guard wouldn't change the observable behavior. Happy to revisit as a separate cleanup if a maintainer thinks it's worth it.

be/src/agent/agent_task.cpp
```diff
@@ -1103,8 +1103,7 @@ void run_replicate_snapshot_task(const std::shared_ptr<ReplicateSnapshotAgentTas
 
     Status res;
     if (replicate_snapshot_req.tablet_type == TTabletType::TABLET_TYPE_LAKE) {
-        res = exec_env->lake_replication_txn_manager()->replicate_snapshot(replicate_snapshot_req,
-                                                                           replicate_snapshot_pool);
+        res = exec_env->lake_replication_txn_manager()->replicate_snapshot(replicate_snapshot_req, replicate_file_pool);
```
xiangguangyxg: Should the `replicate_snapshot` task continue to use `replicate_snapshot_pool`, while internal file copies use a newly added `replicate_file_pool`?

wxl24life: Yes — outer REPLICATE_SNAPSHOT agent task still runs on `replicate_snapshot_pool` (`agent_server.cpp:635`, resolved via `get_thread_pool(REPLICATE_SNAPSHOT)`, unchanged); only the per-file copies inside `replicate_lake_remote_storage` use the new `replicate_file_pool`. Added inline `/*param=*/` hints at both call sites (this commit) so the role of each pool is obvious without reading the callee signatures.

CelerData-Reviewer:
@codex review

xiangguangyxg (dismissed):

docs/en/administration/management/BE_parameters/shared_lake_other.md
```diff
@@ -402,6 +402,15 @@ This topic introduces the following types of BE configurations:
...
 - Introduced in: -
 
+### lake_replication_file_copy_threads
+
+- Default: 0
+- Type: Int
+- Unit: -
+- Is mutable: No
+- Description: Number of threads in the dedicated thread pool used by lake-to-lake (shared-data) cross-cluster replication for per-file copy. `0` means `cpu_cores * 4` (the same default semantics as `replication_threads`); negative values mean `-value * cpu_cores`. This pool is intentionally separate from the agent-task `replicate_snapshot` pool so that per-file copy sub-tasks can be awaited from the outer task without tripping the thread-pool self-deadlock guard. The pool is built once at startup and has no runtime resize hook, so CN restart is required to change its size.
+- Introduced in: -
```
kevincai: 1. need ja translation
2. introduced version: v4.1.2

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

PR #73674, opened 2026-05-22
[BugFix] FILES()/LOAD respect Parquet isAdjustedToUTC=false for INT64 timestamps

### LINKED ISSUE #73672: [BUG] SELECT FROM FILES() shifts INT64 Parquet timestamps with isAdjustedToUTC=false by session timezone offset

The Arrow-based read path used by `SELECT FROM FILES()` (and broker/stream LOAD)
shifts INT64 Parquet timestamps written with `isAdjustedToUTC=false` by the
session timezone offset, breaking the round-trip even when StarRocks itself
both wrote and read the file.

1. **Write** — Let StarRocks write a Parquet file via `INSERT INTO FILES()`.
   StarRocks emits INT64 + `Timestamp(isAdjustedToUTC=false, MICROS)` by default.

   ```sql
   SET time_zone = 'Asia/Seoul';
   INSERT INTO FILES(
     "path"           = "s3://<bucket>/tz-rt/",
     "format"         = "parquet",
     "compression"    = "uncompressed",
     "aws.s3.endpoint"= "<endpoint>",
     "aws.s3.access_key" = "<ak>",
     "aws.s3.secret_key" = "<sk>",
     "aws.s3.enable_path_style_access" = "true",
     "aws.s3.region"  = ""
   )
   SELECT 1 AS id, CAST('2026-05-21 15:30:00' AS DATETIME) AS ts;
   ```

2. **Read** — `SELECT FROM FILES()` against the same file in three different
   session timezones:

   ```sql
   SET time_zone = 'Asia/Seoul';
   SELECT * FROM FILES("path"="s3://<bucket>/tz-rt/*", "format"="parquet", /* same s3 props */);
   -- → 2026-05-22 00:30:00   (+9h shift)

   SET time_zone = 'Asia/Shanghai';
   SELECT * FROM FILES(...);
   -- → 2026-05-21 23:30:00   (+8h shift)

   SET time_zone = 'UTC';
   SELECT * FROM FILES(...);
   -- → 2026-05-21 15:30:00   (identity, the only correct result)

   SET time_zone = 'America/Los_Angeles';
   SELECT * FROM FILES(...);
   -- → 2026-05-21 08:30:00   (-7h shift)
   ```

3. **File on disk is spec-compliant** — pyarrow inspection of the file
   StarRocks wrote in step 1:

   ```
   schema:
     id int8
     ts timestamp[us]                     # Arrow surface: no tz attached
   parquet logical type:
     ts: Timestamp(isAdjustedToUTC=false, timeUnit=microseconds)
   row_group[0].ts.min / .max:
     2026-05-21 15:30:00
   raw INT64 (row 0): 1779377400000000
     # = 1779377400 s = 2026-05-21 15:30:00 (wall-clock, NOT UTC of KST 15:30)
   created_by: parquet-cpp-arrow version 19.0.1 starrocks-4.0.6-45fdede
   ```

   The raw integer matches the wall-clock interpretation, so the WRITE path is
   spec-compliant. The shift is introduced on READ.

Per the Parquet TIMESTAMP logical type spec
(https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#timestamp):

> If the *isAdjustedToUTC* attribute is set to `true`, the value is the time
> elapsed [...] since the Unix epoch. [...] If *isAdjustedToUTC* is `false`,
> the value is the [...] number of timeunits elapsed since some unspecified
> instant. [...] The reader of an unadjusted timestamp **should not apply any
> timezone conversion**.

`SELECT * FROM FILES()` must therefore return `2026-05-21 15:30:00`
independent of session `time_zone`, matching Trino, Spark, Impala, Apache
Arrow, and DuckDB, all of which read timezone-naive Parquet timestamps as
wall-clock values without any session-tz shift.

### Real behavior (Required)

| session `time_zone`               | output                | delta from raw wall-clock |
|-----------------------------------|-----------------------|---------------------------|
| `Asia/Seoul` (UTC+9)              | `2026-05-22 00:30:00` | **+9 h**                  |
| `Asia/Shanghai` (UTC+8)           | `2026-05-21 23:30:00` | **+8 h**                  |
| `UTC`                             | `2026-05-21 15:30:00` | 0 (identity)              |
| `America/Los_Angeles` (UTC-7 PDT) | `2026-05-21 08:30:00` | **-7 h**                  |

The output drifts with the session timezone offset, i.e. the reader applies
`(session_tz - UTC)` to the raw INT64 even though `isAdjustedToUTC=false`. The
result is identity only when the session happens to be UTC.

The same path is used by broker/stream LOAD via `parquet_scanner`, so loads
into native tables are also shifted.

- Repro cluster: **4.0.6** (`select current_version()` → `4.0.6-45fdede`)
- Code path on upstream `main`:
  `be/src/column/arrow/arrow_to_starrocks_converter.cpp` lines 145 & 874
  (introduced by #50448 in 2024-08, partially guarded but not fixed by
  #73460). All 3.x / 4.x branches that include #50448 are affected.

### Additional context

Cross-engine comparison on the same file written by StarRocks:

| engine          | reads as                            | shift    |
|-----------------|-------------------------------------|----------|
| Trino 470       | `2026-05-21 15:30:00`               | none     |
| Spark 4.0       | `2026-05-21 15:30:00`               | none     |
| Impala 4.4      | `2026-05-21 15:30:00`               | none     |
| Apache Arrow 19 | `2026-05-21 15:30:00`               | none     |
| DuckDB 1.1      | `2026-05-21 15:30:00`               | none     |
| StarRocks 4.0.6 | `2026-05-22 00:30:00` (KST session) | **+9 h** |

ronjay1827:
## Why I'm doing

`SELECT FROM FILES()` and broker/stream LOAD shift INT64 Parquet timestamps written with `isAdjustedToUTC=false` (Arrow surfaces these as timezone-naive, i.e. empty `arrow::TimestampType::timezone()`) by the session timezone offset. The output is identity only when the session is UTC; otherwise it drifts by `(session_tz - UTC)`, breaking the round-trip even when StarRocks itself both writes and reads the file.

Per [Parquet `LogicalTypes.md` — TIMESTAMP](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#timestamp): a timezone-naive timestamp "must be displayed the same way, regardless of the local time zone in effect." Trino, Spark, Impala, Apache Arrow, and DuckDB all honor this and read the same file as the wall-clock value with no shift.

History:

- PR #50448 (2024-08) introduced the unconditional overwrite of `mutable_timezone` with the session tz in `convert_arrow_array_to_column` to fix a complex-type case for timezone-aware timestamps.
- PR #73460 added `conv_ctx != nullptr && !conv_ctx->timezone.empty()` guards, but `conv_ctx->timezone` is always populated with the session tz, so the guards never tripped on the normal SELECT path and the naive case kept shifting.

Reproduction on StarRocks 4.0.6 (`INSERT INTO FILES` → `SELECT FROM FILES` on the same Parquet file, value `'2026-05-21 15:30:00'`):

| session `time_zone`               | output                | delta from wall-clock |
|-----------------------------------|-----------------------|------------------------|
| `Asia/Seoul` (UTC+9)              | `2026-05-22 00:30:00` | **+9 h**               |
| `Asia/Shanghai` (UTC+8)           | `2026-05-21 23:30:00` | **+8 h**               |
| `UTC`                             | `2026-05-21 15:30:00` | 0 (identity)           |
| `America/Los_Angeles` (UTC-7 PDT) | `2026-05-21 08:30:00` | **-7 h**               |

Full repro (raw INT64 dump + pyarrow metadata + cross-engine comparison): #73672.

## What I'm doing

Restrict the session-timezone rewrite in the Arrow-based read path to timezone-aware columns only.

- `be/src/column/arrow/arrow_to_starrocks_converter.cpp` — `convert_arrow_array_to_column`: add `!mutable_timezone.empty()` to the existing guard so only `isAdjustedToUTC=true` Arrow timestamps (non-empty Arrow tz) are rectified to the session tz. Preserves #50448's complex-type fix while restoring spec compliance for naive timestamps.
- `be/src/column/arrow/arrow_to_starrocks_converter.cpp` — `ArrowConverter<TIMESTAMP, DATETIME>::apply`: when the Arrow tz is empty (`isAdjustedToUTC=false`), use `"UTC"` so `cctz` applies offset 0 and the wall-clock value round-trips byte-for-byte. Drops the `ctx->timezone` fallback that re-introduced the session-tz shift. Matches Trino's `DateTimeZone.UTC` pattern in `ColumnReaderFactory.java` for the same case.

Behaviour for timezone-aware columns (`isAdjustedToUTC=true`, e.g. Arrow tz = `"UTC"`) is unchanged — the existing session-tz rectification still applies.

### INT96 rework

The naive carve-out above would regress legacy INT96 timestamps on its own: the writer normalizes INT96 to a UTC instant (`level_builder.cpp` subtracts the offset for INT96), but Arrow decodes INT96 as timezone-naive — byte-identical to an INT64 `isAdjustedToUTC=false` column — so the no-shift rule would wrongly stop shifting INT96 to the session timezone (caught by `ParquetScannerTest.int96_timestamp` in CI).

- `be/src/exec/parquet_reader.cpp` — `_init_parquet_reader`: record top-level columns whose Parquet physical type is INT96 in `_int96_columns`.
- `be/src/exec/parquet_reader.cpp` — `_rectify_int96_timezone()` (called from `get_batch`): re-tag those columns' Arrow type from timezone-naive to `"UTC"` before conversion, so the converter still rectifies INT96 to the session timezone, matching the native reader's `Int96ToDateTimeConverter`. Only the Arrow type metadata changes; the underlying buffers are aliased unchanged (`data->Copy()` copies the `ArrayData` descriptor, not the buffers), and the `RecordBatch` is rebuilt only when a column was actually re-tagged.

**Out of scope — nested INT96.** `_int96_columns` records top-level columns only (`dot_vector.size() == 1`). An INT96 timestamp nested inside a STRUCT/ARRAY/MAP still decodes as naive and is read as wall-clock, diverging from the native reader for that case. This is a deliberate limitation: INT96 has been deprecated since Parquet 2.0 and nested occurrences are rare. Documenting it here so it is an explicit decision rather than a silent gap.

**Why no compatibility flag.** The shift removal is unconditional by design: wall-clock preservation for `isAdjustedToUTC=false` is what the Parquet spec mandates and what Trino/Spark/Impala/Arrow/DuckDB already do, so correctness is not made opt-in. There is also no performance trade-off to gate — the INT96 re-tag is metadata-only (no per-row work). If a user turns out to depend on the old shifted reads, a session-variable escape hatch can be added in a follow-up PR.

### Docs

`docs/{en,zh,ja}/sql-reference/sql-functions/table-functions/files.md` — document the read-side mapping of Parquet TIMESTAMP to DATETIME (instant vs. local semantics by `isAdjustedToUTC`; top-level INT96 treated as a UTC instant, nested INT96 read as wall-clock), with a behavior-change note that earlier versions shifted local-semantics values by the session offset. Wording mirrors the existing unloading-side semantics note in the same file.

Suggested release-note bullet for the per-version `[Doc]` release-notes PRs (3.5.x / 4.0.x / 4.1.x):

> `FILES()` and file-based loading now preserve the wall-clock value when reading timezone-naive Parquet `TIMESTAMP` (`INT64` with `isAdjustedToUTC = false`) columns, per the Parquet specification. Previously a session time zone offset was applied, so returned values could differ from what was written. Columns with `isAdjustedToUTC = true` and top-level legacy INT96 timestamps are unaffected and continue to be converted to the session time zone. [#73674](https://github.com/StarRocks/starrocks/pull/73674)

## Tests

- `be/test/column/arrow_to_starrocks_converter_test.cpp` — `test_timestamp_convert_naive_preserves_wall_clock`: asserts identity for session tz in `{UTC, Asia/Shanghai, Asia/Seoul, America/Los_Angeles}` against a timezone-naive Arrow TIMESTAMP array (raw 0 → `1970-01-01 00:00:00` regardless of session).
- `be/test/exec/file_scanner/parquet_scanner_test.cpp` — `int96_timestamp_session_timezone_shift`: reads the existing `int96_timestamp.parquet` through the full `ParquetScanner` → `ParquetReaderWrap` → arrow converter path under a UTC session (baseline) and a fixed `-08:00` session, asserting the INT96 instants shift by the session offset and explicitly differ from the UTC render (`EXPECT_NE`). Fails if `_rectify_int96_timezone()` is removed — locks in the INT96 no-regression guarantee.
- **Test-harness fix (found while adding the test above):** `create_parquet_scanner` assigned `query_globals.time_zone` without the thrift setter, so `__isset.time_zone` stayed false and `RuntimeState` silently fell back to the default `+08:00` — the timezone argument never took effect and every "UTC" scanner test actually ran under `+08:00`. Switched to `query_globals.__set_time_zone(...)` (the pattern used by `time_functions_test.cpp`) and corrected the baselines that had encoded `+08:00` renders mislabeled as UTC: `int96_timestamp` (`23:59:59.009999`/`15:04:05` → true-UTC `15:59:59.009999`/`07:04:05`, verified against the file's raw instants via pyarrow) and `datetime` (`15:04:05`* → `07:04:05`*, the file stores `isAdjustedToUTC=true` instants written as `15:04:05` Asia/Shanghai). Without this fix a non-UTC INT96 test is impossible — both sessions silently render identically.
- Existing `test_timestamp_convert_uses_context_timezone` still passes as a regression guard for the timezone-aware case (`arrow::TimestampType(SECOND, "UTC")` + session `Asia/Shanghai` → `1970-01-01 08:00:00`), confirming #50448 / #73460 behaviour is preserved.
- Existing `ParquetScannerTest.int96_timestamp` now genuinely guards the UTC render of INT96 (corrected baselines above).

Fixes #73672

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/column/arrow/arrow_to_starrocks_converter.cpp
+++ b/be/src/column/arrow/arrow_to_starrocks_converter.cpp
@@ -136,13 +136,17 @@ Status convert_arrow_array_to_column(ConvertFuncTree* conv_func, size_t num_elem
                                              chunk_start_idx, chunk_filter, conv_ctx);
     }
 
-    // for timestamp type, state->timezone which is specified by user. convert function
-    // obtains timezone from array. thus timezone in array should be rectified to
-    // state->timezone.
+    // Per Parquet LogicalTypes.md TIMESTAMP section, an INT64 timestamp with
+    // isAdjustedToUTC=false is surfaced by Arrow as timezone-naive (empty
+    // `timezone`, see arrow/type.h TimestampType docstring) and must be
+    // displayed identically regardless of session timezone. Only override the
+    // column timezone when it is already timezone-aware (isAdjustedToUTC=true
+    // -> non-empty Arrow tz). This preserves PR #50448's fix for complex-type
+    // timezone-aware timestamps while restoring spec compliance for naive ones.
     if (array->type_id() == ArrowTypeId::TIMESTAMP) {
         auto* timestamp_type = down_cast<arrow::TimestampType*>(array->type().get());
         auto& mutable_timezone = (std::string&)timestamp_type->timezone();
-        if (conv_ctx != nullptr && !conv_ctx->timezone.empty()) {
+        if (conv_ctx != nullptr && !conv_ctx->timezone.empty() && !mutable_timezone.empty()) {
             mutable_timezone = conv_ctx->timezone;
         }
     }
@@ -859,21 +863,16 @@ struct ArrowConverter<AT, LT, is_nullable, is_strict, DateOrDateTimeATGuard<AT>,
             }
         } else if constexpr (at_is_datetime<AT>) {
             auto timezone = concrete_type->timezone();
-            if (timezone.empty()) {
-                // Quote from https://github.com/apache/arrow/blob/4743e181596b9ee45c6b063bcf59fdf9eb72418f/cpp/src/arrow/type.h#L1217
-
-                /// If a TimestampType is constructed without a timezone (or, equivalently, if the
-                /// timezone supplied is an empty string) then the resulting Arrow field (column) is
-                /// considered "timezone-naive".  The producer of a timezone-naive column may populate
-                /// its constituent integer arrays with datetime values from any timezone; the consumer
-                /// of a timezone-naive column should make no assumptions about the interoperability or
-                /// comparability of the values of such a column with those of any other timestamp
-                /// column or datetime value.
-
-                // When the parquet timezone is empty, populate data with runtime timezone instead.
-                if (ctx != nullptr && !ctx->timezone.empty()) {
-                    timezone = ctx->timezone;
-                }
+            const bool is_timezone_naive = timezone.empty();
+            if (is_timezone_naive) {
+                // Arrow timezone-naive = Parquet isAdjustedToUTC=false. Per
+                // Parquet LogicalTypes.md TIMESTAMP, the value must be
+                // displayed "the same way, regardless of the local time zone
+                // in effect." Use UTC so the cctz lookup yields offset=0 and
+                // convert_datetime becomes a no-op for the wall-clock case.
+                // Matches Trino's `DateTimeZone.UTC` pattern in
+                // ColumnReaderFactory.java for the same case.
+                timezone = "UTC";
             }
 
             cctz::time_zone ctz;
--- a/be/src/exec/parquet_reader.cpp
+++ b/be/src/exec/parquet_reader.cpp
@@ -143,8 +143,20 @@ Status ParquetReaderWrap::_init_parquet_reader() {
             auto parquet_schema = _file_metadata->schema();
             for (int i = 0; i < parquet_schema->num_columns(); i++) {
                 auto column_desc = parquet_schema->Column(i);
-                std::string field_name = column_desc->path()->ToDotVector()[0];
+                const auto dot_vector = column_desc->path()->ToDotVector();
+                const std::string& field_name = dot_vector[0];
                 _map_column_nested[field_name].push_back(i);
+                // Record top-level INT96 columns so _rectify_int96_timezone() can re-tag them
+                // as UTC instants. Arrow surfaces INT96 as timezone-naive, identical to an
+                // INT64 isAdjustedToUTC=false wall-clock column, so they are otherwise
+                // indistinguishable downstream.
+                // INT96 leaves nested inside complex types (dot_vector.size() > 1) are
+                // deliberately not re-tagged: INT96 is a legacy physical type and rarely
+                // nested, and re-tagging would require rebuilding nested ArrayData trees.
+                // Such leaves are read as wall-clock (no session-timezone shift).
+                if (dot_vector.size() == 1 && column_desc->physical_type() == ::parquet::Type::INT96) {
+                    _int96_columns.insert(field_name);
+                }
             }
         }
 
@@ -314,7 +326,45 @@ Status ParquetReaderWrap::read_record_batch(const std::vector<SlotDescriptor*>& 
     return Status::OK();
 }
 
+void ParquetReaderWrap::_rectify_int96_timezone() {
+    // _batch is guaranteed non-null by the caller: get_batch() is only reached after a
+    // successful read_record_batch()/next_batch(), which always populates _batch.
+    if (_int96_columns.empty()) {
+        return;
+    }
+    auto fields = _batch->schema()->fields();
+    auto columns = _batch->columns();
+    bool changed = false;
+    for (size_t i = 0; i < fields.size(); ++i) {
+        if (fields[i]->type()->id() != arrow::Type::TIMESTAMP) {
+            continue;
+        }
+        if (_int96_columns.find(fields[i]->name()) == _int96_columns.end()) {
+            continue;
+        }
+        auto ts_type = std::static_pointer_cast<arrow::TimestampType>(fields[i]->type());
+        if (!ts_type->timezone().empty()) {
+            continue;
+        }
+        // INT96 is a UTC-normalized instant but Arrow decodes it as timezone-naive. Tag it as
+        // UTC so the downstream converter overrides it with the session timezone and shifts the
+        // value (matching the native parquet reader's Int96ToDateTimeConverter), instead of
+        // treating it as a wall-clock INT64 isAdjustedToUTC=false column. The underlying int64
+        // buffer is shared unchanged; only the Arrow type metadata gains the "UTC" zone.
+        auto utc_type = arrow::timestamp(ts_type->unit(), "UTC");
+        auto data = columns[i]->data()->Copy();
+        data->type = utc_type;
+        columns[i] = arrow::MakeArray(data);
+        fields[i] = fields[i]->WithType(utc_type);
+        changed = true;
+    }
+    if (changed) {
+        _batch = arrow::RecordBatch::Make(arrow::schema(fields), _batch->num_rows(), columns);
+    }
+}
+
 const std::shared_ptr<arrow::RecordBatch>& ParquetReaderWrap::get_batch() {
+    _rectify_int96_timezone();
     _current_line_of_batch += _batch->num_rows();
     _current_line_of_group += _batch->num_rows();
     return _batch;
--- a/be/test/column/arrow_to_starrocks_converter_test.cpp
+++ b/be/test/column/arrow_to_starrocks_converter_test.cpp
@@ -1248,6 +1248,34 @@ PARALLEL_TEST(ArrowConverterTest, test_timestamp_convert_uses_context_timezone) 
     ASSERT_EQ(column->get_data()[0], string_to_datetime<TYPE_DATETIME>("1970-01-01 08:00:00"));
 }
 
+// Per Parquet LogicalTypes.md TIMESTAMP section, an INT64 timestamp with
+// isAdjustedToUTC=false is surfaced by Arrow as timezone-naive (empty
+// `timezone`) and must be displayed "the same way, regardless of the local
+// time zone in effect." Confirms that the convert path no longer applies the
+// session timezone offset to such columns (the bug introduced by PR #50448
+// and only partially guarded by PR #73460).
+PARALLEL_TEST(ArrowConverterTest, test_timestamp_convert_naive_preserves_wall_clock) {
+    auto type = std::make_shared<arrow::TimestampType>(arrow::TimeUnit::SECOND, "");
+    size_t counter = 0;
+    auto array = create_constant_datetime_array<arrow::TimestampType, false>(1, 0, type, counter);
+    ConvertFuncTree cf(get_arrow_converter(ArrowTypeId::TIMESTAMP, TYPE_DATETIME, false, false));
+    ASSERT_TRUE(cf.func != nullptr);
+
+    // The raw value (0) is 1970-01-01 00:00:00 in wall-clock semantics.
+    // No matter what session timezone is set, the result must be identical.
+    for (const auto* session_tz : {"UTC", "Asia/Shanghai", "Asia/Seoul", "America/Los_Angeles"}) {
+        SCOPED_TRACE(std::string("session_tz=") + session_tz);
+        ArrowConvertContext ctx;
+        ctx.timezone = session_tz;
+        auto column = TimestampColumn::create();
+        Filter filter(1, 1);
+        ASSERT_STATUS_OK(
+                convert_arrow_array_to_column(&cf, array->length(), array.get(), column.get(), 0, 0, &filter, &ctx));
+        ASSERT_EQ(column->get_data()[0], string_to_datetime<TYPE_DATETIME>("1970-01-01 00:00:00"))
+                << "naive timestamp must not be shifted by session timezone";
+    }
+}
+
 template <bool is_nullable>
 std::shared_ptr<arrow::Array> create_const_decimal_array(size_t num_elements,
                                                          const std::shared_ptr<arrow::Decimal128Type>& type,
--- a/be/src/exec/parquet_reader.h
+++ b/be/src/exec/parquet_reader.h
@@ -28,6 +28,7 @@
 #include <cstdint>
 #include <map>
 #include <string>
+#include <unordered_set>
 
 #include "column/vectorized_fwd.h"
 #include "common/status.h"
@@ -83,6 +84,11 @@ private:
     Status next_selected_row_group();
     // _init_parquet_reader initializes the underlying parquets reader.
     Status _init_parquet_reader();
+    // Re-tag INT96 timestamp columns of the current _batch as UTC. Legacy INT96 carries no
+    // isAdjustedToUTC flag and Arrow decodes it as timezone-naive, indistinguishable from an
+    // INT64 isAdjustedToUTC=false (wall-clock) column. Tagging INT96 as UTC keeps its instant
+    // semantics so the converter shifts it to the session timezone, matching the native reader.
+    void _rectify_int96_timezone();
 
     const int32_t _num_of_columns_from_file;
     int64_t _num_rows = 0;
@@ -97,6 +103,8 @@ private:
 
     // For nested column type, it's consisting of multiple physical-columns
     std::map<std::string, std::vector<int>> _map_column_nested;
+    // Top-level column names whose parquet physical type is INT96 (see _rectify_int96_timezone).
+    std::unordered_set<std::string> _int96_columns;
     std::vector<int> _parquet_column_ids;
     int _total_groups{0}; // groups in a parquet file
     int _current_group{0};
--- a/be/test/exec/file_scanner/parquet_scanner_test.cpp
+++ b/be/test/exec/file_scanner/parquet_scanner_test.cpp
@@ -137,7 +137,10 @@ private:
             query_options.__set_batch_size(chunk_size);
         }
         auto query_globals = TQueryGlobals();
-        query_globals.time_zone = timezone;
+        // Use the thrift setter: a plain member assignment leaves __isset.time_zone false, so
+        // RuntimeState silently falls back to the default time zone (+08:00) and the `timezone`
+        // argument never takes effect.
+        query_globals.__set_time_zone(timezone);
         RuntimeState* state = _obj_pool.add(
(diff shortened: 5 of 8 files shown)


CelerData-Reviewer:
@codex review

ronjay1827:
Fixes the ParquetScannerTest.int96_timestamp regression in c227e8174e0.

Arrow surfaces INT96 as timezone-naive, the same as INT64 isAdjustedToUTC=false, so the naive no-shift carve-out wrongly stopped shifting INT96 (a UTC instant). The fix re-tags top-level INT96 columns as UTC in ParquetReaderWrap::get_batch so they shift to the session timezone matching the native Int96ToDateTimeConverter while genuinely naive INT64 isAdjustedToUTC=false columns stay unshifted. The converter change in this PR is unchanged. Thank You @kevincai

CelerData-Reviewer:
@codex review

kevincai (commented):
Reviewed the correctness end-to-end (not just the diff): I traced read routing and the write path to confirm the fix actually round-trips rather than relocating the drift.

- Routing is correct: `FILES()` reads map to `FileScanNode` -> `FileConnector` (builds `ParquetScanner`) -> `ParquetReaderWrap` -> `convert_arrow_array_to_column`. So both `FILES()` and broker/stream LOAD genuinely hit this path; the title scope is accurate.
- Round-trip verified against the writer: `file_writer.cpp` emits INT64 as `LogicalType::Timestamp(false, MICROS)` and `level_builder.cpp` writes the wall-clock value with no offset for INT64. So old read = shift = drift; new read = no shift = identity. Correct.
- INT96 rework is necessary, not incidental: the writer subtracts the offset for INT96 (it is a UTC instant), and Arrow decodes it as naive. Without `_rectify_int96_timezone()` the naive fix would regress INT96 by dropping its shift. Good catch.
- Consistency restored: the native reader (`column_converter.cpp`) already treats naive as no-shift; this makes the Arrow path agree.

Verdict: correct and well-reasoned. Two things should be addressed before merge (see inline): the undocumented + untested INT96 rework, and a release note given the three-branch backport. Nested INT96 is an acceptable but undeclared limitation. Inline comments have the detail.

be/src/exec/parquet_reader.cpp
```diff
@@ -314,7 +322,43 @@ Status ParquetReaderWrap::read_record_batch(const std::vector<SlotDescriptor*>&
     return Status::OK();
 }
 
+void ParquetReaderWrap::_rectify_int96_timezone() {
```
kevincai, line 329: **P1 - undocumented and untested.** This entire INT96 rework (`_rectify_int96_timezone()`, the `_int96_columns` set, the schema rebuild) is not mentioned in the PR description, and it is the riskiest code in the PR: it rebuilds the `RecordBatch`, aliases buffers via `data->Copy()`, and mutates `_batch`. There is no test for it. Please (1) update the PR body to cover the INT96 changes, and (2) add a BE test for an INT96 Parquet column round-tripping under a non-UTC session, asserting the session-tz shift is still applied. That locks in the no-regression claim that this helper exists to guarantee.

be/src/exec/parquet_reader.cpp
```diff
@@ -143,8 +143,16 @@ Status ParquetReaderWrap::_init_parquet_reader() {
...
                 auto column_desc = parquet_schema->Column(i);
-                std::string field_name = column_desc->path()->ToDotVector()[0];
+                const auto dot_vector = column_desc->path()->ToDotVector();
+                const std::string& field_name = dot_vector[0];
                 _map_column_nested[field_name].push_back(i);
+                // Record top-level INT96 columns so _rectify_int96_timezone() can re-tag them
+                // as UTC instants. Arrow surfaces INT96 as timezone-naive, identical to an
+                // INT64 isAdjustedToUTC=false wall-clock column, so they are otherwise
+                // indistinguishable downstream.
+                if (dot_vector.size() == 1 && column_desc->physical_type() == ::parquet::Type::INT96) {
```
kevincai, line 157: **P2 - nested INT96 silently not handled.** `_int96_columns` only records top-level columns (`dot_vector.size() == 1`). An INT96 timestamp nested inside a struct/list/map will still be decoded as naive and read as wall-clock (no shift), diverging from the native `Int96ToDateTimeConverter`. Acceptable given INT96 is legacy and rarely nested, but please state explicitly in the PR description that nested INT96 is out of scope so it is a deliberate decision rather than a silent gap.

be/src/exec/parquet_reader.cpp
```diff
@@ -314,7 +322,43 @@ Status ParquetReaderWrap::read_record_batch(const std::vector<SlotDescriptor*>&
...
+        changed = true;
+    }
+    if (changed) {
+        _batch = arrow::RecordBatch::Make(arrow::schema(fields), _batch->num_rows(), columns);
+    }
+}
+
 const std::shared_ptr<arrow::RecordBatch>& ParquetReaderWrap::get_batch() {
+    _rectify_int96_timezone();
     _current_line_of_batch += _batch->num_rows();
```
kevincai, line 368: Minor: `_rectify_int96_timezone()` defends against `_batch == nullptr`, but the very next line dereferences `_batch->num_rows()` unconditionally. The `+=` deref is pre-existing so this is not a regression, but the asymmetry reads oddly - either drop the null check in the helper or guard the caller too, for consistency.

be/src/column/arrow/arrow_to_starrocks_converter.cpp
```diff
@@ -859,21 +863,16 @@ struct ArrowConverter<AT, LT, is_nullable, is_strict, DateOrDateTimeATGuard<AT>,
...
+            const bool is_timezone_naive = timezone.empty();
+            if (is_timezone_naive) {
+                // Arrow timezone-naive = Parquet isAdjustedToUTC=false. Per
+                // Parquet LogicalTypes.md TIMESTAMP, the value must be
+                // displayed "the same way, regardless of the local time zone
+                // in effect." Use UTC so the cctz lookup yields offset=0 and
+                // convert_datetime becomes a no-op for the wall-clock case.
+                // Matches Trino's `DateTimeZone.UTC` pattern in
+                // ColumnReaderFactory.java for the same case.
+                timezone = "UTC";
```
kevincai, line 875: **P1/P2 - behavior change needs a release note.** This is correctly labeled `behavior_changed` and backports to 3.5/4.0/4.1. After upgrade, any non-UTC session reading naive-timestamp Parquet (including third-party files whose producer intended session-local semantics) will see values shift by the session offset. The checklist leaves "needs user documentation" unchecked - for a spec-semantics change shipping to three release branches, please add a short release note / doc entry in `docs/en` and `docs/zh` so users are not surprised by changed values after upgrade.

be/test/column/arrow_to_starrocks_converter_test.cpp
```diff
@@ -1248,6 +1248,34 @@ PARALLEL_TEST(ArrowConverterTest, test_timestamp_convert_uses_context_timezone)
     ASSERT_EQ(column->get_data()[0], string_to_datetime<TYPE_DATETIME>("1970-01-01 08:00:00"));
 }
 
+// Per Parquet LogicalTypes.md TIMESTAMP section, an INT64 timestamp with
+// isAdjustedToUTC=false is surfaced by Arrow as timezone-naive (empty
+// `timezone`) and must be displayed "the same way, regardless of the local
+// time zone in effect." Confirms that the convert path no longer applies the
+// session timezone offset to such columns (the bug introduced by PR #50448
+// and only partially guarded by PR #73460).
+PARALLEL_TEST(ArrowConverterTest, test_timestamp_convert_naive_preserves_wall_clock) {
```
kevincai, line 1257: Good coverage for the naive INT64 case. Missing: a test for the INT96 path added in `parquet_reader.cpp` (re-tagged to UTC, must still shift under a non-UTC session) and ideally an explicit timezone-aware regression assertion alongside it. The INT96 rework is the riskiest part of this PR and currently has zero tests.

be/src/exec/parquet_reader.cpp
ronjay1827: Thanks — both addressed in bf5222145ff. (1) The PR description now has a dedicated "INT96 rework" section covering `_int96_columns`, `_rectify_int96_timezone()`, and the batch rebuild, plus an explicit out-of-scope statement for nested INT96. (2) Added `ParquetScannerTest.int96_timestamp_session_timezone_shift`: it drives the full `ParquetScanner` → `ParquetReaderWrap` → arrow-converter path against the existing `int96_timestamp.parquet` under both a UTC and a fixed `-08:00` session, asserting the instants shift by the session offset and differ from the UTC baseline (`EXPECT_NE`); it fails if `_rectify_int96_timezone()` is removed.

While adding it we found `create_parquet_scanner` never propagated its timezone argument (plain assignment left `__isset.time_zone` false, so `RuntimeState` silently fell back to the default `+08:00`). The harness now uses `__set_time_zone`, and the `int96_timestamp`/`datetime` baselines were corrected to true-UTC renders, verified against the files' raw instants via pyarrow (details in the Tests section of the description). Re-validated locally: `ParquetScannerTest.*` 21/21 and `ArrowConverterTest.*` 7/7 pass.

be/src/exec/parquet_reader.cpp
ronjay1827: Declared explicitly in bf5222145ff — in the PR description (What I'm doing → INT96 rework → out of scope), as a code comment at the `_int96_columns` recording site, and in the user docs (`files.md` now states nested INT96 is read as wall-clock with no session shift).

be/src/exec/parquet_reader.cpp
ronjay1827: Done in bf5222145ff — dropped the null check in the helper. `get_batch()` already relies on a non-null `_batch`, so the helper now documents that invariant in a comment instead of half-guarding it.

be/src/column/arrow/arrow_to_starrocks_converter.cpp
ronjay1827: Addressed in bf5222145ff. Added the read-side timezone semantics to the `FILES()` reference in docs/en, docs/zh, and docs/ja — instant vs. local semantics by `isAdjustedToUTC`, top-level INT96 as a UTC instant, nested INT96 as wall-clock — with a behavior-change note that earlier versions shifted local-semantics values by the session offset, and checked the user-documentation boxes.

For `release-3.5/4.0/4.1.md`: those files appear to be written by dedicated per-version `[Doc]` PRs at release time (e.g. #73026, #71791), so instead of editing them here I put a suggested release-note bullet in the PR description for the maintainers to pick up — happy to move it into the release-notes files directly if you prefer.

be/test/column/arrow_to_starrocks_converter_test.cpp
ronjay1827: Added `ParquetScannerTest.int96_timestamp_session_timezone_shift` in bf5222145ff (details in the `parquet_reader.cpp` thread) — it covers the INT96 path end-to-end including the re-tag, with the explicit timezone-aware assertion alongside the UTC baseline. I kept it in `parquet_scanner_test.cpp` rather than this converter test because the re-tagging happens in `ParquetReaderWrap`, which the converter unit tests never construct.

be/src/exec/parquet_reader.cpp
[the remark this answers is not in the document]
ronjay1827: Correct — INT96 leaves inside complex columns are not re-tagged. As of bf5222145ff this is an explicit, documented limitation (PR description, a code comment at the recording site, and the `files.md` docs) rather than a silent gap; the native reader keeps its existing behavior for that case.

ronjay1827:
Done.  rebased onto main. Thank You @kevincai

CelerData-Reviewer:
@codex review

kevincai (approved):

wyb (approved):

EsoragotoSpirit:
@ronjay1827 Please involve the documentation team in the review when your PR has changes to documents.

ronjay1827:
Thanks for the heads-up @EsoragotoSpirit . This PR updates the FILES() docs (en/ja/zh). Could the documentation team please review the doc changes?
(cc @DanRoscigno)

PR #73686, opened 2026-05-22
[BugFix] Wire physical rowids through PK index publish for shared segments

xiangguangyxg:
## Why I'm doing:

Tablet split in shared-data mode lets a child tablet share the parent's segment files. The publish path reads such shared segments via a range-pruned `ChunkIterator` that starts at a non-zero physical offset in the source segment file, while reload reads back by physical rowid. Publish was encoding each row's segment position as `chunk_row_offset_in_segment + i_in_chunk`, which equals the chunk-local index for non-shared segments (where the iterator starts at row 0) but skips the range-start offset for shared ones — producing wrong `rss_rowid` entries and `PersistentIndexMemtable<12> insert found duplicate key` errors on subsequent publish.

This PR aligns publish and reload on physical rowids end-to-end in shared-data publish, condition update, and column-mode partial update paths.

## What I'm doing:

`SegmentPKIterator` captures the segment's physical rowid base (= the iterator's `range_start`) from the first emit's `rowid_buffer.front()` into `_physical_rowid_base` (`std::optional<uint32_t>` so `0` is distinguishable from "unset"). The base is set once per segment; subsequent `_load()` calls fall through to the plain `get_next(chunk)` form — no per-emit rowid-buffer overhead. The base is exposed read-only via `physical_rowid_base()`.

Each `SegmentPKChunkRef` carries one field — `physical_rowid_offset`, chunk[0]'s position in the source segment file (= base + rows-emitted-before-this-chunk). Downstream consumers (`parallel_upsert`, condition update, `_handle_column_upsert_mode`) use it directly; chunk[i]'s physical rowid is `physical_rowid_offset + i`. For non-shared segments the base is 0, so behavior is unchanged.

**Condition update** (`_do_update_with_condition`) snapshots `chunk_physical_rowid_offset` per chunk during the parallel compare phase. In phase 2 it materializes `winner_rowids = chunk_physical_rowid_offset + winner_local_indices[k]` for the (possibly fragmented) winners and submits via the existing `PrimaryIndex::upsert(rssid, const vector<uint32_t>& rowids, pks, stat, ctx)` overload. `PrimaryIndex` itself is unmodified — it treats rowids as opaque uint32 values; no `physical_` terminology leaks into the basic component.

**Column-mode partial update** (`ColumnModePartialUpdateHandler::_load_update_state`) reads each update segment's base via `pk_iters[i]->physical_rowid_base()` after consuming the iterator and passes it explicitly to `ColumnPartialUpdateState::build_rss_rowid_to_update_rowid(uint32_t base)`. The build step keeps `rss_rowid_to_update_rowid` as logical offsets (consumed by `_update_source_chunk_by_upt` via `append_selective` on the iterator-emitted chunk) while translating `insert_rowids` to physical positions (consumed via full-file `fetch_values_by_rowid`). The two non-lake callers pass `/*base=*/0` explicitly — the default was removed so a future shared-segment caller forgetting the base becomes a compile error instead of silent base-0 corruption.

**Contiguity contract** — the design relies on each emitted chunk's physical rowids forming a single contiguous run (so `physical_rowid_offset + i` is well-defined). This is enforced structurally by 8 `DCHECK`s at `Rowset::get_each_segment_iterator` entry on the `SegmentReadOptions` fields that would introduce sparsity: `delvec_loader`, `delete_predicates`, `pred_tree`, `pred_tree_for_zone_map`, `runtime_filter_preds`, `rowid_range_option`, `ranges`, `short_key_ranges`. Documented contract that delvec-aware reads must use a separate iterator factory.

**Auto-increment conflict resolution** drops a now-unreachable `UINT32_MAX` sentinel and an associated `DCHECK` (the dense/physical mismatch case it defended against can no longer arise).

## Notes on the redesign

The first iteration threaded a `shared_ptr<vector<uint32_t>> physical_rowids` through five layers (~800 lines of plumbing). After confirming the underlying iterator only does range filtering on PK-sorted segments — so emitted rowids are always contiguous — the design was rebased to use a scalar base + per-chunk offset.

Following review feedback the iterator's segment base is now captured from `rowid_buffer.front()` on the **first emit only** (subsequent emits don't pay the rowid-buffer cost); the per-emit runtime contiguity validator was dropped in favor of the structural `DCHECK` wall at the iterator-construction boundary; the sparse `PrimaryIndex::upsert(rssid, base, indices, pks, ...)` overload added in an earlier round was reverted, keeping `PrimaryIndex` free of `physical_` terminology; and the iterator's member/accessor was renamed `_physical_rowid_offset` / `physical_rowid_offset()` → `_physical_rowid_base` / `physical_rowid_base()` to lexically distinguish the immutable segment-wide base from the per-chunk `SegmentPKChunkRef::physical_rowid_offset` that advances by chunk.

Net diff: 13 files, +377 / -86, behavior-preserving for non-shared segments.

## Verification

- `/codex-review:code` — APPROVED across multiple iterations.
- `/simplify` — converged with "no further simplifications".
- GitHub CI — BE Architecture / BUILD / BE UT all pass.
- TSP cluster end-to-end validation (shared-data, 3 CN): a 2M-row PK table with `ORDER BY(id)` was split into 16 sub-tablets, then exercised across five non-overlapping id ranges — untouched, full-row upsert, column-mode partial update of `c2`, column-mode partial update of `c3`, and delete — followed by manual `COMPACT`. Final comprehensive invariant check: 1,600,000 rows, 0 violations, 0 duplicate PKs, with the compaction pass re-verifying 0 violations across 12 successive checks. The directly-fixed `_handle_column_upsert_mode` path was confirmed correct: untouched columns survived intact through partial updates over shared post-split segments.

## Tests

- `be/test/storage/lake/segment_pk_iterator_test.cpp` (new, 5 tests): non-shared base=0, shared range filter (`physical_rowid_base() == range_start`), multi-emit single load (base set once on first emit), empty inner emit preserves base, lazy-load across `next()` calls (segment base unchanged while per-chunk `physical_rowid_offset` advances).
- Existing `condition_update_test` (in particular `test_condition_update_no_sst_parallel_fragmented_winners`) exercises the rowids-form `PrimaryIndex::upsert` end-to-end on fragmented winner ranges.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [ ] Parameter changes
- [ ] Policy changes
- [ ] Feature removed
- [ ] Miscellaneous

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/lake/column_mode_partial_update_handler.cpp
```diff
@@ -115,14 +115,26 @@ Status ColumnModePartialUpdateHandler::_load_update_state(const RowsetUpdateStat
 
     // Parallel query PK index: each segment's PKs are loaded chunk-by-chunk (lazy)
     // and each chunk is queried against the index in parallel via thread pool.
+    // Also collect each row's segment-local physical rowid so INSERT-row reads
+    // by fetch_values_by_rowid land on the right rows when the source iterator
+    // is range-pruned (sparse physical rowids).
     std::vector<std::vector<uint64_t>> rss_rowids_per_segment;
+    std::vector<std::vector<uint32_t>> physical_rowids_per_segment;
```
luohaha: What is `physical_` mean?

xiangguangyxg: for non-shared segments, logical rowid and physical rowid are the same.
for shared segments, logical rowid and physical rowid are different, logical rowid starts from picked data rows but physical rowid starts from the beginning of the whole segment data.

be/src/storage/lake/lake_primary_index.cpp
```diff
@@ -492,38 +510,66 @@ Status LakePrimaryIndex::parallel_get(ThreadPoolToken* token, SegmentPKIterator*
...
+            size_t num_rows = current.chunk->num_rows();
+            size_t begin_rowid = current.chunk_row_offset_in_segment;
+            // chunk_row_offset_in_segment is the dense per-segment index used to
+            // land this chunk's results at the right slot of the per-segment
+            // output vector; do NOT mistake it for a physical rowid.
+
+            if (physical_rowids_per_segment != nullptr) {
+                if (current.physical_rowids == nullptr || current.physical_rowids->size() != num_rows) {
+                    std::lock_guard<std::mutex> l(mutex);
+                    status.update(Status::InternalError(strings::Substitute(
```
luohaha: Why we need so much check?

luohaha: Can we consolidate this check into a single entry point? This will help avoid a lot of duplicate check code.

luohaha: And when this issue happen, could it be recoverable?

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/lake/column_mode_partial_update_handler.cpp
```diff
@@ -398,7 +404,18 @@ Status ColumnModePartialUpdateHandler::execute(const RowsetUpdateStateParams& pa
...
+            // segment file so downstream readers (e.g. fetch_values_by_rowid) land on
+            // the right rows. Base is 0 for non-shared segments (no-op). The output
+            // slot was just resized to empty above, and `_partial_update_states` is
+            // cleared right after this loop, so no clear() of either side is needed.
+            const auto& logical_offsets = _partial_update_states[upt_id].insert_rowids;
+            const uint32_t segment_base = _partial_update_states[upt_id].upt_segment_physical_rowid_base;
+            auto& physical_rowids_out = (*insert_rowids_by_segment)[upt_id];
+            physical_rowids_out.reserve(logical_offsets.size());
+            for (uint32_t logical_offset : logical_offsets) {
+                physical_rowids_out.push_back(segment_base + logical_offset);
```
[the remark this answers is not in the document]
xiangguangyxg: fixed

be/src/storage/lake/lake_primary_index.cpp
xiangguangyxg: done

be/src/storage/lake/rowset_update_state.cpp
```diff
@@ -121,8 +169,12 @@ void SegmentPKIterator::next() {
     }
 }
 
-std::pair<ChunkPtr, size_t> SegmentPKIterator::current() {
-    return std::pair<ChunkPtr, size_t>(std::move(_pk_column_chunk), _begin_rowid_offsets[_current_pk_column_idx]);
+SegmentPKChunkRef SegmentPKIterator::current() {
+    SegmentPKChunkRef ref;
+    ref.chunk_row_offset_in_segment = _begin_rowid_offsets[_current_pk_column_idx];
+    ref.physical_rowid_base = segment_physical_rowid_base() + static_cast<uint32_t>(ref.chunk_row_offset_in_segment);
```
luohaha: `chunk_row_offset_in_segment` and `physical_rowid_base` are very confused

be/src/storage/lake/rowset_update_state.h
```diff
@@ -97,8 +120,16 @@ class SegmentPKIterator {
...
+    // Returns the most-recently-loaded chunk plus both views of its position
+    // in the source segment (see SegmentPKChunkRef doc). The returned chunk
+    // is moved out — done() will report empty until the next _load().
+    SegmentPKChunkRef current();
+
+    // First physical rowid emitted by the underlying iterator for this segment,
+    // i.e. the range_start of the iterator's tablet_range filter. Returns 0
+    // before any non-empty emit (which is also the correct base for a non-
+    // shared segment iterating from row 0).
+    uint32_t segment_physical_rowid_base() const { return _segment_physical_rowid_base.value_or(0); }
```
luohaha: Could we hide this concept inside `SegmentPKIterator`

be/src/storage/rowset_column_update_state.h
```diff
@@ -73,8 +73,18 @@ struct ColumnPartialUpdateState {
...
+    // For SHARED update segments (post-split), the row's physical position in
+    // the segment file is upt_segment_physical_rowid_base + insert_rowids[k];
+    // for non-shared segments the base is 0 and logical == physical. Downstream
+    // readers (e.g. fetch_values_by_rowid) must use the physical position.
     std::vector<uint32> insert_rowids;
+    // First physical rowid emitted by the source iterator for this update
+    // segment (i.e., the iterator's range_start; 0 for non-shared segments).
+    // Combined with `insert_rowids[k]` to compute each insert row's physical
+    // position in the update segment file.
+    uint32_t upt_segment_physical_rowid_base = 0;
```
luohaha: We don't need this, because segment generated by column mode partial update would be spiltted.

be/src/storage/lake/rowset_update_state.h
xiangguangyxg: Made `_segment_physical_rowid_base` fully private (commit 283d2f50):

- Renamed `_segment_physical_rowid_base` → `_segment_physical_rowid_offset`
- Removed the public accessor `segment_physical_rowid_base()`
- The only external use was in `batch_parallel_get_rss_rowids`, which needed the segment-wide range start to populate the per-segment output for `ColumnPartialUpdateState`. That call now derives the value from the first chunk's `physical_rowid_offset`: since the first chunk's `logical_rowid_offset == 0`, its `physical_rowid_offset` equals the segment-wide range start.

The iterator now exposes only the two `SegmentPKChunkRef` fields; its internal `_segment_physical_rowid_offset` member is no longer reachable from outside.

be/src/storage/lake/rowset_update_state.cpp
xiangguangyxg: Renamed and re-introduced the second field (commit 283d2f50):

- `physical_rowid_base` → `physical_rowid_offset`
- Re-added `logical_rowid_offset` (it was dropped in an earlier simplifier pass; reinstated after identifying a consumer that requires it — see below)

The two offsets serve different consumers:

- `physical_rowid_offset` — chunk[0]'s position in the source segment file. Used by PK index upserts, delvec writers, and `fetch_values_by_rowid` (which opens the full segment without a range filter).
- `logical_rowid_offset` — chunk[0]'s position in the iterator's emit sequence. Used by `ColumnModePartialUpdateHandler::_update_source_chunk_by_upt`, which calls `append_selective` to copy from an iterator-emitted (range-filtered) chunk; that operation requires chunk-local indices, not segment-file positions.

For non-shared segments the two values are equal; for shared segments they differ by the iterator's range_start. Because the two consumers have opposite indexing requirements, both fields are necessary.

The `_base` suffix in the previous naming was the likely source of confusion: it suggested an offset value to be added, which read similarly to other position-like fields. The `_offset` suffix is symmetric across both fields, and the `physical_`/`logical_` prefix names the coordinate system.

be/src/storage/rowset_column_update_state.h
xiangguangyxg: I'd like to confirm the scenario you're considering, because the code I read suggests this field is still needed.

Tracing the cross-publish path:

- `tablet_reshard_helper.cpp` `set_all_data_files_shared(TxnLogPB_OpWrite*)` marks all entries of `op_write->mutable_rowset()->mutable_shared_segments()` as true.
- `tablet_reshard.cpp:387` `convert_txn_log_for_splitting` invokes that helper for each split child, including the op_write txn log.
- `update_manager.cpp:925` `publish_column_mode_partial_update` constructs `ColumnModePartialUpdateHandler` and reads `params.op_write.rowset()` directly.
- I grepped `be/src/storage/lake/` for any split-time special handling of column-mode partial update (reject, abort, or per-sub-tablet physical slicing) and didn't find it.

That suggests the following scenario is reachable: a source tablet has an in-flight column-mode partial update load when it is split; on cross-publish, each sub-tablet reads the source's update segment marked shared (range filter applied), so the iterator emits chunk-local logical positions while `fetch_values_by_rowid` requires segment-file physical positions. `upt_segment_physical_rowid_offset` provides the translation needed to bridge that gap.

If there is an upper-layer guarantee that prevents this scenario (for example, FE blocking column-mode partial update writes during split), removing the field plus a `DCHECK(shared_segments(i) == false)` contract would also work. From the BE code alone I couldn't find such a guarantee — could you point me to where it lives if you know?

CelerData-Reviewer:
@codex review

be/test/storage/lake/segment_pk_iterator_test.cpp
```diff
@@ -0,0 +1,247 @@
...
+        EXPECT_EQ(expected_physical_offsets[emit_index], ref.physical_rowid_offset);
+        EXPECT_EQ(expected_logical_offsets[emit_index], ref.logical_rowid_offset);
+        // Difference is the segment-wide base, which must NOT change across
+        // _load() calls. 200 here = the iterator's range_start.
+        EXPECT_EQ(200u, ref.physical_rowid_offset - ref.logical_rowid_offset);
+        EXPECT_EQ(2u, ref.chunk->num_rows());
+        ++emit_index;
+        iter.next();
+    }
+    EXPECT_EQ(expected_chunk_bases.size(), emit_index);
```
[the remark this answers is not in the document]
xiangguangyxg: Fixed in current HEAD `fbd5e03a041`. The stale `expected_chunk_bases` on line 244 of commit `283d2f50` was caught by a follow-up simplifier pass and changed to `expected_physical_offsets.size()`. Confirmed via:

```
$ git show HEAD:be/test/storage/lake/segment_pk_iterator_test.cpp | grep -nE "expected_chunk_bases|expected_physical_offsets"
229:    const std::vector<uint32_t> expected_physical_offsets = {200, 202, 204};
233:        ASSERT_LT(emit_index, expected_physical_offsets.size());
235:        EXPECT_EQ(expected_physical_offsets[emit_index], ref.physical_rowid_offset);
244:    EXPECT_EQ(expected_physical_offsets.size(), emit_index);
```

No stale `expected_chunk_bases` reference remains.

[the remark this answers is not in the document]
xiangguangyxg: fixed

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/rowset_column_update_state.h
```diff
@@ -71,12 +71,28 @@ struct ColumnPartialUpdateState {
...
+    // (the iterator's range_start; 0 for non-shared segments and for the non-lake
+    // path). Added to each insert row's logical offset to produce its physical
+    // position in the segment file.
+    uint32_t upt_segment_physical_rowid_offset = 0;
+
+    // build `rss_rowid_to_update_rowid` and `insert_rowids` from `src_rss_rowids`.
+    // The two outputs use different coordinate systems on purpose (see field docs):
+    // rss_rowid_to_update_rowid keeps logical offsets, insert_rowids are translated
+    // to physical positions via upt_segment_physical_rowid_offset.
     void build_rss_rowid_to_update_rowid() {
```
luohaha: pass `upt_segment_physical_rowid_offset` as arg

CelerData-Reviewer:
@codex review

be/src/storage/rowset_column_update_state.h
xiangguangyxg: done

be/src/storage/lake/rowset_update_state.h
```diff
@@ -136,6 +173,11 @@ class SegmentPKIterator {
     ChunkUniquePtr _pk_column_chunk;
     // For no lazy load, we can load whole pk column and encode at once.
     MutableColumnPtr _standalone_pk_column;
+    // First physical rowid emitted by the underlying iterator for this segment.
+    // Set once on the first non-empty emit and never reset across subsequent
+    // _load() / next() calls. std::optional disambiguates "not yet set" from
+    // the legitimate value 0 (non-shared segment iterating from physical row 0).
+    std::optional<uint32_t> _physical_rowid_offset;
```
luohaha: `_physical_rowid_offset` is too close to `physical_rowid_offset`, use another name

luohaha: I think `_physical_rowid_offset` is immutable, but `physical_rowid_offset` will be changed each time, so you absolutely need different name for them.

be/src/storage/primary_index.cpp
```diff
@@ -1511,25 +1512,42 @@ Status PrimaryIndex::upsert(uint32_t rssid, uint32_t rowid_start, const Column&
     }
 }
 
-Status PrimaryIndex::upsert(uint32_t rssid, const std::vector<uint32_t>& rowids, const Column& pks, IOStat* stat,
-                            ParallelPublishContext* ctx) {
+Status PrimaryIndex::upsert(uint32_t rssid, uint32_t physical_rowid_offset, const std::vector<uint32_t>& indices,
```
luohaha: No need to change this

luohaha: `PrimaryIndex` is a basic components, it shouldn't care about `physical_`

be/src/storage/lake/lake_primary_index.cpp
```diff
@@ -509,21 +509,22 @@ Status LakePrimaryIndex::batch_parallel_get_rss_rowids(ThreadPoolToken* token,
...
+        size_t segment_logical_offset = 0;
         for (; !pk_iter->done(); pk_iter->next()) {
             auto current = pk_iter->current();
-            size_t num_rows = current.first->num_rows();
-            size_t begin_rowid = current.second;
-
             auto slot = std::make_unique<RssRowidSlot>();
-            slot->begin_rowid = begin_rowid;
-            slot->count = num_rows;
+            slot->begin_rowid = segment_logical_offset;
```
luohaha, line 520: Why not use `slot->begin_rowid = current.physical_rowid_offset` ?

be/src/storage/lake/rowset_update_state.cpp
```diff
@@ -40,27 +40,74 @@
...
+        return Status::InternalError(
+                strings::Substitute("SegmentPKIterator: cross-emit non-contiguous rowids; "
+                                    "expected $0, got $1 (delvec/predicate iterator unsupported on this path)",
+                                    expected_next_rowid, rowid_buffer.front()));
+    }
+    // Intra-emit continuity. A single get_next() returning e.g. {5, 7, 9}
+    // (which could come from a future iterator applying delete bitmap or
+    // row predicates) would silently produce wrong physical rowids under
+    // `base + i` arithmetic downstream.
+    for (size_t i = 1; i < rowid_buffer.size(); ++i) {
```
luohaha: This is very waste, we don't need check row by row

be/src/storage/lake/rowset_update_state.cpp
```diff
@@ -40,27 +40,74 @@
...
-                                   _pk_column_chunk->num_rows() >= config::pk_index_parallel_execution_min_rows)) {
-                    break;
-                }
+            }
+            if (rowid_buffer.size() != chunk_container->num_rows()) {
+                return Status::InternalError(
+                        strings::Substitute("SegmentPKIterator: get_next rowids/rows mismatch: $0 vs $1",
+                                            rowid_buffer.size(), chunk_container->num_rows()));
+            }
+            RETURN_IF_ERROR(_validate_rowid_continuity(rowid_buffer, expected_next_rowid));
```
luohaha: i think we don't need to get `rowid_buffer` and run `_validate_rowid_continuity` just for correctness check under release mode.

xiangguangyxg: Removed `_validate_rowid_continuity` entirely (commit 894074726). `_load()` now reads `rowid_buffer` only on the first non-empty emit to capture the segment base; subsequent emits use plain `get_next(chunk)`. Contiguity is enforced structurally by the boundary `DCHECK`s in `Rowset::get_each_segment_iterator`.

be/src/storage/lake/rowset_update_state.cpp
xiangguangyxg: Done — the row-by-row check is gone together with the validator. See the related reply above.

be/src/storage/lake/lake_primary_index.cpp
xiangguangyxg: Not applied. `slot->begin_rowid` indexes the per-segment flat result array sized by total LOGICAL rows (`last->begin_rowid + last->count`), which `build_rss_rowid_to_update_rowid` later iterates by `upt_row_id` in `[0, src_rss_rowids.size())`. For shared segments where `current.physical_rowid_offset = range_start + logical_offset`, using it as `begin_rowid` would over-size the array by `range_start` and leave entries `[0, range_start)` zero-initialized; downstream code would then read those zeros as legitimate `(rssid=0, rowid=0)` mappings, producing silent corruption. Kept the local logical counter.

be/src/storage/primary_index.cpp
xiangguangyxg: Done. Reverted the sparse `upsert(rssid, physical_rowid_offset, indices, pks, ...)` overload and restored the original `upsert(rssid, const std::vector<uint32_t>& rowids, pks, stat, ctx)` — `PrimaryIndex` no longer has any `physical_` terminology. The caller in `update_manager.cpp` materializes absolute rowids and passes them in.

be/src/storage/lake/rowset_update_state.h
xiangguangyxg: Renamed `_physical_rowid_offset` → `_physical_rowid_base`, accessor `physical_rowid_offset()` → `physical_rowid_base()`. The immutable segment-wide base is now lexically distinguished from the per-chunk `SegmentPKChunkRef::physical_rowid_offset` field which advances by chunk.

luohaha (approved):

kevincai (approved):

PR #73702, opened 2026-05-22
[BugFix] Handle mixed range-colocate × hash JOIN in optimizer

xiangguangyxg:
## Why I'm doing:

F2 follow-up from the range-colocate stack (`colocate.md` follow-ups). Mixed range-colocate × hash-distributed JOIN previously surfaced to MySQL clients as `ERROR 1064 (HY000): Unknown error`. Repro:

```sql
set enable_range_distribution = true;
create table cd  (k1 int, k2 int) order by(k1, k2)
    properties('replication_num'='1', 'colocate_with'='grp:k1');
create table dim (k1 int, x int) distributed by hash(k1) buckets 3
    properties('replication_num'='1');
select count(*) from cd join dim on cd.k1 = dim.k1;
-- pre-fix: ERROR 1064 (HY000): Unknown error
-- post-fix: succeeds with INNER JOIN (BUCKET_SHUFFLE)
```

Captured exception stack:

```
java.lang.IllegalStateException                  // null message
  at GroupExpression.getCost:211
  at ChildOutputPropertyGuarantor.enforceChildDistribution:340
  at ChildOutputPropertyGuarantor.transToBucketShuffle:331
  at ChildOutputPropertyGuarantor.transToBucketShuffleJoin:300
  at ChildOutputPropertyGuarantor.visitPhysicalJoin:619
```

`null`-message `IllegalStateException` gets re-rendered by `MysqlErrPacket` as the generic `Unknown error`.

## What I'm doing:

**Primary fix** in `ChildOutputPropertyGuarantor.convertRangeToHashShuffle`. The helper updates `requiredChildrenProperties[i]` and `childrenOutputProperties[i]` to the post-enforcer hash spec but did not publish the enforcer `GroupExpression` into `childrenBestExprList[i]`. Downstream `transToBucketShuffleJoin` reads `childrenBestExprList` directly, gets the stale original range scan, then calls `getCost(Hash(SHUFFLE_JOIN))` on it — `lowestCostTable` doesn't contain that property → `Preconditions.checkState` fires with no message.

Publishing the enforcer into the third list alongside the existing two list updates makes the converted child visible to downstream bucket/shuffle enforcement.

**Defense in depth** in `OutputPropertyDeriver.visitPhysicalJoin`:
- A mixed (Range, Hash) pair reaching property derivation now throws a typed `StarRocksPlannerException` with a descriptive message. The deriver cannot install exchange enforcers itself, so silently producing a derived output would risk downstream misclassification — failing loud is the correct behaviour if the upstream Guarantor invariant ever breaks again.
- The two pre-existing terminal `IllegalStateException` throws are upgraded to the same typed exception so any unsupported child distribution combination surfaces as a real planner error instead of `Unknown error`.

**Tests**:
- New `RangeColocateMixedJoinPlanTest` (13 tests) — end-to-end shared-data plan tests covering both join orientations, the F2 repro, non-colocate keys, every supported join type (INNER / LEFT OUTER / LEFT SEMI / LEFT ANTI / RIGHT OUTER / FULL OUTER), shuffle hint, `disable_colocate_join`, plus a regression for range×range colocate. Includes an optimizer-level invariant that walks the post-optimization `OptExpression` and asserts no non-colocate join child carries a `RangeDistributionSpec`.
- `Config.enable_range_distribution` and the session variable are saved in `@BeforeAll` and restored in `@AfterAll` so the test does not leak into subsequent suites.

Plan and code both went through `/codex-review` — plan: APPROVED after 3 rounds; code: APPROVED after 3 rounds.

Fixes the F2 follow-up tracked in the internal range-colocate design doc.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Plan-shape-only optimizer fix: a previously-failing query now produces a working shuffle/bucket join plan. No SQL syntax, session variable, config, or API surface changes.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
@@ -248,6 +248,10 @@ public class ChildOutputPropertyGuarantor extends PropertyDeriverBase<Void, Expr
 
         requiredChildrenProperties.set(childIndex, newChildInputProperty);
         childrenOutputProperties.set(childIndex, newChildInputProperty);
+        // Publish the enforcer into childrenBestExprList; downstream enforcements
+        // (transToBucketShuffleJoin) read this list directly and need the converted
+        // child whose lowestCostTable contains the new hash output property.
+        childrenBestExprList.set(childIndex, pair.first);
         return pair.first;
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/OutputPropertyDeriver.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/OutputPropertyDeriver.java
@@ -27,6 +27,8 @@ import com.starrocks.qe.ConnectContext;
 import com.starrocks.qe.SessionVariable;
 import com.starrocks.server.GlobalStateMgr;
 import com.starrocks.sql.ast.JoinOperator;
+import com.starrocks.sql.common.ErrorType;
+import com.starrocks.sql.common.StarRocksPlannerException;
 import com.starrocks.sql.optimizer.base.CTEProperty;
 import com.starrocks.sql.optimizer.base.ColumnRefSet;
 import com.starrocks.sql.optimizer.base.DistributionCol;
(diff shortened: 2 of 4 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

stephen-shelby (approved):

srlch (approved):

PR #73721, opened 2026-05-22
[Refactor] Canonicalize percentile compression literals in FunctionAnalyzer

eshishki:
## Why I'm doing:

`percentile_approx` and `percentile_approx_weighted` currently split
compression handling between FE and BE.

FE can pass several compression expression shapes downstream, while BE still
owns part of the validation and fallback behavior at execution time. That makes
the effective compression harder to reason about in planner-side logic and
leaves more than one place responsible for the same compression contract.

This PR moves the compression literal contract to the analyzer so downstream
code (planner, MV rewrite, BE) sees a single canonical compression value.

## What I'm doing:

Move percentile compression canonicalization to `FunctionAnalyzer`:

- Add FE-side compression constants in `PercentileCompression`.
- Canonicalize compression arguments for `percentile_approx` and
  `percentile_approx_weighted` during analysis.
- Inject an explicit default compression when the optional compression argument
  is omitted.
- Accept an integer literal or a user variable bound to an integer constant for
  the compression argument.
- Reject compression shapes that are not a literal at analysis time — column
  references (including CTE-projected user variables), casts, and arithmetic —
  up front, so they never reach BE as a non-canonical value.
- Make BE consume the FE-canonicalized compression value and use it in the
  serialize-format paths that previously constructed a default
  `PercentileValue`.
- Keep a BE fallback for legacy arities during rolling upgrades from older FEs.

For out-of-range integer compression literals, this PR follows the current user
documentation: values outside the accepted compression range are canonicalized
to the default compression. That includes `0`, which previously failed later
with `compression must be positive`.

That behavior choice is not required by the refactor itself. If reviewers prefer
to preserve the existing hard-fail for non-positive values, we can keep that
validation behavior while still leaving compression canonicalization in
`FunctionAnalyzer`.

## Testing

`test_percentile_approx` and `test_percentile_approx_weighted` lock in the
contract: a pre-existing case that passed compression as `v2 + 1` over a
CTE-projected user variable now asserts the `compression must be an integer
literal` analyzing error, and a companion case passes compression as a direct
user variable (`@v2`) and succeeds. `AnalyzeAggregateTest` and `AggregateTest`
cover literal, NULL, out-of-range, cast, and arithmetic compression at the
analyzer/optimizer level.

Related to #73636

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/agg/percentile_approx.h
+++ b/be/src/exprs/agg/percentile_approx.h
@@ -54,10 +54,21 @@ class PercentileApproxAggregateFunctionBase
 protected:
     static constexpr double MIN_COMPRESSION = 2048.0;
     static constexpr double MAX_COMPRESSION = 10000.0;
-    // Used only as a rolling-upgrade fallback when an old FE ships a legacy
-    // arity (without the canonicalized compression literal). For new-FE calls
-    // FunctionAnalyzer is the single source of truth and DEFAULT lives there.
+    // Kept on BE for rolling upgrades from pre-canonicalization FEs. For
+    // new-FE calls FunctionAnalyzer is the single source of truth and DEFAULT
+    // lives there.
     static constexpr double DEFAULT_COMPRESSION_FACTOR = 10000.0;
+
+    static double clamp_compression_factor(double compression) {
+        if (LIKELY(std::isfinite(compression) && compression >= MIN_COMPRESSION && compression <= MAX_COMPRESSION)) {
+            return compression;
+        }
+        // Old FEs can still ship explicit compression literals without the
+        // analyzer-side [MIN, MAX] clamp during a rolling upgrade.
+        LOG(WARNING) << "Compression factor out of range. Using default compression factor: "
+                     << DEFAULT_COMPRESSION_FACTOR;
+        return DEFAULT_COMPRESSION_FACTOR;
+    }
 
 public:
     virtual double get_compression_factor(FunctionContext* ctx) const = 0;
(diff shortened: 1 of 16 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

eshishki:
Fixed in `a1701d5cbe6`. You're right, and the root cause is the merge-phase arity collapse: `SplitAggregateRule.appendConstantColumns` passes only the constant args to the merge aggregator, so a non-constant weight column is dropped and `percentile_approx_weighted(v, weight_col, q, compression)` reaches merge as `[state, q, compression]` (3 args) — indistinguishable *by arity* from a legacy 3-arg call with a constant weight `[state, w, q]`. The `num_args <= 3` legacy guard then returned DEFAULT, so the merge state ran with a different compression than the partial digests.

Rather than guess from arity (or value ranges), all three `merge()` overrides now adopt the compression from the **deserialized incoming state** — compression is serialized inside the TDigest blob, and every partial digest for a group was built with the user compression in the update phase (where the phase keeps all args and arity is reliable), so the merge target adopts it by construction. `clamp_compression_factor` guards a garbage/empty blob. `get_compression_factor` stays for the update/convert paths.

PR #73723, opened 2026-05-22
[Feature] Add percentile_hash(value, compression) overload

### LINKED ISSUE #73725: [Feature] Support explicit compression in percentile_hash
## Feature request

`percentile_hash(value)` builds percentile digests with the legacy storage compression. That keeps existing callers compatible, but it gives MV authors no function-level way to store a percentile digest at an explicit compression when a later `percentile_approx(..., compression)` query asks for that precision.

**Describe the solution you would like**

Add a compression-aware overload:

```sql
percentile_hash(value, compression)
```

The existing one-argument `percentile_hash(value)` should keep its current behavior. The compression argument should use the same analyzer-side percentile compression canonicalization as `percentile_approx` so BE receives a constant canonical compression value.

**Describe alternatives you have considered**

Keeping the one-argument function only preserves compatibility but cannot represent the requested stored precision in an MV definition. Changing its default compression would alter existing behavior and storage semantics.

**Additional context**

This is the feature slice of the percentile MV compression work discussed in #73636. PR #73723 implements the overload separately from the MV rewrite policy changes so it can be reviewed and merged independently.

eshishki:
## Why I'm doing:

`percentile_hash(value)` always builds a percentile digest with the legacy
storage compression.

That is fine for existing callers, but it leaves no function-level way to build
a percentile digest with an explicit compression when the caller needs to store
a digest at a different precision.

## What I'm doing:

Add a compression-aware overload for percentile hash:

- Register `percentile_hash(DOUBLE, DOUBLE)` with opcode `130003`.
- Add the BE implementation
  `PercentileFunctions::percentile_hash_with_compression`.
- Reuse the analyzer-side percentile compression canonicalization for the new
  overload, so BE receives a constant canonical compression value.
- Keep the existing one-argument `percentile_hash(value)` behavior unchanged.
- Add a BE unit test covering the new overload and verifying the serialized
  digest keeps the requested compression.

This PR is currently stacked on #73721 because it reuses the analyzer-side
compression canonicalization added there. Until #73721 is merged, the GitHub
diff against `main` includes that base refactor as well.

Fixes #73725

Related to #73636

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/agg/percentile_approx.h
+++ b/be/src/exprs/agg/percentile_approx.h
@@ -54,10 +54,21 @@ class PercentileApproxAggregateFunctionBase
 protected:
     static constexpr double MIN_COMPRESSION = 2048.0;
     static constexpr double MAX_COMPRESSION = 10000.0;
-    // Used only as a rolling-upgrade fallback when an old FE ships a legacy
-    // arity (without the canonicalized compression literal). For new-FE calls
-    // FunctionAnalyzer is the single source of truth and DEFAULT lives there.
+    // Kept on BE for rolling upgrades from pre-canonicalization FEs. For
+    // new-FE calls FunctionAnalyzer is the single source of truth and DEFAULT
+    // lives there.
     static constexpr double DEFAULT_COMPRESSION_FACTOR = 10000.0;
+
+    static double clamp_compression_factor(double compression) {
+        if (LIKELY(std::isfinite(compression) && compression >= MIN_COMPRESSION && compression <= MAX_COMPRESSION)) {
+            return compression;
+        }
+        // Old FEs can still ship explicit compression literals without the
+        // analyzer-side [MIN, MAX] clamp during a rolling upgrade.
+        LOG(WARNING) << "Compression factor out of range. Using default compression factor: "
+                     << DEFAULT_COMPRESSION_FACTOR;
+        return DEFAULT_COMPRESSION_FACTOR;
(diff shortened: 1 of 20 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

eshishki:
Fixed in `0e3b785eab4` (rebased onto current main). The compression validation lived inside `analyzeBuiltinAggFunction`, which `FunctionAnalyzer.analyze` only invokes for `AggregateFunction` — but `percentile_hash` is a scalar function, so the check never ran. Moved the block to the scalar `analyze()` path. A non-constant compression arg is now rejected up front (`clampCompressionLiteral` → "compression must be an integer literal"), so BE's `get_const_value<TYPE_DOUBLE>(columns[1])` always gets a constant in range; out-of-range / NULL still clamp to DEFAULT. Added `AnalyzeAggregateTest.testPercentileHashFunction` covering non-constant / non-integer / CAST rejection and in-range / out-of-range / NULL clamping.

PR #73726, opened 2026-05-22
[Enhancement] Store BackendStatus.lastSuccessReportTabletsTime as epoch ms and tidy TimeUtils

kevincai:
The prior round-trip kept the timestamp as a "yyyy-MM-dd HH:mm:ss" string and re-parsed it back to a long inside ReportHandler / MetaRecoveryDaemon. With longToTimeString(long) being session-zoned, the parse side needed a separate fixed-zone helper so the cycle did not depend on the default session time_zone. The wall-clock was a useless intermediate; the field is in-memory only (no @SerializedName) so no persistence-compat concern blocks switching the storage type.

Rename BackendStatus.lastSuccessReportTabletsTime -> lastSuccessReportTabletsTimeMs and store it as a long epoch-ms (0 means "never reported"). Internal callers compare the long directly; display sites route through a single
BackendStatus.getLastSuccessReportTabletsTimeStr() accessor that renders 0 as "N/A".

The SHOW BACKENDS contract is preserved. BackendStatus only escapes the FE into the "Status" column of SHOW BACKENDS / SHOW PROC '/backends' - it is never persisted - so the previous default-Gson serialization is replaced with an explicit
BackendStatus.toShowBackendsJson() that writes the documented shape {"lastSuccessReportTabletsTime":"<formatted>"} directly. BackendsProcDir calls it in place of `new Gson().toJson(...)`.

Tidy up on TimeUtils:

  - longToTimeStringInStorageZone(long) had only this one caller; remove it.
  - timeStringToLongInStorageZone(String) stays for LoadsSystemTable.pickTime, which still parses legacy wall-clock strings produced by older BE versions over RPC; clarify the javadoc to say that is its only remaining use.
  - longToTimeString(long) retains its fast-path for non-positive sentinels and reuses the pre-parsed DATETIME_TO_STRING_FORMAT .withZone(...) to avoid pattern reparse on the SHOW / proc hot path.
  - Collapse the parse-with-zone logic into a private helper; the public parse surface is just timeStringToLong(String) and timeStringToLongInStorageZone(String).

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/system/information/LoadsSystemTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/system/information/LoadsSystemTable.java
@@ -253,9 +253,9 @@ public class LoadsSystemTable {
             // Prefer the new UTC epoch-ms fields when the BE sets them - those are
             // already comparable to LoadJob.finishTimestamp (System.currentTimeMillis).
             // Fall back to parsing the legacy wall-clock string fields only when an
-            // older BE hasn't been upgraded yet; that path keeps the historical
-            // (buggy on non-Asia/Shanghai sessions) semantics rather than introducing
-            // a third interpretation.
+            // older BE hasn't been upgraded yet; parse those in the fixed cluster
+            // storage zone (+08:00) to match how the legacy BE produced them,
+            // regardless of the current session time_zone.
             filter.loadStartTimeFrom = pickTime(
                     request.isSetLoad_start_time_from_ms() ? request.getLoad_start_time_from_ms() : null,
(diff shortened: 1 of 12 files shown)


CelerData-Reviewer:
@codex review

gengjun-git (approved):

stdpain (approved):

HangyuanLiu (approved):

PR #73728, opened 2026-05-22
fix(be): correctly read cgroupsv2 memory limit

### LINKED ISSUE #73727: [BUG] BE/CN mem_limit cgroupsv2 logic is incorrect
1. Deploy a BE/CN node with `resources.limits.memory` set
2. Do not explicitly set a `mem_limit` value (or use a % maybe)

* BE/CN detects cgroupsv2 memory limit and uses it

### Real behavior (Required)
* BE/CN will fail to detect cgroupsv2 memory limit

4.0.8 (bug still exists on `main`)

### Technical Details / Context
I believe this was regressed in #43270. `/sys/fs/cgroup/kubepods/memory.max` will only exist on the _Node_, i.e. it's for kubelet to use.

That PR made it an `&&` for existence, so we'll always short-circuit since it won't exist.

The Doris implementation has diverged quite a bit: https://github.com/apache/doris/blob/5960d4cea0e6d2a08f62b45a07a662395d2bbcfc/be/src/common/cgroup_memory_ctl.cpp#L83

milas:
## Why I'm doing:

BE/CN does not correctly detect cgroupsv2

## What I'm doing:

Fixes #73727

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

Not sure of a good way to test this!

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

Don't know what backports make sense!

### DIFF
--- a/be/src/common/system/mem_info.cpp
+++ b/be/src/common/system/mem_info.cpp
@@ -163,10 +163,29 @@ void MemInfo::set_memlimit_if_container() {
                 StringParser::string_to_int<int64_t>(limit_in_bytes_str.data(), limit_in_bytes_str.size(), &result);
     } else if (fs.f_type == CGROUP2_SUPER_MAGIC) {
         // cgroup v2
-        // Read from /sys/fs/cgroup/memory/memory.max
+        std::string cgroup_info;
+        std::string cgroup_path = "";
+        if (FileUtil::read_whole_content("/proc/self/cgroup", cgroup_info)) {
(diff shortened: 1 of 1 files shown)


kevincai (changes requested):
need to double confirm if it is a real issue.

PR #73745, opened 2026-05-23
[Enhancement] Iceberg parquet: short-circuit SELECT DISTINCT via dict page

### LINKED ISSUE #73746: [Feature] Skip data-page scan for SELECT DISTINCT on iceberg parquet low-cardinality columns
## Feature request

`SELECT DISTINCT col FROM iceberg_t` and `SELECT col FROM iceberg_t GROUP BY col` over iceberg parquet are slow on low-cardinality columns (`country`, `region`, `status` — tens of distinct values across millions of rows). All data pages of the column get fetched and decoded even though parquet already stores the answer in the per-row-group dictionary page.

New session variable `enable_iceberg_dict_page_shortcut` (default OFF). When enabled, eligible single-column DISTINCT / GROUP-BY queries skip data-page IO and return the dictionary values directly. Queries that aren't eligible silently fall back to the normal scan.

```sql
SET enable_iceberg_dict_page_shortcut = true;
SELECT DISTINCT country FROM iceberg_catalog.db.events;
```

n/a

eshishki:
## Why I'm doing:

`SELECT DISTINCT col FROM iceberg_t` and `SELECT col FROM iceberg_t GROUP BY col` over iceberg parquet currently scan every row of the column: decode RLE codes into strings, push them up to HashAggregate, dedup into a HashSet. For low-cardinality columns (`country`, `region`, `status`, `device_type`) the answer is already sitting on disk — the column-chunk dictionary page contains all distinct values of that row group, but we ignore it and recompute row by row.

On typical low-card columns this means scanning megabytes of data pages to recover ~50 dictionary entries that were one IO away.

## What I'm doing:

A new FE rule (`ApplyIcebergDictPageShortcutRule`, sits after `LowCardinalityRewriteRule` / `ApplyMinMaxStatisticRule` in `QueryOptimizer.physicalRuleRewrite`) recognises the pattern

```
HashAggregate(group_by=[col], aggregations=∅)
  ← [Project(identity)?]
    ← PhysicalIcebergScanOperator(data_slots={col}, parquet, no predicate)
```

and sets a hint on the scan via `ScanOptimizeOption.dictPageShortcutHint`. The hint propagates through `THdfsScanNode` to BE. The rule is a thin pattern-match + one setter — it deliberately doesn't reuse `LowCardinalityRewriteRule` (which works with FE-shipped `global_dict`, rewrites scan output type to int dict-codes, and is off by default for lake) or `ApplyMinMaxStatisticRule` (which targets compressed-key min/max statistics on the agg, not a scan hint). The semantics are different enough that mixing them in either existing pass would be more confusing than helpful.

The parquet `GroupReader` decides per row group in `prepare()` — after column readers are set up but before coalesced PAGES IO planning — whether the shortcut is safe. If yes it materialises the dict-page values via a new `ScalarColumnReader::materialize_dictionary_values()` (reuses the existing column converter so all types are supported, not just strings) and `get_next()` emits the prepared values in `chunk_size`-bounded batches.

Per-RG safety gates:

- **Writer trust**: `FileMetaData.writer_version().application_ == "parquet-mr"` (exact match). parquet-mr is the only writer with explicit prune of unreferenced dict entries (`DictionaryValuesWriter.toDictPageAndClose` writes only `lastUsedDictionarySize`). arrow-cpp/pyarrow/DuckDB writers can leave unused dict entries — silent over-approximation risk.
- **Column chunk fully dict-encoded** via existing `column_all_pages_dict_encoded()` (same gate Trino/Iceberg use for dict-aware predicate filtering).
- **No scan predicate** (`scan.getPredicate() == null`). In principle a predicate that partition pruning, RG zone-map and page-index fully resolve at row-group level (RG either entirely included or entirely excluded, no per-row residual) is safe; but `PredicateFilterEvaluator` currently only signals "RG dropped" or "_range narrowed by page-index", not "predicate proved TRUE for the entire RG". Without that signal we can't runtime-distinguish "fully resolved" from "residual remains", so the conservative gate bails on any scan predicate.
- **No per-slot conjuncts at runtime** (`conjunct_ctxs_by_slot` empty on BE). Runtime filters propagated from upstream joins land on the scan's slot map at execution time, after FE planning is done. If one arrives, the normal pipeline filters rows by it; the shortcut would emit dict values that runtime-filtered rows would never have produced, over-approximating the result. The FE gate catches static predicates; this catches RF arrivals.
- **Deletion bitmap empty for this RG's row range** — uses `DeletionBitmap::get_range_cardinality()`, so clean RGs of partially-deleted v2 position-delete / v3 deletion-vector files still qualify.
- **Page-index did not narrow the RG**: `_range.span_size() == num_rows`.
- **`Statistics.null_count` is set** — needed to decide whether NULL belongs in the result; unset → fallback.
- **`dictionary_page_offset` declared** and **`null_count != num_rows`** — parquet-mr may omit the dict page entirely for all-NULL RGs even when encoding stats claim dict encoding.
- **Single non-partition data-slot** in scan; walker accepts only identity Project between agg and scan (non-row-preserving ops like TopN/Sort/Limit disqualify).

Anything else falls back to the standard row pipeline.

**Off by default** behind `enable_iceberg_dict_page_shortcut` (session variable). Without explicit `SET enable_iceberg_dict_page_shortcut=true` no plan or scan behavior changes.

Also adds `DictDecoder<T>::get_dict_values(Column*)` for non-Slice types so the shortcut works for INT/BIGINT/DECIMAL/DATE etc. (the existing override only covered Slice).

Fixes #73746

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Test plan

- FE: `IcebergDictPageShortcutTest` — 8 plan-explain cases (DISTINCT/GROUP-BY get hint; multi-col, predicate, agg fns, inner LIMIT/ORDER, scan LIMIT, flag-off all rejected).
- BE: `dict_encoding_test` — 3 new cases for `DictDecoder<T>::get_dict_values` on int / varchar / empty dict.
- Manual smoke: end-to-end `SET enable_iceberg_dict_page_shortcut=true; SELECT DISTINCT col FROM iceberg_t;` + `EXPLAIN VERBOSE` to verify `dict-page shortcut: hint=true` line, plus runtime profile bytes-read.

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/be/src/connector/hive_connector.cpp
+++ b/be/src/connector/hive_connector.cpp
@@ -459,6 +459,9 @@ void HiveDataSource::_init_tuples_and_slots(RuntimeState* state) {
     if (hdfs_scan_node.__isset.can_use_count_opt) {
         _use_count_opt = hdfs_scan_node.can_use_count_opt;
     }
+    if (hdfs_scan_node.__isset.dict_page_shortcut_hint) {
+        _dict_page_shortcut_hint = hdfs_scan_node.dict_page_shortcut_hint;
+    }
     if (hdfs_scan_node.__isset.use_partition_column_value_only) {
         _use_partition_column_value_only = hdfs_scan_node.use_partition_column_value_only;
     }
@@ -858,6 +861,7 @@ Status HiveDataSource::_init_scanner(RuntimeState* state) {
     scanner_params.use_min_max_opt = _use_min_max_opt;
     scanner_params.can_use_any_column = _can_use_any_column;
     scanner_params.use_count_opt = _use_count_opt;
+    scanner_params.dict_page_shortcut_hint = _dict_page_shortcut_hint;
     scanner_params.all_conjunct_ctxs = _all_conjunct_ctxs;
     if (!_disable_column_access_path_hints && !_column_access_paths.empty()) {
         scanner_params.column_access_paths = &_column_access_paths;
--- a/be/src/connector/hive_connector.h
+++ b/be/src/connector/hive_connector.h
@@ -189,6 +189,8 @@ private:
     // that placeholder column from the data file.
     bool _can_use_any_column = false;
     bool _use_count_opt = false;
+    // Hint for parquet dict-page shortcut. Propagated to HdfsScannerParams.
+    bool _dict_page_shortcut_hint = false;
     const HiveTableDescriptor* _hive_table = nullptr;
 
     bool _has_scan_range_indicate_const_column = false;
--- a/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
+++ b/be/src/exec/hdfs_scanner/hdfs_scanner.cpp
@@ -202,6 +202,7 @@ Status HdfsScanner::_build_scanner_context() {
     ctx.use_min_max_opt = _scanner_params.use_min_max_opt;
     ctx.can_use_any_column = _scanner_params.can_use_any_column;
     ctx.use_count_opt = _scanner_params.use_count_opt;
+    ctx.dict_page_shortcut_hint = _scanner_params.dict_page_shortcut_hint;
     ctx.use_file_metacache = _scanner_params.use_file_metacache;
     ctx.use_file_pagecache = _scanner_params.use_file_pagecache;
     ctx.timezone = _runtime_state->timezone();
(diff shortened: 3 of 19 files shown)


CelerData-Reviewer:
@codex review

PR #73763, opened 2026-05-24
[Enhancement] account collecting aggregates' off-pool heap in agg-state memory

eshishki:
## Why I'm doing:

The "collecting" aggregates -- `array_agg`, `array_union_agg`, `group_concat`,
`map_agg` -- keep their per-group state off the aggregator mem pool: a `Column`
(`array_agg`/`map_agg` value column), a `std::string` (`group_concat`), or a
`phmap` hash set/map, all allocated through the **default** allocator. That heap
is visible to the query `MemTracker` (via the allocation hook) but **not** to
the operator-level agg-state counter `_agg_state_mem_usage`, which is the only
thing fed into `Aggregator::memory_usage()`.

So for these aggregates `memory_usage()` reported ~0 for the bulk of their
footprint. `memory_usage()` drives:
- **memory-limited streaming** pre-aggregation (`LimitedMemAggState::has_limited`
  compares `memory_usage()` to the limit) -- affected on `main` today;
- **spill** revocable-bytes / mem-table sizing -- affected once the spill
  operators read `memory_usage()` (see sibling PR #73758).

Either way the operator under-counted these states and could keep building a
huge in-memory hash table without the accounting reflecting it.

## What I'm doing:

Charge the off-pool heap into `_agg_state_mem_usage` via
`FunctionContext::add_mem_usage`, mirroring `percentile`/HLL/DataSketches:
capacity-based deltas around `update`/`merge`/`reset` (and the up-front
`reserve` on the single-state `group_concat` path). Three commits by mechanism:

1. **single-column** (`array_agg`/`array_union_agg` non-distinct, `group_concat`)
   -- exact `Column::memory_usage()` / `std::string::capacity()` deltas. The
   `std::string` SSO inline buffer lives in the state struct (already in the mem
   pool), so only a heap buffer is reported.
2. **container** (`map_agg`, `array_agg` V2 / `group_concat` V2 with ORDER BY) --
   sum of `data_columns` memory; for `map_agg` also an **approximate** size of
   the open-addressing hash map (`capacity * (sizeof(value_type) + 1)`; phmap
   exposes no exact byte count). `map_agg` string keys live in the mem pool
   (already counted), so they are excluded.
3. **distinct dedup set** (`array_agg(DISTINCT)`, `array_unique_agg`) --
   approximate size of the phmap set, guarded by `if constexpr (is_distinct)`.

**No double count:** this heap is not in the mem pool (`hash_map_memory_usage`)
nor the counting allocator (`allocator_memory_usage`), and `add_mem_usage` is
independent of the query `MemTracker` that catches the malloc. These are the
exact terms `Aggregator::memory_usage()` already sums.

**Deliberately out of scope:** `count(distinct)`/`sum(distinct)` already account
their set through the `AggregateStateAllocator` (-> `allocator_memory_usage`), so
they are untouched. For distinct array variants the dedup set is materialized
into the value column lazily at output; that transient copy is not separately
charged (the build-time set drives sizing, and the query `MemTracker` still sees
it) -- same as `percentile`/HLL, which don't account output materialization.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

(Query results are unchanged. Only the internal memory-accounting accuracy
improves, so memory-limited streaming / spill may trigger slightly earlier and
more correctly under memory pressure.)

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/agg/array_agg.h
+++ b/be/src/exprs/agg/array_agg.h
@@ -96,8 +96,20 @@ struct ArrayAggAggregateState {
         set.clear();
     }
 
-    // Off-pool heap charged into the operator's agg-state memory so it shows in Aggregator::memory_usage().
-    int64_t mem_usage() const { return data_column.memory_usage(); }
+    // Off-pool heap charged into the operator's agg-state memory so it shows in
+    // Aggregator::memory_usage(): the value column, plus (distinct mode) the dedup hash set,
+    // approximated from its slot count (phmap exposes no exact byte count). For string keys the
+    // key bytes live in the operator mem pool (counted there); only the set's slots/control
+    // bytes are added here, so there is no double counting. Distinct mode materializes the set
+    // into the value column lazily at output (get_data_column); that transient copy is not charged
+    // here -- only the build-time set drives sizing, and the query MemTracker still sees it.
+    int64_t mem_usage() const {
+        int64_t usage = data_column.memory_usage();
+        if constexpr (is_distinct) {
+            usage += static_cast<int64_t>(set.capacity()) * (sizeof(typename MyHashSet::value_type) + 1);
+        }
+        return usage;
+    }
 
     ColumnType data_column; // Aggregated elements for array_agg
     size_t null_count = 0;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

PR #73821, opened 2026-05-25
[Enhancement] Add avrocpp-based Avro scanner for routine load

eshishki:
## Why I'm doing:

Routine-load Avro (Kafka) runs on a separate legacy stack — C `libavro` + `libserdes` (`AvroScanner` + `formats/avro/{binary,numeric,nullable}_column`) — that the rest of the engine no longer uses. It has real gaps:

- **No native STRUCT/MAP** — an Avro `record` is flattened to a JSON string and a `map` to JSON, instead of `STRUCT`/`MAP` columns.
- **Avro logical types are ignored** — `date`/`timestamp`/`decimal` are read as raw int/long/bytes (the readers never inspect `logicalType`), so they don't become `DATE`/`DATETIME`/`DECIMAL`.
- **Two Avro engines to maintain** — `libavro` for routine load, `avrocpp` everywhere else (FILES()/external, `hdfs_scanner_avro`).

## What I'm doing:

Add `AvroStreamScanner`, a routine-load Avro scanner on the **`avrocpp`** stack that reuses the **same column readers** as the FILES()/external path — so routine load gets native `STRUCT`/`MAP`/`ARRAY` and correct logical types, on a single Avro engine. `libserdes` stays, used only to resolve the schema id from the Confluent Schema Registry.

- `ConfluentAvroDecoder` — strips Confluent framing, resolves the writer schema by id (cached), decodes one message into a `GenericDatum`.
- jsonpath navigation ported onto `GenericDatum` — functional parity with the legacy path, with two deliberate tightenings where the legacy reader silently loads the parent value instead: a jsonpath that overshoots a scalar leaf (`$.id.x` on a `long` `id`), and a root array index on a non-array root (`$[0]` on a record). Both now fail the row with an explicit error. Everything else (union peeling, `$`/array-index selectors, missing field → NULL) behaves identically.
- **Avro is Kafka-only.** Both decoders (legacy and native) require Confluent wire framing + a Confluent Schema Registry — a Kafka-ecosystem mechanism with no Pulsar equivalent (Pulsar's native per-topic schema is unrelated and not Confluent-framed, so it would not decode). Pulsar behavior is unchanged in this PR: `PulsarTaskInfo` still maps every non-json format (including avro) to `FORMAT_CSV_PLAIN`, as before. The native reader applies to Kafka jobs only.
- `avro.use_native_reader` is validated through `Util.getBooleanPropertyOrDefault`, consistent with the other routine-load boolean properties.

**Selection.** Per-job: the `avro.use_native_reader` property in `CREATE ROUTINE LOAD` overrides the FE-wide default `enable_routine_load_native_avro_reader` (default `false`). The choice is resolved at job creation and persisted, so it is uniform across all BE nodes and stable across config changes — flipping the FE config affects only newly created jobs; existing jobs keep their reader.

⚠️ **Behaviour change.** Enabling the native reader loads `record`/`map` as `STRUCT`/`MAP`, and `date`/`timestamp`/`decimal` as interpreted `DATE`/`DATETIME`/`DECIMAL` instead of legacy raw values. Hence default-off and per-job opt-in.

**Compatibility.** Decode/registry failure behavior kept 1:1 with legacy (filtered row + error file + task fail). jsonpaths and schema-registry URL/auth unchanged; `schema evolution` stays unsupported (same limit). Upgrade-safe: `optional` thrift field (new ordinal), persisted FE field has `@SerializedName`; new-FE/old-BE and old-FE/new-BE both fall back to the legacy scanner. Enable native only after the BE upgrade completes.

**Tests.** Unit tests: decoder round-trip, jsonpath navigation, STRUCT/MAP filling, logical-`date`. The reused logical-type/complex `avrocpp` readers are already covered on the FILES path (`avro_cpp_scanner_test`). Docs updated (`RoutineLoad.md`, FE config reference).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/file_scanner/avro_stream_scanner.cpp
+++ b/be/src/exec/file_scanner/avro_stream_scanner.cpp
@@ -1,0 +1,257 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "exec/file_scanner/avro_stream_scanner.h"
+
+#include <fmt/format.h>
+
+#include <avrocpp/Exception.hh>
+#include <avrocpp/GenericDatum.hh>
+#include <avrocpp/Types.hh>
+
+#include "base/time/timezone_utils.h"
+#include "column/adaptive_nullable_column.h"
+#include "column/chunk.h"
+#include "column/column_helper.h"
+#include "column/nullable_column.h"
+#include "common/runtime_profile.h"
+#include "exec/file_scanner/json_scanner.h" // JsonScanner::parse_json_paths
+#include "fs/fs.h"
+#include "gutil/casts.h"
+#include "runtime/runtime_state.h"
+#include "runtime/runtime_state_helper.h"
+#include "runtime/stream_load/stream_load_pipe.h"
+#include "util/byte_buffer.h"
+
+namespace starrocks {
+
+AvroStreamScanner::AvroStreamScanner(RuntimeState* state, RuntimeProfile* profile, const TBrokerScanRange& scan_range,
+                                     ScannerCounter* counter)
+        : FileScanner(state, profile, scan_range.params, counter),
+          _scan_range(scan_range),
+          _max_chunk_size(state->chunk_size()) {
+    _file_format_str = "avro";
+}
+
+AvroStreamScanner::~AvroStreamScanner() = default;
+
+Status AvroStreamScanner::open() {
+    RETURN_IF_ERROR(FileScanner::open());
+    if (_scan_range.ranges.empty()) {
+        return Status::OK();
+    }
+
+    if (!TimezoneUtils::find_cctz_time_zone(_state->timezone(), _timezone)) {
+        return Status::InvalidArgument(fmt::format("can not find cctz time zone {}", _state->timezone()));
+    }
+
+    const TBrokerRangeDesc& range_desc = _scan_range.ranges[0];
+
+    if (!_scan_range.params.__isset.confluent_schema_registry_url) {
+        return Status::InternalError("'confluent_schema_registry_url' not set");
+    }
+    _decoder = std::make_unique<ConfluentAvroDecoder>(_scan_range.params.confluent_schema_registry_url);
+    RETURN_IF_ERROR(_decoder->init());
+
+    if (range_desc.__isset.jsonpaths) {
+        std::string jsonpaths = avro_preprocess_jsonpaths(range_desc.jsonpaths);
+        RETURN_IF_ERROR(JsonScanner::parse_json_paths(jsonpaths, &_json_paths));
+    }
+
+    RETURN_IF_ERROR(create_column_readers());
+    RETURN_IF_ERROR(create_sequential_file(range_desc, _scan_range.broker_addresses[0], _scan_range.params, &_file));
+    ++_counter->num_files_read;
+    return Status::OK();
+}
+
+Status AvroStreamScanner::create_column_readers() {
+    _column_readers.resize(_src_slot_descriptors.size());
+    _field_names.resize(_src_slot_descriptors.size());
+    for (size_t i = 0; i < _src_slot_descriptors.size(); ++i) {
+        auto* slot_desc = _src_slot_descriptors[i];
+        if (slot_desc == nullptr) {
+            continue;
+        }
+        _column_readers[i] = avrocpp::ColumnReader::get_nullable_column_reader(slot_desc->col_name(), slot_desc->type(),
+                                                                               _timezone, !_strict_mode);
+        // Materialize the avro field name once; fill_row() reuses it per message instead of rebuilding a
+        // std::string from the pmr-backed col_name() string_view for every cell.
+        _field_names[i] = slot_desc->col_name();
+    }
+    return Status::OK();
+}
+
+Status AvroStreamScanner::create_src_chunk(ChunkPtr* chunk) {
+    SCOPED_RAW_TIMER(&_counter->init_chunk_ns);
+    *chunk = std::make_shared<Chunk>();
+    for (size_t i = 0; i < _src_slot_descriptors.size(); ++i) {
+        auto* slot_desc = _src_slot_descriptors[i];
+        if (slot_desc == nullptr) {
+            continue;
+        }
+        auto column = ColumnHelper::create_column(slot_desc->type(), true, false, 0, true);
+        (*chunk)->append_column(std::move(column), slot_desc->id());
+    }
+    return Status::OK();
+}
+
+Status AvroStreamScanner::fill_row(const avro::GenericDatum& datum, Chunk* chunk) {
+    const bool use_jsonpath = !_json_paths.empty();
+
+    const avro::GenericRecord* record = nullptr;
+    if (!use_jsonpath) {
+        const avro::GenericDatum* d = &datum;
+        while (d->type() == avro::AVRO_UNION) {
+            d = &d->value<avro::GenericUnion>().datum();
+        }
+        if (d->type() != avro::AVRO_RECORD) {
+            return Status::InternalError("avro message top-level value is not a record");
+        }
+        record = &d->value<avro::GenericRecord>();
+    }
+
+    for (size_t i = 0; i < _src_slot_descriptors.size(); ++i) {
+        auto* slot_desc = _src_slot_descriptors[i];
+        if (slot_desc == nullptr) {
+            continue;
+        }
+        auto* column = down_cast<AdaptiveNullableColumn*>(chunk->get_column_raw_ptr_by_slot_id(slot_desc->id()));
+
+        if (use_jsonpath) {
+            if (i >= _json_paths.size()) {
+                column->append_nulls(1);
+                continue;
+            }
+            auto extracted = avro_extract_field(datum, _json_paths[i]);
+            if (extracted.status().is_not_found()) {
+                column->append_nulls(1);
+                continue;
+            }
+            RETURN_IF_ERROR(extracted.status());
+            RETURN_IF_ERROR(_column_readers[i]->read_datum_for_adaptive_column(*extracted.value(), column));
+        } else {
+            // GenericRecord::hasField/field take const std::string&; reuse the name materialized once in
+            // create_column_readers() (col_name() is a pmr-backed std::string_view).
+            const std::string& field_name = _field_names[i];
+            if (record->hasField(field_name)) {
+                RETURN_IF_ERROR(_column_readers[i]->read_datum_for_adaptive_column(record->field(field_name), column));
+            } else {
+                column->append_nulls(1);
+            }
+        }
+    }
+    return Status::OK();
+}
+
+Status AvroStreamScanner::parse_one_message(const uint8_t* data, size_t size, Chunk* chunk) {
+    const size_t rows_before = chunk->num_rows();
+
+    Status st = _decoder->decode(data, size, &_datum);
+    if (!st.ok()) {
+        _counter->num_rows_filtered++;
+        RuntimeStateHelper::append_error_msg_to_file(_state, "", st.to_string());
+        return st;
+    }
+
+    // decode() converts avrocpp exceptions to Status itself; the readers below it do not, and a datum
+    // shape the guards don't anticipate must cost one filtered row, never the BE.
+    try {
+        st = fill_row(_datum, chunk);
+    } catch (const avro::Exception& e) {
+        st = Status::DataQualityError(fmt::format("avro read failed: {}", e.what()));
+    }
+    if (!st.ok()) {
+        if (_counter->num_rows_filtered++ < MAX_ERROR_LINES_IN_FILE) {
+            RuntimeStateHelper::append_error_msg_to_file(_state, "", st.to_string());
+        }
+        chunk->set_num_rows(rows_before);
+        return st;
+    }
+    return Status::OK();
+}
+
+StatusOr<ChunkPtr> AvroStreamScanner::get_next() {
+    SCOPED_RAW_TIMER(&_counter->total_ns);
+    // open() succeeds without initializing _file when the scan range carries no ranges.
+    if (_file == nullptr) {
+        return Status::EndOfFile("no avro ranges to read");
+    }
+    ChunkPtr src_chunk;
+    RETURN_IF_ERROR(create_src_chunk(&src_chunk));
+
+    auto* stream = down_cast<StreamLoadPipeInputStream*>(_file->stream().get());
+
+    Status st = Status::OK();
+    while (src_chunk->num_rows() < static_cast<size_t>(_max_chunk_size)) {
+        ByteBufferPtr buf;
+        {
+            ++_counter->file_read_count;
+            SCOPED_RAW_TIMER(&_counter->file_read_ns);
+            auto res = stream->pipe()->read();
+            if (!res.ok()) {
+                st = res.status();
+                break;
+            }
+            buf = std::move(res.value());
+        }
+        if (buf == nullptr || buf->remaining() == 0) {
+            continue;
+        }
+        // Confluent framing requires one message per buffer; StreamLoadPipe::read() returns the whole
+        // buffer that KafkaConsumerPipe::append_json enqueued, so each buffer is exactly one message.
+        RETURN_IF_ERROR(
+                parse_one_message(reinterpret_cast<const uint8_t*>(buf->ptr), buf->remaining(), src_chunk.get()));
+    }
+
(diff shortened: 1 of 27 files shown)


CelerData-Reviewer:
@codex review

be/src/exec/file_scanner/avro_stream_scanner.cpp
```diff
@@ -0,0 +1,232 @@
...
+                parse_one_message(reinterpret_cast<const uint8_t*>(buf->ptr), buf->remaining(), src_chunk.get()));
+    }
+
+    if (src_chunk->num_rows() == 0) {
+        if (st.is_end_of_file()) {
+            return Status::EndOfFile("EOF of reading avro stream, nothing read");
+        }
+        // Timeout at the start of a chunk, or any other error: surface it.
+        if (!st.ok()) {
+            return st;
```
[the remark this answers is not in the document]
eshishki: Thanks — fixed. A non-timeout/non-EOF read error is now returned immediately regardless of how many rows were already filled, matching the legacy `AvroScanner`. Amended in `e39bef2`.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/load/routineload/PulsarTaskInfo.java
```diff
@@ -139,6 +139,8 @@ public TRoutineLoadTask createRoutineLoadTask() throws StarRocksException {
         tRoutineLoadTask.setMax_batch_size(Config.max_routine_load_batch_size);
         if (!routineLoadJob.getFormat().isEmpty() && routineLoadJob.getFormat().equalsIgnoreCase("json")) {
             tRoutineLoadTask.setFormat(TFileFormatType.FORMAT_JSON);
+        } else if (!routineLoadJob.getFormat().isEmpty() && routineLoadJob.getFormat().equalsIgnoreCase("avro")) {
+            tRoutineLoadTask.setFormat(TFileFormatType.FORMAT_AVRO);
```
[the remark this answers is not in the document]
eshishki: N/A: Pulsar never emits FORMAT_AVRO — `PulsarTaskInfo` maps every non-json format to CSV_PLAIN. The explicit avro-for-Pulsar reject was dropped (`874b6cc`) to avoid a behavior change.

fe/fe-core/src/main/java/com/starrocks/sql/ast/CreateRoutineLoadStmt.java
```diff
@@ -621,6 +631,13 @@ public void checkJobProperties() throws StarRocksException {
             } else if (format.equalsIgnoreCase("avro")) {
                 format = "avro";
                 jsonPaths = jobProperties.get(JSONPATHS);
+                // Resolve the reader choice once, at creation time, so flipping the FE-wide default
+                // later only affects newly created jobs and never existing ones.
+                if (jobProperties.containsKey(AVRO_USE_NATIVE_READER)) {
+                    useNativeAvroReader = Boolean.parseBoolean(jobProperties.get(AVRO_USE_NATIVE_READER));
+                } else {
```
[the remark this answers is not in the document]
eshishki: Already resolved via `Util.getBooleanPropertyOrDefault`, which rejects non-boolean values.

docs/en/loading/RoutineLoad.md
```diff
@@ -406,13 +469,23 @@ The data type mapping between the Avro data fields you want to load and the Star
 
 | Avro           | StarRocks                                                    |
 | -------------- | ------------------------------------------------------------ |
-| record         | Load the entire RECORD or its subfields into StarRocks as JSON. |
+| record         | Legacy reader: the entire RECORD or its subfields as JSON. Native reader (`avro.use_native_reader = true`): `STRUCT`. |
```
kevincai, line 472: Would like to know the rationale for this change. not sure if STRUCT is feasible enough to hold an Avro record, which can be nested multiple levels.

eshishki: Nested STRUCT is fully supported in StarRocks (since v3.1) — a struct field can itself be a STRUCT/ARRAY/MAP to arbitrary depth, e.g. `STRUCT<a INT, b STRUCT<c INT, d INT>, c MAP<INT,INT>, d ARRAY<INT>>`. The native reader reuses the **same** avrocpp column readers as the FILES()/external path: `StructColumnReader` builds one child reader per `type_desc.children`, so a nested Avro `record` recursively maps onto a nested `STRUCT` — the exact code that already handles multi-level STRUCT for ORC/Parquet/Avro on FILES().

The rationale: if I have a fully-qualified Avro record with typed fields, I don't want it flattened into a JSON string that then has to be text-parsed on every query — I want real columnar columns with proper types. Nested STRUCT works well for this.

If a user *does* want the JSON-blob behavior, it's still achievable per-column at load time — routine load `COLUMNS` accepts expressions, so you can write `my_json_col = to_json(my_record)` (or `CAST(my_record AS JSON)`). So this change adds the typed-column option without taking the JSON option away.

The only real constraint here is not nesting depth but schema being declared in the table DDL (fields map by name); dynamic/unknown-field schema evolution stays unsupported, same as the legacy reader.

kevincai:
@eshishki the ci workflow is dead during the github action incident. You need to rebase the code and push again to trigger the CI workflow again.

eshishki:
Rebased `feat/avro-routine-load-cpp` onto current `main` and force-pushed (`874b6ccd559` → `1c6a61c1dc4`). CI is picking up the new head. Thanks for the heads-up.

CelerData-Reviewer:
@codex review

be/src/exec/file_scanner/avro_stream_scanner.cpp
```diff
@@ -0,0 +1,246 @@
...
+    return Status::OK();
+}
+
+StatusOr<ChunkPtr> AvroStreamScanner::get_next() {
+    SCOPED_RAW_TIMER(&_counter->total_ns);
+    ChunkPtr src_chunk;
+    RETURN_IF_ERROR(create_src_chunk(&src_chunk));
+
+    auto* stream = down_cast<StreamLoadPipeInputStream*>(_file->stream().get());
+
```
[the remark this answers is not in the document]
eshishki: Thanks — fixed. `get_next()` now returns `EndOfFile` early when `_file` is null (empty `_scan_range.ranges`), matching the other scanners' defensive behavior. Amended in `d08d10ea24c`.

fe/fe-core/src/main/java/com/starrocks/sql/ast/CreateRoutineLoadStmt.java
```diff
@@ -621,6 +633,11 @@ public void checkJobProperties() throws StarRocksException {
             } else if (format.equalsIgnoreCase("avro")) {
                 format = "avro";
                 jsonPaths = jobProperties.get(JSONPATHS);
+                // Resolve the reader choice once, at creation time, so flipping the FE-wide default
+                // later only affects newly created jobs and never existing ones.
+                useNativeAvroReader = Util.getBooleanPropertyOrDefault(jobProperties.get(AVRO_USE_NATIVE_READER),
+                        Config.enable_avro_routine_load_native_reader,
+                        AVRO_USE_NATIVE_READER + " should be a boolean");
```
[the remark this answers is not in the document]
eshishki: Thanks — the PR body was inaccurate, updated. This PR intentionally keeps Pulsar behavior unchanged (no behavior change): `PulsarTaskInfo` still maps every non-json format, including avro, to `FORMAT_CSV_PLAIN` as before. The native reader applies to Kafka jobs only.

fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
```diff
@@ -444,6 +448,7 @@ protected void setOptional(CreateRoutineLoadStmt stmt) throws StarRocksException
                 jobProperties.put(CreateRoutineLoadStmt.JSONPATHS, "");
             }
             this.confluentSchemaRegistryUrl = stmt.getConfluentSchemaRegistryUrl();
+            this.useNativeAvroReader = stmt.getUseNativeAvroReader();
```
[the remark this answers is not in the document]
eshishki: Thanks — fixed. `jobPropertiesToSql()` now emits `avro.use_native_reader` when the job has it resolved (Avro jobs), so SHOW CREATE ROUTINE LOAD replays with the persisted reader choice; non-Avro jobs are unchanged. Added a UT. Amended in `23d0ae95522`.

docs/en/administration/management/FE_parameters/user_query_loading.md
```diff
@@ -217,6 +217,15 @@ Starting from version 3.3.0, the system defaults to refreshing one partition at
 - Description: Whether to enable automatic collection for the NDV information of the ARRAY type.
 - Introduced in: v4.0
 
+### `enable_avro_routine_load_native_reader`
+
+- Default: false
+- Type: Boolean
+- Unit: -
+- Is mutable: Yes
+- Description: The default Avro reader for Routine Load jobs that do not set the `avro.use_native_reader` property. When `true`, such jobs use the native (avrocpp) reader, which loads Avro `record`/`map` fields as `STRUCT`/`MAP` columns and interprets Avro logical types (`date` → DATE, `timestamp` → DATETIME, `decimal` → DECIMAL); when `false`, the legacy reader is used. The reader choice is resolved when a job is created, so changing this item only affects newly created jobs, not existing ones.
```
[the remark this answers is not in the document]
eshishki: Thanks — added the zh and ja translations for both the FE config entry and the RoutineLoad section in `2cf449b2917`.

be/src/formats/avro/cpp/complex_column_reader.cpp
```diff
@@ -85,6 +85,14 @@ Status MapColumnReader::read_datum(const avro::GenericDatum& datum, Column* colu
...
 
+    // Avro map keys are always strings, and the key path below appends raw bytes into a binary key
+    // column. FILES() always types the key column VARCHAR (inferred schema), but routine load types it
+    // straight from the table, so e.g. MAP<INT,...> reaches here: reject it instead of corrupting it.
+    if (_key_reader != nullptr && UNLIKELY(!keys_column->data_column_raw_ptr()->is_binary())) {
+        return Status::DataQualityError(
+                fmt::format("avro map keys are strings; the key type of map column '{}' cannot hold them", _col_name));
+    }
+
     const auto& map = datum.value<avro::GenericMap>();
```
wyb, line 107: type-mismatched datum into a `STRUCT/MAP/ARRAY` column will cause BE crash.
for example, table column type is Map but the data in avro stream is INT.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
wyb:
```suggestion
    public static boolean enable_routine_load_native_avro_reader = false;
```

fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
wyb, line 1815: For Avro jobs with null field, `SHOW CREATE` should emit "avro.use_native_reader"="false" to preserve current behavior.

be/src/exec/file_scanner/avro_stream_scanner.cpp
```diff
@@ -0,0 +1,250 @@
...
+                continue;
+            }
+            RETURN_IF_ERROR(extracted.status());
+            RETURN_IF_ERROR(_column_readers[i]->read_datum_for_adaptive_column(*extracted.value(), column));
+        } else {
+            // GenericRecord::hasField/field take const std::string&; reuse the name materialized once in
+            // create_column_readers() (col_name() is a pmr-backed std::string_view).
+            const std::string& field_name = _field_names[i];
+            if (record->hasField(field_name)) {
+                RETURN_IF_ERROR(_column_readers[i]->read_datum_for_adaptive_column(record->field(field_name), column));
```
wyb, line 148: better to catch `avro::Exception`.

be/src/formats/avro/cpp/complex_column_reader.cpp
eshishki: Fixed — all three complex readers (struct/array/map) now validate the datum kind up front and return a `DataQualityError`, consistent with the scalar readers: strict mode filters the row, non-strict turns it into NULL — instead of the crashing `value<T>()` cast. Added a UT covering struct/map/array with a mismatched scalar datum.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
eshishki: Done — renamed to `enable_routine_load_native_avro_reader` (config, docs in all locales, PR description).

fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
eshishki: Fixed — for Avro jobs the property is now always emitted: the resolved value when present, and an explicit `"false"` for jobs persisted before this field existed, so replaying the shown DDL keeps the legacy reader regardless of the current FE-wide default. Extended the UT.

be/src/exec/file_scanner/avro_stream_scanner.cpp
eshishki: Done — `parse_one_message` now catches `avro::Exception` around the row fill and converts it to a per-row data error going through the existing filtered-row path (the decoder already converts its own exceptions to Status).

eshishki:
@wyb while addressing the data-parsing error findings from this review, we audited how strict mode and per-row errors are actually handled across the routine-load formats — the native reader inherits this landscape, so it seemed worth validating end-to-end. Result of the code walk:

| # | Path | Error class | non-strict (default) | strict | Offset | Errors reach FE | Job outcome |
|---|------|------------|----------------------|--------|--------|-----------------|-------------|
| 1 | CSV | structural (column count, UTF-8) | skip row, `filtered++` | same | advances | yes | PAUSED post-factum via `max_error_number` (default 0); **skipped rows are lost** |
| 2 | CSV | value conversion | NULL, row loads | skip row, `filtered++` | advances | strict: yes | strict: PAUSED post-factum via `max_error_number`, **skipped rows are lost**; non-strict: values become NULL — expected non-strict semantics, load continues |
| 3 | JSON | fatal: broken document | `filtered++`, **task fails** | same | **stays** | no | reason has `"parse error."` → with `pause_on_fatal_parse_error=true` PAUSED; **default false → infinite retry** |
| 4 | JSON | row-level construct/type | conversion → NULL; structural row issues → skip, `filtered++` | skip row, `filtered++` | advances | yes | PAUSED post-factum via `max_error_number`; **skipped rows are lost** |
| 5 | Avro (both readers) | decode: framing/registry/binary | `filtered++`, **task fails** | same | **stays** | no | reason never matches `"parse error"` → **infinite retry, no user-visible state; `pause_on_fatal_parse_error` has no effect for Avro** |
| 6 | Avro (both readers) | value conversion / kind mismatch | NULL, row loads | `filtered++`, **task fails** (no skip) | strict: stays | no | strict: **infinite retry** (row not lost, job stuck); non-strict: values become NULL — expected non-strict semantics, load continues |

Problems this exposes (none introduced by this PR — the native reader is bug-for-bug with the legacy reader here; rows 5/6 behave identically on both):

1. **Strict mode loses data in CSV/JSON**: the row is dropped, the batch still commits and the offset moves past it; the `max_error_number` pause fires only after the commit, so by the time the job pauses the rows are unrecoverable. Strict mode pauses louder, but protects nothing.
2. **Avro errors are invisible**: a fatal Avro error fails the task without advancing the offset (the right conservative stream behavior), but then retries forever with no user-visible job state change — the job stays RUNNING, the reason only in `OtherMsg`.
3. **Error classification is substring matching on the abort message** (`TxnStatusChangeReason.fromString` → `reason.contains("parse error")`) instead of an explicit error code/flag in the attachment. Fragile by construction — and exactly why `pause_on_fatal_parse_error` works for JSON yet silently does nothing for Avro: the Avro messages just don't contain the magic words.
4. **The knobs mean different things per format**: `strict_mode` = "filtered instead of NULL" in CSV/JSON but "fail the task" in Avro; `max_error_number` only functions where the scanner skips (rows 1/2/4); `pause_on_fatal_parse_error` only where the message text matches (row 3). Same job properties, three different semantics.
5. **The strict-mode default is ambiguous in the code**: `CreateRoutineLoadStmt` initializes `strictMode = true`, but `checkJobProperties()` always overwrites it with `RoutineLoadJob.DEFAULT_STRICT_MODE = false` — a dead initializer that suggests default-true was intended at some point. The effective `false` also violates POLA next to INSERT, which defaults to strict.

At minimum we would like to fix the Avro infinite retry (row 5 and the strict case of row 6): fail the task and keep the offset as today, but surface it as a user-visible PAUSED with the reason instead of retrying forever. We can do that in this PR or as a follow-up — maintainer's call. We are also happy to fix the rest of the table once there is an agreed unified policy on what the error-handling semantics should be across formats.

eshishki:
@wyb After PR https://github.com/StarRocks/starrocks/pull/74901 should i close this PR?

wyb:
> @wyb After PR #74901 should i close this PR?

No need to close this PR. We can use the cpp implementation in both files() and routine load later.

PR #73828, opened 2026-05-25
[Enhancement] Add CANCEL ALTER TABLE ... FORCE for publish-stuck alter jobs (phase 2)

luohaha:
## Why I'm doing:

Phase-1 (#73553) added `ADMIN SKIP COMMITTED TRANSACTION` so an operator can unblock a publish-stuck load or lake-compaction transaction. Phase-1 explicitly deferred alter / schema-change support because the three lake alter subclasses all refuse to cancel once they have reached `FINISHED_REWRITING` — exactly the state in which a publish-stuck alter sits, blocking every subsequent load/compaction on the table.

This PR closes that gap: a lake alter (heavy schema change, lightweight async schema change, or metadata alter) whose publish RPC cannot make progress (missing `txnlog`, lost segment / SST, persistent object-storage outage) can now be force-cancelled by the same operator who would otherwise have to drop the partition.

## What I'm doing:

Add a `FORCE` modifier on the existing `CANCEL ALTER TABLE` SQL plus the FE plumbing to make it safe.

**SQL**

```sql
CANCEL ALTER TABLE [COLUMN|ROLLUP|OPTIMIZE] FROM <db>.<tbl> (<jobId>) FORCE
```

Operator flow: `SHOW ALTER` → grab the stuck `JobId` → enable the kill switch → `CANCEL ALTER ... FORCE` → disable the kill switch.

**Scope (phase 2)**

- Covers all three lake alter subclasses: `LakeTableSchemaChangeJob`, `LakeTableAsyncFastSchemaChangeJob`, `LakeTableAlterMetaJob`.
- Non-lake alter subclasses are unchanged (their `cancelImpl(force)` defaults to the existing single-arg form, so existing behaviour is preserved).
- Same kill switch as phase 1: `enable_admin_skip_committed_txn` (default `false`). One FE config covers the whole escape-hatch surface, so operators only flip one switch during recovery.

**Mechanism**

- `AlterJobV2` gains a force-aware cancel: `cancel(errMsg, force)` and a default `cancelImpl(errMsg, force)` that delegates to the existing single-arg form. Subclasses that don't need the escape hatch don't need any change.
- `LakeTableAlterMetaJobBase.cancelImpl(force=true)` bypasses the `FINISHED_REWRITING` guard. Covers both `LakeTableAlterMetaJob` and `LakeTableAsyncFastSchemaChangeJob` via inheritance — neither has shadow tablets to clean up; cancel is a pure state transition + edit-log entry.
- `LakeTableSchemaChangeJob.cancelImpl(force=true)` likewise bypasses the `FINISHED_REWRITING && tableExists()` guard. The existing `removeShadowIndex()` inside `persistStateChange` cleans up the shadow tablets, which is what unblocks the `publish_log_version` path (FE stops scheduling vtxn copies once the shadow index is gone).
- `AlterJobV2.forceSkippedAtCommitted`: gson-persisted audit marker, set to `true` when the job was force-cancelled. Pure observability — does not affect the cancel mechanics.
- `SchemaChangeHandler.cancel` routes to `cancel(reason, force=true)` when the stmt has `FORCE`; otherwise unchanged. Config gate enforces opt-in.

**BE failpoint for integration testing**

Added `DEFINE_FAIL_POINT(lake_publish_version_rpc_fail)` inside `LakeServiceImpl::publish_version` (`lake_service.cpp`). This is the FE→BE RPC that lake alter jobs (heavy / async-fast / meta) call when transitioning `FINISHED_REWRITING → FINISHED`. Enabling the failpoint pins publish failing so the alter parks at `FINISHED_REWRITING`, which is exactly the state the FORCE escape hatch is designed for.

The failpoint is compiled in only when BE is built with `ENABLE_FAULT_INJECTION=ON` (default `OFF`); in standard Release builds it expands to a no-op so production is unaffected.

For QA / e2e: enable on each lake CN via brpc HTTP (shared-data clusters have no BE registered, so the existing `ADMIN ENABLE FAILPOINT ... ON BACKEND` SQL cannot address CN — pre-existing limitation of `FailPointExecutor`, tracked as follow-up):

```bash
curl -X POST -d '{"fail_point_name":"lake_publish_version_rpc_fail","trigger_mode":{"mode":1}}' \
  http://<cn-host>:<brpc-port>/PInternalService/update_fail_point_status
```

End-to-end verified on a fault-injection shared-data cluster: the alter parks at `FINISHED_REWRITING`, plain `CANCEL ALTER` is rejected with a hint to retry with `FORCE`, and `CANCEL ALTER TABLE ... FORCE` succeeds — for both `LakeTableSchemaChangeJob` (bloom_filter shadow rewrite) and `LakeTableAlterMetaJob` (file_bundling toggle).

**Concurrency / safety**

- The FORCE path goes through the same `persistStateChange` + edit-log flow as the existing cancel paths, so replay and audit are unchanged.
- For heavy schema change, shadow indices are removed atomically with the state transition — no half-applied schema is left visible.
- For metadata alter / lightweight schema change, the alter's catalog effect never lands (since the job did not reach `FINISHED`), so cancelling at `FINISHED_REWRITING` is equivalent to "alter never happened".
- No data loss for loads / compactions that follow: once the alter job is `CANCELLED`, the partition is unblocked and subsequent publish operations proceed normally.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

luohaha:
rerun

luohaha:
rerun sql-tester

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJobBase.java
```diff
@@ -487,29 +487,84 @@ protected AgentBatchTask getBatchTask() {
...
+                        // the stale FE-visible version (= commitVersion-1) and
+                        // BE will still try to apply the cancelled alter's
+                        // txn_log when materializing the load's new version. The
+                        // updateVisibleVersion() call mirrors what the normal
+                        // FINISHED_REWRITING → FINISHED transition does, just
+                        // moved into the cancel callback.
+                        boolean advanceVersionForForce = (jobState == JobState.FINISHED_REWRITING) && force;
                         persistStateChange(this, JobState.CANCELLED, () -> {
+                            if (advanceVersionForForce) {
+                                updateVisibleVersion(table);
```
kevincai: Two concerns with reusing `updateVisibleVersion(table)` on the force-cancel path:

**(Medium-High) `metadataSwitchVersion` leader/replay divergence.** `updateVisibleVersion()` does more than bump the version:
```java
if (enableFileBundling() || disableFileBundling()) {
    physicalPartition.setMetadataSwitchVersion(commitVersion);
}
```
For an `ENABLE_FILE_BUNDLING` alter (enable OR disable -- both make the condition true) the live force-cancel sets `metadataSwitchVersion = commitVersion`, but the new replay CANCELLED branch hand-rolls the bump and omits it (`physicalPartition.updateVisibleVersion(commitVersion, finishedTimeMs)` with no `setMetadataSwitchVersion`). So a leader FE and a replayed/restarted FE diverge on `metadataSwitchVersion` for the same partition -- the exact leader/replay-consistency class of bug the `forceSkippedAtCommitted` machinery otherwise avoids. Semantically the format did not switch at `commitVersion` (the `no_op_publish` writes V-1 content tagged as V, the alter is discarded), so setting the vacuum guard here is arguably wrong and needlessly pins vacuum's retain version; the replay path's omission looks correct.

**(Medium) hard precondition can throw inside the edit-log applier.** `updateVisibleVersion()` does `Preconditions.checkState(getVisibleVersion() == commitVersion - 1, ...)`, while the schema-change live path and both replay branches use a soft `if (getVisibleVersion() == commitVersion - 1)` guard. Because `EditLog.logEdit` writes the journal (`submitLog` + `waitInfinity`) before running `walApplier.apply()`, a thrown precondition lands after the CANCELLED entry is durably persisted but before `table.setState(NORMAL)`, leaving the live FE inconsistent with its own journal and the table stuck out of NORMAL. Rare given the stuck-publish invariant, but this feature exists for pathological states.

Suggestion for both: in the force branch bump the visible version only (mirror the guarded inline bump in `LakeTableSchemaChangeJob.cancelImpl` and the replay branch) instead of calling full `updateVisibleVersion()`, and add a UT asserting `getMetadataSwitchVersion()` stays 0 after a force-cancelled file_bundling alter.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJobBase.java
```diff
@@ -525,6 +580,63 @@ protected boolean cancelImpl(String errMsg) {
...
+
+            // Follow the same single-vs-aggregate dispatch as the normal
+            // lakePublishVersion() so that the no-op publish writes metadata
+            // in the format consistent with how the partition's data is
+            // already stored on OSS. Without this, an alter on a
+            // file_bundling=true table would have its no-op publish written
+            // per-tablet while subsequent loads expect a bundle file,
+            // forcing BE into the (working but fragile) mixed-format
+            // recovery code path.
+            boolean useAggregatePublish = enableFileBundling() || (isFileBundling && !disableFileBundling());
```
kevincai: (Low-Medium / Question) This dispatch keys off the job's *target* bundling state, not the partition's *current* format. For an `ENABLE_FILE_BUNDLING` (enabling) alter being force-cancelled, `enableFileBundling()` is true, so the no-op publish is sent as an aggregate/bundle write -- but the partition's existing V-1 data is still per-tablet (the enable never took effect). The comment argues this matches how the partition's data is already stored, yet for a cancel the already-stored format is the pre-alter one. Can you confirm BE writes a correct V from V-1 content in this case, or should the dispatch key off the current/pre-alter format rather than the alter's target?

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJobBase.java
luohaha: Agreed on both points — fixed in f7d647d.

**metadataSwitchVersion divergence:** you're right that the live path's `updateVisibleVersion(table)` recorded `setMetadataSwitchVersion(commitVersion)` for an `ENABLE_FILE_BUNDLING` alter while the replay branch did not. A force-cancel discards the alter (the `no_op_publish` writes V-1 content tagged as V), so no format switch happened at `commitVersion` and recording one is wrong — it would diverge leader vs. replay and needlessly pin vacuum's retained version. I extracted `advanceVisibleVersionForForceSkip()` which bumps the visible version only (no `metadataSwitchVersion`) and is now called from **both** the live cancel callback and the replay CANCELLED branch, so they are byte-identical.

**Hard precondition:** also addressed — the new helper uses a soft `if (getVisibleVersion() == commitVersion - 1)` guard instead of `Preconditions.checkState`, exactly because it runs inside the edit-log applier after the CANCELLED entry is already journaled; a thrown precondition there would leave the live FE inconsistent with its own journal.

Added `testForceCancelFileBundlingAlterDoesNotSetMetadataSwitchVersion` asserting `getMetadataSwitchVersion()` stays 0 after a force-cancelled file_bundling alter.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJobBase.java
luohaha: Good catch — keying off the alter's *target* format was wrong for a cancel. Fixed in f7d647d to dispatch on the partition's *current* (pre-alter) format via `table.isFileBundling()`.

Rationale: a cancel writes V from V-1 content, so V must be written in V-1's format. The alter never reached FINISHED, so `updateCatalog()` has not flipped the table's `file_bundling` flag yet — `table.isFileBundling()` is therefore the current/pre-alter format. So:
- enabling alter cancelled while V-1 is still per-tablet → `useAggregatePublish=false` (per-tablet), matching the data;
- disabling alter cancelled while V-1 is still bundled → `useAggregatePublish=true` (aggregate), matching the data.

This avoids asking BE to write an aggregate bundle over per-tablet V-1 content (your concern). The two dispatch UTs now key off the table's current `file_bundling`.

CelerData-Reviewer:
@codex review

banmoy:
LakeRollupJob also needs force cancel

luohaha:
@banmoy No in this pr

kevincai (approved):

xiangguangyxg (approved):

luohaha:
### Not backporting to branch-4.1 (closing #74210)

The mergify backport (#74210) hit cherry-pick conflicts that turned out to be **structural, not mechanical**, so we are intentionally not backporting this to branch-4.1 for now.

**Why:** phase-2 is built on main's alter-job persistence refactor, which branch-4.1 does not have:

- `AlterJobV2.copyForPersist()` / `copyBaseFields()`
- `AlterJobV2.persistStateChange(...)`
- the `AlterJobV2(AlterJobV2)` copy constructor
- the serialized `forceSkippedAtCommitted` marker (the leader/replay-consistency core of this feature)

branch-4.1 still uses the older `writeEditLog(this)` model (it logs the live job object directly instead of a `copyForPersist()` snapshot). The conflicting files were exactly the persistence-dependent core (`AlterJobV2`, `LakeTableSchemaChangeJob`, `LakeTableAlterMetaJobBase`, `Utils`); the SQL plumbing (grammar / analyzer / handler / `CancelAlterTableStmt`) cherry-picked cleanly.

**The trade-off:** a correct 4.1 backport would require one of:

1. **Re-implementing** phase-2 against 4.1's older `writeEditLog(this)` framework — re-deriving the leader/replay version-bump consistency by hand and re-verifying the no-op-publish path on a 4.1 shared-data cluster; or
2. **First backporting the persistence refactor** to 4.1, then cherry-picking this on top.

Both are high-risk on a release branch relative to the benefit (this is an operator-only escape hatch gated behind `enable_admin_skip_committed_txn`). Taking the incoming side as-is is not viable either — it introduces `abstract copyForPersist()` and breaks every alter subclass in 4.1.

So we are leaving 4.1 on main only for now and can revisit if a 4.1 user actually needs it.

PR #73833, opened 2026-05-25
[BugFix] Avoid external metadata calls during stats replay

### LINKED ISSUE #73832: [Bug] FE journal replayer can block on external statistics cache refresh

This is easiest to observe with an external catalog table when the connector metadata path is slow or stuck.

1. Configure an external Iceberg/Hive catalog.
2. Run external table statistics collection so the leader writes external statistics metadata journals such as `OP_ADD_EXTERNAL_BASIC_STATS_META` or `OP_ADD_EXTERNAL_HISTOGRAM_STATS_META`.
3. On a follower/observer FE, make the external catalog table resolution path slow or unavailable, for example HMS/HDFS metadata lookup blocks while resolving the table.
4. Let the FE replay the external statistics metadata journal.

Observed stack shape from a production jstack, sanitized:

```text
"replayer" ...
  at com.starrocks.connector.CachingIcebergCatalog.getTable(...)
  at com.starrocks.connector.iceberg.IcebergMetadata.getTable(...)
  at com.starrocks.server.MetadataMgr.getTable(...)
  at com.starrocks.statistic.AnalyzeMgr.refreshConnectorTableBasicStatisticsCache(...)
  at com.starrocks.persist.EditLog.loadJournal(...)
```

The blocking metadata call was in the path used by `OP_ADD_EXTERNAL_BASIC_STATS_META`. The same replay pattern exists for external basic stats add/remove and external histogram add/remove.

Journal replay should not depend on external catalog/HMS/HDFS availability. Replaying external statistics metadata should update FE in-memory metadata and keep statistics caches correct without blocking the replay thread on connector table resolution.

### Real behavior (Required)

`EditLog.loadJournal()` calls `AnalyzeMgr.refreshConnectorTableBasicStatisticsCache()` / `expireConnectorTableAndColumnStatistics()` / histogram variants during replay. Those methods call `MetadataMgr.getTable()`, which can block on external catalog or HDFS metadata I/O. If it blocks, the FE journal replayer stops progressing and the follower/observer can fall behind.

Observed on StarRocks 4.0.10. The affected replay code path is also present on current main at the time of reporting.

### Proposed fix

During replay, avoid resolving external table metadata and avoid synchronous/async refresh work that can block, reorder, or be dropped. Replay should:

1. Apply the external statistics metadata journal synchronously to `AnalyzeMgr`.
2. Synchronously invalidate connector statistics caches without external I/O.
3. Let normal query-time cache loading refresh statistics lazily.
4. Expose `replayed_journal_id` as an FE metric to make replay stalls observable.

yakirgb:
## Why I'm doing:

Fixes #73832

FE journal replay currently refreshes or expires external table statistics caches by resolving the external table through `MetadataMgr.getTable()`. For Iceberg/Hive catalogs this can call HMS/HDFS during replay. If the external metadata path is slow or stuck, the FE replayer thread can stop progressing.

## What I'm doing:

- Avoid resolving external table metadata during replay of external basic stats and external histogram stats journals.
- Replay the external stats metadata synchronously, then synchronously invalidate connector statistics caches without external I/O.
- Let connector statistics reload lazily through the normal query/cache path.
- Add `starrocks_fe_replayed_journal_id` so replay progress is observable.
- Add FE unit tests covering replay add/remove for external basic stats and histograms, plus the new replayed journal metric.

Fixes #73832

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

## Test:

```bash
cd fe && mvn -pl fe-core -am -Dant.python=python3 -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -Dtest='com.starrocks.metric.MetricRepoTest#testReplayedJournalIdMetric,com.starrocks.persist.EditLogTest#testReplayExternalBasicStatsInvalidatesCacheWithoutRefresh+testReplayExternalHistogramStatsInvalidatesCacheWithoutRefresh' test
```


CelerData-Reviewer:
@codex review

yakirgb:
Addressed the concrete review items in `1cf9011f53`:

- Added the `starrocks_fe_replayed_journal_id` entry to the Japanese metric docs.
- Added the missing `@Test` annotation for `testPlanAdvisorMetricsExposure()`.

Verification:

```bash
cd fe && mvn -pl fe-core -am -Dant.python=python3 -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -Dtest='com.starrocks.metric.MetricRepoTest#testReplayedJournalIdMetric+testPlanAdvisorMetricsExposure' test
```

Result: `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`.

For the cache invalidation comments: I kept the current full connector stats invalidation in this PR. The granular cache key is `ConnectorTableColumnKey(table.getUUID(), column)`, but the external stats journal only carries catalog/db/table/column. Reconstructing the exact table UUID would require resolving the external `Table` again; for Iceberg/Paimon/Hive-like tables that can re-enter the metadata path this PR is specifically removing from FE replay. Batching/deduping invalidation at the replay-loop level could be a follow-up, but I kept this bug fix focused on removing blocking external metadata calls from journal replay.

fe/fe-core/src/main/java/com/starrocks/persist/EditLog.java
```diff
@@ -985,50 +985,36 @@ public void loadJournal(GlobalStateMgr globalStateMgr, JournalEntity journal)
...
-                    // We don't need to refresh statistics when checkpointing
                     if (!GlobalStateMgr.isCheckpointThread()) {
-                        globalStateMgr.getAnalyzeMgr().refreshConnectorTableHistogramStatisticsCache(
-                                histogramStatsMeta.getCatalogName(), histogramStatsMeta.getDbName(),
-                                histogramStatsMeta.getTableName(),
-                                Lists.newArrayList(histogramStatsMeta.getColumn()), true);
+                        // Do not resolve external table metadata during replay. Connector metadata calls may block
+                        // on HMS/HDFS. The journal updates stats metadata synchronously; caches reload lazily.
+                        globalStateMgr.getStatisticStorage().expireAllConnectorHistogramStatistics();
                     }
```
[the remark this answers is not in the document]
yakirgb: The replay-cost concern is valid, but I kept this as full-cache invalidation deliberately for this bug fix.

The granular histogram cache key is `ConnectorTableColumnKey(table.getUUID(), column)`. The external histogram journal only carries catalog/db/table/column, not the exact connector table UUID. For Iceberg/Paimon/Hive-like tables, reconstructing that UUID requires resolving the external `Table`, which can call back into HMS/HDFS and reintroduce the replay-blocking path this PR removes.

So this path trades repeated memory-only invalidation for avoiding external metadata I/O on the replayer thread. Batching/deduping invalidation at the replay-loop level is a reasonable follow-up, but I would not add that extra replay plumbing in this focused fix.

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
```diff
@@ -532,6 +532,15 @@ public Long getValue() {
         };
         STARROCKS_METRIC_REGISTER.addMetric(maxJournalId);
 
+        GaugeMetric<Long> replayedJournalId = (GaugeMetric<Long>) new GaugeMetric<Long>(
+                "replayed_journal_id", MetricUnit.NOUNIT, "replayed journal id of this frontend") {
+            @Override
+            public Long getValue() {
+                return GlobalStateMgr.getCurrentState().getReplayedJournalId();
```
kevincai, line 575: what will this metric be on the FE leader node? always 0?

fe/fe-core/src/main/java/com/starrocks/persist/EditLog.java
```diff
@@ -985,50 +985,36 @@ public void loadJournal(GlobalStateMgr globalStateMgr, JournalEntity journal)
...
-                    // statistic, and the follower's should refresh cache here.
-                    // We don't need to refresh statistics when checkpointing
                     if (!GlobalStateMgr.isCheckpointThread()) {
-                        globalStateMgr.getAnalyzeMgr().refreshConnectorTableBasicStatisticsCache(
-                                basicStatsMeta.getCatalogName(),
-                                basicStatsMeta.getDbName(), basicStatsMeta.getTableName(),
-                                basicStatsMeta.getColumns(), true);
+                        // Do not resolve external table metadata during replay. Connector metadata calls may block
+                        // on HMS/HDFS. The journal updates stats metadata synchronously; caches reload lazily.
+                        globalStateMgr.getStatisticStorage().expireAllConnectorTableColumnStatistics();
```
kevincai: looks like brutal to invalidate all tables/columns stats for all dbs, could be possibly move the cache refresh out of the current replay thread, do it in a dedicated thread pool. A sudden invalidation of all the caches, could hurt the system badly.

yakirgb: Fixed in `ad48b5471e`. I removed the full connector stats/histogram cache invalidation path from replay. The journal metadata update still happens synchronously, but connector table resolution and the targeted refresh/expire work now run in a dedicated `replay-external-stats-cache-pool` thread. That keeps HMS/HDFS metadata calls off the replay thread while preserving the previous per-table/per-column cache scope instead of invalidating all connector stats caches.

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
yakirgb: Fixed in `ad48b5471e`. On the leader this now reports `getMaxJournalId()`; on non-leader FEs it reports `getReplayedJournalId()`. The leader applies writes directly instead of advancing through follower replay, so using max journal ID avoids showing a misleading zero/stale replay value on the active leader. I also updated the metric docs to describe it as the latest applied journal ID.

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
```diff
@@ -532,6 +532,16 @@ public Long getValue() {
         };
         STARROCKS_METRIC_REGISTER.addMetric(maxJournalId);
 
+        GaugeMetric<Long> replayedJournalId = (GaugeMetric<Long>) new GaugeMetric<Long>(
```
kevincai: `LeaderAwareGaugeMetricLong` can help with the metric along with `is_leader` label to distinguish if the metric is from leader.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
yakirgb: Fixed in `75d645550e`. The metric now uses `LeaderAwareGaugeMetricLong`, so it exports the `is_leader` label. I overrode both paths: leaders return `getMaxJournalId()`, and non-leaders return `getReplayedJournalId()` instead of the default non-leader `0`. Docs now list the `is_leader` label, and `MetricRepoTest#testReplayedJournalIdMetric` checks the Prometheus label output.

kevincai (dismissed):

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/persist/EditLog.java
```diff
@@ -985,50 +985,38 @@ public void loadJournal(GlobalStateMgr globalStateMgr, JournalEntity journal)
...
                     globalStateMgr.getAnalyzeMgr().replayAddExternalBasicStatsMeta(basicStatsMeta);
-                    // The follower replays the stats meta log, indicating that the master has re-completed
-                    // statistic, and the follower's should refresh cache here.
-                    // We don't need to refresh statistics when checkpointing
                     if (!GlobalStateMgr.isCheckpointThread()) {
-                        globalStateMgr.getAnalyzeMgr().refreshConnectorTableBasicStatisticsCache(
-                                basicStatsMeta.getCatalogName(),
-                                basicStatsMeta.getDbName(), basicStatsMeta.getTableName(),
-                                basicStatsMeta.getColumns(), true);
+                        // Do not resolve external table metadata during replay. Connector metadata calls may block
```
stephen-shelby, line 989: Before settling on the fix, could you share some numbers on the actual replay impact?

How long does a single table block on getTable during replay (normal vs. slow metastore)?
Does every external table stall, or only on a connector-metadata cache miss?
Is it reproducible steadily, or only when HMS / object storage is degraded?
This tells us whether we need to avoid getTable on replay entirely, or just add a timeout.

stephen-shelby: Thanks for working on this. Two concerns with the background-refresh approach: the queue is unbounded + single-threaded (newDaemonFixedThreadPool(1, Integer.MAX_VALUE, ...)), so tasks pile up if the metastore is slow/down; and it still calls getTable (just on the background thread), so it keeps retrying against a degraded metastore and never drains.

Also, getTable isn't slow in every case — it only goes remote on a connector-metadata cache miss against a slow/unreachable metastore, so most replays won't block here.

We opened [#74371](https://github.com/StarRocks/starrocks/issues/74371) to mitigate the same issue behind an opt-in config (enable_external_stats_lazy_refresh_on_replay, default off): persist the table UUID and invalidate the cache by UUID on replay — no getTable, nothing to queue. Might be worth comparing.

gengjun-git:
Why not simply invalidate the cache directly and leave the refresh work to the caller?

fe/fe-core/src/main/java/com/starrocks/persist/EditLog.java
yakirgb: Real impact from the incident this PR was opened for (Taboola, 4.0.10, shared-data, ~40 CNs):

- Jstack confirmed the replayer thread BLOCKED on a per-key Caffeine monitor inside `CachingIcebergCatalog.getTable`, while another thread (a user `SHOW COLUMNS` from the mysql nio pool) held that monitor parked inside `HadoopInputFile.getLength → $Proxy43.getFileInfo` against a silent NameNode. Replayer was blocked ~7h before we noticed; `starrocks_fe_max_journal_id` kept advancing (BDBJE replication thread is independent of the application replayer), so the standard metric did not surface it.
- Not every external-table replay stalls — only on a connector-metadata cache miss against a degraded/silent HMS/HDFS. In our case the underlying amplifier was `ipc.client.rpc-timeout.ms=0` in the mounted `core-site.xml` (separate config bug fixed on our side), so the NN socket read blocked forever.
- A `getTable` timeout alone wouldn't fully fix this: the Caffeine per-key monitor (`computeIfAbsent`) is held across the loader's I/O, so even a fast-fail loader still holds the monitor; the replayer would block on the same key. The structural fix is to keep `getTable` off the replayer thread entirely.

yakirgb: Agreed on both concerns:

1. The single-threaded `Integer.MAX_VALUE` queue is the wrong shape. Under a degraded HMS, tasks accumulate and the submitting replayer eventually blocks on enqueue too — exactly what the PR was trying to avoid.
2. Right — pushing `getTable` to a background thread relocates the hang, it doesn't eliminate it. Retries against a dead HMS never drain and the cache stays stale either way.

#74371's approach (persist `tableUUID` on the leader at analyze time, invalidate the connector cache by UUID on replay — no `getTable`, no queue) is the cleaner fix. It also matches @gengjun-git's comment on this PR ("invalidate directly and leave refresh to the caller") and how internal-table stats replay already works (invalidate by table id).

Happy to close #73833 in favor of #74371. One ask: keep the `starrocks_fe_replayed_journal_id` gauge (with the `is_leader` label) — that was the missing observability that made the incident invisible. Can ride along on #74371 or split as a standalone follow-up.

yakirgb:
@gengjun-git that's the right shape. The blocker for a plain invalidate path is that the granular cache key is `ConnectorTableColumnKey(table.getUUID(), column)` (`CachedStatisticStorage.java:215`) and the journal entry only carries catalog/db/table/column — so per-table invalidate currently still needs `getTable` to recover the UUID, which puts us back on the same replayer hang.

#74371 closes that gap by persisting the UUID into `ExternalBasicStatsMeta` / `ExternalHistogramStatsMeta` at analyze time on the leader, so replay can invalidate by UUID directly with no `getTable` call. I'll pivot this PR to that approach (or close in favor of #74371 if it lands first).

kevincai:
favor of https://github.com/StarRocks/starrocks/pull/74371 and close this PR.

PR #73834, opened 2026-05-25
[Feature] add BigQuery JDBC catalog support

yuriolive:
## Why I'm doing:

BigQuery is widely used but not supported as a JDBC catalog source in StarRocks. This blocks users who want to federate queries from StarRocks into BigQuery datasets without a full ETL pipeline.

## What I'm doing:

Adds a new `BigQuerySchemaResolver` and wires it into the JDBC catalog factory, enabling StarRocks to connect to Google BigQuery via JDBC.

Compatible with the Simba/Google BigQuery JDBC driver (`com.simba.googlebigquery.jdbc.Driver` / `com.simba.googlebigquery.jdbc42.Driver`).

**Type mapping:**

| BigQuery type | StarRocks type | Notes |
|---|---|---|
| INT64 | BIGINT | |
| FLOAT64 | DOUBLE | |
| BOOL | BOOLEAN | Handles both `Types.BIT` and `Types.BOOLEAN` (Simba driver quirk) |
| NUMERIC(≤38,d) | DECIMAL(p,s) | |
| BIGNUMERIC | VARCHAR | Precision > 38 cannot fit in DECIMAL(38,x) |
| STRING | VARCHAR(max) | |
| BYTES | VARBINARY | |
| DATE | DATE | |
| TIME | TIME | |
| DATETIME | DATETIME | Civil time, no TZ conversion |
| TIMESTAMP | DATETIME | UTC epoch — converted to query timezone in JDBCScanner |
| JSON | JSON | |
| STRUCT / ARRAY / GEOGRAPHY / INTERVAL | VARCHAR | Serialised string representation |

**`JDBCScanner` changes:** BigQuery `TIMESTAMP` columns are UTC epoch. Added `isBigQueryDriver` flag and per-column tracking so `TIMESTAMP` values are shifted from UTC to the session query timezone (same mechanism already used for Postgres `timestamptz`). BigQuery `DATETIME` (civil time) columns are explicitly excluded from this conversion.

**Usage:**
```sql
CREATE EXTERNAL CATALOG bigquery_catalog
PROPERTIES (
  "type" = "jdbc",
  "jdbc.driver_class" = "com.simba.googlebigquery.jdbc.Driver",
  "jdbc.driver_url" = "https://<host>/GoogleBigQueryJDBC42.jar",
  "jdbc.uri" = "jdbc:bigquery://https://www.googleapis.com/bigquery/v2:443;ProjectId=my-project;OAuthType=3",
  "jdbc.user" = "",
  "jdbc.password" = ""
);
```

Auth lives entirely in the JDBC URL (`OAuthType`, `OAuthPvtKeyPath`, etc.).

image

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

## Files changed

- `fe/.../connector/jdbc/BigQuerySchemaResolver.java` — new resolver (~190 lines)
- `fe/.../connector/jdbc/JDBCMetadata.java` — factory wiring (3 edits, +5 lines)
- `java-extensions/.../jdbcbridge/JDBCScanner.java` — BigQuery TIMESTAMP UTC conversion
- `fe/.../connector/jdbc/BigQuerySchemaResolverTest.java` — 19 FE tests
- `java-extensions/.../jdbcbridge/JDBCScannerTest.java` — 8 scanner tests for BigQuery TIMESTAMP handling

## Known limitations

- BigQuery native partitioning is not exposed via JDBC metadata; whole table is treated as one partition
- STRUCT and ARRAY columns are returned as VARCHAR (serialised by the JDBC driver); no nested type support
- Auth is JDBC-URL-only — convenience properties for service account JSON path are not yet added


fe/fe-core/src/main/java/com/starrocks/connector/jdbc/JDBCMetadata.java
```diff
@@ -137,6 +137,8 @@ private JDBCSchemaResolver createSchemaResolver() {
             return new OracleSchemaResolver(properties);
         } else if (driverClass.contains("sqlserver")) {
             return new SqlServerSchemaResolver();
+        } else if (driverClass.contains("bigquery") || driverClass.contains("googlebigquery")) {
```
GavinMar: These two checks are somewhat redundant. If `contains("googlebigquery")` is true, `contains("bigquery")` will also definitely be true.

java-extensions/jdbc-bridge/src/main/java/com/starrocks/jdbcbridge/JDBCScanner.java
```diff
@@ -79,6 +81,8 @@ public JDBCScanner(String driverLocation, JDBCScanContext scanContext) {
         this.scanContext = scanContext;
         this.isOracleDriver = scanContext.getDriverClassName().toLowerCase(Locale.ROOT).contains("oracle");
         this.isPostgresDriver = scanContext.getDriverClassName().toLowerCase(Locale.ROOT).contains("postgresql");
+        String driverClassLower = scanContext.getDriverClassName().toLowerCase(Locale.ROOT);
+        this.isBigQueryDriver = driverClassLower.contains("bigquery") || driverClassLower.contains("googlebigquery");
```
GavinMar: ditto

java-extensions/jdbc-bridge/src/main/java/com/starrocks/jdbcbridge/JDBCScanner.java
```diff
@@ -297,6 +314,11 @@ public List<Object[]> getNextChunk() throws Exception {
                         Timestamp timestampValue =
                                 resultObject instanceof Timestamp ? (Timestamp) resultObject : resultSet.getTimestamp(i + 1);
                         dataColumn[resultNumRows] = convertPostgresTimestampWithTimezoneValue(timestampValue);
+                    } else if (shouldConvertBigQueryTimestampColumn(i)) {
+                        Timestamp timestampValue =
+                                resultObject instanceof Timestamp ? (Timestamp) resultObject : resultSet.getTimestamp(i + 1);
+                        // BigQuery TIMESTAMP is UTC; reuse the same UTC→queryTZ adjustment.
+                        dataColumn[resultNumRows] = convertPostgresTimestampWithTimezoneValue(timestampValue);
```
GavinMar: The function name `convertPostgresTimestampWithTimezoneValue` seems inappropriate now; it's recommended to change it to a more generic name.

yuriolive: Fixed

java-extensions/jdbc-bridge/src/main/java/com/starrocks/jdbcbridge/JDBCScanner.java
yuriolive: Fixed

fe/fe-core/src/main/java/com/starrocks/connector/jdbc/JDBCMetadata.java
yuriolive: Fixed

GavinMar (approved):

anthonyhungnguyen:
We deployed this PR (rebased onto a 4.1.x release) to a production cluster against real BigQuery and hit two issues worth folding into the PR:

**1. `getTables`/`getColumns` return empty results with the current Simba driver (1.7.0.1001) when an explicit catalog is passed.**

`BigQuerySchemaResolver` calls `connection.getMetaData().getTables(connection.getCatalog(), dbName, ...)`. Against driver 1.7.0.1001 this always returns zero rows — `SHOW TABLES` is empty for every dataset even though `SHOW DATABASES` works. Passing `null` as the catalog (JDBC "current catalog", i.e. the URI `ProjectId`) fixes it:

```java
connection.getMetaData().getTables(null, dbName, null, defaultTableTypes);
connection.getMetaData().getColumns(null, dbName, tblName, "%");
```

Matrix we tested against a live project: `("<project>", null)` → 0 rows, `("<project>", "%")` → 0 rows, `(null, null)` → all tables, `(null, "%")` → all tables. Older driver versions may have accepted the explicit catalog, but the current downloadable driver does not.

**2. FE-side driver loading needs documenting.**

The FE resolves the driver via `Class.forName` from its own classpath (like the bundled mariadb/ojdbc drivers) — `driver_url` is only downloaded for checksum/BE distribution, never loaded into the FE process. So the Simba jar has to be placed in `fe/lib`. And since the driver's dependency tree (protobuf 4.x, grpc 1.76, gax 2.74, guava 33) collides with FE's own versions, a plain jar-of-all-deps fails at class initialization (`Could not initialize class com.google.api.gax.rpc.EndpointContext` on connect). We had to repackage the driver as a shaded jar with all dependencies relocated (leaving only `com.simba.*` public) before FE-side catalog creation worked. Happy to share the shade config if useful.

PR #73886, opened 2026-05-26
[Refactor] Extract libcurl global init/cleanup into CurlGlobalGuard RAII

Hyper-FF:
## Why I'm doing:

curl_global_init was called in main without a matching curl_global_cleanup; wrap it in a stack-allocated RAII guard so cleanup runs on normal shutdown and the init/cleanup pair lives in one place. Mirrors the AwsSdkGuard approach already used for the AWS SDK.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/util/curl_global_guard.cpp
+++ b/be/src/util/curl_global_guard.cpp
@@ -1,0 +1,33 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "util/curl_global_guard.h"
+
+#include <curl/curl.h>
+
+namespace starrocks {
+
+int CurlGlobalGuard::init() {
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

be/src/util/curl_global_guard.cpp
```diff
@@ -12,13 +12,22 @@
...
+
+int CurlGlobalGuard::init() {
+    auto ret = curl_global_init(CURL_GLOBAL_ALL);
+    _initialized = (ret == CURLE_OK);
+    return ret;
 }
+
+CurlGlobalGuard::~CurlGlobalGuard() {
+    if (_initialized) {
+        curl_global_cleanup();
```
kevincai, line 29: This may crash the process because the instance will be destroyed when leaving the main(), However, other global static variables/objects can still reference the object for pending ops.

be/src/util/curl_global_guard.h
```diff
@@ -14,23 +14,27 @@
...
-#include "gen_cpp/FrontendService_types.h"
-
 namespace starrocks {
 
-class SchemaCollationCharacterSetApplicabilityScanner : public SchemaScanner {
+// RAII guard around curl_global_init / curl_global_cleanup. Call init() once
+// at process startup; failure to initialize is reported via the return code so
+// the caller can choose how to fail. The destructor runs curl_global_cleanup
+// only if init() succeeded.
+class CurlGlobalGuard {
```
alvin-phoenix-ai, line 23: Please move those files into `base/` directory.

PR #73907, opened 2026-05-27
[BugFix] Track physical rowid base for shared-segment PK index publish (backport #73686)

xiangguangyxg:
Backport of #73686 to branch-4.1.

## Why I'm doing:

Tablet split in shared-data mode lets a child tablet share the parent's segment files. The publish path reads such shared segments via a range-pruned `ChunkIterator` that starts at a non-zero physical offset in the source segment file, while reload reads back by physical rowid. Publish was encoding each row's segment position as `chunk_row_offset_in_segment + i_in_chunk`, which equals the chunk-local index for non-shared segments (where the iterator starts at row 0) but skips the range-start offset for shared ones — producing wrong `rss_rowid` entries and `PersistentIndexMemtable<12> insert found duplicate key` errors on subsequent publish.

## What I'm doing:

`SegmentPKIterator` captures `_segment_physical_rowid_offset` from the underlying iterator's first emitted rowid and propagates it through `SegmentPKChunkRef`. Each `SegmentPKChunkRef` carries two offsets:
- `physical_rowid_offset` — chunk[0]'s position in the source segment file (for PK index upserts, delvec, and `fetch_values_by_rowid` full-segment reads).
- `logical_rowid_offset` — chunk[0]'s position in the iterator's emit sequence (for consumers that index into iterator-emitted range-filtered chunks).

Three enforcement layers guard the contiguity assumption (release-safe `Status::InternalError` on non-contiguous rowids, 8 boundary DCHECKs on `SegmentReadOptions`). Condition update phase 2 uses a new sparse `PrimaryIndex::upsert(rssid, physical_rowid_offset, indices, pks, ...)` overload. Column-mode partial update threads the per-segment base through `ColumnPartialUpdateState::upt_segment_physical_rowid_offset`.

See #73686 for full design discussion and the redesign rationale.

## Backport notes

Adapted to branch-4.1 during cherry-pick:
- `primary_index.cpp` sparse upsert: branch-4.1's `build_persistent_keys` returns a raw `const Slice*` (main returns `StatusOr`), so the `ASSIGN_OR_RETURN` was changed to a direct assignment.
- Conflict in the `PrimaryIndex::upsert` body resolved in favor of the new sparse implementation (branch-4.1 still had the old vector-overload body).

All 8 `SegmentReadOptions` fields referenced by the boundary DCHECKs, `ChunkHelper::new_chunk`, and the `pk_index_parallel_execution_min_rows` / `pk_column_lazy_load_threshold_bytes` config entries exist on branch-4.1.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [x] This is a backport pr


xiangguangyxg:
Closing in favor of mergify auto-backport #74031 (now CI-fixed). Both target branch-4.1; keeping one.

PR #73911, opened 2026-05-27
[Feature] Add rewrite_equality_delete_files Iceberg table procedure

### LINKED ISSUE #73947: [Feature Request] Iceberg: rewrite_equality_delete_files procedure to compact equality deletes into position deletes
### Problem

StarRocks reads Iceberg v2 tables with equality-delete files by applying them as a LEFT ANTI JOIN at query time. Equality deletes accumulate over time, so every read keeps paying the anti-join cost, and there is no way from StarRocks to compact them away.

### Proposal

Add an Iceberg table procedure `rewrite_equality_delete_files()`, invoked via:

```sql
ALTER TABLE <iceberg_table> EXECUTE rewrite_equality_delete_files();
```

It converts the live equality-delete files of the current snapshot into position-delete files. The set of visible rows stays identical (deleted keys, NULL included, must not resurrect), but subsequent reads no longer pay the equality-delete anti-join.

Tracked by #73911.

eshishki:
> **Stacked on #73900** (#67321 has merged). Until it merges, this PR's diff includes its commits (partition-scope read fix + its temporary fixture commit). Please review/merge bottom-up; I'll rebase as it lands. The convert-specific commits are the top two: `Extract shared builders` → `Add procedure`.

## Why I'm doing:

Iceberg v2 tables accumulate equality-delete files (CDC/Flink upserts). Every read applies them with an anti-join that gets more expensive as they pile up, and StarRocks today has no way to compact them. This adds an admin procedure that converts equality deletes into position deletes — reads stop paying the anti-join, and the resulting position deletes can later be cleaned by normal file maintenance.

## What I'm doing:

Add the `rewrite_equality_delete_files` Iceberg table procedure:

```sql
ALTER TABLE catalog.db.tbl EXECUTE rewrite_equality_delete_files();
```

It converts a table's live equality-delete files into position-delete files by reusing the StarRocks join engine, then atomically removes the equality deletes and adds the position deletes in a single `RewriteFiles` commit.

The conversion is the dual of the read path: per `(equalityIds, scope)` a broadcast **LEFT SEMI JOIN** keeps exactly the rows an equality delete removes — null-safe key equality plus `data.$data_sequence_number < delete.$data_sequence_number`, and `$partition_id` for partitioned-spec deletes so a delete only matches its own partition. The match predicate and GLOBAL/PARTITIONED scope split are shared with the read path via `IcebergEqualityDeleteScanBuilder` (extracted here) so the two paths can't drift. The join projects `$file_path/$pos` (+ partition cols) into an `IcebergDeleteSink`; multiple legs are UNIONed (distinct).

Adding the position deletes and removing the equality deletes happens in one atomic `RewriteFiles` commit, through a new class **`ConvertEqualityDeleteRewriteFiles`** (subclass of Iceberg's `BaseRewriteFiles`).

Why a subclass. A concurrent commit could let deleted rows come back two ways. The stock `RewriteFiles` already handles one: if someone concurrently removes the equality-delete files we're dropping, `failMissingDeletePaths` aborts the commit. It does **not** handle the other: a concurrent rewrite can compact away a data file our new position deletes point at, and then those positions mask nothing — the rows resurface. Our subclass adds that missing check in the commit retry loop — if a referenced data file is gone, the commit fails and retries.

The files we remove are exactly the equality deletes the scan actually applied. v1 rewrites the whole table (no `WHERE`).

## Tests

- FE unit tests: the commit-validation semantics (`ConvertEqualityDeleteRewriteFiles`), the conversion job, and the planner shape.
- `test_iceberg_rewrite_equality_delete_files` — e2e SQL test in the `register_table` form against the `eq_del_convert` fixture **already hosted on the CI OSS bucket for #67321** (`iceberg_ci_db/eq_del_convert/metadata/v4.metadata.json`): 6 visible rows before and after the conversion; NULL-keyed deletes must not resurrect. `register_table` pins `v4.metadata.json`, so each run's conversion commit advances only its own per-run catalog entry and never mutates the shared fixture files (the local `file://` fixture and its checked-in binaries are gone).

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

Fixes #73947


CelerData-Reviewer:
@codex review

eshishki:
Heads-up on the SQL integration test (`test_iceberg_rewrite_equality_delete_files`) — it will be **red in CI and can't be turned green from my side**, same root cause as #67321.

The fixture is a local `file://` hadoop catalog: the case does `shell: cp … /tmp/sr_eq_fixture/warehouse`, then `create external catalog … "iceberg.catalog.warehouse"="file:///tmp/sr_eq_fixture/warehouse"`. In SQL-Tester CI the cluster is multi-node (FE and BE on separate nodes) and `shell:` runs via `subprocess` on the GitHub runner — a third machine. So the `cp` lands on the runner's `/tmp`, the FE reads `file:///tmp/...` from its own node where the namespace dir doesn't exist → analysis fails with `Unknown database`. Not a code bug — it works end-to-end on a single-node deployment, which is where I validated it.

No test tag fixes this (`@native`/`@cloud` select storage mode, not node count); every iceberg fixture test that passes in CI builds its table over an `oss://${oss_bucket}` warehouse. Hosting this in the CI external HMS+OSS catalog needs access I don't have — pointers welcome.

PR #73962, opened 2026-05-28
[BugFix] Re-record per-partition coordinator claim on every sender's open

luohaha:
## Why I'm doing:

In shared-data mode, the lake per-partition coordinator added in #71992 elects the smallest `sender_id` whose open RPC recorded a claim for a given partition (`LakeTabletsChannel::_partition_coordinator`). Only the elected coordinator returns `txn_logs` in its EOS response, and the FE-coord `OlapTableSink` of that sender writes the partition's `combined_txn_log` file to OSS.

The election only works if **every** sender's claim makes it into `_partition_coordinator`. It doesn't today: `LoadChannel::open` (`be/src/runtime/load_channel.cpp`) creates the `LakeTabletsChannel` and calls `channel->open(...)` for the **first** sender that lands on a given `TabletsChannelKey`; every subsequent non-incremental open finds the channel already in `_tablets_channels` and falls into the no-op default branch — so `_record_coordinator_claims` never sees those senders' params.

For a non-partitioned shared-data PK table with pipeline parallelism N, all N senders share the same `TabletsChannelKey` (only the doubleWrite optimize-table path sets non-zero `sink_id`). So on each storage BE only the first sender to arrive is recorded. When that first opener is not sender 0 — which happens whenever RPC scheduling reorders opens, even on the local NodeChannel — `_should_enter_collect_path(0)` returns false at EOS time, and that BE's `txn_logs` get returned in the response of whichever sender did win the race. Across the storage BEs of one load, multiple `OlapTableSink` instances end up with different subsets of the partition's logs and all of them race to write the same shared-data OSS `combined_txn_log` path under `CREATE_OR_OPEN_WITH_TRUNCATE`. The last writer wins and the file ends up missing every losing sink's NodeChannel-worth of tablets.

Downstream, the `aggregate_publish_version` for that partition then trips `txn log list does not contain txn log of tablet` (transactions.cpp:158), `publish_version` returns InternalError, and the partition retries forever:

```
ERROR PublishVersionDaemon.publishPartitionBatch
  Fail to publish partition <pid> ... error msg: txn log list does not contain txn log of tablet <tid>
```

I observed this on a 4.0.10 cluster after the cherry-pick of #71992: ~3% of `.logs` files in one partition were missing 1–10 NodeChannels' worth of tablet entries, with the lost groups exactly matching the 2-tablet-per-BE assignment.

## What I'm doing:

Add a virtual `TabletsChannel::update_open(const PTabletWriterOpenRequest&)` with a no-op default. Override it on `LakeTabletsChannel` to invoke `_record_coordinator_claims(params)`. In `LoadChannel::open`'s `else-if (is_incremental)` chain, when the channel already exists and the request is *not* incremental, call `update_open` so the late-arriving sender's partition claim still lands in `_partition_coordinator`. Smallest-sender-id-wins election then converges to sender 0 across all storage BEs, all `txn_logs` flow back to a single `OlapTableSink`, and there is exactly one writer of the partition's `combined_txn_log`.

The change is shared-data-only by construction: `LocalTabletsChannel` inherits the no-op base implementation.

A regression test (`LakeTabletsChannelPerPartitionCoordinatorTest.test_late_arriving_sender_0_open_claims_via_update_open`) simulates sender 1 opening first via `channel->open` and sender 0 arriving via `channel->update_open` (what `LoadChannel::open`'s new else branch funnels into), then drives both senders through `add_chunk` + EOS and asserts sender 0 (smallest id) returns all 4 tablet logs while sender 1 returns 0.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Policy changes: every sender's non-incremental open is now applied to the lake per-partition coordinator map (previously only the first opener's)
- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/runtime/lake_tablets_channel.cpp
+++ b/be/src/runtime/lake_tablets_channel.cpp
@@ -82,6 +82,11 @@ public:
     Status incremental_open(const PTabletWriterOpenRequest& params, PTabletWriterOpenResult* result,
                             std::shared_ptr<OlapTableSchemaParam> schema) override;
 
+    // Re-applied non-incremental open: only re-records the per-partition
+    // coordinator claim. See base class comment for the rationale and
+    // `_record_coordinator_claims` for what is recorded.
+    void update_open(const PTabletWriterOpenRequest& params) override { _record_coordinator_claims(params); }
+
     void cancel(const std::string& reason) override;
 
     void abort() override;
--- a/be/src/runtime/load_channel.cpp
+++ b/be/src/runtime/load_channel.cpp
@@ -172,6 +172,19 @@ void LoadChannel::open(const LoadChannelOpenContext& open_context) {
             COUNTER_UPDATE(_index_num, 1);
         } else if (request.is_incremental()) {
             st = it->second->incremental_open(request, response, _schema);
+        } else {
+            // Channel already created by another sender's non-incremental open.
+            // We must still apply this sender's open params so that per-sender
+            // state derived from open (e.g. the lake per-partition coordinator
+            // claim) reflects every sender, not just the first opener.
+            // Without this, `LakeTabletsChannel::_partition_coordinator` only
+            // records the first opener on each storage BE, causing
+            // `_should_enter_collect_path(sender_id)` to return false for the
+            // other senders' EOS handlers — their EOS responses then carry no
+            // txn_logs back to the FE-coord OlapTableSink, and the combined
+            // txn_log file ends up missing the contributions from whichever
+            // NodeChannel was not coordinated by sender 0.
+            it->second->update_open(request);
         }
     }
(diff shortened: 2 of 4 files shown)


CelerData-Reviewer:
@codex review

meegoo (approved):

kevincai (approved):

luohaha:
https://github.com/Mergifyio backport branch-4.1.1

luohaha:
https://github.com/Mergifyio backport branch-4.0.11

PR #73970, opened 2026-05-28
[BugFix] Recreate pipe_file_list when _statistics_ is dropped

meegoo:
## Why I'm doing:

`RepoCreator` (the daemon that owns `_statistics_.pipe_file_list`) caches
`databaseExists` and `tableExists` in `static` fields and only flips them from
`false` to `true`. Once both are `true`, every subsequent `run()` short-circuits
without re-checking the metastore.

That means after any of the following the table is never re-created until the
FE leader restarts:

- `DROP DATABASE _statistics_`
- `DROP TABLE _statistics_.pipe_file_list`
- The table being lost for any other reason (e.g. shared-data storage volume
  failure during a schema change on `column_statistics`)

For users who actively use PIPE this is dangerous: `pipe_file_list` is the
ledger of already-loaded files. If it is missing, active PIPEs may re-ingest
files that were already loaded.

## What I'm doing:

Live-check existence from the metastore on every `run()`, mirroring the
pattern in `TableKeeper`:

- Always call `checkDatabaseExists()` / `checkTableExists()` instead of trusting
  the cached flags.
- Reset `tableExists` when the database is gone.
- Set `tableExists = true` only after either observing the table or completing
  `createTable()` successfully.
- `synchronized run()` and `volatile` flags to keep the accessor reads honest.

The new `checkTableExists()` uses `LocalMetastore.mayGetTable(...)` — the same
helper `TableKeeper` uses.

Added `testRecreateAfterDrop` to cover the drop / restore / recreate cycle.
The existing `testCreator` was updated to mock `checkTableExists` since the
mock-table state now needs to be visible to the daemon between iterations.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/pipe/filelist/RepoCreator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/pipe/filelist/RepoCreator.java
@@ -44,11 +44,12 @@ public class RepoCreator {
                 LOG.warn("database not exists: " + FileListTableRepo.FILE_LIST_DB_NAME);
                 return;
             }
-            if (!checkTableExists()) {
+            tableExists = checkTableExists();
+            if (!tableExists) {
                 createTable();
+                tableExists = true;
                 LOG.info("table created: " + FileListTableRepo.FILE_LIST_TABLE_NAME);
             }
-            tableExists = true;
             correctTable();
         } catch (Exception e) {
             LOG.error("error happens in RepoCreator: ", e);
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

sevev (approved):

HangyuanLiu (approved):

PR #73975, opened 2026-05-28
[DO NOT MERGE] combined fix #73962 + failpoint #73974 for e2e repro

luohaha:
Temporary combined branch used to drive a TSP build that has both the bugfix (https://github.com/StarRocks/starrocks/pull/73962) and the reproduction fail point (https://github.com/StarRocks/starrocks/pull/73974) in one binary. Used only to verify on a real shared-data cluster that the fail point still no-ops the bug after the fix is applied. **This PR is not intended for merge.**


luohaha:
repro work done — closing this scratch PR

PR #73977, opened 2026-05-28
[Enhancement] Support `ANALYZE TABLE` statistics collection for `ENGINE=file` external tables

dirtysalt:
## Why I'm doing:
`ENGINE=file` external tables (backed by HDFS/OSS parquet/orc files) previously returned hardcoded cardinality=1 to the optimizer regardless of actual data size, because `ANALYZE TABLE` silently no-op'd on them. This caused suboptimal query plans — especially join ordering — whenever these tables were involved.

## What I'm doing:

**Collection path** — route `ANALYZE TABLE` on a FileTable through the existing external-table statistics pipeline (`ExternalFullStatisticsCollectJob → external_column_statistics`), which already works for Hive/Iceberg/Paimon:

```
ANALYZE TABLE file_tbl
  → AnalyzeStmtAnalyzer: mark isExternal=true
  → ExternalAnalyzeStatus (catalogName = default_catalog)
  → ExternalFullStatisticsCollectJob
      SELECT COUNT, MIN, MAX, NDV... FROM `default_catalog`.`db`.`file_tbl` WHERE 1=1
      → write rows into external_column_statistics keyed by table.getUUID()
```

**UUID alignment** — the external stats storage/cache infrastructure identifies tables by a `catalog.db.table.id` string (parsed by `StatisticsUtils.getTableByUUID`). FileTable previously inherited the base `Table.getUUID()` that returns a bare numeric id, which the parser would reject. `FileTable` now overrides `getUUID()` to emit the same 4-part format; legacy serialised instances without `dbName` fall back to the numeric id transparently.

**Partition handling** — FileTable has no partition metadata and is absent from `ConnectorPartitionTraits`.  StatisticsCollectJobFactory` now treats it as a single unpartitioned segment (matching `DefaultTraits` behaviour for unpartitioned tables), so `WHERE 1=1` is used and no partition-key predicates are generated.

**Optimizer read path** — `StatisticsCalculator.computeFileTableScanNode` previously returned hardcoded unknowns. It now calls `getConnectorTableStatisticsSync` (synchronous cache load) so the optimizer sees real cardinality on the first `EXPLAIN` after `ANALYZE`, without needing a cache warm-up round.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/FileTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/FileTable.java
@@ -64,15 +64,33 @@ public class FileTable extends Table {
     @SerializedName(value = "fp")
     private Map<String, String> fileProperties = Maps.newHashMap();
 
+    @SerializedName(value = "dn")
+    private String dbName;
+
     public FileTable() {
         super(TableType.FILE);
     }
 
     public FileTable(long id, String name, List<Column> fullSchema, Map<String, String> properties)
             throws DdlException {
+        this(id, name, null, fullSchema, properties);
+    }
+
+    public FileTable(long id, String name, String dbName, List<Column> fullSchema, Map<String, String> properties)
+            throws DdlException {
         super(id, name, TableType.FILE, fullSchema);
+        this.dbName = dbName;
         this.fileProperties = properties;
         validate(properties);
+    }
+
+    @Override
+    public String getUUID() {
+        // Null-safe: old serialized instances that pre-date this field fall back to the legacy numeric ID.
+        if (dbName == null) {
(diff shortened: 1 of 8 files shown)


CelerData-Reviewer:
@codex review

mark-phoenix-ai:
External table is deprecated. We shouldn't optimize more.

PR #73998, opened 2026-05-28
[Enhancement] Allow data skew detection when table row count may be inaccurate

martinr0x:
## Why I'm doing:
We often see skew optimization rules not being applied because `DataSkew.getSkewCandidates` does not report skew. This is because statistics for the table sometimes report `isTableRowCountMayInaccurate()=true`, even though ColumnStatistics and MCVs are skewed and correct. 

There are multiple reasons why `isTableRowCountMayInaccurate()=true`, but one reason is due to cache eviction in tableStatsCache.
TableStatsCache is a fixed size cache in `CachedStatisticStorage`. Evicted entries trigger async reloads that may return empty results to the caller (mostly due to `enable_sync_statistics_load=false`). Due to this`getPartitionRows` may return a rowCount of 0 and we end up with a tableRowCount=0 in `computeOlapScanNode`, resulting in `isTableRowCountMayInaccurate = true`.

## What I'm doing:
Adds an session variable `enable_skew_detect_with_inaccurate_stats` that bypasses the `isTableRowCountMayInaccurate` guard in DataSkew. When enabled, skew detection proceeds based on available MCV/histogram data even when the row count is flagged as potentially inaccurate.

The check is implemented inside DataSkew itself, so all consumers (skew join elimination, aggregation skew handling, window function optimization) benefit without any call-site changes.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
@@ -562,6 +562,7 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
     public static final String SKEW_JOIN_MAX_OTHER_SIDE_OVERLAP_ROW_COUNT = "skew_join_max_other_side_overlap_row_count";
     public static final String SKEW_JOIN_MCV_SINGLE_THRESHOLD = "skew_join_mcv_single_threshold";
     public static final String SKEW_JOIN_MCV_MIN_INPUT_ROWS = "skew_join_mcv_min_input_rows";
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

trueeyu (approved):

kevincai (approved):

martinr0x:

PR #74005, opened 2026-05-28
[BugFix] Fix lake PCU crash and silent corruption under schema drift

zhangheihei:
## Why I'm doing:
```
*** Aborted at 1779975211 (unix time) try "date -d @1779975211" if you are using GNU date ***
PC: @          0x802f047 starrocks::ScalarColumnWriter::append_array_offsets(starrocks::Column const&)
*** SIGSEGV (@0x0) received by PID 975880 (TID 0x7f482625f700) LWP(976454) from PID 0; stack trace: ***
    @     0x7f4900f8120b __pthread_once_slow
    @          0xdd8c9d4 google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @     0x7f4900766cfd PosixSignals::chained_handler(int, siginfo*, void*) [clone .part.10]
    @     0x7f490076778d JVM_handle_linux_signal
    @     0x7f4900f8a630 (/usr/lib64/libpthread-2.17.so+0xf62f)
    @          0x802f047 starrocks::ScalarColumnWriter::append_array_offsets(starrocks::Column const&)
    @          0x8233cec starrocks::ArrayColumnWriter::append(starrocks::Column const&)
    @          0x7595afa starrocks::SegmentWriter::append_chunk(starrocks::Chunk const&)
    @          0x7aa0720 starrocks::lake::ColumnModePartialUpdateHandler::execute(starrocks::lake::RowsetUpdateStateParams const&, starrocks::lake::MetaFileBuilder*, std::vector<std::vector<unsigned int, std::allocator<unsigned int> >, std::allocator<std::vector<unsigned int, std:@
    @          0xa0d042c starrocks::ThreadPool::dispatch_thread()
    @          0xa0c76a1 starrocks::Thread::supervise_thread(void*)
    @     0x7f4900f82ea5 start_thread
    @     0x7f48fe880b0d __clone
```

## What I'm doing:

Fixes #issue
Column-mode partial update (PCU) on shared-data tables records the column id
(cid) of each target column in the txn log at write time. When a concurrent
ALTER TABLE ADD/DROP COLUMN commits between write and publish, the tablet
metadata schema is updated with new column ordering, causing those frozen cids
to no longer match the current schema layout.

This bug is reachable only when cloud_native_fast_schema_evolution_v2 = false
(the default for existing lake tables). In this mode, ALTER dispatches
LakeTableAsyncFastSchemaChangeJob that directly rewrites the tablet metadata
schema via op_alter_metadata, independently of any data transaction. A PCU
whose write completed before the ALTER but publishes after it will then see a
schema where cids have shifted.

When FSE v2=true, this bug cannot occur: ALTER only updates the FE catalog
without touching tablet metadata, and schema upgrades are piggybacked on data
transactions themselves — guaranteeing that each transaction's publish always
sees the same schema version it wrote with.

The stale cid caused three distinct symptoms:

- Crash: an ARRAY target column whose cid shifted to a non-array column
  would segfault inside ArrayColumnWriter::append (down_cast on the wrong
  column type).
- Silent data corruption: a fixed-length target column whose cid shifted
  would write its values into an unrelated column that happened to occupy
  the old cid position, with no error surfaced.
- Conditional update (merge_condition) with PCU compared against the wrong
  column once the condition column's cid drifted.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/storage/lake/column_mode_partial_update_handler.cpp
+++ b/be/src/storage/lake/column_mode_partial_update_handler.cpp
@@ -344,29 +344,33 @@ Status ColumnModePartialUpdateHandler::execute(const RowsetUpdateStateParams& pa
 
     const auto& txn_meta = params.op_write.txn_meta();
 
+    // cid may shift across schema versions; recompute it from uid against the current
+    // tablet schema. partial_update_column_ids in txn_meta is kept for compatibility only.
+    DCHECK_EQ(txn_meta.partial_update_column_ids_size(), txn_meta.partial_update_column_unique_ids_size());
     std::vector<ColumnId> update_column_ids;
     std::vector<ColumnUID> unique_update_column_ids;
-    for (ColumnId cid : txn_meta.partial_update_column_ids()) {
-        if (cid >= params.tablet_schema->num_key_columns()) {
-            if (!params.tablet_schema->column(cid).is_auto_increment()) {
-                update_column_ids.push_back(cid);
-            }
-        }
-    }
-    for (uint32_t uid : txn_meta.partial_update_column_unique_ids()) {
-        auto cid = params.tablet_schema->field_index(uid);
+    for (int i = 0; i < txn_meta.partial_update_column_unique_ids_size(); ++i) {
+        const uint32_t uid = txn_meta.partial_update_column_unique_ids(i);
+        const auto cid = params.tablet_schema->field_index(uid);
         if (cid == -1) {
             std::string msg = strings::Substitute("column with unique id:$0 does not exist. tablet:$1", uid,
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

kevincai (approved):

PR #74030, opened 2026-05-29
[BugFix] Encode __ROW_ID__ for GROUP BY-only incremental MV

Youngwb:
## Why I'm doing:

An incremental materialized view whose query has GROUP BY but no aggregate — e.g. `SELECT k FROM t GROUP BY k` (a DISTINCT-on-keys MV) — is accepted at CREATE but crashes on the second (incremental) refresh:

```
Type check failed. hash join equal predicate type mismatch:
left VARCHAR, right BIGINT, predicate: __ROW_ID__ = from_binary
```

Root cause: `IVMAnalyzer.checkAggregate` decided the `__ROW_ID__` strategy from *"are there aggregates"*, while `IvmDeltaAggregateRule` (the refresh-side rewrite) keys on *"is there GROUP BY"*. The two disagree exactly for GROUP BY without aggregates:

- CREATE: `checkAggregate` sees an empty aggregate list, returns early, and the MV is built with an `AUTO_INCREMENT` `BIGINT __ROW_ID__`.
- REFRESH: `IvmDeltaAggregateRule` fires on the GROUP BY and encodes the group keys as `FROM_BINARY(ENCODE_ROW_ID(...))` → `VARCHAR`, then joins delta against the MV on `__ROW_ID__`.

`BIGINT = VARCHAR` fails the planner's type check (`TypeChecker` during `InsertPlanner.buildExecPlan`), so refresh never completes.

## What I'm doing:

Fixes #issue: N/A — no public tracking issue.

Align the analyzer's row-id decision to GROUP BY, matching the refresh path:

- Any query with GROUP BY encodes its group keys into `__ROW_ID__` (`QUERY_COMPUTED`), regardless of whether it has aggregates.
- Only a query with no GROUP BY stays `AUTO_INCREMENT` (append-only projection).
- The aggregate-rewrite loop is skipped when there are no aggregates (`getAggregate()` is non-null post-analysis, so it just no-ops), producing a DISTINCT-on-keys MV whose refresh plan (`IvmDeltaAggregateRule` with an empty aggregation map) already degrades to a dedup-by-row-id join.

Aggregate-with-GROUP-BY and no-GROUP-BY projection paths are unchanged.

### Tests

- **`IVMBasedMvRefreshProcessorIcebergTest.testIVMWithGroupByNoAggregate`** — end-to-end: generates the incremental refresh plan for `SELECT id FROM t0 GROUP BY id` across 3 runs. On the unpatched analyzer this reproduces the exact crash (`TypeChecker` rejects `__ROW_ID__ = from_binary`, BIGINT vs VARCHAR, in `InsertPlanner.buildExecPlan`); with the fix the plan joins on the encoded VARCHAR `__ROW_ID__` and refresh succeeds.
- **`IVMAnalyzerTest.testGroupByWithoutAggregateYieldsQueryComputedStrategy`** — analyzer-level: `SELECT id FROM t0 GROUP BY id` yields `QUERY_COMPUTED` (was `AUTO_INCREMENT` before the fix).

Both verified RED on the unpatched code (the refresh test throws the TypeChecker exception; the analyzer test asserts `AUTO_INCREMENT`) and GREEN after. Full `IVMAnalyzerTest` (26) and `IVMBasedMvRefreshProcessorIcebergTest` (51) green; `mvn checkstyle:check` clean.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [x] Miscellaneous: a GROUP BY-only incremental MV that previously crashed at refresh now works; such MVs are created with a VARCHAR `__ROW_ID__` instead of BIGINT.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/mv/IVMAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/mv/IVMAnalyzer.java
@@ -313,16 +313,19 @@ public class IVMAnalyzer {
 
     private boolean checkAggregate(SelectRelation selectRelation) throws AnalysisException {
         List<FunctionCallExpr> aggregateExprs = selectRelation.getAggregate();
-        if (CollectionUtils.isEmpty(aggregateExprs)) {
+        List<Expr> groupByExprs = selectRelation.getGroupBy();
+
+        // Gate on GROUP BY, not aggregates: the refresh path (IvmDeltaAggregateRule) keys on
+        // GROUP BY, so a GROUP BY-only query must also get QUERY_COMPUTED row ids here — else the
+        // AUTO_INCREMENT MV it would otherwise build crashes refresh on a row-id type mismatch.
+        if (CollectionUtils.isEmpty(groupByExprs)) {
+            if (CollectionUtils.isNotEmpty(aggregateExprs)) {
+                throw new SemanticException("IVMAnalyzer requires group by expressions for incremental view maintenance.");
+            }
             return false;
         }
 
-        List<Expr> groupByExprs = selectRelation.getGroupBy();
-        if (CollectionUtils.isEmpty(groupByExprs)) {
-            // If there are no group by expressions, we cannot apply IVM optimizations.
-            throw new SemanticException("IVMAnalyzer requires group by expressions for incremental view maintenance.");
-        }
-        // new aggregate functions
+        // getAggregate() is non-null post-analysis, so for no aggregates this loop just no-ops.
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

stephen-shelby (approved):

Astralidea (approved):

PR #11315, opened 2022-09-16
[Enhancement] chore: fixing redundant AVX2 macro in runtime filter

mapleFU:
## What type of PR is this：

- [x] Tool

## Which issues of this PR fixes ：

Fixes #

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] I have added user document for my new feature or new function

PR #74036, opened 2026-05-29
[BugFix] Fix combined ALTER TABLE on external Iceberg re-executing queued actions

GavinMar:
## Why I'm doing:

A single `ALTER TABLE` on an external Iceberg table that carries more than one schema clause (e.g. `ADD COLUMN a, ADD COLUMN b, DROP COLUMN c`) fails with a misleading error such as:

```
ERROR 1064 (HY000): Cannot add column, name already exists: multi_col_a
```

Root cause is in `ConnectorAlterTableExecutor`. Each `visitXxxClause` only queues a deferred action into the `actions` list, but the overridden `visit(ParseNode, ConnectContext)` re-ran the **entire** `actions` list on every clause dispatch without ever clearing it. So for N clauses, action #1 runs N times, action #2 runs N-1 times, etc. The second time an `addColumn` action runs against the same Iceberg transaction, Iceberg's `SchemaUpdate` rejects it with `name already exists`.

This affects external **Iceberg** catalogs only (the sole subclass of `ConnectorAlterTableExecutor`). Internal/OLAP tables use a separate `AlterJobExecutor` path and are unaffected. The defect has existed since the class was introduced, so all maintained branches are impacted.

Reported in StarRocksTest issue #11315.

## What I'm doing:

- Collect the deferred actions while visiting all clauses, then run each queued action exactly once after the loop completes, inside `applyClauses()`.
- Remove the redundant `visit(ParseNode, ConnectContext)` override and fall back to the default `AstVisitor` dispatch.
- Add a regression test `AlterTableTest#testCombinedAddAndDropColumn` covering a combined add/drop-column ALTER against an Iceberg table. The existing `IcebergMetadataTest.testAlterTable` could not catch this because it uses a fully mocked catalog that swallows the duplicate `addColumn` calls.

Verified end-to-end on a TSP cluster: the same combined ALTER fails on the baseline build and succeeds after applying this fix.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorAlterTableExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorAlterTableExecutor.java
@@ -45,6 +45,11 @@ public class ConnectorAlterTableExecutor implements AstVisitorExtendInterface<Vo
             for (AlterClause c : alterClauses) {
                 visit(c, null);
             }
+            // Run queued actions only after every clause is visited; running inside the
+            // loop above would re-execute every previously queued action on each pass.
+            for (Runnable r : actions) {
+                r.run();
+            }
         } catch (StarRocksConnectorException e) {
             throw new DdlException(e.getMessage(), e.getCause());
         }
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

dirtysalt (approved):
LGTM! 🚀

Youngwb (approved):

PR #74037, opened 2026-05-29
[BugFix] Pass tablet metadata when applying PK index compaction output sstables

xiangguangyxg:
## Why I'm doing:

In shared-data (cloud-native) mode, publishing a primary-key table can fail permanently with `Invalid argument: metadata is null when loading delvec from file`. `LakePersistentIndex::apply_opcompaction` reopened a PK-index compaction **output** sstable via `PersistentIndexSstable::new_sstable(sstable_pb, location, cache)` without passing the tablet metadata. When the parallel-compaction "move/passthrough" path carries an input sstable's embedded delete vector (delvec) onto the output sstable, `init()` cannot load that delvec without the metadata and returns the error above, which wedges publish (observed retrying indefinitely).

## What I'm doing:

Thread the tablet metadata (`TabletMetadataPtr`) through `apply_opcompaction` and the `publish_primary_compaction` / `light_publish_primary_compaction` / `early_sst_compact` paths, and pass it down to `new_sstable` so the embedded delvec loads — mirroring what `LakePersistentIndex::init()` and the earlier fix in `sample_keys_from_sstable` (#69846) already do.

The metadata is passed by `const TabletMetadataPtr&` (by reference) throughout, so no `shared_ptr` is copied along the chain; the only by-value capture is the necessary one in the async `early_sst_compact` callback (capturing by reference there would dangle).

Added a regression test (`LakePersistentIndexTest.test_apply_opcompaction_output_sstable_with_delvec`) that reproduces the failure (an op_compaction output sstable carrying a delvec) and verifies the fix.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -675,7 +675,9 @@ StatusOr<AsyncCompactCBPtr> LakePersistentIndex::early_sst_compact(
     ASSIGN_OR_RETURN(auto cb,
                      compact_mgr->async_compact(
                              result.candidate_filesets, metadata, is_merge_base_level,
-                             [&, result](const std::vector<PersistentIndexSstablePB>& sstables) {
+                             // Capture `result` and `metadata` by value: this callback runs
+                             // asynchronously on the compaction thread pool.
+                             [&, result, metadata](const std::vector<PersistentIndexSstablePB>& sstables) {
                                  // 4. Merge output sstables into current index.
                                  //    reuse `apply_opcompaction` to do this.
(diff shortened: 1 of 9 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

kevincai (approved):

PR #72789, opened 2026-05-06
[BugFix] Make Locker rollback exception-safe and fix unlock order

kevincai:
Several correctness fixes in the FE lock-manager wrappers on Locker:

1. lockTablesWithIntensiveDbLock and tryLockTablesWithIntensiveDbLock only caught LockException, so any unchecked exception thrown after the DB intention lock had been acquired would leak that lock (and any partially-acquired table locks), stalling later DB-WRITE operations. Both methods now use a try / finally pattern that routes every failure through rollbackPartialIntensiveLock.

2. unLockTablesWithIntensiveDbLock released the DB intention lock BEFORE releasing the table locks, violating hierarchical-locking ordering. The window allowed a concurrent X-on-DB request (e.g. DROP DATABASE) to acquire the DB and drop metadata while the caller still held a table lock against it. Release order is now tables first, then DB intention, matching the acquire order in reverse.

3. unLockTableWithIntensiveDbLock(LockParams, LockType) was allocating a fresh Locker to perform the release. It worked only because Locker identity collapses by threadId; cross-thread or refactored callers would silently leak locks. It now uses this.

4. LockManager.notifyVictim used timeRemain < 0 for its timeout check while the other two timeout sites in lock() / lockAcquireSlowPath() use <= 0. At timeRemain == 0 the inconsistent boundary caused one extra notify-and-sleep round before timing out. Aligned to <= 0.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
[...]

PR #74041, opened 2026-05-29
[BugFix] PlannerMetaLocker: skip unlock when lock() never succeeded (POST-1561)

meegoo:
Companion to the Locker-level rollback that already landed upstream (#72423, #72789). Those PRs ensure that
`Locker.lock(Table|Tables)WithIntensiveDbLock` releases anything it partially acquired before re-throwing. This commit closes the remaining symptom at the PlannerMetaLocker layer.

Pre-fix behavior (the SmartNews 4.0.9 incident):

  caller: try { plannerMetaLocker.lock(); ... } finally { unlock(); }

  1. lock() iterates (db, tables) entries calling inner Locker.lockTablesWithIntensiveDbLock on each.
  2. The inner call throws on some entry (e.g. ERR_NO_FILES_FOUND in the FILES() planning path forcing the planner to bail, or a real LockException). Inner Locker rolls itself back — good.
  3. PlannerMetaLocker.lock() does NOT roll back any earlier (db, tables) entries that already succeeded; it just lets the exception propagate.
  4. The caller's finally calls unlock(), which blindly iterates the full `tables` map and releases entries that were never acquired.
  5. LockManager.release sees a Locker with no matching LockHolder and throws IllegalMonitorStateException: "Attempt to unlock lock, not locked by current locker" — surfaced to the client as ERROR 5600, masquerading as the original ERR_NO_FILES_FOUND.

Fix:

* `PlannerMetaLocker` now records `heldEntries` on successful lock()/tryLock(). unlock() only releases entries in `heldEntries`, and is a no-op when null.
* lock() wraps its multi-db loop in try/catch: on failure it releases the earlier (db, tables) entries it had successfully acquired (in reverse order, best-effort with logging), clears `heldEntries`, and re-throws. Inner-layer rollback (already in upstream Locker) handles the failing entry itself; this outer rollback only addresses entries that fully succeeded before the failing one.
* AutoCloseable.close() → unlock() is now safe to call when lock() was never invoked or threw.

Tests:

* `PlannerMetaLockerRollbackTest` (new): unlock-without-lock, close-without-lock, normal lock/unlock round-trip, single-db rollback (inner fails), multi-db rollback (first db succeeds, second throws — asserts the first db's release happened).
* `LockerIntensiveLockRollbackTest` (new): regression test pinning the contract that the Locker-layer fix from #72423/#72789 provides — six cases (mid-loop failure, first-rid failure, db-intent failure, single-table variant, WRITE mode, success path), each verified by trying to acquire an exclusive WRITE on the rid afterward from a fresh Locker (any leaked IS/READ would time out the attempt).

Related: POST-1561.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/PlannerMetaLocker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/PlannerMetaLocker.java
@@ -74,14 +74,33 @@ public class PlannerMetaLocker implements AutoCloseable {
     private Map<Long, Set<Long>> tables = Maps.newTreeMap(Long::compareTo);
 
     /**
-     * Entries successfully locked by the most recent {@link #lock()} or {@link #tryLock} call.
-     * Null when no acquisition is currently held (initial state, after a failed {@code lock()}
-     * that rolled itself back, or after {@link #unlock()}). {@code unlock()} only releases entries
-     * recorded here, so it never tries to release rids that were never locked by this instance
-     * — which would otherwise surface as {@code IllegalMonitorStateException: "Attempt to unlock
-     * lock, not locked by current locker"}. See POST-1561 / PR following #69284.
+     * Entries successfully locked by the outermost {@link #lock()} / {@link #tryLock} that
+     * is currently active on this instance. {@code unlock()} uses this snapshot to release
+     * the right rids, regardless of how deeply nested the lock/unlock calls are.
+     *
+     * <p>Null when {@link #lockDepth} is 0 (no acquisition currently in scope, or already
+     * fully released). See {@code lockDepth} for the depth semantics.
      */
     private List<Map.Entry<Long, Set<Long>>> heldEntries;
+
+    /**
+     * Number of currently-pending {@code lock()}/{@code tryLock()} acquisitions on this
+     * instance — i.e., how many matching {@code unlock()} calls are still owed.
+     *
+     * <p>Nested acquisitions happen for real on the optimistic-retry path:
+     * {@link com.starrocks.sql.InsertPlanner#buildExecPlanWithRetry} unlocks during planning
+     * then re-locks before validation, and {@link com.starrocks.sql.StatementPlanner#reAnalyzeStmt}
+     * additionally calls {@code lock()}/{@code unlock()} on the SAME instance inside the retry
+     * iteration. The inner pair must not release the outer acquisition's LockManager refCount,
+     * and the inner {@code unlock()} must not clear {@link #heldEntries} (otherwise the outer
+     * {@code unlock()} would become a no-op and leak the locks).
+     *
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

meegoo:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/PlannerMetaLocker.java
```diff
@@ -69,6 +73,35 @@ public class PlannerMetaLocker implements AutoCloseable {
...
+     * iteration. The inner pair must not release the outer acquisition's LockManager refCount,
+     * and the inner {@code unlock()} must not clear {@link #heldEntries} (otherwise the outer
+     * {@code unlock()} would become a no-op and leak the locks).
+     *
+     * <p>{@code lock()} increments after a successful acquire; {@code unlock()} decrements
+     * unconditionally when in-scope and clears {@code heldEntries} once it hits zero. A failed
+     * {@code lock()} (partial-acquire that rolled itself back) does NOT increment, so the
+     * outer state — if any — is preserved untouched.
+     */
+    private int lockDepth = 0;
```
kevincai: it is tricky to use `lockDepth` to handle the lock reentrance issue. Think about some alternatives. e.g., record the <db-id, table-ids> in a list. and unlock the list in reverse order. As long as the underlying lock allows reentrant lock, the symmetric lock/unlock should be able to clean the lock clearly.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/PlannerMetaLocker.java
```diff
@@ -91,8 +117,11 @@ public boolean tryLock(long timeout, TimeUnit unit) {
                     return false;
                 }
                 lockedDbs.add(database);
+                lockedEntries.add(entry);
             }
             isLockSuccess = true;
+            // Record this acquisition so the matching unlock() releases exactly what it locked.
+            heldStack.push(lockedEntries);
```
kevincai, line 121: The `lockedDbs` is not necessary any more, giving that the lockedEntries records everything needed in the `finally` statement. And the rollback lock in the finally shall be exception safe, avoid leaking lockedEntries if previous dbEntries unlock excepted.

kevincai (approved):

HangyuanLiu (approved):

meegoo:
@mergify backport branch-4.1

PR #74059, opened 2026-05-30
[BugFix] Sum composite-rowset stats when batching op_writes in PK multi-statement

xiangguangyxg:
## Why I'm doing:

`MetaFileBuilder::add_rowset` accumulates the op_writes of a multi-statement / batch transaction into a single composite rowset. It merges their segments (and segment sizes, metas, encryption metas, shared flags, bundle offsets) but **never sums `num_rows`, `data_size`, or `num_dels`** — the first op_write's values (from the initial `CopyFrom`) survive. So the composite rowset reports only the first statement's counts while physically holding every statement's segments (e.g. `num_rows=1` for a rowset whose segments contain `1 + 10000 = 10001` rows).

On a **PRIMARY KEY** table this corrupts the read path — rows in the under-counted segments become unreadable and range scans miscount — until a later compaction rewrites the rowset and recomputes `num_rows`. It is reached by any PK multi-statement / batch transaction (`PrimaryKeyTxnLogApplier::apply(TxnLogVector)`); a write cross-published to split children during a tablet reshard is one trigger. The non-PK batch path already sums correctly, so only PK tables are affected.

Surfaced while end-to-end testing concurrent multi-statement writes during a tablet split (reshard work, #74013), but the bug is **pre-existing and broader than reshard**. Confirmed by decoding the on-disk `TabletMetadataPB`: a composite rowset with `num_rows=1` over two segments whose `segment_metas` hold `1` and `10000` rows.

## What I'm doing:

In the accumulation branch of `add_rowset`, sum `num_rows`/`data_size`/`num_dels` into the pending composite rowset, and mark it `overlapped` (a composite spanning multiple op_writes may carry overlapping keys — matching the non-PK batch path's `set_overlapped(all_segments.size() > 1)`).

Adds `MetaFileTest.test_add_rowset_sums_composite_stats`, which builds a composite rowset from a 1-row + 10000-row op_write and asserts the summed `num_rows`/`data_size` (it fails on the pre-fix first-op-write-only behavior).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [ ] Miscellaneous: data-correctness fix (composite-rowset row/size counts on PK tables)

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/meta_file.cpp
+++ b/be/src/storage/lake/meta_file.cpp
@@ -971,6 +971,19 @@ void MetaFileBuilder::add_rowset(const RowsetMetadataPB& rowset_pb, const std::m
                     _pending_rowset_data.assigned_segment_idx + get_segment_idx(rowset_pb, i));
         }
     } else {
+        // The first op_write was CopyFrom'd above (carrying its num_rows/data_size/num_dels);
+        // subsequent op_writes only append their segments below, so their row/size/del counts
+        // must be SUMMED into the composite rowset. Otherwise the rowset keeps just the first
+        // op_write's counts while holding every op_write's segments — a multi-statement / batch
+        // (incl. cross-publish) PK txn would then report e.g. num_rows=1 for a 10001-row rowset,
+        // corrupting reads until compaction rewrites the rowset. A composite spanning multiple
+        // op_writes can also carry overlapping keys, so mark it overlapped (as the non-PK batch
+        // path does via set_overlapped(all_segments.size() > 1)).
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #74068, opened 2026-05-31
[UT] Stabilize and speed up CreateMaterializedViewTest

kevincai:
Combines two changes against the same test file:

1. Fix the flaky hang caused by @Mocked TaskManager racing TaskCleaner.

The three testCreateAsync_* tests used @Mocked TaskManager, which makes
JMockit replace every public method body with a class-wide stub for the
duration of the test. TaskManager.taskUnlock() is public;
tryTaskLock()/takeTaskLock() are private and therefore left untouched.
If the TaskCleaner daemon was inside removeExpiredTasks (the real
tryLock had already incremented the hold count) when the test method
started, the finally's taskUnlock() landed in the JMockit stub instead
of taskLock.unlock(). The hold count never returned to zero,
TaskCleaner became the perpetual fair-lock owner, and every later test
that dropped a real-task MV wedged in @AfterEach ->
dropMaterializedView -> TaskManager.getTask -> takeTaskLock for the
rest of the JVM.

Replace with a narrow MockUp<TaskManager> that only stubs the public
executeTask(String) overload these tests actually verify;
taskUnlock() and the other TaskManager methods keep their real bodies
and the strike window disappears. Reproduced twice (2026-05-13,
2026-05-29) with identical 2:06 elapsed-time fingerprint. See
trouble-shoot/fe-ut-timeout/findings-v2.html for the full diagnosis.

2. Replace two unconditional Thread.sleep(4000L) calls with Awaitility
polls and harden waitingTaskFinish().

The sleeps trace back to PR #9233 (Jul 2022) where one test needed a
periodic-refresh task to fire at least once; later refactors folded
the sleep into two helpers now used by ~24 callers, almost none of
which need to wait at all.

  getMaterializedViewKeysChecked()  -- 20 callers, schema-only
    poll: await up to 5s, interval 200ms, until MV is visible as
    a MaterializedView in LocalMetastore

  getMaterializedViewChecked()      -- 3 callers
    poll: same MV-visible condition, 5s / 200ms

The MV-visible poll is near-zero cost in the common case because
createMaterializedView is synchronous and the MV is already in
nameToTable by the time the helper looks.
Removing the second sleep also exposed a pre-existing race in
waitingTaskFinish(): it captured getMatchedTaskRunStatus(null) once
at the start and then re-checked taskRuns.get(0), so the loop could
return immediately by observing a stale, already-finished history
entry that belonged to some unrelated task. Harden the helper:

  - take an mvTaskName parameter and pass it to the task-name filter
    in TGetTasksParams, so only this MV's task runs are considered
  - poll (Awaitility, atMost 15s, interval 500ms) until at least one
    matching run exists AND all matching runs are in a finish state,
    so any pending/running submission drains before we return
  - return a freshly re-fetched list (not the stale snapshot)

Both call sites updated to pass the task name.

Result on local ./run-fe-ut.sh --test CreateMaterializedViewTest:
  Surefire time elapsed: 112.0 s -> 28.0 s (-75%)
  Wall clock:            3:00   -> 1:08    (-62%)
  Pass/fail:             201/0  -> 201/0   (unchanged)
  4s-plateau tests:      24     -> 0       (no fixed sleeps remain)

## Why I'm doing:

Introduced in #31258

```
2026-05-29T10:28:25.3058656Z 2026-05-29 18:28:24 [main] WARN  TaskManager:860 - task lock is held by: {"id":78,"name":"TaskCleaner","stack":["java.base@17.0.8/java.lang.Thread.sleep(Native Method)","app//com.starrocks.common.util.Daemon.run(Daemon.java:104)"]}
2026-05-29T10:28:25.3060607Z 2026-05-29 18:28:24 [main] WARN  TaskManager:879 - fail to get TaskManager lock after retry 1 times
2026-05-29T10:28:32.2939484Z 2026-05-29 18:28:29 [main] WARN  TaskManager:860 - task lock is held by: {"id":78,"name":"TaskCleaner","stack":["java.base@17.0.8/java.lang.Thread.sleep(Native Method)","app//com.starrocks.common.util.Daemon.run(Daemon.java:104)"]}
2026-05-29T10:28:32.2940748Z 2026-05-29 18:28:29 [main] WARN  TaskManager:879 - fail to get TaskManager lock after retry 2 times
2026-05-29T10:28:35.3007306Z 2026-05-29 18:28:34 [main] WARN  TaskManager:860 - task lock is held by: {"id":78,"name":"TaskCleaner","stack":["java.base@17.0.8/java.lang.Thread.sleep(Native Method)","app//com.starrocks.common.util.Daemon.run(Daemon.java:104)"]}

...

2026-05-29T11:40:18.3396721Z 2026-05-29 19:40:10 [main] WARN  TaskManager:879 - fail to get TaskManager lock after retry 862 times
2026-05-29T11:40:18.3397705Z 2026-05-29 19:40:15 [main] WARN  TaskManager:860 - task lock is held by: {"id":78,"name":"TaskCleaner","stack":["java.base@17.0.8/java.lang.Thread.sleep(Native Method)","app//com.starrocks.common.util.Daemon.run(Daemon.java:104)"]}
2026-05-29T11:40:18.3398631Z 2026-05-29 19:40:15 [main] WARN  TaskManager:879 - fail to get TaskManager lock after retry 863 times
```

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/test/java/com/starrocks/analysis/CreateMaterializedViewTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/analysis/CreateMaterializedViewTest.java
@@ -47,7 +47,6 @@ import com.starrocks.common.MaterializedViewExceptions;
 import com.starrocks.common.tvr.TvrTableSnapshot;
 import com.starrocks.common.util.DateUtils;
 import com.starrocks.common.util.PropertyAnalyzer;
-import com.starrocks.common.util.ThreadUtil;
 import com.starrocks.persist.CreateTableInfo;
 import com.starrocks.persist.OperationType;
 import com.starrocks.persist.metablock.SRMetaBlockReaderV2;
@@ -56,6 +55,7 @@ import com.starrocks.qe.ConnectContext;
 import com.starrocks.qe.DDLStmtExecutor;
 import com.starrocks.qe.StmtExecutor;
 import com.starrocks.scheduler.Constants;
+import com.starrocks.scheduler.SubmitResult;
 import com.starrocks.scheduler.TaskBuilder;
 import com.starrocks.scheduler.TaskManager;
 import com.starrocks.scheduler.persist.TaskRunStatus;
@@ -96,16 +96,16 @@ import com.starrocks.sql.plan.ExecPlan;
 import com.starrocks.statistic.StatisticsMetaManager;
 import com.starrocks.system.Backend;
 import com.starrocks.system.SystemInfoService;
+import com.starrocks.thrift.TGetTasksParams;
 import com.starrocks.type.ScalarType;
 import com.starrocks.type.Type;
 import com.starrocks.utframe.StarRocksAssert;
 import com.starrocks.utframe.UtFrameUtils;
-import mockit.Expectations;
 import mockit.Mock;
 import mockit.MockUp;
-import mockit.Mocked;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.awaitility.Awaitility;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #74087, opened 2026-05-31
[BugFix] Backfill uncollected external table columns on partial ANALYZE

eshishki:
## Why I'm doing:

`MetadataMgr#getTableStatistics` merges internal (collected via `ANALYZE`)
statistics with connector/manifest statistics for external tables. That
merge is all-or-nothing over the set of requested columns:

- **No stats collected** → every column is unknown, the method falls back
  to connector/manifest stats, so columns inherit manifest min/max. Fine.
- **Stats collected for *some* columns** (a partial `ANALYZE`) → the method
  returns the internal stats as-is. Collected columns are fine, but the
  *uncollected* columns stay `unknown` and the connector/manifest fallback
  is skipped for the whole table.

So a partial `ANALYZE` leaves the uncollected columns with no min/max at
all — strictly worse than collecting nothing, where those same columns
would have inherited manifest min/max. The fallback only kicks in when
*every* column is unknown.

## What I'm doing:

**Commit 1 (BugFix):** when internal stats cover only part of the requested
columns, fetch connector metadata and backfill the `unknown` columns from
it. The row count is also taken from the connector (the queried snapshot)
so it stays consistent with the backfilled columns; collected columns and
any histogram already attached to a backfilled column are preserved. Adds a
unit test for the merge helper.

**Commit 2 (Refactor):** the fix exposed three near-duplicate column-merge
branches in `getTableStatistics` (histogram-over-connector combine, the new
partial backfill, and the histogram attach in
`getTableStatisticsFromInternalStatistics`). Collapse the method into one
flat path — fast-return when every column is collected, otherwise fetch
connector metadata once and reuse the backfill helper for both the
all-unknown and partially-collected cases. Behavior across the previous
fallback / single-table / blacklist / null-connector branches is preserved.

The two commits are kept separate on purpose: the BugFix is self-contained
and cherry-pick friendly for backports, the Refactor is main-only. Happy to
split into two PRs if maintainers prefer.

**User-visible behavior change:** for external tables with a *partial*
`ANALYZE` (statistics collected for only some of the queried columns), the
uncollected columns now contribute connector/manifest min/max to the
optimizer instead of being treated as unknown, and the table row count for
that estimate now comes from connector metadata rather than from the
collected stats. Consequently, cardinality estimates — and therefore query
plans (join order, runtime filter generation, broadcast vs. shuffle, …) —
can change for such tables. Estimates are generally more accurate; tables
that were fully analyzed, or had no statistics at all, are unaffected.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/server/MetadataMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/MetadataMgr.java
@@ -25,7 +25,6 @@ import com.google.common.cache.RemovalCause;
 import com.google.common.cache.RemovalListener;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
 import com.starrocks.catalog.BaseTableInfo;
 import com.starrocks.catalog.BasicTable;
 import com.starrocks.catalog.Column;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby:
@Youngwb @dirtysalt what do you think of this strategy?

PR #74108, opened 2026-06-01
[BugFix] Add decorrelated jitter to lake vacuum retry backoff

starrocks-xupeng:
## Why I'm doing:

`calculate_retry_delay()` in `be/src/storage/lake/vacuum.cpp` currently
returns a fully deterministic exponential backoff
(`min_delay * 2^attempted_retries`). When multiple CNs hit the same S3
throttling (`Please reduce your request rate`) at the same time, they
all back off by the same amount and retry in lockstep, which keeps the
prefix-level QPS high and prolongs the throttled state.

Decorrelated jitter spreads retries across a wider, ever-growing
window, and each CN's window diverges from the others over successive
attempts.

### Example: 100 CN concurrently throttled, base = 100ms

| | First retry timing | First retry concurrent rate |
| --- | --- | --- |
| **Before** (deterministic) | all 100 CN retry at exactly 100ms | ∞ instantaneously (all 100 in a single instant) |
| **After** (decorrelated jitter) | uniformly distributed in `[100, 300]ms` | ~500 QPS (100 CN spread over a 200ms window) |

S3 per-prefix DELETE rate limit is 3500 QPS, so the post-change rate
comfortably stays well below it. Successive retries de-correlate
further: each CN's `last_delay` diverges from the others, so the
window each CN samples from on the next iteration is different.

## What I'm doing:

- Move `calculate_retry_delay()` out of the anonymous namespace in
  `vacuum.cpp` into `starrocks::lake` so it can be reached from the
  unit test via a forward declaration without exposing it in
  `vacuum.h`.
- Replace the deterministic backoff with **decorrelated jitter** (AWS
  Architecture Blog "Exponential Backoff And Jitter"):
  `next_delay = min(cap, rand([base, last_delay * 3]))`, where
  `base = lake_vacuum_retry_min_delay_ms` and
  `cap = base * 2^lake_vacuum_retry_max_attempts`. Caller passes
  `last_delay` (= `base` on the first attempt) and feeds the returned
  value back on the next call.
- Update `delete_files_with_retry()` to track `last_delay` across loop
  iterations.
- Use `butil::fast_rand_less_than()` (brpc thread-local PRNG) instead
  of `std::rand()` to avoid contention on the global rand state.
- Switch the shift expressions to `1L << ...` to avoid signed-int
  shift narrowing if `lake_vacuum_retry_max_attempts` is raised in
  the future.
- Add `LakeVacuumTest2.test_calculate_retry_delay_jitter` covering:
  - first-call window `[base, min(cap, base*3)]` with 200 samples;
  - cap clamping (delays must not exceed cap when `last_delay * 3`
    overflows the cap);
  - a 20-step simulated retry chain stays within `[base, cap]`.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/storage/lake/vacuum.cpp
+++ b/be/src/storage/lake/vacuum.cpp
@@ -14,6 +14,7 @@
 
 #include "storage/lake/vacuum.h"
 
+#include <butil/fast_rand.h>
 #include <butil/time.h>
 #include <bvar/bvar.h>
 
@@ -99,6 +100,29 @@ static bvar::PassiveStatus<int> g_queued_delete_file_tasks("lake_vacuum_queued_d
                                                            get_num_delete_file_queued_tasks, nullptr);
 static bvar::PassiveStatus<int> g_active_delete_file_tasks("lake_vacuum_active_delete_file_tasks",
                                                            get_num_active_file_queued_tasks, nullptr);
+
+// Decorrelated jitter (AWS Architecture Blog "Exponential Backoff And Jitter"):
+//   next_delay = min(cap, rand([base, last_delay * 3]))
+// where cap = base * 2^max_retries.
+//
+// Pure helper: all knobs (base, max_retries) are passed in by the caller so the function has no
+// hidden dependencies on global config and is trivial to unit-test. Caller passes
+// last_delay (= base on the first attempt) and feeds the returned value back on the next call.
+//
+// Compared to deterministic backoff, retry timestamps of independent CNs grow decorrelated over
+// successive attempts even when they all start throttled at the same moment, keeping the
+// per-prefix request rate below the remote storage's limit. The returned value is always
+// in [base, cap], so the caller-configured minimum delay is preserved as a floor.
+int64_t calculate_retry_delay(int64_t last_delay, int64_t base, int64_t max_retries) {
+    int64_t cap = base * (1L << max_retries);
+    int64_t upper = std::min(cap, last_delay * 3);
+    if (upper <= base) {
+        return base;
+    }
+    int64_t range = upper - base + 1;
+    return base + static_cast<int64_t>(butil::fast_rand_less_than(range));
+}
+
 namespace {
 
 // Delete .vi files for segments in a rowset using segment_metas metadata.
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg (dismissed):

be/src/storage/lake/vacuum.cpp
```diff
@@ -99,6 +100,25 @@ static bvar::PassiveStatus<int> g_queued_delete_file_tasks("lake_vacuum_queued_d
...
+
+// Decorrelated jitter (AWS Architecture Blog "Exponential Backoff And Jitter"):
+//   next_delay = min(cap, rand([base, last_delay * 3]))
+// where base = lake_vacuum_retry_min_delay_ms and cap = base * 2^lake_vacuum_retry_max_attempts.
+// Caller passes last_delay (= base on the first attempt) and feeds the returned value back on
+// the next call. Compared to deterministic backoff, retry timestamps of independent CNs grow
+// decorrelated over successive attempts even when they all start throttled at the same moment,
+// keeping the per-prefix request rate below the remote storage's limit.
+int64_t calculate_retry_delay(int64_t last_delay) {
+    int64_t base = config::lake_vacuum_retry_min_delay_ms;
```
kevincai: move the two configvars out, passed in as input parameters, so this `calculate_retry_delay` can be purely static and independent. It will be the caller's responsibility how these 3 parameters set.

starrocks-xupeng: Done — `calculate_retry_delay` is now a pure function taking `(last_delay, base, max_retries)`. `delete_files_with_retry` reads the two configvars once at the top of the function and passes them down. Unit test simplified accordingly (no more config mutation / DeferOp restore). Pushed as part of the latest amend.

be/src/storage/lake/vacuum.cpp
```diff
@@ -99,6 +100,16 @@ static bvar::PassiveStatus<int> g_queued_delete_file_tasks("lake_vacuum_queued_d
...
 static bvar::PassiveStatus<int> g_active_delete_file_tasks("lake_vacuum_active_delete_file_tasks",
                                                            get_num_active_file_queued_tasks, nullptr);
+
+// Equal jitter: returned delay is uniformly distributed in [base/2, base], where
+// base = lake_vacuum_retry_min_delay_ms * 2^attempted_retries. Random component avoids
+// synchronized retries across CNs that would otherwise prolong throttling.
+int64_t calculate_retry_delay(int64_t attempted_retries) {
+    int64_t base = config::lake_vacuum_retry_min_delay_ms * (1L << attempted_retries);
+    int64_t half = base / 2;
+    return half + static_cast<int64_t>(butil::fast_rand_less_than(half + 1));
```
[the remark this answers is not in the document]
starrocks-xupeng: Re-reading the code, this concern doesn't apply: `delete_files_with_retry` initializes `last_delay = base` and feeds the returned value back on every iteration; the function body returns `base + rand_less_than(upper - base + 1)` which is always >= `base`. So the configured `lake_vacuum_retry_min_delay_ms` is preserved as a hard floor — the first retry sleeps in `[base, min(cap, base*3)]`, never below `base`. Added a note in the function comment to make this contract explicit.

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

PR #74112, opened 2026-06-01
[BugFix] Add vacuum batch-size + retry-count metrics

starrocks-xupeng:
## Why I'm doing:

The existing lake vacuum bvars (`lake_vacuum_deleted_files`,
`lake_vacuum_del_file_fails`, `lake_vacuum_del_file_latency`) expose
cumulative or per-event values but no per-minute aggregates. When tuning
`lake_vacuum_min_batch_delete_size` or rolling out retry-related changes,
operators have to derive trends manually from the raw counters.

Two new 60s-window aggregates surface the answers directly:

- *Did bumping `lake_vacuum_min_batch_delete_size` actually raise the
  average batch size on the wire?* — needs per-minute batch volume.
- *Is the cluster currently hitting S3 throttling and retrying?* — needs
  per-minute retry counts.

## What I'm doing:

- Add `lake_vacuum_del_file_batch_size_minute`: recent (60s) mean of
  files-per-batch across `DeleteObjects` calls. Implemented as a
  `bvar::Window<bvar::IntRecorder>` so the exposed value tracks the
  current batch size after `lake_vacuum_min_batch_delete_size` tuning
  instead of the lifetime average dominated by historical samples.
  Recording happens once per logical batch (before retries) so that
  retried batches do not contribute duplicate samples that would bias
  the mean during throttling — the very condition operators are trying
  to read.
- Add `lake_vacuum_del_file_retries_minute`: number of delete retries
  triggered in the last 60s. Surfaces transient S3 throttling pressure
  (RequestRate / try-again responses) and is the direct signal for
  whether jitter / backoff changes reduce retry frequency.
- Assignment sites:
  - `delete_single_batch` lambda head (once per logical batch, before
    retries): `g_del_file_batch_size << batch.size();`
  - `delete_files_with_retry` retry branch (each retry trigger):
    `g_del_file_retries << 1;`
- Both windows use the 60s convention already established for every
  other `lake` module bvar window (`metacache.cpp`, `segment.cpp`),
  and follow the existing `_minute` naming suffix.

No unit test: bvar is a brpc-validated library and these are pure
metric assignments — there is no application-level logic to test.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5


CelerData-Reviewer:
@codex review

starrocks-xupeng:
Addressed the Codex P2 ("Count each retried delete batch"): moved `g_del_file_batch_size << paths.size()` from the `delete_single_batch` lambda entry into the `delete_files_with_retry` attempt loop, so every attempt (including retries) is recorded. The metric now reflects the real wire-level DeleteObjects batch volume — which is the case that matters most during throttling.

Also flipped the PR type to BugFix so the change auto-backports.

CelerData-Reviewer:
@codex review

be/src/storage/lake/vacuum.cpp
```diff
@@ -102,6 +102,19 @@ static bvar::PassiveStatus<int> g_queued_delete_file_tasks("lake_vacuum_queued_d
 static bvar::PassiveStatus<int> g_active_delete_file_tasks("lake_vacuum_active_delete_file_tasks",
                                                            get_num_active_file_queued_tasks, nullptr);
 
+// Distribution of files-per-batch across DeleteObjects calls. IntRecorder exposes the
+// instantaneous mean as the metric value plus *_count / *_sum / *_qps sub-metrics, so
```
kevincai: This comment is inaccurate. I checked `brpc/src/bvar/recorder.h`: an exposed `IntRecorder` does **not** publish `*_count` / `*_sum` / `*_qps` sub-metrics -- that is `LatencyRecorder` behavior. `IntRecorder::describe()`/`get_value()` return `_combiner.combine_agents()`, i.e. a single cumulative average. The header even notes (line ~173) that `IntRecorder is often used as the source of data and not exposed`, the intended pattern being to wrap it in a `Window`. Please correct the comment to match what the type actually exposes.

be/src/storage/lake/vacuum.cpp
```diff
@@ -102,6 +102,19 @@ static bvar::PassiveStatus<int> g_queued_delete_file_tasks("lake_vacuum_queued_d
...
+// operators can directly tell if lake_vacuum_min_batch_delete_size tuning actually changes
+// the average batch size on the wire (instead of having to derive it from a 60s-windowed sum).
+static bvar::IntRecorder g_del_file_batch_size("lake_vacuum_del_file_batch_size");
```
kevincai: A bare exposed `IntRecorder` reports the **lifetime** mean (sum/num since process start), not a recent value. After the BE has been up for a while, a change to `lake_vacuum_min_batch_delete_size` barely moves the cumulative average -- which is exactly the trend-detection the PR set out to provide. To get a recent average, wrap it in a window like the metacache/segment convention and leave the recorder anonymous:

```cpp
static bvar::IntRecorder g_del_file_batch_size;
static bvar::Window<bvar::IntRecorder> g_del_file_batch_size_minute(
        "lake_vacuum", "del_file_batch_size_minute", &g_del_file_batch_size, 60);
```

or use a windowed `Adder` sum of `batch.size()` divided by the delete-call QPS (the approach the PR description actually describes). Note also the inconsistency within this PR: `g_del_file_retries` is anonymous with only its `Window` exposed, but this recorder is exposed raw -- please make them consistent.

Separately, the metric name here is `lake_vacuum_del_file_batch_size` (no `_minute`), but the PR description says `lake_vacuum_del_file_batch_size_minute` and describes a windowed sum. The description and code need to be reconciled.

be/src/storage/lake/vacuum.cpp
```diff
@@ -164,8 +177,12 @@ Status delete_files_with_retry(FileSystem* fs, std::span<const std::string> path
     const int64_t max_retries = config::lake_vacuum_retry_max_attempts;
     int64_t last_delay = base;
     for (int64_t attempted_retries = 0; /**/; attempted_retries++) {
+        // Record on every attempt (including retries) so the metric reflects the actual
+        // DeleteObjects batch volume sent on the wire, not just the unique batch count.
+        g_del_file_batch_size << paths.size();
```
kevincai: Recording at the top of the retry loop means a batch retried N times contributes N identical samples, over-weighting batches that fail and retry (typically during throttling) -- the very condition that skews the average operators will read. For an *average batch size* this is a bias; if the intent is wire volume a windowed sum/QPS expresses it cleanly, and if the intent is typical batch size, record once before the loop instead.

kevincai (commented):
Thanks for adding these. The retry-count metric looks good. The batch-size metric has a couple of issues I left inline: as written it reports a lifetime cumulative mean rather than a recent window, so it won't surface the tuning trend the PR is after, and the accompanying comment describes IntRecorder behavior that isn't accurate. Details inline.

CelerData-Reviewer:
@codex review

starrocks-xupeng:
Thanks Kevin — all four points addressed in the latest amend (push: `c325d01142`). Below is the actual code.

```diff
- // Distribution of files-per-batch across DeleteObjects calls. IntRecorder exposes the
- // instantaneous mean as the metric value plus *_count / *_sum / *_qps sub-metrics, so
- // operators can directly tell if lake_vacuum_min_batch_delete_size tuning actually changes
- // the average batch size on the wire (instead of having to derive it from a 60s-windowed sum).
- static bvar::IntRecorder g_del_file_batch_size("lake_vacuum_del_file_batch_size");
+ // Recent (60s) mean of files-per-batch across DeleteObjects calls. The IntRecorder is wrapped
+ // in a Window so the exposed value tracks current batch size after
+ // lake_vacuum_min_batch_delete_size tuning instead of the lifetime average dominated by
+ // historical samples. Recording happens once per logical batch (before retries) so that retried
+ // batches do not contribute duplicate samples that would bias the mean during throttling -- the
+ // very condition operators are trying to read.
+ static bvar::IntRecorder g_del_file_batch_size;
+ static bvar::Window<bvar::IntRecorder> g_del_file_batch_size_minute("lake_vacuum", "del_file_batch_size_minute",
+                                                                     &g_del_file_batch_size, 60);

  Status delete_files_with_retry(FileSystem* fs, std::span<const std::string> paths) {
      for (int64_t attempted_retries = 0; /**/; attempted_retries++) {
-         // Record on every attempt (including retries) so the metric reflects the actual
-         // DeleteObjects batch volume sent on the wire, not just the unique batch count.
-         g_del_file_batch_size << paths.size();
          auto st = fs->delete_files(paths);

      auto delete_single_batch = [fs](std::span<const std::string> batch) -> Status {
+         g_del_file_batch_size << batch.size();
```

| Your comment | Fix |
| --- | --- |
| (line 106) IntRecorder doesn't expose *_count/_sum/_qps -- LatencyRecorder does | Comment rewritten; only describes "recent 60s mean" now. |
| (line 109) Bare IntRecorder reports lifetime mean | Wrapped in `bvar::Window<bvar::IntRecorder>`, 60s window matching metacache/segment convention. |
| (line 109) Style inconsistency vs `g_del_file_retries` | `g_del_file_batch_size` is now anonymous; only the windowed `_minute` variant is exposed, matching `g_del_file_retries` / `g_del_file_retries_minute`. |
| (line 109) Metric name vs PR description mismatch | Metric name is now `lake_vacuum_del_file_batch_size_minute`; PR description updated to describe a `Window<IntRecorder>` mean rather than "sum / QPS". |
| (line 182) Recording in retry loop over-weights throttled batches | Recording moved back to `delete_single_batch` lambda entry (once per logical batch, before retries). |

Verified on the test cluster after rebuild: with `lake_vacuum_min_batch_delete_size = 5`, the new `lake_vacuum_del_file_batch_size_minute` reads `5` (cumulative `lake_vacuum_deleted_files` also climbing), confirming the windowed mean reflects current batch size rather than lifetime average.

be/src/storage/lake/vacuum.cpp
starrocks-xupeng: done

be/src/storage/lake/vacuum.cpp
starrocks-xupeng: done

be/src/storage/lake/vacuum.cpp
starrocks-xupeng: done

kevincai (approved):

xiangguangyxg (approved):

PR #74113, opened 2026-06-01
[Enhancement] Fail load writes fast instead of exiting BE on delta writer hang

meegoo:
## Why I'm doing:
When the shared async delta writer thread pool is saturated because of a slow or hung disk, ThreadPoolExecutor::submit() used to call LOG(FATAL) and take the whole BE down, killing writes and queries on every healthy disk too.

## What I'm doing:
Instead, shed load gracefully:
- submit() no longer crashes; on give-up it runs the consumer inline (matching brpc's own fallback) so already-queued work still makes progress.
- AsyncDeltaWriter::write()/commit() and lake::AsyncDeltaWriter write()/finish() reject new work with a retryable ServiceUnavailable error once the pool stays full with no progress for be_exit_after_disk_write_hang_second, so loads fail fast and the FE can retry or reroute.

The legacy exit behavior is preserved behind the new mutable config enable_load_fail_fast_when_disk_write_hang (default true); set it to false to restore exiting the BE on a disk write hang.

Add ThreadPoolExecutor::is_overloaded(), a ThreadPool::max_queue_size() getter, unit tests, and BE config docs (en/zh/ja).

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -627,8 +627,21 @@ CONF_mInt64(max_queueing_memtable_per_tablet, "2");
 // 0 means disable
 CONF_mInt64(stale_memtable_flush_time_sec, "0");
 
-// delta writer hang after this time, be will exit since storage is in error state
+// The async delta writer thread pool is treated as "hung" once it stays saturated
+// (queue full and no task completes) for longer than this many seconds, which usually
+// indicates a slow or stuck disk.
+// - When enable_load_fail_fast_when_disk_write_hang is true (default), new load writes
+//   are failed fast with a retryable error after this timeout.
+// - When it is false, the BE keeps the legacy behavior and exits the process after this
+//   timeout so the storage error becomes visible.
 CONF_Int32(be_exit_after_disk_write_hang_second, "60");
+
+// When the async delta writer thread pool stays saturated for longer than
+// be_exit_after_disk_write_hang_second (typically a slow or hung disk), fail new load
+// writes fast with a retryable error instead of exiting the whole BE process. This keeps
+// the node serving the other (healthy) disks and lets the FE retry or reroute the load.
+// Set to false to restore the legacy behavior of exiting the BE on a disk write hang.
+CONF_mBool(enable_load_fail_fast_when_disk_write_hang, "true");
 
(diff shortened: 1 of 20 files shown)


dirtysalt (dismissed):
LGTM! 🚀

be/src/common/config.h
```diff
@@ -627,9 +627,22 @@ CONF_mInt64(max_queueing_memtable_per_tablet, "2");
 // 0 means disable
 CONF_mInt64(stale_memtable_flush_time_sec, "0");
 
-// delta writer hang after this time, be will exit since storage is in error state
+// The async delta writer thread pool is treated as "hung" once it stays saturated
+// (queue full and no task completes) for longer than this many seconds, which usually
+// indicates a slow or stuck disk.
+// - When enable_load_fail_fast_when_disk_write_hang is true (default), new load writes
```
kevincai, line 633: can give a better name/alias if in the fail fast path, the be won't exit. `be_exit_after_xxx` looks scary.

be/src/common/util/bthreads/executor.cpp
```diff
@@ -33,20 +33,50 @@ int ThreadPoolExecutor::submit(void* (*fn)(void*), void* args) {
...
             break;
         }
-        LOG(INFO) << "async_delta_writer is busy, retry after " << _busy_sleep_ms << "ms";
+        VLOG(2) << "async_delta_writer is busy, retry after " << _busy_sleep_ms << "ms";
         bthread_usleep(_busy_sleep_ms * 1000);
     }
-    LOG_IF(FATAL, !st.ok()) << "BE exit since submit write fail err=" << st;
-    return st.ok() ? 0 : -1;
+    if (!st.ok()) {
+        if (starrocks::config::enable_load_fail_fast_when_disk_write_hang) {
```
kevincai, line 49: doesn't sound like doing the right thing. Hijack a common used threadpool interface just for loading traffic??

PR #74117, opened 2026-06-01
[BugFix] Prevent StarMgrMetaSyncer from reaping range-colocate PACK shard groups

xiangguangyxg:
## Why I'm doing:

In shared-data mode, range-distribution colocate creates a **PACK shard group** per colocate range. This PACK shard group is tracked only in `ColocateRangeMgr` and is **not attached to any `PhysicalPartition`** (unlike the per-partition SPREAD shard groups).

`StarMgrMetaSyncer.deleteUnusedShardAndShardGroup()` reaps StarOS shard groups that are absent from its "FE-known" set, which it builds solely from `PhysicalPartition.getShardGroupIds()` via `getAllPartitionShardGroupId()`. Because PACK shard groups never appear in that set, they were misclassified as orphans and permanently deleted once older than `shard_group_clean_threshold_sec` (default 3600s) — and `cleanOneGroup()` first deleted the live tablets that are dual members of `[SPREAD, PACK]`.

After deletion, `ColocateRangeMgr` still references the now-deleted PACK shard group, so subsequent `createShards` / `createShardsForSplit` fail with `NOT_EXIST: shard group <id> not exist`, surfacing on both `CREATE TABLE` (joining an existing colocate group) and tablet reshard split jobs.

## What I'm doing:

**Primary fix — recognize PACK shard groups as FE-known:**

- `ColocateRangeMgr.getAllPackShardGroupIds()` enumerates the PACK shard group ids across colocate groups.
- `ColocateTableIndex.getAllPackShardGroupIds()` exposes them under the index read lock.
- `StarMgrMetaSyncer.getAllPartitionShardGroupId()` unions them into the FE-known set.

This only enlarges the protected set, so it can only prevent erroneous deletions; genuine-orphan cleanup behavior is unchanged.

**Encapsulate `ColocateRangeMgr` behind `ColocateTableIndex` (review feedback):**

- All production access now goes through the lock-guarded `ColocateTableIndex.getColocateRanges()`, which returns an immutable snapshot, so callers can neither mutate the range list nor read it lock-free.
- Migrated the 6 former direct callers — `ColocateChecker`, `SplitTabletJobFactory`, `SplitTabletJob`, `RangeColocateScanDispatch`, `OlapTable`, `LocalMetastore`.
- `getColocateRangeMgr()` is now `@VisibleForTesting` (zero production callers; remaining users are cross-package tests).

**Note on orphaned range entries:** a crash between the range-update and add-table journal records can leave a `ColocateRangeMgr` entry with no live colocate group. Such a PACK shard group is intentionally kept protected (a benign leak — the `grpId` is never reused via `getNextId()`) rather than reaped, because reaping it without a per-shard liveness check could delete the half-created table's live tablets.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/ColocateRangeMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/ColocateRangeMgr.java
@@ -19,11 +19,12 @@ import com.google.gson.annotations.SerializedName;
 import com.starrocks.common.Range;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * Manages colocate ranges for range distribution colocate groups.
@@ -136,20 +137,29 @@ public class ColocateRangeMgr {
     }
 
     /**
-     * Returns all PACK shard group ids currently tracked across every colocate group.
+     * Returns the PACK shard group ids tracked for the given colocate groups.
      *
-     * <p>These PACK shard groups are created by FE but are not attached to any
+     * <p>PACK shard groups are created by FE but are not attached to any
      * {@code PhysicalPartition}, so callers that build the set of FE-known shard groups
-     * from partitions alone must union this in (e.g. {@code StarMgrMetaSyncer}); otherwise
+     * from partitions alone must union these in (e.g. {@code StarMgrMetaSyncer}); otherwise
      * a live PACK shard group would be misclassified as an orphan and reaped.
      *
+     * <p>The caller passes the colocate group ids that are still live (referenced by the
+     * colocate table index). Restricting to those ids skips stale range entries that can
+     * linger after a crash between the range-update and add-table journal records, so a
+     * genuinely orphaned PACK shard group is still eligible for cleanup.
+     *
+     * @param colocateGroupIds the live colocate group ids to collect PACK shard groups for
      * @return a new set of PACK shard group ids (empty if none); never null
      */
-    public Set<Long> getAllPackShardGroupIds() {
-        return colocateGroupToRanges.values().stream()
-                .flatMap(List::stream)
-                .map(ColocateRange::getShardGroupId)
-                .collect(Collectors.toSet());
+    public Set<Long> getPackShardGroupIds(Collection<Long> colocateGroupIds) {
+        Set<Long> packShardGroupIds = new HashSet<>();
+        for (Long colocateGroupId : colocateGroupIds) {
+            for (ColocateRange range : colocateGroupToRanges.getOrDefault(colocateGroupId, Collections.emptyList())) {
+                packShardGroupIds.add(range.getShardGroupId());
+            }
+        }
+        return packShardGroupIds;
     }
 
     // ---- Initialize ----
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/ColocateRangeMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/ColocateRangeMgr.java
@@ -22,6 +22,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * Manages colocate ranges for range distribution colocate groups.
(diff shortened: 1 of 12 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/catalog/ColocateRangeMgr.java
```diff
@@ -133,6 +135,23 @@ public boolean containsColocateGroup(long colocateGroupId) {
...
+     * Returns all PACK shard group ids currently tracked across every colocate group.
+     *
+     * <p>These PACK shard groups are created by FE but are not attached to any
+     * {@code PhysicalPartition}, so callers that build the set of FE-known shard groups
+     * from partitions alone must union this in (e.g. {@code StarMgrMetaSyncer}); otherwise
+     * a live PACK shard group would be misclassified as an orphan and reaped.
+     *
+     * @return a new set of PACK shard group ids (empty if none); never null
+     */
+    public Set<Long> getAllPackShardGroupIds() {
```
kevincai: this class may have confused lock design. It doesn't have any lock mechanism to ensure its private map integrity, yet it exposes a handful public interface that can access the private map. This will assume the caller has the responsibility to mutable each other properly. Need to rethink about this design, make the module self-sustainable.

fe/fe-core/src/main/java/com/starrocks/catalog/ColocateTableIndex.java
```diff
@@ -164,6 +164,24 @@ public ColocateRangeMgr getColocateRangeMgr() {
...
+    /**
+     * Returns all PACK shard group ids tracked by the range-colocate metadata.
+     *
+     * <p>PACK shard groups are created by FE but not attached to any {@code PhysicalPartition},
+     * so {@code StarMgrMetaSyncer} must union these into its FE-known shard group set to avoid
+     * reaping live PACK shard groups as orphans.
+     *
+     * @return a new set of PACK shard group ids (empty if none); never null
+     */
+    public Set<Long> getAllPackShardGroupIds() {
```
kevincai, line 197: If all the interaction to ColocateRangeMgr is delegated through this ColocateTableIndex, interface `public ColocateRangeMgr getColocateRangeMgr() ` should not be public at all, or at most just for Unit Test

fe/fe-core/src/test/java/com/starrocks/lake/StarMgrMetaSyncerTest.java
```diff
@@ -896,6 +915,57 @@ public void deleteShardGroup(List<Long> groupIds) throws StarClientException {
...
     }
 
+    // A range-colocate PACK shard group lives in ColocateRangeMgr (surfaced via
+    // getAllPackShardGroupIds), not on any PhysicalPartition. It must be reported as FE-known
+    // and never reaped by deleteUnusedShardAndShardGroup, even when older than the threshold.
+    @Test
+    public void testPackShardGroupNotReapedWhenStillReferenced() {
+        boolean oldValue = Config.meta_sync_force_delete_shard_meta;
+        Config.meta_sync_force_delete_shard_meta = false;
+        Config.shard_group_clean_threshold_sec = 0;
```
kevincai, line 926: this config var is not restored after the test.

fe/fe-core/src/main/java/com/starrocks/catalog/ColocateRangeMgr.java
starrocks-xupeng: +1

xiangguangyxg: Thanks — I've encapsulated ColocateRangeMgr behind ColocateTableIndex. All production access now goes through ColocateTableIndex, which owns the read/write lock: the new `getColocateRanges()` delegate acquires the read lock and returns an immutable snapshot, so callers can neither mutate the map nor read it lock-free. After migrating the callers there are now zero production sites reaching the unsynchronized map directly. I kept the single-owner-lock model (ColocateTableIndex's lock guards all of its sub-state) rather than adding a second lock inside ColocateRangeMgr, to stay consistent with the rest of the class.

fe/fe-core/src/main/java/com/starrocks/catalog/ColocateTableIndex.java
xiangguangyxg: Migrated all 6 production callers (ColocateChecker, SplitTabletJobFactory, SplitTabletJob, RangeColocateScanDispatch, OlapTable, LocalMetastore) to the lock-guarded `ColocateTableIndex.getColocateRanges()` delegate, so production callers of `getColocateRangeMgr()` are now zero. I kept the getter only as a `@VisibleForTesting` seam: its remaining users are tests in `com.starrocks.alter.reshard` and `com.starrocks.planner`, which are different packages from `com.starrocks.catalog`, so package-private wouldn't be reachable. This follows the prevailing fe-core convention — `@VisibleForTesting` paired with the minimum visibility the test's package needs (package-private when same-package, e.g. `StarMgrMetaSyncer.getAllPartitionShardGroupId()`; public when cross-package, ~160 such cases).

fe/fe-core/src/test/java/com/starrocks/lake/StarMgrMetaSyncerTest.java
xiangguangyxg: Fixed — the test now saves and restores both `meta_sync_force_delete_shard_meta` and `shard_group_clean_threshold_sec`.

kevincai (approved):

starrocks-xupeng (approved):

sevev (approved):

PR #73845, opened 2026-05-26
[Enhancement] Alter light_weight_tablet_creation + skip CreateReplicaTask in schema change/rollup

starrocks-xupeng:
## Why I'm doing:

Build on the table property and CREATE TABLE skip from #73422 by wiring up the ALTER path and the corresponding CreateReplicaTask skip for shadow tablets in lake schema change / rollup jobs. The shadow tablet's version 1 metadata is materialized on demand by the CN-side fallback (already on main, see #72270 and #72364) when the first read or publish hits it.

## What I'm doing:

**ALTER TABLE property change**
* `AlterTableClauseAnalyzer` validates the value (true/false) and rejects the property on non-cloud-native tables.
* `AlterJobExecutor.visitModifyTablePropertiesClause` now routes the property through `SchemaChangeHandler.processLakeTableAlterMeta` so the cloud-native dispatch picks it up.
* `SchemaChangeHandler.createAlterMetaJob` handles `light_weight_tablet_creation` as a pure FE property (mirrors `lake_compaction_max_parallel`): no-op when the value is unchanged, otherwise routes to `LocalMetastore.alterTableProperties`.
* `LocalMetastore.alterLightWeightTabletCreation` appends an applier that calls `OlapTable.setLightWeightTabletCreation`. The change is persisted by the existing `ModifyTablePropertyOperationLog` under `OP_ALTER_TABLE_PROPERTIES`.
* `TableProperty.buildProperty(OP_ALTER_TABLE_PROPERTIES)` now invokes `buildLightWeightTabletCreation` so replay on follower / FE restart syncs the typed field from the properties map.

**Schema change / rollup skip**
* `LakeTableSchemaChangeJob.runPendingJob` and `LakeRollupJob.runPendingJob` snapshot `table.isLightWeightTabletCreation()` and skip `CreateReplicaTask` construction plus `sendAgentTaskAndWait` when enabled. The job still advances through `addShadowIndexToCatalog`, `watershedTxnId` allocation, and `persistStateChange(WAITING_TXN)`.

**Tests**
* `CreateLakeTableTest.testAlterLightWeightTabletCreation`: toggle on/off/no-op + invalid value rejection.
* `LakeTableSchemaChangeJobTest.testLightWeightTabletCreationSkipsSendTask`: light-weight schema change advances to `WAITING_TXN` without invoking `sendAgentTaskAndWait`.
[...]

PR #74132, opened 2026-06-02
[Enhancement] Backfill v1 metadata when light_weight_tablet_creation is altered to false

starrocks-xupeng:
## Why I'm doing:

Build on #73422 / #73845. When a table has `light_weight_tablet_creation` flipped from `true` to `false`, every PhysicalPartition that has never been published (visibleVersion == 1) lacks the v1 metadata and schema file in object storage. A downgrade target that does not include the CN-side fallback would fail to read those tablets. Backfill those files at ALTER time so the table behaves identically to a table that was created without the property.

## What I'm doing:

**Mechanics**
* `TabletTaskExecutor.CreateTabletOption` gains a `backfill` flag. The light-weight early-return in `buildPartitionsSequentially` / `buildPartitionsConcurrently` is relaxed so a backfill caller can drive `CreateReplicaTask` dispatch even while the table still reports `isLightWeightTabletCreation() == true`.
* `LocalMetastore.buildPartitions` exposes an overload with the backfill flag. When set, the alive-node precheck always throws on zero alive nodes (backfill needs CNs).
* `LocalMetastore.alterLightWeightTabletCreation` detects the `true -> false` transition, filters PhysicalPartitions to those with `visibleVersion == PARTITION_INIT_VERSION` (already-published partitions have v2+ in object storage and need no v1), and runs the backfill synchronously under the caller's existing table WRITE lock. If backfill throws, the edit log is not written and the property stays `true`. Empty tables short-circuit without dispatch.

**Tests**
* `CreateLakeTableTest.testAlterLightWeightTabletCreationTrueToFalseTriggersBackfill`: a freshly created light-weight table altered to `false` invokes `TabletTaskExecutor` with `option.backfill = true` for its unpublished partition.
* `CreateLakeTableTest.testAlterLightWeightTabletCreationFalseToTrueSkipsBackfill`: `false -> true` must not trigger backfill.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [ ] Parameter changes
- [ ] Policy changes
- [ ] Feature removed
- [ ] Miscellaneous: ALTER on an opt-in property; only takes effect when a user explicitly flips the property from `true` to `false`.

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java
@@ -1155,6 +1155,16 @@ public class LocalMetastore implements ConnectorMetadata, MVRepairHandler, Memor
             metaChanged = true;
         }
 
+        // The light_weight_tablet_creation snapshot from the lock-free phase decides whether
+        // the loop in buildPartitions sends CreateReplicaTask. If a concurrent ALTER flipped
+        // this property between the snapshot and the commit, the lock-free phase used a stale
+        // value and the new partition would land without v1 metadata under a table that now
+        // claims it has v1 in object storage - which silently undermines the downgrade
+        // backfill in alterLightWeightTabletCreation. Treat this as a meta change and retry.
+        if (olapTable.isLightWeightTabletCreation() != copiedTable.isLightWeightTabletCreation()) {
+            metaChanged = true;
+        }
+
         if (metaChanged) {
             throw new DdlException("Table[" + tableName + "]'s meta has been changed. try again.");
         }
@@ -2018,6 +2028,11 @@ public class LocalMetastore implements ConnectorMetadata, MVRepairHandler, Memor
 
     void buildPartitions(Database db, OlapTable table, List<PhysicalPartition> partitions,
                          ComputeResource computeResource) throws DdlException {
+        buildPartitions(db, table, partitions, computeResource, false);
+    }
+
+    void buildPartitions(Database db, OlapTable table, List<PhysicalPartition> partitions,
+                         ComputeResource computeResource, boolean backfill) throws DdlException {
         if (partitions.isEmpty()) {
             return;
         }
@@ -2033,7 +2048,7 @@ public class LocalMetastore implements ConnectorMetadata, MVRepairHandler, Memor
                 }
             }
         }
-        if (numAliveNodes == 0 && !table.isLightWeightTabletCreation()) {
+        if (numAliveNodes == 0 && (backfill || !table.isLightWeightTabletCreation())) {
             if (RunMode.isSharedDataMode()) {
                 throw new DdlException("no alive compute nodes");
             } else {
@@ -2053,6 +2068,7 @@ public class LocalMetastore implements ConnectorMetadata, MVRepairHandler, Memor
         option.setEnableTabletCreationOptimization(table.isCloudNativeTableOrMaterializedView()
                 && (Config.lake_enable_tablet_creation_optimization || table.isFileBundling()));
         option.setGtid(GlobalStateMgr.getCurrentState().getGtidGenerator().nextGtid());
+        option.setBackfill(backfill);
 
         try {
             GlobalStateMgr.getCurrentState().getConsistencyChecker().addCreatingTableId(table.getId());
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java
```diff
@@ -4149,9 +4155,36 @@ private void alterLightWeightTabletCreation(OlapTable table,
...
+                    partitionsToBackfill.add(pp);
+                }
+            }
+        }
+        if (partitionsToBackfill.isEmpty()) {
+            return;
+        }
+        LOG.info("backfilling tablet metadata for table {}.{}: {} physical partitions",
+                db.getFullName(), table.getName(), partitionsToBackfill.size());
+        buildPartitions(db, table, partitionsToBackfill, computeResource, true /* backfill */);
```
banmoy, line 4195: `buildPartitions` here dispatches `CreateReplicaTask` and then blocks in `waitForFinished` (up to `tablet_create_timeout_second` per batch) **while the table WRITE lock is held** — this whole alter runs under `AlterJobExecutor.processLakeTableAlterMeta` → `lockTablesWithIntensiveDbLock(..., WRITE)`. So every load / query / DDL on the table is blocked for the entire backfill.

`addPartitions` / `addSubPartitions` deliberately avoid exactly this: they snapshot a shadow copy under a READ lock, release it, run `buildPartitions` **outside** the lock, then re-acquire WRITE only to commit the metadata. Could the backfill do the same?

Caveat: the WRITE lock also serializes this backfill against a concurrent publish, so it can't simply be moved outside the lock without re-validating partition state afterwards.

fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java
```diff
@@ -4149,9 +4155,36 @@ private void alterLightWeightTabletCreation(OlapTable table,
...
+        }
         appliers.add(() -> table.setLightWeightTabletCreation(newValue));
     }
 
+    private void backfillLightWeightTabletMetadata(Database db, OlapTable table) throws DdlException {
+        ConnectContext ctx = ConnectContext.get();
+        ComputeResource computeResource = ctx != null ? ctx.getCurrentComputeResource()
+                                                      : WarehouseManager.DEFAULT_RESOURCE;
+        List<PhysicalPartition> partitionsToBackfill = new ArrayList<>();
+        for (Partition partition : table.getPartitions()) {
```
banmoy, line 4183: This only iterates partitions already committed to the live table, which opens a race with concurrent partition creation (`ADD PARTITION` / `addSubPartitions`):

1. `addPartitions` snapshots a shadow copy under a READ lock — capturing `light_weight_tablet_creation = true` — then **releases the lock** and runs `buildPartitions` outside it, so it skips `CreateReplicaTask` based on that stale snapshot.
2. Concurrently this ALTER takes the WRITE lock and backfills. The in-flight partition isn't committed yet, so it's **invisible** to this loop; the ALTER then sets the property to `false` and commits.
3. `addPartitions` re-acquires WRITE and commits its new partition. `checkIfMetaChange` (LocalMetastore.java:1131) compares only index count / schemaHash / distribution — **not** `lightWeightTabletCreation` — so the stale-snapshot commit is not rejected.

Result: the new partition lands with `visibleVersion == 1` while the table is `light_weight_tablet_creation = false`, materialized by neither path. It self-heals on first write via the CN-side fallback, but not if it is never written before a downgrade — which is precisely the scenario this backfill exists to prevent.

Adding `lightWeightTabletCreation` to the `checkIfMetaChange` comparison would turn this into a safe "meta has been changed. try again." retry.

fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java
starrocks-xupeng: it's on purpose, because this operation only occurs under extreme situation, so for simplicity, we do the operation under the lock

fe/fe-core/src/main/java/com/starrocks/server/LocalMetastore.java
starrocks-xupeng: done

banmoy (approved):

kevincai (approved):

xiangguangyxg (approved):

PR #74147, opened 2026-06-02
[Enhancement] Record per-file size of lake DCG .cols files to skip stat on open

luohaha:
## Why I'm doing:

In shared-data mode, each delta-column-group (DCG) `.cols` file is itself a full segment, so its byte length is recoverable from the segment trailer at read time. However, unlike regular segments — whose size is persisted in `RowsetMetadataPB.segment_size` and threaded through `FileInfo.size` so that opening the segment can skip a stat/`HeadObject` — `DeltaColumnGroupVerPB` stored no size at all. As a result `Segment::new_dcg_segment()` constructed a `FileInfo` without `.size`, and every open of a `.cols` segment paid an extra `HeadObject` RPC against object storage (non-trivial latency on S3/OSS/OBS, etc.).

## What I'm doing:

Carry the already-computed `.cols` file size through to the reader so the optimization that regular segments enjoy also applies to DCG files:

- `gensrc/proto/lake_types.proto`: add `repeated int64 column_file_sizes = 6 [packed]` to `DeltaColumnGroupVerPB`, strictly 1:1 with `column_files`.
- `be/src/storage/lake/meta_file.{h,cpp}`: `MetaFileBuilder::append_dcg` takes a parallel `file_sizes` vector, writes one size per file, and carries forward old sizes (or `0` = unknown) when retaining existing entries so the array stays aligned with `column_files`.
- `be/src/storage/lake/column_mode_partial_update_handler.cpp`: collect `segment_file_size` (already produced by `SegmentWriter::finalize()`) and pass it to `append_dcg`.
- `be/src/storage/delta_column_group.{h,cpp}`: new `_column_file_sizes` member + accessor; lake `load()` reads the field.
- `be/src/storage/rowset/segment.cpp`: `new_dcg_segment` sets `FileInfo.size` when the recorded size is `> 0`, otherwise falls back to discovering it.

**Backward compatibility:** a size of `0` (or an absent `column_file_sizes`, e.g. data written before this field existed) is treated as "unknown" and transparently falls back to the previous stat/`HeadObject` behavior. No migration required.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/delta_column_group.cpp
+++ b/be/src/storage/delta_column_group.cpp
@@ -142,6 +142,9 @@ Status DeltaColumnGroup::load(int64_t version, const DeltaColumnGroupVerPB& dcg_
     for (const auto& encryption_meta : dcg_ver_pb.encryption_metas()) {
         _encryption_metas.push_back(encryption_meta);
     }
+    for (const auto& column_file_size : dcg_ver_pb.column_file_sizes()) {
+        _column_file_sizes.push_back(column_file_size);
+    }
     for (const auto& ucids : dcg_ver_pb.unique_column_ids()) {
         _column_uids.emplace_back();
         for (const auto& cid : ucids.column_ids()) {
--- a/be/src/storage/delta_column_group.h
+++ b/be/src/storage/delta_column_group.h
@@ -105,6 +105,10 @@ public:
 
     const std::vector<std::string>& encryption_metas() const { return _encryption_metas; }
 
+    // Per-file size of each `.cols` file, 1:1 with relative_column_files(). May be empty (and
+    // individual entries may be 0) for data written before the size field existed.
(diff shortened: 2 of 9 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg (dismissed):

sevev (approved):

xiangguangyxg (approved):

HangyuanLiu (approved):

PR #74150, opened 2026-06-02
[Enhancement] Add FE config to print load profiles to fe profile log

luohaha:
## Why I'm doing:

Load profiles (stream load, routine load, broker load, merge-commit load) share the **same** in-memory `ProfileManager` store as query profiles. That store is a single `LinkedHashMap` capped by `profile_info_reserved_num` (default 500) and evicted **FIFO across all profile types**. A burst of one type silently evicts the other, and once a profile is evicted it is **unrecoverable** — there is no durable store. This makes it hard to retrieve a useful load profile after the fact.

## What I'm doing:

Add a mutable FE config `enable_print_load_profile_to_log` (default `false`). When enabled, load profiles are additionally printed to `fe.log` at INFO level when they are pushed to `ProfileManager`, so they remain recoverable from the log even after eviction.

- The hook is the single central `ProfileManager.pushProfile`, which already eagerly formats the profile string — so logging reuses that string at no extra cost.
- Only profiles with `QUERY_TYPE = "Load"` (stream/routine/broker/merge load) are printed; query profiles are not affected.
- The default path adds only one config comparison.
- A load profile is printed only when it is actually collected (i.e. `enable_profile = true` or the load crosses the big-load profile threshold), matching the subset `ProfileManager` would otherwise have held.

Docs updated for `FE_parameters` in en/zh/ja.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

> New config `enable_print_load_profile_to_log`, default `false` — behavior is unchanged unless explicitly enabled.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
@@ -38,11 +38,13 @@ import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.gson.Gson;
 import com.starrocks.common.Config;
 import com.starrocks.common.FeConstants;
 import com.starrocks.memory.MemoryTrackable;
 import com.starrocks.memory.estimate.Estimator;
 import com.starrocks.qe.ConnectContext;
+import com.starrocks.qe.QueryDetail;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -72,6 +74,12 @@ import java.util.concurrent.locks.ReentrantReadWriteLock.WriteLock;
  */
 public class ProfileManager implements MemoryTrackable {
     private static final Logger LOG = LogManager.getLogger(ProfileManager.class);
+    // Dedicated profile logger writing to fe.profile.log, whose JSON layout uses the much larger
+    // sys_log_json_profile_max_string_length cap, so large profiles are not truncated (unlike fe.log,
+    // which caps the JSON message field at sys_log_json_max_string_length).
+    private static final Logger PROFILE_LOG = LogManager.getLogger("profile");
+    // Single-line JSON, matching the query profile log written by StmtExecutor.
+    private static final Gson GSON = new Gson();
     private static ProfileManager INSTANCE = null;
     public static final String QUERY_ID             = ProfileKeyDictionary.QUERY_ID;
     public static final String START_TIME           = ProfileKeyDictionary.START_TIME;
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

gengjun-git:
Can this log be printed to fe.profile.log?

luohaha:
> Can this log be printed to fe.profile.log?

Good idea

fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
```diff
@@ -229,6 +233,11 @@ public String pushProfile(ProfilingExecPlan plan, RuntimeProfile profile) {
 
         LOG.debug("push profile for query: {}, remove profile for query: {}", queryId, removedQueryId);
 
+        if (Config.enable_print_load_profile_to_log
+                && "Load".equals(element.infoStrings.get(QUERY_TYPE))) {
+            PROFILE_LOG.info("Load profile for query {}:\n{}", queryId, profileString);
```
banmoy: how about following the same json format as query profile log? Does anywhere depend on the log format?
Image

luohaha: Good point — done. I checked the repo and nothing in-tree reads/parses `fe.profile.log`; it's effectively write-only for external collectors, which expect one JSON object per line (the query profile log already writes single-line `GSON.toJson(queryDetail)`). A multi-line plain-text entry would have broken that, so I switched the load-profile write to emit a single-line `QueryDetail` JSON via the same `profile` logger, matching the query profile log format. `isQuery` is set to false, the profile string goes in the `profile` field, and numeric execution stats that loads don't expose are left at their defaults (same as query profiles serialize unavailable values). Updated in 9cc6074.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
luohaha: done

fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
```diff
@@ -229,9 +237,39 @@ public String pushProfile(ProfilingExecPlan plan, RuntimeProfile profile) {
...
+            detail.setLatency(element.endTimeMs - element.startTimeMs);
+        }
+        detail.setDatabase(info.get(DEFAULT_DB));
+        detail.setSql(info.get(SQL_STATEMENT));
+        detail.setUser(info.get(USER));
+        if (info.get(WAREHOUSE_CNGROUP) != null) {
+            detail.setWarehouse(info.get(WAREHOUSE_CNGROUP));
+        }
+        detail.setState("Aborted".equals(info.get(QUERY_STATE))
+                ? QueryDetail.QueryMemState.FAILED : QueryDetail.QueryMemState.FINISHED);
```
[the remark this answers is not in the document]
luohaha: Good catch — fixed in 65750d7. The earlier code only special-cased the stream-load spelling `Aborted` and defaulted everything else to FINISHED, so merge-commit `ABORTED`/`CANCELLED` were mislabeled as success. Replaced it with a `toLoadState()` helper that maps case-insensitively across all load spellings: `Finished`→FINISHED, `Cancelled`→CANCELLED, `Aborted`→FAILED, and anything else (e.g. broker `Running`, merge intermediate states, or unknown) → RUNNING — so a non-success load is never recorded as FINISHED.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
```diff
@@ -229,9 +237,54 @@ public String pushProfile(ProfilingExecPlan plan, RuntimeProfile profile) {
...
+        return detail;
+    }
+
+    // Map the QUERY_STATE string recorded by the various load paths to a QueryMemState. Spellings differ:
+    // stream/routine load use "Finished"/"Aborted", broker load "Finished"/"Running", and merge commit the
+    // upper-case TaskState name ("FINISHED"/"ABORTED"/"CANCELLED"/...). Anything not recognized as a terminal
+    // success/cancel/abort is treated as RUNNING so a non-success load is never recorded as FINISHED.
+    private static QueryDetail.QueryMemState toLoadState(String state) {
+        if ("Finished".equalsIgnoreCase(state)) {
+            return QueryDetail.QueryMemState.FINISHED;
```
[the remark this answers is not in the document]
luohaha: You're right. `LoadLoadingTask.buildFinishedTopLevelProfile()` records `Query State = Finished` from its `finally` block whenever the task stops running, even if `actualExecute()` threw — so a failed broker load carries a profile-only `Finished` marker. At `ProfileManager.pushProfile` there's no coordinator/load result available, only that unreliable string, so the state can't be determined correctly at this layer for every load path.

Rather than risk recording failed loads as successful in the durable log, I removed the derived `state` field entirely (and the `toLoadState` helper) in 131f00f. The QueryDetail JSON still carries queryId, isQuery=false, timing, db, sql, user, warehouse and the full profile text — and the authoritative state remains inside that embedded profile under "Query State". This avoids promoting an unreliable marker into a structured, machine-parsed field.

CelerData-Reviewer:
@codex review

banmoy (approved):

fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
```diff
@@ -229,9 +237,43 @@ public String pushProfile(ProfilingExecPlan plan, RuntimeProfile profile) {
...
+            detail.setLatency(element.endTimeMs - element.startTimeMs);
+        }
+        detail.setDatabase(info.get(DEFAULT_DB));
+        detail.setSql(info.get(SQL_STATEMENT));
+        detail.setUser(info.get(USER));
+        if (info.get(WAREHOUSE_CNGROUP) != null) {
+            detail.setWarehouse(info.get(WAREHOUSE_CNGROUP));
+        }
+        detail.setProfile(profileString);
+        return detail;
```
gengjun-git, line 269: You should set the state param, even though is not exactly correct. Because the fe.profile.log is subscribed by outer system, the format should be consistent.

fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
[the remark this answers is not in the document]
luohaha: Update: keeping the `state` field after all (e63ce31). Downstream consumers of the profile log (BYOC / manager) expect it, so dropping it would be a compatibility break. It still maps the profile's own QUERY_STATE the same way SHOW PROFILELIST presents it — Finished→FINISHED, Cancelled→CANCELLED, Aborted→FAILED, everything else→RUNNING. The broker-load caveat you raised is now documented in `toLoadState`: `LoadLoadingTask` writes "Finished" from its finally block even on failure, so such a load is reported FINISHED here too. That's an upstream imprecision in the profile itself (also visible via SHOW PROFILELIST), and fixing it accurately belongs in the load task that builds the profile rather than in this log serializer — I'd rather not special-case/guess at this layer. Can scope that as a follow-up if desired.

fe/fe-core/src/main/java/com/starrocks/common/util/ProfileManager.java
luohaha: done

CelerData-Reviewer:
@codex review

gengjun-git (approved):

HangyuanLiu (approved):

PR #74158, opened 2026-06-02
[Refactor] Remove lock_manager_enabled config and hardwire LockManager

gengjun-git:
## Why I'm doing:

`lock_manager_enabled` has defaulted to `true` and is non-mutable, so `LockManager` is already the only lock implementation used in production. The legacy `Database`-level `rwLock` path that it guarded is dead code that nobody maintains. Keeping the dual paths adds noise and confuses readers (e.g. `@Deprecated` "before the new LockManager is fully enabled" comments that no longer apply).

## What I'm doing:

Remove the `lock_manager_enabled` config and the now-dead legacy lock path:

- Drop `Config.lock_manager_enabled`.
- Collapse every `lock_manager_enabled` branch in `Locker` to the `LockManager` implementation and drop the `@Deprecated` legacy db-lock fallbacks. Method signatures are unchanged (heavily-used public API).
- Delete `LockUtils` (only used by the legacy `Locker` branches).
- Remove `Database.rwLock` / `slowLockLogStats` and their accessors. Neither field is `@SerializedName`, so there is no persistence/serialization impact.
- Collapse `SysFeLocks.listLocks` to the `LockManager` branch and delete `resolveLockInfo`.
- Remove `LockChecker.checkSlowLocks` (it scanned the legacy db `rwLock`, which is never held under `LockManager`); keep `checkDeadlocks` (JVM-level, independent of the flag) and `getLockWaiterInfoJsonArray`.
- Keep `QueryableReentrantReadWriteLock` — still used by `RoutineLoadJob` and `BinlogManager` for their own internal locks.
- Update `fe_locks` docs (en/zh/ja) to drop the `lock_manager_enabled` conditional wording.

Tests are migrated off the removed flag. The routine-load finish test (`GlobalTransactionMgrTest#testCommitRoutineLoadTransaction`) previously set `lock_manager_enabled=false` to skip the intention-lock path because it used the same id (`1`) for both the db and the table; under `LockManager` that is an illegal `WRITE` request on a rid already holding an `INTENTION_EXCLUSIVE` lock. It now uses a table id distinct from the db id so the hierarchical lock (IX on db + WRITE on table) is valid.

Upgrade/downgrade compatibility: a stale `lock_manager_enabled = false` left in `fe.conf` after upgrade is silently ignored (unknown config keys are not applied), and FE keeps running on `LockManager` — the same behavior as today's default. There is no observable runtime behavior change for any deployment that did not explicitly disable the lock manager.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [x] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

HangyuanLiu (approved):

wangsimo0 (approved):

gengjun-git:
@mergify backport branch-4.1

PR #74165, opened 2026-06-02
[BugFix] Do not throw exception in RewriteMultiDistinctRule when cbo_cte_reuse=false

kaijianding:
## Why I'm doing:

`java.sql.SQLSyntaxErrorException: [count(distinct 732: spu_id), count(distinct if(733: spu_status = 1, 732: spu_id, null))] is unsupported when cbo_cte_reuse is disabled`

## What I'm doing:

`return false` instead of throw exception.

Sql can actually run and get correct result, be/cn can handle like: multi_distinct(id1, id2)

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteMultiDistinctRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteMultiDistinctRule.java
@@ -22,8 +22,6 @@ import com.starrocks.catalog.Function;
 import com.starrocks.catalog.FunctionSet;
 import com.starrocks.common.FeConstants;
 import com.starrocks.sql.ast.expression.ExprUtils;
-import com.starrocks.sql.common.ErrorType;
-import com.starrocks.sql.common.StarRocksPlannerException;
 import com.starrocks.sql.optimizer.OptExpression;
 import com.starrocks.sql.optimizer.OptimizerContext;
 import com.starrocks.sql.optimizer.Utils;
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteMultiDistinctRule.java
```diff
@@ -185,8 +178,7 @@ private boolean useCteToRewrite(OptExpression input, OptimizerContext context) {
         }
 
         if (!context.getSessionVariable().isCboCteReuse() && !canRewriteByMultiFunc) {
-            throw new StarRocksPlannerException(ErrorType.USER_ERROR,
-                    "%s is unsupported when cbo_cte_reuse is disabled", distinctAggOperatorList);
+            return false;
```
stephen-shelby: I think the second branch of this change is problematic.

In `useCteToRewrite`, returning `false` means "use the MultiFunc rewriter". The new line:

```java
if (!isCboCteReuse() && !canRewriteByMultiFunc) {
    return false;
}
```

returns `false` (→ MultiFunc) exactly when `canRewriteByMultiFunc == false` — i.e. it routes to the MultiFunc rewriter the very cases (`group_concat` / json / complex / `array_agg(decimal)`) that were just determined *not* to be rewritable by multi-distinct functions. That's self-contradictory.

Concretely, with `cbo_cte_reuse = false`:

- `select count(distinct k1), group_concat(distinct k4) from baseall` → succeeds (the convertible `count` becomes `multi_distinct_count`, leaving a single `group_concat(distinct)`).
- `select group_concat(distinct k1), group_concat(distinct k4) from baseall` → **throws** `The query contains multiple distinct agg functions, each can't have multi columns.`

So for the case of multiple non-convertible distincts, this PR doesn't make the query run — it replaces the original, clear `"unsupported when cbo_cte_reuse is disabled"` error with one that is both more obscure and inaccurate (each distinct above is a **single** column, so "multi columns" doesn't describe the actual situation).

The first change (the multi-column-distinct branch) looks fine; this is specifically about the second branch.

stephen-shelby (changes requested):

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteMultiDistinctRule.java
kaijianding: ok, the second change is reverted.

PR #74167, opened 2026-06-02
[Enhancement] Avoid group by skew optimization when high distinct NDV

OliLay:
## Why I'm doing:

Currently, when `enableDistinctColumnBucketization` is true and all group by columns are skewed (or we have low cardinality), we rewrite the plan from
- (a) a 3-stage-agg: shuffle (groups), distinct global agg, global agg
- (b) a 4-stage-agg: shuffle (groups, distinct col), distinct global, local agg, shuffle (groups), local agg

to a 4-stage-agg: shuffle (groups, salting col), agg, agg, shuffle (groups)

In case (b), if `distinct col` already has a high NDV the original 4-stage-agg plan provides good distribution. Adding the salt rewrite here can even hurt the query performance because we have to compute the hash, it adds an extra agg and shuffle without making the distribution better (+ `multi_distinct_count` has to consider another dedup state due to the salting). We had a case where the rewrite caused 3x peak memory and almost 4x query runtime compared to the normal 4-stage-agg plan.

## What I'm doing:

Therefore, I am adding a guard in `GroupByCountDistinctDataSkewEliminateRule` that prevents the rewrite from application if we have a high NDV distinct column and the 3-stage-agg is _not_ an option (hence we have to decide between the 4-stage-agg and the 4-stage-agg with salting). In this case we rather want to apply the normal 4-stage-agg without salting.
 
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/GroupByCountDistinctDataSkewEliminateRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/GroupByCountDistinctDataSkewEliminateRule.java
@@ -69,6 +69,39 @@ public class GroupByCountDistinctDataSkewEliminateRule extends TransformationRul
         return INSTANCE;
     }
 
+    // Returns true when `SplitMultiPhaseAggRule` will pick the 4-stage shuffle-by-(g, d) shape
+    // as the sole alternative and the distinct column already has more distinct values than
+    // the number of buckets. In that case the (g, d) shuffle on its own distributes at least
+    // as well as the salt rewrite, so the rewrite is pure overhead.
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

trueeyu (approved):

kevincai (approved):

martinr0x:

PR #74175, opened 2026-06-02
[BugFix] Fix count on iceberg tables for rewriting simple agg to hdfs…

### LINKED ISSUE #74174: [Bug] 4.1.1 select count(*) fails on large iceberg v3 tables fails when enable_rewrite_simple_agg_to_hdfs_scan=true
1. SET enable_rewrite_simple_agg_to_hdfs_scan = true
2. Run select count(*) from large_iceberg_table;

you will get a 
```
ERROR 1064 (HY000): Internal error. Detail: deserialize chunk data failed. column slot id: 22, column row count: 8, expected row count: 1. There is probably a bug here.
```

The count returns

### Real behavior (Required)
an error is thrown

### Suspected root cause
The count(*) rewrite declares the synthetic ___count___ placeholder with a NULL type, but the BE count optimization fills it as BIGINT:

FE: RewriteSimpleAggToHDFSScanRule → new Column("___count___", NullType.NULL) → slot type resolves to BOOLEAN (1 byte).
BE: HdfsScannerContext::append_or_update_count_column_to_chunk fills it as BIGINT (8 bytes).
When the plan ships ___count___ across a GATHER exchange un-aggregated (1-phase agg, chosen for large scans), the receiver builds the column from the slot type (BOOLEAN) and decodes the 8-byte BIGINT value as 8 one-byte BOOLEAN rows → row-count mismatch.

Note: I am not sure if this error is specific to iceberg v3 of 4.1.1 or we just happen to encounter it now. It is possible that it only happens when the plan creates an exchange for the count and also hits this rewrite rule. 

 - You can get the StarRocks version by executing SQL `select current_version()`
`4.1.1`

kyle-goodale-klaviyo:
We encountered a bug where counts on very large iceberg tables would fail. More details in the issue. 

## Why I'm doing:
Fixing the issue.

## What I'm doing:
Changing the slot type for the ___count___ value returned from the RewriteSimpleAggToHDFSScanRule from Null to BigINT so that it aligns with the actual type.

Fixes #74174

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteSimpleAggToHDFSScanRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteSimpleAggToHDFSScanRule.java
@@ -123,7 +123,7 @@ public class RewriteSimpleAggToHDFSScanRule extends TransformationRule {
             // generate a placeholder column for scan node.
             // ___count___ must be the column name for backend code.
             String metaColumnName = "___" + aggCall.getFnName() + "___";
-            Column c = new Column(metaColumnName, NullType.NULL);
+            Column c = new Column(metaColumnName, IntegerType.BIGINT);
             c.setIsAllowNull(true);
             ColumnRefOperator placeholderColumn =
                     columnRefFactory.create(metaColumnName, aggCall.getType(), aggCall.isNullable());


CelerData-Reviewer:
@codex review

dirtysalt (approved):
LGTM! 🚀

silverbullet233 (approved):

PR #74180, opened 2026-06-02
[Refactor] Centralize L2/L3 cache-size fallbacks in CpuInfo and size agg two-level by L3

eshishki:
## Why I'm doing

When the CPU does not report a cache size, BE falls back to a fixed default —
1 MiB for L2, 32 MiB for L3. That fallback is copy-pasted in four places, each
re-implementing the same guard:

- runtime filter — bitset vs bloom choice
- adaptive partition hash join — sizing
- join hash-map selector — range-direct mapping
- aggregation — `two_level_memory_threshold`, the point at which a single-level
  hash table is converted to two-level

These defaults should sit in one place, the `CpuInfo` getters. They could not,
because `get_l3_cache_size()` had no default and could return 0 or a negative
`sysconf` value, so each caller added its own copy.

The aggregation one is also the only change in behavior here: it was a plain
32 MiB constant (`33554432`) that never read the CPU. On a server with an L3
larger than 32 MiB it switched to two-level too early — while the table still
fit in L3. Sized to the real L3, it now waits.

## What I'm doing

1. Harden the getters: add the L3 default, reject a negative `sysconf` result,
   and have `get_l3_cache_size()` return L3, else L2, else the default — always
   positive, so callers stop re-guarding it. (A CPU with no shared L3 reports a
   real L2, which is its last-level cache.)
2. Point the three duplicated `DEFAULT_L2/L3_CACHE_SIZE` fallbacks at the getters.
3. Size the aggregation `two_level_memory_threshold` to the detected L3 instead
   of the fixed 32 MiB. Debug builds keep the tiny value so tests still hit the
   two-level path.

No new config. Aside from that threshold, behavior is unchanged on any host that
reports an L3.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/common/system/cpu_info.h
+++ b/be/src/common/system/cpu_info.h
@@ -89,13 +89,19 @@ public:
 
     static long get_l2_cache_size() {
         const auto& cache_sizes = CpuInfo::get_cache_sizes();
-        return cache_sizes[CpuInfo::L2_CACHE] ? cache_sizes[CpuInfo::L2_CACHE] : DEFAULT_L2_CACHE_SIZE;
+        return cache_sizes[CpuInfo::L2_CACHE] > 0 ? cache_sizes[CpuInfo::L2_CACHE] : DEFAULT_L2_CACHE_SIZE;
     }
 
+    // Returns the size of the last-level cache. A CPU with no shared L3 (e.g.
+    // ARM Cortex-A, some small x86) still reports a real L2, which is then the
+    // last-level cache, so fall back to it; fall back to a fixed default only
+    // when neither level is reported (cache info unreadable). Always positive,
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

alvin-phoenix-ai:
@codex review

alvin-phoenix-ai (approved):

stdpain (approved):

kevincai (approved):

PR #74207, opened 2026-06-03
[BugFix] Add decorrelated jitter to lake vacuum retry backoff (backport #74108)

starrocks-xupeng:
## Why I'm doing:

Cherry-pick of #74108 to `branch-3.5`. The mergify auto-backport PR
(#74206) failed because the test file diverged between `main` and
`branch-3.5`, so I cherry-picked manually and only kept the test
content that PR #74108 actually added (`test_calculate_retry_delay_jitter`
+ the forward declaration); none of the other unrelated tests that
exist only on `main` are included.

`calculate_retry_delay()` in `be/src/storage/lake/vacuum.cpp` currently
returns a fully deterministic exponential backoff
(`min_delay * 2^attempted_retries`). When multiple CNs hit the same S3
throttling (`Please reduce your request rate`) at the same time, they
all back off by the same amount and retry in lockstep, which keeps the
prefix-level QPS high and prolongs the throttled state.

Decorrelated jitter spreads retries across a wider, ever-growing
window, and each CN's window diverges from the others over successive
attempts.

## What I'm doing:

- Replace deterministic backoff with **decorrelated jitter** (AWS
  Architecture Blog "Exponential Backoff And Jitter"):
  `next_delay = min(cap, rand([base, last_delay * 3]))`, where
  `cap = base * 2^max_retries`.
- `calculate_retry_delay` is now a pure helper taking
  `(last_delay, base, max_retries)`; the two config values
  (`lake_vacuum_retry_min_delay_ms`, `lake_vacuum_retry_max_attempts`)
  are read once at the top of `delete_files_with_retry` and passed
  down.
- Returned delay is always in `[base, cap]` so the configured minimum
  delay is preserved as a hard floor.
- Use `butil::fast_rand_less_than` (brpc thread-local PRNG) to avoid
  `std::rand` global lock contention.
- Add `LakeVacuumTest2.test_calculate_retry_delay_jitter` covering
  first-call window, cap clamping, and a 20-step simulated retry chain.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/vacuum.cpp
+++ b/be/src/storage/lake/vacuum.cpp
@@ -14,6 +14,7 @@
 
 #include "storage/lake/vacuum.h"
 
+#include <butil/fast_rand.h>
 #include <butil/time.h>
 #include <bvar/bvar.h>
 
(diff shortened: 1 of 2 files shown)


kevincai (approved):

trueeyu:

PR #74218, opened 2026-06-03
[BugFix] Non-root compound predicates yield NotPushDown instead of EOF

satanson:
## Why I'm doing:

Fixes a wrong-result issue in predicate pushdown for compound predicates under `UNION`.

 When a pushed-down predicate contains an impossible nested branch, for example:

```sql
  (measures = 'current' OR measures = 'cumulated')
  OR
  (measures = 'current' AND measures = 'cumulated')
```
The nested AND branch is always false, but the whole OR predicate is still valid. Previously, BE predicate normalization could convert the nested false branch into EndOfFile, then propagate that as scan-level EOF, causing OlapScanNode to emit no rows incorrectly.

**Root Cause**
ChunkPredicateBuilder::build_olap_filters() returned Status::EndOfFile for empty/full value ranges regardless of whether the builder represented the root scan predicate or a nested compound predicate.

For nested compound predicates, EOF should not mean “the whole scan is empty.” It should only mean this compound predicate cannot be safely pushed down.

## What I'm doing:

  For EOF-producing branches in build_olap_filters():

  - If _is_root_builder is true, keep returning Status::EndOfFile.
  - If _is_root_builder is false, return a local not-pushdown status instead.
  - In _normalize_compound_predicate(), handle the not-pushdown status by returning false, so the compound predicate is left for normal expression evaluation instead of being pushed to
    storage.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exec/olap_scan_prepare.cpp
+++ b/be/src/exec/olap_scan_prepare.cpp
@@ -191,6 +191,16 @@ Status prune_scan_ranges_by_partition_conjuncts(RuntimeState* state, const Tuple
 // Util methods.
 // ------------------------------------------------------------------------------------
 
+static constexpr std::string_view kNotPushDownPredicateStatus = "not push down predicate";
+
+static Status not_push_down_predicate_status() {
+    return Status::NotSupported(kNotPushDownPredicateStatus);
+}
+
+static bool is_not_push_down_predicate_status(const Status& status) {
+    return status.is_not_supported() && status.message() == kNotPushDownPredicateStatus;
+}
+
 static bool ignore_cast(const SlotDescriptor& slot, const Expr& expr) {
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

ZiheLiu (approved):

stdpain:
@satanson could u add a SQLTest ?

trueeyu (approved):

PR #74220, opened 2026-06-03
[Feature] Support max_pt function in odps catalog

adios666:
## Why I'm doing:
MaxCompute(ODPS) supports 'max_pt' function to get max non-empty partition of odps table in lexicographical order. Users migrated from ODPS get used to use 'max_pt' when querying ODPS table. It is necessary to be compatible SQL usage with ODPS.
Getting max partition is an expensive operation. Including get all partitions, sort partitions lexicographically by top-k. Considering most ODPS tables are t+1 updating, 'max_pt' value can be cached like `partitionCache`.

## What I'm doing:
Implement 'max_pt' meta function and max partition cache in `OdpsMetadata`. Partitions loaded from `partitionCache` will be sorted lexicographically. Empty partition is filtered. Values of max partitions are saved in `maxPartitionCache`, which is similar to `partitionCache`. `maxPartitionCache` is updated together with `partitionCache`.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorMetadata.java
@@ -236,6 +236,10 @@ public interface ConnectorMetadata {
         return Lists.newArrayList();
     }
 
+    default String getMaxPartitionValue(Table table, boolean nonEmptyPartition) {
+        return null;
+    }
+
     /**
      * Get partition info at a specific snapshot identified by the request context.
      * Default implementation ignores the context and falls back to the live-snapshot variant.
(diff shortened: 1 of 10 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/connector/ConnectorMetadata.java
```diff
@@ -236,6 +236,10 @@ default List<PartitionInfo> getPartitions(Table table, List<String> partitionNam
         return Lists.newArrayList();
     }
 
+    default String getMaxPartitionValue(Table table, boolean nonEmptyPartition) {
```
stephen-shelby, line 239: The cross-connector ConnectorMetadata.getMaxPartitionValue isn't needed — max_pt is ODPS-only, so resolve the leaf metadata via DelegatingConnectorMetadata.delegateFor() and instanceof OdpsMetadata (same pattern as MVPCTMetaRepairer.resolveLeafMetadata + instanceof HiveMetadata). That drops the changes to ConnectorMetadata, CatalogConnectorMetadata, and MetadataMgr. Also: the boolean nonEmptyPartition arg is dead (only true is ever passed); the unconditional partitions.get(0) in the refresh paths will throw for non-partitioned/empty tables; and maxPartitionCache is redundant with partitionCache — computing the max on demand removes the 3 duplicated, hand-synced code paths.

PR #74226, opened 2026-06-03
[BugFix] Fix hudi slice reader for varchar type (backport #58521)

HangyuanLiu:
## Why I'm doing:

Backport of #58521 to `branch-3.5`.

When `hudi_mor_force_jni_reader = true`, querying a HUDI MOR table whose schema contains `char`/`varchar` columns (including `char`/`varchar` nested inside `struct`/`map`/`array`) fails on `branch-3.5`:

```
java.io.IOException: Failed to open the hudi MOR slice reader.
    at com.starrocks.hudi.reader.HudiSliceScanner.open(HudiSliceScanner.java:221)
Caused by: java.lang.IllegalArgumentException: char type is specified without length:
    char,char,...,map<char,array<int>>,struct<...,b:char>>
    at org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils$TypeInfoParser.parseType
```

Root cause: in `ColumnType`, the reverse map `PRIMITIVE_TYPE_VALUE_STRING_MAPPING` (TypeValue -> hive type string) is built by iterating `PRIMITIVE_TYPE_VALUE_MAPPING`. Since `string`, `varchar` and `char` all map to `TypeValue.STRING`, inserting `varchar`/`char` *before* the reverse map is generated lets the reverse entry for `STRING` be overwritten to `"char"`. `HudiScannerUtils.mapColumnTypeToHiveType()` then serializes every STRING column as a length-less `char`, which Hive's `TypeInfoParser` rejects.

The issue was first fixed in #34397, regressed by #58292 (paimon time type) which reordered the static initializer, and re-fixed on `main` by #58521. `branch-3.5` carries the #58292 regression but is missing #58521, making it the only release branch currently affected.

## What I'm doing:

Move the `varchar`/`char` -> `STRING` mapping to *after* `PRIMITIVE_TYPE_VALUE_STRING_MAPPING` is generated, so `STRING` reverse-maps to `"string"` and char columns are handed to the Hive SerDe as `string`. Includes the `TestHudiSliceScanner` regression test from the original PR.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [x] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

> Note: target base is already `branch-3.5`, so the 3.5 box is intentionally left unchecked. 3.4/3.3 are not affected yet (the #58292 regression backports #58339/#58338 are still open), so this fix is not auto-backported there; it must accompany #58292 if/when those land.

### DIFF
--- a/java-extensions/hudi-reader/src/test/java/com/starrocks/hudi/reader/TestHudiSliceScanner.java
+++ b/java-extensions/hudi-reader/src/test/java/com/starrocks/hudi/reader/TestHudiSliceScanner.java
@@ -1,0 +1,332 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.hudi.reader;
+
+import com.starrocks.jni.connector.OffHeapTable;
+import com.starrocks.utils.Platform;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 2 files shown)


gengjun-git (approved):

tiannan-sr:
@Mergifyio backport branch-3.5.18

trueeyu:

PR #74264, opened 2026-06-04
[Enhancement] Record eligibility_skipped on single-partition pre-split resolve failure

xiangguangyxg:
## Why I'm doing:

The single-partition Sample-Based Tablet Pre-Split hook flow (INSERT-from-FILES and Broker Load) short-circuited **silently** when `PreSplitTargets.findEligibleTarget` could not resolve a single-physical-partition / single-base-tablet target. The `eligibility_skipped{reason}` counter is only emitted by `TabletPreSplitCoordinator.maybeAct`, which the single-partition flow never reaches on a resolve failure.

As a result, the most common operational case — a re-load against a partition that was already split (so it now has more than one base tablet) — produced **no metric at all**, so any operator dashboard built on `eligibility_skipped` under-counted on the multi-tablet branch. This was surfaced concretely by the post-merge regression run (a second load on an already-pre-split partition emitted no bvar despite correctly skipping pre-split).

## What I'm doing:

`PreSplitTargets.findEligibleTarget` now records the failing `SkipReason` into the **existing** `eligibility_skipped{reason}` counter before returning `null`:

- `metadata_not_resolved` — the table does not resolve to a single physical partition, or its base index is missing (e.g. an alter raced the load);
- `multiple_base_index_tablets` — the partition has zero or multiple base-index tablets (the already-split re-load case).

The recording is **centralized in the resolver, not the hooks**. Its only callers are the two single-partition hook flows, which all skip on a `null` target, so neither hook can forget the metric — and both hook bodies stay byte-identical to `main` (the whole change lives in `PreSplitTargets`). This mirrors `TabletPreSplitCoordinator.skipEligibility`, which also records internally. (The sibling table-level gate `findEligibleTable` stays pure because its multi-partition caller needs the reason for its own `Skipped` outcome and records it itself.)

- **No new counter and no new `SkipReason` values** — both labels are already valid and documented for `eligibility_skipped`; this just makes them reachable from the single-partition path. `target-non-empty` was already covered (a single non-empty tablet reaches `maybeAct` → `partition_not_empty`).
- The base-index/tablet resolution is inlined into `findEligibleTarget` (resolving the index once and naming each failure at its check), removing the now-unused `findSingleBaseTabletId` helper.

Tests: `PreSplitTargetsTest` covers every `findEligibleTarget` branch (no-unique-partition, missing base index, multiple tablets, zero tablets, and the eligible happy path), counter-asserting the recorded reason; `BrokerLoadPreSplitHookTest`'s single-partition resolve tests were strengthened to pass the table-level gate (so they genuinely reach `findEligibleTarget`) and assert the `eligibility_skipped` counter increments exactly once with no delegation. All pre-split FE suites green; checkstyle clean.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/PreSplitTargets.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/PreSplitTargets.java
@@ -32,22 +32,27 @@ import java.util.List;
  * that contract in one place so every hook (and any future caller) checks the
  * same shape.
  *
- * <p>Each helper returns a sentinel ({@code null} or {@code -1}) for "no
- * match" so callers can chain early returns without intermediate booleans.
- * The eligibility gate inside {@link TabletPreSplitCoordinator#maybeAct}
+ * <p>The eligibility gate inside {@link TabletPreSplitCoordinator#maybeAct}
  * re-runs equivalent checks against the resolved {@code physicalPartitionId};
- * the helpers here only let the hook skip the pipeline construction.
+ * the helpers here only let the hook skip the pipeline construction. Because
+ * the single-partition hook flow short-circuits before {@code maybeAct}, the
+ * skip must still be attributed to a {@link SkipReason} so operators can see
+ * why pre-split did not run.
  *
  * <p><b>Two slices</b>:
  * <ul>
  *   <li>{@link #findEligibleTable} — table-only structural gate (range
  *       distribution, NORMAL state, no MV/rollup, supported sort key) used
- *       by the multi-partition coordinator's defensive re-check. The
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

luohaha (approved):

HangyuanLiu (approved):

PR #74330, opened 2026-06-04
[BugFix] support meta scan for schema change (backport #72901)

Wenjun7J:
## Why I'm doing:

Backport https://github.com/StarRocks/starrocks/pull/72901 to branch-3.5.

After some schema change like add column, the new added column could not be found in the old segment files, so the meta scan will throw a not found error now. This may cause background statistics analyze failed.

## What I'm doing:

The first four schema changes are relatively straightforward:

- DROP COLUMN: the new schema no longer reads this column, so the old physical column in existing segments can be ignored.
- RENAME COLUMN: only the name changes; the column uid stays the same, so the reader should be located by uid.
- REORDER COLUMN: the ordinal/cid changes, but the column uid stays the same, so locating the reader by uid is sufficient.
- ADD COLUMN: old segments do not have the new physical column, so the meta reader can synthesize the default value or NULL.

The remaining cases are harder:

1. ALTER COLUMN TYPE min/max values come from the segment zone map, and those values are encoded using the old physical type. We would need explicit type conversion logic.
2. Complex/derived/access-path changes, such as flat JSON paths, extended columns, or nested type layout changes.

These 2 types enhancement could be processed in independent prs in future.

Backport notes:

- Cherry-picked from https://github.com/StarRocks/starrocks/commit/149e174b5ae1bd7e9f1d34517cc02b8f6ef8e8ce.
- Resolved conflicts in `be/src/storage/meta_reader.cpp` and `be/src/storage/meta_reader.h`.
- Kept the branch-3.5 collect path and did not backport the main-only virtual-column collect path.
- Omitted `column_size` and `column_compressed_size` test coverage because branch-3.5 does not have those meta functions.

Validation:

- `git diff --check HEAD~1..HEAD`
- `cmake --build be/ut_build_Debug --target Storage -j 8` on branch-4.1

Fixes https://github.com/StarRocks/starrocks/pull/72901

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/meta_reader.cpp
+++ b/be/src/storage/meta_reader.cpp
@@ -243,7 +243,7 @@ StatusOr<ColumnReader*> SegmentMetaCollecter::_get_column_reader(ColumnId cid) c
 }
 
 bool SegmentMetaCollecter::_is_missing_default_column(const TabletColumn& column) const {
-    return !column.is_extended() && _segment->is_default_column(column);
+    return _segment->is_default_column(column);
 }
 
 Status SegmentMetaCollecter::_init_return_column_iterators() {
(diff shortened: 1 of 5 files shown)


Youngwb (approved):

trueeyu:

PR #74341, opened 2026-06-04
[Enhancement] Add journal_replay_lag metric for leader-follower replay skew

gengjun-git:
## Why I'm doing:

There is currently no metric that exposes how far each Follower/Observer FE lags behind the Leader in replaying the BDBJE edit-log journal. Operators can only read `ReplayedJournalId` from `SHOW FRONTENDS` and subtract by hand, and the existing `max_journal_id` gauge is a plain per-node gauge (no way to reliably pick "the Leader's authoritative value" from it), so building a journal-replay skew alert is not possible from existing metrics.

## What I'm doing:

Add a new **Leader-only** FE gauge metric `starrocks_fe_journal_replay_lag`. On the Leader it emits one series per **alive non-Leader** frontend, labeled by `host`, with value `leader_max_journal_id - replayed_journal_id` (clamped to 0). Followers/Observers emit no series for this metric, so it is unambiguous and needs no `is_leader` label.

Implementation notes:
- The computation lives in `MetricRepo`, collected during `getMetric()` (alongside the other dynamic gauges). The per-frontend build is factored into a `@VisibleForTesting` pure helper so it is unit-testable without reflection.
- Existing metrics (`max_journal_id`, `meta_log_count`, …) are **not** changed.
- `Frontend.replayedJournalId` is made `volatile`: it is written by the heartbeat thread and now also read by metric collection, so a non-volatile 64-bit read would be unsafe under the Java memory model.
- Added a unit test covering Leader/non-Leader, dead-frontend skip, clamp-to-0, and Prometheus output format.
- Documentation added for `en`, `zh`, and `ja`.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4


docs/en/administration/management/monitoring/metric_details/i-p.md
```diff
@@ -235,17 +200,12 @@ For more information on how to build a monitoring service for your StarRocks clu
...
 - Description: Memory used by jit compiled function cache.
 
-## `lake_vacuum_del_file_batch_size_minute`
-
-- Unit: Count (files per batch)
-- Type: Gauge
-- Description: Recent (60s) mean of files per `DeleteObjects` batch sent by Vacuum on shared-data clusters.
-
-## `lake_vacuum_del_file_retries_minute`
+## `starrocks_fe_journal_replay_lag`
```
kevincai, line 203: I guess with this in hand, the metric added in this PR, may not be necessary any more. https://github.com/StarRocks/starrocks/pull/73833

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
kevincai, line 1509: the `isLeader` is always `true` in non-testing code path, then it is not necessary.

kevincai, line 1519: or, in order to report the whole set of frontends, a label `'alive={true|false}'` can be provided to distinguish the live ones and dead ones. The metrics can be calculated based on label filtering. So from the producer, it creates all the necessary metrics, and from the consumer side (grafana/prometheus), it can decide whether to include those dead ones.

kevincai, line 1525: what's the behavior of same metric name with different labels registered as separate Metric? will the output (both plaintext and json format) be compatible with prometheus? I am not quite sure.

fe/fe-core/src/main/java/com/starrocks/system/Frontend.java
kevincai, line 71: not necessary to be `volatile` at all, metric collection can be tolerant with the inaccuracy.

docs/en/administration/management/monitoring/metric_details/i-p.md
gengjun-git: OK, wait this pr to merge. https://github.com/StarRocks/starrocks/pull/73833

gengjun-git:
duplicate with https://github.com/StarRocks/starrocks/pull/73833

PR #74344, opened 2026-06-04
[BugFix] Reject partial update targeting a same-transaction-modified table in explicit txn

meegoo:
## Why I'm doing:

In an explicit multi-statement transaction (`BEGIN ... COMMIT`) on a shared-data primary-key table, all statements are batch-applied together as multiple `op_write`s on one tablet at COMMIT. A **partial-update INSERT** writes only a subset of columns and implicitly **reads the target table at publish time** to fill the columns it does not write.

When that table was **already modified earlier in the same transaction**, the batch publish path cannot correctly merge the partial update against the uncommitted same-transaction writes and **silently corrupts data**:

- A column-upsert that inserts new keys calls `apply_opwrite` mid-batch, which bumps `next_rowset_id` and upserts new keys at a non-batch-offset rssid. This desyncs the primary index from the pending rowset's final id (`set_final_rowset`), so an earlier statement's rows resolve to the wrong rowset and read another row's data.
- Two or more row-mode partial updates collide on `replace_segments` keys: the key is a per-`op_write` *local* segment id and `add_rowset` merges it verbatim (without the `assigned_segment_idx` offset that `segment_metas` get), so `set_final_rowset` applies a rewritten segment onto the wrong segment position.

This matches the existing transaction guardrails that already forbid `SELECT` reading a same-transaction-modified table (5307) and restrict `UPDATE`/`DELETE` ordering (5306). Partial-update INSERT was the remaining gap.

## What I'm doing:

Reject it at the FE during explicit-transaction validation. `ExplicitTxnStatementValidator` now raises a new error **`5308 ERR_EXPLICIT_TXN_PARTIAL_UPDATE_ON_MODIFIED_TABLE`** when a partial-update INSERT targets a table already in the transaction's `modifiedTableIds`.

Still allowed: a single partial update in a transaction, partial updates on different tables, and plain full inserts (the supported multi-statement case). Adds an FE unit test (`ExplicitTxnTest.testPartialUpdateOnModifiedTableRejected`) and a SQL-Tester suite (`test/sql/test_partial_update_multi_stmt_txn`).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

banmoy:
This limitation is not obvious. Should add documents for it.

CelerData-Reviewer:
@codex review

banmoy (approved):

kevincai (approved):

HangyuanLiu (approved):

PR #74356, opened 2026-06-05
[DO NOT MERGE] repro: fix #73962 + lottery (random_device seed, matches #74001)

luohaha:
Same as #73985 but with random_device seed aligned to PR #74001 for a strictly clean comparison.


luohaha:
repro work done — closing this scratch PR

PR #74371, opened 2026-06-05
[Enhancement] Add opt-in lazy refresh of external stats cache on journal replay

stephen-shelby:
## Why I'm doing:

When followers replay external-table statistics journals (`OP_ADD/REMOVE_EXTERNAL_BASIC|HISTOGRAM_STATS_META`), the cache-refresh path resolves the external table via `MetadataMgr.getTable` (`ConnectorColumnStatsCacheLoader` → `StatisticsUtils.getTableByUUID` → `MetadataMgr.getTable`). For Hive/Iceberg/etc. this calls the Hive Metastore / object storage, so a slow or stuck metastore can stall the FE journal replayer thread. The same path runs during **restart recovery** (replaying the incremental edit log), so a degraded metastore can also slow down — or block — FE startup.

Skipping the refresh entirely is not acceptable: the connector stats cache TTL is `statistic_update_interval_sec * 2` (48h by default, with `enable_statistic_cache_refresh_after_write=false`), so followers would otherwise serve stale stats for up to two days.

## What I'm doing:

Persist the table UUID (the connector stats cache key) into `ExternalBasicStatsMeta` / `ExternalHistogramStatsMeta` at analyze time on the leader, where the table is already resolved. On replay, followers **invalidate** the cache by that UUID directly — no `getTable`, no external I/O, no replayer blocking — and reload lazily on the next query. This mirrors how internal-table stats replay already invalidates by table id.

Gated behind a new mutable FE config `enable_external_stats_lazy_refresh_on_replay` (**default `false`**), so default behavior is unchanged. When enabled:

- new journals (with a persisted UUID) → invalidate by UUID (non-blocking);
- old journals (no UUID) → fall back to the legacy eager refresh.

The UUID is always persisted regardless of the flag (best-effort, never breaks analyze), so enabling the flag takes effect immediately for existing metas without a re-analyze.

Related to #73832 (an alternative to #73833 that avoids resolving external metadata on the replayer instead of offloading it to a background thread).

## Compatibility

`tableUUID` is a new optional gson field on the stats meta (journal + image). Upgrade: older journals/images deserialize it as `null` → eager-refresh fallback. Downgrade: older FEs ignore the unknown field. No required/ordinal concerns.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -2503,6 +2503,14 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true)
     public static boolean enable_statistic_cache_refresh_after_write = false;
 
+    @ConfField(mutable = true, comment = "When replaying external-table statistics journals on followers " +
```
gengjun-git, line 2506: Why not default to true, I think this enhancement has no bad effect.

stephen-shelby: this pr changes default behavior. the original behavior was that stats would be reloaded when replaying edit log. but now it just invalidates the cache. the cache is no longer reloaded.

kevincai: vote for making it `true` as default. this is preferred and recommended configuration for all clusters even though it is a behavior change.

stephen-shelby: The slowness of the table retrieval interface is only a minor issue., but if the statistics is not updated in a timely manner in the follower, and the wrong plan is selected, more serious consequences may occur, such as broadcasting large tables, which can lead to OOM (Out Of Memory) errors.

Youngwb (approved):

fe/fe-core/src/main/java/com/starrocks/common/Config.java
gengjun-git: How about using `synchronous().refresh(key)`, this will trigger refresh in background(do not block replay thread), and before the new value reached, get(key) will always return old value.

gengjun-git (approved):

fe/fe-core/src/main/java/com/starrocks/common/Config.java
stephen-shelby: This will block the loading of the internal table. For the time being, I am more inclined to keep the behavior unchanged.

PR #74379, opened 2026-06-05
[BugFix] Do not freeze CatalogRecycleBin when cluster snapshot keeps failing

xiangguangyxg:
## Why I'm doing:

In shared-data mode, when automated cluster snapshot is ON but **no snapshot job has ever reached `FINISHED`** (e.g. every image upload fails with an object-storage 403), `ClusterSnapshotMgr.getSafeDeletionTimeMs()` returned `0`. That made `isDeletionSafeToExecute()` always false, so `CatalogRecycleBin` stopped erasing **anything**. With a daily full `INSERT OVERWRITE`, overwritten partitions/tablets piled up in the recycle bin and FE memory grew without bound. Disabling cluster snapshot let the bin drain and FE memory recovered — confirming the snapshot safety gate was the cause.

## What I'm doing:

Redefine the safe-deletion boundary as the earliest consistency point that must stay restorable:
- the kept finished snapshots (historical "keep the two most-recent completed" behavior preserved byte-for-byte), and
- any in-progress job.

A job that ended in `ERROR` protects nothing, and when nothing needs protecting the method returns the current time so the recycle bin drains normally. `getSafeDeletionTimeMs()` is a single early-breaking pass (no extra hot-path cost). The AlterJob tablet path (`isTableSafeToDeleteTablet`) is fixed by the same change since it routes through `getSafeDeletionTimeMs()`.

Snapshot-health observability so a persistently failing snapshot is detectable:
- gauges `cluster_snapshot_last_finished_time`, `cluster_snapshot_consecutive_failures`
- a `WARN` logged on each failed automated snapshot (snapshots run infrequently, so this is a low-rate recurring signal)

The non-atomic "remote marker written but `FINISHED` not yet persisted" orphan window is intentionally out of scope. The general `recycle_bin_*` size gauges were moved to a separate follow-up PR per review.

Validated end-to-end on a TSP shared-data cluster: with snapshot failing every cycle and zero ever FINISHED, the recycle bin drained 23→0 (real `erasePartition` log entries), `last_finished_time`==0 throughout, and the WARN fired as designed.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5


CelerData-Reviewer:
@codex review

xiangguangyxg:
Note for reviewers / backporters (this PR is labeled for backport to 3.5 / 4.0 / 4.1):

1. **Doc `Introduced in` version**: the new config `cluster_snapshot_failure_warn_threshold` is documented as `Introduced in: v4.2` (correct for `main`). The cherry-picked backports will carry this string verbatim, so the backport PRs to branch-3.5 / 4.0 / 4.1 should adjust it to the version/patch appropriate for each branch.
2. **Backport conflicts**: `ClusterSnapshotMgr.getSafeDeletionTimeMs()` is substantially rewritten; the auto-backports may need manual conflict resolution on the older branches (please confirm the same code path exists, especially on 3.5).

The non-atomic "remote marker written but FINISHED not yet persisted" orphan window is intentionally out of scope for this fix.

xiangguangyxg:
## End-to-end validation (TSP shared-data cluster, build from this PR)

Reproduced the production scenario on a real 1 FE + 3 CN shared-data cluster (`main-8e767a8`): automated cluster snapshot pointed at a non-existent bucket so **every upload fails** (`NoSuchBucketException`, same class as the incident's 403), exactly like "no snapshot job ever reaches FINISHED".

**Result — recycle bin drains while snapshots keep failing:**

| time | recycle_bin_partition_count | consecutive_failure_count | last_finished_time |
|---|---|---|---|
| 17:06 | 23 | 8 | 0 |
| 17:07 | **0** | 9 | 0 |
| 17:08 | 0 | 10 | 0 |
| 17:09 | 0 | 11 | 0 |

- `recycle_bin_partition_count`: **23 → 0** (FE log shows 23 real `CatalogRecycleBin.erasePartition` "Removed partition" entries) — under the old code this stayed pinned forever.
- `cluster_snapshot_last_finished_time` == 0 throughout (zero successful snapshots — the exact production case).
- `cluster_snapshot_consecutive_failure_count` climbed 8→11 (snapshots failing every interval).
- New threshold WARN fired as designed: `ClusterSnapshotJobScheduler ... Automated cluster snapshot has failed N times in a row ... The recycle bin is NOT blocked, but no new restore point is being produced. Check snapshot storage volume permissions/connectivity.`
- All 5 new gauges (`recycle_bin_{partition,table,database}_count`, `cluster_snapshot_last_finished_time`, `cluster_snapshot_consecutive_failure_count`) and the `cluster_snapshot_failure_warn_threshold` config registered and behaved correctly.

Confirms the fix end-to-end: a persistently-failing cluster snapshot no longer freezes the recycle bin.

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotJobScheduler.java
```diff
@@ -110,6 +111,17 @@ protected void runAfterCatalogReady() {
             // set last start time when job has been created and begin to submit
             lastAutomatedJobStartTimeMs = runningJob.getCreatedTimeMs();
             runningJob.run(this);
+
+            if (runningJob.isError()) {
+                int failures = GlobalStateMgr.getCurrentState().getClusterSnapshotMgr().getConsecutiveFailureCount();
+                int threshold = Config.cluster_snapshot_failure_warn_threshold;
+                if (threshold > 0 && failures >= threshold) {
```
kevincai: will this counter be reset? or after reaching the threshold, this logging message will be printed for every cycle if it is not recovered.

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
```diff
@@ -585,6 +585,53 @@ public Long getValue() {
         };
         STARROCKS_METRIC_REGISTER.addMetric(metaLogCount);
 
+        GaugeMetric<Long> recycleBinPartitionCount = new GaugeMetric<Long>(
+                "recycle_bin_partition_count", MetricUnit.NOUNIT, "partition count in catalog recycle bin") {
```
kevincai: 1. name is a little bit confusing. these metrics are Gauge, not Counter, `*_count` could be confusing. Using `*_num` or `*_size` instead.

2. Worth splitting into a separate PR, those recyle_bin_* metrics can be independent to the ClusterSnapshot bugfix.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/lake/snapshot/ClusterSnapshotJobScheduler.java
xiangguangyxg: Good point. Snapshots run infrequently (default `automated_cluster_snapshot_interval_seconds`=600s), so I simplified rather than add a flag: removed the threshold config entirely and just log one WARN per failed attempt — a low-rate recurring signal (≤ a few per hour), not per-fast-cycle spam. The `cluster_snapshot_consecutive_failures` gauge carries the precise count for alerting, and the consecutive count is naturally reset to 0 once a snapshot succeeds (it's computed as trailing-ERRORs-since-last-FINISHED).

fe/fe-core/src/main/java/com/starrocks/metric/MetricRepo.java
xiangguangyxg: Both addressed:
1. Naming — renamed the gauge off `_count` (Gauge, not Counter): `cluster_snapshot_consecutive_failure_count` → `cluster_snapshot_consecutive_failures`.
2. Split — removed the `recycle_bin_*` gauges (and their `CatalogRecycleBin` accessors/tests/docs) from this PR entirely. They'll go in a separate, main-only enhancement PR named `recycle_bin_*_num` (also off `_count`). This PR now keeps only the bugfix + the snapshot-health signal intrinsic to it. I'll link the follow-up PR here.

xiangguangyxg:
Follow-up PR for the split-out `recycle_bin_*` size gauges (renamed to `*_num`): #74440

HangyuanLiu (approved):

srlch (approved):

PR #74394, opened 2026-06-05
[Enhancement] opt(lake pindex): origin-skip + parallel del-apply on cold rebuild

luohaha:
## Why I'm doing:

On shared-data cold-start, rebuilding the cloud-native PK index from a rowset's delete files (`LakePersistentIndex::load_dels`) was slow: each del file did an index `get()` whose SST reads are cold, and the files were applied serially. On cold 300-table runs `rebuild_index_del_cost` was the dominant cost (seconds).

Two observations make this avoidable:
1. A del file whose `origin_rowset_id` equals the rowset being rebuilt never needs the `get()`: the skip filter it would build is a no-op, so every key is erased anyway. The `get()` (and its cold SST reads) is pure waste for those files.
2. The only order-dependent step is the memtable `replay_erase`. Reading/decoding the del bytes, extracting keys, and running the non-origin `get()` are all parallel-safe against the same pre-erase snapshot.

## What I'm doing:

Rewrote `load_dels` into two phases:

- **`load_one` (parallel-safe):** read + decode the del file, extract keys, and — only for non-origin del files — run `get()` and build the skip filter. Origin del files skip `get()` entirely (origin-skip), avoiding the cold SST reads that dominated the cost.
- **`erase_one` (serial):** the order-dependent memtable `replay_erase`, replayed in `del_idx` order.

When the two-phase path is enabled (existing `should_parallel_rebuild_prefetch` gate: parallel execution on, `num_del_files > 1`, update mem-tracker below the rebuild ratio), Phase 1 runs every file's `load_one` concurrently so the non-origin `get()`s overlap, then Phase 2 erases serially. Otherwise it falls back to a single-pass `load_one`+`erase_one` loop holding one decoded column at a time.

This is correct even when del files share keys: `replay_erase` writes `(rowset_version, NULL)` and every del file of a rowset uses the same `rowset_version`, so the erase is idempotent and the parallel result equals the serial result.

Phase 1 reuses the existing `pk_index_execution_thread_pool` (registered as `cloud_native_pk_index_execution`). `get()` runs its SST lookups synchronously and never submits back to that pool, so there is no pool-in-pool nesting — no dedicated pool, config, or metric is added.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

This is a performance optimization of the PK-index cold rebuild path; the rebuilt index is identical to before.

## Checklist:

- [ ] This is a backport pr

Added BE unit tests in `lake_persistent_index_test.cpp`:
- `test_load_dels_non_origin_filters_stale_deletes` — non-origin del runs `get()`+filter; stale deletes (key rssid newer than `origin_rowset_id + op_offset`) are skipped, older ones erased.
- `test_load_dels_origin_erases_all_keys` — origin del erases every key regardless of rssid (origin-skip semantics), contrasting the non-origin case on identical input.
- `test_load_dels_parallel_matches_single_pass` — two key-sharing del files (one origin, one non-origin) produce an identical index via the parallel and single-pass paths.

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -945,59 +945,68 @@ Status LakePersistentIndex::load_dels(const RowsetPtr& rowset, const Schema& pke
         return pkc;
     };
 
-    // Apply one decoded del column to the index. Order-dependent: replay_erase mutates state
-    // observed by later files' get(), so callers must invoke this sequentially in del_idx order.
-    auto apply_one = [&](int del_idx, MutableColumnPtr& pkc) -> Status {
-        const auto& del = rowset->metadata().del_files(del_idx);
-        // We can't insert delete operation to index directly, because some delete operation is
-        // older than current item, and we need to igore these delete operations.
-        std::vector<IndexValue> found_values(pkc->size(), IndexValue(NullIndexValue));
-        std::vector<bool> filter(pkc->size(), false);
-        auto generate_filter_fn = [&]() {
-            if (rowset->id() != del.origin_rowset_id()) {
-                // del file in origin rowset doesn't need to skip.
-                for (int i = 0; i < pkc->size(); i++) {
-                    if (found_values[i] != IndexValue(NullIndexValue) &&
-                        found_values[i].get_rssid() > del.origin_rowset_id() + del.op_offset()) {
-                        // Use `rowset_id + op_offset` as delete file's rssid.
-                        // delete operation is too old for this key.
-                        filter[i] = true;
-                    }
-                }
-            }
-        };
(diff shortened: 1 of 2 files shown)


kevincai:
@mergifyio rebase

CelerData-Reviewer:
@codex review

kevincai (dismissed):

xiangguangyxg (dismissed):

CelerData-Reviewer:
@codex review

xiangguangyxg (approved):

kevincai (approved):

PR #74397, opened 2026-06-05
[BugFix] Fix ReduceCastRule dropping a fractional-truncating cast

### LINKED ISSUE #51545: cast bigint to string return wrong result
select cast(cast((9-1)/3+1 as bigint) as string)

return 3
### Real behavior (Required)
return 3.6666666666666665
 - 3.3.3

Hyper-FF:
## Why I'm doing:

ReduceCastRule reduced cast(cast(<floating> as <integer>) as <outer>) by removing the intermediate integer cast. That cast truncates the fractional part, so dropping it exposed the untruncated floating-point value to the outer cast. For example cast(cast((9-1)/3+1 as bigint) as string) returned "3.6666666666666665" instead of "3".

The existing slot-size guard missed this because DOUBLE and BIGINT share the same size. Skip the reduction when the grandchild is a floating-point type and the intermediate cast is not, unless the outer type is also an integer type, in which case it re-applies the truncation and the chain remains safe to reduce.

## What I'm doing:

Fixes #51545

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

https://claude.ai/code/session_01F3uo4oXbH83Lcp8vj21aBC

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/ReduceCastRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/ReduceCastRule.java
@@ -147,7 +147,6 @@ public class ReduceCastRule extends TopDownScalarOperatorRewriteRule {
     public boolean checkCastTypeReduceAble(Type parent, Type child, Type grandChild) {
         int parentSlotSize = parent.getTypeSize();
         int childSlotSize = child.getTypeSize();
-        int grandChildSlotSize = grandChild.getTypeSize();
 
         if (parent.isDecimalOfAnyVersion() || child.isDecimalOfAnyVersion() || grandChild.isDecimalOfAnyVersion()) {
             return false;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

HangyuanLiu:
Thanks for fixing #51545 — the direction is right, but I think the guard is **incomplete**: it only covers the case where the *intermediate* cast is an integer. When the middle cast is the lossy one — i.e. `child` is `float` — `!child.isFloatingPointType()` short-circuits the new condition, so the chain is still reduced.

**Why both guards miss it.** `checkCastTypeReduceAble` compares `getTypeSize()`, and `getTypeSize(FLOAT) == getTypeSize(DOUBLE) == 8`. So for `child=FLOAT, grandChild=DOUBLE`, the slot-size guard's `childSlotSize < grandChildSlotSize` is `8 < 8 == false` and never fires, for *any* `parent`. Together with the new guard short-circuiting, `cast(cast(<double> as float) as <anything>)` is still reduced and loses the double→float narrowing.

**Verified locally with this PR's guard applied** — all four reduce, dropping the middle float:

```
double->float->int     => cast(<double> as int)
double->float->bigint  => cast(<double> as bigint)
double->float->double  => cast(<double> as double)
double->float->varchar => cast(<double> as varchar)
```

Wrong result, same shape as #51545: `cast(cast(16777217.5 as float) as int)` should be `16777218` (float rounds `16777217.5 → 16777218.0`, then truncates), but the reduced `cast(16777217.5 as int)` gives `16777217`.

Repro (passes against the current PR ⇒ bug still present):

```java
@Test
public void testFloatMiddleCastStillReduced() {
    ScalarOperator op = new CastOperator(IntegerType.INT,
            new CastOperator(FloatType.FLOAT, ConstantOperator.createDouble(16777217.5)));
    ScalarOperator r = new ReduceCastRule().apply(op, null);
    assertEquals(OperatorType.CONSTANT, r.getChild(0).getOpType()); // middle float cast is gone
    assertTrue(r.getChild(0).getType().isDouble());
}
```

There's also a wider family the type-triple approach won't catch, e.g. `cast(cast(<bigint> as float) as int)`: `grandChild` isn't floating-point (new guard doesn't apply) and `bigint`/`float` are both size 8 (slot-size guard doesn't fire), yet `bigint→float` is lossy.

**Suggestion.** Rather than enumerating `(grandChild, child, parent)` shapes, gate the reduction on whether the *middle* conversion is lossless — which is exactly what `visitBinaryPredicate` already does in this same file (~L134):

```java
if (!TypeManager.isImplicitlyCastable(cast1ChildType, child1Type, true)) {
    return operator;
}
```

`isImplicitlyCastable(grandChild, child, /*strict=*/true)` separates all of these cleanly:

| middle g→c | lossless | reduce |
|---|---|---|
| double→float | no | ❌ |
| bigint→float | no | ❌ |
| double→bigint | no | ❌ (original #51545) |
| int→bigint | yes | ✅ |
| float→double | yes | ✅ |

The existing `float→int→int` relaxation (outer truncation re-applies) can stay as an explicit exception layered on top of the lossless check, instead of being merged into one combined predicate.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

PR #74408, opened 2026-06-05
[BugFix] Self-heal non-PK replica stuck with a permanent version hole

xiangguangyxg:
## Why I'm doing:

A clone racing with publish can skip a middle version on a non-primary-key replica while later versions keep landing, leaving a **permanent hole** in the replica's version chain. The BE honestly reports `version = max_continuous_version` with `version_miss = true`, but the FE finish path keeps advancing `Replica.version` past the hole (`lastFailedVersion` stays `-1`), so the FE believes the replica is healthy. As a result:

- queries keep routing to the holed replica and fail with `capture_consistent_versions error: version not found ... tablet_max_version:<frozen>`;
- the replica still counts toward write quorum;
- `TabletChecker` sees the tablet as HEALTHY and never schedules a repair, so the replica can **never self-heal**.

Related (chaos test that surfaced it): StarRocks/StarRocksTest#11386. Reproduced on `main`.

## What I'm doing:

Detect the stuck hole during tablet report and route the replica into the **existing** recovery path (`tabletRecoveryMap -> handleRecoverTablet -> setBad -> clone`). Marking it bad excludes it from query routing and write-target selection and lets `TabletScheduler` clone a fresh copy from a healthy sibling; the cloned replica then replaces the bad one. No new journal op, replay change, thrift/config change.

Detection (`ReportHandler.needRecoverVersionMiss`) is deliberately conservative:

- **Two-report debounce.** Acts only when the **same** below-visible continuous version is reported with `version_miss=true` on two consecutive reports, tracked by a new non-persisted, leader-local `Replica.versionMissBaselineVersion` that resets on any healthy/advancing report — so a transient out-of-order publish (or a prior healthy report at the same version) is not mistaken for a stuck hole.
- **Genuine source required.** Acts only when another replica on a different backend is `NORMAL`, non-bad, non-error, `lastFailedVersion <= 0`, alive, and `lastReportVersion >= visibleVersion`. `lastReportVersion` comes straight from the BE report, so (unlike the FE-masked `Replica.version`) it is not inflated by the masking that causes this bug — guaranteeing single-replica / all-holed tablets are never stranded and `recover_with_empty_tablet` is never engaged.
- **Excludes** primary-key, colocate, and restore tablets.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
@@ -675,6 +675,25 @@ public class ReportHandler extends LeaderDaemon implements MemoryTrackable {
                         continue;
                     }
 
+                    // A replica can end up with a permanent hole in its version chain that normal publish
+                    // can never fill (BE keeps reporting version_miss with a frozen continuous version).
+                    // Route it into the existing recovery path so it is marked bad and cloned. Skip restore
+                    // tables, like needRecover above (here olapTable is already resolved, so check it directly).
+                    if (olapTable.getState() != OlapTable.OlapTableState.RESTORE) {
+                        MaterializedIndex materializedIndex = physicalPartition.getIndex(tabletMeta.getIndexId());
+                        LocalTablet localTablet = materializedIndex == null
+                                ? null : (LocalTablet) materializedIndex.getTablet(tabletId);
+                        if (localTablet != null && needRecoverVersionMiss(olapTable, localTablet, replica,
+                                physicalPartition.getVisibleVersion(), backendTabletInfo)) {
+                            LOG.warn("replica {} of tablet {} on backend {} misses version persistently and "
+                                            + "cannot self-heal, mark it for recovery. replica in FE: {}, "
+                                            + "report version {}, partition visible version {}",
+                                    replica.getId(), tabletId, backendId, replica,
+                                    backendTabletInfo.getVersion(), physicalPartition.getVisibleVersion());
+                            tabletRecoveryMap.put(tabletMeta.getDbId(), tabletId);
+                        }
+                    }
+
                     TStorageMedium storageMedium = storageMediumMap.get(physicalPartition.getParentId());
                     if (storageMedium != null && backendTabletInfo.isSetStorage_medium()) {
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
```diff
@@ -841,6 +860,92 @@ private static boolean needRecover(Replica replicaInFe, TTabletInfo backendTable
...
+        }
+        if (replica.getState() != ReplicaState.NORMAL || replica.isBad()) {
+            return false;
+        }
+        // Replicas reported as unusable (used==false) are already handled by needRecover().
+        if (backendTabletInfo.isSetUsed() && !backendTabletInfo.isUsed()) {
+            return false;
+        }
+        // Primary-key tablets recover version errors through their own is_error_state path.
+        if (olapTable.getKeysType() == KeysType.PRIMARY_KEYS) {
```
kevincai, line 892: should these filters be moved before any actual check or setter? those replicas won't be fixed under any condition, so check and early return before any getter/setter.

kevincai (dismissed):

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
xiangguangyxg: Good point — done in 58902ccb83. Moved the unconditional filters (replica state/bad, used==false, primary-key, colocate) above the version-miss debounce, so those never-recoverable replicas early-return before touching the baseline getter/setter. The baseline reset still runs for every actionable (NORMAL/non-bad/non-PK/non-colocate) replica on each report, so the two-report debounce semantics are unchanged.

kevincai (approved):

HangyuanLiu (approved):

PR #74409, opened 2026-06-05
[BugFix] Resolve sort-key arity from materialized schema in colocate tablet split

xiangguangyxg:
## Why I'm doing:

A range-distribution colocate table whose sort key is implicit — a PRIMARY KEY table, or a duplicate-key table created without an explicit `ORDER BY` — never splits its oversized tablets. Observed on a shared-data cluster: a ~48 GiB single-tablet PRIMARY KEY colocate table (`colocate_with` = the full primary key) auto-triggered a split 229 times in one day; every split job ran to `FINISHED`, yet the tablet's data size stayed byte-for-byte constant and it never shrank.

Root cause: the data-driven split path `get_tablet_split_ranges_impl` validated `colocate_column_count` against the **raw** `TabletSchemaPB.sort_key_idxes_size()`. For such tables `sort_key_idxes` is empty in the raw PB — the "empty `sort_key_idxes` ⇒ sort key is the key columns" convention is only applied when the schema is materialized. So the arity was read as 0, any `colocate_column_count > 0` was rejected with `InvalidArgument`, and `split_tablet` fell back to an identical-tablet publish — copying the tablet but never actually splitting it. Auto-split then re-fired every cycle with no progress. The external-boundaries split path already avoided this by materializing the schema via `GlobalTabletSchemaMap`.

## What I'm doing:

Resolve the sort-key arity from the materialized `TabletSchema` (via `GlobalTabletSchemaMap`, which applies the empty-`sort_key_idxes` ⇒ key-columns convention) and validate `colocate_column_count` against that, mirroring the external-boundaries path. The FE mirrors the same convention in `MetaUtils.getRangeDistributionColumns`.

Added a BE unit test (`tablet_splitter_test.cpp::data_driven_colocate_split_resolves_empty_sort_key_idxes`): a PRIMARY KEY metadata with empty `sort_key_idxes` and `colocate_column_count == key arity` now passes the arity gate (it fails later at segment loading rather than being wrongly rejected at the gate).

Audited the rest of the BE split/merge code: this is the only site reading raw `sort_key_idxes` where the convention matters; the range/sort-key helpers and the segment-metadata filter all operate on materialized schemas and are unaffected.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1

### DIFF
--- a/be/src/storage/lake/tablet_splitter.cpp
+++ b/be/src/storage/lake/tablet_splitter.cpp
@@ -662,10 +662,18 @@ Status get_tablet_split_ranges_impl(TabletManager* tablet_manager, const TabletM
     if (split_count < 2) {
         return Status::InvalidArgument("Invalid split count, it is less than 2");
     }
-    if (colocate_column_count < 0 || colocate_column_count > tablet_metadata->schema().sort_key_idxes_size()) {
+    // Resolve the sort-key arity from the materialized TabletSchema, not the raw TabletSchemaPB.
+    // A range-distributed table created without an explicit ORDER BY (e.g. a PRIMARY KEY table)
+    // leaves sort_key_idxes empty in the PB; the "empty sort_key_idxes => sort key is the key
+    // columns" convention (which the FE mirrors in MetaUtils.getRangeDistributionColumns) is only
+    // applied when the schema is materialized. Reading the raw PB here would see arity 0 and reject
+    // every colocate split, silently falling back to an identical tablet so the oversized tablet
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

luohaha (approved):

HangyuanLiu (approved):

PR #74412, opened 2026-06-05
[BugFix] fix data race on MaterializedIndexMeta updateSchemaBackendId

kevincai:
updateSchemaBackendId was a plain HashSet mutated under the shared READ lock by both the BE-report send path (add) and task-finish RPCs (remove), which run concurrently. Concurrent add/add and add/remove on a non-thread- safe HashSet can lose entries or, during resize, corrupt the set.

Make the set a ConcurrentHashMap.newKeySet() and collapse the former check-then-act (hasUpdateSchemaTask + addUpdateSchemaBackend) into a single atomic addUpdateSchemaBackendIfAbsent so two senders cannot both observe "no task" and both enqueue a duplicate task. The send path now reserves before building the task and releases the reservation on failure so a backend is never left permanently marked. Table-scoped READ locks at the call sites are unchanged; thread-safety is encapsulated in the class.

Also narrow the finish path (LeaderImpl.finishUpdateSchemaTask) from a full DB READ lock to a table-scoped intensive READ lock. It only touches one known table's indexMeta, so the DB-wide lock was broader than needed; the intensive lock matches the send path and lets DDL/ALTER on unrelated tables in the same DB proceed concurrently.

Add a concurrency regression test that hammers reserve/remove from 16 threads and asserts no exception and an empty set afterward.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndexMeta.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedIndexMeta.java
@@ -48,10 +48,10 @@ import com.starrocks.sql.common.MetaUtils;
 import com.starrocks.thrift.TStorageType;
 
 import java.io.IOException;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 // MaterializedIndexMeta is used to store metadata of a materialized index.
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg (approved):

HangyuanLiu (approved):

PR #74429, opened 2026-06-06
[BugFix] Report real vacuum watermark when retain-boundary metadata is gone

luohaha:
## Why I'm doing:

For shared-data (lake) tables, a partition's `min_retain_version` is clamped to its
`PhysicalPartition.metadataSwitchVersion` (set by a `file_bundling` enable/disable `ALTER`) until
vacuum "crosses" that version. The FE clears the switch version only when a vacuum round reports
`vacuumed_version >= switch_version` (`AutovacuumDaemon`).

If the tablet metadata at and below the retain boundary has already been vacuumed away by an earlier
run, `collect_files_to_vacuum` walks from `min_retain_version` and either:

- never enters the loop, because `min_version` has already advanced past `min_retain_version`; or
- breaks immediately, because the first `get_tablet_metadata(min_retain_version)` returns `NotFound`.

In both cases `skip_check_grace_timestamp` stays false and the function returns
`final_retain_version - 1` (== `min_retain_version - 1`). That watermark is **below** the switch
version, so the FE never clears `metadataSwitchVersion`. The switch is then stranded forever:

- old-format metadata can never be reclaimed (extra object-storage usage), and
- `ALTER ... SET("file_bundling"=...)` is blocked by `allowUpdateFileBundling()` (switch != 0).

This is reachable in practice: a vacuum round legitimately crosses the switch and clears it
**in memory** (the clear is not journaled, only persisted at the next FE image checkpoint); during
`switch == 0` later rounds irreversibly delete the switch version's metadata; then an FE
failover/restart before the next checkpoint reverts `metadataSwitchVersion` from the stale image
while the deletions are permanent. From then on vacuum can only ever return `switch - 1`.

## What I'm doing:

In `collect_files_to_vacuum`, when no tablet metadata exists at or below the retain boundary
(tracked by a new `read_any_metadata` flag), report the real cleaned watermark
`max(final_retain_version, min_version - 1)` instead of `final_retain_version - 1`. Everything below
`min_version` is gone, so cleanup has reached at least `min_version - 1`, and the boundary version
itself no longer exists. This lets the FE clear the stranded switch version on the next vacuum round.

Normal paths are unaffected: the walk always reads at least one metadata, so `read_any_metadata` is
true and the existing behavior is preserved.

Note: this is the CN-side defensive fix (it also self-heals already-stranded partitions and is robust
to FE failover reverting the switch). A follow-up FE PR will make the `metadataSwitchVersion` clear
durable (journaled) so the switch state never goes inconsistent across failover in the first place.

Fixes the stranded `metadataSwitchVersion` / "vacuum stuck at `min_retain_version - 1`" issue.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/vacuum.cpp
+++ b/be/src/storage/lake/vacuum.cpp
@@ -427,6 +427,11 @@ static Status collect_files_to_vacuum(TabletManager* tablet_mgr, std::string_vie
     auto min_version = std::max<int64_t>(1, tablet_info.min_version());
     // grace_timestamp <= 0 means no grace timestamp
     auto skip_check_grace_timestamp = grace_timestamp <= 0;
+    // Whether the walk read at least one tablet metadata at or below |min_retain_version|. It stays
+    // false when the loop never runs (|min_version| has already advanced past |min_retain_version|)
+    // or when the very first read returns NotFound, i.e. the metadata at and below the retain
+    // boundary has already been vacuumed away by a previous run.
+    bool read_any_metadata = false;
     size_t extra_file_size = 0;
     int64_t prepare_vacuum_file_size = 0;
     // Starting at |*final_retain_version|, read the tablet metadata forward along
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

sevev (approved):

kevincai (approved):

PR #74460, opened 2026-06-08
[Enhancement] Support combined txn log / file bundling for FRONTEND_STREAMING loads

luohaha:
## Why I'm doing:

Merge-commit (batch write) loads begin their transaction with `LoadJobSourceType.FRONTEND_STREAMING`, which is **not** included in `LakeTableHelper.isLoadingTransaction()`:

```java
private static boolean isLoadingTransaction(TransactionState.LoadJobSourceType sourceType) {
    return sourceType == BACKEND_STREAMING ||
            sourceType == ROUTINE_LOAD_TASK ||
            sourceType == INSERT_STREAMING ||
            sourceType == BATCH_LOAD_JOB;   // FRONTEND_STREAMING missing
}
```

Both `supportCombinedTxnLog()` and the `fileBundling` gate in `DatabaseTransactionMgr.beginTransaction()` depend on `isTransactionSupportCombinedTxnLog()` → `isLoadingTransaction()`. So for a merge-commit load, `useCombinedTxnLog` is computed as `false` **even when the target table has file bundling enabled**.

The transaction then falls back to per-tablet txn log writes (`kWriteTxnLog`) on the BE: every commit writes one txn log object per tablet to object storage. For a high-frequency merge-commit workload (e.g. `merge_commit_interval_ms=600`) spanning dozens of tablets, this multiplies the object-storage request rate and makes the `finish` phase highly sensitive to object-storage latency spikes — which can push the `add_chunk` RPC past its timeout and abort the load (`tablet writer add chunk timeout ... WaitWriteTime`).

## What I'm doing:

Classify `FRONTEND_STREAMING` (FE streaming load / merge-commit batch write) as a loading transaction so it also skips per-tablet txn log writes and benefits from combined txn log, reducing object-storage request pressure.

Per review feedback, the loading-transaction classification is moved off `LakeTableHelper.isLoadingTransaction()` onto the `LoadJobSourceType` enum itself (a per-constant flag + `isLoadingTransaction()` method), so adding a new source type forces setting the flag and the classification cannot be silently missed. No behavior change from the refactor.

File bundling read/write is unaffected and was already correct for these loads: `useAggregatePublish` at publish time is driven solely by `table.isFileBundling()`, and the read path is load-type agnostic. This change only reduces the number of txn log objects written before publish.

Verified on a shared-data cluster with `lake_metadata_cache_limit=0` (forcing bundled-metadata reads from object storage on every query). Both import types that produce `FRONTEND_STREAMING` were exercised against a `file_bundling` table and confirmed to commit as `FRONTEND_STREAMING`, publish to `VISIBLE`, and read back correctly:
- **Merge commit / batch write** (`enable_merge_commit`) — 3 loads, 30000 rows.
- **Parallel (channel) transaction stream load** (`/api/transaction/*` with `channel_id`/`channel_num`) — 10000 rows.

Note: `MULTI_STATEMENT_STREAMING` was evaluated and intentionally **not** included. Testing showed that enabling combined txn log for transaction stream load makes publish fail with a combined-txn-log 404 (the BE transaction-stream-load write path does not produce the combined log that publish expects), leaving the txn stuck in COMMITTED. Supporting it requires propagating the combined-txn-log strategy into that BE write path first, which is separate work.

Updated `LakeTableHelperTest` accordingly (`testSupportCombinedTxnLog`, `testIsTransactionSupportCombinedTxnLog`).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1 
  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
@@ -361,20 +361,12 @@ public class LakeTableHelper {
     }
 
     public static boolean supportCombinedTxnLog(TransactionState.LoadJobSourceType sourceType) {
-        return RunMode.isSharedDataMode() && Config.lake_use_combined_txn_log && isLoadingTransaction(sourceType);
-    }
-
-    private static boolean isLoadingTransaction(TransactionState.LoadJobSourceType sourceType) {
-        return sourceType == TransactionState.LoadJobSourceType.BACKEND_STREAMING ||
-                sourceType == TransactionState.LoadJobSourceType.ROUTINE_LOAD_TASK ||
-                sourceType == TransactionState.LoadJobSourceType.INSERT_STREAMING ||
-                sourceType == TransactionState.LoadJobSourceType.FRONTEND_STREAMING ||
-                sourceType == TransactionState.LoadJobSourceType.BATCH_LOAD_JOB;
+        return RunMode.isSharedDataMode() && Config.lake_use_combined_txn_log && sourceType.isLoadingTransaction();
     }
 
     // for now, only loading txn and compaction txn support combined txn log
     public static boolean isTransactionSupportCombinedTxnLog(TransactionState.LoadJobSourceType sourceType) {
-        return isLoadingTransaction(sourceType) || sourceType == TransactionState.LoadJobSourceType.LAKE_COMPACTION;
+        return sourceType.isLoadingTransaction() || sourceType == TransactionState.LoadJobSourceType.LAKE_COMPACTION;
     }
 
     // if one of the tables in tableIdList is LakeTable with file bundling, return true
--- a/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
@@ -368,6 +368,7 @@ public class LakeTableHelper {
         return sourceType == TransactionState.LoadJobSourceType.BACKEND_STREAMING ||
                 sourceType == TransactionState.LoadJobSourceType.ROUTINE_LOAD_TASK ||
                 sourceType == TransactionState.LoadJobSourceType.INSERT_STREAMING ||
+                sourceType == TransactionState.LoadJobSourceType.FRONTEND_STREAMING ||
                 sourceType == TransactionState.LoadJobSourceType.BATCH_LOAD_JOB;
     }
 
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
banmoy: how about moving this function to LoadJobSourceType so that we can be aware of updating it easily when adding a new type next time?

fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
```diff
@@ -368,6 +368,7 @@ private static boolean isLoadingTransaction(TransactionState.LoadJobSourceType s
         return sourceType == TransactionState.LoadJobSourceType.BACKEND_STREAMING ||
                 sourceType == TransactionState.LoadJobSourceType.ROUTINE_LOAD_TASK ||
                 sourceType == TransactionState.LoadJobSourceType.INSERT_STREAMING ||
+                sourceType == TransactionState.LoadJobSourceType.FRONTEND_STREAMING ||
```
banmoy: also need to add MULTI_STATEMENT_STREAMING

luohaha: https://github.com/StarRocks/starrocks/pull/74479

fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
luohaha: Done — moved the loading-transaction classification onto `LoadJobSourceType` itself as a per-constant flag with an `isLoadingTransaction()` method, so adding a new source type now forces setting it.

fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
luohaha: Added in this PR. #74479 is closed and folded in here, so this PR now covers both FRONTEND_STREAMING and MULTI_STATEMENT_STREAMING.

banmoy (approved):

fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
luohaha: Correction: I tested MULTI_STATEMENT_STREAMING on a shared-data cluster and it does **not** work yet — enabling combined txn log for transaction stream load makes publish fail with a combined-txn-log 404 (`...log/<txn>.logs` not found), leaving the txn stuck in COMMITTED and data invisible. Reproduced even with a single-table, non-file_bundling target, so it's the combined-txn-log write path for transaction stream load that's missing, not a bundling issue. I've removed MULTI_STATEMENT_STREAMING from this PR; it needs the BE transaction-stream-load path to actually write the combined log first. (#74479 stays closed.)

meegoo (approved):

kevincai (approved):

PR #74464, opened 2026-06-08
[BugFix] Bound IVM adaptive refresh when the first delta exceeds mv_max_rows_per_refresh

Youngwb:
## Why I'm doing:

When `mv_max_rows_per_refresh` (or `mv_max_bytes_per_refresh`) is set, an asynchronous IVM refresh is supposed to split a large base-table delta into multiple bounded task runs. In `MVIVMRefreshProcessor.getBaseTableChangedDeltaAdaptive` the loop accumulates each delta trait's added rows/bytes and stops once it reaches the cap — but the `break` fired **before** `toSnapshot` advanced, and `toSnapshot` was seeded to the **end** of the full delta. So when the first (oldest) delta trait alone already met the cap, `toSnapshot` stayed at the range end, the whole backlog was refreshed in a single run, and no follow-up run was scheduled — the cap had no effect on the per-refresh size.

Separately, the follow-up condition only checked the row cap, so reaching the **byte** cap also failed to schedule the next run.

Note: the defaults (`mv_max_rows_per_refresh = 100000000`, `mv_max_bytes_per_refresh = 20GB`) are large, so default refreshes are unaffected; only explicitly-lowered caps were impacted.

## What I'm doing:

- Always advance `toSnapshot` by at least one delta trait before breaking, so a single delta that already exceeds the cap becomes its own bounded batch and the remainder defers to a follow-up task run (forward progress preserved; a delta cannot be split below one snapshot).
- Make the follow-up condition honor the byte cap as well as the row cap, matching the break condition.
- Extract the decision into a pure `computeAdaptiveDelta` and add `MVIVMRefreshProcessorTest` covering fine-grained batching, the oversized-first-delta regression, the byte-cap follow-up, and the empty-stats single-trait (unchanged) case.

Fixes #issue: N/A — no public issue; found while reviewing IVM adaptive-refresh batching.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
@@ -14,6 +14,7 @@
 
 package com.starrocks.scheduler.mv.ivm;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

Astralidea (approved):

HangyuanLiu (approved):

Youngwb:
@mergify backport branch-4.1

PR #74466, opened 2026-06-08
[BugFix] Support sub-1% sampling percent in TABLE SAMPLE and histogram analyze

stephen-shelby:
## Why I'm doing:

On large tables, histogram analyze (`ANALYZE FULL TABLE ... UPDATE HISTOGRAM`)
derives the sample ratio from `histogram_max_sample_row_count` (default 10M).
When the table is large enough, that ratio falls below 1%. The FE then computed
the sample percent as `(int)(ratio * 100)`, which truncates any sub-1% ratio to
`0` and emits `SAMPLE('percent'='0')`. The parser rejects this with:
com.starrocks.sql.parser.ParsingException: ... Detail message: invalid percent which should in (0, 100).

So histogram collection simply fails on sufficiently large tables. The sampling
probability was an integer across the whole path (FE clause → thrift → BE), so
fractional percents could not be represented even if the FE stopped truncating.

## What I'm doing:

Make the sampling probability a `double` end-to-end, and keep wire/back-compat
via a new thrift field:

- **thrift**: add `TTableSampleOptions.probability_percent_v2` (`double`); keep
  the legacy `i64 probability_percent` so an old BE during rolling upgrade still
  reads a valid (rounded, clamped) value.
- **FE `TableSampleClause`**: parse/store percent as `double`, allow fractional
  values in the open range `(0, 100)`, set both thrift fields, and render
  percents without a trailing `.0` (e.g. `0.5`, `10`).
- **FE `HistogramStatisticsCollectJob`**: emit the true (possibly sub-1%) percent
  instead of truncating to an integer. Also fix the same latent truncation-to-0
  in `HyperStatisticSQLs` by clamping to `[1, 100]` like its sibling builders.
- **BE `data_sample` / `segment_iterator` / `olap_chunk_source`**: prefer the new
  `double` field (falling back to the legacy field), and use `double` throughout
  the bernoulli and histogram-sampling paths.

Added FE UTs, BE UTs, and SQL tests covering sub-1% sampling.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/pipeline/scan/olap_chunk_source.cpp
+++ b/be/src/exec/pipeline/scan/olap_chunk_source.cpp
@@ -962,9 +962,11 @@ void OlapChunkSource::_update_counter() {
 
     // Data sampling
     if (_params.sample_options.enable_sampling) {
+        double sample_percent = _params.sample_options.__isset.probability_percent_v2
+                                        ? _params.sample_options.probability_percent_v2
+                                        : static_cast<double>(_params.sample_options.probability_percent);
         _runtime_profile->add_info_string("SampleMethod", to_string(_params.sample_options.sample_method));
-        _runtime_profile->add_info_string("SamplePercent",
-                                          std::to_string(_params.sample_options.probability_percent) + "%");
+        _runtime_profile->add_info_string("SamplePercent", std::to_string(sample_percent) + "%");
         COUNTER_UPDATE(ADD_CHILD_TIMER(_runtime_profile, "SampleTime", parent_name),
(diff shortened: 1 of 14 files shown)


CelerData-Reviewer:
@codex review

PR #74467, opened 2026-06-08
[BugFix] Record exported row count in audit log for SELECT INTO OUTFILE

gengjun-git:
## Why I'm doing:

For `SELECT ... INTO OUTFILE` over the MySQL protocol, the audit log recorded `ReturnRows=0` even when N rows were actually exported. OUTFILE does not deliver result rows to the client, so `ConnectContext.returnRows` (accumulated per result batch) stays 0; the BE-reported exported row count was fed into that counter only on the Arrow Flight SQL path, and `ConnectProcessor.auditAfterExec` then overwrote the audit `ReturnRows` field with the zero context counter. The client `Query OK, N rows affected` was correct, but the audit log under-reported.

## What I'm doing:

Widen the branch in `StmtExecutor.processQueryStatisticsFromResult` so OUTFILE queries (not only Arrow Flight SQL) take the row count from the BE-reported `QueryStatistics.returnedRows` — the number of rows actually written to file, summed across split files — guarded against null. Regular queries are unaffected: they already accumulate `returnRows` per result batch and do not enter this branch, so there is no double counting.

Verified end-to-end on a running cluster exporting to S3 (MinIO): same `SELECT ... INTO OUTFILE` went from audit `ReturnRows=0` (before) to `ReturnRows=7` (after). Also verified accurate across query shapes against a real table — `WHERE` filters, empty result (0), `LIMIT`, `ORDER BY`+`LIMIT`, `GROUP BY`, full scan, and a 1,000,000-row export split into 3 files — in every case `ReturnRows` equals the actual number of exported rows.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
@@ -2078,7 +2078,8 @@ public class StmtExecutor {
     /**
      * The query result batch will piggyback query statistics in it
      */
-    private void processQueryStatisticsFromResult(RowBatch batch, ExecPlan execPlan, boolean isOutfileQuery) {
+    @VisibleForTesting
+    void processQueryStatisticsFromResult(RowBatch batch, ExecPlan execPlan, boolean isOutfileQuery) {
         if (batch != null && parsedStmt.getOrigStmt() != null && parsedStmt.getOrigStmt().getOrigStmt() != null) {
             statisticsForAuditLog = batch.getQueryStatistics();
             statisticsForAuditLogFromPlaceholder = false;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

wyb (approved):

srlch (approved):

PR #74479, opened 2026-06-08
[Enhancement] Support combined txn log / file bundling for multi-statement streaming loads

luohaha:
## Why I'm doing:

Multi-statement streaming loads begin their transaction with `LoadJobSourceType.MULTI_STATEMENT_STREAMING` (see `TransactionLoadAction` / `StreamLoadMultiStmtTask`), which is **not** included in `LakeTableHelper.isLoadingTransaction()`:

```java
private static boolean isLoadingTransaction(TransactionState.LoadJobSourceType sourceType) {
    return sourceType == BACKEND_STREAMING ||
            sourceType == ROUTINE_LOAD_TASK ||
            sourceType == INSERT_STREAMING ||
            sourceType == BATCH_LOAD_JOB;   // MULTI_STATEMENT_STREAMING missing
}
```

As a result, `supportCombinedTxnLog()` returns `false` for these loads, so `useCombinedTxnLog` is computed as `false`. The transaction then writes one txn log object **per tablet** to object storage (`kWriteTxnLog`) instead of a single combined txn log per transaction, multiplying the object-storage request rate on commit.

Note that a multi-statement streaming load goes through the exact same explicit-transaction commit path (`TransactionStmtExecutor.beginStmt` → `loadData` → `commitStmt`) as an explicit `BEGIN ... INSERT ... COMMIT` transaction, which uses `INSERT_STREAMING` and **already** supports combined txn log. The two differ only by source-type label, so this is the natural symmetric gap.

This is a follow-up to #74460, which added `FRONTEND_STREAMING` for the same reason.

## What I'm doing:

Add `MULTI_STATEMENT_STREAMING` to `isLoadingTransaction()` so multi-statement streaming loads also benefit from combined txn log, reducing object-storage request pressure on commit.

Updated `LakeTableHelperTest` accordingly (`testSupportCombinedTxnLog`, `testIsTransactionSupportCombinedTxnLog`).

File bundling read/write is unaffected and was already correct for these loads: `useAggregatePublish` at publish time is driven solely by `table.isFileBundling()`, and the read path is load-type agnostic. This change only reduces the number of txn log objects written before publish.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/LakeTableHelper.java
@@ -368,6 +368,7 @@ public class LakeTableHelper {
         return sourceType == TransactionState.LoadJobSourceType.BACKEND_STREAMING ||
                 sourceType == TransactionState.LoadJobSourceType.ROUTINE_LOAD_TASK ||
                 sourceType == TransactionState.LoadJobSourceType.INSERT_STREAMING ||
+                sourceType == TransactionState.LoadJobSourceType.MULTI_STATEMENT_STREAMING ||
                 sourceType == TransactionState.LoadJobSourceType.BATCH_LOAD_JOB;
     }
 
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

banmoy (approved):

luohaha:
Folded into #74460 together with the enum refactor, per review discussion. Closing this one.

PR #74489, opened 2026-06-08
[Enhancement] Add ORC meta-tier sampling for tablet pre-split

xiangguangyxg:
## Why I'm doing:

The tablet pre-split meta tier (Tier 1) reads file-footer statistics on the FE to predict the target tablet layout without issuing a BE query. Until now it only handled **Parquet** sources; **ORC** sources fell through to the data tier (Tier 2), which issues a BE sample sub-query per load. ORC footers carry per-stripe min/max statistics analogous to Parquet row groups, so they can be sampled directly on the FE — skipping the sub-query. This is a performance optimization, not a coverage gap (ORC pre-split already worked via the data tier).

## What I'm doing:

- Add `OrcStripeStatisticsReader`, the ORC analogue of `ParquetRowGroupStatisticsReader`: reads per-stripe min/max/row-count for the sort-key column. Conservative integer-only type window (BYTE/SHORT/INT/LONG → TINYINT/SMALLINT/INT/BIGINT); any other type, missing/all-null stripe stats, or ambiguous column match falls back to the data tier via `MetaTierUnavailableException`; real I/O failures surface as `StarRocksException`. The footer read is pinned to the load's snapshotted file length (`ReaderOptions.maxLength`), matching the Parquet reader's snapshot guarantee.
- Add `MetaTierFormat`, an enum centralizing the format→reader dispatch and the supported-format window. Both the INSERT-from-FILES and Broker Load providers now route through it instead of each hard-coding a Parquet-only reject gate.
- Pin `org.apache.orc:orc-core` as a direct `fe-core` dependency (Maven `dependencyManagement` + Gradle mirror); it was previously only pulled in transitively.

`ParquetMetadataSampler` is unchanged — it already consumes the format-agnostic `RowGroupStatistics`, so the boundary-planning logic is shared. No new metric, no new config, no BE change.

A go/no-go feasibility check confirmed empirically that the runtime non-shaded `orc-core` 1.9.1 reads ORC footers with the FE's bundled `protobuf-java` 3.25.5 with no version conflict.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

The load result is identical; only the internal sampling path changes (FE footer read instead of a BE sample sub-query) for ORC sources that were already eligible for pre-split.

## Checklist:

- [ ] This is a backport pr


CelerData-Reviewer:
@codex review

xiangguangyxg:
## TSP end-to-end validation (shared-data)

Built this PR merged onto `main` and ran it on a TSP shared-data cluster (1 FE + 3 CN, FE reports `main-5265869`). The differential signal is the FE metric `tablet_pre_split_tier_used{tier="meta_tier"|"data_tier"}` (diffed before/after each load), plus target tablet count (1→N) and row count. Each target is an empty, unpartitioned range-distribution table (1 tablet); load is a straight `INSERT INTO <t> SELECT * FROM FILES(...)`.

| Case | Source | Sort key | `tier_used` Δ | tablets | rows | Result |
|------|--------|----------|---------------|---------|------|--------|
| A | ORC (4 disjoint-range files) | `BIGINT` | **meta_tier +1** | 1→3 | 20,000,000 | ✅ ORC meta tier hit — the new path, no BE sample sub-query |
| B | same ORC files | `VARCHAR` (ORDER BY) | **data_tier +1**, meta_tier +0 | 1→3 | 20,000,000 | ✅ string is out-of-window → `MetaTierUnavailableException` → data-tier fallback |
| C | Parquet (4 disjoint-range files) | `BIGINT` | data_tier +1 | 1→3 | 20,000,000 | ✅ pre-split works; tier is data_tier (see note) |
| D | ORC | hash `DISTRIBUTED BY` | none (skipped) | 4 (unchanged) | 20,000,000 | ✅ `eligibility_skipped{reason="not_range_distribution"}` +1, sampler not invoked |

**Key proof (A vs B):** identical ORC data — a `BIGINT` sort key routes through **meta_tier**, a `VARCHAR` sort key falls back to **data_tier**. This isolates exactly the new `OrcStripeStatisticsReader` and its integer type-window, end-to-end through the real load path, with the BE sample sub-query skipped on the meta-tier path.

Final cumulative counters: `sampler_invocations=4`, `tier_used{meta_tier}=1`, `tier_used{data_tier}=3`, `eligibility_skipped{not_range_distribution}=1` (data_tier=3 includes one earlier single-stripe ORC run + B + C).

**Note on Case C (not a regression):** StarRocks' own `INSERT INTO FILES(parquet)` writer annotates integers (`Int(bitWidth=32, isSigned=true)`, confirmed via pyarrow), and the pre-existing `ParquetRowGroupStatisticsReader` window admits only *unannotated* ints, so it falls back to data tier. This PR routes Parquet through that same unchanged reader (via `MetaTierFormat.PARQUET.read`), so the behavior is identical to `main`; Parquet pre-split itself still works (1→3). A positive Parquet-meta control needs externally-written unannotated-int Parquet and is covered by existing unit tests + prior regressions.

Also worth noting for reviewers: well-encoded data (sequential int + low-cardinality string) packs 20M rows into a single ORC stripe; the meta tier needs ≥2 non-overlapping stripes to place boundaries, hence the multiple disjoint-range files above.

kevincai (approved):

srlch (approved):

xiangguangyxg:
@Mergify backport branch-4.1

PR #74499, opened 2026-06-08
[Enhancement] Lift the DCG x gap NotSupported in shared-data tablet merge

xiangguangyxg:
## Why I'm doing:

Shared-data range-distribution tablet **merge** had a known limitation: when a child tablet had compacted away its share before merge (leaving a row-coverage *gap* on a canonical shared rowset) **and** that same shared rowset carried a DCG (partial-column-update) **same-column conflict** that needs a physical `.cols` rebuild, `merge_tablet` returned `Status::NotSupported`.

The merge already synthesizes a *gap delvec* (`compute_synthesized_gap_specs`) that masks exactly those gap rowids, so they are never returned at read time — but the DCG rebuild's row-window coverage check did not consult that bitmap and rejected the hole. FE scheduling worked around it by requiring all children fully compacted before merging any tablet with active partial-update DCGs.

## What I'm doing:

Let the DCG rebuild consult the **same** gap bitmap that drives delvec masking, so a coverage hole that coincides with masked gap rowids is accepted instead of failing.

- Hoist `compute_synthesized_gap_specs` into `merge_tablet` (PK-only) and pass the one `std::vector<CanonicalGapSpec>` to **both** `merge_dcg_meta` and `merge_delvecs`. The bitmap that masks the delvec and the bitmap that accepts DCG coverage are now the *same object* — they cannot diverge.
- Extract the coverage logic into a pure, unit-tested `tablet_reshard_helper::reconcile_windows_with_gap`. It accepts a hole only after proving it is fully contained in `gap_bits` (`Roaring::containsRange`); unmasked holes, distinct-owner overlaps, and zero-row targets still error. A `nullptr` bitmap reproduces the original strict contiguous-coverage check verbatim.
- In `rebuild_dcg_for_target_segment`, fill the accepted gap windows from the canonical **base segment** (real, already-indexed values). The rows are provably delvec-masked and never returned, and using base values avoids any secondary/vector-index validity questions that synthetic defaults would raise.

**Core invariant:** a gap row is filled only after proving its window ⊆ `gap_bits`, and `gap_bits` is exactly what `merge_delvecs` writes to the delvec — so every filled row is provably masked.

Because gap synthesis stays PK-only and `nullptr` keeps strict coverage, non-PK and no-gap merges are byte-for-byte unchanged. The FE "compact-all-before-merge for active-DCG tablets" workaround can be dropped in a follow-up FE change (out of scope here).

Tests: 7 pure-helper unit tests for `reconcile_windows_with_gap` + 3 end-to-end gap×DCG rebuild tests (first/middle/last child compacted); existing 26 dcg/gap reshard tests unchanged. All 36 pass.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Range-distribution split/merge is a new feature gated behind `enable_range_distribution` and `tablet_reshard_enable_tablet_merge` (both default off, not yet GA); no existing user observes a change. The only effect is that a merge that previously returned `NotSupported` on this gated path now completes correctly.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/tablet_merger.cpp
+++ b/be/src/storage/lake/tablet_merger.cpp
@@ -844,17 +844,20 @@ const TabletSchemaPB* resolve_rowset_schema_pb(const TabletMetadataPB& new_metad
     return nullptr;
 }
 
-struct DcgRowWindow {
-    size_t old_tablet_index;
-    Range<rowid_t> range;
-};
+using tablet_reshard_helper::DcgRowWindow;
 
 // Step C — compute row windows in the target segment for every source old tablet
-// rowset that references it. Coverage over [0, num_rows_of_target) is validated.
+// rowset that references it. Holes left by a compacted-away old tablet are
+// accepted as is_gap windows when they fall within |gap_bits| (the same
+// synthesized bitmap merge_delvecs masks); a nullptr |gap_bits| keeps the strict
+// contiguous-coverage requirement. The opened base segment and its row count are
+// returned so the caller can fill gap rows from it.
 Status compute_row_windows_for_source_rowsets(TabletManager* tablet_manager, int64_t new_tablet_id,
                                               const RowsetMetadataPB& target_rowset, int target_segment_position,
                                               const TabletSchemaCSPtr& full_tablet_schema,
                                               const std::vector<DcgSourceRowsetReference>& source_references,
+                                              const roaring::Roaring* gap_bits,
+                                              std::shared_ptr<Segment>* out_base_segment,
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

luohaha (approved):

PR #74502, opened 2026-06-08
[BugFix] Materialize sort-key schema only for colocate-aware tablet split

xiangguangyxg:
## Why I'm doing:

Follow-up to #74409. That PR made `get_tablet_split_ranges_impl` validate `colocate_column_count` against a **materialized** `TabletSchema` (so PRIMARY KEY / no-explicit-`ORDER BY` range-colocate tables, whose raw `sort_key_idxes` is empty, are no longer wrongly rejected and silently degraded to an identical, un-split tablet). But it did the materialization **unconditionally**, which has two problems on the non-colocate split path:

- A non-colocate split (`colocate_column_count == 0`) does not need the sort-key arity at all, yet now materializes the schema (and pollutes the global schema map) on every split.
- `GlobalTabletSchemaMap::emplace` has a `DCHECK_NE(invalid_id(), id)`. Synthetic metadata without a schema id (e.g. `LakeTabletReshardTest.test_tablet_splitting`) therefore **aborts the BE unit tests in a DEBUG build**. This is exactly what failed the branch-4.1 backport CI (#74491), which runs a DEBUG build. The #74409 main CI happened to run with `DCHECK` compiled out, so it didn't catch this.

## What I'm doing:

Gate the materialization and the arity check on `colocate_column_count > 0`. A non-colocate split keeps its original behavior (it never consulted the sort-key arity), and the colocate-aware path is unchanged. This removes the needless per-split materialization and the debug-build abort, and keeps `main` consistent with the branch-4.1 fix.

No new test: the colocate path is already covered by `tablet_splitter_test.cpp::data_driven_colocate_split_resolves_empty_sort_key_idxes` (added in #74409), and the non-colocate path is exercised by `LakeTabletReshardTest.test_tablet_splitting` (which aborts in a DEBUG build without this gate).

The equivalent gating is already carried by the branch-4.1 backport (#74491), so this PR does **not** need to be backported to 4.1.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/be/src/storage/lake/tablet_splitter.cpp
+++ b/be/src/storage/lake/tablet_splitter.cpp
@@ -662,18 +662,25 @@ Status get_tablet_split_ranges_impl(TabletManager* tablet_manager, const TabletM
     if (split_count < 2) {
         return Status::InvalidArgument("Invalid split count, it is less than 2");
     }
-    // Resolve the sort-key arity from the materialized TabletSchema, not the raw TabletSchemaPB.
-    // A range-distributed table created without an explicit ORDER BY (e.g. a PRIMARY KEY table)
-    // leaves sort_key_idxes empty in the PB; the "empty sort_key_idxes => sort key is the key
-    // columns" convention (which the FE mirrors in MetaUtils.getRangeDistributionColumns) is only
-    // applied when the schema is materialized. Reading the raw PB here would see arity 0 and reject
-    // every colocate split, silently falling back to an identical tablet so the oversized tablet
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

luohaha (approved):

PR #74533, opened 2026-06-09
[Enhancement] Make ANN refine a user-controlled, quantization-keyed option

sevev:
## Why I'm doing:

Refining an ANN result (re-ranking candidates by recomputing the exact distance on the full-precision vectors) was hard-wired to the IVFPQ index family via the `use_ivfpq` flag. HNSW now also supports quantization (`quantizer` = sq4/sq8/pq), whose index distances are likewise approximate, so the flag was keyed on the wrong axis: a quantized HNSW silently ranked on lossy distances with no way to ask for accuracy, while IVFPQ always paid for refine even when the user was happy to trust the index distance.

## What I'm doing:

Key the refine path on whether the index is **quantized**, and gate it behind a new session variable so users choose the accuracy/cost trade-off.

- **`enable_vector_index_refine`** (session, default `false`): when on, a quantized-index query refines; when off, it trusts the (lossy) index distance. No effect on a non-quantized index, whose distance is already exact.
- FE derives `is_quantized` (IVFPQ, or HNSW with a non-flat `quantizer`) and sets the new `TVectorSearchOptions.refine_distance = is_quantized && enable_vector_index_refine`.
- BE renames the `use_ivfpq` gate to `refine_distance` (1:1 — it always meant "run the refine path"); it now also covers quantized HNSW. The over-fetch knobs (`k_factor` / `pq_refine_factor`) are unchanged.
- `EXPLAIN` shows `Refine: ON/OFF` instead of `IVFPQ: ON/OFF`.

**Rolling-upgrade compatibility.** `use_ivfpq` (thrift ordinal 8) is deprecated but kept reserved and treated as a synonym of `refine_distance` during the deprecation window: FE sets both fields to the same value, and BE reads `refine_distance || use_ivfpq`. So a version-skewed FE/BE — in either upgrade order — picks the same execution path instead of silently switching between the trust and refine layouts. (An old FE always sets `use_ivfpq=true` for IVFPQ, so a new BE keeps refining IVFPQ for it.)

**Distance-range correctness fix.** Previously a `approx_*_distance(...) <= c` bound was folded entirely into the ANN range and the predicate dropped, so on the refine path the bound was only enforced on the lossy index distance via `range_search` (and not at all when a segment's `.vi` was missing, since that path reads the rows and recomputes the distance above the scan). The refine path now also keeps the bound as a Filter re-applied on the recomputed exact distance above the scan, removing false positives the lossy prefilter let through. The trust path is unchanged (it folds the bound into the range and accepts the lossy result). This corrects `test_vector_index_ivfpq`, whose recorded range results previously contained out-of-range rows.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

  - [ ] I have added documentation for my new feature or new function (deferred to a follow-up doc PR)
- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

luohaha (approved):

meegoo (approved):

HangyuanLiu (approved):

PR #74550, opened 2026-06-09
[UT] Add SQL integration tests for light_weight_tablet_creation

starrocks-xupeng:
## Why I'm doing:

Cover the user-visible surface of `light_weight_tablet_creation` (the property introduced in #73422 / #73845 / #74132) end-to-end against a real shared-data cluster. The existing FE unit tests already exercise the analyzer / metastore / alter-job code paths in isolation; this PR adds SQL integration tests so we also lock down behavior across distribution and partition variants, ALTER (both directions, including the true → false backfill path) and schema-change / ADD ROLLUP.

## What I'm doing:

`test/sql/test_light_weight_tablet_creation/` splits the cases into two files:

**`T/test_basic` — 12 cases** (DDL + DML on light-weight tables)
* Hash distribution with and without explicit BUCKETS
* Random distribution
* Range partitioned, with and without explicit BUCKETS
* Range distribution on a PRIMARY KEY table (uses `admin set frontend config ('enable_range_distribution' = 'true')`)
* `ADD PARTITION`
* `TRUNCATE TABLE`
* `INSERT OVERWRITE`
* Auto partition by `date_trunc('day', ...)` (INSERT-triggered)
* Explicit `LIST` partition
* Expression `LIST` auto partition (INSERT-triggered)

**`T/test_alter` — 5 cases** (ALTER paths)
* `ALTER ... SET ('light_weight_tablet_creation' = 'true')` from a `false` table
* `ALTER ... SET ('light_weight_tablet_creation' = 'invalid')` must error with `Property light_weight_tablet_creation must be bool type(false/true)`
* `ALTER ... SET ('light_weight_tablet_creation' = 'false')` from a `true` table (exercises the #74132 backfill path; verifies the table is still usable after backfill)
* Schema Change `MODIFY COLUMN` on a light-weight table, waiting via `function: wait_alter_table_finish()`
* `ADD ROLLUP` on a light-weight table, waiting via `function: wait_alter_table_finish(\"ROLLUP\", 8)`

## Cases intentionally not integrated

* Dynamic partition — wall-clock-driven scheduler + `CURDATE()` make the diff non-deterministic; existing tests already cover the basic CREATE-with-`dynamic_partition.enable=true` shape.
* Non-shared-data OLAP rejection — the SQL harness runs against a shared-data cluster, so the negative case cannot be constructed.
* FE restart edit-log replay — the harness has no way to restart FE.
* `SHOW PARTITIONS` / `SHOW PROC` / `SHOW TABLET` output and OSS-side `ossutil` checks from the manual scripts — they reference variable partition / tablet IDs or external tooling that the diff-based harness cannot validate.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/test/sql/test_light_weight_tablet_creation/R/test_alter
+++ b/test/sql/test_light_weight_tablet_creation/R/test_alter
@@ -1,0 +1,165 @@
+-- name: test_light_weight_tablet_creation_alter @cloud @sequential
+create database db_${uuid0};
+-- result:
+-- !result
+use db_${uuid0};
+-- result:
+-- !result
+CREATE TABLE t_alter_basic (
+    c0 INT,
+    c1 STRING
+) DUPLICATE KEY(c0)
+DISTRIBUTED BY HASH(c0) BUCKETS 4
+PROPERTIES("light_weight_tablet_creation" = "false");
+-- result:
+-- !result
+ALTER TABLE t_alter_basic SET("light_weight_tablet_creation" = "true");
+-- result:
+-- !result
+SHOW CREATE TABLE t_alter_basic;
+-- result:
+t_alter_basic	CREATE TABLE `t_alter_basic` (
+  `c0` int(11) NULL COMMENT "",
+  `c1` varchar(65533) NULL COMMENT ""
+) ENGINE=OLAP 
+DUPLICATE KEY(`c0`)
+COMMENT "OLAP"
+DISTRIBUTED BY HASH(`c0`) BUCKETS 4 
+PROPERTIES (
+"cloud_native_fast_schema_evolution_v2" = "true",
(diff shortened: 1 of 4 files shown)


starrocks-xupeng:
@mergify rebase

huailiu1122:
@mergify rebase

xiangguangyxg (approved):

kevincai (approved):

banmoy (approved):

PR #74551, opened 2026-06-09
[BugFix] Support sub-1% sampling percent in TABLE SAMPLE and histogram analyze

stephen-shelby:
## Why I'm doing:

On large tables, histogram analyze (`ANALYZE FULL TABLE ... UPDATE HISTOGRAM`) derives the sample ratio from `histogram_max_sample_row_count` (default 10M). When the table is large enough, that ratio falls below 1%, and the FE computed the sample percent as `(int)(ratio * 100)`, which truncates any sub-1% ratio to `0` and emits `SAMPLE('percent'='0')`. The parser rejects this with `invalid percent which should in (0, 100)`, so histogram collection simply fails on sufficiently large tables. The sampling probability was an integer across the whole path (FE clause → thrift → BE), so fractional percents could not be represented even if the FE stopped truncating.

In addition, allowing sub-1% percents makes the BE "sample at least one block/page" fallback fire frequently, which exposed two latent bugs in that fallback:
- `BlockDataSample` used the random block *index* directly as a row id (biasing the sample toward the head of the table), and used an inclusive upper bound that could point one block past the end.
- `PageDataSample` used an inclusive upper bound that could index one page past the end.

## What I'm doing:

**1. Make the sampling probability a `double` end-to-end, keeping wire/back-compat:**
- **thrift**: add `TTableSampleOptions.probability_percent_v2` (`double`); keep the legacy `i64 probability_percent` so an old BE during a rolling upgrade still reads a valid value (rounded and clamped to `[1, 99]`).
- **FE `TableSampleClause`**: parse/store the percent as `double`, allow fractional values in the open range `(0, 100)`, populate both thrift fields, and render percents without a trailing `.0` (e.g. `0.5`, `10`).
- **FE `HistogramStatisticsCollectJob`**: emit the true (possibly sub-1%) percent via `BigDecimal` instead of truncating to an integer — this avoids both binary-float noise (e.g. `0.49999999999999994`) and scientific notation that the parser cannot consume.
- **BE `data_sample` / `segment_iterator` / `olap_chunk_source`**: prefer the new `double` field (falling back to the legacy field for an old FE), and use `double` throughout the bernoulli and histogram-sampling paths.

**2. Fix the data-sampling fallback so it stays block-aligned and in-bounds:**
- `BlockDataSample`: convert the random block index to a row offset (instead of using it directly as a row id) and clamp to `[0, num_blocks - 1]`.
- `PageDataSample`: clamp the page index to `[0, num_pages - 1]`.

**3. Misc:** clamp the hyper sample-hint percent to `[1, 99]` (the parser rejects `100`) to match `TableSampleClause.toThrift`, and tidy up imports.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/ast/TableSampleClause.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/ast/TableSampleClause.java
@@ -89,10 +89,8 @@ public final class TableSampleClause implements ParseNode {
 
     public void toThrift(TTableSampleOptions msg) {
         msg.setEnable_sampling(true);
-        // Keep the legacy integer field populated for backward compatibility with old BE that only reads it.
-        // Round to the nearest integer in (0, 100) so old BE still produces a sane (non-zero) sample.
-        msg.setProbability_percent(Math.max(1L, Math.min(99L, Math.round(randomProbabilityPercent))));
-        // New field carries the exact (possibly sub-1%) value; new BE prefers this.
+        // BE is always upgraded before FE, so a new FE only ever talks to a BE that already understands
+        // this field. The legacy integer probability_percent is intentionally left unset.
         msg.setProbability_percent_v2(randomProbabilityPercent);
         msg.setSample_method(sampleMethod.toThrift());
         msg.setRandom_seed(randomSeed);
--- a/fe/fe-core/src/main/java/com/starrocks/sql/ast/TableSampleClause.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/ast/TableSampleClause.java
@@ -22,6 +22,7 @@ import com.starrocks.thrift.TTableSampleOptions;
 import org.apache.commons.lang3.EnumUtils;
 import org.apache.logging.log4j.util.Strings;
 
+import java.math.BigDecimal;
 import java.util.Map;
 import java.util.concurrent.ThreadLocalRandom;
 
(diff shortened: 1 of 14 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/ast/TableSampleClause.java
```diff
@@ -86,7 +89,11 @@ public void analyzeProperties(Map<String, String> properties) throws AnalysisExc
 
     public void toThrift(TTableSampleOptions msg) {
         msg.setEnable_sampling(true);
-        msg.setProbability_percent(randomProbabilityPercent);
+        // Keep the legacy integer field populated for backward compatibility with old BE that only reads it.
+        // Round to the nearest integer in (0, 100) so old BE still produces a sane (non-zero) sample.
+        msg.setProbability_percent(Math.max(1L, Math.min(99L, Math.round(randomProbabilityPercent))));
```
trueeyu: Is it necessary to set old item?

trueeyu (approved):

Youngwb (approved):

PR #74563, opened 2026-06-09
[BugFix] Keep key attribute on MODIFY COLUMN for DUPLICATE/UNIQUE tables

### LINKED ISSUE #74553: MODIFY key column without keyness on DUPLICATE tables, failing with misleading "Can not change aggregation type"

```sql
CREATE TABLE test_db.t (
  id   INT          NOT NULL,
  name VARCHAR(12)  NOT NULL,
  v    FLOAT        NOT NULL
)
DUPLICATE KEY(id, name)

-- Change ONLY the comment of the key column `name`, keeping the exact same type.
-- Note: the KEY keyword is not restated (a very natural thing to omit).
ALTER TABLE test_db.t MODIFY COLUMN name VARCHAR(12) COMMENT 'full name';
```

Result:

```
ERROR 1064 (HY000): Can not change aggregation type
```

Workaround — explicitly restate `KEY`:

```sql
ALTER TABLE test_db.t MODIFY COLUMN name VARCHAR(12) KEY COMMENT 'full name';  -- succeeds
```

The same problem reproduces on UNIQUE KEY tables when modifying a key column without restating `KEY`.
It does **not** happen on PRIMARY KEY or AGGREGATE tables (see root cause).

Modifying only the comment (or only the type) of an existing **key** column should succeed without
the caller having to restate the `KEY` keyword. A `MODIFY COLUMN` should preserve the column's
existing keyness rather than implicitly turning a key column into a value column.

This is what PRIMARY KEY and AGGREGATE tables already do, and what the test
`RangeDistributionGuardTest#testModifyNonSortKeyColumnAllowedOnRangeDistribution` documents as the
intended behavior ("MODIFY of a ... column with **no keyness change** must still succeed").

### Real behavior (Required)

The statement fails with a context-free, misleading error:

```
Can not change aggregation type
```

I have never specified any aggregation type. Because `KEY` was omitted, the engine treated the
existing key column as a value column (an unintended **keyness flip**) and implicitly assigned it an
aggregation type, which then conflicts with the original column's `null` aggregation type.

- Reproduced on `3.5.11` (`select current_version()`).
- Confirmed still present on `main`.

sfwang218:
## Why I'm doing:

`MODIFY COLUMN` on an existing **key** column of a DUPLICATE or UNIQUE table silently dropped the key attribute when the `KEY` keyword was omitted (for example, when changing only a column comment), and then failed with the misleading error `Can not change aggregation type` — even though the user never specified any aggregation type.

```sql
CREATE TABLE test_db.t (
  id   INT          NOT NULL,
  name VARCHAR(12)  NOT NULL,
  v    FLOAT        NOT NULL
)
DUPLICATE KEY(id, name)

-- Change ONLY the comment of the key column `name`, keeping the same type:
ALTER TABLE test_db.t MODIFY COLUMN name VARCHAR(12) COMMENT 'full name';
-- ERROR 1064 (HY000): Can not change aggregation type
```

The same reproduces on UNIQUE KEY tables. Workaround was to explicitly restate `KEY`.

**Root cause:** `SchemaChangeHandler#processModifyColumn` derives the key attribute per table model. PRIMARY KEY tables restore the key flag from the base column for backward compatibility, and AGGREGATE KEY tables do so implicitly (no aggregate ⇒ key). The DUPLICATE and UNIQUE branches did neither: with `KEY` omitted, `modColumn.isKey()` was `false`, so the column was assigned an implicit aggregation type (`NONE` / `REPLACE`). `Column#checkSchemaChangeAllowed` then compared it against the original key column (`aggregationType == null`) and threw. The key set of DUPLICATE/UNIQUE tables is fixed at creation time and cannot be changed by `MODIFY COLUMN`, so a key column should never be implicitly turned into a value column.

## What I'm doing:

1. **Primary fix** — in the UNIQUE and DUPLICATE branches of `processModifyColumn`, restore `isKey` from the base column when `KEY` is omitted, mirroring the documented backward-compatibility behavior of the PRIMARY KEY branch. After this, the repro succeeds without restating `KEY`, and key-attribute handling is consistent across all four table models (PRIMARY / AGGREGATE / UNIQUE / DUPLICATE).

2. **Secondary fix** — make the aggregation-type mismatch error in `Column#checkSchemaChangeAllowed` actionable by including the column name and the `from → to` aggregation types, matching the sibling type-change message in the same method. This still fires for a genuine aggregation change (e.g. AGGREGATE value column `SUM` → `MAX`).

Added `ModifyKeyColumnKeepKeyTest` covering: DUPLICATE key-column comment change, UNIQUE key-column comment change, and the improved actionable message on a genuine AGGREGATE aggregation change. All 3 pass.

Fixes #74553

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -779,6 +779,13 @@ public class SchemaChangeHandler extends AlterHandler {
                 throw new DdlException("Can not assign aggregation method on column in Unique data model table: " +
                         modColumn.getName());
             }
+            Column baseColumn = olapTable.getBaseColumn(modColumn.getName());
+            if (baseColumn != null && baseColumn.isKey()) {
+                // backward compatibility, same as the PRIMARY KEY branch: the key set of a unique table is
+                // fixed at creation time and cannot be changed by MODIFY COLUMN, so keep an existing key
+                // column as a key column when the KEY keyword is omitted in the modify column clause.
+                modColumn.setIsKey(true);
+            }
             if (!modColumn.isKey()) {
                 modColumn.setAggregationType(AggregateType.REPLACE, true);
             }
@@ -786,6 +793,13 @@ public class SchemaChangeHandler extends AlterHandler {
             if (null != modColumn.getAggregationType()) {
                 throw new DdlException("Can not assign aggregation method on column in Duplicate data model table: " +
                         modColumn.getName());
+            }
+            Column baseColumn = olapTable.getBaseColumn(modColumn.getName());
+            if (baseColumn != null && baseColumn.isKey()) {
+                // backward compatibility, same as the PRIMARY KEY branch: the key set of a duplicate table is
+                // fixed at creation time and cannot be changed by MODIFY COLUMN, so keep an existing key
+                // column as a key column when the KEY keyword is omitted in the modify column clause.
+                modColumn.setIsKey(true);
             }
             if (!modColumn.isKey()) {
                 modColumn.setAggregationType(AggregateType.NONE, true);
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -727,6 +727,48 @@ public class SchemaChangeHandler extends AlterHandler {
         }
     }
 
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

PR #74564, opened 2026-06-09
[BugFix] meta_tool: report true per-column data size from ordinal index

luohaha:
## Why I'm doing:

`meta_tool dump_column_size` reported each column's "compressed bytes". For a
complex type (`array` / `map` / `struct`) it scanned **each sub-column separately**
(via a per-leaf access path) and summed the per-sub-column
`compressed_bytes_read_request`.

The sub-columns of a complex type share structural streams (null flags, offsets).
Reading any one leaf also reads those shared streams to reconstruct the column, so
each sub-column's reported size included them, and **summing the per-sub-column
sizes double-counted the shared streams** — the per-column total over-reported
complex columns. On a real segment this showed up as near-identical byte counts
repeated across the sub-columns of one complex column.

## What I'm doing:

Scan the whole column once and report a single total per column, instead of
scanning each sub-column and summing.

- A full-column read touches every underlying stream (elements, keys/values, null
  flags, offsets, dictionary pages) exactly once, so the reported size is correct
  and free of double-counting.
- It naturally includes dictionary pages (the read decodes via the dict page), which
  a per-data-page sum would miss.
- The full per-column structure is still printed via the recursive `ColumnMetaPB`
  DebugString.

Verified end-to-end on a TSP shared-nothing cluster: a 200k-row table with
`array`/`map` columns now reports one additive total per column (no per-sub-column
duplication); the sum of per-column compressed bytes (1,751,429) matches the segment
file size (1,757,472) within the index/footer overhead that is not column data.

This only changes the `meta_tool` diagnostic output; no production code path is
affected.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [x] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/tools/meta_tool.cpp
+++ b/be/src/tools/meta_tool.cpp
@@ -1479,52 +1479,25 @@ Status SegmentDump::dump_column_size() {
             return Status::OK();
         };
 
-        if (tablet_column.subcolumn_count() == 0) {
-            // regular column
-            read_the_segment().ok();
+        // Scan the whole column once and report a single total. For complex types
+        // (ARRAY/MAP/STRUCT) this reads every underlying stream (elements, keys/values,
+        // null flags, offsets, dictionary pages) exactly once, so the reported size is
+        // correct and free of double-counting. Previously each sub-column was scanned
+        // separately and the per-sub-column sizes summed, which over-counted the
+        // structural streams (null/offset) shared across sub-columns. The full column
+        // structure is still printed via the recursive ColumnMetaPB DebugString below.
+        read_the_segment().ok();
 
-            auto compession_desc = CompressionTypePB_descriptor()->FindValueByNumber(column_meta.compression());
-            auto encoding_desc = EncodingTypePB_descriptor()->FindValueByNumber(column_meta.encoding());
+        auto compession_desc = CompressionTypePB_descriptor()->FindValueByNumber(column_meta.compression());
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

luohaha:
Thanks for the review. I've reworked the fix (force-pushed) to address both points at once.

Instead of summing per-sub-column ordinal-index page sizes, `dump_column_size` now **scans the whole column once** and reports a single total per column. A full-column read touches every underlying stream exactly once, so:

- **Dictionary pages** (P2 #1) are included — the read decodes via `_load_dict_page()`, so the `dict_page` bytes are counted.
- **Structural streams** for ARRAY/MAP/STRUCT (P2 #2) — null flags and offsets — are read once as part of the single scan, so they are neither omitted nor double-counted.

This also removes the original double-counting root cause: previously each sub-column was scanned separately and the per-sub-column sizes summed, which counted the shared null/offset streams once per sub-column. The full per-column structure is still printed via the recursive `ColumnMetaPB` DebugString.

kevincai (approved):

xiangguangyxg (approved):

PR #74572, opened 2026-06-09
[Enhancement] Unify lake no-SST condition update onto pk_index_execution pool

luohaha:
## Why I'm doing:

While analysing a slow `PublishTablet` trace for a lake PK table with condition
update, the `condition_update_compare_phase_us` counter dominated (~4.5s) while
every recorded sub-operation summed to a couple of milliseconds. Two problems
surfaced:

1. The non-SST condition-merge compare phase in `_do_update_with_condition`
   submits per-chunk compare tasks to `lake_partial_update_thread_pool`, while the
   sibling SST path (`_do_update_with_condition_parallel`) and branch-4.1 both use
   `pk_index_execution_thread_pool`. The no-SST path therefore contends with
   row-mode partial-update publishes instead of staying with the other PK-index
   work, and the two condition-merge paths are inconsistent.
2. `TRACE_COUNTER_*` reads a thread-local current trace that pool worker threads do
   not inherit. Counters incremented inside the compare task were silently dropped
   from the publish trace, which is exactly why the breakdown "went dark" where the
   time was actually spent.

## What I'm doing:

In `UpdateManager::_do_update_with_condition` (non-SST path):

- **Unify the thread pool:** route the compare tasks to
  `pk_index_execution_thread_pool`, matching the SST path and branch-4.1.
- **Single-chunk fast path:** hold the first chunk's compare task; if it is the
  only chunk, run it inline on the publish thread instead of paying a thread-pool
  round-trip (the common small-publish case). Once a second chunk proves there is
  real parallel work, flush the held task and submit every chunk from then on. The
  token is always drained before returning so worker closures never outlive the
  locals they capture.
- **Propagate the trace:** capture the publish thread's `Trace::CurrentTrace()` and
  `ADOPT_TRACE` it inside each worker task, mirroring the existing pattern in
  `lake_persistent_index.cpp`. Worker-side counters (e.g.
  `process_condition_update_count`, `get_column_values`/segment-IO timings) now
  flow back into the publish trace.

Test: `test_condition_update_no_sst_trace_propagation` forces >1 per-segment chunk
so the compare runs on the pool, verifies the merge result, and asserts
`process_condition_update_count` is visible in the adopted trace (would be 0 without
the propagation fix). Existing no-SST tests already cover the single-chunk inline
path and the multi-chunk parallel path for correctness.

No user-facing, config, or interface change; this is internal task scheduling plus
observability. The inline + trace-propagation improvements would also be useful on
branch-4.1 (which already uses `pk_index_execution_thread_pool`); happy to follow up
with a backport if desired.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/update_manager.cpp
+++ b/be/src/storage/lake/update_manager.cpp
@@ -1314,11 +1314,22 @@ Status UpdateManager::_do_update_with_condition(const RowsetUpdateStateParams& p
     std::vector<uint32_t> read_column_ids;
     read_column_ids.push_back(condition_column);
 
+    // Only parallelize when the segment iterator will actually split into more than one
+    // compare chunk. SegmentPKIterator splits at pk_index_parallel_execution_min_rows, so a
+    // rowset smaller than that yields a single chunk per segment, and comparing it inline on
+    // the publish thread is cheaper than a thread-pool round-trip. The rowset-wide row count
+    // is a safe lower bound for the per-segment count: if the whole rowset is under the
+    // threshold, every segment is too.
     std::unique_ptr<ThreadPoolToken> token;
-    if (config::enable_pk_index_parallel_execution) {
-        token = GlobalEnv::GetInstance()->lake_partial_update_thread_pool()->new_token(
+    if (config::enable_pk_index_parallel_execution &&
+        params.op_write.rowset().num_rows() >= config::pk_index_parallel_execution_min_rows) {
+        token = GlobalEnv::GetInstance()->pk_index_execution_thread_pool()->new_token(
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg (approved):

kevincai (approved):

PR #74574, opened 2026-06-09
[Enhancement] Add PLAIN_ENCODING_DELTA_OFFSET for string columns

luohaha:
## Why I'm doing:

High-cardinality string/varchar columns fall back to plain (non-dictionary) encoding,
whose page offset trailer is stored as **absolute offsets**. Absolute offsets increase
monotonically and compress poorly under LZ4: the low bytes never repeat, and the two
high zero-bytes form only a 2-byte run that is below LZ4's 4-byte minimum match length,
so they survive as literals. As a result the offset trailer (~4 bytes/row) is essentially
incompressible and dominates the compressed size of high-cardinality string columns.

## What I'm doing:

Add a new column encoding **`PLAIN_ENCODING_DELTA_OFFSET`**: same as `PLAIN_ENCODING` for
strings, but the page offset trailer stores **per-value deltas (string lengths)** instead
of absolute offsets. Deltas are near-constant for fixed-ish strings, so identical 4-byte
words repeat and LZ4 collapses them; the uncompressed trailer keeps the same size.

- The format is recorded in the **column meta encoding**, not via any in-trailer flag — the
  count field stays a plain element count. `EncodingInfo::get` dispatches the decoder by
  encoding, so a BE that does not understand `PLAIN_ENCODING_DELTA_OFFSET` **fails to open
  the segment with a clear error instead of silently misreading it**.
- `BinaryPlainPageBuilder`/`Decoder` take a `delta_offset` ctor flag; the decoder
  reconstructs absolute offsets once per page in `init()` (little- and big-endian), leaving
  the per-row read path unchanged. Dictionary columns are unaffected.
- `StringColumnWriter::speculate_string_encoding` selects the new encoding instead of
  `PLAIN_ENCODING` when the mutable config `enable_binary_plain_delta_offset` is on
  (default `false`).

### Measured benefit

A/B on the same TSP shared-nothing cluster, 5M-row `varchar id` column, only the offset
encoding differs:

| id data | absolute (baseline) | delta | reduction |
|---|--:|--:|--:|
| high-cardinality (`enum,long-number`) | 75,645,274 | 56,814,739 | **-24.9%** |
| low-cardinality prefix (`persona,N`) | 41,059,263 | 21,091,158 | **-48.6%** |

Row-by-row `JOIN` verification showed identical reads (0 mismatches). The saving is roughly
the compressed cost of the offset trailer (~4 bytes/row); the percentage is larger when the
character data itself compresses well.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

The new config defaults to `false`, so existing deployments are unaffected unless it is
explicitly enabled. Because the new encoding is recorded per column, a BE without support
fails to open such a segment rather than misreading it; enable only after the whole cluster
is upgraded, and note that segments written with it are not readable after downgrading to a
version without support.

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

luohaha:
Reworked the design (force-pushed) to address both P1 comments at the root, dropping the high-bit-in-count approach entirely.

The format is now a distinct column encoding **`PLAIN_ENCODING_DELTA_OFFSET`** (new `EncodingTypePB` value, appended at ordinal 8) recorded in the column metadata, instead of stealing the high bit of the page trailer's count field.

- **P1: gate delta pages before writing the new trailer flag** — there is no trailer flag anymore; the count field is a plain count. Cross-version safety now comes from the encoding itself: a BE that does not know `PLAIN_ENCODING_DELTA_OFFSET` returns `EncodingInfo::get` → `"fail to find valid type encoding"` and **fails to open the segment (clear error) instead of misreading it**. The config is still default-off; docs note it must not be enabled until the cluster is fully upgraded and that such segments are not readable after a downgrade.
- **P1: use the generated config forward header** — `binary_plain_page.h` no longer includes `common/config.h` at all (the delta flag is a ctor parameter). The config is consulted only in `column_writer.cpp`, which reads it via the generated `config_rowset_fwd.h` (added to the manifest and regenerated). `check_common_config_header_includes.sh` now passes.

Also added handling in `storage_page_decoder.cpp` (new encoding maps to the base data decoder like `PLAIN_ENCODING`) and updated the unit test to drive the format via the ctor flag.

CelerData-Reviewer:
@codex review

luohaha:
Addressed the P3 comment: corrected the `enable_binary_plain_delta_offset` config comment (and the regenerated `config_rowset_fwd.h`) — it no longer mentions an in-trailer flag. The comment now states the format is identified by the `PLAIN_ENCODING_DELTA_OFFSET` column encoding recorded in segment metadata, and that a BE without support fails to open the segment instead of misreading.

Also added unit tests to cover the new paths:
- `encoding_info_test`: resolves `PLAIN_ENCODING_DELTA_OFFSET` for VARCHAR/CHAR and round-trips through the builder/decoder the traits create.
- `column_reader_writer_test`: end-to-end high-cardinality string write/read, asserting the speculated column encoding (`PLAIN_ENCODING` with the config off, `PLAIN_ENCODING_DELTA_OFFSET` with it on) and identical reads — covers `StringColumnWriter::speculate_string_encoding` and the read-side dispatch.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg (dismissed):

be/src/storage/rowset/binary_plain_page.h
```diff
@@ -109,9 +115,22 @@ class BinaryPlainPageBuilder final : public PageBuilder {
...
+        // monotonically increasing absolute offsets. The format is identified by the column's
+        // encoding (PLAIN_ENCODING_DELTA_OFFSET), not by any in-trailer flag.
+        if (_delta_offset) {
+            uint32_t prev = 0;
+            for (uint32_t off : _offsets) {
+                put_fixed32_le(&_buffer, off - prev);
+                prev = off;
+            }
+        } else {
+            for (uint32_t _offset : _offsets) {
```
wyb:
```suggestion
            for (uint32_t offset : _offsets) {
```

be/src/storage/rowset/binary_plain_page.h
wyb: line 42: update comment

luohaha:
Thanks @wyb, addressed both in 25c336b:

1. Renamed the plain-branch loop variable `_offset` → `offset` (local, not a member — dropped the leading-underscore prefix), per your suggestion.
2. Updated the `BinaryPlainPage` trailer doc comment (the `// Trailer / Offsets` block) to note the offset section holds per-value deltas (string lengths) for `PLAIN_ENCODING_DELTA_OFFSET`, with the absolute offset being the prefix sum — the trailer size is unchanged, only the values differ.

CelerData-Reviewer:
@codex review

wyb:
`BinaryDictPageBuilder::reset()` will fall back to `PLAIN_ENCODING` if the dict size is larger than 1M.


BinaryDictPageBuilder::reset() is actually invoked from ScalarColumnWriter::finish_current_page(), at be/src/storage/rowset/column_writer.cpp:703:

  _page_builder->reset();   // _page_builder is PageBuilder*; for DICT columns it points to BinaryDictPageBuilder

  Call chain

  reset() is a virtual method on PageBuilder, so the call goes through the base-class pointer _page_builder->reset() and dispatches to BinaryDictPageBuilder::reset(). The flow:

  1. ScalarColumnWriter::append() writes data; once the page is full it triggers
  2. → finish_current_page(): first finish() flushes the current data page out (_push_back_page), then calls _page_builder->reset() to reset the builder for the next page (column_writer.cpp:703)
  3. → BinaryDictPageBuilder::reset() (binary_dict_page.cpp:128)

luohaha:
Good catch raising the dict-fallback path — I confirmed it is intentionally **unaffected** by this change, and that is both correct and the safe choice:

- The config only changes what `StringColumnWriter::speculate_string_encoding()` returns up front, i.e. the **column-level** encoding. A column either becomes `PLAIN_ENCODING_DELTA_OFFSET` (written by `BinaryPlainPageBuilder(opts, /*delta=*/true)` directly — no `BinaryDictPageBuilder` involved), or it stays `DICT_ENCODING`.
- For `DICT_ENCODING` columns, `BinaryDictPageBuilder` is untouched: both the dictionary page (`binary_dict_page.cpp:72`) and the plain-fallback data pages it switches to when the dict exceeds 1M (`binary_dict_page.cpp:131`) construct `BinaryPlainPageBuilder(_options)` with the default `delta_offset=false`. The reader side (`BinaryDictPageDecoder` → `BinaryPlainPageDecoder<Type>(_data)`, `:206`) is also default `false`. So those fallback pages keep absolute offsets and round-trip correctly.

So the dict-fallback plain pages do **not** get the delta size win, by design. The reason it is deliberately scoped out rather than "also enabled there":

> A `DICT_ENCODING` column carries no per-page encoding signal that its plain-fallback pages are delta-encoded (the page's in-band header is just `PLAIN_ENCODING`). If we delta-encoded them, both `BinaryDictPageDecoder` and any **older BE** reading that column would misread the offsets — the exact silent-corruption hazard the column-encoding (`PLAIN_ENCODING_DELTA_OFFSET`) approach was chosen to avoid. The whole point of using a distinct column encoding is that an unknowing reader fails fast via `EncodingInfo::get` instead of misreading.

Covering the dict-fallback pages would require a format/version bump on the dict page header itself (with its own cross-version handling), which I'd prefer to keep out of this PR. Happy to do it as a follow-up if you think the extra coverage is worth it — let me know.

wyb (approved):

xiangguangyxg (approved):

Issue #73594, opened 2026-05-20
Shared-data: Workload Identity for ADLS2 storage volume is silently converted to Managed Identity, breaking authentication on CN


aadrii:
### Steps to reproduce the behavior (Required)

**Context:** StarRocks running in shared-data mode on AKS, with Azure Workload Identity configured on the FE and CN pods. ADLS Gen2 used as the backing storage.

**ServiceAccount used (with Workload Identity annotation):**

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: starrocks-sa
  namespace: sr-s3
  annotations:
    azure.workload.identity/client-id: "<client-id>"
  labels:
    azure.workload.identity/use: "true"

```

**Helm chart values used to deploy the cluster:**

```yaml
starrocks:
    starrocksCluster:
        enabledBe: false
        enabledCn: true
        componentValues:
            podLabels:
                azure.workload.identity/use: "true"
    starrocksFESpec:
        serviceAccount: starrocks-sa
    starrocksCnSpec:
        serviceAccount: starrocks-sa

```

**Verification that Workload Identity is correctly injected in the pod:**

```bash
$ kubectl exec -n <ns> <fe-pod> -- env | grep -i azure
AZURE_TENANT_ID=<tenant-id>
AZURE_FEDERATED_TOKEN_FILE=/var/run/secrets/azure/tokens/azure-identity-token
AZURE_AUTHORITY_HOST=https://login.microsoftonline.com/
AZURE_CLIENT_ID=<client-id>

$ kubectl exec -n <ns> <fe-pod> -- ls /var/run/secrets/azure/tokens/
azure-identity-token
```

The federated token file is present and readable inside the pod.

**Reproduction steps:**

1. Deploy a shared-data StarRocks cluster on AKS with Azure Workload Identity enabled on the FE and CN ServiceAccounts. The user-assigned Managed Identity has `Storage Blob Data Contributor` on the target storage account, with a federated credential pointing to the K8s ServiceAccount used by the pods.

2. Create a storage volume using the `azure.adls2.oauth2_token_file` property (introduced in PR #62754):

```sql
   CREATE STORAGE VOLUME adls2_storage
   TYPE = ADLS2
   LOCATIONS = ("adls2://<container>/<path>")
   PROPERTIES (
     "enabled" = "true",
     "azure.adls2.endpoint" = "https://<storage-account>.dfs.core.windows.net/",
[...]

PR #74581, opened 2026-06-09
[BugFix] Fix typo in azure_adls2_oauth2_client_endpoint config field name

Gui13:
missing doc

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [x] 3.4

## What problem does this PR solve?

When deploying StarRocks in `shared_data` mode with Azure ADLS2 storage using OAuth2 authentication (`cloud_native_storage_type = ADLS2`), the FE fails to start with:

```
SemanticException: Storage params is not valid {..., "azure.adls2.oauth2_client_endpoint":"", ...}
```

The `azure.adls2.oauth2_client_endpoint` parameter is always empty, even when `azure_adls2_oauth2_client_endpoint` is correctly set in `fe.conf`.

**Root cause**: `Config.java` declares the field as `azure_adls2_oauth2_oauth2_client_endpoint` (duplicate `oauth2_`), so the value from `fe.conf` is parsed under a different key, while `SharedDataStorageVolumeMgr` reads from the misnamed field and always gets an empty string.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -3319,8 +3319,8 @@ public class Config extends ConfigBase {
     public static String azure_adls2_oauth2_client_id = "";
     @ConfField
     public static String azure_adls2_oauth2_client_secret = "";
-    @ConfField
-    public static String azure_adls2_oauth2_oauth2_client_endpoint = "";
+    @ConfField(aliases = {"azure_adls2_oauth2_oauth2_client_endpoint"})
+    public static String azure_adls2_oauth2_client_endpoint = "";
 
     // gcp gs
     @ConfField
--- a/fe/fe-core/src/main/java/com/starrocks/server/SharedDataStorageVolumeMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/SharedDataStorageVolumeMgr.java
@@ -642,7 +642,7 @@ public class SharedDataStorageVolumeMgr extends StorageVolumeMgr {
                 params.put(CloudConfigurationConstants.AZURE_ADLS2_OAUTH2_CLIENT_SECRET,
                         Config.azure_adls2_oauth2_client_secret);
                 params.put(CloudConfigurationConstants.AZURE_ADLS2_OAUTH2_CLIENT_ENDPOINT,
-                        Config.azure_adls2_oauth2_oauth2_client_endpoint);
+                        Config.azure_adls2_oauth2_client_endpoint);
                 break;
             case "gs":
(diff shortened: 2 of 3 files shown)


Gui13:
I think this fixes https://github.com/StarRocks/starrocks/issues/73594

At least, I can now provision a oauth2 adls2 storage with this. I'll confirm that the managed identity works tomorrow.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3320,7 +3320,7 @@ public class Config extends ConfigBase {
     @ConfField
     public static String azure_adls2_oauth2_client_secret = "";
     @ConfField
-    public static String azure_adls2_oauth2_oauth2_client_endpoint = "";
+    public static String azure_adls2_oauth2_client_endpoint = "";
```
kevincai, line 3323: shall put the typo name as alias, otherwise this will be a break change if the user already uses this typo name in the production.

Gui13: Good idea, this will make it more clear that it was a typo.

Gui13: Done.

kevincai: didn't see the new commit, do you forget to push the new changes?

Gui13:
> I think this fixes #73594
> 
> At least, I can now provision a oauth2 adls2 storage with this. I'll confirm that the managed identity works tomorrow.

Ok I have reviewed the code before testingn, and the token file is not even declared in the configuration file. Also, from what I understand from the code, declaring it is not sufficient: the library used to transfer the configuration from FE to BE needs also to be changed. That's the Starlet lib?

And then we have to make sure that the starlet lib can actually use the token file instead of the hardcoded token. I don't know about starlet so IDK what this work entails. 

Right now we use service provider with password, and this means we have to change it every year :/

fe/fe-core/src/main/java/com/starrocks/common/Config.java
Gui13: Yep, sorry, amended the commit so my git push failed. I forced it ans it went through.

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

Issue #74586, opened 2026-06-09
[Enhancement] Move spillable pipeline operators off the busy-poller onto the event scheduler


eshishki:
### The problem

A blocked pipeline driver is woken one of two ways: a legacy **poller** thread that spins re-checking every blocked driver (`O(N)` `is_not_blocked()` calls, never sleeps while any driver is blocked), or the **event scheduler** that wakes a driver only when the state it waits on changes. Most operators use the event scheduler. **The spillable operators — aggregation, hash join, sort — never adopted it:** they block on spill IO but subscribe to no wakeup, so the poller must spin-poll them.

On a production node under steady load the poller thread (`pip_poll_com`) burns ~0.51 core continuously ≈ 21 % of all useful pipeline work on the host, with 83 % of drivers blocked and the node otherwise ~99 % idle. It scales with nothing — it spins as long as one driver is blocked, and the spilled drivers are what keep it spinning.

```
Poller (legacy — what spilled ops use today)     Event scheduler (where we move them)
  one thread, forever:                             whoever changes the state notifies the sleeper:
  while true:                                        spiller flush/restore completion ─┐
    for d in blocked_drivers:  # O(N), spins         channel enqueue/drain ────────────┼─► sink/source_trigger()
      if d.is_not_blocked(): ready.put(d)            build-side load done ──────────────┘           │
    # never sleeps while any driver is blocked                                       re-check one predicate, re-queue
```

### The fix — a base event layer, then each family opts in

**Base (PR 1)** adds the shared machinery, inert until an operator opts in:

- **`BlockReason`** — a typed reason a parked branch is waiting on (`WAIT_FLUSH`, `WAIT_RESTORE`, `WAIT_CHANNEL`, `WAIT_LATCH`, `WAIT_BUILD`). Each operator declares which reasons its subscriptions cover; a `named<>()` compile guard rejects returning a reason the operator never subscribed to, and a park-time check is the runtime backstop. A plain bool cannot express "this branch has no wakeup"; the type can.
- **Per-spiller event-notification layer** — a sink list and a source list on the spiller; an IO completion fires the matching trigger. A lifetime guard (`GuardedCompletion`) closes the decrement→notify use-after-free window (notify runs before the query-lifetime pin is released).
[...]

PR #74587, opened 2026-06-09
[Enhancement] Add the spill-event scheduler base infrastructure

eshishki:
## Why I'm doing:

The spillable operators — aggregation, hash join, sort — still run without the event scheduler: they block on spill IO but subscribe to no wakeup, so the busy-poller has to spin-poll them. Eventizing each family needs the same shared machinery: a typed block reason and a notification layer on the spiller. This PR is that common base — inert on its own, so every plan keeps running exactly as before. Problem statement, measurements and the full plan are in #74586.

## What I'm doing:

Under the event scheduler, whoever changes the state a driver waits on notifies that driver directly, instead of a poller thread re-checking it on every spin. This PR adds the two shared pieces that let a spillable operator be woken that way, all unused until an operator subscribes:

- **`BlockReason`** — a typed reason a parked branch is waiting on (`WAIT_FLUSH`, `WAIT_RESTORE`, `WAIT_CHANNEL`). An operator returns the reason it is blocked on and declares a `covered_wakeups()` mask of the reasons its subscriptions actually wake. `named<R, Mask>()` is a compile-time guard: returning a reason outside the mask fails to compile. At park time `verify_block_reason_covered()` is the runtime backstop (DCHECK in debug; a fail-soft metric + rate-limited warning in release). A plain bool cannot express "this branch has no wakeup"; the type can.
- **Per-spiller event-notification layer** (`SpillEventObservable`) — a sink-side and a source-side observer list on the `Spiller`. An IO completion fires the matching trigger: `sink_trigger()` wakes an `OUTPUT_FULL` sleeper, `source_trigger()` an `INPUT_EMPTY` one. Completions go through `complete_io` (publish → decrement the in-flight count → notify); task bodies use `defer_complete_io`, whose RAII completion runs that order and then releases the task's query-lifetime pin — the pin is what keeps the notified observers alive, so the ordering is welded inside one destructor. It is deliberately not `PipeObservable` (whose `defer_notify_sink` would wake the sink list with the wrong trigger and lose the wakeup).

Also reworks the spill-process pump (channel + operator) for one-shot notify and adds the `SpillableFlatSinkMixin` shared by the agg/join sinks — both behavior-preserving on the existing poller path.

The interior-operator driver machinery (for the spilled hash-join probe) is not part of this PR; it moved into #74589 per review.

Part of #74586

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/compute_env/spill/spiller.h
```diff
@@ -169,6 +170,44 @@ class Spiller : public std::enable_shared_from_this<Spiller> {
...
+        std::forward<PublishFn>(publish_fn)();
+        if (wake_sink) {
+            notify_sink_observers();
+        }
+        notify_source_observers();
+        decrease_in_flight_io();
+        if (wake_sink) {
+            notify_sink_observers();
+        }
+        notify_source_observers();
```
stdpain, line 197: I don't understand why we need to notify twice.

stdpain:
Could we separate the changes required for the intermediate spill operator?

From my perspective, the changes for Sort and Aggregation are relatively low risk, and we could merge them before the 4.2 release.

For Join, I'd like to do some additional testing before merging, as it may require more validation.

be/src/compute_env/spill/spiller.h
eshishki: Good point — the double notify wasn't actually needed. The first notify was the data wakeup, the second was the release wakeup for PENDING_FINISH drivers gating on the in-flight count (a driver woken before the decrement re-parks, and the decrement alone wakes nobody). But since the whole completion runs inside the task's query-lifetime guard window, a single notify after the decrement covers both cases: the predicates are already published and the count is already at its new value.

Done in 02ffb69, together with the RAII shape you suggested: `Spiller::defer_complete_io(publish_fn, wake_sink, guard)` returns a completion object whose destructor runs publish → decrement → notify and then `guard.scoped_end()`, in that order — the guard's query pin is what keeps the notified observers alive across the notify, so the ordering is welded inside the destructor instead of being spelled out at every call site. The submit-failure compensation keeps calling the immediate `complete_io` directly: that path never entered the guard window, so it must not run `scoped_end`.

be/src/exec/query_cache/conjugate_operator.cpp
```diff
@@ -79,6 +79,26 @@ Status ConjugateOperator::set_finishing(RuntimeState* state) {
...
+    auto reason = _sink_op->block_reason();
+    if (reason != pipeline::BlockReason::NONE) {
+        return reason;
+    }
+    return _source_op->block_reason();
+}
+
+uint32_t ConjugateOperator::covered_wakeups() const {
+    return _sink_op->covered_wakeups() | _source_op->covered_wakeups();
+}
```
stdpain: This is the query cache code, which won't work with spill, so you can discard these modifications.

eshishki: Removed from the base PR. One finding from verifying this: for the aggregation the combination really is inert — the wrappers don't forward `spillable()`/`set_execute_mode`, so a cached pre-cache agg never engages spill in auto mode (cache stays correct, spill just doesn't kick in). A cache-wrapped spillable join probe is reachable though — its spill decision is made by the unwrapped build sink — so the forwarding moved into the join PR together with the rest of the interior-operator machinery, to be validated there.

eshishki:
Done. The base PR now carries only the edge canon (everything sort and aggregation need); the interior-operator machinery (`INTERMEDIATE_BLOCK`, `supports_intermediate_wakeup` + the all-factory gate, `WAIT_LATCH`/`WAIT_BUILD`, whole-chain `pending_finish`, the query-cache wrapper forwarding) moved into the join PR as its own first commit. Sort and aggregation sit on the slimmed base and are independent of the join.

be/src/exec/pipeline/spill_process_operator.cpp
```diff
@@ -43,9 +81,12 @@ void SpillProcessOperator::close(RuntimeState* state) {
 
 StatusOr<ChunkPtr> SpillProcessOperator::pull_chunk(RuntimeState* state) {
     if (!_channel->current_task()) {
+        // acquire_spill_task is now non-blocking. An empty queue is only transient (has_output gates us,
+        // and the writer's enqueue notify reschedules us), so return a no-op instead of an error; a
+        // genuinely closed channel shows up through is_finished()/set_finished() upstream.
         bool res = _channel->acquire_spill_task();
         if (!res) {
-            return Status::InternalError("couldn't get task from spill channel");
+            return nullptr;
```
stdpain, line 89: I still don't understand what kind of case would trigger this branch. If has_output is not satisfied, then pull_chunk will not be called.

stdpain:
@eshishki The conflict needs to be resolved.

be/src/exec/pipeline/spill_process_operator.cpp
eshishki: You're right that in steady state `has_output()` gates the pump and the normal enqueue→notify path never lands here. The branch is reachable because `has_output()` and `acquire_spill_task()` read two different things:

- `has_output()` → `has_task()` reads the counter `_task_count`;
- `acquire_spill_task()` reads the queue itself (`_spill_tasks.try_get`),

and the writer updates them in sequence, not atomically together. In `_add_spill_task_locked` the counter is bumped *before* the task lands in the queue:

```cpp
_task_count.fetch_add(1);     // has_task() is already true here
_spill_tasks.put(...);        // queue stays empty until this returns
```

The pump reads neither under the channel mutex, so a pump that re-evaluates `has_output()` concurrently with an in-flight enqueue — e.g. right after `on_current_task_finished()`, or when woken by a spiller flush-completion notify on the source list — can observe `has_task() == true` while `try_get()` still returns empty.

The counter leading the enqueue is intentional: the sink blocks on `has_task()` for `WAIT_CHANNEL`, so the channel has to report "busy" as early as possible. If the counter lagged the `put`, there'd be a window where the task is queued but `has_task()` is false, and the sink could observe the channel as free and finish early. So we favor the sink's correctness and let the pump absorb the transient.

Since `acquire_spill_task()` is now non-blocking (it used to block until a task arrived), an empty queue here is a legal transient rather than an impossible state. `return nullptr` just yields the round with no chunk; the writer's `notify_source_observers()` after the `put` re-drives the pump and the task is then dequeued. The old `return Status::InternalError(...)` assumed the blocking-acquire semantics where an empty queue could only mean shutdown — under non-blocking acquire it would turn a benign race into a spurious query failure.

stdpain (dismissed):

CelerData-Reviewer:
@codex review

stdpain (approved):

PR #74589, opened 2026-06-09
[Enhancement] Eventize the spillable hash join build and probe

eshishki:
## Why I'm doing:

The spillable hash join blocks on spill IO and on its build-side load but subscribes to no wakeup, so the busy-poller spin-polls it (the cost is in #74586, the shared base in #74587). This PR opts the hash join into the event scheduler. The probe is an interior operator, so it is the case the base's `INTERMEDIATE_BLOCK` driver state was added for.

## What I'm doing:

**Stacked on #74587 (base).** Two commits: the first brings the interior-operator support in the driver core (moved out of the base PR per review) — the `INTERMEDIATE_BLOCK` driver state with its park/release checks, `supports_intermediate_wakeup()` on operator and factory with the all-factory event-scheduler gate, the `WAIT_LATCH`/`WAIT_BUILD` block reasons, whole-chain `pending_finish` aggregation, and the query-cache wrapper forwarding (a cache-wrapped spilling probe is reachable — its spill decision is made by the unwrapped build sink — so the wrappers must keep it visible to these aggregations). The second commit is the join wiring itself:

The build is a sink edge: it parks `OUTPUT_FULL` on a full writer (`WAIT_FLUSH`) or a busy spill-process channel (`WAIT_CHANNEL`), subscribes to the spiller's sink list, and shares `SpillableFlatSinkMixin` — same as the agg sinks.

The probe is an interior operator (it sits between its source and sink edges). It declares an interior wakeup unconditionally, so the base classifies it `INTERMEDIATE_BLOCK` and parks it instead of cycling the ready queue — this stops the probe's busy-spin even with the flag off (its per-query driver-schedule count drops from ~3.4M to ~7k).

While parked it names one of four reasons, each woken by its own event:

- `WAIT_BUILD` — waiting for the build to finish its phase; woken by the join builder's observable (the same build-done handoff the non-spill join already uses).
- `WAIT_LATCH` — waiting for the build-side partition load tasks of the current batch; woken when the last load task completes.
- `WAIT_FLUSH` / `WAIT_RESTORE` — spilling / restoring its own probe partitions; woken by the probe-spiller's sink / source lists.

Gated on `enable_spill_join_events` (default false, mutable): with the probe already parked (above), the flag additionally moves it — and the build — off the poller onto events, so a spill/build completion wakes the driver directly instead of the poller re-checking it. Off, the join stays on the poller. Query results are unchanged.

Part of #74586

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

PR #74590, opened 2026-06-09
[Enhancement] Eventize the spillable sort

eshishki:
## Why I'm doing:

Spillable sort (ORDER BY / TOP-N) blocks on spill IO but subscribes to no wakeup, so the busy-poller spin-polls it (the cost is in #74586, the shared base in #74587). This PR opts the sort into the event scheduler.

## What I'm doing:

**Stacked on #74587 (base).**

The partition sort sink is a sink edge: it parks `OUTPUT_FULL` on a full writer (`WAIT_FLUSH`) or a busy spill-process channel (`WAIT_CHANNEL`), subscribes to the spiller's sink list, and shares `SpillableFlatSinkMixin` — same as the agg / join sinks.

The merge sources (single-merge and parallel-merge) park `INPUT_EMPTY` on a per-partition restore (`WAIT_RESTORE`). A sort merges all spilled partitions, so unlike the single-spiller agg source a merge source subscribes to every partition spiller's source list — one subscription per spilled partition — and any partition's restore or flush-all completion wakes it. The parallel-merge source also coordinates its merge stages through the merger's own observable, as the non-spill parallel sort already does.

Gated on `enable_spill_sort_events` (default false, mutable): off, the sort fragment runs on the poller as before; on, the sort drivers are woken by events instead of polled. Query results are unchanged.

Part of #74586

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/exec/pipeline/sort/local_merge_sort_source_operator.h
```diff
@@ -47,6 +48,22 @@ class LocalMergeSortSourceOperator final : public SourceOperator {
 
     bool is_finished() const override;
 
+    // True while any partition spiller still has restore/flush IO in flight. The source subscribes to every
+    // partition spiller's source list (see prepare), so this holds the source driver -- and its observer,
+    // which those spillers hold a raw pointer to -- alive until no in-flight IO can fire a notify against it.
+    // The UAF in the complete_io decrement->notify window is closed by the IO task's query-lifetime guard
+    // (GuardedCompletion runs the completion before releasing the query pin); pending_finish only guards
+    // against a premature park-in-FINISH lost wakeup.
```
stdpain: I don't understand why it's necessary to override the `pending_finish` interface. The current spill task uses a query context guard to ensure the operator hasn't been destructed.

eshishki: You're right — we re-traced it and removed the override (724fbef). The IO task's query-context pin does transitively keep the driver and its observer alive (FragmentContexts are destroyed only by `~QueryContext`, drivers only by `~FragmentContext`), the completion notify runs before the pin is released, and a task that starts after the query died fails the weak-lock in `scoped_begin` and abandons without notifying. So `pending_finish` was redundant on the merge sources; same for the agg sources and the spill-process pump, removed as well.

The overrides that remain are of a different kind: the spillable sinks must drain their spill channel before `close()` (the queued channel tasks capture the operator's own state, e.g. the aggregator's streaming chunks, and are executed later by the spill-process pump), and the join probe waits on build-side load tasks that capture the operator raw. Those are not notify-lifetime guards.

be/src/exec/pipeline/sort/spillable_partition_sort_sink_operator.h
```diff
@@ -31,7 +40,15 @@ class SpillablePartitionSortSinkOperator final : public PartitionSortSinkOperato
 
     void close(RuntimeState* state) override;
 
-    bool need_input() const override { return !is_finished() && !_chunks_sorter->is_full(); }
+    // The sink sleeps OUTPUT_FULL on need_input(): a full writer (spiller is_full(), woken by the
+    // flush-completion defer on the sink list) or the spill-process channel still holding a task (woken by the
+    // channel handshake). The two reasons are read separately -- ChunksSorter::is_full() collapses them into a
+    // single bool, which the SpillableFlatSinkMixin predicate keeps distinct for block_reason().
+    bool need_input() const override { return spill_sink_need_input(); }
+    bool pending_finish() const override { return spill_sink_pending_finish(); }
```
stdpain: we also don't need override pending_finish in sort sink ops.

eshishki: Right — removed. The drain barrier already exists in the operator itself: set_finishing() takes a _sort_context->ref() and the matching unref runs inside the flush-all callback, which execute() queues as the channel's final task, so the sorters stay alive until the channel and the flush IO have drained no matter when the sink closes. The override only duplicated that, so it is gone and the sink keeps just need_input()/block_reason() from the mixin.

One thing we hit while verifying this: the cancelled branch of set_finishing() returns before taking that ref, while _chunks_sorter->cancel() may still queue a task capturing the raw sorter pointer, so on cancellation the sorters can be freed while the spill-process driver still holds that task. The race pre-dates this PR (it is reachable on main under the poller).

stdpain:
@eshishki could u rebase this PR ? I want to push for this PR to be merged.

CelerData-Reviewer:
@codex review

be/src/exec/pipeline/sort/spillable_partition_sort_sink_operator.h
```diff
@@ -69,7 +96,10 @@ class SpillablePartitionSortSinkOperatorFactory final : public PartitionSortSink
     Status prepare(RuntimeState* state) override;
     void close(RuntimeState* state) override;
 
-    bool support_event_scheduler() const override { return false; }
+    // Event-scheduler opt-in, gated on enable_spill_sort_events (default false). This is the only false edge
+    // among the sort fragment's factories (the source and spill-process factories already report true), so
+    // the kill-switch demotes the whole sort fragment back to the poller without a rebuild.
+    bool support_event_scheduler() const override { return config::enable_spill_sort_events; }
```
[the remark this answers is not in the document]
eshishki: Thanks — fixed in `f166c86`. Both sort merge sources now report ready on a partition spiller task error via `SortContext::spiller_task_status()` (`has_output()` returns true, `pull_chunk()` `RETURN_IF_ERROR`s it), so the error propagates instead of the source re-parking `WAIT_RESTORE`. Mirrors the agg/NLJ `RETURN_TRUE_IF_SPILL_TASK_ERROR` + `task_status()` pattern, scanned across sort's N partition spillers.

stdpain (approved):

trueeyu (approved):

PR #74605, opened 2026-06-10
[Enhancement] Record IVM source version/timestamp ranges in MV task run EXTRA_MESSAGE

Youngwb:
## Why I'm doing:

An incremental (IVM) materialized view refresh consumes a specific snapshot range from each base table, but that range is currently only visible in FE logs. Exposing it on the task run makes every refresh auditable via `information_schema.task_runs.EXTRA_MESSAGE` ("which slice of the source did this refresh consume?"), and it is the data source for the `IMV_SOURCE_*` columns of the planned MV refresh-jobs system table.

## What I'm doing:

Record two new JSON sub-fields in the MV task run extra message for every IVM refresh, right after `MVIVMRefreshProcessor` stages the per-table TVR deltas:

- `imvSourceVersionRange`: `{"catalog.db.tbl": {"start": "100", "end": "128"}}` — the TVR version (snapshot id) range consumed from each base table. `TvrVersion` MIN/MAX sentinels are rendered as `"MIN"`/`"MAX"` (readable, instead of leaking `Long.MIN_VALUE`/`Long.MAX_VALUE` literals).
- `imvSourceTimestampRange`: same shape, values are the commit times of those versions in epoch millis. Resolution is delegated to `Table.getVersionCommitTimeMillis(version)` — a polymorphic seam whose base implementation returns empty and which each table format overrides (Iceberg resolves the snapshot's commit time); endpoints that cannot be resolved (sentinel versions, expired/unknown snapshots, formats with no version commit time) are omitted, so a table degrades to `{}` instead of failing the refresh. This keeps `MVIVMRefreshProcessor` format-agnostic, so other IVM-capable table formats only need to override that one method.

Recording goes through the existing best-effort `updateTaskRunStatus()` helper (null-guarded, exception-swallowing), so it can never fail a refresh. Both setters cap map entries with `MvUtils.shrinkToSize`, mirroring `pinnedSnapshotIdMap`. PCT runs leave both fields as `{}`.

Example after an incremental refresh:

```sql
SELECT get_json_string(EXTRA_MESSAGE, '$.imvSourceVersionRange')
FROM information_schema.task_runs ORDER BY FINISH_TIME DESC LIMIT 1;
-- {"iceberg_cat.mv_db.t1":{"start":"4912319873255147742","end":"6987413294356100802"}}
```

Tests:
- FE UT: `MVTaskRunExtraMessageTest` (shrink-to-size + JSON serialization of the new fields), `IVMBasedMvRefreshProcessorIcebergTest#testImvSourceVersionRangeRecordedOnExtraMessage` (range values + degradation to `{}` for unresolvable snapshots, plus the empty base-`Table` resolution), `#testImvSourceTimestampRangeRecordedOnExtraMessage` (commit times against real Iceberg snapshots on the mock catalog), and `#testImvSourceRangesNotRecordedOnPctFallback` (no stale ranges when hybrid falls back to PCT).
- SQL: `test_ivm_task_run_source_range` — PCT baseline run records nothing; incremental run records both ranges.

Fixes #issue: N/A — observability enhancement, no tracking issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
@@ -16,6 +16,7 @@ package com.starrocks.scheduler.mv.ivm;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.starrocks.catalog.BaseTableInfo;
@@ -50,7 +51,9 @@ import com.starrocks.scheduler.TaskRunContext;
 import com.starrocks.scheduler.mv.BaseTableSnapshotInfo;
 import com.starrocks.scheduler.mv.MVRefreshExecutor;
 import com.starrocks.scheduler.mv.MVRefreshProcessor;
+import com.starrocks.scheduler.persist.MVTaskRunExtraMessage;
 import com.starrocks.server.GlobalStateMgr;
+import com.starrocks.server.MetadataMgr;
 import com.starrocks.sql.StatementPlanner;
 import com.starrocks.sql.analyzer.Analyzer;
 import com.starrocks.sql.analyzer.AnalyzerUtils;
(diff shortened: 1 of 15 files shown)


CelerData-Reviewer:
@codex review

test/sql/test_materialized_view_ivm/T/test_ivm_task_run_source_range
```diff
@@ -0,0 +1,45 @@
...
+
+-- Incremental refresh records the consumed snapshot range and its commit times.
+[UC]REFRESH MATERIALIZED VIEW test_mv1 WITH SYNC MODE;
+SELECT count(*) FROM test_mv1;
+SELECT count(*) FROM information_schema.task_runs WHERE TASK_NAME = '${TASK_NAME}' AND get_json_string(EXTRA_MESSAGE, '$.refreshMode') = 'INCREMENTAL' AND EXTRA_MESSAGE LIKE '%"imvSourceVersionRange":{"mv_iceberg_${uuid0}.mv_ice_db_${uuid0}.t_src":{"start":%';
+SELECT count(*) FROM information_schema.task_runs WHERE TASK_NAME = '${TASK_NAME}' AND get_json_string(EXTRA_MESSAGE, '$.refreshMode') = 'INCREMENTAL' AND EXTRA_MESSAGE LIKE '%"imvSourceTimestampRange":{"mv_iceberg_${uuid0}.mv_ice_db_${uuid0}.t_src":{"start":%';
+
+drop materialized view test_mv1;
+drop table mv_iceberg_${uuid0}.mv_ice_db_${uuid0}.t_src force;
+drop database mv_iceberg_${uuid0}.mv_ice_db_${uuid0} force;
```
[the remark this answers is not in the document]
Youngwb: Good catch — fixed in 198cffe3cec: added `drop catalog mv_iceberg_${uuid0};` to both T and R, matching the newer tests in this suite (e.g. `test_ivm_with_iceberg_mv_max_rows_per_refresh`, `test_ivm_with_iceberg_multi_refresh_tvr_advance`).

Youngwb:
@codex review

fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
```diff
@@ -122,6 +126,7 @@ public ProcessExecPlan getProcessExecPlan(TaskRunContext taskRunContext) throws
                 tvrTableSnapshotInfo.setTvrSnapshot(changedVersionRange);
             }
         }
+        recordImvSourceRangesOnTaskRun();
```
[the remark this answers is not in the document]
Youngwb: Confirmed real — reproduced it with a regression test (`testImvSourceRangesNotRecordedOnPctFallback`: stage deltas, fail IVM plan generation, hybrid falls back to PCT) which showed the stale `{start:0,end:2}` left on the run. Fixed in 8108b75ff7d by moving `recordImvSourceRangesOnTaskRun()` to after `prepareRefreshPlan()` succeeds, i.e. recording only once the run is committed to the IVM path; the javadoc now pins that ordering. Side effect: SKIPPED (no-change) runs no longer record ranges, which is consistent — they consume nothing.

Youngwb:
@codex review

Youngwb:
rerun

Youngwb:
Re-run the workflow

fe/fe-core/src/main/java/com/starrocks/catalog/Table.java
```diff
@@ -841,6 +841,14 @@ public boolean isTemporal() {
         return false;
     }
 
+    /**
+     * Commit time in epoch millis of the given TVR version (for lake tables, the snapshot id);
+     * empty when the table format cannot resolve it.
+     */
+    public Optional<Long> getVersionCommitTimeMillis(long version) {
+        return Optional.empty();
```
stephen-shelby: I'd avoid adding getVersionCommitTimeMillis(long version) to the base Table: it's lake-specific (only IcebergTable overrides it, everything else returns empty), long version isn't consistent across formats (OLAP visible-version vs Iceberg snapshot id), and Table otherwise has no version concept. Since the range already comes from TvrVersionRange / TvrTableSnapshotInfo, resolving version→time fits better on a narrower lake/IVM seam rather than the base class.

HangyuanLiu: +1

Youngwb: Agreed on all three points — base `Table` has no version concept, `long` means different things per format (OLAP visible-version vs snapshot id), and only Iceberg resolves it. Moved the resolution off `Table` onto a narrow capability interface `VersionCommitTimeResolver { Optional<Long> getVersionCommitTimeMillis(long) }` (051a5d7): `IcebergTable implements` it, base `Table` is untouched, and `MVIVMRefreshProcessor` gates on `table instanceof VersionCommitTimeResolver` so it stays format-agnostic without hard-binding to a concrete type. The version-space-is-table's-own caveat is documented on the interface. A future IVM-capable format just implements the interface; non-lake tables resolve to `{}` as before.

Youngwb: Reworked to avoid touching the table classes entirely (eb31393): instead of a method on `Table`/`IcebergTable`, commit-time resolution now lives on `ConnectorMetadata.getVersionCommitTimeMillis(dbName, table, version)` — a `default`-empty method alongside the existing `getTableVersionRange`/`getCurrentTvrSnapshot`/`listTableDeltaTraits`, which `IcebergMetadata` overrides. `Table` and `IcebergTable` are now untouched, `MVIVMRefreshProcessor` calls it through `MetadataMgr` with no `instanceof`, and the connector framework dispatches by format. A future IVM-capable format just overrides the same method in its own metadata impl.

stephen-shelby (approved):

HangyuanLiu (approved):

PR #74657, opened 2026-06-10
[BugFix] Heal version chain when cancelling lake index fast-path alter in FINISHED_REWRITING

meegoo:
## Why I'm doing:

Cancelling a lake-table index fast-path `ALTER TABLE ... ADD/DROP INDEX` (handled by `LakeTableIndexFastPathJobBase` on shared-data tables) while the job is in `FINISHED_REWRITING` corrupts the version chain and **hangs all subsequent loads on that table forever**.

On `RUNNING -> FINISHED_REWRITING` the job reserves a commit version `V` for each partition and bumps `nextVersion` to `V+1`, but `V` is **not yet published to BE** and FE `visibleVersion` is still `V-1`. If the job is cancelled in `FINISHED_REWRITING` — which can happen **without any manual force-cancel**, e.g. a publish timeout in `runFinishedRewritingJob` throws `AlterCancelException` and auto-cancels via `run()` — the fast-path cancel releases the table straight back to `NORMAL` without a no-op publish and without advancing `visibleVersion`. This leaves `nextVersion=V+1`, `visibleVersion=V-1`, and no `tablet_metadata_<V>` on BE. Later loads commit at `V+1` whose publish base is the missing `V`, so they can never publish and hang in `COMMITTED`.

`LakeTableIndexFastPathJobBase` is the only lake `AlterJobV2` subclass missing the guard that its siblings `LakeTableAlterMetaJobBase` and `LakeTableSchemaChangeJob` already implement (and which their code comments document as introduced precisely to avoid this measured failure).

## What I'm doing:

Mirror the sibling lake alter jobs' cancel handling in `LakeTableIndexFastPathJobBase`:

- Override the force-aware `cancelImpl(String errMsg, boolean force)`:
  - While in `FINISHED_REWRITING`, a non-force cancel is rejected (`return false`) — the operator must force (`ADMIN SKIP ...`).
  - A force cancel out of `FINISHED_REWRITING` first runs a no-op publish (new `lakePublishVersionWithSkip`) of the reserved version `V` for every partition's `VISIBLE` materialized-index tablets (publish format follows the table's current `isFileBundling()`), then advances `visibleVersion` to `V` via the shared `AlterJobV2.advanceVisibleVersionForForceSkip` inside the locked `persistStateChange` closure, before releasing the table to `NORMAL`. If the no-op publish RPC fails, the job stays in `FINISHED_REWRITING` for operator retry.
- Set `forceSkippedAtCommitted=true` before `persistStateChange` so `copyForPersist` records it and replay on follower FEs reproduces the same version advance, keeping leader and follower FEs consistent.

FE-only change; no on-disk format or interface change.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [x] Policy changes: cancelling a lake index fast-path alter in `FINISHED_REWRITING` now requires force (`ADMIN SKIP ...`); a force cancel performs a no-op publish + version advance to heal the version chain — bringing this job in line with the other lake alter jobs.
- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableIndexFastPathJobBase.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableIndexFastPathJobBase.java
@@ -382,9 +382,52 @@ public abstract class LakeTableIndexFastPathJobBase extends AlterJobV2 {
 
     @Override
     protected boolean cancelImpl(String errMsg) {
+        return cancelImpl(errMsg, false);
+    }
+
+    @Override
+    protected boolean cancelImpl(String errMsg, boolean force) {
         if (jobState.isFinalState()) {
             return false;
         }
+
+        // A job sitting in FINISHED_REWRITING has already reserved a commit
+        // version V per partition and bumped nextVersion to V+1 (see
+        // updateNextVersion at the FINISHED_REWRITING transition), but has NOT
+        // yet published V to BE. Cancelling here without healing the version
+        // chain leaves a hole at V: the table flips back to NORMAL, the next
+        // load commits at V+1, its publish base is V, and BE can never
+        // materialize tablet_metadata_<V> for the discarded alter — so
+        // subsequent loads hang on publish forever. Mirror the sibling lake
+        // alter jobs (LakeTableAlterMetaJobBase, LakeTableSchemaChangeJob):
+        //   - a non-force cancel in FINISHED_REWRITING is REFUSED; the operator
+        //     must opt in via ADMIN SKIP COMMITTED TRANSACTION (force=true);
+        //   - a force cancel performs a no-op publish that writes V (V-1
+        //     content tagged as V) on BE, then advances FE's VisibleVersion to
+        //     V before releasing the table back to NORMAL.
+        boolean tableStillExists = tableExists();
+        if (jobState == JobState.FINISHED_REWRITING && tableStillExists && !force) {
+            return false;
+        }
+
+        boolean advanceVersionForForce = force && jobState == JobState.FINISHED_REWRITING && tableStillExists;
+        if (advanceVersionForForce) {
+            if (!lakePublishVersionWithSkip(errMsg)) {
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableIndexFastPathJobBase.java
```diff
@@ -382,36 +382,166 @@ protected boolean lakePublishVersion() {
...
+        // carry a null (or empty) map, and with no reserved version there is no
+        // version-chain hole to heal — so skip the special handling entirely
+        // (refuse nothing, publish nothing) rather than NPE in
+        // lakePublishVersionWithSkip's commitVersionMap.keySet().
+        boolean hasReservedVersion = commitVersionMap != null && !commitVersionMap.isEmpty();
+        boolean tableStillExists = tableExists();
+        if (jobState == JobState.FINISHED_REWRITING && tableStillExists && hasReservedVersion && !force) {
+            return false;
+        }
+
```
meegoo, line 420: **Dispatch gap: the new force-cancel path is unreachable from SQL**

The only call site that passes `force=true` into an alter job is
`SchemaChangeHandler.cancel()`, and it gates FORCE with:

```java
if (force && !(schemaChangeJobV2 instanceof LakeTableSchemaChangeJobBase
        || schemaChangeJobV2 instanceof LakeTableAlterMetaJobBase)) {
    throw new DdlException(
            "CANCEL ALTER TABLE ... FORCE is only supported on shared-data (lake) tables. ...");
}

luohaha (approved):

kevincai (approved):

PR #74658, opened 2026-06-10
[Enhancement] Allow concurrent ADD PARTITION during metadata-only shared-data ALTER

meegoo:
## Why I'm doing:

Partition creation is blocked or destructive while a shared-data table is under an ALTER that is actually metadata-only and safe to run concurrently with it:

- **Manual `ALTER TABLE ... ADD PARTITION`** is rejected with `Table[...] is under SCHEMA_CHANGE` for any non-NORMAL state.
- **Automatic partition creation during load** *cancels* the running schema-change job — so a load that auto-creates an expression partition kills a user's running lake `ADD INDEX` / `DROP INDEX` job, even though that job has no real conflict with partition creation.
- **The dynamic-partition scheduler** skips its cycle for any non-NORMAL state.

Two ALTER families are provably safe against concurrent partition creation: FSE V2 synchronous add/drop column (transient `UPDATING_META`, set-and-restored inside one WRITE-lock + edit-log) and the lake `ADD/DROP INDEX` fast-path jobs (`LakeTableIndexFastPathJobBase`: owned-tablet snapshot taken once, no table-level shadow meta before FINISHED, idempotent table-level catalog flip, FE-only cancel).

## What I'm doing:

- `AlterJobV2.allowConcurrentPartitionCreation()` — capability method, default `false`; overridden `true` in `LakeTableIndexFastPathJobBase` (covers `LakeTableAddIndexJob`, `LakeTableDropIndexJob`, and the bloom-filter-columns property jobs built on them). Behavioral method, not serialized state.
- `AlterJobMgr.unfinishedAlterJobsAllowConcurrentPartitionCreation(tableId)` — shared lock-free guard helper over the schema-change + rollup handlers; empty job set with a non-NORMAL state is treated as NOT tolerable (anomaly conservatism).
- `Config.enable_concurrent_add_partition_during_alter` — mutable, default `true`, the kill switch.
- Relax the guards: **G1** `AlterJobExecutor.visitAlterTableStatement`, **G2** `FrontendServiceImpl.cancelConflictingAlterJobs` (skip the cancel), **G3** `DynamicPartitionScheduler` (don't skip add-partition).
- Relax the **authoritative** table-state check inside `LocalMetastore.addPartitions` (`checkTableForAddPartitions`), which all three paths converge on. Consistency is still guaranteed by `checkIfMetaChange` under the table WRITE lock; the coarse state check is the only thing skipped, and only for the tolerable states.

Only `ADD PARTITION` is relaxed. `DROP PARTITION`, full-rewrite schema change (`SchemaChangeJobV2` / `LakeTableSchemaChangeJob`), rollup, `OPTIMIZE`, and all non-partition DDL keep today's behavior. Setting the config to `false` restores the legacy behavior at every guard.

> Note vs. the design plan: the plan assumed `LocalMetastore.addPartitions` had no table-state assertion. It does (`checkTableState` at both the pre-lock entry and the under-lock re-check), so relaxing only G1/G2/G3 was insufficient — the `LocalMetastore` relaxation is the linchpin and was added after an e2e test caught it.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


meegoo:
### Shared-data cluster validation ✅

Validated on a fresh shared-data cluster (1 FE + 3 CN), with both FE and CN running this PR's build. The test runs an A/B comparison on the **same** binary, flipping only `enable_concurrent_add_partition_during_alter`, while a lake `ADD INDEX` fast-path job runs concurrently with 80 partition-creating loads (each load auto-creates a distinct expression partition → the G2 `cancelConflictingAlterJobs` path).

| | `ADD INDEX` job | index present | new partitions created during the job |
|---|---|---|---|
| **ON** (default `true`) | **FINISHED** | ✅ | **82** (all 80 concurrent loads succeeded) |
| **OFF** (`false`, legacy) | **CANCELLED** | ❌ | 3 (the first load cancelled the job) |

- **ON:** the index job finishes and the index + all concurrently-created partitions coexist — the headline fix.
- **OFF:** legacy behavior reproduced exactly (the first partition-creating load cancels the running `ADD INDEX` job).

The only variable between the two phases is the config flag, confirming the change is the cause and that the mutable flag is a working kill switch.

FE-side unit coverage: `ConcurrentAddPartitionDuringAlterTest` (helper combinations), `LakeTableIndexFastPathJobTest#testAllowConcurrentPartitionCreation` (capability), and `ConcurrentAddPartitionDuringAlterE2ETest` (UtFrame shared-data e2e of the `UPDATING_META` branch — which is what caught that `LocalMetastore.addPartitions` enforces `checkTableState` at both the pre-lock and under-lock points, the gap the design note missed).

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

sevev (approved):

HangyuanLiu (approved):

PR #74673, opened 2026-06-11
[BugFix] FILES()/LOAD respect Parquet isAdjustedToUTC=false for INT64 timestamps (backport #73674)

ronjay1827:
## Why I'm doing:

Backport of #73674 to `branch-3.5`. The mergify auto-backport (#74646) was closed because the cherry-pick of 1481abcc57ef73aeeec61b6ccd4c77d1a4d805e9 conflicted with this branch's source layout; this PR replaces it with a manually resolved cherry-pick of the same change set.

Backports the fix for #73672 (FILES()/LOAD shifted Parquet INT64 `isAdjustedToUTC=false` timestamps by the session time zone).

## What I'm doing:

Commit 1 is the cherry-pick of #73674 resolved for this branch:

- main moved the Arrow converter to `be/src/column/arrow/`; this branch keeps `be/src/exec/arrow_to_starrocks_converter.cpp`, where the timezone-naive (empty Arrow tz) fallback now resolves to `"UTC"` instead of the session time zone.
- The unconditional session-timezone rectify that #73674 guards lives in `ParquetScanner::convert_array_to_column` (`be/src/exec/parquet_scanner.cpp`) on this branch — the converter-side block from #50448 is not on release branches — so the `isAdjustedToUTC=false` guard is applied there: only rectify the Arrow column time zone when it is already timezone-aware.
- `be/src/exec/parquet_reader.cpp/.h` (top-level INT96 re-tagged as UTC via `_rectify_int96_timezone()` so INT96 keeps the session-timezone shift) and `docs/{en,zh,ja}` `files.md` apply as on main.
- `convert_array_to_column` takes `ColumnPtr&` on this branch (main/branch-4.1 take `Column*`), so the adapted converter test follows this branch's existing call pattern.

Commit 2 adapts the two regression tests from #73674 to this branch's test helpers: `ArrowConvertContext` carries a `RuntimeState*` here (not a `timezone` string), so the converter test injects the session time zone through `RuntimeState(TQueryGlobals.__set_time_zone)` and drives the production entry `ParquetScanner::convert_array_to_column`; the scanner test helper gets the `get_chunk` timezone parameter and the `query_globals.__set_time_zone` fix from #73674, with the `int96_timestamp`/`datetime` baselines refreshed to their true-UTC renders.

Verified locally (BE UT, dev-env-ubuntu:3.5.18): `ArrowConverterTest.test_timestamp_convert_naive_preserves_wall_clock`, `ParquetScannerTest.int96_timestamp_session_timezone_shift`, `ParquetScannerTest.int96_timestamp`, and `ParquetScannerTest.datetime` all pass.

Identical change set is backported to the other release branches — branch-4.1: #74671, branch-4.0: #74672 — only file paths and test-helper details differ per branch.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

### DIFF
--- a/be/src/exec/arrow_to_starrocks_converter.cpp
+++ b/be/src/exec/arrow_to_starrocks_converter.cpp
@@ -685,19 +685,16 @@ struct ArrowConverter<AT, LT, is_nullable, is_strict, DateOrDateTimeATGuard<AT>,
             }
         } else if constexpr (at_is_datetime<AT>) {
             auto timezone = concrete_type->timezone();
-            if (timezone.empty()) {
-                // Quote from https://github.com/apache/arrow/blob/4743e181596b9ee45c6b063bcf59fdf9eb72418f/cpp/src/arrow/type.h#L1217
-
-                /// If a TimestampType is constructed without a timezone (or, equivalently, if the
-                /// timezone supplied is an empty string) then the resulting Arrow field (column) is
-                /// considered "timezone-naive".  The producer of a timezone-naive column may populate
-                /// its constituent integer arrays with datetime values from any timezone; the consumer
-                /// of a timezone-naive column should make no assumptions about the interoperability or
-                /// comparability of the values of such a column with those of any other timestamp
-                /// column or datetime value.
-
-                // When the parquet timezone is empty, populate data with runtime timezone instead.
-                timezone = ctx->state->timezone();
+            const bool is_timezone_naive = timezone.empty();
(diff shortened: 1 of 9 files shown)


kevincai (approved):

trueeyu:

PR #74683, opened 2026-06-11
[Refactor] Move pipeline fragment completion handling to query lifecycle

alvin-phoenix-ai:
## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/pipeline/query_context_manager.cpp
+++ b/be/src/exec/pipeline/query_context_manager.cpp
@@ -23,15 +23,34 @@
 #include "common/config_rpc_client_fwd.h"
 #include "common/system/master_info.h"
 #include "common/thread/thread.h"
+#include "common/thread/threadpool.h"
 #include "common/util/thrift_client_cache.h"
+#include "compute_env/workgroup/pipeline_executor_set.h"
+#include "compute_env/workgroup/work_group.h"
 #include "exec/pipeline/fragment_context.h"
 #include "exec/pipeline/fragment_context_manager.h"
+#include "exec/pipeline/primitives/driver_executor.h"
 #include "exec/pipeline/query_context.h"
+#include "gen_cpp/FrontendService.h"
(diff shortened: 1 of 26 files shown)


be/src/exec/pipeline/query_context_manager.cpp
```diff
@@ -218,6 +237,55 @@ void QueryContextManager::count_down_fragments(QueryContext* query_ctx) {
...
+                        [&](FrontendServiceConnection& client) { client->reportFragmentFinish(res, params); });
+            }
+        };
+
+        auto runnable = std::make_shared<RpcRunnable>(fe_addr, res, params);
+        (void)execution_services(state).streaming_load_thread_pool->submit(runnable);
+    }
+
+    fragment_ctx->destroy_pass_through_chunk_buffer();
+    count_down_fragments(query_ctx.get());
```
stdpain, line 286: I don't understand why the query context needs to be retrieved from the query context manager. I think it would be more logical to put this functionality within the query context itself.

stdpain: Getting the query context is a relatively heavy operation, and we should avoid it as much as possible.

stdpain: I believe the approach regarding the iteration direction after `QueryContext` is to manage their lifecycles in the same way as bthread IDs. We should avoid coupling with the query context manager as much as possible. Currently, it appears that the direction is moving toward tighter coupling with the query context manager, which is contrary to our goal.

PR #74694, opened 2026-06-11
[BugFix] Abort BE vacuum tasks once the FE caller's timeout elapses

starrocks-xupeng:
## Why I'm doing:

The FE gives up waiting for a vacuum RPC after its brpc timeout (`LakeService.TIMEOUT_VACUUM`, 1 hour), marks the partition vacuum as failed and re-dispatches it shortly after. But the BE-side vacuum task is never cancelled: it keeps running as a zombie, occupying one of the few workers of the `RELEASE_SNAPSHOT` thread pool (5 by default) for hours, while nobody reads its response. On clusters with partitions that accumulated a huge number of versions, zombie tasks can exhaust the whole pool: newly dispatched vacuum requests pile up in the queue, vacuum throughput collapses cluster-wide, and the version backlog keeps growing.

## What I'm doing:

- Add `optional int64 timeout_ms = 11` to `VacuumRequest` (`gensrc/proto/lake_service.proto`): the maximum duration the FE caller waits for the request.
- FE: `AutovacuumDaemon#vacuumPartitionImpl` fills `timeoutMs` with `LakeService.TIMEOUT_VACUUM`, the brpc timeout of the vacuum RPC, so the BE deadline matches exactly how long the FE actually waits.
- BE: `LakeServiceImpl::vacuum` anchors an absolute deadline (`butil::gettimeofday_ms() + timeout_ms`) at the time the request is received and passes it to `lake::vacuum` (new `deadline_ms` parameter, default `0` = no deadline, all other callers unchanged).
- BE: `vacuum_impl` checks the deadline once at entry, bringing it forward by `min(5min, 1/10 of the FE timeout)`: a task that already exceeded the deadline while queued — or has so little of its window left that the walk could only time out mid-way without advancing any metadata — aborts before doing any work, instead of occupying a worker and burning FS list QPS on a round that cannot finish. `collect_files_to_vacuum` then checks the exact deadline on each iteration of the version-chain walk (the dominant cost for high-version-count partitions) and aborts with `Status::TimedOut`, freeing the worker. Aborting between walk iterations leaves the metadata chain untouched, so the next vacuum round resumes from the same state.
- BE: new mutable config `lake_vacuum_enable_task_timeout` (default `true`) gates the deadline: when set to `false` the BE ignores `timeout_ms` and vacuum tasks always run to completion.
- Requests without `timeout_ms` (older FE versions) carry no deadline and run to completion, exactly as before.
- UT: `test_vacuum_deadline_expired_mid_walk` (deadline expires mid-walk via a mocked clock on the new `vacuum:check_deadline` sync point: nothing is deleted, and a follow-up run without deadline converges normally), `test_vacuum_deadline_window_too_small_to_start` (a task that starts with less than the minimum useful window left aborts at entry without walking the version chain), `test_vacuum_task_deadline_exceeded` (handler threads `timeout_ms` into the task and returns `TIMEOUT`; a request without the field is unaffected, and so is any request when `lake_vacuum_enable_task_timeout` is off), and `testVacuumRequestCarriesTimeout` (FE fills the field).

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/storage/lake/vacuum.cpp
+++ b/be/src/storage/lake/vacuum.cpp
@@ -17,6 +17,7 @@
 #include <butil/fast_rand.h>
 #include <butil/time.h>
 #include <bvar/bvar.h>
+#include <fmt/format.h>
 
 #include <optional>
 #include <set>
@@ -174,6 +175,23 @@ bool should_retry(const Status& st, int64_t attempted_retries) {
     }
     auto message = st.message();
     return MatchPattern(message, config::lake_vacuum_retry_pattern.value());
+}
+
+// Returns Status::TimedOut once |deadline_ms| (milliseconds since the Epoch) has passed.
+// deadline_ms <= 0 means no deadline. The deadline is anchored at the time the BE received
+// the vacuum request, so it also expires for tasks that waited too long in the thread pool
+// queue: by then the FE caller has given up waiting and would re-dispatch the partition,
+// continuing would only keep a vacuum worker occupied for a response nobody reads.
+Status check_vacuum_deadline(int64_t deadline_ms) {
+    if (deadline_ms <= 0) {
+        return Status::OK();
+    }
+    int64_t now_ms = butil::gettimeofday_ms();
+    TEST_SYNC_POINT_CALLBACK("vacuum:check_deadline", &now_ms);
+    if (now_ms >= deadline_ms) {
+        return Status::TimedOut(fmt::format("vacuum task deadline exceeded, now={}, deadline={}", now_ms, deadline_ms));
+    }
+    return Status::OK();
 }
 
 Status delete_files_with_retry(FileSystem* fs, std::span<const std::string> paths) {
@@ -417,7 +435,7 @@ static Status collect_files_to_vacuum(TabletManager* tablet_mgr, std::string_vie
                                       AsyncFileDeleter* datafile_deleter, AsyncFileDeleter* metafile_deleter,
                                       AsyncSharedFileDeleter* shared_file_deleter, int64_t* total_datafile_size,
                                       int64_t* vacuumed_version, int64_t* extra_datafile_size,
-                                      const TabletRetainInfo& retain_info) {
+                                      const TabletRetainInfo& retain_info, int64_t deadline_ms) {
     auto t0 = butil::gettimeofday_ms();
     auto meta_dir = join_path(root_dir, kMetadataDirectoryName);
     auto data_dir = join_path(root_dir, kSegmentDirectoryName);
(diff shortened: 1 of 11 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
```diff
@@ -318,6 +318,10 @@ private void vacuumPartitionImpl(Database db, OlapTable table, PhysicalPartition
             vacuumRequest.deleteTxnLog = needDeleteTxnLog;
             vacuumRequest.enableFileBundling = fileBundling;
             vacuumRequest.enableSharedFileCleanup = enableSharedFileCleanup;
+            // The longest this FE waits for the response (the brpc timeout of the vacuum RPC).
+            // The BE checks it periodically during execution and aborts the task once it has
+            // elapsed, instead of running on as a zombie that no caller is waiting for.
+            vacuumRequest.timeoutMs = LakeService.TIMEOUT_VACUUM;
```
xiangguangyxg, line 324: better to make vacuum timeout configurable

kevincai: this is the BRPC framework issue, where the rpc time is injected during the auto-code generation.

be/src/storage/lake/vacuum.cpp
```diff
@@ -808,6 +830,9 @@ Status vacuum_impl(TabletManager* tablet_mgr, const VacuumRequest& request, Vacu
     if (UNLIKELY(request.grace_timestamp() <= 0)) {
         return Status::InvalidArgument("value of grace_timestamp is zero or nagative");
     }
+    // The task may have stayed in the thread pool queue long enough that the FE caller
+    // already timed out and gave up, abort without doing any work in that case.
+    RETURN_IF_ERROR(check_vacuum_deadline(deadline_ms));
```
kevincai: Try give some buffer check about the deadline before actually launching the task, e.g. if remain time window to the deadline is less than 5 mins or 1/10 of the timeout, abort it at the earliest. Don't waste the walk of the meta and then aborted for timeout without doing anything.

be/src/storage/lake/vacuum.cpp
```diff
@@ -437,6 +455,9 @@ static Status collect_files_to_vacuum(TabletManager* tablet_mgr, std::string_vie
     // Starting at |*final_retain_version|, read the tablet metadata forward along
     // the |prev_garbage_version| pointer until the tablet metadata does not exist.
     while (version >= min_version) {
+        if (auto st = check_vacuum_deadline(deadline_ms); !st.ok()) {
+            return Status::TimedOut(fmt::format("{} tablet_id={}", st.message(), tablet_id));
```
kevincai, line 459: there are AsyncDeleter running in background, need to consider how to deal with the inflight deletes, This TimedOut returned to the caller who will RETURN_IF_ERROR immediately. Double check if it has problems.

starrocks-xupeng: it's safe, even without this change, the situation will still happen

be/src/storage/lake/vacuum.cpp
starrocks-xupeng: done

kevincai (approved):

xiangguangyxg (approved):

wangruin (approved):

PR #74699, opened 2026-06-11
[Enhancement] Reduce copies and allocations in spill restore path

eshishki:
## Why I'm doing:

Spill restore performs avoidable allocations and copies on every restored record:

1. Each record read from disk is copied twice before decoding: `ColumnarSerde::deserialize` does `serialize_buffer.resize(attachment_size)` (allocation) + `read_fully` (memcpy from the reader buffer) and only then decodes.
2. A fresh `SerdeContext` is created on every IO task submission and every yield re-submission, so its serialize buffer regrows from scratch each time.
3. In-memory mem-table chunks are deep-cloned (`clone_unique`) when served through `RawChunkInputStream`, even when the stream is the sole consumer. This path is hot: with default mem-table size, unflushed partition tails account for ~8% of restored rows on a spilled shuffle join, and it also serves the partition-split (recursive repartition) read.

## What I'm doing:

- `BlockReader::read_view(fallback, count)`: serve restore reads as zero-copy views into the internal read buffer when possible; fall back to copying into the caller's buffer otherwise. `ColumnarSerde::deserialize` consumes the view directly (encode levels via `UNALIGNED_LOAD32` since the view base can be unaligned, negative/short `attachment_size` rejected).
- Keep one `SerdeContext` in `SpillIOTaskContext` for the lifetime of an IO task (restore and flush sides), instead of recreating it per (re)submission. The buffer now lives as long as the IO task — bounded by `io_tasks_per_scan_operator` restore tasks plus one flush task per spiller.
- `MemTableReadMode::{SHARED, EXCLUSIVE}` for `UnorderedMemTable::as_input_stream`: a single-consumer read (`read_shared=false`) hands chunks over by move; `RawChunkInputStream` returns a chunk without cloning only when its whole column tree is exclusively owned (`ExclusiveOwnershipChecker`, conservative for unknown column types). Multi-reader topologies (broadcast, NL join) keep the copying SHARED mode. Mem tracker consumption is retained until reset/destruction so spill memory is never understated; a second exclusive acquire returns `InternalError` instead of silently losing rows.
- Wire up the `RowsRestoreFromMemTable` / `BytesRestoreFromMemTable` profile counters: they were registered but never incremented (always 0); `BytesRestoreFromMemTable` unit fixed from `UNIT` to `BYTES`. They now report the in-memory restore volume.
- Remove dead `SerdeContext&` parameters from the partition spill/split internals.

Measured (200M⋈200M spilled shuffle join, `spill_mode=force`, single-node A/B on one host, baseline = main + counters only):

| | baseline | patched | Δ |
|---|---|---|---|
| wall (5 runs) | 13.89 s | 13.76 s | within noise |
| query allocated memory | 325.8 GB | 310.3 GB | **−15.5 GB (−4.8%)** |
| with forced partition splits (4 MB mem tables) | 892.3 GB | 867.7 GB | **−24.6 GB (−2.8%)** |

Breakdown of the −15.5 GB: ~10.4 GB from `read_view` (= `BytesRestoreFromLocalDisk`, the volume previously copied into `serialize_buffer`), ~1 GB from the mem-table move (= `BytesRestoreFromMemTable`, the volume previously cloned), the remainder from `SerdeContext` reuse. The same volume of memcpy traffic is eliminated alongside the allocations. Wall time is unchanged on a single query (the bottleneck is flush/IO); the win is allocator and memory-bandwidth pressure. Query results are byte-identical with spill on/off and across all runs.

A moved chunk transfers `extra_data`/schema/slot maps verbatim, while `clone_unique` rebuilds them; both are correct for spill consumers.

Tests: 4 new cases in `be/test/compute_env/spill/spill_test.cpp`: `read_view` buffered/boundary/fallback/EOF/negative-count paths golden-checked against `read_fully`; move-vs-clone including a unique `NullableColumn` wrapping a shared child (must clone); SHARED/EXCLUSIVE read modes incl. tracker retention and counter values; double exclusive acquire returns an error. `compute_env_test`: 104/104 passed.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/compute_env/spill/block_manager.h
+++ b/be/src/compute_env/spill/block_manager.h
@@ -17,6 +17,7 @@
 #include <memory>
 #include <optional>
 
+#include "base/container/raw_container.h"
 #include "base/metrics.h"
 #include "base/string/slice.h"
 #include "common/runtime_profile.h"
@@ -105,11 +106,24 @@ public:
     // if the Block has reached the end, should return EndOfFile status
     virtual Status read_fully(void* data, int64_t count);
 
+    // read exactly `count` bytes like read_fully, but prefer returning a
+    // zero-copy view into the internal read buffer. When the data cannot be
+    // served from the buffer (buffer read disabled or `count` exceeds the
+    // buffer capacity), the bytes are copied into `fallback` and the returned
+    // slice points into it. The returned slice is invalidated by any
+    // subsequent read on this reader and by any other use of `fallback`, so
+    // it must be fully consumed before either is touched again and must not
+    // be stored.
+    [[nodiscard]] StatusOr<Slice> read_view(raw::RawString* fallback, int64_t count);
+
     virtual std::string debug_string() = 0;
 
     virtual const Block* block() const = 0;
 
 protected:
+    Status _init_readable();
+    void _update_io_metrics(int64_t io_ns, int64_t read_len);
+
     const Block* _block = nullptr;
     std::unique_ptr<io::InputStreamWrapper> _readable;
     size_t _length = 0;
--- a/be/src/compute_env/spill/block_reader.cpp
+++ b/be/src/compute_env/spill/block_reader.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstring>
+
 #include "base/string/slice.h"
 #include "common/statusor.h"
 #include "compute_env/spill/block_manager.h"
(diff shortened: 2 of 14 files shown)


CelerData-Reviewer:
@codex review

PR #74702, opened 2026-06-11
[Enhancement] Derive guard predicates from multi-branch OR to enable expression short-circuit

dirtysalt:
## Why I'm doing:

Multi-branch OR queries where each branch shares a common pivot column (e.g. `l_shipdate`) cause the BE to union ALL predicate columns into `active_columns`. With 10 branches each using different payload column pairs, nearly the entire table is decoded upfront — late materialization degenerates into eager full-table scan.

```
Query shape:
  (l_shipdate BETWEEN '1998-01-01' AND '1998-01-31' AND (l_discount > 0.09 OR l_tax > 0.07))
  OR (l_shipdate BETWEEN '1997-11-01' AND '1997-11-30' AND (l_quantity < 10 OR l_extendedprice > 90000))
  OR ... (8 more branches, each with different column pairs)

BE behavior without guard:
  active_columns = UNION(all predicate columns) = 14/16
  → all 14 columns decoded for all rows before expression evaluation
  → ColumnReadTime dominates (19.4s), narrow IO ≈ wide IO (189.9 GB vs 191.7 GB)
```

The guard predicate makes the pivot column (`l_shipdate`) a single-slot conjunct in `conjunct_ctxs_by_slot`, which the BE evaluates first. This enables short-circuit: rows that don't match any branch's date range skip the full OR expression evaluation.

## What I'm doing:

Add `DeriveGuardPredicateRule` — a one-pass transformation called from `PushDownPredicateScanRule` that extracts a "guard" predicate from OR branches sharing a common pivot column.

**Transformation:**

```
Before:
  (shipdate_range_1 AND branch_1_cols)
  OR (shipdate_range_2 AND branch_2_cols)
  OR ...

After:
  (shipdate_range_1 OR shipdate_range_2 OR ...)   -- guard (single-slot, enters conjunct_ctxs_by_slot)
  AND ((shipdate_range_1 AND branch_1_cols)        -- original OR preserved
    OR (shipdate_range_2 AND branch_2_cols)
    OR ...)
```

The guard is purely additive (`AND(guard, original)`), so correctness is preserved.

**Guard eligibility rules:**

- Range predicates (GE, LE, GT, LT): always collected
- IN predicates: always collected
- EQ on numeric/date/datetime: collected (high-cardinality, selective)
- EQ on VARCHAR/CHAR/BOOLEAN: excluded — low-cardinality payload columns should stay LAZY

**Two defensive guardrails:**

1. **EQ-only guard skip** — If a column's guard would consist purely of `=`/`IN` predicates (no range predicates), skip it. `ScalarRangePredicateExtractor` already produces equivalent `col IN (...)` lists. Not a compromise.

2. **Minimum column threshold (≥ 3 distinct columns)** — Skip guard when the OR involves ≤ 2 columns. With only 2 columns both are already ACTIVE regardless of the guard, so short-circuit benefit is marginal. The real reason: guards added during MV refresh are baked into the MV's stored plan, and MV matching (which is not guard-aware) fails when predicate structures differ between query and stored plan. This threshold preserves the guard for the target production use case (14 columns across 10 branches), and the 3-column SELECTIVE case still triggers the guard. **Compromise — MV matching is not guard-aware yet.**

**Architecture:**

The rule is a standalone class (`static` methods), not a `ScalarOperatorRewriteRule` inside the fixpoint loop. It's called once from `PushDownPredicateScanRule`, after `ScalarRangePredicateExtractor` and the scalar rewriter complete. This avoids infinite rewrite loops where the guard's output was fed back into the rewriter.

**Known limitation — IO not reduced:**

The guard reduces CPU (expression short-circuit) but does not reduce IO — the BE still marks all branch-local columns as ACTIVE because they appear in the remaining OR expression. Reducing IO requires BE-side lazy column loading during expression evaluation, tracked separately. This FE-side guard is the prerequisite that provides the column classification signal.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/DeriveGuardPredicateRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/DeriveGuardPredicateRule.java
@@ -34,8 +34,8 @@ import java.util.Set;
 /**
  * Derive guard predicates from OR expressions where every branch contains
  * a comparison on a common pivot column.  The guard is a single-slot predicate
- * that enters conjunct_ctxs_by_slot in the BE, making the pivot column ACTIVE
- * while branch-local columns remain LAZY.
+ * that enters conjunct_ctxs_by_slot in the BE, enabling early evaluation and
+ * expression short-circuit on the pivot column before the full OR is evaluated.
  * <p>
  * This is a one-pass transformation (like ScalarRangePredicateExtractor),
  * NOT a ScalarOperatorRewriteRule.  It is called directly from
@@ -299,6 +299,12 @@ public class DeriveGuardPredicateRule {
      * </ul>
      */
     private static boolean isGuardCandidatePredicate(ScalarOperator op) {
+        // Nondeterministic predicates (e.g. rand(), uuid()) must not be
+        // duplicated — the guard and original expression would evaluate
+        // different values, potentially filtering rows incorrectly.
+        if (Utils.hasNonDeterministicFunc(op)) {
+            return false;
+        }
         if (op instanceof BinaryPredicateOperator) {
             return isGuardCandidateBinaryPredicate((BinaryPredicateOperator) op);
         }
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/DeriveGuardPredicateRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/DeriveGuardPredicateRule.java
@@ -14,6 +14,7 @@
 
 package com.starrocks.sql.optimizer.rewrite.scalar;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.starrocks.sql.ast.expression.BinaryType;
 import com.starrocks.sql.optimizer.Utils;
 import com.starrocks.sql.optimizer.base.ColumnRefSet;
@@ -21,7 +22,6 @@ import com.starrocks.sql.optimizer.operator.scalar.BinaryPredicateOperator;
 import com.starrocks.sql.optimizer.operator.scalar.CompoundPredicateOperator;
 import com.starrocks.sql.optimizer.operator.scalar.InPredicateOperator;
 import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
-import com.starrocks.sql.optimizer.rewrite.ScalarOperatorRewriteContext;
 import com.starrocks.type.Type;
 
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

satanson (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/DeriveGuardPredicateRule.java
```diff
@@ -0,0 +1,346 @@
...
+        // Nondeterministic predicates (e.g. rand(), uuid()) must not be
+        // duplicated — the guard and original expression would evaluate
+        // different values, potentially filtering rows incorrectly.
+        if (Utils.hasNonDeterministicFunc(op)) {
+            return false;
+        }
+        if (op instanceof BinaryPredicateOperator) {
+            return isGuardCandidateBinaryPredicate((BinaryPredicateOperator) op);
+        }
+        if (op instanceof InPredicateOperator) {
```
stephen-shelby, line 311: `InPredicateOperator` guard candidates do not exclude `NOT IN` / subquery IN.

stephen-shelby:
I have a few concerns about the derived guard predicate here:

1. The generated guard is appended as a normal predicate: `guard AND original_or`. Since the guard is logically implied by the original OR, it should probably be marked as derived/redundant, similar to predicates produced by `ScalarRangePredicateExtractor`. Otherwise, downstream logic may treat it as a real user predicate: stats estimation can apply the same selectivity twice, and MV rewrite / predicate split may include the guard in matching predicates.

2. The guard subtree reuses predicate nodes from the original OR branches. `colPreds` are collected from the original expression tree and then passed into `Utils.compoundAnd(colPreds)`, while the original OR is also kept in the final expression. Since `ScalarOperator` is mutable and carries flags such as redundant / notEvalEstimate, sharing the same node between the guard and the original predicate can be fragile. If we mark the guard as derived/redundant or a later rule mutates it, it may accidentally affect the original OR predicate. Could we clone the guard subtree, or at least ensure only the top-level generated guard node is marked/mutated?

3. Should this apply to all scan types in `PushDownPredicateScanRule.SUPPORT`? The benefit seems to depend on BE-side column loading / lazy materialization / expression short-circuit behavior. For scan types like schema/meta/view/table-function scans, the guard may not provide the intended benefit and only adds extra predicate complexity/evaluation cost. Maybe we should restrict it to scan types that can actually benefit from this guard, or add a session/config gate before enabling it broadly.

stephen-shelby (dismissed):

PR #74729, opened 2026-06-12
[Enhancement] Support property.kafka_partition_discovery for routine load

wyb:
## Why I'm doing:

  By default, specifying `kafka_partitions` in `CREATE ROUTINE LOAD` pins the
  consumed partition list, so partitions added to the Kafka topic later are
  never consumed. Since `kafka_offsets` requires `kafka_partitions`, seeding
  exact starting offsets (e.g. when recreating a job to continue from
  recorded progress) always disables partition auto-discovery.

## What I'm doing:

  Add a new data source property "property.kafka_partition_discovery":

  - CREATE: when set to "true" together with kafka_partitions/kafka_offsets,
    the listed partitions only seed the starting offsets; the job keeps
    fetching the partition list from the broker, so newly added partitions
    are consumed automatically. The seed partitions are validated against
    the broker once at creation time: they are never pinned, so the
    prepare()-time check would not cover them, and a mistyped partition
    would otherwise be accepted and its seeded offset silently discarded.
  - ALTER: setting the property to "true" unpins the partition list of an
    existing job while keeping its progress, and offsets may be specified
    for partitions beyond the pinned set (validated against the broker).
    Setting it back to "false" is rejected; recreate the job to pin again.
  - The flag is derived from the persisted customProperties (same pattern
    as kafka_default_offsets): no metadata format change, and edit-log
    replay stays deterministic across leader and followers.
  - The value must be "true" or "false"; at CREATE the property is only
    accepted together with kafka_partitions.

  Behavior change: the default starting offset for partitions discovered
  after the job already has consuming progress (i.e. partitions added to
  the topic mid-flight) is now OFFSET_BEGINNING instead of OFFSET_END when
  property.kafka_default_offsets is not set, so messages produced during
  the discovery interval are not silently skipped. This applies to all
  auto-discovery jobs, not only partition-discovery ones; a job's initial
  discovery keeps the start-from-latest default, and jobs that explicitly
  set property.kafka_default_offsets are unaffected.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/load/routineload/KafkaRoutineLoadJob.java
```diff
@@ -713,6 +722,13 @@ public static KafkaRoutineLoadJob fromCreateStmt(CreateRoutineLoadStmt stmt) thr
                 db.getId(), tableId, stmt.getKafkaBrokerList(), stmt.getKafkaTopic());
         kafkaRoutineLoadJob.setOptional(stmt);
         kafkaRoutineLoadJob.checkCustomProperties();
+        if (stmt.isKafkaPartitionDiscovery()) {
+            // the seed partitions are never pinned into customKafkaPartitions, so the
+            // prepare()-time check skips them; validate once against the broker here,
+            // otherwise a mistyped partition would silently discard its seeded offset
+            kafkaRoutineLoadJob.checkCustomPartition(stmt.getKafkaPartitionOffsets().stream()
+                    .map(p -> p.first).collect(Collectors.toList()));
+        }
```
banmoy, line 734: **Create-time seed validation runs with `DEFAULT_RESOURCE`, not the job's warehouse (shared-data)**

This `checkCustomPartition` runs inside `fromCreateStmt`, which executes *before* `prepare()`. The compute resource is only acquired in `prepare()`:

```java
// RoutineLoadJob#prepare()  ("call before first scheduling")
this.computeResource = warehouseManager.acquireComputeResource(
        CRAcquireContext.of(this.warehouseId, this.computeResource));
```

and `RoutineLoadMgr.createRoutineLoadJob()` only calls `fromCreateStmt()` + `addRoutineLoadJob()`, never `prepare()`. So here `computeResource` is still `WarehouseManager.DEFAULT_RESOURCE` (the field initializer), even though `warehouseId` was already set in `setOptional()`.

`getAllKafkaPartitions` routes by `computeResource`, not `warehouseId`:

```java
// KafkaUtil#genPKafkaLoadInfo
if (RunMode.isSharedDataMode()) {
    kafkaLoadInfo.warehouseId = computeResource.getWarehouseId();  // == DEFAULT here
}
```

and the proxy selects nodes from that warehouse, throwing if none are alive:

```java
// KafkaUtil#sendProxyRequest
if (nodeIds.isEmpty()) {
    throw new LoadException("...No alive backends or compute nodes");
}
```

**Impact (shared-data only):** for a job created with `PROPERTIES("warehouse" = "<non-default>")`, this create-time validation is routed to the **default** warehouse instead of the job's warehouse. If the default warehouse has no live CN (common in multi-warehouse deployments), `CREATE ROUTINE LOAD ... "property.kafka_partition_discovery" = "true"` fails even though the target warehouse is healthy. A normal pinned-partition job is unaffected, because its `checkCustomPartition` runs inside `prepare()` *after* `acquireComputeResource()`. Even when the default warehouse has live CNs, per-warehouse network/ACL isolation could make the validation result wrong. Shared-nothing mode is unaffected.

Note: the existing tests don't catch this because `getAllKafkaPartitions` is mocked.

wyb: fixed

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

banmoy (approved):

gengjun-git (approved):

srlch (approved):

PR #74730, opened 2026-06-12
[BugFix] Fix MV refresh for joins of same-named base tables across databases

Youngwb:
## Why I'm doing:

A materialized view whose query joins two distinct base tables that share the same **unqualified name** across different databases/catalogs (e.g. `db_a.tj JOIN db_b.tj`) fails to refresh. `CREATE` and the initial baseline refresh are accepted, but every subsequent refresh aborts with `java.lang.IllegalStateException: Duplicate key tj`, leaving the MV stale. First observed on an INCREMENTAL (IVM) iceberg MV, but the same collision affects any partitioned MV with such a join — including native OLAP base tables.

## What I'm doing:

The refresh path identified base tables by their **unqualified name** in three places, which collide when two distinct base tables share a name:

- **IVM** (`MVIVMRefreshProcessor.bindBaseTableTvrVersionRanges`): the per-base-table TVR delta map was keyed by table name (`Collectors.toMap`, no merge) → crash, and the subsequent lookup mis-assigned the delta.
- **PCT pushdown** (`MVPCTRefreshPlanBuilder`): relations were grouped and the to-refresh partition map looked up by `Table.getTableIdentifier()`, which returns the **bare name for OLAP tables**, so two same-named tables were conflated and partition predicates pushed onto one relation only.
- **PCT scope** (`PCTRefreshScopeCalculator.buildScope`): the diagnostic `RefBasePartitionsToRefreshMap` was built with `Collectors.toMap` keyed by name → crash (hit first, by the IVM hybrid baseline refresh).

Fix: the execution paths (IVM delta bind, PCT partition pushdown) now key base tables by the **base-table object**. Object identity is unique per base table (OLAP by table id, external by catalog+uuid), stable across the snapshot/relation boundary, and matches the existing `mv.getRefBaseTablePartitionSlots()` keying right beside the old lookup. A self-join (one identity) is still correctly distinguished from two same-named tables (two identities). The `buildScope` diagnostic stays keyed by table name; same-named tables now **merge** in that map instead of throwing — it is diagnostic-only, so refresh correctness is unaffected and the existing name-keyed format is preserved.

The fix applies uniformly across base-table types: native OLAP (keyed by table id) and external iceberg/hive/hudi/delta (keyed by `catalog + db + stable identifier`, i.e. `name:uuid` for iceberg/delta, `name:createTime` for hive/hudi). Validated end-to-end on a dev cluster — the native-OLAP and Hive same-name-join PCT MVs refresh to 2 rows initially and 3 after appending to both base tables (previously aborted). Added FE unit tests for the IVM bind and the PCT scope (including the same-name merge), a native-OLAP PCT regression, and SQL-Tester e2e cases (iceberg IVM / iceberg PCT / iceberg self-join / native-OLAP PCT / Hive PCT) consolidated in one file. Commits are split by component (IVM / PCT / e2e).

Fixes #issue: N/A — reported via an internal tracker.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/ivm/MVIVMRefreshProcessor.java
@@ -65,6 +65,7 @@ import com.starrocks.sql.optimizer.rule.transformation.materialization.MvUtils;
 import com.starrocks.sql.plan.ExecPlan;
 import org.apache.commons.collections4.CollectionUtils;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -554,30 +555,36 @@ public final class MVIVMRefreshProcessor extends MVRefreshProcessor {
     private InsertStmt buildInsertPlan(InsertStmt insertStmt) throws AnalysisException {
         QueryStatement queryStatement = insertStmt.getQueryStatement();
         Multimap<String, TableRelation> tableRelations = AnalyzerUtils.collectAllTableRelation(queryStatement);
-        Map<String, TvrVersionRange> baseTableNameToTvrVersionRangeMap = snapshotBaseTables.values()
-                .stream()
-                .map(snapshotInfo -> (TvrTableSnapshotInfo) snapshotInfo)
-                .map(snapshotInfo -> {
-                    BaseTableInfo baseTableInfo = snapshotInfo.getBaseTableInfo();
-                    TvrVersionRange tvrVersionRange = snapshotInfo.getTvrSnapshot();
-                    if (tvrVersionRange == null) {
-                        throw new SemanticException("Base table %s.%s does not have a valid tvr version range",
-                                baseTableInfo.getDbName(), baseTableInfo.getTableName());
-                    }
-                    return Maps.immutableEntry(baseTableInfo.getTableName(), tvrVersionRange);
-                })
-                .collect(Collectors.toMap(entry -> entry.getKey(), Map.Entry::getValue));
+        bindBaseTableTvrVersionRanges(snapshotBaseTables.values(), tableRelations);
+        return insertStmt;
+    }
+
+    @VisibleForTesting
+    static void bindBaseTableTvrVersionRanges(Collection<BaseTableSnapshotInfo> snapshotInfos,
+                                              Multimap<String, TableRelation> tableRelations) {
+        // Key by the base-table object, not its unqualified name: two same-named tables from different
+        // databases have distinct table identity, while a self-join collapses to one entry.
+        Map<Table, TvrVersionRange> tvrRangeByTable = Maps.newHashMap();
+        for (BaseTableSnapshotInfo snapshotInfo : snapshotInfos) {
+            TvrTableSnapshotInfo tvrInfo = (TvrTableSnapshotInfo) snapshotInfo;
+            TvrVersionRange tvrVersionRange = tvrInfo.getTvrSnapshot();
+            if (tvrVersionRange == null) {
+                BaseTableInfo baseTableInfo = tvrInfo.getBaseTableInfo();
+                throw new SemanticException("Base table %s.%s does not have a valid tvr version range",
+                        baseTableInfo.getDbName(), baseTableInfo.getTableName());
+            }
+            tvrRangeByTable.put(tvrInfo.getBaseTable(), tvrVersionRange);
+        }
         for (Map.Entry<String, TableRelation> entry : tableRelations.entries()) {
             TableRelation tableRelation = entry.getValue();
             Table table = tableRelation.getTable();
-            if (!baseTableNameToTvrVersionRangeMap.containsKey(table.getName())) {
+            TvrVersionRange tvrVersionRange = tvrRangeByTable.get(table);
+            if (tvrVersionRange == null) {
                 throw new SemanticException("Base table %s.%s is not found in the changed version ranges",
                         tableRelation.getName().getDb(), tableRelation.getName().getTbl());
             }
-            TvrVersionRange tvrVersionRange = baseTableNameToTvrVersionRangeMap.get(table.getName());
             tableRelation.setTvrVersionRange(tvrVersionRange);
         }
-        return insertStmt;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


CelerData-Reviewer:
@codex review

HangyuanLiu (commented):
Thanks for the fix — keying by identifier is the right direction and it's solid for the Iceberg/external-table case (self-join behavior and the unit tests' discriminating power both check out). One concern on completeness, plus a few smaller items.

### 🔴 No-op for native (OLAP) tables — same-name cross-DB join still crashes

`buildScope` keys by `entry.getKey().getBaseTable().getTableIdentifier()` (PCTRefreshScopeCalculator.java:77). `getBaseTable()` returns the `Table`, and `Table.getTableIdentifier()` is just `return name` (Table.java:298) — OlapTable/MaterializedView don't override it. So a partitioned native MV over `db1.orders a JOIN db2.orders b` keys both bases to `"orders"`, and `Collectors.toMap` (no merge fn) still throws the same `Duplicate key`. Both tables do become ref base tables (matched by bare name via `getMvPartitionExpr` → `areTableNamesEqual`), so it's reachable. It's pre-existing rather than a regression, but the "affects any partitioned MV" wording doesn't hold for native tables.

The unique id is one method away — on `BaseTableSnapshotInfo`, `getBaseTableInfo().getTableIdentifier()` returns `String.valueOf(tableId)` for native tables (unique) and `...:uuid` for external, whereas the currently-used `getBaseTable().getTableIdentifier()` degrades to bare `name` for native.

⚠️ But it can't be a one-line change: the consumer `MVPCTRefreshPlanBuilder` looks up by `relation.getTable().getTableIdentifier()`, where only the `Table` is available (no `BaseTableInfo`). Switching only `buildScope` to `getBaseTableInfo()` would make the producer/consumer keys diverge → partition predicates silently stop pushing down instead of crashing. Both sides need a single key that is (a) unique for native tables and (b) reachable from a bare `Table` on the lookup side — e.g. `table.getId()` on both (please verify external-table id stability first), or thread the snapshot/`BaseTableInfo` through to the lookup.

### Smaller items

- **IVM bind has the same producer/consumer split** (MVIVMRefreshProcessor.java): the map is built from `baseTableInfo.getTableIdentifier()` (:576 → tableId) but probed with `table.getTableIdentifier()` (:582 → name). These are equal only for Iceberg (`name:uuid` on both sides) — safe under the current Iceberg-only gate, but it will break the day IVM accepts native tables with `Base table ... is not found in the changed version ranges`.
- **Trace map not re-keyed**: `mvPlanBuildMessage` is still keyed by `table.getName()`; two same-named tables now both reach push-down and their trace entries overwrite each other, so the persisted PlanBuilderMessage shows only one table's predicate.
- **`RefBasePartitionsToRefreshMap`** in refresh-job history / "Final Plan" logs now carries the identifier (`tj:<uuid>` for Iceberg) instead of the table name for every partitioned MV — minor, but a user-visible diagnostic change despite the "No behavior change" box.
- **null-table check tightened**: moved to an unconditional pre-loop throw over every relation (was: only ref base tables, after the `containsKey` continue). Reachability is low post-analysis, but it also makes the adjacent pre-existing comment ("handled later in the ref-base-table path") inaccurate.

### Test

A native-table case mirroring `test_pct_mv_with_iceberg_same_name_join` (two `default_catalog` OLAP tables in different DBs) would currently still throw `Duplicate key` on this branch — that's exactly what pins down the gap. The unit tests mock `getTableIdentifier()` to distinct uuids, so they structurally can't catch the native bare-name case.

Nice work on the Iceberg path regardless — happy to help validate the native-table fix once the keying is settled.

CelerData-Reviewer:
@codex review

Youngwb:
@codex review

CelerData-Reviewer:
@codex review

Youngwb:
Thanks for the thorough review — the native-table gap and the producer/consumer split were exactly right. Reworked the keying; status per point (HEAD `b09c56e9f8`):

**🔴 Native (OLAP) no-op — fixed.** Both execution paths now key by the **base-table `Table` object** rather than any `getTableIdentifier()` string:
- `MVIVMRefreshProcessor.bindBaseTableTvrVersionRanges` builds `Map<Table, TvrVersionRange>` keyed by `snapshotInfo.getBaseTable()`, probed by `relation.getTable()`.
- `MVPCTRefreshPlanBuilder` groups relations into `Multimap<Table, TableRelation>` and looks up to-refresh partitions by `Table` — matching the sibling `mv.getRefBaseTablePartitionSlots()` (`Map<Table,…>`) right next to it.

On the key choice: I avoided `table.getId()` because the analyzer's external `getTable()` can be a fresher connector-cache instance with a new generated id ≠ the snapshot's (the pinned-TVR path comments on exactly that). The `Table` object sidesteps it — `OlapTable` is `equals`-by-`id` (stable/unique), and the external tables override `equals` to `catalog + db + stable-identifier` (`name:uuid` for iceberg/delta, `name:createTime` for hive/hudi), stable across the snapshot/relation boundary. One key, reachable from a bare `Table` on the lookup side, correct for both.

Verified end-to-end on a cluster: native-OLAP `db_a.tj JOIN db_b.tj` and a Hive equivalent now refresh 2 → 3 across an append, no `Duplicate key`.

**Smaller items — all addressed now:**
- **IVM producer/consumer split** — fixed; both sides key by the `Table` object, so it no longer leans on the iceberg `name:uuid == name:uuid` coincidence.
- **`RefBasePartitionsToRefreshMap` format** — kept the diagnostic map keyed by `getName()` and only added a merge fn (same-named entries merge instead of `Collectors.toMap` throwing). **No diagnostic format change** → the "No behavior change" box holds.
- **Trace map (`mvPlanBuildMessage`)** — now keyed by the **qualified `db.tbl` relation name**, so two same-named tables keep separate trace entries instead of overwriting (updated the two `PlanBuilderMessage` assertions accordingly, e.g. `tbl1` → `test.tbl1`).
- **null-table check / comment** — corrected the stale comment to point at the relation-grouping loop where the controlled `AnalysisException` is actually thrown.

**Tests:** added native-OLAP and Hive same-name-join PCT cases (`test_pct_mv_with_olap_same_name_join`, `test_pct_mv_with_hive_same_name_join`) next to the iceberg ones — both verified 2 → 3 on a cluster. The unit test now also exercises the same-name **merge** (two distinct mock tables sharing a name → snapshot map keeps both, diagnostic merges to one), not just distinct uuids.

Thanks again!

CelerData-Reviewer:
@codex review

HangyuanLiu (approved):

Astralidea (approved):

Youngwb:
@mergify backport branch-4.1

PR #74710, opened 2026-06-12
[Enhancement] Read DATE/DATETIME footer stats in tablet pre-split meta tier

xiangguangyxg:
## Why I'm doing:

The Sample-Based Tablet Pre-Split *meta tier* reads Parquet/ORC footer column statistics to plan tablet split boundaries cheaply, avoiding a full data-tier `SELECT` sub-query. Until now it only accepted integer/boolean stats, so temporal range/partition sort keys — the dominant bulk-load case — always fell through to the data tier. This extends the meta tier to DATE/DATETIME so those loads get the cheap footer-based split.

## What I'm doing:

Only the two footer readers change; the pipeline below them (`ParquetMetadataSampler`, `BoundaryPlanner`, `Variant`/`DateVariant`, the providers, `MetaTierFormat`) is type-agnostic and untouched.

- **Parquet** (`ParquetRowGroupStatisticsReader`): `INT32 + DATE` → StarRocks `DATE`; `INT64 + TIMESTAMP` with `isAdjustedToUTC=false` (MILLIS/MICROS/NANOS) → `DATETIME`.
- **ORC** (`OrcStripeStatisticsReader`): `DATE` category → StarRocks `DATE`.
- All gated to the value window `[1970-01-01, 9999-12-31]` via the new `MetaTierTemporalWindow`. Anything outside it — plus UTC-adjusted/INT96 Parquet timestamps, ORC `TIMESTAMP`/`TIMESTAMP_INSTANT`, `DECIMAL`, and `FLOAT/DOUBLE` — throws `MetaTierUnavailableException` and falls back to the data tier (never a wrong boundary, never a load failure).

Why the window: for non-negative ticks the FE `Math.floorDiv`/`floorMod` conversion is bit-identical to the BE load's signed division (`Int64ToDateTimeConverter`), so the FE-computed boundary equals the value the load stores; the window also avoids the `yyyy`-formatter year-0 mis-render and the pre-1582 proleptic/hybrid-calendar parity question. Boundary strings round-trip through `DateVariant.getStringValue` (canonical `yyyy-MM-dd` / `yyyy-MM-dd HH:mm:ss[.ffffff]`).

Tests: per-reader DATE/DATETIME success (incl. window edges and NANOS→micros truncation), UTC-adjusted / out-of-window / type-mismatch / all-null fallbacks, ORC `TIMESTAMP` deferral, and a `ParquetMetadataSampler` date-quantile integration test that pins the type-agnostic-pipeline claim.

## What type of PR is this:

[...]

PR #74739, opened 2026-06-12
[Enhancement] Support DECIMAL sort keys in tablet pre-split (data tier)

xiangguangyxg:
## Why I'm doing:

DECIMAL is the only key / sort-key-eligible StarRocks column type that lacked a `Variant` representation. FLOAT/DOUBLE cannot be keys (`AnalyzerUtils.transformTableColumnType` even coerces float/double key columns to `DECIMAL128(38,9)`), so DECIMAL is the last reachable gap. Because `Variant.of(decimalType, …)` had no decimal case and threw `IllegalArgumentException`, the Sample-Based Tablet Pre-Split **data tier** caught it in `FilesSampleSubqueryExecutor.decodeCell` and recorded `SkipReason.SAMPLE_FAILED` — a decimal sort key got **no pre-split** at all (the load still succeeded against the single tablet — fail-safe, but no parallel write fan-out). This closes that gap for a single decimal sort key.

## What I'm doing:

- **New `DecimalVariant`** (`com.starrocks.catalog`), BigDecimal-backed:
  - `getStringValue()` returns canonical `BigDecimal.toPlainString()` (no scientific notation) so the BE `datum_from_string` parser accepts it.
  - Carries the column `ScalarType` (precision/scale), which activates the previously-dead decimal precision/scale check in `BoundaryPlanner.validateValueAgainstColumn` and matches the BE external-boundary validator in `tablet_splitter.cpp` (exact precision/scale required, else identical-fallback).
  - `equals`/`hashCode` are value-based and consistent with `compareTo` (BigDecimal.compareTo is scale-independent), so `Tuple` equality/dedup (e.g. `ColocateRangeMgr.hasBoundaryAt`, any `Set<Tuple>`) treats the same boundary value written at different scales as equal.
- **Route** `DECIMALV2 / DECIMAL32 / DECIMAL64 / DECIMAL128 / DECIMAL256` through `Variant.of`.
- **Register** `DecimalVariant` in the Gson polymorphic adapter so journaled decimal split boundaries (`SplittingTablet.newTabletRanges`) replay correctly.
- **Fix a pre-existing bug** in `TypeDeserializer`: the DECIMALV2 proto-deserialize path passed `precision` as the scale argument (`createDecimalV2Type(precision, precision)`); it now uses `scale`, and both decimal proto arms guard a null precision/scale with a clear message instead of an opaque unboxing NPE.

This alone fixes the data-tier decimal no-op: a decimal sort key now gets pre-split via the data tier.

**BE is unchanged** — `datum_from_string` already parses every decimal type, `tablet_splitter.cpp` already requires an exact precision/scale match (else it falls back to an identical tablet), and `string_to_decimal` rounds over-scale fractional digits half-up. So the FE keeps the exact sampled value and the BE applies the single authoritative rounding to both the boundary and the loaded data.

**Out of scope (follow-up):** meta-tier footer-reader DECIMAL (Parquet `DecimalLogicalTypeAnnotation`, ORC `DecimalColumnStatistics`) is deferred to a separate PR that builds on #74710 since it edits the same reader methods. Until then a decimal sort key falls back from the meta tier to the data tier, which this PR makes effective.

## What type of PR is this:

- [ ] BugFix
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/DecimalVariant.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/DecimalVariant.java
@@ -86,14 +86,17 @@ public class DecimalVariant extends Variant {
         if (object == null || !(object instanceof DecimalVariant)) {
             return false;
         }
-        // Exact-value equality (BigDecimal.equals is scale-sensitive); ordering uses
-        // compareTo, which is scale-independent. Neither the planner nor the backend
-        // relies on decimal equals.
-        return this.value.equals(((DecimalVariant) object).value);
+        // Value-based equality, kept consistent with compareTo: BigDecimal.compareTo is
+        // scale-independent (1.0 == 1.00), so equals must be too. BigDecimal.equals would be
+        // scale-sensitive and break Tuple equals/hashCode-based dedup (e.g.
+        // ColocateRangeMgr.hasBoundaryAt, any Set<Tuple>) when the same boundary value is
+        // written at different scales.
+        return this.value.compareTo(((DecimalVariant) object).value) == 0;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(value);
+        // stripTrailingZeros canonicalizes scale so values equal under compareTo hash equally.
+        return value.stripTrailingZeros().hashCode();
     }
 }
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/DecimalVariant.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/DecimalVariant.java
@@ -1,0 +1,99 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.catalog;
+
+import com.google.common.base.Preconditions;
+import com.google.gson.annotations.SerializedName;
+import com.starrocks.type.Type;
+
+import java.math.BigDecimal;
+import java.util.Objects;
+
+/*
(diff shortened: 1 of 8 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
## TSP e2e validation — PASS ✅

Shared-data cluster (1 FE + 3 CN, 16c64g, aliyun), build of this PR. Validates that a **DECIMAL sort key now gets data-tier pre-split** — the exact case that previously produced `sampler_failed` and no pre-split.

**Setup**
- `SET enable_range_distribution = true;`
- Target (single tablet, no `DISTRIBUTED BY`):
  ```sql
  CREATE TABLE t_dec (d DECIMAL(18,2), id BIGINT, v INT) DUPLICATE KEY(d) ORDER BY(d);
  ```
- Unloaded 3,000,000 ordered `DECIMAL(18,2)` rows (range `[0.00, 29999.99]`) to OSS Parquet via `INSERT INTO FILES(... TABLE(generate_series(0, 2999999)))`.
- Loaded via bare `INSERT INTO t_dec SELECT * FROM FILES(...)` (triggers the INSERT-from-FILES pre-split hook).

**Results**

| Gate | Observed | Verdict |
|------|----------|---------|
| Tier used | `tablet_pre_split_tier_used{tier="data_tier"}` Δ **+1** (meta tier rejects DECIMAL → data tier); no `sampler_failed` | ✅ |
| Boundaries planned | `tablet_pre_split_boundaries_planned_sum = 2` → K = 3 | ✅ |
| BE accepts decimal boundaries | `tablet_reshard_split_external_boundaries_fallback_total` Δ **= 0** on all 3 CNs, with `tablet_reshard_split_count` +1 on the leader CN (no identical-fallback → exact precision/scale matched) | ✅ |
| Tablet layout | partition **1 → 3** physical tablets, evenly distributed (33% per CN) | ✅ |
| Correctness | `COUNT(*) = 3,000,000`, range preserved `[0.00, 29999.99]` | ✅ |
| Load | `FINISHED` | ✅ |

This flips the Phase-1 ([#74710](https://github.com/StarRocks/starrocks/pull/74710)) "Case E" result — where a DECIMAL sort key hit `sampler_failed{sample_failed}` and got no pre-split — to a successful data-tier split.

Cluster released after the run.

CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #74741, opened 2026-06-12
[Enhancement] Bring public probe/observability HTTP endpoints into enable_http_auth as AuthN-only

iatman:
> Previously the FE/BE health and observability probe endpoints were fully exempt from the enable_http_auth framework. This change brings them into scope at an authentication-only (AuthN-only) level: when enable_http_auth=true they require valid HTTP Basic credentials but no RBAC privilege; when enable_http_auth=false they remain anonymous (fully backward compatible).

When enable_http_auth=true, most external FE/BE HTTP endpoints require HTTP Basic Auth (verified against the FE user store; BE delegates to the FE via the checkAuth RPC). The public probe / observability endpoints are authentication-only (AuthN-only): they require valid Basic credentials when the flag is on, require no RBAC privilege (any authenticated user may call them), and stay anonymous when the flag is off:

- FE: /api/health, /api/idle_status, /api/v2/feature, /metrics
- BE: /api/health, /metrics, /metrics/memory

The following stay fully anonymous regardless of the flag:

- /api/oauth2 (FE) — the OAuth2 callback is itself the authentication handshake, so it can never require Basic Auth.
- /api/bootstrap (FE) — polled by the FE-to-FE heartbeat (HeartbeatMgr) using the cluster token, not HTTP Basic; gating it would break peer-FE heartbeats. (It still accepts the cluster token to unlock extra fields.)
- Token-gated internal transport (tablet clone, load-error file download, Stream Load / transaction endpoints) — protected by their own token checks.

▎ Note: /api/bootstrap is intentionally excluded from the AuthN-only set (an earlier revision listed it; corrected here).

information_schema.fe_metrics no longer depends on HTTP auth

The BE fe_metrics scanner previously scraped each FE's HTTP /metrics?type=json, which would 401 once enable_http_auth=true. It now fetches metrics over a new FrontendService.getFeMetrics Thrift RPC. The FE returns the identical JSON payload, so the table's columns are unchanged, and the view works regardless of enable_http_auth. information_schema.be_metrics is unaffected (it reads BE process metrics in-process, never over HTTP).

Impact:
- No behavior change when enable_http_auth=false (the default).
- When enabled, the listed probes require Basic auth, so Prometheus scrapers, k8s liveness/readiness probes, and bootstrap pollers must supply credentials.
- The /metrics per-table / per-MV / per-user-connection breakdowns keep their existing admin authorization check.

Tests:
- FE RestBaseActionTest: probes gated by the flag; OAuth2 never gated.
- BE metrics_action_test: probes are need_auth()=true with NONE privilege.

Docs: updated the enable_http_auth description in the en/zh/ja FE and BE parameter docs to describe the AuthN-only behavior and the OAuth2 exemption.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/http/rest/FeatureAction.java
+++ b/fe/fe-core/src/main/java/com/starrocks/http/rest/FeatureAction.java
@@ -14,6 +14,7 @@
 
 package com.starrocks.http.rest;
 
+import com.starrocks.common.Config;
 import com.starrocks.common.Version;
 import com.starrocks.feature.ProductFeature;
 import com.starrocks.http.ActionController;
@@ -32,10 +33,12 @@ public class FeatureAction extends RestBaseAction {
         controller.registerHandler(HttpMethod.GET, "/api/v2/feature", new FeatureAction(controller));
     }
 
-    // Public endpoint: returns server version + feature flag list, no sensitive data.
+    // Returns server version + feature flag list, no sensitive data. Historically
+    // anonymous; gated for backward compatibility so it requires Basic auth (AuthN-only,
+    // no privilege check) only when the operator opts in via `enable_http_auth`.
     @Override
     public boolean needAuth() {
-        return false;
+        return Config.enable_http_auth;
     }
 
     @Override
--- a/fe/fe-core/src/main/java/com/starrocks/http/rest/IdleAction.java
+++ b/fe/fe-core/src/main/java/com/starrocks/http/rest/IdleAction.java
@@ -49,10 +49,12 @@ public class IdleAction extends RestBaseAction {
         controller.registerHandler(HttpMethod.GET, "/api/idle_status", new IdleAction(controller));
     }
 
-    // K8s idle probe; intentionally unauthenticated.
+    // K8s idle probe. Historically anonymous; gated for backward compatibility so it
+    // requires Basic auth (AuthN-only, no privilege check) only when the operator opts
+    // in via `enable_http_auth`.
     @Override
     public boolean needAuth() {
-        return false;
+        return Config.enable_http_auth;
     }
 
     @Override
--- a/gensrc/thrift/FrontendService.thrift
+++ b/gensrc/thrift/FrontendService.thrift
@@ -2439,6 +2439,16 @@ struct TBatchGetTabletMetadataResponse {
     2: optional list<TGetTabletMetadataResponse> responses;
 }
 
+// information_schema.fe_metrics: the BE scanner fetches each FE's metrics over the getFeMetrics
+// RPC (instead of scraping the HTTP /metrics endpoint), so fe_metrics works regardless of
+// `enable_http_auth` and no longer depends on HTTP Basic auth. The RPC takes no arguments —
+// FE process metrics are global with no per-object RBAC to authorize.
+struct TFeMetricsResult {
+    1: optional Status.TStatus status
+    // JSON payload identical to the FE `/metrics?type=json` output, parsed by the BE scanner.
+    2: optional string json_metrics
+}
+
 service FrontendService {
     TGetDbsResult getDbNames(1:TGetDbsParams params)
     TGetTablesResult getTableNames(1:TGetTablesParams params)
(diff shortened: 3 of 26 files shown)


CelerData-Reviewer:
@codex review

wyb:
@codex review

gengjun-git (dismissed):

fe/fe-core/src/main/java/com/starrocks/http/rest/FeatureAction.java
```diff
@@ -32,10 +33,12 @@ public static void registerAction(ActionController controller)
...
     }
 
-    // Public endpoint: returns server version + feature flag list, no sensitive data.
+    // Returns server version + feature flag list, no sensitive data. Historically
+    // anonymous; gated for backward compatibility so it requires Basic auth (AuthN-only,
+    // no privilege check) only when the operator opts in via `enable_http_auth`.
     @Override
     public boolean needAuth() {
-        return false;
+        return Config.enable_http_auth;
```
jaogoy, line 41: It should be false all the time, it's used for other system to check the features supported by this version, and no security issue.

be/src/http/action/health_action.h
```diff
@@ -49,7 +49,10 @@ class HealthAction : public HttpHandler {
 
     void handle(HttpRequest* req) override;
 
-    bool need_auth() const override { return false; }
+    // AuthN-only: require Basic identity when `config::enable_http_auth` is on (the
+    // injected verifier short-circuits when it is off). No extra privilege required —
+    // `required_privilege()` stays NONE.
+    bool need_auth() const override { return true; }
```
jaogoy: How could other systems, such as nginx and LB, do health check with a user?
Especially, it uses a remote call to check the auth (different with FE), it's not recommended for health probe.
We need to tell users about it in our docs.

fe/fe-core/src/main/java/com/starrocks/http/rest/IdleAction.java
```diff
@@ -49,10 +49,12 @@ public static void registerAction(ActionController controller) throws IllegalArg
...
     }
 
-    // K8s idle probe; intentionally unauthenticated.
+    // K8s idle probe. Historically anonymous; gated for backward compatibility so it
+    // requires Basic auth (AuthN-only, no privilege check) only when the operator opts
+    // in via `enable_http_auth`.
     @Override
     public boolean needAuth() {
-        return false;
+        return Config.enable_http_auth;
```
jaogoy, line 57: What do we need to configure k8s or grafana to fetch the idle status with a user?
We need to tell users about it in our docs, also including Metrics.

huailiu1122:
@mergify rebase

huailiu1122:
@mergify rebase

gensrc/thrift/FrontendService.thrift
```diff
@@ -2439,6 +2439,20 @@ struct TBatchGetTabletMetadataResponse {
     2: optional list<TGetTabletMetadataResponse> responses;
 }
 
+// information_schema.fe_metrics: the BE scanner fetches each FE's metrics over this RPC
+// (instead of scraping the HTTP /metrics endpoint), passing the already-authenticated
+// caller identity in auth_info.current_user_ident. This keeps fe_metrics working
+// regardless of `enable_http_auth`, since it no longer depends on HTTP Basic auth.
+struct TFeMetricsRequest {
+    1: optional TAuthInfo auth_info
```
wyb: `auth_info` is not used

wyb (approved):

gengjun-git (approved):

trueeyu (approved):

wyb:
branch-4.1 https://github.com/StarRocks/starrocks/pull/75582

PR #74747, opened 2026-06-12
[Enhancement] Propagate flat_json config ALTER to BE via versioned task

### LINKED ISSUE #73846: [Bug] ALTER TABLE flat_json properties are not propagated to BE

```
CREATE TABLE t (id BIGINT, j JSON)
DUPLICATE KEY(id) DISTRIBUTED BY HASH(id) BUCKETS 1
PROPERTIES (
    'replication_num'           = '1',
    'flat_json.enable'          = 'true',
    'flat_json.sparsity.factor' = '0.5',
    'flat_json.null.factor'     = '0.99',
    'flat_json.column.max'      = '100'
);

ALTER TABLE t SET ('flat_json.sparsity.factor' = '1');

INSERT INTO t
SELECT n,
       CASE WHEN n % 100 < 60
            THEN parse_json('{"a":"v","p":"hit"}')
            ELSE parse_json('{"a":"v"}')
       END
FROM TABLE(generate_series(1, 1000)) AS t(n);
Path .p ends up in 600/1000 = 60% of rows. With sparsity.factor = 1.0, the FE-configured policy requires .p to be SKIPPED (since 600 >= 1000 × 1.0 is false). Path .a at 100% should still be columnized.
```

BE honors the ALTERed sparsity.factor = 1.0:

- .p (60%) → SKIPPED
- .a (100%) → columnized
Same expectation applies to flat_json.enable, flat_json.null.factor, and flat_json.column.max.

### Real behavior (Required)

SHOW CREATE TABLE reflects sparsity.factor = 1.0, but BE keeps using the CREATE-time value 0.5. .p is incorrectly columnized (600 >= 1000 × 0.5 = 500 → TRUE), contradicting the FE-configured policy. The ALTER produces zero BE-side log activity:

```
$ tail -f be.INFO | grep flat_json_config
# (nothing during ALTER)
```

#### Code verification

The single decision branch on BE is [be/src/util/json_flattener.cpp:688](https://github.com/StarRocks/starrocks/blob/main/be/src/util/json_flattener.cpp#L688):

```
if (type_check && node->multi_times <= 0
    && node->hits >= _total_rows * _min_json_sparsity_factory) {
    hit_leaf->emplace_back(node, absolute_path);   // → columnize
    node->remain = false;
    return 1;
} else {
    node->remain = true;                            // → SKIPPED
    return 0;
}
```

_min_json_sparsity_factory is the BE's runtime view of flat_json.sparsity.factor. Before the fix, ALTERing the table-level property never updates that field on any BE — the comparison silently uses the CREATE-time value forever.

#### Root cause
- SchemaChangeHandler.updateFlatJsonConfigMeta(..., TTabletMetaType metaType) accepts a metaType parameter but never dispatches any TabletMetadataUpdateAgentTask.
- TTabletInfo has no flat_json_config_version field, so ReportHandler cannot reconcile FE/BE drift.
- The dangling metaType parameter and the unused TTabletMetaType.FLAT_JSON_CONFIG enum value appear to be surface remnants of an unfinished implementation. No SQL fixture under test/sql/test_semi/ covers ALTER property changes for flat_json.

#### Fix
- PR #73514 follows the existing StarRocks metadata synchronization pattern.

- main (reproducible since flat_json ALTER properties were introduced; affects 4.1 / 4.0 / 3.5 / 3.4)

moweonlee:
> **Squashed, single-history re-submission of #73514** (2 clean commits — the fix + its UT), opened separately per maintainer request so it can be fast-tracked and backported independently of the long review/CI history on #73514.
>
> The change set is **identical to #73514** (17/18 touched files are byte-identical; `ReportHandler.java` differs only in surrounding context after rebasing onto the latest `main`). Re-verified on the latest `main`: FE compiles, `mvn checkstyle:check` clean, `ReportHandlerTest` 20/0, and the touched BE files (`tablet.cpp`, `agent_task.cpp`, `lake/schema_change.cpp`, `lake/txn_log_applier.cpp`) recompile cleanly.

---

Fixes #73846

## Why I'm doing:

`ALTER TABLE ... SET ("flat_json.*" = ...)` is documented as a way to retune flat_json behaviour on an existing table (`flat_json.enable`, `flat_json.sparsity.factor`, `flat_json.null.factor`, `flat_json.column.max`). The FE accepts the ALTER and `SHOW CREATE TABLE` reflects the new values, but the BE never receives the change: subsequent INSERTs, flushes, and compactions keep using the CREATE-time `FlatJsonConfig` indefinitely.

`SchemaChangeHandler.updateFlatJsonConfigMeta(..., TTabletMetaType metaType)` already takes a `metaType` parameter but never uses it. There is no `TabletMetadataUpdateAgentTask` dispatch, no version field on `FlatJsonConfig` or `TTabletInfo`, and no `ReportHandler` reconciliation. The dangling `metaType` parameter and the unused `TTabletMetaType.FLAT_JSON_CONFIG` enum value look like surface remnants of an unfinished implementation that the existing tests do not exercise — no SQL fixture under `test/sql/test_semi/` covers ALTER property changes for flat_json.

### Decision point (where the bug surfaces in the BE)

`be/src/util/json_flattener.cpp:688` is the single decision branch that decides whether a JSON path becomes a sub-column:

```cpp
if (type_check && node->multi_times <= 0
    && node->hits >= _total_rows * _min_json_sparsity_factory) {
    hit_leaf->emplace_back(node, absolute_path);   // → columnize
    node->remain = false;
    return 1;
} else {
    node->remain = true;                            // → SKIPPED (raw remain)
    return 0;
}
```

`_min_json_sparsity_factory` is the BE's runtime view of `flat_json.sparsity.factor`. Before this fix, ALTERing the table-level property never changes that field on any BE; the comparison silently uses the CREATE-time value forever.

## What I'm doing:

Mirror the binlog config pattern, which already solves this exact problem in production for `BinlogConfig`. Two flows close the gap together; neither alone is sufficient.

**Flow 1 — Immediate FE → BE push.** On `ALTER`, the FE
- snapshots partitions under the table read lock,
- bumps `FlatJsonConfig.configVersion` (`incVersion()`),
- persists via `EditLog` (existing `LocalMetastore.modifyFlatJsonMeta`),
- then drops the lock and pushes the new config to every BE that holds a replica, via `createFlatJsonConfigUpdateTask` → `UpdateFlatJsonConfigTask` → `AgentBatchTask` → BE's `agent_task.cpp` new `case TTabletMetaType::FLAT_JSON_CONFIG` → `Tablet::update_flat_json_config` (version-gated, so replays or out-of-order pushes are idempotent).

**Flow 2 — Heartbeat reconciliation.** Every tablet report now carries `TTabletInfo.flat_json_config_version`. On the leader, `ReportHandler.handleSetTabletFlatJsonConfig` compares it against `olapTable.getFlatJsonConfig().getVersion()` and re-pushes via the same task factory when a BE has fallen behind. So a BE that was down (or briefly unreachable) at ALTER time automatically catches up on its next report cycle, without operator intervention.

Both flows are mechanical mirrors of `BinlogConfig`'s `updateBinlogPartitionTabletMeta` + `handleSetTabletBinlogConfig` — same call shape, same locking pattern (intensive table read lock + lock acquired outside the `try` block), same agent task / countdown latch / timeout / error-handling structure.

### Thrift / proto additions (all optional, append-only)

- `TFlatJsonConfig.version`
- `TTabletInfo.flat_json_config_version`
- `FlatJsonConfigPB.version`

Old peers that omit these fields are treated as `version=0` on the receiving side (`__isset.version ? config.version : 0` on BE, `has_version() ? ... : 0` on proto). Mixed-version clusters stay correct: an old BE/FE keeps the pre-fix behaviour (no propagation) but cannot crash a new peer; a new BE seeing legacy on-disk meta initialises at `version=0` and accepts the first push normally.

### Files changed

```
gensrc/thrift/AgentService.thrift                                     (+TFlatJsonConfig.version)
gensrc/thrift/MasterService.thrift                                    (+TTabletInfo.flat_json_config_version)
gensrc/proto/olap_file.proto                                          (+FlatJsonConfigPB.version)

be/src/storage/flat_json_config.h                                     (version field + serde)
be/src/storage/tablet.h                                               (update_flat_json_config decl + fwd)
be/src/storage/tablet.cpp                                             (update_flat_json_config impl, version-gated)
be/src/agent/agent_task.cpp                                           (case FLAT_JSON_CONFIG)

fe/.../catalog/FlatJsonConfig.java                                    (configVersion + incVersion + thrift carry)
fe/.../task/TabletMetadataUpdateAgentTaskFactory.java                 (factory + UpdateFlatJsonConfigTask)
fe/.../alter/SchemaChangeHandler.java                                 (incVersion + updateFlatJsonPartitionTabletMeta)
fe/.../leader/ReportHandler.java                                      (handleSetTabletFlatJsonConfig)
```

## Verification

End-to-end on a single-FE / single-BE local cluster, rebased onto current upstream `main` HEAD + this commit, with `thirdparty` rebuilt from scratch (fmt 10.2.1 / staros v4.2-rc1 / tenann updated). Same scenario run before and after applying the fix; outcomes are opposite at the decision line above.

### Scenario

```sql
CREATE TABLE t (id BIGINT, j JSON)
DUPLICATE KEY(id) DISTRIBUTED BY HASH(id) BUCKETS 1
PROPERTIES (
    'replication_num'           = '1',
    'flat_json.enable'          = 'true',
    'flat_json.sparsity.factor' = '0.5',
    'flat_json.null.factor'     = '0.99',
    'flat_json.column.max'      = '100'
);

ALTER TABLE t SET ('flat_json.sparsity.factor' = '1');

INSERT INTO t
SELECT n,
       CASE WHEN n % 100 < 60
            THEN parse_json('{"a":"v","p":"hit"}')
            ELSE parse_json('{"a":"v"}')
       END
FROM TABLE(generate_series(1, 1000)) AS t(n);
```

Path `.p` ends up in 600/1000 rows = 60%. With `sparsity.factor = 1.0`, the FE-configured policy requires that path to be SKIPPED (`600 >= 1000 × 1.0` → false). Path `.a` at 100% must still be columnized.

### Before the fix

`SHOW CREATE TABLE` reflects `flat_json.sparsity.factor = 1.0`, but the BE-side decision uses the stale CREATE-time `0.5`:

```
JsonPathDeriver: path '.p' columnized: hits=600/1000,  sparsity_factor=0.5   ← !!! must be SKIPPED
JsonPathDeriver: path '.a' columnized: hits=1000/1000, sparsity_factor=0.5
```

Same arithmetic at `json_flattener.cpp:688` then runs `600 >= 1000 × 0.5 = 500` → TRUE → columnize, contradicting the FE-configured policy. The ALTER produced **zero** BE-side log activity:

```
$ tail -f be.INFO | grep flat_json_config
# (nothing during ALTER)
```

### After the fix

`ALTER` immediately produces BE-side activity through Flow 1:

```
agent_server.cpp Submit task success. type=UPDATE_TABLET_META_INFO, signatures=…
agent_task.cpp   update tablet:N flat_json_config
tablet.cpp       set flat_json_config of tablet: N,
                  FlatJsonConfig{flat_json_enable=true,
                                 flat_json_null_factor=0.99,
                                 flat_json_sparsity_factor=1,         ← now matches FE
                                 flat_json_max_column_max=100,
                                 version=1}                           ← version bumped
```

Next INSERT uses the new threshold at the same decision line:

```
JsonPathDeriver: path '.p' SKIPPED:    hits=600/1000,  sparsity_factor=1   ← !!! policy honoured
JsonPathDeriver: path '.a' columnized: hits=1000/1000, sparsity_factor=1
```

`600 >= 1000 × 1.0 = 1000` → FALSE → SKIPPED. Path `.a` still columnizes (100% ≥ 100%). Both branches are exercised; both print `sparsity_factor=1` from `_min_json_sparsity_factory`.

The two `VLOG(2)` lines used to dump `sparsity_factor` are local diagnostic patches only and are **not** part of this PR — verification was done with them enabled, the PR ships without them.

### Backward compatibility check (logical)

- **Old FE + new BE**: old FE never dispatches → new BE's `case FLAT_JSON_CONFIG` never fires → matches pre-fix behaviour. No crash.
- **New FE + old BE**: new FE dispatches `meta_type=FLAT_JSON_CONFIG`, old BE's switch falls through default → silent skip. Pre-fix behaviour preserved.
- **Old on-disk `FlatJsonConfigPB` (no `version`)**: `has_version()` → false → `_flat_json_config_version = 0`. First ALTER then has `BE_version=0 < FE_version=1` → push proceeds normally.
- **FE leader failover mid-ALTER**: `incVersion()` + `EditLog` happen under the table write lock, so the new leader sees the bumped version. Any BE that missed Flow 1 catches up via Flow 2 on its next tablet report.
- **BE reports `version > FE_version`** (shouldn't happen): handler emits a WARN log and does not re-push; never regresses to a stale value.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

Rationale for the boxes:
- **Policy changes**: `ALTER TABLE SET ("flat_json.*" = ...)` now actually takes effect on the BE for new writes and compactions, instead of being silently a no-op against existing tablets.
- **Miscellaneous (compatibility)**: new optional `version` fields on `TFlatJsonConfig`, `TTabletInfo`, `FlatJsonConfigPB`. Behaviour for both directions of mixed-version clusters is described under "Backward compatibility check" above.

## Checklist:

- [ ] This is a backport pr

Notes on the boxes:
- **Tests**: end-to-end behavioural verification at the decision line is recorded in the "Verification" section above. Happy to add a SQL fixture under `test/sql/test_semi/` if reviewers want a check that runs in CI.
- **Docs**: the user-facing page at [`docs/en/using_starrocks/Flat_json/`](https://docs.starrocks.io/docs/using_starrocks/Flat_json/) describes the `flat_json.*` table properties and `ALTER` flow without flagging that the BE silently keeps stale values, so the docs already describe the intended (and now actual) behaviour — no doc change required to *describe* this fix.

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/agent/agent_task.cpp
+++ b/be/src/agent/agent_task.cpp
@@ -983,9 +983,10 @@ void run_update_meta_info_task(const std::shared_ptr<UpdateTabletMetaInfoAgentTa
                 tablet->update_binlog_config(binlog_config);
                 break;
             case TTabletMetaType::FLAT_JSON_CONFIG: {
-                LOG(INFO) << "update tablet:" << tablet->tablet_id() << " flat_json_config";
                 FlatJsonConfig flat_json_config;
                 flat_json_config.update(tablet_meta_info.flat_json_config);
+                LOG(INFO) << "update tablet:" << tablet->tablet_id()
+                          << " flat_json_config, version: " << flat_json_config.get_flat_json_config_version();
                 tablet->update_flat_json_config(flat_json_config);
                 break;
             }
--- a/be/src/agent/agent_task.cpp
+++ b/be/src/agent/agent_task.cpp
@@ -46,6 +46,7 @@
 #include "storage/tablet_manager.h"
 #include "storage/txn_manager.h"
 #include "storage/update_manager.h"
+#include "storage_primitive/flat_json_config.h"
 
 namespace starrocks {
 
@@ -981,6 +982,13 @@ void run_update_meta_info_task(const std::shared_ptr<UpdateTabletMetaInfoAgentTa
                 binlog_config.update(tablet_meta_info.binlog_config);
                 tablet->update_binlog_config(binlog_config);
                 break;
+            case TTabletMetaType::FLAT_JSON_CONFIG: {
+                LOG(INFO) << "update tablet:" << tablet->tablet_id() << " flat_json_config";
+                FlatJsonConfig flat_json_config;
+                flat_json_config.update(tablet_meta_info.flat_json_config);
+                tablet->update_flat_json_config(flat_json_config);
+                break;
+            }
             case TTabletMetaType::ENABLE_PERSISTENT_INDEX:
                 LOG(INFO) << "update tablet:" << tablet->tablet_id()
                           << " enable_persistent_index:" << tablet_meta_info.enable_persistent_index;
--- a/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
@@ -607,11 +607,11 @@ public class AlterJobExecutor implements AstVisitorExtendInterface<Void, Connect
                     properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_COLUMN_MAX)) {
                 if (table.isCloudNativeTable()) {
                     Locker locker = new Locker();
-                    locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(table.getId()), LockType.WRITE);
+                    locker.lockTableWithIntensiveDbLock(db.getId(), table.getId(), LockType.WRITE);
                     try {
                         schemaChangeHandler.processLakeTableAlterMeta(clause, db, (OlapTable) table);
                     } finally {
-                        locker.unLockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(table.getId()), LockType.WRITE);
+                        locker.unLockTableWithIntensiveDbLock(db.getId(), table.getId(), LockType.WRITE);
                     }
                     isSynchronous = false;
                 } else {
--- a/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
@@ -605,11 +605,21 @@ public class AlterJobExecutor implements AstVisitorExtendInterface<Void, Connect
                     properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_NULL_FACTOR) ||
                     properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_SPARSITY_FACTOR) ||
                     properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_COLUMN_MAX)) {
-                boolean isSuccess = schemaChangeHandler.updateFlatJsonConfigMeta(db, table.getId(),
-                        properties, TTabletMetaType.FLAT_JSON_CONFIG);
-                if (!isSuccess) {
-                    throw new DdlException("modify flat json config of FEMeta failed");
-
+                if (table.isCloudNativeTable()) {
+                    Locker locker = new Locker();
+                    locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(table.getId()), LockType.WRITE);
+                    try {
+                        schemaChangeHandler.processLakeTableAlterMeta(clause, db, (OlapTable) table);
+                    } finally {
+                        locker.unLockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(table.getId()), LockType.WRITE);
+                    }
+                    isSynchronous = false;
+                } else {
+                    boolean isSuccess = schemaChangeHandler.updateFlatJsonConfigMeta(db, table.getId(),
+                            properties, TTabletMetaType.FLAT_JSON_CONFIG);
+                    if (!isSuccess) {
+                        throw new DdlException("modify flat json config of FEMeta failed");
+                    }
                 }
             } else if (properties.containsKey(PropertyAnalyzer.PROPERTIES_FOREIGN_KEY_CONSTRAINT)
                     || properties.containsKey(PropertyAnalyzer.PROPERTIES_UNIQUE_CONSTRAINT)) {
--- a/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJob.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJob.java
@@ -16,6 +16,7 @@ package com.starrocks.alter;
 
 import com.google.gson.annotations.SerializedName;
 import com.starrocks.catalog.Database;
+import com.starrocks.catalog.FlatJsonConfig;
 import com.starrocks.catalog.MaterializedIndex;
 import com.starrocks.catalog.OlapTable;
 import com.starrocks.catalog.PhysicalPartition;
@@ -42,6 +43,9 @@ public class LakeTableAlterMetaJob extends LakeTableAlterMetaJobBase {
     @SerializedName(value = "compactionStrategy")
     private String compactionStrategy;
 
+    @SerializedName(value = "flatJsonConfig")
+    private FlatJsonConfig flatJsonConfig;
+
     // for deserialization
     public LakeTableAlterMetaJob() {
         super(JobType.SCHEMA_CHANGE);
@@ -67,6 +71,13 @@ public class LakeTableAlterMetaJob extends LakeTableAlterMetaJobBase {
         this.compactionStrategy = compactionStrategy;
     }
 
+    public LakeTableAlterMetaJob(long jobId, long dbId, long tableId, String tableName,
+                                 long timeoutMs, FlatJsonConfig flatJsonConfig) {
+        super(jobId, JobType.SCHEMA_CHANGE, dbId, tableId, tableName, timeoutMs);
+        this.metaType = TTabletMetaType.FLAT_JSON_CONFIG;
+        this.flatJsonConfig = flatJsonConfig;
+    }
+
     protected LakeTableAlterMetaJob(LakeTableAlterMetaJob job) {
         super(job);
         this.metaType = job.metaType;
@@ -74,6 +85,7 @@ public class LakeTableAlterMetaJob extends LakeTableAlterMetaJobBase {
         this.persistentIndexType = job.persistentIndexType;
         this.enableFileBundling = job.enableFileBundling;
         this.compactionStrategy = job.compactionStrategy;
+        this.flatJsonConfig = job.flatJsonConfig == null ? null : new FlatJsonConfig(job.flatJsonConfig);
     }
 
     @Override
@@ -90,6 +102,10 @@ public class LakeTableAlterMetaJob extends LakeTableAlterMetaJobBase {
         if (metaType == TTabletMetaType.COMPACTION_STRATEGY) {
             return TabletMetadataUpdateAgentTaskFactory.createUpdateCompactionStrategyTask(nodeId, tablets,
                         compactionStrategy);
+        }
+        if (metaType == TTabletMetaType.FLAT_JSON_CONFIG) {
+            return TabletMetadataUpdateAgentTaskFactory.createFlatJsonConfigUpdateTask(nodeId, tablets,
+                        flatJsonConfig);
         }
         return null;
     }
@@ -123,6 +139,9 @@ public class LakeTableAlterMetaJob extends LakeTableAlterMetaJobBase {
                     String.valueOf(compactionStrategy));
             table.getTableProperty().buildCompactionStrategy();
         }
+        if (metaType == TTabletMetaType.FLAT_JSON_CONFIG && flatJsonConfig != null) {
+            table.setFlatJsonConfig(flatJsonConfig);
+        }
     }
 
     @Override
@@ -133,6 +152,9 @@ public class LakeTableAlterMetaJob extends LakeTableAlterMetaJobBase {
         this.persistentIndexType = other.persistentIndexType;
         this.enableFileBundling = other.enableFileBundling;
         this.compactionStrategy = other.compactionStrategy;
+        this.flatJsonConfig = other.flatJsonConfig == null
+                ? null
+                : new FlatJsonConfig(other.flatJsonConfig);
     }
 
     @Override
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -3074,33 +3074,41 @@ public class SchemaChangeHandler extends AlterHandler {
             AgentTaskExecutor.submit(batchTask);
             LOG.info("send update flat_json_config tablet meta task for table {}, partition {}, number: {}",
                     tableName, partitionName, batchTask.getTaskNum());
+            waitForTabletMetaBatchTask(batchTask, countDownLatch, totalTaskNum,
+                    "Failed to update partition[" + partitionName + "] flat_json_config tablet meta.");
+        }
+    }
 
-            long timeout = Config.tablet_create_timeout_second * 1000L * totalTaskNum;
-            timeout = Math.min(timeout, Config.max_create_table_timeout_second * 1000L);
-            boolean ok = false;
-            try {
-                ok = countDownLatch.await(timeout, TimeUnit.MILLISECONDS);
-            } catch (InterruptedException e) {
-                LOG.warn("InterruptedException: ", e);
-            }
+    // Wait for a dispatched tablet-meta batch task and turn a timeout/error into a DdlException.
+    private static void waitForTabletMetaBatchTask(AgentBatchTask batchTask,
+                                                   MarkedCountDownLatch<Long, Set<Long>> countDownLatch,
+                                                   int totalTaskNum, String errMsgPrefix) throws DdlException {
+        long timeout = Config.tablet_create_timeout_second * 1000L * totalTaskNum;
+        timeout = Math.min(timeout, Config.max_create_table_timeout_second * 1000L);
+        boolean ok = false;
+        try {
+            ok = countDownLatch.await(timeout, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            LOG.warn("InterruptedException: ", e);
+        }
 
-            if (!ok || !countDownLatch.getStatus().ok()) {
-                String errMsg = "Failed to update partition[" + partitionName + "] flat_json_config tablet meta.";
-                AgentTaskQueue.removeBatchTask(batchTask, TTaskType.UPDATE_TABLET_META_INFO);
-                if (!countDownLatch.getStatus().ok()) {
-                    errMsg += " Error: " + countDownLatch.getStatus().getErrorMsg();
-                } else {
-                    List<Map.Entry<Long, Set<Long>>> unfinishedMarks = countDownLatch.getLeftMarks();
-                    List<Map.Entry<Long, Set<Long>>> subList =
-                            unfinishedMarks.subList(0, Math.min(unfinishedMarks.size(), 3));
-                    if (!subList.isEmpty()) {
-                        errMsg += " Unfinished mark: " + Joiner.on(", ").join(subList);
-                    }
+        if (!ok || !countDownLatch.getStatus().ok()) {
+            String errMsg = errMsgPrefix;
+            AgentTaskQueue.removeBatchTask(batchTask, TTaskType.UPDATE_TABLET_META_INFO);
+            if (!countDownLatch.getStatus().ok()) {
+                errMsg += " Error: " + countDownLatch.getStatus().getErrorMsg();
+            } else {
+                List<Map.Entry<Long, Set<Long>>> unfinishedMarks = countDownLatch.getLeftMarks();
+                List<Map.Entry<Long, Set<Long>>> subList =
+                        unfinishedMarks.subList(0, Math.min(unfinishedMarks.size(), 3));
+                if (!subList.isEmpty()) {
+                    errMsg += " Unfinished mark: " + Joiner.on(", ").join(subList);
                 }
-                errMsg += ". This operation maybe partial successfully, You should retry until success.";
-                LOG.warn(errMsg);
-                throw new DdlException(errMsg);
             }
+            errMsg += ". This operation maybe partial successfully, You should retry until success.";
+            LOG.warn(errMsg);
+            throw new DdlException(errMsg);
         }
     }
 
@@ -3364,6 +3372,7 @@ public class SchemaChangeHandler extends AlterHandler {
             try {
                 ok = countDownLatch.await(timeout, TimeUnit.MILLISECONDS);
             } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
                 LOG.warn("InterruptedException: ", e);
             }
 
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -2750,6 +2750,15 @@ public class SchemaChangeHandler extends AlterHandler {
                         ? new FlatJsonConfig(olapTable.getFlatJsonConfig())
                         : new FlatJsonConfig();
                 newConfig.buildFromProperties(properties);
+                // The analyzer's enabled-check ran without the table lock, so a concurrent ALTER
+                // may have disabled flat_json since; re-validate against the locked state so a
+                // factor change cannot land on a disabled config (mirrors updateFlatJsonConfigMeta).
+                if (!newConfig.getFlatJsonEnable()
+                        && (properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_NULL_FACTOR)
+                        || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_SPARSITY_FACTOR)
+                        || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_COLUMN_MAX))) {
+                    throw new DdlException("flat JSON configuration must be set after enabling flat JSON.");
+                }
                 newConfig.incVersion();
                 metaType = TTabletMetaType.FLAT_JSON_CONFIG;
                 flatJsonConfig = newConfig;
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -2641,6 +2641,7 @@ public class SchemaChangeHandler extends AlterHandler {
             boolean enableFileBundling = false;
             TTabletMetaType metaType = TTabletMetaType.ENABLE_PERSISTENT_INDEX;
             String compactionStrategy = "";
+            FlatJsonConfig flatJsonConfig = null;
             if (properties.containsKey(PropertyAnalyzer.PROPERTIES_ENABLE_PERSISTENT_INDEX)) {
                 enablePersistentIndex = PropertyAnalyzer.analyzeBooleanProp(properties,
                         PropertyAnalyzer.PROPERTIES_ENABLE_PERSISTENT_INDEX, false);
@@ -2741,16 +2742,34 @@ public class SchemaChangeHandler extends AlterHandler {
                 return null;
             } else if (properties.containsKey(PropertyAnalyzer.PROPERTIES_CLOUD_NATIVE_FAST_SCHEMA_EVOLUTION_V2)) {
                 return processAlterCloudNativeFastSchemaEvolutionV2Property(db, olapTable, properties).orElse(null);
+            } else if (properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_ENABLE)
+                    || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_NULL_FACTOR)
+                    || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_SPARSITY_FACTOR)
+                    || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_COLUMN_MAX)) {
+                FlatJsonConfig newConfig = olapTable.containsFlatJsonConfig()
+                        ? new FlatJsonConfig(olapTable.getFlatJsonConfig())
+                        : new FlatJsonConfig();
+                newConfig.buildFromProperties(properties);
+                newConfig.incVersion();
+                metaType = TTabletMetaType.FLAT_JSON_CONFIG;
+                flatJsonConfig = newConfig;
             } else {
                 throw new DdlException("does not support alter " + properties.entrySet().iterator().next().getKey() +
                         " in shared_data mode");
             }
 
             long timeoutSecond = PropertyAnalyzer.analyzeTimeout(properties, Config.alter_table_timeout_second);
-            alterMetaJob = new LakeTableAlterMetaJob(GlobalStateMgr.getCurrentState().getNextId(),
-                    db.getId(),
-                    olapTable.getId(), olapTable.getName(), timeoutSecond * 1000 /* should be ms*/,
-                    metaType, enablePersistentIndex, persistentIndexType, enableFileBundling, compactionStrategy);
+            if (metaType == TTabletMetaType.FLAT_JSON_CONFIG) {
+                alterMetaJob = new LakeTableAlterMetaJob(GlobalStateMgr.getCurrentState().getNextId(),
+                        db.getId(),
+                        olapTable.getId(), olapTable.getName(), timeoutSecond * 1000 /* should be ms*/,
+                        flatJsonConfig);
+            } else {
+                alterMetaJob = new LakeTableAlterMetaJob(GlobalStateMgr.getCurrentState().getNextId(),
+                        db.getId(),
+                        olapTable.getId(), olapTable.getName(), timeoutSecond * 1000 /* should be ms*/,
+                        metaType, enablePersistentIndex, persistentIndexType, enableFileBundling, compactionStrategy);
+            }
         } else {
             // shouldn't happen
             throw new DdlException("only support alter enable_persistent_index in shared_data mode");
@@ -2901,6 +2920,7 @@ public class SchemaChangeHandler extends AlterHandler {
         if (olapTable == null) {
             return false;
         }
+        List<Partition> partitions = Lists.newArrayList();
         Locker locker = new Locker();
         locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(olapTable.getId()), LockType.READ);
         try {
@@ -2910,6 +2930,7 @@ public class SchemaChangeHandler extends AlterHandler {
             } else {
                 newFlatJsonConfig = new FlatJsonConfig(olapTable.getFlatJsonConfig());
             }
+            partitions.addAll(olapTable.getPartitions());
         } finally {
             locker.unLockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(olapTable.getId()), LockType.READ);
         }
@@ -2958,6 +2979,25 @@ public class SchemaChangeHandler extends AlterHandler {
 
         locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(olapTable.getId()), LockType.WRITE);
         try {
+            // check for concurrent modifications by version
+            if (olapTable.containsFlatJsonConfig()
+                    && olapTable.getFlatJsonConfig().getVersion() != newFlatJsonConfig.getVersion()) {
+                Map<String, String> newProperties = olapTable.getFlatJsonConfig().toProperties();
+                newProperties.putAll(properties);
+                newFlatJsonConfig.buildFromProperties(newProperties);
+                newFlatJsonConfig.setVersion(olapTable.getFlatJsonConfig().getVersion());
+                // Re-validate after rebasing onto the (possibly concurrently-modified) current
+                // config: a factor change must not be silently merged onto a now-disabled config,
+                // which would persist an invalid disabled-with-factor state (the factor is dropped
+                // by toProperties() while disabled, diverging in-memory vs replay/failover).
+                if (!newFlatJsonConfig.getFlatJsonEnable()
+                        && (properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_NULL_FACTOR)
+                        || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_SPARSITY_FACTOR)
+                        || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_COLUMN_MAX))) {
+                    throw new RuntimeException("flat JSON configuration must be set after enabling flat JSON.");
+                }
+            }
+            newFlatJsonConfig.incVersion();
             GlobalStateMgr.getCurrentState().getLocalMetastore().modifyFlatJsonMeta(db, olapTable, newFlatJsonConfig);
         } catch (Exception e) {
             isModifiedSuccess = false;
@@ -2965,7 +3005,103 @@ public class SchemaChangeHandler extends AlterHandler {
             locker.unLockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(olapTable.getId()), LockType.WRITE);
         }
 
+        if (isModifiedSuccess) {
+            try {
+                for (Partition partition : partitions) {
+                    updateFlatJsonPartitionTabletMeta(db, olapTable.getName(), partition.getName(),
+                            olapTable.getFlatJsonConfig());
+                }
+            } catch (DdlException e) {
+                LOG.warn("Failed to execute updateFlatJsonPartitionTabletMeta", e);
+            }
+        }
+
         return isModifiedSuccess;
+    }
+
+    /**
+     * Update one specified partition's flat_json_config by partition name of table
+     */
+    public void updateFlatJsonPartitionTabletMeta(Database db,
+                                                  String tableName,
+                                                  String partitionName,
+                                                  FlatJsonConfig flatJsonConfig) throws DdlException {
+        // be id -> Set<tablet id>
+        Map<Long, Set<Long>> beIdToTabletId = Maps.newHashMap();
+        OlapTable olapTable = (OlapTable) GlobalStateMgr.getCurrentState().getLocalMetastore()
+                .getTable(db.getFullName(), tableName);
+        if (olapTable == null) {
+            throw new DdlException("Table[" + tableName + "] does not exist");
+        }
+
+        Locker locker = new Locker();
+        locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(olapTable.getId()), LockType.READ);
+        try {
+            Partition partition = olapTable.getPartition(partitionName);
+            if (partition == null) {
+                throw new DdlException(
+                        "Partition[" + partitionName + "] does not exist in table[" + olapTable.getName() + "]");
+            }
+
+            for (PhysicalPartition physicalPartition : partition.getSubPartitions()) {
+                MaterializedIndex baseIndex = physicalPartition.getLatestBaseIndex();
+                for (Tablet tablet : baseIndex.getTablets()) {
+                    for (Replica replica : ((LocalTablet) tablet).getImmutableReplicas()) {
+                        Set<Long> tabletSet = beIdToTabletId.computeIfAbsent(replica.getBackendId(), k -> Sets.newHashSet());
+                        tabletSet.add(tablet.getId());
+                    }
+                }
+            }
+        } finally {
+            locker.unLockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(olapTable.getId()), LockType.READ);
+        }
+
+        int totalTaskNum = beIdToTabletId.keySet().size();
+        if (totalTaskNum == 0) {
+            return;
+        }
+        MarkedCountDownLatch<Long, Set<Long>> countDownLatch = new MarkedCountDownLatch<>(totalTaskNum);
+        AgentBatchTask batchTask = new AgentBatchTask();
+        for (Map.Entry<Long, Set<Long>> kv : beIdToTabletId.entrySet()) {
+            countDownLatch.addMark(kv.getKey(), kv.getValue());
+            TabletMetadataUpdateAgentTask task = TabletMetadataUpdateAgentTaskFactory
+                    .createFlatJsonConfigUpdateTask(kv.getKey(), kv.getValue(), flatJsonConfig);
+            task.setLatch(countDownLatch);
+            batchTask.addTask(task);
+        }
+        if (!FeConstants.runningUnitTest) {
+            AgentTaskQueue.addBatchTask(batchTask);
+            AgentTaskExecutor.submit(batchTask);
+            LOG.info("send update flat_json_config tablet meta task for table {}, partition {}, number: {}",
+                    tableName, partitionName, batchTask.getTaskNum());
+
+            long timeout = Config.tablet_create_timeout_second * 1000L * totalTaskNum;
+            timeout = Math.min(timeout, Config.max_create_table_timeout_second * 1000L);
+            boolean ok = false;
+            try {
+                ok = countDownLatch.await(timeout, TimeUnit.MILLISECONDS);
+            } catch (InterruptedException e) {
+                LOG.warn("InterruptedException: ", e);
+            }
+
+            if (!ok || !countDownLatch.getStatus().ok()) {
+                String errMsg = "Failed to update partition[" + partitionName + "] flat_json_config tablet meta.";
+                AgentTaskQueue.removeBatchTask(batchTask, TTaskType.UPDATE_TABLET_META_INFO);
+                if (!countDownLatch.getStatus().ok()) {
+                    errMsg += " Error: " + countDownLatch.getStatus().getErrorMsg();
+                } else {
+                    List<Map.Entry<Long, Set<Long>>> unfinishedMarks = countDownLatch.getLeftMarks();
+                    List<Map.Entry<Long, Set<Long>>> subList =
+                            unfinishedMarks.subList(0, Math.min(unfinishedMarks.size(), 3));
+                    if (!subList.isEmpty()) {
+                        errMsg += " Unfinished mark: " + Joiner.on(", ").join(subList);
+                    }
+                }
+                errMsg += ". This operation maybe partial successfully, You should retry until success.";
+                LOG.warn(errMsg);
+                throw new DdlException(errMsg);
+            }
+        }
     }
 
     // return true means that the modification of FEMeta is successful,
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/FlatJsonConfig.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/FlatJsonConfig.java
@@ -36,11 +36,15 @@ public class FlatJsonConfig implements Writable {
     @SerializedName("flatJsonColumnMax")
     private int flatJsonColumnMax;
 
+    @SerializedName("configVersion")
+    private long configVersion;
+
     public FlatJsonConfig(boolean enabled, double nullFactor, double sparsityFactor, int columnMax) {
         this.flatJsonEnable = enabled;
         this.flatJsonNullFactor = nullFactor;
         this.flatJsonSparsityFactor = sparsityFactor;
         this.flatJsonColumnMax = columnMax;
+        this.configVersion = 0;
     }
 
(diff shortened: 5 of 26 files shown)


fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
```diff
@@ -573,11 +573,21 @@ public Void visitModifyTablePropertiesClause(ModifyTablePropertiesClause clause,
...
-                boolean isSuccess = schemaChangeHandler.updateFlatJsonConfigMeta(db, table.getId(),
-                        properties, TTabletMetaType.FLAT_JSON_CONFIG);
-                if (!isSuccess) {
-                    throw new DdlException("modify flat json config of FEMeta failed");
-
+                if (table.isCloudNativeTable()) {
+                    Locker locker = new Locker();
+                    locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(table.getId()), LockType.WRITE);
+                    try {
+                        schemaChangeHandler.processLakeTableAlterMeta(clause, db, (OlapTable) table);
```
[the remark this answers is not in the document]
moweonlee: Same as the resolved thread on #73514: a multi-property flat_json ALTER is rejected earlier in AlterTableClauseAnalyzer ("Can only set one table property at a time") for both lake and shared-nothing — flat_json.* isn't in the same-prefix allow-list (only dynamic_partition. / binlog.). So the shared-nothing path doesn't accept a multi-property flat_json statement either; lake isn't newly rejecting a previously-supported form.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -2731,16 +2752,131 @@ public boolean updateFlatJsonConfigMeta(Database db, Long tableId, Map<String, S
...
+        try {
+            Partition partition = olapTable.getPartition(partitionName);
+            if (partition == null) {
+                throw new DdlException(
+                        "Partition[" + partitionName + "] does not exist in table[" + olapTable.getName() + "]");
+            }
+
+            for (PhysicalPartition physicalPartition : partition.getSubPartitions()) {
+                MaterializedIndex baseIndex = physicalPartition.getLatestBaseIndex();
+                for (Tablet tablet : baseIndex.getTablets()) {
```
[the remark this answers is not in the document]
moweonlee: Same as the resolved thread on #73514: base-index tablets are updated synchronously; rollup / MV-index tablets are reconciled by the heartbeat path (handleSetTabletFlatJsonConfig), which iterates every reported tablet, not just the base index — the same versioned-reconciliation design as binlog. So rollup tablets converge.

moweonlee:
This is the clean, squashed re-submission of #73514 (per @kevincai's suggestion): identical logic in 2 commits on the latest `main`, fully re-verified.

@kevincai @gengjun-git @meegoo — could you take a look when you have a moment? The change follows the existing binlog-config propagation pattern (FE version bump + immediate push to BEs, plus heartbeat reconciliation for stragglers), and all version fields on the Thrift/proto structs are optional for mixed-version compatibility. Happy to rebase or address feedback right away. Thanks! 🙏

moweonlee:
I've added some extra context on the linked issue to make this PR easier to review.

- **Why it matters / how we ran into it** → [#73846 context & motivation](https://github.com/StarRocks/starrocks/issues/73846#issuecomment-4751968723)
- **Design rationale** (why it mirrors `binlog`, and why reconciliation is needed) → [design notes](https://github.com/StarRocks/starrocks/issues/73846#issuecomment-4751981031)

**tl;dr:** the FE accepts `ALTER TABLE ... SET ("flat_json.*" = ...)` and `SHOW CREATE TABLE`
reflects the new values, but the BE never receives them, so existing tablets keep using the
CREATE-time `FlatJsonConfig`. This PR wires up the missing FE → BE propagation using the same
pattern `binlog` config already uses (versioned tablet task + `ReportHandler` reconciliation),
and adds the BE-propagation test coverage that is missing today.

We rely on flat_json across several scenarios and would like to help close this gap, so any
review attention here would be much appreciated, and we're happy to iterate on anything the
maintainers prefer.

**On PR size:** if this is easier to review in smaller pieces, we're glad to split it up — for
example landing the core `ALTER → BE` push first and the heartbeat-based reconciliation as a
follow-up, or staging it as interface / BE-apply → FE-push → reconciliation. If you have a
preferred decomposition, just let us know the shape you'd like and we'll restructure it.

CelerData-Reviewer:
@codex review

kevincai (commented):
Two findings from a local review of this PR (one correctness, one performance). Posted as inline comments below.

Generated with [Claude Code](https://claude.com/claude-code)

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJob.java
```diff
@@ -123,6 +139,9 @@ protected void updateCatalog(Database db, OlapTable table, boolean isReplay) {
                     String.valueOf(compactionStrategy));
             table.getTableProperty().buildCompactionStrategy();
         }
+        if (metaType == TTabletMetaType.FLAT_JSON_CONFIG && flatJsonConfig != null) {
+            table.setFlatJsonConfig(flatJsonConfig);
```
kevincai, line 143: **Lake flat_json config is not persisted durably.**

This branch only sets the transient `flatJsonConfig` field. Unlike the sibling `ENABLE_PERSISTENT_INDEX` and `COMPACTION_STRATEGY` branches just above (which call `table.getTableProperty().modifyTableProperties(...)`), it never writes the flat_json values into the serialized `properties` map.

`TableProperty.flatJsonConfig` has no `@SerializedName`, so it is not serialized directly; it is rebuilt from the `properties` map via `buildFlatJsonConfig()` in `gsonPostProcess()`. As a result, the config applied here is not durable: once the async alter job is GC'd and a new FE image is written, the flat_json config (enable/factors/version) is lost/reverted on reload for cloud-native tables.

The shared-nothing path avoids this because `modifyFlatJsonMeta` logs `flatJsonConfig.toProperties()`. Consider mirroring the compaction-strategy path here, e.g. `modifyTableProperties(...)` with `flatJsonConfig.toProperties()` (so `flat_json.version` and the factors land in the persisted map) before/instead of the bare `setFlatJsonConfig`.

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
```diff
@@ -2208,6 +2216,63 @@ private static void handleSetTabletBinlogConfig(long backendId, Map<Long, TTable
...
+                    continue;
+                }
+
+                OlapTable olapTable = (OlapTable) GlobalStateMgr.getCurrentState().getLocalMetastore()
+                        .getTable(db.getId(), tableId);
+                if (olapTable == null) {
+                    continue;
+                }
+                Locker locker = new Locker();
+                locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(olapTable.getId()), LockType.READ);
```
kevincai, line 2252: **Per-tablet table lock taken on every report, even for tables without flat_json.**

This reconciliation runs on every BE tablet report and takes a per-table READ lock (`lockTablesWithIntensiveDbLock`) for every tablet, then checks `containsFlatJsonConfig()` only after acquiring the lock. Because an unset `flat_json_config_version` is treated as stale (line ~2226, "Treat unset version as stale...") rather than skipped, tablets belonging to tables that never use flat_json still incur a lock acquire/release per tablet per report cycle.

The analogous `handleSetTabletBinlogConfig` avoids this with an early `if (!tabletInfo.isSetBinlog_config_version()) { continue; }` (line ~2157) before any table lookup or lock. Suggest adding the same early-skip (and handling the orphan-tablet case without locking unrelated tables) to avoid the hot-path lock-traffic regression at scale.

fe/fe-core/src/main/java/com/starrocks/alter/LakeTableAlterMetaJob.java
moweonlee: @kevincai 

 It was Good catch — durability was the right concern. Our earlier `set → restart` tests
passed because the ALTER is still replayed from the editlog; the loss only shows up
after a checkpoint folds it into an image and GCs the journal, which we'd missed.

 Digging in, the write side was actually fine (`setFlatJsonConfig()` →
`modifyTableProperties()` already persists to the map). The real bug is on the read
side: `buildFlatJsonConfig()` runs on every image load and read the factors via
`analyzerDoubleProp`/`analyzeIntProp`, which `properties.remove()` the keys — so each
reload stripped them and they reverted to defaults after the next checkpoint. Common
to both modes (since #57379), not lake-specific.

Fixed in f4f6cb69229a39123b75bc968c4a99c4cea22e2f (read with `get()`, fall back to defaults on a bad value). Verified
on shared-nothing and shared-data: the config survives checkpoint+restart, and
toggling `flat_json.enable` flips the on-disk segment layout (footer dumps).

e4b33e4f4956e6106fdf9f23b2dc27fc7648bf92 separately renders flat_json in `SHOW CREATE TABLE` for cloud-native tables.

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
```diff
@@ -2208,6 +2216,63 @@ private static void handleSetTabletBinlogConfig(long backendId, Map<Long, TTable
...
+    private static void handleSetTabletFlatJsonConfig(long backendId, Map<Long, TTablet> backendTablets) {
+        List<Pair<Long, FlatJsonConfig>> tabletToFlatJsonConfig = Lists.newArrayList();
+
+        TabletInvertedIndex invertedIndex = GlobalStateMgr.getCurrentState().getTabletInvertedIndex();
+        for (TTablet backendTablet : backendTablets.values()) {
+            for (TTabletInfo tabletInfo : backendTablet.tablet_infos) {
+                long tabletId = tabletInfo.getTablet_id();
+                // Treat unset version as stale so orphan tablets get reconciled.
+                long beFlatJsonConfigVersion =
+                        tabletInfo.isSetFlat_json_config_version() ? tabletInfo.flat_json_config_version : 0;
```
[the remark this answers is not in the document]
moweonlee: This is an intentional design choice, not an oversight. Treating an unset/legacy report
as version 0 is what lets the version mechanism reconcile *any* backend that is missing
the config — not just one that's a version behind, but also a backend that never
received it (newly added, re-created replica, or down during the ALTER). Verified on a
local cluster: a BE that was down during `ALTER ... flat_json.enable=true` (so its
tablets had no config at all) received version=1 via reconciliation on its next report
after rejoining, and then columnizes.

If we skipped unset reports instead, such a backend would *silently* never columnize —
there's no warning on the behind path (the WARN only fires when a BE is ahead of the FE).

The trade-off is deliberate: the only downside is a transient re-push to a pre-#74747
BE during a rolling upgrade (it can't apply FLAT_JSON_CONFIG yet, so it ignores the task
harmlessly and self-heals once upgraded). New BEs converge in one report cycle. We chose
the auto-reconcile robustness over avoiding that transient churn. If the mixed-version
churn is a concern, a follow-up can have new BEs report an explicit 0 (vs. unset) so we
skip only truly legacy reports while keeping the reconcile benefit.

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
moweonlee: Fixed in 093daa6 — moved the `containsFlatJsonConfig()` check before the lock so only
flat_json tables acquire it (mirrors the early-skip in `handleSetTabletBinlogConfig`).
Tablets of tables that don't use flat_json now skip without taking the per-tablet lock.
The version comparison stays under the lock, and a re-read null-guards the lock-free
check, so reconciliation behavior is unchanged — `ReportHandlerTest` and a local-cluster
test (a backend missing the config is still reconciled) both pass.

moweonlee:
## flat_json config — progress & verification summary

What began as a small durability fix grew, through review, into a set of changes that
make the table-level `flat_json` config behave consistently across **shared-nothing**
and **shared-data**. Summary of where things stand, with a verification matrix built
along the way.

### Changes so far

| Commit | Type | What it does |
|--------|------|--------------|
| `89753d3` | Enhancement | **Propagate flat_json config ALTER to BE.** Adds `flat_json_config_version` to `FlatJsonConfig` / `TTabletInfo`; `ALTER TABLE SET ("flat_json.*"=...)` dispatches a versioned tablet-meta task to the BE, and `ReportHandler.handleSetTabletFlatJsonConfig` reconciles any backend that fell behind on its next report. Shared-data carries the config through `LakeTableAlterMetaJob` and applies it in lake `schema_change` / `txn_log_applier`. |
| `c486958` | UT | **Reconciliation unit tests** for `handleSetTabletFlatJsonConfig` (stale / orphan / up-to-date / ahead / table-without-flat_json). |
| `f4f6cb6` | Enhancement | **Durability** — `buildFlatJsonConfig()` read the factors via `analyzerDoubleProp`/`analyzeIntProp`, which `properties.remove()` the keys. Since it runs on every image load, factors were stripped and reverted to defaults after the next checkpoint. Common to both modes (since #57379). Now read non-destructively with `get()`. |
| `e4b33e4` | BugFix | **Display** — `LakeTable.getUniqueProperties()` didn't render `flat_json.*`, so `SHOW CREATE TABLE` omitted it for cloud-native tables even though it was applied. Now rendered like `OlapTable`. |
| `093daa6` | Enhancement | **Hot-path lock** — the report-handler reconciliation took a per-tablet table lock for every tablet before checking `containsFlatJsonConfig()`. Moved the check before the lock so non-flat_json tables skip without locking (mirrors `handleSetTabletBinlogConfig`). |

### Functional verification matrix

Method: SQL (`ALTER` / `SHOW CREATE`) + **direct segment-footer dumps** (shared-nothing:
local; shared-data: MinIO object) + BE/CN logs, on real clusters (SN: 1 FE + 2 BE; SD:
3 FE + 2 CN + MinIO). Test data: field `"a"` in 100% of rows, field `"freq"` in 70%.

> Legend — **own column**: field extracted into its own typed sub-column; **not extracted**: field kept in the shared raw-JSON `remain` column (no data loss).

| # | Capability | Scenario | shared-nothing | shared-data |
|---|-----------|----------|----------------|-------------|
| 1 | ALTER → storage (enable) | enable `true→false→true`, insert each step | columnized → **RAW** → columnized | **same** |
| 2 | Sparsity threshold ↓ | `1.0 → 0.5` | `freq`: **not extracted → own column** | **same** |
| 3 | Sparsity threshold ↑ | `0.5 → 1.0` | `freq`: **own column → not extracted** (kept in raw `remain`) | **same** |
| 4 | Durability | factor ALTER → checkpoint + restart | factor survives; new data still columnizes | **same** (same FE-image path, same fix) |
| 5 | No silent ignore / recovery | a node that missed the ALTER | recovers — **version reconciliation** (FE re-push, unset treated as 0) | recovers — **shared-storage reload** (CN); *same outcome, different mechanism* |

Sparsity semantics confirmed against the BE flattener (`json_path_deriver.cpp`): a field
is extracted iff `hits >= total_rows * sparsity_factor`; a field below the threshold is
preserved in the shared `remain` column (no data loss). The flattener is shared BE/CN
code, so rows 1–3 are identical by construction and confirmed by footer dumps in both
modes.

### Design philosophy this validates

- **Config drives actual storage, not just metadata.** Every `ALTER` is verified at the
  segment-footer level (columnized ↔ raw, `remain` ↔ typed column) in both modes, and
  mid-stream changes take effect on subsequently written data.
- **No silent ignore (unset treated as version 0).** A backend/replica that never
  received the config (newly added, recreated, or down during the ALTER) is reconciled
  rather than silently left un-columnized. If unset reports were skipped instead, such a
  node would never columnize — with no warning (the WARN path only fires when a BE is
  *ahead* of the FE).
- **Durability is common; recovery differs by design.** The config is persisted in FE
  metadata (editlog + image) for both modes, so the durability fix applies to both.
  Recovery differs because the source of truth differs: shared-nothing keeps a local
  per-replica copy that can drift → it needs version reconciliation; shared-data keeps
  the config in shared storage as the single source of truth → the CN reloads it, so no
  version reconciliation is needed (shared storage plays that role).

CelerData-Reviewer:
@codex review

moweonlee:
@CelerData-Reviewer @kevincai

Added unit tests to raise the incremental coverage that the Fe Cover Checker flagged
on the new FE lines (44df0e5):

- `CreateLakeTableTest#testCreateLakeTableWithFlatJson` — covers
  `LakeTable.getUniqueProperties()` rendering `flat_json.*` for cloud-native tables
  (enabled with factors, and disabled).
- `LakeTableAlterFlatJsonConfigTest` — covers `LakeTableAlterMetaJob.updateCatalog`
  persisting the flat_json config (incl. version) into the serialized properties map
  so it survives an FE image reload (leader and replay paths).

The BE UT red was a SIGTERM/timeout — no assertion failures in the logs, and the
interrupted tests are unrelated to flat_json — so it looks like a CI flake.

moweonlee:
@CelerData-Reviewer could you trigger CI on the latest commit? Thanks — it
adds the coverage tests and should clear the BE UT timeout.

moweonlee:
## flat_json config — verification summary (updated after rebase)

Rebased onto latest main (conflict in `be/src/storage/tablet.cpp` resolved: kept `flat_json_config.h`, dropped the header main relocated to `common/`). Commits:

| Commit | Type | What it does |
|--------|------|--------------|
| `cd22f3a` | Enhancement | Propagate flat_json config ALTER to BE (versioned tablet-meta task + `ReportHandler` reconciliation; shared-data via `LakeTableAlterMetaJob`). |
| `6bc003b` | UT | Reconciliation unit tests for `handleSetTabletFlatJsonConfig`. |
| `fd86941` | Enhancement | Durability — read factors non-destructively in `buildFlatJsonConfig()` so they survive an image checkpoint (`get()` instead of the `remove()`-ing analyzers). |
| `6099a43` | BugFix | Display — render `flat_json.*` in `SHOW CREATE TABLE` for cloud-native tables. |
| `28384b6` | Enhancement | Narrow the reconciliation table lock to flat_json tables (skip non-flat_json before locking). |
| `96ab597` | UT | Coverage tests (LakeTable display + lake ALTER durability). |

### Functional verification matrix

Method: SQL + **segment-footer dumps** (SN: local; SD: MinIO object) + BE/CN logs, on real clusters (SN 1FE+2BE, SD 3FE+2CN+MinIO). All four `flat_json.*` factors were changed mid-stream and the on-disk layout re-checked.

| # | Capability | Scenario | shared-nothing | shared-data |
|---|-----------|----------|----------------|-------------|
| 1 | `enable` → storage | `true→false→true` | columnized → **RAW** → columnized | **same** |
| 2 | `sparsity.factor` ↓ | `1.0→0.5` (field @70%) | `remain` → **own column** | **same** |
| 3 | `sparsity.factor` ↑ | `0.5→1.0` | own column → **`remain`** | **same** |
| 4 | `null.factor` | json 40% null, `0.5→0.3` | flattened → **RAW** (null-ratio gate) | shared flattener* |
| 5 | `column.max` | 8 sparse fields, `2→10` | **2 sub-cols** → **8 sub-cols** | shared flattener* |
| 6 | Durability | factor ALTER → checkpoint+restart | factor survives; new data columnizes | **same** |
| 7 | No silent ignore / recovery | node missed the ALTER | recovers — version reconciliation | recovers — shared-storage reload |

\* rows 4–5 were footer-verified on shared-nothing; the flattener (`json_path_deriver.cpp`) is shared BE/CN C++ code, so shared-data behaves identically (rows 1–3 confirmed that on both modes). Semantics match the code: a field is extracted iff `hits >= total_rows * sparsity_factor`, and `column.max` caps sub-columns except fields present in every row (always kept); below-threshold fields stay in `remain` (no data loss).

### Design philosophy validated
- **Config drives on-disk layout, not just metadata** — every ALTER re-checked at the footer.
- **No silent ignore** — an unset/legacy report is treated as version 0 and reconciled, so a node that missed the ALTER still columnizes (skipping unset would silently never columnize).
- **Durability common; recovery differs by design** — both modes persist via FE image; SN reconciles by version re-push, SD reloads from shared storage (its single source of truth), so it needs no version reconciliation.

CelerData-Reviewer:
@codex review

kevincai:
@moweonlee unfortunately, the code conflicts with main branch again.

moweonlee:
@kevincai Rebased onto latest main (b0d12f75) — only an include-path conflict from main's `storage/primitive/` → `storage_primitive/` move, no flat_json logic changes. FE compiles cleanly.

moweonlee:
**Verification summary — flat_json config propagation (#74747)**

Beyond the unit tests, I ran empirical verification on two real clusters — **shared-nothing (1 FE + 2 BE)** and **shared-data (3 FE + 2 CN + MinIO)** — checking not just SQL/metadata but **actual on-disk segment footers** (SN local; SD via MinIO → `meta_tool show_segment_footer`).

**All four factors verified to drive on-disk layout** (footer re-dumped after each mid-stream ALTER; identical behavior on both modes):

| Factor | Scenario | Result |
|--------|----------|--------|
| `enable` | `true→false→true` | columnized → RAW → columnized |
| `sparsity.factor` | `1.0↔0.5` (field @70%) | `remain` ↔ own sub-column |
| `null.factor` | 40% null, `0.5→0.3` | flattened → RAW (null-ratio gate) |
| `column.max` | 8 sparse fields, `2→10` | 2 → 8 sub-columns |

**Robustness (all empirical unless noted):**

| Area | Test | Result & evidence |
|------|------|-------------------|
| Correctness | Result invariance | 5 queries identical across columnized / raw / mixed layouts; footers confirm layouts actually differ |
| Correctness | Compaction over mixed-config segments | results unchanged before/after; merged segment re-flattens per current config |
| Correctness | Nested / array / mixed-type JSON | nested `o.x`,`o.y` typed-flattened; results invariant; no crash |
| Correctness | Typed predicate pushdown | `ColumnAccessPath: [/j/o/x(bigint)]` on a nested sub-column |
| Concurrency | Concurrent ALTER ×6 | no error/deadlock; single converged config; version monotonic |
| Concurrency | ALTER during 10k-row INSERT flood | 0 errors; 10000 rows + checksum exact; no torn config |
| Distributed recovery | FE failover (kill leader) | config preserved across new leader; new leader serves ALTER |
| Distributed recovery | SD CN restart | post-restart writes columnize (config reloaded from shared storage) |
| Durability | Multi checkpoint × multi ALTER | config survives 2 image generations (0.44 → 0.77) |
| Durability | Old-image compatibility | missing-version → default (UT `FlatJsonConfigValidationTest`); real old-image artifact not exercised |

**Design properties confirmed:** config drives layout not just metadata; durability is common (FE editlog+image) while recovery differs by design (SN version-reconciliation vs SD shared-storage reload); an unset/legacy report is treated as version 0 so a node that missed the ALTER still columnizes; flatten is query-transparent.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
```diff
@@ -605,11 +605,21 @@ public Void visitModifyTablePropertiesClause(ModifyTablePropertiesClause clause,
...
```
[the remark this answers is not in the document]
moweonlee: Duplicate of https://github.com/StarRocks/starrocks/pull/74747#discussion_r3403503686 — multi-key `flat_json.*` ALTERs are already rejected in `AlterTableClauseAnalyzer` before reaching this path, on both modes. No regression.

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
```diff
@@ -2208,6 +2216,69 @@ private static void handleSetTabletBinlogConfig(long backendId, Map<Long, TTable
...
```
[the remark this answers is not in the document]
moweonlee: Duplicate of the "Treat legacy flat_json tablets as stale" thread resolved on #73514 (https://github.com/StarRocks/starrocks/pull/73514#discussion_r3339925963). The version field is new in this work, so there is no stored version to migrate. Before this fix, a flat_json ALTER never reached the BE at all — an upgraded cluster simply stays in that old state, and this PR is what fixes it going forward. Those old ALTERs happened when no delivery path existed, so there is nothing to replay; back-filling them would be a separate one-time migration, kept out of scope as agreed in that thread.

moweonlee:
**Status update — `bddf49e`**

Two follow-up commits on top of the reviewed history (normal push, no force — the diff since the last review is exactly these two):

1. **`a5b4116` — UT, integration with the pre-existing test suite**: made the branch run green with the existing test code — `StorageValueTypesTest`'s exact-match expectation now includes the `version` field that `FlatJsonConfig::to_string()` renders (this was the BE UT failure on the previous run). Test-only change.

2. **`bddf49e` — convention / code-style cleanup (SonarCloud)**: no functional change, just style and de-duplication:
   - `flat_json.*` SHOW CREATE TABLE rendering deduplicated into `OlapTable.appendFlatJsonProperties()`, shared by shared-nothing and cloud-native tables;
   - `waitForTabletMetaBatchTask()` extracted, restoring the thread interrupt status (java:S2142) there and on the pre-existing catch the scan flagged.
   - Some structural duplication with the binlog update path intentionally remains: the propagation deliberately follows that established pattern, and rewriting the pre-existing binlog/generic methods just to chase the duplication metric would be over-engineering for this PR.

Since code changed, the full verification was re-run on the final tree — no issues:

| Check | Result |
|---|---|
| FE compile (fe-core) | BUILD SUCCESS |
| ReportHandlerTest (reconciliation scenarios) | 20/20 |
| CreateLakeTableTest (SHOW CREATE rendering, incl. flat_json display) | 21/21 |
| FlatJsonConfigValidationTest | 9/9 |
| LakeTableAlterFlatJsonConfigTest | 2/2 |
| BE StorageValueTypesTest expectation | matches the actual `to_string()` output recorded by the last BE UT run, character-for-character |

moweonlee:
@kevincai Ran the full FE and BE UT suites locally on the current head (`bddf49e`):

- FE UT (20,438 tests) and BE UT (`starrocks_test` + `storage_primitive_test`) pass locally; the handful of local failures also reproduce identically on pristine upstream/main, so they look unrelated to this PR.
- `StorageValueTypesTest.FlatJsonConfigUpdateAndSerializationSemantics` — the test that failed in the last BE UT run — now passes with the updated expectation.

Hopefully this round goes through — ready for a CI re-run whenever convenient.

kevincai:
> @kevincai Ran the full FE and BE UT suites locally on the current head (`bddf49e`):
> 
> * FE UT (20,438 tests) and BE UT (`starrocks_test` + `storage_primitive_test`) pass locally; the handful of local failures also reproduce identically on pristine upstream/main, so they look unrelated to this PR.
> * `StorageValueTypesTest.FlatJsonConfigUpdateAndSerializationSemantics` — the test that failed in the last BE UT run — now passes with the updated expectation.
> 
> Hopefully this round goes through — ready for a CI re-run whenever convenient.

np.

CelerData-Reviewer:
@codex review

be/src/agent/agent_task.cpp
```diff
@@ -981,6 +982,13 @@ void run_update_meta_info_task(const std::shared_ptr<UpdateTabletMetaInfoAgentTa
                 binlog_config.update(tablet_meta_info.binlog_config);
                 tablet->update_binlog_config(binlog_config);
                 break;
+            case TTabletMetaType::FLAT_JSON_CONFIG: {
+                LOG(INFO) << "update tablet:" << tablet->tablet_id() << " flat_json_config";
```
kevincai: may worth print the version in the logline.

fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
```diff
@@ -605,11 +605,21 @@ public Void visitModifyTablePropertiesClause(ModifyTablePropertiesClause clause,
...
                     properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_SPARSITY_FACTOR) ||
                     properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_COLUMN_MAX)) {
-                boolean isSuccess = schemaChangeHandler.updateFlatJsonConfigMeta(db, table.getId(),
-                        properties, TTabletMetaType.FLAT_JSON_CONFIG);
-                if (!isSuccess) {
-                    throw new DdlException("modify flat json config of FEMeta failed");
-
+                if (table.isCloudNativeTable()) {
+                    Locker locker = new Locker();
+                    locker.lockTablesWithIntensiveDbLock(db.getId(), Lists.newArrayList(table.getId()), LockType.WRITE);
```
kevincai:
```suggestion
                    locker.lockTableWithIntensiveDbLock(db.getId(), table.getId(), LockType.WRITE);
```

and also the unlock statement: `unLockTableWithIntensiveDbLock()`

fe/fe-core/src/main/java/com/starrocks/catalog/FlatJsonConfig.java
```diff
@@ -56,6 +61,10 @@ public FlatJsonConfig() {
     }
 
     public void buildFromProperties(Map<String, String> properties) {
+        if (properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_VERSION)) {
+            configVersion = Long.parseLong(properties.get(
```
kevincai, line 65: Thought this is an internal info that should not be visible to customer, or at most readonly to customer?

Is it intended to allow user set the version by altering the table property?

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -2938,16 +2959,139 @@ public boolean updateFlatJsonConfigMeta(Database db, Long tableId, Map<String, S
...
```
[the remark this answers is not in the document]
moweonlee: Duplicate of https://github.com/StarRocks/starrocks/pull/74747#discussion_r3403504302 — non-base indexes converge through the heartbeat reconciliation, as covered there. Verified locally with an actual rollup: the base tablet got the synchronous push at ALTER time, the rollup tablet was reconciled on its next report (+4s and +39s across two runs), and both tablet metas then hold the identical config (version 1). The pre-report window only affects the layout of newly written segments, never query results.

fe/fe-core/src/main/java/com/starrocks/catalog/FlatJsonConfig.java
moweonlee: Hi @kevincai, checked both points:

> Thought this is an internal info that should not be visible to customer, or at most readonly to customer?

Correct — it's not visible: `SHOW CREATE TABLE` and `information_schema.tables_config` don't render `flat_json.version` (verified locally). The key only exists as the internal persistence form — `flatJsonConfig` is rebuilt from the properties map on image load, so the version must ride in it to survive FE restart (same pattern as `binlogConfig`).

> Is it intended to allow user set the version by altering the table property?

No — `AlterTableClauseAnalyzer` has no branch for it, so setting `flat_json.version` via ALTER or CREATE is rejected with "Unknown properties" (verified locally).

Thanks for the careful review.

be/src/agent/agent_task.cpp
moweonlee: Done in 2eb70d9 — the log line now prints the config version (emitted after the thrift payload is parsed, so the version is available).

fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
moweonlee: Done in 2eb70d9 — switched to `lockTableWithIntensiveDbLock` / `unLockTableWithIntensiveDbLock`.

moweonlee:
**On the relationship with #75891** (cc @kevincai — re: "Could be similar", and @sevev)

We verified the relationship empirically on a local shared-data cluster (1 FE + 1 CN + MinIO, `enable_file_bundling=true`).

Setup — chosen so the config source is distinguishable from the result:
- global CN config: `enable_json_flat = true` (default)
- table: created with `flat_json.enable = false`
- the table config overrides the global one when consumed correctly, so every segment of this table is **expected to be RAW**; if a segment comes out flattened, the writer used the global default instead of the table config.

Each INSERT loads 1,000 rows of uniform-key JSON, and we check the written segment's footer (`children_columns`) for the physical layout:

| # | BE binary | #74747 | procedure | result | expected RAW? |
|---|---|---|---|---|---|
| 1 | #75891's base | no | CREATE → INSERT right away (metadata cache **warm**) | RAW | ✅ |
| 2 | #75891's base | no | restart CN (cache **cold**) → INSERT into the same table | **flattened** | ❌ global default used |
| 3 | our branch | **yes** | same cold procedure | **flattened** | ❌ same behavior |

We were also able to reproduce the cold-cache behavior #75891 addresses (row 2), and row 3 shows #74747 does not affect it. 

 The two PRs don't overlap — they fill different gaps left by the original flat_json implementation: 

- #74747 completes the delivery side (ALTER propagation to tablet metadata, shared-nothing + shared-data)
-  #75891 completes the shared-data write path that consumes it. 

With both merged, flat_json should behave consistently end to end. Thanks @sevev for catching the consumption path.

moweonlee:
@kevincai this branch (`40259da`) is ready for another CI run whenever convenient — the previous BE coverage gap is addressed with two targeted UTs, and the review comments have been applied.

CelerData-Reviewer:
@codex review

kevincai:
> **On the relationship with #75891** (cc @kevincai — re: "Could be similar", and @sevev)
> 
> We verified the relationship empirically on a local shared-data cluster (1 FE + 1 CN + MinIO, `enable_file_bundling=true`).
> 
> Setup — chosen so the config source is distinguishable from the result:
> 
> * global CN config: `enable_json_flat = true` (default)
> * table: created with `flat_json.enable = false`
> * the table config overrides the global one when consumed correctly, so every segment of this table is **expected to be RAW**; if a segment comes out flattened, the writer used the global default instead of the table config.
> 
> Each INSERT loads 1,000 rows of uniform-key JSON, and we check the written segment's footer (`children_columns`) for the physical layout:
> 
> #	BE binary	#74747	procedure	result	expected RAW?
> 1	#75891's base	no	CREATE → INSERT right away (metadata cache **warm**)	RAW	✅
> 2	#75891's base	no	restart CN (cache **cold**) → INSERT into the same table	**flattened**	❌ global default used
> 3	our branch	**yes**	same cold procedure	**flattened**	❌ same behavior
> We were also able to reproduce the cold-cache behavior #75891 addresses (row 2), and row 3 shows #74747 does not affect it.
> 
> The two PRs don't overlap — they fill different gaps left by the original flat_json implementation:
> 
> * [[Enhancement] Propagate flat_json config ALTER to BE via versioned task #74747](https://github.com/StarRocks/starrocks/pull/74747) completes the delivery side (ALTER propagation to tablet metadata, shared-nothing + shared-data)
> * [[BugFix] Make flat_json config reliable on the shared-data write path #75891](https://github.com/StarRocks/starrocks/pull/75891) completes the shared-data write path that consumes it.
> 
> With both merged, flat_json should behave consistently end to end. Thanks @sevev for catching the consumption path.

Thanks for the comprehensive check!

kevincai:
@moweonlee BE compiling fails. Also the two comments from codex bot looks reasonable to me, please take a look.

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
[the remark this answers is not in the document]
moweonlee: Fixed in c0aaddee328.

A missing `flat_json_config_version` in the report is now mapped to -1 instead of 0, so it compares below any real version — including the version-0 config a table gets at CREATE time, which was the gap here: 0 (unset) == 0 (create-time) meant such tablets were never repaired.

Considered starting versions at 1 instead, but version-0 configs are already persisted in existing clusters; the sentinel covers them retroactively without a migration.

Verified on a live cluster in addition to the new unit test: stripped the config from a replica's tablet meta with `meta_tool`, confirmed the old FE never repairs it across report cycles, and the fixed FE restores the version-0 config on the first report.

fe/fe-core/src/main/java/com/starrocks/alter/AlterJobExecutor.java
```diff
@@ -605,11 +605,21 @@ public Void visitModifyTablePropertiesClause(ModifyTablePropertiesClause clause,
...
-                boolean isSuccess = schemaChangeHandler.updateFlatJsonConfigMeta(db, table.getId(),
-                        properties, TTabletMetaType.FLAT_JSON_CONFIG);
-                if (!isSuccess) {
-                    throw new DdlException("modify flat json config of FEMeta failed");
-
+                if (table.isCloudNativeTable()) {
+                    Locker locker = new Locker();
+                    locker.lockTableWithIntensiveDbLock(db.getId(), table.getId(), LockType.WRITE);
+                    try {
+                        schemaChangeHandler.processLakeTableAlterMeta(clause, db, (OlapTable) table);
```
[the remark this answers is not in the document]
moweonlee: On the multi-property part — I looked into this, and flat_json itself doesn't support multi-property `SET` today, independent of this PR: the analyzer's single-property invariant ([AlterTableClauseAnalyzer.java#L218-L222](https://github.com/StarRocks/starrocks/blob/main/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/AlterTableClauseAnalyzer.java#L218-L222)) exempts only the `dynamic_partition.`/`binlog.` prefixes, so an ALTER carrying more than one `flat_json.*` key is rejected for both lake and non-lake tables before either handler path runs. The non-lake path doesn't accept the grouped form either, so the lake path isn't newly rejecting a supported form.

Tried it on a live shared-data cluster running this branch and it matches the code reading:

```sql
ALTER TABLE mp SET ('flat_json.enable'='true','flat_json.null.factor'='0.2');
-- ERROR 5064: Can only set one table property at a time.
ALTER TABLE mp SET ('flat_json.null.factor'='0.2','flat_json.sparsity.factor'='0.7');
-- ERROR 5064: Can only set one table property at a time.
ALTER TABLE mp SET ('flat_json.enable'='true');  -- OK
```

Same conclusion as the two earlier threads on #73514 ([r3330402837](https://github.com/StarRocks/starrocks/pull/73514#discussion_r3330402837)). Grouping `flat_json.*` like `binlog.*` would be nice to have, but it feels better done as a separate enhancement, decoupled from the config-propagation fix here.

moweonlee:
Rebased onto latest main for the BE compile failure (the arrow 24.0.0 thirdparty upgrade in #75868 — reproduced and confirmed locally with the updated dev image). Both codex findings are addressed above: the missing-version sentinel is fixed in c0aaddee328 with a UT, and the multi-property one turned out to be pre-existing behavior (details in that thread). Ready for another CI run.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -2741,16 +2742,34 @@ public AlterJobV2 createAlterMetaJob(AlterClause alterClause, Database db, OlapT
...
             } else if (properties.containsKey(PropertyAnalyzer.PROPERTIES_CLOUD_NATIVE_FAST_SCHEMA_EVOLUTION_V2)) {
                 return processAlterCloudNativeFastSchemaEvolutionV2Property(db, olapTable, properties).orElse(null);
+            } else if (properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_ENABLE)
+                    || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_NULL_FACTOR)
+                    || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_SPARSITY_FACTOR)
+                    || properties.containsKey(PropertyAnalyzer.PROPERTIES_FLAT_JSON_COLUMN_MAX)) {
+                FlatJsonConfig newConfig = olapTable.containsFlatJsonConfig()
+                        ? new FlatJsonConfig(olapTable.getFlatJsonConfig())
+                        : new FlatJsonConfig();
+                newConfig.buildFromProperties(properties);
```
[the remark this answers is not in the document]
moweonlee: Looking into this. The interleaving reproduces at the FE level with the real code paths (analyzer check while enabled → concurrent disable job's `updateCatalog` → `createAlterMetaJob`), so a fix with regression tests is prepared — currently verifying it against a real shared-data cluster before pushing.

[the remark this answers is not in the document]
moweonlee: Fixed in a050caae674.

Reproduced the interleaving at the FE level with the real code paths (analyzer pass while enabled → concurrent disable job's `updateCatalog()` → `createAlterMetaJob` accepting the factor on the disabled config); the accepted config then diverges between in-memory state, the persisted properties map, and a post-restart rebuild, since `toProperties()` omits factors while disabled. 

Also verified on a live shared-data cluster: normal enable/factor/disable flows are unaffected, and a race harness across the disable windows never hit the interleaving naturally — the window is sub-millisecond, so this is a low-probability path guarded for correctness. The lake branch now re-validates under the table WRITE lock, mirroring `updateFlatJsonConfigMeta()`.

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
```diff
@@ -560,6 +561,9 @@ private static void tabletReport(long backendId, Map<Long, TTablet> backendTable
         // 11. send set table binlog config to be
         handleSetTabletBinlogConfig(backendId, backendTablets);
 
+        // 11.5. flat_json_config reconciliation — re-push when BE-reported version is behind.
+        handleSetTabletFlatJsonConfig(backendId, backendTablets);
```
kevincai, line 565: One thing to consider, this TabletReport doesn't exist in shared-data mode, how does the tablet flatjson get reconciled?

kevincai (approved):

fe/fe-core/src/main/java/com/starrocks/leader/ReportHandler.java
moweonlee: Hi @kevincai — right, `tabletReport()` returns early in shared-data mode (`ReportHandler.java:489`), so this version-based reconciliation path is shared-nothing only.

Shared-data doesn't need the repair loop because there is no per-node copy that can drift: the tablet metadata is a single authoritative copy in object storage, and the flat_json config rides in it end to end —

- **CREATE**: the initial tablet metadata carries the config (`lake/tablet_manager.cpp:268-271`).
- **ALTER**: `LakeTableAlterMetaJob` sends per-tablet update tasks (`LakeTableAlterMetaJob.java:107`); each CN writes a txn log and the config lands on the metadata at publish (`lake/txn_log_applier.cpp:111-116`).
- The FE catalog flips only after every tablet succeeded (`LakeTableAlterMetaJobBase.java:261-270`), and the job is edit-logged, so it survives FE restarts and resumes (`LakeTableAlterMetaJobBase.java:702-722`).

Case by case, the two modes handle the same failures differently — shared-nothing repairs divergence after the fact, shared-data prevents it by deferring the commit:

| Case | Shared-nothing | Shared-data |
|---|---|---|
| FE commit timing | At ALTER, immediately | Only after every tablet applied |
| Node down during ALTER | FE new / BE old → drift, repaired later by this report path | FE old / metadata old → still consistent; the durable job retries until nodes return |
| Node down past job timeout | Drift remains until the BE reports again | Job cancelled, nothing changed on either side — the user just retries |
| FE restart mid-change | Config already committed; BEs converge via report | Job replays from the edit log and resumes |
| Partial application | Per-replica repair via version compare | The next successful ALTER converges all tablets (the applier does an unconditional `CopyFrom`, last-writer-wins) |

Verified on a live shared-data cluster (FE + 2 CN + MinIO built from this branch): the ALTER propagated through the txn-log path (`alter flat_json_config from version 1 to version 2 for tablet id: ...` in the CN log), and CREATE-time configs were present in the initial tablet metadata.

moweonlee:
## flat_json config — verification matrix re-run on the final head (`a050caae674`)

Re-ran the functional matrix from the [earlier summary](https://github.com/StarRocks/starrocks/pull/74747#issuecomment-4885505862) on the current head (now including the lake factor re-validation), on a fresh shared-data cluster (1 FE + 2 CN + MinIO, arm64) built from this branch. Same method: SQL + segment-footer dumps (MinIO object → `meta_tool show_segment_footer`) + CN logs. Shared-nothing rows are from the earlier run — the FE paths they cover are unchanged since, and the flattener is shared BE/CN C++ code.

| # | Capability | Scenario | shared-data (this re-run) | shared-nothing (earlier run) |
|---|-----------|----------|---------------------------|------------------------------|
| 1 | `enable` → storage | `true→false→true` | `[a,freq]` → RAW → `[a,freq]` (footers) | same |
| 2 | `sparsity.factor` ↓↑ | `1.0→0.5→1.0` (field @70%) | `remain` → own column → `remain` | same |
| 3 | `null.factor` | 40% null, `0.5→0.3` | flattened → RAW (null-ratio gate) | same |
| 4 | `column.max` | 8 fields @60% each, `2→10` | `[f1,f5,remain]` (cap=2) → all 8 | same |
| 5 | Durability | factor ALTER → FE restart | factor survives (`0.44`); new data columnizes | same (checkpoint + restart) |
| 6 | Node recovery | CN restart | post-restart writes columnize — config reloaded from shared storage | version reconciliation (SN-specific) |
| 7 | Concurrency | concurrent ALTER ×6 | no errors; single converged config; CN apply log `version 1→2→3→4→5` (monotonic) | same |
| 8 | Guard (this head's fix) | factor ALTER on a disabled table | rejected at analysis (`haven't been enabled`) | same — the analyzer gate is mode-independent |

No regressions. One incidental confirmation along the way: fields present in **100% of rows bypass `column.max`** (always extracted) — matches the deriver's always-extract rule; the cap applies to genuinely sparse fields, which is what row 4 measures (fields @60%).

sevev (approved):

PR #74753, opened 2026-06-13
[Enhancement] Route range-distribution OLAP tables by per-index distribution expressions

xiangguangyxg:
## Why I'm doing:

Range-distribution (shared-data) tables route rows to tablets by per-tablet boundaries stored in sort-key space, but the OLAP table sink could only carry a single partition-level distribution-column set. It therefore could not route different rows to materialized indexes that live in different key spaces. This is the missing sink piece for two future features — the K-tablet shadow-index rewrite job (key-column schema change) and range-distribution rollup — both of which need a base index and a new-key index to coexist in one partition and be routed independently.

## What I'm doing:

Add **per-index distribution routing** to the sink:

- **thrift:** new `TOlapTableIndexSchema.distributed_exprs` (field 9) carrying per-index routing expression trees, evaluated at the sink **sender**. Sender-only: `POlapTableIndexSchema` (proto) is unchanged, so remote write channels never route by it.
- **FE:** `OlapTableSink.createSchema` fills `distributed_exprs` for range-distribution tables with slot-refs over each index's range sort-key columns, gated to the OLAP write-sink path (dictionary / non-write callers do not emit it). For today's base-only range tables this resolves to exactly the columns the partition-level path already used, so routing is **behavior-preserving**. Also adds an optional `targetWriteIndexId` filter (write only one index; schema, partition and loaded-index lists stay 1:1 by meta id).
- **BE:** `OlapTableSchemaParam` parses `distributed_exprs` into per-index `ExprContext`s (prepare/open/close lifecycle); the range sink sender evaluates them once per chunk per index and routes via `RangeRouter`. `RangeRouter::init` validates routing-key types against the boundary types; a new `route_chunk_rows` overload routes from pre-evaluated columns; an empty `distributed_exprs` (K=1) routes to the single tablet. When an index has no `distributed_exprs`, routing falls back to the partition-level path unchanged.

No version gate is needed: StarRocks upgrades BE/CN before FE, so a newly-upgraded FE (the only one that emits the field) never runs against a BE that does not understand it, and an old FE never emits it. Non-range tables and any unset field are byte-for-byte unchanged. This is prerequisite-only: the new capability is dormant for existing tables and consumed by future work.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/OlapTableSink.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/OlapTableSink.java
@@ -164,6 +164,9 @@ public class OlapTableSink extends DataSink {
     private boolean isFromOverwrite = false;
     private boolean isMultiStatementTxn = false;
     private boolean isStreamingLoad = false;
+    // When set, only the index with this id is written (schema + partitions + loaded-index state).
+    // null (default) means write all indexes, i.e. today's behavior.
+    private Long targetWriteIndexId = null;
 
     // Conservative default for RANGE-partitioned tables under stream / routine load.
     // Both planner paths flag streaming ingest via setIsStreamingLoad(true): StreamLoadPlanner
@@ -280,6 +283,10 @@ public class OlapTableSink extends DataSink {
 
     public void setIsStreamingLoad(boolean isStreamingLoad) {
         this.isStreamingLoad = isStreamingLoad;
+    }
+
+    public void setTargetWriteIndexId(Long targetWriteIndexId) {
+        this.targetWriteIndexId = targetWriteIndexId;
     }
 
     public void complete(String mergeCondition) throws StarRocksException {
@@ -411,12 +418,12 @@ public class OlapTableSink extends DataSink {
             }
             tSink.setNum_replicas(numReplicas);
             tSink.setNeed_gen_rollup(dstTable.shouldLoadToNewRollup());
-            tSink.setSchema(createSchema(tSink.getDb_id(), dstTable, tupleDescriptor));
+            tSink.setSchema(createSchema(tSink.getDb_id(), dstTable, tupleDescriptor, targetWriteIndexId, true));
 
             TransactionState txnState = getTransactionState(tSink, dstTable.isOlapExternalTable());
 
             TOlapTablePartitionParam partitionParam = createPartition(tSink.getDb_id(), dstTable, tupleDescriptor,
-                    enableAutomaticPartition, automaticBucketSize, getOpenPartitions(), txnState);
+                    enableAutomaticPartition, automaticBucketSize, getOpenPartitions(), txnState, targetWriteIndexId);
             tSink.setPartition(partitionParam);
             tSink.setLocation(createLocation(dstTable, partitionParam, enableReplicatedStorage, computeResource, txnState));
             tSink.setNodes_info(GlobalStateMgr.getCurrentState().createNodesInfo(computeResource, getSystemInfoService(dstTable)));
@@ -440,7 +447,7 @@ public class OlapTableSink extends DataSink {
                     tSink2.unsetPartition();
                     tSink2.unsetLocation();
                     TOlapTablePartitionParam partitionParam2 = createPartition(tSink2.getDb_id(), dstTable, tupleDescriptor,
-                            false, automaticBucketSize, doubleWritePartitionIds, txnState);
+                            false, automaticBucketSize, doubleWritePartitionIds, txnState, targetWriteIndexId);
                     tSink2.setPartition(partitionParam2);
                     tSink2.setLocation(createLocation(dstTable, partitionParam2, enableReplicatedStorage, computeResource, txnState));
                     tSink2.setIgnore_out_of_partition(true);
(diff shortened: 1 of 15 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
## Real-cluster E2E validation (shared-data)

Built this PR on our internal test platform and validated on a fresh **shared-data** cluster (1 FE + 3 CN, this PR's build).

**What's validated:** the per-index `distributed_exprs` routing path — for each materialized index of a range-distribution table, rows must be routed to the correct range tablet by that index's key. Oracle per case: load a deterministic key pattern → split into K range tablets → **write new rows that span the tablet boundaries** → assert (a) data conservation (COUNT/SUM vs analytic), (b) every row sits inside the tablet whose range covers it — 0 misplaced via `TABLET()` hint, (c) per-tablet counts are exact and sum to the total.

| Case | Key type | tablets | Checks | Result |
|---|---|---|---|---|
| DUPLICATE | INT (k1,k2) | 2 | conservation + post-split write routing (0 misplaced) | ✅ |
| PRIMARY KEY | INT (k1,k2) | 3 | dedup + conservation + routing (0/0/0) + dedup update | ✅ |
| DUPLICATE | VARCHAR(20) leading key | 5 | byte-ordered routing (0×5) + conservation | ✅ |
| AGGREGATE | INT (k1,k2), v1 SUM | 1 | aggregate-on-write correctness | ✅ |

Each "post-split write" exercises the new sink per-index routing into multiple range tablets. Example (PK): table split at boundaries `(39322,7)` / `(78644,3)` into 3 tablets, then 100k new rows spanning all three were written → **0 rows misplaced**, per-tablet counts exact, total conserved at 1.1M rows; a re-write of existing keys confirmed PK upsert. VARCHAR split into 5 tablets at zero-padded boundaries (`026215`,`052429`,`078644`,`098305`) confirmed byte-ordered routing with 0 misplaced.

Range-distribution routing is behavior-preserving for base-only tables; these results confirm loads / queries / splits stay correct with the new per-index routing path active (no regression).

**Guard sanity** (existing range sort-key guards still hold on this build):
- `MODIFY COLUMN k1 BIGINT` on a range sort-key → **rejected** ✓
- `MODIFY COLUMN k1 VARCHAR(40)` (widen) → **allowed** ✓
- `ADD ROLLUP` on a range table → **rejected** ✓

Unit tests: FE `OlapTableSinkTest` (26) and BE `RangeRouterTest` + `TabletSinkSenderRange*` + `OlapTablePartitionParamTest` (incl. per-index routing, K=1 single-tablet, key-type validation) all pass.

kevincai:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

fe/fe-core/src/main/java/com/starrocks/planner/OlapTableSink.java
```diff
@@ -282,6 +285,10 @@ public void setIsStreamingLoad(boolean isStreamingLoad) {
         this.isStreamingLoad = isStreamingLoad;
     }
 
+    public void setTargetWriteIndexId(Long targetWriteIndexId) {
```
srlch, line 288: It seems this function has not been called by any code path in this pr? targetWriteIndexId is always null so far?

xiangguangyxg: Good catch — you're right, there's no production caller in this PR yet, so `targetWriteIndexId` is always null today.

It's a deliberate prerequisite for the follow-up range-rewrite job (the K-tablet shadow-index rewrite that enables key-column schema change / rollup on range-distribution tables). During that job's double-write window the rewrite `INSERT` must write **only** the new-key shadow index and must not re-write the base index — that's exactly what this "write a single target index" filter provides (it keeps the schema/partition/loaded-index lists 1:1 by meta id so BE accepts the load). It's not dead in the test sense — `OlapTableSinkTest` exercises it directly (`testTargetWriteIndexFilterRestrictsToOneIndex` / `...UnsetWritesAllIndexes` / `...ThrowsWhenIndexAbsent`), and unset (the only path today) is a strict no-op.

I kept it in this PR so the rewrite-job PR doesn't have to re-touch `OlapTableSink` routing again, but I'm happy to go either way:
- **Keep**, and I'll add a comment on the setter/field marking it as a prerequisite for the range-rewrite job (so the "no caller yet" is explicit), or
- **Drop it here** and re-introduce it together with its caller in the rewrite-job PR.

Which do you prefer? Happy to remove it now if you'd rather this PR stay strictly to the exercised per-index `distributed_exprs` routing.

srlch: I think keep it is fine.

srlch (approved):

luohaha (approved):

PR #74758, opened 2026-06-13
[BugFix] Reject non-sortable (JSON) columns as rollup key columns

### LINKED ISSUE #74757: [BE crash] SIGSEGV in SegmentWriter::append_chunk sort-key build when ADD ROLLUP includes a JSON column
### Summary

Adding a ROLLUP that includes a JSON column to a DUPLICATE KEY table makes the FE
promote the JSON column to a rollup **sort key**, which the BE cannot handle:
`SchemaChangeWithSorting` crashes with SIGSEGV while building the sort-key tuple.

Worse, the alter job is persisted in the FE edit log, so **every BE restart
re-dispatches the same task and crashes the BE again** — a permanent crash loop
until the job is manually cancelled.

The root cause is a validation gap in the FE rollup key derivation: base-table
CREATE already rejects JSON key columns via `Type.canDistributedBy()`, but the
rollup path uses a narrower ad-hoc check that misses JSON. A fix PR is on the way.

### Environment

| Item | Value |
|---|---|
| Deployment mode | shared-nothing (1 FE + 1 BE, and separately 1 FE + 2 BE; `replication_num=1`) |
| Build environment | `starrocks/dev-env-ubuntu:latest` container (Ubuntu 22.04, GCC toolchain) |
| Architecture | aarch64 (crash is a null-pointer deref, not arch-specific) |
| Build type | Release, source build via `./build.sh` |

```sql
CREATE TABLE t_rollup_test (
    id bigint,
    j json,
    sub_str varchar(20)
) DUPLICATE KEY(id)
DISTRIBUTED BY HASH(id) BUCKETS 2
PROPERTIES ('replication_num'='1');

INSERT INTO t_rollup_test
SELECT n, parse_json(concat('{"k1":"v', n, '","k2":"x"}')), 's'
FROM TABLE(generate_series(1, 100)) AS t(n);

-- control: rollup WITHOUT the JSON column → FINISHED normally
ALTER TABLE t_rollup_test ADD ROLLUP r_no_json (id, sub_str);
-- (wait until SHOW ALTER TABLE ROLLUP reports FINISHED before the next ALTER)

-- trigger: rollup INCLUDING the JSON column → BE SIGSEGV
ALTER TABLE t_rollup_test ADD ROLLUP r_idx (id, j);
```

All three rollup entry paths reproduce the same crash (all funnel into the same
key-derivation code in `MaterializedViewHandler.checkAndPrepareMaterializedView`):

1. auto key derivation — `ADD ROLLUP r (id, j)` (above)
2. explicit dup key — `ADD ROLLUP r (id, j) DUPLICATE KEY(id, j)`
3. inline rollup — `CREATE TABLE ... ROLLUP (r (id, j) DUPLICATE KEY(id, j))`

Notes:
- `flat_json` properties are **not** required; a plain JSON column is enough
  (verified on a table with no `flat_json.*` properties at all).
- AGG/UNIQUE tables are not affected (their rollups copy key flags from the
  base table, which can never have a JSON key) — verified on both models.
- A rollup whose **first** column is JSON fails cleanly with
  `Data type of first column cannot be JSON` (pre-existing check); the crash
  requires at least one sortable column before the JSON column.

Either of:
- the JSON column is added as a **value** column (consistent with base-table
  auto key derivation, which stops key promotion at non-sortable types), or
- the statement is **rejected** with the same error a base table gives:
  `Invalid data type of key column 'j': 'JSON'`.

In no case should the BE crash.

### Real behavior (Required)

BE dies with SIGSEGV; the rollup job stays `RUNNING`/`PENDING`, and the BE
enters a crash loop on every restart while the job exists.

```
PC: SegmentWriter::append_chunk(Chunk const&)
SIGSEGV (@0x20)                       ← null pointer + 0x20 offset
  SegmentWriter::append_chunk
  HorizontalRowsetWriter::_flush_chunk(Chunk const&, SegmentPB*)
  HorizontalRowsetWriter::flush_chunk(...)
  MemTableRowsetWriterSink::flush_chunk(...)
  MemTable::flush(...)
  SchemaChangeWithSorting::process(...)
  SchemaChangeHandler::_convert_historical_rowsets(SchemaChangeParams&)
  SchemaChangeHandler::_do_process_alter_tablet_normal(...)
  SchemaChangeHandler::_do_process_alter_tablet(...)
```

FE side after the crash: `SHOW BACKENDS` reports the BE as
`Alive=false ... java.net.ConnectException: Connection refused`.

All builds are source builds, so `select current_version()` reports a
branch/commit string rather than a release number. Verified matrix:

| Build (commit) | How verified | Result |
|---|---|---|
| pure upstream `c8ca6e38936` (FE + BE both from this commit, zero local changes) | cluster repro | **crash reproduced** |
| FE `main` @ `1e1dbb91dc` (2026-06-11) | cluster repro (before/after fix comparison) | **crash reproduced** without fix; FINISHED with fix |
| `main` tip @ `d70721db3d7` | code inspection | bug **still present** — `MaterializedViewHandler.java` guard is still `isFloatingPointType() \|\| isComplexType()`; no BE-side guard in `chunk_variant_helper.cpp` / `segment_writer.cpp` either |
| 4.1.0-based build (with unrelated local patches) | cluster observation | same crash signature first observed here; root-caused on pure upstream above |

No matching fix or issue found on `main` or in open PRs/issues as of filing
(closest related work: #13305 rejected unsupported column types for sync MVs,
but the rollup DUP-key derivation gap remained).

### Root cause

Verified with an instrumented build: every column appends successfully, then the
crash happens in the sort-key build step right after the append loop. The
instrumentation shows the rollup chunk carries `sort_detail=[id BIGINT, j JSON]`
— i.e. **the JSON column was made a sort/short-key column**, and
`build_variant_tuple_from_chunk_row` cannot build a sort-key datum from a JSON
value (`be/src/storage/chunk_variant_helper.cpp`):

```cpp
for (uint32_t column_index : column_indexes) {        // includes the JSON column
    tuple.emplace(schema->field(column_index)->type(),
                  chunk.get_column_by_index(column_index)->get(row_idx));
}
```

On the FE side, `MaterializedViewHandler.checkAndPrepareMaterializedView`
(DUP_KEYS branch, line ~595 on current main) derives rollup keys with an
ad-hoc guard:

```java
if (column.getType().isFloatingPointType() || column.getType().isComplexType()) {
    break;   // stops key promotion — but JSON passes through and becomes a key
}
```

`isComplexType()` does not include JSON, so the JSON column is promoted to a
key. The explicit `DUPLICATE KEY(...)` path (line ~658) has **no key type
validation at all**. Meanwhile base-table CREATE rejects JSON keys through
`Type.canDistributedBy()` (`ColumnDefAnalyzer`/`CreateTableAnalyzer`), and sync
MVs go through `canBeMVKey()` which delegates to the same predicate — so the
rollup path has simply diverged from the canonical key-type predicate.

Boundary check on the predicate (measured on a live cluster, with the fix
applied): every type rejected by `canDistributedBy()` (JSON / ARRAY / MAP /
STRUCT / DOUBLE) is refused as a rollup key, and every type it allows
(VARBINARY / VARCHAR / INT) passes the BE sort-key build safely — the predicate
boundary aligns with the crash boundary (the crash itself verified with JSON).

### Operational impact

Because the alter task is persisted, a single bad `ADD ROLLUP` can take a
cluster down repeatedly: BE restarts → FE re-dispatches the alter task → BE
crashes at the same spot (observed as BE flapping alive↔down until the job was
removed). Recovery requires `CANCEL ALTER TABLE ROLLUP FROM <table>` before the
BE can stay up.

### Workaround

- Do not include JSON columns in rollup column lists.
- If already hit: `CANCEL ALTER TABLE ROLLUP FROM <table>;`, then restart the BE.

### Additional finding (separate functional gap, same code path)

While investigating, we also found `_convert_historical_rowsets`
(`be/src/storage/schema_change.cpp`) never sets
`writer_context.flat_json_config`, unlike the load (`delta_writer.cpp`),
compaction (`compaction_utils.cpp`), and `tablet_manager.cpp` write paths. So
segments produced by schema-change/rollup silently ignore the table's
`flat_json.*` settings and fall back to global defaults. This is **not** the
cause of the crash (verified: setting it does not prevent the SIGSEGV), but it
is an inconsistency worth fixing alongside. Happy to split this into a separate
issue if preferred.

### Fix direction (PR incoming)

Unify both rollup key paths on the canonical predicate `Type.canDistributedBy()`:

- auto path: stop key promotion at any non-sortable type → column becomes a
  value column (mirrors base-table key derivation);
- explicit path: reject with the same error message as base-table validation
  (`Invalid data type of key column 'j': 'JSON'`).

Verified locally: before fix — SIGSEGV on `ADD ROLLUP (id, j)`; after fix —
auto path FINISHED with `j` as a value column (query results correct), explicit
path returns the error, BE stays alive. Coverage: DUP/AGG/UNIQUE models, inline
rollup, explicit dup-key, two JSON columns, JSON-first edge case, 20k rows ×
8 buckets, drop-and-re-add; 110+ existing FE UTs pass, checkstyle clean.

krokayman:
DUP_KEYS rollup key derivation in MaterializedViewHandler used an ad-hoc
guard (isFloatingPointType || isComplexType) that missed JSON, metric,
variant, and function types. When a JSON column became a rollup sort-key,
BE SchemaChangeWithSorting::process crashed with SIGSEGV @0x20 in
build_variant_tuple_from_chunk_row, which cannot construct a sort-key
datum from a JSON value.

Fix two paths in checkAndPrepareMaterializedView:
- Path 1 (auto key derivation): replace the ad-hoc guard with !canDistributedBy(), mirroring CreateTableAnalyzer key derivation. Non-sortable columns are silently demoted to value columns.
- Path 2 (explicit DUPLICATE KEY): reject non-sortable key columns with the same error message as base-table validation (ColumnDefAnalyzer), so rollup can never produce a key schema the base table itself rejects.

Verified: before fix — SIGSEGV on ADD ROLLUP r(id,j); after fix — FINISHED, BE alive. Both DUP/AGG/UNIQUE table models, inline ROLLUP, explicit DUPLICATE KEY, large datasets (20k rows × 8 buckets), and JSON-as-first-column edge cases tested. 110+ FE unit tests pass, checkstyle clean.

Fixes #74757

## Why I'm doing:

JSON columns cannot be sort/key columns in StarRocks, but the DUP_KEYS
rollup key derivation path silently promoted JSON to a sort key, causing
BE SIGSEGV. The failed alter task is also persisted in the edit log, so
every BE restart re-dispatches the same task and crashes the BE again —
a permanent crash loop until the job is manually cancelled. Tracked in #74757.

## What I'm doing:

Fixes #74757

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/MaterializedViewHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/MaterializedViewHandler.java
@@ -592,7 +592,9 @@ public class MaterializedViewHandler extends AlterHandler {
                         }
                         break;
                     }
-                    if (column.getType().isFloatingPointType() || column.getType().isComplexType()) {
+                    if (!column.getType().canDistributedBy()) {
+                        // JSON and other non-sortable types must not become rollup sort-key
+                        // columns (mirrors base-table key derivation in CreateTableAnalyzer).
                         break;
                     }
                     if (column.getType().isVarchar()) {
@@ -656,6 +658,12 @@ public class MaterializedViewHandler extends AlterHandler {
 
                     Column oneColumn = new Column(baseColumn);
                     if (isKey) {
+                        if (!baseColumn.getType().canDistributedBy()) {
+                            // JSON and other non-sortable types cannot be rollup key columns
+                            // (mirrors base-table key validation in ColumnDefAnalyzer).
+                            throw new DdlException(String.format(
+                                    "Invalid data type of key column '%s': '%s'", rollupColName, baseColumn.getType()));
+                        }
                         hasKey = true;
                         oneColumn.setIsKey(true);
                         oneColumn.setAggregationType(null, false);
--- a/fe/fe-core/src/test/java/com/starrocks/alter/MaterializedViewHandlerTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/alter/MaterializedViewHandlerTest.java
@@ -42,11 +42,13 @@ import com.starrocks.catalog.MaterializedIndex;
 import com.starrocks.catalog.OlapTable;
 import com.starrocks.catalog.PhysicalPartition;
 import com.starrocks.common.jmockit.Deencapsulation;
+import com.starrocks.sql.ast.AddRollupClause;
 import com.starrocks.sql.ast.AggregateType;
 import com.starrocks.sql.ast.CreateSyncMVStmt;
 import com.starrocks.sql.ast.KeysType;
 import com.starrocks.sql.ast.MVColumnItem;
 import com.starrocks.type.IntegerType;
+import com.starrocks.type.JsonType;
 import com.starrocks.type.VarcharType;
 import mockit.Expectations;
 import mockit.Injectable;
@@ -282,6 +284,72 @@ public class MaterializedViewHandlerTest {
             Assertions.assertEquals(null, newMVColumn.getAggregationType());
             Assertions.assertEquals(false, newMVColumn.isAggregationTypeImplicit());
             Assertions.assertTrue(newMVColumn.getType().isVarchar());
+        } catch (Exception e) {
+            Assertions.fail(e.getMessage());
+        }
+    }
+
+    @Test
+    public void testRollupRejectsJsonKeyColumn(@Injectable OlapTable olapTable, @Injectable Database db) {
+        // Explicit DUPLICATE KEY listing a JSON column must be rejected, mirroring
+        // base-table key validation. Otherwise the JSON column becomes a rollup sort
+        // key and the BE crashes while building the sort-key tuple.
+        final long baseIndexMetaId = 1L;
+        final String rollupName = "r_json_key";
+        Column idColumn = new Column("id", IntegerType.BIGINT, true, AggregateType.NONE, "", "");
+        Column jsonColumn = new Column("j", JsonType.JSON, false, AggregateType.NONE, "", "");
+        AddRollupClause addRollupClause = new AddRollupClause(rollupName,
+                Lists.newArrayList("id", "j"), Lists.newArrayList("id", "j"), null, null);
+        new Expectations() {
+            {
+                olapTable.hasMaterializedIndex(rollupName);
+                result = false;
+                olapTable.getKeysType();
+                result = KeysType.DUP_KEYS;
+                olapTable.getSchemaByIndexMetaId(baseIndexMetaId);
+                result = Lists.newArrayList(idColumn, jsonColumn);
+            }
+        };
+        MaterializedViewHandler materializedViewHandler = new MaterializedViewHandler();
+        try {
+            Deencapsulation.invoke(materializedViewHandler, "checkAndPrepareMaterializedView",
+                    addRollupClause, olapTable, baseIndexMetaId);
+            Assertions.fail("expected DdlException for JSON key column");
+        } catch (Exception e) {
+            Assertions.assertTrue(e.getMessage().contains("Invalid data type of key column 'j'"),
+                    e.getMessage());
+        }
+    }
+
+    @Test
+    public void testRollupDemotesJsonToValueColumn(@Injectable OlapTable olapTable, @Injectable Database db) {
+        // Auto key derivation (no explicit DUPLICATE KEY) must stop promoting keys at a
+        // non-sortable type, so the JSON column is added as a value column rather than a key.
+        final long baseIndexMetaId = 1L;
+        final String rollupName = "r_json_value";
+        Column idColumn = new Column("id", IntegerType.BIGINT, true, AggregateType.NONE, "", "");
+        Column jsonColumn = new Column("j", JsonType.JSON, false, AggregateType.NONE, "", "");
+        AddRollupClause addRollupClause = new AddRollupClause(rollupName,
+                Lists.newArrayList("id", "j"), null, null, null);
+        new Expectations() {
+            {
+                olapTable.hasMaterializedIndex(rollupName);
+                result = false;
+                olapTable.getKeysType();
+                result = KeysType.DUP_KEYS;
+                olapTable.getSchemaByIndexMetaId(baseIndexMetaId);
+                result = Lists.newArrayList(idColumn, jsonColumn);
+            }
+        };
+        MaterializedViewHandler materializedViewHandler = new MaterializedViewHandler();
+        try {
+            List<Column> rollupSchema = Deencapsulation.invoke(materializedViewHandler,
+                    "checkAndPrepareMaterializedView", addRollupClause, olapTable, baseIndexMetaId);
+            Assertions.assertEquals(2, rollupSchema.size());
+            Column idResult = rollupSchema.stream().filter(c -> c.getName().equals("id")).findFirst().orElseThrow();
+            Column jsonResult = rollupSchema.stream().filter(c -> c.getName().equals("j")).findFirst().orElseThrow();
+            Assertions.assertTrue(idResult.isKey());
+            Assertions.assertFalse(jsonResult.isKey(), "JSON column must not be a rollup key");
         } catch (Exception e) {
             Assertions.fail(e.getMessage());
         }


CelerData-Reviewer:
@codex review

krokayman:
Thanks for running CI. Flagging the one red check so it's easy to assess.

Failure: LakeReplicationJobTest.testCrashRecoveryWithS3PartitionedPrefix — publish version not finished (FE UT: 19838 / 19839 passed, 86 skipped, 1 failed).

Why it appears unrelated to this PR. This change is FE-only and narrow (+9 / −1 in a single file): it tightens rollup key-column validation in MaterializedViewHandler.checkAndPrepareMaterializedView so a JSON (non-canDistributedBy) column can no longer become a rollup sort key. It does not touch — directly or transitively — the replication path, lake/shared-data storage, S3 prefixing, or publish-version logic that the failing test exercises. The two areas share no call path, so a regression from this diff reaching that test would be surprising.

On the failure itself. publish version not finished is an async-timing symptom (the test waits on a publish that hadn't completed in the allotted window) rather than an assertion on behavior this PR changes. I couldn't find a prior issue tracking this specific test, so I don't want to assert it's a known flake — only that it reads like a timing-sensitive failure in a subsystem this change doesn't touch.

Locally, the rollup / alter / lake-rollup suites relevant to this change pass (MaterializedViewHandlerTest, RollupJobV2Test, AddRollupClauseTest, LakeRollupJobTest, alter.AlterTest, lake.AlterTest), and checkstyle is clean.

Could you re-trigger CI when convenient? If it reproduces on the re-run, I'm glad to dig into the interaction further.

CelerData-Reviewer:
@codex review

krokayman:
Following up after the latest push.

The coverage gate is now green. The earlier FE Incremental Coverage failure is resolved — MaterializedViewHandler.java is at 4/4 (100%), with both new key-validation lines now exercised by the two unit tests added in this push. FE UT and FE Code Style Check also pass.

One unrelated check is still red. SQL-Tester reports 1313 tests run, 1312 passed, 1 failed, and the job page shows Annotations: 1 error — a single failing case, test_window_functions_with_complex_types. The two red entries (SQL-Tester and SQL-Tester Report) are the run job and its reporting job for that same one failure, not two separate failures.


AssertionError: Element counts are not equal:
  First has 5, Second has 6
  First has 6, Second has 5
Why it is unrelated to this change

This PR is FE-only and narrow — 2 files, +77/−1. It makes MaterializedViewHandler.checkAndPrepareMaterializedView reject a JSON (non-canDistributedBy) column from being promoted to a rollup key, and adds the matching unit tests. It changes nothing in query execution, window-function evaluation, or complex-type (map/struct/array/json) handling, and adds no files under test/sql/.

The failing case (test/sql/test_window_function/T/test_window_functions_with_complex_types) runs FIRST_VALUE / LAST_VALUE / LEAD / LAG over complex types and asserts the result set. One detail worth calling out directly: its table is a DUPLICATE KEY table, and this PR's change lives in the DUP_KEYS branch of rollup key derivation — but the two never meet. The modified method only runs during rollup creation (checkAndPrepareMaterializedView, reached via processBatchAddRollup), and this test issues only CREATE TABLE + SELECT ... OVER (...) — there is no ROLLUP or ADD ROLLUP anywhere in it (grep -c rollup → 0). So the modified code is never on the call stack for this test, dup-key table or not.

On the failure itself

The assertion is assert_count_equal, which compares multiset element counts independently of row order — so this is not an output-ordering diff. The same row appears 5 times on one side and 6 on the other (and symmetrically 6 vs 5 for another), i.e. one value is over-counted while a neighbor is under-counted by the same amount. That is the shape you get when a window function with a non-unique ORDER BY resolves a tie to a different row between runs, shifting which rows carry a given value. I could not find a prior issue tracking this specific case, so I won't claim it is a known flake — only that both the failure shape and its location point away from this change.

What passes

All FE checks relevant to this PR are green: FE UT, FE Code Style Check, and FE Incremental Coverage (100%). Locally, MaterializedViewHandlerTest runs 12 / 0 failed (10 existing + the 2 added here), and the broader rollup/alter suites pass.

Could you re-trigger the SQL-Tester stage when convenient? If it reproduces on the re-run, I'm glad to investigate the interaction further.

stdpain:
https://github.com/Mergifyio update main

krokayman:
Hi @HangyuanLiu, @gengjun-git 👋 — I hope you're both having a good week! Please forgive the gentle nudge, and there's genuinely no rush on this.

Whenever you happen to find a quiet moment, I'd be truly grateful if you could take a look. For a little context: this fixes a BE SIGSEGV that occurs when a JSON column is used as a rollup key column (tracked in #74757). It's a small, FE-only change (+77 / −1) — all CI checks are green and FE incremental coverage is at 100%.

I completely understand how busy review queues can get, so please take all the time you need. I'd be more than happy to walk through anything, address feedback, or make any changes you'd like. Thank you so much for your time and for everything you do for the project. 🙏

HangyuanLiu (approved):

stdpain (approved):

PR #74785, opened 2026-06-15
[BugFix] Stop vector index rewrite from polluting the shared table schema

sevev:
## Why I'm doing:

A vector ANN query (`ORDER BY approx_*_distance(v, [...]) LIMIT k`) can intermittently break unrelated statements on the same table with:

```
java.lang.IllegalArgumentException: Multiple entries with same key:
  `__vector_approx_cosine_similarity` float NOT NULL COMMENT ""=N: __vector_approx_cosine_similarity and
  `__vector_approx_cosine_similarity` float NOT NULL COMMENT ""=M: __vector_approx_cosine_similarity
```

Root cause: `RewriteToVectorPlanRule.rewriteOptByDistanceColumn` registers the synthetic distance column by calling `scanOp.getTable().addColumn(...)`. On the whole-phase-lock planning path the scan references the shared catalog `OlapTable`, and `Table.addColumn` appends to `fullSchema` (a `List` with no dedup) while only `nameToColumn` dedups. So every vector query that plans on the live table appends another duplicate `__vector_*` column. Once two duplicates accumulate, building the `Column`-keyed `ImmutableMap` in `RelationTransformer.visitTable` throws `Multiple entries with same key` for any later statement on that table (SELECT/DELETE/INSERT). It is intermittent because analyze-phase column pruning sometimes drops the synthetic columns. The default lock-free path was unaffected because it plans on a per-query `copyOnlyForQuery` snapshot whose `fullSchema` is a private list.

## What I'm doing:

The distance column is per-query synthetic state; it only needs to live in the scan operator's `colRefToColumnMetaMap`/`columnMetaToColRefMap` (`PlanFragmentBuilder` builds the scan slot's `Column` from `colRefToColumnMetaMap`, never from the table schema). The fix creates the column locally and stops adding it to the shared table schema entirely, so no planning path mutates the catalog `OlapTable`. Only `RewriteToVectorPlanRule` changes; a regression test in `VectorIndexTest` forces the lock path and verifies that repeated vector queries leave the shared schema clean and keep planning successfully.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteToVectorPlanRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/RewriteToVectorPlanRule.java
@@ -169,10 +169,15 @@ public class RewriteToVectorPlanRule extends TransformationRule {
                                                      ScalarOperator newPredicate,
                                                      VectorFuncInfo info,
                                                      VectorSearchOptions opts) {
-        // Add index distanceColumn to the scan operator, including table, colRefToColumnMetaMap, and columnMetaToColRefMap.
+        // The distance column is per-query synthetic state: it only needs to live in the scan operator's
+        // colRef maps below (PlanFragmentBuilder builds the scan slot's Column from colRefToColumnMetaMap).
+        // It must NOT be added to the shared catalog table's fullSchema. Table.addColumn appends to
+        // fullSchema (a List that does not dedup) while only nameToColumn dedups, so adding it here
+        // appended a duplicate "__vector_*" column on every vector query that planned on the live table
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

kangkaisen (approved):

stephen-shelby (approved):

sevev:
@mergify backport branch-4.1

PR #74794, opened 2026-06-15
[BugFix] Support table-level override for first load statistics collection

stephen-shelby:
## Why I'm doing:

`enable_statistic_collect_on_first_load` currently behaves like an AND between the table property and the global config. This makes it impossible to disable first-load statistics collection globally while enabling it for selected tables.

Also, the old create-table flow always persisted `enable_statistic_collect_on_first_load=true` into table properties by default, even when users did not explicitly set this property. So we cannot simply use the existence of this key to determine whether the table property was explicitly configured.

The expected behavior is:
- If the table property is explicitly set, use the table-level value.
- If the table property is not explicitly set, fall back to the global config.

## What I'm doing:

Change the first-load statistics collection decision from:

```text
table_value && global_value
```

to:

```text
table property explicitly set ? table_value : global_value
```

Behavior before this change:

```text
table default true + global true  => collect
table default true + global false => skip
table false        + global true  => skip
table false        + global false => skip
table true         + global false => skip
```

Behavior after this change:

```text
table unset        + global true  => collect
table unset        + global false => skip
table false        + global true  => skip
table false        + global false => skip
table true         + global true  => collect
table true         + global false => collect
```

Compatibility handling:
- Add an explicit-set marker in `TableProperty` for new metadata.
- Only persist `enable_statistic_collect_on_first_load` during create table when users explicitly specify it.
- For old metadata:
  - Old `enable_statistic_collect_on_first_load=true` is treated as unset, because it may have been written by the old default create-table flow.
  - Old `enable_statistic_collect_on_first_load=false` is treated as explicitly set, because old default creation would not write `false`.
- This preserves existing behavior for most upgraded tables: if users never configured the table property, the table still follows the global config.

Other changes:
- Remove the global config short-circuit in `LoadJobStatsListener`, so the final decision can be made per table in `StatisticsCollectionTrigger`.
- Add FE unit tests for runtime trigger behavior and metadata compatibility.
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/OlapTable.java
@@ -2214,10 +2214,15 @@ public class OlapTable extends Table {
         return tableProperty.enableStatisticCollectOnFirstLoad();
     }
 
+    public boolean isSetEnableStatisticCollectOnFirstLoad() {
+        return tableProperty != null && tableProperty.isSetEnableStatisticCollectOnFirstLoad();
+    }
+
     public void setEnableStatisticCollectOnFirstLoad(boolean enable) {
         tableProperty.modifyTableProperties(PropertyAnalyzer.PROPERTIES_ENABLE_STATISTIC_COLLECT_ON_FIRST_LOAD,
                 Boolean.valueOf(enable).toString());
         tableProperty.buildEnableStatisticCollectOnFirstLoad();
+        tableProperty.setEnableStatisticCollectOnFirstLoad(enable);
     }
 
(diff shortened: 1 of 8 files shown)


CelerData-Reviewer:
@codex review

Youngwb (approved):

HangyuanLiu (approved):

PR #74802, opened 2026-06-15
[BugFix] Stop double-counting vector index cache memory on the process tracker

sevev:
## Why I'm doing:

The HNSW/tenann vector index lives in the normal heap, so the global allocator hook charges its bytes to the process mem tracker once while the index loads. `VectorIndexCache` then accounted the same bytes a second time: `DynamicCache::update_object_size` did a plain `consume()` on the `vector_index` tracker, which is a child of `process` and therefore propagated the bytes up to `process` again.

A cached index thus appeared twice in the process counter (real heap once + cache label once), inflating it by a full copy of every cached index. On a node with large indexes this pushed the process counter past `mem_limit` and caused spurious `MEM_LIMIT_EXCEEDED` for every query, even though real RSS was far below the limit (e.g. observed `process_mem_bytes` 26.77GB = jemalloc 15.27GB + vector_index 11.43GB, with RSS only 14.7GB).

## What I'm doing:

Fix the accounting without disturbing the self-balancing allocator-hook pair (its alloc `+` / free `-` already cancel on `process`). `DynamicCache` gains `set_mem_tracker_excluding_root()`, which routes the cache's accounting through `consume_without_root`/`release_without_root` so the cache tracker labels usage without re-adding it to the root (process) tracker — the same idiom `segment_replicate_executor` already uses for brpc send buffers. The existing `set_mem_tracker()` (additive) is unchanged, so other `DynamicCache` users keep their current behavior. `VectorIndexCache` opts into the new variant, and `TenANNReader` charges index-load allocations to the process tracker instead of the vector_index tracker (the hook already counts these heap bytes on process once; pointing it at vector_index would double the vector_index label and leak it on eviction). After the fix the process counter reflects the index exactly once and stays balanced across insert/evict cycles, while the `vector_index` tracker still reports the logical index size.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/index/vector/tenann_index_reader.cpp
+++ b/be/src/storage/index/vector/tenann_index_reader.cpp
@@ -83,9 +83,20 @@ Status TenANNReader::init_searcher(const tenann::IndexMeta& meta, const std::str
     auto meta_copy = meta;
     apply_index_reader_cache_options(&meta_copy);
 
-    // Loader runs under vector_index mem tracker and opens the remote file
-    // lazily, so warm-path cache hits skip the OSS/S3 round-trip entirely.
-    auto* tracker = GlobalEnv::GetInstance()->vector_index_mem_tracker();
+    // Loader opens the remote file lazily, so warm-path cache hits skip the OSS/S3
+    // round-trip entirely.
+    //
+    // Charge the load to the process tracker (not the vector_index tracker): the
+    // index lives in the heap, so the allocator hook already accounts these bytes
+    // to process exactly once here. The VectorIndexCache then labels the same bytes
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

kevincai (approved):

PR #74811, opened 2026-06-15
[BugFix] Honor page cache in global late-materialization lookup

sevev:
## Why I'm doing:

The global late-materialization (GLM) fetch path builds its `RowsetReadOptions` without setting `use_page_cache`, so it falls back to the default `false` (`be/src/storage/rowset/rowset_options.h`) and bypasses the `StoragePageCache`. As a result, every column page touched by the per-row-locator lookup is re-read and re-decompressed instead of being served from cache. The normal scan path does set this (see `be/src/connector/lake_connector.cpp`), so the lookup path is inconsistent.

This shows up as wasted CN CPU and lower QPS. On a shared-data ANN top-K workload (`SELECT id ... ORDER BY approx_*_distance(v, q) LIMIT k`), where the LIMIT is already pushed into the index so late materialization yields no row reduction, the extra uncached lookup is pure overhead and costs ~28% QPS. A perf inverted call-graph attributed ~22% of CN CPU to `LZ4_decompress_safe` + `bshuf_*` reached exclusively via `NativeLookUpTask::_late_materialize_by_row_locators` -> `SegmentIterator` -> `read_and_decompress_page`. Results are correct; the impact is purely performance.

## What I'm doing:

Set `use_page_cache` on the `RowsetReadOptions` built in both `OlapScanTabletAdaptor::get_iterator` and `LakeScanTabletAdaptor::get_iterator` (`be/src/exec/pipeline/lookup/tablet_adaptor.cpp`), gated on `config::disable_storage_page_cache` exactly as the normal scan path is. The lake path additionally sets `lake_io_opts.fill_data_cache = true` to mirror the scan behavior. This benefits all queries that go through GLM late materialization, not only vector search.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/pipeline/lookup/tablet_adaptor.cpp
+++ b/be/src/exec/pipeline/lookup/tablet_adaptor.cpp
@@ -20,7 +20,6 @@
 #include "base/status.h"
 #include "base/status_fmt.hpp"
 #include "base/statusor.h"
-#include "common/config.h"
 #include "compute_env/global_dict/fragment_dict_state.h"
 #include "exec/olap_scan_node.h"
 #include "exec/pipeline/fragment_context.h"
@@ -132,6 +131,7 @@ private:
     std::vector<ColumnAccessPathPtr> _column_access_paths;
     ColumnIdToGlobalDictMap* _global_dicts = nullptr;
     OlapScanLazyMaterializationContext* _glm_ctx = nullptr;
+    bool _use_page_cache = false;
 };
 
 class LakeScanTabletAdaptor final : public LookUpTabletAdaptor {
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

be/src/exec/pipeline/lookup/tablet_adaptor.cpp
```diff
@@ -20,6 +20,7 @@
 #include "base/status.h"
 #include "base/status_fmt.hpp"
 #include "base/statusor.h"
+#include "common/config.h"
```
stdpain: instead with
```
config_storage_fwd.h
```

sevev: > instead with
> 
> ```
> config_storage_fwd.h
> ```

updated

luohaha (approved):

stdpain (approved):

sevev:
ignore backport check: 4.1

PR #70360, opened 2026-03-17
[BugFix] Mask user auth strings in audit and SQL redaction

gengjun-git:
## Why I'm doing:
The current credential masking is incomplete for user-related statements in audit/profile SQL output.

`CREATE USER`, `ALTER USER`, and `SET PASSWORD` can still expose auth strings in some paths. In particular, when an auth plugin is specified, `AST2StringVisitor` currently skips masking by default, which may leak native password credentials. At the same time, auth strings for `AUTHENTICATION_LDAP_SIMPLE` are not passwords and should not be masked.

## What I'm doing:
- Extend `SqlCredentialRedactor` to mask:
  - `IDENTIFIED BY ...`
  - `IDENTIFIED BY PASSWORD ...`
  - `IDENTIFIED WITH MYSQL_NATIVE_PASSWORD BY/AS ...`
  - `SET PASSWORD ... = PASSWORD(...)`
- Keep `AUTHENTICATION_LDAP_SIMPLE BY/AS ...` unchanged because its auth string is not a password.
- Align `AST2StringVisitor` with the same masking rules so AST-based SQL output does not leak credentials when an auth plugin is specified.
- Add and update unit tests for the redactor and user statement formatting.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

PR #69383, opened 2026-02-20
[BugFix] audit log no longer has client id

srihithg:
## Why I'm doing:

client_id and client_secret leaking in audit logs. OAuth2 client_id keys (e.g. azure.adls2.oauth2_client_id, azure.blob.oauth2_client_id) were not in the redaction list, and FILES() with only OAuth2 credentials did not set needEncrypt, so raw SQL could be logged. This fix ensures OAuth2 client_id and client_secret are redacted everywhere they are logged.

## What I'm doing:

- **SqlCredentialRedactor**: Add OAuth2 client_id keys (azure.adls2.oauth2_client_id, azure.blob.oauth2_client_id, azure.adls1.oauth2_client_id) to the credential redaction list so they are replaced with *** in logged SQL.
- **AuditEncryptionChecker**: Treat OAuth2 client_id (and existing client_secret) in properties as sensitive in hasSecretInProperties() so needEncrypt is true for FILES() and similar statements with OAuth2 credentials.
- **ConnectProcessor**: formatStmt() already uses SqlCredentialRedactor.redact() when needEncrypt is true and in the fallback path; no code change needed once the above are in place.
- **SimpleExecutor**: Internal SQL is already logged via SqlCredentialRedactor.redact(); no code change needed.
- **PrintableMap**: Add OAuth2 client_id keys to SENSITIVE_KEY so properties printed (e.g. in task JSON) show *** for client_id.
- **Tests**: SqlCredentialRedactorTest includes testRedactAzureOAuth2ClientId and related cases.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [x] Yes, this PR will result in a change in behavior.
- [ ] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [x] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [x] This pr needs auto generate documentation
[...]

PR #74812, opened 2026-06-15
[Enhancement] Skip SQL credential redaction when no credential marker is present

sevev:
## Why I'm doing:

`ConnectProcessor.formatStmt()` redacts every audited statement as defense in depth, so `SqlCredentialRedactor.redact()` runs on the audit hot path for every statement, including every prepared-statement execution (`handleExecute` -> `auditAfterExec` -> `formatStmt`). `redact()` always runs four regex passes and allocates four intermediate strings even when the SQL carries no credential at all.

After #70360 extended the redactor to also mask `IDENTIFIED BY` / `IDENTIFIED WITH` / `SET PASSWORD` clauses, each `redact()` call became four full re2 scans plus four `StringBuilder` allocations. Combined with the unconditional defense-in-depth redaction added in #69383, this runs on every statement and shows up as high FE CPU under prepared-statement workloads, where the same statement is executed at high frequency.

## What I'm doing:

Short-circuit `redact()` with the existing cheap `mayNeedCredentialRedaction()` check (a single lowercase scan for credential markers): when the text has no credential marker, return the input unchanged without running any regex pass. The check already covers every key handled by the redaction patterns (`IDENTIFIED`, `password`, and all credential property keys), so when it returns false none of the patterns can match and the output is identical to before — this is a pure fast path. All `redact()` call sites benefit, including the audit hot path.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/SqlCredentialRedactor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/SqlCredentialRedactor.java
@@ -176,6 +176,14 @@ public class SqlCredentialRedactor {
             return sql;
         }
 
+        // Fast path: most statements carry no credentials. Skip the regex passes (and their
+        // StringBuilder allocations) entirely when there is no credential marker. This guards the
+        // hot audit path (ConnectProcessor.formatStmt redacts every statement as defense in depth),
+        // which is otherwise re-run on every prepared-statement execution and shows up as FE CPU.
+        if (!mayNeedCredentialRedaction(sql)) {
+            return sql;
+        }
+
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

jaogoy (approved):

HangyuanLiu (approved):

PR #74827, opened 2026-06-15
[BugFix] Fix UTF8_ERROR when reading gzip-compressed JSON Hive external tables

stephen-shelby:
## Why

When querying a Hive external table backed by **gzip-compressed JSON files with OpenX SerDe** (`org.openx.data.jsonserde.JsonSerDe`), StarRocks raises `UTF8_ERROR` and the query fails entirely:

```
parse error. Failed to iterate document stream as object.
error: UTF8_ERROR: The input is not valid UTF-8
file = s3://.../part-7-11243.json.gz
```

**Root cause**: `HdfsJsonReader` reads the decompressed stream in fixed 8 MB buffers and passes each buffer directly to simdjson's `iterate_many()`. When a multi-byte UTF-8 character (e.g. a 3-byte CJK character or 4-byte emoji) straddles the physical end of a buffer, simdjson's UTF-8 validator sees an incomplete byte sequence and raises `UTF8_ERROR` — it does not treat it as a truncated document. The existing `truncated_bytes()` carry-over mechanism only handles structurally incomplete JSON records; it cannot recover from a raw UTF-8 byte boundary split.

This only affects the OpenX SerDe path (`HdfsJsonScanner`). The HCatalog SerDe path reads line-by-line and is unaffected.

## What

Before passing each buffer to simdjson, detect any incomplete trailing UTF-8 sequence (1–3 bytes of an unfinished multi-byte character) and exclude those bytes from the current parse window. Add them back into the carry-over count alongside `truncated_bytes()` so they are prepended to the next read, completing the split character.

Changes:
- **`be/src/base/string/utf8.h`** — add `incomplete_trailing_utf8_len()`, a small helper that scans back at most 4 bytes from the buffer end using the existing `UTF8_BYTE_LENGTH_TABLE` to detect an incomplete trailing sequence
- **`be/src/exec/hdfs_scanner/hdfs_scanner_json.cpp`** — call the helper in `_read_and_parse_json()` before `iterate_many`, store the tail length in `_utf8_partial_tail`, and add it to `truncated_bytes()` in the carry-over path
- **`be/src/exec/hdfs_scanner/hdfs_scanner_json.h`** — add `_utf8_partial_tail` member
- **`be/test/exec/hdfs_scanner/hdfs_scanner_json_test.cpp`** — add 3 unit tests covering a 3-byte character split with 1 byte in the first buffer, 2 bytes in the first buffer, and a 4-byte emoji split across the boundary

## Verification

**Algorithm**: `incomplete_trailing_utf8_len()` tested against 15 cases (ASCII, complete multi-byte, each truncation variant 1–3 bytes, stray continuation bytes, invalid leading bytes) — all pass.

**Root cause confirmed**: Using real simdjson 4.6.4 (`libsimdjson.a`), a buffer ending with a partial multi-byte character produces `UTF8_ERROR`; a buffer ending with a cleanly truncated JSON document is handled correctly via `truncated_bytes()`.

**End-to-end**: A simulation replicating the full `HdfsJsonReader::next_record` loop (real simdjson, real `ByteBuffer` flip semantics) sweeping all 57 buffer sizes from 24–80 bytes with CJK/emoji payloads: **8 `UTF8_ERROR` failures without the fix, 0 with the fix**.

**Real cluster**: Reproduced the exact error on a Hive external table (OpenX SerDe, `.json.gz`, 64 MB uncompressed / ~500 KB compressed). `SELECT COUNT(1)` returned `UTF8_ERROR` before the fix; after deploying the patched BE it returned the correct row count `3725`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/base/string/utf8.h
+++ b/be/src/base/string/utf8.h
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include <algorithm>
+#include <cstdint>
 #include <cstring>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

dirtysalt (approved):
LGTM! 🚀

Youngwb (approved):

PR #74828, opened 2026-06-15
[Enhancement] Support Sample-Based Tablet Pre-Split for INSERT-from-table loads

xiangguangyxg:
## Why I'm doing:

Sample-Based Tablet Pre-Split previously had only two consumers — INSERT-from-FILES and Broker Load. `INSERT INTO <range-distributed target> SELECT ... FROM <olap table>` did not trigger pre-split at all, so the target's range partition stayed single-tablet and the load wrote serially through one CN instead of fanning out across the cluster.

While adding that third consumer it became clear the three entry hooks (`InsertFromFilesPreSplitHook`, `InsertFromTablePreSplitHook`, `BrokerLoadPreSplitHook`) each carried a near-verbatim copy of the same downstream flow (single/multi-partition dispatch → submit → sync-await; the multi-partition sample→group→combined-submit; the direct data-tier sampler). This PR therefore both **adds the INSERT-from-table consumer** and **unifies all three hooks** so that flow lives in one place.

## What I'm doing:

**1. New consumer — INSERT-from-table.** When `INSERT INTO <range-dist target> SELECT <*|bare columns> FROM <single OLAP table> [WHERE <deterministic, function-free predicate>]` is eligible, the hook data-tier samples the source via a sub-query, plans K-1 row-quantile boundaries, and submits a tablet reshard through the existing coordinator/pipeline/grouper so the triggering INSERT plans against the post-split layout. It runs pre-plan (holds no locks → cannot deadlock the reshard daemon) and forces the data tier (OLAP sources have no Parquet/ORC footer).

**2. Unified architecture.** The triplicated flow is extracted into one shared package-private `PreSplitFlow` (the `Prepared` carrier + `dispatch` + single/multi-partition flow + direct data-tier sampler). On top of it:
- `InsertPreSplitHook` (single `StmtExecutor` entry, replacing the two old INSERT hooks) selects an `InsertPreSplitSource` strategy — `FilesPreSplitSource` or `TablePreSplitSource` — that resolves the source and builds a `Prepared`.
- `BrokerLoadPreSplitHook` is slimmed to a thin `BrokerLoadJob` entry that builds a `Prepared` and calls the same `PreSplitFlow` (its 8-arg entry signature is unchanged; `BrokerLoadJob.java` is untouched).
- The two old INSERT hooks are deleted; `StmtExecutor` goes from two hook calls to one; `TabletPreSplitCoordinator` Javadoc `@link`s are repointed. Net: **−452 lines** of production pre-split code.

**Eligibility (strict, fail-safe — anything uncertain is silently skipped and the load runs normally):** single plain OLAP source (rejects FILES/JOIN/UNION/subquery/CTE and any modifier such as PARTITION/TABLET/TABLESAMPLE/temporal); identity projection (`SELECT *` or all bare column refs) with by-position/by-name exact-set mapping onto the target sort key; deterministic, function-free `WHERE` only; insertable (non-materialized-view) range-distributed target with no target-partition clause. The sampler sub-query runs under a ROOT statistics context, so the table path performs an explicit source `SELECT` authorization check plus a row-access / column-masking policy gate before sampling; the target `INSERT` + non-default-warehouse `USAGE` authorizations are checked before any side effect.

**Isolation:** the two footer readers (`ParquetRowGroupStatisticsReader` / `OrcStripeStatisticsReader`) and `BrokerLoadJob.java` are zero-change.

**Config:** new FE `enable_tablet_pre_split_for_insert_from_table` (default `true`), documented in en/zh/ja.

### Behavior notes for reviewers (all conservative *skips* — never fail or change the result of a load)

Unifying the hooks moves four conservative-skip gates into the shared path, which the GA FILES / Broker paths now also observe. None can break a previously-working load; the only effect is that pre-split (a transparent layout optimization) is skipped in narrow cases where it was previously attempted:
- **Automatic-partition gate** (now in `PreSplitFlow.dispatch`, so it covers FILES + Broker as well as the table path): a manually list/range-partitioned target is no longer pre-split. Previously the multi-partition flow could pre-create system partitions from sampled values on such a target — undesirable; this is intended hardening.
- **FILES path additionally skips** when the INSERT carries a target-partition clause, carries load `PROPERTIES(...)`, or targets a materialized view (the table path already did).
- The `tablet_pre_split` `eligibility_skipped{reason="disabled_by_config"}` counter is now recorded only after a real INSERT-from-FILES/table candidate is identified (previously it was bumped for every statement, including non-INSERTs).

Load semantics, results, and success/failure are unchanged in every case, so this is classified as no change in behavior.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg:
## Tri-path TSP e2e — ALL PASS ✅

Build **2020** (= branch head `1c1ece9`), shared-data cluster `unify-presplit-e2e` (1 FE 4c32g + 3 CN 16c64g, aliyun cn-zhangjiakou). Each path drives a fresh range-distributed (1-tablet) target; metrics from FE `:8030` + the 3 CN `:8040`.

| Path | tier | tablets | fallback (×3 CN) | rows | result |
|---|---|---|---|---|---|
| **INSERT-from-table** `INSERT INTO t SELECT * FROM <olap>` (1M) | data_tier | 1→3 (cuts 333103 / 667386 ≈ quantile thirds) | 0 | 1M==1M | ✅ |
| **INSERT-from-FILES** `SELECT * FROM FILES(parquet)` | data_tier | 1→3 (composite `(group_id,position_id)` cuts) | 0 | 161,566 | ✅ |
| **Broker Load** `LOAD … FROM parquet` | data_tier | 1→3 (composite cuts) | 0 | 161,566, FINISHED | ✅ |
| **Manual-partition skip** (new auto-partition gate) `INSERT INTO <manual RANGE-partitioned> SELECT *` | — | unchanged (no split) | 0 | 1M==1M | ✅ skipped, load OK |

Aggregate across the run: `tablet_pre_split_sampler_invocations` 0→**3** (exactly one per eligible path — the manually-partitioned case correctly did **not** sample, proving the auto-partition gate short-circuits before sampling), `tier_used{data_tier}`=3, `boundaries_planned`=3; CN `tablet_reshard_split_count`=3 / `output_tablet_count`=9, `split_failed`=0, `tablet_reshard_split_external_boundaries_fallback_total`=**0** on all 3 CNs.

(FILES/Broker used a single parquet shard of the eightfoldai dataset — a single file's row-groups overlap in sort order, so the meta tier correctly falls back to data_tier; that is the expected layout artifact, not a regression. The unified `PreSplitFlow` + strategy layer drives all three GA paths identically end-to-end.)

Cluster released.

xiangguangyxg:
Addressed the Codex P2 on `AbstractSqlSampleSubqueryExecutor.configureSampleContext` ("Preserve the sampler query timeout after warehouse switch") in `9bf2b0b1c05`.

Confirmed real: `ConnectContext.setCurrentWarehouseId` → `setCurrentWarehouse` does `this.sessionVariable = variableMgr.newSessionVariable()` and re-applies only **tracked SET** variables. The pre-submit-budget `query_timeout` was applied via the direct setter `getSessionVariable().setQueryTimeoutS(...)` (not a tracked SET), so setting it *before* the warehouse switch dropped it — an over-budget data-tier sample would then run to the warehouse/default timeout and block the load past the pre-submit budget instead of failing fast + falling back.

Fix: set `query_timeout` **after** `setCurrentWarehouseId`/`setCurrentComputeResource`, on the final session variable. Added `AbstractSqlSampleSubqueryExecutorTest` with an in-order guard (verified it fails on the pre-fix ordering). The tri-path e2e above didn't surface this because the samples were sub-second; it only bites on large/slow sources.

(The two earlier Codex findings on `0b605345e7` — P1 auto-partition pre-create, P2 validate-INSERT-properties — were already fixed in `fd7d072bc92` and are now the shared `passesCommonPreFilters` + `PreSplitFlow.dispatch` gates; the auto-partition one is e2e-validated by the manual-partition skip case above.)

CelerData-Reviewer:
@codex review

xiangguangyxg:
Both Codex P2s on `9bf2b0b1c0` addressed in `1935fe00cd4`:

**1. Temporary-table source** (`TablePreSplitSource.java`) — confirmed real. The hook resolves the source session-aware (sees a session temp table), but the data-tier sampler runs as `UserIdentity.ROOT` in a fresh statistics `ConnectContext` that has no temp tables, so a temp source shadowing a permanent table would be sampled as the *permanent* one → boundaries off. Fix: `resolveSourceTable` now returns null (skip pre-split) when `sourceTable.isTemporaryTable()`. Added `prepareSkipsTemporaryTableSource` (bite-verified: builds a scan context without the gate, null with it).

**2. Rollback checklist** (`stats_storage.md`) — fixed in en/zh/ja: step 1 now sets all three flags (`…_insert_from_files`, `…_broker_load`, `…_insert_from_table`) to `false`, so a downgrade no longer leaves INSERT-from-table pre-split active.

Both are conservative skips (load runs normally; pre-split just doesn't fire), so behavior-change classification is unchanged. FE incremental coverage + param-drift + translation checks remain green.

kevincai (approved):

luohaha (approved):

PR #74837, opened 2026-06-16
[BugFix] Drop Iceberg timestamptz file-level min/max across DST fall-back

eshishki:
## Why I'm doing:

The Iceberg datetime min/max aggregate optimization (#71870) answers
`SELECT MIN/MAX(ts)` from file-level column bounds. For a `TIMESTAMP WITH TIME ZONE`
(timestamptz) column it converts each file's UTC lower/upper bound to session-local
micros one endpoint at a time.

`local = utc + offset(utc)` is non-monotonic across a backward UTC-offset change —
a DST fall-back (e.g. America/Los_Angeles PDT -07:00 -> PST -08:00) or a historical
zone-rule change. When a data file's values straddle such a transition, the true
local min/max can lie strictly inside the file, so converting only the endpoints
yields an inexact bound and `MIN/MAX(ts_tz)` can return a wrong result (too-small MAX
and/or too-large MIN) compared to a full scan. The optimization is on by default
(`enable_min_max_optimization=true`), so this silently returns wrong aggregates.

## What I'm doing:

Skip emitting file-level min/max for a timestamptz column when the file's `[low, high]`
instant range contains an offset-decreasing transition (checked via `ZoneRules`). BE
then reads that file directly through the existing per-file fallback in
`HdfsScannerContext`, while files without such a transition stay optimized; the
aggregate combines them correctly.

Constant-offset and spring-forward ranges keep the optimization with exact endpoints
(a spring-forward offset increase keeps UTC->local monotonic, so the endpoints remain
the true local min/max). `timestamp` without time zone is unaffected. Added unit tests
for the fall-back (dropped), within-single-offset (kept), and spring-forward (kept) cases.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergUtil.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergUtil.java
@@ -44,6 +44,8 @@ import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
+import java.time.zone.ZoneOffsetTransition;
+import java.time.zone.ZoneRules;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


eshishki:
@dirtysalt PTAL

CelerData-Reviewer:
@codex review

PR #74868, opened 2026-06-16
[BugFix] Apply session timezone shift to nested INT96 timestamps

banmoy:
## Why I'm doing:

When `FILES()` or Broker Load reads a Hive/parquet-mr Parquet file in which a datetime is encoded as the legacy **INT96** physical type and **nested inside an ARRAY, MAP, or STRUCT** (e.g. `array<datetime>`), the value loses one session-timezone offset — it is returned 8 hours short under `Asia/Shanghai`. Top-level INT96 columns are correct; only nested leaves are wrong.

INT96 has no `isAdjustedToUTC` flag: the writer normalizes it to a UTC instant, but `parquet::arrow` decodes it as a timezone-naive Arrow `TimestampType` — byte-identical to an INT64 `isAdjustedToUTC=false` wall-clock column. The native reader (`Int96ToDateTimeConverter`) still treats INT96 as a UTC instant and shifts it to the session timezone; the Arrow-based legacy path (`ParquetReaderWrap`) does not, so the two diverge for nested INT96.

This is a regression deliberately left open by #73674. That PR made timezone-naive Arrow timestamps read as wall-clock (correct for INT64 `isAdjustedToUTC=false`), and re-tagged INT96 back to `"UTC"` via `_rectify_int96_timezone()` so it would still shift — but **scoped the re-tag to top-level columns only** (`dot_vector.size() == 1`) and declared nested INT96 explicitly **out of scope** in the PR body and in the maintainer review ("Acceptable given INT96 is legacy and rarely nested ... please state explicitly that nested INT96 is out of scope so it is a deliberate decision rather than a silent gap"). Issue StarRocks/StarRocksTest#11421 is a real user hitting that documented gap. This PR completes it.

## What I'm doing:

Extend the INT96 re-tag in `ParquetReaderWrap` to nested columns, while preserving #73674's core invariant (INT64 `isAdjustedToUTC=false` stays wall-clock, no shift) and re-tagging **only** genuine INT96 leaves.

- `be/src/exec/parquet_reader.cpp` — `_init_parquet_reader()`: drop the `dot_vector.size() == 1` guard. A top-level column is recorded in `_int96_columns` whenever **any** of its leaves is INT96, including leaves nested inside ARRAY/MAP/STRUCT. This is a coarse skip-filter only.
- `be/src/exec/parquet_reader.cpp` — `_rectify_int96_timezone()` resolves the precise per-leaf decision using the `parquet::arrow::SchemaManifest` from the open `FileReader`. The recursive helper `rectify_int96_array_data()` walks the Arrow `ArrayData` tree in lockstep with the parquet `SchemaField` tree (isomorphic by child index: STRUCT child *i*, LIST element child, MAP `struct<key,value>` entries) and re-tags a timezone-naive timestamp leaf to `"UTC"` **only when its parquet column is physically INT96** (`descr->Column(column_index)->physical_type() == INT96`). A sibling INT64 `isAdjustedToUTC=false` (wall-clock) leaf in the same complex column is left untouched — no over-tagging. Matching is purely structural (by child index), never by Arrow field name, which is unreliable for list/map child fields on read.
- The change is metadata-only: nested types are rebuilt bottom-up via the field-preserving Arrow factories (`arrow::list` / `large_list` / `fixed_size_list` / `map` / `struct_`), and buffers are aliased unchanged via `ArrayData::Copy()`, so there is no per-row cost; the `RecordBatch` is rebuilt only when a column was re-tagged.

The downstream converter (`arrow_to_starrocks_converter.cpp`) is unchanged — a re-tagged `"UTC"` leaf takes the same session-timezone rectification path as a top-level INT96 column, matching the native `Int96ToDateTimeConverter`.

### Docs

`docs/{en,zh,ja}/sql-reference/sql-functions/table-functions/files.md` — the read-side mapping note added by #73674 stated that nested INT96 is read as a wall-clock value with no session shift. Updated it to state that StarRocks treats an INT96 column as a UTC instant and converts it to the session time zone **whether it is top-level or nested inside a STRUCT/ARRAY/MAP**, so the contract matches the fixed behavior.

## Tests

- `be/test/exec/file_scanner/parquet_scanner_test.cpp` — `ParquetScannerTest.int96_timestamp_nested_session_timezone_shift`: writes a Parquet file with a top-level timestamp plus `list<timestamp>` and `list<list<timestamp>>`, all INT96 (`enable_deprecated_int96_timestamps()`), and drives the full `ParquetScanner` → `ParquetReaderWrap` → arrow-converter path. Under a `"UTC"` session every level renders `2024-06-15 12:00:00`; under `-08:00` every level renders `2024-06-15 04:00:00` (and never `12:00:00`) — asserting nested leaves shift by the session offset exactly like the top-level column. Fails on `main` (nested unshifted), passes with this fix.
- `be/test/exec/file_scanner/parquet_scanner_test.cpp` — `ParquetScannerTest.int96_timestamp_mixed_nested_no_over_tagging`: writes a `struct<a: INT96, b: INT64 isAdjustedToUTC=false>` via the **low-level parquet column writers** (the high-level Arrow writer's `enable_deprecated_int96_timestamps()` is a global switch and cannot mix encodings in one file). Under `-08:00` the INT96 leaf `a` shifts to `04:00:00` while the INT64 wall-clock leaf `b` stays at `12:00:00` — guarding against over-tagging a sibling leaf in the same complex column.
- `be/test/exec/file_scanner/parquet_scanner_test.cpp` — `ParquetScannerTest.int96_timestamp_map_value_session_timezone_shift`: writes a `MAP<VARCHAR, DATETIME>` whose timestamp values are INT96 (exercising the MAP branch, whose child is a `struct<key,value>`). Under `-08:00` the INT96 map value shifts to `04:00:00` while the VARCHAR key is unaffected.
- Verified live (`Asia/Shanghai`): pre-fix `col_arr_dt[1]` / `col_arr_arr_dt[1][1]` = `12:00:00` (8 h short) while `col_dt_top` = `20:00:00`; post-fix all three = `20:00:00`. UTC baseline unchanged.

Fixes https://github.com/StarRocks/StarRocksTest/issues/11421

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5


CelerData-Reviewer:
@codex review

dirtysalt (dismissed):
LGTM! 🚀

CelerData-Reviewer:
@codex review

kevincai (approved):

dirtysalt (approved):
LGTM! 🚀

PR #74906, opened 2026-06-16
[BugFix] Debounce autovacuum minActiveTxnId before txn-log deletion

kevincai:
A begin-transaction vs autovacuum race could make computeMinActiveTxnId()
momentarily one greater than an in-flight transaction: beginTransaction() draws
an id (advancing the global generator / peekNextTransactionId()) before
registering it into idToRunningTransactionState, and an autovacuum probe landing
in that gap falls back to the peek seed. The resulting min_active_txn_id is sent
to the BE, whose vacuum_txn_log then deletes that transaction's still-needed
combined txn log. Its publish can never proceed afterwards ("Both txn_log and
corresponding tablet_meta missing"), permanently wedging the partition.

Defend in AutovacuumDaemon by never acting on a single, unconfirmed watermark:
keep a per-partition lastMinActiveTxnId and, on a round whose minActiveTxnId is
confirmed by the previous round (non-decreasing), sweep txn logs using that
older, confirmed value. When the value is unconfirmed -- first observation
(including after an FE restart/failover, since lastMinActiveTxnId is in-memory
and resets to 0) or a regression (logged at WARN) -- skip the ENTIRE vacuum round
rather than only the txn-log delete. Skipping only the delete would still let the
round advance lastSuccVacuumVersion, and with
lake_autovacuum_detect_vaccumed_version=true shouldVacuum() would then stop
scheduling the partition, so the confirming follow-up round (and its sweep) might
never run for a partition that goes cold. A full skip leaves lastSuccVacuumVersion
untouched, so the partition stays schedulable and the next round runs with a
confirmed watermark; the only cost is deferring that partition's metadata/data
vacuum by one (rare) cycle.

PhysicalPartition.lastMinActiveTxnId is in-memory runtime state only (no
@SerializedName, so excluded from the image and edit log); it resets to 0 on
restart/failover, which conservatively triggers the first-round skip.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
@@ -155,6 +155,11 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
 
     private final AtomicLong lastSuccVacuumVersion = new AtomicLong(0);
 
+    // Purpose: the previous autovacuum round's computeMinActiveTxnId(), used to debounce a
+    //   transiently-too-high value (begin-vs-vacuum race) before allowing txn-log deletion.
+    // Persistence: in-memory only, NOT persisted (no @SerializedName); resets to 0 on restart/failover.
+    private volatile long lastMinActiveTxnId = 0;
+
     @SerializedName(value = "bucketNum")
     private int bucketNum = 0;
     
@@ -286,6 +291,14 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
 
     public void setLastSuccVacuumVersion(long lastSuccVacuumVersion) {
         this.lastSuccVacuumVersion.set(lastSuccVacuumVersion);
+    }
+
+    public long getLastMinActiveTxnId() {
+        return lastMinActiveTxnId;
+    }
+
+    public void setLastMinActiveTxnId(long lastMinActiveTxnId) {
+        this.lastMinActiveTxnId = lastMinActiveTxnId;
     }
 
     public long getExtraFileSize() {
--- a/fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
@@ -178,6 +178,38 @@ public class AutovacuumDaemon extends FrontendDaemon {
         long minRetainVersion;
         long startTime = System.currentTimeMillis();
         long minActiveTxnId = computeMinActiveTxnId(db, table);
+
+        // Confirmed/lagged-watermark debounce, against a begin-transaction vs autovacuum race:
+        // beginTransaction() draws an id (advancing peekNextTransactionId()) BEFORE registering it in
+        // idToRunningTransactionState, so a probe landing in that gap can compute a minActiveTxnId one
+        // greater than an in-flight txn. Acting on it would let the BE delete that txn's still-needed
+        // combined log and permanently wedge publish on the partition. So we only act on a value confirmed
+        // by the PREVIOUS round (non-decreasing) and sweep txn logs with that older, confirmed value.
+        //
+        // When the current value is NOT confirmed -- first observation (also after FE restart/failover,
+        // since lastMinActiveTxnId is in-memory and resets to 0) or a regression -- we skip the ENTIRE
+        // round, not merely the txn-log delete. Skipping only the delete would still let the round advance
+        // lastSuccVacuumVersion; with lake_autovacuum_detect_vaccumed_version=true, shouldVacuum() then
+        // stops scheduling the partition once lastSuccVacuumVersion >= minRetainVersion, so for a partition
+        // that goes cold the confirming follow-up round (and its sweep) might never run, leaking txn logs.
+        // A full skip leaves lastSuccVacuumVersion untouched, so the partition stays schedulable and the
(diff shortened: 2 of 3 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
```diff
@@ -294,6 +294,35 @@ private void vacuumPartitionImpl(Database db, OlapTable table, PhysicalPartition
...
+        // partition. Defend by never acting on a single, unconfirmed observation: sweep txn logs using the
+        // PREVIOUS round's value (gated by a non-decreasing check), and skip the sweep on any regression.
+        // minActiveTxnId is only consumed by the BE when deleteTxnLog=true, so gating deleteTxnLog is enough.
+        long lastMinActiveTxnId = partition.getLastMinActiveTxnId();
+        long txnLogSweepWatermark = minActiveTxnId;
+        boolean allowTxnLogSweep;
+        if (lastMinActiveTxnId <= 0) {
+            // First observation since this FE became leader / the partition was first seen this run:
+            // there is no confirmed predecessor, so do not sweep txn logs this round.
+            allowTxnLogSweep = false;
```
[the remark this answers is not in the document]
kevincai: 1. It's no correctness impact. 2. The left old txn logs is bounded. 3. It's self-headling, once the partition is written again, the vacuum process will find the correct minActiveTxnId and clean the left ones.

fe/fe-core/src/main/java/com/starrocks/catalog/PhysicalPartition.java
```diff
@@ -155,6 +155,11 @@ public class PhysicalPartition extends MetaObject implements GsonPostProcessable
 
     private final AtomicLong lastSuccVacuumVersion = new AtomicLong(0);
 
+    // Purpose: the previous autovacuum round's computeMinActiveTxnId(), used to debounce a
+    //   transiently-too-high value (begin-vs-vacuum race) before allowing txn-log deletion.
+    // Persistence: in-memory only, NOT persisted (no @SerializedName); resets to 0 on restart/failover.
+    private volatile long lastMinActiveTxnId = 0;
```
banmoy, line 161: how about adding a sanity check when committing transaction. If the txn id < lastMinActiveTxnId, the transaction is not protected and not safe to commit, and fail it aggressively

kevincai: It will be false positive with current code fix. For example, if the TxnId = 740 is running, while the lastMinActiveTxnId bump to 741. In current implementation, the lastMinActiveTxnId won't be reset to 740 until next round of vacuum, which would be up to 3 mins window with the default configuration. So if txnId-740 commit anytime during this time window, it will fail aggressively even there is no vacuum job running.

While current fix can prevent the conflict window from millisecond to 3 mins. This additional check will cause the false positive up to 3 mins for the certain txn, and it only protects those txns that hold for larger than 3 mins without registering itself after got the allocated txn id.

banmoy: You're right, the check can't be based on `lastMinActiveTxnId`. A solution is to add a field on the partition recording the actually-dispatched `txnLogSweepWatermark`, and compare the committing `txnId` against it. If it's below the watermark, just log it or fail the transaction commit aggressively.

kevincai: With the mid term solution in plan and the longer enough conflict window in current implementation, I am opt to no additional band-aid to the code.

banmoy (approved):

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
```diff
@@ -178,6 +178,38 @@ private void vacuumPartitionImpl(Database db, OlapTable table, PhysicalPartition
...
+                                + "(possible begin/vacuum race)",
+                        lastMinActiveTxnId, minActiveTxnId, db.getFullName(), table.getName(), partition.getId());
+            }
+            partition.setLastMinActiveTxnId(minActiveTxnId);
+            partition.setLastVacuumTime(startTime);
+            return;
+        }
+        // Confirmed non-decreasing: sweep txn logs with the previous (older, lower) value.
+        final long txnLogSweepWatermark = lastMinActiveTxnId;
+        partition.setLastMinActiveTxnId(minActiveTxnId);
```
luohaha, line 211: How about `FullVacuumDaemon` path? I think it need same fix too.

kevincai: FullVacuumDaemon has been set to false by default, and it will conflict with autovacuum's lastMinActiveTxnId. Will leave it for now since it is off for now. Will fix it by the txn id allocation solution which will be cleaner and applies to both autoVacuum and fullVacuum.

luohaha (approved):

sevev (approved):

PR #74911, opened 2026-06-16
[Enhancement] Prune Iceberg TopN scans by per-file min/max

eshishki:
> Depends on #74837 (Iceberg timestamptz DST min/max fix) — its commit is the first one in this branch. The change to review is the 4 commits on top.

## Why I'm doing:

A `ORDER BY <col> LIMIT k` query on a large Iceberg table opens every data file and reads its Parquet footer. Only a few files hold the top-k rows; the others are read and then dropped.

DataFusion 54 shipped a new TopK optimization: order row groups by their min/max statistics and exit early once the limit is reached. StarRocks already has most of this through the TopN runtime filter (`enable_topn_runtime_filter`): the sort+limit builds a runtime filter and pushes it to the scan.

This PR makes that runtime filter more effective on Iceberg, behind a new flag `enable_iceberg_topn_scan_pruning` (off by default).

## What I'm doing:

**Commit 1 — order files by min/max.**
We order the scan morsels by each file's min/max on the sort column, taken from the Iceberg manifest: min first for ASC, max first for DESC. The runtime filter then reaches its final value sooner and prunes more files.

**Commit 2 — apply the filter before the footer.**
A file whose min/max cannot pass the current filter has no top-k rows. Today the scan reads such a file's footer first and prunes its row groups only after that. We move the check earlier: before opening the footer, we test the file's manifest min/max against the live filter and skip the whole file. Same check, on the file-level min/max from the manifest instead of the row-group stats in the footer — the footer read is saved.

The query result does not change. The flag only changes which files are read and in what order.

## How

**FE.** A new rule `IcebergTopNScanPruneRule` marks the Iceberg scan under a qualifying TopN (has limit, runtime filter enabled, leading sort key is a non-partition int/date/datetime column). It sets a flag on the scan and ships the sort-key slot id, direction and null order in `THdfsScanNode`. The Iceberg scan-range source then ships the per-file min/max for that column in `THdfsScanRange.min_max_values` — the same bounds the min/max aggregate optimization already uses.

**BE — reorder.** A new `PriorityMorselQueue` orders morsels by the raw int64 bound from the scan range. No decode is needed: the raw Iceberg encodings (int, date as days, timestamp as micros) are monotonic, so raw order equals decoded order. The queue is forced to a single shared work-stealing queue (`can_uniform_distribute() == false`); a per-driver split would shard the global order and defeat the reorder. Split pieces of one file share a scan range, so they get the same key and the per-owner ticket checker still works.

**BE — skip.** `HdfsScanner::open()` calls `should_skip_scan_range_by_topn_min_max()` after the runtime-filter pruner is built and before `do_open()` reads the footer. It decodes the file's min/max into the slot type and checks it two ways: (a) the root-level predicates of the predicate tree, where an already-arrived runtime filter is folded in as a min/max range; (b) a throwaway copy of the runtime-filter pruner, for filters that arrive later. It uses only the root-level predicates, not the full flattened map, so an `OR` is never AND-tested by mistake. The pruner is copied because the Parquet footer path copies it for its own row-group skip — mutating the original would suppress that.

## Numbers

A production Iceberg table: ~1180 data files, 38.5 GB, ~1280 row groups. `ORDER BY <key> DESC LIMIT 100`, cold caches (`enable_scan_datacache=false`, `enable_file_metacache=false`).

| Metric | id OFF | id ON | datetime OFF | datetime ON |
|---|---|---|---|---|
| Total | 1948 ms | **749 ms** | 1911 ms | **714 ms** |
| ExecWallTime | 1497 ms | 300 ms | 1412 ms | 299 ms |
| ReaderInitFooterRead | 767 ms | **7.5 ms** | 763 ms | **10.9 ms** |
| OpenFile | 805 ms | 30 ms | 819 ms | 37.6 ms |
| ScanTime | 857 ms | 46 ms | 895 ms | 51 ms |
| Prepare (manifest planning) | 288 ms | 303 ms | 302 ms | 264 ms |
| MorselQueueType | dynamic | priority | dynamic | priority |
| Eligible / NoBound morsels | — | 1180 / 0 | — | 1182 / 0 |
| Files skipped before footer | 0 | **1156** | 0 | **1158** |
| Scan ranges (total) | 1181 | 1181 | 1182 | 1182 |

Result identical off vs on (sha1), both keys. The whole win is in execution; Prepare (manifest planning, ~290 ms) is a fixed floor the feature does not touch.

## Tested

- e2e SQL test (`test/sql/test_iceberg/test_iceberg_topn_scan_pruning`): a 5-file table with disjoint id/ts ranges. `ORDER BY id|ts DESC LIMIT 3` gives the same result off and on; a string sort key (not eligible) falls back to the dynamic morsel queue; an `OR` predicate on the sort key is a regression test for the predicate-tree fix. Profile asserts the priority morsel queue and `NoBound == 0`.
- BE unit tests for `should_skip_scan_range_by_topn_min_max`: OR-predicate over-skip, all-null and NULLS-FIRST files, root-predicate skip/keep, disabled-path guards. Plus the existing `decode_min_max_endpoint` cases.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

PR #74924, opened 2026-06-17
[Enhancement] Add checksum protection for shared-data tablet metadata and txn log

luohaha:
## Why I'm doing:

In the shared-data architecture, lake tablet metadata and txn log files are
serialized as plain protobuf with no header and no checksum. Silent corruption of these
files (truncation, bit-rot, partial writes on object storage) cannot be detected on read,
and a corrupted tablet meta / txn log can lead to hard-to-diagnose failures.

The shared-nothing engine already protects its metadata with `ProtobufFileWithHeader`,
which carries an Adler-32 checksum, but that path hardcodes `OLAP_FIX_HEADER_MAGIC_NUMBER`
(== 0, a poor magic) and rejects anything else, so it cannot be reused as-is for lake.

## What I'm doing:

Add an opt-in checksummed on-disk format for shared-data tablet metadata and txn logs,
with full backward/forward read compatibility.

- Introduce `LAKE_META_HEADER_MAGIC_NUMBER`, chosen so its little-endian first on-disk
  byte is `0xFE`. A serialized `TabletMetadataPB` / `TxnLogPB` starts with the field-1
  varint tag (`0x08`) and protobuf can never produce a leading `0xFE` byte, so a reader
  can distinguish the new checksummed format from legacy headerless protobuf with
  effectively zero false-positive risk.
- Parameterize `ProtobufFileWithHeader` with `magic` and `allow_plain_protobuf_fallback`
  (both defaulted), so existing shared-nothing callers are byte-for-byte unchanged. When
  fallback is enabled, `load()` parses legacy headerless protobuf when the magic is absent.
- Add BE config `lake_enable_protobuf_file_checksum` (default `false`). It only controls the
  **write** format; readers always auto-detect and verify the checksum when present,
  regardless of the flag.
- Route lake tablet meta, txn log / slog / vlog, and combined txn log through the new
  save/load helpers in `TabletManager`, and fix `vacuum.cpp`'s own metadata reader to use
  the same auto-detecting load path.
- For bundle tablet metadata, protect both layers: a per-tablet page checksum
  (`map<int64,uint32> tablet_meta_page_checksum` in `BundleTabletMetadataPB`, same key as
  `tablet_meta_pages`) plus a footer crc32 over the shared bundle header (schemas, mappings,
  page pointers, and the page-checksum map itself). When the flag is on the footer becomes
  `[bundle_meta][crc32][size | high-bit flag]`; readers detect the layout via the high bit of
  the size field (collision-free, since a legacy bundle size is always well below 2^63) and
  fall back to the legacy `[bundle_meta][size]` when it is absent.
- Adapt `meta_tool` (`print_lake_metadata` / `print_lake_txn_log` /
  `print_lake_combined_txn_log`) to load via the header-aware path so it parses both formats;
  `print_lake_bundle_metadata` already goes through the updated `parse_bundle_tablet_metadata`.

Compatibility note: the read path is always backward compatible (a new BE reads legacy
headerless files via fallback; the checksummed format is auto-detected). The flag defaults to
`false` to stay safe during rolling upgrades and downgrades: while versions are mixed, an older
BE/CN uses the legacy reader and cannot parse newly-written checksummed files. Enable it only
after the whole cluster has been upgraded to a version that understands the format.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

sevev (dismissed):

CelerData-Reviewer:
@codex review

be/src/storage/olap_define.h
```diff
@@ -46,6 +46,15 @@ namespace starrocks {
...
 static const uint64_t OLAP_FIX_HEADER_MAGIC_NUMBER = 0;
 
+// Magic number for shared-data (lake) tablet metadata / txn log files that carry a
+// FixedFileHeader with an Adler-32 checksum. It is chosen so that its little-endian
+// first on-disk byte is 0xFE, which can never be the leading tag byte of a serialized
+// TabletMetadataPB / TxnLogPB (those start with the field-1 varint tag 0x08). Readers can
+// therefore distinguish the new checksummed format from legacy headerless protobuf by
+// inspecting the first bytes, with effectively zero risk of a false positive. The
+// remaining bytes spell "LAKEMTA". Never reuse OLAP_FIX_HEADER_MAGIC_NUMBER (0) here.
+static const uint64_t LAKE_META_HEADER_MAGIC_NUMBER = 0x41544D454B414CFEULL;
```
starrocks-xupeng, line 61: how this is made sure that the start will be `0x08` before

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -578,6 +597,10 @@ Status TabletManager::put_bundle_tablet_metadata(std::map<int64_t, TabletMetadat
 
         (*bundle_meta.mutable_tablet_meta_pages())[tablet_id] =
                 make_page_pointer(current_offset, serialized_buf.size());
+        if (config::lake_enable_protobuf_file_checksum) {
+            (*bundle_meta.mutable_tablet_meta_page_checksum())[tablet_id] =
+                    olap_adler32(ADLER32_INIT, serialized_buf.data(), serialized_buf.size());
```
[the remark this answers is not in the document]
luohaha: Good catch — confirmed the shared `BundleTabletMetadataPB` section (schemas, `tablet_to_schema`, page pointers) is not covered by the per-page checksum.

We're deferring full bundle-header integrity to a follow-up rather than fixing it here, for two reasons:

1. **Backward-compat constraint.** The bundle layout is `[per-tablet pages][BundleTabletMetadataPB][fixed64 size]`, and readers locate everything from the trailing size. Any checksum placed inside the bundle header is self-referential, and any checksum placed outside it shifts the offsets that older BEs rely on. A correct fix therefore requires a bundle file-format change, which is disproportionate to this PR.
2. **A self-checksum field is also fragile here**, not just heavy: `BundleTabletMetadataPB` is all `map` fields, and protobuf map serialization is non-deterministic by default, so a "zero the field, re-serialize, compare" scheme can false-positive unless deterministic serialization is forced.

Net effect of this PR is still strictly positive: the per-tablet metadata payload (the bulk and highest-value content) goes from unchecked to checksummed; the shared header remains exactly as before (unchecked). Added an inline comment documenting the gap, and a backward-compatible whole-bundle checksum (CRC over the exact written bytes, detected via a footer magic) is the planned follow-up.

be/src/storage/olap_define.h
luohaha: To clarify — the detection does not actually depend on the message starting with the field-1 tag `0x08` (protobuf doesn't guarantee field serialization order). The real guarantee is the chosen magic's first on-disk byte `0xFE`:

Every protobuf message begins with a field tag varint, and the wire type is the low 3 bits of that varint, i.e. the low 3 bits of the first byte. `0xFE` has low 3 bits `= 6`, and wire type 6 is not a valid/emitted protobuf wire type (only 0,1,2,3,4,5 exist; 3/4 are the deprecated groups). So no protobuf serializer ever produces a message whose first byte is `0xFE`, regardless of which field is serialized first. That's why a reader can safely use the first byte to tell the new checksummed format apart from a legacy headerless protobuf, with zero false-positive risk.

I've updated the comment on `LAKE_META_HEADER_MAGIC_NUMBER` to state this rationale instead of the misleading "starts with 0x08" wording.

xiangguangyxg (dismissed):

be/src/storage/lake/tablet_manager.cpp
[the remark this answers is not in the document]
luohaha: Update: addressed this in the latest push rather than deferring. The bundle now also protects the shared `BundleTabletMetadataPB` header via a footer crc32 + magic. When `lake_enable_protobuf_file_checksum` is on, the footer is written as `[bundle_meta][crc32(bundle_meta)][magic][size]` (legacy is `[bundle_meta][size]`); the size stays the trailing 8 bytes in both, so readers detect the layout by the magic at the fixed offset before the size and fall back to legacy when it's absent. The crc covers the schemas, tablet_to_schema, page pointers, and the per-page checksum map itself, so a bit flip in that section is now caught instead of silently changing a restored schema.

This sidesteps the self-checksum / map-nondeterminism problem since the crc is computed and verified over the exact written bytes (no re-serialization). Added BE UTs in `tablet_manager_test.cpp` covering: new-format footer magic present + read OK, legacy-format (no magic) read OK with the flag on, footer-crc catching a bundle-header corruption, and the per-page checksum catching a tablet-page corruption.

CelerData-Reviewer:
@codex review

be/src/common/config.h
```diff
@@ -1508,6 +1508,13 @@ CONF_mBool(enable_strict_delvec_crc_check, "true");
 // When the ratio of cumulative level to base level is greater than this config, use base merge.
 CONF_mDouble(lake_pk_index_cumulative_base_compaction_ratio, "0.1");
 CONF_Int32(lake_pk_index_block_cache_limit_percent, "10");
+// When true, shared-data (lake) tablet metadata and txn log files are written with a
+// FixedFileHeader carrying an Adler-32 checksum, so corruption can be detected on read.
+// Readers always auto-detect and verify the checksum when a file has it, regardless of this
+// flag; the flag only controls the write format. Defaults to true. Disable it before
+// downgrading to an older BE/CN that predates this format, because such a binary cannot read
+// the checksummed files.
+CONF_mBool(lake_enable_protobuf_file_checksum, "true");
```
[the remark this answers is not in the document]
luohaha: Agreed, changed the default to `false`. Enabling the write flag now requires an explicit opt-in after the whole cluster has been upgraded to a version that understands the checksummed format, so a rolling upgrade / mixed-version window can't produce files that a not-yet-upgraded BE/CN fails to read.

The new format stays covered regardless of the default — the unit tests toggle `lake_enable_protobuf_file_checksum` on explicitly. Updated the en/zh/ja BE config docs to state the default and the rolling-upgrade / downgrade caveat.

CelerData-Reviewer:
@codex review

be/src/storage/protobuf_file.cpp
```diff
@@ -110,19 +125,37 @@ Status ProtobufFileWithHeader::load(::google::protobuf::Message* message, bool f
...
+    // A legacy headerless file may be shorter than the header, or its first bytes simply won't
+    // match |magic|. In either case, fall back to plain protobuf when allowed. The magic is chosen
+    // so a serialized TabletMetadataPB/TxnLogPB can never be misread as the checksummed format.
+    bool has_header = data.size() >= sizeof(header);
+    if (has_header) {
+        ::memcpy(&header, data.data(), sizeof(header));
+    }
+    if (!has_header || header.magic_number != magic) {
+        if (allow_plain_protobuf_fallback) {
+            return parse_plain();
```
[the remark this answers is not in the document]
luohaha: Good catch — fixed. The legacy fallback now rejects an empty buffer instead of letting `ParseFromArray("")` succeed as an all-default message, so a metadata/txn-log object truncated to 0 bytes returns `Corruption` rather than a silent empty result. A legacy headerless file is never legitimately empty, so this is safe. (The streaming, non-fallback path already returned `Corruption` on a short read.)

Added BE UTs: empty buffer via the static loader and a 0-byte file via the file loader both return `Corruption`.

xiangguangyxg (dismissed):

be/src/storage/protobuf_file.h
```diff
@@ -47,24 +48,44 @@ class ProtobufFile {
...
+              _allow_plain_protobuf_fallback(allow_plain_protobuf_fallback) {}
 
     DISALLOW_COPY_AND_MOVE(ProtobufFileWithHeader);
 
     Status save(const ::google::protobuf::Message& message, bool sync = true);
 
     Status load(::google::protobuf::Message* message, bool fill_cache = true);
 
-    static Status load(::google::protobuf::Message* message, std::string_view data);
+    static Status load(::google::protobuf::Message* message, std::string_view data,
```
kevincai: are the `magic` and `allow_plain_protobuf_fallback` parameters redundant, since they are already provided in the constructor, what's the purpose of letting caller to overwrite the configuration per each call?

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -578,6 +597,13 @@ Status TabletManager::put_bundle_tablet_metadata(std::map<int64_t, TabletMetadat
 
         (*bundle_meta.mutable_tablet_meta_pages())[tablet_id] =
                 make_page_pointer(current_offset, serialized_buf.size());
+        if (config::lake_enable_protobuf_file_checksum) {
```
kevincai: since this is a mutable configuration, shall snapshot the value before making changes, avoid inconsistent behavior in the sam bundle for different tablets.

be/src/storage/lake/tablet_manager.cpp
```diff
@@ -588,7 +614,16 @@ Status TabletManager::put_bundle_tablet_metadata(std::map<int64_t, TabletMetadat
     }
     RETURN_IF_ERROR(meta_file->append(Slice(serialized_buf)));
     std::string fixed_buf;
-    put_fixed64_le(&fixed_buf, serialized_buf.size());
+    uint64_t size_field_value = serialized_buf.size();
+    if (config::lake_enable_protobuf_file_checksum) {
```
kevincai: similar issue here.

be/src/storage/lake/tablet_manager.cpp
luohaha: Good point — fixed. `put_bundle_tablet_metadata` now snapshots `config::lake_enable_protobuf_file_checksum` once at the start and uses that single value for every tablet page's checksum and for the footer, so one bundle file is always internally consistent (no mix of checksummed/unchecksummed pages, and the footer flag always matches the per-page map) even if the flag is flipped concurrently.

be/src/storage/protobuf_file.h
luohaha: This is the `static` overload (`load(message, data, magic, allow_plain_protobuf_fallback)`), not the instance one. Being `static`, it has no `this` and cannot read the constructor-initialized `_magic` / `_allow_plain_protobuf_fallback` members, so the parameters are required here rather than redundant — it doesn't override any instance configuration because there is no instance.

It exists for callers that only have an in-memory buffer and never construct an object: `meta_tool` (passes `LAKE_META_HEADER_MAGIC_NUMBER` + fallback) and `tablet_meta.cpp` / lake `replication_txn_manager.cpp` (defaults). The instance `load(message, fill_cache)` does use the constructor's values.

If the concern is having two configuration entry points, I can instead make the buffer variant an instance method that reuses the constructor config and drop the static overload — happy to do that in a follow-up if you'd prefer a single configuration surface.

kevincai: I see, maybe choose a different name that can be easily distinguished from the member function one. lol.

be/src/storage/lake/tablet_manager.cpp
luohaha: done

be/src/storage/protobuf_file.h
luohaha: Done — renamed the static overload to `load_from_buffer` so it's clearly distinct from the instance `load`. Updated all call sites (meta_tool, tablet_meta, lake replication, and the internal fallback dispatch).

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

PR #74954, opened 2026-06-17
[BugFix] Record audit stats for async materialized view refresh

### LINKED ISSUE #74942: [BugFix] Async materialized view refresh audit log misses CpuCostNs / MemCostBytes / ScanBytes

  1. Create an async materialized view and trigger a refresh (`REFRESH MATERIALIZED VIEW <mv>`).
  2. Look at the corresponding `insert overwrite ... <mv>` entry in `fe.audit.log`.

  **Expected:** the audit entry contains resource fields `CpuCostNs`, `MemCostBytes`, `ScanBytes`, `ScanRows`, the same way a normal query / `INSERT` does.

  **Actual:** those fields are entirely absent from the MV-refresh audit entry (note: not `0`/`-1`, the columns are dropped). A normal query on the same cluster still reports them, e.g. `...|CpuCostNs=617568|MemCostBytes=991440|...`.

  MV refresh audit entries should carry the same execution-statistics fields as other statements.

  ### Real behavior (Required)

  `CpuCostNs` / `MemCostBytes` / `ScanBytes` / `ScanRows` are missing for every async MV refresh.

  - Reproduced on 3.3.x and confirmed still present on 3.5.x (latest).
  - Regression vs 3.2.x (3.2 reports these fields correctly).

  ### Root cause

  This is a regression introduced by #56257 ("[Enhancement] record audit stats for all statements", backported to branch-3.3 via #56282, branch-3.4 via #56283).

  That PR removed the old fallback in `ConnectProcessor.auditAfterExec()` that wrote statistics directly into the audit-event builder, and centralized the write into a new method `StmtExecutor.recordExecStatsIntoContext()`. However `recordExecStatsIntoContext()` is only
  invoked from `execute()`, `executeStmtWithExecPlan()` and `executeStmtWithResultQueue()`.

  Async MV refresh does **not** go through `execute()` — `PartitionBasedMvRefreshProcessor.refreshMaterializedView()` builds an internal `StmtExecutor` and calls `executor.handleDMLStmtWithProfile()` directly, so `recordExecStatsIntoContext()` is never called. The
  statistics object passed to `auditAfterExec(...)` is also ignored (`ConnectProcessor.auditAfterExec` no longer reads its `statistics` parameter). As a result the builder's `cpuCostNs` stays at the default `-1`, and `AuditLogBuilder` drops any field whose value is `-1`,
  so the columns disappear. `ReturnRows` still shows because it is set separately.

  The data itself is available at audit time (`executor.getQueryStatisticsForAuditLog()` returns the real values via `coord.getAuditStatistics()`); it is simply never written into the audit-event builder.

  This is the same family of regression as #56404 (iceberg sink not reporting audit statistics, fixed in #74072 for connector sinks), but on the MV-refresh code path, which neither #74072 nor any other PR covers.

  ### Impact

  Observability only. MV refresh executes normally, resources are still tracked and governed by resource groups; only the audit-log resource columns are missing. No data/functional impact.

spaces-X:
## Why I'm doing:

  Close #74942

  Async materialized view refresh audit-log entries lost the `CpuCostNs` / `MemCostBytes` / `ScanBytes` / `ScanRows` fields. This is a regression introduced by #56257, which centralized audit-stats writing into `StmtExecutor.recordExecStatsIntoContext()` (only called from
  `execute()` / `executeStmtWithExecPlan()` / `executeStmtWithResultQueue()`) and removed the old fallback in `ConnectProcessor.auditAfterExec()`.

  MV refresh does not go through `execute()`: it calls `executor.handleDMLStmtWithProfile()` directly, so `recordExecStatsIntoContext()` is never invoked and the statistics never reach the audit-event builder, even though the data is available via
  `executor.getQueryStatisticsForAuditLog()`.

  ## What I'm doing:

  Flush exec statistics into the audit-event builder on the MV-refresh path before `auditAfterExec`, mirroring what `execute()` does in its `finally` block. `recordExecStatsIntoContext()` is made `public` (it lives in `qe`, the caller in `scheduler`).

  It is intentionally **not** placed inside `handleDMLStmtWithProfile()`: `execute()` also calls `handleDMLStmtWithProfile()` and then calls `recordExecStatsIntoContext()` in its own `finally`, so flushing inside `handleDMLStmtWithProfile()` would double-count for normal
  INSERTs (the builder uses accumulating `add*`). The MV-refresh path uses a fresh `ConnectContext` whose builder starts at the default `-1`, so a single explicit call is correct and not double-counted.

  ## What type of PR is this:
  - [x] BugFix
  - [ ] Feature
  - [ ] Enhancement
  - [ ] Refactor
  - [ ] UT
  - [ ] Doc
  - [ ] Tool

  - [ ] Feature removed

  ## Checklist:
  - [ ] This is a backport PR

    - [x] 4.1
    - [x] 4.0
    - [x] 3.5
    - [ ] 3.4
    - [ ] 3.3

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java
@@ -1356,7 +1356,7 @@ public class StmtExecutor {
      * some statements may execute multiple statement which will also create multiple StmtExecutor, so here
      * we accumulate them into the ConnectContext instead of using the last one
      */
-    private void recordExecStatsIntoContext() {
+    public void recordExecStatsIntoContext() {
         PQueryStatistics execStats = getQueryStatisticsForAuditLog();
         context.getAuditEventBuilder().addCpuCostNs(execStats.getCpuCostNs() != null ? execStats.getCpuCostNs() : 0);
         context.getAuditEventBuilder()
--- a/fe/fe-core/src/main/java/com/starrocks/scheduler/MVTaskRunProcessor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/scheduler/MVTaskRunProcessor.java
@@ -371,6 +371,8 @@ public class MVTaskRunProcessor extends BaseTaskRunProcessor implements MVRefres
             throw e;
         } finally {
             logger.info("[QueryId:{}] finished to refresh mv in DML", ctx.getQueryId());
+            // the MV refresh uses its own fresh ConnectContext whose audit builder starts at the default value.
+            executor.recordExecStatsIntoContext();
             auditAfterExec(mvTaskRunContext, executor.getParsedStmt(), executor.getQueryStatisticsForAuditLog());
             executor.addFinishedQueryDetail();
         }
--- a/fe/fe-core/src/test/java/com/starrocks/scheduler/MVTaskRunProcessorTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/scheduler/MVTaskRunProcessorTest.java
@@ -18,6 +18,7 @@ import com.starrocks.catalog.Database;
 import com.starrocks.catalog.MaterializedView;
 import com.starrocks.common.util.RuntimeProfile;
 import com.starrocks.common.util.UUIDUtil;
+import com.starrocks.plugin.AuditEvent;
 import com.starrocks.qe.ConnectContext;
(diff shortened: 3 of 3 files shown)


CelerData-Reviewer:
@codex review

dontknow9179:
cherry-pick branch should be 3.5+

Youngwb (approved):

HangyuanLiu (approved):

PR #74962, opened 2026-06-17
[Enhancement] Avoid redundant consistent-hash ring walks in scan-range assignment

eshishki:
## Why I'm doing:

While profiling FE-side planning for Iceberg queries with the metadata **caching catalog** enabled, I looked at where time goes once the manifest cache is warm (so `getScanFiles` is cheap). The remaining planning time was dominated by `computeScanRangeAssignment`: on one query (one compute node, ~1183 Iceberg splits) it took **~23 ms on the planning critical path**, plus ~16 ms more in incremental scan-range delivery.

Profiling that time pointed straight at `ConsistentHashRing` — `collectNodes`, `ComputeNode.hashCode`, `TreeMap` navigation. The cause: `get(key, distinctNumber)` is asked for `distinctNumber` **distinct** nodes (callers pass `kCandidateNumber = 3`). When the cluster has fewer physical nodes than that, `get()` can never reach the target, so it never breaks early — it walks the entire virtual ring (256 vnodes/node), and then walks it a **second time**, allocating a `HashSet` per call. One lookup per split makes this O(splits × ring_size) of pure waste.

This is **not just a single-node curiosity**. The scan data cache is **on by default**, and with it `HDFSBackendSelector` skips re-balancing and uses only the **first** ring node. So in the default production config we genuinely need *one* node but ask the ring for *three* — which is exactly what triggers the repeated full-ring walk on any warehouse running a small number of compute nodes.

## What I'm doing:

1. **`HDFSBackendSelector`** — when re-balancing is skipped (data cache on, no forced re-balance), ask the ring for a single candidate instead of `kCandidateNumber`, since only the first node is ever used.
2. **`ConsistentHashRing`** — make `get` robust regardless of caller: cap the target at `min(distinctNumber, nodeCount)` and walk the ring **at most once** (clockwise with wrap-around, shared `visited` set, early-stop). The returned node set and order are unchanged for the multi-node case.

Same query on one compute node: `computeScanRangeAssignment` **~23 ms → ~1 ms** on the critical path (and ~16 ms → ~2 ms in incremental delivery).

## What type of PR is this:

- [ ] BugFix
- [x] Enhancement
- [ ] Feature
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/ConsistentHashRing.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/ConsistentHashRing.java
@@ -32,6 +32,11 @@ public class ConsistentHashRing<K, N> implements HashRing<K, N> {
     HashFunction hashFunction;
 
     TreeMap<Long, VNode> hashRing = new TreeMap<>();
+    // Physical nodes currently on the ring. A node is spread across `virtualNumber`
+    // virtual nodes, so the ring can never hand back more distinct nodes than this set
+    // holds; `get` uses its size to stop walking instead of scanning for nodes that do
+    // not exist.
+    Set<N> physicalNodes = new HashSet<>();
     Funnel<K> keyFunnel;
     Funnel<N> nodeFunnel;
     int virtualNumber;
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

PR #74963, opened 2026-06-17
[Enhancement] Plan cached Iceberg manifests on the consumer thread (avoid ParallelIterable poll wait)

eshishki:
## Why I'm doing:

Continuing to profile warm FE-side Iceberg planning with the metadata caching catalog: even once the manifest cache is warm, `getScanFiles` still cost **~17 ms**. Profiling showed this isn't CPU — the coordinator thread sits in `futex_wait`, waiting on the planning worker pool.

The culprit is upstream `ParallelIterable` (Apache Iceberg 1.10.0). Its consumer `hasNext()` polls the queue and, when it's momentarily empty while tasks run, does `Thread.sleep(10)`. On the cache-hit planning path the per-manifest work is purely in-memory (microseconds), and the whole result fits far under `DEFAULT_MAX_QUEUE_SIZE` (30k) so producers never yield — yet the consumer still sleeps a 10 ms quantum before it notices the queue filled. With no IO to overlap, that sleep *is* the latency rather than being hidden by it.

In `StarRocksIcebergTableScan.planTaskWithCache`, data manifests are already split into cache hits (fully in memory) and cache misses (need IO). The cache-hit branch was assembled through `ParallelIterable` — paying the executor hand-off and the 10 ms poll for zero benefit.

## What I'm doing:

`planTaskWithCache`: assemble the cache-hit scan tasks with a sequential `CloseableIterable.concat`, evaluated lazily on the consumer thread (no executor, no poll-sleep). Cache-**miss** manifests keep parallel planning via `planFileTasks`, where real IO genuinely benefits from it.

No new unit test: the change is behavior-equivalent — same set of `FileScanTask`, only the iteration over already-in-memory cached data changes from parallel to sequential. Existing `IcebergMetadataTest` (`planFiles` -> file-scan-task assertions) covers the planning output.

Same query, warm cache: `getScanFiles` **~17 ms -> ~6 ms**, and the inner `hasMoreOutput` (where the sleep lived) **~16 ms -> ~2 ms**.

## What type of PR is this:

- [ ] BugFix
- [x] Enhancement
- [ ] Feature
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/iceberg/StarRocksIcebergTableScan.java
+++ b/fe/fe-core/src/main/java/org/apache/iceberg/StarRocksIcebergTableScan.java
@@ -42,7 +42,6 @@ import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.CloseableIterator;
 import org.apache.iceberg.metrics.MetricsReporter;
 import org.apache.iceberg.metrics.ScanMetricsUtil;
-import org.apache.iceberg.util.ParallelIterable;
 import org.apache.iceberg.util.SerializationUtil;
 import org.apache.iceberg.util.TableScanUtil;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #74970, opened 2026-06-17
[Refactor][BugFix] Refactor low cardinality aggregation logic and output handling

farhad-celo:
## Why I'm doing:
Currently low cardinality optimization does not differentiate between whether an aggregate function can be optimized (rewritten using dictionary codes) and whether its output is low cardinality (there is a limited support for COUNT / COUNT DISTINCT, but not for general cases). This causes a bug for queries like `ARRAY_AGG(non_low_card_column ORDER BY low_card_column)` for which the framework marks its output as low cardinality. 

## What I'm doing:
Refactoring the rewrite for aggregate functions. This refactor separates the logic of whether an agg should be rewritten versus whether the output is low cardinality. The latter part will be handled by `checkDependsOnExpr()` like other string columns, making `checkDependsOnExpr()` the only source of truth for this decision in the framework.
* DecodeInfo: Adding two new fields: 
  * `inProgressStringAggregations`: Aggregate functions which are being rewritten, but haven't finalized yet.
  * `finalizingStringAggregations`: Aggregate functions which are being finalized in this operator. 
  * When creating output info, `inProgressStringAggregations` will be passed to the caller, but `finalizingStringAggregations` won't. 
* DecodeCollector:
  * Removing specializations for COUNT, COUNT DISTINCT. All aggregates go through the same path now. 
  * populating `inProgressStringAggregations` and `finalizingStringAggregations`.
  * stop the assumption that all string aggregations produce low cardinality outputs (adding `aggregateId` to `context.allStringColumns`). The decision now depends on `checkDependsOnExpr()`.
  *  `inProgressStringAggregations`, `finalizingStringAggregations` are not intersected with `context.allStringColumns`, to ensure rewrites happen if there is any low cardinality inputs to the function.
  * Only populate struct info for aggregates with a struct return type (as opposed to intermediateOrReturnType)
* DecodeContext:
Refactoring and simplifying `AggregateRewriter`. Now it generates the function template based on the parameters in the local stage and uses this function template to rewrite every stage. 
* DecodeRewriter
Refactoring rewrites for PhysicalHashAggregate and PhysicalDistribution to consider new fields added to DecodeInfo. 

Fixes #issue

## What type of PR is this:
- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:
- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lowcardinality/DecodeCollector.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lowcardinality/DecodeCollector.java
@@ -400,13 +400,9 @@ public class DecodeCollector extends OptExpressionVisitor<DecodeInfo, DecodeInfo
                 continue;
             }
             List<CallOperator> aggregateExprs = stringAggregateExpressions.get(aggregateId);
-            for (CallOperator agg : aggregateExprs) {
-                if (agg.getColumnRefs().stream().map(ColumnRefOperator::getId)
-                        .anyMatch(context.allStringColumns::contains)) {
-                    context.stringAggregateExprs.put(aggregateId, aggregateExprs);
-                    context.allStringColumns.add(aggregateId);
-                    break;
-                }
+            if (aggregateExprs.get(0).getColumnRefs().stream().map(ColumnRefOperator::getId)
+                    .anyMatch(context.allStringColumns::contains)) {
+                context.stringAggregateExprs.put(aggregateId, aggregateExprs);
             }
         }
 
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lowcardinality/DecodeCollector.java
```diff
@@ -400,13 +400,9 @@ private void initContext(DecodeContext context) {
                 continue;
             }
             List<CallOperator> aggregateExprs = stringAggregateExpressions.get(aggregateId);
-            for (CallOperator agg : aggregateExprs) {
```
trueeyu, line 403: In the previous implementation, this loop itself was meaningless, right?

farhad-celo: Yes, we only need to check the first stage and the list is always sorted.

trueeyu (approved):

stdpain (approved):

PR #75001, opened 2026-06-18
[BugFix] Render DATETIME microseconds locale-independently

xiangguangyxg:
## Why I'm doing:

The DATETIME canonical text renders its sub-second microsecond component with `String.format("%06d", micros)`, which honors the JVM default locale's digit set. On an FE JVM whose default locale uses non-ASCII digits (e.g. Arabic/Persian), the canonical text becomes e.g. `1970-01-01 00:00:01.٥٠٠١٢٣` instead of `1970-01-01 00:00:01.500123`. `DateUtils.parseStrictDateTime` and the BE `datum_from_string` parser then reject the value, so a date/datetime sort-key boundary fails to parse and the split is silently skipped. Under a normal ROOT/en default locale the output is byte-identical, so the defect is locale-latent.

The same locale-sensitive `String.format("%06d", ...)` appeared at three render sites: the canonical `DateVariant.getStringValue()` and a byte-identical private `renderDateTime` duplicated in both the Parquet and ORC footer readers.

## What I'm doing:

Two commits:

1. `[Refactor]` Extract the two footer readers' byte-identical private `renderDateTime(LocalDateTime)` into a single package-private `static MetaTierTemporalWindow.renderDateTime` (same `com.starrocks.alter.reshard.presplit` package; already the readers' shared temporal helper) and delegate. Pure move, zero behavior change.
2. `[BugFix]` Render the touched DATE/DATETIME boundary path via the `%Y`-based `*_UNIX` `DateTimeFormatter`s (`DATE_FORMATTER_UNIX`, `DATE_TIME_FORMATTER_UNIX`, `DATE_TIME_MS_FORMATTER_UNIX`) — the same family `parseStrictDateTime` already uses. Their numeric fields render as ASCII via `DecimalStyle.STANDARD`, so the output is locale-independent **without** any `String.format`.

### Why the `*_UNIX` formatters (not `Locale.ROOT` on `String.format`)

- A `java.time.format.DateTimeFormatter` renders numeric fields with `DecimalStyle.STANDARD` (ASCII digits) regardless of the JVM default locale — so moving the microsecond fraction onto `DATE_TIME_MS_FORMATTER_UNIX` removes the locale footgun at the root rather than patching each `String.format` with `Locale.ROOT`.
- It also reuses the existing canonical formatter (`ConstantOperator` / `FeExecuteCoordinator` already render datetime-with-micros this way) and keeps the two renderers independent across packages (no `catalog`↔`presplit` coupling) — they each call the shared `DateUtils` layer they already depend on.
- It additionally retires the deprecated `yyyy`-based `DATE_FORMATTER` / `DATE_TIME_FORMATTER` from these sites. Those are `@Deprecated` because the year-of-era `yyyy` mis-renders year 0000; the `%Y`-based `*_UNIX` formatters render the proleptic year (matching the parser).

Output is **byte-identical for all in-range values** (verified, including trailing-zero microseconds e.g. `.500000`, and the `nano == 0` no-fraction branch). The fixed 6-digit microsecond width is preserved (`%f` resolves to a fixed-width fraction in output mode).

`DateVariant`'s `DATE` and `DATETIME` branches both move to the `*_UNIX` family; the `TIME`/default branch (ISO `Instant.toString()`) is unchanged.

### Tests

- New `MetaTierTemporalWindowTest`: forces a non-ASCII-digit locale (`Locale.forLanguageTag("ar-SA-u-nu-arab")`, which pins the Arabic-Indic numbering system) and asserts `renderDateTime` returns ASCII `1970-01-01 00:00:01.500123`; also covers the fixed-6-digit fraction (`.500000`, not `.5`) and the `nano == 0` (no-fraction) branch. Restores the prior locale in `finally`.
- `VariantTest.testDateTimeStringValueIsLocaleIndependent`: same forced-locale assertion for the public canonical site `DateVariant.getStringValue()`.
- Existing footer-reader micro/timestamp/date tests and the existing `VariantTest` DATE/DATETIME tests serve as byte-equivalence regression guards (209 tests green across the touched classes).

No TSP/cluster e2e: the defect only manifests when the FE JVM default locale uses non-ASCII digits (ar/fa), which a TSP cluster cannot reproduce; the forced-locale unit tests are the authoritative verification.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/MetaTierTemporalWindow.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/MetaTierTemporalWindow.java
@@ -14,7 +14,10 @@
 
 package com.starrocks.alter.reshard.presplit;
 
+import com.starrocks.common.util.DateUtils;
+
 import java.time.LocalDate;
+import java.time.LocalDateTime;
 
 /**
  * The value window the meta tier accepts for DATE/DATETIME sort-key boundaries:
@@ -32,9 +35,10 @@ import java.time.LocalDate;
  * </ul>
  *
  * <p>This lives in one place — rather than being duplicated per reader like the
- * integer-stat conversion — because the constraint is format-agnostic: it operates on
- * an already-decoded {@link LocalDate}, identical whether the value came from a Parquet
- * INT32/INT64 stat or an ORC day-of-epoch stat.
+ * integer-stat conversion — because both responsibilities are format-agnostic: the
+ * window check operates on an already-decoded {@link LocalDate}, and the microsecond
+ * datetime renderer on an already-decoded {@link LocalDateTime}, identical whether the
+ * value came from a Parquet INT32/INT64 stat or an ORC day-of-epoch / timestamp stat.
  */
 final class MetaTierTemporalWindow {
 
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #75002, opened 2026-06-18
[BugFix] Zero-extend unsigned INT32 parquet values on load

xiangguangyxg:
## Why I'm doing:

When the BE loads a Parquet column physically stored as `INT32` but annotated as an **unsigned** integer (`UINT_32` / logical `INTEGER` `bitWidth=32 isSigned=false`), it widened the value into a wider StarRocks column by **sign-extending** the signed `int32` source. A high-bit value such as `3000000000` (`0xB2D05E00`) was therefore silently stored as the negative value `-1294967296` — silent data corruption.

`be/src/formats/parquet/utils.cpp` already infers the correct destination type (UINT_32 → `BIGINT`, which can hold the full unsigned range); the bug was purely in value conversion. The converter selection in `ColumnConverterFactory::create_converter` branched only on the physical type (`INT32`) and destination type, never on the unsigned annotation, and `convert_int_to_int` widened with `DestType(src[i])` (i.e. `int64_t(int32_t)`), which sign-extends.

This is a long-standing bug, present since the converter was introduced; it affects `main`, 4.1, 4.0 and 3.5.

## What I'm doing:

For an **unsigned** `INT32` source, reinterpret each element through its unsigned bit pattern before widening — `static_cast<DestType>(static_cast<std::make_unsigned_t<SourceType>>(src[i]))` — so `3000000000` loads as `3000000000` instead of `-1294967296`. The source column stays materialized as the signed `FixedLengthColumn<int32_t>`; only the widening step changes.

- The unsigned decision is read from the Parquet schema element already available in `create_converter`, so the factory signature is unchanged. The predicate `is_unsigned_integer()` (logical `INTEGER.isSigned` when a logical type is present, else legacy `UINT_*` converted type) is promoted to a shared helper in `formats/parquet/utils.{h,cpp}`.
- Zero-extension is applied via a defaulted `kSourceUnsigned` template parameter on the two converters that widen an `INT32` source: `NumericToNumericConverter` (TINYINT/SMALLINT/BIGINT/DOUBLE) and `PrimitiveToDecimalConverter` (DECIMALV2/32/64/128). The branch is `if constexpr`, so there is no per-row runtime cost; signed sources, the `INT64` branch and the `FLOAT`→`DOUBLE` path are unchanged.
- The same converter is used by `StatisticsHelper` to decode footer min/max. To keep statistics consistent with the corrected decode, `get_min_max_value` now derives the statistics sort order from the Parquet annotation (`SortOrder::UNSIGNED` for unsigned integers) instead of the signed StarRocks destination type. This makes legacy (pre-1.10 parquet-mr) signed-ordered deprecated `min`/`max` be **rejected** for unsigned columns rather than trusted and zero-extended into inverted bounds; modern unsigned-ordered `min_value`/`max_value` are accepted and decoded correctly. (Incidentally, footer min/max for unsigned INT32 columns now decode correctly.)

**Behavior change:** This corrects a wrong result (high-bit `UINT_32` values were a bug), so behavior change is marked **No** per the project rubric. Reviewers should be aware that high-bit `UINT_32` values now load as the correct positive value instead of the previous (buggy) negative value.

**Scope / limitations (existing semantics, unchanged):**
- `UINT_64` (and `INTEGER bitWidth=64 isSigned=false`) is out of scope — it maps to a `VARIANT` fallback for schema inference and is not loaded as a plain integer; the `INT64` branch is untouched.
- User-declared narrower destinations (e.g. `UINT_32` → `INT`/`TINYINT`) keep existing overflow/narrowing semantics — this fix only guarantees correct *unsigned decode*; a value that cannot be represented in a too-narrow declared type still overflows as before.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

> Note for maintainers: this is an old bug present on 4.0 and 3.5 as well. I set 4.1 to match the related work line; please advise if you want it also backported to 4.0/3.5.

Tested with BE unit tests (`ColumnConverterTest.UnsignedInt32ZeroExtend` covers BIGINT/DOUBLE/DECIMAL zero-extension across both the logical-type and legacy converted-type encodings, signed non-regression, and logical-type precedence; `ColumnConverterTest.UnsignedInt32StatsRejectsLegacySignedOrder` covers the statistics sort-order rejection) plus a broader parquet reader/stats/metadata regression (208 tests pass).

### DIFF
--- a/be/src/formats/parquet/column_converter.cpp
+++ b/be/src/formats/parquet/column_converter.cpp
@@ -22,6 +22,7 @@
 #include <cstring>
 #include <memory>
 #include <sstream>
+#include <type_traits>
 #include <utility>
 #include <vector>
 
@@ -126,15 +127,21 @@ private:
     int64_t _scale_to_nano_factor = 0;
 };
 
-template <typename SourceType, typename DestType>
+template <typename SourceType, typename DestType, bool kSourceUnsigned = false>
 void convert_int_to_int(const SourceType* __restrict__ src, DestType* __restrict__ dst, size_t size) {
     for (size_t i = 0; i < size; i++) {
-        dst[i] = DestType(src[i]);
+        if constexpr (kSourceUnsigned && std::is_integral_v<SourceType>) {
+            // Reinterpret the source through its unsigned bit pattern before widening so a
+            // high-bit unsigned value is zero-extended instead of sign-extended.
+            dst[i] = DestType(static_cast<std::make_unsigned_t<SourceType>>(src[i]));
+        } else {
+            dst[i] = DestType(src[i]);
+        }
     }
 }
 
 // Support int => int and float => double
-template <typename SourceType, typename DestType>
+template <typename SourceType, typename DestType, bool kSourceUnsigned = false>
 class NumericToNumericConverter final : public ColumnConverter {
 public:
     NumericToNumericConverter() = default;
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #75017, opened 2026-06-18
[BugFix] Honor CREATE DATABASE IF NOT EXISTS on Iceberg REST catalogs

ycra-dev:
## Why I'm doing:

On Iceberg REST catalogs, `CREATE DATABASE IF NOT EXISTS <db>` failed with an error when the database already existed, instead of being silently ignored as the `IF NOT EXISTS` clause promises.

The root cause is an exception-type mismatch at the StarRocks ↔ Iceberg boundary. `IcebergMetadata.createDb()` only relies on StarRocks' `com.starrocks.common.AlreadyExistsException` to drive the `IF NOT EXISTS` handling in the upper layer. However, the Iceberg client (notably the REST catalog) throws its own `org.apache.iceberg.exceptions.AlreadyExistsException`, which is an unrelated class. That connector-level exception leaked through unchanged, so the upper layer never recognized it as "already exists" and the statement errored out.

## What I'm doing:

Translate the connector-level `org.apache.iceberg.exceptions.AlreadyExistsException` thrown by `icebergCatalog.createDB()` into StarRocks' `com.starrocks.common.AlreadyExistsException`, carrying the database name in the message. This lets the existing `IF NOT EXISTS` logic recognize the condition and behave correctly.

- `IcebergMetadata.createDb()`: wrap the `createDB` call in a try/catch and rethrow as StarRocks `AlreadyExistsException`.
- Added a unit test `testCreateDbTranslatesConnectorAlreadyExists` that mocks the catalog to throw the Iceberg exception and asserts the translated StarRocks exception (type + message) is raised.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -319,7 +319,11 @@ public class IcebergMetadata implements ConnectorMetadata {
             throw new AlreadyExistsException("Database Already Exists");
         }
 
-        icebergCatalog.createDB(context, dbName, properties);
+        try {
+            icebergCatalog.createDB(context, dbName, properties);
+        } catch (org.apache.iceberg.exceptions.AlreadyExistsException e) {
+            throw new AlreadyExistsException("Database Already Exists: " + dbName);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


ycra-dev:
Hi @stephen-shelby @Youngwb, gentle ping on this one 🙂

It's been open for ~3 weeks with all checks green and no conflicts. The equivalent fix for 3.5 has already been reviewed and merged separately (the 3.5 codebase differs, so it required its own change), so this PR brings main/4.x into line. Would appreciate a review whenever you have a moment — happy to address any feedback. Thanks!

PR #75023, opened 2026-06-18
[Enhancement] Add hdfs_backend_selector_cache_replica_num variable 

### LINKED ISSUE #75022: Under high concurrency the HDFS Selectors cache affinity creates hot spots that impact overall query throughput
## Enhancement

We observed that when querying Iceberg tables with high concurrency that the data cache affinity of the HDFS Backend selector causes worker skew which has a significant impact on query performance. This may be a niche issue but there is unfortunately no way currently to opt-out of that affinity or spread the load more to get around it. 

We built and deployed a change that adds a configurable session variable `hdfs_backend_selector_cache_replica_num`. Setting this value change the data cache affinity from the current 1 to 1 relation to 1 to N. In our tests we set the value to 5 and saw an immediate difference in query concurrency, latency, and overall cluster utilization. 

I do not know if this is the best way to solve this issue, but we would like some configuration that allows us to better distribute the load for scenarios like this. A better option may be to somehow integrate load of the CN's into the backend selection to only distribute work away from nodes that are processing way more than others.

I've opened a PR with the changes we tested. Here are the results:

Image

Image

kyle-goodale-klaviyo:
## Why I'm doing:
In some cases when running highly concurrent queries against data lake tables the cache affinity of the HDFS backend selector creates worker skew where one or two workers will be overloaded while the rest of the cluster sits with minimal load

## What I'm doing:

Adding hdfs_backend_selector_cache_replica_num to allow configuring the number of CN nodes that get assigned particular data cache files. This more evenly distributes the load at the cost of caching files on additional nodes. The default is 1 which preserves the current behavior.

Fixes #75022

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/HDFSBackendSelector.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/HDFSBackendSelector.java
@@ -62,6 +62,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * Hybrid backend selector for hive table.
@@ -94,6 +95,10 @@ public class HDFSBackendSelector implements BackendSelector {
     private final boolean shuffleScanRange;
     private final boolean useIncrementalScanRanges;
     private final int kCandidateNumber = 3;
+    // Number of top hash-candidate nodes a split may be cached/served on. Each scan range is routed to a
+    // uniformly random one of its top-N hash nodes, spreading a hot data object across N nodes instead of
+    // pinning it to its single hash home. 1 (default) preserves the original single-home placement behavior.
+    private final int cacheReplicaNum;
     // After testing, this value can ensure that the scan range size assigned to each BE is as uniform as possible,
     // and the largest scan data is not more than 1.1 times of the average value
     private final double kMaxImbalanceRatio = 1.1;
@@ -174,6 +179,7 @@ public class HDFSBackendSelector implements BackendSelector {
         this.hdfsScanRangeHasher = new HdfsScanRangeHasher();
         this.shuffleScanRange = shuffleScanRange;
         this.useIncrementalScanRanges = useIncrementalScanRanges;
+        this.cacheReplicaNum = connectContext.getSessionVariable().getHdfsBackendSelectorCacheReplicaNum();
         this.candidateWorkerProvider = initCandidateWorkerProvider();
     }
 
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/qe/HDFSBackendSelector.java
```diff
@@ -359,10 +365,21 @@ private void computeGeneralAssignment() throws StarRocksException {
...
-            ComputeNode node = reBalanceScanRangeForComputeNode(backends, avgNodeScanRangeBytes, scanRangeLocations);
+            List<ComputeNode> backends = hashRing.get(scanRangeLocations, candidateCount);
+            ComputeNode node;
+            if (cacheReplicaNum > 1) {
+                // Spread each split uniformly across its top-N hash candidates so a hot data object is served
+                // by N nodes instead of pinning to its single hash home. Stateless: the only per-call state is
+                // the random draw. The candidate_node (peer datacache) hint below is left untouched, so a chosen
+                // node that lacks the block can still peer-fetch it from the historical cache home.
+                int replicaCount = Math.min(cacheReplicaNum, backends.size());
+                node = backends.get(ThreadLocalRandom.current().nextInt(replicaCount));
```
GavinMar, line 386: The change spreads the execution node across N candidates to improve the hot spot problem, but the peer-cache hint candidate_node is still pinned to a single node (historical top-1). So with datacache sharing enabled, the N execution replicas of a hot split all peer-read from the same one historical
node — an N→1 fan-in that recreates, on the cache-read path, the exact worker skew this feature is meant to remove.

One possible suggestion: spread candidate_node too — since the primary is picked randomly among the top-N, pick the peer hint randomly among the historical top-N as well, instead of always top-1.

GavinMar, line 386: When cache_replica_num > 1, an empty candidate list throws a confusing IllegalArgumentException instead of the intended error. If hashRing.get(...) returns an empty list (no available workers), replicaCount becomes 0 and ThreadLocalRandom.current().nextInt(0) throws
 IllegalArgumentException: bound must be positive, before reaching the node == null check. The original path (reBalanceScanRangeForComputeNode) returns null here and surfaces the clear StarRocksException("Failed to find backend to execute").

Suggestion: guard for an empty backends before the random draw and fall through to the existing node == null error path, so the failure message stays consistent.

fe/fe-core/src/main/java/com/starrocks/qe/HDFSBackendSelector.java
```diff
@@ -359,10 +365,21 @@ private void computeGeneralAssignment() throws StarRocksException {
...
         }
         // assign scan ranges.
+        int candidateCount = Math.max(kCandidateNumber, cacheReplicaNum);
         for (int i = 0; i < remoteScanRangeLocations.size(); ++i) {
             TScanRangeLocations scanRangeLocations = remoteScanRangeLocations.get(i);
-            List<ComputeNode> backends = hashRing.get(scanRangeLocations, kCandidateNumber);
-            ComputeNode node = reBalanceScanRangeForComputeNode(backends, avgNodeScanRangeBytes, scanRangeLocations);
+            List<ComputeNode> backends = hashRing.get(scanRangeLocations, candidateCount);
+            ComputeNode node;
+            if (cacheReplicaNum > 1) {
```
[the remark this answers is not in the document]
kyle-goodale-klaviyo: fixed

fe/fe-core/src/main/java/com/starrocks/qe/HDFSBackendSelector.java
kyle-goodale-klaviyo: Good call, fixed!

kyle-goodale-klaviyo: Good call, fixed!

kyle-goodale-klaviyo:
@GavinMar Good call on those changes. I made the requested changes and update this PR. Please take another look when you get a chance. Thanks!

GavinMar (approved):

GavinMar:
LGTM! 🚀

dirtysalt (approved):
LGTM! 🚀

stdpain (approved):

PR #75032, opened 2026-06-19
[BugFix] Define out-of-range bit shift counts to return 0

Hyper-FF:
## Why I'm doing:

bit_shift_left / bit_shift_right / bit_shift_right_logical computed l << r / l >> r directly. When the shift count is negative or >= the operand bit width, this is C++ undefined behavior; on x86 the hardware masks the count to the operand width and silently returns a wrong (and non-deterministic) value instead of 0.

These SQL functions are registered as ArithmeticExpr operators, so they are evaluated through ArithmeticBinaryOperator in arithmetic_operation.h, not through BitFunctions in bit_functions.cpp. Guard the shift in both the scalar apply() and the JIT generate_ir() paths so an out-of-range count yields 0. The JIT path additionally adapts the i64 BIGINT shift count to the operand width (CreateZExtOrTrunc) to match the shift-instruction operand type. The defensive guard in bit_functions.cpp is kept for the (shadowed) builtin path.

Add regression test test_bit_shift_out_of_range covering all three functions across TINYINT/SMALLINT/INT/BIGINT with negative, == bit width and > bit width shift counts, with operands read from columns so the FE cannot constant-fold them away.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exprs/arithmetic_operation.h
+++ b/be/src/exprs/arithmetic_operation.h
@@ -178,21 +178,35 @@ struct ArithmeticBinaryOperator {
         } else if constexpr (is_bitxor_op<Op>) {
             return l ^ r;
         } else if constexpr (is_bit_shift_left_op<Op>) {
-            return l << r;
+            // A shift count outside [0, bit width) is C++ undefined behavior; on x86 the
+            // hardware masks the count and silently returns a wrong value. Define such
+            // out-of-range counts to return 0. Kept branchless so the per-element loop still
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

PR #75033, opened 2026-06-19
[BugFix] Guard against cyclic view definitions to avoid StackOverflowError

Hyper-FF:
## Why I'm doing:

QueryAnalyzer.visitView re-resolves a view body recursively with no visited-set or depth guard. A view cannot reference itself at CREATE time, but ALTER VIEW analyzes the new body against the current catalog, so an ALTER that closes a cycle (v1 -> v2 -> v1) passes analysis and commits the cycle. A later SELECT then recurses forever and throws StackOverflowError, which is an Error and escapes the catch (SemanticException), aborting the query with an opaque "Unknown error".

Track the ids of views currently being expanded on the resolution path and raise a SemanticException when a view re-enters its own expansion, mirroring the currentRecursiveCTE guard used for recursive CTEs. Diamond view references (the same view in sibling subtrees) are unaffected because the id is popped when its subtree finishes.

## What I'm doing:
reproduce case:
```
CREATE TABLE vt (a INT) DUPLICATE KEY(a) DISTRIBUTED BY HASH(a) BUCKETS 1
  PROPERTIES("replication_num"="1");
CREATE VIEW cyc_v1 AS SELECT a FROM vt;
CREATE VIEW cyc_v2 AS SELECT a FROM cyc_v1;
ALTER VIEW cyc_v1 AS SELECT a FROM cyc_v2;   -- analyzes vs. old v1 → OK, commits the cycle
SELECT * FROM cyc_v1 LIMIT 1;                -- StackOverflowError
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/ConnectContext.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/ConnectContext.java
@@ -238,6 +238,12 @@ public class ConnectContext {
 
     private boolean relationAliasCaseInsensitive = false;
 
+    // Keys of the views currently being expanded on the analysis path, used to detect cyclic view
+    // definitions. It lives on the session (not on a single QueryAnalyzer.Visitor) so the path is
+    // shared across the fresh QueryAnalyzer instances spawned for scalar/IN/EXISTS subqueries; a
+    // cycle routed through a subquery would otherwise reset the per-Visitor set on every hop.
+    private final Set<String> viewExpansionPath = Sets.newHashSet();
+
     private final Map<String, PrepareStmtContext> preparedStmtCtxs = Maps.newHashMap();
 
     // Control whether to read Iceberg caches without populating/updating them for the current execution.
@@ -1273,6 +1279,10 @@ public class ConnectContext {
 
     public boolean isRelationAliasCaseInsensitive() {
         return relationAliasCaseInsensitive;
+    }
+
+    public Set<String> getViewExpansionPath() {
+        return viewExpansionPath;
     }
 
     public void setForwardTimes(int forwardTimes) {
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

kangkaisen:
1. The "cycle" error gets masked by the outer wrapping.

  The cycle SemanticException is thrown in the inner visitView, but each enclosing visitView catches SemanticException and re-wraps it,
  so the user ends up seeing:

  ▎ View v1 references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them: ... contains a
  ▎ cycle ...

  The test only asserts message.contains("cycle"), which passes, but in practice the misleading "invalid table/column/permission"
  prefix is what users see first, and the real "cycle" reason is buried in the nested message. Consider giving the cycle a distinct
  exception type (or short-circuiting it in the outer catch so it passes through without being re-wrapped) so the surfaced error
  clearly states it's a cyclic view definition.

  2. The cycle stays in the catalog; every query keeps failing.

  This is a read-time guard, so it doesn't clean up the catalog. Once the cycle is committed, every SELECT on the view fails and the
  only recovery is another ALTER VIEW to break it. A more complete fix would detect and reject the cycle at ALTER VIEW analysis time
  (fail-fast, keep the catalog clean). This read-time guard is still valuable as a backstop for cycles from any source, but adding the
  ALTER-time check as a follow-up would prevent the bad metadata from being persisted in the first place

CelerData-Reviewer:
@codex review

kangkaisen (approved):

stdpain (approved):

PR #75034, opened 2026-06-19
[BugFix] Don't fail LIST-partitioned queries when a partition value can't be cast

Hyper-FF:
## Why I'm doing:

OptOlapPartitionPruner.listPartitionPrune only caught AnalysisException, but the LIST pruner casts every partition value to the predicate's target type via ListPartitionPruner.castLiteralExpr, which rethrows any failure as StarRocksConnectorException (a RuntimeException). A non-numeric partition value (e.g. 'east') under a CAST(col AS INT) = / IN / range predicate therefore made the exception escape pruning and fail an otherwise-valid query with "can not cast literal value ... to target type int".

Partition pruning is best-effort, so widen the catch to fall back to the unpruned candidate set (all partitions when no PARTITION hint is given) on any exception, letting the scan apply the predicate normally.

## What I'm doing:

Fixes #issue

reproduce case:
```sql
CREATE TABLE t_lp (id INT, region VARCHAR(32))
PARTITION BY LIST(region) (
  PARTITION p_num  VALUES IN ("100","200"),
  PARTITION p_text VALUES IN ("east","west")
)
DISTRIBUTED BY HASH(id) PROPERTIES("replication_num"="1");
INSERT INTO t_lp VALUES (1,"100"),(2,"east");

SELECT * FROM t_lp WHERE CAST(region AS INT) = 100;        -- FAILS
SELECT * FROM t_lp WHERE CAST(region AS INT) IN (100,200); -- FAILS
```
**Observed:** `ERROR 1064 (HY000): can not cast literal value east to target type int(11)`

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/OptOlapPartitionPruner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/OptOlapPartitionPruner.java
@@ -338,7 +338,12 @@ public class OptOlapPartitionPruner {
             } else {
                 return prune;
             }
-        } catch (AnalysisException e) {
+        } catch (Exception e) {
+            // Pruning is best-effort. A non-numeric LIST partition value that fails to cast (or any
+            // other pruning failure) must not abort the query: fall back to the unpruned candidate set
+            // (all partitions when no explicit PARTITION hint was given), and let the scan apply the
+            // predicate. StarRocksConnectorException (a RuntimeException) used to escape the old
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

PR #75038, opened 2026-06-19
[BugFix] Don't rewrite null-safe-equal join OR-disjuncts to UNION ALL

Hyper-FF:
## Why I'm doing:

SplitJoinORToUnionRule rewrites JOIN ON p1 OR p2 into a UNION ALL of two joins, adding a dedup predicate to the second branch so a row matching both disjuncts is not emitted twice. check() admits a disjunct as equi via JoinHelper.getEqualsPredicate, whose isEquivalence() is true for both EQ and EQ_FOR_NULL (<=>), but the dedup predicate is the negation of plain '=': (a != b) OR a IS NULL OR b IS NULL. For NULL operands a <=> b is TRUE (matched the first branch) and the dedup predicate is also TRUE, so the row is emitted by both branches -> duplicate rows.

Reject EQ_FOR_NULL disjuncts in check() so null-safe-equal ORs are left as a single join (the codebase already knows <=> has no simple negation).

## What I'm doing:

```sql
CREATE TABLE jl(k1 INT) DUPLICATE KEY(k1) DISTRIBUTED BY HASH(k1) BUCKETS 1 PROPERTIES("replication_num"="1");
CREATE TABLE jr(v1 INT, v2 INT) DUPLICATE KEY(v1) DISTRIBUTED BY HASH(v1) BUCKETS 1 PROPERTIES("replication_num"="1");
INSERT INTO jl VALUES (NULL);
INSERT INTO jr VALUES (NULL, NULL);

-- default (rule off): correct
SELECT * FROM jl JOIN jr ON jl.k1 <=> jr.v1 OR jl.k1 <=> jr.v2;   -- 1 row (NULL,NULL,NULL)

SET enable_rewrite_or_to_union_all_join = true;                   -- default false
SELECT * FROM jl JOIN jr ON jl.k1 <=> jr.v1 OR jl.k1 <=> jr.v2;   -- BUG: 2 identical rows
```
**Observed:** rule off → 1 row; rule on → **2 duplicate rows**. Plain `=` disjuncts are correct; only `<=>` is broken.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitJoinORToUnionRule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitJoinORToUnionRule.java
@@ -183,6 +183,14 @@ public class SplitJoinORToUnionRule extends TransformationRule {
             return false;
         }
 
+        // The per-branch dedup predicate built in transform() is the negation of plain '=', which
+        // is NOT a correct negation of the null-safe '<=>' (EQ_FOR_NULL): for NULL operands both
+        // the join condition and the dedup predicate are true, so the row is emitted by both
+        // UNION ALL branches (duplicate rows). Refuse to rewrite null-safe-equal disjuncts.
+        if (equalPredicates.stream().anyMatch(p -> p.getBinaryType() == BinaryType.EQ_FOR_NULL)) {
+            return false;
+        }
+
         return true;
     }
 
--- a/fe/fe-core/src/test/java/com/starrocks/sql/optimizer/rewrite/JoinPredicatePushdownTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/sql/optimizer/rewrite/JoinPredicatePushdownTest.java
@@ -264,4 +264,22 @@ public class JoinPredicatePushdownTest extends PlanTestBase {
             connectContext.getSessionVariable().setEnabledRewriteOrToUnionAllJoin(false);
         }
     }
+
+    @Test
+    public void testSplitJoinORToUnionRuleRejectsNullSafeEqual() throws Exception {
+        connectContext.getSessionVariable().setEnabledRewriteOrToUnionAllJoin(true);
+        try {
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/SplitJoinORToUnionRule.java
```diff
@@ -183,6 +183,14 @@ public boolean check(OptExpression input, OptimizerContext context) {
             return false;
         }
 
+        // The per-branch dedup predicate built in transform() is the negation of plain '=', which
+        // is NOT a correct negation of the null-safe '<=>' (EQ_FOR_NULL): for NULL operands both
+        // the join condition and the dedup predicate are true, so the row is emitted by both
+        // UNION ALL branches (duplicate rows). Refuse to rewrite null-safe-equal disjuncts.
+        if (equalPredicates.stream().anyMatch(p -> p.getBinaryType() == BinaryType.EQ_FOR_NULL)) {
```
stephen-shelby, line 190: nit: `<=>` is not inherently impossible to rewrite. It just needs a different dedup predicate from plain `=`. For `a <=> b OR c <=> d`, the safe rewrite would be:

branch1: a <=> b
branch2: c <=> d AND NOT (a <=> b)

or expand `NOT (a <=> b)` as:
a != b OR (a IS NULL AND b IS NOT NULL) OR (a IS NOT NULL AND b IS NULL)

So this conservative rejection is fine for now, but maybe worth mentioning that future support can be added by generating the null-safe dedup predicate.

stdpain (approved):

PR #75045, opened 2026-06-19
[BugFix] Fix TopN RANK boundary bug in chunks_sorter_topn

Hyper-FF:
## Why I'm doing:

Fix an off-by-one bug in be/src/exec/chunks_sorter_topn.cpp::_rank_pruning that caused RANK TopN to drop one row when the rank boundary lies exactly at a chunk boundary (i.e. rank_limit == k * chunk_size + 1). The issue manifested only when merged runs spanned multiple chunks. Root cause

The loop that locates which merged-run chunk contains the rank boundary used a strict > comparison, while the helper _get_run_by_row_id() uses >=. When the boundary index fell exactly at the end of a run/chunk, the wrong run was selected and the peer-group end was computed against an out-of-bounds/wrong row, silently dropping one row from the result. 

Change the comparison in _rank_pruning to match the semantics of _get_run_by_row_id() (use >= / equivalently compare cumulative >= boundary_idx + 1). This ensures the correct run is chosen when the boundary is exactly at a chunk end.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exec/chunks_sorter_topn.cpp
+++ b/be/src/exec/chunks_sorter_topn.cpp
@@ -750,7 +750,7 @@ void ChunksSorterTopn::_rank_pruning() {
 
     const auto& merged_runs = _merged_runs;
     for (int i = 0; i < merged_runs.num_chunks(); ++i) {
-        if (target_index > merged_runs.at(i).num_rows()) {
+        if (target_index >= merged_runs.at(i).num_rows()) {
             target_index -= merged_runs.at(i).num_rows();
         } else {
             index_in_runs = i;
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #75071, opened 2026-06-20
[BugFix] Do not rewrite avg(DISTINCT x) via a sum/count materialized view

Hyper-FF:
## Why I'm doing:

The MV aggregate rewrite turns avg(x) into sum(x)/count(x) so a query average can be
answered from a materialized view exposing sum/count for the same group. But
AggregateFunctionRewriter.canRewriteAggFunction accepted AVG without checking
isDistinct(), and rewriteAvg builds sum/count from the children without propagating
DISTINCT. So avg(DISTINCT x) was rewritten to the non-distinct sum(x)/count(x),
silently dropping DISTINCT and returning a wrong result when x has duplicates in a
group.

Decline the rewrite for distinct aggregates (canRewriteAggFunction returns false when
isDistinct()), so the query keeps its original correct aggregation. canRewriteAggFunction
only ever rewrote AVG, so the sole behavior change is that avg(DISTINCT x) is no longer
mis-rewritten; non-distinct avg(x) still uses the MV rewrite.

Repro: avg(DISTINCT x) over a sum/count MV returned 20 (= 60/3) with rewrite on vs the
correct 25 with rewrite off; now returns 25 in both cases.

Test: test/sql/test_materialized_view_rewrite/{T,R}/test_mv_avg_distinct_rewrite.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/materialization/AggregateFunctionRewriter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/materialization/AggregateFunctionRewriter.java
@@ -59,6 +59,12 @@ public class AggregateFunctionRewriter {
         }
 
         CallOperator aggFunc = (CallOperator) op;
+        // Rewriting avg(DISTINCT x) as sum(x)/count(x) over the MV's non-distinct
+        // sum/count would silently drop DISTINCT and return a wrong result, so decline
+        // the rewrite for any distinct aggregate.
+        if (aggFunc.isDistinct()) {
+            return false;
+        }
         String aggFuncName = aggFunc.getFnName();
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

HangyuanLiu (approved):

PR #75110, opened 2026-06-22
[BugFix] Derive LIST-partition MIN/MAX only from single-value partitions

Hyper-FF:
## Why I'm doing:

PartitionColumnMinMaxRewriteRule answered MIN/MAX(partition_col) on a LIST-partitioned table from the largest/smallest declared VALUES IN (...) value, gated only by hasData(). A non-empty partition is not guaranteed to contain its extreme declared value, so e.g. a partition VALUES IN ('10','20','30') holding only 10 made MAX(k) return 30 -- a value no row holds (silent wrong result).

A non-empty *single-value* LIST partition, however, must contain its one declared value (rows route by value), exactly like the rule's day-wide DATE range path (a one-day interval on a DATE column admits a single value). So restrict the LIST rewrite to single-value partitions: getMin/getMaxListPartitionValue return null for a multi-value partition, and optimizeWithPartitionValues then declines, falling back to the partition-prune / top-N paths that compute MIN/MAX from real data.

Adds ListPartitionCell.valueCount(). The range path is unchanged.

Repro: LIST(k) with p1 IN (1,2,3), p2 IN (10,20,30), data {1,10}: MAX(k) returned 30 with enable_rewrite_partition_column_minmax=true vs the correct 10; now 10. A single-value LIST table (s1 IN (5), s2 IN (15)) still uses the rewrite and correctly returns 15/5.

Test: test/sql/test_list_partition/{T,R}/test_list_partition_minmax_declared_absent.

## What I'm doing:

```sql
CREATE TABLE lpm (k INT) DUPLICATE KEY(k) PARTITION BY LIST(k) (PARTITION p1 VALUES IN ('1','2','3'), PARTITION p2 VALUES IN ('10','20','30')) DISTRIBUTED BY HASH(k) BUCKETS 1 PROPERTIES("replication_num"="1");
INSERT INTO lpm VALUES (1),(10);
SET enable_rewrite_partition_column_minmax=false; SELECT max(k) FROM lpm;  -- 10 (correct)
SET enable_rewrite_partition_column_minmax=true;  SELECT max(k) FROM lpm;  -- 30 (WRONG — declared list max of p2)
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/ListPartitionInfo.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/ListPartitionInfo.java
@@ -173,6 +173,17 @@ public class ListPartitionInfo extends PartitionInfo {
             return CollectionUtils.isEmpty(singleColumnValues) && CollectionUtils.isEmpty(multiColumnValues);
         }
 
+        // Number of values this partition declares in its VALUES IN (...) list.
+        public int valueCount() {
+            if (singleColumnValues != null) {
+                return singleColumnValues.size();
+            }
+            if (multiColumnValues != null) {
+                return multiColumnValues.size();
+            }
+            return 0;
+        }
+
         public ListPartitionValue minValue() {
             if (singleColumnValues != null) {
                 return ListPartitionValue.of(singleColumnValues.stream().min(LiteralExpr::compareTo).get());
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

PR #74043, opened 2026-05-29
[Enhancement] Move slow-publishPartition threshold into Config for live tuning

luohaha:
## Why I'm doing:

The phase-breakdown WARN at `maybeLogSlowPublishPartition` is the right tool for investigating lake-publish latency outliers, but its threshold was a hardcoded `3000` ms constant — it could not be tuned without a code change and FE restart. When chasing sub-second publish-latency jitter on a live cluster, that coarse, immutable threshold leaves the gap between "start publish lake" and the `publish_version` BRPC unexplained.

## What I'm doing:

Surface the threshold as `Config.slow_publish_partition_log_threshold_ms` (mutable, default **3000 ms**) so operators can dial it in via `ADMIN SET FRONTEND CONFIG` against a live cluster without a restart, then restore the default once an investigation completes.

- Replace the `SLOW_PUBLISH_PARTITION_LOG_THRESHOLD_MS = 3000` constant in `PublishVersionDaemon` with a read of `Config.slow_publish_partition_log_threshold_ms`.
- Add the config field in `Config.java` (mutable, default 3000 ms).

No behavior change at the default 3000 ms; lowering it makes the daemon WARN with the `executor_queue / db_lock_wait / fe_prep / rpc` decomposition.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [x] This pr needs user documentation (for new or modified features or behaviors)
  - [x] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

PR #75125, opened 2026-06-22
[Enhancement] Safe leader demotion: drain leader-only work in place on demotion instead of System.exit (tier-3 + followups) + HA bug fixes

gengjun-git:
## Why I'm doing:

When an FE loses the leader role (graceful `ALTER SYSTEM TRANSFER LEADER`, BDBJE re-election, or any LEADER -> FOLLOWER/OBSERVER transition), the old code path fell back to `System.exit(-1)`: the process crashed and relied on the external supervisor to restart it as a follower. That is the opposite of graceful — it maximizes the unavailability window, throws away the warm in-memory state, and risks straddling in-flight WAL writes across the handoff. Issue #63357 asks for the FE to instead **drain leader-only work in place and transition to follower without crashing**.

This is the tier-3 + followups milestone of that project. It builds on the already-merged scaffolding:
- **part 1 — #70911 (MERGED):** `LeaderRoleState`, `LeaderLease`, the admission fence, `leader_demotion_drain_timeout_sec`.
- **part 2 — #71932 (MERGED):** the `LeaderDaemon` framework and the partial `stopLeaderOnlyDaemonThreads` list.

What remained: the rest of the leader-only daemons + the two checkpoint controllers still crashed on demote; the executor-owning daemons could not survive a demote/re-elect cycle; the LEADER -> non-leader transition itself still called `System.exit(-1)`; and there was no fence guaranteeing that no in-memory mutation outlives its WAL record across the handoff. Real-machine HA validation of the resulting stack also surfaced three concrete bugs (a daemon-hang regression, a routine-load slot leak, and a commit-vs-seal race) that this PR fixes.

## What I'm doing:

Fixes #63357.

This PR finishes safe in-place leader demotion and fixes three bugs found during real-machine HA validation.

#### 1. Migrate the remaining tier-3 leader-only daemons + checkpoint controllers onto `LeaderDaemon`
Every daemon still started by `startLeaderOnlyDaemonThreads()` is migrated off the crash-on-demote base class onto `LeaderDaemon`, with a symmetric stop wired into `stopLeaderOnlyDaemonThreads()` in strict reverse-of-start order:
- **Standard migrations:** `PipeListener`, `StarMgrMetaSyncer`, `TabletWriteLogHistorySyncer`, `TabletReshardJobMgr` (its maps are image/gson-persisted, so `onStopped` is a deliberate no-op).
- **State-clearing `onStopped()` hooks:** `PipeScheduler` (resets recovered flag + clears `beSlotMap`, and now calls `Pipe.resetForLeaderHandoff()` per pipe so a re-elected leader re-schedules), `TaskRunStateSynchronizer`, `VectorIndexBuildScheduler`.
- **Special forms:** `TaskManager` (does not extend `FrontendDaemon`; the two schedulers are made `volatile` and rebuilt by `start()`, with a new `stop()`), the `taskCleaner` inline daemon, `AutovacuumDaemon`/`FullVacuumDaemon` (executor made `volatile`, rebuilt on start / shut on stop), `ClusterSnapshotJobScheduler`, `ClusterSnapshotMgr.stopGracefully(timeoutMs)` (now nulls the scheduler field so a re-elected leader rebuilds it with fresh `CheckpointController` deps), and `ExportChecker` (extends `LeaderDaemon`, new static `stopAll(timeoutMs)` that drains PENDING/EXPORTING checkers and closes its three `LeaderTaskExecutor`s).
- Both **`CheckpointController`s** (`GlobalStateMgr` + `StarMgrServer`) migrate from `FrontendDaemon` to `LeaderDaemon`. Because the three checkpoint phases are idempotent system-wide, `onStopped()` only drops leader-session bookkeeping; `onStopRequested()` offers a failure status to wake the in-flight wait and calls `MetaHelper.cancelInFlight()`. Added `StarMgrServer.stopCheckpointController(timeoutMs)`.
- The two non-daemon load schedulers (`pendingLoadTaskScheduler`, `loadingLoadTaskScheduler`) are now drained via their `close(awaitMillis)` APIs and wired into the reverse-start stop sequence.

#### 2. Rebuild executor pools + restart-gate so daemons survive demote/re-elect
Every demotion-drained component that owns a worker pool is hardened against the demote/re-elect race. `LeaderTaskExecutor` and `PriorityLeaderTaskExecutor` save their construction params and rebuild both internal pools inside `synchronized start()` when the previous pool was shutdown, and throw `IllegalStateException` from `start()` if the previous pool is shutdown-but-not-yet-terminated — mirroring the `BatchWriteMgr`/`AlterHandler` restart guard, so a fresh pool never races a lingering one (this fixes `RejectedExecutionException` on re-elected broker-load scheduling, because the `GlobalStateMgr` scheduler fields are final singletons). The same shutdown-vs-terminated gap is closed in `TaskManager`, `AutovacuumDaemon`, `FullVacuumDaemon`, `HeartbeatMgr`, and `PublishVersionDaemon` (each `onStopped` awaits termination up to `leader_demotion_drain_timeout_sec`; keep the reference on timeout so the restart guard fires, null it on clean drain so getters lazily rebuild). `ExportChecker.stopAll` now blocks on actual executor termination.

#### 3. 6-stage in-place demotion orchestrator + journal seal
The LEADER -> FOLLOWER/OBSERVER/UNKNOWN `System.exit(-1)` in `StateChangeExecutor` is replaced by an in-place transition (`System.exit` kept only for unexpected target types). On demotion from LEADER, `transferToNonLeader` runs a fixed ordered pipeline:

```
beginLeaderDemotion -> sealJournalWriter -> stopLeaderOnlyDaemonThreads
  -> flushLeaderSessionState -> switchFrontendType(targetType) -> completeLeaderDemotion
```

Each stage is wrapped by `runDemotionStage`, which logs timing and only `System.exit(-1)`s if a stage throws. `sealJournalWriter` calls `JournalWriter.sealAndStop(timeout)`, then `awaitLeaderWalAppliesDrained`, then advances `replayedJournalId` to the sealed committed watermark; `flushLeaderSessionState` cancels running insert-overwrite jobs. The `LeaderDaemon` framework no longer interrupts a worker mid-cycle: `stop` uses a `stopSignal` monitor, `stopGracefully` waits for the current cycle without interrupting, and a new `onStopRequested()` hook lets subclasses wake their own cancellable waits. A stale-leader-activation rollback path (`rollbackStaleLeaderActivationIfNeeded`) lets a node that lost the BDB master race during activation roll back and restart the replayer instead of exiting.

#### 4. Leader WAL-apply fence: no in-memory mutation outlives its WAL record, no work straddles the handoff
A WAL-apply fence in `GlobalStateMgr` guards the gap between writing a journal record and applying it in memory. New `runWithLeaderJournalAdmission` / `runWithLeaderWalApplyAdmission(OrThrow)` entry points acquire `leaderWalApplyFenceLock`, validate admission (reject when not leader / admission closed / DEMOTING, throwing `JournalWriteException`), count in-flight applies, and return a `LeaderWalApplyRegistration` whose `AutoCloseable` scope records any apply failure and decrements the counter. `EditLog.logEdit/logJsonObject(/OrThrow)` are rewritten to submit-under-admission, wait for the WAL commit, then `applyWithWalFence`; the old per-call `Preconditions` leader check is replaced by this centralized fence. `awaitLeaderWalAppliesDrained` blocks until in-flight applies finish (or throws if one recorded a failure) so demotion never proceeds with a half-applied mutation.

The `JournalWriter` daemon no longer blanket-`System.exit(-1)`s: interrupts/throwables while not RUNNING stop cleanly, restart fails fast if the previous daemon is still alive, and `batchWriteCommit` takes a `shouldRetry` `BooleanSupplier` that `BDBJEJournal` honors to break the BDBJE commit-retry sleep early (in 100ms slices) once the writer leaves RUNNING. Many entry points become handoff-aware: `LeaderImpl.finishTask` rejects with `LEADER_TRANSFERRED`/`INTERNAL_ERROR`; `FrontendServiceImpl` txn begin/commit/prepare/rollback share `rejectIfLeaderLoadTxnAdmissionClosed`; `TabletTaskExecutor` cancels tablet creation when admission closes; `StmtExecutor`/`GlobalTransactionMgr` stop waiting for publish on demotion and bound the transferring-to-leader spin by `thrift_rpc_timeout_ms`; `LeaderOpExecutor` bounds and skips forwarded-journal-replay waits.

#### Fix 1 — `LeaderDaemon`: interval-0 daemons hang after one cycle on `wait(0)`
`LeaderDaemon.loop()` unconditionally called `stopSignal.wait(intervalMs)` at the end of every cycle. For interval-0 daemons (report-handler, resource-report-handler, routine-load-task-scheduler) this is `Object.wait(0)`, which by Java semantics blocks **forever**, so those daemons ran exactly one cycle per leader activation (stuck disk-capacity reporting, stalled routine load after a transfer). Fix: guard the inter-cycle wait with `if (intervalMs > 0)`; interval <= 0 is a tight drain loop that self-paces via its own bounded blocking call in `runAfterLeaseValid()`. `setStop()`/`stopGracefully()` still wake the loop promptly.

#### Fix 2 — `RoutineLoadTaskScheduler`: release BE task slots on leader demotion
Per-backend routine-load task-slot counts are leader-session state. `takeBeTaskSlot()` only ever incremented per-backend usage, demotion never released it, and `updateBeTaskSlot()` does not reset counts for backends already present and alive — so every leader -> follower -> leader cycle leaked slots until `max_routine_load_task_num_per_be` was exhausted (`takeBeTaskSlot` returned -1, "no available be slot ... 16") and no routine-load task could be scheduled again. Fix: added `RoutineLoadMgr.clearBeTaskSlot()` (under `slotLock`, zeroes every per-backend usage in `warehouseNodeTasksNum` and clears every node->jobs set in `warehouseNodeToJobs`, keeping capacity entries intact) and call it from `RoutineLoadTaskScheduler.onStopped()`. The steady-state scheduling path is unchanged.

#### Fix 3 — `BDBJEJournal`: treat closed-transaction `IllegalStateException` as a commit failure
During a leader transfer, if a journal commit is in flight and its BDBJE transaction is invalidated/closed before `commit()` runs (commit-vs-seal race, or a long GC/IO stall), `Transaction.commit()` throws a plain `java.lang.IllegalStateException` ("Transaction ... has been closed") — not a `DatabaseException`. `BDBJEJournal.batchWriteCommit` only caught `DatabaseException` and `JournalWriter.writeOneBatch` only catches `JournalException`, so the unchecked ISE escaped both layers: the in-flight batch's tasks were never `markAbort`'d, waiting `logEdit` callers blocked forever, and the leader WAL-apply fence (`leaderWalApplyInFlight`) leaked. `awaitLeaderWalAppliesDrained` then never reached zero, demotion waited the full `leader_demotion_drain_timeout_sec` (180s default) and called `System.exit(-1)` — defeating safe demotion. Fix: broaden the catch to `catch (DatabaseException | IllegalStateException e)`, converting the closed-transaction ISE into a `JournalException` that flows the existing retry/abort path (while SEALING it fails fast and aborts the current batch so callers unblock and the fence drains in milliseconds, no exit).

> Note: the final consolidating commit ("Fix FE leader demotion handling", `4e4fec4747e`) has an empty git body; its scope is the demotion orchestration + WAL-apply fence described above.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [x] BugFix
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

(Enhancement: safe in-place leader demotion; BugFix: the three regressions/races above.)

- [x] Yes
- [ ] No

- [ ] Parameter changes (default values, similar parameters but with different default values)
- [ ] Policy changes (use new policy to replace old one, functionality automatically pivots to new behavior while deprecating the legacy behavior)
- [ ] Feature removed (remove a feature flag, command line, or function)
- [ ] Miscellaneous (upgrade dependencies, extend time range of a specific functionality, etc.)

- [x] **Policy changes:** on losing the leader role, an FE now performs an ordered in-place drain + transition to FOLLOWER/OBSERVER instead of `System.exit(-1)`. The crash-exit fallback is retained only for unexpected target types and as the demotion self-fence (a stage throwing or the drain timing out). Graceful-transfer unavailability is ~2s (vs ~1.2s for a hard kill) — a small downtime cost for a clean handoff that preserves warm in-memory state and avoids straddling in-flight WAL writes.

## Checklist:

- [x] I have added test cases for my bug fixes or new features.
- [ ] I have added user documentation for any new features or behavior changes. (No user-facing config/syntax changes; `leader_demotion_drain_timeout_sec` already shipped with part 1.)
- [ ] This is a backport pr.

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

**Backport:** main-only, **not** for backport. The safe-leader-demotion feature stack (parts 1 & 2 #70911 / #71932) is only on `main` — it is not present in any released branch (3.5 / 4.0 / 4.1) — so the three bug fixes have nothing to backport to, and the version checkboxes are intentionally left unchecked.

### Unit tests
30 test files changed (+2112 / -41), 4 brand-new, all JUnit5, no `Deencapsulation`/`setAccessible` except for genuinely-private fields in `EditLogFailureVisibleTest` (per the no-reflection-in-UT rule).

**Core demotion state-machine / journal-seal:**
- `server/GlobalStateMgrTest` (+251): leader-transfer flag latched on role-change notification; stale-leader activation rolls back when the BDB master changed (anti-split-brain); seal TIMEOUT throws and does **not** advance the replayed-journal-id (no phantom watermark); seal blocks on the WAL-apply fence before advancing `replayId` on both LEADER_LOST and clean BARRIER_REACHED paths; leader bootstrap idempotent across re-promotion.
- `journal/JournalWriterSealTest` (+93): seal/drain barrier semantics — empty-queue watermark, barrier cuts the batch and aborts post-barrier tasks, commit failure -> LEADER_LOST, commit-retry disabled after seal starts, timeout returns latest committed watermark, clean restart after seal, force-roll reason.
- `journal/bdbje/BDBJEJournalTest` (+61): **Fix 3** at unit level — `testBatchWriteCommitConvertsTransactionClosedToJournalException`.
- `persist/EditLogFailureVisibleTest` (+82): `submitLogOrThrow` rejects with `ADMISSION_CLOSED` / `NOT_LEADER` (but allows `OP_STARMGR` on follower); `waitOrThrow` propagates `WRITER_ABORTED`/`TIMEOUT`; `waitInfinity` does not hang when the task is aborted mid-demotion; `logJsonObject(OrThrow)` closes the WAL-apply fence and skips the applier on abort.

**The three bug fixes (direct mapping):**
- `common/util/LeaderDaemonTest` (+180, NEW): **Fix 1** — `testZeroIntervalKeepsLoopingInsteadOfHangingAfterOneCycle` proves `wait(0)` no longer blocks forever (>=3 cycles); plus graceful-stop-doesn't-interrupt and hard-`setStop`-still-interrupts.
- `load/routineload/RoutineLoadSlotResetTest` (+53, NEW): **Fix 2** — `testOnStoppedReleasesBeTaskSlotsOnDemotion`.
- `journal/bdbje/BDBJEJournalTest` (above): **Fix 3**.

**Daemon/executor lifecycle (stop-on-demotion / clean-restart-on-repromotion):** `LeaderTaskExecutorTest` (+85), `PriorityLeaderTaskExecutorTest` (+109), `TabletTaskExecutorTest` (+73), `TaskManagerTest` (+129), `PublishVersionDaemonTest` (+118), `AutovacuumDaemonTest` (+120, NEW), `FullVacuumDaemonTest` (+116, NEW), `HeartbeatMgrTest` (+96), `ExportCheckerTest` (+61), `CheckpointControllerTest` (+36), `StarMgrServerTest` (+22) — consistent pattern: `onStopped` shuts down & awaits executor termination, `start` rebuilds pools, `start` refuses restart before the old pool terminates, and a logs-on-refuse path.

**Caller-facing demotion semantics:** `LeaderImplTest` (+58, `finishTask` returns `LEADER_TRANSFERRED` while demoting), `FrontendServiceImplTest` (+47, loadTxn RPCs rejected while demoting), `GlobalTransactionMgrTest` (+39, publish-wait stops post-commit on demotion), `LeaderOpExecutorTest` (+78, forwarded stmt returns leader result if follower replay times out), `StmtExecutorTest` (+42, bounded self transfer-to-leader wait), `StreamLoadTaskTest` (+26), `InsertOverwriteJobRunnerEditLogTest` (+29), plus smaller follow-on adjustments in `StarMgrMetaSyncerTest`, `VacuumTest`, `ClusterSnapshotTest`, `VectorIndexBuildSchedulerTest`, `TabletReshardJobMgrTest`.

### Real-machine HA validation (5-node-plan, distributed, not docker)
Cluster `transfer-leader-test` = 3 electable FE (.29/.31/.32) + 2 BE (.27/.28), `replication_num=2`; build = main@036205ca566 + safe-demotion stack + the three fixes (BE built `--without-tenann`); chaos controller on a separate dev box (witness separation).

**Headline:** data safety held across **12 leader transitions** (6 graceful `transferMaster` + 6 ungraceful `kill -9`) + a network-partition fencing test + 4 compound-failure scenarios + an 8-round stream-load run + a 12-step DDL-across-transfer run + the Fix-3 commit-seal race end-to-end check. **LOST = DUP = PHANTOM = 0 in every reconciliation.**

- **① Data-safety oracle + ② ungraceful kill + ④ availability** (6 autocommit writers, unique keys, DUPLICATE-KEY rep=2): attempted 10431 / OK-acked 10283 / UNCERTAIN 148 (12 resolved-present, 136 resolved-absent) / FAIL 0; `table_rows 10295 == distinct_keys 10295`; **LOST 0 / DUP 0 / PHANTOM 0 -> PASS**. Every OK-acked write present exactly once; every UNCERTAIN deterministically present-XOR-absent. Unavailability windows (incl. client reconnect, upper bound): graceful safe-demotion n=6 p50 2.057s / max 2.901s; ungraceful kill n=6 p50 1.169s / max 1.286s. The graceful window is slightly **longer** than a hard kill — the ordered seal+drain+handoff trades ~1s of downtime for a clean handoff; both bounded < 3s. All 6 graceful `je_rc=0`; all 6 killed nodes rejoined alive.
- **③ Stale-leader fencing / split-brain:** iptables-DROP leader<->followers on BDBJE 9010 + RPC 9020 (both directions), drive writes at the partitioned stale leader, then heal. 4 writes at the stale leader -> OK 0 / ERR 2 (InsufficientAcks) / TIMEOUT 2 -> **0 commits**; after heal, **0 PHANTOM** stale-leader keys, old leader rejoined as FOLLOWER, single leader. `fencing(no stale commit)=True`, `no_divergence=True`, `clean_rejoin=True`. BDBJE `SIMPLE_MAJORITY` quorum is the real fencing token — an isolated minority leader physically cannot commit. (The harness prints a spurious "FAIL" only because its in-loop elect flag didn't latch on a silent-partition heartbeat-timeout election ~30–50s; confirmed not a real failure by the post-heal leader change.)
- **B1 stream-load across transfer** (4 loaders, 8 rounds): loads_ok 1838 / uncertain 273 (9 present); `table_rows 5523 == distinct_keys 5523`; **LOST 0 / DUP 0 / PHANTOM 0 -> PASS**.
- **B2 DDL-across-transfer + B4 cross-FE metadata integrity:** 12/12 DDLs `ddl_rc=0` with `cross_fe_consistent=True`; post-run integrity 5/5 tables schema- and count-consistent across all FE -> **PASS**, no metadata divergence.
- **B3 compound failures** (6 writers, 4 scenarios incl. kill follower / kill leader / kill-new-leader-then-immediately-transfer / kill mid-demotion): all 4 recovered; reconcile attempted 1641 / ok 1574 / uncertain 67 / fail 0; `table_rows 1576 == distinct_keys 1576`; **LOST 0 / DUP 0 / PHANTOM 0 -> PASS**.
- **Fix 3 (commit-vs-seal race) end-to-end** (failpoint `bdb_write_jitter` holds a commit open ~10s; transfer driven into that window, FIXED build): drain=180s × 3 trials — each old leader rejoined as FOLLOWER in 0.5–1.5s, `exited=False`, `cluster_ok=True`, sentinel rows ∈ {0,1} (no DUP/phantom), MAX demote wall 2.2s. On the buggy build this stalled ~180s and `System.exit(-1)`'d; on the fixed build the ISE converts to a `JournalException`, the fence drains, and demotion completes in seconds. (Harness "FAIL" label here is inverted — it expected to observe the stall+exit; their absence is the pass condition.)
- **Drain-timing corroboration** (separate cluster): 15/15 + 3/3 graceful rounds PASS, each showing the ACTIVE -> DEMOTING -> INACTIVE staged transition with bounded drains (0.0–10.1s), routine-load committed counts monotonically advancing, BE capacity stable across transfers — consistent with Fix 1 (no daemon hang) and Fix 2 (no slot leak).

**Verified-correct behaviors (not gaps):**
1. An FE that loses write quorum mid-write while still believing it is leader exits via the JournalWriter fail-fast path (`InsufficientAcks` × `RETRY_TIME`, `writerState` still RUNNING, no demotion notice) — owner-confirmed intentional. The graceful path (`writerState` SEALING, clean notification) aborts cleanly with `LEADER_LOST` + `markAbort`, no exit. Two distinct correct paths.
2. A 3-node-min-quorum cascade in an early fence run traced to repeatedly killing the **same** node, whose rejoin did a BDBJE `HARD_RECOVERY` rollback of a locally-durable-but-never-acked write (no data-safety violation); did not reproduce on a healthy cluster. Argues for 5-FE in production.

**Not run this batch / partial:**
- ⑤ quorum-loss / asymmetric partition / follower-lag — needs 5 electable FE; out of scope for this 3FE+2BE cluster.
- The drain-timeout -> `System.exit` self-fence (Fix-3 sub-scenario ⑥-B, drain=5s × 2) was attempted but **not positively triggered** — the in-flight commit resolved within the drain, so the timeout->exit path wasn't exercised on the real machine; both nodes rejoined cleanly, `cluster_ok=True`, no data loss. That self-fence path remains covered only by the unit-level seal/fence tests.
- Methodology caveat: unavailability windows include client-reconnect latency (upper bounds); ungraceful runs always killed the current leader and graceful always transferred to .29, concentrating churn on .29.

## Note on base / sync with main

This branch was forked at `036205ca566` (#74043). It has now been **merged up to current `upstream/main` (`a10c264`, v4.2-rc2)** — it is no longer behind, and GitHub reports it **mergeable** (no conflicts). The merge had a single content conflict, in a test file only — `TaskManagerTest.java`, where main and this branch both appended `@Test` methods at the class end; resolved by keeping both sides' tests. All other 70 files (including the core `EditLog` / `StmtExecutor` / `TaskManager` / `GlobalStateMgr` / journal paths) auto-merged cleanly, and main's autovacuum change #74906 auto-merged into the migrated `AutovacuumDaemon`.

Post-merge verification against the up-to-date code: **FE compiles clean** (`./build.sh --fe`), and the merge-affected + fix unit tests pass — `TaskManagerTest` 55, `BDBJEJournalTest` 20, `LeaderDaemonTest` (common+server) 14, `JournalWriterSealTest` 8, `RoutineLoadSlotResetTest` 1, `AutovacuumDaemonTest`/`FullVacuumDaemonTest`/`VacuumTest` 21 — **0 failures**. The full real-machine 5-node HA validation summarized above was performed on the pre-merge build (the safe-demotion stack + the three fixes); the merge only brought in unrelated main progress + the test-only conflict resolution, so it does not invalidate those results, but a `run-fe-ut.sh` full pass + a smoke HA round are still advisable in CI before merge.


gengjun-git:
Superseded by the two-way split of this work: #75592 (leader-daemon graceful-demotion adaptation + `ALTER SYSTEM TRANSFER LEADER` SQL) and #75591 (the 3 standalone backportable bug fixes). Closing in favor of those two PRs.

PR #75137, opened 2026-06-22
[BugFix] Keep full 128 bits of a LARGEINT literal in JIT codegen

Hyper-FF:
## Why I'm doing:

IRHelper::load_ir_number built a 128-bit constant for a TYPE_LARGEINT / TYPE_DECIMAL128 literal via APInt(128, reinterpret_cast<const int128_t*>(value)[0], true). The matched overload is APInt(unsigned, uint64_t, bool) -- there is no __int128 form -- so the int128 argument was narrowed to uint64_t and the high 64 bits were dropped. Any LARGEINT literal
>= 2^64 was zeroed/wrapped under JIT while the interpreter kept the full value.

Build the APInt from both 64-bit words of the int128 (little-endian, low word first) so the full 128-bit literal is preserved. Values that fit in 64 bits are unaffected (the high word is the correct sign/zero extension). int128 columns were already correct; only literals were wrong.

Repro: sum(v + 18446744073709551616 - v) over a LARGEINT column returned 0 at jit_level=-1 vs the correct 5000*2^64 at jit_level=0; now equal.

Test: test/sql/test_agg/{T,R}/test_jit_largeint_literal.

## What I'm doing:

Fixes #issue
```sql
CREATE TABLE jitli(id INT, v LARGEINT NULL) DUPLICATE KEY(id) DISTRIBUTED BY HASH(id) BUCKETS 1 PROPERTIES("replication_num"="1");
INSERT INTO jitli SELECT generate_series, 1 FROM TABLE(generate_series(1,5000)) g;

SET jit_level=0;  SELECT sum(v + 18446744073709551616 - v) FROM jitli;   -- 92233720368547758080000 (=5000·2^64)
SET jit_level=-1; SELECT sum(v + 18446744073709551616 - v) FROM jitli;   -- 0   (2^64 literal truncated to 0)
-- also: literal 10^20 → interp 500000000000000000000000 vs JIT 38831398157261209600000 (low-64-bit wrap)
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exprs/jit/ir_helper.cpp
+++ b/be/src/exprs/jit/ir_helper.cpp
@@ -115,7 +115,13 @@ StatusOr<llvm::Value*> IRHelper::load_ir_number(llvm::IRBuilder<>& b, const Logi
         return b.getInt64(reinterpret_cast<const int64_t*>(value)[0]);
     case TYPE_LARGEINT:
     case TYPE_DECIMAL128: {
-        llvm::APInt value_128(128, reinterpret_cast<const int128_t*>(value)[0], true);
+        // The APInt(numBits, uint64_t, isSigned) overload would narrow the int128 to 64 bits and
+        // drop the high half (zeroing/wrapping any literal >= 2^64). Build it from both 64-bit
+        // words (little-endian: low word first) so the full 128-bit literal is preserved.
+        const int128_t raw = reinterpret_cast<const int128_t*>(value)[0];
+        const uint64_t words[2] = {static_cast<uint64_t>(static_cast<__uint128_t>(raw)),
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

kangkaisen (approved):

PR #75139, opened 2026-06-22
[BugFix] Bound-check truncated percent-escape in url_decode (OOB read)

Hyper-FF:
## Why I'm doing:

url_decode_slice read value[i+1] and value[i+2] after seeing '%' with no check that two bytes remain. For column input (e.g. url_extract_parameter) the value slice is not NUL-terminated, so a value ending in a truncated escape ("...%" or "...%X") read 1-2 bytes past the slice into adjacent column memory. The result therefore depended on neighbouring rows -- an out-of-bounds read / cross-row information leak (and the over-read bytes were even echoed into the "illegal hex chars" error message).

Add `if (i + 2 >= len) return error` before reading the two hex digits, so a truncated escape yields a clean error instead of reading out of bounds. Complete escapes still decode.

Repro (column input): two tables identical except for the row following 'http://x/p?k1=ab%'; before the fix url_extract_parameter(u,'k1') returned hex 6162AB when the next row began with hex bytes vs NULL otherwise -- the result varied with adjacent memory. After the fix both return NULL (clean), and a complete '%41' still decodes to 'A'.

Test: test/sql/test_string_functions/{T,R}/test_url_extract_parameter_truncated_escape.

## What I'm doing:

Fixes #issue

ASAN UT:
```
  ==19499==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000013
     at pc 0x51c72f bp 0x7ffe09109710 sp 0x7ffe09109708
  READ of size 1 at 0x502000000013 thread T0
      #0 0x51c72e in url_decode_slice                  url_oob_repro.cpp:10   ← char l = value[i+1];
      #1 0x51c72e in seek_param_key_in_query_params    url_oob_repro.cpp:27
      #2 0x51c72e in url_extract_parameter             url_oob_repro.cpp:30
      #3 0x51c72e in main                              url_oob_repro.cpp:36
      #4 libc start ...

  0x502000000013 is located 0 bytes after 3-byte region [0x502000000010,0x502000000013)
  allocated by thread T0 here:
      #0 operator new[](unsigned long)
      #1 0x51c56e in main                              url_oob_repro.cpp:33   ← char* buf = new char[3];  // "ab%"

```

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exprs/string_functions.cpp
+++ b/be/src/exprs/string_functions.cpp
@@ -2781,6 +2781,13 @@ static Status url_decode_slice(const char* value, size_t len, std::string* to) {
     to->reserve(len);
     for (size_t i = 0; i < len; i++) {
         if (value[i] == '%') {
+            // A '%' must be followed by two hex digits. Without this bound check, a truncated
+            // escape ("...%" or "...%X") at the end of the slice reads 1-2 bytes past it. The
+            // slice is not NUL-terminated for column input (e.g. url_extract_parameter), so this
+            // is an out-of-bounds read of adjacent memory whose result depends on neighbouring data.
+            if (i + 2 >= len) {
+                return Status::RuntimeError("decode string contains an incomplete percent-encoding");
+            }
             char l = value[i + 1];
             char r = value[i + 2];
             if ((l < 'A' || l > 'F') && (l < '0' || l > '9')) {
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #75141, opened 2026-06-22
[BugFix] array_map/transform drops NULL rows when all non-null arrays are empty

Hyper-FF:
## Why I'm doing:

In array_map_expr.cpp, when a chunk has some null rows (so it does not take the all-null fast path) but every non-null input array is empty (total_elements_num == 0), the fast path returned a ConstColumn of empty arrays sized num_rows - null_rows and without a null mask. So null input rows were rendered as [] (NULL silently dropped) and the row count was wrong. The adjacent all-null branch correctly uses num_rows and a NullableColumn; array_filter uses a different path and was already correct.

When there are null rows, build num_rows empty arrays and wrap them in a NullableColumn with the per-row null mask (a const column cannot represent the mix of NULL and [] rows). With no null rows the rows are identical, so the const empty-array column is kept.

Repro: array_map(x->x+1, arr) over rows (NULL,[],[]) returned [] for the NULL row (and count(... IS NULL)=0) vs array_filter's correct NULL; now array_map returns NULL too.

Test: test/sql/test_array_fn/{T,R}/test_array_map_null_empty_rows.

## What I'm doing:

Fixes #issue

```
  CREATE TABLE amt(id INT, arr ARRAY<INT>) DUPLICATE KEY(id)
    DISTRIBUTED BY HASH(id) BUCKETS 1 PROPERTIES("replication_num"="1");
  INSERT INTO amt VALUES (1, NULL), (2, []), (3, []);

  SELECT id, array_map(x -> x+1, arr) AS am FROM amt ORDER BY id;
  SELECT count(*) FROM amt WHERE array_map(x -> x+1, arr) IS NULL;   -- expect 1 (the NULL row)
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exprs/array_map_expr.cpp
+++ b/be/src/exprs/array_map_expr.cpp
@@ -398,13 +398,27 @@ StatusOr<ColumnPtr> ArrayMapExpr::evaluate_checked(ExprContext* context, Chunk* 
                     ->get_total_elements_num(result_null_column);
 
     if (total_elements_num == 0) {
-        // if all input rows are empty arrays, return a const empty array column as result
+        // All non-null input arrays are empty. When some rows are null we cannot use a const
+        // column (null rows and empty-`[]` rows differ): build `num_rows` empty arrays and apply
+        // the per-row null mask so null inputs stay null instead of being rendered as `[]` (and
+        // so the row count is correct). With no null rows every row is identical, so a const
+        // empty array column is fine.
         column = ColumnHelper::create_column(type().children[0], true);
         auto aligned_offsets = UInt32Column::create(0);
+        if (result_null_column) {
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

stdpain (approved):

trueeyu (approved):

PR #75143, opened 2026-06-22
[BugFix] bar(): reject negative/huge width instead of growing unbounded (DoS)

Hyper-FF:
## Why I'm doing:

bar(size, min, max, width) read its width/min/max BIGINT arguments into an unsigned size_t. A negative width (e.g. -1) wrapped to SIZE_MAX and slipped past the `width <= 0` guard (which only rejects 0 for an unsigned value); result_width = ratio * width then became ~SIZE_MAX and the inner `for (j < result_width) bar.append(...)` grew a plain std::string without bound. That string is not tracked by the MemTracker, so a single query exhausted BE memory and hung the node (KILL QUERY could not stop the append loop; unrelated to query_mem_limit) -- a DoS.

Read width/min/max/size as signed int64_t (they come from BIGINT and may legitimately be negative), so a negative width now correctly fails `width <= 0` and negative value ranges render correctly. Also bound width with kMaxBarWidth (1000000) so even a large positive width cannot grow the result unbounded.

Repro: SELECT length(bar(v,0,100,-1)) hung with BE RSS climbing past 88GB. After the fix bar(...,-1)/(...,0)/(...,1e11) return a clean "requirement: width ..." error, bar(5,0,10,20) renders "▓▓▓▓▓▓▓▓▓▓", and bar(-5,-10,0,20) renders correctly.

Test: test/sql/test_function/{T,R}/test_bar_negative_width (the unbounded-growth case is intentionally not exercised).

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exprs/utility_functions.cpp
+++ b/be/src/exprs/utility_functions.cpp
@@ -365,6 +365,9 @@ StatusOr<ColumnPtr> UtilityFunctions::get_query_profile(FunctionContext* context
 
 StatusOr<ColumnPtr> UtilityFunctions::bar(FunctionContext* context, const Columns& columns) {
     static std::u8string kBar = u8"\u2593";
+    // Upper bound on the rendered bar length. The per-row result is a plain std::string that is
+    // NOT tracked by the MemTracker, so an unbounded width can exhaust BE memory (DoS).
+    constexpr int64_t kMaxBarWidth = 1000000;
     RETURN_IF(columns.size() != 4, Status::InvalidArgument("expect 4 arguments"));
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #75145, opened 2026-06-22
[BugFix] Restrict time_slice() MV rewrite to sound binary operators

Hyper-FF:
## Why I'm doing:

TimeSliceRewriteEquivalent lets a time_slice() rollup MV answer a query that filters the raw timestamp by rewriting `dt OP c` (bucket-aligned c) to `time_slice(dt) OP c`, but it kept the same operator with no guard. time_slice() floors a timestamp to its bucket start, so only `>=` and `<` are equivalent on the raw column; `=`, `<=` and `>` select the whole bucket [c, c+interval), giving a silently wrong result.

Guard the rewrite with SUPPORTED_BINARY_TYPES = {GE, LT} and decline any other operator. This is stricter than DateTruncEquivalent (which also allows GT, sound for its DATE/day identity-like case); time_slice with a sub-day interval genuinely buckets, so GT is unsound.

Repro: a 5-minute time_slice MV with rows in [00:00,00:05); count(*) WHERE dt='2023-06-01 00:00:00' returned 3 (whole bucket) with the rewrite on vs the correct 1. After the fix `=` and `<=` match the rewrite-off answer while `>=`/`<` still use the MV.

Test: test/sql/test_materialized_view_rewrite/{T,R}/test_mv_time_slice_predicate_rewrite.

## What I'm doing:

Fixes #issue

```
  -- minimal repro: a time_slice() rollup MV over-selects the whole bucket for a `=` filter on raw dt

  CREATE TABLE t (dt DATETIME) DUPLICATE KEY(dt)
  DISTRIBUTED BY HASH(dt) BUCKETS 1 PROPERTIES("replication_num"="1");
  -- two rows in the SAME 5-minute bucket [00:00, 00:05)
  INSERT INTO t VALUES ('2023-06-01 00:00:00'), ('2023-06-01 00:01:00');

  CREATE MATERIALIZED VIEW mv REFRESH MANUAL AS
  SELECT time_slice(dt, INTERVAL 5 MINUTE) AS t, count(*) AS c
  FROM t GROUP BY time_slice(dt, INTERVAL 5 MINUTE);

  SELECT count(*) FROM t WHERE dt = '2023-06-01 00:00:00';
  -- enable_materialized_view_rewrite = false : 1   (correct — only the 00:00:00 row)
  -- enable_materialized_view_rewrite = true  : 2   (WRONG — both rows in the 00:00 bucket)


```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/materialization/equivalent/TimeSliceRewriteEquivalent.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/materialization/equivalent/TimeSliceRewriteEquivalent.java
@@ -14,8 +14,10 @@
 
 package com.starrocks.sql.optimizer.rule.transformation.materialization.equivalent;
 
+import com.google.common.collect.ImmutableSet;
 import com.starrocks.catalog.FunctionSet;
 import com.starrocks.common.AnalysisException;
+import com.starrocks.sql.ast.expression.BinaryType;
 import com.starrocks.sql.optimizer.operator.scalar.BinaryPredicateOperator;
 import com.starrocks.sql.optimizer.operator.scalar.CallOperator;
 import com.starrocks.sql.optimizer.operator.scalar.ColumnRefOperator;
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

PR #75156, opened 2026-06-22
[Doc] Note that ALTER TABLE flat_json properties are not propagated to the BE

moweonlee:
## Why I'm doing

Table-level Flat JSON properties (`flat_json.*`) currently take effect when they are set at
table-creation time. The table-level `ALTER TABLE ... SET ("flat_json.*" = ...)` example in the
docs does not mention this, which can be misleading. This change documents the current behavior
so users know to set these properties at table creation.

## What I'm doing

Add a `:::note` after the table-level `ALTER TABLE` flat_json example clarifying that table-level
Flat JSON configuration currently takes effect when it is set at table-creation time. Updated in
all three language docs (en / zh / ja).

## Checklist
- [x] Documentation only — `docs/en|zh|ja/using_starrocks/Flat_json.md` updated in sync


moweonlee:
Superseded by #75158

PR #75161, opened 2026-06-22
[BugFix] Fix flaky AND-only MATCH queries

SnobbyDragon:
## Why I'm doing:
Starrocks would sometimes return `Match can only used as a pushdown predicate on column with GIN in a single query.` error for valid AND-only MATCH queries. The same query would also sometimes succeed.

This flaky behaviour was due to the error only occurring when a segment was empty (either initialized empty or after being pruned by ranges/bitmap/indices/sampling/etc.)

If the segment was empty, we would skip deleting the MATCH predicate during `_apply_inverted_index` and proceed to `_rewrite_predicates` which would then see the MATCH predicate and cause the error.

## What I'm doing:
Check if the segment is empty before attempting rewriting the predicates.
If the segment is empty, we don't need to rewrite because there are no rows anyways.

With this fix, it is possible that an invalid MATCH query would return 0 rows instead of raising an error.
This case would be very rare because it would require all segments to be empty after pruning, so we would never attempt predicate rewriting + raise the MATCH error. 

## What type of PR is this:

- [X] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/rowset/segment_iterator.cpp
+++ b/be/src/storage/rowset/segment_iterator.cpp
@@ -3628,6 +3628,8 @@ Status SegmentIterator::_init_global_dict_decoder() {
 }
 
 Status SegmentIterator::_rewrite_predicates() {
+    RETURN_IF(_scan_range.empty(), Status::OK());
+
     {
         // in normal case it can always rewrite the predicate,
         // but for JSON extended column, it might be a JsonExtractColumnIterator, so we need to fallback to the orignal predicate


CelerData-Reviewer:
@codex review

srlch (approved):

stdpain (approved):

zxl-wangwang:
When will the fix for this issue be released?
The problems we are currently facing are highly similar to this one.

henriqueng:
@Mergifyio backport branch-4.1

srlch:
@Mergifyio backport branch-4.1

PR #75179, opened 2026-06-23
[BugFix] Enforce query memory limit during table function execution

Wenjun7J:
## Why I'm doing:

A query with `unnest` over array columns can drive BE memory far past `query_mem_limit` and get the **BE OOM-killed**, instead of just failing that one query. Reported on 3.5.18 (shared-nothing).

## What I'm doing:

Wrap the `TableFunction::process()` call in `_process_table_function` with `TRY_CATCH_BAD_ALLOC`, so allocations during the expansion are checked against the query / query_pool / process mem tracker **at allocation granularity**. This mirrors how `ProjectOperator::push_chunk`, `AggregateBlockingSinkOperator::push_chunk`, `ExchangeSinkOperator`, hash/nl-join probes and ~60 other hot paths already guard their per-chunk compute — `TableFunctionOperator` was simply missing it. A runaway table function now returns `MemoryLimitExceeded` instead of crashing the whole BE.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature (see Tests note — not unit-testable in `BE_TEST`)
- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exec/pipeline/table_function_operator.cpp
+++ b/be/src/exec/pipeline/table_function_operator.cpp
@@ -188,13 +188,18 @@ Status TableFunctionOperator::_process_table_function(RuntimeState* state) {
     _next_output_row = 0;
     _next_output_row_offset = 0;
 
-    // A table function (e.g. unnest) materializes the whole expansion of the input
-    // chunk in this single call, with no intermediate memory-limit check. For wide
-    // arrays this one step can allocate tens of GB and get the BE OOM-killed before
-    // the next pipeline-driver limit check fires. Wrap it so allocations are checked
-    // against the query/query_pool/process mem tracker per batch so that a runaway
-    // query fails with MemoryLimitExceeded instead of crashing the whole BE.
-    TRY_CATCH_BAD_ALLOC(_table_function_result = _table_function->process(state, _table_function_state));
+    // A table function (e.g. unnest) materializes the whole expansion of the input chunk in
+    // this single call, with no intermediate memory-limit check. For wide arrays this one step
+    // can allocate tens of GB and get the BE OOM-killed before the next pipeline-driver limit
+    // check fires. For implementations whose process() is exception-safe, wrap it in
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

be/src/exec/pipeline/table_function_operator.cpp
```diff
@@ -187,7 +188,13 @@ Status TableFunctionOperator::_process_table_function(RuntimeState* state) {
...
     _next_output_row_offset = 0;
 
-    _table_function_result = _table_function->process(state, _table_function_state);
+    // A table function (e.g. unnest) materializes the whole expansion of the input
+    // chunk in this single call, with no intermediate memory-limit check. For wide
+    // arrays this one step can allocate tens of GB and get the BE OOM-killed before
+    // the next pipeline-driver limit check fires. Wrap it so allocations are checked
+    // against the query/query_pool/process mem tracker per batch so that a runaway
+    // query fails with MemoryLimitExceeded instead of crashing the whole BE.
+    TRY_CATCH_BAD_ALLOC(_table_function_result = _table_function->process(state, _table_function_state));
```
stdpain: There might be non-exception-safe operations inside. We need to restrict the effect to only certain functions. You can refer to FunctionCallExpr.

CelerData-Reviewer:
@codex review

be/src/exec/pipeline/table_function_operator.cpp
Wenjun7J: set unnest/multiunnest as exception safe, the rest table funcitons still keep unsafe, since they will call the thirdparty package.

stdpain (approved):

satanson (approved):

PR #75186, opened 2026-06-23
[Enhancement] Revive Parquet unsigned integer footer stats for meta-tier pre-split

xiangguangyxg:
## Why I'm doing:

Sample-Based Tablet Pre-Split has a **meta tier** that reads a Parquet file's footer min/max statistics to compute split boundaries cheaply (no data-tier sub-query). Parquet UNSIGNED integers (`UINT_8`/`UINT_16`/`UINT_32`, all physically INT32-backed) were implemented for this path but **reverted mid-PR in #74955**: at that time the BE loaded a Parquet `UINT_32 → BIGINT` by **sign-extending** the int32, so a high-bit value such as `3000000000` was stored as `-1294967296`. The footer's unsigned-magnitude min/max could not be aligned with that (negative) loaded value, so the meta-tier boundary would have been wrong.

PR #75002 fixed the BE root cause: unsigned INT32 is now **zero-extended** on load (`be/src/formats/parquet/column_converter.cpp`), already merged to `main` (`685bc36f66a`) and branch-4.1 (`0daf79ec141`). With that deployed BE-first, the unsigned-magnitude footer boundary aligns with the loaded value, so the meta-tier path is safe to revive.

This PR also fixes a related FILES gap: a bare `SELECT * FROM FILES(<parquet with a UINT_32 column>)` inferred the column as `INT`, silently overflowing high values negative.

## What I'm doing:

**FE — `ParquetRowGroupStatisticsReader` (meta-tier footer reader), value-based:**
- Accept a Parquet `INT32` column annotated UNSIGNED (`!isSigned() && bitWidth <= 32`, i.e. UINT_8/16/32) as a meta-tier sort key, **only** when the file's raw footer positively declares a `TypeDefinedOrder` column order for that leaf (so the footer min/max are unsigned-ordered; legacy/UNDEFINED-order files fall back to the data tier). This is the FE analog of the BE's `SortOrder::UNSIGNED` footer-stats guard from #75002.
- Decode the footer min/max as the true unsigned magnitude (`Integer.toUnsignedLong`), then build a `Variant` of the StarRocks sort-key column type. `IntVariant` range-checks each value; a magnitude that overflows the target column throws and is caught, so the file falls back to the **data tier** — never a wrong split, never a load failure. This admits any integer sort key and lets the per-value range check screen the cases where the loaded value would diverge (e.g. a UINT_32 ≥ 2^31 into an `INT` column).
- Generalize the byte-array-DECIMAL column-order helper `declaresSignedByteArrayOrder` → `declaresTypeDefinedColumnOrder` (now shared by byte-array DECIMAL and unsigned ints) and extract `requiresColumnOrderCheck`.

**BE — `parquet_schema_builder.cpp` (FILES schema inference):**
- For an unsigned Parquet `INT32`, infer a signed StarRocks type wide enough to hold the full unsigned range: `UINT_8 → SMALLINT`, `UINT_16 → INT`, `UINT_32 → BIGINT` (mirrors `variant_integer_desc_from_bitwidth` in `be/src/formats/parquet/utils.cpp`). Signed ints and all non-int annotations are unchanged. `UINT_64`/INT64 are out of scope (cannot fit BIGINT, consistent with #75002). The vectorized load path already zero-extends `UINT_32 → BIGINT` (#75002), so the wider inferred type matches the loaded value.

Scope is limited to the meta-tier Parquet footer reader and the FILES schema-inference primitive (plus their tests). The downstream pre-split pipeline (`BoundaryPlanner`/`Variant`/sampler) and the load-path converter are unchanged — the converter was already fixed in #75002. ORC has no unsigned integer types, so `OrcStripeStatisticsReader` is untouched.

**Deployment ordering (BE-first):** the revived FE meta-tier path relies on #75002 being deployed BE-first (already merged to main/4.1). End-to-end validation must run on a cluster whose BE contains #75002 (`main` ≥ `685bc36f66a`; the 4.1 backport must use 4.1 BE containing `0daf79ec141`).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Both parts correct previously-wrong or missing behavior rather than alter intended behavior. The FE meta-tier change only affects internal tier selection (meta vs data) with identical query results. The BE FILES inference change corrects a lossy inference (UINT_32 high values previously overflowed negative under `INT`); like #75002 (which corrected the loaded values from negative to positive and was classified behavior-change=No), this corrects a bug rather than changing intended behavior. Flagging the user-visible effect explicitly: a bare `SELECT * FROM FILES(<uint parquet>)` now reports a wider type (e.g. `BIGINT` for `UINT_32`) and surfaces high values as their correct positive magnitude instead of overflowed negatives.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #75188, opened 2026-06-23
[BugFix] Unregister txn-state callback when removing multi-statement stream load task

meegoo:
## Why I'm doing:

Multi-statement stream load leaks one entry in `TxnStateCallbackFactory` per transaction, growing without bound until the FE runs out of heap.

The parent `StreamLoadMultiStmtTask` is registered as a transaction-state callback in `StreamLoadMgr.addLoadTask` → `getCallbackFactory().addCallback(task)`. But the explicit transaction created by `TransactionStmtExecutor.beginStmt` carries **no callback id** (only the per-statement `InsertLoadJob` jobIds are added via `addCallbackId`). Therefore `TransactionState.afterStateTransform` never dispatches to the parent — its `afterCommitted` / `afterVisible` / `afterAborted` (which only delegate to sub-tasks) are never invoked, and the parent's callback is **never removed**.

`TxnStateCallbackFactory` is a plain `HashMap` with no expiry/sweep, and `unprotectedRemoveTaskFromDb` (the place every final-state task removal goes through) did not remove the callback. So:

- `idToStreamLoadTask` is bounded by `stream_load_task_keep_max_num`, but the callback map is **not** — it grows monotonically, one entry per multi-statement transaction, forever.
- Each leaked entry pins the parent task, its `ConnectContext` (which holds a full `SessionVariable`, ~520 fields — never nulled) and the sub-task `StreamLoadTask` shells. ~10–25 KB per leak in the common case, unbounded in count.
- The `TxnCallbackCount` memory-tracker metric was observed climbing past 46k entries (~GB-level retained heap) and rising with throughput/uptime.

Ordinary (single-table / sync) stream load is **not** affected — it removes its callback in `afterVisible`/`afterAborted`. Sub-task coordinator memory is already released on current code via `markCommittedByParent(visible) → gcObject()`; this PR only fixes the dangling callback entry.

## What I'm doing:

Remove the callback in `StreamLoadMgr.unprotectedRemoveTaskFromDb`:

```java
GlobalStateMgr.getCurrentState().getGlobalTransactionMgr().getCallbackFactory()
        .removeCallback(streamLoadTask.getId());
```

`unprotectedRemoveTaskFromDb` is called only from `cleanOldStreamLoadTasks` and `cleanSyncStreamLoadTasks`, both of which only remove tasks that have reached a **final state** (`checkNeedRemove` / `isFinalState`). So the callback is guaranteed to be unneeded at this point. For an ordinary `StreamLoadTask` the callback was already removed at terminal state, so this is an idempotent no-op (`removeCallback` is a guarded `Map.remove`). For the multi-statement parent task it is the actual fix. Callback ids are globally unique (`getNextId`), so there is no risk of evicting a different live object's callback.

This caps the callback map at roughly the number of live stream-load tasks (≤ `stream_load_task_keep_max_num`), turning an unbounded leak into a bounded footprint.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

It only stops an internal memory leak. The removed callback was never dispatched, so no user-observable behavior (load semantics, results, SHOW STREAM LOAD / information_schema.loads display) changes.

## Checklist:

- [ ] This is a backport pr

Added `StreamLoadManagerTest#testMultiStmtTaskRemovesTxnCallbackOnCleanup` — a true red/green regression (fails without the fix: `expected <null> but was <StreamLoadMultiStmtTask@...>`; passes with it). All 59 stream-load FE unit tests pass; checkstyle clean.

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/streamload/StreamLoadMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/streamload/StreamLoadMgr.java
@@ -582,6 +582,19 @@ public class StreamLoadMgr implements MemoryTrackable {
             }
         }
 
+        // Unregister the txn-state callback that was registered in addLoadTask(), otherwise the task
+        // leaks in TxnStateCallbackFactory forever.
+        // - For an ordinary StreamLoadTask the callback is already removed in afterVisible/afterAborted,
+        //   so this is an idempotent no-op.
+        // - For a StreamLoadMultiStmtTask the explicit transaction carries no callback id, so the parent
+        //   task's afterCommitted/afterVisible/afterAborted are never dispatched and its callback is never
+        //   removed. Without this call every multi-statement stream load would leave one dangling entry
+        //   (and the StreamLoadTask sub-task shells it references) in the callback map permanently.
+        // This is only reached after the task has reached a final state (see checkNeedRemove /
+        // isFinalState in the callers), so the callback is guaranteed to be unneeded here.
+        GlobalStateMgr.getCurrentState().getGlobalTransactionMgr().getCallbackFactory()
+                .removeCallback(streamLoadTask.getId());
+
         if (streamLoadTask instanceof StreamLoadTask) {
             warehouseLoadStatusInfoBuilder.withRemovedJob((StreamLoadTask) streamLoadTask);
         }
--- a/fe/fe-core/src/test/java/com/starrocks/load/streamload/StreamLoadManagerTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/load/streamload/StreamLoadManagerTest.java
@@ -18,6 +18,7 @@ import com.google.common.collect.Lists;
 import com.starrocks.backup.CatalogMocker;
 import com.starrocks.catalog.Database;
 import com.starrocks.common.AnalysisException;
+import com.starrocks.common.Config;
 import com.starrocks.common.FeConstants;
 import com.starrocks.common.StarRocksException;
 import com.starrocks.common.jmockit.Deencapsulation;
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

starrocks-xupeng (approved):

HangyuanLiu (approved):

banmoy:
A txn-state callback's lifetime should follow the **transaction's lifecycle**: registered when the txn begins, unregistered when the txn reaches a terminal state. That's how every other load task already works.

  This PR instead removes the callback in `unprotectedRemoveTaskFromDb`, which is only reached from the background sweep (`cleanOldStreamLoadTasks` / `cleanSyncStreamLoadTasks`) when the task is **evicted from the map**. That's not the txn lifecycle — it's a janitor that runs much later (after
  `stream_load_task_keep_max_second`, or only when the task map overflows). So the callback and the object graph it pins linger long after the txn is done, and `TxnCallbackCount` only returns to baseline at sweep time, not at commit/abort.

  Cleaning up here is fine as a **backstop**, but I don't think it should be the primary mechanism.

  Suggested options 

  - **Option 1:** don't register the parent in `TxnStateCallbackFactory` at all. The parent's id never enters the transaction's `callbackIdList`, so its callback is never dispatched — it's dead weight today. If it's never a dispatch target, it shouldn't be registered,
  and there's nothing to leak or to clean up.
  - **Option 2:** unregister the parent's callback at its terminal-state transition (in `commitTxn` after it reaches COMMITED, and on the abort/rollback paths), so it's freed the moment the txn finishes.

kevincai (approved):

PR #74107, opened 2026-06-01
[Refactor] Make lake segment_metas the canonical per-segment metadata

xiangguangyxg:
## Why I'm doing:

The lake `RowsetMetadataPB` / `TxnLogPB.OpWrite` historically store per-segment and per-del-file metadata as several positionally-aligned parallel arrays (`segments`, `segment_size`, `segment_encryption_metas`, `bundle_file_offsets`, `shared_segments` on the rowset; `dels`, `rewrite_segments`, `del_encryption_metas`, `shared_dels` on the op-write). Parallel arrays are fragile: every producer/reader must keep N arrays index-aligned, guard length mismatches, and defensively `std::min()` over them. This consolidates them into structured per-record messages (Phase 1).

## What I'm doing:

- Rename the 9 legacy parallel-array fields with a `deprecated_` prefix, **keeping the same field numbers** (wire-compatible), and make the structured fields canonical:
  - `RowsetMetadataPB.segment_metas[]` — `SegmentMetadataPB` gains `filename` / `size` / `encryption_meta` / `shared` / `bundle_file_offset` (fields 8–12).
  - `OpWrite.dels_meta[]` and `OpWrite.rewrite_segments_meta[]` (`FileMetaPB`).
- Migrate all BE readers and producers to the structured fields.
- Add a normalizer (`lake_proto_normalizer`) hooked at the `TabletManager` load/save choke points:
  - **before save**: rebuild the deprecated legacy arrays from the structured fields, so a BE rolled back to a pre-feature version can still read the metadata (done on a throwaway copy; the cached in-memory object is not mutated);
  - **after load**: back-fill the structured fields from the legacy arrays for data written by a pre-feature BE, then clear the legacy arrays in memory — they are never read again, so the cached metadata holds only the structured fields (no duplicate per-segment file-attribute copies) and can never go stale relative to them.
- Add `SegmentFileInfo::to_proto(segment_idx, *)` and a `to_file_meta_pb(FileInfo, *)` helper to unify the segment / del / sst / lcrm metadata producers, and remove the now-redundant `write_sort_key_fields_to(SegmentMetadataPB*)` overloads.

This is Phase 1 (dual-write legacy for rollback). Phase 2 will gate off the legacy dual-write once rollback below this version is unsupported; Phase 3 will remove the deprecated fields from the proto.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
[...]

PR #75193, opened 2026-06-23
[BugFix] Preserve segment filenames when dual-writing legacy arrays from un-normalized metadata

xiangguangyxg:
## Why I'm doing:

PR #74107 made `SegmentMetadataPB.filename` the canonical per-segment field and the
`deprecated_*` parallel arrays (`RowsetMetadataPB.deprecated_segments`, ...) a derived **dual-write**
kept only for version-rollback compat: `normalize_*_after_load()` back-fills the structured fields from
the legacy arrays then clears them; `normalize_*_before_save()` rebuilds the legacy arrays from the
structured fields. after-load was wired into every S3 load entry point but **not** the paths that
receive metadata/txn-logs over RPC from another BE.

On a shared-data cluster with `file_bundling=true`, under a **mixed BE-version transition** (rolling
upgrade *or* rollback), the aggregate publish/compaction and the import/load paths can permanently
erase segment filenames:

- **OLD producer → NEW persister**: the old node sends legacy-shaped metadata (`segment_metas` carries
  sort-key fields but no `filename`; real names only in `deprecated_segments`). The persister's
  `before_save` sees `segment_metas` non-empty, clears `deprecated_segments`, and rebuilds it from the
  empty `filename()` → `[""]`. Both name sources are now empty → permanent loss; the tablet resolves to
  `s3://.../<partition>/data/` and starlet reports the object does not exist.
- **NEW producer → OLD persister** (the rollback direction; aggregate publish predates #74107 and
  exists in 4.0): the new node sends canonical-only metadata; an old persister with no normalizer
  persists `deprecated_segments` empty → unreadable by old BEs.

## What I'm doing:

A minimal fix that does **not** change the on-disk format or the dual-write contract:

1. **Non-destructive, idempotent before-save** (`lake_proto_normalizer.cpp`): back-fill `segment_metas`
   (and `dels_meta` / `rewrite_segments_meta`) from the legacy arrays **before** clearing+rebuilding
   them. after-load extends `segment_metas` to cover every legacy entry; before-save uses **no-extend**
   so a partial-compaction-trimmed segment is never resurrected, and **fail-closes** (`Corruption`)
   rather than silently truncate if it ever sees un-normalized longer-legacy input.
2. **After-load received metadata/txn-logs on entry at every foreign-input point** (so the no-extend
   before-save always sees complete `segment_metas`, and the in-memory metacache holds the canonical
   form): the aggregate persist choke points `put_bundle_tablet_metadata` and `put_combined_txn_log`
   (the latter also covers the **normal multi-node combined-txn import write**), the non-bundling
   `repair_tablet_metadata` path, and **snapshot restore** (`LakeSnapshotLoader`). `put_tablet_metadata`
   caches the original object, so its foreign callers normalize on entry rather than deferring to a
   save-time copy.
3. **Producer dual-writes the RPC payload at all four cross-BE send sites** (aggregate publish,
   resharding, compaction scheduler, and the **import tablet-writer `add_txn_logs`**) so a persister of
   any version receives old-readable bytes.

Tests: unit tests in `lake_proto_normalizer_test.cpp` (back-fill / idempotency / fail-closed /
both-empty / op_write dels+rewrite); regression tests in `tablet_manager_test.cpp` (legacy-shaped and
sparse-count `put_bundle_tablet_metadata`), `lake_service_test.cpp` (aggregate-publish payload
dual-write, aggregate-compaction sparse receive + on-disk dual-write, non-bundling repair preserves
legacy filenames), and `lake_tablets_channel_test.cpp` (import writer's returned txn logs are
dual-written).

**Mixed-version cluster e2e** (shared-data, TSP): on a 4.1.x cluster with one CN held at **4.1.1** (pre-#74107 legacy producer), running the identical `file_bundling` aggregate-publish + `lake_use_combined_txn_log` workload, the only variable being the #74107 **persister** CN version — **without the fix** (official `4.1.2` persisters) the exact reported symptom reproduces (`starlet err Object s3://.../<tablet>/data/ does not exist`; DUP & AGG tables unreadable, PRIMARY KEY returned 9216/30720 rows, ~70% lost); **with the fix** (`branch-4.1` + this PR persisters) all three tables are fully intact (30720/30720 rows, every value-invariant holds, full scans clean). Write-up in the comment below.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/runtime/lake_snapshot_loader.cpp
+++ b/be/src/runtime/lake_snapshot_loader.cpp
@@ -24,6 +24,7 @@
 #include "runtime/exec_env.h"
 #include "runtime/snapshot_loader.h"
 #include "storage/lake/filenames.h"
+#include "storage/lake/lake_proto_normalizer.h"
 #include "storage/lake/tablet.h"
 #include "storage/storage_env.h"
 
@@ -311,6 +312,12 @@ Status LakeSnapshotLoader::restore(const ::starrocks::RestoreSnapshotsRequest* r
                 return Status::Corruption(fmt::format("failed to parse tablet meta {}", full_remote_file));
             }
             meta->set_id(restore_info.tablet_id());
+            // The snapshot blob is parsed directly (it never went through a load path), so a backup taken
+            // on a pre-feature BE is legacy-shaped. Normalize it on entry exactly like a disk-loaded one
+            // (after-load back-fills segment_metas from the legacy arrays) so both the cached object and
+            // put_tablet_metadata's dual-write before-save carry the segment filenames. put_tablet_metadata
+            // caches the original object, so this must run here on the cached object, not on a copy inside.
+            lake::normalize_tablet_metadata_after_load(meta.get());
             RETURN_IF_ERROR(StorageEnv::GetInstance()->lake_tablet_manager()->put_tablet_metadata(meta));
         }
 
--- a/be/src/runtime/lake_tablets_channel.cpp
+++ b/be/src/runtime/lake_tablets_channel.cpp
@@ -47,6 +47,7 @@
 #include "storage/lake/async_delta_writer.h"
 #include "storage/lake/delta_writer.h"
 #include "storage/lake/delta_writer_finish_mode.h"
+#include "storage/lake/lake_proto_normalizer.h"
 #include "storage/memtable.h"
 #include "storage/memtable_flush_executor.h"
 #include "storage/primitive/tablet_info.h"
@@ -150,7 +151,18 @@ private:
         void add_txn_logs(const std::vector<TxnLogPtr>& logs) {
             _response->mutable_lake_tablet_data()->mutable_txn_logs()->Reserve(logs.size());
             for (auto& log : logs) {
-                _response->mutable_lake_tablet_data()->add_txn_logs()->CopyFrom(*log);
+                // Dual-write the deprecated legacy arrays into the txn log shipped back to the load
+                // coordinator, so a coordinator BE that lacks the segment_metas refactor (mixed-version)
+                // still persists a combined txn log an old reader can resolve. DeltaWriter emits only the
+                // structured segment_metas; before-save adds the legacy dual-write. Mirrors the compaction
+                // worker. On the (practically unreachable) normalize failure, fail the load rather than
+                // ship a non-dual-written log.
+                TxnLogPB normalized(*log);
+                if (auto st = starrocks::lake::normalize_txn_log_before_save(&normalized); st.ok()) {
+                    _response->mutable_lake_tablet_data()->add_txn_logs()->Swap(&normalized);
+                } else {
+                    update_status(st);
+                }
             }
         }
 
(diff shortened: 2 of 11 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg:
## End-to-end validation on a mixed-version shared-data cluster

Validated this fix on a real shared-data (存算分离) cluster with **mixed BE versions across the #74107 boundary** — reproducing the exact data-loss symptom and confirming the fix eliminates it.

### Why 4.1.x
`#74107` (make `segment_metas` canonical) landed in `branch-4.1` **between 4.1.1 and 4.1.2**, which gives three clean points around the bug:
- **4.1.1** — pre-#74107: produces legacy-shaped metadata (`deprecated_segments` only, no `segment_metas.filename`) → the "old worker".
- **4.1.2** — has #74107 **without** this fix → the destructive `before_save`.
- **branch-4.1 + this PR** — has #74107 **with** the fix.

(The validation is intentionally done within 4.1.x rather than main↔4.1.1: the 4.1→4.2 data-plane chunk wire format is incompatible in unrelated ways, and 4.1.1→4.1.2 is the actual upgrade boundary where users hit this.)

### Setup
1 FE + 3 CN shared-data cluster. One CN held on **4.1.1** (legacy producer) for both runs. Identical workload each run: 30 rounds of `INSERT` into DUP / PRIMARY KEY / AGG tables with `file_bundling=true` (aggregate publish) and `lake_use_combined_txn_log=true`, followed by `COMPACT`. The **only** variable is the version of the #74107 **persister** CNs.

### Result (controlled experiment)

| Run | #74107 persister CNs | Outcome |
|-----|----------------------|---------|
| **With this fix** | `branch-4.1` + this PR | ✅ **30720 / 30720 rows** in all 3 tables, every value-invariant holds, full-table scans clean — **zero loss** |
| **Control (no fix)** | official `4.1.2` | ❌ `starlet err Object s3://.../<db>/<table>/<tablet>/data/ does not exist` on DUP & AGG (tablets unreadable); PRIMARY KEY returned 9216 / 30720 (~70% rows lost) |

The control reproduces the **exact reported symptom** — an empty segment filename collapsing the object path to `.../data/`. Same workload, same topology, same 4.1.1 producer CN; the only difference is this fix → **the fix is what prevents the loss, end to end.**

Backport to `branch-4.1`: #75227 (the build that was validated above).

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

kevincai:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

kevincai (approved):

luohaha (approved):

PR #68123, opened 2026-01-19
[Enhancement] Support segment-level splitting for parallel compaction

meegoo:
## Why I'm doing:

The current parallel compaction can only group rowsets at the rowset level. For large rowsets with many segments (e.g., 8GB+ with 8+ segments), a single subtask must process all segments sequentially, limiting parallelism and increasing memory pressure.

## What I'm doing:

- Add segment range mode to Rowset class for partial segment loading
- Add SubtaskType (NORMAL/LARGE_ROWSET_PART) to distinguish subtask types
- Implement large rowset split with "all-or-nothing" semantics:
  * Split criteria: size >= 2x max_rowset_size, segments >= 4, overlapped
  * All subtasks from same split must succeed for apply
- Add unit tests for new functionality

Fixes #66457

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

PR #75206, opened 2026-06-23
[BugFix] Index segment metadata by true position in segment-range mode

banmoy:
## Why I'm doing:

In segment-range mode (the rowsets built by parallel split-compaction sub-tasks, e.g. `tablet_parallel_compaction_manager::execute_subtask_segment_range`), `Rowset::load_segments` packs the loaded segments densely from vector position 0. So a loaded-vector position is **not** the segment's index in `_metadata->segment_metas()` once the sub-range starts past segment 0.

`Rowset::read`, `get_each_segment_iterator`, and `get_each_segment_iterator_with_delvec` decided whether to apply the tablet key-range filter by reading `_metadata->segment_metas(i).shared()` with the loop position `i` directly. For a sub-range `[k, k+1)` with `k > 0` this reads `segment_metas(0).shared()` instead of `segment_metas(k).shared()` — the wrong segment's flag.

This only bites on a split / merged / resharded tablet, where a rowset can carry a tablet range and **non-uniform** per-segment `shared` flags (a boundary-straddling segment is `shared=true`, a contained sibling is `shared=false`). There, the mis-read skips the tablet-range filter on a genuinely-shared segment and **leaks a sibling tablet's rows** into the read / compaction output. It is a no-op on tablets that were never split (no `shared` segments, no `range()`).

Introduced by #68123

## What I'm doing:

- Record each loaded segment's **true** `segment_metas()` position in a new `Rowset::LoadedSegment{segment, segment_meta_pos}` at load time (the position is the natural loop variable in `load_segments`, so there is no offset arithmetic to get wrong), and have all three read paths consult `shared()` via `segment_meta_pos` instead of the loop position. Binding the position to the segment is also robust to partial-compaction trimming and lost segments, where a simple `start + i` offset would not be.
- Add a `DCHECK` in `load_segments` asserting the recorded position matches the loaded segment's identity (`get_segment_idx`), so any future mis-indexing trips immediately in debug / UT builds.
- The change is a no-op outside segment-range mode (`segment_meta_pos` equals the loop position when the range starts at 0).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [ ] Miscellaneous: corrects read/compaction output on split tablets undergoing parallel split-compaction; no interface, parameter, or policy change

## Checklist:

- [ ] This is a backport pr

  - [X] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -162,11 +162,11 @@ Status Rowset::add_partial_compaction_segments_info(TxnLogPB_OpCompaction* op_co
         op_compaction->mutable_output_rowset()->mutable_range()->CopyFrom(metadata().range());
     }
 
-    std::vector<SegmentPtr> segments;
+    std::vector<LoadedSegment> segments;
     LakeIOOptions lake_io_opts{.fill_data_cache = true};
     SegmentReadOptions seg_options;
     seg_options.lake_io_opts = lake_io_opts;
-    std::pair<std::vector<SegmentPtr>, std::vector<SegmentPtr>> not_used_segments;
+    std::pair<std::vector<LoadedSegment>, std::vector<LoadedSegment>> not_used_segments;
     RETURN_IF_ERROR(load_segments(&segments, seg_options, &not_used_segments));
 
     bool clear_file_size_info = false;
@@ -181,7 +181,7 @@ Status Rowset::add_partial_compaction_segments_info(TxnLogPB_OpCompaction* op_co
     for (size_t i = 0; i < metadata().next_compaction_offset(); ++i) {
         auto* segment_meta = output_rowset->add_segment_metas();
         segment_meta->CopyFrom(metadata().segment_metas(i));
-        StatusOr<int64_t> size_or = already_compacted_segments[i]->get_data_size();
+        StatusOr<int64_t> size_or = already_compacted_segments[i].segment->get_data_size();
         int64_t file_size = 0;
         if (size_or.ok()) {
             file_size = *size_or;
@@ -192,7 +192,7 @@ Status Rowset::add_partial_compaction_segments_info(TxnLogPB_OpCompaction* op_co
                          << " for tablet " << _tablet_id
                          << " when collecting uncompacted segment info, error: " << size_or.status();
         }
-        uncompacted_num_rows += already_compacted_segments[i]->num_rows();
+        uncompacted_num_rows += already_compacted_segments[i].segment->num_rows();
         uncompacted_data_size += file_size;
     }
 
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/lake/rowset.cpp
```diff
@@ -626,12 +647,18 @@ Status Rowset::load_segments(std::vector<SegmentPtr>* segments, SegmentReadOptio
...
     auto check_status_at_index = [&](StatusOr<SegmentPtr>& segment_or, const std::string& seg_name, int seg_id,
-                                     int target_idx) -> Status {
+                                     int target_idx, int32_t meta_pos) -> Status {
         if (segment_or.ok()) {
+            // The segment was loaded with id get_segment_idx(metadata(), meta_pos); assert the
+            // recorded metadata index still matches its identity, so a wrong meta_pos trips
+            // immediately in debug/UT builds instead of silently reading another segment's metadata.
+            DCHECK(get_segment_idx(metadata(), meta_pos) == segment_or.value()->id())
+                    << "loaded segment / metadata index mismatch: meta_pos=" << meta_pos
+                    << " segment_id=" << segment_or.value()->id();
```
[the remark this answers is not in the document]
banmoy: Good catch — confirmed against `TabletManager::load_segment`. The NOTE at `tablet_manager.cpp:1629` documents that after partial compaction a cached `Segment`'s id can differ from the requested `segment_id` (the metacache keys by file path, and the cached object keeps its original id), so this `DCHECK` could false-abort debug/`BE_TEST` builds on a perfectly valid cached file.

Removed the `DCHECK` in 50e37f9. There's no salvageable id-based cross-check here, since the segment's identity is exactly what can be stale; correctness instead relies on recording `segment_meta_pos` at load time, which is unconditional.

xiangguangyxg (approved):

be/src/storage/lake/rowset.h
```diff
@@ -151,12 +151,23 @@ class Rowset : public BaseRowset {
...
     [[nodiscard]] StatusOr<std::vector<SegmentPtr>> segments(const LakeIOOptions& lake_io_opts);
 
+    // Pairs a loaded segment with its position in _metadata->segment_metas(). load_segments() packs
+    // its result densely (segment-range start, partial-compaction trim, lost segments), so an
+    // element's index there is NOT its metadata position; consult per-segment metadata (e.g. shared())
+    // via `segment_meta_pos` -- the segment_metas() array position, not Segment::id()/segment_idx.
+    struct LoadedSegment {
+        SegmentPtr segment;           // nullptr if the segment was skipped, filtered out, or lost
+        int32_t segment_meta_pos = 0; // position in _metadata->segment_metas()
+    };
```
xiangguangyxg, line 161: could segment_meta_pos be moved to Segment?

banmoy: prefer not currently because segment_meta_pos is only used when loading segments in rowset. Could do it in the future if we find it's needed in more places

CelerData-Reviewer:
@codex review

meegoo (approved):

PR #75207, opened 2026-06-23
[BugFix] Fix pre-1970 Parquet timestamp load corrupting sub-second DATETIME

xiangguangyxg:
## Why I'm doing:

When loading a Parquet `INT64` column annotated `TIMESTAMP` (`isAdjustedToUTC=false`) into a StarRocks `DATETIME`, a **pre-1970 value with a nonzero sub-second part** was decoded to a corrupt garbage value instead of the real wall clock.

`Int64ToDateTimeConverter::convert` splits the signed epoch tick with C++ truncating division:

```cpp
int64_t seconds     = src_data[i] / _second_mask;
int64_t nanoseconds = (src_data[i] % _second_mask) * _scale_to_nano_factor;
```

For a negative tick whose sub-second remainder is nonzero, `nanoseconds` is negative. `timestamp::of_epoch_second` then packs the result via a bitwise OR (`from_julian_and_time`), so the negative microsecond corrupts the packed Julian field — e.g. `1969-12-31 23:59:59.500` loaded as a year-41222 garbage value. Whole-second negatives and all post-1970 values were unaffected.

## What I'm doing:

Borrow a whole second when the sub-second remainder is negative, so `nanoseconds` stays in `[0, NANOSECS_PER_SEC)` — the floor split the FE boundary computation already uses (`Math.floorDiv`/`Math.floorMod`). `of_epoch_second` then receives a non-negative sub-second and packs the correct value. The borrow is unit-agnostic (MILLIS/MICROS/NANOS) and runs before the UTC whole-second offset, so it composes with the timezone-adjusted branch unchanged.

Added a regression test (`Int64PreEpochTimestampSubSecond`) that loads a Parquet file holding `1969-12-31 23:59:59.500000` in both MILLIS and MICROS units: it decoded to a garbage value before the fix and to the correct wall clock after. The existing post-1970 `Int64_2_Timestamp` test is unchanged (14/14 `ColumnConverterTest` pass).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

xiangguangyxg:
@codex review

kevincai (approved):

srlch (approved):

PR #75246, opened 2026-06-24
[BugFix] Fix nested dict expr exchange translation

dontknow9179:
## Why I'm doing:

Nested derived dictionary expressions can be rebuilt differently across exchange fragments when the expression chain is flattened to the base string column.

For null-sensitive expressions such as nested `CASE`/`IF`, the producer may build a derived dictionary through an intermediate dict expression, while the consumer rebuilds the same output slot from the original string column. With `NULL` inputs, those two paths can produce incompatible dictionary translations and cause runtime dict decode failures such as `Dict Decode failed, Dict can't take cover all key`.

For example:

```sql
SELECT CASE WHEN subq.wip_status IS NULL THEN '-' ELSE subq.wip_status END AS wip_status
FROM (
    SELECT DISTINCT CASE WHEN lot_state IN ('created') THEN 'Created' ELSE 'WIP' END AS wip_status
    FROM t_wip_lot
) subq;
```

The derived slots are roughly:

```text
slot13 = CASE WHEN lot_state IN ('created') THEN 'Created' ELSE 'WIP' END
slot14 = IF(slot13 IS NULL, '-', slot13)
```

If the producer keeps the intermediate dict expression, it builds `slot14` from `slot13`:

```text
slot13 code 0 -> NULL      -> IF(NULL IS NULL, '-', NULL)           -> '-'       // enters slot14 dict
slot13 code 2 -> 'Created' -> IF('Created' IS NULL, '-', 'Created') -> 'Created'
slot13 code 3 -> 'WIP'     -> IF('WIP' IS NULL, '-', 'WIP')         -> 'WIP'
```

If the consumer flattens `slot14` to the base `lot_state` dictionary, the synthetic NULL takes a different path:

```text
lot_state code 0 -> NULL      -> CASE(NULL)      -> 'WIP'     -> IF('WIP' IS NULL, '-', 'WIP')         -> 'WIP'
lot_state code 7 -> 'created' -> CASE('created') -> 'Created' -> IF('Created' IS NULL, '-', 'Created') -> 'Created'
```

So the producer-side `slot14` dictionary can contain `'-'`, while the consumer-side flattened dictionary does not. That mismatch makes cross-fragment dictionary translation fail. The numeric codes above are illustrative; the important part is that the intermediate dictionary's synthetic NULL is evaluated by different expression chains.

## What I'm doing:

- Prevent null-sensitive nested global dictionary expressions from being flattened to the base string column.
- Preserve dependent derived dictionary expressions across exchange boundaries so both producer and consumer build the same dictionary chain.
- Fix verbose explain fallback printing for missing dict-expression dependencies so nested dict exprs show their source slot clearly.
- Add FE plan coverage and an end-to-end SQL regression for nested `CASE` dictionary expressions across exchange.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5


Wenjun7J:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lowcardinality/DecodeContext.java
```diff
@@ -63,6 +66,13 @@
  *  e. Generate the global dictionary expression
  */
 class DecodeContext {
+    private static final Set<String> NULL_TOLERANT_LOW_CARD_FUNCTIONS = ImmutableSet.of(
+            FunctionSet.IF,
+            FunctionSet.IFNULL,
+            FunctionSet.COALESCE,
+            FunctionSet.NULLIF,
+            FunctionSet.CONCAT_WS);
```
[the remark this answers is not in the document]
dontknow9179: Unlike CONCAT_WS, CONCAT is null-propagating: concat(NULL, 'x') returns NULL, and NULL is represented by the fixed null code rather than as a regular global-dictionary entry. So, evaluating CONCAT over the intermediate dictionary's synthetic NULL should not introduce an extra producer-only dictionary key.

satanson (dismissed):

fe/fe-core/src/main/java/com/starrocks/sql/plan/PlanFragmentBuilder.java
```diff
@@ -1258,8 +1258,8 @@ private static Map<Integer, Expr> getGlobalDictsExprs(Map<Integer, ScalarOperato
             if (!nodeRefs.keySet().containsAll(columnRefs)) {
                 nodeRefs = Maps.newHashMap(nodeRefs);
                 for (ColumnRefOperator f : columnRefs) {
-                    nodeRefs.computeIfAbsent(f, k -> new SlotRef(f.getName(),
-                            new SlotDescriptor(new SlotId(f.getId()), f.getName(), f.getType(), true)));
+                    nodeRefs.computeIfAbsent(f, k -> new SlotRef(f.toString(),
+                            new SlotDescriptor(new SlotId(f.getId()), f.getName(), f.getType(), f.isNullable())));
```
stdpain: Why we need change this ?

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/lowcardinality/DecodeRewriter.java
```diff
@@ -486,33 +488,11 @@ public OptExpression visitPhysicalDistribution(OptExpression optExpression, Colu
...
-                        dictMap.put(context.stringRefToDictRefMap.get(useRefs).getId(),
-                                context.stringRefToDicts.get(useRefs.getId()));
-                    }
-                }
-            }
+        ColumnRefSet inputColumns = expandStructUseDictExprs(info.inputStringColumns);
+        ColumnRefSet usedDictExprs = expandStructUseDictExprs(fragmentUseDictExprs);
+        usedDictExprs.union(inputColumns);
+        for (int sid : usedDictExprs.getColumnIds()) {
+            collectBaseDictForStringRef(sid, dictMap);
```
stdpain: Isn't our goal to avoid merging expressions? Why are we changing these?

CelerData-Reviewer:
@codex review

stdpain (approved):

satanson (approved):

PR #75254, opened 2026-06-24
[BugFix] Fix NPE in OlapScanNode.getBucketNums when selected partition is concurrently removed

singer-bin:
## Why I'm doing:

`OlapScanNode.getBucketNums()` dereferences `olapTable.getPartition(pid)` without a null check on the HASH-distribution path:

```java
for (Long pid : getSelectedPartitionIds()) {
    bucketNum = olapTable.getPartition(pid).getDistributionInfo().getBucketNum(); // NPE if getPartition() == null
}
```

The selected partition ids are captured at **planning** time. When the scheduler later builds the colocate / local-bucket-shuffle assignment (`ExecutionFragment.getOrCreateColocatedAssignment` -> `ColocatedBackendSelector.Assignment`), it calls `getBucketNums()`. If a concurrent **INSERT OVERWRITE** has swapped/removed that partition in between (INSERT OVERWRITE atomically replaces partitions, so the old partition id no longer exists on the live `OlapTable`), `getPartition(pid)` returns `null` and the coordinator preprocessing crashes with an NPE. The query is then cancelled.

**Steps to reproduce (observed on a customer cluster, 3.3.x):**
- A bucketed/colocate-eligible table is queried by a multi-join SQL whose predicate prunes to a single partition (e.g. `WHERE year=2023 AND term='...'`).
- The same table is being rewritten by a periodic `INSERT OVERWRITE`.
- When the query's scheduling phase races with the partition swap, it fails intermittently.

**Stack trace (3.3.x line numbers):**
```
java.lang.NullPointerException: null
    at com.starrocks.qe.ColocatedBackendSelector$Assignment.<init>(ColocatedBackendSelector.java:186)
    at com.starrocks.qe.scheduler.dag.ExecutionFragment.getOrCreateColocatedAssignment(ExecutionFragment.java:170)
    at com.starrocks.qe.scheduler.assignment.BackendSelectorFactory.create(BackendSelectorFactory.java:82)
    at com.starrocks.qe.scheduler.assignment.LocalFragmentAssignmentStrategy.assignScanRangesToWorker(...)
    at com.starrocks.qe.CoordinatorPreprocessor.computeFragmentInstances(CoordinatorPreprocessor.java:248)
    ...
    at com.starrocks.qe.StmtExecutor.execute(StmtExecutor.java:646)
```

**Frequency:** intermittent — only when query scheduling races with the partition swap of the pruned partition.

Note: setting `disable_colocate_join = true` does **not** avoid this. With colocate disabled the planner falls back to a **local bucket-shuffle join** (`ChildOutputPropertyGuarantor.transToBucketShuffleJoin`), which goes through the exact same `getOrCreateColocatedAssignment` -> `Assignment` path and hits the same NPE.

## What I'm doing:

Add a null guard in `OlapScanNode.getBucketNums()`: if the selected partition was concurrently removed, fall back to the table's default distribution bucket number instead of dereferencing `null`. This turns a hard coordinator crash into graceful behavior.

**Scope / honesty note:** this is a **defensive fix at the symptom layer**, not a root-cause fix. The underlying issue is that the live `OlapTable` referenced by an in-flight query can have its partition metadata mutated by a concurrent `INSERT OVERWRITE` (the query does not operate on a fully consistent partition snapshot for this code path). The same root cause has surfaced at other call sites of `olapTable.getPartition(id)` in the scheduler/statistics paths. Fixing the snapshot/locking model is a larger change tracked separately; this PR removes one concrete, customer-impacting crash.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
@@ -337,7 +337,16 @@ public class OlapScanNode extends AbstractOlapTableScanNode {
         int bucketNum = distInfo.getBucketNum();
         if (getSelectedPartitionIds().size() <= 1) {
             for (Long pid : getSelectedPartitionIds()) {
-                bucketNum = olapTable.getPartition(pid).getDistributionInfo().getBucketNum();
+                // The selected partition id is captured at planning time. It may be concurrently
+                // removed (e.g. by INSERT OVERWRITE swapping partitions) before the scheduler
+                // builds the colocate/local-bucket-shuffle assignment, in which case
+                // getPartition() returns null. Fall back to the table's default bucket number
+                // instead of throwing NPE, so the query fails gracefully (or succeeds) rather
+                // than crashing the coordinator preprocessing.
+                Partition partition = olapTable.getPartition(pid);
+                if (partition != null) {
+                    bucketNum = partition.getDistributionInfo().getBucketNum();
+                }
             }
         }
         return bucketNum;
--- a/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
@@ -340,14 +340,23 @@ public class OlapScanNode extends AbstractOlapTableScanNode {
                 // The selected partition id is captured at planning time. It may be concurrently
                 // removed (e.g. by INSERT OVERWRITE swapping partitions) before the scheduler
                 // builds the colocate/local-bucket-shuffle assignment, in which case
-                // getPartition() returns null. Fall back to the table's default bucket number
-                // instead of throwing NPE, so the query fails gracefully (or succeeds) rather
-                // than crashing the coordinator preprocessing.
+                // getPartition() returns null. Keep the table default here and fall back to the
+                // actual scan-range bucket sequences below, instead of throwing NPE and crashing
+                // the coordinator preprocessing.
                 Partition partition = olapTable.getPartition(pid);
                 if (partition != null) {
                     bucketNum = partition.getDistributionInfo().getBucketNum();
                 }
             }
+        }
+        // The table default can be 0 for inferred distributions (e.g. DISTRIBUTED BY HASH(k)
+        // without an explicit BUCKETS, where inferDistribution() only fills the partition copy).
+        // If the selected partition was concurrently removed we would otherwise return 0, and
+        // ExecutionFragment.getBucketSeqAssignment() would allocate zero-length arrays and fail
+        // with "bucketSeq exceeds bucketNum". Derive a non-zero count from the bucket sequences
+        // already materialized in the scan ranges.
+        if (bucketNum <= 0 && !bucketSeq2locations.isEmpty()) {
+            bucketNum = Collections.max(bucketSeq2locations.keySet()) + 1;
         }
         return bucketNum;
     }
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
```diff
@@ -337,7 +337,16 @@ public int getBucketNums() {
         int bucketNum = distInfo.getBucketNum();
         if (getSelectedPartitionIds().size() <= 1) {
             for (Long pid : getSelectedPartitionIds()) {
-                bucketNum = olapTable.getPartition(pid).getDistributionInfo().getBucketNum();
```
satanson, line 340: it seems a robust way is to deep copy such distribution info before entering planer

singer-bin: You're right - to be upfront, this PR is a defensive guard at the symptom layer, not a root-cause fix. The real problem is that the in-flight query reads the *live* `OlapTable`: `getSelectedPartitionIds()` is captured at planning time, but a concurrent `INSERT OVERWRITE` swaps/removes those partitions before the scheduler builds the colocate/bucket-shuffle assignment, so `getPartition(pid)` returns null. Deep-copying the distribution info (or snapshotting the selected partitions' metadata) before entering the planner is indeed the robust fix and would also cover the other `getPartition()` call sites that share this race.

I've opened #75256 to track the root-cause snapshot work. Could you advise which direction you'd prefer for this PR: (a) keep a hardened guard here as an interim while the snapshot fix is tracked in #75256, or (b) pivot this PR to do the distribution-info snapshot directly? Happy to go either way.

fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
```diff
@@ -337,7 +337,16 @@ public int getBucketNums() {
...
+                // The selected partition id is captured at planning time. It may be concurrently
+                // removed (e.g. by INSERT OVERWRITE swapping partitions) before the scheduler
+                // builds the colocate/local-bucket-shuffle assignment, in which case
+                // getPartition() returns null. Fall back to the table's default bucket number
+                // instead of throwing NPE, so the query fails gracefully (or succeeds) rather
+                // than crashing the coordinator preprocessing.
+                Partition partition = olapTable.getPartition(pid);
+                if (partition != null) {
+                    bucketNum = partition.getDistributionInfo().getBucketNum();
+                }
```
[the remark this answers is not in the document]
singer-bin: Good catch - the table default can be 0 for inferred distributions, which would later surface as `bucketSeq exceeds bucketNum`. Fixed in the latest commit: when the default is 0 (selected partition concurrently removed), `getBucketNums()` now derives a non-zero count from the bucket sequences already materialized in the scan ranges (`bucketSeq2locations`) instead of returning the table default. Added a unit test covering the inferred-distribution (default 0) case.

satanson (approved):

singer-bin:
hi @stdpain gentle ping on this one when you get a chance — already got an approval from @satanson and CI is green. Would love your eyes on it so we can move it forward. Thanks!

stdpain:
This may not be the result we want. In fact, it leads to an unexpected result. @kangkaisen Could u help to check ?

PR #75257, opened 2026-06-24
[BugFix] Fix use-after-free of PipelineObserver in global runtime filter receive path

spaces-X:
## Why I'm doing:

A global RuntimeFilter delivered asynchronously on the `RuntimeFilterWorker` thread can dereference a **destructed `PipelineObserver`**, crashing the BE/CN process:

```
*** SIGSEGV received by PID ... ; stack trace: ***
PC: RuntimeFilterProbeDescriptor::set_runtime_filter(RuntimeFilter const*)
    RuntimeFilterProbeDescriptor::set_shared_runtime_filter(...)
    RuntimeFilterPort::receive_shared_runtime_filter(...)
    RuntimeFilterWorker::_receive_total_runtime_filter(...)
    RuntimeFilterWorker::execute()
```
(observed as a crash inside `set_runtime_filter` / `PipelineObserver::source_trigger -> _do_update`, signal address `@0x0` or a small offset.)

### Root cause
During `PipelineDriver::prepare()` the driver registers a ready-observer callback that **captures its own `PipelineObserver*`** into every global-RF `RuntimeFilterProbeDescriptor`:

```cpp
auto* observer = this->observer();
desc->add_observer(_runtime_state, [observer]() { observer->source_trigger(); });
```

The descriptor lives in the fragment's object pool and **outlives the driver**, but the callback is **never detached** when the driver is torn down. When a query finishes, the driver (and its `PipelineObserver`, owned via `unique_ptr`) is destroyed while the descriptor keeps the stale callback. A late RF delivered on the `RuntimeFilterWorker` thread then iterates `_ready_observers` and invokes the callback over the freed observer → use-after-free.

The existing `fragment_prepared()` guard only covers the *prepare* side — it is a one-way latch that stays `true` after prepare — so it does **not** protect this *teardown* window.

This is reliably reproducible under concurrent query finish + late global RF delivery, independent of host load, and is not OOM/liveness related.

## What I'm doing:

- Key each ready-observer by its `owner` (the `PipelineObserver*`) and add `RuntimeFilterProbeDescriptor::remove_observer(owner)`.
- `PipelineDriver::finalize()` detaches its observer from every global-RF descriptor before teardown.
- Guard `_ready_observers` with a mutex: `set_runtime_filter()` notifies on the `RuntimeFilterWorker` thread while `remove_observer()` runs on the driver-executor thread; the lock serializes them, so an in-flight notify keeps the observer alive until it finishes, and any notify after detach sees no stale callback.

The matching pipeline timer observer is already removed via the global-RF timer `unschedule` in `finalize()`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport PR

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

> Affected since the event-scheduler runtime-filter integration (#54691). 3.3.x does not have the event scheduler observer path and is not affected.

### DIFF
--- a/be/src/exec/runtime/pipeline_driver.cpp
+++ b/be/src/exec/runtime/pipeline_driver.cpp
@@ -233,7 +233,10 @@ Status PipelineDriver::prepare(RuntimeState* runtime_state) {
 #endif
                     auto* observer = this->observer();
                     DCHECK(observer != nullptr || !runtime_state->enable_event_scheduler());
-                    desc->add_observer(_runtime_state, [observer]() { observer->source_trigger(); });
+                    // Key the callback by `observer` so it can be detached in finalize(); the
+                    // descriptor outlives this driver, so a late runtime filter must not invoke a
+                    // callback capturing the destructed observer (use-after-free).
+                    desc->add_observer(_runtime_state, observer, [observer]() { observer->source_trigger(); });
                 }
             }
 
@@ -860,6 +863,20 @@ void PipelineDriver::_try_to_release_buffer(RuntimeState* state, size_t operator
 }
 
 void PipelineDriver::finalize(RuntimeState* runtime_state, DriverState state) {
+    // Global runtime filters are delivered asynchronously on the RuntimeFilterWorker thread and
+    // the probe descriptors outlive this driver (they live in the fragment's object pool). This
+    // driver registered a callback capturing its PipelineObserver into each descriptor during
+    // prepare(); detach them before teardown, otherwise a late runtime filter would invoke that
+    // callback over the destructed observer (use-after-free, crashing in set_runtime_filter ->
+    // PipelineObserver::source_trigger -> _do_update). The fragment_prepared() guard only covers
+    // the prepare side and stays true once set, so it does not protect this teardown window. The
+    // matching timer observer is removed via the global RF timer unschedule below.
+    if (runtime_state->enable_event_scheduler()) {
+        for (auto* desc : _global_rf_descriptors) {
+            desc->remove_observer(observer());
+        }
+    }
+
     stop_timers();
     int64_t time_spent = 0;
     // The driver may be destructed after finalizing, so use a temporal driver to record
--- a/be/src/exec/runtime_filter/runtime_filter_probe.cpp
+++ b/be/src/exec/runtime_filter/runtime_filter_probe.cpp
@@ -147,10 +147,18 @@ std::string RuntimeFilterProbeDescriptor::debug_string() const {
 
 RuntimeFilterProbeDescriptor::~RuntimeFilterProbeDescriptor() = default;
 
-void RuntimeFilterProbeDescriptor::add_observer(RuntimeState* state, ReadyObserver observer) {
+void RuntimeFilterProbeDescriptor::add_observer(RuntimeState* state, const void* owner, ReadyObserver observer) {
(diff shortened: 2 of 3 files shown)


CelerData-Reviewer:
@codex review

alvin-phoenix-ai:
@codex review

stdpain:
In this path, the fragment ctx has not been released. Therefore, the pipeline driver has not been destroyed either. Can you provide a definitive conclusion?
```
    // capture shared ptr here
    query_ctx = runtime_services->query_context_mgr->get(query_id);

    auto& probe_finst_ids = params.probe_finst_ids();
    for (const auto& pb_finst_id : probe_finst_ids) {
        auto fragment_ctx = query_ctx->fragment_mgr()->get(finst_id);
        fragment_ctx = query_ctx->fragment_mgr()->get(finst_id);
        // pipeline driver and observer still live here.
        fragment_ctx->runtime_filter_port()->receive_shared_runtime_filter(params.filter_id(), shared_rf);
```

spaces-X:
> In this path, the fragment ctx has not been released. Therefore, the pipeline driver has not been destroyed either. Can you provide a definitive conclusion?
> 
> ```
>     // capture shared ptr here
>     query_ctx = runtime_services->query_context_mgr->get(query_id);
> 
>     auto& probe_finst_ids = params.probe_finst_ids();
>     for (const auto& pb_finst_id : probe_finst_ids) {
>         auto fragment_ctx = query_ctx->fragment_mgr()->get(finst_id);
>         fragment_ctx = query_ctx->fragment_mgr()->get(finst_id);
>         // pipeline driver and observer still live here.
>         fragment_ctx->runtime_filter_port()->receive_shared_runtime_filter(params.filter_id(), shared_rf);
> ```

@stdpain You're right, thanks for catching this — I owe a correction.

I re-traced the ownership chain: on this path _receive_total_runtime_filter holds query_ctx and fragment_ctx as shared_ptr, and fragment_ctx → Pipeline (vector<DriverPtr>) → PipelineDriver → _observer is a strong chain that's only released in ~FragmentContext. So the driver and its observer are indeed alive throughout receive_shared_runtime_filter. This is not a use-after-free — my original framing was wrong.

The actual issue is a data race on the unsynchronized observer container: set_runtime_filter() (on the RuntimeFilterWorker thread) iterates the observer list while add_observer() (driver prepare()) is concurrently push_back-ing into it, so the reader can walk a reallocated/stale buffer and pick up a garbage pointer (hence the @0x0 / small-offset fault on a live object). The window is opened when set_fragment_prepared(true) is set before the observers are registered.

And this is already fixed on main by #65423 ("Correctly set prepared flag in fragment execution"), which moves set_fragment_prepared(true) out of prepare() to the end of execute() after the drivers are submitted. So main/branch-3.5/4.0 are not affected.

What actually happened on our side: an internal 3.5 fork hit a backport-ordering problem — the #65200 backport got re-applied on top of the #65423 backport, which re-introduced the premature set_fragment_prepared(true) and silently reverted the fix. That regression is what we saw in production, and I mistakenly took it for an unfixed upstream bug.

So this PR is redundant against main — I'll close it. Sorry for the noise, and thanks again for the sharp review.

PR #75255, opened 2026-06-24
[BugFix] Surface isolation property in SHOW FUNCTIONS output

wyb:
## Why I'm doing:

  `ScalarFunction.getProperties()` and `AggregateFunction.getProperties()`,
  which build the Properties column of `SHOW [FULL] FUNCTIONS`, only emitted
  fid/file/md5/symbol/type and never the isolation flag. A UDF/UDAF
  created with "isolation" = "shared" therefore showed no isolation entry,
  so users could not tell whether the property was set.

## What I'm doing:

  Always emit "isolation" ("shared" or "isolated") in the properties JSON,
  computed from the persisted isolationType field, so the state is visible
  for any function without recreating it.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [x] Yes, this PR will result in a change in behavior.
- [ ] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [x] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

PR #75276, opened 2026-06-24
[BugFix] Surface isolation property in SHOW FUNCTIONS output (backport #75255)

wyb:
## Why I'm doing:

Backport of #75255 to branch-3.5.

When a UDF/UDAF is created with `"isolation" = "shared"`, the `isolation` setting was not shown in the `SHOW [FULL] FUNCTIONS` Properties column, so users could not tell whether a function was created with shared isolation.

## What I'm doing:

Surface the `isolation` entry in `Function#getProperties()` (the source of the `SHOW [FULL] FUNCTIONS` Properties column) for scalar and aggregate Java UDFs — emitting `"shared"` when the function is shared and `"isolated"` otherwise. The value is computed from the persisted `isolationType` field, so existing functions display correctly after upgrade without needing to be recreated.

Backport notes (differences from the original #75255 on main):
- Dropped the `synthesizePropertiesFromFields` changes — that helper belongs to `SHOW CREATE FUNCTION` (#73519), which is not present on this branch.
- The regression test is written as a direct `Function#getProperties()` unit test, because this branch's `StarRocksAssert.withFunction` test helper does not apply the `isolation` property (it constructs a mock function with default isolation), so an end-to-end `SHOW FULL FUNCTIONS` assertion cannot exercise the shared path here.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/AggregateFunction.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/AggregateFunction.java
@@ -470,6 +470,10 @@ public class AggregateFunction extends Function {
         properties.put(CreateFunctionStmt.MD5_CHECKSUM, checksum);
         properties.put(CreateFunctionStmt.SYMBOL_KEY, symbolName == null ? "" : symbolName);
         properties.put(CreateFunctionStmt.TYPE_KEY, getBinaryType().name());
+        // isolationType defaults to true (isolated); surface it so users can tell whether the
+        // function was created with isolation = "shared".
+        properties.put(CreateFunctionStmt.ISOLATION_KEY,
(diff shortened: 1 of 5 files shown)


satanson (approved):

wyb:
https://github.com/Mergifyio backport branch-3.5.19

trueeyu:

PR #75280, opened 2026-06-24
[Enhancement] Compute Iceberg foreground row-count from manifests without enumerating files

dirtysalt:
## Why I'm doing:
The CBO front-end statistics entry `IcebergMetadata#getTableStatistics()` always called `triggerIcebergPlanFilesIfNeeded()`, which runs a full `planFiles()` (opens every matching data manifest and deserializes every DataFile entry, O(files)) just to obtain a row count.

This has two costs:
- FE planning is blocked on a full O(files) plan even when only a row count is needed — expensive and memory-heavy for million-file tables.
- It eagerly materializes the per-query `splitTasks` cache, which forces the execution-time `getRemoteFilesAsync()` path to replay one giant pre-built list instead of streaming, effectively defeating incremental scan range delivery.

## What I'm doing:
`getTableStatistics()` is restructured into three explicit paths, each with a clear accuracy/cost trade-off:

```text
                ┌── Path A: column-stats enabled (opt-in, default OFF)
                │   → planFiles ⟶ per-column bounds + Puffin NDV
                │   Cost: O(files) + per-column metadata. Accuracy: best (NDV + min/max).
                │
getTableStatistics
                ├── Path B: incremental delta + fallback, column-stats off
                │   → planFiles via IncrementalAppendScan ⟶ delta-only/exact cardinality
                │   Cost: O(delta-files), cheap. Accuracy: exact.
                │
                └── Path C: whole-table snapshot (default, column-stats off, non-delta)
                    → filterManifests (O(manifests), in-memory) ⟶ sum addedRows+existingRows
                    Cost: O(manifests), ZERO manifest-file opens, ZERO DataFile enumeration.
                    Accuracy: manifest addedRows+existingRows is EXACT for total live files
                    regardless of write style. Partition-aware pruning at manifest-granularity
                    yields a safe over-estimate (typically <5x vs 100x for snapshot total-records).
```

### Key trade-off

| | manifest-pruned (Path C) | planFiles (Path A/B) |
|---|---|---|
| Row count for whole table | **exact** (manifest list = complete live-file catalogue) | exact |
| Row count for delta range | unreliable (MergeAppend/compaction rewrites manifests) | **exact** (added_snapshot_id on DataFile) |
| Partition pruning | manifest-granularity (over-estimate, safe) | file-granularity (precise) |
| File enumeration | **none** | O(matching-files) |
| Incremental delivery | **unlocked** (splitTasks never populated) | defeated (eagerly materializes splitTasks) |

### Manifest row-count guard (`canCountRowsFromManifest`)
Path C checks that every matching manifest carries `addedRowsCount` / `existingRowsCount`
before summing. These fields were added in Iceberg's v2 manifest-list format (~0.10, 2021);
all modern writers (StarRocks, Spark, Trino, Flink) write them, so this guard passes for
essentially all real tables. If it fails (v1 manifest list or pre-0.10 metadata), Path C
transparently falls back to Path B for exact cardinality via planFiles. The guard is
extracted into `canCountRowsFromManifest(ManifestFile)` for explicit readability.

### Benefits
- Foreground statistics cost drops from `O(files) + O(manifest opens)` to `O(1 manifest-list read) + O(manifests)` for the default path.
- Incremental scan range delivery is genuinely unlocked: file planning is deferred to the delivery phase and streamed instead of eagerly materialized in the front-end.

### Limitations
- Partition-pruned row count is a manifest-granularity over-estimate (safe direction for CBO).
- This fully unlocks incremental delivery for the common single-table / `prepareMetadata`-off case. When `enable_parallel_prepare_metadata` is ON (multi-table joins), `prepareMetadata` still eagerly runs `planFiles()` and pre-fills the same `splitTasks` cache key, so delivery is defeated there.

### Follow-up: can we drop `prepareMetadata`?
`prepareMetadata` (gated by `enable_parallel_prepare_metadata`, default OFF) only runs for multi-table joins and exists to parallel-warm connector metadata. It is the last remaining front-end path that eagerly runs `planFiles()` and pre-fills `splitTasks` under the same cache key the delivery path reads.

Two incremental options for a future PR (kept out of this one to limit blast radius):
1. Make `prepareMetadata` warm only manifest-level metadata (or a lazy source) instead of a full file enumeration — keeping the parallel-prefetch benefit without eager materialization.
2. Decouple incremental delivery from the `splitTasks` cache entirely, so `getRemoteFilesAsync` always takes the lazy/streaming iterator regardless of front-end warming; the `splitTasks` cache then only serves the non-incremental sync `getRemoteFiles` path.

Note: `PredicateSearchKey` equality only considers `{db, table, version, predicate, enableColumnStats}` and ignores fieldNames/MOR, which is exactly why stats/prepareMetadata entries collide with the delivery key.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -1676,29 +1676,63 @@ public class IcebergMetadata implements ConnectorMetadata {
             return StatisticsUtils.buildDefaultStatistics(columns.keySet());
         }
 
+        // Default path (column statistics disabled): derive row count from manifest metadata
+        // (O(manifests)) without enumerating DataFiles. This avoids triggering a full planFiles()
+        // in the foreground, so it no longer pre-populates splitTasks and incremental scan range
+        // delivery (getRemoteFilesAsync) stays on its lazy/streaming path.
+        if (!session.getSessionVariable().enableIcebergColumnStatistics()) {
+            try (Timer ignored = Tracers.watchScope(EXTERNAL, "ICEBERG.calculateCardinality")) {
+                long rowCount = getManifestPrunedRowCount(icebergTable, predicate, version);
+                return statisticProvider.buildRowCountStatistics(columns, rowCount);
+            }
+        }
+
         GetRemoteFilesParams params = GetRemoteFilesParams.newBuilder()
                 .setPredicate(predicate)
                 .setLimit(limit)
                 .setTableVersionRange(version)
-                .setEnableColumnStats(session.getSessionVariable().enableIcebergColumnStatistics())
+                .setEnableColumnStats(true)
                 .build();
 
         PredicateSearchKey key = PredicateSearchKey.of(
                 icebergTable.getCatalogDBName(), icebergTable.getCatalogTableName(), params);
         triggerIcebergPlanFilesIfNeeded(key, icebergTable);
+        return statisticProvider.getTableStatistics(icebergTable, columns, session, params);
+    }
 
-        if (!session.getSessionVariable().enableIcebergColumnStatistics()) {
-            List<FileScanTask> icebergScanTasks = splitTasks.get(key);
-            if (icebergScanTasks == null) {
-                throw new StarRocksConnectorException("Missing iceberg split task for table:[{}.{}]. predicate:[{}]",
-                        icebergTable.getCatalogDBName(), icebergTable.getCatalogTableName(), predicate);
-            }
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

Youngwb (approved):

Astralidea (approved):

dirtysalt:
@mergify backport branch-4.0

PR #74323, opened 2026-06-04
[Enhancement] Detect and reassign range-colocate tablets misplaced across PACK shard groups

xiangguangyxg:
## Why I'm doing:

Range-distribution colocate tablets in shared-data mode each belong to a SPREAD shard group (per partition) and one PACK shard group (per colocate range). The originating user-driven Level-1 split leaves one child in the **old** PACK shard group, so after post-publish range reclassification a tablet can be range-aligned yet still sit in the wrong PACK shard group — query correctness is preserved, but compute-side host co-residency for the colocate join is lost on that bucket.

The StarOS shard-group reassignment RPC needed to fix this shipped in `4.2-rc2`. This PR detects the misplaced tablets and migrates them into their expected PACK shard group.

## What I'm doing:

- **Backstop**: once a colocate group is range-aligned, `ColocateChecker` queries each NORMAL-state table's tablets' actual membership and reassigns any misplaced tablet into its expected PACK shard group. The stable flip is **gated on placement repair**: the group is marked stable only after a membership re-read confirms nothing needs repair; a StarOS query/RPC failure, an incomplete response, or a just-issued reassignment keeps it unstable and retried next tick (the checker never revisits a stable group, so marking it stable while mis-placed would leak permanently). `getShardInfo` is chunked, and an incomplete response is treated as an unread membership (never "no groups") to avoid bogus add-only reassignments.
- **Build fix**: bump `staros.version` `4.1-rc4` → `4.2-rc2` in `fe/build.gradle.kts`. The upstream bump (#74873) updated only `fe/pom.xml`, leaving the Gradle build (and Gradle-based CI) on the pre-delta client.

Deferred follow-ups: immediate reassignment inside `SplitTabletJob.applyColocateRangeSplitResult` (latency optimization — the backstop already converges within one checker cycle); gating group stability on actual StarOS placement convergence via `queryShardGroupStable` (F7).

Unit/integration tests in `ColocateCheckerTest`, `ColocateRangeUtilsTest`, `StarOSAgentTest`.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
[...]

PR #75290, opened 2026-06-24
[Enhancement] Gate range-colocate stability on StarOS placement convergence

xiangguangyxg:
## Why I'm doing:

Range-colocate groups in shared-data mode are marked **stable** by `ColocateChecker` once every table is range-aligned and PACK shard-group **membership** has been repaired (#74323). That is sufficient for query **correctness** — `RangeColocateScanDispatch` routes scan ranges by tablet range, not by StarOS worker placement — but it is **not** sufficient for the colocate-join optimization to actually deliver host-local execution: StarOS may still be migrating shards onto co-resident workers in the background (e.g. the transient window after `reassignShardGroups` is issued but before the StarOS `ShardScheduler` finishes placement). Marking the group stable during that window lets the optimizer pick the colocate plan (`RangeDistributionSpec.canColocate` gates on `isGroupUnstable`) while execution silently pays a network hop on the still-migrating bucket, with no plan-level signal.

## What I'm doing:

Add a final placement-convergence gate, on top of the existing range-colocate stability machinery (#74323):

- **`StarOSAgent.queryShardGroupStable(shardGroupIds, workerGroupId)`** — a thin wrapper (modeled on `getShardInfo`) over the StarOS `QueryShardGroupStable` RPC, present in the pinned `staros.version` (`4.2-rc2` on main, `4.1.1` on `branch-4.1`). It returns, positionally aligned with the input ids, whether each PACK shard group has converged onto co-resident workers in the queried worker group.
- **`ColocateChecker.processGroup`** now marks a group stable only when `reconcilePackPlacement(...) && isPlacementConverged(...)`. After membership settles, `isPlacementConverged` queries convergence for every PACK shard group of the colocate group (in bounded batches) and marks stable only when all report converged. The worker group is resolved via the reshard-daemon idiom `WarehouseManager.getBackgroundComputeResource(tableId)` — the default warehouse worker group in the open-source build.
- **Fail-closed**: anything other than an all-converged response keeps the group unstable for the next tick — a `false` for any group (still migrating), a short/size-mismatched response, a `StarClientException` from the query, or a failure resolving the worker group (`getBackgroundComputeResource` throws `ErrorReportException` when the warehouse has no available compute nodes). The whole convergence check is wrapped in one `try/catch` that fails closed on any exception (logging the throwable so the stack trace is preserved). A group that never converges simply stays unstable and the colocate join falls back to a correct shuffle plan — no livelock, no correctness risk. The convergence query is short-circuited (not issued) until membership is settled.

The checker's per-RPC StarOS batch cap (shared by the membership read and the convergence query) is the runtime-mutable `Config.tablet_reshard_colocate_checker_batch_size` (default 1000), so it can be tuned without an FE restart.

Tests: `StarOSAgentTest` covers the wrapper passthrough + exception propagation; `ColocateCheckerTest` covers not-converged / converged (asserting the default worker group is queried) / query-failure / short-response / worker-group resolution failure, plus a regression that convergence is **not** queried while PACK membership is still being repaired.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

This is an internal placement gate: it only tightens *when* a range-colocate group becomes stable, which affects optimizer eligibility for the colocate-join plan (a plan-shape concern with identical SQL results). No SQL syntax, config, session-variable, metric, or external API changes.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
@@ -30,6 +30,7 @@ import com.starrocks.catalog.Table;
 import com.starrocks.catalog.Tablet;
 import com.starrocks.catalog.TabletRange;
 import com.starrocks.catalog.Tuple;
+import com.starrocks.common.Config;
 import com.starrocks.common.DdlException;
 import com.starrocks.common.Range;
 import com.starrocks.common.StarRocksException;
@@ -86,16 +87,6 @@ import java.util.stream.Collectors;
  */
 public class ColocateChecker {
     private static final Logger LOG = LogManager.getLogger(ColocateChecker.class);
-
-    // Cap on tablets per getShardInfo RPC during PACK reconcile. A local safety bound on this
-    // (potentially large) sweep — getShardInfo is not globally chunked, and other callers pass
-    // smaller, bounded lists.
-    private static final int GET_SHARD_INFO_BATCH_SIZE = 1000;
-
-    // Cap on PACK shard group ids per queryShardGroupStable RPC. Colocate ranges (one PACK group
-    // each) accumulate across Level-1 splits, so bound the convergence query like the membership
-    // read above.
-    private static final int QUERY_SHARD_GROUP_STABLE_BATCH_SIZE = 1000;
 
     /**
      * Invoked from {@link TabletReshardJobMgr#runAfterCatalogReady} on every scheduler tick.
@@ -183,7 +174,7 @@ public class ColocateChecker {
     }
 
     /**
-     * Final stability gate (F7): after range alignment and PACK membership repair, asks StarOS whether
+     * Final stability gate: after range alignment and PACK membership repair, asks StarOS whether
      * every PACK shard group of this colocate group has actually converged onto co-resident workers —
      * placement is done, not merely that membership was assigned. Range alignment + membership are
      * sufficient for query correctness, but the colocate-join optimization only pays off when execution
@@ -210,9 +201,10 @@ public class ColocateChecker {
         try {
             long workerGroupId = resolveWorkerGroupId(colocateTableIndex, peers);
             StarOSAgent starOSAgent = GlobalStateMgr.getCurrentState().getStarOSAgent();
-            for (int batchStart = 0; batchStart < packGroupIds.size(); batchStart += QUERY_SHARD_GROUP_STABLE_BATCH_SIZE) {
+            int batchSize = Math.max(1, Config.tablet_reshard_colocate_checker_batch_size);
+            for (int batchStart = 0; batchStart < packGroupIds.size(); batchStart += batchSize) {
                 List<Long> batch = packGroupIds.subList(batchStart,
-                        Math.min(batchStart + QUERY_SHARD_GROUP_STABLE_BATCH_SIZE, packGroupIds.size()));
+                        Math.min(batchStart + batchSize, packGroupIds.size()));
                 List<Boolean> stable = starOSAgent.queryShardGroupStable(batch, workerGroupId);
                 // A size mismatch would let allMatch pass on a subset and flip the group stable while a
                 // group is still migrating — fail closed so placement convergence is never faked.
@@ -336,9 +328,10 @@ public class ColocateChecker {
         List<Long> tabletIds = new ArrayList<>(tabletIdToRange.keySet());
         Map<Long, List<Long>> tabletIdToGroupIds = new HashMap<>();
         // Query membership in bounded batches so one very large table cannot produce an oversized RPC.
-        for (int batchStart = 0; batchStart < tabletIds.size(); batchStart += GET_SHARD_INFO_BATCH_SIZE) {
+        int batchSize = Math.max(1, Config.tablet_reshard_colocate_checker_batch_size);
+        for (int batchStart = 0; batchStart < tabletIds.size(); batchStart += batchSize) {
             List<Long> batch = tabletIds.subList(batchStart,
-                    Math.min(batchStart + GET_SHARD_INFO_BATCH_SIZE, tabletIds.size()));
+                    Math.min(batchStart + batchSize, tabletIds.size()));
             try {
                 for (ShardInfo shardInfo : starOSAgent.getShardInfo(batch, StarOSAgent.DEFAULT_WORKER_GROUP_ID)) {
                     tabletIdToGroupIds.put(shardInfo.getShardId(), shardInfo.getGroupIdsList());
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
@@ -201,7 +201,9 @@ public class ColocateChecker {
         try {
             long workerGroupId = resolveWorkerGroupId(colocateTableIndex, peers);
             StarOSAgent starOSAgent = GlobalStateMgr.getCurrentState().getStarOSAgent();
-            int batchSize = Math.max(1, Config.tablet_reshard_colocate_checker_batch_size);
+            // queryShardGroupStable computes each group's stability server-side, so keep this batch small
+            // to bound per-RPC latency; the full result is assembled across repeated calls.
+            int batchSize = Math.max(1, Config.tablet_reshard_colocate_checker_convergence_batch_size);
             for (int batchStart = 0; batchStart < packGroupIds.size(); batchStart += batchSize) {
                 List<Long> batch = packGroupIds.subList(batchStart,
                         Math.min(batchStart + batchSize, packGroupIds.size()));
@@ -328,7 +330,7 @@ public class ColocateChecker {
         List<Long> tabletIds = new ArrayList<>(tabletIdToRange.keySet());
         Map<Long, List<Long>> tabletIdToGroupIds = new HashMap<>();
         // Query membership in bounded batches so one very large table cannot produce an oversized RPC.
-        int batchSize = Math.max(1, Config.tablet_reshard_colocate_checker_batch_size);
+        int batchSize = Math.max(1, Config.tablet_reshard_colocate_checker_membership_batch_size);
         for (int batchStart = 0; batchStart < tabletIds.size(); batchStart += batchSize) {
             List<Long> batch = tabletIds.subList(batchStart,
                     Math.min(batchStart + batchSize, tabletIds.size()));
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
```diff
@@ -161,20 +166,95 @@ private void processGroup(ColocateTableIndex colocateTableIndex, long colocateGr
...
+     * {@code reconcilePackPlacement} settled, which already proved every member table is a live NORMAL
+     * OlapTable this pass, and {@code getBackgroundComputeResource} resolves the table's warehouse
+     * worker group (table-independent in OSS) — so the first member table is a fine representative.
+     */
+    private long resolveWorkerGroupId(ColocateTableIndex colocateTableIndex,
+                                      List<ColocateTableIndex.GroupId> peers) {
+        for (ColocateTableIndex.GroupId peerGroupId : peers) {
+            List<Long> tableIds = colocateTableIndex.getAllTableIds(peerGroupId);
+            if (!tableIds.isEmpty()) {
+                return GlobalStateMgr.getCurrentState().getWarehouseMgr()
```
starrocks-xupeng, line 249: getBackGroundComputeResource will throw exception, will it be handled by outer layer?

xiangguangyxg: Yes — it was already caught by the outer layer: `runOneCycle` wraps each `processGroup(...)` call in a per-group `try { ... } catch (Exception e) { LOG.warn(...) }`, so an `ErrorReportException` from `getBackgroundComputeResource` (e.g. warehouse has no available compute nodes) is caught, the group is not marked stable, and it's retried next tick — fail-closed, blast radius one group / one tick.

That said, every other external call in this file (`getShardInfo`, `getTable`, `queryShardGroupStable`) is caught *locally* and converted to a fail-closed return, so relying on the distant outer catch here was inconsistent. Fixed in bc37dbcb91b: `isPlacementConverged` now resolves the worker group inside a `try` and returns `false` (group stays unstable, retried next tick) on failure, matching the adjacent `queryShardGroupStable` handling. Added a regression test (`testStaysUnstableWhenWorkerGroupResolutionFails`).

xiangguangyxg: Follow-up: simplified the handling into a single try/catch over the whole convergence check (resolve worker group + batched query) in `isPlacementConverged` — it now catches any exception, fails closed (group stays unstable, retried next tick), and logs the throwable so the stack trace is preserved. Now in commit 4ecc2efdcbb (force-pushed; the earlier bc37dbcb91b was amended into it).

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
```diff
@@ -92,6 +92,11 @@ public class ColocateChecker {
     // smaller, bounded lists.
     private static final int GET_SHARD_INFO_BATCH_SIZE = 1000;
 
+    // Cap on PACK shard group ids per queryShardGroupStable RPC. Colocate ranges (one PACK group
+    // each) accumulate across Level-1 splits, so bound the convergence query like the membership
+    // read above.
+    private static final int QUERY_SHARD_GROUP_STABLE_BATCH_SIZE = 1000;
```
kevincai: this check is expensive, consider a much smaller size and make it mutable in runtime.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
```diff
@@ -161,20 +166,99 @@ private void processGroup(ColocateTableIndex colocateTableIndex, long colocateGr
...
+                    && isPlacementConverged(colocateTableIndex, peers, expectedRanges, colocateGroupId)) {
                 colocateTableIndex.markAllGroupsWithSameColocateGroupIdStable(colocateGroupId, true);
                 LOG.info("marked colocate group id {} stable across {} peer GroupIds",
                         colocateGroupId, peers.size());
             }
         }
     }
 
+    /**
+     * Final stability gate (F7): after range alignment and PACK membership repair, asks StarOS whether
```
kevincai: remove those `F*` from the comments, update your claude memory, remove those markers that is only visible to internal

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
```diff
@@ -161,20 +166,99 @@ private void processGroup(ColocateTableIndex colocateTableIndex, long colocateGr
...
+     *
+     * <p>Queried in bounded batches (PACK groups accumulate as colocate ranges split). Best-effort /
+     * eventually-consistent: a {@code false} for any group (still migrating), a short response, or a
+     * query failure returns {@code false}, leaving the group unstable for the next tick. A group that
+     * never converges simply stays unstable and the colocate join falls back to a correct shuffle plan
+     * — no livelock, no correctness risk.
+     *
+     * @return {@code true} iff every PACK shard group reports placement-converged.
+     */
+    private boolean isPlacementConverged(ColocateTableIndex colocateTableIndex,
```
kevincai, line 191: also consider caching the result in a limited local cache with TTL, so the same shardGroup won't be checked again and again in a short time period.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
xiangguangyxg: Done in 8f08c691662 — unified both batch caps (the `getShardInfo` membership read and the `queryShardGroupStable` convergence query) into one runtime-mutable `Config.tablet_reshard_colocate_checker_batch_size` (default 1000), read once per call with a `Math.max(1, ...)` guard.

On "much smaller": the cap bounds ids *per RPC*, and the id count here is one colocate group's PACK-range count (typically a handful), so it's ~1 RPC/tick regardless of the cap — the dominant cost is the per-tick *frequency*, not the batch size. The cap is now tunable at runtime if you'd still like to lower it; the frequency is what the TTL cache (your other comment) would address.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
xiangguangyxg: Done in 8f08c691662 — removed the `(F7)` marker from the `isPlacementConverged` javadoc (internal design-doc label) and scrubbed the `F5`/`F7` references from the PR description too.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
xiangguangyxg: Agreed — this targets the real cost (the per-tick frequency). Deferring the TTL cache to a follow-up to keep this PR focused on the gate plus the now-mutable cap; will add it if perf shows it's needed. When implemented it'll be a *negative* cache (cache only not-converged results, never converged): staros's predicate is best-effort / eventually-consistent and explicitly warns against treating a single `true` as durable, so caching a `true` could trigger a premature stable flip. Happy to fold it into this PR instead if you'd prefer it not deferred.

kevincai: can implement the cache in separate PR, but whether a negative cache? think about it again.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
```diff
@@ -161,20 +157,100 @@ private void processGroup(ColocateTableIndex colocateTableIndex, long colocateGr
...
+        // nodes) and the StarOS query can throw StarClientException. Anything unexpected propagates to
+        // runOneCycle, which logs it with a stack trace rather than masking it as "not converged".
+        try {
+            long workerGroupId = resolveWorkerGroupId(colocateTableIndex, peers);
+            StarOSAgent starOSAgent = GlobalStateMgr.getCurrentState().getStarOSAgent();
+            int batchSize = Math.max(1, Config.tablet_reshard_colocate_checker_batch_size);
+            for (int batchStart = 0; batchStart < packGroupIds.size(); batchStart += batchSize) {
+                List<Long> batch = packGroupIds.subList(batchStart,
+                        Math.min(batchStart + batchSize, packGroupIds.size()));
+                List<Boolean> stable = starOSAgent.queryShardGroupStable(batch, workerGroupId);
```
kevincai, line 210: this is exactly where the expensive cost comes. Each groupId check on server side is pontentially expensive. A 1 ms cost of single group check latency, will result a 1 second latency of batch op with 1000 ids. So if the cost increases to ~20ms, this interface latency can potentially up to 20s. Reduce the batch size is exactly what we want, to make each RPC call in reasonable latency, the entire result can be assembled by repeated calls.

xiangguangyxg: You're right — I was reasoning about total work, not per-RPC latency. Since each group's stability is computed server-side, a 1000-id batch is one long RPC (~1s at 1ms/group, ~20s at 20ms). Fixed in fea930da44e: the convergence query now has its own runtime-mutable cap `tablet_reshard_colocate_checker_convergence_batch_size` (default 64) to bound per-RPC latency, assembling the full result across repeated calls; the cheap `getShardInfo` membership read keeps a separate `tablet_reshard_colocate_checker_membership_batch_size` (default 1000). Both are runtime-tunable — happy to drop the convergence default lower if you'd prefer.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
xiangguangyxg: On negative vs positive cache — thought about it more. I lean negative-only (cache only not-converged results) because F7's stable flip is terminal: the checker never revisits a group once it's marked stable, so the flip should ride on a *fresh* all-true read. A negative cache just throttles the common 'still migrating' polling and, when it does flip, flips on fresh data. Caching converged=true would let the flip ride on a <=TTL-stale true — bounded, and arguably tolerable since the flip is already point-in-time, but it weakens flip-time freshness for no real gain (a converged group exits the loop anyway). I'll finalize this in the separate cache PR; open to your view there.

kevincai (approved):

srlch (approved):

PR #75322, opened 2026-06-25
[BugFix] Fix multi-column NOT IN dropping wrong rows on NULL tuple columns

Hyper-FF:
## Why I'm doing:

A row-value NOT IN such as (a, b) NOT IN (SELECT x, y FROM t) must follow SQL three-valued logic per column: a NULL on either side of a column comparison makes that tuple comparison UNKNOWN (not a definite mismatch), so the outer row must be dropped, not returned. Previously the multi-column case wrongly returned such rows (e.g. (1,1) NOT IN {(1,NULL)} kept (1,1)), and a (NULL,NULL) build row failed to drop anything. Single-column NOT IN was already correct.

Root cause is in the NULL_AWARE_LEFT_ANTI_JOIN execution. When pairing a probe row with the NULL-containing build rows, the executor only inspected the FIRST key column (key_columns[0]), so a NULL in any non-first tuple column was missed and the probe row was treated as "no match". Multi-column build-side null-awareness was effectively unimplemented.

BE: in _probe_from_ht_for_null_aware_anti_join_with_other_conjunct, pair the probe row with every build row that has a NULL in ANY key column, using build_key_nulls (the per-row OR of all key columns' null flags) instead of only key_columns[0]. Single-key builds, which do not materialize build_key_nulls, keep using the existing first-column path.

FE: in QuantifiedApply2JoinRule, keep the NULL_AWARE_LEFT_ANTI_JOIN and its equi keys (so the join still runs as a hash join), and additionally carry the per-column null-aware check (l = r) OR l IS NULL OR r IS NULL (only for nullable columns) as an OTHER join conjunct. This routes the join through the null-aware filter path and lets it keep/drop the null-build pairs per three-valued logic. Non-nullable columns stay as plain equi keys.

The FE rewrite, by example:

  (x, y) NOT IN (SELECT a, b FROM s)
    => NULL_AWARE_LEFT_ANTI_JOIN
         equi keys:      x = a AND y = b                       (drive the hash join)
         OTHER conjunct: (x = a OR x IS NULL OR a IS NULL)
                     AND (y = b OR y IS NULL OR b IS NULL)     (per-column 3VL on null-build pairs)

Add SQLTest test_subquery/test_multi_column_not_in_null covering NULL in any tuple column, 3-column tuples, (NULL,NULL) build rows, empty subquery, positive IN, and NOT NULL columns.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/join/join_hash_map.hpp
+++ b/be/src/exec/join/join_hash_map.hpp
@@ -1702,6 +1702,21 @@ void JoinHashMap<LT, CT, MT>::_probe_from_ht_for_null_aware_anti_join_with_other
                 MATCH_RIGHT_TABLE_ROWS()
                 RETURN_IF_CHUNK_FULL_FOR_NULLAWARE_OTHER_CONJUCTS()
             }
+        } else if (!_table_items->build_key_nulls.empty()) {
+            // Multi-column NULL_AWARE_LEFT_ANTI_JOIN: build_key_nulls[j] is the OR of every key
+            // column's null flag for build row j. Pair the probe row with EVERY build row that has
+            // a NULL in ANY key column (not just key_columns[0]) so the null-aware other-conjunct
+            // can apply per-column three-valued logic. Without this, a NULL in a non-first key
+            // column was missed and a multi-column NOT IN wrongly emitted rows it must drop.
+            const auto& null_array = _table_items->build_key_nulls;
+            size_t j = _probe_state->cur_nullaware_build_index;
+            while (j < _table_items->row_count + 1) {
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

PR #75329, opened 2026-06-25
[BugFix] Preserve key column keyness on MODIFY COLUMN for DUPLICATE/UNIQUE tables

### LINKED ISSUE #74553: MODIFY key column without keyness on DUPLICATE tables, failing with misleading "Can not change aggregation type"

richerduong:
## Why I'm doing:

`ALTER TABLE ... MODIFY COLUMN` on a **key** column of a DUPLICATE or UNIQUE table, without restating the `KEY` keyword, fails with a misleading `Can not change aggregation type` error, even when only the column comment or type is changed. Omitting `KEY` is a very natural thing to do, and the error gives no hint about the real cause.

```sql
CREATE TABLE t (id INT NOT NULL, name VARCHAR(12) NOT NULL, v FLOAT NOT NULL)
DUPLICATE KEY(id, name) DISTRIBUTED BY HASH(id) BUCKETS 1 PROPERTIES ("replication_num" = "1");

-- Change only the comment of the key column `name`, keeping the exact same type:
ALTER TABLE t MODIFY COLUMN name VARCHAR(12) COMMENT 'full name';
-- ERROR 1064 (HY000): Can not change aggregation type

-- Workaround today: restate KEY explicitly
ALTER TABLE t MODIFY COLUMN name VARCHAR(12) KEY COMMENT 'full name';  -- succeeds
```

The same problem reproduces on UNIQUE KEY tables. It does not happen on PRIMARY KEY or AGGREGATE tables (see root cause).

## What I'm doing:

Root cause: in `SchemaChangeHandler.processModifyColumn`, the column rebuilt from the `MODIFY COLUMN` clause is non-key when `KEY` is omitted. The UNIQUE and DUPLICATE branches then assign it an aggregation type (`REPLACE` or `NONE`), silently demoting the existing key column to a value column. That implicitly assigned aggregation type later clashes with the original key column's `null` aggregation type in `Column.checkSchemaChangeAllowed`, which throws `Can not change aggregation type`.

A keyness flip is not a supported `MODIFY COLUMN` operation. The range-distribution guard in the same method already rejects it explicitly. This change mirrors the existing PRIMARY KEY backward-compatibility behavior: when the base column is a key and `KEY` is omitted, preserve its keyness instead of demoting it. PRIMARY KEY and AGGREGATE tables already behave this way.

Adds `ModifyColumnKeynessTest` covering DUPLICATE and UNIQUE key-column modifications without the `KEY` keyword, plus a value-column case to guard against an over-broad fix. The two key-column tests fail with `Can not change aggregation type` on the unpatched code and pass with the fix; the value-column test passes in both cases.

Fixes #74553

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -780,7 +780,18 @@ public class SchemaChangeHandler extends AlterHandler {
                         modColumn.getName());
             }
             if (!modColumn.isKey()) {
-                modColumn.setAggregationType(AggregateType.REPLACE, true);
+                // backward compatibility: if the user does not restate the KEY attribute when modifying an
+                // existing key column (e.g. changing only its comment or type), preserve its keyness instead
+                // of silently demoting it to a value column. A keyness flip is not a supported MODIFY COLUMN
+                // operation (see the range-distribution guard below), and the implicit demotion would otherwise
+                // surface as a misleading "Can not change aggregation type" error. This mirrors the behavior of
+                // the primary key branch above.
+                Column baseColumn = olapTable.getBaseColumn(modColumn.getName());
+                if (baseColumn != null && baseColumn.isKey()) {
+                    modColumn.setIsKey(true);
+                } else {
+                    modColumn.setAggregationType(AggregateType.REPLACE, true);
+                }
             }
         } else {
             if (null != modColumn.getAggregationType()) {
@@ -788,7 +799,15 @@ public class SchemaChangeHandler extends AlterHandler {
                         modColumn.getName());
             }
             if (!modColumn.isKey()) {
-                modColumn.setAggregationType(AggregateType.NONE, true);
+                // backward compatibility: preserve the keyness of an existing key column when the KEY attribute
+                // is omitted, rather than silently demoting it to a value column. See the matching comment in the
+                // unique key branch above.
+                Column baseColumn = olapTable.getBaseColumn(modColumn.getName());
+                if (baseColumn != null && baseColumn.isKey()) {
+                    modColumn.setIsKey(true);
+                } else {
+                    modColumn.setAggregationType(AggregateType.NONE, true);
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


richerduong:
Hi @gengjun-git and @meegoo, could you please review this?

PR #75345, opened 2026-06-25
[Enhancement] Enable tablet pre-split for column-list INSERTs that keep all sort keys

luohaha:
## Why I'm doing:

Sample-Based Tablet Pre-Split silently skips any INSERT that specifies a target column list. `InsertPreSplitHook.passesCommonPreFilters` returns false when `getTargetColumnNames() != null`, so the extremely common shape

```sql
INSERT INTO t (c1, c2, ...) SELECT * FROM FILES(...)
```

never pre-splits — the entire load lands in the single initial range tablet, losing the parallelism pre-split exists to provide. This is a quiet footgun: the same statement without the column list pre-splits fine, and nothing tells the user why one is slow.

## What I'm doing:

Allow a target column list as long as it keeps pre-split well-defined:

- **Source-agnostic gate** (`targetColumnListIncludesAllSortKeys`): the list must name every range-distribution (sort) key column. An omitted key column is written with its default for every row, collapsing the data on that key and making split boundaries degenerate. Non-key columns may be omitted.
- **INSERT-from-FILES**: `schemasAlignForByPositionInsert` is generalized to align the FILES schema against the *effective* target columns (the explicit list when present, else the base schema). With all sort keys present and aligned, the by-name sampler reads exactly the FILES column the by-position load writes into each key — so a full **or partial-but-key-complete** list is correct.
- **INSERT-from-table**: its column mapping (`InsertSelectSourceColumns`) still assumes the full base schema in order, so that path accepts only a full, in-order identity list (`targetColumnListIsFullIdentity`) for now; partial/reordered lists are skipped there (a follow-up can generalize the table mapping).

The blanket `getTargetColumnNames()` reject moves out of `passesCommonPreFilters` (which runs before target resolution) into a resolution-time gate, where the sort-key columns are available.

Tests: the two gates are unit-tested directly (`InsertPreSplitHookColumnListTest`), plus FILES partial-list alignment cases in `InsertPreSplitHookFilesTest`. The two obsolete `testTargetColumnListShortCircuits` cases (premise: any column list skips) are removed.

> Local FE unit tests are not runnable in my environment; relying on CI to execute the added tests.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/FilesPreSplitSource.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/FilesPreSplitSource.java
@@ -34,6 +34,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * INSERT-from-FILES pre-split source. Matches only the strict bare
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg:
@codex review

luohaha:
Thanks @chatgpt-codex-connector — both points addressed. The relaxed gate is now `targetColumnListIsPreSplitSafe`, which validates an explicit list against the resolved target schema **before** dispatching: it rejects duplicate names, unknown/generated names, and any list that omits a required column (no default, not nullable, not auto-increment, not generated — mirroring `InsertAnalyzer`'s "must be explicitly mentioned" rule). So the hook no longer reshards tablet metadata for a statement the analyzer would later reject (e.g. `INSERT INTO t(k)` missing a NOT-NULL `v`, or `INSERT INTO t(k, typo)`). Added unit tests for the unknown/duplicate/missing-required cases.

xiangguangyxg (approved):

sevev (approved):

PR #75351, opened 2026-06-25
[Feature] support alembic muilt MetaData for Starrocks python client

lang07123:
starrocks  alembic  support muilt MetaData

## Why I'm doing:
I have a monorepo project , so that had muilt target_metadata,
but `autogen_context.metadata` type of  `MetaData | Sequence[MetaData] | None` , so when I set target_metadata to ` target_metadata = [WebBase.metadata, SdkBase.metadata, AppsBase.metadata]`, I got an error `AttributeError: 'list' object has no attribute 'tables'`

image

## What I'm doing:
I fix it when set target_metadata to a list.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


jaogoy (dismissed):

jaogoy (commented):
Thanks for the fix. I traced the data flow and the core approach is correct. And a few things worth addressing before merge.

## Suggestions before merge

* **Add a test (main gap)**. This is a [Feature] PR but test/unit/test_compare_views.py / test_compare_mvs.py have a clean pattern that mocks autogen_context.metadata. A case where metadata is a list of 2+ MetaData (each contributing a view/MV) asserting the union is compared would directly guard this.

* **Duplicate-key handling diverges from Alembic core.** Core's table_key_to_table raises ValueError on duplicate table keys across MetaData objects. Here the `*_name_to_table` dict comprehensions silently let the last one win on a duplicate (schema, name). Low risk for the disjoint-monorepo case, but consider mirroring core's check (or at least a warning) so a duplicated view across two bases doesn't silently resolve to one definition.

* **Prefer alembic.util.to_list() over isinstance(x, Sequence)**. It's exactly the normalization core uses, dedups the repeated guard (currently inlined 4×), and avoids edge cases (str is also a Sequence). If you keep isinstance, use collections.abc.Sequence rather than typing.Sequence.

CelerData-Reviewer:
@codex review

lang07123:
> Thanks for the fix. I traced the data flow and the core approach is correct. And a few things worth addressing before merge.
> 
> ## Suggestions before merge
> * **Add a test (main gap)**. This is a [Feature] PR but test/unit/test_compare_views.py / test_compare_mvs.py have a clean pattern that mocks autogen_context.metadata. A case where metadata is a list of 2+ MetaData (each contributing a view/MV) asserting the union is compared would directly guard this.
> * **Duplicate-key handling diverges from Alembic core.** Core's table_key_to_table raises ValueError on duplicate table keys across MetaData objects. Here the `*_name_to_table` dict comprehensions silently let the last one win on a duplicate (schema, name). Low risk for the disjoint-monorepo case, but consider mirroring core's check (or at least a warning) so a duplicated view across two bases doesn't silently resolve to one definition.
> * **Prefer alembic.util.to_list() over isinstance(x, Sequence)**. It's exactly the normalization core uses, dedups the repeated guard (currently inlined 4×), and avoids edge cases (str is also a Sequence). If you keep isinstance, use collections.abc.Sequence rather than typing.Sequence.

sorry , I'm bad at english, but I think you are right. so I changed that. @jaogoy

PR #75357, opened 2026-06-25
[BugFix] Reserve dict size + 1 for NULL-absorbing dict-mapping groupby keys

Hyper-FF:
## Why I'm doing:

A group-by on a low-cardinality dict-mapping column uses the compressed group-by key path, whose key range is sized by ApplyMinMaxStatisticRule as (0, baseDictSize). For a NULL-absorbing mapping expression (IS NULL / ifnull / coalesce / nullif / concat_ws) the dict's NULL slot is mapped to an extra value, so the derived dict can hold baseDictSize + 1 codes. That extra code overflows the packed compressed-key width, the producer emits a code the decode dict cannot cover, and the query fails with "Dict Decode failed, Dict can't take cover all key".

Reserve baseDictSize + 1 for such keys. A single expression maps each input code (including the NULL slot) to at most one output, so the derived dict is always <= baseDictSize + 1, which makes the bound exact. Non NULL-absorbing mappings keep baseDictSize unchanged.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #75382, opened 2026-06-26
[Enhancement] Estimate NDV for Iceberg columns when Puffin stats are unavailable

dirtysalt:
## Why I'm doing:
When querying Iceberg external tables without Puffin statistics files, the CBO optimizer hardcodes `distinctValuesCount = 1` for every column. This causes catastrophically wrong query plans: JOIN selectivity is estimated as 1.0 (full cross-join), and GROUP BY output cardinality is wildly overestimated. The result is that the optimizer routinely picks the wrong join strategy (broadcast vs. shuffle) and misestimates aggregation output size, regardless of the actual data distribution.

## What I'm doing:
Replace the hardcoded `NDV = 1` fallback with a three-tier estimation strategy in `IcebergStatisticProvider.buildColumnStatistic()`. The Puffin path is untouched — this only affects the `else` branch when Puffin is absent.

```
NDV source priority (unchanged):
  1. Puffin sketch  ← no change
  2. [NEW] estimateNdvFallback()
       Tier 1 — range-based (integer/date/timestamp, min/max already in IcebergFileStats → zero extra IO)
                NDV ≤ min(rowCount, max − min + 1)
       Tier 2 — column-size-based (compressed bytes / min-bytes-per-distinct-value)
                NDV ≤ min(rowCount, columnSizeBytes / typeMinBytes)
       Tier 3 — type-fraction last resort (rowCount × fraction, fraction = 0.3 for INT, 0.5 for STRING, 0.01 for BOOL)
```

The key design point for Tier 1: `IcebergFileStats` already carries `minValues`/`maxValues` from DataFile metadata when column stats are enabled. The estimation reads them with zero additional IO — if they're absent (fast path with `IcebergFileStats(recordCount)` only), `getMinValue()` returns empty and the code falls through to Tier 2/3 automatically. No parameter signature changes.

When both Tier 1 and Tier 2 are available, the code takes `min(rangeNdv, sizeNdv)` to stay conservative: range gives a tight upper bound for low-cardinality columns (e.g. a status column with values 1–5 gives NDV ≤ 5), while size-based caps the estimate when the range is too wide to be informative.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/cost/IcebergStatisticProvider.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/cost/IcebergStatisticProvider.java
@@ -344,7 +344,9 @@ public class IcebergStatisticProvider {
         Optional<Double> minVal = icebergStats.getMinValue(fieldId);
         Optional<Double> maxVal = icebergStats.getMaxValue(fieldId);
         if (minVal.isPresent() && maxVal.isPresent() && isRangeEstimableType(fieldId, icebergStats)) {
-            long range = Math.max(0L, (long) (maxVal.get() - minVal.get())) + 1;
+            double diff = maxVal.get() - minVal.get();
+            // Guard against overflow: very wide BIGINT ranges may exceed Long.MAX_VALUE as a double
+            long range = (diff >= (double) Long.MAX_VALUE) ? Long.MAX_VALUE : Math.max(0L, (long) diff) + 1;
             double rangeNdv = Math.min(rowCount, range);
             // Take min with size-based to stay conservative when range >> actual NDV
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

Youngwb (approved):

stephen-shelby (approved):

PR #75390, opened 2026-06-26
[Enhancement] Optimize MERGE INTO planning: predicate propagation, EnforceUnique elision, and anti-skew shuffle

### LINKED ISSUE #73684: Support MERGE INTO statement on Iceberg V2 tables using Merge-On-Read
- [x] https://github.com/StarRocks/starrocks/pull/74725
- [x] https://github.com/StarRocks/starrocks/pull/73707
## Feature request

StarRocks now supports `INSERT`, `DELETE`, and `UPDATE` on Iceberg V2 tables, but `MERGE INTO` — the standard way to express *upserts* and *conditional row mutations against a source feed* in a single atomic statement — is not available. Users who need this pattern (CDC apply, slowly-changing-dimension refresh, batch reconciliation against a staged delta) have to chain `DELETE` + `INSERT` or `UPDATE` + `INSERT`, which (a) splits the operation across statements so readers can observe the partial state, (b) duplicates the join logic each side, and (c) silently does the wrong thing when a source row matches more than one target row — `UPDATE` last-write-wins and `INSERT` may double-insert without any error surface. Today this gap forces users to keep an external engine around solely for `MERGE`.

Support `MERGE INTO` on Iceberg format-V2 tables under the Merge-On-Read model. The write must be atomic and follow the Iceberg V2 row-delta semantics so the result is interoperable with Spark and other Iceberg-aware engines.

```sql
MERGE INTO target [AS t] USING source [AS s] ON <join_cond>
WHEN MATCHED     [AND <cond>] THEN UPDATE SET col = expr, ...
WHEN MATCHED     [AND <cond>] THEN DELETE
WHEN NOT MATCHED [AND <cond>] THEN INSERT (col, ...) VALUES (expr, ...)
WHEN NOT MATCHED [AND <cond>] THEN INSERT *
```

Initial scope:
- `WHEN MATCHED ... THEN UPDATE` / `THEN DELETE` and `WHEN NOT MATCHED ... THEN INSERT (cols) VALUES (...)`, `INSERT VALUES (...)` (positional), `INSERT *` (column-name match with the source).
- Multiple `WHEN MATCHED` / `WHEN NOT MATCHED` branches, each with an optional `AND` predicate; first matching branch wins.
- Duplicate-match detection: if the same target row matches more than one source row and any matching branch is `UPDATE` / `DELETE`, the statement fails at execution time (Iceberg semantics).
- Conflict-detection filter built from the *target-side* `IcebergScanNode`, so self-merge (source and target both reference the same Iceberg table) gets the right filter.
- Source relation: base table, subquery, `VALUES`, or another Iceberg table.
- Reuses the existing UPDATE row-delta sink and commit path (conflict detection + serializable isolation).

Non-goals for this phase: partition column updates; `WHEN NOT MATCHED BY SOURCE`; Iceberg V1 / V3; equality deletes; Copy-On-Write mode; correlated subqueries inside the source.

- Two-statement `DELETE` + `INSERT`: not atomic, readers see the in-between state; cannot express "matched-but-with-AND-condition" routing.
- UPSERT via UDF or stored procedure: not supported on Iceberg and would have to materialize the join twice.
- Keep MERGE in an external Iceberg-aware engine: forces users to operate two engines against the same table.

Merge-On-Read with a single planned join is preferred: it matches the Iceberg V2 spec, is cheap for selective merges, and reuses the existing row-delta machinery instead of growing a parallel commit path.

GavinMar:
## Why I'm doing:

The base MERGE INTO pipeline (#74725) always scans the full target, always runs the runtime EnforceUnique duplicate check, and lets the optimizer prune the synthesized source–target join's shuffle columns. These leave avoidable I/O, an unnecessary bucket-shuffle exchange, and a data-skew risk on the path that feeds the duplicate check and the Iceberg row-delta sink.

## What I'm doing:

Plan-time optimizations on top of the base MERGE INTO pipeline:

- **Predicate propagation** (`MergeIntoPredicateDeriver`): push source-side predicates to the target through ON equalities, so the target scan reads less data. Derivation is skipped across a type-mismatched ON key: the analyzed AST keeps bare `SlotRef`s even where an implicit cast is later inserted, and a non-injective coercion would make the pushed-down predicate unsound.
- **Cardinality elision**: skip the runtime EnforceUnique duplicate check when the source's uniqueness candidates fully cover the ON key set — a provably-unique source matches each target row at most once, so the check is redundant. Coverage is credited only when the ON-key sides are type-equivalent (same reason as above), and a unique key is related to a source output column by `SlotRef` column name; a non-`SlotRef` key (e.g. `GROUP BY f(x)`) is never assumed to prove uniqueness, so the check is conservatively kept. The check sits directly above the merge join and reuses its distribution, so eliding it drops the per-row dedup-tracking node itself; no separate exchange is involved.
- **Anti-skew shuffle preservation**: carry a `preserveShuffleColumns` hint (`JoinRelation` → `LogicalJoinOperator`/`PhysicalJoinOperator`, honored by `PruneShuffleColumnRule`, `DecodeRewriter`, and the HashJoin impl rule) so the multi-column ON-key shuffle that feeds the duplicate check and the row-delta sink is not collapsed to a single, possibly value-skewed column — a stability guard. The generic `SkewShuffleJoinEliminationRule` is left untouched; instead the skew-join broadcast rewrite is disabled for the MERGE plan inside the planner, because it would split the join into a broadcast branch that replicates the target side and break the per-driver co-location the duplicate check relies on (that one is a correctness guard for the dedup path). Duplicate-match correctness otherwise rests on the shuffle join itself.
- **Keep the duplicate check on the join distribution**: place `EnforceUniqueRowLocatorNode` directly above the merge join and disable post-join passthrough (`setCanLocalShuffle(false)`) so each driver's seen-set stays co-located. The MERGE join must also be backend-executable: the BE `exec_factory` builds an operator only for `HASH_JOIN_NODE` and `CROSS_JOIN_NODE`/`NESTLOOP_JOIN_NODE`, so a sort-merge join (`join_implementation_mode=merge`) has no executor and is rejected at the FE with a clear error — on every path, including the elision path that skips the duplicate-match node.

Tested with `PruneShuffleColumnRuleTest`, targeted cases in `MergeIntoPlanTest` and `MergeIntoAnalyzerIcebergTest`, and the `test_iceberg_merge_into` SQL-tester case (unpartitioned + partitioned). Verified end-to-end on a 1 FE + 3 BE cluster.

Fixes #73684

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


GavinMar:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/MergeIntoPlanner.java
```diff
@@ -221,7 +246,172 @@ private void setupIcebergMergeSink(ExecPlan execPlan, List<String> colNames,
...
+        if (sourceRelation instanceof TableRelation tableRelation) {
+            return tableRelation.getName().getTbl();
+        }
+        return null;
+    }
+
+    private static boolean isSlotFromRelation(SlotRef slotRef, String relationName) {
+        return slotRef.getTblName() != null
+                && slotRef.getTblName().getTbl() != null
+                && slotRef.getTblName().getTbl().equalsIgnoreCase(relationName);
```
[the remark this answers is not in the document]
GavinMar: fixed

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/MergeIntoAnalyzer.java
```diff
@@ -282,6 +284,22 @@ private static void analyzeIcebergTable(MergeIntoStmt stmt, IcebergTable iceberg
         // Analyze the query statement
         new QueryAnalyzer(session).analyze(queryStatement);
 
+        // Propagate source-side predicates to the target through type-equivalent ON equalities, now
+        // that analysis has resolved slot types. This is done post-analysis on purpose: the analyzed
+        // AST equality keeps bare SlotRefs even across an implicit cast (e.g. VARCHAR = INT), so a
+        // pre-analysis, name-only rewrite could derive a wrong target predicate; deriveTargetPredicate
+        // only maps across equi-keys whose slot types match. Re-analyze the combined ON predicate so
+        // the new AND node and the derived conjunct get full type/legality state.
+        Expr derivedTargetPredicate = MergeIntoPredicateDeriver.deriveTargetPredicate(
```
Youngwb: why need this, optimizer already have OnPredicateMoveAroundRule

Youngwb:
I think this PR is bundling several largely independent optimizations into one change: source-to-target predicate derivation, EnforceUnique elision based on source uniqueness, shuffle-column pruning control, skew-join interaction, and duplicate-check distribution handling. These have different correctness arguments and regression surfaces, so they should be split into separate PRs.

For each optimization, I also think we should first check whether the existing optimizer framework can express it, or extend the optimizer capability there, instead of adding MERGE-specific parallel logic.

For example, predicate derivation overlaps with `OnPredicateMoveAroundRule` / predicate pushdown. If expression-based propagation across equality predicates is missing, it should preferably be added as a general optimizer capability with the right type, determinism, projection, and outer-join-direction guards, rather than introducing `MergeIntoPredicateDeriver` only for MERGE.

Similarly, preserving multi-column shuffle keys should ideally be modeled as a general optimizer property/rule guard when a downstream operator depends on the distribution shape, not as a MERGE-only special path unless there is a strict MERGE-specific correctness requirement.

Could we split this PR and make each piece either reuse or extend the relevant optimizer rule first?

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/MergeIntoAnalyzer.java
GavinMar: > why need this, optimizer already have OnPredicateMoveAroundRule

Make sense. Compared to the existing `OnPredicateMoveAroundRule`, this only adds support for modulo operations. However, this scenario is limited and can be achieved through subsequent improvements to the general optimizer rules. There is no need to repeat the same logic here.

CelerData-Reviewer:
@codex review

PR #75395, opened 2026-06-26
[Enhancement] Cache Iceberg manifest column statistics selectively

Youngwb:
## Why I'm doing:

The FE Iceberg manifest cache (`CachingIcebergCatalog` data/delete file cache) stores per-file column statistics — lower/upper bounds, value/null/nan counts — for **every** column of every cached data file. On wide tables with many data files these statistics maps dominate the FE heap (a ~1M-file, ~170-column table can hold tens of GB of column-stat maps, while the data-file objects themselves are a negligible fraction). Yet file-level min/max pruning is only effective on **clustered** columns (partition / sort keys); statistics for the remaining columns cost memory without enabling any data skipping.

## What I'm doing:

When manifest column statistics are cached (the existing catalog property `iceberg_manifest_cache_with_column_statistics`, default `true`), keep statistics only for the columns where file-level min/max pruning is effective:

- partition source columns, unioned across **all** partition specs (so partition evolution stays covered);
- sort-order source columns;
- identifier columns (equality-delete correctness).

Statistics for other columns are dropped via `copyWithStats(keepSet)`, collapsing the manifest cache footprint on wide tables while preserving data skipping on the columns that matter. Scans that explicitly request column statistics (`shouldReturnColumnStats()` — CBO column statistics / min-max aggregation) and tables with an empty keep set still cache all columns. No new property is introduced, and `enable_cache_data_file_identifier_column_statistics` and the equality-delete code path are unchanged.

Also corrects the documented default of `iceberg_manifest_cache_with_column_statistics` (the docs said `false`; the code default is `true`) in en/zh/ja.

Fixes #issue: N/A — FE memory optimization, no tracking issue.

## What type of PR is this:

- [ ] BugFix
- [x] Enhancement
- [ ] Feature
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

Youngwb:
@codex review

Youngwb:
@codex review

Youngwb:
Follow-up on the design: selective caching is now keyed on the existing `iceberg_manifest_cache_with_column_statistics` (default `true`) instead of `enable_cache_data_file_identifier_column_statistics` — the latter keeps its original meaning (it is not touched by this PR).

When `iceberg_manifest_cache_with_column_statistics=true`, only the prune-effective keep-set (partition source columns across all specs, sort-order columns, identifier columns) is cached; scans that explicitly request column statistics (`shouldReturnColumnStats()` — CBO column stats / min-max aggregation) and tables with an empty keep set still cache all columns. Docs (en/zh/ja) and the corrected default were updated accordingly.

Youngwb:
@codex review

Youngwb:
Good catch — fixed. The earlier guard in `getIdentifierFieldIds()` only covered cache **misses** (newly read manifests); a column-stats-requesting scan hitting an already-pruned cache entry would still get partial stats, as you note.

Fix: `useCache()` now returns `false` whenever `shouldReturnColumnStats()` is true, so such scans (CBO column-statistics collection and min/max-aggregation pushdown) **bypass the manifest cache entirely** — both hits and misses — and read full per-column stats via the non-cache path (`newManifestGroup().planFiles()`, inherited from Iceberg `DataScan`). That path is already the one these scans take when `iceberg_manifest_cache_with_column_statistics=false`, so it's not a new code path. The now-redundant guard inside `getIdentifierFieldIds()` was removed (it is only reached for non-stats scans now).

One correction on scope: the **metadata-delete conflict check** is not affected — it runs on a separate `nativeTable.newScan().includeColumnStats()` in `IcebergMetadata`, which does not consult the manifest data-file cache. And min/max-aggregation correctness was verified end-to-end on a cluster (selective-cache warm-up followed by `MAX(non_key_col)` returned the exact value; it gracefully falls back to a real scan when a column's bounds are absent).

Youngwb:
@codex review

Youngwb:
Fixed. `statsKeepColumnIds` now unions **all** sort orders via `table.sortOrders().values()` (symmetric to the all-specs handling), so source columns from evolved/historical sort orders stay in the keep set and keep pruning their files from the cache. Added a unit-test case with two sort orders where the older sort key (absent from the current order) is retained.

Youngwb:
@codex review

stephen-shelby (approved):

dirtysalt (approved):
LGTM! 🚀

PR #75403, opened 2026-06-26
[BugFix] Fix small-LIMIT hang on connector tables with incremental scan ranges

### LINKED ISSUE #75296: [Bug] Query with small LIMIT on external/connector table hangs until query_timeout when enable_connector_incremental_scan_ranges=true
Prerequisites: an external catalog table (Hive/Iceberg/Delta, Parquet on object storage) whose scan produces more scan ranges than one incremental batch, and a selective predicate that matches a few rows which are not all in the first delivered batch.

```
-- 1. Make incremental delivery use multiple rounds even for a modest table
--    (default is 500; lowering it makes the repro deterministic on small tables)
SET connector_incremental_scan_ranges_size = 1;
SET enable_connector_incremental_scan_ranges = true;   -- default
-- 2. Selective point-lookup with a small LIMIT over the external table
SELECT *
FROM hive_catalog.db.parquet_tbl
WHERE id = '<a value that matches a few rows>'
LIMIT 1;
```
The query returns the matching row(s) quickly (the LIMIT short-circuit should make it faster, not slower).

### Real behavior (Required)
The query hangs and eventually fails with:

```
ERROR 5024 (53400): Query reached its timeout of <N> seconds, please increase the 'query_timeout' session variable and retry
```
The query never completes regardless of how long query_timeout is. EXPLAIN ANALYZE/profile shows the scan fragment never finishes; scan operators are parked with the morsel queue empty but has_more = true.

### Root cause
With incremental scan ranges, the FE tells each BE scan instance "more ranges may come" via a sentinel (TScanRangeParams.empty=true, has_more=true), and ScanOperator::is_finished() keeps the operator alive while _morsel_queue->has_more() is true. The terminal has_more=false sentinel is only stamped inside HDFSBackendSelector.computeScanRangeAssignment, which DefaultCoordinator.assignIncrementalScanRangesToDeployStates calls only while scanNode.hasMoreScanRanges() is true.

When a scan reaches its limit before the source is drained, hasMoreScanRanges() (= !reachLimit && scanRangeSource.hasMoreOutput()) flips to false via the reachLimit path (BE-reported closed_scan_nodes). The coordinator then stops delivering without ever broadcasting the terminal has_more=false sentinel to the still-parked instances, so those scan operators wait forever. (Source-exhaustion does not hit this because the last real round stamps has_more=false itself.)

This was introduced by the limit short-circuit added in #62188 (backport #62257, commit bd48411e3ec), which layered reach_limit / closed_scan_nodes on top of incremental scan ranges (#50189). Before that PR, delivery always ran to source exhaustion, so the terminal sentinel was always sent.

### Suggested fix

FE: when incremental delivery stops because the scan reached its limit (not source-exhaustion), still flush the terminal has_more=false sentinel to the fragment's instances exactly once (in DefaultCoordinator.assignIncrementalScanRangesToDeployStates).
BE (defense-in-depth): once _morsel_queue->reach_limit() is set, ScanOperator::is_finished() should stop blocking on has_more().

### Workaround: 
```
SET [GLOBAL] enable_connector_incremental_scan_ranges = false;
```

 - 4.0.11

xhumanoid:
## Why I'm doing:

introduce additional checks for scan terminations

## What I'm doing:

Fixes #75296

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/pipeline/fragment_executor.cpp
+++ b/be/src/exec/pipeline/fragment_executor.cpp
@@ -1104,8 +1104,10 @@ Status FragmentExecutor::append_incremental_scan_ranges(ExecEnv* exec_env, const
     }
     FragmentContextPtr fragment_ctx = query_ctx->fragment_mgr()->get(instance_id);
     if (fragment_ctx == nullptr) {
-        return Status::InternalError(fmt::format("FragmentContext not found for query_id: {}, instance_id: {}",
-                                                 print_id(query_id), print_id(instance_id)));
+        // The fragment instance may have already finished (e.g. its scan reached the limit and the
+        // instance was torn down) before this incremental / terminal scan-range delivery arrived.
+        // A late delivery to a finished instance is a no-op, not an error.
+        return Status::OK();
     }
     RuntimeState* runtime_state = fragment_ctx->runtime_state();
 
--- a/be/src/exec/pipeline/scan/scan_operator.cpp
+++ b/be/src/exec/pipeline/scan/scan_operator.cpp
@@ -232,8 +232,19 @@ bool ScanOperator::is_finished() const {
         return false;
     }
 
+    // Once this scan has produced enough rows to satisfy its pushed-down limit, it must not keep
+    // waiting for more incremental scan ranges (`has_more`) nor for the remaining morsels to be
+    // drained. When connector incremental scan ranges are enabled the FE may short-circuit delivery
+    // after the limit is reached; if it fails to deliver the terminal `has_more=false` sentinel and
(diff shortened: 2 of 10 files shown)


CelerData-Reviewer:
@codex review

PR #75426, opened 2026-06-26
[BugFix] Fix GLM deserialize column failed on cross-side outer-join expressions

Hyper-FF:
## Why I'm doing:

Under global late materialization, a query whose deferred expression references columns from BOTH sides of an outer join (e.g. t.amt + mk.rate over mk RIGHT JOIN [bucket] t) failed with
"deserialize column failed, slot_id: NN, data_size: NN".

Root cause: a projection above the FETCH conservatively widens all of its outputs to nullable when an outer join exists below it (ProjectNode.hasNullableGenerateChild). This also widens the preserved side's row-position (row-id) virtual columns in the shared slot descriptor, even though the runtime column stays non-nullable. The GLM FETCH request path is the only consumer that rebuilds a column purely from the slot descriptor over the non-self-describing ColumnArraySerde format, so the sender serialized a non-nullable layout while the remote LookUp receiver built a nullable column and overran the buffer.

Make the FETCH request path conform to the slot descriptor on both ends:
- fetch_task: reconcile every request column to its descriptor nullability before serializing (ColumnHelper::update_column_nullable).
- lookup_operator: build the row-source column from its descriptor like the other request columns, and assert it carries no nulls, instead of forcing it non-nullable.

Add a SQL regression test covering RIGHT/LEFT/FULL/INNER joins with unmatched (null-producing) rows under global late materialization.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/pipeline/fetch_task.cpp
+++ b/be/src/exec/pipeline/fetch_task.cpp
@@ -23,6 +23,7 @@
 #include "base/status_fmt.hpp"
 #include "base/time/time.h"
 #include "base/utility/defer_op.h"
+#include "column/column_helper.h"
 #include "common/brpc/brpc_stub_cache.h"
 #include "exec/pipeline/fetch_processor.h"
 #include "gen_cpp/internal_service.pb.h"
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #75431, opened 2026-06-26
[BugFix] Cache Iceberg REST vended-credential tables and keep their credentials fresh

Youngwb:
## Why I'm doing:

PR #69434 introduced a cache bypass for Iceberg REST catalogs with vended credentials to avoid serving stale STS tokens. It fixed the expiry symptom but made **every `getTable()` hit the REST catalog and Lake Formation directly**, generating hundreds of `GetDataAccess` requests per minute and triggering AWS `Rate exceeded` failures (reported on #70988 by multiple users running AWS Glue / S3 Tables + Lake Formation).

The underlying problem is in `CachingIcebergCatalog`: credential-bearing `Table` objects were discarded on refresh instead of being kept current. This PR fixes that at the source and restores caching, using only **provider-agnostic** mechanisms.

This builds on and supersedes #70988: it keeps that PR's core `reload()` fix, adds a fix for the background-refresh path, and replaces the timing-based test with deterministic ones.

## What I'm doing:

Fixes #issue: N/A — connector caching regression introduced by #69434; tracked via #70988, no separate GitHub issue.

1. **`reload()` always returns the freshly loaded table.** The removed `metadataFileLocation` comparison returned the *old* cached `Table` (carrying its expiring FileIO/STS token) whenever metadata was unchanged — the common case — even though `delegate.getTable()` had already fetched a fresh one. The comparison saved no remote round-trip; it only threw away fresh credentials.

2. **Background `refreshTableUnderLock()` renews credentials when metadata is unchanged.** Previously it loaded a fresh table every cycle but discarded it unless the snapshot changed, so a frequently-refreshed-but-unchanged table kept its expiring credentials. The snapshot-delta partition/file caches stay valid when metadata is unchanged, so we keep gating the expensive delta maintenance on the metadata comparison, but now swap the already-loaded table into the cache so its renewed credentials replace the expiring ones (no extra remote cost — the `getTable()` call already happened).

3. **Revert the cache bypass from #69434.** Remove the `!delegate.isVendedCredentialsEnabled()` condition from `getTable()`, and delete the now-unused `isVendedCredentialsEnabled()` from `IcebergCatalog` and `IcebergRESTCatalog`.

4. **Cap the table-cache TTL for REST catalogs (default-safe, provider-agnostic).** Under the long default `iceberg_meta_cache_ttl_sec` (24h), an *idle* entry can be served past its ~1h token before `refreshAfterWrite` replaces it (Caffeine returns the stale value on the access that triggers the async reload). To make the default safe without operator tuning, the **`tables`** cache hard-expiry is capped at `3000s` when the delegate is a REST catalog. This bounds only the table cache (the partition/file caches keep the full TTL) and only for REST catalogs.

We deliberately do **not** parse per-cloud token-expiry properties to decide this: the markers differ per provider (`s3.session-token-expires-at-ms`, `gcs.oauth2.token-expires-at`, others for Azure/OSS/etc.) and are incomplete/version-dependent, so a hardcoded check would silently fail to protect unlisted providers. The cap is a coarse, provider-agnostic bound instead. Deployments with tokens shorter than the cap should still lower `iceberg_meta_cache_ttl_sec` to match.

Tests: `CachingIcebergCatalogTest` now covers (a) a REST catalog served from the cache across calls, (b) `reload()` installing the fresh table even when `metadataFileLocation` is identical, (c) `refreshTableUnderLock()` swapping in the reloaded table (renewing credentials) when metadata is unchanged, and (d) the REST table-cache TTL being capped while a non-REST catalog keeps the configured TTL. The reload/refresh tests drive the cache executor deterministically (no `Thread.sleep`).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

**Behavior change**: Iceberg REST catalogs with vended credentials use the Caffeine table cache again (as before #69434); cached `Table` objects are refreshed with current credentials on every refresh cycle, including the background refresh path. For REST catalogs, the table cache's hard-expiry is additionally capped at 3000s (the partition/file caches and non-REST catalogs are unaffected).

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
@@ -77,6 +77,9 @@ public class CachingIcebergCatalog implements IcebergCatalog {
     // Only emit the partition-load INFO line when a refresh exceeds this many partitions, so the log
     // remains useful for diagnosing slow loads on large tables without flooding for normal-sized ones.
     private static final int PARTITION_LOAD_LOG_THRESHOLD = 10000;
+    // REST tables embed short-lived vended credentials (~1h) in their FileIO; cap the table-cache TTL so an
+    // idle entry can't outlive its token. Provider-agnostic: bounds cache lifetime, no per-cloud expiry parse.
+    private static final long REST_TABLE_CACHE_MAX_TTL_SEC = 3000;
     private static final ThreadLocal<ConnectContext> TABLE_LOAD_CONTEXT = new ThreadLocal<>();
     private final String catalogName;
     private final IcebergCatalog delegate;
@@ -107,8 +110,12 @@ public class CachingIcebergCatalog implements IcebergCatalog {
         this.databases = newCacheBuilderWithMaximumSize(
                 icebergProperties.getIcebergMetaCacheTtlSec(),
                 NEVER_CACHE, DEFAULT_CACHE_NUM).build();
+        long tableCacheTtlSec = icebergProperties.getIcebergMetaCacheTtlSec();
+        if (delegate instanceof IcebergRESTCatalog) {
+            tableCacheTtlSec = Math.min(tableCacheTtlSec, REST_TABLE_CACHE_MAX_TTL_SEC);
+        }
         this.tables = newCacheBuilder(
-                icebergProperties.getIcebergMetaCacheTtlSec(),
+                tableCacheTtlSec,
                 icebergProperties.getIcebergTableCacheRefreshIntervalSec())
                 .executor(executorService)
                 .maximumWeight(tableCacheSize)
@@ -133,14 +140,7 @@ public class CachingIcebergCatalog implements IcebergCatalog {
                     @Override
                     public Table reload(IcebergTableName key, Table oldValue) {
                         try {
-                            BaseTable updateTable = 
-                                    (BaseTable) delegate.getTable(new ConnectContext(), key.dbName, key.tableName);
-                            TableOperations newOps = updateTable.operations();
-                            TableOperations oldOps = ((BaseTable) oldValue).operations();
-                            if (oldOps.current().metadataFileLocation().equals(newOps.current().metadataFileLocation())) {
-                                return oldValue;
-                            }
-                            return updateTable;
+                            return delegate.getTable(new ConnectContext(), key.dbName, key.tableName);
                         } catch (Exception e) {
                             LOG.warn("refresh table {}.{} failed", key.dbName, key.tableName, e);
                             return oldValue;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

Youngwb:
Thanks @chatgpt-codex-connector — both findings are valid, and I've reworked the PR to address them by **not** detecting per-cloud credential expiry in code at all.

**On the P2 (GCS) point specifically:** it's correct, and it generalizes — the expiry marker differs per provider (`s3.session-token-expires-at-ms` (ms), `gcs.oauth2.token-expires-at` (ms), and Azure/OSS/COS/OBS use other markers or none). Any hardcoded key list is inherently incomplete, so for an unlisted provider the check would silently return "not expired" and keep serving a dead token — worse than having no check, because it looks handled. So rather than add GCS (and chase the rest), I removed the S3-only `vendedCredentialExpired` helper entirely and reverted `reload()`'s failure path to the upstream `return oldValue` (which self-heals on the next successful background reload once the catalog recovers).

**On the P1 (idle table) point:** agreed it's real with the default 24h TTL. The fix is the **provider-agnostic** bound: configure `iceberg_meta_cache_ttl_sec` below the credential lifetime (e.g. `3400` for AWS LF 1h tokens) so an idle entry hard-expires before its token does, plus `iceberg_table_cache_refresh_interval_sec=600`. `expireAfterWrite` eviction doesn't care which cloud minted the token, so this closes the idle hole without per-provider parsing. This is now documented in the PR description as the requirement.

What this PR keeps, all provider-agnostic:
1. `reload()` always returns the freshly loaded table (drops the `metadataFileLocation` short-circuit that discarded fresh credentials).
2. `refreshTableUnderLock()` now swaps the reloaded table into the cache when metadata is unchanged, so the background refresh renews credentials instead of discarding them (the snapshot-delta partition/file caches stay gated on the metadata comparison, so no extra cost).
3. Revert the #69434 cache bypass.

So the "never serve an expired credential" guarantee is bounded by TTL (provider-agnostic) rather than by an in-code expiry check that can't be made complete.

Youngwb:
@codex review

Youngwb:
Addressed the idle-table point (P1/P2) in `6994befd974` — but provider-agnostically, not by parsing per-cloud expiry.

The `tables` cache hard-expiry is now **capped at 3000s when the delegate is a REST catalog**, so under the default 24h `iceberg_meta_cache_ttl_sec` an idle entry can no longer be served past a typical ~1h vended token. The cap:
- bounds only the **`tables`** cache (the partition/file caches keep the full TTL, so planning-time manifest caching is unaffected);
- applies only to **REST** catalogs (Hive/Glue/Hadoop keep 24h — they have no embedded credential);
- caps cache *lifetime* rather than reading any cloud's expiry marker, so it can't silently miss a provider.

It's a coarse bound, not a hard guarantee for arbitrarily short tokens: deployments using tokens shorter than the cap should still lower `iceberg_meta_cache_ttl_sec` to match (documented in the PR description). Test `testRestTableCacheTtlIsCapped` asserts REST is capped and non-REST is not.

Youngwb:
@codex review

dirtysalt (approved):
LGTM! 🚀

Astralidea (approved):

PR #75432, opened 2026-06-26
[BugFix] Preserve sub-second when loading pre-1970 ORC TIMESTAMP

xiangguangyxg:
## Why I'm doing:

`OrcTimestampHelper::orc_ts_to_native_ts_before_unix_epoch` (`be/src/formats/orc/utils.h`) hardcoded the microsecond argument to `0`, so loading **any pre-1970 ORC `TIMESTAMP` with a non-zero sub-second** dropped the fraction, e.g. `1965-03-02 12:00:00.500000` was loaded as `1965-03-02 12:00:00.000000`. It affects both the plain `TIMESTAMP` and the `TIMESTAMP WITH LOCAL TIME ZONE` (instant) read paths and is independent of tablet pre-split.

The `after_unix_epoch` path already carries the sub-second through (`nanoseconds / 1000`); only the `before_unix_epoch` branch threw it away. liborc hands the load path a clean, floored `(seconds, nanoseconds)` pair (`nanoseconds` in `[0, 1e9)`, `instant = data + nanoseconds/1e9`): the writer's `+1` (`ColumnWriter.cc`) and reader's `-1` (`ColumnReader.cc`) for negative sub-second values cancel.

## What I'm doing:

`be/src/formats/orc/utils.h`, in `orc_ts_to_native_ts_before_unix_epoch`: pass `nanoseconds / NANOSECS_PER_USEC` as the microsecond instead of `0`, mirroring the unguarded `after_unix_epoch` sibling. The row load path always supplies a non-negative `nanoseconds`, so this simply restores the sub-second. The dispatcher and the `after_unix_epoch` path are untouched, so timezone composition, whole-second negatives, and post-1970 values are unchanged.

This helper is also shared by the ORC stripe min/max statistics decoder (`orc_min_max_decoder.cpp`), whose pre-1970 sub-second handling has separate, pre-existing quirks (the ORC nanos field is stored with a `+1` offset the decoder does not undo; its truncating-division remainder can be negative; the before-epoch instant branch ignores the reader offset). To keep this change scoped to the data load path and leave predicate-pushdown bounds **byte-for-byte unchanged**, the decoder now explicitly drops the sub-second for negative-epoch bounds (its prior behavior) rather than letting the now-sub-second-preserving helper expose those quirks. Correctly decoding pre-1970 stripe-stats sub-second is left as a follow-up.

BE-only; FE and the pre-split pipeline are not touched.

**Behavior change: No** — values only, correcting a lossy result (the old output was a bug, not intended behavior); no SQL syntax / config / interface change. Same rationale as the companion Parquet fix #75207.

### Tests

`be/test/formats/orc/orc_chunk_reader_test.cpp` — `TestTimestampPreEpochSubSecond` drives the real `OrcChunkReader` load path: pre-1970 sub-second on the plain (`.500000`, `.123456`) and instant (`.500000`) paths, plus whole-second-negative and post-1970 sub-second as no-regression guards. Verified RED→GREEN on a real build (before the fix the sub-second cases dropped to `.000000` and the instant case dropped its fraction; after the fix all pass and the pre-existing `TestTimestamp` still passes). `OrcMinMaxDecoderPreEpochTimestampDropsSubSecond` covers the decoder guard: a pre-1970 stripe-stats bound decodes to its whole second (sub-second dropped, pruning unchanged). BE module-boundary check clean.

Companion to the merged Parquet fix #75207; together they cover pre-1970 sub-second temporal loads.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/formats/orc/orc_min_max_decoder.cpp
+++ b/be/src/formats/orc/orc_min_max_decoder.cpp
@@ -137,6 +137,23 @@ static Status decode_datetime_min_max(const orc::Type* orc_type, const orc::prot
             }
             int64_t secs = ms / 1000;
             ns += (ms - secs * 1000) * 1000000L;
+            // orc_ts_to_native_ts now carries the sub-second into the before-epoch result for the
+            // data load path. Pre-1970 stripe-stats sub-second is a separate, pre-existing concern
+            // here (the ORC nanos field is stored with a +1 offset this decoder does not undo, the
+            // remainder above can be negative, and the before-epoch instant branch ignores the
+            // offset). Dropping the min sub-second does not affect pruning, but dropping the max
+            // sub-second understates the max bound; keep dropping it for negative-epoch bounds to
+            // leave pruning byte-for-byte unchanged rather than regress it.
+            //
+            // TODO: decode pre-1970 stripe-stats sub-second correctly in a dedicated stats path:
+            //   - subtract the +1 nanos offset (mirror liborc's reader);
+            //   - floor secs toward -inf (not toward zero) so the ms remainder stays non-negative;
+            //   - apply the instant tz offset on the negative-epoch branch;
+            //   - round conservatively for bounds (floor the min, ceil the max) so [min, max] still
+            //     contains the true value range instead of understating the max.
+            if (secs < 0) {
+                ns = 0;
+            }
             OrcTimestampHelper::orc_ts_to_native_ts(&min, utc_tzinfo, tz_offset_in_seconds, secs, ns, is_instant);
         }
 
@@ -148,6 +165,10 @@ static Status decode_datetime_min_max(const orc::Type* orc_type, const orc::prot
             }
             int64_t secs = ms / 1000;
             ns += (ms - secs * 1000) * 1000000L;
+            // See the minimum branch: keep dropping the sub-second for negative-epoch bounds.
+            if (secs < 0) {
+                ns = 0;
+            }
             OrcTimestampHelper::orc_ts_to_native_ts(&max, utc_tzinfo, tz_offset_in_seconds, secs, ns, is_instant);
         }
 
--- a/be/src/formats/orc/utils.h
+++ b/be/src/formats/orc/utils.h
@@ -86,7 +86,10 @@ public:
                                                       int64_t nanoseconds) {
         cctz::time_point<cctz::sys_seconds> t = CCTZ_UNIX_EPOCH + cctz::seconds(seconds);
         const auto tp = cctz::convert(t, tz);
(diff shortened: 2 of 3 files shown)


xiangguangyxg:
@codex review

xiangguangyxg:
@codex review

xiangguangyxg:
@codex review

be/src/formats/orc/orc_min_max_decoder.cpp
```diff
@@ -137,6 +137,14 @@ static Status decode_datetime_min_max(const orc::Type* orc_type, const orc::prot
             }
             int64_t secs = ms / 1000;
             ns += (ms - secs * 1000) * 1000000L;
+            // orc_ts_to_native_ts now carries the sub-second into the before-epoch result for the
+            // data load path. Pre-1970 stripe-stats sub-second is a separate, pre-existing concern
+            // here (the ORC nanos field is stored with a +1 offset this decoder does not undo, the
+            // remainder above can be negative, and the before-epoch instant branch ignores the
+            // offset), so keep dropping it for negative-epoch bounds to leave pruning unchanged.
```
kevincai: shall add TODO here, otherwise this will be a mystery why implementing like this.

dropping the min nano doesn't affect the prune result, but dropping the max nano will do.

```
  // results. Drop it for negative-epoch bounds to keep pruning byte-for-byte unchanged.
  //
  // TODO: decode pre-1970 stripe-stats sub-second correctly in a dedicated stats path:
  //   - subtract the +1 nanos offset (mirror liborc's reader);
  //   - floor secs toward -inf (not toward zero) so the ms remainder stays non-negative;
  //   - apply the instant tz offset on the negative-epoch branch;
  //   - round conservatively for bounds (floor the min, ceil the max) so [min, max] still
  //     contains the true value range instead of understating the max.
```

xiangguangyxg: Done in 4330ff7 — added the TODO with the dedicated-stats-path recipe you outlined (subtract the +1 nanos offset, floor secs toward -inf, apply the instant tz offset on the negative-epoch branch, and floor-min / ceil-max for conservative bounds), and noted that it's the max sub-second whose dropping would understate the bound. Thanks for the catch.

kevincai (approved):

srlch (approved):

PR #75444, opened 2026-06-26
[Enhancement] Self-enable and tune TopN runtime-filter back-pressure on the scan path

Hyper-FF:
## Why I'm doing:

For ORDER BY ... LIMIT (TopN), the sort builds a stream runtime filter (RF) that tightens as top-k candidates arrive and is pushed to the scan to skip rows that cannot enter the top-k. On the shared-data (lake/connector) path this was largely ineffective: the scan reads far ahead of the RF, so a near-distinct GROUP BY ... ORDER BY ... LIMIT floods the aggregation before the RF can prune (observed with connector_io_tasks_per_scan_operator=16: ~6.4M rows through the scan before the RF arrived).

Root causes and fixes:
- Back-pressure only gated output, not IO submission, and the row budget is not concurrency-aware, so N readers overshoot it Nx. Clamp scan read-ahead to a small IO-task count until the RF actually arrives; full DOP resumes once it does.
- Back-pressure was gated on the FE topn_filter_back_pressure_mode flag (default 0), so the path that needs it most never ran it. Scans now self-enable back-pressure whenever a TopN RF targets them, for both shared-nothing (olap) and shared-data (lake/connector) scans, controlled by the enable_topn_filter_back_pressure session variable (default true).
- The throttle was never wired into the event scheduler (only re-checked by the fallback poller). Arm an event-scheduler wakeup timer at the throttle deadline.
- Release was selectivity-based and went stale once zonemap pruning emptied the pulled chunks. Release on actual RF arrival (rf_arrived latch).
- TopN/stream-build RFs push down only as zonemap page-range pruning, so still evaluate them per chunk to trim within-page survivors.

Tunability: every knob is a session variable carried on TQueryOptions, defaults matching the tuned behavior (topn_filter_back_pressure_io_tasks, enable_topn_filter_back_pressure, topn_back_pressure_max_rounds/num_rows/ throttle_time_ms/throttle_time_upper_bound_ms).

Pipeline-breaker guard: when a TopN RF reaches a scan only across a non-aggregation deterministic pipeline breaker (blocking sort, analytic/window) the RF cannot arrive in time, so FE marks the scan (topn_filter_back_pressure_disabled) and BE skips back-pressure on both the FE-driven and self-enable paths. Aggregation is not treated as a breaker (a near-distinct GROUP BY degrades to a streaming pre-aggregation, which is exactly the case this optimization targets).

Diagnostics: enable_topn_runtime_filter_diagnostic_log (BE, hot-reloadable) traces RF build/publish, arrival timing, per-segment zonemap prune effect, and observed selectivity.

Validated on a shared-data cluster (TPCH tpch_100g.lineitem group-by-then-TopN, connector_io_tasks=16): update-phase aggregation ~13x faster and ~5.4x less allocated memory vs the feature off; adaptive IO tasks ON shows no regression; the io-task clamp and enable switch change behavior monotonically at runtime.

## What I'm doing:

Fixes #issue

  | Metric                 | Baseline | Patched (default) | Improvement  |
  |------------------------|----------|-------------------|--------------|
  | Query total time       | 182 ms   | 94 ms             | ~1.9× faster |
  | Query peak memory      | 224.8 MB | 28.9 MB           | ~7.8× less   |
  | Query allocated memory | 1.78 GB  | 332 MB            | ~5.4× less   |

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/connector_scan_node.cpp
+++ b/be/src/exec/connector_scan_node.cpp
@@ -101,6 +101,8 @@ Status ConnectorScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
             _back_pressure_throttle_time = tnode.lake_scan_node.back_pressure_throttle_time;
             _back_pressure_throttle_time_upper_bound = tnode.lake_scan_node.back_pressure_throttle_time_upper_bound;
         }
+        _topn_filter_back_pressure_disabled = tnode.lake_scan_node.__isset.topn_filter_back_pressure_disabled &&
+                                              tnode.lake_scan_node.topn_filter_back_pressure_disabled;
     }
 
     return Status::OK();
--- a/be/src/exec/olap_scan_node.cpp
+++ b/be/src/exec/olap_scan_node.cpp
@@ -153,6 +153,8 @@ Status OlapScanNode::init(const TPlanNode& tnode, RuntimeState* state) {
         _back_pressure_throttle_time = tnode.olap_scan_node.back_pressure_throttle_time;
         _back_pressure_throttle_time_upper_bound = tnode.olap_scan_node.back_pressure_throttle_time_upper_bound;
     }
+    _topn_filter_back_pressure_disabled = tnode.olap_scan_node.__isset.topn_filter_back_pressure_disabled &&
+                                          tnode.olap_scan_node.topn_filter_back_pressure_disabled;
 
     _estimate_scan_and_output_row_bytes();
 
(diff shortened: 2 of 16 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

kangkaisen (approved):

PR #75449, opened 2026-06-26
[BugFix] Fix NLJoin crash from build-side column nullability mismatch (backport #75343)

Joob1n:
## Why I'm doing:

The query will crash when the query has following case:
- use NL-join
- build-side's rows less than normal chunk size
- the slot-descriptor is not nullable and the runtime-status is nullable

Then the query will append a nullable column-value into a not nullable column of output chunk.

Like this SQL case:
```
create table t5 (x int null, v double null)
duplicate key(x) distributed by hash(x) buckets 4 properties("replication_num"="1");

insert into t5 select g%100000, cast(g%500 as double)
from table(generate_series(1,20000)) t(g);

with scd as (select 'test' as type, 0 as lo, 100000 as hi)
select a.*, b.v from scd a
left join t5 b on b.x between a.lo and a.hi
left join t5 c on c.x between a.lo and a.hi
order by b.v limit 5;
```
A middle-layer TopN between of the double-layer NL-join with a not nullable column. Because the TopN has a outer-join as sub-node, Optimizer will let the TopN sets the runtime-status as nullable. Then StarRocks get a diff-status between slot-descriptor and runtime-status. Then BE builds the output chunk's nullability from a transient build-chunk pointer that has already gone null after the only chunk is consumed, so the column is created as non-nullable while the build data is nullable, and appending it crashes the BE.

crash stack in main:
```
*** Aborted at 1782453455 (unix time) try "date -d @1782453455" if you are using GNU date ***
PC: @     0x74d8626819fc pthread_kill
*** SIGABRT (@0x57c82) received by PID 359554 (TID 0x74d6d41e6640) LWP(360386) from PID 359554; stack trace: ***
    @         0x203382d2 starrocks::BinaryColumnBase::append(starrocks::Column const&, unsigned long, unsigned long)
    @         0x19318be4 starrocks::Column::append(starrocks::Column const&)
    @         0x1a2b49d5 starrocks::pipeline::NLJoinProbeOperator::_permute_probe_row(std::shared_ptr const&)
    @         0x1a2b44f4 starrocks::pipeline::NLJoinProbeOperator::_permute_chunk_for_other_join(unsigned long)
    @         0x1a2b60c7 starrocks::pipeline::NLJoinProbeOperator::_pull_chunk_for_other_join(unsigned long)
    @         0x1a2b5afc starrocks::pipeline::NLJoinProbeOperator::pull_chunk(starrocks::RuntimeState*)
    @         0x1c8e9093 starrocks::pipeline::PipelineDriver::process(starrocks::RuntimeState*, int)
    @         0x1b104227 starrocks::pipeline::GlobalDriverExecutor::_worker_thread()
```

crash stack in 3.5:
```
*** SIGABRT (@0x175c5) received by PID 95685 (TID 0x716673be4640) LWP(96543) from PID 95685; stack trace: ***
    @     0x716810024e96 __assert_fail
    @         0x15480806 starrocks::BinaryColumnBase const& down_cast const&, starrocks::Column const>(starrocks::Column const&)
    @         0x15468a1c starrocks::BinaryColumnBase::append(starrocks::Column const&, unsigned long, unsigned long)
    @         0x1531bc9e starrocks::Column::append(starrocks::Column const&)
    @         0x19980bf0 starrocks::pipeline::NLJoinProbeOperator::_permute_probe_row(std::shared_ptr const&)
    @         0x199806af starrocks::pipeline::NLJoinProbeOperator::_permute_chunk_for_other_join(unsigned long)
    @         0x1998245c starrocks::pipeline::NLJoinProbeOperator::_pull_chunk_for_other_join(unsigned long)
    @         0x19981fe7 starrocks::pipeline::NLJoinProbeOperator::pull_chunk(starrocks::RuntimeState*)
```

## What I'm doing:

The fix reads nullability from a stable build chunk instead.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

> Backport of #75343 to branch-3.5.

### DIFF
--- a/be/src/exec/pipeline/nljoin/nljoin_probe_operator.cpp
+++ b/be/src/exec/pipeline/nljoin/nljoin_probe_operator.cpp
@@ -272,11 +272,15 @@ ChunkPtr NLJoinProbeOperator::_init_output_chunk(size_t chunk_size) const {
         MutableColumnPtr new_col = ColumnHelper::create_column(slot->type(), nullable);
         chunk->append_column(std::move(new_col), slot->id());
     }
+    // A null _curr_build_chunk means the previous round just consumed all build chunks, so use the
+    // first build chunk to decide column nullability.
+    // NOTE: this can still be wrong if column info differs between build chunks.
+    Chunk* ref_build_chunk = _num_build_chunks() > 0 ? _cross_join_context->get_build_chunk(0) : _curr_build_chunk;
     for (size_t i = _probe_column_count; i < _col_types.size(); i++) {
         SlotDescriptor* slot = _col_types[i];
         bool nullable = right_to_nullable | _col_types[i]->is_nullable();
-        if (_curr_build_chunk) {
-            nullable |= _curr_build_chunk->is_column_nullable(slot->id());
+        if (ref_build_chunk) {
+            nullable |= ref_build_chunk->is_column_nullable(slot->id());
         }
         MutableColumnPtr new_col = ColumnHelper::create_column(slot->type(), nullable);
         chunk->append_column(std::move(new_col), slot->id());
--- a/be/test/CMakeLists.txt
+++ b/be/test/CMakeLists.txt
@@ -70,6 +70,7 @@ set(EXEC_FILES
         ./exec/pipeline/sink/table_function_table_sink_operator_test.cpp
         ./exec/pipeline/sink/memory_scratch_sink_operator_test.cpp
         ./exec/pipeline/limit_operator_test.cpp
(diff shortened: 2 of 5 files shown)


stdpain (approved):

trueeyu:

PR #75451, opened 2026-06-26
[BugFix] Fix heap-use-after-free in OrderedPartitionExchanger when previous chunk is mutated downstream (backport #75279)

martinr0x:
## Why I'm doing this

Fix: https://github.com/StarRocks/StarRocksTest/issues/11482

## Why I'm doing this

A heap-use-after-free (caught by ASAN) can crash the BE when a window function with a skew hint on `PARTITION BY` runs at DOP>1:

```
heap-use-after-free in Cow<Column>::ImmutPtr::operator->()
  READ  : Chunk::num_rows()      <- OrderedPartitionExchanger::accept()     (thread T296)
  FREED : Chunk::append_column() <- Analytor::_output_result_chunk()        (thread T295)
```

**Root cause:** `OrderedPartitionExchanger::accept()` hands a chunk to the downstream sources (`add_chunk`) and then retains references into that same chunk — `_previous_chunk` and `_previous_partition_columns` (which alias the chunk's partition-key columns for `SlotRef` keys). On the next `accept()`, it reads `_previous_chunk->num_rows()` and `compare_at()` on those columns to test the partition join between consecutive chunks. But the chunk has already been consumed by the downstream `AnalyticSinkOperator`, whose `_output_result_chunk()` reuses the input chunk as its output and mutates it on another thread:

- `append_column()` reallocates the chunk's `_columns` vector → the reported UAF on `num_rows()`.
- On the LIMIT path, `set_num_rows()` → `as_mutable_raw_ptr()->resize()` resizes every column in place (bypassing COW) → races with `compare_at()` and can index out of bounds.

The exchanger violates the pipeline contract that a chunk handed downstream is owned by the consumer. The bug has existed since `OrderedPartitionExchanger` was introduced (#35486); the analytic operator already reused and mutated the input chunk by then (#34312). It only triggers with a window skew hint + `PARTITION BY` + DOP>1, which is why it escaped routine testing.

## What I'm doing

Stop retaining anything that aliases a chunk handed downstream. `accept()` only needs the previous chunk's last-row partition key to compare against the next chunk's first row, so:

- Replace `ChunkPtr _previous_chunk` / `Columns _previous_partition_columns` (aliasing the chunk) with an owned single-row copy of the boundary key, plus a `bool _has_previous` sentinel.
- Add `_clone_partition_key_row()` which deep-copies just the last row of each partition-key column.
- Clone the boundary row before the chunk is published via `add_chunk()` — i.e. while it is still exclusively owned by the current driver thread — so the clone itself is race-free without locks. After publishing, the exchanger only ever reads its own copy.

The clone is one row × a few key columns per chunk (negligible vs. the sort+window work), and is robust against any downstream in-place chunk mutation, not just the two known today.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exec/pipeline/exchange/local_exchange.cpp
+++ b/be/src/exec/pipeline/exchange/local_exchange.cpp
@@ -187,6 +187,14 @@ void OrderedPartitionExchanger::close(RuntimeState* state) {
 Status OrderedPartitionExchanger::accept(const ChunkPtr& chunk, const int32_t sink_driver_sequence) {
     DCHECK_EQ(sink_driver_sequence, 0);
 
+    const size_t cur_num_rows = chunk->num_rows();
+    // Drop empty chunks: they carry no rows to route and no partition boundary. This also guards the
+    // `cur_num_rows - 1` boundary-row indexing below from wrapping. Matches the other row-partition
+    // exchangers, which all early-return on empty input.
+    if (cur_num_rows == 0) {
+        return Status::OK();
+    }
+
     Columns partition_columns(_partition_exprs.size());
     for (size_t i = 0; i < partition_columns.size(); ++i) {
         ASSIGN_OR_RETURN(partition_columns[i], _partition_exprs[i]->evaluate(chunk.get()));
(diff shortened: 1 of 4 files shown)


martinr0x:
Closed because it was already backported https://github.com/StarRocks/starrocks/pull/75424

PR #75026, opened 2026-06-18
[BugFix] Fix Spurious Alembic Regeneration for Syntax, Property

chris-celerdata:
## Why I'm doing:
There are two cases of spurious autogenerate diffs on StarRocks < 4.0.6: view/MV definitions that are semantically identical but syntactically different from StarRocks' own canonicalization and `enable_statistic_collect_on_first_load` table property generating phantom changes because StarRocks only emits the property in `tables_config` when it is explicitly false. 

## What I'm doing:
**View/MV canonicalization**: When comparing view or MV definitions, create a temporary VIEW to obtain the databases' canonical form before comparison. The regex-based normalizer is reserved as a fallback and has some bug fixes, but is not as comprehensive as direct comparison. 
**`enable_statistic_collect_on_first_load`**: Adds a default of true so an absent value matches 'true'. Adds a set `_SKIP_IMPLICIT_RESET_PROPERTIES` such that when the property is absent, the DB value is left untouched rather than reset to the default. This matches the absent = true approach. 
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [x] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [x] This pr needs user documentation (for new or modified features or behaviors)
  - [x] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

PR #75480, opened 2026-06-27
[BugFix] Stabilize Python client integration tests

jx2lee:
## Why I'm doing:

Python client integration/system tests currently fail with newer StarRocks and Alembic versions due to reflection output changes, parser edge cases, Alembic autogenerate behavior changes.

## What I'm doing:

Fix Python client test failures caused by newer StarRocks reflection output and Alembic behavior changes.
- `uv run pytest test --log-cli-level=WARNING`
- `uv run pytest -s test/integration`
- `uv run pytest -s test/system`

or success `uv run pytest` in python client repository dir.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/contrib/starrocks-python-client/starrocks/alembic/compare.py
+++ b/contrib/starrocks-python-client/starrocks/alembic/compare.py
@@ -21,6 +21,10 @@ from alembic.autogenerate import comparators
 from alembic.autogenerate.api import AutogenContext
 from alembic.ddl import DefaultImpl
 from alembic.operations.ops import AlterColumnOp, AlterTableOp, UpgradeOps
+try:
+    from alembic.util import DispatchPriority
+except ImportError:
+    DispatchPriority = None  # type: ignore[assignment,misc]
 from sqlalchemy import Column
 from sqlalchemy.engine.reflection import Inspector
 from sqlalchemy.exc import ArgumentError
@@ -197,7 +201,7 @@ def compare_complex_type(impl: DefaultImpl, inspector_type: sqltypes.TypeEngine,
     return True
 
 
-def comparators_dispatch_for_starrocks(dispatch_type: str):
+def comparators_dispatch_for_starrocks(dispatch_type: str, *, priority=None):
     """
     StarRocks-specific dispatch decorator.
 
@@ -231,10 +235,17 @@ def comparators_dispatch_for_starrocks(dispatch_type: str):
             # StarRocks dialect, execute actual logic
             return func(*args, **kwargs)
 
+        registration_options = {}
+        if priority is not None and DispatchPriority is not None:
+            registration_options["priority"] = priority
+
         # Register to Alembic dispatch system
-        return comparators.dispatch_for(dispatch_type)(wrapper)
+        return comparators.dispatch_for(dispatch_type, **registration_options)(wrapper)
 
     return decorator
+
+
+_SCHEMA_COMPARATOR_PRIORITY = DispatchPriority.LAST if DispatchPriority is not None else None
 
 
 # ==============================================================================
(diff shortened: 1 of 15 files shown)


jaogoy (commented):
Thanks! Two robustness suggestions on the production code (not the tests):

1. `bucket_size` default (common/defaults.py) — Config.default_automatic_bucket_size has flip-flopped: 1G (3.2) → 4G (3.3–4.0) → 1G (4.1+), and it's frozen into a table's properties at create time. So no single constant is correct everywhere — this value is right for 4.1+ but wrong for 3.3–4.0, and for such a table with no bucket_size in metadata the comparator will emit a spurious ALTER ... SET ("bucket_size"="1073741824") that actually shrinks it. Two options, I'd recommend B:

    A. Choose the default by server version.
    B. (recommended) Don't diff bucket_size unless it's explicitly declared in metadata — the only approach correct across all versions and needing no version table. It does mean special-casing it rather than the current uniform default-based comparison.

2. REFRESH SCHEDULE → ASYNC — normalizing is good (SCHEDULE is just a display alias for ASYNC + EVERY; SR's parser treats them as synonyms). But the anchored pre-parse regex ^REFRESH\s+SCHEDULE misses REFRESH DEFERRED SCHEDULE EVERY (...) (a valid SHOW CREATE form), which then falls back with a warning. Suggest handling it in the grammar instead:

`async_scheme: ("ASYNC"i | "SCHEDULE"i) ASYNC_DETAILS?`

The transformer already emits canonical ASYNC ..., so no transformer change is needed and both forms parse cleanly without preprocessing. Keep the string-rewrite in the reflection.py fallback, and a REFRESH DEFERRED SCHEDULE EVERY (...) test would be worth adding.

jx2lee:
@jaogoy Thanks for reviewing. I picked option B for `bucket_size`: only diff it when explicitly declared in metadata.

Also added grammar-level handling for `REFRESH SCHEDULE`, and deferred schedule parser case, and removed the now-unneeded `_REFRESH_SCHEDULE_PATTERN`.

jaogoy:
It's better to pass in a list to ignore the implicit comparison.

contrib/starrocks-python-client/starrocks/alembic/compare.py
```diff
@@ -1779,6 +1820,11 @@ def _compare_table_properties_impl(
     for key in all_keys:
         conn_value = normalized_conn.get(key)
         meta_value = normalized_meta.get(key)
+
+        # Only diff bucket_size when metadata declares it; its implicit default is version-dependent.
+        if skip_implicit_bucket_size and key == "bucket_size" and meta_value is None:
```
chris-celerdata: Can you take a look at the approach taken in #75026? 
```py
# Properties never implicitly reset when absent from metadata. Needed for properties
# whose DB representation is asymmetric (e.g. only stored when false), so a missing
# metadata entry should leave the DB value alone rather than reset it to the default.
_SKIP_IMPLICIT_RESET_PROPERTIES: frozenset = frozenset({
    'enable_statistic_collect_on_first_load',
})
```

The buckets property is similar to this property.

chris-celerdata (changes requested):
Thanks for the contribution! Can the dependencies be updated in `requirements-dev.txt` due to these changes (for non `uv` development)?

jx2lee:
> Thanks for the contribution! Can the dependencies be updated in `requirements-dev.txt` due to these changes (for non `uv` development)?

@chris-celerdata Thanks for cheking, I don't think this PR adds a new dependency; the changes is not related to package.
If i missed dependency, please point me to it and will update requirements dev. Thanks !

contrib/starrocks-python-client/starrocks/alembic/compare.py
jx2lee: it make sense. I fixed with `_TABLE_SKIP_IMPLICIT_RESET_PROPERTIES` constant.

PR #75488, opened 2026-06-28
[BugFix] Release global meta lock on dumpImage partial db-lock failure

kevincai:
dumpImage() acquired the per-db READ locks inside the try block and released every db in lockedDbMap in the finally. If lockDatabase() threw mid-loop (deadlock victim or interrupt), the finally's unlock loop reached a db that was never locked and threw IllegalMonitorStateException, aborting the finally before unlock() ran. This stranded the global meta lock indefinitely, blocking all future checkpoints and DDL.

Track only the dbs actually locked and release just those in an inner finally; move the global unlock() into an outer finally that always runs, so the global lock is released even if db-lock acquisition or release throws.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/server/GlobalStateMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/server/GlobalStateMgr.java
@@ -2957,26 +2957,36 @@ public class GlobalStateMgr {
                 lockedDbMap.put(dbId, db);
             }
 
-            // lock all dbs
-            for (Database db : lockedDbMap.values()) {
-                locker.lockDatabase(db.getId(), LockType.READ);
-            }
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

gengjun-git (approved):

HangyuanLiu (approved):

PR #75491, opened 2026-06-28
[Enhancement] Balance load-primary compute nodes for skewed shared-data shards

luohaha:
## Why I'm doing:

In shared-data mode, the load-primary compute node for each tablet is chosen as its
StarOS shard owner (`OlapTableSink.createLocation` → `getAllComputeNodeIdsAssignToTablets`
/ `getPrimaryComputeNodeIdByShard`). Freshly created shards — e.g. the presplit/reshard
shards of a RANGE-distributed Primary-Key table — can all land on a single worker at
creation time. A load that runs immediately afterwards then routes **every** tablet's
delta-writer, memtable flush and load-spill merge to that one node, while the other
compute nodes sit idle and backpressure the scan.

This was measured on a 3-BE shared-data cluster loading a 1.5B-row PK table via
`INSERT INTO ... SELECT * FROM FILES()`: with 6 tablets balanced 2-per-BE in storage,
the RANGE table's write phase still ran entirely on one BE (~2× CPU, all 6 tablets'
spill-merge there, the other two BEs ~98% idle), making the import ~2.4× slower than
the equivalent HASH table (~1507s vs ~627s) even though both had an identical balanced
tablet layout. The post-load `SHOW TABLETS` ownership was balanced 2/BE — the skew is
purely at *load time*, because the load-primary follows the (skewed) shard owner.

## What I'm doing:

In shared-data mode the write node need not be the shard owner — segment data is
persisted to object storage, so any alive compute node can serve as the load primary.
`createLocation` now collects the natural per-tablet load primaries and, when that
assignment is more skewed than an even spread (some node owns more than `ceil(n/k)`
tablets over `k` alive nodes), reassigns them round-robin across the alive compute
nodes. Already-balanced assignments (the common HASH-bucketing case) are returned
unchanged, so HASH loads are unaffected. Guarded by a new mutable FE config
`lake_balance_load_primary_compute_nodes` (default `true`) for easy rollback.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/Config.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/Config.java
@@ -1236,6 +1236,12 @@ public class Config extends ConfigBase {
     @ConfField(mutable = true)
     public static boolean lake_enable_tablet_creation_optimization = false;
 
+    @ConfField(mutable = true, comment = "In shared-data mode, balance per-tablet load-primary compute " +
+            "nodes across alive nodes when the natural (StarOS shard owner) assignment is skewed, e.g. a " +
+            "freshly presplit/reshard RANGE table whose new shards all live on one worker. Prevents the " +
+            "whole load write phase (delta-writer/flush/spill-merge) from concentrating on one node.")
+    public static boolean lake_balance_load_primary_compute_nodes = true;
+
     @ConfField(mutable = true, comment = "Max retry times for failed create tablet tasks in shared-data mode. " +
             "Only explicitly failed tasks are retried (not timeouts). Set 0 to disable retry.")
     public static int lake_create_tablet_max_retries = 1;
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

luohaha:
Superseded by a root-cause fix. Instead of rebalancing load-primary compute nodes at load-plan time, pre-split now spreads its new shards across compute nodes at shard-creation time: `SplitTabletJob` marks the pre-split entry points with a `spreadNewShards` flag and `StarOSAgent.createShardsForSplit` drops the `WITH_SHARD` PACK pin for pre-split (the source tablet is empty, so there is no warm cache to reuse). The load is then balanced from the start, and post-fix `SHOW TABLETS` and load-time placement agree. Closing in favor of that approach; the load-side rebalance here would only ever be a no-op once shards are created spread. Replacement fix: luohaha/starrocks#17 (stacked on the pre-split feature branch).

PR #75501, opened 2026-06-29
[BugFix] Link librdkafka/libpulsar on Runtime to fix gold linker build

wangruin:
## Why I'm doing:

`be/src/runtime/routine_load/data_consumer.cpp` and `data_consumer_group.cpp` are
compiled into the `Runtime` library and reference `librdkafka`/`libpulsar` symbols.
After `RoutineLoadTaskExecutor` moved to the `Orchestration` module (#75461), the
consumer that pulls `data_consumer.o` into the link is `Orchestration`, which was
only reachable as a transitive (PRIVATE) dependency and ended up *after* the
thirdparty group in the final `starrocks_be` link line.

The GNU gold linker (forced on CentOS where `glibc < 2.29` and `GCC >= 14`, see
`be/CMakeLists.txt`) does a single left-to-right pass and does not backtrack into
the already-processed `librdkafka`/`libpulsar` archives, so it fails with:

```
be/src/runtime/routine_load/data_consumer.cpp:679: error: undefined reference to 'pulsar::Consumer::close()'
be/src/runtime/routine_load/data_consumer.cpp:338: error: undefined reference to 'RdKafka::err2str[abi:cxx11](RdKafka::ErrorCode)'
be/src/runtime/routine_load/data_consumer.cpp:65:  error: undefined reference to 'RdKafka::Conf::create(RdKafka::Conf::ConfType)'
...
make[2]: *** [.../starrocks_be] Error 1
```

Ubuntu CI passes because it uses lld, which resolves symbols globally regardless of
group boundaries, so this is only reproducible on the CentOS7 + GCC>=14 (gold) build.

## First attempt (rejected)

The first version declared the thirdparty dependency on the consuming library:

```cmake
# be/src/runtime/CMakeLists.txt
target_link_libraries(Runtime PRIVATE librdkafka_cpp librdkafka libpulsar)
```

This did clear the `pulsar::*` / `RdKafka::*` undefined references, but because
`librdkafka`/`libpulsar` are already listed in the thirdparty `--start-group`, it
added a **second, mis-ordered copy** of those archives into the link line. That
copy pulled `tinycthread` out of `librdkafka` next to the one in `libserdes`, and
left `librdkafka`'s SASL backend without its `sasl_*` provider:

```
ld.gold: error: librdkafka.a(tinycthread.c.o): multiple definition of 'mtx_init'
ld.gold: libserdes.a(tinycthread.o): previous definition here
...
./librdkafka-2.11.0/src/rdkafka_sasl_cyrus.c:489: error: undefined reference to 'sasl_dispose'
./librdkafka-2.11.0/src/rdkafka_sasl_cyrus.c:548: error: undefined reference to 'sasl_client_new'
...
```

So the dependency-declaration approach was dropped: duplicating thirdparty archives
just trades one link error for another.

## What I'm doing:

Leave the thirdparty group untouched (single copy, original order) and instead list
`Orchestration` explicitly inside the `--start-group/--end-group` block in
`STARROCKS_LINK_LIBS`, ahead of the thirdparty group:

```cmake
    Runtime
    Orchestration
    DataWorkflows
```

This keeps the routine-load consumer in the static library group, so `data_consumer.o`
is pulled while the group is still being processed and its kafka/pulsar references
resolve against the single thirdparty archive set that follows. It touches only the
link order (no duplicated thirdparty archives) and does not change behavior on
Ubuntu/lld.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/be/CMakeLists.txt
+++ b/be/CMakeLists.txt
@@ -552,6 +552,7 @@ set(STARROCKS_LINK_LIBS
     FileSystem
     Formats
     Runtime
+    Orchestration
     DataWorkflows
     Udf
     Util


kevincai (dismissed):

CelerData-Reviewer:
@codex review

kevincai (approved):

alvin-phoenix-ai (approved):

PR #75528, opened 2026-06-29
[BugFix] Fix sync MV rewrite losing a rollup column for same-column min/max

stephen-shelby:
## Why I'm doing:

When a query takes two aggregates on the **same base column** of a sync materialized view / rollup (e.g. `min(c)` and `max(c)`), the optimizer fails during cost estimation with `missing statistic of col: ... mv_min_c` (also surfaces as `Invalid plan: ... required cols cannot obtain from input cols`).

Each aggregate produces its own `RewriteContext` that shares the base column ref but maps to a different rollup column (`mv_min_c` / `mv_max_c`). The contexts are applied **sequentially**: the first one removes the shared base column from the `LogicalOlapScanOperator` (and replaces it in the intermediate `LogicalProjectOperator`). The second (sibling) context can then no longer find the base column, so its rollup column is never projected. The rewritten aggregate ends up referencing a column the scan/project does not produce, and the optimizer throws while costing the plan.

## What I'm doing:

Rewrite all aggregates that share the same base column as **one group, in a single pass**, instead of one context at a time:

- `MaterializedViewRule`: group the rewrite contexts by their shared base query column, and rewrite each group in a single pass.
- `MaterializedViewRewriter`: a rewrite now operates on the whole group, so the scan removes the base column **once** and adds **every** rollup column derived from it together (`mv_min_c` + `mv_max_c`); the project and aggregate likewise rewrite all aggregates of the group. Because the base column is no longer removed out from under a sibling, the scan/project rewrite stays at the plain "base column present -> replace" form, with no per-operator fallback.
- Add plan UTs in `MVRewriteTest`: same-column `min(c), max(c)`; regression controls (single min, single max, different-column min/max, two mins, min+sum); and two `min/max` pairs.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/mv/MaterializedViewRewriter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/mv/MaterializedViewRewriter.java
@@ -48,13 +48,30 @@ import java.util.Map;
 
 import static com.starrocks.catalog.Function.CompareMode.IS_IDENTICAL;
 
-public class MaterializedViewRewriter extends OptExpressionVisitor<OptExpression, MaterializedViewRule.RewriteContext> {
+/**
+ * Rewrites a query plan to read a sync materialized view (rollup) index.
+ *
+ * <p>The unit of rewrite is a group of {@link MaterializedViewRule.RewriteContext} that all share the
+ * same base query column (their {@code queryColumnRef} is identical). All aggregates on that base column
+ * are rewritten together, so the base column is removed from the scan exactly once and every rollup column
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

Youngwb (approved):

HangyuanLiu (approved):

PR #75538, opened 2026-06-29
[BugFix] Apply CAST semantics for string->number schema-change conversion

Hyper-FF:
## Why I'm doing:

ALTER TABLE ... MODIFY COLUMN from VARCHAR to an integer/float type went through StringToOtherTypeConverter, which used the lenient TypeInfo::from_string. That silently 32-bit-wrapped out-of-range values ('99999999999' -> 1215752191) and defaulted unparseable ones ('abc' -> 0), storing them as NON-NULL — diverging from query-time CAST and INSERT, which yield NULL for the same inputs.

Parse integer/float targets with StringParser (mirroring cast_int_from_string_fn / cast_float_from_string_fn) and store NULL on parse-failure/overflow. The converted column is nullable, so NULL is representable. Fixes the VARCHAR->INT case  and the numeric part of the VARCHAR->numeric generalization.

## What I'm doing:

Fixes #issue

```
source VARCHAR value   target T    query-time CAST   stored after MODIFY COLUMN (WRONG)
'not-a-date'           DATE        NULL              1400-01-01
'9999-99-99'           DATE        NULL              1400-01-01
''                     DATE        NULL              1400-01-01
NULL                   DATE        NULL              NULL          (correct)
'abc'                  INT         NULL              0
'99999999999999999999' INT         NULL (overflow)   -1
'xyz'                  DOUBLE      NULL              0
''                     DOUBLE      NULL              0
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/column/type_converter.cpp
+++ b/be/src/column/type_converter.cpp
@@ -14,12 +14,14 @@
 
 #include "column/type_converter.h"
 
+#include <cmath>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
 #include <string>
 
 #include "base/hash/unaligned_access.h"
+#include "base/string/string_parser.hpp"
 #include "column/type_converter_detail.h"
 #include "gutil/strings/substitute.h"
 #include "types/datetime_value.h"
@@ -29,8 +31,6 @@
 #include "types/storage_type_traits.h"
 #include "types/timestamp_value.h"
 #include "types/type_info.h"
-#include "base/string/string_parser.hpp"
-#include <cmath>
 
 namespace starrocks {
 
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

kangkaisen (approved):

PR #75542, opened 2026-06-29
[Enhancement] Build vector index for shared-data bundle segments

sevev:
## Why I'm doing:

In shared-data mode, small-batch loads trigger file bundling: each tablet in a partition writes its single segment as a slice into one shared physical file. Bundle segments previously skipped vector index creation entirely, so their data fell back to brute-force vector search until compaction later rewrote them into standalone segments.

The root cause was a `.vi` filename collision. The `.vi` name was derived from the segment filename, but for a bundle that filename is the shared bundle file (identical across the tablets sharing it), so multiple tablets resolved to the same `.vi` path and would overwrite each other. To avoid that, the writer simply skipped index creation for bundle segments.

## What I'm doing:

Make the `.vi` filename include the tablet id: `gen_vector_index_filename` and `gen_vector_index_path_from_segment_path` now take a required `tablet_id`, so each tablet's bundle slice gets a distinct `.vi` (`<segment_base>_<tablet_id>_<index_id>.vi`). With naming unambiguous, the writer now builds bundle vector indexes for both modes — sync inline, and async via the deferred lake build task — always deriving the `.vi` path from the shared bundle filename so the writer, reader, vacuum/GC, txn-abort cleanup, and the build task all agree on the path. Making `tablet_id` a required parameter (no default) forces every call site to update, so a missed site fails to compile rather than silently mis-naming a `.vi` (which could let GC delete a live index).

It also unifies build-threshold resolution into a single `resolve_vector_index_build_threshold` shared by the inline (`VectorIndexWriter`) and async (lake) paths, applying the IVFPQ `nlist` floor last so a user-supplied `index_build_threshold` can never drop below `nlist`, which faiss requires to train (the async path previously lacked this floor). Shared-nothing naming (`IndexDescriptor::vector_index_file_path`) is untouched, and there are no FE, proto, or scheduler changes.

Covered by unit tests under `be/test/storage/lake` (filenames, rowset, shared-data vector index, vector index build task, lake service) and `be/test/storage/index`.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/index/vector/vector_index_writer.cpp
+++ b/be/src/storage/index/vector/vector_index_writer.cpp
@@ -14,7 +14,11 @@
 
 #include "storage/index/vector/vector_index_writer.h"
 
+#include <algorithm>
+#include <cerrno>
 #include <cmath>
+#include <cstdlib>
+#include <limits>
 
 #include "column/nullable_column.h"
 #include "column/raw_data_visitor.h"
@@ -46,30 +50,10 @@ void VectorIndexWriter::create(const std::shared_ptr<TabletIndex>& tablet_index,
 }
 
 Status VectorIndexWriter::init() {
-    // Step 1: enforce minimum threshold by algorithm type.
-    // IVFPQ requires training points >= nlist. If the input is below this,
-    // faiss will throw, so we floor the threshold at nlist to skip the build
-    // rather than crash. Other algorithms (HNSW etc.) keep the default
-    // threshold from config so callers that want "build immediately" must
-    // explicitly request it via index_build_threshold below.
-    auto index_type_it = _tablet_index->common_properties().find("index_type");
-    if (index_type_it != _tablet_index->common_properties().end()) {
-        std::string index_type = index_type_it->second;
-        std::transform(index_type.begin(), index_type.end(), index_type.begin(), ::tolower);
-        if (index_type == "ivfpq") {
-            auto nlist_it = _tablet_index->index_properties().find("nlist");
-            if (nlist_it != _tablet_index->index_properties().end()) {
-                auto nlist = static_cast<uint32_t>(std::atoi(nlist_it->second.c_str()));
-                _start_vector_index_build_threshold = std::max(_start_vector_index_build_threshold, nlist);
-            }
-        }
-    }
-
-    // Step 2: user-specified property has final control over threshold.
-    auto find_result = _tablet_index->common_properties().find("index_build_threshold");
-    if (find_result != _tablet_index->common_properties().end()) {
-        _start_vector_index_build_threshold = std::atoi(find_result->second.c_str());
-    }
+    // Threshold resolution is the single source of truth shared with the async build
+    // path (lake::get_vector_index_build_threshold); see resolve_vector_index_build_threshold
+    // for the precedence rules (config default -> user override -> IVFPQ nlist floor).
+    _start_vector_index_build_threshold = resolve_vector_index_build_threshold(*_tablet_index);
     return Status::OK();
 }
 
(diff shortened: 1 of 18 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/lake/filenames.h
```diff
@@ -103,28 +103,33 @@ inline std::string gen_segment_filename(int64_t txn_id) {
...
+// Generate vector index filename from segment filename. The tablet_id disambiguates
+// segments that share a physical file across tablets (file bundling), where the recorded
+// segment filename is the shared bundle name and the offset alone is not on the read path;
+// without it those tablets would collide on the same .vi path and overwrite each other.
+// Non-bundled segments carry it too (redundant but keeps naming uniform, no special-casing).
+// e.g. "0123_abcd.dat" + tablet 9 + index 123 -> "0123_abcd_9_123.vi"
+inline std::string gen_vector_index_filename(std::string_view segment_filename, int64_t tablet_id, int64_t index_id) {
     if (segment_filename.ends_with(".dat")) {
-        return fmt::format("{}_{}.vi", segment_filename.substr(0, segment_filename.size() - 4), index_id);
+        return fmt::format("{}_{}_{}.vi", segment_filename.substr(0, segment_filename.size() - 4), tablet_id, index_id);
```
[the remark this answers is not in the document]
sevev: The vector index for shared-data is not release yet, so there are no compatibility issues.

be/src/storage/lake/filenames.h
```diff
@@ -103,28 +103,33 @@ inline std::string gen_segment_filename(int64_t txn_id) {
...
     }
-    return fmt::format("{}_{}.vi", segment_filename, index_id);
+    return fmt::format("{}_{}_{}.vi", segment_filename, tablet_id, index_id);
```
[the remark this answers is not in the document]
sevev: the vector index for shared-data is not release yet and there are no compatibility issues

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/lake/vacuum.cpp
```diff
@@ -157,11 +157,11 @@ int64_t calculate_retry_delay(int64_t last_delay, int64_t base, int64_t max_retr
 namespace {
 
 // Delete .vi files for segments in a rowset using segment_metas metadata.
-static Status delete_rowset_vi_files(AsyncFileDeleter* deleter, const std::string& base_dir,
+static Status delete_rowset_vi_files(AsyncFileDeleter* deleter, const std::string& base_dir, int64_t tablet_id,
                                      const RowsetMetadataPB& rowset) {
     for (const auto& segment_meta : rowset.segment_metas()) {
         for (int64_t vi_id : segment_meta.vector_index_ids()) {
-            auto vi_name = gen_vector_index_filename(segment_meta.filename(), vi_id);
+            auto vi_name = gen_vector_index_filename(segment_meta.filename(), tablet_id, vi_id);
```
luohaha, line 164: How about old files? Will they become orphan files?

luohaha (approved):

be/src/storage/lake/vacuum.cpp
sevev: > How about old files? Will they become orphan files?

This feature is not release yet so there is no historical issues.

srlch:
@codex review

be/src/storage/rowset/segment_iterator.cpp
```diff
@@ -1110,8 +1110,8 @@ Status SegmentIterator::_init_ann_reader() {
 #ifdef WITH_TENANN
         std::string index_path;
         if (_opts.belonged_to_cloud_native) {
-            index_path =
-                    lake::gen_vector_index_path_from_segment_path(_segment->file_name(), tablet_index_meta->index_id());
+            index_path = lake::gen_vector_index_path_from_segment_path(_segment->file_name(), _opts.tablet_id,
+                                                                       tablet_index_meta->index_id());
```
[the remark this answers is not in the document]
sevev: > **P2 Badge Preserve vector-index paths across tablet ownership changes**
> 
> Using the reader's current `_opts.tablet_id` to derive the `.vi` name makes existing metadata unstable when the same `SegmentMetadataPB` is reused under another tablet: split/merge code carries `filename` and `vector_index_ids` forward without recording the tablet that originally wrote the index, so a post-split tablet probes `<segment>_<newTablet>_<index>.vi` while the live file was written as `<segment>_<oldTablet>_<index>.vi` (and pre-upgrade files use the old `<segment>_<index>.vi` form). In those cases `_init_reader_from_file` treats the new path as NotFound and disables ANN, so all affected rowsets fall back to brute force until compaction rewrites them; please persist/derive a stable owner/name or try the legacy/original path before disabling the index.
> 
> Useful? React with 👍 / 👎.

The vector index for shared-data is not release yet, so there are no compatibility issues.

be/src/storage/index/vector/vector_index_writer.cpp
```diff
@@ -46,30 +50,10 @@ void VectorIndexWriter::create(const std::shared_ptr<TabletIndex>& tablet_index,
...
-            }
-        }
-    }
-
-    // Step 2: user-specified property has final control over threshold.
-    auto find_result = _tablet_index->common_properties().find("index_build_threshold");
-    if (find_result != _tablet_index->common_properties().end()) {
-        _start_vector_index_build_threshold = std::atoi(find_result->second.c_str());
-    }
+    // Threshold resolution is the single source of truth shared with the async build
```
srlch, line 53: Is it the relative change with the bundle segments?

sevev: > Is it the relative change with the bundle segments?

It is a small refactor to reduce duplicate logic code and it has little to do with the logic of bundle segments actually.

srlch (approved):

PR #75544, opened 2026-06-29
[BugFix] Fix NullPointerException in NativeAccessController for unqualified external-catalog tables

Martozar:
## Why I'm doing:

An unqualified table reference over an external (e.g. Iceberg) catalog can reach
`NativeAccessController.checkTableAction` with a `TableName` whose `db` is `null`. The
unconditional `Preconditions.checkNotNull(tableName.getDb())` then throws a
`NullPointerException` during planning, which surfaces to clients as the opaque
`1064 (HY000): Unknown error`.

## What I'm doing:

Resolve a `null` db to the session's current database (`context.getDatabase()`) before the
privilege check, mirroring the existing null-catalog fallback on the line above. The check
stays fail-closed: if neither the `TableName` db nor the current database is set it still
throws via `Preconditions.checkNotNull(db)`, rather than letting a `null` db flow into
`checkObjectTypeAction`. Privilege enforcement is unchanged — the check runs against the
database the unqualified name resolves to.

Test: `AuthorizationMgrTest.checkTableActionResolvesNullDbToCurrentDatabase` covers null-db
resolution to the current database (`SELECT` granted → allowed) and the fail-closed path
(no privilege → `AccessDeniedException`, not NPE).

Note: the sibling `checkAnyActionOnTable` has the identical `checkNotNull(getDb())` pattern;
left unchanged as I could not reproduce a `null` db reaching it — happy to follow up if a
maintainer prefers.

## What type of PR is this:

- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool
- [x] BugFix

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
+++ b/fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
@@ -90,10 +90,15 @@ public class NativeAccessController implements AccessController {
     public void checkTableAction(ConnectContext context, TableName tableName, PrivilegeType privilegeType)
             throws AccessDeniedException {
         String catalog = tableName.getCatalog() == null ? InternalCatalog.DEFAULT_INTERNAL_CATALOG_NAME : tableName.getCatalog();
-        Preconditions.checkNotNull(tableName.getDb());
+        // An unqualified table reference can reach this check with a null db (for example a table over an
+        // external catalog whose database was not resolved onto the TableName). Fall back to the session's
+        // current database, mirroring the null-catalog handling above. Keep the check fail-closed: if neither
+        // is available the privilege check still throws rather than letting a null db flow downstream.
+        String db = tableName.getDb() != null ? tableName.getDb() : context.getDatabase();
+        Preconditions.checkNotNull(db);
 
         checkObjectTypeAction(context, privilegeType, ObjectType.TABLE,
-                Arrays.asList(catalog, tableName.getDb(), tableName.getTbl()));
+                Arrays.asList(catalog, db, tableName.getTbl()));
     }
 
     @Override
--- a/fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
+++ b/fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
@@ -89,11 +89,12 @@ public class NativeAccessController implements AccessController {
     @Override
     public void checkTableAction(ConnectContext context, TableName tableName, PrivilegeType privilegeType)
             throws AccessDeniedException {
-        String catalog = tableName.getCatalog() == null ? InternalCatalog.DEFAULT_INTERNAL_CATALOG_NAME : tableName.getCatalog();
-        // An unqualified table reference can reach this check with a null db (for example a table over an
-        // external catalog whose database was not resolved onto the TableName). Fall back to the session's
-        // current database, mirroring the null-catalog handling above. Keep the check fail-closed: if neither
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
```diff
@@ -90,10 +90,15 @@ public void checkAnyActionOnDb(ConnectContext context, String catalogName, Strin
     public void checkTableAction(ConnectContext context, TableName tableName, PrivilegeType privilegeType)
             throws AccessDeniedException {
         String catalog = tableName.getCatalog() == null ? InternalCatalog.DEFAULT_INTERNAL_CATALOG_NAME : tableName.getCatalog();
```
jaogoy: It seems that it need to use `getCurrentCatalog()` when null rather then internalCatalog directly.

fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
```diff
@@ -90,10 +90,15 @@ public void checkAnyActionOnDb(ConnectContext context, String catalogName, Strin
...
-        Preconditions.checkNotNull(tableName.getDb());
+        // An unqualified table reference can reach this check with a null db (for example a table over an
+        // external catalog whose database was not resolved onto the TableName). Fall back to the session's
+        // current database, mirroring the null-catalog handling above. Keep the check fail-closed: if neither
+        // is available the privilege check still throws rather than letting a null db flow downstream.
+        String db = tableName.getDb() != null ? tableName.getDb() : context.getDatabase();
```
jaogoy, line 98: It's good to do fail-closed check in this funciton.
But, it'll be better to make the `TableName` more correct in the outer invoker to set Tablename properly.

jaogoy:
@StarRocks-Reviewer review it

fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
Martozar: Good catch — done in a96b844. `checkTableAction` now resolves a null catalog to `context.getCurrentCatalog()` (which itself defaults to the internal catalog), instead of hardcoding `InternalCatalog.DEFAULT_INTERNAL_CATALOG_NAME`. This is strictly more correct: an unqualified table over an external catalog now resolves against the session's actual current catalog, matching the null-db fallback to `context.getDatabase()`.

fe/fe-core/src/main/java/com/starrocks/authorization/NativeAccessController.java
Martozar: Agreed the fail-closed guard is the safety net, not the ideal fix. I traced where the null-db `TableName` originates for the SELECT repro: it comes from the `tableNameTableObj` map built during analysis + the optimizer's `ScanColumnCollector` (`ColumnPrivilege.check`), and is then consumed by several authz call sites (`checkTableAction`/`checkViewAction`/`checkColumnAction`). So normalizing at the source means ensuring relation resolution and the optimizer populate the catalog/db onto every `TableName` — that touches the analyzer/planner query path fairly broadly, rather than being a single clean invoker.

I'd prefer to keep this contained, defensive resolution in the controller for this bugfix (it also protects the other unqualified-name entry points), and file a follow-up for normalizing `TableName` in the analyzer/optimizer if you'd like that tracked separately. Happy to go either way if you feel the source fix belongs in this PR.

Issue #74913, opened 2026-06-16
Add IF NOT EXISTS / IF EXISTS support for idempotent plugin install/uninstall


s4ch1n:
### Problem

`INSTALL PLUGIN` and `UNINSTALL PLUGIN` have no idempotency guards. Running either statement in an automated or repeated context (bootstrap scripts, CI pipelines) causes hard failures. There is no way today to express "install this plugin if it isn't already there" or "remove it if it exists." Every caller must pre-check SHOW PLUGINS or catch exceptions in application code, neither is appropriate for SQL-level orchestration. 

```
-- Second run of a setup script fails with - StarRocksException: plugin audit_plugin has already been installed.
INSTALL PLUGIN FROM '/opt/starrocks/plugins/audit_plugin.zip';

-- Teardown script fails if plugin was never installed - Plugin audit_plugin does not exist
UNINSTALL PLUGIN audit_plugin;
```

### Proposed Syntax
```
-- Install only if not already installed; silently succeeds if already present
INSTALL PLUGIN IF NOT EXISTS FROM 'path/to/plugin.zip' [PROPERTIES(...)];

-- Uninstall only if installed; silently succeeds if already absent
UNINSTALL PLUGIN IF EXISTS plugin_name;
```

This follows the exact same IF NOT EXISTS / IF EXISTS pattern used by every other DDL in StarRocks — CREATE TABLE, DROP TABLE, CREATE FUNCTION, DROP
FUNCTION, DROP CATALOG, DROP RESOURCE GROUP, etc.

The flag does not change what happens in the normal path — it only suppresses the error in the idempotent path.

kmozaid:
Please assign this issue to me, I will work on it.

kmozaid:
I have started working on it, committed changes to my fork repo https://github.com/kmozaid/starrocks/commit/e33549c051096d4d0985e81db2cef6cee4b8a8c3.

- I need to test the changes 

Once tested thoroughly, I will raise a PR.

s4ch1n:
@kmozaid  I forgot to update here, I actually have a fix ready for this since we use this feature internally. I was hoping to get the PR up today, but didn't get a chance. I'll get it out early next week.

PR #75563, opened 2026-06-29
[Enhancement] Make plugin install/uninstall idempotent

s4ch1n:
## Why I'm doing:

Add IF NOT EXISTS / IF EXISTS support for idempotent plugin install/uninstall

https://github.com/StarRocks/starrocks/issues/74913

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/plugin/PluginMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/plugin/PluginMgr.java
@@ -143,6 +143,9 @@ public class PluginMgr implements Writable {
             }
 
             if (checkDynamicPluginNameExist(info.getName())) {
+                if (stmt.isIfNotExists()) {
+                    return null;
+                }
                 throw new StarRocksException("plugin " + info.getName() + " has already been installed.");
             }
 
@@ -169,6 +172,12 @@ public class PluginMgr implements Writable {
 
     public void uninstallPluginFromStmt(UninstallPluginStmt stmt) throws IOException, StarRocksException {
         String pluginName = stmt.getPluginName();
+        if (!checkDynamicPluginNameExist(pluginName)) {
+            if (stmt.isIfExists()) {
+                return;
+            }
+            throw new DdlException("Plugin " + pluginName + " does not exist");
+        }
         int typeId = uninstallPlugin(pluginName);
 
         GlobalStateMgr.getCurrentState().getEditLog().logUninstallPlugin(new UninstallPluginLog(pluginName), wal -> {
--- a/fe/fe-core/src/main/java/com/starrocks/plugin/PluginMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/plugin/PluginMgr.java
@@ -144,7 +144,8 @@ public class PluginMgr implements Writable {
 
             if (checkDynamicPluginNameExist(info.getName())) {
                 if (stmt.isIfNotExists()) {
-                    return null;
+                    LOG.info("install plugin[{}] which already exists", info.getName());
+                    return info;
                 }
                 throw new StarRocksException("plugin " + info.getName() + " has already been installed.");
             }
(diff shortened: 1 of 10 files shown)


s4ch1n:
@HangyuanLiu FE Incremental coverage report seems stale, It should be more than 80%, Is there a way to check for staleness in the FE coverage pipeline ? I don't have access to CI logs to check

dirtysalt (dismissed):
LGTM! 🚀

fe/fe-core/src/main/java/com/starrocks/plugin/PluginMgr.java
```diff
@@ -143,6 +143,9 @@ public PluginInfo installPlugin(InstallPluginStmt stmt) throws IOException, Star
             }
 
             if (checkDynamicPluginNameExist(info.getName())) {
+                if (stmt.isIfNotExists()) {
```
gengjun-git, line 146: Better to print a log here, and return the existing PluginInfo instead of null

fe/fe-core/src/main/java/com/starrocks/plugin/PluginMgr.java
```diff
@@ -169,6 +172,12 @@ public PluginInfo installPlugin(InstallPluginStmt stmt) throws IOException, Star
 
     public void uninstallPluginFromStmt(UninstallPluginStmt stmt) throws IOException, StarRocksException {
         String pluginName = stmt.getPluginName();
+        if (!checkDynamicPluginNameExist(pluginName)) {
+            if (stmt.isIfExists()) {
```
gengjun-git, line 177: Better to print a log here

fe/fe-core/src/main/java/com/starrocks/plugin/PluginMgr.java
```diff
@@ -169,6 +173,13 @@ public PluginInfo installPlugin(InstallPluginStmt stmt) throws IOException, Star
...
+                LOG.info("uninstall plugin[{}] which does not exist", pluginName);
+                return;
+            }
+            throw new DdlException("Plugin " + pluginName + " does not exist");
```
osscm, line 181: we can use `String.format`

osscm:
@s4ch1n @dirtysalt  - trying to understand, how does this plugin zip in installed? adding to the plugin folder? and is there any other check SR does before adding to the plugins....

@dirtysalt 
It is not about this PR, but in general, from security and safety purpose.
Should we have pre-check about what that plugin is trying to do, or this is SQL to do admin operations, instead of adding it manually to the build?

thanks

s4ch1n:
@osscm I'll try to answer some of the questions. 

> how does this plugin zip in installed? adding to the plugin folder? and is there any other check SR does before adding to the plugins....

A plugin zip can be installed from either a local path or a remote http/https URL. Once fetched, StarRocks extracts it and moves it into $STARROCKS_HOME/plugins/<plugin-name>/. [More info ](https://docs.starrocks.io/docs/sql-reference/sql-statements/cluster-management/plugin/INSTALL_PLUGIN/)

Before installing, SR runs a PLUGIN system privilege check, validates required fields in plugin.properties (name, type, version, classname), and confirms the plugin class can actually be loaded - among other checks.

> in general, from security and safety purpose. Should we have pre-check about what that plugin is trying to do, or this is SQL to do admin operations, instead of adding it manually to the build?

From a security standpoint though, By default, only `root` and [db_admin roles](https://github.com/StarRocks/starrocks/blob/main/fe/fe-core/src/main/java/com/starrocks/authorization/AuthorizationMgr.java#L200)  hold the PLUGIN privilege, so it's already restricted to admin-tier roles. Once loaded, the plugin class runs with the full permissions of the FE JVM process there's no sandboxing or static analysis of intent, but given the lack of sandboxing, admins should treat INSTALL PLUGIN as equivalent to deploying arbitrary code onto the FE, and vet the plugin source themselves beforehand.

gengjun-git (approved):

dirtysalt (approved):
LGTM! 🚀

HangyuanLiu (approved):

PR #75573, opened 2026-06-30
[BugFix] Resolve MODIFY COLUMN keyness consistently for comment-only detection

xiangguangyxg:
## Why I'm doing:

A keyness demotion expressed as `MODIFY COLUMN <col> <type>` (omitting `KEY`) on a shared-data range-distribution DUPLICATE/UNIQUE table was being classified as a comment-only no-op, so an invalid keyness flip (demoting the only key, or leaving a value before a key) slipped through unrejected instead of being rejected synchronously.

The masking came from `isCommentOnlyModification` inheriting an existing key column's key flag whenever the clause omitted `KEY`. That contradicted the documented behavior (a full-definition MODIFY COLUMN on a non-aggregation key column must repeat `KEY` to keep the column a key) and every other MODIFY COLUMN path (`processModifyColumn`, `resolveModifyColumnKeyness`, the analyzer, the range-distribution guards), all of which treat omitting `KEY` as a demotion.

## What I'm doing:

- `isCommentOnlyModification` now resolves the post-modify keyness through the shared `resolveModifyColumnKeyness` helper (PK inherits an existing key; AGG with no aggregation method is a key; DUP/UNIQUE take the clause's `KEY` flag), so a keyness flip is surfaced as a real schema change instead of being masked as comment-only.
- Extracted `routesModifyColumnToRangeRewrite(table, clause, postFlipBaseSchema)` = `needsRangeRewriteSchemaChange(...) && rangeRewriteKeySchemaIsValid(...)`, used by both the keyness-flip rejection-suppression gate in `processModifyColumn` and the range-rewrite routing site in `analyzeAndCreateJob`, so the two conditions that must mirror each other share one definition and cannot drift. Both sites pass the post-flip base index schema.
- A pure comment change on a key column still uses the bare `MODIFY COLUMN <col> COMMENT "..."` metadata-only form; a full-definition comment edit on a non-aggregation key column repeats `KEY` (matching the existing ALTER TABLE reference).

Verified locally: `RangeRewriteRoutingTest`, `SchemaChangeHandlerTest`, `LakeRangeRewriteSchemaChangeJobTest`, `RangeDistributionGuardTest`, and `AlterTest` all pass (170 tests); `mvn checkstyle:check -pl fe-core` reports 0 violations.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -741,11 +741,13 @@ public class SchemaChangeHandler extends AlterHandler {
             // Leave the "column does not exist" error to processModifyColumn.
             return false;
         }
-        // The KEY designation is table-level and is not repeated in the MODIFY COLUMN clause, so a rebuilt
-        // column inherits the stored key flag the same way processModifyColumn does.
-        if (!modColumn.isKey()) {
-            modColumn.setIsKey(oriColumn.isKey());
-        }
+        // Resolve the post-modify keyness exactly as processModifyColumn does (PK inherits an existing key;
+        // AGG with no aggregation method is a key; DUP/UNIQUE take the clause's KEY flag), so that a keyness
+        // flip is surfaced to the comparison below as a real change instead of being masked as comment-only.
+        // Per the ALTER TABLE reference, a full-definition MODIFY COLUMN on a non-aggregation key column must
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

PR #75578, opened 2026-06-30
[BugFix] Exclude NULL rows from NOT MATCH answered by GIN inverted index

### LINKED ISSUE #75577: [Bug]NOT MATCH on a column with a GIN inverted index wrongly returns NULL rows

On a column with an untokenized GIN inverted index, with a NULL value present:

CREATE TABLE t (
    id INT NOT NULL,
    c  VARCHAR(30) NULL,
    INDEX idx_c (c) USING GIN ("parser" = "none")
)
DUPLICATE KEY(id)

INSERT INTO t VALUES (1, 'hello'), (2, 'world'), (3, NULL);

SELECT id, c FROM t WHERE c NOT MATCH 'hello' ORDER BY id;

Only (2, 'world'). Under SQL three-valued logic, NOT MATCH over a NULL value is
NULL (not TRUE), so the NULL row must not be returned. The same query on a column
without the GIN index returns only id=2.

### Real behavior (Required)

Both (2, 'world') and (3, NULL) are returned. Adding a GIN index changes the query
result, which breaks index transparency.

main (reproduced on latest main). Affects versions that support the GIN inverted index.

cccaarr:
## Why I'm doing:

On a column with a GIN inverted index, a NOT MATCH predicate returns NULL rows
that it should not.

- Before: SELECT * FROM t WHERE c NOT MATCH 'hello'  ->  returns 'world' AND the NULL row
- After:  SELECT * FROM t WHERE c NOT MATCH 'hello'  ->  returns 'world' (NULL excluded)

NULL rows are kept in a separate null bitmap, not in the term postings. The NOT
path subtracted only the matched rows from the candidate set, so NULL rows
survived. Under SQL three-valued logic, NOT MATCH over a NULL value is NULL (not
TRUE), so those rows must not be returned.

## What I'm doing:

In ColumnExprPredicate::seek_inverted_index, the NOT branch now also reads the
null bitmap via iterator->read_null(...) and subtracts it (after subtracting the
matched rows), mirroring ColumnNotNullPredicate::seek_inverted_index. The
positive branch is unchanged.

Added a SQL regression test (test/sql/test_index/test_gin_index, case
test_gin_not_match_excludes_null) for NOT MATCH over a GIN column with a NULL row.

Fixes #75577 

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/storage/primitive/column_expr_predicate.cpp
+++ b/be/src/storage/primitive/column_expr_predicate.cpp
@@ -390,6 +390,9 @@ Status ColumnExprPredicate::seek_inverted_index(const std::string& column_name, 
 
     if (with_not) {
         *row_bitmap -= roaring_opt.value();
+        roaring::Roaring null_roaring;
+        RETURN_IF_ERROR(iterator->read_null(column_name, &null_roaring));
+        *row_bitmap -= null_roaring;
     } else {
         *row_bitmap &= roaring_opt.value();
     }
--- a/test/sql/test_index/R/test_gin_index
+++ b/test/sql/test_index/R/test_gin_index
@@ -251,3 +251,60 @@ PROPERTIES (
 drop database test_gin_index;
 -- result:
 -- !result
+-- name: test_gin_not_match_excludes_null @native
+create database test_gin_null_${uuid0};
+-- result:
+-- !result
+use test_gin_null_${uuid0};
+-- result:
+-- !result
+ADMIN SET FRONTEND CONFIG ("enable_experimental_gin" = "true");
+-- result:
+-- !result
+CREATE TABLE t_null (
+    id int NOT NULL,
+    c  varchar(30) NULL,
+    INDEX idx_c (c) USING GIN ("parser" = "none")
+)
+DUPLICATE KEY(id)
+DISTRIBUTED BY HASH(id) BUCKETS 1
+PROPERTIES ("replication_num" = "1");
+-- result:
+-- !result
+insert into t_null values (1, 'hello'), (2, 'world'), (3, NULL), (4, 'hxllo');
+-- result:
+-- !result
+[ORDER] select id, c from t_null where c match 'hello';
+-- result:
+1	hello
+-- !result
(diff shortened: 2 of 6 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/primitive/column_expr_predicate.cpp
```diff
@@ -390,6 +390,9 @@ Status ColumnExprPredicate::seek_inverted_index(const std::string& column_name,
 
     if (with_not) {
         *row_bitmap -= roaring_opt.value();
+        roaring::Roaring null_roaring;
+        RETURN_IF_ERROR(iterator->read_null(column_name, &null_roaring));
```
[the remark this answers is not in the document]
cccaarr: Fixed.

CelerData-Reviewer:
@codex review

test/sql/test_index/R/test_gin_index
```diff
@@ -250,4 +250,39 @@ PROPERTIES (
...
+use test_gin_null_${uuid0};
+-- result:
+-- !result
+ADMIN SET FRONTEND CONFIG ("enable_experimental_gin" = "true");
+-- result:
+-- !result
+CREATE TABLE t_null (
+    id int NOT NULL,
+    c  varchar(30) NULL,
+    INDEX idx_c (c) USING GIN ("parser" = "none")
```
srlch, line 267: Also need `imp_lib = 'builtin'` to cover the `builtin` version. The default value is `CLucence`.

srlch (approved):

satanson (approved):

test/sql/test_index/R/test_gin_index
cccaarr: Fixed.

PR #75583, opened 2026-06-30
[BugFix] Serve flat-JSON subfield reads for keys whose name contains '.'

Hyper-FF:
## Why I'm doing:

A JSON key whose name contains a literal '.' (e.g. {"a.b": 7}) is never written as a flat sub-column -- the flat-JSON path scheme uses '.' as the level separator, so such keys are kept in the `remain` blob (json_path_deriver skips them). On read, get_json_*(j,'$."a.b"') / a pushed PruneSubfield access path asks for the typed subfield "a.b". Two read paths mishandled it and returned NULL, which -- because the read also drives pushed-down predicates/joins/grouping -- became silent row loss and wrong aggregation, not just a wrong projected value:

1. JsonFlatPath::split_path split the requested path on every '.', shredding the FE-quoted key "a.b" into a bogus nested path "a"->"b" that matches nothing. Make it quote-aware: a "..."-wrapped level is one key (quotes stripped, interior '.' kept). No existing flat-leaf path starts with '"', so only the previously broken FE-wrapped dotted keys change behavior.

2. _extract_with_hyper (the default lazy/direct path, enable_lazy_dynamic_flat_json) gave up on a dotted key (in_flat=false), routing it through an empty-target "reconstruct whole JSON + real_path" flow that could not materialize the value from remain. Instead, wrap the dotted key in quotes and keep it as a flat path level so it rides the now quote-aware split and extracts from remain. A key whose name itself contains a '"' is still not representable and keeps the old fallback.

Keys colliding with the writer's reserved physical sub-columns ('nulls'/'remain') remain a separate, intentionally out-of-scope case.

## What I'm doing:

```
mysql>   select get_json_int(j,'$."a.b"')  from t ;
+----------------------------+
| get_json_int(j, '$."a.b"') |
+----------------------------+
|                          7 |
|                          8 |
+----------------------------+
2 rows in set (0.02 sec)

mysql> set cbo_prune_json_subfield = true;
Query OK, 0 rows affected (0.00 sec)

mysql>   select get_json_int(j,'$."a.b"')  from t ;
+----------------------------+
| get_json_int(j, '$."a.b"') |
+----------------------------+
|                       NULL |
|                       NULL |
+----------------------------+
2 rows in set (0.03 sec)

mysql> 
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/column/flat_json/json_flat_path.cpp
+++ b/be/src/column/flat_json/json_flat_path.cpp
@@ -22,17 +22,17 @@
 namespace starrocks {
 
 std::pair<std::string_view, std::string_view> JsonFlatPath::split_path(const std::string_view& path) {
-    size_t pos = 0;
-    pos = path.find('.', pos);
-    std::string_view key;
-    std::string_view next;
-    if (pos == std::string::npos) {
-        key = path;
-    } else {
-        key = path.substr(0, pos);
-        next = path.substr(pos + 1);
+    // A quoted level like "a.b" is a single key whose interior '.' is part of the name (the FE wraps a
+    // key containing the '.' separator in quotes, see SubfieldAccessPathNormalizer.formatJsonPath);
+    // resume the separator search after the closing quote, then strip the wrapping quotes so the level
+    // key equals the raw vpack object key, which carries none.
+    size_t scan = (!path.empty() && path.front() == '"') ? path.find('"', 1) : 0;
+    size_t dot = path.find('.', scan);
+    std::string_view key = path.substr(0, dot);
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

kangkaisen (approved):

PR #75595, opened 2026-06-30
[Enhancement] Align Hive text scanner behavior with Hive (quote/escape, ESCAPED BY, text maps, row splitting)

Joob1n:
## Why I'm doing:

When StarRocks reads Hive text tables (OpenCSVSerde or LazySimpleSerDe), some results are different from what Hive returns on the same data:

**1. Quotes are ignored (OpenCSVSerde).** A separator inside a quoted field is treated as a column break:

```sql
-- Hive
CREATE TABLE default.sr_csv_repro (name string, atime string)
ROW FORMAT SERDE 'org.apache.hadoop.hive.serde2.OpenCSVSerde'
WITH SERDEPROPERTIES ("separatorChar" = ",", "quoteChar" = "\"")
STORED AS TEXTFILE;

INSERT INTO default.sr_csv_repro
VALUES ('/data/INFANTS, PETS/report.csv', '2026-06-26 10:00:00');

-- Hive output:      name=/data/INFANTS, PETS/report.csv   atime=2026-06-26 10:00:00
-- StarRocks output: name=/data/INFANTS                    atime=PETS/report.csv
```

**2. `ESCAPED BY` is ignored (LazySimpleSerDe).** For a table created with `ESCAPED BY '\\'`, the line `a\,b,c` should give two fields, `a,b` and `c`. StarRocks splits at the escaped comma and returns `a\` and `b`.

**3. Map columns always read as NULL.** Hive text maps look like `k1:v1|k2:v2` (no braces). StarRocks only understood the `{k:v}` format, so every Hive text map column came back as NULL.

**4. A newline inside quotes breaks row counts.** Hive always cuts rows at physical line ends and then parses each line on its own, so a "quoted newline" becomes two all-null rows. StarRocks joined the two lines into one row — even `count(*)` differed from Hive.

**5. The `\N` null marker is checked at the wrong time.** Hive checks `\N` on the raw bytes *before* removing escapes: `\N` is NULL, while `\\N` is the literal string `\N`. StarRocks checked after unescaping, which inverts both cases.

## What I'm doing:

Follow the same two-layer design Hive itself uses (line splitting first, then the SerDe):

**1. FE** passes the quote/escape characters to the BE through `TextFileFormatDesc`:
- OpenCSVSerde: `quoteChar` / `escapeChar`, including their defaults (`"` / `\`), for both HMS and AWS Glue;
- LazySimpleSerDe: the `escape.delim` property written by `ESCAPED BY`.

**2. BE** first cuts rows at physical line boundaries (the same well-tested path that plain text tables already use), then splits the fields of one line with a small per-SerDe splitter:
- `split_hive_lazy_simple_line`: the escape char makes the next byte literal, and `\N` is checked on the raw bytes before unescaping;
- `split_hive_open_csv_line`: byte-compatible with the opencsv 2.3 parser that Hive 0.14–4.0 bundles, keeping its exact rules — for example, the escape char only escapes a quote or another escape (so `a\,b` still splits), and a line that ends inside quotes keeps only the finished fields (an all-null row if none finished).

**3. `MapConverter`** learns the Hive text map format: entries split on the collection delimiter, key/value at the first mapkey delimiter, and an empty field is an empty map `{}`.

Only the Hive text scanner is affected. Broker load / stream load / `FILES()` use a different scanner and are not touched.

**How this is verified:**
- `opencsv23_golden.tsv`: all 3906 possible input lines (length 0–5 over `{a , " \ space}`) were run through Hive's own bundled opencsv 2.3 parser; the new splitter must produce identical output for every one of them.
- New scanner tests: escaped separators, `\N` vs `\\N`, text maps, quoted newlines, CRLF/BOM, scan-range splits.
- A 17-case end-to-end audit against a real Hive 3.1 cluster now matches Hive exactly.

**Hive version note:** the opencsv 2.3 semantics above hold for every Hive release from 0.14 through 4.0. Hive 4.1 replaced opencsv 2.3 with 5.9 ([HIVE-28248](https://issues.apache.org/jira/browse/HIVE-28248), a mechanical dependency bump with no test-expectation changes on Hive's side), which silently alters a few edge-case parsing rules in Hive itself — e.g. the escape char can now also protect a separator (`a\,b` no longer splits), and an escape becomes effective at the start of an unquoted field. This PR targets the 2.3 semantics (all current mainstream deployments); aligning with 4.1+ would be a follow-up with a golden fixture regenerated against opencsv 5.9. LazySimpleSerDe's splitting semantics are unchanged across all Hive versions.

**Performance** (single-threaded bench `be/src/bench/csv_scanner_bench.cpp`, Scan level = split + convert to typed Columns, real file I/O):

**1. v2 vs v1 on the same data (no escape/quote — v1 is the untouched production path):**

| Scenario | v1 | v2 | v2 vs v1 |
|---|---|---|---|
| LazySimple, 2GB | 14381 ms | 15692 ms | +9.1% |
| LazySimple, 10GB | 69295 ms | 72911 ms | +5.2% |
| ARRAY/MAP columns, 2GB | 29492 ms | 30164 ms | +2.3% |
| ARRAY/MAP columns, 10GB | 144607 ms | 146597 ms | +1.4% |

Tables without quote/escape still take the v1 path in production (`_use_v2` is only set when `enclose`/`escape` is configured), so existing tables don't even pay this single-digit overhead. v2's convert time is identical to v1 (<1%); the overhead is entirely in the serde-aware field splitting.

**2. Cost of escape handling (v2 only — v1 never parsed these correctly):**

| Scenario | no escape | `ESCAPED BY` | cost |
|---|---|---|---|
| LazySimple, 2GB | 15692 ms | 19322 ms | +23.1% |
| LazySimple, 10GB | 72911 ms | 94629 ms | +29.8% |
| ARRAY/MAP with escaped separators inside elements, 2GB | 30164 ms | 35222 ms | +16.8% |
| ARRAY/MAP with escaped separators inside elements, 10GB | 146597 ms | 181046 ms | +23.5% |

Concentrated in the convert stage (where null-detection/unescaping now lives). This is the price of correct results on data that was previously parsed wrong, not a regression on any working path.

**3. Why line-first splitting (vs the earlier `more_rows()` state-machine approach):**

An earlier iteration of this PR routed the Hive path through `CSVReader::more_rows()`, the generic per-character escape/enclose state machine. Besides not matching Hive's quoted-newline semantics, it was much slower (LazySimple, 256MB, same machine):

| | parse throughput | Scan-level overhead vs v1 |
|---|---|---|
| `more_rows()` state machine (earlier iteration) | 227 MiB/s (~4.3× slower than v1) | ~+51% |
| line-first + per-serde splitter (this PR) | 649 MiB/s (~1.5× slower than v1) | +5-9% |

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

be/src/exec/hdfs_scanner/hdfs_scanner_text.cpp
```diff
@@ -396,25 +450,111 @@ Status HdfsTextScanner::_parse_csv(int chunk_size, ChunkPtr* chunk) {
...
+    size_t rows_read = 0;
+    CSVRow row;
+    for (; rows_read < chunk_size; rows_read++) {
+        // next_record(CSVRow&) drives CSVReader::more_rows, which is quote/escape
+        // aware: separators inside enclosed fields are kept as data. We call it
+        // through the base CSVReader pointer because HdfsScannerCSVReader declares
+        // its own next_record(Record*), which would otherwise hide this overload.
+        // EOF is returned on its own call (with empty columns), exactly like the
+        // Record* path used by _parse_csv.
+        Status status = _reader->next_record(row);
```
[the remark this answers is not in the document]
Joob1n: v2 will use the Hive-behavior: blank line will be filled by null columns. How about this? @Wenjun7J

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/connector/hive/TextFileFormatDesc.java
```diff
@@ -32,17 +32,30 @@ public class TextFileFormatDesc {
 
     private final int skipHeaderLineCount;
 
+    // OpenCSVSerde quote (enclose) and escape characters. 0 means unset, in which
+    // case the BE keeps using the naive split_record path.
+    private final int enclose;
+
+    private final int escape;
```
Wenjun7J, line 39: Could this be expanded to more than 2 characters if we need to support it in the future?

Wenjun7J (approved):

fe/fe-core/src/main/java/com/starrocks/connector/hive/TextFileFormatDesc.java
Joob1n: OpenCSV and LazySimple does not support multi-chars delimited case. So we don't need to do anything for that. Until StarRocks supports the Non-default serialization.null.format or other SerDe

GavinMar (approved):

dirtysalt (approved):
LGTM! 🚀

Youngwb (approved):

PR #75635, opened 2026-07-01
[WIP][Enhancement] Support bounded incremental vacuum in shared-data lake

starrocks-xupeng:
## Why I'm doing:

In shared-data mode, a partition that accumulates a very large backlog of historical metadata versions (a "death-spiral" partition) can time out the one-shot AutoVacuum RPC before it deletes anything, so it never converges — the backlog only grows. The one-shot path walks the whole `prev_garbage_version` chain in a single round with no upper bound on work.

This adds a bounded, resumable ("incremental") vacuum protocol: each round commits the version range the previous round proposed (and the FE persisted) and then proposes the next bounded range plus a resume cursor, doing at most a fixed amount of work per round. A large backlog now drains across many rounds instead of repeatedly timing out one RPC. It works for both file-bundling and non-bundling tables, since lake metadata versions are partition-wide.

## What I'm doing:

- **lake_service.proto**: add `max_versions_per_round` to `VacuumRequest` (its presence alone selects the incremental protocol — no mode flag) and a new `VacuumStatePB` message — `{to_delete_low, to_delete_high, next_propose_start_version, alive_boundary_version}` — carried by both `VacuumRequest` (the range to commit + resume cursor + pass retain floor) and `VacuumResponse` (the newly proposed range), so the round-trip state is one unit. All fields optional/additive.
- **BE `vacuum.cpp`**:
  - `propose_metadata_range`: a compute-only walk that resumes a top-down descent from a cursor, caps it at `max_versions_per_round`, and reports the next delete range `[to_delete_low, to_delete_high)` (high = the intersection retain floor = MIN of the per-tablet retain boundaries), the resume cursor, and — on a fresh round — the pass retain floor. A tablet that folded the entry version away anchors its walk by stepping down to its nearest existing version, so a per-tablet hole does not zero the whole round. The legacy `collect_files_to_vacuum` / `vacuum_tablet_metadata` one-shot path is left untouched.
  - `commit_metadata_range`: physically delete a persisted `[to_delete_low, to_delete_high)` range idempotently, data-first then metadata, without consulting `grace_timestamp`. It anchors each tablet's `prev_garbage_version` chain by entering at (or just below) `to_delete_high`, stepping down to the nearest existing version when batch publish folded a partition-level version away per-tablet, and reads the pass retain floor to spare shared files still referenced by surviving tablets.
  - `VacuumTabletMetaVerionRange::intersect_low` takes the MAX of the per-tablet low ends so a bounded round never deletes bundled metadata below a tablet that has not yet collected it.
  - `vacuum_impl`: in incremental mode commit the previous range then propose the next and return it to the FE.
- **FE `PhysicalPartition` / `VacuumState`**: hold the per-partition vacuum state (to-delete range, resume cursor, pass retain floor) in one object. It persists in the FE image via `@SerializedName` and, between snapshots, via an `OP_MODIFY_PARTITION_VACUUM_STATE` edit log the coordinator writes each round under the table WRITE lock, so an in-flight pass survives an FE restart / failover and resumes from its cursor instead of restarting. The op is `@IgnorableOnReplayFailed` — the state is reconstructible (a lost round just restarts the pass, at worst redoing idempotent work), so a replay failure must not block FE startup.
- **FE `AutovacuumDaemon`**: drive the protocol — each round relays the state, commits the previous round's proposed range and proposes the next; advance the success watermark to the pass retain floor only when a pass drains; keep scheduling a partition while a pass is in flight. When a partition spans several nodes, the per-node proposals are recombined (MIN floor / MAX low / MAX cursor) into the same partition-level state a single node would compute — tablet-to-node routing is left unchanged.
- **Config**: `lake_autovacuum_max_versions_per_round` (default 500) caps the per-round work; the incremental protocol is always on and cannot be disabled (values `<= 0` are treated as `1`), because the legacy one-shot path (which walks the chain top-down) cannot safely resume after a bounded round leaves holes in it.
- **Observability (`VacuumVersion`)**: surface `lastSuccVacuumVersion` (the pass retain floor) as a `VacuumVersion` column in `SHOW PARTITIONS` and a `VACUUM_VERSION` column in `information_schema.partitions_meta` (new `TPartitionMetaInfo.vacuum_version` thrift field + SystemTable column + FE data source + BE schema scanner slot, appended so existing column indexes stay stable), so `VisibleVersion - VacuumVersion` reads as the version-span of the un-reclaimed backlog. Persisted via `@SerializedName`.
- **Carried BugFix (`cal_new_base_version`)**: this branch also carries the write-side fix that stops `cal_new_base_version` from adopting a metacache-only tablet-metadata version as `prev_garbage_version` under file bundling + batch-publish retry (adds `CacheOptions.skip_meta_cache`, reads durable storage for the index-version existence check). It is a standalone main-based fix tracked in its own PR and is included here so the incremental path can be tested end-to-end on PK + file-bundling tables.
- **Docs**: document the new config and the `partitions_meta` column in `en` / `zh` / `ja`.

Validated on a shared-data test cluster — bounded rounds progress band by band and converge (see the FE/BE `incremental vacuum` round logs). Unit tests are a follow-up.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

PR #75640, opened 2026-07-01
[BugFix] Skip stale IDG index for DCG-overlaid columns after column-mode partial update

meegoo:
## Why I'm doing:

On a shared-data (lake) **PRIMARY KEY** table, a **column-mode partial update**
(`partial_update_mode = column`) of a column that carries a **fast-path IDG index**
(`ALTER TABLE ... ADD INDEX ... USING BITMAP`, or `bloom_filter_columns` / NGRAMBF added via the
fast path) returned **silently wrong query results** — no error, no crash.

Reproduction (lake PK table, `v1` values seen by the index, then column-mode update `v1 → 777`
for a subset):

| table | index on the updated column | `count(*) WHERE v1=777` | correct? |
|---|---|---|---|
| no index | none | 2 | ✅ |
| `CREATE TABLE ... INDEX ... USING BITMAP` (footer) | footer | 2 | ✅ |
| `ALTER TABLE ... ADD INDEX ... USING BITMAP` (fast-path IDG) | IDG `.idx` | **0** | ❌ **WRONG** |

The updated rows were **missing** from equality queries on the new value, and still counted under
their old value, until the tablet's next compaction healed it.

**Root cause:** a column-mode partial update writes the updated column into a Delta Column Group
(`.cols`) and does **not** rebuild the column's index. At read time the DCG-overlaid column is
served by a DCG column iterator over `.cols` (the new values), but the read path still probed the
fast-path IDG index using the **base** segment id (`rowset_id + segment_id()`) and surfaced that
**stale** `.idx` (built over the pre-update base values) to the pruning gates
(`_apply_bitmap_index` for BITMAP; `has_{original,ngram}_bloom_filter_index()` set from the IDG
probe in `ScalarColumnIterator::init` for bloom/ngram). The IDG entry's version is the ADD-INDEX
version (`<= query_version`), so a later column-mode update does not filter it out, and the gate
pruned rows against values the column no longer holds.

A traditional footer index is unaffected: it lives on the base segment's `ColumnReader`, which the
DCG iterator's `.cols` reader does not inherit, so `has_bitmap_index()` is false there and the DCG
column falls back to a correct unindexed scan. Only the IDG path re-attached the base index to the
DCG column.

## What I'm doing:

When a column is served from a Delta Column Group for the query version, do **not** wire the IDG
loader for that column — matching the (correct) footer-index behavior. `idg_loader == nullptr` is
already the documented signal that keeps a reader on the footer-only path, so the DCG-overlaid
column falls back to a correct unindexed scan; base (non-overlay) columns are untouched and keep
the fast-path IDG index. Two read sites:

- `_init_column_iterator_by_cid` (covers bloom / ngram bloom): in the DCG branch, clear
  `iter_opts.idg_loader` before initializing the DCG column iterator, so the IDG probe in
  `ScalarColumnIterator::init` does not set `_has_idg_{original,ngram}_bf`.
- `_apply_bitmap_index` (covers BITMAP): only pass `opts.idg_loader` when the column is served from
  the base segment (`_get_dcg_segment(ucid) == nullptr`); for a DCG-overlaid column leave it unset
  so `new_bitmap_index_iterator` stays on the footer path.

The fast path for pure-base columns is fully preserved (no regression to the ADD INDEX fast-path
read capability). This covers all three IDG fast-path flavors (BITMAP, plain BLOOM_FILTER,
NGRAMBF). GIN inverted indexes are a separate index family and are not part of the IDG fast path,
so they are untouched here.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/rowset/segment_iterator.cpp
+++ b/be/src/storage/rowset/segment_iterator.cpp
@@ -1643,6 +1643,19 @@ Status SegmentIterator::_init_column_iterator_by_cid(const ColumnId cid, const C
         // create delta column iterator
         // TODO io_coalesce
         _column_iterators[cid] = std::move(col_iter);
+        // The column is served from a Delta Column Group (.cols) for this query
+        // version (column-mode partial update overlay). A fast-path IDG index
+        // (.idx sidecar) is keyed by the BASE segment id and was built over the
+        // BASE column values; it does NOT describe the overlaid DCG values and
+        // is not rebuilt on a column-mode update. Probing it here would let the
+        // DCG column iterator surface the STALE base index via
+        // has_{original,ngram}_bloom_filter_index(), and the pruning gate would
+        // filter rows against values the column no longer holds -> silently
+        // wrong results. Drop the IDG context so the DCG column falls back to an
+        // unindexed (correct) scan, matching footer-index behavior (a footer
+        // index lives on the base ColumnReader, which the DCG reader does not
+        // inherit, so it is never consulted for DCG-overlaid columns).
+        iter_opts.idg_loader = nullptr;
         opts.encryption_info = dcg_encryption_info;
         ASSIGN_OR_RETURN(auto dcg_file, _opts.fs->new_random_access_file(opts, dcg_filename));
         iter_opts.read_file = dcg_file.get();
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

sevev (approved):

luohaha (approved):

luohaha:
I think it doesn't make sense that index become useless after partial update. Need to improve it later.

PR #75653, opened 2026-07-01
[Enhancement] Improve Iceberg external table statistics: persistent predicate columns, E2E auto-ANALYZE wiring, and stats-source metrics

dirtysalt:
## Why I'm doing:
External table statistics collection has two gaps. First, the set of predicate/join/group-by columns recorded during queries is only held in memory (`ConnectorTableTriggerAnalyzeMgr`), so it is lost on FE restart and invisible to other FE nodes in HA setup — causing the background auto-ANALYZE job to analyze the wrong or full set of columns. Second, there is no observability into whether a query used background ANALYZE results (accurate) vs. foreground metadata estimation (approximate).

## What I'm doing:

**1. Persistent `external_predicate_columns` table**

Introduce `ExternalPredicateColumnsStorage`, backed by a new system table `_statistics_.external_predicate_columns` (PRIMARY KEY on `table_uuid, column_name`, string-keyed like `external_column_statistics`). `PredicateColumnsMgr` now routes connector-table column usages into this storage instead of the in-memory map in `ConnectorTableTriggerAnalyzeMgr`. The storage follows the same `restore → vacuum → persist` daemon cycle as `PredicateColumnsStorage`, ensuring durability across restarts. A `queryGlobalState()` method hits the table directly (not the in-memory cache) so the auto-collect job sees columns observed by all FE nodes.

**2. E2E auto-ANALYZE column scoping**

`StatisticsCollectJobFactory.createExternalAnalyzeJob()` now calls `ExternalPredicateColumnsStorage.queryGlobalState(tableUUID)` when no explicit column list is given, so the generated ANALYZE job is narrowed to predicate/join/group-by columns rather than all columns.

```
Query touches col_a, col_b in WHERE/JOIN
       ↓
ExternalPredicateColumnsStorage.record(uuid, col)   ← persisted across restarts, visible to all FEs
       ↓
Auto-ANALYZE job: ANALYZE TABLE t(col_a, col_b)     ← scoped to predicate columns only
```

**3. Stats-source counters**

Add two `LongCounterMetric` entries to `MetricRepo`, aligned with the existing `StatsSource` enum so monitoring maps directly to query profile output: `external_stats_analyze_total` (stats served from background ANALYZE) and `external_stats_table_metadata_total` (stats served from foreground metadata estimation). Counters are incremented in `MetadataMgr.getTableStatistics()` (refactored to single-exit), guarded by `!table.isNativeTableOrMaterializedView()` so only external tables are counted.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

Youngwb (approved):

Astralidea (approved):

fe/fe-core/src/main/java/com/starrocks/statistic/columns/ExternalPredicateColumnsStorage.java
```diff
@@ -0,0 +1,346 @@
...
+
+    private static final Logger LOG = LogManager.getLogger(ExternalPredicateColumnsStorage.class);
+
+    public static final String DATABASE_NAME = StatsConstants.STATISTICS_DB_NAME;
+    public static final String TABLE_NAME = "external_predicate_columns";
+    public static final String TABLE_FULL_NAME = DATABASE_NAME + "." + TABLE_NAME;
+
+    private static final String TABLE_DDL =
+            "CREATE TABLE " + TABLE_NAME + "(" +
+            "table_uuid  VARCHAR(65530) NOT NULL," +
```
stephen-shelby, line 70: PK may exceed 128-byte limit: table_uuid (catalog.db.table.id) + column_name can easily exceed the default primary_key_limit_size (128 bytes), causing the insert to be rejected.

fe/fe-core/src/main/java/com/starrocks/statistic/columns/ExternalPredicateColumnsStorage.java
```diff
@@ -0,0 +1,346 @@
...
+                persistBatch(tableUUID, batch);
+                count += batch.size();
+            }
+        }
+
+        lastPersist = nextPersist;
+        LOG.info("[ExternalPredicateCols] persisted {} entries elapsed={}", count, watch);
+    }
+
+    private void persistBatch(String tableUUID, List<Map.Entry<String, ColumnEntry>> batch) {
```
stephen-shelby, line 251: Error handling issue: persistBatch() in persist() has no try/catch — one table's failure aborts the whole loop, blocking persistence for all other tables in that cycle, and lastPersist never advances so the failure repeats every cycle.

stephen-shelby:
external predicate column table's schema miss fe_id.

PR #63553, opened 2025-09-25
[BugFix] change CHECK to DCHECK in nullablecolumn to prevent the crash

murphyatwork:
## What I'm doing:

```
F20250923 23:42:04.217981 47716112008960 nullable_column.cpp:426] Check failed: _null_column->size() == _data_column->size() (4096 vs. 0)
    @          0x45e4183 starrocks::failure_function()
    @          0xb64b24a google::LogMessage::Fail()
    @          0xb64cc84 google::LogMessageFatal::~LogMessageFatal()
    @          0x46a0aad starrocks::NullableColumn::check_or_die() const
    @          0x7aa102d starrocks::JsonMergeIterator::next_batch(starrocks::SparseRange<unsigned int> const&, starrocks::Column*)
    @          0x704c3db starrocks::SegmentIterator::_read(starrocks::Chunk*, std::vector<unsigned int, std::allocator<unsigned int> >*, unsigned long)
    @          0x703b539 starrocks::SegmentIterator::_do_get_next(starrocks::Chunk*, std::vector<unsigned int, std::allocator<unsigned int> >*)
    @          0x7047351 starrocks::SegmentIterator::do_get_next(starrocks::Chunk*)
    @          0x70c6865 starrocks::ProjectionIterator::do_get_next(starrocks::Chunk*)

```

In `JsonMergeIterator`, the `Column::check_or_die` function is used to validate data and previously relied on `CHECK` to enforce this. We have now replaced `CHECK` with `DCHECK` to prevent crashes in release builds.

This change is a temporary workaround to address the crash issue but does not resolve the underlying problem.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
[...]

PR #75680, opened 2026-07-01
[BugFix] Handle NOT NULL -> nullable flat-JSON columns in compaction read path (fix JsonMergeIterator CHECK crash)

srihithg:
## Why I'm doing:

We hit a BE/CN crash loop on a shared-data cluster after a user ran
`ALTER TABLE t MODIFY COLUMN c json NULL` on a flat-JSON column that was originally `NOT NULL`.
Fast schema evolution makes that ALTER metadata-only, so the old segments on object storage are
never rewritten — they stay physically NOT NULL with no null sub-stream. The next background
compaction then reads those segments into a now-nullable output column and dies here:

```
CHECK((_null_iter == nullptr && null_column == nullptr) || (_null_iter != nullptr && null_column != nullptr));
```

The reader has `_null_iter == nullptr` (segment has no null stream) but the output is nullable
(`null_column != nullptr`), so the CHECK fires and SIGABRTs the whole process. Compaction retries,
crashes again, and the node never recovers. Reproduced on 4.0.12.

The thing is, this state is perfectly legal. A NOT NULL segment can't contain nulls, so the reader
just needs to fill the null map with "not null" instead of aborting.

## What I'm doing:

Fix is read-side only, all in `be/src/storage/rowset/json_column_iterator.cpp`. No writer, FE,
thrift or proto changes.

All 6 read methods on `JsonMergeIterator` and `JsonFlatColumnIterator` (both `next_batch`
overloads and `fetch_values_by_rowid` on each) now handle the two mismatch cases instead of
CHECKing:

- Segment NOT NULL, output nullable (the schema-evolution case): after the merge/read succeeds,
  append not-null flags to the output's null column for exactly the rows produced
  (`json_column->size() - before`), then `update_has_null()`. Same
  `append_value_multiple_times(&NOT_NULL, n)` idiom the flat-JSON writer already uses.
- Segment nullable, output non-nullable: there's nowhere to put the null flags, so return
  `Status::InternalError` instead of killing the process.

While in there I fixed two adjacent problems in the same family:
`JsonFlatColumnIterator::next_batch(size_t*)` never had the CHECK at all and would silently
under-fill the null column in release builds, and
`JsonFlatColumnIterator::fetch_values_by_rowid` down_cast the destination to `JsonColumn`
whenever the segment had no null stream — UB if the destination was actually nullable. Also moved
`check_or_die()` behind the success path, since running it on a half-filled column after a failed
read would just abort in a different place.

Tests added to `FlatJsonColumnCompactTest`:

- `CompactJsonNotNullToNullableSchemaEvolution` — NOT NULL flat segment read into a nullable
  column through all three read paths. This SIGABRTs without the fix. Asserts row count, that
  every synthesized null byte is 0, and that the JSON values round-trip.
- `CompactJsonMixedNullabilitySegments` — NOT NULL + nullable + NOT NULL segments merged into one
  output, so the backfill also runs against a non-empty destination and sits next to a real NULL.
- `CompactJsonNullableSegmentToNonNullableOutputError` — the inverse mismatch returns
  `InternalError` on all three paths instead of crashing.

Scope note: only the flat-JSON read paths (`JsonMergeIterator` / `JsonFlatColumnIterator`)
couple the segment's stored nullability to the output column like this. Non-flat JSON segments
are read through `ScalarColumnIterator`, and dynamic flattening (`JsonDynamicFlatIterator`)
reads through a `clone_empty()` proxy -- both maintain the output null column through the
Column append API and have no nullability-mismatch coupling, so they need no change.

The mismatch policy lives in two file-local helpers (`read_json_null_stream` /
`finish_json_read`) shared by all six read methods, so the planned ARRAY/MAP follow-up can
reuse the same shape instead of copy-pasting it.

One heads-up for reviewers: `array_column_iterator.cpp` (3x) and `map_column_iterator.cpp` (1x)
have the exact same CHECK and the same crash vector for ARRAY/MAP columns after the same kind of
ALTER. I'm deliberately not touching them here to keep this backportable and easy to review —
follow-up PR coming.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

(3.4 and 3.3 need manual backports — same as the earlier compaction crash fixes #63553 / #70539.)

### DIFF
--- a/be/src/storage/rowset/json_column_iterator.cpp
+++ b/be/src/storage/rowset/json_column_iterator.cpp
@@ -45,6 +45,56 @@
 
 namespace starrocks {
 
+namespace {
+
+struct JsonOutputColumns {
+    JsonColumn* json_column;
+    NullColumn* null_column; // nullptr when the output column is non-nullable
+};
+
+// Split the output column into its JSON data part and (optional) null part.
+JsonOutputColumns extract_json_output(Column* output) {
+    if (output->is_nullable()) {
+        auto* nullable = down_cast<NullableColumn*>(output);
+        return {down_cast<JsonColumn*>(nullable->data_column_raw_ptr()),
+                down_cast<NullColumn*>(nullable->null_column_raw_ptr())};
+    }
+    return {down_cast<JsonColumn*>(output), nullptr};
+}
+
+// Read the segment's null stream into the output null column via `read_nulls`.
+// The segment's stored nullability and the output column's nullability can
+// legitimately differ after fast schema evolution, so mismatches must not crash:
+// a segment with a null stream cannot be read into a non-nullable output, fail
+// with InternalError instead.
+template <typename ReadNullsFn>
+Status read_json_null_stream(Column* output, NullColumn* null_column, ColumnIterator* null_iter,
+                             ReadNullsFn read_nulls) {
+    if (null_iter != nullptr && null_column != nullptr) {
+        RETURN_IF_ERROR(read_nulls(null_column));
+        down_cast<NullableColumn*>(output)->update_has_null();
+    } else if (null_iter != nullptr && null_column == nullptr) {
+        return Status::InternalError("flat json: segment has a JSON null stream but the output column is non-nullable");
+    }
+    return Status::OK();
+}
+
+// Call after a successful data read. When the segment was written NOT NULL (no
+// null stream) but the output column is nullable -- a JSON column relaxed to
+// nullable by metadata-only fast schema evolution -- every row produced is
+// non-null, so append matching not-null flags to keep the null column in sync.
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #75687, opened 2026-07-02
[BugFix] Rewrite empty bundled partial-update segments instead of keeping them live

luohaha:
## Why I'm doing:

When file bundling is enabled (`enable_file_bundling = true`, the default for shared-data tables since 4.0), one physical `.dat` file holds the segments of several sibling tablets of a partition. A tablet whose partial-update slice received **no upserts** (e.g. it only had deletes) still gets a **0-row segment that points into the shared bundle file**.

`RowsetUpdateState::rewrite_segment` short-circuited on `rowset_meta.num_rows() == 0` and did nothing, so that raw bundle segment was left as a **live rowset segment**. It is not valid standalone data — its shared-ness is encoded by `bundle_file_offset`, not the `shared` flag — and, worse, it pins the shared bundle file. Once a sibling tablet **with** rows rewrites its own slice and orphans the bundle file, vacuum deletes the physical file while the empty rowset still references it. Any later publish on that tablet then fails at `prepare_primary_index -> Rowset::load_segments`:

```
Not found: load_segments failed ... starlet err Object s3://.../<txn>_<uuid>.dat does not exist
```

which wedges the whole table's publish queue. The lost segment is a 0-row slice, so no rows are actually lost, but publish never recovers.

## What I'm doing:

Drop the `num_rows == 0` short-circuit in `RowsetUpdateState::rewrite_segment`. The empty slice is now rewritten into a **private (non-bundled) segment** and the raw bundle segment is **orphaned**, so no rowset keeps a dangling reference to the shared bundle file. Truly segment-less rowsets (deletes only, no segment) are still short-circuited by the pre-existing `rewrite_segments_meta_size() == 0` check, so behavior for them is unchanged.

Reproduce: a 3-tablet file-bundled primary-key table; one tablet upserts a row while the other two only delete (their upsert segment is an empty bundled slice); then let vacuum delete the orphaned bundle file. Before the fix the two delete-only tablets wedge publish with "segment does not exist"; after the fix publish succeeds.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/rowset_update_state.cpp
+++ b/be/src/storage/lake/rowset_update_state.cpp
@@ -476,8 +476,19 @@ Status RowsetUpdateState::rewrite_segment(uint32_t segment_id, int64_t txn_id, c
     ASSIGN_OR_RETURN(auto fs, FileSystemFactory::CreateSharedFromString(root_path));
     std::shared_ptr<TabletSchema> tablet_schema = std::make_shared<TabletSchema>(params.metadata->schema());
     // get rowset schema
-    if (!params.op_write.has_txn_meta() || params.op_write.rewrite_segments_meta_size() == 0 ||
-        rowset_meta.num_rows() == 0) {
+    //
+    // NOTE: do NOT short-circuit on `rowset_meta.num_rows() == 0`. Under file bundling one physical
+    // .dat holds the segments of several sibling tablets, so a tablet whose slice received no rows
+    // still gets a (0-row) segment that points into the shared bundle file. If we skip the rewrite
+    // here, that raw bundle segment is left as a live rowset segment. It is not valid standalone
+    // data (its "shared-ness" is encoded by bundle_file_offset) and, worse, it pins the shared
+    // bundle file: once a sibling tablet with rows rewrites its own slice and orphans the bundle
+    // file, vacuum deletes the physical file while this empty rowset still references it, wedging
+    // publish with "segment does not exist". Letting the rewrite run turns the 0-row slice into a
+    // private (non-bundled) segment and orphans the raw bundle segment, so nothing keeps a dangling
+    // reference to a shared file. Truly segment-less rowsets are still short-circuited by the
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

luohaha:
Confirmed, this is a real P1. `SegmentRewriter::rewrite_partial_update()` short-circuits with `fs::copy_file(src.path, dest->path)` when there are no unmodified columns, and that copy ignores `src.bundle_file_offset` — for a bundled source it copies the entire physical bundle (all sibling slices) into the new private segment.

Dropping the `num_rows == 0` guard here is therefore unsafe: it routes empty bundled slices into that copy-only path. Reworking this PR to avoid rewriting empty bundled segments (the empty slice is dropped/orphaned rather than rewritten), and keeping the vacuum-side protection in the companion PR that flags bundled orphans as shared so `collect_garbage_files` never deletes a bundle file a sibling still references. Will update this PR accordingly.

luohaha:
Closing in favor of #75689, which fixes the same bug with a smaller, lower-risk change.

Per the Codex review here, dropping the `num_rows == 0` guard in `rewrite_segment` routes empty bundled slices into `SegmentRewriter::rewrite_partial_update`'s copy-only fast path (`fs::copy_file` ignoring `bundle_file_offset`), which would copy the whole physical bundle. Rather than rework the rewrite/publish path, #75689 keeps the wedge from happening by flagging bundled segments as shared when they are orphaned, so `collect_garbage_files` routes them through the alive-checked shared-file deleter and never deletes a bundle file a sibling still references. That fully prevents the "segment does not exist" publish wedge without touching the rewrite path.

PR #75689, opened 2026-07-02
[BugFix] Flag bundled segments as shared when orphaning them for vacuum

luohaha:
## Why I'm doing:

When file bundling is enabled (`enable_file_bundling = true`, the default for shared-data tables since 4.0), one physical `.dat` holds the segments of several sibling tablets of a partition, and a segment's shared-ness is encoded by `bundle_file_offset` — so vacuum's `is_shared_segment()` is `shared() || has_bundle_file_offset()`.

Consider a 3-tablet bundled primary-key table and one load where tablet A upserts a row while tablets B and C only delete rows. B and C get a **0-row bundled segment** (their slice of the shared `.dat`), which `rewrite_segment` leaves as a live rowset segment (it skips empty rowsets). Tablet A rewrites its own slice during publish and **orphans the raw bundle segment**.

`orphan_files` entries are `FileMetaPB`, which carries only `name` / `size` / `shared` / `encryption_meta` — **not** `bundle_file_offset`. The orphan-producing paths (`RowsetUpdateState::rewrite_segment`, `MetaFileBuilder::apply_column_mode_partial_update`, `MetaFileBuilder::apply_opcompaction_with_conflict`) copied only `segment.shared()`, so tablet A's orphan landed with `shared=false`. `collect_garbage_files` then routes it through the **plain deleter** and deletes the physical bundle file — while B and C still reference it in a live rowset. Their next publish fails at `prepare_primary_index -> Rowset::load_segments`:

```
Not found: load_segments failed ... Object s3://.../<txn>_<uuid>.dat does not exist
```

which wedges the whole table's publish queue. `collect_garbage_files` cannot recover the bundle info from an orphan `FileMetaPB` (no offset, no rowset/index to feed `is_shared_segment`), so the `shared` flag is the only signal on the orphan path — it must be set correctly when the orphan is produced.

## What I'm doing:

Set the orphan `FileMetaPB.shared` to `is_shared_segment()`'s definition (`shared() || has_bundle_file_offset()`) in all three orphan-producing paths. `collect_garbage_files`'s existing `file.shared()` check then routes bundled orphans through the alive-checked shared-file deleter, so `collect_alive_shared_files` vetoes deletion while any sibling still references the bundle file. The bundle file is GC'd only once no tablet references it (e.g. after the empty slices are compacted away). Add a regression test.

Reproduced end-to-end on a 3-tablet file-bundled PK table (1 tablet upsert + 2 tablets delete-only, then accelerated vacuum): before the fix the delete-only tablets wedge publish with "segment does not exist"; after the fix publish succeeds.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/meta_file.cpp
+++ b/be/src/storage/lake/meta_file.cpp
@@ -253,9 +253,10 @@ void MetaFileBuilder::apply_column_mode_partial_update(const TxnLogPB_OpWrite& o
     for (const auto& segment_meta : op_write.rowset().segment_metas()) {
         FileMetaPB file_meta;
         file_meta.set_name(segment_meta.filename());
-        if (segment_meta.has_shared()) {
-            file_meta.set_shared(segment_meta.shared());
-        }
+        // Mirror is_shared_segment(): a bundled segment (bundle_file_offset set) is shared with
+        // sibling tablets. The orphan FileMetaPB only carries `shared`, so encode bundling into it
+        // too, otherwise vacuum deletes the shared bundle file while a sibling still references it.
+        file_meta.set_shared(segment_meta.shared() || segment_meta.has_bundle_file_offset());
         _tablet_meta->mutable_orphan_files()->Add(std::move(file_meta));
     }
 }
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg (approved):

kevincai (approved):

PR #75696, opened 2026-07-02
[Enhancement] Hash partition_name in external table statistics to avoid primary key size overflow

dirtysalt:
## Why I'm doing:
Background collection of column statistics for external tables (Iceberg/Hive/Paimon) writes one row per (table, partition, column) into the internal `_statistics_.external_column_statistics` table, whose primary key is `(table_uuid, partition_name, column_name)`.

For tables that use partition evolution (e.g. a hashed/bucketed partition column), the resulting `partition_name` string can be long. Combined with `table_uuid` and `column_name`, the encoded primary key can exceed the BE's `primary_key_limit_size` (128 bytes by default), and the write fails with `primary key size exceed the limit`. Because the failed partition's stats never get persisted, the very next query re-triggers collection for the same partition, and it fails again — the table gets stuck in a permanent collection-failure loop and never accumulates usable statistics.

`partition_name` is only ever used internally as an opaque dedup/lookup key — it's never shown to users and no code path does substring/prefix matching on it. So there's no need to preserve the original value; replacing it with a fixed-length hash removes the size cap concern entirely, independent of how long or deeply nested the underlying partition scheme is.

## What I'm doing:
- Add `StatisticUtils.hashExternalPartitionName()`, which hashes a partition name into a fixed-length, prefixed string (`v2h-<16 hex chars>`). Applied at the single place `external_column_statistics.partition_name` is actually written (`ExternalFullStatisticsCollectJob`, shared by both full and sample collection), so the stored key length is now bounded regardless of the source partition name.
- Update the one read path that filters by `partition_name` (`PartitionSelector.getExternalTablePartitionStats`, used for materialized-view refresh partition selection) to hash its lookup keys and translate the results back to the caller's original partition names.
- The main CBO statistics-loading query never filtered or grouped by `partition_name`, so it's unaffected by the format change — except for one migration concern:

  ```
  Before migration                         After migration
  +---------------------------------+      +---------------------------------+
  | partition_name = "tenant_id=    |      | partition_name =                |
  |   hash-2c933fdc" (raw, long)    |      |   "v2h-a1b2c3d4e5f60789" (hashed)|
  +---------------------------------+      +---------------------------------+
        same logical partition, but a different primary key now -> both rows
        coexist, so a naive SUM(row_count) across all rows for the table would
        double count this partition
  ```

  Since a primary-key table can't overwrite the old raw-format row with the new hash-keyed one (different key), the aggregate query now adds a `partition_name LIKE 'v2h-%'` filter so only hashed rows are counted. This avoids any destructive migration (no TRUNCATE/DELETE): pre-existing raw-format rows simply stop being read and are naturally superseded as tables get re-collected.
- While touching the IN-clause construction in `PartitionSelector`, also fixed a latent bug: multiple partition names were joined into one comma-separated string and wrapped in a single pair of quotes (`IN ('a,b,c')`), which only ever matched a single-partition lookup.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/partition/PartitionSelector.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/transformation/partition/PartitionSelector.java
@@ -79,6 +79,7 @@ import com.starrocks.sql.optimizer.rewrite.ScalarOperatorRewriter;
 import com.starrocks.sql.optimizer.transformer.ExpressionMapping;
 import com.starrocks.sql.optimizer.transformer.SqlToScalarOperatorTranslator;
 import com.starrocks.sql.parser.SqlParser;
+import com.starrocks.statistic.StatisticUtils;
 import com.starrocks.thrift.TResultBatch;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

PR #75697, opened 2026-07-02
[Enhancement] Route VARCHAR sort keys through meta-tier tablet pre-split

xiangguangyxg:
## Why I'm doing:

Sample-Based Tablet Pre-Split's meta tier (the footer-statistics path) could not handle VARCHAR sort keys — string columns always fell back to the more expensive data-tier sample. Extending the meta tier to strings also surfaced a pre-existing correctness bug: the FE string comparator (`StringUtils.compareStringWithUTF8ByteArray`) compared UTF-8 bytes as **signed** Java bytes and carried a trailing-NUL "equal" special case, so its ordering diverged from the BE, which routes range-distributed rows by an **unsigned** `memcmp` (`be/src/base/string/memcmp.h`). For a non-ASCII VARCHAR range-distribution key this could make `RangeDistributionPruner` prune a tablet the BE had actually filled, dropping rows from query results. Meta-tier string boundaries are only safe once FE ordering matches BE ordering.

## What I'm doing:

1. **Fix the FE string comparator to match BE (query-correctness bug fix).** `StringUtils.compareStringWithUTF8ByteArray` now delegates to `Arrays.compareUnsigned` — unsigned UTF-8 byte order, shorter-prefix-is-smaller tie-break, no trailing-NUL special case — which is byte-identical to BE `memcompare`. Its sole caller is `StringVariant.compareToImpl`, so this aligns every consumer of `Variant`/`Tuple` ordering (pre-split boundaries, query-time `RangeDistributionPruner`, range-colocate dispatch) with how the BE routes rows, for both VARCHAR and CHAR keys.

2. **Parquet meta-tier VARCHAR.** `ParquetRowGroupStatisticsReader` no longer marks BINARY+UTF8 (string) chunk statistics truncated. parquet-cpp / parquet-mr do not truncate chunk-level `Statistics` (parquet-mr's `parquet.statistics.truncate.length` defaults to `Integer.MAX_VALUE`; parquet-cpp drops oversized stats entirely, which the existing `isEmpty()` guard already routes to the data tier). Pre-split is a pure planning optimization — the BE routes every row by its true value — so a VARCHAR sort key now flows through the meta tier.

3. **ORC meta-tier VARCHAR.** `OrcStripeStatisticsReader` accepts the ORC STRING/VARCHAR categories (unpadded, unsigned-byte-ordered stripe stats) for a VARCHAR sort key and decodes the stripe min/max. ORC returns `null` from `getMinimum()`/`getMaximum()` exactly when a value exceeded 1024 bytes and only a truncated bound was retained; that stripe is marked truncated and falls back to the data tier.

**CHAR is intentionally deferred to the data tier** (both readers): the BE pads/truncates a `CHAR(N)` value to its fixed width before it routes the row (`pad_char_values`, `be/src/exec/data_sinks/tablet_sink.cpp`), so raw footer/stripe stats would not match the routed key and could produce a skewed split. The data tier samples the post-cast rows and is unaffected. This also matches the existing deferral of the ORC CHAR source category.

Tests: new `StringUtilsTest`; `VariantTest` gains a non-ASCII ordering case and updates the trailing-NUL case to BE parity; `RangeDistributionPrunerTest` gains a non-ASCII pruning regression; the Parquet/ORC reader tests assert meta-tier VARCHAR reads, ORC's >1024-byte truncation fallback, and CHAR-sort-key data-tier fallback.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/OrcStripeStatisticsReader.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/OrcStripeStatisticsReader.java
@@ -30,6 +30,7 @@ import org.apache.orc.DecimalColumnStatistics;
 import org.apache.orc.IntegerColumnStatistics;
 import org.apache.orc.OrcFile;
 import org.apache.orc.Reader;
+import org.apache.orc.StringColumnStatistics;
 import org.apache.orc.StripeInformation;
 import org.apache.orc.StripeStatistics;
 import org.apache.orc.TimestampColumnStatistics;
@@ -65,10 +66,15 @@ import java.util.Objects;
  *       a pre-1970 sub-second — and packs the calendar date through the same proleptic conversion as
  *       the boundary parse). TIMESTAMP_INSTANT is deferred (the load applies a session-tz offset this
  *       reader cannot reproduce).</li>
+ *   <li>ORC STRING/VARCHAR -> StarRocks VARCHAR, using the stripe min/max when ORC
+ *       kept them exact (getMinimum()/getMaximum() non-null); a truncated bound (value
+ *       > 1024 bytes -> getMinimum()/getMaximum() null) marks the stripe truncated -> data
+ *       tier. StarRocks CHAR (the BE pads/truncates it to a fixed width before routing) and the
+ *       ORC CHAR source category (space-padded) are both deferred -> data tier.</li>
  * </ul>
- * Everything else — STRING/CHAR/VARCHAR (data-tier fallback anyway), BOOLEAN,
- * FLOAT/DOUBLE, TIMESTAMP_INSTANT (load applies a session-tz offset this reader cannot reproduce),
- * non-matching-precision/scale DECIMAL, and any complex type — makes the reader throw
+ * Everything else -- BOOLEAN, FLOAT/DOUBLE, TIMESTAMP_INSTANT (load applies a session-tz offset this
+ * reader cannot reproduce), non-matching-precision/scale DECIMAL, ORC CHAR, and any complex type --
+ * makes the reader throw
  * {@link MetaTierUnavailableException} so the pipeline falls back to data tier. That is
  * NOT a load failure. Pure I/O failures surface as {@link StarRocksException}.
  * Note: a legacy ORC file lacking modern UTC stats (pre-1.5-era footer) decodes through the JVM
@@ -153,10 +159,10 @@ public final class OrcStripeStatisticsReader {
      * Reject ORC/StarRocks type pairings outside meta tier's supported window
      * eagerly, before iterating stripes. Anything outside the window means "fall
      * back to data tier", not "fail the load". The supported window is the unannotated
-     * integer categories, ORC DATE → StarRocks DATE, ORC DECIMAL → a same-precision/scale
-     * StarRocks DECIMAL, and ORC TIMESTAMP → StarRocks DATETIME. String, boolean,
-     * floating-point, TIMESTAMP_INSTANT, non-matching DECIMAL, and complex types are deferred
-     * (fall back to data tier).
+     * integer categories, ORC DATE -> StarRocks DATE, ORC DECIMAL -> a same-precision/scale
+     * StarRocks DECIMAL, ORC TIMESTAMP -> StarRocks DATETIME, and ORC STRING/VARCHAR ->
+     * StarRocks VARCHAR. Boolean, floating-point, TIMESTAMP_INSTANT, non-matching DECIMAL,
+     * a CHAR target, and complex types are deferred (fall back to data tier).
      */
     private static void rejectIncompatibleTypeMapping(
             TypeDescription fieldType, Column sortKeyColumn) throws MetaTierUnavailableException {
@@ -172,6 +178,12 @@ public final class OrcStripeStatisticsReader {
             // Plain ORC TIMESTAMP is local (no timezone); the BE load stores its UTC wall clock
             // verbatim (no session-tz offset, unlike TIMESTAMP_INSTANT) → StarRocks DATETIME.
             case TIMESTAMP -> starRocksPrimitive == PrimitiveType.DATETIME;
+            // ORC STRING/VARCHAR categories are unpadded; their stripe min/max are
+            // unsigned-byte-ordered (Text/WritableComparator), matching BE VARCHAR routing.
+            // Target CHAR is excluded (the BE pads/truncates CHAR to its fixed width before
+            // routing, so raw stripe stats would not match the routed key), as is the ORC CHAR
+            // source category (space-padded) -> data tier.
+            case STRING, VARCHAR -> starRocksPrimitive == PrimitiveType.VARCHAR;
             default -> false;
         };
         if (!compatible) {
@@ -218,6 +230,10 @@ public final class OrcStripeStatisticsReader {
         if (sortKeyStatistics instanceof TimestampColumnStatistics timestampStatistics
                 && timestampStatistics.getNumberOfValues() > 0) {
             return convertTimestampStripe(timestampStatistics, rowCount, location);
+        }
+        if (sortKeyStatistics instanceof StringColumnStatistics stringStatistics
+                && stringStatistics.getNumberOfValues() > 0) {
+            return convertStringStripe(stringStatistics, rowCount, location);
         }
         // Absent / all-null stats (no presence flag on ORC numeric/date stats, so an
         // empty stripe is detected via getNumberOfValues() == 0) → missing min/max.
@@ -295,6 +311,32 @@ public final class OrcStripeStatisticsReader {
         } catch (RuntimeException conversionFailure) {
             throw new MetaTierUnavailableException(String.format(
                     "ORC decimal stats value not representable for sort-key column \"%s\": %s",
+                    location.starRocksColumn.getName(), conversionFailure.getMessage()));
+        }
+        return new RowGroupStatistics(
+                new Tuple(List.of(minVariant)), new Tuple(List.of(maxVariant)), rowCount, /*truncated=*/ false);
+    }
+
+    private static RowGroupStatistics convertStringStripe(
+            StringColumnStatistics stringStatistics, long rowCount, SortKeyLocation location)
+            throws MetaTierUnavailableException {
+        // ORC returns null minimum/maximum exactly when the value was truncated (only a
+        // lower/upper BOUND is retained, for strings longer than 1024 bytes). A truncated
+        // bound is a widened endpoint, unsafe for boundary placement -- mark truncated so
+        // the pipeline falls back to data tier.
+        String min = stringStatistics.getMinimum();
+        String max = stringStatistics.getMaximum();
+        if (min == null || max == null) {
+            return new RowGroupStatistics(null, null, rowCount, /*truncated=*/ true);
+        }
+        Variant minVariant;
+        Variant maxVariant;
+        try {
+            minVariant = Variant.of(location.starRocksColumn.getType(), min);
+            maxVariant = Variant.of(location.starRocksColumn.getType(), max);
+        } catch (RuntimeException conversionFailure) {
+            throw new MetaTierUnavailableException(String.format(
+                    "ORC string stats value not representable for sort-key column \"%s\": %s",
                     location.starRocksColumn.getName(), conversionFailure.getMessage()));
         }
         return new RowGroupStatistics(
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/ParquetRowGroupStatisticsReader.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/ParquetRowGroupStatisticsReader.java
@@ -72,8 +72,11 @@ import java.util.Objects;
  *   <li>Parquet INT64 with TIMESTAMP annotation, {@code isAdjustedToUTC=false}
  *       (MILLIS/MICROS/NANOS) → StarRocks DATETIME. UTC-adjusted timestamps are
  *       deferred (the load applies a session-tz offset this reader cannot reproduce).</li>
- *   <li>Parquet BINARY with UTF8 (string) annotation → StarRocks CHAR/VARCHAR
- *       (always marked truncated → forces data-tier fallback for string sort keys)</li>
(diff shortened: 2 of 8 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/ParquetRowGroupStatisticsReader.java
```diff
@@ -284,23 +291,21 @@ private static RowGroupStatistics convertBlock(BlockMetaData block, SortKeyLocat
...
+        // even under a hypothetical custom-truncating writer, a widened min/max can only
+        // skew split boundaries or trigger a conservative overlap fallback — it can never
+        // misroute or drop data. All numeric/temporal stats are exact too, so this reader
+        // never marks a Parquet row group truncated.
         return new RowGroupStatistics(
                 new Tuple(List.of(minVariant)),
                 new Tuple(List.of(maxVariant)),
                 rowCount,
-                truncated);
+                /*truncated=*/ false);
```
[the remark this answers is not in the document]
xiangguangyxg: Thanks — I looked at this closely and I'm keeping `truncated=false` for string chunk stats, for three reasons:

1. **Chunk-level stats are not truncated by the dominant writers.** parquet-mr's chunk `Statistics` truncation is `parquet.statistics.truncate.length`, which defaults to `Integer.MAX_VALUE` (no truncation). The 64-byte `parquet.columnindex.truncate.length` only truncates the **ColumnIndex**, which this reader never reads. parquet-cpp and the StarRocks writer don't truncate chunk-level `Statistics` either (it drops them past `max_statistics_size`, which the `isEmpty()/!hasNonNullValue()` guard just above already routes to the data tier). The TSP e2e here used StarRocks-unloaded Parquet and got exact bounds with the cut landing on the true row-group min.

2. **The suggested gate has no signal to key off.** "Prove both bounds are exact" would require the Thrift `is_min_value_exact`/`is_max_value_exact` fields, but both dominant writers leave them **absent** (not `false`) — empirically and by parquet-mr bytecode. Gating meta-tier on them would reject ~100% of real files, i.e. the string meta tier would never engage. That flag-gate was evaluated during design and dropped for exactly this reason.

3. **Correctness is preserved even under a deliberately-truncating writer.** parquet-mr truncation only ever *widens* a bound (min → a ≤ prefix, max → an incremented ≥ prefix), so `[reported_min, reported_max] ⊇ [true_min, true_max]`. Pre-split only *chooses split boundaries*; the BE routes every loaded row to a tablet by its true value (unsigned `memcmp`), independent of these stats. A widened boundary can therefore only make a split uneven, or produce a boundary the BE rejects → `external_boundaries_fallback_total` increments and the split falls back — never a misrouted or dropped row. The comment on this line documents that tradeoff, and it was reviewed in the plan phase.

So the worst case under the non-default writer you describe is a *less-even* split, which is a planning-quality concern, not a correctness one — and the data-tier fallback still fires whenever the footer omits the bounds. Happy to revisit if there's a writer that both truncates chunk stats *and* reliably sets the exactness flags.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/OrcStripeStatisticsReader.java
```diff
@@ -172,6 +177,10 @@ private static void rejectIncompatibleTypeMapping(
             // Plain ORC TIMESTAMP is local (no timezone); the BE load stores its UTC wall clock
             // verbatim (no session-tz offset, unlike TIMESTAMP_INSTANT) → StarRocks DATETIME.
             case TIMESTAMP -> starRocksPrimitive == PrimitiveType.DATETIME;
+            // ORC STRING/VARCHAR categories are unpadded; their stripe min/max are
+            // unsigned-byte-ordered (Text/WritableComparator), matching BE VARCHAR/CHAR.
+            // ORC CHAR (space-padded) is intentionally excluded → data tier.
+            case STRING, VARCHAR -> starRocksPrimitive.isCharFamily();
```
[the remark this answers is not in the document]
xiangguangyxg: Resolved: I've kept `CHAR` on the data tier. Both readers now accept only a `VARCHAR` target for the string meta tier (`OrcStripeStatisticsReader`: `case STRING, VARCHAR -> starRocksPrimitive == VARCHAR`; `ParquetRowGroupStatisticsReader`: string gate `== VARCHAR`), so a `CHAR(N)` sort key falls back to data-tier sampling of the post-cast rows — matching the BE trailing-space removal + fixed-width pad/truncate you cited. Added `charSortKeyFallsBackToDataTier` regression tests to both reader test classes.

xiangguangyxg:
### TSP shared-data e2e validation — PASS

Built this PR and deployed a shared-data cluster (1 FE + 3 CN). Validated that a **VARCHAR range sort key now flows through the meta tier** for both Parquet and ORC, end to end, with data integrity and BE boundary parity.

**Setup**
- Range-distributed target: `CREATE TABLE t (k VARCHAR(32), id BIGINT, v INT) DUPLICATE KEY(k) ORDER BY(k)` (no `DISTRIBUTED BY` → `DISTRIBUTE_TYPE=RANGE`, 1 tablet), with `enable_range_distribution=true`.
- Fixtures: 3 ordered, non-overlapping VARCHAR bands (`lpad(id,10,'0')`, ids `[0,10^6)`, `[2·10^6,3·10^6)`, `[4·10^6,5·10^6)`), each unloaded as one row group via `INSERT INTO FILES(...)` with `pipeline_dop=1` — once as Parquet, once as ORC.
- Load: `INSERT INTO t SELECT * FROM FILES(...)` (single-level glob), snapshotting FE `:8030` and CN `:8040` pre-split metrics before/after.

**Results (3,000,000 rows per table)**

| Signal | Parquet | ORC |
|---|---|---|
| `tablet_pre_split_tier_used{tier="meta_tier"}` | Δ +1 | Δ +1 |
| `tablet_pre_split_tier_used{tier="data_tier"}` | 0 | 0 |
| CN `tablet_reshard_split_external_boundaries_fallback_total` | 0 on all 3 CNs | 0 on all 3 CNs |
| `tablet_reshard_split_total` / `split_output_tablet_count` / `split_failed` | 2 / 4 / 0 (cumulative across both loads) | — |
| Row count | 3,000,000 | 3,000,000 |
| MIN(k) / MAX(k) | `0000000000` / `0004999999` | `0000000000` / `0004999999` |

**Boundary correctness (Parquet table shown; ORC identical):** the cut point landed **exactly on the decoded footer min of band 2 (`0002000000`)**:

```
TabletId  DataSize  RowCount  Range
10288     8.1MB     1000000   (-∞, [0002000000])      -- band 1
10289     16.2MB    2000000   [[0002000000], +∞)      -- bands 2+3
```

This demonstrates the full path: the reader decodes the string footer/stripe min → the meta tier plans it as a boundary → the BE accepts it (`fallback_total = 0`) → rows route by unsigned-byte order (band 1 below the cut, bands 2+3 above), with the total preserved (1M + 2M = 3M). Repeated at both 0.9M and 3M rows; both hit the meta tier with zero fallback.

**Covered by unit tests instead of e2e (both correctly route to the data tier):**
- ORC string stats truncated beyond 1024 bytes (`getMinimum()/getMaximum()` null) → `truncatedStringStatsFallBackToDataTier`.
- `CHAR(N)` sort key (BE pads/truncates before routing) → `charSortKeyFallsBackToDataTier` in both reader tests.

Cluster released; temporary fixtures cleaned up.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/OrcStripeStatisticsReader.java
```diff
@@ -65,10 +66,15 @@
...
+ *       kept them exact (getMinimum()/getMaximum() non-null); a truncated bound (value
+ *       > 1024 bytes → getMinimum()/getMaximum() null) marks the stripe truncated → data
+ *       tier. StarRocks CHAR (the BE pads/truncates it to a fixed width before routing) and the
+ *       ORC CHAR source category (space-padded) are both deferred → data tier.</li>
  * </ul>
- * Everything else — STRING/CHAR/VARCHAR (data-tier fallback anyway), BOOLEAN,
- * FLOAT/DOUBLE, TIMESTAMP_INSTANT (load applies a session-tz offset this reader cannot reproduce),
- * non-matching-precision/scale DECIMAL, and any complex type — makes the reader throw
+ * Everything else — BOOLEAN, FLOAT/DOUBLE, TIMESTAMP_INSTANT (load applies a session-tz offset this
+ * reader cannot reproduce), non-matching-precision/scale DECIMAL, ORC CHAR, and any complex type —
```
kevincai: Nit: the comments/javadoc added in this PR use non-ASCII typography -- right arrow (U+2192) and em dash (U+2014) -- here and in the other added comment blocks (`rejectIncompatibleTypeMapping`, `convertStringStripe`, `ParquetRowGroupStatisticsReader#convertBlock`, and the new/updated tests). Please use the ASCII equivalents `->` and `--` in the added lines to keep the source ASCII-clean and grep/terminal-friendly.

Generated with [Claude Code](https://claude.ai/code)

fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/OrcStripeStatisticsReader.java
```diff
@@ -172,6 +178,12 @@ private static void rejectIncompatibleTypeMapping(
             // Plain ORC TIMESTAMP is local (no timezone); the BE load stores its UTC wall clock
             // verbatim (no session-tz offset, unlike TIMESTAMP_INSTANT) → StarRocks DATETIME.
             case TIMESTAMP -> starRocksPrimitive == PrimitiveType.DATETIME;
+            // ORC STRING/VARCHAR categories are unpadded; their stripe min/max are
+            // unsigned-byte-ordered (Text/WritableComparator), matching BE VARCHAR routing.
+            // Target CHAR is excluded (the BE pads/truncates CHAR to its fixed width before
+            // routing, so raw stripe stats would not match the routed key), as is the ORC CHAR
+            // source category (space-padded) → data tier.
+            case STRING, VARCHAR -> starRocksPrimitive == PrimitiveType.VARCHAR;
```
kevincai, line 186: This method's javadoc (lines 163-165 above) still lists String among the deferred types: "String, boolean, floating-point, TIMESTAMP_INSTANT, non-matching DECIMAL, and complex types are deferred (fall back to data tier)." With this new case, ORC STRING/VARCHAR -> StarRocks VARCHAR is now inside the supported window (the class-level javadoc was already updated to say so). Please update the method javadoc to match.

Generated with [Claude Code](https://claude.ai/code)

fe/fe-core/src/main/java/com/starrocks/common/util/StringUtils.java
```diff
@@ -83,35 +84,16 @@ public static boolean areColumnNamesEqual(String columName, String toCheck) {
...
-            if (s2Bytes[i] == 0x00) {
-                return 0;
-            } else {
-                return -1;
-            }
-        } else {
-            return 0;
-        }
+        return Arrays.compareUnsigned(
+                s1.getBytes(StandardCharsets.UTF_8), s2.getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
xiangguangyxg: Thanks — I traced this end to end and it isn't reachable, so I'm keeping the comparator as-is. The scenario requires a *persisted* CHAR range boundary that carries NUL padding while the query literal doesn't, but on both sides the stored value is unpadded:

- **BE strips the trailing NUL for CHAR before a boundary is ever emitted.** `be/src/column/datum_convert.cpp:95-97`:
  ```cpp
  // If type is TYPE_CHAR, strip its tailing '\0'
  if (type == TYPE_CHAR) { slice.size = strnlen(slice.data, slice.size); }
  ```
  A BE-reshard boundary round-trips through `DatumVariant::to_proto` → `datum_to_string` (CHAR via the VARCHAR stringifier), so the persisted `TabletRange` value is `"abc"`, not `"abc\0\0"`.
- **FE side is unpadded too.** The query literal is `Variant.of(column.getType(), expr.getStringValue())` → `new StringVariant(type, value)` with no width/padding, and FE pre-split defers a CHAR target to the data tier (this PR), whose sample decodes the raw source string. GSON persists `StringVariant.value` verbatim.

The NUL padding from `pad_char_values` is transient at BE load-time *routing* only; it is never the value FE stores or compares. So the boundary and the literal are symmetric (both unpadded), and the removed trailing-NUL equality case could not have fired for CHAR-key pruning — dropping it is a no-op there. (The pre-existing FE-unpadded vs BE-padded-routing mismatch for CHAR is a separate matter, and is exactly why meta-tier defers CHAR; the old special case didn't address it either.)

fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/OrcStripeStatisticsReader.java
xiangguangyxg: Fixed. The `rejectIncompatibleTypeMapping` javadoc now includes `ORC STRING/VARCHAR -> StarRocks VARCHAR` in the supported window and drops String from the deferred list (deferred now reads "Boolean, floating-point, TIMESTAMP_INSTANT, non-matching DECIMAL, a CHAR target, and complex types"). Thanks for catching the stale method doc.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/OrcStripeStatisticsReader.java
xiangguangyxg: Fixed. Replaced the right-arrow (U+2192) and em dash (U+2014) with ASCII `->` and `--` in all lines this PR adds — across `rejectIncompatibleTypeMapping`, `convertStringStripe`, the class javadocs, `ParquetRowGroupStatisticsReader#convertBlock`, and the new/updated tests. I scoped the change to the lines this PR touches and left the pre-existing comment blocks unchanged.

kevincai (approved):

srlch (approved):

PR #75702, opened 2026-07-02
[Enhancement] Name shared-data vector index files by the segment's owning tablet id

sevev:
## Why I'm doing:

Follow-up to #75542, which added vector-index (`.vi`) support for shared-data bundle-file
segments. That PR names each `.vi` `<segment_base>_<tablet_id>_<index_id>.vi`, embedding the
tablet id to disambiguate segments that share one physical bundle file across tablets. It uses
the id of the tablet **touching** the segment, which is not reader-agnostic.

After a tablet split, a single segment is shared by multiple child tablets
(`SegmentMetadataPB.shared`, carried verbatim via `CopyFrom`). Each child reads the shared
segment with its **own** tablet id, so it derives a different `.vi` name than the one the writer
actually produced:

- **Read**: the child looks up `..._<child_id>_<index_id>.vi`, which does not exist → NotFound
  → silent fall back to brute-force vector search (results stay correct, but slow).
- **GC / vacuum**: orphan protection and deletion also reconstruct the name from the local
  tablet id, so the real `.vi` (named after the writer) is neither protected nor collected under
  the name every referencing tablet computes — it can be deleted while still referenced, or
  leaked.

## What I'm doing:

Record the **writing** tablet's id in a new persisted field
`SegmentMetadataPB.vector_index_tablet_id` and derive every `.vi` name from that recorded owner
instead of the local tablet id. Because the field is carried verbatim across split via
`CopyFrom`, every child that shares the segment resolves the exact same `.vi` the writer wrote.

- The `.vi` filename **format is unchanged** (`<segment_base>_<owner_tablet_id>_<index_id>.vi`);
  only the source of the embedded id changes from "the tablet touching the segment" to "the
  recorded owner".
- The owner is stamped at **write time**, in the same step that records `vector_index_ids`: the
  tablet writers (normal write and the deferred-async initial write) and the segment rewrite
  (partial-update / auto-increment). `to_proto` persists it and the publish apply carries it
  verbatim. From there it is threaded through `FileInfo` / `SegmentReadOptions` and consumed by
  the read path (`segment_iterator`), the async build (`vector_index_build_task`), and GC
  (`vacuum` / `transactions`).
- **Invariant, no fallback**: a segment that records `vector_index_ids` always records this owner
  too, and every consumer reads the owner only for a vector-indexed segment and `DCHECK`s its
  presence — there is no fall back to the reading tablet id. A future write path that records ids
  but forgets the owner fails loudly in tests instead of silently degrading to brute-force reads
  and mis-named GC. This PR is a follow-up to #75542 (unreleased) and ships together with it, so
  no released version ever persisted a vector-indexed segment without the owner.
- A **split-shared** `.vi` is named by the recorded owner, so it is one physical file referenced
  by every sibling tablet. It follows its segment's shared-file lifecycle in GC: routed through
  the shared-file deleter / retained while any sibling still references it, and delay-deleted like
  the shared segment. A **bundle-only** `.vi` stays per-tablet (the filename embeds the writer id)
  and keeps ordinary per-tablet deletion.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/fs/fs.h
+++ b/be/src/fs/fs.h
@@ -81,6 +81,11 @@ struct FileInfo {
     // rewrite), so the apply path can persist them into SegmentMetadataPB::vector_index_ids and
     // keep async builds / vacuum / reads in sync. Empty for files with no vector index.
     std::vector<int64_t> vector_index_ids;
+    // Tablet id under which this segment's .vi files were written; embedded in the .vi filename so
+    // it stays stable and reader-agnostic across tablet split (see the field comment on
+    // SegmentMetadataPB::vector_index_tablet_id). -1 when no vector index; otherwise set together
+    // with vector_index_ids and persisted into SegmentMetadataPB::vector_index_tablet_id.
+    int64_t vector_index_tablet_id = -1;
 
     // Cache key uniquely identifying this FileInfo as a *slice* of a physical file. Caches keyed
     // on file identity (e.g. lake metacache for Segments) must use this rather than `path` so two
--- a/be/src/fs/fs.h
+++ b/be/src/fs/fs.h
@@ -81,11 +81,6 @@ struct FileInfo {
     // rewrite), so the apply path can persist them into SegmentMetadataPB::vector_index_ids and
     // keep async builds / vacuum / reads in sync. Empty for files with no vector index.
     std::vector<int64_t> vector_index_ids;
-    // Tablet id under which this segment's .vi files were written; embedded in the .vi filename so
-    // it stays stable and reader-agnostic across tablet split (see the field comment on
-    // SegmentMetadataPB::vector_index_tablet_id). -1 when no vector index; otherwise set together
-    // with vector_index_ids and persisted into SegmentMetadataPB::vector_index_tablet_id.
-    int64_t vector_index_tablet_id = -1;
 
     // Cache key uniquely identifying this FileInfo as a *slice* of a physical file. Caches keyed
     // on file identity (e.g. lake metacache for Segments) must use this rather than `path` so two
--- a/be/src/fs/fs.h
+++ b/be/src/fs/fs.h
@@ -76,16 +76,6 @@ struct FileInfo {
     std::shared_ptr<FileSystem> fs;
     // It is used to store the file offset of the bundle file.
     std::optional<int64_t> bundle_file_offset;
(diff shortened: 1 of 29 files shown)


CelerData-Reviewer:
@codex review

sevev:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/fs/fs.h
```diff
@@ -81,6 +81,11 @@ struct FileInfo {
     // rewrite), so the apply path can persist them into SegmentMetadataPB::vector_index_ids and
     // keep async builds / vacuum / reads in sync. Empty for files with no vector index.
     std::vector<int64_t> vector_index_ids;
+    // Tablet id under which this segment's .vi files were written; embedded in the .vi filename so
+    // it stays stable and reader-agnostic across tablet split (see the field comment on
+    // SegmentMetadataPB::vector_index_tablet_id). -1 when no vector index; otherwise set together
+    // with vector_index_ids and persisted into SegmentMetadataPB::vector_index_tablet_id.
+    int64_t vector_index_tablet_id = -1;
```
xiangguangyxg: could vector_index_ids and  vector_index_tablet_id be moved to SegmentFileInfo, because they are belong to segment

CelerData-Reviewer:
@codex review

be/src/fs/fs.h
sevev: > could vector_index_ids and vector_index_tablet_id be moved to SegmentFileInfo, because they are belong to segment

updated.

xiangguangyxg (dismissed):

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

luohaha (approved):

xiangguangyxg (approved):

PR #75719, opened 2026-07-02
[Enhancement] Elide the MERGE duplicate check when the source is provably unique

GavinMar:
## Why I'm doing:

The base MERGE INTO pipeline always inserts a runtime `EnforceUniqueRowLocatorNode` to reject a target row matched by more than one source row. When the source is already unique on a key set that the ON clause fully covers, at most one source row can match each target row, so that runtime check is provably redundant — a wasted per-row dedup-tracking node on the path that feeds the row-delta sink.

## What I'm doing:

Elide the duplicate check when source uniqueness over the ON keys guarantees an at-most-one match (`canElideEnforceUnique`). Uniqueness is credited only conservatively, so a wrong elision can never drop a needed check:

- The source must be **DISTINCT**, or **GROUP BY** whose keys map to output columns by `SlotRef` column name; a non-`SlotRef` key (e.g. `GROUP BY f(x)`) is not credited (no fragile `toSql()` text matching).
- The two sides of an ON equi-key must be **type-equivalent**. The analyzed AST keeps bare `SlotRef`s across an implicit cast, so a non-injective coercion (e.g. `VARCHAR = INT`) is rejected — distinct source values (`'1'`, `'01'`) could otherwise match one target row.
- When the source and target are referenced by the **same identifier** (e.g. a USING subquery aliased to the target's bare name), no key is credited, so a target-only ON predicate (`target.k = target.k`) cannot be miscounted as covering a source key.

The backend-executability check on the join shape still runs on the elide path, so `join_implementation_mode=merge` is rejected with a clear error even when the duplicate check is elided.

Tested with targeted cases in `MergeIntoPlanTest` (elide / keep, type mismatch, non-SlotRef GROUP BY, grouping sets, name collision, and merge-join rejection on the elide path).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/MergeIntoPlanner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/MergeIntoPlanner.java
@@ -36,12 +36,18 @@ import com.starrocks.planner.SlotDescriptor;
 import com.starrocks.planner.SlotId;
 import com.starrocks.planner.TupleDescriptor;
 import com.starrocks.qe.ConnectContext;
+import com.starrocks.sql.analyzer.AnalyzerUtils;
 import com.starrocks.sql.analyzer.SemanticException;
 import com.starrocks.sql.ast.MergeIntoStmt;
 import com.starrocks.sql.ast.QueryRelation;
+import com.starrocks.sql.ast.Relation;
 import com.starrocks.sql.ast.SelectList;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

PR #75726, opened 2026-07-02
[BugFix] Don't delete pass-through-reused PK-index sstables when dropping a compaction during tablet reshard

xiangguangyxg:
## Why I'm doing:

In shared-data mode, when a tablet that has a pending (committed-but-unpublished) persistent-index compaction is split or merged, the reshard cross-publish path (`convert_txn_log_for_splitting` / `convert_txn_log_for_merging`) drops that compaction and asynchronously deletes its output files, on the assumption that compaction outputs are freshly-written and unreferenced.

That assumption is broken by the persistent-index parallel compaction "full contain / only do move" optimization, which re-emits an input sstable as its output verbatim — the same physical file, with only its `fileset_id` re-stamped. So the dropped compaction's `output_sstables` can list still-live, shared base-fileset filenames. Deleting them removes PK-index sstables that are still referenced by the whole split family, permanently stalling publishes with `prepare_primary_index: load primary index failed: ... .sst does not exist` and leaving load transactions stuck in `COMMITTED`.

## What I'm doing:

When collecting a dropped compaction's output files for deletion, skip any output sstable whose filename also appears in the op's `input_sstables` (the pass-through reuse). This mirrors the opposite-direction guard already in `MetaFileBuilder::remove_compacted_sst`, and covers both `op_compaction` and `op_parallel_compaction`, so both split and merge cross-publish are fixed. Filename equality is the correct key because the physical path is derived from the filename alone; `fileset_id` cannot discriminate since the move re-stamps it.

Additional changes:
- Rename `collect_compaction_output_file_paths` → `collect_compaction_output_files` to match `append_compaction_output_files`.
- Log (at INFO) which files a reshard drop deletes; previously this deletion was only visible at VLOG level.
- Add a regression test covering the pass-through skip for both `op_compaction` and `op_parallel_compaction`, singular and plural output fields.

Fixes #64986

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/tablet_reshard.cpp
+++ b/be/src/storage/lake/tablet_reshard.cpp
@@ -21,6 +21,7 @@
 
 #include "base/utility/defer_op.h"
 #include "common/logging.h"
+#include "gutil/strings/join.h"
 #include "storage/lake/metacache.h"
 #include "storage/lake/tablet_manager.h"
 #include "storage/lake/tablet_merger.h"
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
@Mergify backport branch-4.1

luohaha (approved):

srlch (approved):

PR #75728, opened 2026-07-02
[Enhancement] Support byte-based edit log roll trigger (edit_log_roll_bytes)

### LINKED ISSUE #75718: [Enhancement] BDBJE journal bloat and slow FE restart: every async MV refresh persists the full AsyncRefreshContext to the edit log (O(#base partitions) per entry)
## Enhancement

Every successful async materialized view refresh task run writes one `OP_CHANGE_MATERIALIZED_VIEW_REFRESH_SCHEME` edit log entry that contains a full copy of the MV's `AsyncRefreshContext` - all accumulated per-partition version maps covering every live base-table partition, not just the partitions refreshed by that task run. With many base partitions and periodic refresh, the BDBJE journal bloats and FE restart slows down until, on Kubernetes, the startup probe can kill the FE mid-replay in a self-reinforcing crash loop.

### Current behavior

On each successful task run, [`MVVersionManager.java#L126-L130`](https://github.com/StarRocks/starrocks/blob/0fbae22cfff/fe/fe-core/src/main/java/com/starrocks/scheduler/mv/MVVersionManager.java#L126-L130) builds a `ChangeMaterializedViewRefreshSchemeLog`, whose constructor copies the full context ([`ChangeMaterializedViewRefreshSchemeLog.java#L54`](https://github.com/StarRocks/starrocks/blob/0fbae22cfff/fe/fe-core/src/main/java/com/starrocks/persist/ChangeMaterializedViewRefreshSchemeLog.java#L54): `refreshScheme.getAsyncRefreshContext().copy()`, five maps deep-copied in [`MaterializedView.java#L501-L515`](https://github.com/StarRocks/starrocks/blob/0fbae22cfff/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java#L501-L515)), and the whole object is serialized as one GSON JSON journal entry.

Entry size is therefore O(P), where P is the number of live base table partitions, at roughly 165-170 bytes per partition (measured with a GSON replica of this serialization):

| live base partitions P | size of ONE refresh-scheme entry |
|---|---|
| 365 | ~61 KB |
| 1,000 | ~167 KB |
| 10,000 | ~1.7 MB |

With the default `partition_refresh_number = 1`, one refresh cycle over N stale partitions writes N such entries (O(N x P) bytes per cycle), there is no config to cap or expire the maps, and the journal roll / checkpoint trigger is count-only (`edit_log_roll_num = 50000`, no byte-based trigger) - so with MB-sized entries the backlog between two checkpoint images can reach tens of GB. On a live cluster, even a single idle MV on a 24-row unpartitioned Iceberg base produced ~3.8 journal entries / ~3.3 KiB per refresh cycle with zero data change.

At restart the FE single-thread replays everything after the last image before `isReady` and opens no ports until then; `CheckpointController` only starts after ready, so if the startup probe (default 300s in the operator) kills the FE mid-replay, no image progress is made and the next boot replays the same journal from scratch. The FE's own 60s warning already names this failure mode: "There are too many BDB logs to replay, because of previous failure of checkpoint". All code references are from current main and were re-verified there.

### Expected behavior

1. As a backstop, a byte-based journal roll trigger (`edit_log_roll_bytes`, default off) so that oversized entries cannot grow the inter-image backlog unbounded regardless of entry count - submitted as #75728 with benchmark numbers.
2. The refresh-scheme log carries only the delta (partitions updated/dropped by this task run) and merges into the existing map on replay, keeping the full map materialized only in checkpoint images. This is the model Apache Doris uses for MTMV refresh (`OP_ALTER_MTMV` carries only the refreshed partitions' snapshots; `MTMVRefreshSnapshot.updateSnapshots()` merges on replay). Since the ongoing TvR work also touches `AsyncRefreshContext`, I would like maintainer feedback on the direction before implementing this part.

ronjay1827:
## Why I'm doing:

Journal rolling (and therefore checkpointing and old-journal cleanup) is triggered only by entry count (`edit_log_roll_num`, default 50000). Workloads whose individual edit log entries are large can accumulate a very large replay backlog between two checkpoints while staying under the count threshold. One such workload is async materialized view refresh, which persists the MV's full `AsyncRefreshContext` on every refresh, producing entries proportional to the number of base table partitions (~167 KB at 1,000 partitions per entry). A bloated backlog directly extends FE restart time (single-threaded replay before any port opens) and can end in a startup-probe crash loop on Kubernetes. See #75718 for analysis and measurements.

## What I'm doing:

Add an opt-in, size-based journal roll trigger complementing the count-based one:

- `Config.java`: new mutable config `edit_log_roll_bytes` (long, default `0` = disabled).
- `JournalWriter.java`: accumulate the estimated size of committed journal tasks in `rollJournalBytes`; roll when the accumulated size reaches `edit_log_roll_bytes`. The count-based trigger and the `ALTER SYSTEM CREATE IMAGE` force-roll behavior are unchanged; both counters reset on roll. The roll reason is logged (`rollEditBytes N >= edit_log_roll_bytes M`).
- `JournalWriterRollTest.java`: size trigger fires and resets / size accumulates across batches / disabled trigger changes nothing / count trigger regression.
- Docs: `edit_log_roll_bytes` entry in `FE_parameters/log_server_meta.md` (en/zh/ja).

Upgrading clusters see no behavior change: the trigger is disabled by default.

## Benchmark:

Single-FE benchmark (fresh meta dir, NVMe): 16,000 x `ALTER VIEW` with a 64 KB definition, mirroring the repeated-large-entry pattern of #75718 (the live catalog stays one view, so checkpoint images stay small). The FE is killed with `kill -9` after the load and restarted:

|  | baseline (`edit_log_roll_bytes=0`) | `edit_log_roll_bytes=33554432` (32 MB) |
|---|---|---|
| journal entries pending replay at restart | 16,079 | 212 |
| journal rolls / checkpoint images during load | 0 / 0 | 20 / ~161 KB each |
| restart: image load | - (no image) | 119 ms |
| restart: journal replay | 15,237 ms | **310 ms (49x)** |
| restart wall clock (process start -> queryable) | 18.8 s | 4.6 s |
| write throughput during load | 96 stmts/s | 97 stmts/s (no regression) |

Replay work is bounded by the threshold instead of growing with total written volume; with multi-GB backlogs (the crash-loop scenario in #75718) the gap grows linearly.

## Out of scope / notes:

- Roll is evaluated at batch boundaries (same as the count trigger), so a single oversized batch can exceed the threshold before rolling.
- `rollJournalBytes` uses `JournalTask.estimatedSizeByte()` (journal id + payload buffer); BDBJE storage overhead is not counted.
- Neither `rollJournalCounter` nor `rollJournalBytes` is reset on leader transfer (`init()`) - the new counter deliberately mirrors the existing counter's behavior.

Fixes #75718

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

kevincai (approved):
While this change looks good, a small edit log roll bytes, would still cause too frequent checkpoint traffic in the background, this would have impact to the FE memory footprint and CPU usage too.

Worth to figure out the huge editlog, and trim unnecessary context from the editlog.

gengjun-git (approved):

PR #75734, opened 2026-07-02
[BugFix] Report per-physical bucket count for lake SHOW PARTITIONS/partitions_meta

xiangguangyxg:
## Why I'm doing:

On shared-data (`run_mode = shared_data` / lake) tables, `SHOW PARTITIONS` and `information_schema.partitions_meta` report every physical partition's **Buckets** as the table-level distribution default instead of the physical partition's own bucket count.

Reproduce on any shared-data cluster:

```sql
CREATE TABLE db.t (k1 INT, v1 BIGINT) DUPLICATE KEY(k1) DISTRIBUTED BY RANDOM BUCKETS 3;
ADMIN EXECUTE ON FRONTEND 'metastore.addPhysicalPartition("db","t","t",5,2)';
SHOW PARTITIONS FROM db.t;   -- BUG: all rows show Buckets=3; EXPECTED: 3,5,5
```

The tablets are actually created correctly (the two added physical partitions each have 5 tablets, verifiable via `SHOW PROC '/dbs/db/t/partitions/<physicalPartitionId>/<indexId>/'`), so this is purely a read/reporting bug, not a data issue.

## What I'm doing:

The per-physical bucket count is stored in `PhysicalPartition.bucketNum` (set by the write path `createPhysicalPartition` → `setBucketNum(distributionInfo.getBucketNum())`). The logical `Partition` holds a single shared `DistributionInfo` whose `bucketNum` stays at the table default and is not updated when a physical partition with a different bucket count is added.

The OLAP read path (`PartitionsProcDir.getOlapPartitionInfo`) already prefers `physicalPartition.getBucketNum()` when `> 0`, falling back to the distribution default otherwise. The lake read path and `information_schema.partitions_meta` used `distributionInfo.getBucketNum()` directly — the bug.

This PR makes both lake read paths mirror the OLAP path:

- `PartitionsProcDir.getLakePartitionInfo` (the `SHOW PARTITIONS` Buckets column).
- `InformationSchemaDataSource.genPartitionMetaInfo` (the `partitions_meta` BUCKETS column) — the physical partition is already in scope there, and the same change also aligns the OLAP rows of `partitions_meta` with what `SHOW PARTITIONS` already reports.

Both use `physicalPartition.getBucketNum() > 0 ? physicalPartition.getBucketNum() : <table default>`. The `> 0` guard preserves the fallback for legacy metadata deserialized before `PhysicalPartition.bucketNum` existed.

This only corrects a wrong reported value — no internal FE/BE logic consumes these values for decisions (verified: `SHOW PARTITIONS` output is display-only, the `partitions_meta` `WHERE Buckets` filter runs in-memory over the same displayed value, and the only internal `partitions_meta` consumer, `PartitionSelector`, selects `PARTITION_NAME` only).

Tests:
- `PartitionsProcDirTest`: lake table reports per-physical Buckets (default `3`, added `5`, and a `bucketNum == 0` physical falls back to the table default), plus an OLAP regression guard.
- `LakeInformationSchemaDataSourceTest`: `partitions_meta` reports per-physical Buckets after `addPhysicalPartition` with a bucket count different from the table default.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/proc/PartitionsProcDir.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/proc/PartitionsProcDir.java
@@ -411,7 +411,8 @@ public class PartitionsProcDir implements ProcDirInterface {
                 .stream().map(Column::getName).collect(Collectors.toList()))); // Partition key
         partitionInfo.add(findRangeOrListValues(tblPartitionInfo, partition.getId())); // List or Range
         partitionInfo.add(distributionKeyAsString(table, partition.getDistributionInfo())); // DistributionKey
-        partitionInfo.add(partition.getDistributionInfo().getBucketNum()); // Buckets
+        partitionInfo.add(physicalPartition.getBucketNum() > 0 ?
+                physicalPartition.getBucketNum() : partition.getDistributionInfo().getBucketNum()); // Buckets
         partitionInfo.add(new ByteSizeValue(physicalPartition.storageDataSize())); // DataSize
         long storageSize = physicalPartition.storageDataSize() + physicalPartition.getExtraFileSize();
         partitionInfo.add(new ByteSizeValue(storageSize)); // StorageSize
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

srlch (approved):

luohaha (approved):

PR #75742, opened 2026-07-02
[BugFix] CN nodes should emit metrics with starrocks_cn_ prefix instead of starrocks_be_

### LINKED ISSUE #75741: CN nodes emit Prometheus metrics with starrocks_be_ prefix instead of starrocks_cn_
1. Deploy StarRocks with both BE and CN StatefulSets in the same Kubernetes namespace (e.g. using the StarRocks Operator)
2. Scrape Prometheus metrics from both BE and CN pods' `/metrics` endpoints
3. Query any `starrocks_be_*` metric — it returns results from **both** BE and CN pods
4. There is no way to distinguish BE vs CN metrics by metric name or labels

**Example:** Deploy a cluster with 3 BE pods and 2 CN pods. Query Prometheus:

```promql
starrocks_be_disks_total_capacity
```

Returns 5 time series — all with `starrocks_be_` prefix:

```
starrocks_be_disks_total_capacity{pod="my-cluster-be-0", ...} 107374182400
starrocks_be_disks_total_capacity{pod="my-cluster-be-1", ...} 107374182400
starrocks_be_disks_total_capacity{pod="my-cluster-be-2", ...} 107374182400
starrocks_be_disks_total_capacity{pod="my-cluster-cn-0", ...} 53687091200
starrocks_be_disks_total_capacity{pod="my-cluster-cn-1", ...} 53687091200
```

The CN pods (`cn-0`, `cn-1`) emit metrics with the `starrocks_be_` prefix, identical to actual BE pods. The only way to tell them apart is by parsing the pod name, which is fragile and deployment-specific.

Similarly, scraping the `/metrics` endpoint directly on a CN pod returns:

```
# TYPE starrocks_be_process_mem_bytes gauge
starrocks_be_process_mem_bytes 1073741824
# TYPE starrocks_be_process_thread_num gauge
starrocks_be_process_thread_num 128
# TYPE starrocks_be_compaction_deltas_total counter
starrocks_be_compaction_deltas_total{type="base"} 0
```

All metrics use `starrocks_be_` even though this is a CN process.

CN nodes should emit metrics with `starrocks_cn_` prefix, matching the convention where FE uses `starrocks_fe_` prefix:

```
# Expected from CN pod:
starrocks_cn_process_mem_bytes 1073741824
starrocks_cn_process_thread_num 128
starrocks_cn_compaction_deltas_total{type="base"} 0
starrocks_cn_disks_total_capacity 53687091200
```

This would allow users to:
- Build separate Grafana dashboards/panels for BE vs CN
- Set different alert thresholds for BE and CN
- Aggregate metrics correctly for capacity planning
- Use `{__name__=~"starrocks_(be|cn)_.*"}` to query both when needed

### Real behavior (Required)
CN nodes emit all metrics with `starrocks_be_` prefix, identical to BE nodes. This makes monitoring dashboards mix BE and CN data, causes false positive alerts, and breaks capacity planning aggregations.

**Root cause:** `ProcessMetricsRegistry("starrocks_be")` is hardcoded in `be/src/service/service_be/starrocks_be.cpp:162`, ignoring the `as_cn` boolean that is already available at that callsite. The code already knows whether it's running as CN:

```cpp
void start_be(const std::vector<StorePath>& paths, bool as_cn) {
    std::string process_name = as_cn ? "CN" : "BE";  // as_cn is used here for logging...
    // ...but ignored here:
    static auto* process_metrics_registry = new ProcessMetricsRegistry("starrocks_be");  // always "starrocks_be"
```

Similarly, `TableMetricsManager` in `be/src/common/util/table_metrics.h` hardcodes `MetricRegistry _metrics{"starrocks_be"}`.

All versions — the prefix has been hardcoded since CN was introduced. Verified on 3.x deployed via StarRocks Operator on GKE.

vizsmurali:
## Why I'm doing:

CN (Compute Node) processes emit all Prometheus metrics with the `starrocks_be_` prefix, making it impossible to distinguish BE from CN metrics without relying on external labels like pod name or Kubernetes `job` label. This causes:
- Mixed BE+CN data in monitoring dashboards
- False positive alerts on BE-specific thresholds triggered by CN pods
- Inaccurate capacity planning when metrics are aggregated

### The Problem

In a cluster with 3 BE pods and 2 CN pods, querying any `starrocks_be_*` metric returns results from **all 5 pods**:

```promql
starrocks_be_disks_total_capacity
```

```
starrocks_be_disks_total_capacity{pod="my-cluster-be-0"} 107374182400
starrocks_be_disks_total_capacity{pod="my-cluster-be-1"} 107374182400
starrocks_be_disks_total_capacity{pod="my-cluster-be-2"} 107374182400
starrocks_be_disks_total_capacity{pod="my-cluster-cn-0"} 53687091200   # <-- CN pod with BE prefix!
starrocks_be_disks_total_capacity{pod="my-cluster-cn-1"} 53687091200   # <-- CN pod with BE prefix!
```

Scraping the `/metrics` endpoint directly on a CN pod shows the same issue — every metric uses `starrocks_be_`:

```
# TYPE starrocks_be_process_mem_bytes gauge
starrocks_be_process_mem_bytes 1073741824
# TYPE starrocks_be_process_thread_num gauge
starrocks_be_process_thread_num 128
```

### Root Cause

The `as_cn` flag is already available at the metric registry creation site but is ignored:

```cpp
void start_be(const std::vector<StorePath>& paths, bool as_cn) {
    std::string process_name = as_cn ? "CN" : "BE";  // as_cn used for logging...
    // ...but hardcoded here:
    static auto* process_metrics_registry = new ProcessMetricsRegistry("starrocks_be");
```

Additionally, `TableMetricsManager` in `table_metrics.h` hardcodes its own `MetricRegistry _metrics{"starrocks_be"}`, so even if the root registry prefix were fixed, table-level metrics would still use `starrocks_be_`.

## What I'm doing:

Use the existing `as_cn` flag to set the metric registry prefix to `"starrocks_cn"` when running in CN mode, and propagate the prefix to `TableMetricsManager` which had its own hardcoded `"starrocks_be"` registry.

### Changes (4 files, ~10 lines):

1. **`be/src/service/service_be/starrocks_be.cpp`**: Use `as_cn ? "starrocks_cn" : "starrocks_be"` for `ProcessMetricsRegistry`
2. **`be/src/common/util/table_metrics.h`**: Accept prefix via constructor parameter instead of hardcoding `"starrocks_be"`
3. **`be/src/common/metrics/process_metrics_registry.cpp`**: Forward the registry name to both `_root_registry` and `_table_metrics_mgr`
4. **`be/test/http/metrics_action_test.cpp`**: Update test expectations — registry created with prefix `"test"` now correctly produces `test_table_*` metrics instead of `starrocks_be_table_*`

### After This Fix

BE pods continue to emit metrics with `starrocks_be_` prefix (unchanged):
```
starrocks_be_process_mem_bytes 2147483648
starrocks_be_disks_total_capacity 107374182400
starrocks_be_compaction_deltas_total{type="base"} 42
```

CN pods now emit metrics with `starrocks_cn_` prefix:
```
starrocks_cn_process_mem_bytes 1073741824
starrocks_cn_disks_total_capacity 53687091200
starrocks_cn_compaction_deltas_total{type="base"} 0
```

This matches the convention where FE uses `starrocks_fe_` prefix, and allows users to:
- Build separate Grafana dashboards for BE vs CN without relying on pod name parsing
- Set different alert thresholds for BE-specific and CN-specific metrics
- Aggregate correctly: `sum(starrocks_be_disks_total_capacity)` gives BE-only capacity
- Query both when needed: `{__name__=~"starrocks_(be|cn)_process_mem_bytes"}`

Fixes #75741

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [x] Miscellaneous: CN metric names change from `starrocks_be_*` to `starrocks_cn_*`

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/common/metrics/process_metrics_registry.cpp
+++ b/be/src/common/metrics/process_metrics_registry.cpp
@@ -19,6 +19,6 @@
 namespace starrocks {
 
 ProcessMetricsRegistry::ProcessMetricsRegistry(std::string root_registry_name)
-        : _root_registry(std::move(root_registry_name)) {}
+        : _root_registry(root_registry_name), _table_metrics_mgr(std::move(root_registry_name)) {}
 
 } // namespace starrocks
--- a/be/src/common/util/table_metrics.h
+++ b/be/src/common/util/table_metrics.h
@@ -51,7 +51,8 @@ using TableMetricsPtr = std::shared_ptr<TableMetrics>;
 // so the real deletion will be done asynchronously by daemon thread.
 class TableMetricsManager {
 public:
-    TableMetricsManager() : _last_cleanup_ts(MonotonicSeconds()) {}
+    explicit TableMetricsManager(std::string prefix = "starrocks_be")
+            : _metrics(std::move(prefix)), _last_cleanup_ts(MonotonicSeconds()) {}
 
     MetricRegistry* metric_registry() { return &_metrics; }
 
@@ -72,7 +73,7 @@ public:
 private:
     bool can_install_metrics();
 
-    MetricRegistry _metrics{"starrocks_be"};
+    MetricRegistry _metrics;
     using MutexType = starrocks::bthreads::BThreadSharedMutex;
     using MetricsMap =
             phmap::parallel_flat_hash_map<uint64_t, TableMetricsPtr, phmap::Hash<uint64_t>, phmap::EqualTo<uint64_t>,
--- a/be/src/service/service_be/starrocks_be.cpp
+++ b/be/src/service/service_be/starrocks_be.cpp
@@ -159,7 +159,7 @@ void start_be(const std::vector<StorePath>& paths, bool as_cn) {
 
     int start_step = 1;
     // Metric singletons keep registry back-pointers, so the process registry must outlive shutdown.
-    static auto* process_metrics_registry = new ProcessMetricsRegistry("starrocks_be");
+    static auto* process_metrics_registry = new ProcessMetricsRegistry(as_cn ? "starrocks_cn" : "starrocks_be");
 
     auto daemon = std::make_unique<Daemon>();
     daemon->init(as_cn, paths, process_metrics_registry);
--- a/be/test/http/metrics_action_test.cpp
+++ b/be/test/http/metrics_action_test.cpp
@@ -156,14 +156,14 @@ TEST_F(MetricsActionTest, prometheus_output_with_table_metrics) {
     ProcessMetricsRegistry registry("test");
     enable_table_metrics(&registry);
     s_expect_response =
-            "# TYPE starrocks_be_table_load_bytes counter\n"
-            "starrocks_be_table_load_bytes{table_id=\"42\"} 0\n"
-            "# TYPE starrocks_be_table_load_rows counter\n"
-            "starrocks_be_table_load_rows{table_id=\"42\"} 0\n"
-            "# TYPE starrocks_be_table_scan_read_bytes counter\n"
-            "starrocks_be_table_scan_read_bytes{table_id=\"42\"} 0\n"
-            "# TYPE starrocks_be_table_scan_read_rows counter\n"
-            "starrocks_be_table_scan_read_rows{table_id=\"42\"} 11\n";
+            "# TYPE test_table_load_bytes counter\n"
+            "test_table_load_bytes{table_id=\"42\"} 0\n"
+            "# TYPE test_table_load_rows counter\n"
+            "test_table_load_rows{table_id=\"42\"} 0\n"
+            "# TYPE test_table_scan_read_bytes counter\n"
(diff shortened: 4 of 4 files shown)


kevincai (changes requested):
This is not acceptable, it will break existing grafana dashboard and all the monitoring alerts immediately and also introduce document chaos between `starrocks_be_` and `starrocks_cn_`, this is why we still use the same prefix for cn node at the first place.

Whether running as a CN or as a BE is not that important. The host and the cluster is more important than just distinguishing cn from be in practice.

PR #75755, opened 2026-07-02
[BugFix] Escape jsonpaths value in SHOW CREATE ROUTINE LOAD output

srihithg:
## Why I'm doing:

`SHOW CREATE ROUTINE LOAD` emits the `jsonpaths` property value raw into a double-quoted SQL string. A real `jsonpaths` value contains double quotes, e.g. `["$.id", "$.name"]`, which terminate the enclosing `"..."` early and produce non-runnable DDL — copy/paste of the output fails to parse. The same generated DDL is used in replay/migration and tooling paths, so the job definition is effectively unrecoverable from the output. The sibling `json_root` and `merge_condition` properties have the same latent bug (a merge condition can legitimately contain a string literal like `src.op = "delete"`).

## What I'm doing:

Escape backslashes and double quotes in the `jsonpaths`, `json_root`, and `merge_condition` values in `RoutineLoadJob.jobPropertiesToSql()` via a small targeted helper (`\` first, then `"`).

A targeted escape is used instead of `StringEscapeUtils.escapeJava` (the convention used for `enclose`/`escape`) deliberately: `escapeJava` emits `\uXXXX` for non-ASCII characters, which the parser's `escapeBackSlash()` cannot decode. That would turn a CJK jsonpath like `["$.用户名"]` into a silently wrong path (loading NULLs) instead of failing loudly at parse time. The targeted escape round-trips all inputs exactly. `enclose`/`escape` are single ASCII characters sourced from CSV configs, so `escapeJava` remains safe there and they are left unchanged. The `envelope` and warehouse-name values in the same method are keyword-like/identifier values and are intentionally out of scope as low risk.

Added `ShowCreateRoutineLoadDDLTest` with round-trip tests: the emitted properties SQL is embedded into a full `CREATE ROUTINE LOAD` statement, parsed back, and the recovered values are asserted equal to the originals (double-quoted jsonpaths, non-ASCII jsonpaths, backslashes, `json_root` + `merge_condition` with quoted literals, and the empty default). Without the fix, 4 of the 5 tests fail with the reported symptom (`Unexpected input 'ON'` syntax error).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
@@ -1799,6 +1799,14 @@ public abstract class RoutineLoadJob extends AbstractTxnStateChangeCallback
         return sb.toString();
     }
 
+    // Escape a value for embedding in a double-quoted SQL string literal: backslash first, then
+    // double quote. Pairs with the parser's escapeBackSlash() so the emitted DDL parses back to
+    // the original value. escapeJava is unsuitable here: it emits \\uXXXX for non-ASCII chars,
+    // which escapeBackSlash() cannot decode, silently corrupting e.g. CJK jsonpaths.
+    private static String escapeForDoubleQuotedSql(String value) {
+        return value.replace("\\", "\\\\").replace("\"", "\\\"");
+    }
+
     public String jobPropertiesToSql() {
         StringBuilder sb = new StringBuilder();
         sb.append("(\n");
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

stephen-shelby (approved):

PR #75758, opened 2026-07-02
[BugFix] Iceberg _row_id: use the file-level first_row_id as the reader base

### LINKED ISSUE #75757: Iceberg v3 _row_id is shifted for every row group after the first within a data file
1. Create an Iceberg v3 (row lineage) table whose data file contains more than one Parquet row group, e.g. from Spark:
```sql
CREATE TABLE local.db.t (id INT, payload STRING) USING iceberg
TBLPROPERTIES ('format-version'='3', 'write.parquet.row-group-size-bytes'='2048');
INSERT INTO local.db.t SELECT /*+ REPARTITION(1) */ id, repeat(uuid(), 4) FROM range(0, 1500);
-- one 1500-row data file with 15 row groups of 100 rows
```
2. Query the row lineage column from StarRocks:
```sql
SELECT id, _row_id FROM t WHERE id IN (99, 100, 1499) ORDER BY id;
SELECT count(*), count(distinct _row_id), min(_row_id), max(_row_id) FROM t;
```

`_row_id` = the file's `first_row_id` + file-local row position — contiguous `0..1499`, matching Spark's reader: `99→99, 100→100, 1499→1499`, `max(_row_id)=1499`.

### Real behavior (Required)
Every row group after the first is shifted up by its row-group start: `99→99, 100→200, 1499→2899`, `max(_row_id)=2899` for a 1500-row file.

`GroupReader::_create_column_readers` seeds `IcebergRowIdReader` with `_row_group_first_row_id` (= file `first_row_id` + row-group start) as the base, but `IcebergRowIdReader::read_range` adds a file-local row index that already includes the row-group start, so the row-group start is double-counted. The reader's zone-map filters use the same shifted base, so `_row_id` predicates are evaluated against the shifted range too, and computed values diverge from the physical `_row_id` column of compacted files.

main, reproduced at `188da6e1ec5`; `branch-4.1` carries the same code.

eshishki:
## Why I'm doing:

`SELECT _row_id` on an Iceberg v3 table returns wrong values for any data file
with more than one Parquet row group: GroupReader seeds IcebergRowIdReader with
`_row_group_first_row_id` (= file first_row_id + row-group start) as the base,
while the read range already carries file-local positions, so the row-group
start is double-counted. On a 3000-row fixture (2 files x 15 row groups of 100
rows) `max(_row_id)` comes out 4399 instead of 2999, and id=100 maps to
_row_id=200. The reader's zone-map filters share the shifted base, so _row_id
predicates are evaluated against the shifted range too, and computed values
diverge from the physical _row_id column of compacted files.

## What I'm doing:

- Seed IcebergRowIdReader with the scan range's file-level first_row_id; the
  zone-map formulas (base + rg_first_row) then line up with the emitted values
  by construction. The legacy (non-lineage) GLM lookup branch had the same
  double count — its base is the file-local position origin, plain 0.
- `_row_group_first_row_id`, the GroupReader constructor that carried it and
  the per-row-group counter in FileReader have no remaining users — dropped,
  so the per-row-group base cannot be reintroduced by accident.
- BE UT: a multi-row-group lineage case
  (TestIcebergRowIdSecondRowGroupUsesFileLevelFirstRowId) plus a rework of the
  legacy-path test that encoded the shifted semantics.
- SQL e2e test `test_iceberg_row_id_multi_row_group` in the register_table
  form. Its fixture tree is checked in temporarily under
  `test/sql/test_iceberg/data/row_id_multi_rg` (embedded paths point at
  file:/tmp/sr_rowid_fixture/warehouse) and needs to be hosted on the CI OSS
  bucket under `iceberg_ci_db/rowid_multi_rg`; the fixture commit gets
  dropped once hosted.

Fixes #75757

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5


eshishki:
@Youngwb the SQL case follows the register_table pattern and needs its fixture hosted: the tree is checked in under `test/sql/test_iceberg/data/row_id_multi_rg` (embedded paths are `file:/tmp/sr_rowid_fixture/warehouse`, so rewriting + upload via your existing tooling, target `oss://${oss_bucket}/iceberg_ci_db/rowid_multi_rg`). Once hosted, I'll drop the binary-fixture commit.

The fixture is an Iceberg v3 table: two 1500-row data files of 15 row groups (100 rows each), first_row_id 0 and 1500 — the multi-row-group layout is what reproduces the shifted `_row_id`.

CelerData-Reviewer:
@codex review

be/src/formats/parquet/group_reader.cpp
```diff
@@ -527,9 +516,16 @@ Status GroupReader::_create_column_readers() {
...
                 std::optional<int64_t> first_row_id = std::nullopt;
                 if (_param.scan_range != nullptr && _param.scan_range->__isset.first_row_id) {
-                    first_row_id = std::optional<int64_t>(_row_group_first_row_id);
+                    // IcebergRowIdReader emits `first_row_id + i` where `i` is a file-local row
+                    // index (it already includes the row-group start), so the base must be the
+                    // file-level first_row_id. A row-group-level base double-counts the row-group
+                    // start and shifts the emitted _row_id for every row group after the first,
+                    // diverging from the physical _row_id column of compacted files. The file-level
+                    // base also keeps the reader's zone-map filters (base + rg_first_row) correct.
+                    first_row_id = std::optional<int64_t>(_param.scan_range->first_row_id);
```
dirtysalt, line 525: as described in PlanNodes.thrift

```
    // Iceberg v3 row lineage: first row id of the data file, used to compute _row_id
    // as first_row_id + row_position for non-compacted files.
    // The _last_updated_sequence_number fallback value is passed via the extended_columns map.
    37: optional i64 first_row_id;
```

first row id is supposed to firt row of file not row group.

dirtysalt: do you mean 

```
                if (_param.scan_range != nullptr && _param.scan_range->__isset.first_row_id) {
                    first_row_id = std::optional<int64_t>(_row_group_first_row_id);
                } else if (use_legacy_lookup_row_id) {
                    first_row_id = std::optional<int64_t>(_row_group_first_row_id);
                }
                ColumnReaderPtr row_id_reader =
                        reader != nullptr ? std::make_unique<IcebergRowIdReader>(std::move(reader), first_row_id)
                                          : std::make_unique<IcebergRowIdReader>(first_row_id);
```

we are supposed to pass first row id of file, but not first row id ot row group, to initialize `IcebergRowIdReader`?

dirtysalt:
@eshishki I think it should be a bug.

And about sql-test fix, you can cherry-pick this commit
- create a filesystem iceberg catalog
- upload data files to oss
- call alter table to add data files to iceberg table.

https://github.com/dirtysalt/starrocks/commit/e00187d9e8b32f9153bc28bbae8efcc8f735636f

dirtysalt (dismissed):
LGTM! 🚀

GavinMar (approved):

eshishki:
Thanks @dirtysalt — adopted your approach. I cherry-picked your commit and squashed everything into the single bugfix commit: the test now provisions a hadoop catalog in the CI bucket, uploads the two parquet files, and commits them via `add_files`, so the pre-hosted fixture tree is gone and no manual hosting is needed.

be/src/formats/parquet/group_reader.cpp
eshishki: Yes, exactly. `IcebergRowIdReader` adds a file-local row index — the ranges passed to `read_range()` are file-relative and already include the row-group start. So the base must be the file-level `first_row_id` from the scan range. The old code passed `_row_group_first_row_id` (= file `first_row_id` + row-group start), which counted the row-group start twice: every row group after the first emitted shifted `_row_id` values. The fix passes `_param.scan_range->first_row_id` directly and drops `_row_group_first_row_id` altogether, so a per-row-group base cannot sneak back in. For the legacy lookup path (no lineage columns) the base is plain 0, since GLM keys rows there by file-local position.

dirtysalt (approved):
LGTM! 🚀

Youngwb (approved):

PR #75760, opened 2026-07-03
[Enhancement] Store union-aggregate state once in IVM materialized views

Youngwb:
## Why I'm doing:

Three-union IVM support (`bitmap_union` / `hll_union` / `percentile_union`) recently landed. For those aggregates an incremental MV stores the same state twice: the visible column is an identity finalize of the merged state, and a hidden `__AGG_STATE_` column holds the very same state that refresh merges into. For BITMAP/HLL sketches that doubles the state storage for no benefit.

## What I'm doing:

Collapse the visible column and the hidden state column into one for union-family aggregates: the select item itself becomes the intermediate aggregate under the user's alias, so a single visible column is both the queryable output and the refresh merge target. Finalizable aggregates (`sum`/`avg`/…) keep the visible-finalize + hidden-state pair, and a wrapped aggregate such as `bitmap_count(bitmap_union(x))` still keeps its hidden state column.

Collapsing relocates union state out of the appended `__AGG_STATE_` section, which broke the refresh side's positional pairing — it zipped prefix-filtered schema positions against id-sorted aggregate `ColumnRef`s, two independently derived orders that only align while every state column sits appended in aggregate order. This replaces that with an explicit binding keyed by `ColumnRef` id: `InsertPlanner` stashes the ordered INSERT outputs on `TvrOptContext`; `IvmRewriter` resolves each output through pass-through projections and, where it lands on a bare aggregation output ref, records `refId -> state column name`; `IvmDeltaAggregateRule` looks up each aggregate's column by id and asserts the types match, so a mis-binding fails loudly instead of merging one aggregate's state into another's column. Nothing is persisted — refresh re-derives the rewritten statement from the original query, so the binding is rebuilt each run.

Incremental union MVs created before this change use the old two-column layout and will fail refresh with a column-count mismatch; recreate them. Three-union IVM is not in a released version, so no supported upgrade is affected.

Fixes #issue: N/A — internal IVM storage optimization, no tracking issue.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/InsertPlanner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/InsertPlanner.java
@@ -582,6 +582,11 @@ public class InsertPlanner {
         try (Timer ignore2 = Tracers.watchScope("Optimizer")) {
             OptimizerContext optimizerContext = OptimizerFactory.initContext(session, columnRefFactory);
             optimizerContext.setSourceTablesCount(sourceTablesCount);
+            if (session.getSessionVariable().isEnableIVMRefresh()) {
+                // Position i of outputColumns writes targetTable fullSchema[i]; IvmRewriter relies on
+                // this pairing to bind aggregates to MV state columns (bindStateColumnsForAggregate).
+                optimizerContext.getTvrOptContext().setIvmInsertOutputColumns(outputColumns);
+            }
             Optimizer optimizer = OptimizerFactory.create(optimizerContext);
             optimizedPlan = optimizer.optimize(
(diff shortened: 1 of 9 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

Astralidea (approved):

PR #75763, opened 2026-07-03
[Enhancement] Make lake_autovacuum_parallel_partitions mutable with fair LRU vacuum scheduling

starrocks-xupeng:
## Why I'm doing:

`lake_autovacuum_parallel_partitions` controls how many partitions AutoVacuum can process
concurrently in a shared-data cluster. Today the value is baked into a
`BlockingThreadPoolExecutorService` at `AutovacuumDaemon` construction time, with two
consequences:

1. It cannot be tuned without restarting the FE.
2. The pool is *blocking*: once it is saturated, `execute()` blocks the single auto-vacuum
   daemon thread, so the whole vacuum loop stalls behind the busiest partitions.

Fixing (2) means the daemon can no longer lean on the pool to throttle — it has to cap
concurrency itself with a non-blocking "outer gate". A non-blocking gate only submits up to the
cap each round and then stops, so the *order* candidates are considered in now matters: with a
fixed metastore order, partitions near the front would always be submitted first and those near
the back could starve. That is why the new scheduling also has to be fair
(least-recently-vacuumed first). (The old blocking pool did not have this problem — it simply
blocked the daemon thread until every candidate was eventually submitted.)

Separately, `AutovacuumDaemon` extended `FrontendDaemon` (a raw `Thread`, started once). On
leader demotion it was never stopped, and because a `Thread` cannot be restarted, once this FE
was demoted and later re-elected AutoVacuum would silently never run again. Vacuum is a
leader-only job, so it should follow the same lease-checked, restartable lifecycle as the other
leader daemons (e.g. `PublishVersionDaemon`).

## What I'm doing:

- **`Config.java`**: mark `lake_autovacuum_parallel_partitions` as `mutable = true` so it can be
  tuned at runtime (default stays `8`).
- **`AutovacuumDaemon`**: replace the construction-time `BlockingThreadPoolExecutorService` with a
  lazily-created, non-blocking `ThreadPoolExecutor` (`ThreadPoolManager.newDaemonFixedThreadPool`)
  that is **resized in place** by a `ConfigRefreshDaemon` listener (`adjustExecutorService`),
  mirroring `PublishVersionDaemon#getTaskExecutor`. The listener is registered once per daemon
  instance; `adjustExecutorService` ignores non-positive values so the pool is never resized to an
  illegal size.
- **Outer concurrency gate**: `scheduleVacuumRound()` bounds in-flight partitions against
  `vacuumingPartitions.size()` vs. the (re-read-every-round) config, instead of relying on the pool
  to block. This makes the config take effect on the next round and guarantees a slow vacuum never
  blocks the daemon thread. The effective value is clamped to `MAX_PARALLEL_PARTITIONS = 256` (well
  below the 4096 pool queue) so the pool's `BlockedPolicy` never fills its queue and never blocks the
  daemon, no matter how large the mutable config is set; `submitPendingCandidates` also stops the
  round on a rejected submission rather than blocking on every remaining candidate. Setting the config
  to `0` or a negative value disables AutoVacuum entirely (documented in the config reference).
- **Queue-driven, fair (LRU) scheduling**: `collectVacuumCandidates()` gathers every partition that
  needs vacuuming (excluding in-flight ones), sorts them oldest-`lastVacuumTime`-first (LRU), and
  caches them in a queue. Each round drains the queue to fill free slots (re-checking in-flight and
  `shouldVacuum` before submitting) and only collects again once the queue is empty — this mirrors
  the original "finish the batch, then re-select" behavior while avoiding a full table-walk every
  round. When a collection finds nothing to vacuum, scanning backs off for 30s so an idle cluster is
  not walked every round. LRU keeps things fair: a partition that keeps losing the race only gets
  older and rises to the front on the next collection, so nothing starves.
- **Leak fix**: if task submission throws (e.g. `RejectedExecutionException` when the queue is
  saturated), roll the partition id back out of `vacuumingPartitions` — otherwise the never-run
  task's `finally` never removes it and the partition would stay "in flight" forever until an FE
  restart.
- **Run AutoVacuum as a `LeaderDaemon`**: migrate `AutovacuumDaemon` from `FrontendDaemon` to
  `LeaderDaemon`, so each round runs only while the leader lease is valid, the daemon self-stops on
  demotion, and the same instance is restartable on re-election (fixing the latent bug where a
  demoted-then-re-elected FE would never vacuum again). `runAfterCatalogReady()` becomes
  `runAfterLeaseValid()`; a new `onStopped()` shuts down and nulls the executor and releases the
  `vacuumingPartitions` reservations of queued-but-never-run tasks (submissions are wrapped in a
  `VacuumTask` carrying the partition id so the drained queue can be identified), while a
  still-running task keeps its reservation until its own `finally` releases it — `shutdownNow()`
  does not wait for running tasks, so a straggler surviving demotion and re-election must keep its
  entry or it could run concurrently with (and then corrupt the tracking of) the new leader
  session's vacuum of the same partition (`executorListenerRegistered` stays set so the config
  listener is registered exactly once per instance); `GlobalStateMgr#stopLeaderOnlyDaemonThreads()`
  now stops it on demotion. The no-op `acquireBackgroundComputeResource()` call is dropped — its
  result was never read (`vacuumPartitionImpl` resolves the compute resource per table).
- **Docs** (`en`/`zh`/`ja` `shared_lake_other.md`): update the `lake_autovacuum_parallel_partitions`
  "Is mutable" field from `No` to `Yes` and document that `0`/negative disables AutoVacuum.
- **UT** (`VacuumTest`): add tests for pool resize (grow / shrink / non-positive guard), candidate
  collection excluding in-flight partitions + LRU ordering, the scheduling gates plus
  task-submission rollback on `RejectedExecutionException`, queue refill only when empty plus the
  empty-collection back-off, and `onStopped` releasing queued-but-never-run reservations while a
  still-running task keeps its entry until its own `finally` releases it. `VacuumTest` passes 16/16.

Invariant preserved: `vacuumPartitionImpl` (the actual per-partition vacuum RPC path) is unchanged.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/docs/zh/administration/management/FE_parameters/shared_lake_other.md
+++ b/docs/zh/administration/management/FE_parameters/shared_lake_other.md
@@ -370,8 +370,8 @@ ADMIN SET FRONTEND CONFIG ("key" = "value");
 - 默认值: 8
 - 类型: Int
 - 单位: -
-- 是否可变: No
-- 描述: 存算分离集群中可同时进行 AutoVacuum 的分区最大数量。AutoVacuum 是 Compactions 后的垃圾回收。
+- 是否可变: Yes
+- 描述: 存算分离集群中可同时进行 AutoVacuum 的分区最大数量。AutoVacuum 是 Compactions 后的垃圾回收。设置为 `0` 或负数将完全禁用 AutoVacuum。
 - 引入版本: v3.1.0
 
 ### `lake_autovacuum_partition_naptime_seconds`
--- a/fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
@@ -27,7 +27,8 @@ import com.starrocks.catalog.TabletInvertedIndex;
 import com.starrocks.catalog.TabletMeta;
 import com.starrocks.common.Config;
 import com.starrocks.common.FeConstants;
-import com.starrocks.common.util.FrontendDaemon;
+import com.starrocks.common.ThreadPoolManager;
+import com.starrocks.common.util.LeaderDaemon;
 import com.starrocks.common.util.concurrent.lock.LockType;
 import com.starrocks.common.util.concurrent.lock.Locker;
 import com.starrocks.lake.LakeAggregator;
@@ -47,21 +48,23 @@ import com.starrocks.server.WarehouseManager;
 import com.starrocks.system.ComputeNode;
 import com.starrocks.system.SystemInfoService;
 import com.starrocks.warehouse.cngroup.ComputeResource;
-import org.apache.hadoop.util.BlockingThreadPoolExecutorService;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.stream.Collectors;
 
-public class AutovacuumDaemon extends FrontendDaemon {
+public class AutovacuumDaemon extends LeaderDaemon {
     private static final Logger LOG = LogManager.getLogger(AutovacuumDaemon.class);
 
     private static final long MILLISECONDS_PER_SECOND = 1000;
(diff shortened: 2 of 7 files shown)


sevev (dismissed):

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
```diff
@@ -47,45 +48,133 @@
...
-            Config.lake_autovacuum_parallel_partitions, 0, 1, TimeUnit.HOURS, "autovacuum");
+
+    // Lazily created on the leader (see getExecutorService()). It is resized in place when
+    // lake_autovacuum_parallel_partitions changes, never rebuilt.
+    private ThreadPoolExecutor executorService = null;
+    private boolean executorListenerRegistered = false;
 
     public AutovacuumDaemon() {
-        super("auto-vacuum", 2000);
+        super("auto-vacuum", 10000 /* 10s */);
```
kevincai: why change this from 2s to 10s?

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
```diff
@@ -47,45 +48,133 @@
...
+                    // here, otherwise this partition would stay "in flight" forever and never be vacuumed
+                    // again until an FE restart.
+                    vacuumingPartitions.remove(partitionId);
+                    LOG.warn("Failed to submit vacuum task for partition {}", partitionId, e);
+                }
+            }
+        }
+    }
 
+    private List<VacuumCandidate> collectVacuumCandidates() {
```
kevincai, line 243: As stated, this could be expensive to collect the full candidates and just for a few slot scheduling. A better strategy could be: sampling collection for most of the rounds and only full collecting at a much longer fixed rate as a fallback.

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
starrocks-xupeng: reverted

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
starrocks-xupeng: done

xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
```diff
@@ -47,45 +48,179 @@
...
+    // A non-blocking fixed-size pool resized in place via the ConfigRefreshDaemon listener (mirrors
+    // PublishVersionDaemon#getTaskExecutor). It must never block the caller: concurrency is bounded by the
+    // outer gate in scheduleVacuumRound(), not by this pool. Created lazily so that unit tests exercising
+    // only vacuumPartitionImpl() never register a listener on the global ConfigRefreshDaemon.
+    private ThreadPoolExecutor getExecutorService() {
+        if (executorService == null) {
+            int numThreads = Math.max(1, Config.lake_autovacuum_parallel_partitions);
+            executorService = ThreadPoolManager.newDaemonFixedThreadPool(
+                    numThreads, EXECUTOR_QUEUE_SIZE, "auto_vacuum", true);
+            executorService.allowCoreThreadTimeOut(true);
```
kevincai: **The pool can block this daemon's own thread, despite the "must never block" contract.**

`ThreadPoolManager.newDaemonFixedThreadPool(...)` installs `BlockedPolicy(poolName, 60)` as the rejection handler, which blocks the submitting thread (this daemon's worker) for up to 60s via `queue.offer(r, 60, SECONDS)` once the `EXECUTOR_QUEUE_SIZE = 4096` queue saturates, before it ever throws `RejectedExecutionException`. So the "must never block" invariant above is not actually guaranteed by this pool.

Two changes remove the safety margin now that `lake_autovacuum_parallel_partitions` is `mutable = true`:

- `adjustExecutorService()` (L122-L131) only guards `newNumThreads <= 0` — there is no upper bound. Unlike `PublishVersionDaemon.adjustTaskExecutor()`, which clamps to a 4096 hard limit, an operator can now set this above the 4096 queue size at runtime, letting in-flight submissions exceed the queue and trigger `BlockedPolicy`.
- `submitPendingCandidates()` (L206-L217) does not `break` after catching `RejectedExecutionException`, so a saturated queue can block the daemon up to 60s *per remaining candidate* in a single round.

Consider clamping the config to a hard maximum (mirroring `PublishVersionDaemon`) and/or constructing the pool with a non-blocking rejection policy so the "must never block" contract holds even under aggressive tuning.

<sub>🤖 Generated with [Claude Code](https://claude.ai/code) — react 👍 if useful, 👎 if not.</sub>

CelerData-Reviewer:
@codex review

docs/zh/administration/management/FE_parameters/shared_lake_other.md
```diff
@@ -370,8 +370,8 @@ ADMIN SET FRONTEND CONFIG ("key" = "value");
 - 默认值: 8
 - 类型: Int
 - 单位: -
-- 是否可变: No
-- 描述: 存算分离集群中可同时进行 AutoVacuum 的分区最大数量。AutoVacuum 是 Compactions 后的垃圾回收。
+- 是否可变: Yes
```
kevincai, line 373: why this is a "YES/NO" in Chinese version, shall be "是/否". @EsoragotoSpirit

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
```diff
@@ -47,45 +48,189 @@
...
+    protected void onStopped() {
+        // Leader demotion: release leader-session-only state so a follower does not retain it.
+        // executorListenerRegistered stays set so the ConfigRefreshDaemon listener is registered
+        // exactly once per instance across demote/re-elect cycles (mirrors PublishVersionDaemon).
+        ThreadPoolExecutor executor = executorService;
+        if (executor != null) {
+            executor.shutdownNow();
+            executorService = null;
+        }
+        vacuumingPartitions.clear();
```
kevincai: **Potential double-submit race on leader demotion.** `onStopped()` calls `shutdownNow()` and immediately clears `vacuumingPartitions`, but never awaits termination of the in-flight vacuum tasks. A task interrupted during demotion still runs its `finally { vacuumingPartitions.remove(partition.getId()); }` afterward (L328). If the same FE is re-elected leader and the next round re-adds that partition id, the lagging removal drops the live entry and the partition can be submitted twice (concurrent vacuum of the same partition).

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
starrocks-xupeng: done

fe/fe-core/src/main/java/com/starrocks/lake/vacuum/AutovacuumDaemon.java
starrocks-xupeng: Good catch — fixed in 7d300247c81, and the race is actually one step wider than described: clearing the set wholesale already lets a re-elected session submit the partition while the interrupted straggler is still running (interrupts don't help a task stuck in a non-interruptible section, e.g. waiting on a table lock), and the straggler's late `finally` then drops the new session's live entry, allowing a third concurrent submission.

Fix: submissions are wrapped in a `VacuumTask` carrying the partition id, and `onStopped()` no longer clears `vacuumingPartitions` — it only releases the reservations of queued-but-never-run tasks returned by `shutdownNow()`:

```java
for (Runnable task : executor.shutdownNow()) {
    if (task instanceof VacuumTask) {
        vacuumingPartitions.remove(((VacuumTask) task).partitionId);
    }
}
```

A still-running task keeps its entry until its own `finally` releases it. `ThreadPoolExecutor` guarantees every submitted task either runs `run()` to completion (its `finally` releases the entry) or is returned by `shutdownNow()` (the loop above releases it), so every reservation is released exactly once, and a re-elected session keeps skipping the partition until the straggler actually exits — no `awaitTermination` needed, so demotion never blocks on a stuck vacuum.

Covered by the reworked `testOnStoppedReleasesLeaderSessionState`: a latch-blocked straggler keeps its reservation across `onStopped()` while the queued task's reservation is released, and the straggler's own `finally` releases the last entry.

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

PR #75764, opened 2026-07-03
[BugFix] Reconstruct intermediate flat-JSON object instead of returning NULL

Hyper-FF:
## Why I'm doing:

get_json_string(j,'$.o') on a flat_json.enable=true table returned NULL when the nested object `o` was flattened into leaf sub-columns (o.inner) and the segment had no remain. cbo_prune_json_subfield / cbo_json_v2_rewrite rewrite the call into a typed read of the extended column j.o; Segment::_new_extended_column_iterator found no sub-column named exactly "o" and short-circuited to a NULL default, before the JsonExtractIterator that would reconstruct the object from its o.* leaves.

Treat a field that is a prefix of an existing flat sub-column as present (an intermediate object node with flattened descendants) so the reconstruction path runs. Also fixes json_query / -> on such intermediate objects.

Add regression test test/sql/test_json/test_flat_json_intermediate_object.

## What I'm doing:

```
create table t(id int, j json) duplicate key(id)
  distributed by hash(id) buckets 1
  properties("replication_num"="1","flat_json.enable"="true");
insert into t select generate_series,
  parse_json(concat('{"o":{"inner":', generate_series, '},"leaf":', generate_series, '}'))
  from table(generate_series(1,300));

mysql> select get_json_string(j,'$.o') from t order by id limit 2;
+---------------------------+
| get_json_string(j, '$.o') |
+---------------------------+
| NULL                      |
| NULL                      |
+---------------------------+
2 rows in set (0.05 sec)

mysql> set cbo_prune_json_subfield=false;
Query OK, 0 rows affected (0.00 sec)

mysql> select get_json_string(j,'$.o') from t order by id limit 2;
+---------------------------+
| get_json_string(j, '$.o') |
+---------------------------+
| {"inner": 1}              |
| {"inner": 2}              |
+---------------------------+
2 rows in set (0.03 sec)
```

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/rowset/segment.cpp
+++ b/be/src/storage/rowset/segment.cpp
@@ -594,7 +594,22 @@ StatusOr<ColumnIteratorUPtr> Segment::_new_extended_column_iterator(const Tablet
         std::string_view leaf = paths.back();
         may_contains = column_reader->get_remain_filter()->test_bytes(leaf.data(), leaf.size());
     }
-    if (column_reader->is_flat_json() && !may_contains) {
+    // An intermediate node (e.g. "o") whose descendants are stored as flattened sub-columns
+    // (e.g. "o.inner") is present in this segment even though no sub-column is named exactly
+    // "o". Such a node must be reconstructed by the JsonExtractIterator below, not reported as
+    // absent -- otherwise extracting an intermediate JSON object through a json-path-rewrite
+    // extended column (e.g. get_json_string(j, '$.o')) wrongly returns NULL.
+    bool has_flatten_descendant = false;
+    if (sub_readers) {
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #75779, opened 2026-07-03
[BugFix] Read Iceberg column added by ADD COLUMN before a new snapshot

GavinMar:
## Why I'm doing:

`ALTER TABLE ... ADD COLUMN` on an Iceberg table is a metadata-only commit: it advances the current schema but does not produce a new snapshot. `StarRocksIcebergTableScan` extends Iceberg's `DataScan`, whose `useSnapshotSchema()` returns `true`, so `useSnapshot(snapshotId)` rebinds the scan to the schema recorded on that snapshot. For a normal (current-snapshot) read, that recorded schema predates the metadata-only `ADD COLUMN`, so pushing a predicate down on the newly added column fails to bind and the query errors out:

```
ERROR 1064 (HY000): Cannot find field 'xxx' in struct: struct<...>
```

The failure is conditional: it only triggers when the new column is referenced in a predicate and no write has committed a snapshot after the `ADD COLUMN`. A pure projection of the new column already works (missing columns are silently backfilled with NULL), and any subsequent write (INSERT/MERGE) commits a new snapshot whose schema includes the column, after which reads recover. `REFRESH EXTERNAL TABLE` and rebuilding the catalog do not help, since this is a schema-binding issue, not a cache issue.

Fixes https://github.com/StarRocks/StarRocksTest/issues/11432

## What I'm doing:

Override `useSnapshot` in `StarRocksIcebergTableScan`: when the target snapshot is the table's current snapshot, scan with the current table schema (so a predicate on a column added by a metadata-only `ADD COLUMN` binds correctly). Older snapshots continue to use the per-snapshot schema, preserving time-travel semantics.

Verification:
- FE UT `IcebergMetadataTest#testGetRemoteFileWithPredicateOnColumnAddedAfterSnapshot` reproduces the exact `Cannot find field` error without the fix and passes with it.
- SQL test `test_iceberg_read_new_column_after_add_column` added to `test_iceberg_alter_table` (predicate on the new column right after `ADD COLUMN`, and again after a write).
- End-to-end on a hive Iceberg catalog: the reported queries error before the fix and return correct results after, on the same cluster.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/iceberg/StarRocksIcebergTableScan.java
+++ b/fe/fe-core/src/main/java/org/apache/iceberg/StarRocksIcebergTableScan.java
@@ -138,6 +138,21 @@ public class StarRocksIcebergTableScan
     }
 
     @Override
+    public TableScan useSnapshot(long scanSnapshotId) {
+        // The parent binds the scan to the target snapshot's recorded schema (DataScan#useSnapshotSchema).
+        // A metadata-only ADD COLUMN advances the schema without a new snapshot, so a current-snapshot
+        // read would miss the new column and a filter on it fails with "Cannot find field". Delegate to
+        // the parent (inheriting its validation and context assembly), then rebind only the schema: the
+        // current snapshot uses the current table schema; older snapshots keep theirs for time travel.
+        TableScan scan = super.useSnapshot(scanSnapshotId);
+        Snapshot currentSnapshot = table().currentSnapshot();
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

PR #75789, opened 2026-07-03
[BugFix] Fix query cache normalization crash for tables with sub-partitions

dirtysalt:
## Why I'm doing:
When a table's logical partition has been split into multiple physical sub-partitions and `enable_query_cache=true`, planning certain queries throws `Preconditions.checkState` inside `OlapScanNode.toNormalForm()` while building the query-cache digest.

Root cause: to build the cache key, `toNormalForm()` re-derived the physical partition ids by expanding every sub-partition of each selected logical partition, then zipped that list *positionally* against a separately-tracked list of scanned partition versions. If a logical partition has any empty sub-partition (pruned away earlier because it has no matching tablets), the re-derived id list ends up longer than the version list:

```
logical partition P
 ├─ sub-partition P1 (non-empty, scanned) ──► version recorded
 └─ sub-partition P2 (empty, pruned)      ──► no version recorded

toNormalForm() re-expands P -> [P1, P2]   (2 ids)
                versions   -> [v1]        (1 version)
                size mismatch -> crash
```

A secondary issue in the same area: the loop that filters out empty partitions could push the same logical partition id multiple times (once per non-empty sub-partition it has), which is a latent correctness bug for other logical-id consumers (e.g. bucket-count lookup).

## What I'm doing:
- `OlapScanNode.toNormalForm()` no longer re-derives physical-partition-id → version pairs by expanding logical ids. It now reads directly from `scanPartitionVersions`, a map that is already populated with exactly the physical partitions actually scanned and their versions — no re-derivation, no positional pairing, no mismatch possible. Added a lightweight sanity check (every scanned physical partition must belong to a selected logical partition) as a cheap guard against future drift.
- `PlanFragmentBuilder`'s empty-partition filtering now dedupes the resulting logical partition id list (`Set` instead of `List`), and `OlapScanNode#setSelectedPartitionIds` keeps `selectedPartitionNum` in sync with it.
- Removed the now-unused flat `selectedPartitionVersions` list/getter, superseded by `scanPartitionVersions`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
@@ -130,7 +130,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 import javax.annotation.Nullable;
 
@@ -142,7 +141,6 @@ public class OlapScanNode extends AbstractOlapTableScanNode {
 
     private final List<TScanRangeLocations> result = new ArrayList<>();
     private final List<String> selectedPartitionNames = Lists.newArrayList();
-    private List<Long> selectedPartitionVersions = Lists.newArrayList();
     private final HashSet<Long> scanBackendIds = new HashSet<>();
     private final List<String> unUsedOutputStringColumns = new ArrayList<>();
     // a bucket seq may map to many tablets, and each tablet has a TScanRangeLocations.
@@ -296,14 +294,11 @@ public class OlapScanNode extends AbstractOlapTableScanNode {
 
     public void setSelectedPartitionIds(List<Long> selectedPartitionIds) {
         this.selectedPartitionIds = selectedPartitionIds;
+        this.selectedPartitionNum = selectedPartitionIds.size();
     }
 
     public List<String> getSelectedPartitionNames() {
         return selectedPartitionNames;
-    }
-
-    public List<Long> getSelectedPartitionVersions() {
-        return selectedPartitionVersions;
     }
 
     public List<Expr> getPrunedPartitionPredicates() {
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

satanson (dismissed):

sevev (dismissed):

CelerData-Reviewer:
@codex review

dirtysalt:
## Why update `testPruneLimit`?

The test expects `partitions=7/7` but the plan now outputs `partitions=1/7`. Root cause:

1. `updateScanInfo()` receives 7 partition IDs from the optimizer, so `selectedPartitionNum` is initially set to 7
2. `PlanFragmentBuilder.visitPhysicalOlapScan()` then filters out empty partitions — it checks `partitionToScanTabletMap` (keyed by physical partition id) and only retains logical partitions whose physical sub-partitions have non-empty selected tablets. In this `LIMIT 2` query, the optimizer selects only 1 tablet (enough to satisfy LIMIT 2), so only 1 partition has a non-empty entry in the map
3. The filtering loop reduces `selectedPartitionIds` to 1 element (the single non-empty partition)
4. Before this PR: `setSelectedPartitionIds()` only updated `selectedPartitionIds` but NOT `selectedPartitionNum` — the old `selectedPartitionNum = 7` was **stale**, causing the incorrect display `partitions=7/7`
5. This PR adds `selectedPartitionNum = selectedPartitionIds.size()` in `setSelectedPartitionIds()`, so it now correctly reflects the actual number of scanned partitions → `partitions=1/7`

The same reasoning applies to `tabletRatio`: `totalTabletsNum` was always only accumulated for non-empty partitions (inside the same filtering loop), so it was 48 (one partition`'s tablets) not 336 (all 7 partitions). The new `tabletRatio=1/48` is just accurate — the previous `1` was matching as a substring of `1/336`.

This is purely a **display-accuracy fix** — the actual scan behavior is unchanged.

HangyuanLiu (approved):

satanson (approved):

PR #44326, opened 2024-04-18
[Enhancement] Support hive partition filter and quantity limit

zenoyang:
## Why I'm doing:

Through partition filter restrictions, avoid unreasonable queries causing cluster BE or FE to be unavailable.

## What I'm doing:
Fixes #issue

1. When the number of partitions after partition filtering is greater than > `scan_hive_partition_num_limit`, an exception is thrown.
2. When the query does not have a partition predicate, an exception is thrown
3. When the partition column is used as a function parameter or the constant cannot be folded, causing partition pruning is invalid, an exception is thrown.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 3.3
  - [ ] 3.2
  - [ ] 3.1
  - [ ] 3.0
  - [ ] 2.5

PR #75790, opened 2026-07-03
[Enhancement] Support partition filter requirement and partition count limit for Iceberg and Delta Lake

dirtysalt:
## Why I'm doing:

see also:
- https://github.com/StarRocks/starrocks/pull/44326
- https://github.com/StarRocks/starrocks/pull/68480

Hive/Hudi/Paimon already let users guard against accidental full-table scans on huge external tables via `allow_hive_without_partition_filter` (reject queries with no partition filter) and `scan_lake_partition_num_limit` (cap how many partitions a single scan may touch). Iceberg and Delta Lake had no equivalent protection, so a `SELECT *` on a large lake table could silently trigger a massive scan and destabilize the cluster.

## What I'm doing:

- Generalize `allow_hive_without_partition_filter` into `allow_lake_without_partition_filter` (old name kept as a backward-compatible alias), and extend the partition-filter-required check to Iceberg and Delta Lake.
- Extend `scan_lake_partition_num_limit` enforcement to Iceberg and Delta Lake as well.

The tricky part is that these two checks can't be enforced the same way for every catalog type, because Iceberg/Delta Lake enumerate splits lazily/incrementally instead of upfront like Hive/Paimon:

```
Hive / Hudi / Paimon (eager)             Iceberg / Delta Lake (lazy/incremental)
────────────────────────────             ───────────────────────────────────────
Optimize time:                           Optimize time:
 ├─ list partitions (cheap)                ├─ predicate-shape check only
 ├─ check filter required  ✓               │   (allow_lake_without_partition_filter)
 └─ check partition count  ✓             Scan-range dispatch (execution):
                                           └─ check partition count as each new
Nothing left to check at dispatch time       partition streams in
                                              (scan_lake_partition_num_limit)
```

- The **filter-required** check only needs the predicate shape plus static partition-column metadata, so it still runs at optimize time in `OptExternalPartitionPruner`, same as Hive.
- The **partition-count-limit** check needs the actual set of partitions touched, which for Iceberg/Delta is only known as files stream in during scan-range dispatch. It's enforced in a new shared `ConnectorScanRangeSource.checkPartitionNumLimit()` helper, called from both `IcebergConnectorScanRangeSource` and `DeltaConnectorScanRangeSource` as each new distinct partition is discovered.
- One resulting behavior difference worth flagging: under the default incremental dispatch mode, an Iceberg/Delta query that busts the limit may fail *mid-flight* (after some scan ranges are already dispatched to BE) rather than being rejected up front like Hive/Paimon.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
@@ -1048,6 +1048,8 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
 
     public static final String ALLOW_HIVE_WITHOUT_PARTITION_FILTER = "allow_hive_without_partition_filter";
 
+    public static final String ALLOW_LAKE_WITHOUT_PARTITION_FILTER = "allow_lake_without_partition_filter";
+
     public static final String SCAN_HIVE_PARTITION_NUM_LIMIT = "scan_hive_partition_num_limit";
 
     public static final String SCAN_OLAP_PARTITION_NUM_LIMIT = "scan_olap_partition_num_limit";
@@ -3143,8 +3145,8 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
     @VarAttr(name = ENABLE_EXPR_PRUNE_PARTITION, flag = VariableMgr.INVISIBLE)
     private boolean enableExprPrunePartition = true;
 
-    @VarAttr(name = ALLOW_HIVE_WITHOUT_PARTITION_FILTER)
-    private boolean allowHiveWithoutPartitionFilter = true;
+    @VarAttr(name = ALLOW_LAKE_WITHOUT_PARTITION_FILTER, alias = ALLOW_HIVE_WITHOUT_PARTITION_FILTER)
+    private boolean allowLakeWithoutPartitionFilter = true;
 
     // For the maximum number of partitions allowed to be scanned in a single olap table, 0 means no limit.
     @VarAttr(name = SCAN_OLAP_PARTITION_NUM_LIMIT)
@@ -5782,12 +5784,12 @@ public class SessionVariable implements Serializable, Writable, Cloneable {
         this.enableExprPrunePartition = enableExprPrunePartition;
     }
 
-    public boolean isAllowHiveWithoutPartitionFilter() {
-        return allowHiveWithoutPartitionFilter;
+    public boolean isAllowLakeWithoutPartitionFilter() {
+        return allowLakeWithoutPartitionFilter;
     }
 
-    public void setAllowHiveWithoutPartitionFilter(boolean allowHiveWithoutPartitionFilter) {
-        this.allowHiveWithoutPartitionFilter = allowHiveWithoutPartitionFilter;
+    public void setAllowLakeWithoutPartitionFilter(boolean allowLakeWithoutPartitionFilter) {
+        this.allowLakeWithoutPartitionFilter = allowLakeWithoutPartitionFilter;
     }
 
     public int getScanOlapPartitionNumLimit() {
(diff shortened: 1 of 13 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
```diff
@@ -3143,8 +3145,8 @@ public void setConnectorSinkSortScope(String connectorSinkSortScope) {
     @VarAttr(name = ENABLE_EXPR_PRUNE_PARTITION, flag = VariableMgr.INVISIBLE)
     private boolean enableExprPrunePartition = true;
 
-    @VarAttr(name = ALLOW_HIVE_WITHOUT_PARTITION_FILTER)
-    private boolean allowHiveWithoutPartitionFilter = true;
+    @VarAttr(name = ALLOW_LAKE_WITHOUT_PARTITION_FILTER, alias = ALLOW_HIVE_WITHOUT_PARTITION_FILTER)
```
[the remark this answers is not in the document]
dirtysalt: Will fix in another pr: https://github.com/StarRocks/starrocks/pull/75813

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/qe/SessionVariable.java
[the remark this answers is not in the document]
dirtysalt: will fix in https://github.com/StarRocks/starrocks/pull/75813

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/OptExternalPartitionPruner.java
```diff
@@ -213,6 +214,10 @@ private static boolean canPartitionPrune(ScalarOperator partitionPredicate, Set<
                 }
             }
         } else if (partitionPredicate instanceof CompoundPredicateOperator) {
+            CompoundPredicateOperator cpo = (CompoundPredicateOperator) partitionPredicate;
+            if (cpo.isNot()) {
```
stephen-shelby, line 218: why change this behavior?

dirtysalt: it's recommended by codex, and I think it makes sense. @stephen-shelby 


> Guard unary NOT before running the shared filter check

> When allow_lake_without_partition_filter=false, an Iceberg/Delta predicate such as WHERE NOT (date = '2020-01-01') reaches this new shared check because it contains a partition column. checkPartitionPredicates() then calls canPartitionPrune() on a CompoundPredicateOperator.NOT, whose implementation reads child 1 even though NOT has only one child, so these queries fail with an INTERNAL_ERROR instead of a clean partition-filter validation/pruning result. Please handle unary NOT before applying the binary compound logic.

dirtysalt: Or do you think it's unecessary?

stephen-shelby (approved):

Youngwb (approved):

PR #75814, opened 2026-07-03
[Enhancement] Cap per-compaction input bytes for lake primary-key compaction

luohaha:
## Why

Under a skewed realtime-ingest workload where writes concentrate on one large hot primary-key tablet (e.g. a PK-range distributed table with a zipf tenant head), the lake `PrimaryCompactionPolicy` per-compaction input cap is

```
max(update_compaction_result_bytes, tablet_data_size * update_compaction_ratio_threshold)
```

On a large hot tablet the data-proportional term dominates and lets a single compaction grow to multiple GB. Publishing that compaction is synchronous (compact_mapper read + primary-index apply — measured ~14s for a ~2GB rowset) and it runs on the tablet's **serialized** version chain, so all queued realtime-ingest publishes wait behind it. Result: a long realtime-ingest tail (per-batch p99). A HASH-distributed table scatters the hot tenants across many tablets/nodes, so no single tablet builds a compaction that large and the tail does not appear.

## What

Add a mutable BE config `lake_pk_compaction_max_result_bytes`:

- Default `0` = **disabled**, preserving the exact legacy cap and behavior.
- When `> 0`, the effective per-compaction input-bytes cap is clamped to this value, bounding each compaction's synchronous publish cost so it cannot stall queued ingest publishes on a hot PK tablet.

The clamp is applied in `PrimaryCompactionPolicy::pick_rowset_indexes` at both the primary-level and real-time other-level pick loops. No behavior change unless the operator sets the config.

## Test

Config-gated, default-off, no behavior change at default. Evaluated on a skewed range-vs-hash realtime-ingest benchmark (draft; measuring whether the range ingest tail closes to within the hash baseline).

### DIFF
--- a/be/src/common/config.h
+++ b/be/src/common/config.h
@@ -458,6 +458,13 @@ CONF_mInt64(update_compaction_result_bytes, "1073741824");
 CONF_mInt32(update_compaction_delvec_file_io_amp_ratio, "2");
 // This config defines the maximum percentage of data allowed per compaction
 CONF_mDouble(update_compaction_ratio_threshold, "0.5");
+// Absolute upper bound (bytes) on the input size picked for a single lake primary-key
+// compaction. 0 = disabled (legacy behavior: cap =
+// max(update_compaction_result_bytes, tablet_data_size * update_compaction_ratio_threshold)).
+// When > 0 the effective per-compaction result-bytes cap is clamped to this value so that a
+// single large compaction on a hot PK tablet cannot produce a multi-second synchronous publish
+// that stalls the tablet's serialized ingest version chain (realtime-ingest tail latency).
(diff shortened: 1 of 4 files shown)


luohaha:
Superseded by #75816. Measured (build 2323) that setting `lake_pk_compaction_max_result_bytes=512MB` as a GLOBAL change regressed the HASH control cluster (ingest P99 6.2s → 118.6s): shrinking per-compaction output caused compaction churn/pileup so publishes backed up. Abandoning the *shrink-size* approach; #75816 instead **defers** an over-large compaction (kept full-size, just not submitted until read-pressure emergency), so it doesn't stall the ingest version chain without causing churn. Keeping this PR draft/won't-merge for the record.

PR #75861, opened 2026-07-06
[BugFix] Remove useless per-rowid seeks in ArrayColumnIterator::fetch_values_by_rowid offsets-only path

### LINKED ISSUE #75860: Offsets-only array late materialization loads all element pages due to useless per-rowid seeks in ArrayColumnIterator::fetch_values_by_rowid

1. Create a duplicate-key table with an ARRAY column and load enough rows that the array elements span many data pages, with a low-selectivity predicate column:

```sql
CREATE TABLE t_repro (
    id   BIGINT,
    flag TINYINT,
    tags ARRAY<STRING>
) DUPLICATE KEY(id)

INSERT INTO t_repro
SELECT g, IF(g % 200 = 0, 1, 0),
       array_repeat(concat('tag_', g, '_', repeat('x', 40)), 20)
FROM TABLE(generate_series(1, 2000000)) AS s(g);
```

2. Run a query that only uses the array length, so the FE pushes down an offsets-only `ColumnAccessPath`, and late materialization fetches the surviving rows via `fetch_values_by_rowid`:

```sql
SET enable_profile = true;
SELECT SUM(array_length(tags)) FROM t_repro WHERE flag = 1;  -- 10000 surviving rows
```

3. Check the OLAP_SCAN counters with `ANALYZE PROFILE FROM '<QueryId>', 0`.

In the offsets-only path (`_access_values == false`), element data is never read, so late materialization should only load the size sub-column pages. `ReadPagesNum` should be in the low hundreds and `LateMaterialize` sub-millisecond for 10K rows.

### Real behavior (Required)

`ArrayColumnIterator::fetch_values_by_rowid` performs a per-rowid `seek_to_ordinal_and_calc_element_ordinal` on the size iterator plus a `seek_to_ordinal` on the element iterator even in the offsets-only branch. Neither seek has any consumer in this mode (no `next_batch`/`next_dict_codes` follows; the loop only needs the array sizes already fetched in bulk at the top of the function). Each rowid pays one size-page prefix decode plus one element page load, which is a real page-cache miss / disk read when consecutive rowids land on different pages.

Measured on the repro above (2M rows, 10K surviving rows):

| Counter | Actual (before fix) | After removing the dead seeks |
|---|---|---|
| ReadPagesNum | 10238 | 185 |
| LateMaterialize | 57.45ms | 0.42ms |
| Hot query time | ~61ms | ~11ms |
| SUM result | 200000 | 200000 (unchanged) |

About 98% of the pages read are element pages loaded by the dead seeks and thrown away.

main (ebc11d20059)

cccaarr:
## Why I'm doing:

When a query only uses `array_length()` / `cardinality()` on an array column, the FE pushes down an offsets-only `ColumnAccessPath` and `ArrayColumnIterator::init` sets `_access_values = false`. In this mode element data is never read. However, `fetch_values_by_rowid` (the late-materialization path) still performs a per-rowid `seek_to_ordinal_and_calc_element_ordinal` on the size iterator and a `seek_to_ordinal` on the element iterator in its offsets-only branch. Neither seek has any consumer: no `next_batch`/`next_dict_codes` follows (contrast with the `_access_values == true` branch), and the loop only needs the array sizes already fetched in bulk by `_array_size_iterator->fetch_values_by_rowid` at the top of the function. Each rowid pays one size-page prefix decode plus one element page load — a real page-cache miss / disk read whenever consecutive rowids fall on different pages.

Measured on a 2M-row table with 10K rows surviving the predicate (`SELECT SUM(array_length(tags)) FROM t_repro WHERE flag = 1`):

| Counter | Before | After |
|---|---|---|
| ReadPagesNum | 10238 | 185 |
| LateMaterialize | 57.45ms | 0.42ms |
| Hot query time | ~61ms | ~11ms |
| SUM result | 200000 | 200000 (unchanged) |

~98% of the pages read before the fix were element pages loaded by the dead seeks and thrown away.

## What I'm doing:

- Remove the three dead seek lines from the offsets-only branch of `ArrayColumnIterator::fetch_values_by_rowid`; the loop now only sums the already-fetched array sizes to size the const default elements column.
- Add `ColumnReaderWriterTest.test_array_fetch_by_rowid_offsets_only`: writes a 20000-row `ARRAY<INT>` column spanning multiple data pages, reads it back through an offsets-only access path with scattered rowids, and asserts the returned array lengths match what was written and the elements column stays a const default column sized by the sum of the fetched array sizes. The test locks semantics, not implementation, and passes both before and after the fix.

Fixes #75860

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/storage/rowset/array_column_iterator.cpp
+++ b/be/src/storage/rowset/array_column_iterator.cpp
@@ -236,9 +236,6 @@ Status ArrayColumnIterator::fetch_values_by_rowid(const rowid_t* rowids, size_t 
 
         size_t size_to_read = 0;
         for (size_t i = 0; i < size; ++i) {
-            RETURN_IF_ERROR(_array_size_iterator->seek_to_ordinal_and_calc_element_ordinal(rowids[i]));
-            size_t element_ordinal = _array_size_iterator->element_ordinal();
-            RETURN_IF_ERROR(_element_iterator->seek_to_ordinal(element_ordinal));
             size_to_read += array_size.get_data()[i];
         }
 
--- a/be/test/storage/rowset/column_reader_writer_test.cpp
+++ b/be/test/storage/rowset/column_reader_writer_test.cpp
@@ -44,6 +44,7 @@
 #include "column/binary_column.h"
 #include "column/chunk_factory.h"
 #include "column/column.h"
+#include "column/column_access_path.h"
 #include "column/column_helper.h"
 #include "column/datum_convert.h"
 #include "column/fixed_length_column.h"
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

kevincai:
shall this fix be backported to 4.1/4.0/3.5?

srlch:
@codex review

cccaarr:
> shall this fix be backported to 4.1/4.0/3.5?

Yes, backport needed

srlch (approved):

wyb (approved):

wanpengfei-git:
@Mergifyio backport branch-4.1 branch-4.0 branch-3.5

PR #75864, opened 2026-07-06
[BugFix] Fix nullsFraction clamp in LargeOrCalculatingVisitor

stephen-shelby:
## Why I'm doing:

In `PredicateStatisticsCalculator.LargeOrCalculatingVisitor.computeOrPredicateStatistics` (the averaging-heuristic path used when an OR predicate has more than `DEFAULT_OR_OPERATOR_LIMIT` (16) disjuncts), the merged `nullsFraction` for a column was computed as:

```java
double nulls = Math.max(1,
        (columnStatistic.getNullsFraction() + rightColumnStatistic.getNullsFraction()) / 2);
```

`nullsFraction` is always a ratio in `[0, 1]`, so the average of two such ratios can never exceed `1`. This means `Math.max(1, avg)` always evaluates to the constant `1.0`, regardless of the actual inputs — it looks like a copy-paste of the NDV floor on the line above (`Math.max(1, ...)` is correct there since a distinct-value count must be at least 1, but the same floor doesn't make sense for a fraction).

In effect, any OR predicate with more than 16 disjuncts forces `nullsFraction` to 100% for every column touched by more than one arm, regardless of actual null content, which can badly skew downstream selectivity/cardinality estimates.

## What I'm doing:

Changes the incorrect `Math.max(1, ...)` floor to `Math.min(1.0, ...)`, since a ratio needs an upper cap, not a floor of 1 (the merged value is normally already the plain average of the two arms - the cap only guards against an out-of-range input, since `nullsFraction` isn't type-enforced to stay within `[0, 1]`). Added a regression test that ANDs 17 `(a IS NULL) OR (a IS NOT NULL)` groups together (to route through this visitor) and asserts the merged `nullsFraction` is `0.5`, not `1.0`.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/PredicateStatisticsCalculator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/statistics/PredicateStatisticsCalculator.java
@@ -549,7 +549,9 @@ public class PredicateStatisticsCalculator {
                 columnBuilder.setMaxValue(Math.max(columnStatistic.getMaxValue(), rightColumnStatistic.getMaxValue()));
                 double distinct = Math.max(1,
                         (columnStatistic.getDistinctValuesCount() + rightColumnStatistic.getDistinctValuesCount()) / 2);
-                double nulls = Math.max(1,
+                // nullsFraction is a ratio, not a count like distinct/NDV above it - cap at 1
+                // instead of flooring at 1, which forced every merge to the constant 1.0.
+                double nulls = Math.min(1.0,
                         (columnStatistic.getNullsFraction() + rightColumnStatistic.getNullsFraction()) / 2);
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

HangyuanLiu (approved):

Youngwb (approved):

stephen-shelby:
@Mergifyio backport branch-4.0

stephen-shelby:
https://github.com/Mergifyio backport branch-3.5

PR #75883, opened 2026-07-06
[Enhancement] Reassign colocate-boundary split children to the new PACK group immediately

xiangguangyxg:
## Why I'm doing:

On shared-data range-colocate tables, a user-driven colocate-boundary tablet split splices a new
`ColocateRange` and creates a new PACK shard group for the right side of the boundary. The split's
child tablets are created *before* the boundary is known, so they all join the **old** PACK group.
The child whose range now belongs to the new PACK group therefore stays in the old one until the
background reconciler (`ColocateChecker`) migrates it — one to two scheduler ticks later. During that
window the colocate-join optimization degrades on that single bucket (a network hop), and the group
stays unstable so the optimizer will not pick the colocate plan for the group.

Query correctness is never at stake (scan routing depends on the tablet range, not on PACK
membership), and the background reconciler already converges to the correct placement. This change
simply removes the cross-tick delay for the originating split.

## What I'm doing:

- Extract the reconcile primitive — read a set of tablets' actual PACK-group membership from StarOS,
  detect the ones whose membership no longer matches the `ColocateRange` their range belongs to, and
  reassign each with the minimal add/remove delta — into a shared package-static helper
  `ColocateChecker.reconcileTabletPackPlacement(...)`. The background reconciler now delegates to it;
  its behavior is unchanged.
- In `SplitTabletJob.applyColocateRangeSplitResult`, right after the range splice creates the new PACK
  shard group, call that helper on this split's child tablets so the mis-placed child is moved into the
  new PACK group immediately instead of on the next reconcile tick. Only children that map cleanly into
  exactly one post-split `ColocateRange` are reconciled; a child whose range straddles a boundary has no
  single well-defined PACK group and is left to the background reconciler.
- The immediate reassignment is **best-effort**: the split is already published, so a membership-read
  or reassign failure is caught and logged and never aborts the job. Membership is not journaled
  (StarOS persists it); on leader switch or mid-failure the background reconciler re-converges to the
  same terminal placement. Actual membership is re-read via `getShardInfo` (ground truth) rather than
  derived, because concurrent split jobs on the same group can change the range-to-PACK mapping between
  shard creation and this step.

This does not change SQL, configuration, or any external interface. It only changes *when* the
reassignment RPC is issued — the background reconciler already reaches the same terminal placement — so
behavior change is marked **No**.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/ColocateChecker.java
@@ -348,7 +348,33 @@ public class ColocateChecker {
         if (tabletIdToRange.isEmpty()) {
             return true;
         }
+        return reconcileTabletPackPlacement(tabletIdToRange, expectedRanges, colocateColumnCount,
+                db.getFullName() + "." + table.getName());
+    }
 
+    /**
+     * Reads the actual PACK shard-group membership of the given tablets from StarOS in bounded
+     * batches, finds the ones whose membership disagrees with the {@link ColocateRange} their range
+     * belongs to, and reassigns each into its expected PACK shard group with the minimal add/remove
+     * delta. Shared by the placement backstop ({@link #reconcileTablePackPlacement}) and the
+     * post-publish split path ({@code SplitTabletJob}).
+     *
+     * <p>Best-effort per tablet. Catches the expected StarOS checked failures — the batched
+     * {@link StarOSAgent#getShardInfo} {@link StarClientException} and the per-tablet
+     * {@link StarOSAgent#reassignShardGroups} {@link DdlException} — but is intentionally not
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

kevincai (approved):
While this change looks good, the adjustment of the packGroup on startOS side will not trigger an immediate reschedule. It still takes time to reconcile in background by the ShardChecker, this could be delay up to 10 seconds with the default configuration.

srlch (approved):

PR #75888, opened 2026-07-06
[Enhancement] Add file-level Bernoulli sampling for Iceberg external table SAMPLE ANALYZE

dirtysalt:
## Why I'm doing:
Today, `ANALYZE ... SAMPLE` on an Iceberg external table just picks the latest N partitions and does a **full scan** of them. Cost is unbounded whenever a single partition is large — a table with a few TB-scale partitions makes SAMPLE ANALYZE effectively as expensive as FULL, defeating the point of sampling. We want a sampling strategy that:

- Works at the file level across the whole table (not scoped to a handful of partitions),
- Has a hard, predictable cost cap regardless of table size,
- Gets more accurate over time without needing a separate full scan,
- Requires zero BE changes.

## What I'm doing:
Add a Bernoulli file-sampling layer that FE applies to the scan plan right after Iceberg's file list is split into scan units, before scan ranges are handed to BE. Each unit is independently kept/dropped by a probability `ratio`; dropped units cost zero IO and BE never knows sampling happened.

```
read total row_count (Iceberg manifest-list summary fields, zero data IO)
        │
        ▼
round loop (inside one ANALYZE execution, resumes from last run's round):
    rowsThisRound = min(row_count, cap << round)      // cap doubles each round
    ratio = min(1.0, rowsThisRound / row_count)
    → run one sampled collection pass, upsert into stats table
    → stop once ratio hits 1.0 (full coverage) or max rounds reached
        │
        ▼
planFiles() → split into scan units (existing parallelism logic, unchanged)
        → Bernoulli filter(ratio)   // FE-only, drops units before scan ranges are built
        → scan ranges → BE (scans only the kept units)
```

Key pieces:
- **Ratio is row-count calibrated, with an absolute cap**: `ratio = min(1.0, targetRows / totalRows)`, where `totalRows` comes from Iceberg manifest-list summary fields (no data files opened). This bounds a single round's cost by an absolute row budget instead of a bare percentage, so cost stays predictable even for astronomically large tables.
- **Progressive rounds, warm-started across scheduled runs**: within one ANALYZE execution, the target row budget doubles each round until it reaches full coverage or a max-round limit. The round index is persisted in `AnalyzeStatus` properties so the *next* scheduled run resumes from where the last one left off, instead of always restarting at the smallest ratio.
- **Sampling ratio flows through session variables, not SQL text**: the ANALYZE job sets two invisible session variables on its own connection; the Iceberg connector reads them when building the scan. A tricky gotcha worth flagging: the connector has a query-planning-time file-list cache keyed without the sampling ratio — if left alone, cost-estimation would populate that cache with the *unsampled* file list under the same key, and the sampled execution would silently hit that cache and skip sampling entirely. Sampled requests now bypass that cache.
- **Statistics computed from the same sampled pass**: row count is rescaled by `1/ratio`; NDV and min/max are used as-is from the sampled rows (a conservative, CBO-safe direction — no rescaling attempted).
- Hive, Delta, and other external table types are untouched; they keep the existing "latest N partitions, full scan" sampling behavior.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/docs/en/administration/management/FE_parameters/stats_storage.md
+++ b/docs/en/administration/management/FE_parameters/stats_storage.md
@@ -79,6 +79,33 @@ This topic introduces the following types of FE configurations:
 - Description: Duration in seconds for a single process profile collection. When `proc_profile_cpu_enable` or `proc_profile_mem_enable` is set to `true`, AsyncProfiler is started, the collector thread sleeps for this duration, then the profiler is stopped and the profile is written. Larger values increase sample coverage and file size but prolong profiler runtime and delay subsequent collections; smaller values reduce overhead but may produce insufficient samples. Ensure this value aligns with retention settings such as `proc_profile_file_retained_days` and `proc_profile_file_retained_size_bytes`.
 - Introduced in: v3.2.12
 
+### `statistic_external_sample_max_rows_per_round`
+
+- Default: 2000000
+- Type: Long
+- Unit: rows
+- Is mutable: Yes
+- Description: Maximum rows targeted per sampling round in the Iceberg external-table statistics collection loop (`ANALYZE SAMPLE`). The first round aims for this many rows; subsequent rounds double the cap within the same run until either the full table is scanned or `statistic_external_sample_max_rounds` is reached. Larger values improve NDV accuracy at the cost of additional scan IO per round.
+- Introduced in: -
+
+### `statistic_external_sample_max_rounds`
+
+- Default: 4
+- Type: Int
+- Unit: -
+- Is mutable: Yes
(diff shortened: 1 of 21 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

PR #75889, opened 2026-07-06
[BugFix] Fix CN crash in NativeLookUpTask when late materialization yields no rows

luohaha:
## Why I'm doing:

Under concurrent realtime ingest + query on a **primary-key (lake) table**, the global-lazy-materialization (GLM) lookup path can crash the CN with a **SIGSEGV**, taking down every CN of the warehouse (queries then fail with `Warehouse ... is not available`).

Reproduced on a shared-data cluster (build `main-f3900ce`) running a skewed 300 GB PK-table benchmark. `cn.out` (identical on all CNs):
```
PC: starrocks::pipeline::NativeLookUpTask::_late_materialize_by_row_locators(...)
*** SIGSEGV (@0x18) received ... ***
    NativeLookUpTask::_late_materialize_by_row_locators
    NativeLookUpTask::process
    LookUpProcessor::process
    LookUpOperator::_try_to_trigger_io_task
    ScanExecutor::worker_thread
```
`addr2line` on the crashing PC resolves to `std::__shared_ptr<starrocks::Schema>::get()` — a null `SchemaPtr` deref; fault address `0x18` is the `Chunk::_schema` member offset.

**Root cause.** In `NativeLookUpTask::_late_materialize_by_row_locators`:
```cpp
accumulator.finalize();
auto accumulated = accumulator.pull();     // nullptr when nothing was materialized
for (const auto& slot : slots) {
    size_t column_index = accumulated->schema()->get_field_index_by_name(...);  // null deref -> SIGSEGV @0x18
```
`ChunkAccumulator::pull()` returns `nullptr` when its output is empty, and the result was dereferenced unconditionally.

The empty accumulator arises because the GLM lookup is **two-phase and not atomic**: a position scan produces row locators (`rssid`s captured at version V), and a later FETCH materializes them. Concurrent ingest (upserts/deletes/inserts) keeps changing the tablet between the two phases, so a fetch batch can resolve to zero rows (and, on the remote-fetch path, surfaces as `not found lake rssid`). This is a **general GLM bug** — the same `EXPLAIN` plan (`MERGING-EXCHANGE` -> `FETCH lookup node` fed by a `TOP-N` position scan) is generated for both HASH- and RANGE-distributed PK tables.

Note: reproduced **with lake compaction disabled** (`lake_compaction_max_tasks=0`), so compaction is only an accelerant, not the cause; concurrent ingest alone triggers it. A layout that spreads a hot, heavily-written key across many tablets/nodes (e.g. range distribution under data skew) reproduces it fastest.

## What I'm doing:

Guard the null result from `pull()`, distinguishing two cases so rows are never silently dropped:

1. **Empty lookup request (no row locators).** `_build_row_id_range` returns an empty locator set only when its input has 0 rows (for >=1 row it always flushes at least one locator). Return an empty chunk carrying the requested slots — the correct result.
2. **Non-empty locators that materialized nothing.** Under a version-consistent read a located row must be readable, so producing nothing means the tablet changed under us. Return an error so the query retries on a consistent snapshot rather than silently returning fewer rows. (A missing rowset already surfaces as an error earlier via `LakeScanTabletAdaptor::get_iterator` -> `not found lake rssid`, so this keeps the same loud-failure contract.)

Scope note: this PR fixes only the **crash**. The separate `not found lake rssid` failure (a GLM two-phase snapshot-consistency gap under concurrent writes) is a larger, independent issue and is intentionally out of scope.

Added a regression UT (`NativeLookUpTaskTest.EmptyRowLocatorsReturnsEmptyChunk`) that drives the empty-locator path — the exact `pull() == nullptr` condition that crashed — and asserts an empty, correctly-shaped chunk is returned.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

(It replaces an undefined crash with the intended result: an empty chunk for an empty lookup, or a proper retryable error.)

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

Main-only: the affected code path (`NativeLookUpTask::_late_materialize_by_row_locators` + `ChunkAccumulator`) does not exist on branch-4.1/4.0/3.5, so no backport is needed.

## Test note

Reviewed against the existing `glm_manager_tablet_adaptor_test` harness conventions. I could not run `run-be-ut.sh` in my environment (no `thirdparty/installed`); please rely on CI for the BE UT run.

### DIFF
--- a/be/src/exec/pipeline/lookup_request.cpp
+++ b/be/src/exec/pipeline/lookup_request.cpp
@@ -37,6 +37,7 @@
 #include "exprs/expr_executor.h"
 #include "exprs/expr_factory.h"
 #include "runtime/chunk_accumulator.h"
+#include "runtime/chunk_helper.h"
 #include "runtime/descriptors.h"
 #include "runtime/runtime_state.h"
 #include "storage_primitive/range.h"
@@ -867,6 +868,28 @@ auto NativeLookUpTask::_late_materialize_by_row_locators(RuntimeState* state, co
 
     accumulator.finalize();
     auto accumulated = accumulator.pull();
+    if (accumulated == nullptr) {
+        // pull() returns nullptr when the accumulator produced no chunk, i.e. nothing was
+        // materialized. Previously this null was dereferenced unconditionally below
+        // (accumulated->schema()), which crashes the CN with SIGSEGV.
+        //
+        // There are two ways to get here:
+        //  - There were no row locators to look up (empty lookup request). _build_row_id_range
+        //    returns an empty locator set only when its input has 0 rows, so an empty chunk with
+        //    the requested slots is the correct result.
+        //  - Row locators existed but none could be materialized. Under a version-consistent read
+        //    a located row must be readable, so this indicates the tablet changed under us between
+        //    the position scan and this fetch (concurrent ingest/updates/deletes, or
+        //    compaction/GC). Fail loudly so the query retries on a consistent snapshot rather than
+        //    silently returning fewer rows. (A missing rowset already surfaces as an error earlier
+        //    via _tablet_adaptor->get_iterator -> "not found lake rssid".)
+        if (!row_locators.empty()) {
+            return Status::InternalError(
+                    "late materialization produced no rows for a non-empty set of row locators; "
+                    "the tablet changed between the position scan and the lookup fetch");
+        }
+        return ChunkPtr(RuntimeChunkHelper::new_chunk(slots, 0));
+    }
 
     for (const auto& slot : slots) {
         size_t column_index = accumulated->schema()->get_field_index_by_name(slot->col_name());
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

stdpain:
This might not be the root cause. Can you write a stable test case to reproduce the problem?

luohaha:
Root-cause analysis of the underlying PK-index/rowset inconsistency (why the empty materialization happens) is filed as #75993. This PR is the defensive null-guard for the crash; #75993 covers the actual repair (reconciling PK-index positions with the split child's rowset set).

CelerData-Reviewer:
@codex review

PR #75891, opened 2026-07-06
[BugFix] Make flat_json config reliable on the shared-data write path

sevev:
## Why I'm doing:

On shared-data clusters the table property `flat_json.*` was unreliable on the **write path**: the lake segment writer read `FlatJsonConfig` only from the in-memory tablet-metadata cache (`get_latest_cached_tablet_metadata`) and, on a miss, silently fell back to the global `config::enable_json_flat` (default `true`), ignoring the table property. With file-bundling on (the default) a brand-new table's first load hit this **deterministically**, so `flat_json.enable` and the sparsity/null/column factors were effectively ignored for that data.

## What I'm doing:

Fix the shared-data **write and compaction consumption paths** so the flatten decision honors the FE-authoritative table config instead of a best-effort cache.

- **Carry `FlatJsonConfig` in the load plan** (`TOlapTableSink` / `PTabletWriterOpenRequest`). The lake segment writer uses this FE-authoritative value and only falls back to the metadata cache / global config when it is absent (property-less table or rolling upgrade). Spill-merged loads build their final segments through cloned writers (`LoadSpillPipelineMergeIterator` → `TabletWriter::clone()`), so the clone propagates the config too.
- **Version-consistent compaction.** Compaction and version-specific rewrites now take flat_json from the exact metadata version being read (`VersionedTablet::new_writer_with_schema` uses the `_metadata` it already holds) instead of the best-effort latest cache, so the decision is version-consistent and independent of cache warmth.
- **Move `TFlatJsonConfig` into `Types.thrift`** so it can be shared by the load sink and the agent/create-tablet paths without an include cycle. The new sink/plan fields are optional and wire-compatible.

### Scope / relationship to #74747

Propagating an `ALTER TABLE ... SET("flat_json.*"=...)` change to the BE is a **separate problem** handled by #74747 (which covers both shared-nothing and shared-data). This PR **does not touch the ALTER path** — it focuses solely on the write-path/consumption defect that #74747 does not address (cache-miss fallback to the global config). The two PRs are complementary.

### Tests

- `AlterTabletMetaTest.test_compaction_writer_uses_versioned_flat_json_config` — a writer created off a specific metadata version (compaction / version-specific rewrite) picks up flat_json from *that* version's metadata, and is left null when the version has none.
- `AlterTabletMetaTest.test_clone_writer_propagates_flat_json_config` — `TabletWriter::clone()` (spill-merge path) inherits the plan-carried flat_json config, so spilled loads do not silently fall back to the cache/global config.

Both pass under ASAN (`AlterTabletMetaTest` 11/0). End-to-end on a shared-data cluster: a new table created with an explicit `flat_json.*` property has that property honored on its first load (with file-bundling on), instead of silently using the global default on a metadata-cache miss.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Rationale:
- **Policy change**: shared-data loads/compactions now honor the table's `flat_json.*` property for the flatten decision instead of silently using the global default on a metadata-cache miss.
- **Miscellaneous (compatibility)**: `TFlatJsonConfig` moved to `Types.thrift` and new optional `TOlapTableSink` / `PTabletWriterOpenRequest` fields. All additive/optional; mixed-version clusters fall back to prior behavior for property-less tables and during rolling upgrade.

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/storage/lake/general_tablet_writer.cpp
+++ b/be/src/storage/lake/general_tablet_writer.cpp
@@ -229,6 +229,12 @@ StatusOr<std::unique_ptr<TabletWriter>> HorizontalGeneralTabletWriter::clone() c
                                                             _flush_pool, _bundle_file_context, _global_dicts);
     RETURN_IF_ERROR(writer->open());
     writer->set_auto_flush(auto_flush());
+    // Propagate the plan-carried Flat JSON policy so spill-merged segments (built through cloned
+    // writers, see LoadSpillPipelineMergeIterator) keep using the FE-authoritative value instead of
+    // falling back to the best-effort tablet-metadata cache.
+    if (flat_json_config() != nullptr) {
+        writer->set_flat_json_config(flat_json_config());
+    }
     // Propagate the force-inline flag: spilled sorted schema changes merge through cloned writers
     // (LoadSpillPipelineMergeIterator). Without this, clones default to false and defer .vi building
     // even though the schema-change job stamped the shadow tablets' vibv as already built, so the
@@ -246,8 +252,13 @@ Status HorizontalGeneralTabletWriter::reset_segment_writer(bool eos) {
     opts.is_compaction = _is_compaction;
     opts.vector_index_build_threshold = get_vector_index_build_threshold(_schema);
 
-    if (auto metadata = _tablet_mgr->get_latest_cached_tablet_metadata(_tablet_id);
-        metadata && metadata->has_flat_json_config()) {
+    if (flat_json_config() != nullptr) {
+        // Plan-carried, FE-authoritative Flat JSON policy (shared-data). Preferred over the
+        // best-effort tablet-metadata cache so the flatten decision does not depend on cache
+        // warmth and honors the current table property (including after ALTER, via the plan).
+        opts.flat_json_config = flat_json_config();
+    } else if (auto metadata = _tablet_mgr->get_latest_cached_tablet_metadata(_tablet_id);
+               metadata && metadata->has_flat_json_config()) {
         opts.flat_json_config = std::make_shared<FlatJsonConfig>();
         opts.flat_json_config->update(metadata->flat_json_config());
     }
@@ -566,8 +577,13 @@ StatusOr<std::shared_ptr<SegmentWriter>> VerticalGeneralTabletWriter::create_seg
     opts.is_compaction = _is_compaction;
     opts.vector_index_build_threshold = get_vector_index_build_threshold(_schema);
 
-    if (auto metadata = _tablet_mgr->get_latest_cached_tablet_metadata(_tablet_id);
-        metadata && metadata->has_flat_json_config()) {
+    if (flat_json_config() != nullptr) {
+        // Plan-carried, FE-authoritative Flat JSON policy (shared-data). Preferred over the
+        // best-effort tablet-metadata cache so the flatten decision does not depend on cache
+        // warmth and honors the current table property (including after ALTER, via the plan).
+        opts.flat_json_config = flat_json_config();
+    } else if (auto metadata = _tablet_mgr->get_latest_cached_tablet_metadata(_tablet_id);
+               metadata && metadata->has_flat_json_config()) {
         opts.flat_json_config = std::make_shared<FlatJsonConfig>();
         opts.flat_json_config->update(metadata->flat_json_config());
     }
--- a/be/src/storage/lake/versioned_tablet.cpp
+++ b/be/src/storage/lake/versioned_tablet.cpp
@@ -20,6 +20,7 @@
 #include "storage/lake/tablet_writer.h"
 #include "storage/lake/update_manager.h"
 #include "storage/tablet_schema_map.h"
+#include "storage_primitive/flat_json_config.h"
 #include "storage_primitive/tablet_basic_info.h"
 
 namespace starrocks::lake {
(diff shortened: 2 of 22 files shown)


CelerData-Reviewer:
@codex review

kevincai:
Could be similar in this PR: #74747

CelerData-Reviewer:
@codex review

be/src/storage/lake/versioned_tablet.cpp
```diff
@@ -45,25 +46,36 @@ StatusOr<std::unique_ptr<TabletWriter>> VersionedTablet::new_writer(WriterType t
...
-                                                                 max_rows_per_segment, is_compaction, flush_pool);
+            writer = std::make_unique<VerticalGeneralTabletWriter>(_tablet_mgr, id(), tablet_schema, txn_id,
+                                                                   max_rows_per_segment, is_compaction, flush_pool);
         }
     }
+    // Flat JSON policy for writers created off a specific metadata version (compaction / internal
+    // rewrites): take it from THIS version's metadata (_metadata), which we already hold, rather than
+    // the best-effort latest-metadata cache. This makes the flatten decision version-consistent (it
+    // uses the config of the exact version being read/compacted) and independent of cache warmth.
+    if (_metadata != nullptr && _metadata->has_flat_json_config()) {
```
kevincai, line 73: This makes the compaction **writer** version-consistent, but the compaction **reader** was not updated to match. `TabletReader::init_compaction_column_paths` still derives the flat-JSON column paths from `get_latest_cached_tablet_metadata(...)` even though the reader already holds the exact version in `_tablet_metadata`:

https://github.com/StarRocks/starrocks/blob/15239df45a3f77bfc7e5a8545d4ad6922d1dc1a9/be/src/storage/lake/tablet_reader.cpp#L272-L280

Before this PR both sides read the cache, so they agreed. Now the writer uses the version's config while the reader still uses the cache. On a cold-cache miss -- the exact scenario this PR targets -- `get_latest_cached_tablet_metadata` returns null, so the path deriver falls back to default flatten factors while the writer uses the version's factors, and the two halves of the same compaction run disagree. `_tablet_metadata` is already in scope in the reader; using it (as the writer now does) would keep both sides consistent.

_(Note: `tablet_reader.cpp` is outside this PR's diff, so this is anchored to the writer-side change that introduces the asymmetry.)_

be/src/storage/lake/general_tablet_writer.cpp
```diff
@@ -246,8 +252,13 @@ Status HorizontalGeneralTabletWriter::reset_segment_writer(bool eos) {
...
     opts.vector_index_build_threshold = get_vector_index_build_threshold(_schema);
 
-    if (auto metadata = _tablet_mgr->get_latest_cached_tablet_metadata(_tablet_id);
-        metadata && metadata->has_flat_json_config()) {
+    if (flat_json_config() != nullptr) {
+        // Plan-carried, FE-authoritative Flat JSON policy (shared-data). Preferred over the
+        // best-effort tablet-metadata cache so the flatten decision does not depend on cache
+        // warmth and honors the current table property (including after ALTER, via the plan).
+        opts.flat_json_config = flat_json_config();
+    } else if (auto metadata = _tablet_mgr->get_latest_cached_tablet_metadata(_tablet_id);
```
kevincai, line 260: This path now correctly prefers the plan-carried config and only falls back to the cache. The column-mode partial-update write path does not get the same treatment: its delta-column-group writer still reads flat-JSON config only from `get_latest_cached_tablet_metadata(...)`, with no plan/version fallback, even though the version-specific `params.metadata` is in scope (it is used just below for the DCG loader):

https://github.com/StarRocks/starrocks/blob/15239df45a3f77bfc7e5a8545d4ad6922d1dc1a9/be/src/storage/lake/column_mode_partial_update_handler.cpp#L142-L147

That leaves the same cache-reliability defect this PR fixes elsewhere on the column-mode load path: a cold-cache miss silently drops the table's `flat_json` policy for delta-column writes. Consider deriving it from `params.metadata->flat_json_config()` there.

_(Note: `column_mode_partial_update_handler.cpp` is outside this PR's diff, so this is anchored to the `reset_segment_writer` plan-vs-cache logic it should mirror.)_

be/src/storage/lake/versioned_tablet.cpp
sevev: Good catch — you're right this introduced a writer/reader asymmetry. Fixed in b2fcf4c: `TabletReader::init_compaction_column_paths` now derives the flat-JSON config from `_tablet_metadata` (the exact version already in scope) instead of `get_latest_cached_tablet_metadata(...)`, mirroring the writer side (`VersionedTablet::new_writer_with_schema`). So on a cold-cache miss both halves of the same compaction now use the same version's factors.

be/src/storage/lake/general_tablet_writer.cpp
sevev: Done in b2fcf4c: `ColumnModePartialUpdateHandler::_prepare_delta_column_group_writer` now derives flat_json from `params.metadata->flat_json_config()` (the version-specific metadata already in scope, used just below for the DCG loader) instead of the best-effort cache, so a cold-cache miss no longer silently drops the table's flat_json policy on the column-mode delta-column write path.

be/src/storage/lake/versioned_tablet.cpp
sevev: > This makes the compaction **writer** version-consistent, but the compaction **reader** was not updated to match. `TabletReader::init_compaction_column_paths` still derives the flat-JSON column paths from `get_latest_cached_tablet_metadata(...)` even though the reader already holds the exact version in `_tablet_metadata`:
> 
> https://github.com/StarRocks/starrocks/blob/15239df45a3f77bfc7e5a8545d4ad6922d1dc1a9/be/src/storage/lake/tablet_reader.cpp#L272-L280
> 
> Before this PR both sides read the cache, so they agreed. Now the writer uses the version's config while the reader still uses the cache. On a cold-cache miss -- the exact scenario this PR targets -- `get_latest_cached_tablet_metadata` returns null, so the path deriver falls back to default flatten factors while the writer uses the version's factors, and the two halves of the same compaction run disagree. `_tablet_metadata` is already in scope in the reader; using it (as the writer now does) would keep both sides consistent.
> 
> _(Note: `tablet_reader.cpp` is outside this PR's diff, so this is anchored to the writer-side change that introduces the asymmetry.)_

updated.

be/src/storage/lake/general_tablet_writer.cpp
sevev: > This path now correctly prefers the plan-carried config and only falls back to the cache. The column-mode partial-update write path does not get the same treatment: its delta-column-group writer still reads flat-JSON config only from `get_latest_cached_tablet_metadata(...)`, with no plan/version fallback, even though the version-specific `params.metadata` is in scope (it is used just below for the DCG loader):
> 
> https://github.com/StarRocks/starrocks/blob/15239df45a3f77bfc7e5a8545d4ad6922d1dc1a9/be/src/storage/lake/column_mode_partial_update_handler.cpp#L142-L147
> 
> That leaves the same cache-reliability defect this PR fixes elsewhere on the column-mode load path: a cold-cache miss silently drops the table's `flat_json` policy for delta-column writes. Consider deriving it from `params.metadata->flat_json_config()` there.
> 
> _(Note: `column_mode_partial_update_handler.cpp` is outside this PR's diff, so this is anchored to the `reset_segment_writer` plan-vs-cache logic it should mirror.)_

updated.

kevincai:
@sevev does the `Schema Compatility` check hold or not?

PR #75900, opened 2026-07-06
[Tool] harden pr-checker checkbox parsing and fix stale-payload label race

huailiu1122:
## Why I'm doing:
Two related weaknesses in `pr-checker.yml` bit real PRs this week:

1. **Exact-literal checkbox matching.** behavior-check compares the body against exact strings via `contains()`, so an extra space after `[ ]` fails the format check no matter what the author ticks. #75888 / #75825 were blocked this way (`- [ ]  No, ...` with two spaces — checking the box in the UI keeps the double space, so the error is unrecoverable from the UI), and the error message gives no hint about what is wrong.
2. **Stale-payload label race.** Version labels are driven by the frozen `edited`-event payload under `cancel-in-progress: true`. Ticking checkboxes one by one fires a burst of events whose runs can start out of order; the surviving run may hold a stale body snapshot and *remove* freshly added version labels. #75861 lost its `4.1/4.0/3.5` labels this way (timeline: `4.1` labeled 05:00:30 → unlabeled 05:03:04 by the stale run) and merged with no labels, so auto-backport never triggered.

## What I'm doing:
- Parse the **live** PR body (`gh api .../pulls/N`) instead of the event payload in `behavior-check` and `backport-check-subtask`, eliminating the stale-snapshot race: whichever run survives cancellation now reads the current state.
- Replace exact-literal `contains()` with whitespace/case-tolerant line regexes (`- [x]` / `- [X]` / extra spaces all accepted; matching is line-anchored so prose mentions no longer count).
- Emit a specific error per failure mode (template line missing / both checked / neither checked / change type missing) instead of one generic "check the format" message.
- Keep job/matrix names unchanged so any required-check references stay valid.

**Validation (replay against 380 recent real PRs, running the actual new bash logic):**
- behavior-check: 379/380 identical outcomes. The single divergence is #75825 — checked Yes + double-space No line; old logic wrongly blocked it, new logic passes it. Zero previously-passing PRs regress.
- version-label add/remove decisions: 570/570 identical (114 eligible PRs × 5 versions).
- Synthetic edge cases (double space, uppercase X, CRLF, both/neither checked, template deleted, Yes without type, `4.1` vs `4.10`) all classified correctly.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [x] Tool

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/.github/workflows/pr-checker.yml
+++ b/.github/workflows/pr-checker.yml
@@ -145,7 +145,9 @@ jobs:
           VERSION: ${{ matrix.version }}
         run: |
           body=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}" --jq '.body // ""')
-          has_label=$(gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/labels" --jq "[.[].name] | contains([\"${VERSION}\"])")
+          # any(. == $v), not contains(): jq array containment matches strings
+          # by substring, so labels like "4.1-merged" or "4.1.0" would count as "4.1"
+          has_label=$(gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/labels" --jq "any(.[].name; . == \"${VERSION}\")")
           version_re=${VERSION//./\\.}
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #75904, opened 2026-07-06
[BugFix] Read durable metadata in cal_new_base_version to avoid dangling prev_garbage_version

starrocks-xupeng:
## Why I'm doing:

For a Primary Key table with file bundling, aggregate publish caches each tablet's metadata (`skip_write_tablet_metadata`) and writes only one durable bundle at the batch's final version; the intermediate per-tablet versions are cache-only. When a batch publish is **retried** after a prior attempt advanced the primary index but before the durable bundle was written, `cal_new_base_version` probes remote storage for the primary-index version via `get_tablet_metadata` — which, for an aggregation partition, hits the metacache first and returns OK for a version that has **no durable bundle**. It then adopts that non-durable version as the publish base and records it as `prev_garbage_version`.

Once the metacache is evicted (or the BE restarts) that reference dangles: `get_tablet_metadata` now returns NotFound, the tablet-metadata vacuum walk (which descends `prev_garbage_version`) breaks at that point, and every older version below it is left permanently unreclaimed. This exists on main.

## What I'm doing:

- **`storage/lake/options.h`**: add `CacheOptions.skip_meta_cache` (default `false`). When true it bypasses the tablet-metadata metacache lookup and reads straight from durable storage, letting a caller tell "durably persisted in remote storage" apart from "only in the metacache".
- **`storage/lake/tablet_manager.cpp`**: honor `skip_meta_cache` in both metacache lookups on the read path — `get_tablet_metadata(path, ...)` and `get_single_tablet_metadata`. The durable bundle read path is unchanged, so a version that only exists as a cache entry now reads as NotFound.
- **`storage/lake/transactions.cpp`**: `cal_new_base_version` reads the index version with `CacheOptions{.skip_meta_cache = true}` (the `fill_meta_cache` / `fill_data_cache` flags stay `true`, so a durable read still warms the cache as before). A cache-only version now returns NotFound, so the primary index is rebuilt from the durable base instead of adopting a non-durable version — `prev_garbage_version` only ever references a durably persisted version.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/transactions.cpp
+++ b/be/src/storage/lake/transactions.cpp
@@ -24,6 +24,7 @@
 #include "gutil/strings/join.h"
 #include "storage/lake/filenames.h"
 #include "storage/lake/metacache.h"
+#include "storage/lake/options.h"
 #include "storage/lake/replication_txn_manager.h"
 #include "storage/lake/tablet.h"
 #include "storage/lake/tablet_manager.h"
@@ -96,8 +97,16 @@ int64_t cal_new_base_version(int64_t tablet_id, TabletManager* tablet_mgr, int64
     if (index_version > version) {
         // There is a possibility that the index version is newer than the version in remote storage.
         // Check whether the index version exists in remote storage. If not, clear and rebuild the index.
+        // skip_meta_cache forces this to read DURABLE storage only: with file bundling a version can sit in
+        // the metacache (its metadata cached during publish) without a durable bundle written yet -- e.g. a
+        // batch publish advanced the primary index and cached the metadata but had not written the bundle,
+        // and is now being retried. A plain cache-hitting read would treat such a version as "exists in
+        // remote", adopt it as the base, and record it as prev_garbage_version, leaving a dangling reference
+        // (NotFound while vacuum walks the chain) once the cache is evicted.
         auto gtid = txns.size() == 1 ? txns[0].gtid() : txns[index_version - base_version - 1].gtid();
-        auto res = tablet_mgr->get_tablet_metadata(tablet_id, index_version, true, gtid);
+        auto res = tablet_mgr->get_tablet_metadata(
+                tablet_id, index_version,
+                CacheOptions{.fill_meta_cache = true, .fill_data_cache = true, .skip_meta_cache = true}, gtid);
         if (res.ok()) {
             version = index_version;
         } else {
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/lake/transactions.cpp
```diff
@@ -95,8 +96,16 @@ int64_t cal_new_base_version(int64_t tablet_id, TabletManager* tablet_mgr, int64
     if (index_version > version) {
         // There is a possibility that the index version is newer than the version in remote storage.
         // Check whether the index version exists in remote storage. If not, clear and rebuild the index.
+        // skip_meta_cache forces this to read DURABLE storage only: with file bundling a version can sit in
```
kevincai, line 100: what's the impact of this? What code path will touch this `cal_new_base_version()` call. If the file bundling version cache before publish, can it be reordered to `cache-after-publish` so it won't break the rule that if found from local cache, it must be persisted to remote.

luohaha (dismissed):

CelerData-Reviewer:
@codex review

be/src/storage/lake/transactions.cpp
starrocks-xupeng: data & meta under this version will be leaked forever for bundling table

no ,because each meta is written into cache independently for each tablet, then publish will only write 1 bundled meta to S3

be/src/storage/lake/transactions.cpp
```diff
@@ -96,8 +97,16 @@ int64_t cal_new_base_version(int64_t tablet_id, TabletManager* tablet_mgr, int64
     if (index_version > version) {
```
xiangguangyxg, line 97: Under what circumstances will index_version > version occur, and why would this happen? Are there any more fundamental issues that haven't been resolved?

starrocks-xupeng: a transaction publishes but failed, some tablets might already wrote tablet meta into metacache, when it retries, it might be grouped into a batch publish, so in this case, index_version will be higher than base version, because index_version is from cached tablet meta

luohaha (approved):

xiangguangyxg (approved):

kevincai (approved):

PR #75909, opened 2026-07-06
[BugFix] Truncate CAST(x AS CHAR(N)) to N characters (MySQL-compatible)

HangyuanLiu:
## Why I'm doing:

`CAST(x AS CHAR(N))` did not truncate its result to N characters, unlike MySQL. Amazon QuickSight SPICE incremental refresh emits MySQL-style SQL such as `cast(cast(1775580223839 as char(10)) as decimal)`, expecting the 13-digit millisecond epoch to be truncated to a 10-digit second epoch (`1775580223`). StarRocks returned the full `1775580223839`, which then overflowed downstream `FROM_UNIXTIME` / `DATE_ADD` and broke the incremental-refresh filter. (Jira POST-1335)

## What I'm doing:

Make `CAST(x AS CHAR(N))` truncate to the first N characters (MySQL-compatible), consistently on both the FE constant-folding path and the BE execution path. Scoped to **bounded `CHAR(N)` only** — `VARCHAR(N)` and unbounded `CHAR` are left unchanged, because StarRocks uses `VARCHAR(N)` pervasively for implicit casts / type unification and MySQL has no `VARCHAR` cast target, so truncating it would risk silent data corruption in existing plans.

- **FE constant folding** (`ConstantOperator.castTo`): truncate to N code points for bounded `CHAR(N)`.
- **FE cast reduction** (`ReduceCastRule`): stop eliminating `CAST(string AS CHAR(N))` when the source string may exceed N. `matchesType()` treats all string types as identical, so the rule previously dropped the cast before `FoldConstantsRule` / the BE could truncate.
- **BE execution** (`VectorizedCastToStringExpr`): truncate the result to N UTF-8 characters for bounded `CHAR(N)` across all source types (via `skip_leading_utf8`). The common `VARCHAR` path is unchanged and has zero added overhead.

Examples:

```
CAST(1775580223839 AS CHAR(10))                    -> 1775580223
CAST(CAST(1775580223839 AS CHAR(10)) AS DECIMAL)   -> 1775580223
CAST('hello world' AS CHAR(5))                     -> hello
CAST(1775580223839 AS VARCHAR(10))                 -> 1775580223839   (unchanged)
CAST('hello world' AS CHAR)                        -> hello world     (unchanged)
```

Jira: POST-1335

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exprs/cast_expr.cpp
+++ b/be/src/exprs/cast_expr.cpp
@@ -31,6 +31,7 @@
 #include <unordered_map>
 #include <utility>
 
+#include "base/string/utf8.h"
 #include "base/time/date_func.h"
 #include "base/types/int128.h"
 #include "base/types/numeric_types.h"
@@ -1519,6 +1520,16 @@ class VectorizedCastToStringExpr final : public Expr {
 public:
     DEFINE_CAST_CONSTRUCT(VectorizedCastToStringExpr);
     StatusOr<ColumnPtr> evaluate_checked(ExprContext* context, Chunk* ptr) override {
+        ASSIGN_OR_RETURN(ColumnPtr result, _evaluate_to_string(context, ptr));
+        // MySQL semantics: CAST(x AS CHAR(N)) truncates to N characters. Only bounded CHAR(N)
+        // (len >= 0) is affected; VARCHAR(N) and wildcard CHAR keep the full value.
+        if (type().type == TYPE_CHAR && type().len >= 0) {
+            return _truncate_to_char_len(std::move(result), type().len);
+        }
+        return result;
+    }
+
+    StatusOr<ColumnPtr> _evaluate_to_string(ExprContext* context, Chunk* ptr) {
         ASSIGN_OR_RETURN(ColumnPtr column, _children[0]->evaluate_checked(context, ptr));
         if (ColumnHelper::count_nulls(column) == column->size() && column->size() != 0) {
             return ColumnHelper::create_const_null_column(column->size());
@@ -1622,6 +1633,33 @@ private:
     // In SR, behaviors of both cast(string as varchar(n)) and cast(string as char(n)) keep the same: neglect
     // of the length of char/varchar and return input column directly.
     ColumnPtr _evaluate_string(ExprContext* context, ColumnPtr&& column) { return Column::mutate(std::move(column)); }
+
+    // Truncate each non-null value to its first `len` UTF-8 characters (MySQL CHAR(N) semantics).
+    // Nulls stay null; const-ness is preserved.
+    static ColumnPtr _truncate_to_char_len(ColumnPtr column, int len) {
+        if (column->only_null()) {
+            return column;
+        }
+        ColumnViewer<TYPE_VARCHAR> viewer(column);
+        ColumnBuilder<TYPE_VARCHAR> builder(viewer.size());
+        for (int row = 0; row < viewer.size(); ++row) {
+            if (viewer.is_null(row)) {
+                builder.append_null();
+                continue;
+            }
+            Slice s = viewer.value(row);
+            // Byte length is an upper bound on the character count: <= len bytes already fits.
+            if (s.size <= static_cast<size_t>(len)) {
+                builder.append(s);
+                continue;
+            }
+            const char* begin = s.data;
+            const char* end = s.data + s.size;
+            const char* truncated_end = skip_leading_utf8(begin, end, static_cast<size_t>(len));
+            builder.append(Slice(begin, truncated_end - begin));
+        }
+        return builder.build(column->is_constant());
+    }
 
     ColumnPtr _evaluate_time(ExprContext* context, const ColumnPtr& column) {
         ColumnViewer<TYPE_TIME> viewer(column);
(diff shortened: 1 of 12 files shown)


CelerData-Reviewer:
@codex review

be/src/exprs/cast_expr.cpp
```diff
@@ -1623,6 +1634,33 @@ class VectorizedCastToStringExpr final : public Expr {
...
+            }
+            Slice s = viewer.value(row);
+            // Byte length is an upper bound on the character count: <= len bytes already fits.
+            if (s.size <= static_cast<size_t>(len)) {
+                builder.append(s);
+                continue;
+            }
+            const char* begin = s.data;
+            const char* end = s.data + s.size;
+            const char* truncated_end = skip_leading_utf8(begin, end, static_cast<size_t>(len));
```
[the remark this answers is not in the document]
HangyuanLiu: Good catch — confirmed. `skip_leading_utf8` advances by the UTF-8 lead-byte width, so for a truncated multi-byte char, invalid UTF-8, or raw VARBINARY bytes it can return a pointer past `end`, and the subsequent `Slice(begin, truncated_end - begin)` would over-read.

Fixed in 462af3a by clamping `truncated_end` to `end` before appending, plus a regression test `charTruncateHandlesInvalidUtf8WithoutOverrun` (a 2-byte value whose leading `0xE4` advertises a 3-byte char, cast to `CHAR(1)`). Valid UTF-8 is unaffected since characters end on boundaries `<= end`.

Seaven (dismissed):

CelerData-Reviewer:
@codex review

stdpain (approved):

be/src/exprs/cast_expr_tpl.hpp
```diff
@@ -1521,6 +1522,16 @@ class VectorizedCastToStringExpr final : public Expr {
 public:
     DEFINE_CAST_CONSTRUCT(VectorizedCastToStringExpr);
     StatusOr<ColumnPtr> evaluate_checked(ExprContext* context, Chunk* ptr) override {
+        ASSIGN_OR_RETURN(ColumnPtr result, _evaluate_to_string(context, ptr));
+        // MySQL semantics: CAST(x AS CHAR(N)) truncates to N characters. Only bounded CHAR(N)
+        // (len >= 0) is affected; VARCHAR(N) and wildcard CHAR keep the full value.
+        if (type().type == TYPE_CHAR && type().len >= 0) {
```
stephen-shelby, line 1528: [blocking] This gate only checks the target type, so it also fires on implicit casts the planner generates for write paths (InsertPlanner/UpdatePlanner/load castToSlot add a cast to the column type whenever the source is non-string). Truncation happens before OlapTableSink's length check, so loads that used to fail now silently store truncated data.

Verified: CREATE TABLE t (c CHAR(10)); INSERT INTO t VALUES(1775580223839) now plans as constant exprs: '1775580223' — before this PR the full value reached the sink and the insert failed with "String (length=13) is too long". Note MySQL truncates only for explicit CAST; implicit assignment in strict mode raises "Data too long".

Suggestion: restrict truncation to explicit casts (CastOperator.isImplicit on FE; a flag in TExprNode for BE), and add an INSERT regression test.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/scalar/ReduceCastRule.java
```diff
@@ -170,6 +171,23 @@ public boolean checkCastTypeReduceAble(Type parent, Type child, Type grandChild)
...
+    // A cast to a bounded CHAR(N) truncates its input to the first N characters (MySQL semantics),
+    // so it is not a no-op and must not be reduced away.
+    //
+    // The source type's declared length is deliberately not consulted: it is not an upper bound on
+    // the runtime value. A cast to VARCHAR(N) does not enforce N (string-to-string cast returns the
+    // input column unchanged), so CAST(CAST(x AS VARCHAR(10)) AS CHAR(10)) carries a value that may
+    // exceed 10 characters even though the source type says otherwise. Deciding which producers do
+    // honour their declared length would be an open-ended audit, so every bounded CHAR(N) cast over
+    // a string source is kept and the truncation is left to constant folding or the BE.
+    private static boolean isTruncatingCharCast(Type target, Type source) {
```
stephen-shelby, line 183: blocking] Now that this cast survives, ListPartitionPruner sees it for the first time — and castLiteralExpr (~line 507) still maps partition values through it without truncating, so matching partitions get pruned and queries silently lose rows.

Verified with a UT: partitions {'hello world', 'hi'}, predicate cast(part_col AS CHAR(5)) = 'hello' → pruner.prune() returns [], i.e. the 'hello world' partition (every row of which matches under the new semantics) is pruned and the query returns 0 rows. EQ/IN/range paths are all affected, OLAP and external tables both.

Suggestion: truncate in castLiteralExpr (mirror truncateToCodePoints), or bail out of pruning when the cast target is a bounded CHAR.

PR #75916, opened 2026-07-06
[BugFix] Capture real rowsets for GLM late materialization on lake scans

luohaha:
## Why

On lake (shared-data) scans, `_morsel->rowsets()` is **empty by design**: the lake `TabletReader` uses the morsel's rowset list only when non-empty, and otherwise loads rowsets from the tablet metadata at the scan version (`be/src/storage/lake/tablet_reader.cpp`). The lake scan path wraps everything in the dynamic morsel queue, which never calls `set_rowsets()`, so the list is always empty.

But `LakeDataSource::init_tablet_reader` captured `_morsel->rowsets()` — the **empty** vector — into the global late-materialization (GLM) context. With a zero-rowset capture, `NativeLookUpTask::_late_materialize_by_row_locators` never uses the pinned snapshot and always falls back to **live** tablet metadata. Under concurrent ingest/compaction this races:

- rowsets removed by compaction/vacuum → position-scan rssids become unresolvable → `not found lake rssid` (query error);
- a locator resolves to a rowset whose row was deleted → `get_next` returns EOF → empty accumulator → `accumulated->schema()` null deref → **CN SIGSEGV** (`@0x18`).

An instrumented build confirmed `captured_n=0` in 100/100 `not found lake rssid` cases.

## What

In `LakeDataSource::init_tablet_reader`, capture the same rowsets the reader will actually read: use `_morsel->rowsets()` when non-empty, otherwise load them from the tablet metadata via `lake::Rowset::get_rowsets(...)` at the scan version. This pins a real snapshot so late materialization resolves locators against the version the scan committed to, immune to concurrent writes.

## Relationship to #75889

This is the **root-cause** fix. #75889 is a complementary **defensive guard** that turns the empty-accumulator case into a handled result instead of a crash. Both are worth having: this PR removes the cause; #75889 hardens the code path.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

The affected code path (`NativeLookUpTask::_late_materialize_by_row_locators` + `ChunkAccumulator` GLM) does not exist on branch-4.1/4.0/3.5, so this is **main-only — no backport**.

- [ ] 4.1
- [ ] 4.0
- [ ] 3.5
- [ ] 3.4

### DIFF
--- a/be/src/connector/lake_connector.cpp
+++ b/be/src/connector/lake_connector.cpp
@@ -49,6 +49,7 @@
 #include "storage/chunk_helper.h"
 #include "storage/column_predicate_rewriter.h"
 #include "storage/flat_json_metrics.h"
+#include "storage/lake/rowset.h"
 #include "storage/lake/table_schema_service.h"
 #include "storage/lake/tablet.h"
 #include "storage/predicate_parser.h"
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #75924, opened 2026-07-06
[BugFix] Fix MV rewrite serving stale results after Iceberg rollback_to_snapshot

### LINKED ISSUE #71977: [BugFix] MV query rewrite returns stale data after Iceberg base rollback when mv_rewrite_staleness_second is set
## Summary

When a materialized view's base is an Iceberg table and the user opts in to
`PROPERTIES("mv_rewrite_staleness_second" = "<N>")` (with `N > 0`), a
`rollback_to_snapshot` on the base makes the MV-rewrite path serve **stale
MV data** until the MV's next refresh tick completes.

Root cause: `MaterializedView.isStalenessSatisfied()` computes
`mvStaleness = (base_ts - mv_ts) / 1000` using
`IcebergPartitionTraits.maxPartitionRefreshTs()` which reads
`currentSnapshot().timestampMillis()`. After rollback the iceberg current
snapshot's timestamp is **older** than the MV's `lastRefreshTime`, producing
a **negative** `mvStaleness`. The check `if (mvStaleness > maxMVRewriteStaleness)`
never fails for negative values, so the shortcut returns `true` — the MV
is treated as "fresh" and picked for rewrite. In effect the MV is **ahead
of** (inconsistent with) the base rather than behind, but the code only
checks the signed time difference, not direction.

Auto refresh is not affected (refresh path passes `isQueryRewrite=false`,
shortcut is double-gated by `isQueryRewrite && isStalenessSatisfied()`), so
the MV eventually converges on the next refresh tick; the damage is
localized to the rewrite window. The MV is not marked inactive (FE has no
notification for iceberg rollback).

## Environment

- StarRocks main (observed on feature branch `codex/mv-on-iceberg`)
- Iceberg external catalog (hadoop; same logic applies to hive / REST)
- Applicable to any MV that explicitly sets `mv_rewrite_staleness_second > 0`

## Reproduction

SQL case (commit `3b9b677d77e` on `codex/mv-on-iceberg`,
`test/sql/test_mv_on_iceberg/T/test_mv_iceberg_staleness_shortcut_rowlevel`):

```sql
-- On an Iceberg catalog:
CREATE TABLE t1 (id int, dt date, val int) PARTITION BY (dt);
INSERT INTO t1 VALUES (1, '2024-01-01', 100);                -- snap_a
-- capture snap_a id (e.g. from v2.metadata.json / $snapshots)
INSERT INTO t1 VALUES (2, '2024-01-01', 200),
                      (3, '2024-01-01', 300);                -- snap_b (same partition)

-- On default_catalog:
CREATE MATERIALIZED VIEW mv_sp
PARTITION BY dt
REFRESH DEFERRED MANUAL
PROPERTIES (
  "replication_num" = "1",
  "mv_rewrite_staleness_second" = "3600"   -- opt-in
)
AS SELECT dt, sum(val) AS sv, count(*) AS c
   FROM iceberg_cat.db1.t1
   GROUP BY dt;

REFRESH MATERIALIZED VIEW mv_sp WITH SYNC MODE;   -- MV holds snap_b: sv=600, c=3

-- Roll base back:
ALTER TABLE iceberg_cat.db1.t1
  EXECUTE rollback_to_snapshot(snapshot_id = <snap_a_id>);
REFRESH EXTERNAL TABLE iceberg_cat.db1.t1;

-- Now base is {(1, 100)} again but MV still physically holds (600, 3).
-- Partition '2024-01-01' exists on BOTH sides, so partition pruning is a
-- no-op.

SET enable_materialized_view_rewrite = false;
SELECT dt, sum(val), count(*) FROM iceberg_cat.db1.t1 GROUP BY dt;
-- ► 2024-01-01, 100, 1       (base truth)

SET enable_materialized_view_rewrite = true;
SELECT dt, sum(val), count(*) FROM iceberg_cat.db1.t1 GROUP BY dt;
-- ► 2024-01-01, 600, 3       (stale MV returned — bug)
```

`EXPLAIN` of the rewrite query:

```
  0:OlapScanNode
     TABLE: mv_sp
     partitions=1/1
     rollup: mv_sp
     MaterializedView: true
```

Rewrite hits MV directly, no compensation / union.

## Expected behavior

One of the following:
1. Rewrite misses the MV (MV's fingerprint differs from base's), query falls
   back to base scan and returns `100, 1`; or
2. The MV is marked inactive with a clear `INACTIVE_REASON` indicating base
   snapshot regressed.

Either way, the rewrite path must not return a value derived from a base
snapshot that is no longer the current one.

## Actual behavior

- Rewrite **hits** the MV and returns stale aggregation (`600, 3` vs truth
  `100, 1`, off by 6×).
- `IS_ACTIVE = true`, `INACTIVE_REASON` is empty — FE has no signal that
  the base regressed.
- Next auto refresh tick (e.g. `REFRESH ASYNC EVERY(INTERVAL 1 MINUTE)`)
  silently fixes MV contents (refresh path not gated by the shortcut).
  Verified: MV converges from `(600, 3)` back to `(100, 1)` after one tick.
- `REFRESH ... WITH SYNC MODE` also fixes MV contents.
- For `REFRESH DEFERRED MANUAL`, window is unbounded until user
  triggers SYNC.

## Precise trigger conditions (all must hold)

1. MV property `mv_rewrite_staleness_second > 0` (default `-1`, so **default
   MV is not affected**).
2. Base is Iceberg (or any connector with non-monotonic snapshot timestamps).
3. Base undergoes rollback / time-travel that moves the current snapshot to
   an older one.
4. Query hits the rewrite path while the MV has not yet been re-refreshed.

## Code pointers

- `fe/fe-core/.../connector/partitiontraits/IcebergPartitionTraits.java:72`
  — returns `currentSnapshot().timestampMillis()`, non-monotonic.
- `fe/fe-core/.../catalog/MaterializedView.java:1091-1112`
  — `maxBaseTableRefreshTimestamp()` aggregates per-base `maxPartitionRefreshTs`.
- `fe/fe-core/.../catalog/MaterializedView.java:1134-1163`
  — `isStalenessSatisfied()` computes `mvStaleness = (base_ts - mv_ts)/1000`;
  negative values silently pass `mvStaleness > maxMVRewriteStaleness`.
- `fe/fe-core/.../catalog/MaterializedView.java:1078, 1192`
  — shortcut gate `isQueryRewrite && isStalenessSatisfied()`: fires only on
  rewrite path, not on refresh.

### Minimal

Treat any negative `mvStaleness` (i.e. base has regressed relative to MV)
as non-fresh:

```java
long mvStaleness = (baseTableRefreshTimestamp - mvRefreshTimestamp) / 1000;
if (mvStaleness < 0) {
    LOG.warn("MV {} is ahead of base (base_ts regressed); treat as stale.", getName());
    return false;
}
if (mvStaleness > this.maxMVRewriteStaleness) {
    return false;
}
return true;
```

### Preferred

Use `snapshotId` as the identity/fingerprint for Iceberg freshness instead
of the non-monotonic timestamp. The MV records the snapshot id it
refreshed from; the rewrite path compares the id against
`icebergTable.currentSnapshot().snapshotId()`. Any difference (forward or
backward) is non-fresh. Additionally emit `INACTIVE_REASON` if the
fingerprint indicates regression, closing the observability gap where
DBAs have no way to be alerted about an iceberg rollback.

### Out of scope for this fix

Cross-engine (Spark / Flink / Trino) row-level DML + rollback is a
related scenario but depends on Iceberg v2 reader merge-on-read which is
tracked separately.

## Tests / evidence

- `codex/mv-on-iceberg` branch commits:
  - `3b9b677d77e` — same-partition rollback variant (this bug, 6× stale)
  - `2b3945f2c15` — cross-partition rollback variant (pruning rescues correctness but shortcut still fires)
- R files lock current (buggy) behavior; any fix will flip them.

## Regression checklist

- [ ] Default MV (no `mv_rewrite_staleness_second`) after rollback — still
      returns correct result (unchanged).
- [ ] Opt-in MV after rollback, rewrite query — returns truthful base
      aggregation (flipped).
- [ ] Auto refresh still converges MV contents on next tick (unchanged).
- [ ] Optional: MV becomes inactive with a clear reason on fingerprint
      regression.

---

osscm:
## Why I'm doing:

By default, StarRocks only rewrites a query against an async materialized view (MV) if every base table is unchanged since the MV's last refresh. `mv_rewrite_staleness_second` relaxes that: it lets the optimizer still use the MV as long as the elapsed time between the base tables' most recent refresh/modification and the MV's own last refresh is within the configured budget.

`isStalenessSatisfied()` computed `mvStaleness = (baseTableRefreshTimestamp - mvRefreshTimestamp) / 1000` and only rejected the MV as fresh when that value exceeded `mv_rewrite_staleness_second`. It never checked for the base table's refresh timestamp going *backwards* relative to the MV's own last refresh.

For Iceberg base tables, `IcebergPartitionTraits.maxPartitionRefreshTs()` returns the raw `currentSnapshot().timestampMillis()`. This is normally monotonically increasing, but Iceberg's `rollback_to_snapshot`/`rollback_to_timestamp` procedures can move a table's current snapshot - and therefore its timestamp - backwards in time. When that happens, `mvStaleness` goes negative, which always satisfies `mvStaleness > staleness_budget` as false, so the MV is treated as fresh and used for rewrite even though the underlying data has been rolled back to a different state than what the MV last captured.

Concretely, `getUpdatedPartitionNamesOfExternalTable()` / `getUpdatedPartitionNamesOfOlapTable()` short-circuit to "no partitions need checking" whenever `isStalenessSatisfied()` returns true, so the bug skips the actual per-partition staleness comparison entirely.

## What I'm doing:

Reject the MV outright whenever `mvStaleness < 0` instead of falling through to the budget comparison, so a rolled-back base table forces the planner to fall through to the normal per-partition comparison instead of blindly trusting a stale MV.

This only affects MVs that explicitly set `mv_rewrite_staleness_second` (default is `0`, i.e. disabled) and, in practice, is only reachable through Iceberg's `rollback_to_snapshot`/`rollback_to_timestamp` procedures, since other connectors' `maxPartitionRefreshTs()` implementations don't have a documented operation that moves their timestamp backwards.

Fixes #71977

## What type of PR is this:

- [x] BugFix

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.3

## Test plan
- [x] Added `MaterializedViewTest#testIsStalenessSatisfiedWithNegativeStaleness` and `#testIsStalenessSatisfiedWithPositiveStalenessWithinBudget`, exercising the new negative-staleness branch and confirming the existing positive-staleness-within-budget path is unaffected.
- [x] `mvn -pl fe-grammar,fe-parser,fe-core -am test-compile` clean.
- [x] `mvn -pl fe-core checkstyle:check` clean.
- [x] Both new tests pass: `Tests run: 2, Failures: 0, Errors: 0`.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/MaterializedView.java
@@ -1252,8 +1252,17 @@ public class MaterializedView extends OlapTable implements GsonPreProcessable, G
 
         long baseTableRefreshTimestamp = baseTableRefreshTimestampOpt.get();
         long mvStaleness = (baseTableRefreshTimestamp - mvRefreshTimestamp) / 1000;
+        ZoneId currentTimeZoneId = TimeUtils.getTimeZone().toZoneId();
+        if (mvStaleness < 0) {
+            // A base table's refresh timestamp regressed below the MV's own refresh timestamp, e.g. after an
+            // Iceberg `rollback_to_snapshot`/`rollback_to_timestamp`. Treat as outdated rather than trusting a
+            // negative staleness, otherwise the MV would be served as fresh without re-checking base tables.
+            LOG.debug("MV is outdated because base tables' lastRefreshTime {} is before MV's lastRefreshTime {}",
+                    DateUtils.formatTimeStampInMill(baseTableRefreshTimestamp, currentTimeZoneId),
+                    DateUtils.formatTimeStampInMill(mvRefreshTimestamp, currentTimeZoneId));
+            return false;
+        }
         if (mvStaleness > this.maxMVRewriteStaleness) {
-            ZoneId currentTimeZoneId = TimeUtils.getTimeZone().toZoneId();
             LOG.debug("MV is outdated because MV's staleness {} (baseTables' lastRefreshTime {} - " +
                             "MV's lastRefreshTime {}) is greater than the staleness config {}",
                     DateUtils.formatTimeStampInMill(baseTableRefreshTimestamp, currentTimeZoneId),
--- a/fe/fe-core/src/test/java/com/starrocks/catalog/MaterializedViewTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/catalog/MaterializedViewTest.java
@@ -56,6 +56,8 @@ import com.starrocks.type.IntegerType;
 import com.starrocks.utframe.StarRocksAssert;
 import com.starrocks.utframe.StarRocksTestBase;
 import com.starrocks.utframe.UtFrameUtils;
+import mockit.Mock;
+import mockit.MockUp;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.MethodOrderer.MethodName;
@@ -1772,5 +1774,45 @@ public class MaterializedViewTest extends StarRocksTestBase {
         Assertions.assertEquals(
                 "{\"default_catalog.db.t_olap\":\"" + TimeUtils.longToTimeString(versionTimeMillis) + "\"}",
                 mv.getBaseTableRefreshVersionTimesJson());
+    }
+
+    /**
+     * Test that isStalenessSatisfied() treats a negative staleness (base table's refresh timestamp
+     * regressed below the MV's own refresh timestamp, e.g. after an Iceberg rollback_to_snapshot) as
+     * outdated rather than incorrectly falling through to "fresh".
+     */
+    @Test
+    public void testIsStalenessSatisfiedWithNegativeStaleness() {
+        MaterializedView.MvRefreshScheme refreshScheme = new MaterializedView.MvRefreshScheme();
+        refreshScheme.setLastRefreshTime(2_000_000L);
+        MaterializedView mv = new MaterializedView(1000, 100, "mv_staleness_test", columns, KeysType.AGG_KEYS,
+                null, null, refreshScheme);
+        mv.setMaxMVRewriteStaleness(600);
+
+        new MockUp<MaterializedView>() {
+            @Mock
+            public Optional<Long> maxBaseTableRefreshTimestamp() {
+                // base table's refresh timestamp is before the MV's own refresh timestamp
+                return Optional.of(1_000_000L);
+            }
+        };
+        Assertions.assertFalse(mv.isStalenessSatisfied());
+    }
+
+    @Test
+    public void testIsStalenessSatisfiedWithPositiveStalenessWithinBudget() {
+        MaterializedView.MvRefreshScheme refreshScheme = new MaterializedView.MvRefreshScheme();
+        refreshScheme.setLastRefreshTime(1_000_000L);
+        MaterializedView mv = new MaterializedView(1000, 100, "mv_staleness_test2", columns, KeysType.AGG_KEYS,
+                null, null, refreshScheme);
+        mv.setMaxMVRewriteStaleness(600);
+
+        new MockUp<MaterializedView>() {
+            @Mock
+            public Optional<Long> maxBaseTableRefreshTimestamp() {
+                return Optional.of(1_000_000L + 500 * 1000L);
+            }
+        };
+        Assertions.assertTrue(mv.isStalenessSatisfied());
     }
 }


dirtysalt (approved):
LGTM! 🚀

HangyuanLiu (approved):

PR #75940, opened 2026-07-07
[Enhancement] Restrict shared-data primary key tables to the cloud-native persistent index

luohaha:
## Why I'm doing:

For shared-data (cloud-native) primary key tables, the table-level index configuration
(`enable_persistent_index` / `persistent_index_type`) that still allows the in-memory index
and the local-disk (`LOCAL`) persistent index is being deprecated. Going forward these tables
should only use the cloud-native persistent index, which is the only index type that supports
the full feature set (parallel publish, SST-based apply, parallel upsert, major compaction, etc.).

## What I'm doing:

Add FE-side restrictions so that shared-data primary key tables can only use the cloud-native
persistent index. This is intentionally FE-only: the BE code paths are left untouched, so existing
tables that were created with a `LOCAL` or in-memory index keep working, and migrating an existing
`LOCAL` table to `CLOUD_NATIVE` via `ALTER` is still allowed.

- CREATE: for cloud-native primary key tables, reject an explicit non-`CLOUD_NATIVE`
  `persistent_index_type`, and force `CLOUD_NATIVE` regardless of
  `enable_cloud_native_persistent_index_by_default`. (The in-memory index was already rejected
  for all primary key tables at CREATE.) The old "LOCAL requires a compute node with storage_path"
  check becomes unreachable and is removed.
- ALTER: reject disabling the persistent index (`enable_persistent_index = false`) and reject
  switching `persistent_index_type` to `LOCAL`. Migration to `CLOUD_NATIVE` is still permitted.
- Remove the now-unused `PropertyAnalyzer.analyzePersistentIndexType` helper (its only remaining
  caller was CREATE, which no longer needs it) and its dedicated unit test.

Backward compatibility: existing `LOCAL`/in-memory shared-data PK tables are still loaded and
served by the unchanged BE; only new CREATE/ALTER requests are restricted.

Fixes N/A

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -2646,6 +2646,22 @@ public class SchemaChangeHandler extends AlterHandler {
                         PropertyAnalyzer.PROPERTIES_ENABLE_PERSISTENT_INDEX, false);
                 persistentIndexType = properties.getOrDefault(PropertyAnalyzer.PROPERTIES_PERSISTENT_INDEX_TYPE,
                         TableProperty.CLOUD_NATIVE_INDEX_TYPE);
+                // Shared-data primary key tables only support the cloud-native persistent index.
+                // The local-disk persistent index and the in-memory index are deprecated: disabling
+                // the persistent index or switching to LOCAL is no longer allowed (migrating an
+                // existing LOCAL table to CLOUD_NATIVE is still permitted). These restrictions only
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

sevev (approved):

xiangguangyxg (approved):

PR #75947, opened 2026-07-07
[BugFix] Keep runtime filter local and singleton for range-colocate joins

xiangguangyxg:
## Why I'm doing:

A range-distributed colocate join buckets its data by range containment, but the BE selects a runtime-filter partition by hashing the join key. With the session variable `enable_pipeline_level_multi_partitioned_rf` enabled, the planner emitted a hash-partitioned runtime-filter layout for such a join, so the BE routed probe rows into partitions that had been assigned by range. The wrong partitioned bloom filter was then tested and matching rows were silently dropped, returning incomplete join results.

The default configuration is unaffected: with `enable_pipeline_level_multi_partitioned_rf` off (the default) the local layout is already `SINGLETON`, so the bug does not fire. The hole is reachable only with that non-default session variable, and only after a colocate group has more than one range.

## What I'm doing:

Keep a range-colocate join's runtime filter local and singleton, so the BE never hash-routes probe rows over range-bucketed data:

- force the local layout to `SINGLETON`;
- clear `pipeline_level_multi_partitioned` so the build side merges a single filter (`num_hash_partitions() == 0`) and the probe skips partition selection;
- never push the filter across an exchange, because a global runtime filter is merged by concatenating each build instance's disjoint partition and no singleton or hash layout can probe that correctly over range-bucketed data.

The range-colocate join is detected in `PlanFragmentBuilder` (where the child distribution specs are available) and the flag is propagated to every `RuntimeFilterDescription` the join builds. Non-range-colocate joins (classic hash colocate, bucket-shuffle) are unaffected. FE-only; no BE or thrift change.

Tests:
- `RuntimeFilterLayoutTest#testLayoutForRangeColocateJoin` — a flagged runtime filter uses `SINGLETON` local layout, is not multi-partitioned, and cannot be pushed across an exchange; a hash-colocate filter keeps its partitioned bucket layout (control).
- `RangeColocateMixedJoinPlanTest#rangeColocateJoinIsFlaggedForRuntimeFilterSingleton` — a range × range colocate join is flagged and each runtime filter it builds stays local; a mixed range × hash bucket-shuffle join is not flagged.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/JoinNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/JoinNode.java
@@ -93,6 +93,10 @@ public abstract class JoinNode extends PlanNode implements RuntimeFilterBuildNod
 
     // The partitionByExprs which need to check the probe side for partition join.
     protected List<Expr> probePartitionByExprs;
+    // True for a colocate join whose colocate columns are range-distributed. Such
+    // joins bucket data by range, which no hash-partitioned runtime-filter layout
+    // can match, so their runtime filters must stay singleton.
+    protected boolean colocateWithRangeDistribution = false;
     protected boolean canLocalShuffle = false;
     protected Map<SlotId, Expr> commonSlotMap;
 
@@ -152,6 +156,14 @@ public abstract class JoinNode extends PlanNode implements RuntimeFilterBuildNod
         return this.probePartitionByExprs;
     }
 
+    public void setColocateWithRangeDistribution(boolean colocateWithRangeDistribution) {
+        this.colocateWithRangeDistribution = colocateWithRangeDistribution;
+    }
+
+    public boolean isColocateWithRangeDistribution() {
+        return this.colocateWithRangeDistribution;
+    }
+
     @Override
     public void buildRuntimeFilters(IdGenerator<RuntimeFilterId> runtimeFilterIdIdGenerator, DescriptorTable descTbl,
                                     ExecGroupSets execGroupSets) {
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
Closing this after cluster e2e validation showed the targeted issue is not reachable in practice.

**Validation.** I built shared-data TSP clusters (unpatched main and this change) and exercised a range-distributed colocate join with runtime filters across **4 and 30 `ColocateRange`s**, `pipeline_dop` 1/2/4, and both full and selective builds, with `enable_pipeline_level_multi_partitioned_rf = true`.

**Result.** On the unpatched build the join returns correct results in every configuration (0% rows dropped; e.g. `t1 ⋈ t2` = 2,000,000 and the selective `t1 ⋈ (t2 where k1 ≤ 1,000,000)` = exactly 1,000,000). The multi-partitioned runtime filter is applied per pipeline driver over co-located data, so it stays consistent with the range placement regardless of the (hash-named) bucket layout.

**Why the guarded path is unreachable.** A range-colocate join's runtime filter is always **local**: a COLOCATE join requires both children to be direct scans, and as soon as an exchange is introduced in the probe subtree the join degrades to a bucket-shuffle (non-colocate) over hash-distributed data. So the global/remote concatenated-RF path this change guarded against does not occur for range-colocate joins.

Since the current behavior is already correct, forcing these runtime filters to `SINGLETON` would only remove a working, more-selective per-range/per-driver filter — a performance regression for no correctness benefit. Closing.

PR #75951, opened 2026-07-07
[BugFix] Consolidate load-spill parallel merge results in flush order

luohaha:
## Why I'm doing:

In the shared-data load-spill parallel merge, `LoadSpillPipelineMergeContext::merge_task_results()`
consolidates each merge task's cloned tablet writer into the parent writer **in the order tasks were
registered** via `add_merge_task()`.

Under **concurrent eager merge** (parallel flush), a flush thread generates a merge batch
(`LoadChunkSpiller::generate_merge_input_batch`, which pulls a slot-ordered, contiguous batch of block
groups under the block-groups lock) and then registers the task (`add_merge_task`, under a *different*
lock) — the two steps are **not atomic**. So two flush threads can register their tasks **out of slot
(memtable flush) order**: e.g. thread A pulls slots [0,1], is preempted, thread B pulls slots [2,3] and
registers first, then A registers.

`merge_other_writer()` appends each task's segments / del files in the iterated order and (for PK tables)
derives the delete rssid / `op_offset` from the running segment count. Consolidating out of flush order
therefore lets an **earlier-flush row win over a later one** — i.e. wrong last-writer-wins for upsert
dedup, and (once the op-aware PK delete path is in) wrong upsert/delete resolution.

## What I'm doing:

Give each merge batch a stable order key and sort by it before consolidating:

- Record the batch's smallest `slot_idx` on `LoadSpillMergeInputBatch` (batches cover a monotonic,
  contiguous slot range, so the min slot uniquely orders them), set in `generate_merge_input_batch`
  right after the block groups are slot-sorted.
- Expose it via `TabletInternalParallelMergeTask::slot_idx()`.
- `stable_sort` `_merge_tasks` by it in `merge_task_results()` before the consolidation loop.

This restores flush order regardless of registration order, without adding a lock on the flush hot path
(the actual merge still runs in parallel; only the final, already-serialized consolidation is ordered).
Only the concurrent eager-merge path was affected — the final merge generates and registers tasks
single-threaded and was already ordered.

BE UT `SpillMemTableSinkTest.test_merge_task_results_orders_by_slot`: registers three tasks (writers with
distinct segment row counts) in scrambled order and asserts the parent writer's segments come out in slot
order. Fails without the fix, passes with it.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/compute_env/load_spill/load_chunk_spiller.cpp
+++ b/be/src/compute_env/load_spill/load_chunk_spiller.cpp
@@ -402,6 +402,10 @@ StatusOr<LoadSpillMergeInputBatch> LoadChunkSpiller::generate_merge_input_batch(
     std::sort(groups.begin(), groups.end(),
               [](const BlockGroupPtrWithSlot& a, const BlockGroupPtrWithSlot& b) { return a.slot_idx < b.slot_idx; });
 
+    // This batch consumes a prefix of the (now slot-sorted) groups, so groups[0] is its smallest slot.
+    // Record it so the consolidation step can restore flush order regardless of task registration order.
+    result_batch.slot_idx = groups[0].slot_idx;
+
     // Tracks the last group index included in this task (used for cleanup at end)
     int64_t stop_idx = -1;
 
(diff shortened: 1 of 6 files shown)


xiangguangyxg (approved):

CelerData-Reviewer:
@codex review

wyb (approved):

PR #75961, opened 2026-07-07
[BugFix] Rebuild the MV query for the PCT fallback in AUTO refresh mode

Youngwb:
## Why I'm doing:

When an incremental-or-auto materialized view is analyzed, the IVM rewrite runs against the query AST **in place**. An aggregate block prepends `__ROW_ID__ = encode(group keys)` to the select list before the rest of the shape is validated. In `AUTO` refresh mode, if a later check rejects an unsupported shape, the `SemanticException` is caught and swallowed and the MV falls back to `PCT` — but the query AST has already been mutated, so the `PCT` MV is built from the half-rewritten query and carries a spurious hidden `__ROW_ID__` column.

`INCREMENTAL` mode is unaffected: it rethrows the exception (CREATE fails) instead of falling back, so the fallback branch is only reachable under `AUTO`.

## What I'm doing:

Fixes #issue: N/A — internal robustness fix for the IVM AUTO-mode PCT fallback.

In `MaterializedViewAnalyzer`, the `AUTO` fallback branch now re-parses the pre-trial query SQL — captured in `inlineViewDef` before the IVM trial runs — and rebuilds the `PCT` MV from that original query instead of the mutated one. This covers every shape whose IVM trial mutates the AST before bailing (aggregate / sub-query / union over an unsupported inner relation).

Note on testing: `AUTO` refresh mode is disabled in my environment and I do not have a cluster to exercise it, so this is not covered by a new end-to-end test. The change is confined to the `AUTO` fallback branch, which is unreachable under `INCREMENTAL` (that path rethrows rather than falling back), so existing IVM tests are unaffected.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/MaterializedViewAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/MaterializedViewAnalyzer.java
@@ -384,7 +384,13 @@ public class MaterializedViewAnalyzer {
                     statement.setRowIdStrategy(result.rowIdStrategy());
                     statement.setCurrentRefreshMode(result.currentRefreshMode());
                 } else {
-                    // if not ivm, set query statement directly
+                    // AUTO mode swallows the IVM SemanticException and falls back to PCT, but the trial may
+                    // have prepended __ROW_ID__ to the query in place before bailing. Re-parse the pre-trial
+                    // SQL so the PCT MV is built from the original query, not the half-rewritten one.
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #75971, opened 2026-07-07
[BugFix] Count container node/table overhead in memory Estimator

Youngwb:
## Why I'm doing:

`Estimator` (used by FE memory tracking and, notably, the Iceberg metadata cache weigher `CachingIcebergCatalog#weighContentFiles`) under-counts containers. Its `Map` and `Collection` branches sum only the container's shallow size plus the **sampled key/value/element objects**, and omit the container's own internals: the per-entry wrappers (`HashMap.Node`, tree/linked entries) and the backing table/array.

On wide maps this omission dominates. Measured on an Iceberg data-file cache (each `DataFile` carries several per-column statistics maps): `HashMap.Node` + bucket arrays were **~35% of the cache heap and completely invisible to the weigher**, so the cache retained roughly **2.4x** its configured `maximumWeight` and could OOM the FE.

## What I'm doing:

Add the missing structural overhead to the estimate:
- **Maps**: per-entry node/entry size (HashMap/ConcurrentHashMap 32B, LinkedHashMap 40B, TreeMap 40B) plus the backing hash table (`capacity * ref`, capacity grown by load factor; tree maps have no table).
- **Stand-alone collections**: hash-set backing map (node + table), tree-set nodes, array-list backing array, linked-list nodes. Not applied to `Map` `keySet()`/`values()` views (their backing is charged once on the owning map).

Sizes assume compressed oops (FE heap < 32G). Validated against JOL `GraphLayout` retained size — `HashMap`/`HashSet`/`ArrayList` estimates move from **0.53–0.90x to 0.98–0.99x** of the true size. Added `EstimatorTest` cases asserting the estimate is within [-15%, +25%] of `GraphLayout`.

Fixes #issue: N/A — FE memory-accounting accuracy fix (no separate tracking issue).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

Weigher-bounded caches (e.g. the Iceberg metadata cache) now account for their true footprint, so for the same `*_memory_usage_ratio` they hold fewer entries / evict earlier. This lowers FE memory (respecting the configured limit) but may increase cache misses; users wanting the previous footprint can raise the ratio.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/memory/estimate/Estimator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/memory/estimate/Estimator.java
@@ -336,10 +336,11 @@ public class Estimator {
         // Handle collections. Non-JDK implementations (Iceberg SerializableMap, Guava Immutable*)
         // typically wrap their real storage in a field; the element-sampling path would skip that
         // storage's own structure entirely, so walk their fields instead — the wrapped JDK container
-        // is then estimated (and charged) when the recursion reaches it.
+        // is then estimated (and charged) when the recursion reaches it. Cache views are excluded
+        // (see isCacheView) and stay on the sampling path.
         if (obj instanceof Collection) {
             Collection<?> collection = (Collection<?>) obj;
-            if (!isJdkContainer(clazz)) {
+            if (!isJdkContainer(clazz) && !isCacheView(clazz)) {
                 return estimateObject(obj, maxDepth, sampleSize);
             }
             return estimateCollection(collection, maxDepth, sampleSize) + collectionBackingOverhead(collection);
@@ -348,7 +349,7 @@ public class Estimator {
         // Handle maps. keySet()/values() are views, so their backing is added once here via
         // mapStructureOverhead rather than inside estimateCollection.
         if (obj instanceof Map<?, ?> map) {
-            if (!isJdkContainer(clazz)) {
+            if (!isJdkContainer(clazz) && !isCacheView(clazz)) {
                 return estimateObject(obj, maxDepth, sampleSize);
             }
             return shallow(obj) +
@@ -500,6 +501,15 @@ public class Estimator {
         return clazz.getName().startsWith("java.util.");
     }
 
+    // A cache's keySet()/values()/asMap() view keeps a back-reference to the owning cache, so
+    // field-walking it would re-traverse and double-count the whole cache (and hand a size-0 count
+    // down to the array sampler for an empty cache). Sample these views instead of field-walking.
+    private static boolean isCacheView(Class<?> clazz) {
+        String name = clazz.getName();
+        return name.startsWith("com.github.benmanes.caffeine.")
+                || name.startsWith("com.google.common.cache.");
+    }
+
     // Bytes of a hash table's backing array: capacity (next power of two that keeps the load factor)
     // times a reference, plus the array header. Sized from the entry count, matching maps built via
     // the sizing/copy constructors (the dominant population); put-grown small maps retain a larger
@@ -609,6 +619,9 @@ public class Estimator {
      * @return a list of sample elements
      */
     private static List<Object> getSamples(Collection<?> collection, int sampleSize) {
+        if (sampleSize <= 0) {
+            return Collections.emptyList();
+        }
         int size = collection.size();
         List<Object> samples = new ArrayList<>(Math.min(sampleSize, size));
 
@@ -659,6 +672,9 @@ public class Estimator {
      * @return a list of sample elements
      */
     private static List<Object> getSamplesFromArray(Object array, int length, int sampleSize) {
+        if (sampleSize <= 0) {
+            return Collections.emptyList();
+        }
         List<Object> samples = new ArrayList<>(Math.min(sampleSize, length));
 
         if (length <= sampleSize) {
--- a/fe/fe-core/src/main/java/com/starrocks/memory/estimate/Estimator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/memory/estimate/Estimator.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

kevincai (dismissed):

CelerData-Reviewer:
@codex review

Youngwb:
Thanks for the continued review @chatgpt-codex-connector.

**Earlier two points — addressed in `d7fb890`:**
- *Test payloads*: switched the new tests from `String` to a small `IntBox` value type whose fields the estimator can see without `--add-opens`; verified against JOL `GraphLayout` **without** opening `java.base/java.lang` (matching the FE-UT surefire argLine), `IntBox` maps/sets read ~1.00 vs `String`'s 0.65.
- *Immutable/view maps*: node+table overhead is now restricted to concrete JDK hash/tree/list types, so `Map.of` / Guava `ImmutableMap` / `singletonMap` (e.g. `ColumnDict.dict`) get no node overhead and are not over-charged.

**Latest two points (retained capacity of pre-sized / sparse containers):** valid in principle, but the true backing-array length (`HashMap.table.length`, `ArrayList.elementData.length`) lives in `java.base` internals that are **not reflectively accessible under the FE-UT/production surefire argLine** — the same constraint that made `String` payloads unusable above — so it cannot be read reliably here. `hashTableBytes`/the array estimate instead reconstruct the capacity a naturally-grown container holds at the current `size()`, which is **exact for insertion-built containers** — the case this change targets (the Iceberg per-column statistics `HashMap`s and cached `Set<DataFile>` are all built via `put`/`add`). Pre-sized-then-sparse containers remain a lower-bound, but that is still a strict improvement over the previous behavior, which charged **zero** structural overhead. Reading the real capacity would need `--add-opens java.base/java.util`, which we don't want to require.

gengjun-git (dismissed):

Youngwb:
Update in `c44bb91` — end-to-end validation on a real cluster caught a gap in the previous revision, now fixed.

**What E2E showed (same cluster, plain main vs +fix, Iceberg metadata cache, `estimateSize` vs GC-verified retained delta):** the previous revision barely moved the Iceberg numbers (0.65 → 0.69 on a 105-column full-stats table). Root cause: Iceberg's statistics maps are **wrapper Maps** (`SerializableMap` / `SerializableByteBufferMap`) holding the real `HashMap` in a field. Restricting the structural charge to concrete JDK types gave wrappers zero, and the Map branch only samples `keySet()`/`values()` — it never visits the wrapped `HashMap` object — so the node/table bytes stayed invisible.

**Fix:** non-JDK `Map`/`Collection` implementations are now **field-walked** instead of element-sampled, so the JDK container they wrap is reached and charged normally. This also handles Guava `ImmutableMap` naturally (its internal arrays are counted exactly — no phantom node charge), keeping the earlier immutable-map concern addressed without a type blocklist. `java.util` non-concrete maps (`Map.of`, singleton, unmodifiable views) keep the previous sampling behavior. Also dropped the 16-slot floor in `hashTableBytes` — sizing/copy-constructed maps (the dominant population here) are right-sized, and a 1-entry wrapper map now estimates to its exact 192 bytes.

JOL `GraphLayout` verification (no `--add-opens`): HashMap / HashSet / ArrayList / wrapper-map(1000-entry) / wrapper-map(1-entry) all at **0.98–1.00×** retained size; added `testWrapperMapCountsInnerMapStructure` mirroring the Iceberg shape. Will follow up with the cluster A/B numbers for the reworked build.

Youngwb:
## E2E validation (C phase): v3 fix confirmed on cluster

Same cluster/methodology as A/B above (marginal-delta method via a second empty catalog to isolate class-loading noise; `enable_query_trigger_analyze=false`).

| Stage | Version | full-stats (unpartitioned) | pruned (id=4, 554,854 files) |
|---|---|---|---|
| A | plain main `3df21e6` | est/real = **0.65** (3,390,315 / 5,224,152) | est/real = **1.21** (889,650,329 / 732,395,968) |
| B | +P1 old `d7fb890` ❌ | **0.69** (3,476,061 / 5,008,376) | **1.30** (953,346,463 / 733,518,416) |
| **C** | **+P1 fixed `e53d081`** | **1.01** (5,446,392 / 5,389,960) | **1.70** (1,243,585,936 / 733,178,128) |

full-stats hits the ~1.0 target. The pruned case moves further from 1.0 than the earlier hand estimate (~1.3-1.4), but in the safe direction (over-, not under-, estimate) — a pre-existing, P1-independent sampling bias in the pruned path (already ~+21% over on plain main) compounds with the now-correct HashMap.Node accounting. Real retained bytes for the pruned case (733M) match A/B almost exactly, confirming actual heap usage is unaffected by this change — only the estimate moved.

### Bug found during this pass, fixed in `e53d081`

Deploying the original `c44bb91` build made `/api/memory_usage` return `{"/ by zero"}` on every call. Root cause: `CachingIcebergCatalog.estimateDataFileCacheSize` passes the connector's `dataFileCache.asMap().keySet()` — a Caffeine `KeySetView`, not a JDK container — into the estimator. The new non-JDK field-walk branch walked that view, which back-references the whole cache:
- an **empty** cache passes `sampleSize=0` down to the array sampler, which divides `length / sampleSize` → `ArithmeticException`, crashing the whole endpoint for any catalog with an empty cache;
- a **non-empty** cache gets re-traversed through the view, double-counting every cached value.

Fixed by excluding cache views (Caffeine, Guava `Cache`) from the field-walk — they stay on the existing sampling path — and guarding the samplers against a non-positive sample size. Added `EstimatorTest.testCacheViewNotFieldWalked` mirroring the connector's actual call shape (this is the case the earlier JOL-based unit tests didn't cover, since they only exercised isolated containers, not cache views). Also fixed two pre-existing collection-test expectations that hadn't been updated for the ArrayList backing-array charge (unrelated to this bug, already stale at `c44bb91`).

`EstimatorTest`: 96/96 passing.

gengjun-git (approved):

Astralidea (approved):

PR #56058, opened 2025-02-19
[Enhancement] Support Paimon view

zyl891229:
## Why I'm doing:
Paimon has supported view to Hive Catalog
[https://github.com/apache/incubator-paimon/pull/4340](https://github.com/apache/incubator-paimon/pull/4340)

## What I'm doing:
Implement Paimon view in Starrocks, includes the following：
1. Support create/replace/drop Paimon view;
2. Support show/desc Paimon view;
3. Support read Paimon view;

Optimized the return value to make the appearance of the table building information clearer
`show create table` adds catalog and database to the return values of ext table and ext view
 
Fixes #56080

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [ ] 3.4
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1
  - [ ] 3.0

---

> [!NOTE]
> Introduces connector-level Paimon view support and improves external DDL output clarity.
> 
> - New `PAIMON_VIEW` type with `PaimonView` and view detection in `Table`; authorizer and MV analyzer updated to handle it
> - `PaimonMetadata`: lists views, creates views from `CreateViewStmt`, fetches and caches views, and drops views; falls back to view when table missing
> - `ColumnTypeConverter`: helpers to convert to/from Paimon schemas (`fromPaimonSchemas`, `toPaimonDataType`, `toPaimonRowType`)
> - `AstToStringBuilder`: `SHOW CREATE` for external tables/views now includes fully-qualified ``catalog.db.name``
[...]

PR #46071, opened 2024-05-22
[Feature] support iceberg view based on rest catalog

stephen-shelby:
## What I'm doing:
1. support create/replace/drop iceberg view based on the rest catalog
2. support show/desc iceberg view based on the rest catalog
3. support read iceberg view based on the rest catalog
Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 3.3
  - [ ] 3.2
  - [ ] 3.1
  - [ ] 3.0
  - [ ] 2.5

PR #40921, opened 2024-02-06
[BugFix] Fix hive view query table/view/alias name are not case-insensitive

Youngwb:
## Why I'm doing:
when hive view's text contanis like this :
`select t1.col1, t1.col2 from T1`
This will result in analyze error
## What I'm doing:
set relation name case insensitive to true  when using hive view
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [x] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 3.2
  - [x] 3.1
  - [ ] 3.0
  - [ ] 2.5

PR #75976, opened 2026-07-07
[BugFix] Make relation alias case-insensitive for all connector views

### LINKED ISSUE #75974: Iceberg/Paimon view query fails with "references invalid table(s) or column(s)" for mixed-case relation aliases

External engines such as Spark and Trino resolve identifiers case-insensitively, so an Iceberg (or Paimon) view they create can reference a relation alias in a different case than it was declared. Create such a view, for example one whose body is:

```sql
SELECT t.id, T.data FROM orders T
```

Register that catalog in StarRocks as an external catalog and query the view:

```sql
SELECT * FROM iceberg_catalog.db.my_view;
```

The view resolves and the query runs, the same as an equivalent Hive view. Hive views whose bodies reference mixed-case relation aliases already work.

### Real behavior (Required)

Planning fails with `View my_view references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them: ...`. When analyzing an external view body StarRocks makes relation-alias matching case-insensitive so that definitions authored by case-insensitive engines still resolve, but the switch in `QueryAnalyzer.visitView` is gated on `isHiveView()`. Iceberg and Paimon views are connector views that are not `isHiveView()`, so their mixed-case definitions are resolved case-sensitively and fail. Paimon views are affected in the same way.

The switch was introduced in #40921 (2024-02), when Hive views were the only connector views. Iceberg views (#46071, which also added the `isConnectorView()` predicate) and Paimon views (#56058) were added later without widening the switch. #70217 fixed the same shape of omission for another Hive-view-only special case that Paimon views needed.

main, and reproduced on all release branches from branch-3.2 to branch-4.1.

cc @Youngwb (author of #40921 / #26094) @stephen-shelby (author of Iceberg view support, #46071)

shyundev:
## Why I'm doing:

Querying an Iceberg or Paimon view fails with `references invalid table(s) or column(s)` when the view's stored definition references a relation alias in a different case than it was declared (for example `SELECT t.id FROM orders T`). Such mixed-case definitions are common because the external engines that author these views (Spark, Trino, Hive) resolve identifiers case-insensitively. StarRocks already accounts for this by making relation-alias matching case-insensitive while it analyzes an external view body, but the switch in `QueryAnalyzer.visitView` was gated on `isHiveView()`, so Iceberg and Paimon views (which are also connector views) never got it and are resolved case-sensitively.

The switch was introduced in #40921, when Hive views were the only connector views; Iceberg views (#46071) and Paimon views (#56058) were added later without widening it. #70217 fixed the same shape of omission for another Hive-view-only special case.

## What I'm doing:

Guard the relation-alias case-insensitivity on `isConnectorView()` (Hive/Iceberg/Paimon) instead of `isHiveView()`, both where it is enabled and where it is restored, so all connector views get the same handling. Internal (Olap) views are unaffected because they are not connector views: their definitions are normalized by StarRocks at creation time, so they never need this. Case-insensitive alias matching is a superset of case-sensitive matching, so consistently-cased definitions (including views authored through the StarRocks dialect) resolve exactly as before.

Adds an Iceberg regression test mirroring the existing Hive one (`HiveViewTest.testQueryHiveViewCaseInsensitive`): a view whose body references a mixed-case relation alias resolves, and alias matching outside the view body stays case-sensitive after the flag is restored. Paimon views take exactly the same `isConnectorView()` path in `visitView`, so they are covered by the same guard; there is no query-level Paimon view mock infrastructure yet to write an equivalent Paimon test against.

Fixes #75974

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/QueryAnalyzer.java
@@ -1591,8 +1591,8 @@ public class QueryAnalyzer {
             boolean isRelationAliasCaseInSensitive = false;
             if (ConnectContext.get() != null) {
                 isRelationAliasCaseInSensitive = ConnectContext.get().isRelationAliasCaseInsensitive();
-                // For hive view, relation alias is case-insensitive
-                if (node.getView().isHiveView()) {
+                // For connector view (hive/iceberg/paimon), relation alias is case-insensitive
+                if (node.getView().isConnectorView()) {
                     ConnectContext.get().setRelationAliasCaseInSensitive(true);
                 }
             }
@@ -1608,7 +1608,7 @@ public class QueryAnalyzer {
                         "function(s) or definer/invoker of view lack rights to use them: " + e.getMessage(), e);
             } finally {
                 viewExpansionStack.remove(viewKey);
-                if (ConnectContext.get() != null && node.getView().isHiveView()) {
+                if (ConnectContext.get() != null && node.getView().isConnectorView()) {
                     ConnectContext.get().setRelationAliasCaseInSensitive(isRelationAliasCaseInSensitive);
                 }
             }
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

PR #75977, opened 2026-07-07
[BugFix] Fix heavy-expr pushdown when a heavy expr references a heavy common sub-expression

### LINKED ISSUE #75975: Planning fails with "Cannot convert ColumnRefOperator to Expr" when a heavy expr references a heavy common sub-expression

With the default `push_down_heavy_exprs = true`, plan a query where a heavy expression (one containing an inline regexp) references a heavy common sub-expression:

```sql
CREATE TABLE t (s STRING);
-- Insert a row so the optimizer keeps the scan (an empty table can be folded away).
INSERT INTO t VALUES ('a');

-- regexp_count(s, 'b') is a heavy common sub-expression:
-- it appears in column x's sum and again as column y, so it is shared (a CSE).
SELECT
    regexp_count(s, 'a') + regexp_count(s, 'b') AS x,
    regexp_count(s, 'b') AS y
FROM t;
```

The query plans and runs successfully.

### Real behavior (Required)

Planning fails (the slot id depends on the schema):

```
Getting analyzing error. Detail message: Cannot convert ColumnRefOperator to Expr, please check the input expression: 4: regexp_count.
```

The heavy common sub-expression is removed from the projection's commonSubOperatorMap to be pushed to the scan (since #67477), but the heavy expression that references it remains, so its column ref can no longer be resolved during translation.

Present since the feature was introduced (4.0.1). On 4.0.5+ (which carry #67477), all 4.1.x, and main it fails with the error above; on 4.0.1-4.0.4 the same query overflows the stack during planning instead.

cc @satanson (author of the heavy-expr pushdown feature #64727 and of #64927 / #67477)

shyundev:
## Why I'm doing:

With the default `push_down_heavy_exprs = true`, a query where a heavy expression references a heavy common sub-expression fails to plan:

```sql
SELECT
    regexp_count(s, 'a') + regexp_count(s, 'b') AS x,
    regexp_count(s, 'b') AS y
FROM t;
```

```
Getting analyzing error. Detail message: Cannot convert ColumnRefOperator to Expr, please check the input expression: 4: regexp_count.
```

`regexp_count(s, 'b')` is shared, so it is removed from the projection's commonSubOperatorMap to be pushed down to the scan (since #67477), but the heavy expression that references it stays, and its column ref can no longer be resolved during translation. More fundamentally, the scan computes each pushed expression independently from base columns, so an expression that references another pushed expression's slot cannot itself be pushed down.

Affects main, 4.1.x, and 4.0.5+ (on 4.0.1-4.0.4 the same query overflows the stack during planning instead).

## What I'm doing:

Exclude from the pushdown every heavy expression (or heavy common sub-expression) that references a heavy common sub-expression — directly, or through the light common sub-expressions that are inlined before the pushdown (light ones can be chained, so the check inlines them recursively). Excluded expressions stay in the ProjectNode and read the referenced sub-expression's slot.

Heavy common sub-expressions computed purely from base columns are still pushed down and materialized by the scan, so the expensive shared regexp keeps being evaluated in the scan and only the dependent expressions fall back to the ProjectNode. This reuses the mechanism that already serves light expressions referencing a pushed-down heavy common sub-expression; independent heavy expressions are pushed down exactly as before.

Tests: three `PushDownHeavyExprsTest` cases (direct reference, nested heavy CSEs, reference through a light CSE) and a `test_push_down_heavy_exprs_common_subexpr` SQL-Tester case asserting identical results with the pushdown ON and OFF.

Fixes #75975

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/plan/PlanFragmentBuilder.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/plan/PlanFragmentBuilder.java
@@ -513,6 +513,24 @@ public class PlanFragmentBuilder {
                 return Collections.emptyMap();
             }
 
+            // The scan computes each pushed expression independently from base columns, so an expression that
+            // references a heavy common sub-expression (directly, or through the light common sub-expressions
+            // inlined before the pushdown) cannot be pushed down. Keep it in the ProjectNode, where it reads
+            // the referenced sub-expression's slot; heavy common sub-expressions computed purely from base
+            // columns are still pushed down and materialized by the scan.
+            Map<ColumnRefOperator, ScalarOperator> lightCommonSubExprs = new HashMap<>(commonSubExprs);
+            heavyCommonSubExprs.keySet().forEach(lightCommonSubExprs::remove);
+            ReplaceColumnRefRewriter lightCommonSubExprInliner =
+                    new ReplaceColumnRefRewriter(lightCommonSubExprs, true);
+            ColumnRefSet heavyCommonSubExprRefs = new ColumnRefSet(heavyCommonSubExprs.keySet());
+            Predicate<ScalarOperator> referencesHeavyCommonSubExpr = expr ->
+                    lightCommonSubExprInliner.rewrite(expr).getUsedColumns().isIntersect(heavyCommonSubExprRefs);
+            heavyExprs.values().removeIf(referencesHeavyCommonSubExpr);
+            heavyCommonSubExprs.values().removeIf(referencesHeavyCommonSubExpr);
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

PR #61335, opened 2025-07-28
[Enhancement] Support iceberg rest catalog vending credentials for GCP

Youngwb:
## What I'm doing:
Support iceberg rest catalog vending credentials for GCP
create statement eg.
use polaris as rest catalog 
```
CREATE EXTERNAL CATALOG polaris_gcp
PROPERTIES (   
    "iceberg.catalog.uri"  = "http://xxxx/api/catalog",   
    "type"  =  "iceberg",   
    "iceberg.catalog.type"  =  "rest",   
    "iceberg.catalog.warehouse" = "iceberg_catalog",
    "iceberg.catalog.security" = "oauth2",
    "iceberg.catalog.oauth2.credential" = "xxx:xxx",
    "iceberg.catalog.oauth2.scope"='PRINCIPAL_ROLE:ALL',
    "iceberg.catalog.rest.nested-namespace-enabled"="true"
 );
```

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

PR #75979, opened 2026-07-07
[BugFix] Fix GCS vended credentials ignored due to gcs-connector 3.x config rename

### LINKED ISSUE #75978: Iceberg REST catalog GCS vended credentials ignored — gcs-connector 3.x config keys not set, falls back to ADC

1. Run an Iceberg REST catalog on GCS that vends storage credentials (e.g. Apache Polaris; server returns `gcs.oauth2.token` / `gcs.oauth2.token-expires-at` in the `loadTable` response `config` when the client sends `X-Iceberg-Access-Delegation: vended-credentials`).
2. Create the catalog exactly as documented (vended credential form, no `StorageCredentialParams`):
```sql
CREATE EXTERNAL CATALOG polaris_catalog
PROPERTIES(
  "type" = "iceberg",
  "iceberg.catalog.type" = "rest",
  "iceberg.catalog.uri" = "http://<polaris>/api/catalog",
  "iceberg.catalog.security" = "oauth2",
  "iceberg.catalog.oauth2.credential" = "xxx:xxx",
  "iceberg.catalog.oauth2.scope" = "PRINCIPAL_ROLE:ALL",
  "iceberg.catalog.warehouse" = "<warehouse>"
);
```
3. `SELECT count(*) FROM polaris_catalog.<db>.<table>;`

FE reads `gs://` metadata files using the vended `gcs.oauth2.token` (feature from #61335), query succeeds.

### Real behavior (Required)

```
ERROR 1064 (HY000): Failed to get status for file: gs://<bucket>/<table>/metadata/snap-xxx.avro
```
fe.warn.log shows the GCS request was authenticated as the VM's default service account (ADC), not the vended token:
```
Caused by: GoogleJsonResponseException: 403 Forbidden
GET https://storage.googleapis.com/storage/v1/b/<bucket>/o/...
"<node-sa>@<project>.iam.gserviceaccount.com does not have storage.objects.get access ..."
at ...GoogleHadoopFileSystem.getFileStatus
at org.apache.iceberg.hadoop.HadoopInputFile.lazyStat
```
Verified with arthas on the live FE that `IcebergCachingFileIO.buildConfFromProperties` receives properties **containing** `gcs.oauth2.token` — the token is present but never used.

Root cause (two stacked issues):
1. The vended token is wired via the gcs-connector **2.x** property `fs.gs.auth.access.token.provider.impl`. The bundled gcs-connector **3.0.13** does not read that key; 3.x requires `fs.gs.auth.type=ACCESS_TOKEN_PROVIDER` + `fs.gs.auth.access.token.provider` (see `HadoopCredentialsConfiguration.AUTHENTICATION_TYPE_SUFFIX` / `ACCESS_TOKEN_PROVIDER_SUFFIX`). Auth silently falls back to application-default credentials. Note `GCPCloudCredential` already uses 3.x-style `fs.gs.auth.type` for the SERVICE_ACCOUNT_JSON_KEYFILE branch — only the access-token branch still uses 2.x keys.
2. Hadoop's `FileSystem` cache is keyed by (scheme, authority, ugi) and ignores `Configuration`, so even a corrected conf never reaches an already-created `GoogleHadoopFileSystem`; vended tokens rotate per table/expiry, so the cache is unsound on this path (`fs.gs.impl.disable.cache=true` needed).

With both fixed (patched classes hot-swapped into a running 4.1.1 FE), the same catalog works: count(*), full scans, and partition aggregations over a 3B-row Iceberg table return results identical to Trino 482 reading the same Polaris catalog with vended credentials. PR follows.

- 4.1.1-14b7e3f (shared-data, K8s operator 1.11.5), also present on main

pushpavanthar:
Fixes #75978

## Why I'm doing:

Iceberg REST catalog vended credentials for GCS (added in #61335) do not work: queries against an Iceberg REST catalog (e.g. Apache Polaris) backed by GCS fail on FE metadata reads with

```
Failed to get status for file: gs://<bucket>/<table>/metadata/snap-xxx.avro
Caused by: GoogleJsonResponseException: 403 Forbidden — <node-SA> does not have storage.objects.get access
```

even though the REST server vends a valid `gcs.oauth2.token` in the `loadTable` response (verified with a manual `loadTable` call carrying `X-Iceberg-Access-Delegation: vended-credentials` — the response contains `config.gcs.oauth2.token`, `gcs.oauth2.token-expires-at` and a `storage-credentials` block).

Root cause is two stacked issues:

1. **gcs-connector 2.x → 3.x config rename.** The vended token is wired using the 2.x property `fs.gs.auth.access.token.provider.impl`. The gcs-connector 3.x bundled with StarRocks (3.0.13 shaded) does not read that key at all — 3.x requires `fs.gs.auth.type=ACCESS_TOKEN_PROVIDER` plus `fs.gs.auth.access.token.provider=<class>` (`HadoopCredentialsConfiguration.AUTHENTICATION_TYPE_SUFFIX` / `ACCESS_TOKEN_PROVIDER_SUFFIX`). With the key ignored, auth silently falls back to application-default credentials. (Notably `GCPCloudCredential` already uses the 3.x style `fs.gs.auth.type` for the SERVICE_ACCOUNT_JSON_KEYFILE branch — only the access-token branch was left on 2.x keys.)

2. **Hadoop FileSystem cache.** `FileSystem` instances are cached by (scheme, authority, ugi) and ignore `Configuration`, so once a `GoogleHadoopFileSystem` is created for a bucket without the token, later reads never see the corrected configuration — and since vended tokens rotate per table and expiry, a cached FS is unsound for this path in general. The fix sets `fs.gs.impl.disable.cache=true` on the vended-token path only.

## What I'm doing:

- `GCPCloudConfigurationProvider`: add constants for the gcs-connector 3.x keys (`fs.gs.auth.type`, `fs.gs.auth.access.token.provider`) and `fs.gs.impl.disable.cache`.
- `IcebergCachingFileIO.buildConfFromProperties`: when `gcs.oauth2.token` is present, set the 3.x auth-type/provider keys and disable the FS cache; keep the legacy 2.x `.impl` key for backward compatibility with 2.x connectors.
- `GCPCloudCredential.tryGenerateHadoopConfiguration`: same for the `hasAccessToken()` branch.

## Verification

Tested against Apache Polaris (Iceberg REST, `security=oauth2`, warehouse on GCS), StarRocks 4.1.1 shared-data on GKE, with a catalog created exactly per the docs (`iceberg.catalog.security=oauth2` + credential + scope). Before the fix: 403 as node SA on every read. After the fix (patched classes hot-swapped into the running FE):

- `SELECT count(*)` (FE manifest read path): correct result
- `SELECT * ... LIMIT 5` (CN scan): correct rows
- `GROUP BY partition` over a 3-billion-row Iceberg table: counts byte-identical to Trino 482 (which reads the same catalog with vended credentials) — three partitions, 3.0–3.1B rows each

Fixes the GCS half of vended-credential support; no changes to S3/Azure paths.

## What type of PR is this:
- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
@@ -240,6 +240,9 @@ public class IcebergCachingFileIO implements FileIO, HadoopConfigurable {
                 copied.set(GCPCloudConfigurationProvider.LEGACY_ACCESS_TOKEN_PROVIDER_IMPL_KEY,
                         GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_IMPL);
                 copied.set(GCPCloudConfigurationProvider.DISABLE_FS_CACHE_KEY, "true");
+                // The base conf may carry catalog-level impersonation, which gcs-connector would apply
+                // on top of the vended token; vended tokens typically lack IAM impersonation permission.
+                copied.unset(GCPCloudConfigurationProvider.IMPERSONATION_SERVICE_ACCOUNT_KEY);
                 copied.set(GCPCloudConfigurationProvider.ACCESS_TOKEN_KEY, entry.getValue());
                 copied.set(GCPCloudConfigurationProvider.TOKEN_EXPIRATION_KEY,
                         properties.getOrDefault(GCPCloudConfigurationProvider.GCS_ACCESS_TOKEN_EXPIRES_AT,
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
@@ -233,7 +233,13 @@ public class IcebergCachingFileIO implements FileIO, HadoopConfigurable {
                 return copied;
             } else if (entry.getKey().equals(GCPCloudConfigurationProvider.GCS_ACCESS_TOKEN)) {
                 // Handle GCS access token
-                copied.set("fs.gs.auth.access.token.provider.impl", GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_IMPL);
+                copied.set(GCPCloudConfigurationProvider.AUTH_TYPE_KEY,
+                        GCPCloudConfigurationProvider.AUTH_TYPE_ACCESS_TOKEN_PROVIDER);
+                copied.set(GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_KEY,
+                        GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_IMPL);
+                copied.set(GCPCloudConfigurationProvider.LEGACY_ACCESS_TOKEN_PROVIDER_IMPL_KEY,
+                        GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_IMPL);
+                copied.set(GCPCloudConfigurationProvider.DISABLE_FS_CACHE_KEY, "true");
                 copied.set(GCPCloudConfigurationProvider.ACCESS_TOKEN_KEY, entry.getValue());
                 copied.set(GCPCloudConfigurationProvider.TOKEN_EXPIRATION_KEY,
                         properties.getOrDefault(GCPCloudConfigurationProvider.GCS_ACCESS_TOKEN_EXPIRES_AT,
--- a/docs/en/data_source/catalog/iceberg/iceberg_catalog.md
+++ b/docs/en/data_source/catalog/iceberg/iceberg_catalog.md
@@ -775,6 +775,10 @@ If you choose Google GCS as storage for your Iceberg cluster, take one of the fo
 
 - To choose REST catalog with vended credential (supported from v4.0 onwards), you do not need to configure `StorageCredentialParams`.
 
+  :::note
+  When vended credentials are used, StarRocks accesses GCS directly with the token vended by the REST catalog. Any `gcp.gcs.impersonation_service_account` configured on the catalog is ignored for that access.
+  :::
+
 `StorageCredentialParams` for Google GCS:
 
 ###### gcp.gcs.service_account_email
--- a/docs/ja/data_source/catalog/iceberg/iceberg_catalog.md
+++ b/docs/ja/data_source/catalog/iceberg/iceberg_catalog.md
@@ -774,6 +774,10 @@ Iceberg クラスターのストレージとして Google GCS を選択した場
 
 - REST カタログで Vended Credential（v4.0以降でサポート）を選択する場合、`StorageCredentialParams` を設定する必要はありません。
 
+  :::note
+  Vended Credential を使用する場合、StarRocks は REST カタログから払い出されたトークンで GCS に直接アクセスします。このとき、カタログに設定された `gcp.gcs.impersonation_service_account` は無視されます。
+  :::
+
 Google GCS 用の `StorageCredentialParams`:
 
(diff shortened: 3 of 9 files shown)


pushpavanthar:
@Youngwb @stephen-shelby PTAL

CelerData-Reviewer:
@codex review

kevincai (commented):
shall we update the document about these config rename?

pushpavanthar:
> shall we update the document about these config rename?

@kevincai  The renamed keys (`fs.gs.auth.type`, `fs.gs.auth.access.token.provider`) are internal gcs-connector Hadoop properties that StarRocks sets itself — they were never part of the user-facing docs, so there's nothing to rename there.
That said, this PR does change one user-visible behavior: when REST catalog vended credentials are in use, `gcp.gcs.impersonation_service_account` is now ignored (the vended token is used directly, since vended tokens typically lack IAM impersonation permission). I've documented that in the GCS section of the Iceberg catalog page in en/zh/ja.

pushpavanthar:
recheck

pushpavanthar:
recheck

kevincai (dismissed):

kevincai:
@pushpavanthar the failed UT should be related to this code change.

pushpavanthar:
@kevincai fixed the UT, please review

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
```diff
@@ -233,7 +233,16 @@ public Configuration buildConfFromProperties(Map<String, String> properties, Str
...
             } else if (entry.getKey().equals(GCPCloudConfigurationProvider.GCS_ACCESS_TOKEN)) {
                 // Handle GCS access token
-                copied.set("fs.gs.auth.access.token.provider.impl", GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_IMPL);
+                copied.set(GCPCloudConfigurationProvider.AUTH_TYPE_KEY,
+                        GCPCloudConfigurationProvider.AUTH_TYPE_ACCESS_TOKEN_PROVIDER);
+                copied.set(GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_KEY,
+                        GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_IMPL);
+                copied.set(GCPCloudConfigurationProvider.LEGACY_ACCESS_TOKEN_PROVIDER_IMPL_KEY,
+                        GCPCloudConfigurationProvider.ACCESS_TOKEN_PROVIDER_IMPL);
+                copied.set(GCPCloudConfigurationProvider.DISABLE_FS_CACHE_KEY, "true");
```
kevincai, line 242: what's the reason to disable the cache explicitly?

kevincai (approved):

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/io/IcebergCachingFileIO.java
pushpavanthar: Hadoop's FileSystem cache is keyed by (scheme, authority, ugi) and ignores conf, so a cached gs:// instance keeps the old token even after we set a new one. Vended tokens rotate per table and expire in ~1h, so reads start failing with 403 after expiry - hit this while testing against Polaris. Disabling the cache only on the vended path makes each FileIO pick up the current token.

pushpavanthar:
@Youngwb @stephen-shelby pls review

pushpavanthar:
@kevincai can you pls add someone relevant to review this PR so that we can have this PR merged ASAP?

Youngwb (approved):

PR #75984, opened 2026-07-07
[BugFix] Fix Spurious Alembic Regeneration for Syntax, Property using SQLGlot

chris-celerdata:
## Why I'm doing:
This PR is for the same issue as #75026. That PR had some problems with SQL parsing through regex. This PR uses SQLGlot to fix this for the fallback approach. 
## What I'm doing:
Use SQLGlot to create a tree for comparison, instead of comparing using string expressions. 
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

contrib/starrocks-python-client/starrocks/alembic/compare.py
```diff
@@ -558,8 +609,17 @@ def _compare_view_definition_and_columns(
...
-    meta_def_norm = TableAttributeNormalizer.normalize_sql(meta_definition, remove_qualifiers=True)
+    if conn is not None and schema is not None:
+        canonical_meta = _get_canonical_sql_via_temp_view(conn, schema, meta_definition)
+        if canonical_meta is not None:
+            conn_def_norm = TableAttributeNormalizer.normalize_sql(conn_definition)
+            meta_def_norm = TableAttributeNormalizer.normalize_sql(canonical_meta)
+            logger.debug("View definition comparison via temp-view canonicalization.")
+        else:
+            logger.debug("Temp-view canonicalization failed for %r; falling back to AST/regex normalizer.", view_fqn)
+            conn_def_norm, meta_def_norm = _normalize_definitions_for_compare(conn_definition, meta_definition)
```
jaogoy, line 620: If you want to use SQLGlot, why use 'temp view' first?

contrib/starrocks-python-client/starrocks/common/sql_canonical.py
```diff
@@ -0,0 +1,118 @@
...
+    # Unquote + case-fold identifiers (string literals are exp.Literal, so left untouched).
+    if isinstance(node, exp.Identifier):
+        name = node.this
+        if isinstance(name, str):
+            node.set("this", name.lower())
+        node.set("quoted", False)
+        return node
+
+    # Strip catalog/db/table qualifiers ("db.t.col" -> "col").
+    if remove_qualifiers and isinstance(node, exp.Column):
```
jaogoy: Does `remove_qualifiers` mean `a.b.c` == `x.y.c`?

jaogoy: Can SQLGlot know whether `a.b.c` equals `c` when parsing?

contrib/starrocks-python-client/starrocks/common/sql_canonical.py
```diff
@@ -0,0 +1,118 @@
...
+                node.set(key, None)
+        return node
+    if remove_qualifiers and isinstance(node, exp.Table):
+        for key in ("db", "catalog"):
+            if node.args.get(key) is not None:
+                node.set(key, None)
+        return node
+
+    # "INNER JOIN" -> "JOIN"; "<dir> OUTER JOIN" -> "<dir> JOIN".
+    if isinstance(node, exp.Join):
```
jaogoy, line 76: Can't `tree.sql(dialect=_DIALECT,...)` canonicalize these keywords, including cases?

contrib/starrocks-python-client/starrocks/alembic/compare.py
chris-celerdata: SQLGlot looks for semantic equivalence. It can be overzealous and collapse statements even when there are changes. For example, `WHERE a = 1 AND a = 1 → WHERE a = 1` or `WHERE x > 1 + 2 → WHERE x > 3`. The view definition stored in StarRocks changes, but the SQLGlot analysis does not.

contrib/starrocks-python-client/starrocks/common/sql_canonical.py
chris-celerdata: It cannot know; it doesn't have the schemas or database structure at parse time. StarRocks always re-qualifies stored view defs to `db.tbl.col` (even join aliases), while models can write columns bare or table-qualified, so the fallback path must drop the injected db prefix or every view diffs spuriously. I've fixed the over-stripping to drop only catalog/db and keep the table/alias, so a.id/b.id stay distinct.

contrib/starrocks-python-client/starrocks/common/sql_canonical.py
chris-celerdata: tree.sql(normalize=True) renders join keywords verbatim (it doesn't collapse INNER JOIN→JOIN or LEFT OUTER JOIN→LEFT JOIN). And StarRocks stores the opposite canonical form from what a model usually writes — it rewrites JOIN→INNER JOIN and LEFT JOIN→LEFT OUTER JOIN. So a model written with JOIN/LEFT JOIN mismatches the stored INNER/LEFT OUTER form under normalize=True alone; this branch collapses both to one form so they compare equal.

contrib/starrocks-python-client/starrocks/common/sql_canonical.py
```diff
@@ -54,27 +54,25 @@ def is_available() -> bool:
...
-    # Strip catalog/db/table qualifiers ("db.t.col" -> "col").
-    if remove_qualifiers and isinstance(node, exp.Column):
-        for key in ("table", "db", "catalog"):
-            if node.args.get(key) is not None:
-                node.set(key, None)
-        return node
-    if remove_qualifiers and isinstance(node, exp.Table):
+    # Strip only the catalog/db qualifier StarRocks injects ("db.t.col" -> "t.col"),
+    # keeping the table/alias so distinct join columns (`a.id` vs `b.id`) don't collapse.
+    if remove_qualifiers and isinstance(node, (exp.Column, exp.Table)):
```
jaogoy, line 68: It will lower the possibility of incorrect normalization, but there are still some corner exceptions:
1.  If there are same table names and column names across different databases, some sql may be wrong.
2. If a user use both `table.column` and `column` in a SQL, the current normalization is not enough.

jaogoy (approved):

contrib/starrocks-python-client/starrocks/common/sql_canonical.py
chris-celerdata: I think if this is documented and only exists on the fallback path, then it is reasonable. Unfortunately, there is not a good way to solve this without knowing the whole database structure.

nshangyimingCD (approved):

PR #70281, opened 2026-03-14
[BugFix] Fix CN crash when scanning empty tablet with physical split enabled

sfwang218:
## Why I'm doing:

Fixes #70280 

## What I'm doing:

When a tablet has no rowsets (empty tablet) but PhysicalSplitMorselQueue
is created with need_split=true, accessing _tablet_rowsets[_tablet_idx][_rowset_idx]
in _cur_rowset() causes array out-of-bounds access, resulting in SIGSEGV.
 
This fix adds:
1. Boundary checks in _cur_rowset() and related functions to handle empty
   inner vectors gracefully
2. Early return in TabletReader::open() when _rowsets is empty to prevent
   creating PhysicalSplitMorselQueue with empty rowsets
3. Unit test for empty tablet scenario with split enabled

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

PR #75985, opened 2026-07-08
[BugFix] Null-safe SparseRangeIterator::has_more() — fix physical-split empty-tablet CN crash (#75203)

### LINKED ISSUE #75203: [Bug] CN crash in PhysicalSplitMorselQueue::_try_get_split_from_single_tablet on 4.0.11 — empty-tablet null-deref persists despite #70281 (fault @0x0)
### Summary
We hit a recurring CN SIGSEGV in PhysicalSplitMorselQueue::_try_get_split_from_single_tablet() on 4.0.11, shared-data mode. This is the same crash path as #70280 ("CN crash when scanning empty tablet with physical split enabled"), but the fix #70281 is already present in our 4.0.11 build (the fix commit is an ancestor of the 4.0.11 tag) and the crash still occurs. The fault address is @0x0 here vs @0x8 in #70280, which suggests a sibling unguarded null/OOB access in the same function that #70281 did not cover.

### Version
4.0.11 RELEASE (build 9559176 distro ubuntu arch aarch64), shared-data (CN on Kubernetes), enable_lake_tablet_internal_parallel = true.

### Steps to reproduce
Shared-data cluster, 4.0.11, lake tablet-internal-parallel (physical split) enabled (default).
Run ordinary scans (we saw it on routine INSERT … SELECT ETL and serving queries) against tables that have empty tablets (empty buckets/partitions, common post-compaction or on sparsely-populated buckets).
Intermittently, a scan that assigns a physical split to an empty tablet crashes the CN. It is non-deterministic (depends on a scan landing on an empty tablet) and was not tied to one query — it rolled across ~8 CNs as different scans tripped it.

### Expected behavior
The scan handles the empty tablet and completes normally (as intended by #70281).

### Real behavior
CN crashes. Representative stack:

```
query_id:019ef00b-…, fragment_instance:019ef00b-…, plan_node_id:10
*** Aborted at 1782143790 (unix time) ***
PC: @  0xbeccd3c  starrocks::pipeline::PhysicalSplitMorselQueue::_try_get_split_from_single_tablet()
*** SIGSEGV (@0x0) received by PID 27 … from PID 0; stack trace: ***
    @  google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @  PosixSignals::chained_handler(int, siginfo_t*, void*)
    @  JVM_handle_linux_signal
    @  0xbeccd3c  starrocks::pipeline::PhysicalSplitMorselQueue::_try_get_split_from_single_tablet()
    @  0xbecd37c  starrocks::pipeline::PhysicalSplitMorselQueue::try_get()
    @  0xd39fe5c  starrocks::lake::TabletReader::open(starrocks::TabletReaderParams const&)
    @  0xc1712d0  starrocks::connector::LakeDataSource::init_tablet_reader(starrocks::RuntimeState*)
    @  0xc171b10  starrocks::connector::LakeDataSource::open(starrocks::RuntimeState*)
    @  0xbebbd64  starrocks::pipeline::ConnectorChunkSource::_open_data_source(starrocks::RuntimeState*, bool*)
    @  0xbebc988  starrocks::pipeline::ConnectorChunkSource::_read_chunk(...)
    @  0xbec2a34  starrocks::pipeline::ChunkSource::buffer_next_batch_chunks_blocking(...)
    @  0xae48380  starrocks::pipeline::ScanOperator::_trigger_next_scan(...)
    @  0xbe218f4  starrocks::workgroup::ScanExecutor::worker_thread()
    @  0xd9c4a24  starrocks::ThreadPool::dispatch_thread()
    @  0xd9baf38  starrocks::Thread::supervise_thread(void*)
```

Surrounding error context on the same CN at crash time also references the lake segment/delvec load path (storage/lake/rowset.cpp:270 load_segments, storage/lake/lake_delvec_loader.cpp:49 lake::get_del_vec, storage/rowset/segment_iterator.cpp _init_column_iterator), which may be relevant to where the null arises for an empty/just-published tablet.

### Relationship to #70280 / #70281
#70281 fixed the OOB on _tablet_rowsets[_tablet_idx][_rowset_idx] in _cur_rowset() for empty tablets, and it is in our 4.0.11 build.
We still crash in the same function with @0x0 (vs @0x8), so there appears to be an additional unguarded path when need_split=true is set for a tablet with no rowsets.
Is this covered by the in-progress rewrite #63310 (PhysicalSplitMorselQueue V2), or should _try_get_split_from_single_tablet get an additional empty-tablet guard on the 4.0/4.1 branches?

### Workaround we settled on
SET GLOBAL enable_lake_tablet_internal_parallel = false; — disables the physical-split path so PhysicalSplitMorselQueue is never constructed. No restart, reversible; costs intra-tablet scan parallelism.

srihithg:
> **Revised per review (thanks @kevincai).** This PR went through two rounds:
> 1. An earlier revision guarded `_cur_scan_morsel()` / the tail of `try_get()`, which is
>    downstream of the crash and did not fix it. `addr2line` on the reported PC shows the
>    fault is in `SparseRangeIterator::has_more()`, inlined into
>    `_try_get_split_from_single_tablet()`.
> 2. This revision fixes the crash at the source (transient lake segment-load failures are
>    now surfaced as a retryable Status instead of being swallowed) **and** keeps a null-safe
>    `has_more()` guard as a backstop. See "What I'm doing" below for why the guard alone is
>    not sufficient.

## Why I'm doing:

https://docs.google.com/document/d/13W1-3RqB316jsgXSRAVCkv0bwx8wr1kziz7q0pvXhsk/edit?tab=t.0

Shared-data CN nodes crash with `SIGSEGV @0x0` inside
`PhysicalSplitMorselQueue::_try_get_split_from_single_tablet()` on 4.0.11 (issue #75203),
even with #70281 present. `addr2line` on the reported PC resolves to:

```
std::vector<starrocks::Range<uint32_t>>::size() const   (stl_vector.h)
  (inlined by) SparseRangeIterator<uint32_t>::has_more() const   (storage_primitive/range.h)
  (inlined by) PhysicalSplitMorselQueue::_try_get_split_from_single_tablet()
```

It fires non-deterministically when a physical-split scan
(`enable_lake_tablet_internal_parallel = true`, the default) hits an empty/edge tablet.

## What I'm doing:

**Root cause.** The split loop evaluates `_segment_range_iter.has_more()` in its `while`
condition. `lake::Rowset::get_segments()` is *not safe to call twice*: on a transient
segment-load failure it swallows the error and returns `{}` **without caching**, so a later
call at the same cursor can return a non-empty result (an empty→non-empty "flip"). When the
first call returns empty, `_init_segment()` early-returns before assigning
`_segment_range_iter`, leaving it **default-constructed** (`_range == nullptr`). A subsequent
loop iteration then reaches `has_more()`, which did an unconditional dereference:

```cpp
inline bool SparseRangeIterator<T>::has_more() const {
    return _index < _range->_ranges.size();   // _range == nullptr -> vector::size() at addr 0 -> @0x0
}
```

`#70281` added bounds checks to `_cur_rowset()` / `_cur_segment()`, but the iterator
dereference in `has_more()` was left unguarded, so the crash persists.

**Fix (two layers).**

1. **Root cause — stop swallowing the load failure.** New `BaseRowset::get_segments_checked()`
   surfaces a segment-load failure as its real (retryable) `Status` instead of returning `{}`.
   `lake::Rowset` overrides it to do the real lazy load and keys the cache on an explicit
   `_segments_loaded` flag, so a genuinely zero-segment rowset loads exactly once while a
   transient failure leaves the flag `false` and a retry re-attempts. `get_segments()` becomes
   a thin best-effort shim over it. `PhysicalSplitMorselQueue::_init_segment()` primes the load
   via `get_segments_checked()` and `RETURN_IF_ERROR`s on its status, so a transient failure
   fails the query cleanly **before** the loop can reach `has_more()` on an uninitialized
   iterator.

2. **Backstop — null-safe `has_more()`.** A default-constructed iterator has no ranges, so it
   must report `false` rather than dereference a null `_range`:

   ```cpp
   return _range != nullptr && _index < _range->_ranges.size();
   ```

**Why the guard alone is not enough.** If we only add the `has_more()` guard, the crash is
gone but the bug becomes a *silent row-drop*: on the empty→non-empty flip, `has_more()`
returns `false`, `_next_segment()` advances past the segment that just loaded, and the query
returns missing rows with **no error**. Fixing the swallow at the source turns that transient
failure into a clean, retryable query error; the guard remains only as defense-in-depth.

## How this was reproduced and verified

Built the affected tests under ASAN in the StarRocks dev-env image (aarch64).

- **`storage_primitive_test`** — a new regression test constructs a default
  `SparseRangeIterator<>` and calls `has_more()`. WITHOUT the guard: deterministic SEGV at
  `range.h` in `std::vector<Range>::size()` (matching the issue's `addr2line`); WITH it, the
  test passes.
- **`exec/pipeline/scan/physical_split_morsel_queue_test`** — path-level regression
  (`TransientEmptyRowset` + `test_transient_get_segments_null_range_75203`) drives the exact
  empty→non-empty flip. With the source-level prime, the path now **scans** the segment
  (`try_get` returns a real morsel) instead of dropping it, so its assertion is non-null.
- **`storage/lake/tablet_reader_test`** — drives a real `PhysicalSplitMorselQueue` over a lake
  rowset with a fail-once-then-succeed injected on the segment load; `try_get()` now returns
  the injected retryable error instead of crashing.

```
==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008
    #1 SparseRangeIterator<unsigned int>::has_more() const   be/src/storage_primitive/range.h
SUMMARY: AddressSanitizer: SEGV ... in std::vector<starrocks::Range<unsigned int>>::size() const
```

(The fault address is a codegen artifact — `@0x0` in the production `-O1` build where GCC
fuses the two-field load, `@0x8` under `-O0`/ASAN where the operand loads are sequential —
same function, same null `_range`.)

Fixes #75203

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

(Normal-path behavior is unchanged; the fix only prevents a null-pointer dereference on an
empty/edge-tablet path that previously crashed the CN.)

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4


CelerData-Reviewer:
@codex review

be/src/storage/query/split_morsel_queue.cpp
```diff
@@ -215,7 +215,23 @@ StatusOr<MorselPtr> PhysicalSplitMorselQueue::try_get() {
...
+    // _tablet_rowsets with an out-of-range cursor here dereferences a garbage
+    // ScanMorsel and crashes the CN with SIGSEGV @0x0.
+    //
+    // Return an error rather than nullptr: rowid_range is non-null here, so rows have
+    // already been taken from the split walk. nullptr means end-of-splits to callers
+    // (e.g. lake TabletReader breaks its collection loop), which would silently skip
+    // those rows. A non-OK status makes lake TabletReader fall back to the non-split
+    // collector path and makes pipeline callers fail the query instead.
     auto* scan_morsel = _cur_scan_morsel();
+    if (scan_morsel == nullptr || _tablet_idx >= _tablet_rowsets.size()) {
```
kevincai: this doesn't look like the root cause of the crash. It crashed at `_try_get_split_from_single_tablet` before reaching here. Definsive code here doesn't help the previous nullptr crash.

kevincai:
```
$ addr2line -e starrocks_be -aipfC 0xbeccd3c 0xbecd37c
0x000000000beccd3c: std::vector<starrocks::Range<unsigned int>, std::allocator<starrocks::Range<unsigned int> > >::size() const at /usr/include/c++/12/bits/stl_vector.h:988 (discriminator 8)
 (inlined by) starrocks::SparseRangeIterator<unsigned int>::has_more() const at be/build_Release/src/exec/be/src/storage/range.h:391 (discriminator 8)
 (inlined by) starrocks::pipeline::PhysicalSplitMorselQueue::_try_get_split_from_single_tablet() at be/build_Release/src/exec/be/src/exec/pipeline/scan/morsel.cpp:353 (discriminator 8)
0x000000000becd37c: starrocks::pipeline::PhysicalSplitMorselQueue::try_get() at be/build_Release/src/exec/be/src/exec/pipeline/scan/morsel.cpp:412 (discriminator 1)
```

CelerData-Reviewer:
@codex review

be/src/storage/query/split_morsel_queue.cpp
srihithg: You're right, thanks for catching that. My first fix was in the wrong place and didn't actually address the crash. I reworked the PR to fix it at the real source and added a test that reproduces the original crash and passes with the fix. Let me know if you'd rather I take a different approach.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage_primitive/range.h
```diff
@@ -392,7 +392,12 @@ inline SparseRangeIterator<T>::SparseRangeIterator(const SparseRange<T>* r) : _r
 
 template <typename T>
 inline bool SparseRangeIterator<T>::has_more() const {
-    return _index < _range->_ranges.size();
+    // A default-constructed iterator (_range == nullptr) has no ranges to traverse.
+    // Guard the null before dereferencing: PhysicalSplitMorselQueue can reach
+    // _segment_range_iter.has_more() while the iterator is still default-constructed
+    // (empty/edge tablet, see issue #75203); without this check that dereferenced a
+    // null _range -> std::vector::size() at address 0 -> CN SIGSEGV @0x0.
+    return _range != nullptr && _index < _range->_ranges.size();
```
kevincai, line 398: what code path leads to this nullptr range in the Iterator and still called the `has_more()`. This is a defensive code fix, but the initial point why it can trigger this nullptr, is not clear.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage_primitive/range.h
```diff
@@ -392,7 +392,18 @@ inline SparseRangeIterator<T>::SparseRangeIterator(const SparseRange<T>* r) : _r
...
+    // A default-constructed iterator (_range == nullptr) has no ranges to traverse.
+    // Guard the null before dereferencing: PhysicalSplitMorselQueue can reach
+    // _segment_range_iter.has_more() while the iterator is still default-constructed.
+    // This happens when _init_segment() early-returns before it reaches
+    // `_segment_range_iter = _segment_scan_range.new_iterator()` -- e.g. a transient
+    // lake Rowset::get_segments() failure returns an empty (uncached) segment list, so
+    // _cur_segment() is null for that call; the immediate retry then succeeds and the
+    // while-condition's checks 1-3 fall through to this has_more() with _range still null
+    // (issue #75203). Without this guard, _range->_ranges.size() dereferences null ->
+    // std::vector::size() read at offset 0/8 of a null object -> CN SIGSEGV (@0x0 in the
```
kevincai: make the comment compact and tight, it is a defensive code change.

CelerData-Reviewer:
@codex review

kevincai (approved):

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

dirtysalt (approved):
LGTM! 🚀

PR #75445, opened 2026-06-26
[BugFix] Narrow UNNEST result struct type consistently with its input array element

Hyper-FF:
## Why I'm doing:

PruneSubfieldsForComplexType.visitPhysicalTableFunction narrows the unnest input
array element type (via the access group setUnnest() propagates output->input) but
the matching narrow of the unnest OUTPUT column is gated by pruneForComplexType's
scanRefs check, which an unnest result column never satisfies. So the input element
became struct<...> with fewer fields while the output/returnTypes kept the original
width, and the unnest emitted rows whose struct width differed from its input
element -> StructColumn::append field-count mismatch (BE abort under DCHECK; silent
under-read of the unguarded struct serde in release). Narrow the unnest result
column directly by its own visited access group so input and output stay in sync.

## What I'm doing:

run case:

```
139976525506112 struct_column.cpp:169] Check failed: _fields.size() == struct_column.fields().size() (1 vs. 2) F20260623 07:51:15.003372 139976565671488 
struct_column.cpp:169] Check failed: _fields.size() == struct_column.fields().size() (1 vs. 2) F20260623 07:51:15.003516 139976364844608 struct_column.cpp:169
] Check failed: _fields.size() == struct_column.fields().size() (1 vs. 2) 
F00000000 00:00:00.000000 139976565671488 logging.cc:1962] RAW: Check data_->num_chars_to_log_ > 0 && data_->message_text_[data_->num_chars_to_log_ - 1] == '\
n' failed: 
[1782201075.3][thread: 139976585754176] je_mallctl execute purge success
[1782201075.3][thread: 139976585754176] je_mallctl execute dontdump success
*** Aborted at 1782201074 (unix time) try "date -d @1782201074" if you are using GNU date ***
[1782201075.3][thread: 139976525506112] je_mallctl execute purge success
[1782201075.3][thread: 139976525506112] je_mallctl execute dontdump success
[1782201075.3][thread: 139978091955776] je_mallctl execute purge success
[1782201075.3][thread: 139978091955776] je_mallctl execute dontdump success
[1782201075.3][thread: 139976565671488] je_mallctl execute purge success
[1782201075.3][thread: 139976565671488] je_mallctl execute dontdump success
PC: @     0x7f525e3909fc pthread_kill
*** SIGABRT (@0x2117f) received by PID 135551 (TID 0x7f4ed6abb640) LWP(136379) from PID 135551; stack trace: ***
[...]

PR #76002, opened 2026-07-08
[BugFix] Prune UNNEST output struct by its input array group, not the output's own

jenwitteng:
## Why I'm doing:

`PruneSubfieldsForComplexType` can declare an UNNEST result column's struct type **narrower** than the element type BE actually materializes for the UNNEST **input** array. #75445 fixed the common case, but a residual case remains: when the input array is kept **fully materialized** by another consumer (classically `SELECT t.*`, or any access that retains all subfields), the input element stays full (`struct<sf1,sf2,sf3>`) while the output is pruned to only the subfield that output reads (`struct<sf3>`).

BE then materializes the full element but FE declares a narrower output tuple, so the widths disagree at the exchange and shuffle deserialization fails with:

```
encode size does not equal when decoding
```
(or `Expected remains size N, but get 0`; a `StructColumn::append` DCHECK abort under debug builds).

**Root cause:** #75445 narrows the output by the **output's own** access group (`pruneUnnestResultType`). That is correct only when the input array is pruned to the same group — it holds in the case #75445 tested and in the shared-array case (where #75012 makes input and output share the same group object), but not when the input array's group is a **superset** of the output's group.

## What I'm doing:

Narrow the UNNEST output by the **input** array column's access group — the union of every access to that array, which is exactly what BE materializes — and **skip pruning entirely** when the input is not a real prune boundary: a fully-materialized / "select all" access path, or a derived (non-scan) column. `canSafelyPruneUnnestOutput` recurses through stacked UNNESTs so a nested-array chain (`UNNEST(arr2d)` then `UNNEST(arr1d)`) is pruned only when it bottoms out at a pruned scan column.

Gated behind the existing `enable_prune_complex_types` / `enable_prune_complex_types_in_unnest` (both must be on to reach this path).

### Reproduction (on current `main`)

```sql
CREATE TABLE complex_nested (
    id int,
    nested struct<n_id int, xxx array<struct<sf1 varchar(100), sf2 int, sf3 varchar(100)>>>
) ENGINE=OLAP partition by id distributed by hash(id) buckets 10

SET enable_prune_complex_types = true;
SET enable_prune_complex_types_in_unnest = true;

-- SELECT x.* keeps `nested` (and its array nested.xxx) fully materialized.
SELECT n.sf3, x.* FROM complex_nested x, unnest(x.`nested`.xxx) AS t(n) WHERE n.sf3 IS NOT NULL;
```

Before: `returnTypes: [struct<sf3 varchar(100)>]` while the projection/scan keep the input element full `struct<sf1,sf2,sf3>` → width mismatch on shuffle.
After: `returnTypes: [struct<sf1 varchar(100), sf2 int(11), sf3 varchar(100)>]`.

### Tests

Adds three plan tests to `PruneComplexSubfieldTest`, all passing with the fix (the middle one fails on `main` without it):

| Test | Shape | Before | After |
|------|-------|--------|-------|
| `testUnnestOutputMatchesInputWhenArrayAlsoAccessedDirectly` | array read directly **and** unnested | pass | pass |
| `testUnnestOutputMatchesFullyMaterializedInput` | input kept full via `SELECT t.*` | **fail** | pass |
| `testPruneStackedUnnestOnNestedArray` | stacked UNNEST of a 2D array | pass | pass |

Full `PruneComplexSubfieldTest` (62 tests) is green with the fix.

Related: #75445 (output/input width consistency), #75012 (`setUnnest` access-group merge).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Miscellaneous: fixes an incorrect pruned output type that crashed shuffle deserialization

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/PruneComplexTypeUtil.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/PruneComplexTypeUtil.java
@@ -164,6 +164,13 @@ public class PruneComplexTypeUtil {
             return unnestColRefMap.containsKey(columnRefOperator);
         }
 
+        // Returns the input array column that an UNNEST output column was produced from, or null if
+        // the given column is not an UNNEST output. Used to walk stacked UNNESTs (UNNEST of an
+        // UNNEST output) when deciding whether an output can be pruned in lockstep with its input.
+        public ColumnRefOperator getUnnestInput(ColumnRefOperator output) {
+            return unnestColRefMap.get(output);
+        }
+
         public ComplexTypeAccessGroup getVisitedAccessGroup(ColumnRefOperator columnRefOperator) {
             return accessGroups.get(columnRefOperator);
         }
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/PruneSubfieldsForComplexType.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rule/tree/PruneSubfieldsForComplexType.java
@@ -198,7 +198,7 @@ public class PruneSubfieldsForComplexType implements TreeRewriteRule {
                     // every subfield, so the output must stay full too. Pruning it would make FE
                     // declare a narrower struct than BE materializes for that input, producing
                     // "encode size does not equal when decoding" on shuffle deserialization.
-                    if (!canSafelyPruneUnnestOutput(input, context)) {
+                    if (!canSafelyPruneUnnestOutput(input, context, optExpression)) {
                         continue;
                     }
                     // Prune the output to the INPUT array column's access group (not the output's own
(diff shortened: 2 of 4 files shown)


jenwitteng:
@Hyper-FF  this extends your #75445 for the case where the UNNEST input array is kept fully materialized by a sibling consumer (e.g. SELECT t.*) — there the output's own access group is a strict subset of the input array's, so narrowing by the output group under-declares the struct. Curious if you hit this class of case; happy to adjust the approach.

CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

stdpain:
https://github.com/Mergifyio backport  branch-4.1.3 branch-4.0.13

PR #76004, opened 2026-07-08
[BugFix] Keep decoded CTE-consume output columns as string refs

Hyper-FF:
## Why I'm doing:

For low-cardinality string columns, DecodeRewriter.visitPhysicalCTEConsume rewrote every entry of the CTE output column map to its dict ref unconditionally, without checking whether the column still arrives at the consume in dict form. If a column was decoded inside the CTE producer (e.g. by lead() with a non-null default) it is emitted by the produce fragment as a string slot, but the consume was rewritten to reference the (now absent) dict slot. The query then fails at BE with "Runtime error: slot_id N not found" (debug builds: Check failed: is_slot_exist(slot_id)).

Only rewrite a CTE output column to its dict ref when the producer column is still in dict form (DecodeInfo.inputStringColumns) -- the same guard order-by, group-by and partition-by columns already use.

Add an FE plan test and a SQL regression test.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

stdpain (approved):

trueeyu (approved):

PR #76011, opened 2026-07-08
[Enhancement] Support UTC-adjusted Parquet TIMESTAMP and ORC TIMESTAMP_INSTANT sort keys on the tablet pre-split meta tier

xiangguangyxg:
## Why I'm doing:

Sample-Based Tablet Pre-Split has a "meta tier" that plans split boundaries from
Parquet/ORC footer statistics (no sub-query), falling back to a "data tier"
sub-query for types it cannot decode safely. Two high-frequency timestamp
flavors were still rejected by the meta tier and paid the data-tier cost:

- **Parquet INT64 TIMESTAMP with `isAdjustedToUTC=true`** -- the default modern
  Parquet writing (Spark / pyarrow with a timezone), a UTC-normalized instant.
- **ORC `TIMESTAMP_INSTANT`** (timestamp with local time zone).

Both are the same case: the BE load adds the load session-timezone offset to the
stored UTC instant to obtain the local DATETIME wall clock, so the footer's raw
UTC min/max did not match the value the BE routes/stores, and the reader
conservatively deferred to the data tier. This was the last common gap in the
meta-tier type coverage line (DATE/DATETIME/DECIMAL/VARCHAR/CHAR/local-TIMESTAMP
already covered).

## What I'm doing:

FE-only. When the load session timezone resolves to a **fixed offset**, the two
footer readers reproduce the exact offset the BE load applies and lift these two
flavors onto the meta tier:

- New shared helper `MetaTierTemporalWindow.fixedLoadOffset(String)` resolves the
  load timezone to a constant `ZoneOffset` **only** for a fixed-offset zone
  (`ZoneId.getRules().isFixedOffset()`, offset sampled at `Instant.EPOCH`);
  a DST / named / alias (CST, PRC) / unresolvable / absent timezone returns empty.
- `ParquetRowGroupStatisticsReader` accepts INT64 + TIMESTAMP with
  `isAdjustedToUTC=true`, and `OrcStripeStatisticsReader` accepts
  `TIMESTAMP_INSTANT`, **only** when that fixed offset is present; each decodes the
  raw UTC footer min/max, adds the constant offset (order-preserving, so min/max
  are preserved), then applies the existing `[0001-01-01, 9999-12-31]` window gate
  on the offset-adjusted **local** date.
- The load timezone is threaded from each load kind's scan context, sourced from
  the pre-split hook's own `ConnectContext` session timezone -- the same object
  `JobSpec` builds the BE query globals from -- so the FE boundary and the BE
  stored value use the identical timezone by construction, for both
  INSERT-from-FILES and Broker Load.

**Conservative first version (fail-safe):** only fixed-offset load timezones are
lifted. A DST / named zone (including the FE default `Asia/Shanghai`) applies a
per-instant offset the BE row load computes with cctz that a single scalar cannot
reproduce; those keep using the data tier. Any unresolvable case -> data tier.
Getting the meta tier for these types therefore requires an explicit fixed-offset
`SET time_zone` (e.g. `+08:00`). Pre-split boundaries are advisory: the BE routes
every row by its true value, so this is never a wrong split or a load failure.

**Behavior change: No.** This only changes which planning tier computes the
split boundaries (footer stats vs sub-query); the loaded data, the resulting
tablet layout (by value), and all user-facing surfaces are unchanged.

Verified: 197 focused FE unit tests pass (fixed-offset accept + correct offset
boundary, sub-second preserved, offset crossing the date, DST/named/null rejected,
out-of-window-after-offset rejected, both load kinds); `mvn checkstyle` clean.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/BrokerLoadPreSplitHook.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/BrokerLoadPreSplitHook.java
@@ -145,8 +145,11 @@ public final class BrokerLoadPreSplitHook {
             PreSplitMetrics.recordEligibilitySkip(tableLevelSkip);
             return;
         }
+        // The load session timezone. This same context feeds JobSpec.fromBrokerLoadJobSpec ->
+        // loadPlanner.getContext() for the BE query globals, so it matches the offset the BE applies to
+        // a UTC-adjusted / TIMESTAMP_INSTANT value. A non-fixed zone -> the readers defer to data tier.
         BrokerLoadScanContext scanContext = new BrokerLoadScanContext(
-                brokerDesc, fileGroups, fileStatuses, computeResource);
+                brokerDesc, fileGroups, fileStatuses, computeResource, context.getSessionVariable().getTimeZone());
         PreSplitFlow.Prepared prepared = new PreSplitFlow.Prepared(
                 scanContext,
                 MetaUtils.getRangeDistributionColumns(targetTable),
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/BrokerLoadRowGroupStatisticsProvider.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/BrokerLoadRowGroupStatisticsProvider.java
@@ -86,7 +86,7 @@ final class BrokerLoadRowGroupStatisticsProvider implements RowGroupStatisticsPr
                 MetaTierFormat format = MetaTierFormat.fromBrokerFormatType(
                         Load.getFormatType(declaredFormat, brokerFileStatus.path), brokerFileStatus.path);
                 FileStatus hadoopFileStatus = PreSplitHadoopAccess.toHadoopFileStatus(brokerFileStatus);
-                aggregated.addAll(format.read(hadoopFileStatus, hadoopConfig, sortKeyColumn));
+                aggregated.addAll(format.read(hadoopFileStatus, hadoopConfig, sortKeyColumn, context.loadTimeZone()));
             }
         }
         return aggregated;
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/BrokerLoadScanContext.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/BrokerLoadScanContext.java
@@ -38,12 +38,18 @@ import java.util.Objects;
  * <p>The {@code brokerDesc} may be {@code null} for HDFS-style direct loads
  * that don't go through a broker — matching the same nullability rule the
  * load planner already follows.
+ *
+ * <p>{@code loadTimeZone} is the load session timezone (the same value the BE
+ * query globals use). The meta-tier readers use it to reproduce the BE's
+ * offset for a UTC-adjusted / TIMESTAMP_INSTANT value; a non-fixed / null
+ * zone -> data tier.
  */
 public record BrokerLoadScanContext(
         BrokerDesc brokerDesc,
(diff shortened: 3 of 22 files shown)


xiangguangyxg:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg:
## TSP e2e (shared-data) — PASS

Cluster-validated the FE<->BE load-timezone handshake that unit tests cannot see.

**Setup**
- Build: TSP 2467, `--merge-pr` on PR head, Release, `--linux-distro ubuntu`.
- Cluster: shared-data, 1 FE + 3 CN, `--linux-type ubuntu` (matches the build), aliyun; released after the run.
- Fixtures (pyarrow): one tz-aware UTC timestamp column, 3 ordered non-overlapping bands (UTC `2020-01-01`, `2021-06-01`, `2022-12-01`), 1,000,000 rows/band, written both ways:
  - Parquet -> INT64 TIMESTAMP with `isAdjustedToUTC=true` (`DESC FILES` infers `datetime`).
  - ORC -> `TIMESTAMP_INSTANT` / timestamp with local time zone (`DESC FILES` infers `varchar`).
- Load: `SET enable_range_distribution = true;` + `INSERT INTO <range-dist table> SELECT * FROM FILES(...)`.

**Positive — fixed-offset load timezone `SET time_zone = '+08:00'`** (the meta tier reproduces the BE's constant +8h offset):

| Table | source | `tier_used` | `external_boundaries_fallback_total` (3 CN) | tablets | COUNT | MIN / MAX (stored, local) |
|---|---|---|---|---|---|---|
| `t_parquet` | Parquet `isAdjustedToUTC=true` | **meta_tier** (Δ+1) | **0 / 0 / 0** | 1 -> 2 | 3,000,000 | `2020-01-01 08:00:00` / `2022-12-01 08:00:00.999999` |
| `t_orc` | ORC `TIMESTAMP_INSTANT` | **meta_tier** (Δ+1) | **0 / 0 / 0** | 1 -> 2 | 3,000,000 | `2020-01-01 08:00:00` / `2022-12-01 08:00:00.999999` |

The stored MIN/MAX are the UTC footer values shifted by exactly the load's `+08:00` offset (`2020-01-01T00:00:00Z` -> `2020-01-01 08:00:00`), so the FE-planned meta-tier boundary equals the BE-routed local value — the split completed on the meta tier with **zero** boundary fallback on every CN.

**Negative — DST load timezone `SET time_zone = 'America/New_York'`** (gate holds; per-instant offset a scalar cannot reproduce):

| Table | source | `tier_used` | result |
|---|---|---|---|
| `t_parquet_ny` | Parquet `isAdjustedToUTC=true` | **data_tier** (Δ+1) | 3,000,000 rows, MIN `2019-12-31 19:00:00` (UTC -5h EST), no error |
| `t_orc_ny` | ORC `TIMESTAMP_INSTANT` | **data_tier** (Δ+1) | 3,000,000 rows, MIN `2019-12-31 19:00:00`, no error |

A DST/named zone falls back to the data tier (no meta-tier promotion), the load still succeeds, and no regression — exactly the conservative, fail-safe behavior intended.

**Net:** `tier_used{meta_tier}` +2 (both fixed-offset loads), `tier_used{data_tier}` +2 (both DST loads), `external_boundaries_fallback_total` = 0 on all CNs for the meta-tier loads.

kevincai (approved):

srlch (approved):

PR #76040, opened 2026-07-08
[Enhancement] Support right-outer/semi/anti and full-outer range-colocate joins

xiangguangyxg:
## Why I'm doing:

Shared-data (存算分离) RANGE-distribution colocate tables only produced the shuffle-free colocate-join plan for INNER / LEFT OUTER / LEFT SEMI / LEFT ANTI. RIGHT OUTER / RIGHT SEMI / RIGHT ANTI / FULL OUTER fell back to a hash shuffle even when both sides are range-colocated on the join key. This was the last open join-type gap in the range-colocate optimizer.

## What I'm doing:

Extend the range-colocate join planner to RIGHT OUTER / RIGHT SEMI / RIGHT ANTI / FULL OUTER. FE-optimizer-only:

- `ChildOutputPropertyGuarantor.canRangeColocateJoin`: accept the four additional equi-join families (CROSS, NULL-aware anti, and ASOF joins stay rejected). The join-hint short-circuit, `disable_colocate_join` kill switch, structural `canColocate`, and position-preserving join-key pairing gates are unchanged.
- `OutputPropertyDeriver.visitPhysicalJoin`: for a range × range colocate join, derive the dominated output distribution — the right-side range spec for right-family joins, the null-relaxed left-side spec for full-outer, the left-side spec otherwise — mirroring the existing hash `computeColocateJoinOutputProperty` mapping.

**Why it is correct** (the same invariant INNER/LEFT already rely on): a range-colocate join is bucket-local — rows sharing a colocate key land in the same ColocateRange/bucket on every table in the group, PACK-colocated on one CN. The unmatched right rows a right/full-outer join must emit — including NULL-key rows, which sort into the first ColocateRange consistently on both the FE range geometry and the BE ingest routing — are produced by the right scan node of the same COLOCATE fragment, the identical mechanism hash colocate already uses for right/full. No scheduler or BE change is needed.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Plan-shape optimization only: the colocate plan returns results byte-identical to the pre-existing hash-shuffle plan (that identity is exactly what the shared-data cluster e2e below verifies).

## Checklist:

- [ ] This is a backport pr

## Test coverage

FE unit tests (45/45 pass, checkstyle clean):
- `ChildOutputPropertyGuarantorRangeTest` — allowlist accepts RIGHT OUTER/SEMI/ANTI + FULL OUTER; CROSS and NULL-aware left anti stay rejected; join hints, `disable_colocate_join`, arity/swapped/partial-prefix pairing still fall back to shuffle.
- `OutputPropertyDeriverRangeTest` — dominated-side output derivation (left / right / full-outer null-relaxed).
- `RangeColocateMixedJoinPlanTest` — range×range RIGHT OUTER and FULL OUTER on the colocate key plan as `colocate: true`; `[shuffle]` hint and `disable_colocate_join` fall back to shuffle.

Shared-data cluster e2e (ship gate): for RIGHT OUTER / RIGHT SEMI / RIGHT ANTI / FULL OUTER (incl. NULL-safe `<=>`) the colocate result set is compared byte-identical against the `set disable_colocate_join = true` hash-shuffle oracle, over matched / left-only / right-only / duplicate / NULL keys spanning multiple ColocateRanges with right-only buckets inside populated partitions.

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
@@ -106,9 +106,15 @@ public class ChildOutputPropertyGuarantor extends PropertyDeriverBase<Void, Expr
      *     {@code HINT_JOIN_SKEW}, or {@code HINT_JOIN_BUCKET} disables
      *     colocate. Only {@code null} or {@code HINT_JOIN_COLOCATE} lets
      *     the range fast path proceed.
-     * <li>Join-type allowlist: INNER, LEFT OUTER, LEFT SEMI, LEFT ANTI.
-     *     Right-family and full-outer joins are rejected (NULL-key runtime
-     *     correctness validation deferred past P2).
+     * <li>Join-type allowlist: INNER, LEFT/RIGHT OUTER, LEFT/RIGHT SEMI,
+     *     LEFT/RIGHT ANTI, FULL OUTER. All equi-join families are colocate-safe
+     *     because a colocate range join is bucket-local: rows sharing a colocate
+     *     key (NULL included — NULL sorts into the first ColocateRange on every
+     *     table in the group) always land in the same bucket, so the unmatched
+     *     right rows a right/full-outer join must emit are produced by the right
+     *     scan node in the same fragment. CROSS, NULL-aware anti, and ASOF joins
+     *     stay rejected (no covering equijoin key / cross-bucket NULL semantics /
+     *     inequality match).
      * <li>{@code disable_colocate_join} session kill switch (shared with
      *     hash colocate).
      * <li>Structural {@link RangeDistributionSpec#canColocate}: same group,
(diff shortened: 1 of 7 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
### Cluster e2e validation (shared-data, 1 FE + 3 CN)

Built this PR and validated on a shared-data cluster that the range-colocate plan returns results **byte-identical to the hash-shuffle oracle** for every added join type, over matched / left-only / right-only / duplicate / NULL keys.

**Single ColocateRange** (key-coverage): `RIGHT OUTER`, `FULL OUTER`, right-semi (`EXISTS`), right-anti (`NOT EXISTS`) each plan as `colocate: true` and return results identical to the `set disable_colocate_join=true` oracle. `FULL OUTER ... ON a <=> b` (null-safe) also plans colocate and is correct.

**Multiple ColocateRanges with a right-only bucket**: forced a Level-1 split (group `IsStable=true`, 2 buckets), then arranged the left table to have **zero rows** in the upper bucket while the right table has rows there. `RIGHT OUTER` / `FULL OUTER` plan `colocate: true` and correctly emit the right-only bucket's rows NULL-padded, byte-identical to the oracle — confirming the colocate fragment's right scan node scans empty-left buckets (the same mechanism hash colocate uses; no `isRightOrFullBucketShuffle` needed).

This is the behavior-identity the "No behavior change" classification rests on.

kevincai (approved):

srlch (approved):

PR #76046, opened 2026-07-08
[BugFix] Guard expression partition prune with monotonic function check

eshishki:
## Why I'm doing:

Expression-based range partition pruning ("further prune", `enable_expr_prune_partition`, on by default) evaluates a predicate `f(partition_col) cmp const` by mapping each candidate partition's two boundary values through `f` and intersecting the result with the predicate range (`PartitionColPredicateEvaluator.visitCall`). That is only sound when `f` is monotonic, but `PartitionColPredicateExtractor.visitCall` admits **any** FE-foldable single-column function without a monotonicity check. A non-monotonic function wraps around inside a partition's span, so both boundaries can map to the same value while the partition covers the whole domain — and the partition is pruned with matching rows inside.

Reproduction (default settings):

```sql
create table t (dt date, v int)
partition by range(dt) (
  partition p2024 values [('2024-01-01'), ('2025-01-01')),
  partition p2025 values [('2025-01-01'), ('2026-01-01')))
distributed by hash(v) properties("replication_num"="1");

insert into t values ('2024-05-15', 1);
select * from t where month(dt) = 5;  -- returns empty
```

Both partitions map to month range [1, 1] (`month` of Jan 1 for every year boundary), [1, 1] does not intersect [5, 5], both partitions are pruned, the row silently disappears. Same class: `day()`, `hour()`, `dayofweek()`, `dayofyear()`. The endpoint swap in the evaluator handles *decreasing* monotonic functions but masks wraparound instead of failing.

## What I'm doing:

Gate `PartitionColPredicateExtractor.visitCall` with `OperatorFunctionChecker.onlyContainMonotonicFunctions` — the same registry already used for this purpose by `ListPartitionPruner` (generated-column conjunct deduction) and `PartitionSelector`. Monotonic expressions (`date_trunc`, `days_add`, `last_day`, `datediff`, order-preserving `date_format`, ...) keep pruning — every pruning expression in the existing `FurtherPartitionPruneTest` suite is registry-flagged and stays green. Non-monotonic expressions now degrade to scan-level predicates instead of pruning incorrectly, so affected queries scan more partitions — correctness over performance.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/PartitionColPredicateExtractor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/rewrite/PartitionColPredicateExtractor.java
@@ -27,6 +27,7 @@ import com.starrocks.sql.optimizer.operator.scalar.ConstantOperator;
 import com.starrocks.sql.optimizer.operator.scalar.InPredicateOperator;
 import com.starrocks.sql.optimizer.operator.scalar.IsNullPredicateOperator;
 import com.starrocks.sql.optimizer.operator.scalar.LikePredicateOperator;
+import com.starrocks.sql.optimizer.operator.scalar.OperatorFunctionChecker;
 import com.starrocks.sql.optimizer.operator.scalar.ScalarOperator;
 import com.starrocks.sql.optimizer.operator.scalar.ScalarOperatorVisitor;
 import com.starrocks.sql.optimizer.rewrite.scalar.FoldConstantsRule;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):

stdpain:
This can be used as a hotfix.

stephen-shelby (approved):

stephen-shelby:
Followed the repro in the description (month(dt) = 5) but couldn't get it to actually drop the row. mapRangeBoundValue folds bounds via FoldConstantsRule(needMonotonicFunc=true), which already skips folding for any function without isMonotonic=true — month()/day()/hour() etc. aren't annotated as such, so it already falls back to full-scope (no pruning) instead of computing [1,1].

How did you reproduce the empty result? Which build/version?

eshishki:
@stephen-shelby You're right — the repro doesn't work. Here's how the mistake happened: I saw that the extractor admits any foldable single-column call, assumed the evaluator folds whatever it receives, and wrote the repro from that reading instead of running it. The `needMonotonicFunc=true` fold in `mapRangeBoundValue` already rejects `month()` and friends, the bounds don't fold, the partition falls back to full scope, and no row is lost. There was no bug.

Is the change worth anything then? Honestly, very little. The new gate calls the same check one layer earlier (`onlyContainMonotonicFunctions` delegates to `ScalarOperatorEvaluator.isMonotonicFunction`), so pruning behavior is identical before and after — it only skips building candidate ranges that would all come out full-scope anyway. Worse, the code comment I added describes a wrong-pruning scenario that cannot actually happen. The only real value here is the new tests pinning "non-monotonic predicates must not prune", which would catch a future regression in either layer.

Suggestions: revert this, and close the backports (#76089, #76091, #76092) — there is nothing to fix on release branches.

Sorry for the noise, and thanks for checking this properly.

PR #76047, opened 2026-07-08
[Enhancement] Support add/drop of a key column for shared-data range-distribution tables

xiangguangyxg:
## Why I'm doing:

Shared-data (lake) range-distribution tables previously hard-rejected any `ALTER TABLE` that added or dropped a key (range sort-key) column, because the persisted per-tablet range boundaries live in the sort-key column space. HASH-distributed tables already support these operations; this brings range distribution to parity.

## What I'm doing:

Extends the existing range-rewrite schema-change job (`LakeRangeRewriteSchemaChangeJob`) to column-set changes:

- **ADD COLUMN** of a key column with a constant `DEFAULT` on Duplicate / Aggregate / Unique Key tables — the added key column is carried in the shadow index (`__starrocks_shadow_`-prefixed, hidden from the user namespace until the atomic flip strips it) and materialized from its constant default by the load path; a K-tablet shadow is sampled by the new sort key and atomically flipped in.
- **DROP COLUMN** of a key (range sort-key) column on Duplicate and Aggregate (without a `REPLACE` / `REPLACE_IF_NOT_NULL` value column) tables — the rewrite re-sorts and, for Aggregate, re-aggregates under the reduced key.

The universal identity guards are preserved, matching HASH: Primary Key add/drop key, Unique Key drop key, and Aggregate drop-key-with-`REPLACE` stay rejected. Routing applies only to single-index cloud-native range tables (no rollup / synchronous materialized view, non-colocate, non-auto-increment); a drop that would orphan a secondary index is rejected. FE-only — reuses the existing `ADD COLUMN` / `DROP COLUMN` syntax (no grammar/BE/thrift/proto change). Docs (en/zh/ja) updated.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -2221,6 +2236,60 @@ private AlterJobV2 createRangeRewriteJob(Database db, OlapTable olapTable, List<
...
+        List<Column> shadowSchema = new ArrayList<>(postAddSchema.size());
+        for (Column column : postAddSchema) {
+            Column copy = column.deepCopy();
+            if (!baseNames.contains(column.getName())) {
+                // A routed added key column's default must be materializable by fillShadowColumns
+                // (CONST/NULL scalar). Key columns are scalar (complex ARRAY/MAP/STRUCT are forbidden as
+                // keys), so this holds, but guard explicitly rather than rely on it.
+                if (copy.getDefaultExpr() != null && copy.getDefaultExpr().hasExprObject()) {
+                    throw new DdlException("ADD COLUMN of a range key column requires a scalar constant "
+                            + "default; complex/expression defaults are not supported. Column: " + column.getName());
```
kevincai, line 2282: **Potential bug: the "scalar constant default" guard is applied to every newly-added column, not just the key column.**

This branch is entered for any column absent from the pre-add base schema (`!baseNames.contains(...)`) with no `column.isKey()` check, then throws `"ADD COLUMN of a range key column requires a scalar constant default..."` whenever `getDefaultExpr().hasExprObject()` is true. The adjacent comment documents the intent as scoped to the *key* column ("A routed added key column's default...; Key columns are scalar"), but the check actually runs for every newly-added column.

Impact: a single multi-column `ADD COLUMN` clause that bundles a range key column with a non-key value column having a complex (`ARRAY`/`MAP`/`STRUCT`) constant default is rejected with a misleading, mis-attributed error. For example, on `t(k1 int, v1 int) order by(k1)`:

```sql
ALTER TABLE t ADD COLUMN (k2 INT KEY DEFAULT '0', v2 ARRAY<INT> DEFAULT '[]');
```

routes here because `k2` joins the key-derived sort key, and then wrongly rejects on `v2` — a plain value column that would otherwise be a legal add.

Suggested fix: gate the `hasExprObject()` check on `column.isKey()`.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -2523,6 +2598,81 @@ public int getAsInt() {
...
+                    }
+                    // (c) processDropColumn() above already removed any secondary index (BITMAP/GIN/NGRAM)
+                    //     on the dropped column from newIndexes -- a copy -- not from OlapTable.indexes
+                    //     itself. The routed early return never commits newIndexes back onto the table
+                    //     (that happens in finalAnalyze, which this path bypasses), so silently continuing
+                    //     would leave OlapTable.indexes referencing a now-missing column. Reject instead.
+                    Column droppedColumnForIndexCheck = olapTable.getColumn(dropColumnClause.getColName());
+                    if (droppedColumnForIndexCheck != null && olapTable.getIndexes().stream()
+                            .anyMatch(index -> index.getColumns().stream()
+                                    .anyMatch(c -> c.equalsIgnoreCase(droppedColumnForIndexCheck.getColumnId())))) {
```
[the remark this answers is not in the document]
xiangguangyxg: Fixed in `781de0d`. The routed `DROP COLUMN` now rejects dropping a range sort-key column that is a plain bloom-filter column — checked via `OlapTable.getBfColumnNames()`, since plain bloom filters live on `bfColumns` rather than in `getIndexes()` — and asks the user to remove it from `bloom_filter_columns` first. Verified on a shared-data cluster: dropping such a column is now rejected. Regression test: `testDropRangeSortKeyColumnWithBloomFilterRejected`.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
xiangguangyxg: Fixed in `781de0d`. The scalar-constant-default check is now gated on `column.isKey()`, so it only applies to the added key column. A multi-column `ADD COLUMN` that bundles a key column with a non-key value column carrying a complex default is no longer rejected. Verified on a shared-data cluster: `ADD COLUMN (k2 INT KEY DEFAULT '0', v2 ARRAY<INT> DEFAULT [1,2,3])` completes and the array default is materialized as `[1,2,3]` on all historical rows. Regression test: `testMultiColumnAddKeyDoesNotRejectNonKeyComplexDefault`.

xiangguangyxg:
## End-to-end validation on a shared-data cluster

Validated on shared-data (存算分离) clusters, **3 FE + 3 CN**, built from this PR (`main-781de0d`). All scenarios below pass.

### Add / drop a key column — data-model matrix

| Scenario | Result |
|---|---|
| DUPLICATE, DROP a range sort-key column (multi-tablet) | all rows kept and re-sorted under the reduced sort key; row count + `SUM(value)` conserved |
| DUPLICATE, ADD a key column with a constant `DEFAULT` (multi-tablet) | default materialized on every historical row; row count + sum conserved |
| AGGREGATE, DROP a range sort-key column (multi-tablet) | rows re-aggregate under the reduced key; `SUM` conserved (e.g. 500k rows → 100k, every group sum correct) |
| UNIQUE, ADD a key column with a `DEFAULT` | default materialized on every historical row; count conserved |
| PRIMARY KEY add/drop key · UNIQUE drop key · AGGREGATE drop-key with a `REPLACE` value column | all rejected |

### Review-comment fixes

| Scenario | Result |
|---|---|
| DROP a range sort-key column that is also in `bloom_filter_columns` | rejected with a clear message (remove it from `bloom_filter_columns` first) |
| `ADD COLUMN (k2 INT KEY DEFAULT '0', v2 ARRAY<INT> DEFAULT [1,2,3])` | accepted; the array default materialized as `[1,2,3]` on all historical rows |

### Online-rewrite flip on an empty partition (bug fix)

| Scenario | Before this PR | After |
|---|---|---|
| Empty table: DROP a key column, then ADD a key column | the second ALTER hangs in the publish/flip (repeated 404 on a missing source log) | both complete |
| Empty table: `ORDER BY` reorder ×2 | the second ALTER hangs | both complete |

This confirms the fix is in the shared online-rewrite path, so it also covers sort-key reorder, keyness flip, rollup, and materialized-view sort-key rewrite on empty partitions.

### Durability

| Scenario | Result |
|---|---|
| Concurrent writes during the rewrite window | no writes lost; the added column's default materialized on rows written concurrently (e.g. 3,000,000 + 150 concurrent inserts, all present, all carry the default) |
| 3-FE leader failover mid-rewrite (leader killed during the watershed phase) | a follower is elected, the schema-change job replays from the journal to completion; row count + sum + materialized default intact |

---

All FE unit suites and the SQL integration tests are green. The only red check, `docusaurus-build`, is unrelated CI infrastructure — that job merges several in-flight docs branches before building and one carries a corrupt `docs/docusaurus/package.json`; the broken-link check over this PR's docs reports zero errors.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/LakeOnlineRewriteJobBase.java
```diff
@@ -823,6 +855,31 @@ protected RewriteStatus classifyRewrite(long physicalPartitionId) {
...
+        // A non-empty rewrite whose DPP_NORMAL_ALL counter was somehow unreported (loadedRows left 0) still
+        // has commit infos, so BE loads its real source instead of dropping the partition's data.
+        TxnCommitAttachment attachment = txnState.getTxnCommitAttachment();
+        if (!(attachment instanceof InsertTxnCommitAttachment)
+                || ((InsertTxnCommitAttachment) attachment).getLoadedRows() != 0) {
+            return false;
+        }
+        Set<TabletCommitInfo> commitInfos = txnState.getTabletCommitInfos();
+        return commitInfos == null || commitInfos.isEmpty();
+    }
```
kevincai, line 886: **The two-signal data-loss guard here is effectively dead code in production — it collapses to `loadedRows == 0` alone.**

`rewriteSourceEmpty()` is only reachable after every partition's rewrite txn is `VISIBLE` (the loop at L757-793 runs only once `classifyRewrite()` returns `DONE`, and `DONE` requires `TransactionStatus.VISIBLE`, L838). But the `VISIBLE` transition in `DatabaseTransactionMgr.finishTransaction()` (L1471) and `finishTransactionNew()` (L2411) upserts `copiedState` into the txn map and then calls `resetTransactionStateTabletCommitInfos(copiedState)`, which sets `tabletCommitInfos = null` (`TransactionState.resetTabletCommitInfos()`, L638-641). That reset runs on the **leader** too — it's a memory optimization applied as the tail of the finish, not just the follower-null case this comment describes.

So by the time this method calls `txnState.getTabletCommitInfos()`, it is already `null` in production, `commitInfos == null` is unconditionally `true`, and the guard degrades to `loadedRows == 0` alone — exactly the single point of failure the second signal was meant to backstop. If `loadedRows` is ever misreported as `0` for a rewrite that actually wrote data, BE synthesizes an empty `op_schema_change@W` and silently drops the partition's pre-watershed data (`be/src/storage/lake/transactions.cpp` L231, `alter_version == 1 || txn.shadow_rewrite_source_empty()`).

Compounding it: `LakeRangeRewriteSchemaChangeJobTest.testSourceEmptyFlagReflectsRewriteLoadedRows` hand-injects `setTabletCommitInfos()` on a fresh `TransactionState` that never goes through the finish/reset flow, so the "commit infos present ⇒ not empty" case passes without exercising the production path.

Suggested direction: either sample the commit-info signal **before** the txn finishes (while `tabletCommitInfos` is still populated), or drop it and document `loadedRows == 0` as the sole (persisted, failover-safe) signal — and update the test to drive the real finish path so it reflects production.

xiangguangyxg: Good catch — you're right, it was dead code. `resetTransactionStateTabletCommitInfos` runs in the VISIBLE finish (`finishTransaction`/`finishTransactionNew`, right before `notifyVisible()`) on the leader too, and `rewriteSourceEmpty()` only runs after the rewrite txn is VISIBLE (`classifyRewrite` -> `DONE`), so `getTabletCommitInfos()` is always null here and the guard collapsed to `loadedRows == 0`.

Since the commit-info signal is not obtainable in this flow (it is cleared before control returns from the rewrite INSERT), I dropped the second signal and documented `loadedRows == 0` as the sole, persisted (failover-safe) signal. It is authoritative: a lake load that wrote rows reports them via `DPP_NORMAL_ALL` (the counter behind "N rows affected"), so `0` uniquely means an empty rewrite — a non-empty rewrite reporting `0` would break INSERT row-counting everywhere, which is out of scope to defend here. I also removed the misleading test that hand-injected `setTabletCommitInfos()` on a `TransactionState` that never went through the finish/reset path. Fixed in `1a077bb`.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -2220,6 +2235,65 @@ private AlterJobV2 createRangeRewriteJob(Database db, OlapTable olapTable, List<
...
+                // keys), so this holds, but guard explicitly rather than rely on it. Scope the guard to the
+                // KEY column: a multi-column ADD may also bundle a non-key value column with a complex
+                // (ARRAY/MAP/STRUCT) constant default. That value column is shadow-prefixed and materialized
+                // by the same fillShadowColumns path -- its deep-copied shadow Column carries the default as
+                // a string (Column.getDefaultValue() == defaultExpr.toSql()), which fillShadowColumns emits
+                // as a VARCHAR constant and then casts to the column type -- so it needs no key-only reject.
+                if (copy.isKey() && copy.getDefaultExpr() != null && copy.getDefaultExpr().hasExprObject()) {
+                    throw new DdlException("ADD COLUMN of a range key column requires a scalar constant "
+                            + "default; complex/expression defaults are not supported. Column: " + column.getName());
+                }
```
kevincai, line 2283: **Bug: a bundled non-key column with a complex (ARRAY/MAP/STRUCT) default NPEs the routed range ADD**

This guard only rejects a complex/expression default when `copy.isKey()`. The comment above argues a bundled non-key value column with a complex default is safe because "its deep-copied shadow Column carries the default as a string (`Column.getDefaultValue() == defaultExpr.toSql()`), which fillShadowColumns emits as a VARCHAR constant". But `fillShadowColumns` does not call `getDefaultValue()` — it routes such a column to `materializeShadowColumnDefault`, which calls `Column.calculatedDefaultValue()`. Unlike `getDefaultValue()`, `calculatedDefaultValue()` has no `hasExprObject()` branch and returns `null` for a complex default, so `ConstantOperator.createVarchar(null)` throws NPE via `Objects.requireNonNull`.

Repro: `ALTER TABLE t ADD COLUMN (k1 INT KEY, v1 ARRAY<INT> DEFAULT [1,2,3])` on a shared-data range table (fast_schema_evolution on) whose sort key is key-derived. The single `AddColumnsClause` routes here via `addTouchesKeyDerivedRangeSortKey`; the `alterClauses.size() > 1` check only blocks combining multiple clauses, not multiple columns in one clause; and `v1` slips past this `isKey()` guard. The NPE hits the internal rewrite INSERT and any concurrent double-write while the shadow index is registered.

Suggest extending the guard to reject a complex/expression default on any bundled routed-add column, not just the key column.

xiangguangyxg: Thanks — you're right that relying on `calculatedDefaultValue()` here was fragile, and I've hardened it. Two parts:

1. It is not actually a live NPE today. The column reaching `materializeShadowColumnDefault` is the **deep-copied** shadow column, and `Column`'s copy ctor sets `this.defaultValue = column.getDefaultValue()` (Column.java:297), which for a complex default returns `defaultExpr.toSql()`. So the copy's `defaultValue` is populated (e.g. `"[1,2,3]"`) and `calculatedDefaultValue()` returns it via its `defaultValue != null` branch — not null. Verified on a shared-data cluster with your exact repro: `ADD COLUMN (k2 INT KEY DEFAULT '0', v2 ARRAY<INT> DEFAULT [1,2,3])` completes and materializes `v2 = [1,2,3]` on every historical row (a `createVarchar(null)` would have cancelled the job, not produced `[1,2,3]`).

2. But that only works via the copy ctor's incidental `defaultValue` population, which is exactly the subtlety you (and several earlier reviewers) flagged. So in `eca8b101` I made `materializeShadowColumnDefault` robust by construction: the CONST branch now falls back to `getDefaultValue()` (which renders the expr-object default via `toSql()`) when `calculatedDefaultValue()` is null, so a bundled non-key value column with a complex constant default is materialized as a VARCHAR constant (cast to the column type downstream) regardless of whether `defaultValue` was pre-populated — it can no longer NPE. `calculatedDefaultValue()` is still tried first so a time-function default (e.g. `CURRENT_TIMESTAMP`) is still evaluated to the txn time.

I kept the `isKey()` scoping (per your earlier comment, a non-key value column with a legal complex default should not be rejected) — it is now safely materialized rather than rejected.

fe/fe-core/src/main/java/com/starrocks/sql/InsertPlanner.java
```diff
@@ -847,15 +847,20 @@ private OptExprBuilder fillShadowColumns(ColumnRefFactory columnRefFactory, Inse
                 String originName = Column.removeNamePrefix(targetColumn.getName());
                 Optional<Column> optOriginColumn = outputFullSchema.stream()
                         .filter(c -> c.nameEquals(originName, false)).findFirst();
-                Preconditions.checkState(optOriginColumn.isPresent());
-                Column originColumn = optOriginColumn.get();
-                ColumnRefOperator originColRefOp = outputColumns.get(outputFullSchema.indexOf(originColumn));
+                if (optOriginColumn.isPresent()) {
```
luohaha, line 850: Replacing the old `Preconditions.checkState(optOriginColumn.isPresent())` with this `if/else` is required for the range ADD-key case, but it also silently weakens the invariant for the pre-existing HASH shadow-column (MODIFY COLUMN) path: any future bug that produces a shadow column without its same-named origin will now quietly emit a DEFAULT/NULL value instead of failing fast.

Consider gating the origin-absent branch so it is only taken for a genuinely-new added column (or at least a `LOG.warn`), so the "new add" and "missing origin for a MODIFY" cases stay distinguishable.

fe/fe-core/src/main/java/com/starrocks/alter/LakeOnlineRewriteJobBase.java
```diff
@@ -758,7 +772,24 @@ protected void runRunningJob() throws AlterCancelException {
...
+        // the absent source -- this is the empty-at-W>1 case (a repeat online rewrite of an empty range
+        // partition; W==1 empty is already handled by BE). Read from the rewrite txn's (persisted)
+        // InsertTxnCommitAttachment, the same attachment classifyRewrite already relies on, so a leader
+        // failover that re-runs this recovers it; and journaled by the persist below.
+        for (long physicalPartitionId : partitionStates.keySet()) {
+            PartitionRewriteState state = partitionStates.get(physicalPartitionId);
+            if (state.sourceEmpty == null && state.rewriteTxnId != null) {
+                TransactionState txnState = GlobalStateMgr.getCurrentState().getGlobalTransactionMgr()
+                        .getTransactionState(dbId, state.rewriteTxnId);
+                if (txnState != null) {
```
luohaha, line 788: `state.sourceEmpty` is only set when `getTransactionState(...)` returns non-null. If the rewrite txn has been GC'd by the time a leader failover re-runs this (or the attachment is unexpectedly not an `InsertTxnCommitAttachment`), the flag stays `null` -> at flip `Boolean.TRUE.equals(sourceEmpty)` is `false` -> BE takes the "source op_write MUST exist" branch. For a partition empty at W > 1 that reintroduces exactly the 404 flip-wedge this PR fixes.

The window is narrow and arguably no worse than pre-PR behavior, but it is silent. A `LOG.warn` when the txn state is unexpectedly unavailable here would make the degradation diagnosable.

fe/fe-core/src/main/java/com/starrocks/alter/LakeRangeRewriteSchemaChangeJob.java
```diff
@@ -285,26 +298,78 @@ private List<Tuple> planBoundaries(String dbName, String tableName, String parti
...
+     */
+    private static String projectSortKeyColumn(Column column, Set<String> sourceColumnNames) {
+        String backquoted = ParseUtil.backquote(column.getName());
+        if (sourceColumnNames.contains(column.getName())) {
+            return backquoted;
+        }
+        // Backslash must be escaped before the double quote, or a trailing backslash in the
+        // default would swallow the closing quote and a mid-string backslash would be silently
+        // reinterpreted by the StarRocks string-literal lexer.
+        String literal = column.getDefaultValue() == null
```
luohaha, line 367: This builds the sampling projection from `column.getDefaultValue()` (raw literal), whereas `InsertPlanner.materializeShadowColumnDefault` uses `calculatedDefaultValue()` (which evaluates time-function defaults). For a plain scalar constant these agree, and for an added key column the sampled value is immaterial to boundaries (constant across rows -> no distinction -> K collapses), so there is no correctness impact today.

But if a key column with a special CONST default such as `DEFAULT CURRENT_TIMESTAMP` ever reached here, the sampling SQL would emit `CAST("CURRENT_TIMESTAMP" AS datetime)` (a literal-string cast -> likely NULL) rather than the evaluated value. I believe the `hasExprObject()` key-column guard in `buildRoutedAddKeyColumnJob` rejects such defaults first -- could you add a test confirming a `DEFAULT CURRENT_TIMESTAMP` key ADD is rejected (or handled)? The current suite only covers scalar constants and `uuid()` (VARY).

gensrc/proto/lake_types.proto
```diff
@@ -584,6 +584,12 @@ message TxnInfoPB {
     // anchored as op_schema_change@W. W == 1 means the base partition was empty at the watershed
     // (PARTITION_INIT_VERSION == 1), so there is no source op_write and BE synthesizes an empty one.
     optional int64 shadow_rewrite_alter_version = 13;
+    // TXN_SHADOW_REWRITE flip publish: true when the rewrite produced no source op_write because the base
+    // partition was empty at the watershed. FE sets it from the rewrite txn having zero tablet commit infos.
```
luohaha: Nit: this says FE sets the flag "from the rewrite txn having zero tablet commit infos," but the FE code actually derives it from `loadedRows == 0` -- because `tabletCommitInfos` is reset to null at VISIBLE (as the Java comment in `rewriteSourceEmpty` correctly explains). Please align this proto comment with the Java rationale so a future reader doesn't try to "fix" the FE to consult the (now-null) commit-info list.

fe/fe-core/src/main/java/com/starrocks/sql/InsertPlanner.java
xiangguangyxg: Fixed in `e0b826c`. The origin-absent branch now materializes a default only for a genuinely-new column — gated on `outputBaseSchema` having no committed column of that name — and otherwise `Preconditions.checkState` fails fast, restoring the invariant the old `checkState` enforced for the MODIFY path (a MODIFY shadow always retains its origin, so a missing one now throws instead of silently emitting a wrong DEFAULT). I used a hard guard rather than a `LOG.warn` here because this branch is hit on every concurrent double-write during a range ADD (the new column is materialized for each load), so a WARN would be log spam; the checkState only fires for the genuinely-corrupt case.

fe/fe-core/src/main/java/com/starrocks/alter/LakeOnlineRewriteJobBase.java
xiangguangyxg: Added a `LOG.warn` in `e0b826c` when the rewrite txn state is unavailable here. Agree on your framing: this loop runs the same tick `classifyRewrite` returned DONE (which requires the txn VISIBLE), so it is normally present, and a replay that finds it evicted re-runs the rewrite (`classifyRewrite` -> NEEDS_RUN) rather than reaching here with a null — so the null path is effectively unreachable and, if hit, degrades to the pre-PR flip-publish-retry (recoverable), never data loss. The warn just makes that rare case diagnosable instead of silent.

fe/fe-core/src/main/java/com/starrocks/alter/LakeRangeRewriteSchemaChangeJob.java
xiangguangyxg: Added `testAddKeyColumnWithCurrentTimestampDefault` in `e0b826c`. Confirmed behavior: `ADD COLUMN k_ts DATETIME KEY DEFAULT CURRENT_TIMESTAMP` is **handled** (routes + builds), not rejected — `hasExprObject()` is false for `CURRENT_TIMESTAMP`, so the key-column scalar-default guard does not reject it. It is safe: the added key column is constant across all historical rows (materialized to the evaluated txn time via `calculatedDefaultValue`), so it adds no distinction to the sort key and K collapses; and `projectSortKeyColumn` null-guards (`getDefaultValue() == null ? "NULL" : ...`), so the degenerate sampling literal cannot NPE. No boundary/correctness impact, as you noted.

gensrc/proto/lake_types.proto
xiangguangyxg: Fixed in `e0b826c` — aligned the proto comment with the Java rationale: FE derives the flag from `loadedRows == 0`, and the comment now explicitly notes it does NOT consult tablet commit infos (reset to null at the VISIBLE finish before FE reads it), pointing to the `rewriteSourceEmpty` comment so a future reader does not try to "fix" it to read the now-null list.

luohaha (approved):

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -4438,13 +4618,15 @@ private static boolean hasOnlyTypeChangeForVarcharLengthFastPath(Column oriColum
...
+     *       always shifts the range sort key), a {@link ModifyColumnClause} that flips a column's
+     *       keyness AND that flip shifts the range sort key, a {@link DropColumnClause} whose
+     *       column is in the range sort key, or an {@link AddColumnClause}/{@link AddColumnsClause}
+     *       that adds a key column joining a key-derived range sort key.</li>
      * </ul>
      *
-     * <p>Column-set-changing operations (add/drop key) remain out of scope and are rejected
-     * elsewhere; this predicate only decides between routing and the existing keyness-flip /
-     * sort-key-reorder rejections.
+     * <p>A drop or add that shifts the range sort key is now routed here; keysType filtering
```
banmoy, line 4627: **Gap: a full-column `ORDER BY` that permutes the key columns on a key-derived range table isn't routed here — it falls to the plain reorder path and keeps stale tablet boundaries.**

`needsRangeRewriteSchemaChange` is only consulted on the `changeSortKeyColumn(...) == true` branch. `changeSortKeyColumn` returns `orderedColumns.size() != baseSchema.size()`, so an `ORDER BY` that lists **all** columns returns false → the `else` branch → `processReorderColumn`, which rejects only `PRIMARY_KEYS` and has no range-distribution guard. (The guard lives in the sibling `processModifySortKeyColumn`, which only the subset path reaches.) So this predicate is never consulted for a full reorder.

Precondition — the table was created **without** an explicit `ORDER BY` (`sortKeyIdxes == null`), so the range sort key is derived from `isKey()` order. With an explicit `ORDER BY`, `calculateShortKey` remaps `sortKeyIdxes` by name and the sort key is preserved through a reorder, so there is no issue.

For such a table, `processReorderColumn` reorders the schema and `calculateShortKey` re-derives the range sort key from the new key order — so the range sort key **does** shift. But the plain `LakeTableSchemaChangeJob` copies each shadow tablet's range verbatim (`LakeTableAlterJobV2Builder`: `if (table.isRangeDistribution()) shadowTablet.setRange(originTablet.getRange())`) instead of re-sampling. After the flip the sort key is the new order while the stored boundary tuples are still in the old key-column space, so `RangeDistributionPruner` (FE) and `RangeRouter` (BE) compare new-order key tuples against old-order boundaries → silent wrong pruning and mis-routed loads, with no error.

Repro (lake, key-derived range table, ≥2 keys):
```sql
-- t(k1,k2,v1) DUPLICATE KEY(k1,k2), range distribution, no ORDER BY; loaded so tablets hold (k1,k2) boundaries
ALTER TABLE t ORDER BY (k2, k1, v1);   -- all columns, key order permuted
```
`ALTER TABLE t ORDER BY (k2)` (subset) is correctly routed to `LakeRangeRewriteSchemaChangeJob` and re-samples — only the full-column form slips through.

This is pre-existing (these lines aren't touched by this PR), but it's the one sort-key-shifting ALTER on a range table this PR's routing does not cover. Suggest routing it to the rewrite job (or rejecting it) when a full reorder changes the range key order, mirroring the subset path.

(Traced statically through the paths above; worth a quick e2e to confirm the wrong-routing symptom.)

xiangguangyxg: Confirmed — good catch, and your trace is correct: `changeSortKeyColumn` returns `orderedColumns.size() != baseSchema.size()`, so a full-column `ORDER BY` returns false and goes to `processReorderColumn` (which only rejects `PRIMARY_KEYS`, no range guard), landing on the plain `LakeTableSchemaChangeJob` that copies each shadow tablet range verbatim. On a key-derived range table that permutes the key order, the stored boundary tuples then no longer match the new sort-key order → silent wrong pruning/routing. The subset path is the only one that reaches `needsRangeRewriteSchemaChange`.

As you noted, this is pre-existing and untouched by this PR (which scopes to ADD/DROP of a key column); it lives in the reorder-routing lane. I will address the full-column-reorder case as a focused follow-up — routing it to `LakeRangeRewriteSchemaChangeJob` (re-sample) when the reorder shifts a key-derived range key, mirroring the subset path, with a lake e2e confirming boundaries are re-sampled. Filing it separately to keep this PR reviewable; I can link the follow-up here.

xiangguangyxg: Follow-up opened: #76256.

After analysis I scoped it to the **correctness fix** rather than the full route+re-sample: routing a full reorder would require a schema-reorder-aware rewrite job (a larger change), whereas the actual harm here is the *silent* boundary corruption. So #76256 rejects the corrupting case — a full-column `ORDER BY` that permutes the key columns of a key-derived (`getSortKeyIdxes() == null`) cloud-native range table — with a message pointing to the subset `ORDER BY (<sort-key cols>)` form (which already re-samples via the rewrite job). Safe cases are untouched: a full reorder that keeps the key order (value columns only), explicit-`ORDER BY` tables (remapped by name), and non-range tables. Tests cover both the reject and the value-only-allowed case. Full route+re-sample support can follow as an enhancement if desired.

banmoy (approved):

srlch (approved):

PR #76051, opened 2026-07-08
[Enhancement] Support IS NULL and <=> shapes in OR range derivation

eshishki:
## Why I'm doing:

`ScalarRangePredicateExtractor` derives single-column filters from OR-of-conjunctions predicates by unioning per-branch constraints, and predicate pushdown carries them to the scans for partition pruning and zone-map filtering:

```sql
(v1 = 1 AND v2 = 2) OR (v1 = 3 AND v2 = 4)
-- derives: v1 IN (1, 3), v2 IN (2, 4)
```

Two constraint shapes are invisible to it: `IS NULL` and `<=>` (EQ_FOR_NULL). A branch that constrains a column only with such a shape produces no descriptor for that column, so the column is not "constrained in every branch" and the whole derivation for it is dropped:

```sql
(v1 = 1 AND v2 = 2) OR (v1 IS NULL AND v2 = 4)
-- derives: v2 IN (2, 4), nothing for v1
-- implied and pushable: (v1 = 1) OR (v1 IS NULL)
```

Such shapes are common in BI-generated filters over nullable columns, and in plans produced by the MV transparent-union rewrite itself: the stale-partition compensation predicate is exactly `col IN (<stale days>) OR col IS NULL`.

### The mechanism

Each `ValueDescriptor` describes the set of values on which the source conjunct can evaluate to TRUE. This change adds one bit to that set: whether NULL belongs to it (`admitsNull`).

| conjunct | can be TRUE on | `admitsNull` |
|---|---|---|
| `E = x`, `E < x`, `E IN (...)` | some non-null values | no: a comparison on NULL is UNKNOWN, never TRUE |
| `E IS NULL` | {NULL} | yes |
| `E <=> NULL` | {NULL} | yes |
| `E <=> x` (non-null x) | {x} | no: on NULL it is FALSE |
| `(E IN (a,b)) OR (E IS NULL)` | {a, b, NULL} | yes |

How the flag combines:

| | value sets | NULL alternative | example |
|---|---|---|---|
| **OR** | union | allowed if **any** branch allows it | `col = 1 OR col IS NULL` → {1} + NULL |
| **AND** | intersection | allowed only if **every** conjunct allows it | `(col IN ('a','b') OR col IS NULL) AND col <= 'x'` → {a, b}, NULL gone: `col <= 'x'` is never TRUE on NULL |

When the intersection leaves nothing — no values and no NULL — the conjunction can never be TRUE and the scan folds to an empty set: `v1 <=> NULL AND v1 = 3`.

The derivation of the NULL alternative is gated by a new session variable `cbo_derive_predicate_null_alternative` (default true). With it disabled, `IS NULL` / `<=>` conjuncts produce no descriptors and the extractor behaves exactly as before.

### Why it is safe

The extractor emits in three modes, each with its own obligation, all satisfied by the set model above:

1. **Append** (the normal case): the derived conjunct is the intersection of the sets, so it holds on every row the original conjunction accepts. The original conjuncts stay verbatim; the derived one is additive.
2. **Whole-predicate replacement** (only for a pure OR of simple shapes): requires exact equivalence. On this path there is no AND, so nothing eliminates NULL — the emitted form is `(values) OR (col IS NULL)`, exactly the descriptor's set.
3. **Contradiction** (empty value set and no NULL alternative): the conjunction can never be TRUE, the scan folds to empty.

Two scope guarantees: the extractor runs only on filter predicates (join/scan predicate pushdown), where a non-TRUE result discards the row — so "never TRUE" is the only fact the derivation relies on; and descriptors are never collected under `NOT` (existing guard), where the polarity of the reasoning would flip.

### Plan-level effects

- Derived single-column IN/range filters now also appear for OR predicates with `IS NULL` / `<=>` branches. A plan can show a bare `l_shipdate IN ('1998-01-02','1998-01-05')` next to the original `(... IN ...) OR (... IS NULL)`: the bare IN is the jointly-implied fact in the simple form that partition pruning and join equivalence transfer consume; the OR form is not transferable across a join equality, the IN is.
- Contradicting conjunctions (`k1 > '2020-06-30' AND k1 IS NULL`) fold to an empty scan.
- In MV transparent-union rewrite, a compensation branch whose predicate is proven empty is pruned. Example from the updated tests: the compensation reads Hive days `{1998-01-02, 1998-01-05}`, the derived IN transfers through the join equality `l_shipdate = o_orderdate` to the other join side, whose partition domain contains no such dates — the compensation join is provably empty, the UNION disappears, and the query is answered from the MV alone. The proof uses partition metadata only, not data.

Affected plan tests updated: `FurtherPartitionPruneTest` (4 contradiction cases now expect an empty scan), `MvTransparentUnionRewriteHiveTest` (4 blocks with a pruned compensation branch, 2 expects with the extra derived IN), `MaterializedViewWithPartitionTest` (extra derived conjunct in a union-rewrite plan). A new SQL test (`test_null_predicate_derive`) verifies row-level results for the new shapes — including NULL rows on the preserved side of a left join, a full outer join with a never-true ON condition, and null-aware `NOT IN` subqueries whose result set contains NULLs — with the derivation enabled and disabled.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

PR #76041, opened 2026-07-08
[BugFix] Fix ASAN crash for BOOLEAN nested in Avro complex-type columns

banmoy:
## Why I'm doing:

Fix the daily ASAN crash reported in StarRocks/StarRocksTest#11522:

```
be/src/gutil/casts.h:78: To down_cast(From*) [with To = FixedLengthColumn<signed char>*; ...]:
Assertion `f == __null || dynamic_cast<To>(f) != __null' failed.
```

It aborts BE when an Avro routine load targets a native complex-type column (STRUCT/MAP/ARRAY) that contains a `BOOLEAN` field/element/value.

## What I'm doing:

`BooleanColumn` is `FixedLengthColumn<uint8_t>`, but the nested (non-adaptive) `add_nullable_column` in `be/src/formats/avro/nullable_column.cpp` dispatched `TYPE_BOOLEAN` through `add_nullable_numeric_column<int8_t>`. `add_numeric_column<int8_t>` then `down_cast`s the uint8 column to `FixedLengthColumn<int8_t>*`, which trips the `dynamic_cast` assert under ASAN/DEBUG.

The adaptive path (`add_adpative_nullable_column`, used for top-level columns) already dispatched `TYPE_BOOLEAN` through the correct `<uint8_t>` — which is why only complex types crashed and a plain top-level `BOOLEAN` column was fine. The fix dispatches the nested path through `<uint8_t>` too. `add_numeric_column<uint8_t>` is already explicitly instantiated, so no other change is needed.

Added regression test `AvroScannerTest.test_struct_with_boolean`; the full `AvroScannerTest` suite (26 tests) passes on a clean ASAN build.

Notes:
- The crash-reachable STRUCT/MAP paths were made reachable by #74901 (main only). The nested ARRAY path carries the same latent mismatch since #19866.
- In opt builds the assert is compiled out and `int8_t`/`uint8_t` are layout-identical, so for valid boolean data this is an ASAN/DEBUG crash rather than a release data-correctness issue.

Fixes StarRocks/StarRocksTest#11522

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
[...]

PR #76066, opened 2026-07-09
[BugFix] Fix ASAN crash for BOOLEAN nested in Avro complex-type columns (backport #76041)

banmoy:
## Why I'm doing:

Manual backport of #76041 to branch-4.0. The mergify auto-backport (#76065) could not apply: on main the avro scanner test lives at `be/test/exec/file_scanner/avro_scanner_test.cpp` and depends on the native STRUCT target reader from #74901, which is not in 4.0, so the test hunk produced a large rename/content conflict.

## What I'm doing:

`BooleanColumn` is `FixedLengthColumn<uint8_t>`, but the nested `add_nullable_column` in `be/src/formats/avro/nullable_column.cpp` dispatched `TYPE_BOOLEAN` through `add_nullable_numeric_column<int8_t>`. `add_numeric_column<int8_t>` then `down_cast`s the uint8 column to `FixedLengthColumn<int8_t>*`, tripping the `dynamic_cast` assert (`gutil/casts.h`) under ASAN/DEBUG when a BOOLEAN is nested inside a native complex-type target column of an Avro routine load (StarRocks/StarRocksTest#11522).

The one-line fix (dispatch `TYPE_BOOLEAN` through `<uint8_t>`, matching the adaptive path) is identical to #76041; `add_numeric_column<uint8_t>` is already explicitly instantiated on 4.0.

On branch-4.0 only the nested **ARRAY** path is reachable (native STRUCT/MAP target support came later on main via #74901), so the regression test is `AvroScannerTest.test_array_with_boolean` in `be/test/exec/avro_scanner_test.cpp` (4.0's path), using only helpers already present on 4.0. It drives the same `add_nullable_column` `TYPE_BOOLEAN` dispatch that the fix corrects.

Fixes StarRocks/StarRocksTest#11522

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/formats/avro/nullable_column.cpp
+++ b/be/src/formats/avro/nullable_column.cpp
@@ -138,7 +138,7 @@ static Status add_nullable_column(Column* column, const TypeDescriptor& type_des
                                   const avro_value_t& value) {
     switch (type_desc.type) {
     case TYPE_BOOLEAN:
-        return add_nullable_numeric_column<int8_t>(column, type_desc, name, value);
+        return add_nullable_numeric_column<uint8_t>(column, type_desc, name, value);
     case TYPE_BIGINT:
         return add_nullable_numeric_column<int64_t>(column, type_desc, name, value);
     case TYPE_INT:
--- a/be/test/exec/avro_scanner_test.cpp
+++ b/be/test/exec/avro_scanner_test.cpp
@@ -841,6 +841,66 @@ TEST_F(AvroScannerTest, test_array) {
     ASSERT_EQ(4294967298, array[1].get_int64());
 }
 
+// Regression for issue #11522: a BOOLEAN element nested inside a native ARRAY target column.
+// BooleanColumn is FixedLengthColumn<uint8_t>, but the nested add_nullable_column dispatched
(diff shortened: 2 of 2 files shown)


banmoy:
Closing: branch-4.0 does not need this backport.

Issue #11522 requires a genuine uint8 BooleanColumn to reach the nested add_nullable_column TYPE_BOOLEAN path. On 4.0 that never happens: native STRUCT/MAP targets are not supported (pre-#74901), and _construct_avro_types maps an ARRAY<BOOLEAN> element to VARCHAR, so the int8->uint8 dispatch fix has no effect here. CI confirmed the backported test aborts on 4.0 for an unrelated reason (element built as VARCHAR/BinaryColumn but filled as BOOLEAN), which is a distinct pre-existing ARRAY<BOOLEAN> mismatch, not #11522.

The fix stays on main (#76041) and 4.1 (#76064).

PR #76094, opened 2026-07-09
[Enhancement] Support load spill and eager PK-index SST for separate-sort-key cloud-native PK tables

luohaha:
## Why I'm doing:

Shared-data (cloud-native) primary key tables whose **sort key differs from the
primary key** currently disable both **load spill** and **eager PK-index SST
generation**. The root cause is not a memory-layout limitation but an ordering
conflict: the spill merge must order output by the sort key, so duplicate
primary keys are no longer physically adjacent and the aggregate iterator (which
dedups by PK on adjacent rows) cannot collapse them. As a result these tables
lose the memory relief of load spill and push the entire PK-index build onto the
serialized publish path.

## What I'm doing:

Enable load spill + eager PK-index SST for separate-sort-key PK tables, keeping
**1 segment : 1 SST**:

- **`PkTabletUnsortSSTWriter`** (new, in `pk_tablet_unsort_sst_writer.{h,cpp}`):
  buffers primary keys (which now arrive in sort-key order) in an in-memory
  btree, sorting them and collapsing duplicate PKs within a segment. The dedup
  winner is the **last-flushed** row, decided by a per-row ordering key
  (`slot_idx`); the loser's rowid is recorded into a per-segment delete vector.
  The SST stores only the rowid — `rssid`/`version` are still supplied by
  `shared_rssid`/`shared_version` at ingest, so the read/ingest path is
  **unchanged**. When the map reaches the memtable threshold (mirroring
  `is_memtable_full`) it spills to an intermediate SST; at segment finalize the
  intermediate SSTs are k-way merged into the single final SST, resolving
  cross-SST duplicate PKs and appending losers to the delvec.
- **slot_idx ordering channel**: `BlockGroupIterator` encodes `slot_idx` into the
  high 32 bits of `rssid_rowids`; the heap merge propagates it per row when
  `need_rssid_rowids` is set. The pipeline merge iterator disables aggregation
  and turns on `need_rssid_rowids` for separate-sort-key PK.
- **delvec plumbing**: the per-segment dedup delvec is carried in
  `OpWrite.seg_delvecs` (parallel to `ssts`) and folded into `new_deletes` at
  publish, so losing rows are masked together with historical-duplicate deletes.
- **correctness gate**: load spill for a separate-sort-key PK table is enabled
  only when eager PK-index build is supported (cloud-native index + PK-encoding
  preconditions); otherwise the sort-key-ordered merge would leave duplicate PKs
  to be resolved by the publish upsert with a wrong (sort-key-order) tie-break,
  so spilling stays disabled and the in-memory memtable path handles it.

This also applies to compaction of such tables (same writer path); compaction
input has no duplicate PKs, so the unsort writer there is a pure sort with no
delvec.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/delta_writer.cpp
+++ b/be/src/storage/lake/delta_writer.cpp
@@ -37,6 +37,7 @@
 #include "runtime/mem_tracker.h"
 #include "runtime/runtime_env.h"
 #include "storage/chunk_helper.h"
+#include "storage/del_vector.h"
 #include "storage/delta_writer.h"
 #include "storage/lake/filenames.h"
 #include "storage/lake/load_spill_pipeline_merge_context.h"
@@ -267,12 +268,14 @@ private:
 
     int64_t _max_buffer_size;
 
-    // Snapshot the in-transaction upsert/delete-order feature flag once for the whole load.
-    // lake_enable_pk_preserve_txn_delete_order is a mutable BE config, but it must not change
-    // mid-load: it drives both whether the spill path keeps the __op column (fixing LoadChunkSpiller's
-    // serde/schema from the first chunk) and whether finish() emits the del_op_offsets array. Reading
-    // it live in each place could tear a single load across the legacy and op-aware paths.
-    const bool _preserve_txn_delete_order = config::lake_enable_pk_preserve_txn_delete_order;
+    // Effective "preserve in-transaction upsert/delete order" for this load, snapshotted ONCE (in
+    // build_schema_and_writer, once the schema is known) via pk_preserve_txn_delete_order_enabled(): the
+    // config OR the separate-sort-key load-spill path (which always needs op-aware ordering). It must not
+    // change mid-load: it drives both whether the spill path keeps the __op column (fixing
+    // LoadChunkSpiller's serde/schema from the first chunk) and whether finish() emits del_op_offsets.
+    // Reading the mutable configs live in each place could tear a single load across the legacy and
+    // op-aware paths.
+    bool _preserve_txn_delete_order = false;
 
     std::unique_ptr<TabletWriter> _tablet_writer;
     std::unique_ptr<MemTable> _mem_table;
@@ -401,15 +404,23 @@ const DictColumnsValidMap* DeltaWriterImpl::global_dict_columns_valid_info() con
 //    WHY: Condition merge requires reading old values during ingestion, which conflicts with spilling
 // 2. Partial updates are involved
 //    WHY: Partial updates need to merge with existing data, requiring all columns in memory
-// 3. Separate sort keys are used
-//    WHY: Sort key handling requires special memory layout incompatible with spilling
+// 3. Separate sort keys are used AND eager PK-index build is NOT supported for this schema
+//    Note: separate-sort-key tables normally DO spill -- the spill orders the merge output by the
+//    sort key (primary keys are not adjacent), so duplicate primary keys are resolved by the unsort
+//    SST writer, which runs under eager build. Spilling is disabled only for the narrow case where
+//    eager build is unavailable (V1 single non-VARCHAR/CHAR key, or metadata not yet cached), because
+//    there is then no unsort SST writer to resolve dups and lazy publish rebuild would tie-break by
+//    sort-key order instead of flush order. In that case the load falls back to the non-spill memtable
+//    path, which is correct: for PK tables the memtable sorts by primary key and removes duplicates
+//    first, then re-sorts by the sort key (see MemTable::_sort).
 //
 // For non-PK tables: Always allow spilling if globally enabled (no special constraints)
 bool DeltaWriterImpl::should_enable_load_spill() const {
     return config::enable_load_spill &&
            (_tablet_schema->keys_type() != KeysType::PRIMARY_KEYS ||
             ((config::ignore_merge_condition_inside_same_transaction || _merge_condition.empty()) &&
-             !is_partial_update() && !_tablet_schema->has_separate_sort_key()));
+             !is_partial_update() &&
+             (!_tablet_schema->has_separate_sort_key() || pk_index_eager_build_supported(*_tablet_schema))));
 }
 
 Status DeltaWriterImpl::build_schema_and_writer() {
@@ -418,6 +429,8 @@ Status DeltaWriterImpl::build_schema_and_writer() {
         ASSIGN_OR_RETURN([[maybe_unused]] auto tablet, _tablet_manager->get_tablet(_tablet_id));
         RETURN_IF_ERROR(init_tablet_schema());
         RETURN_IF_ERROR(init_write_schema());
+        // Snapshot the effective preserve-order decision once, now that the schema is known.
+        _preserve_txn_delete_order = pk_preserve_txn_delete_order_enabled(*_tablet_schema);
         if (_tablet_schema->keys_type() == KeysType::PRIMARY_KEYS) {
             _tablet_writer = std::make_unique<HorizontalPkTabletWriter>(_tablet_manager, _tablet_id, _write_schema,
                                                                         _txn_id, nullptr, false /** no compaction**/,
@@ -432,6 +445,12 @@ Status DeltaWriterImpl::build_schema_and_writer() {
         }
         RETURN_IF_ERROR(_tablet_writer->open());
         if (should_enable_load_spill()) {
+            // Eager PK-index build (the unsort SST writer that a separate-sort-key spill load needs to
+            // dedup non-adjacent duplicate primary keys) is enabled at merge time by try_enable, which
+            // fires for any spilled load and, for separate-sort-key tables, regardless of size
+            // (merge_blocks_to_segments / init_parallel_merge). A single flush never spills and takes the
+            // direct write_single_flush* path, which does not need it. So there is no build-time force
+            // here; the decision follows the data volume (whether the load actually spills).
             if (_load_spill_block_mgr == nullptr || !_load_spill_block_mgr->is_initialized()) {
                 // Pass txn_id to LoadSpillBlockManager so that spill files are placed under
                 // the flat layout <tablet_root>/load_spill_txns/<txn_id_hex>_<load_id>_<frag_id>_<seq>,
@@ -845,11 +864,13 @@ StatusOr<TxnLogPtr> DeltaWriterImpl::finish_with_txnlog(DeltaWriterFinishMode mo
         // Carry the per-del op_offset (parallel to dels_meta, index by del_id) so the apply/persist
         // path can preserve in-transaction upsert/delete ordering. A kUnknownDelOpOffset entry (spill /
         // concurrent flush) keeps the array aligned and is read as "not recorded" -> max segment id.
-        // Downgrade-safety gate: only emit the array when explicitly enabled. While disabled (the
-        // default), it stays empty so apply/persist/rebuild uniformly fall back to the legacy "delete
-        // after all segments" path -- this never writes the new on-disk state that a pre-fix BE
-        // (rollback / not-yet-upgraded node / cross-version OpReplication target) would misread into a
-        // duplicate primary key. See config::lake_enable_pk_preserve_txn_delete_order.
+        // Only emitted when the effective preserve-order snapshot is on (pk_preserve_txn_delete_order_enabled
+        // -- the config, which defaults on, OR any separate-sort-key table). The separate-sort-key path
+        // MUST emit it: its parallel merge can split a key's DELETE and later re-UPSERT across tasks, and
+        // without the offsets the earlier del file would erase the later re-insert. This is new on-disk
+        // state (del_op_offsets) that a pre-feature BE (rollback / not-yet-upgraded node / cross-version
+        // OpReplication target) would misread; its read side must be deployed fleet-wide before relying on
+        // it (see the seg_delvecs read-side backport).
         if (_preserve_txn_delete_order) {
             for (size_t i = 0; i < del_idx; ++i) {
                 op_write->add_del_op_offsets(i < del_op_offsets.size() ? del_op_offsets[i] : kUnknownDelOpOffset);
@@ -861,6 +882,24 @@ StatusOr<TxnLogPtr> DeltaWriterImpl::finish_with_txnlog(DeltaWriterFinishMode mo
     }
     for (auto& sst_range : _tablet_writer->sst_ranges()) {
         op_write->add_sst_ranges()->CopyFrom(sst_range);
+    }
+    // Per-segment dedup delete vectors from the unsort SST writer (separate-sort-key PK). Kept
+    // parallel to `ssts` by index; an entry with no data means that segment had no dedup losers.
+    // The bitmap version is a placeholder here; publish stamps it with the real tablet version.
+    // Only the separate-sort-key path can leave intra-segment duplicate-PK losers, so only it emits
+    // seg_delvecs: the common sort-key==PK path would otherwise append one empty entry per SST (and hand
+    // a pre-feature publisher an unknown field for no reason). Publish guards missing indices as empty,
+    // so emitting nothing here is equivalent to emitting all-empty entries.
+    if (_tablet_schema->has_separate_sort_key()) {
+        for (const auto& seg_del_rowids : _tablet_writer->seg_delvecs()) {
+            auto* seg_delvec_pb = op_write->add_seg_delvecs();
+            if (!seg_del_rowids.empty()) {
+                DelVector dv;
+                dv.init(0 /* version stamped at publish */, seg_del_rowids.data(), seg_del_rowids.size());
+                seg_delvec_pb->set_version(0);
+                seg_delvec_pb->set_data(dv.save());
+            }
+        }
     }
     op_write->mutable_rowset()->set_num_rows(_tablet_writer->num_rows());
     op_write->mutable_rowset()->set_data_size(_tablet_writer->data_size());
--- a/be/src/storage/lake/load_spill_pipeline_merge_iterator.cpp
+++ b/be/src/storage/lake/load_spill_pipeline_merge_iterator.cpp
@@ -40,6 +40,14 @@ LoadSpillPipelineMergeIterator::LoadSpillPipelineMergeIterator(LoadChunkSpiller*
     // to maintain correctness (combine duplicate keys, apply aggregate functions).
     if (schema->keys_type() == KeysType::DUP_KEYS) {
         _do_agg = false;
+    } else if (schema->keys_type() == KeysType::PRIMARY_KEYS &&
+               _parent_writer->tablet_schema()->has_separate_sort_key()) {
+        // Separate sort key: the merge orders output by the sort key, so primary keys are NOT
+        // adjacent and the aggregate iterator cannot collapse duplicate PKs on the right axis.
+        // Disable aggregation and instead carry a per-row ordering key (rssid_rowids) so the unsort
+        // SST writer resolves duplicate PKs by last-flushed-wins and records losers into a delvec.
+        _do_agg = false;
+        _need_rssid_rowids = true;
     } else {
         _do_agg = true;
     }
@@ -59,9 +67,9 @@ Status LoadSpillPipelineMergeIterator::_generate_next_task() {
     // - true (sort): Whether to sort during merge (always true for correctness)
     // - _do_agg: Whether to perform aggregation (depends on table keys type)
     // - _final_round: Whether this merges to final tablet vs intermediate blocks
-    ASSIGN_OR_RETURN(auto batch, _spiller->generate_merge_input_batch(config::load_spill_max_merge_bytes,
-                                                                      config::load_spill_memory_usage_per_merge,
-                                                                      true /*sort*/, _do_agg, _final_round));
+    ASSIGN_OR_RETURN(auto batch, _spiller->generate_merge_input_batch(
+                                         config::load_spill_max_merge_bytes, config::load_spill_memory_usage_per_merge,
+                                         true /*sort*/, _do_agg, _final_round, _need_rssid_rowids));
 
     // nullptr merge_itr indicates no more block groups available - iteration complete
     if (batch.merge_itr != nullptr) {
@@ -84,7 +92,7 @@ Status LoadSpillPipelineMergeIterator::_generate_next_task() {
         auto task = std::make_unique<LoadSpillMergeInputBatch>(std::move(batch));
         _current_task = std::make_shared<lake::TabletInternalParallelMergeTask>(
                 std::move(writer), std::move(task), _spiller->schema().get(), _quit_flag,
-                _spiller->spiller()->metrics().write_io_timer, _op_aware);
+                _spiller->spiller()->metrics().write_io_timer, _op_aware, _need_rssid_rowids);
     } else {
         // No more data to merge - signal iteration completion by returning nullptr.
         // Pipeline operators check has_more() which tests for nullptr to know when to stop.
--- a/be/src/storage/lake/pk_tablet_sst_writer.h
+++ b/be/src/storage/lake/pk_tablet_sst_writer.h
@@ -33,13 +33,39 @@ class PersistentIndexSstableStreamBuilder;
 class DefaultSSTWriter {
 public:
     virtual ~DefaultSSTWriter() = default;
-    virtual Status append_sst_record(const Chunk& data) { return Status::OK(); }
+    // `rssid_rowids`, when non-null, carries one per-row ordering key aligned with `data`'s rows.
+    // Its high 32 bits encode the source memtable flush order (slot_idx); writers that must resolve
+    // duplicate primary keys by "last-flushed wins" (see PkTabletUnsortSSTWriter) use it. Writers
+    // whose input is already in primary-key order (PkTabletSSTWriter) ignore it.
+    // `column_indexes`, when non-null, gives the tablet column id of each column in `data` (used by
+    // the vertical-compaction key group, whose chunk carries [sort-key columns, then PK columns]).
+    // The unsort writer uses it to locate the PK columns; other writers ignore it because `data`
+    // already has the PK columns at positions [0, num_key_columns).
+    virtual Status append_sst_record(const Chunk& data, const std::vector<uint64_t>* rssid_rowids = nullptr,
+                                     const std::vector<uint32_t>* column_indexes = nullptr) {
+        return Status::OK();
+    }
+    // Feed DELETE rows to writers that resolve op order themselves (the separate-sort-key unsort
+    // writer). Default is a no-op: other writers receive deletes via the caller's del-file path.
+    virtual Status append_delete_records(const Chunk& data, const std::vector<uint64_t>* rssid_rowids,
+                                         const std::vector<uint32_t>* column_indexes = nullptr) {
+        return Status::OK();
+    }
     virtual Status reset_sst_writer(const std::shared_ptr<LocationProvider>& location_provider,
                                     const std::shared_ptr<FileSystem>& fs) {
         return Status::OK();
     }
     virtual StatusOr<std::pair<FileInfo, PersistentIndexSstableRangePB>> flush_sst_writer() { return Status::OK(); }
     virtual bool has_file_info() const { return false; }
+    // Rowids (within the just-flushed segment) that lost primary-key dedup and must be masked by a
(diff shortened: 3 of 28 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/lake/update_manager.cpp
```diff
@@ -396,7 +396,23 @@ Status UpdateManager::publish_primary_key_tablet(const TxnLogPB_OpWrite& op_writ
...
+        // and must be masked by its delete vector. _do_update below appends historical-duplicate
+        // deletes on top via push_back, so this seed is preserved.
+        if (local_id < op_write.seg_delvecs_size() && !op_write.seg_delvecs(local_id).data().empty()) {
+            const auto& seg_delvec_pb = op_write.seg_delvecs(local_id);
+            DelVector writer_dv;
+            RETURN_IF_ERROR(
+                    writer_dv.load(metadata->version(), seg_delvec_pb.data().data(), seg_delvec_pb.data().size()));
+            auto& seg_deletes = new_deletes[rssid];
+            for (const auto rowid : *writer_dv.roaring()) {
+                seg_deletes.push_back(rowid);
```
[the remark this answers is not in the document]
luohaha: Fixed. `init_new_deletes` now sort+dedups each `new_deletes` vector before the `cur_old + cur_add == cur_new` cardinality check, gated on `op_write.seg_delvecs_size() > 0` (the separate-sort-key unsort path where one incoming segment can hold duplicate PKs). See update_manager.cpp (`dedupe_new_deletes`).

be/src/storage/lake/tablet_writer.cpp
```diff
@@ -20,35 +20,41 @@
...
+    // Eager PK index build only supports the cloud-native persistent index. When metadata is not yet
+    // cached we optimistically allow it (matches the historical try_enable behavior).
     if (metadata != nullptr) {
-        // Eager PK index build only supports cloud native pk index.
         if (!metadata->enable_persistent_index() ||
             metadata->persistent_index_type() != PersistentIndexTypePB::CLOUD_NATIVE) {
-            return;
+            return false;
         }
     }
```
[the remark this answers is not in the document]
luohaha: Fixed. `pk_index_eager_build_supported` now returns false for a separate-sort-key schema when `metadata == nullptr` (a cache miss no longer optimistically enables the path). The sort-key==PK path keeps its historical optimistic behavior. Since the gate lives in this shared helper, it covers both `DeltaWriterImpl::should_enable_load_spill` and `TabletWriter::try_enable_pk_index_eager_build`.

be/src/storage/lake/delta_writer.cpp
```diff
@@ -855,6 +863,18 @@ StatusOr<TxnLogPtr> DeltaWriterImpl::finish_with_txnlog(DeltaWriterFinishMode mo
     for (auto& sst_range : _tablet_writer->sst_ranges()) {
         op_write->add_sst_ranges()->CopyFrom(sst_range);
     }
+    // Per-segment dedup delete vectors from the unsort SST writer (separate-sort-key PK). Kept
+    // parallel to `ssts` by index; an entry with no data means that segment had no dedup losers.
+    // The bitmap version is a placeholder here; publish stamps it with the real tablet version.
+    for (const auto& seg_del_rowids : _tablet_writer->seg_delvecs()) {
+        auto* seg_delvec_pb = op_write->add_seg_delvecs();
```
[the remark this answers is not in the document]
luohaha: Fixed. Added `config::lake_enable_separate_sort_key_load_spill` (default **false**), mirroring `lake_enable_pk_preserve_txn_delete_order`. The whole separate-sort-key spill+eager path is gated on it in `pk_index_eager_build_supported`, so when off these tables fall back to the pre-feature load path and never emit `seg_delvecs` (field 22). Enable only after the cluster is fully upgraded and no rollback to a pre-fix BE is expected. Documented in the shared-data lake BE parameters (en/zh/ja).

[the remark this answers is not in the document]
luohaha: Update: per author decision the feature is now **enabled by default** rather than gated — the `lake_enable_separate_sort_key_load_spill` config has been removed (`4c8f2f9`). The `seg_delvecs` (field 22) is therefore always emitted on this path; the rolling-upgrade downgrade caveat is accepted for the 4.1.4 rollout (the eager unsort SST + field-22 masking ship together in the same version).

CelerData-Reviewer:
@codex review

be/src/storage/lake/load_spill_pipeline_merge_iterator.cpp
```diff
@@ -40,6 +40,14 @@ LoadSpillPipelineMergeIterator::LoadSpillPipelineMergeIterator(LoadChunkSpiller*
...
     if (schema->keys_type() == KeysType::DUP_KEYS) {
         _do_agg = false;
+    } else if (schema->keys_type() == KeysType::PRIMARY_KEYS &&
+               _parent_writer->tablet_schema()->has_separate_sort_key()) {
+        // Separate sort key: the merge orders output by the sort key, so primary keys are NOT
+        // adjacent and the aggregate iterator cannot collapse duplicate PKs on the right axis.
+        // Disable aggregation and instead carry a per-row ordering key (rssid_rowids) so the unsort
+        // SST writer resolves duplicate PKs by last-flushed-wins and records losers into a delvec.
+        _do_agg = false;
+        _need_rssid_rowids = true;
```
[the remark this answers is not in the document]
luohaha: Fixed in `087b1a7`. The unsort SST writer is now op-aware: DELETE rows are fed via `append_pk_index_deletes` and reconciled against the upserts by `rssid_rowid` order (latest op per PK wins), so aggregation is no longer required to preserve DELETE/UPSERT order. A PK whose latest op is a DELETE is written as a persistent-index tombstone; a DELETE superseded by a later UPSERT is dropped. Added UT `test_unsort_sst_writer_op_aware_reconcile`.

be/src/storage/lake/pk_tablet_unsort_sst_writer.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+Status PkTabletUnsortSSTWriter::append_sst_record(const Chunk& data, const std::vector<uint64_t>* rssid_rowids,
+                                                  const std::vector<uint32_t>* column_indexes) {
+    if (_wf == nullptr) {
+        return Status::InternalError("unsort pk sst writer not initialized");
+    }
+    const size_t num_rows = data.num_rows();
+    if (rssid_rowids == nullptr || rssid_rowids->size() != num_rows) {
+        return Status::InternalError(fmt::format("unsort pk sst writer requires per-row order keys, expected {} got {}",
+                                                 num_rows, rssid_rowids == nullptr ? 0 : rssid_rowids->size()));
```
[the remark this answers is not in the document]
luohaha: Fixed in `087b1a7`. `append_records` now returns early for an empty chunk, so the 0-row anchor write no longer trips the per-row-order-key check. The op-aware path still writes the 0-row anchor for a delete-only batch so the eager SST stays 1:1 with segments.

be/src/storage/lake/delta_writer.cpp
```diff
@@ -862,6 +870,18 @@ StatusOr<TxnLogPtr> DeltaWriterImpl::finish_with_txnlog(DeltaWriterFinishMode mo
...
+        if (!seg_del_rowids.empty()) {
+            DelVector dv;
+            dv.init(0 /* version stamped at publish */, seg_del_rowids.data(), seg_del_rowids.size());
+            seg_delvec_pb->set_version(0);
+            seg_delvec_pb->set_data(dv.save());
```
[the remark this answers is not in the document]
luohaha: Per author decision the feature is enabled by default (the `lake_enable_separate_sort_key_load_spill` gate was removed). The eager unsort SST + the `seg_delvecs`/tombstone masking ship together in the same version (4.1.4), so a same-version publisher always understands both; the rolling-upgrade caveat vs a pre-feature BE is accepted for this rollout.

CelerData-Reviewer:
@codex review

be/src/storage/lake/delta_writer.cpp
```diff
@@ -401,15 +402,25 @@ const DictColumnsValidMap* DeltaWriterImpl::global_dict_columns_valid_info() con
...
 // For non-PK tables: Always allow spilling if globally enabled (no special constraints)
 bool DeltaWriterImpl::should_enable_load_spill() const {
     return config::enable_load_spill &&
            (_tablet_schema->keys_type() != KeysType::PRIMARY_KEYS ||
             ((config::ignore_merge_condition_inside_same_transaction || _merge_condition.empty()) &&
-             !is_partial_update() && !_tablet_schema->has_separate_sort_key()));
+             !is_partial_update() &&
+             (!_tablet_schema->has_separate_sort_key() ||
+              pk_index_eager_build_supported(*_tablet_schema,
+                                             _tablet_manager->get_latest_cached_tablet_metadata(_tablet_id)))));
```
luohaha: **TOCTOU between this spill gate and eager-build enablement can silently produce wrong data.**

This check runs once at writer setup with `get_latest_cached_tablet_metadata()`. Eager build is enabled *later* (`LoadSpillPipelineMergeContext::init_parallel_merge()` / `merge_blocks_to_segments()`), and `pk_index_eager_build_supported()` re-fetches metadata there. If the cache entry is evicted mid-load, the second fetch returns null, the separate-sort-key branch returns false, and the writer silently falls back to `DefaultSSTWriter`. Consequences:

- duplicate PKs reach publish unresolved and are tie-broken in sort-key order (wrong dedup);
- worse, the op-aware branch in `write_one_merged_chunk` feeds DELETE rows to `append_delete_records`, which is a **no-op** on `DefaultSSTWriter` — deletes are silently dropped.

Suggestion: latch the decision — call `force_set_enable_pk_index_eager_build()` on the writer at the moment spill is enabled for a separate-sort-key table, or return `InternalError` in the merge path when `has_separate_sort_key() && !enable_pk_index_eager_build()` so this becomes a load failure instead of silent corruption.

gensrc/proto/lake_types.proto
```diff
@@ -399,6 +399,13 @@ message TxnLogPB {
         // (see config::lake_enable_pk_preserve_txn_delete_order). The persisted per-del order
         // lives on DelfileWithRowsetId.op_offset, derived from this at apply/persist time.
         repeated uint32 del_op_offsets = 21;
+        // Parallel to `ssts` (index by in-rowset segment ordinal): per-segment delete vector produced
+        // by the unsort SST writer for separate-sort-key PK tables, masking the rows that lost
+        // primary-key dedup inside the segment (their winner is what the SST points to). Each entry is
+        // self-contained (version + serialized roaring bitmap). Empty/absent for the sort-key==PK eager
+        // path (no intra-segment duplicates). At publish these rowids are folded into the segment's
+        // delete vector together with historical-duplicate deletes.
+        repeated DelvecDataPB seg_delvecs = 22;
```
luohaha, line 431: **Rolling-upgrade window: an old BE/CN publishing this txn log drops `seg_delvecs`.**

Field 22 is unknown to older binaries, so a non-upgraded node publishing a txn log written by an upgraded node would skip the dedup-loser masking entirely and expose duplicate rows. This requires write-on-new / publish-on-old (possible after failover or rescheduling in shared-data). The PR is marked "no behavior change" and has no version gating — at minimum this deserves a note in the PR description; ideally gate the new path on a config operators flip after the fleet is fully upgraded.

be/src/storage/lake/update_manager.cpp
```diff
@@ -334,6 +334,121 @@ static DelInterleavePlan build_del_interleave_plan(const TxnLogPB_OpWrite& op_wr
...
+    // rows -- they are only masked by a delete vector, not removed. `_do_update(read_only=true)` then
+    // reads that segment's PK column (duplicates included) and `parallel_get` looks up every row, so a
+    // duplicate PK that ALSO exists in a prior version has its historical rowid pushed into
+    // `new_deletes` once per occurrence. Left as-is, `cur_add` (the raw vector size, e.g. old_rowid
+    // twice) would exceed the deduplicated bitmap cardinality and trip the
+    // `cur_old + cur_add != cur_new` consistency check below. Deduplicate the rowids here so the count
+    // matches the bitmap; the delvec content is unchanged (init/add_dels build a roaring bitmap that
+    // dedups anyway). Gated on the unsort path to avoid the sort cost on the common sort-key==PK path,
+    // whose segments never carry duplicate PKs.
+    const bool dedupe_new_deletes = op_write.seg_delvecs_size() > 0;
```
luohaha: **This gate fires for *all* eager-SST loads, not just the unsort path.**

`finish_with_txnlog` adds a `seg_delvecs` entry (empty `DelvecDataPB`) for **every** SST — including the sort-key==PK streaming path — because `flush_segment_writer` unconditionally does `_seg_delvecs.emplace_back(take_deleted_rowids())`. So `seg_delvecs_size() > 0` is true for every eager-build load, contradicting the comment above ("Gated on the unsort path to avoid the sort cost on the common sort-key==PK path"). Functionally harmless (dedupe is idempotent), but the sort cost lands on the common path anyway and every eager load carries N empty proto entries in the txn log.

Suggestion: gate on "any entry has non-empty `data()`" (precise: a segment with intra-segment duplicates always records losers, so its delvec is non-empty), or fix the comment.

luohaha (commented):
Reviewed the full diff. The core algorithm (unsort SST writer + slot_idx ordering channel + per-segment dedup delvec) looks correct, and the publish-side delvec cardinality regression test shows good rigor. Left inline comments; the ones I'd prioritize before merge are: (1) latching the eager-build decision to close the metadata-cache TOCTOU window (silent wrong dedup + dropped DELETEs), (2) an end-to-end DELETE test through spill+publish, and (3) making the E2E test data non-monotonic so sort-key order actually differs from PK order.

be/test/storage/lake/pk_tablet_sst_writer_test.cpp
```diff
@@ -984,4 +985,413 @@ INSTANTIATE_TEST_SUITE_P(PkEncodingType, PkTabletSSTWriterBigintKeyTest,
...
+    ASSERT_EQ(dels.size(), 1u);
+    EXPECT_EQ(dels[0], 0u);
+}
+
+// Op-aware reconciliation (separate-sort-key spill path): DELETE rows are fed via append_delete_records
+// and reconciled against the upserts by rssid_rowid order (latest op per PK wins). An UPSERT that lost
+// to a later DELETE (or a later UPSERT) is masked by the delete vector; a DELETE superseded by a later
+// UPSERT contributes nothing. (A PK whose latest op is a DELETE is written into the SST as a
+// persistent-index tombstone; that on-disk encoding is covered by code review.)
+TEST_F(PkTabletSSTWriterTest, test_unsort_sst_writer_op_aware_reconcile) {
```
luohaha, line 1100: **Test gap: no end-to-end DELETE test through spill + publish.**

This test covers writer-level reconciliation only. The tombstone SST encoding → ingest → publish-erase chain, the 0-row anchor segment for delete-only batches, and cross-batch DELETE/UPSERT ordering are all untested end-to-end — this is the highest-risk untested path in the PR. Suggest an E2E case on `PkTabletUnsortWriterE2ETest` that loads upserts + deletes (op-aware) with spill forced, publishes, and verifies both row count and surviving values.

be/test/storage/lake/pk_tablet_sst_writer_test.cpp
```diff
@@ -984,4 +985,413 @@ INSTANTIATE_TEST_SUITE_P(PkEncodingType, PkTabletSSTWriterBigintKeyTest,
...
+        CHECK_OK(_tablet_mgr->put_tablet_metadata(*_tablet_metadata));
+    }
+    void TearDown() override { remove_test_dir_ignore_error(); }
+
+    Chunk generate_data(int64_t chunk_size, int64_t start_index = 0) {
+        auto c0 = BinaryColumn::create();
+        auto c1 = Int32Column::create();
+        for (int i = 0; i < chunk_size; i++) {
+            c0->append(fmt::format("key_{:10d}", start_index + i));
+            c1->append((start_index + i) * 3);
```
luohaha: **The E2E data is blind to two classes of bugs.**

`c1 = i*3` is ascending and the space-padded `key_{:10d}` PKs also sort numerically, so sort-key order **equals** PK order in every E2E test — the "keys arrive out of PK order" premise is never actually exercised (e.g. a bug that used map-iteration position instead of the stored rowid would still pass). Also, duplicate rows are byte-identical (same `i` → same key and value), so picking the wrong dedup winner is invisible to row-count assertions.

Suggest: generate `c1` descending (or shuffled) so sort-key order ≠ PK order, and give duplicate PKs distinct values that are verified on read-back.

be/src/storage/lake/pk_tablet_writer.cpp
```diff
@@ -101,7 +116,13 @@ Status HorizontalPkTabletWriter::reset_segment_writer(bool eos) {
     // reset sst file writer
     if (_pk_sst_writer == nullptr) {
         if (enable_pk_index_eager_build()) {
-            _pk_sst_writer = std::make_unique<PkTabletSSTWriter>(tablet_schema(), _tablet_mgr, _tablet_id);
+            if (tablet_schema()->has_separate_sort_key()) {
+                // Separate sort key: primary keys arrive in sort-key (not PK) order, so buffer+sort
+                // them in the unsort writer instead of streaming into a sorted SST.
+                _pk_sst_writer = std::make_unique<PkTabletUnsortSSTWriter>(tablet_schema(), _tablet_mgr, _tablet_id);
```
luohaha, line 190: **Test gap:** horizontal compaction of a separate-sort-key table with eager build also routes through the unsort writer via this branch (`HorizontalPkTabletWriter::write(data, rssid_rowids)` with real rssid|rowid as the order key), but only the vertical path is covered by `test_vertical_compaction_orderby_eager_sst`. Worth a horizontal-compaction E2E case too.

be/src/storage/lake/pk_tablet_unsort_sst_writer.cpp
```diff
@@ -0,0 +1,367 @@
...
+    auto* update_mgr = _tablet_mgr->update_mgr();
+    if (update_mgr != nullptr && update_mgr->mem_tracker() != nullptr &&
+        update_mgr->mem_tracker()->limit_exceeded_by_ratio(config::memory_urgent_level) &&
+        _map_mem_usage >= static_cast<size_t>(config::l0_min_mem_usage)) {
+        return true;
+    }
+    return false;
+}
+
+Status PkTabletUnsortSSTWriter::flush_map_to_intermediate_sst() {
```
luohaha, line 196: **Perf note:** intermediate spill SSTs are written to the tablet's `sst_location`, which on shared-data is **remote object storage** — each map overflow adds an OSS PUT here, a GET in `merge_intermediates_into`, and an async DELETE at flush. Fine as a first cut, but the load-path latency cost is real; the perf validation should include a run that actually overflows `l0_max_mem_usage` to measure it.

be/src/storage/lake/pk_tablet_unsort_sst_writer.cpp
```diff
@@ -0,0 +1,367 @@
...
+        if (pos == column_indexes.size()) {
+            return Status::InternalError(
+                    fmt::format("unsort pk sst writer: primary-key column {} missing from key group", pk_col));
+        }
+        out->append_column(data.get_column_by_index(pos), static_cast<SlotId>(pk_col));
+    }
+    return Status::OK();
+}
+
+bool PkTabletUnsortSSTWriter::is_map_full() const {
```
luohaha, line 176: **Memory sizing note:** each cloned parallel-merge-task writer owns its own unsort map, so peak memory is up to `l0_max_mem_usage` (100 MB default) × N concurrent merge tasks before spilling kicks in. It is charged to the merge mem tracker, but it's a notable addition to a feature whose purpose is memory relief — worth mentioning in the class header comment.

be/src/storage/lake/pk_tablet_sst_writer.h
```diff
@@ -33,34 +33,64 @@ class PersistentIndexSstableStreamBuilder;
...
-    virtual Status append_sst_record(const Chunk& data) { return Status::OK(); }
+    // `rssid_rowids`, when non-null, carries one per-row ordering key aligned with `data`'s rows.
+    // Its high 32 bits encode the source memtable flush order (slot_idx); writers that must resolve
+    // duplicate primary keys by "last-flushed wins" (see PkTabletUnsortSSTWriter) use it. Writers
+    // whose input is already in primary-key order (PkTabletSSTWriter) ignore it.
+    // `column_indexes`, when non-null, gives the tablet column id of each column in `data` (used by
+    // the vertical-compaction key group, whose chunk carries [sort-key columns, then PK columns]).
+    // The unsort writer uses it to locate the PK columns; other writers ignore it because `data`
+    // already has the PK columns at positions [0, num_key_columns).
+    virtual Status append_sst_record(const Chunk& data, const std::vector<uint64_t>* rssid_rowids = nullptr,
```
luohaha, line 44: Nit: default arguments on virtual functions bind statically (to the static type of the pointer), which is fragile if base/override defaults ever diverge. They're consistent here, but consider dropping the defaults on the overrides (or on the interface) to remove the trap.

be/src/storage/lake/pk_tablet_writer.cpp
```diff
@@ -65,6 +66,13 @@ Status HorizontalPkTabletWriter::write(const Chunk& data, SegmentPB* segment, bo
     return Status::OK();
 }
 
+Status HorizontalPkTabletWriter::append_pk_index_deletes(const Chunk& data, const std::vector<uint64_t>& rssid_rowids) {
+    // DELETE rows are only fed to the eager PK-index SST writer (to resolve the latest op per key);
+    // they are NOT written to a segment. Only the unsort SST writer acts on them; other SST writers
+    // no-op (deletes reach them via the caller's del file instead).
+    return _pk_sst_writer->append_delete_records(data, &rssid_rowids);
```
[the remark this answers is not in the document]
luohaha: Fixed in dcd05f4. `HorizontalPkTabletWriter::append_pk_index_deletes` now lazily initializes the segment/SST writer when `_seg_writer == nullptr` (the same path `write()` takes), so a delete-first or delete-only merge chunk sets it up before feeding deletes; guarding on `_seg_writer == nullptr` means the following upsert `write()` reuses the same writer instead of resetting and dropping the deletes. Confirmed by an end-to-end mixed upsert/delete load on a cluster (which reproduced the SIGSEGV in `tablet_par_mrg` before the fix). Note the winning deletes now go to a del file rather than an SST tombstone (86d572b).

be/test/storage/lake/pk_tablet_sst_writer_test.cpp
luohaha: Added in f78f219: `PkTabletUnsortWriterE2ETest.test_load_spill_op_aware_delete` -- an op-aware upsert+delete load through DeltaWriter -> spill -> publish -> read-back, including DELETEs of rows written by an EARLIER transaction (the del-file path). Note the tombstone->ingest->publish-erase chain no longer exists: 86d572b routes winning DELETEs to a del file instead of an SST tombstone (a tombstone shadowed the key in the index but never masked the pre-existing segment row). The 0-row anchor for a delete batch and historical-delete masking are now covered end to end.

be/test/storage/lake/pk_tablet_sst_writer_test.cpp
luohaha: Fixed in f78f219: the new delete E2E test's sort key DECREASES as the PK increases, so sort-key order is the reverse of PK order (the out-of-PK-order premise is actually exercised), and updated rows get a distinct new value that is verified on read-back -- so a wrong dedup winner is visible, not hidden behind byte-identical dups.

be/src/storage/lake/pk_tablet_unsort_sst_writer.cpp
luohaha: Documented in 107916a: added a Memory note to the class header -- peak usage is up to `l0_max_mem_usage` x concurrent merge tasks before spilling, charged to the load-spill merge mem tracker.

be/src/storage/lake/pk_tablet_unsort_sst_writer.cpp
luohaha: Acknowledged. Honest caveat: the current UT/E2E data does not overflow `l0_max_mem_usage`, so the intermediate-SST spill + `merge_intermediates_into` path (remote PUT/GET/async-DELETE) is not yet exercised by a committed test. The perf validation will include a run that lowers `l0_max_mem_usage` (or loads enough) to actually overflow and measure the remote-IO cost; a dedicated UT for the overflow path is worth adding too.

be/src/storage/lake/pk_tablet_writer.cpp
luohaha: Added in 80d4b98: `test_horizontal_compaction_orderby_eager_sst` -- forces horizontal compaction (all columns in one group) of the separate-sort-key table and asserts the eager SST is built, the row count is preserved, and the built index is PK-keyed (re-upsert dedups instead of growing). Mirrors the vertical case.

be/src/storage/lake/delta_writer.cpp
luohaha: I will make cloud native pk index become the only choice so we don't need to get latest cached tablet metadata any more:
https://github.com/StarRocks/starrocks/pull/76260

CelerData-Reviewer:
@codex review

be/src/storage/lake/spill_mem_table_sink.cpp
```diff
@@ -54,12 +54,30 @@ SpillMemTableSink::~SpillMemTableSink() {
...
+    // preserve_txn_delete_order snapshot. That writer resolves duplicate primary keys by flush order in
+    // its own per-PK reconciliation (a DELETE loses to a later re-UPSERT of the same key), and routes
+    // only winning deletes to a by-key del file. The legacy path taken when this returns false
+    // (_split_upserts_deletes -> flush_chunk_with_deletes) instead splits deletes off at the memtable
+    // BEFORE reconciliation and writes them to a del file applied "after all segments", which erases a
+    // key re-upserted in a LATER flush -- silent data loss for any delete-then-reinsert straddling a
+    // memtable-flush boundary. Forcing op-awareness here fixes that; seg_delvecs still emit and
+    // del_op_offsets stay gated off (preserve is still false), so on-disk state / downgrade safety are
+    // unchanged.
+    return _keep_op_column || writer_needs_order_keys();
```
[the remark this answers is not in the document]
luohaha: Fixed in 6a0d1fc. Root cause: the separate-sort-key unsort SST writer ALWAYS resolves duplicate PKs by flush order (it cannot do the legacy delete-after-upserts), so this path must always run with preserve semantics. Centralized that into `pk_preserve_txn_delete_order_enabled(schema)` = config OR (lake_enable_separate_sort_key_load_spill AND separate sort key), snapshotted once per load into `_preserve_txn_delete_order`. It now gates BOTH keep_op_column() and the del_op_offsets serialization, so del_op_offsets ARE emitted for the separate-sort-key path -- a DELETE and a later re-UPSERT that split across parallel merge tasks now resolve by flush order (the earlier del file gets its real op_offset, not 'after all segments'). Downgrade-safe: the path is opt-in behind lake_enable_separate_sort_key_load_spill (same gate as seg_delvecs). Confirming with a long-run harness whose load_spill_max_merge_bytes is lowered so small batches fan out into multiple final-merge tasks, exercising exactly this cross-task path.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

luohaha:

PR #76098, opened 2026-07-09
[BugFix] Fix LoadChunkSpiller init race that crashes BE during load spill

sevev:
## Why I'm doing:

`LoadChunkSpiller::_prepare()` lazily creates the spiller and its serde, but used
`_spiller == nullptr` as its "already initialized" flag **without any lock**. Multiple
memtable-flush threads share a single `LoadChunkSpiller` and call `spill()` concurrently
on the first spill (see the parallel-flush comment on `LoadChunkSpiller::spill`).

Because `_spiller` is assigned *before* the serde's encode context is created, a racing
thread could observe a non-null `_spiller`, skip initialization, and call
`ColumnarSerde::serialize()` while the encode context was still null. `serialize()` then
dereferenced the null context in `_get_encode_levels()` and crashed the BE/CN with
`SIGSEGV`. Observed on a shared-data (CN) production cluster during load spill:

```
PC: starrocks::spill::ColumnarSerde::serialize(...)
*** SIGSEGV (@0x50) received ... ***
    ColumnarSerde::serialize
    LoadChunkSpiller::_do_spill
    LoadChunkSpiller::spill
    lake::SpillMemTableSink::flush_chunk
    MemTable::flush
    FlushToken::_flush_memtable
    MemtableFlushTask::run
```

## What I'm doing:

- Serialize the one-time initialization in `_prepare()` with a `std::mutex`, and gate
  readiness on a separate `std::atomic<bool> _prepared` that is stored with release
  ordering **only after** the spiller and its serde are fully constructed; readers load
  it with acquire ordering. Readiness is no longer inferred from `_spiller != nullptr`.
- Build the spiller on a local and publish it via `std::move`, so a half-initialized
  spiller is never observable through `_spiller`.
- Defense in depth: `ColumnarSerde::serialize()` now returns an error instead of
  dereferencing a null encode context, mirroring the existing null guard in
  `_max_serialized_size()`.
- Add a regression test that drives many threads through the first spill of a fresh
  `LoadChunkSpiller` and asserts all concurrent spills succeed.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/compute_env/load_spill/load_chunk_spiller.cpp
+++ b/be/src/compute_env/load_spill/load_chunk_spiller.cpp
@@ -147,39 +147,39 @@ StatusOr<size_t> LoadChunkSpiller::spill(const Chunk& chunk, int64_t slot_idx) {
 }
 
 Status LoadChunkSpiller::_prepare(const ChunkPtr& chunk_ptr) {
-    // Fast path: already fully initialized. The acquire load pairs with the release store
-    // below, so a thread that observes `true` is guaranteed to see the fully-constructed
-    // `_spiller` and its prepared serde.
-    if (_prepared.load(std::memory_order_acquire)) {
-        return Status::OK();
-    }
-    // Slow path: serialize the one-time initialization. Multiple memtable-flush threads may
-    // reach here concurrently on the first spill of this LoadChunkSpiller.
-    std::lock_guard<std::mutex> l(_init_lock);
-    if (_prepared.load(std::memory_order_relaxed)) {
-        return Status::OK();
-    }
-    // Build everything on a local first, so a half-initialized spiller is never published
-    // through `_spiller`. The previous code assigned `_spiller` before preparing its serde,
-    // which let a racing thread observe a non-null spiller whose serde encode context was
-    // still null and crash in ColumnarSerde::serialize().
-    spill::SpilledOptions options;
-    options.encode_level = 7;
-    // Leave options.wg unset (nullptr): the spill framework resolves it to the reserved
-    // default workgroup in Spiller::prepare(), so this load path no longer needs ExecEnv.
-    auto spiller = _spiller_factory->create(options);
-    RETURN_IF_ERROR(spiller->prepare(_runtime_state.get()));
-    DCHECK(_profile != nullptr) << "LoadChunkSpiller profile is null";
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

be/src/compute_env/load_spill/load_chunk_spiller.cpp
```diff
@@ -147,23 +147,38 @@ StatusOr<size_t> LoadChunkSpiller::spill(const Chunk& chunk, int64_t slot_idx) {
...
-        _spiller->set_metrics(metrics);
-        // 2. prepare serde
-        if (const_cast<spill::ChunkBuilder*>(&_spiller->chunk_builder())->chunk_schema()->empty()) {
-            const_cast<spill::ChunkBuilder*>(&_spiller->chunk_builder())->chunk_schema()->set_schema(chunk_ptr);
-            RETURN_IF_ERROR(_spiller->serde()->prepare());
-        }
+    // Fast path: already fully initialized. The acquire load pairs with the release store
+    // below, so a thread that observes `true` is guaranteed to see the fully-constructed
+    // `_spiller` and its prepared serde.
+    if (_prepared.load(std::memory_order_acquire)) {
```
luohaha: Use `std::call_once(_load_once_flag...` instead?

sevev: Done in 7308d39 — switched `_prepare()` to `std::call_once`, dropping the manual `std::mutex` + `std::atomic<bool>`. The init body still builds the spiller on a local and publishes it via `std::move` only after its serde is prepared, and the outcome is cached in `_prepare_status` so a failed init is reported to every concurrent flush thread instead of leaving a null/half-built spiller reachable. `LoadChunkSpillerTest` (14 cases, incl. the concurrent-first-spill regression) passes under ASAN.

CelerData-Reviewer:
@codex review

luohaha (approved):

meegoo (approved):

sevev:
ignore backport check: 4.0

PR #76104, opened 2026-07-09
[BugFix] Fix range-colocate null-safe join dropping matches under bucket-shuffle

xiangguangyxg:
## Why I'm doing:

A range-distribution colocate join on a **null-safe** key (`a <=> b`) that cannot
colocate — `disable_colocate_join=true`, or the two tables in different colocate
groups — was planned as `FULL/INNER/... JOIN (BUCKET_SHUFFLE(S))` with the **range
table as the local/stay side**. Bucket shuffle re-buckets the other side with
crc32, which does not match the range tablet layout, so **every match was silently
dropped — including non-NULL equal keys** (e.g. `10 <=> 10`), turning the join into
an all-NULL-padded result.

The same query with `=`, or `<=>` with colocate enabled (`COLOCATE` plan), or on
plain hash tables (`PARTITIONED`), all returned correct results — only the range
bucket-shuffle `<=>` path was wrong.

Root cause (FE optimizer): a range scan natively satisfies a hash `SHUFFLE_JOIN`
required property (`RangeDistributionSpec.isSatisfyHashShuffle`) — this is what
enables the legitimate `COLOCATE` plan. When the join cannot colocate,
`ChildOutputPropertyGuarantor` normalized each range child via
`convertRangeToHashShuffle` (SourceType `SHUFFLE_JOIN`). For a `<=>` key the
required distribution is already null-strict, so the `enforceNullStrict` inside
that helper is a no-op: the enforced `SHUFFLE_JOIN[null-strict]` matches exactly
what the un-exchanged range scan is cached as satisfying, so plan extraction
discards the shuffle exchange and reuses the range scan — a `SHUFFLE_HASH_BUCKET`
join with a range stay side. A plain `=` key is null-relaxed, so `enforceNullStrict`
shifts the property to null-strict (which the range scan was not costed for),
forcing the exchange — which is why the bug was `<=>`-specific.

## What I'm doing:

A range-distributed scan must never be the local (stay) side of a bucket-shuffle
join. In `ChildOutputPropertyGuarantor.visitPhysicalJoin`, when both children are
range but `canRangeColocateJoin` is false, force a full **PARTITIONED** shuffle on
both sides via `enforceChildShuffleDistribution` (SourceType `SHUFFLE_ENFORCE`,
which a range scan does **not** natively satisfy) and return, instead of
`convertRangeToHashShuffle`. This guarantees the shuffle exchange is materialized
on both sides regardless of the key's null-strictness.

Unchanged:
- Range-colocate joins that **can** colocate still plan as `COLOCATE` (the
  `canRangeColocateJoin` early return runs first) — for `<=>` as well as `=`.
- Range × hash joins still use bucket-shuffle with the **hash** table as the stay
  side (a valid crc32 stay side), via the existing single-range-child path.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
@@ -562,9 +562,12 @@ public class ChildOutputPropertyGuarantor extends PropertyDeriverBase<Void, Expr
 
         // Range-colocate dispatch:
         // - both children range + canRangeColocateJoin → skip exchange (return).
-        // - otherwise, normalize any range child to hash SHUFFLE_JOIN via
-        //   convertRangeToHashShuffle, then fall through to the existing
-        //   hash/hash compatibility path unchanged.
+        // - both children range but NOT colocate → force a full (PARTITIONED)
+        //   shuffle on both sides and return (see below).
+        // - exactly one range child (range × hash) → normalize the range child to
+        //   hash SHUFFLE_JOIN via convertRangeToHashShuffle, then fall through to the
+        //   existing hash/hash compatibility path unchanged (the hash side is a valid
+        //   crc32 bucket-shuffle stay side).
         DistributionSpec leftRawSpec = leftChildOutputProperty.getDistributionProperty().getSpec();
         DistributionSpec rightRawSpec = rightChildOutputProperty.getDistributionProperty().getSpec();
         boolean leftIsRange = leftRawSpec instanceof RangeDistributionSpec;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
### Cluster e2e (shared-data, 1 FE + 3 CN)

Reproduced the bug and verified the fix on a shared-data cluster with two range-distribution colocate tables (`order by(k1)`, same `colocate_with` group, `k1` nullable), 1000 matched keys + 1 left-only + 1 right-only + 1 NULL each. Signature = counts over `cl FULL OUTER JOIN cr ON cl.k1 <=> cr.k1`.

**Before (current main, build of `5cc7b51`)** — `disable_colocate_join=true`:
- plan: `FULL OUTER JOIN (BUCKET_SHUFFLE(S))`
- `matched_nonnull=0`, `left_only=1000`, `right_only=1000` — **all matches dropped** (even non-NULL `10 <=> 10`).

**After (this PR, build of `d08a569`)** — `disable_colocate_join=true`:
- plan: `FULL OUTER JOIN (PARTITIONED)`
- `total=1002, matched_nonnull=999, left_only=1, right_only=1, null_null=1` — **correct**, byte-identical to the `COLOCATE` oracle (`disable_colocate_join=false`) and to the `=` result.

Genuine range-colocate `<=>` joins still plan as `COLOCATE` (unchanged); `=` joins are unchanged. FE plan regression test `RangeColocateNullSafeJoinPlanTest` covers `<=>` under `disable_colocate_join`, different colocate groups, and the preserved colocate case.

fe/fe-core/src/main/java/com/starrocks/sql/optimizer/ChildOutputPropertyGuarantor.java
```diff
@@ -562,9 +562,12 @@ public Void visitPhysicalJoin(PhysicalJoinOperator node, ExpressionContext conte
 
         // Range-colocate dispatch:
         // - both children range + canRangeColocateJoin → skip exchange (return).
-        // - otherwise, normalize any range child to hash SHUFFLE_JOIN via
-        //   convertRangeToHashShuffle, then fall through to the existing
-        //   hash/hash compatibility path unchanged.
+        // - both children range but NOT colocate → force a full (PARTITIONED)
```
kevincai, line 565: still use non-ascii chars in the comments. suggest add a rule in the AI memory, don't write comments with non-ascii chars.

kevincai (approved):

srlch (approved):

PR #76106, opened 2026-07-09
[BugFix] Surface local-disk (datacache) write bytes/time in lake compaction stats

luohaha:
## Why I'm doing:

The lake compaction profile (`SHOW PROC '/compactions'` and the BE-log statistics) reports **0 bytes written** even when a compaction writes many segments — e.g. observed `write_segment_count=76` with `write_*_mb=0` and `write_*_sec=0` on a compaction that read 72&nbsp;GB.

Root cause: with a write-back **datacache**, compaction output lands on **local disk first** and is uploaded to remote storage **asynchronously**, so `bytes_write_remote` / `io_ns_write_remote` are 0 at write time. The writer and compaction stats only surfaced the *remote* write counters, so the write side of the profile is effectively invisible — you can't tell how much was written or how long writing took, which makes it impossible to account for a compaction's wall-clock time.

The `StarletOutputStream` io stats already track `bytes_write_local_disk` / `io_ns_write_local_disk`, and the `kBytesWriteLocalDisk` / `kIONsWriteLocalDisk` stat-name constants already exist — they were just never exported by the writable file's `get_numeric_statistics()`, so nothing downstream (writer stats → compaction profile) could observe them.

## What I'm doing:

Plumb the local-disk write bytes/time through end to end (purely additive, observability only — no behavior change):
- `StarletOutputStream::get_numeric_statistics()`: also `append(kIONsWriteLocalDisk, …)` and `append(kBytesWriteLocalDisk, …)` (reserve 4).
- `OlapWriterStatistics` + `collect_writer_stats()`: add `bytes_write_local_disk` / `write_local_disk_ns` and collect them.
- `CompactionTaskStats`: carry the two fields (with matching `operator+`/`operator-`) and emit `write_local_mb` / `write_local_sec` in the compaction profile JSON.

After this, a datacache-backed compaction that writes N segments reports the real local write volume/time instead of `0`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

(Observability only — adds two profile fields; no execution-path change.)

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

## Verification
Verifying on a shared-data cluster with datacache: before, a compaction writing 76 segments reports `write_*_mb=0`; after, the profile reports `write_local_mb` = the real bytes written. (On-cluster verification in progress; will comment the before/after profile.)

### DIFF
--- a/be/src/compute_env/staros/starlet_filesystem.cpp
+++ b/be/src/compute_env/staros/starlet_filesystem.cpp
@@ -303,9 +303,15 @@ public:
         }
         const auto& io_stats = (*stream_st)->get_io_stats();
         auto stats = std::make_unique<io::NumericStatistics>();
-        stats->reserve(2);
+        stats->reserve(4);
         stats->append(kIONsWriteRemote, io_stats.io_ns_write_remote);
         stats->append(kBytesWriteRemote, io_stats.bytes_write_remote);
+        // Also surface local-disk (datacache) writes: with a write-back datacache the bytes land on
+        // local disk first and are uploaded asynchronously, so bytes_write_remote stays 0 at write
+        // time. Without this the compaction/writer profile reports 0 bytes written despite writing
+        // real segments.
+        stats->append(kIONsWriteLocalDisk, io_stats.io_ns_write_local_disk);
+        stats->append(kBytesWriteLocalDisk, io_stats.bytes_write_local_disk);
         return std::move(stats);
     }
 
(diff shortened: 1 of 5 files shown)


luohaha:
Closing — this PR is based on a wrong premise. There is no write-back datacache mode, so compaction output is written directly to remote storage; `bytes_write_remote` should therefore be non-zero and the real bug is that it is reported as 0, NOT that a local-write counter is missing. Surfacing `bytes_write_local_disk` (a read-side datacache-fill counter) would not fix the observed 0. Re-diagnosing the actual cause of `bytes_write_remote=0`.

PR #76108, opened 2026-07-09
[BugFix] Report actual finalized segment bytes as lake compaction write volume

luohaha:
## Why I'm doing it

The cloud-native (lake) compaction profile reports **`write_remote_mb = 0`** even when the compaction writes real output segments. This makes the compaction profile / `show proc '/compactions'` useless for judging how much data a compaction actually wrote — the write side of the IO accounting is silently blank.

Root cause: `CompactionTaskStats::collect(const OlapWriterStatistics&)` sources the written volume from `OlapWriterStatistics::bytes_write_remote`. That field is filled from the underlying output stream's numeric counter (`kBytesWriteRemote`), which is **not populated on every write path** — for shared-data segment writes it stays `0`. So the profile reports 0 bytes despite flushing real segments.

Meanwhile the tablet writer already knows the ground-truth size: `SegmentWriter::finalize()` returns the finalized `segment_size`, which is accumulated into `_data_size` and stored on each `SegmentFileInfo`. That value is authoritative (it is the actual on-storage segment size) and is available at every flush site.

## What I'm doing

- Add `OlapWriterStatistics::bytes_written` — the sum of finalized `segment_size` values, the ground truth for bytes written.
- Accumulate it at each segment flush site alongside the existing `_data_size += segment_size` (general horizontal/vertical writers + PK writer).
- In `CompactionTaskStats::collect(...)`, prefer `bytes_written`, falling back to the stream counter `bytes_write_remote` when the segment size is unavailable, so no path regresses.

Net effect: the compaction profile's `write_remote_mb` now reflects the real bytes written.

## What type of PR is this

- [x] BugFix

- [x] Yes — an observability field that previously read 0 now reports the real value. No functional/data-path change.

### DIFF
--- a/be/src/storage/lake/compaction_task_context.cpp
+++ b/be/src/storage/lake/compaction_task_context.cpp
@@ -40,7 +40,11 @@ void CompactionTaskStats::collect(const OlapReaderStatistics& reader_stats) {
 
 void CompactionTaskStats::collect(const OlapWriterStatistics& writer_stats) {
     write_segment_count = writer_stats.segment_count;
-    write_segment_bytes = writer_stats.bytes_write_remote;
+    // Prefer the actual finalized segment bytes (ground truth from SegmentWriter::finalize()). The
+    // underlying stream's bytes_write_remote counter is not populated on every write path (it stays
+    // 0 e.g. for shared-data segment writes), which made the profile report 0 bytes written despite
+    // writing real segments. Fall back to the stream counter if the segment size is unavailable.
+    write_segment_bytes = writer_stats.bytes_written > 0 ? writer_stats.bytes_written : writer_stats.bytes_write_remote;
     io_ns_write_remote = writer_stats.write_remote_ns;
 }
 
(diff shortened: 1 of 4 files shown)


luohaha:
### Verified on a live shared-data cluster

Deployed this build to a 3-CN shared-data cluster and compared cloud-native (lake) compactions of the **same PK table** across the restart onto the fixed binary, reading `write_remote_mb` from the compaction stats (`vertical_compaction_task.cpp` finish log + `show proc '/compactions'`):

| build | txn | write segs | input bytes | `write_remote_mb` |
|---|---|---|---|---|
| before fix | 89 | 4 | 11.9 GB | **0** |
| before fix | 89 | 15 | 15.4 GB | **0** |
| before fix | 1100 | 2 | 1.07 GB | **0** |
| before fix | 1210 | 2 | 4.4 GB | **0** |
| **with fix** | 2039 | 1 | 315 MB | **161** |
| **with fix** | 2039 | 1 | 314 MB | **155** |

Before: every compaction reported `write_remote_mb = 0` regardless of how much was actually written (the `bytes_write_remote` starlet counter is not populated on the shared-data segment-write path). After: it reports the real finalized segment volume. No functional change — observability only.

PR #76109, opened 2026-07-09
[Enhancement] Pre-place IVM refresh __op as a fixed output column; recognize PK-target MV at refresh

Youngwb:
## Why I'm doing:

At IVM refresh, the `__op` sink control column (UPSERT/DELETE) for a PRIMARY KEY target materialized view was created **inside the optimizer** (`IvmRewriter.appendPkLoadOpColumn`), *after* `InsertPlanner` had already frozen the sink's output columns — so on the refresh path the optimizer-produced `__op` never reached the sink. Append-only PK-target MVs (aggregate / dedup `QUERY_COMPUTED` MVs) stayed correct only because a PRIMARY KEY table sink defaults to UPSERT, so the dropped `__op` was harmless there. Separately, `isPrimaryKeyTargetMv` read `optimizerContext.getStatement()`, which is not the `InsertStmt` at refresh, so it failed to recognize a PK-target MV at refresh.

## What I'm doing:

- Pre-place `__op` as a fixed trailing output column in `InsertPlanner` **before** optimize, so it flows into the sink tuple by position like any other column. `IvmRewriter` binds its value to `__ACTION__` instead of creating a new column ref (with a fallback that creates one when `__op` was not pre-placed, so a direct insert and `IvmRewriterTest` are unaffected). `bindStateColumnsForAggregate` excludes the pre-placed `__op` so the positional MV-state binding stays aligned.
- Fix `isPrimaryKeyTargetMv` to resolve the target MV via `loadTargetMv` (the tvr context), matching `bindStateColumnsForAggregate`, so a PK-target MV is recognized at refresh.

This routes the refresh `__op` through the normal positional `outputColumns`/tuple machinery, aligning the create-time and refresh paths. Append-only MV results are unchanged — the refresh plan now carries an explicit `__op` = UPSERT column that was previously dropped and defaulted at the sink.

Covered by the existing `IVMBasedMvRefreshProcessorIcebergTest` (refresh plan structure) and `IvmRewriterTest` (append-only `__op` bind/fallback); both green. Also validated end-to-end on a cloud-native cluster: append-only IVM refresh (dup projection / join / union / aggregate forms) and PK projection refresh (insert / value-update / filter move-out-and-back / delete / change-PK), MV equals base with diff=0.

Fixes #issue: N/A — internal IVM refresh correctness/refactor, no change to MV results.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/InsertPlanner.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/InsertPlanner.java
@@ -113,6 +113,7 @@ import com.starrocks.sql.plan.ExecPlan;
 import com.starrocks.sql.plan.PlanFragmentBuilder;
 import com.starrocks.thrift.TPartialUpdateMode;
 import com.starrocks.thrift.TResultSinkType;
+import com.starrocks.type.IntegerType;
 import com.starrocks.type.NullType;
 import com.starrocks.type.Type;
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

Astralidea (approved):

PR #69617, opened 2026-02-28
[BugFix] Guard against iterator UB in get_column_values when rssid not found

luohaha:
## Why I'm doing:

When the PrimaryIndex contains stale entries pointing to rowsets that have been compacted away, `TabletUpdates::get_column_values` crashes with SIGSEGV. The root cause is that `upper_bound(rssid)` returns `begin()` when the requested rssid is smaller than all keys in `rssid_to_rowsets`, and then `--iter` on `begin()` is undefined behavior. In practice this causes the iterator to point to the map's sentinel node, leading to dereference of corrupted memory.

Core dump analysis confirmed:
- `rssid = 432750` from PrimaryIndex lookup
- `rssid_to_rowsets` min key = `476512` (old rowsets already compacted)
- `--begin()` UB → dereference of corrupted `RowsetMetaPB` pointer `0x80804509150b1880` → SIGSEGV

## What I'm doing:

Add a bounds check before decrementing the `upper_bound` iterator. When `iter == begin()`, return `InternalError` with a descriptive message instead of crashing.

## What type of PR is this:
- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Does this PR entail a change in behavior?
- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

## If yes, please specify the type of change:
- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:
- [x] I have added test cases for my bug fix or my new feature
- [ ] This PR needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
- [ ] This is a backport PR

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the PR will be auto-backported to the target branch
  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

PR #76119, opened 2026-07-09
[BugFix] Fix BE crash in primary key auto-increment partial update apply

wyb:
## Why I'm doing:

For a primary-key table with an AUTO_INCREMENT column, a partial update that
omits that column applies through SegmentRewriter::rewrite_auto_increment, which
rebuilds the full segment from three sources: the keys / user-provided columns
re-read from the partial segment, the auto-increment column, and the missing
columns read from history.

rewrite_auto_increment swallowed the status of the partial-segment read: the
iterator was taken with StatusOr::value() (no ok() check) and get_next's status
was dropped (only a DCHECK guarded the row count, compiled out in release). A
transient read failure left read_chunk short, so the rebuilt segment recorded a
segment num_rows of 0 while its value columns still held N rows of real data.
The rowset meta and primary index kept the commit-time counts, so the
inconsistency stayed dormant until a later partial update of the same keys read
that segment via get_column_values, which silently skipped the 0-row segment,
returned an empty read column, and dereferenced a null source in
append_selective / SIMDGather (src=0x0) -- a crash that replays the pending
apply on every restart.

Shared-nothing only; the lake path (rewrite_auto_increment_lake)
already checks the read.

## What I'm doing:

Root cause (write side), be/src/storage/rowset/segment_rewriter.cpp:
- rewrite_auto_increment no longer swallows read errors: create the iterator
  with ASSIGN_OR_RETURN (an iterator-creation failure now returns cleanly
  instead of aborting in value()), propagate get_next's status, and always
  validate the read row count against the segment, with the expected/actual
  counts and segment path in the error. This mirrors rewrite_auto_increment_lake.
  Also closes a latent crash where itr->close() ran on a possibly-null iterator.

Read side (stops the crash on already-corrupt data),
be/src/storage/tablet_updates.cpp:
- get_column_values returns Status::Corruption instead of silently skipping when
  a resolved rssid points to a 0-row segment (both the main read loop and the
  auto-increment default block). Corruption is non-retryable, so the tablet
  enters error state (recoverable via clone) rather than an endless crash-loop.
  Generalizes #69617, which only guarded the rssid-underflow / out-of-range case.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/storage/rowset/segment_rewriter.cpp
+++ b/be/src/storage/rowset/segment_rewriter.cpp
@@ -165,14 +165,25 @@ Status SegmentRewriter::rewrite_auto_increment(const std::string& src_path, cons
     seg_options.chunk_size = num_rows;
     seg_options.temporary_data = true;
 
-    auto res = rowset->segments()[segment_id]->new_iterator(src_schema, seg_options);
-    auto& itr = res.value();
+    ASSIGN_OR_RETURN(auto itr, rowset->segments()[segment_id]->new_iterator(src_schema, seg_options));
 
     if (itr) {
         auto st = itr->get_next(read_chunk);
-        DCHECK_EQ(read_chunk->num_rows(), num_rows);
+        itr->close();
+        // Do NOT swallow the read error: a transient read failure (page crc / decompress / io) here would
+        // otherwise leave read_chunk short, and the downstream append_chunk would silently emit a segment whose
+        // key columns have fewer rows than the value columns (segment num_rows=0 while value columns=N).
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

sevev (approved):

PR #76121, opened 2026-07-09
[BugFix] Detect range-colocate joins in plan-feedback colocate detection

banmoy:
## Why I'm doing:

`JoinTuningGuide.isColocateJoin` (the plan-feedback / Plan Advisor tuning guide) only recognized
classic hash-`LOCAL` colocate joins. A range-colocate join carries `RangeDistributionSpec`
(`DistributionType.RANGE_LOCAL`) required properties, whose `isShuffle()` is `false`, so it was
misclassified as **non-colocate**. As a result the Plan Advisor silently skipped the colocate
build/probe-swap tuning guide for range-colocate joins — no wrong results and no crash (the
`isShuffle()` guard short-circuits before the `HashDistributionSpec` cast), just a missed
optimization.

This "is a required property a colocate join?" check was hand-rolled in three consumers
(`PlanFragmentBuilder`, `SPMPlan2SQLBuilder`, `JoinTuningGuide`), each with its own
`instanceof` / `SourceType` matching. When range-colocate join was added (#71860) the first two were
updated and this third one in the plan-feedback subsystem was missed — duplicated derived logic
drifting out of sync.

## What I'm doing:

Fix the detection and remove the duplication that caused it: extract the check into a single helper
`DistributionSpecHelper.supportColocate(DistributionSpec)` — a required property is colocate when its
spec is hash-`LOCAL` (classic colocate) or a `RangeDistributionSpec` (range colocate) — and route all
three consumers through it. Range-colocate joins are now correctly detected, and the check lives in
one place so it cannot drift again.

The behavior delta is confined to the range-colocate path, which is gated behind the off-by-default
experimental `enable_range_distribution` flag — no change under any default/released configuration.

Tests: `DistributionSpecHelperTest#supportColocate*` covers detection per spec (range / hash-`LOCAL` /
shuffle / other). Verified green together with the planner (`RangeColocateMixedJoinPlanTest`) and SPM
(`SPMPlan2SQLTest`) consumers that route through the helper.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed
- [ ] Miscellaneous: gated behind the off-by-default experimental `enable_range_distribution`; the Plan Advisor can now emit the colocate build/probe-swap guide for range-colocate joins

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5
 
### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/feedback/guide/JoinTuningGuide.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/feedback/guide/JoinTuningGuide.java
@@ -19,8 +19,7 @@ import com.starrocks.sql.ast.JoinOperator;
 import com.starrocks.sql.optimizer.OptExpression;
 import com.starrocks.sql.optimizer.base.DistributionProperty;
 import com.starrocks.sql.optimizer.base.DistributionSpec;
-import com.starrocks.sql.optimizer.base.HashDistributionDesc;
-import com.starrocks.sql.optimizer.base.HashDistributionSpec;
+import com.starrocks.sql.optimizer.base.DistributionSpecHelper;
 import com.starrocks.sql.optimizer.base.PhysicalPropertySet;
 import com.starrocks.sql.optimizer.operator.physical.PhysicalDistributionOperator;
 import com.starrocks.sql.optimizer.operator.physical.PhysicalHashJoinOperator;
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg (approved):

stephen-shelby (approved):

PR #76123, opened 2026-07-09
[BugFix] Bump table optimistic version on tablet split so concurrent queries re-plan

luohaha:
## Why I'm doing:

During a tablet reshard (split) on a shared-data RANGE PK table, installing the new materialized indexes changes the partition's tablet layout. A query planned concurrently captures the old layout: `OlapScanNode.computeScanRangeLocations()` fills `scanTabletIds` from `PhysicalPartition.getLatestIndex()`, then `mapTabletsToPartitions()` re-reads `getLatestIndex()`. The split installs a new `MaterializedIndex` between the two reads, so a `scanTabletId` (a just-created split child) is absent from the rebuilt map and planning throws:

```
Invalid tablet id: '<id>'. The tablet may have been dropped. Selected partitions:[...]
```

`StatementPlanner`'s planning-retry loop should re-plan on concurrent metadata change, but its guard `OptimisticVersion.validateTableUpdate` only reacts to schema changes (`lastSchemaUpdateTime`) or `UPDATING_META` state — a tablet reshard bumps neither, so it rethrows instead of retrying.

## What I'm doing:

Bump `olapTable.lastSchemaUpdateTime` at the split's visible-commit point (`addNewMaterializedIndexes()`, which runs on both the leader path `runRunningJob()` and the replay path `replayCleaningJob()`), mirroring what `MergePartitionJob` already does at its partition-replace commit point. The existing retry loop then detects the change and re-plans against the new tablet layout instead of hard-failing.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

---

### Verification

Reproduced on a shared-data RANGE PK lake table under sustained ingest with dynamic reshard splitting and a concurrent Q1–Q5 query load: the `Invalid tablet id` failure occurs deterministically in the split index-swap window. With this fix the planning-retry loop re-plans against the new tablet layout and the failure no longer occurs during concurrent splits.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/SplitTabletJob.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/SplitTabletJob.java
@@ -740,6 +740,18 @@ public class SplitTabletJob extends TabletReshardJob {
                             newIndex.getMetaId() == olapTable.getBaseIndexMetaId());
                 }
             }
+
+            // Installing the new materialized indexes changes the partition's tablet layout:
+            // PhysicalPartition.getLatestIndex() now returns a different tablet set. A query that
+            // captured the old layout during planning (OlapScanNode fills scanTabletIds from
+            // getLatestIndex(), then mapTabletsToPartitions() re-reads it) would otherwise hard-fail
+            // at plan build ("Invalid tablet id ... may have been dropped") or hand a CN a stale
+            // tablet/version whose metadata object no longer exists. Bump the table's optimistic
+            // version so StatementPlanner's retry loop (OptimisticVersion.validateTableUpdate) detects
+            // the change and re-plans against the new layout. Mirrors MergePartitionJob's bump at its
+            // partition-replace commit point. This runs on both the leader (runRunningJob) and the
+            // replay path (replayCleaningJob), so followers re-plan too.
+            olapTable.lastSchemaUpdateTime.set(System.nanoTime());
         }
     }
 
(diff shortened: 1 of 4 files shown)


fe/fe-core/src/main/java/com/starrocks/alter/reshard/SplitTabletJob.java
```diff
@@ -740,6 +740,18 @@ private void addNewMaterializedIndexes() {
...
+            // PhysicalPartition.getLatestIndex() now returns a different tablet set. A query that
+            // captured the old layout during planning (OlapScanNode fills scanTabletIds from
+            // getLatestIndex(), then mapTabletsToPartitions() re-reads it) would otherwise hard-fail
+            // at plan build ("Invalid tablet id ... may have been dropped") or hand a CN a stale
+            // tablet/version whose metadata object no longer exists. Bump the table's optimistic
+            // version so StatementPlanner's retry loop (OptimisticVersion.validateTableUpdate) detects
+            // the change and re-plans against the new layout. Mirrors MergePartitionJob's bump at its
+            // partition-replace commit point. This runs on both the leader (runRunningJob) and the
+            // replay path (replayCleaningJob), so followers re-plan too.
+            olapTable.lastSchemaUpdateTime.set(System.nanoTime());
```
xiangguangyxg, line 754: MergeTabbletJob also needs this fix

luohaha: Good catch — done. `MergeTabletJob` has the same structure (`addNewMaterializedIndexes()` called from both `runRunningJob()` and `replayCleaningJob()`), so I applied the identical `lastSchemaUpdateTime` bump there at its visible-commit point (commit f718f16).

xiangguangyxg (dismissed):

CelerData-Reviewer:
@codex review

kevincai (approved):

xiangguangyxg (approved):

PR #76130, opened 2026-07-09
[BugFix] Resolve sparse segment_idx rssids in lake GLM late-materialization 

luohaha:
## Why I'm doing:

On a shared-data RANGE PK table, global lazy-materialization (GLM) resolves a scan-emitted
row locator `rssid` back to its rowset + segment via
`LakeScanLazyMaterializationContext::get_rowset()`. That resolver assumed each rowset owns the
**dense** rssid range `[rowset.id, rowset.id + segment_metas_size)` and derived the segment as
`rssid - rowset.id`.

The assumption breaks after a tablet **SPLIT**. A split child keeps only the segments it owns and
**preserves each kept segment's original `segment_idx`** (so the inherited cloud-native PK-index
rssids stay stable), which leaves gaps — e.g. a child that drops the low-key segment keeps segments
with `segment_idx {1, 2}` while `segment_metas_size() == 2`. The scan tags those rows with
`rssid = rowset.id + get_segment_idx()` (sparse), but `get_rowset` checked `rssid < rowset.id + 2`, so:

- the top segment (`rssid == id+2`) resolved to no rowset → the row's late materialization failed
  with **`not found lake rssid`** (before #75889's null guard, a CN SIGSEGV);
- a surviving lower rssid could resolve to the **wrong physical segment position**, returning wrong data.

This is the actual root cause of the `not found lake rssid` errors seen on split children — the defect
is in the read/resolution path, not in the PK-index contents.

## What I'm doing:

Fix `LakeScanLazyMaterializationContext::get_rowset()` to be sparse-`segment_idx`-aware: match the
rssid against each segment's real `rowset.id + get_segment_idx(i)` and return the segment's **physical
position `i`**, with a fast reject on the rowset's owned span `[id, id + get_rowset_id_step)`.

This resolves sparse rssids correctly **without losing any row**, at the true source of the error, so
no write-side PK-index rewrite or entry drop is needed. Range-only in effect because hash-bucket
tablets never split; the OLAP path is unaffected (it allocates dense dynamic rssids).

> Supersedes the earlier base-level SST-merge reconcile approach on this branch: that dropped index
> entries on the write side using the *same* dense-interval assumption, which masked the symptom while
> risking dropping valid sparse-`segment_idx` entries. This PR fixes the resolver directly instead.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

---

### Verification

Added a unit test (`glm_manager_tablet_adaptor_test.cpp`) that builds a split-shaped lake rowset with
sparse `segment_idx {1, 2}` and asserts `get_rowset` resolves `id+1`→position 0 and `id+2`→position 1
(the case the pre-fix dense logic missed → `not found lake rssid`), plus a mixed dense/sparse
multi-rowset case and the unresolved-gap cases.

### DIFF
--- a/be/src/connector/lake/lake_global_late_materialization_context.cpp
+++ b/be/src/connector/lake/lake_global_late_materialization_context.cpp
@@ -39,13 +39,32 @@ lake::RowsetPtr LakeScanLazyMaterializationContext::get_rowset(const std::vector
     lake::RowsetPtr target;
     int32_t segment_id = 0;
 
+    // Resolve an rssid (rowset id + a segment's segment_idx) back to its rowset and the segment's
+    // PHYSICAL position in segment_metas. segment_idx is NOT necessarily dense over [0, num_segments):
+    // a tablet split keeps only the segments a child owns while preserving each kept segment's
+    // ORIGINAL segment_idx (so the inherited cloud-native PK-index rssids stay stable), which leaves
+    // gaps -- e.g. a child that drops the low-key segment keeps segments with segment_idx {1,2} while
+    // segment_metas_size()==2. Matching by segment_idx and returning the physical position is therefore
+    // required: the old dense `rssid - rowset.id()` offset would fail to resolve the top segment
+    // (rssid == id+2 falls outside [id, id+2)), surfacing as "not found lake rssid", and would resolve a
+    // surviving lower rssid to the wrong physical segment. See issue #75993.
+    const auto urssid = static_cast<uint32_t>(rssid);
     for (const auto& rowset : rowsets) {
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

stdpain (approved):

xiangguangyxg (approved):

PR #76145, opened 2026-07-09
[Feature] Support static extension access control for external catalogs

### LINKED ISSUE #76144: [Feature] Support static extension access control for external catalogs
## Motivation

External catalogs currently select only native, allow-all, or Ranger access control in `LazyConnector`. Static FE extensions are loaded before catalog initialization and can register capabilities, but there is no supported path for an extension to provide an `AccessController`.

That leaves custom external authorization with two poor choices: carrying a core patch/fork, or enforcing policy outside the engine. The latter cannot use StarRocks' existing authorization, row-filter, and column-masking integration points.

## Proposal

Add a small static-extension hook for external catalogs:

- Introduce `AccessControllerFactory`, which receives a `ConnectorContext` and returns an `AccessController`.
- Add `extension` as a `catalog.access.control` value.
- When configured, resolve the factory from `ExtensionManager` and install its controller for that catalog.
- Fail catalog initialization if the extension is absent or returns no controller.

This preserves the existing `native`, `allowall`, `ranger`, and Hive Ranger paths unchanged.

## Why this boundary

The factory returns StarRocks' existing `AccessController` contract, so authorization checks and policy expressions remain inside the normal FE analysis and rewrite flow. The hook deliberately does not prescribe a policy engine, network protocol, identity provider, or policy format.

## Proof of concept

A companion draft PR adds the hook, a focused `LazyConnector` unit test, and documentation for extension authors and the catalog property.

## Non-goals

- Replacing Ranger.
- Adding a policy-engine implementation to the StarRocks tree.
- Changing native authorization semantics.
- Adding a new dynamic-plugin type.

XavierGeerinck:
## Why I'm doing:

External catalogs can currently select only native, allow-all, or Ranger access controllers. Static FE extensions can register capabilities at startup, but an extension cannot currently provide an `AccessController` to an external catalog.

This forces custom external authorization to carry a core patch or operate outside the FE authorization and policy-rewrite path.

## What I'm doing:

- Add the `AccessControllerFactory` static-extension capability for external catalogs.
- Add `catalog.access.control = extension`.
- Install the factory-created controller for that catalog and fail initialization when the factory is absent or returns null.
- Document the extension registration pattern, concrete `ExternalAccessController` implementation, and catalog-property dispatch in English, Chinese, and Japanese.
- Add focused extension-context and connector unit tests.

Fixes #76144

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

## Validation

```text
PYTHON=python3 mvn -pl fe-core -am \
  -Dtest=com.starrocks.connector.LazyConnectorTest,com.starrocks.extension.DefaultExtensionContextTest \
  -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false \
  -DtrimStackTrace=false -T 2 test
```

Passed with Java 17: 24 tests, 0 failures, 0 errors.


jaogoy (commented):
Overall this is a clean, well-scoped extension hook. Returning the existing `AccessController` contract keeps authorization, row-filtering and column-masking inside the normal FE analysis/rewrite flow, and the existing `native` / `allowall` / `ranger` paths are untouched. A few things worth addressing before merge:

**1. Give a clear, actionable error when the factory isn't registered.**
`ExtensionManager.getComponent(AccessControllerFactory.class)` does *not* return `null` for a missing extension. Because the argument is an interface, `DefaultExtensionContext.get()` falls through to the DI path and throws `IllegalStateException("Failed to create instance of ...AccessControllerFactory ... no @Inject annotation or no-arg constructor found")`. That message is misleading for the actual cause (no extension registered). Please add an explicit registration/null check with a clear message before calling `createAccessController`, so that `catalog.access.control=extension` without a loaded extension fails with something actionable. The existing `Objects.requireNonNull` already handles the returns-null case well.

**2. The doc example should extend `ExternalAccessController`.**
External-catalog controllers follow the established `RangerAccessController extends ExternalAccessController (implements AccessController)` convention. The current example (`new CustomAccessController(...)`) and the unit test (mocking a bare `AccessController`) nudge authors toward implementing `AccessController` directly. Please show `class CustomAccessController extends ExternalAccessController` so authors follow the same pattern as the built-in external controllers.

**3. Document how to select among multiple extension controllers via catalog properties.**
The factory already receives the full `CREATE EXTERNAL CATALOG` property map through `context.getProperties()` (passed through unchanged from `CatalogMgr.createCatalog` → `ConnectorContext`). This means a single registered factory can dispatch to different controllers without any new interface parameter — the author just reads a custom property. Please extend the docs to show this pattern, e.g.:

```sql
CREATE EXTERNAL CATALOG ... PROPERTIES (
    "catalog.access.control" = "extension",
    "extension.access.control.type" = "authZ-1"
);
```

```java
ctx.register(AccessControllerFactory.class, connectorContext -> {
    String type = connectorContext.getProperties().get("extension.access.control.type");
    return switch (type) {
        case "authZ-1" -> new AuthZ1AccessController(connectorContext);
        case "authZ-2" -> new AuthZ2AccessController(connectorContext);
        default -> throw new IllegalArgumentException("unknown extension.access.control.type: " + type);
    };
});
```

This mirrors how the Ranger path carries its sub-detail in a dedicated property (`ranger.plugin.hive.service.name`) rather than encoding it into the `catalog.access.control` value, so it stays consistent with existing conventions. (If you'd rather support a compact `catalog.access.control=extension:authZ-1` form later, that's a reasonable follow-up, but it requires changing `LazyConnector`'s exact-match dispatch and can be done separately.)

**4. Broaden test coverage.**
The current test only asserts happy-path wiring with everything mocked. Please also cover:
- an end-to-end enable using a small *real* `ExternalAccessController` (assert an actual `check...` call is delegated to it, not just that `setAccessControl` was invoked);
- factory returns `null` → catalog init fails;
- factory / extension not registered → catalog init fails with the clear error from point 1.

XavierGeerinck:
Hi @jaogoy ! Addressed as much as possible, please let me know if more is required :)

PR #76149, opened 2026-07-09
[BugFix] Fix analysis time nullability for ROLLUP keys

chris-celerdata:
## Why I'm doing:
#75684 introduced a bug for `GROUP BY ROLLUP / CUBE / GROUPING SETS` produces NULLs in the grouping key columns, while `SelectAnalyzer` derives output field nullability solely from the expression itself (`item.getExpr().isNullable()`, i.e. the base column's nullability). This widening normally happens at the optimizer level, but due to ArrowFlight, the schema needs to be correct at analysis time. 
## What I'm doing:
Fix the bug for nullable columns when there are grouping clauses. 
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/SelectAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/SelectAnalyzer.java
@@ -92,6 +92,7 @@ public class SelectAnalyzer {
 
         List<Expr> groupByExpressions = new ArrayList<>(
                 analyzeGroupBy(groupByClause, analyzeState, sourceScope, outputScope, outputExpressions));
+        widenGroupingKeyNullability(groupByClause, outputScope, outputExpressions, groupByExpressions);
 
         boolean distinctWithoutGroupBy = selectList.isDistinct() && groupByExpressions.isEmpty();
         if (selectList.isDistinct()) {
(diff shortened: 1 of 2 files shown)


fe/fe-core/src/main/java/com/starrocks/sql/analyzer/SelectAnalyzer.java
```diff
@@ -635,6 +636,32 @@ private List<Expr> analyzeGroupBy(GroupByClause groupByClause, AnalyzeState anal
...
     }
 
+    /**
+     * ROLLUP/CUBE/GROUPING SETS produce super-aggregate rows where grouping-key columns are NULL,
+     * regardless of whether the underlying column is declared NOT NULL. The optimizer's Repeat
+     * operator already accounts for this widening at plan time, but analysis-time output field
+     * nullability (used by views, CTAS, etc.) is otherwise derived solely from the grouping
+     * expression itself, so it must be widened here too.
+     */
+    private void widenGroupingKeyNullability(GroupByClause groupByClause, Scope outputScope,
```
stephen-shelby, line 652: Thanks for the fix! Before this goes in I'd like you to confirm a few things

Is this really Arrow-Flight-specific, or does the analysis-time nullability change also affect normal query result metadata / other consumers? Want to understand the blast radius.
The description says the optimizer already widens the grouping keys to nullable at plan time — but when I check the plan for a NOT NULL grouping key it still shows NOT NULL. Could you point to where the optimizer actually widens it?
Could you verify end-to-end that this genuinely fixes the mismatch — i.e. for a NOT NULL column used as a ROLLUP key, do the Arrow Flight prepare-time schema and the execute-time result schema actually agree now? Just want to be sure we're not moving the mismatch to the other side.

chris-celerdata: Blast radius (Q1): analysis-time only. Affects the Arrow Flight prepare schema and CREATE MATERIALIZED VIEW (MaterializedViewAnalyzer reads Field.isNullable() — correct direction). Not MySQL/JDBC result metadata (null-flag hardcoded to 0 in MysqlCodec), CREATE VIEW/CTAS (read originExpression), or execution.

"Optimizer widens at plan time" (Q2): I was wrong — the logical plan keeps the key NOT NULL (what you saw). Widening actually happens at fragment-build time on the tuple slots above the Repeat (hasNullableGenerateChild in PlanFragmentBuilder → getOutputExprs(), and on the BE via thrift has_outer_join_child), not in the logical plan, which is exactly why it has to be re-derived at analysis time here.

End-to-end (Q3) + gap: the original change widened only the scope Field, but Arrow prepare reads the output Expr (getOutputExpression().isNullable()). Through a view it works (visitView rebuilds outer fields from the widened body); a direct GROUP BY ROLLUP/CUBE/GROUPING SETS was still reported NOT NULL at prepare while delivering NULLs — verified live. Fixed by widening the expression too:
```java
outputFields.get(i).setNullable(true);
if (outputExpr instanceof SlotRef) {
    ((SlotRef) outputExpr).setNullable(true);
}
```
Now all four view/direct × ROLLUP/GROUPING SETS cases agree prepare-vs-execute (added direct ROLLUP/CUBE/GROUPING SETS tests + a plain-GROUP BY negative guard). count(*) stays prepare=NOT NULL/execute=nullable — benign (never actually null).

stephen-shelby (approved):

HangyuanLiu (approved):

PR #76165, opened 2026-07-10
[BugFix] Bound and expose the Iceberg partition cache memory

Youngwb:
## Why I'm doing:

`CachingIcebergCatalog`'s partition cache stores one whole-table partition map (`Map<String, Partition>`) per entry, but was bounded only by **entry count** (`maximumSize`), not bytes. A handful of tables with many partitions (a table can have millions) can therefore grow it without limit. It is also the **only** metadata cache left out of `estimateSize()`, so its footprint is invisible in `/api/memory_usage` and the minute-level memory logs.

This is a real growth path, not a corner case: materialized views over Iceberg tables track base-table partitions, so MV rewrite/refresh populates the partition cache across many base tables.

Alongside it, the auxiliary maps `tableLatestAccessTime` / `tableLatestRefreshTime` / `tableRefreshLockMap` were likewise absent from `estimateSize()`.

## What I'm doing:

- Add a weigher + `maximumWeight` bound to the partition cache, sized by a new catalog property `iceberg_partition_cache_memory_usage_ratio` (default `0.1`), mirroring the data/delete/table file caches. The weigher samples the partition values (sample size 3) so weighing stays cheap on large maps.
- Include the partition cache and the three auxiliary maps in `estimateSize()`, so their bytes are reported. This benefits from the recently merged Estimator container-overhead fix (#75971), which makes the estimate accurate.

Fixes #issue: N/A — FE memory-safety/observability fix for the Iceberg partition cache (no separate tracking issue).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

New catalog property `iceberg_partition_cache_memory_usage_ratio` (default 0.1). The partition cache is now bounded by memory weight instead of a fixed 100k entry count, so under memory pressure it may hold fewer entries than before; the ratio is tunable and can be raised to restore the previous footprint.

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/CachingIcebergCatalog.java
@@ -147,8 +147,13 @@ public class CachingIcebergCatalog implements IcebergCatalog {
                         }
                     }
                 });
-        this.partitionCache = newCacheBuilderWithMaximumSize(
-                icebergProperties.getIcebergMetaCacheTtlSec(), NEVER_CACHE, DEFAULT_CACHE_NUM).build(
+        long partitionCacheSize = Math.round(Runtime.getRuntime().maxMemory() *
+                icebergProperties.getIcebergPartitionCacheMemoryUsageRatio());
+        this.partitionCache = newCacheBuilder(icebergProperties.getIcebergMetaCacheTtlSec(), NEVER_CACHE)
+                .executor(executorService)
+                .maximumWeight(partitionCacheSize)
+                .weigher((Weigher<IcebergTableName, Map<String, Partition>>) this::weighPartitionEntry)
+                .build(
                     new com.github.benmanes.caffeine.cache.CacheLoader<IcebergTableName, Map<String, Partition>>() {
                         @Override
                         public Map<String, Partition> load(IcebergTableName key) throws Exception {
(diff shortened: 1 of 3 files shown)


dirtysalt (approved):
LGTM! 🚀

Youngwb:
### Cluster E2E verification

Verified both halves of this change on a real cluster (Iceberg catalog over HMS, `orders` table with 2406 partitions + a small `littel` table, FE `MaxHeap` = 25.6 GB). `partitionCache` is populated only via MV partition-tracking (plain queries / `SHOW PARTITIONS` / `partitions_meta` / `REFRESH EXTERNAL TABLE` do not touch it), so the cache was filled by refreshing two MVs to a total of 2408 partitions.

**1. Visibility — `partitionCache` now counted in `estimateSize`** (same build, before/after):

| state | estimateSize (B) | PartitionNames |
|---|---|---|
| baseline (empty partitionCache) | 3,193 | 0 |
| after refreshing MVs | 323,725 | 2408 |

Δ = **+320,532 B** attributable to the 2408-partition cache. Before this PR the same 2408 partitions left `estimateSize` unchanged (the cache was excluded from the sum → invisible to the memory tracker / weigher).

**2. Byte-bound eviction — `maximumWeight` + weigher** (same table, same MV refresh, only `iceberg_partition_cache_memory_usage_ratio` differs):

| catalog | ratio | cap | PartitionNames | estimateSize (B) |
|---|---|---|---|---|
| default | 0.1 | 2.56 GB | 2408 | 323,725 |
| tiny | 1e-6 | ~27 KB | 0 (the 320 KB `orders` entry exceeds the cap → evicted) | 2,528 |

The old `maximumSize` (count-based) bound could not react to per-entry byte size; the new weight-based bound evicts oversized entries as expected.

CI is green (BUILD / FE UT incl. `testPartitionCacheCountedInEstimateSize` / SQL-Tester / FE Code Style).

Astralidea (approved):

PR #76193, opened 2026-07-10
[Enhancement] Cluster iceberg rewrite_manifests output by order-preserving partition ranges

dontknow9179:
## Why I'm doing:

`rewrite_manifests` clustered data files into manifests with `hash(partition) % targetManifestClusters` (clusters capped at 100). When a table has more distinct partitions than clusters, partitions that are far apart in sort order hash into the same manifest, so each manifest's partition min/max bound spans the whole value range. Wide, overlapping bounds defeat manifest-level pruning at read time: a partition predicate can no longer skip manifests, which is the main benefit rewrite_manifests is supposed to preserve.

## What I'm doing:

Replace hash-mod clustering with order-preserving range bucketing. Collect the distinct top-level partition values of the snapshot, sort them with Iceberg's canonical comparator, and assign each an equal-width bucket so sorted-adjacent partitions land in the same manifest. Every manifest then covers a narrow, non-overlapping partition range, restoring effective manifest pruning while still bounding the number of concurrent manifest writers.

Details:
- NULL partitions cluster into their own manifest.
- Files written under an older partition spec are coerced into the current spec before clustering; an unknown spec (partition-spec evolution under a concurrent commit) aborts the run so the next maintenance pass re-clusters against the new spec.
- Values not seen during the up-front scan (e.g. a concurrent append to the same spec) fall into their sort-order neighbor bucket, preserving the ordering instead of polluting an unrelated bucket.
- Only data manifests are counted and clustered; delete manifests are left untouched.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergUtil.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergUtil.java
@@ -30,9 +30,15 @@ import com.starrocks.thrift.TExprMinMaxValue;
 import com.starrocks.thrift.TExprNodeType;
 import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.FileScanTask;
+import org.apache.iceberg.GenericManifestFile;
+import org.apache.iceberg.InternalData;
+import org.apache.iceberg.ManifestFile;
 import org.apache.iceberg.Schema;
+import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.TableProperties;
+import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.io.FileIO;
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

Wenjun7J (approved):

CelerData-Reviewer:
@codex review

Youngwb (dismissed):

CelerData-Reviewer:
@codex review

Youngwb (approved):

wyb (approved):

PR #76194, opened 2026-07-10
[BugFix] Avoid query 404 from torn split reads and batch-publish intermediate versions

luohaha:
## Why I'm doing:

On a shared-data RANGE PK table under concurrent tablet split + batch publish, a read query can 404 on a partition metadata object `.../meta/<tablet>_<version>.meta` that maps to no materialized state. Two independent mechanisms feed the same symptom:

1. **Torn (version, index) read during a split.** Lock-free query planning reads the tablet set (`PhysicalPartition.getLatestIndex()`) and the visible version (`getVisibleVersion()`) at two independent points. A concurrent split flips the partition from `(prevVersion, parentIndex)` to `(commitVersion, childIndex)` between them, so planning can pair the still-present parent index with the already-bumped commit version and route a scan to a `<tablet>_<version>.meta` that never existed.

2. **Unmaterialized intermediate version under batch publish.** With `lake_enable_batch_publish_version=true`, BE materializes one metadata object at the batch's final `new_version` and skips intermediate versions, while FE stepped `visibleVersion` per-txn through every intermediate (`LakeTableTxnLogApplier.applyVisibleLog` → `updateVisibleVersion`). A lock-free planner can sample an intermediate visible version whose `.meta` BE never wrote → 404.

## What I'm doing:

Both are fixed with two small, orthogonal, **lock-free** changes — no cached, derived `(version, index)` snapshot on `PhysicalPartition` (which would need a manual refresh at every version/index mutation site, an easy-to-miss invariant). Query planning stays lock-free.

- **Split — pair the scan version with the selected index** (commit 1). `OlapScanNode.computeTabletInfo` and the range-colocate path in `PlanFragmentBuilder.visitPhysicalOlapScan` now capture the selected index and the visible version together and thread the captured version through `addScanRangeLocations`, instead of re-reading `getVisibleVersion()` deep inside it (which widened the window). The two facts are read **index first, then version**. A split bumps the version (`setVisibleVersion`) before adding the child index, both inside one catalog-write-lock critical section (`SplitTabletJob.addNewMaterializedIndexes`), so reading `index → version` bounds the only remaining lock-free torn outcome to `(commitVersion, split-parent)`. The split publishes `commitVersion` on the parent tablets and the parent stays readable until `CLEANING` (the orphan-shard-group physical-deletion grace, #76195), so that rare sub-statement pairing returns correct data rather than a 404.

- **Batch — advance the visible version once, at batch end** (commit 2). `applyVisibleLogBatch` no longer steps `visibleVersion` per-txn. Each txn's per-txn bookkeeping (dataVersion, compaction score, reshard signal) still runs, but the visible version is validated for per-partition continuity and advanced **exactly once, at batch end** — jumping straight from the batch's start to its final materialized `new_version`, never through an unmaterialized intermediate. A lock-free reader therefore sees either the pre-batch or the final version, never an intermediate one. Runs under the db write lock; both the leader (`finishTransactionBatch`) and follower replay (`replayUpsertTransactionStateBatch`) reach it via `updateCatalogAfterVisibleBatch`, so the guarantee holds symmetrically on every FE that plans queries locally. Batch publish stays ON (no throughput regression).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

---

### Verification

- **Split** — `ScanNodeComputeScanRangeTest`: `testFinalizeStatsThreadsCapturedVisibleVersion` drives `finalizeStats() → getScanRangeLocations() → computeTabletInfo()`, and `testExecPlanBuildsOlapScanRangesWithCapturedVersion` builds an exec plan that runs `PlanFragmentBuilder.visitPhysicalOlapScan`'s inline scan-range path — both exercise the captured-version threading. Safety of the residual `(commitVersion, split-parent)` pairing relies on the parent-readability grace in #76195.
- **Batch** — `LakeTableTxnLogApplierTest`: `testApplyVisibleLogBatchAdvancesVersionOnceToFinal` asserts a 3-txn contiguous batch advances the partition to the final version only, and `testApplyVisibleLogBatchRejectsNonContiguousVersions` asserts a non-contiguous batch trips the continuity check without advancing.
- **End-to-end** — an earlier iteration of these fixes ran **224,605 queries with 0 `.meta` 404s** on a 3-CN shared-data cluster (RANGE PK table driven 1 → 300+ tablets under sustained ingest + a concurrent Q1–Q5 loop, batch publish ON), where the same workload hit the 404 within minutes without the fix. A re-run of the soak is recommended before merge.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
@@ -613,7 +613,8 @@ public class OlapScanNode extends AbstractOlapTableScanNode {
                                       MaterializedIndex index,
                                       List<Tablet> tablets,
                                       List<TKeyRange> partitionRanges,
-                                      long localBeId) throws StarRocksException {
+                                      long localBeId,
+                                      long partitionVisibleVersion) throws StarRocksException {
         boolean enableQueryTabletAffinity =
                 ConnectContext.get() != null && ConnectContext.get().getSessionVariable().isEnableQueryTabletAffinity();
         boolean skipDiskCache = ConnectContext.get() != null && ConnectContext.get().getSessionVariable().isSkipLocalDiskCache();
@@ -622,7 +623,9 @@ public class OlapScanNode extends AbstractOlapTableScanNode {
         int schemaHash = olapTable.getSchemaHashByIndexMetaId(index.getMetaId());
         String schemaHashStr = String.valueOf(schemaHash);
         ConnectContext ctx = ConnectContext.get();
-        long visibleVersion = chooseScanVersion(physicalPartition.getVisibleVersion(), physicalPartition.getId(),
+        // Use the visibleVersion the caller captured together with the selected index, so a concurrent
+        // split cannot make us pair this index with a mismatched version (404 on a live tablet).
+        long visibleVersion = chooseScanVersion(partitionVisibleVersion, physicalPartition.getId(),
                 ctx == null ? null : ctx.getScanVersionOverride());
         scanPartitionVersions.put(physicalPartition.getId(), visibleVersion);
         String visibleVersionStr = String.valueOf(visibleVersion);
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

PR #76195, opened 2026-07-10
[BugFix] Recycle a superseded reshard index instead of deleting it immediately

luohaha:
## Why I'm doing:

A query planned against a materialized index that is later **superseded** by a tablet reshard can 404 on a `<tablet>_<version>.meta` once that index's shards are physically reclaimed while the query is still reading. This affects a **tablet-split parent** and a **tablet-merge child** index.

Key detail: a split **reuses the parent index's shard group for the child**, so the group is never orphaned — the group-level GC (`StarMgrMetaSyncer.deleteUnusedShardAndShardGroup`) never reaps it. The parent's shards are actually reclaimed **per-shard** by `StarMgrMetaSyncer.syncTableMetaInternal`, which lists every shard in the group and drops the ones not referenced by an index still on the partition. The reshard cleanup used to **detach the old index from its partition immediately**, so on the very next sync tick the parent shards became unreferenced and were dropped (`forceDeleteData=true`) — with no grace for in-flight reads.

## What I'm doing:

Route the superseded index's retention through the **`CatalogRecycleBin`** (one retention rule) at the **natural granularity — recycle a `MaterializedIndex` directly** — but **defer the actual detachment to erase time** rather than detaching at reshard cleanup.

When a tablet split/merge replaces an index, `recycleOldMaterializedIndexes()` **leaves the old index installed** on its live physical partition and only parks a lightweight removal record (`RecycleMaterializedIndexInfo` = db / table / physical-partition / index ids) via `CatalogRecycleBin.recycleMaterializedIndex(...)`:

- **While retained**, the old index stays a normal (non-latest) version on the partition, which is what protects its shards:
  - reads/writes **never pick it** — every scan resolves the partition via `getLatestIndex(indexMetaId)` and every load via `getLatestMaterializedIndices(...)`, both of which return only the **new child**; `storageRowCount` uses `getLatestVisibleIndices` (no double-count/double-write);
  - `syncTableMetaInternal` enumerates it via `getAllMaterializedIndices`, **subtracts its tablets, and keeps its shards**. This lands the protection on the actual (per-shard) reaper, uniformly for split (shared group) and merge.
- **On erase** (retention elapsed), the recycle bin **detaches** the index (under the table write lock) and drops its tablets from the inverted index; `StarMgrMetaSyncer` then reclaims the now-unreferenced shards **per-shard** (`dropTabletAndDeleteShard`) on its next cycle — still gated by cluster-snapshot safety (#75638). It **never** does a partition-directory delete, which for a split would destroy the live **child** tablets that share the parent's object-storage directory.

This makes the earlier group-level protection unnecessary and **removes** the `getRecycledIndexShardGroupIds` union in `StarMgrMetaSyncer.getAllPartitionShardGroupId()` and the recycled-index re-registration in `CatalogRecycleBin.addTabletToInvertedIndex()` (the "still installed" invariant covers both).

**Retention:** reuses the existing **`partition_recycle_retention_period_secs`** (default **30 min**), floored to a positive value so an entry can't become erasable before its recycle is durable.

**Locking:** `eraseMaterializedIndex` collects the erasable entries under the recycle-bin monitor, then detaches + journals **outside** it, preserving the established **catalog-lock → recycle-bin-monitor** order (`detach()` takes the table write lock).

**Persistence:** the recycle is **not journaled** — it's rebuilt from the reshard job's replay and persisted in the recycle-bin image (additive, `SRMetaBlockEOFException`-tolerant for older images). Only the leader-driven erase is journaled, via `OP_ERASE_MATERIALIZED_INDEX`, which now carries the index's **full coordinates** so `replayEraseMaterializedIndex` detaches it self-containedly (idempotent / null-safe). The index keeps its own id, so there are no synthetic ids to allocate.

**Scope / follow-up:** covers tablet **split** (parent) and **merge** (child) via the shared `TabletReshardJob.recycleOldMaterializedIndexes`. Schema-change / rollup origin-index cleanup is **not yet** routed through this path (follow-up).

**Compatibility:** this changes the recycle-bin image + adds an editlog op, so it is **main-only and not backported** to 4.1.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [ ] Parameter changes
- [ ] Policy changes
- [ ] Feature removed
- [ ] Miscellaneous

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

---

### Verification

**Unit tests** (`FE UT` green in CI):
- `SplitTabletJobTest` / `MergeTabletJobTest`: after a split/merge runs to `FINISHED`, the superseded index is **left installed** on the partition (`physicalPartition.getIndex(oldId) != null`) and **scheduled for removal** (`recycleBin.isMaterializedIndexRecycled(oldId)`), while `getLatestBaseIndex()` resolves to the new child and the old tablets stay addressable in the inverted index. The optimistic-version-bump and replay tests keep passing.
- `CatalogRecycleBinTest.testRecycleAndEraseMaterializedIndex`: recycle → scheduled (idempotent) → not-yet-expired retained → expired erased (bookkeeping cleared) → null-safe replay.

**On-cluster** (shared-data, 1 FE + 3 CN, this build; `partition_recycle_retention_period_secs`=30–120s, `star_mgr_meta_sync_interval_sec`=30s to make the per-shard reaper run frequently):
- **Split correctness / read routing:** force-split a 100k-row range-distributed PK tablet. Query returns 100000 throughout; `EXPLAIN` shows `tabletList=<new-child>` (`tabletRatio=1/1`) — the still-installed old parent is **never scanned**. FE log confirms the old index is parked (`Put materialized index N to recycle bin`).
- **Retention holds while the reaper runs (the bug this fixes):** across multiple `star_mgr_meta_sync_interval_sec` cycles during retention, the old parent tablet stays present and the query stays 100000 — `syncTableMetaInternal` does **not** reap the still-installed old index's shards.
- **Erase after retention:** `Detached recycled materialized index N from partition … ; 1 tablets unregistered` → `Erased materialized index N`; the old tablet leaves `SHOW TABLETS`, the shards are reclaimed per-shard, and the query is **still 100000** (child intact).
- **FE restart mid-retention:** the parked index re-parks as **still installed** (replay of the reshard job), query stays 100000, scan still routes to the new child, no feature-related replay errors; a subsequent erase then detaches it cleanly.
- **SHOW-family during retention:** `SHOW TABLETS` / `SHOW PARTITIONS` / `SHOW PROC '/dbs/..'` / `SHOW CREATE TABLE` all succeed (the parked index is visible in `SHOW TABLETS` but never read — cosmetic only).

Merge shares the same `recycleOldMaterializedIndexes` retention path and is covered by `MergeTabletJobTest`; it was not separately reproduced on-cluster because the synthetic single-tablet splits produced identical (1:1) children with no range boundaries to merge back.


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -4609,6 +4609,21 @@ public class Config extends ConfigBase {
...
+     * queries plan against the children (the latest index) immediately; this grace only keeps the
+     * parent tablets' metadata alive long enough for read queries that were planned against them
+     * just before the split to finish scanning. Without it, an in-flight scan can 404 on the
+     * parent's {@code <tablet>_<version>.meta} once CLEANING deletes it. Set to 0 to disable
+     * (restore the previous immediate-cleanup behavior).
+     */
+    @ConfField(mutable = true, comment = "Grace period (ms) a split job waits in CLEANING before "
+            + "removing the split-away parent tablets, so in-flight read queries planned against "
+            + "them can finish before their metadata is reclaimed. 0 disables the grace.")
+    public static long lake_reshard_split_parent_retention_grace_ms = 60000;
```
xiangguangyxg:
```suggestion
    public static long tablet_reshard_parent_retention_grace_ms = 60000;
```

fe/fe-core/src/main/java/com/starrocks/alter/reshard/SplitTabletJob.java
```diff
@@ -331,6 +332,17 @@ protected void runCleaningJob() {
...
+        // planned against them just before this split finished can still open their metadata.
+        // The children were already published before CLEANING, so new queries route to them; this
+        // grace only prevents an in-flight scan from 404-ing on a parent tablet's <id>_<version>.meta
+        // that CLEANING would otherwise delete out from under it. Stay in CLEANING (the scheduler
+        // re-polls) until the grace elapses.
+        long graceMs = Config.lake_reshard_split_parent_retention_grace_ms;
+        if (graceMs > 0 && System.currentTimeMillis() - stateStartedTimeMs < graceMs) {
+            return;
+        }
+
```
xiangguangyxg: tablet merge also need this.

xiangguangyxg: The issue of materailized index being replaced and deleted is a fundamental problem, and schema changes also have this issue. Fixing it shouldn't be done in this job because it increases job time and occupies table state. Consider adding grace time during the actual physical deletion, this would also benefit schema changes.

luohaha: Agreed — reworked exactly along these lines. The grace is no longer in the reshard job; the whole CLEANING-parking approach is gone. Instead I added an **orphan-time** retention grace at the shared physical-delete chokepoint `StarMgrMetaSyncer.deleteUnusedShardAndShardGroup() -> dropTabletAndDeleteShard()`: record when a shard group first leaves the FE live set, and defer physical deletion until it has been orphaned for `lake_orphan_shard_group_retention_grace_seconds` (default 1800s = 30min, aligned with `lake_autovacuum_grace_period_minutes`; 0 disables). Because tablet split (parent), tablet merge (child) **and** schema-change / rollup (origin) all funnel their old-index physical reclamation through this one path, this covers all of them uniformly — and it no longer lengthens the reshard job or occupies the table state. (This also subsumes your other note that tablet merge needs the same fix.) The old `shard_group_clean_threshold_sec` grace didn't help here because it is keyed on shard-group *create* time, already expired for a long-lived superseded index.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
luohaha: The knob moved out of the reshard job to the shared lake-GC physical-delete path, so it's no longer reshard-specific — renamed to `lake_orphan_shard_group_retention_grace_seconds` (seconds, matching the neighboring `shard_group_clean_threshold_sec`; default 1800). It now also governs schema-change / rollup origin-index cleanup, not just reshard.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3266,6 +3266,22 @@ public class Config extends ConfigBase {
...
+     * a now-superseded index -- a tablet-split parent, a tablet-merge child, or a schema-change /
+     * rollup origin index -- from failing (404 on a &lt;tablet&gt;_&lt;version&gt;.meta) when that index's
+     * shards are reclaimed out from under it. Set to 0 to disable. Default 1800s (30 min), aligned
+     * with lake_autovacuum_grace_period_minutes; both should exceed the maximum expected query time.
+     */
+    @ConfField(mutable = true, comment = "Grace period in seconds, measured from when a shard group becomes " +
+            "orphaned, before its tablet metadata is physically deleted in shared-data mode. Protects in-flight " +
+            "queries planned against a superseded index (tablet-split parent / merge child / schema-change " +
+            "origin). 0 disables. Default 1800 (30 min).")
+    public static long lake_orphan_shard_group_retention_grace_seconds = 1800L;
```
xiangguangyxg: shard_group_clean_retention_grace_seconds ? mirror to other shard group clean config

luohaha: Done — renamed to `shard_group_clean_retention_grace_seconds` to mirror `shard_group_clean_threshold_sec`, and relocated the doc entry right next to it in en/zh/ja. (b2c2c67)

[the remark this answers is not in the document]
luohaha: Documented: added the config to `docs/{en,zh,ja}/administration/management/FE_parameters/shared_lake_other.md` next to `shard_group_clean_threshold_sec`, covering default (1800), type, unit, mutability, `0`-disables semantics, and introduced-in. Also renamed the knob to `shard_group_clean_retention_grace_seconds` per review.

starrocks-xupeng: can FE recycle bin help? because this feature functions like a recycle bin? all shard groups will be delayed 1hour + 30 minutes

fe/fe-core/src/main/java/com/starrocks/lake/StarMgrMetaSyncer.java
```diff
@@ -407,10 +432,15 @@ private void deleteUnusedShardAndShardGroup() {
...
                         SHARD_GROUP_DELETE_COUNTER.increase(1L);
+                        orphanShardGroupFirstSeenMs.remove(shardGroupId);
                     }
                 }
             }
         } while (nextShardGroupId != 0);
+
+        // Drop clocks for groups that are no longer orphaned (came back into the FE live set) so a
+        // group orphaned again later restarts its grace, and keep the map bounded.
+        orphanShardGroupFirstSeenMs.keySet().retainAll(orphansThisRun);
```
starrocks-xupeng: retainAll 在 list 中途失败后仍会执行（StarMgrMetaSyncer.java 新增的方法末尾）。分页循环里 listShardGroup 抛 DdlException 会 break，此时 orphansThisRun 只含已处理页的组，末尾 retainAll 会把未列出页上孤儿组的时钟误清，下轮重新计时。方向是保守的（只延迟不误删），但 StarMgr 持续抖动时物理回收会被无限推迟。建议：list 异常 break 时跳过 retainAll（加一个 completed 标志）。

luohaha: Good catch — fixed in 11a284f. Added a `listedCompletely` flag, set to `false` in the `catch (DdlException) { ... break; }`, and guarded the trailing `retainAll` with it. So a partial listing no longer resets the grace clocks of orphaned groups on the un-listed pages.

fe/fe-core/src/main/java/com/starrocks/common/Config.java
luohaha: The FE recycle bin (`CatalogRecycleBin`) only handles explicitly DROPPED catalog objects (db/table/partition/index) via `catalog_trash_expire_second`; the shards here belong to a *superseded* index (tablet split parent / merge child / schema-change / rollup origin) that is removed logically through `deleteMaterializedIndexBy*`, and their physical reclamation funnels through the StarMgr GC path (`StarMgrMetaSyncer.deleteUnusedShardAndShardGroup -> dropTabletAndDeleteShard`), not the recycle bin — so the recycle bin is not in this path. On the delay: the 1h create-time gate (`shard_group_clean_threshold_sec`) and the 30min orphan grace are independent AND-conditions, not additive. For the target case (a long-lived index that is then superseded) the create-time gate is already long past, so only the ~30min orphan grace adds delay.

xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/common/Config.java
```diff
@@ -3266,6 +3266,22 @@ public class Config extends ConfigBase {
...
+     * Retention grace, measured from when a shard group first becomes orphaned (leaves the FE live
+     * set), before its tablets' metadata is physically deleted from object storage. Unlike
+     * shard_group_clean_threshold_sec (keyed on shard-group create time, hence already expired for a
+     * long-lived index the moment it is superseded), this protects an in-flight query planned against
+     * a now-superseded index -- a tablet-split parent, a tablet-merge child, or a schema-change /
+     * rollup origin index -- from failing (404 on a &lt;tablet&gt;_&lt;version&gt;.meta) when that index's
+     * shards are reclaimed out from under it. Set to 0 to disable. Default 1800s (30 min), aligned
+     * with lake_autovacuum_grace_period_minutes; both should exceed the maximum expected query time.
+     */
+    @ConfField(mutable = true, comment = "Grace period in seconds, measured from when a shard group becomes " +
```
kevincai: check if this can be squash into existing RecycleBin reclamation so the entire system will follow the same rule instead of patching scattered different places.

Delete -> RecycleBin(Retention Period) -> Physical Clean (fe meta clean up) -> StarMgr meta clean

kevincai (changes requested):

fe/fe-core/src/main/java/com/starrocks/common/Config.java
luohaha: Thanks — the "one retention rule for the whole system" instinct is the right one, and it's the same thing @starrocks-xupeng flagged (this does look like a recycle bin). I did look at routing it through `CatalogRecycleBin`, but it doesn't fit this case cleanly, for three concrete reasons:

1. **Granularity.** `CatalogRecycleBin` only recycles db / table / partition (`recycleDatabase` / `recycleTable` / `recyclePartition`) — there is no path to recycle an individual materialized index. The superseded index here lives *inside a still-live partition* (the table and partition are not dropped); it is removed logically via `deleteMaterializedIndexBy*`. Feeding it into the recycle bin would mean building a new index-level recycle granularity — that is *more* new machinery, not less.

2. **Lifecycle.** The recycle bin models a user `DROP` with a recovery window (`RECOVER`). A schema-change / reshard / rollup origin index is never user-recoverable — it is an internal artifact of a completed DDL. Putting it in the "recoverable trash" conflates two different meanings of "deleted".

3. **Trigger.** Recycle-bin retention is keyed on the drop-request time, and there is no explicit drop event for a superseded index. The race we need to cover is keyed on *when the shard group becomes orphaned to StarMgr*, which is exactly the reference point this grace uses.

On "patching scattered different places": this isn't a new parallel subsystem — it is an in-place gate at the one place where all four sources (split parent / merge child / schema-change origin / rollup origin) converge for physical reclamation (`StarMgrMetaSyncer.deleteUnusedShardAndShardGroup -> dropTabletAndDeleteShard`). The new grace is ANDed with the existing gates at that same site: `isSafeToDelete()` (the cluster-snapshot guard from #75638) is still checked first and can only veto, and `shard_group_clean_threshold_sec` is still the final condition — the grace can only *further delay* deletion, never bypass either. Routing through the recycle bin would instead require each of those four DDL jobs to enqueue their superseded indexes into a new recycle path; that would be the scattered version.

There is also precedent that this safety gating belongs at the StarMgr layer *coordinated with* the recycle bin rather than *subsumed into* it: `eraseDatabase` is already gated by the cluster snapshot to prevent premature `StarMgrMetaSyncer` deletion, and #75638 wired reshard safety into `isSafeToDelete()` at this same chokepoint.

If we do want to unify everything under a single retention mechanism, I think that is a larger refactor (introducing an index-level recycle entry and moving all four DDL paths onto it) and out of scope for this targeted bugfix, which also needs to backport to 4.1. Happy to file a follow-up for the unification if you'd like. WDYT?

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/reshard/TabletReshardJob.java
```diff
@@ -289,4 +295,38 @@ public void replay() {
...
+            MaterializedIndex oldIndex, long virtualPartitionId, long virtualPhysicalPartitionId) {
+        Partition virtualPartition = new Partition(virtualPartitionId, virtualPhysicalPartitionId,
+                "reshard_recycle_" + oldIndex.getId(), oldIndex, table.getDefaultDistributionInfo());
+        RecyclePartitionInfo recyclePartitionInfo = new RecycleUnPartitionInfo(dbId, table.getId(),
+                virtualPartition, DataProperty.DEFAULT_DATA_PROPERTY, (short) 1, null);
+        // Never user-recoverable: an internal artifact of a completed reshard, not a user DROP.
+        recyclePartitionInfo.setRecoverable(false);
+        // Retain for the configured grace so in-flight reads finish; 0 falls back to the recycle bin's
+        // global catalog_trash_expire_second.
+        recyclePartitionInfo.setRetentionPeriod(Config.shard_group_clean_retention_grace_seconds);
```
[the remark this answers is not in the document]
luohaha: Good catch — confirmed and fixed the description (49f/…). For a non-recoverable partition, `getAdjustedRecycleTimestamp()` returns `0` when `retentionPeriod <= 0`, so `timeExpired()` treats it as already expired and it is erased on the next recycle-bin cycle. So a value `<= 0` **disables** the grace (near-immediate cleanup, reverting to the pre-fix behavior — still gated by `shard_group_clean_threshold_sec` and cluster-snapshot safety); it does **not** fall back to `catalog_trash_expire_second`.

This is the intended "disable" escape hatch, so I corrected the `Config` javadoc/`@ConfField` comment, the helper comment, and the en/zh/ja docs to state it accurately rather than changing the behavior. The default (`1800s`) is unaffected and protects reads as intended.

CelerData-Reviewer:
@codex review

docs/en/administration/management/FE_parameters/shared_lake_other.md
```diff
@@ -534,6 +534,15 @@ This topic introduces the following types of FE configurations:
 
 - Introduced in: -
 
+### `shard_group_clean_retention_grace_seconds`
```
kevincai: can we reuse the same config for the partitions: `partition_recycle_retention_period_secs`?

luohaha: Done in 39be36d — reused `partition_recycle_retention_period_secs` (default 30 min) and deleted the new `shard_group_clean_retention_grace_seconds` config + its en/zh/ja docs. It's the same knob `dropPartitionWithRetention` already uses for recycled partitions, so the virtual partition follows the same retention rule.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/TabletReshardJob.java
```diff
@@ -289,4 +295,40 @@ public void replay() {
...
+        RecyclePartitionInfo recyclePartitionInfo = new RecycleUnPartitionInfo(dbId, table.getId(),
+                virtualPartition, DataProperty.DEFAULT_DATA_PROPERTY, (short) 1, null);
+        // Never user-recoverable: an internal artifact of a completed reshard, not a user DROP.
+        recyclePartitionInfo.setRecoverable(false);
+        // Retain for the configured grace so in-flight reads finish. A value <= 0 disables the grace:
+        // for a non-recoverable partition getAdjustedRecycleTimestamp() then returns 0, so the entry is
+        // eligible for erase on the next recycle-bin cycle (reverting to the pre-fix immediate cleanup,
+        // still gated by shard_group_clean_threshold_sec and cluster-snapshot safety).
+        recyclePartitionInfo.setRetentionPeriod(Config.shard_group_clean_retention_grace_seconds);
+        GlobalStateMgr.getCurrentState().getRecycleBin().recyclePartition(recyclePartitionInfo);
```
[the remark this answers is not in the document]
luohaha: Fixed in 39be36d. The retention passed to `setRetentionPeriod` is now floored to a positive value (`Math.max(partition_recycle_retention_period_secs, 1)`), so `getAdjustedRecycleTimestamp()` returns the real recycle time (never 0) and the entry can't become erasable in the window before the reshard job's FINISHED update makes it durable — no `logErasePartition` for an un-added partition, so `replayErasePartition()` can't NPE. The default (30 min) keeps that window firmly closed. (Kept 'always recycle' — a delete-on-disable branch is replay-unsafe since the -1 id sentinel can't distinguish 'chose delete' from 'not yet processed'.)

fe/fe-core/src/main/java/com/starrocks/alter/reshard/TabletReshardJob.java
```diff
@@ -289,4 +295,40 @@ public void replay() {
...
+     * </ul>
+     *
+     * <p>The virtual-partition ids are allocated once on the leader and persisted on the reshard job
+     * (see {@link ReshardingMaterializedIndex}), so the replay path passes the same ids and rebuilds an
+     * identical recycle-bin entry rather than re-allocating.
+     */
+    protected static void recycleSupersededMaterializedIndex(long dbId, OlapTable table,
+            MaterializedIndex oldIndex, long virtualPartitionId, long virtualPhysicalPartitionId) {
+        Partition virtualPartition = new Partition(virtualPartitionId, virtualPhysicalPartitionId,
+                "reshard_recycle_" + oldIndex.getId(), oldIndex, table.getDefaultDistributionInfo());
```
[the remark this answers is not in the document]
luohaha: Fixed in 39be36d. The virtual partition is now named `reshard-recycled-index-<virtualPartitionId>`. The hyphens make it illegal as a user partition name (which must match `^[a-zA-Z]\\w*$`), so it can never match a real dropped partition in `disableRecoverPartitionWithSameName`; the globally-unique virtual-partition id also keeps it distinct.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/reshard/TabletReshardJob.java
```diff
@@ -289,4 +298,80 @@ public void replay() {
...
+                    GlobalStateMgr.getCurrentState().getNextId());
+        }
+        long virtualPartitionId = reshardingIndex.getRecycledVirtualPartitionId();
+        long virtualPhysicalPartitionId = reshardingIndex.getRecycledVirtualPhysicalPartitionId();
+
+        // Hyphens make this name illegal as a user partition name (which must match ^[a-zA-Z]\w*$), so
+        // it can never collide with a real dropped partition in disableRecoverPartitionWithSameName;
+        // the globally-unique virtual-partition id also keeps it distinct.
+        Partition virtualPartition = new Partition(virtualPartitionId, virtualPhysicalPartitionId,
+                "reshard-recycled-index-" + virtualPartitionId, oldIndex, table.getDefaultDistributionInfo());
```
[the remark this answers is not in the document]
luohaha: I don't think this one applies. `handleGetTabletMetadata`/`getCloudTabletMeta` returns **version-1 "initial empty metadata"** — the schema + config a CN uses to construct a fresh empty `TabletMetadataPB` when a tablet is first created (the method rejects version > 1 with `NOT_IMPLEMENTED`, and `TCloudTabletMeta` carries no rowsets/historical schemas). It is not the in-flight-scan read path: a scan of the superseded index reads its already-persisted `.meta`/rowsets **from object storage** (the 404 this PR fixes is a *storage* 404), which the recycle-bin retention keeps alive. The old index's tablets never legitimately request "initial empty metadata".

Also note this path resolves by tablet id first (`getTabletMeta(tabletId)`), and this change now **keeps** the old tablets in the inverted index (previously they were deleted immediately), so that lookup actually succeeds where it didn't before. Happy to reconsider if you have a concrete scan path that calls this RPC for an already-created tablet.

fe/fe-core/src/main/java/com/starrocks/alter/reshard/TabletReshardJob.java
```diff
@@ -289,4 +298,80 @@ public void replay() {
...
+     * (see {@link ReshardingMaterializedIndex}); the replay path reuses the same ids and rebuilds an
+     * identical recycle-bin entry rather than re-allocating.
+     */
+    protected static void recycleSupersededMaterializedIndex(long dbId, OlapTable table,
+            MaterializedIndex oldIndex, ReshardingMaterializedIndex reshardingIndex) {
+        // Allocate the virtual-partition ids once on the leader; the replay path finds them already set.
+        if (reshardingIndex.getRecycledVirtualPartitionId() == -1L) {
+            reshardingIndex.setRecycledVirtualPartitionIds(
+                    GlobalStateMgr.getCurrentState().getNextId(),
+                    GlobalStateMgr.getCurrentState().getNextId());
```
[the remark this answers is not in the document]
luohaha: Good catch — fixed in 3b19f00. The allocation guard now treats `<= 0` (either id) as "not yet allocated", not just `== -1`, so a reshard job serialized before these fields existed (Gson deserializes the missing primitive longs as 0 via Unsafe, bypassing the -1 initializer) allocates fresh ids on the leader instead of recycling as partition 0. Real `getNextId()` ids are always > 0. Added `ReshardingMaterializedIndexTest` covering the fresh (-1) and legacy-JSON (0) cases.

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/persist/OperationType.java
```diff
@@ -707,6 +707,11 @@ public class OperationType {
     public static final short OP_GRANT_ROLE_TO_GROUP = 20501;
     public static final short OP_REVOKE_ROLE_FROM_GROUP = 20502;
 
+    // Physically erase a materialized index parked in the CatalogRecycleBin (e.g. a superseded index
+    // retired by a tablet reshard). The recycle itself is not journaled -- it is rebuilt from the
+    // reshard job's replay and persisted in the recycle-bin image; only the leader-driven erase is.
+    public static final short OP_ERASE_MATERIALIZED_INDEX = 20503;
```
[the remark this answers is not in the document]
luohaha: Confirmed and fixed in 172ede9 — good catch, you're right. `buildIgnorableOperations()` `System.exit(-1)`s for any op above `OP_TYPE_EOF` (20000) that isn't whitelisted, so `20503` would abort the FE the moment that guard runs. Renumbered to `13558` (next free id in the normal < 20000 range).

FWIW my on-cluster tests didn't surface it because `OP_*` are compile-time constants (referencing them never initializes `OperationType`) and `IGNORABLE_OPERATIONS` is only touched on a replay failure, which the tests didn't hit. `OperationTypeTest` exercises that guard and would have caught it — it just hasn't run because FE UT is currently blocked in CI (unrelated backport-blacklist).

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -269,6 +273,84 @@ public synchronized void recyclePartition(RecyclePartitionInfo recyclePartitionI
...
+     * Idempotent: a re-run/replay of the reshard cleanup that re-invokes this is a no-op, so it can be
+     * called unconditionally without a duplicate-id precondition. Keyed by the index's own id.
+     */
+    public synchronized void recycleMaterializedIndex(RecycleMaterializedIndexInfo info) {
+        long indexId = info.getIndexId();
+        if (idToIndex.containsKey(indexId)) {
+            return;
+        }
+        idToIndex.put(indexId, info);
+        idToRecycleTime.put(indexId, System.currentTimeMillis());
```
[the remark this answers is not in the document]
luohaha: Real gap, thanks. With the defaults it doesn't bite — `partition_recycle_retention_period_secs` (30 min) is ~144x smaller than `tablet_reshard_history_job_keep_max_ms` (3 days), so the reshard job (which rebuilds `idToIndex` on replay) always outlives the retention, and the index is normally erased long before the job is pruned; a checkpoint also captures `idToIndex` well within 3 days. The window you describe needs retention > job-keep (misconfig) or no checkpoint for the full job-keep period, and even then `StarMgrMetaSyncer.isSafeToDelete()` still gates the physical reclamation for cluster-snapshot safety.

That said, relying on the reshard job (which, unlike a DROP op, is pruned on a timer) is fragile, so I'll close it properly by **journaling the recycle** (a leader-only `OP_RECYCLE_MATERIALIZED_INDEX`, replayed idempotently) so `idToIndex` is durable independent of the reshard job and image timing. Will push as a follow-up.

luohaha:
@codex review

The approach changed to **deferred index removal** — the superseded reshard index now stays installed on its partition for the whole retention window and is only detached at recycle-bin erase (so the per-shard `syncTableMetaInternal` reaper protects its shards). Please review the latest diff.

luohaha:
@kevincai — heads-up on a design change to this PR since we last talked.

We'd agreed to route the superseded index's retention through the `CatalogRecycleBin` (one retention rule), recycling a `MaterializedIndex` natively. While validating that, I found the **group-level** protection doesn't actually work for a **split**:

- A split **reuses the parent's shard group** for the child (`SplitTabletJobFactory.createMaterializedIndex` passes `oldIndex.getShardGroupId()`; the new child shards are created into that same SPREAD group). So the group is never orphaned, and the group-level reaper `deleteUnusedShardAndShardGroup` never touches it — unioning the recycled index's shard group into the live set is a no-op for split.
- The parent's shards are actually reclaimed **per-shard** by `StarMgrMetaSyncer.syncTableMetaInternal` (`forceDeleteData=true`, every sync tick), which the group-level protection did not delay — so the split read-race was effectively still open.

New approach (**deferred index removal**): instead of detaching the old index at reshard-cleaning time and re-registering its tablets, we **leave the old index installed on its partition for the whole retention window** and only park a lightweight removal record in the recycle bin.

- While installed, reads/writes never pick it — every scan resolves via `getLatestIndex`, every load via `getLatestMaterializedIndices`, row-count via `getLatestVisibleIndices` (all return only the new child).
- `syncTableMetaInternal` enumerates it via `getAllMaterializedIndices`, subtracts its tablets, and **keeps its shards** — i.e. the protection now lands on the actual per-shard reaper.
- On erase (retention elapsed), the recycle bin detaches it under the table write lock; `StarMgrMetaSyncer` then reaps the now-unreferenced shards per-shard — never a partition-directory delete.

This covers split (shared group) and merge uniformly, and lets us drop the group-level `getRecycledIndexShardGroupIds` union in `StarMgrMetaSyncer` plus the inverted-index re-registration on image load. `eraseMaterializedIndex` detaches outside the recycle-bin monitor to preserve the established catalog-lock → recycle-bin-monitor order; the erase journal carries full coordinates so replay is self-contained.

Validated: FE UT + on-cluster (split correctness with reads routed to the new child; mid-retention the per-shard syncer runs but does **not** reap the still-installed old index's shards; erase-after-retention with data intact; FE-restart mid-retention replays it as still-installed). CI is green. The PR title/description are updated. Would appreciate your review of the new shape.

xiangguangyxg (dismissed):

fe/fe-core/src/main/java/com/starrocks/persist/OperationType.java
```diff
@@ -686,6 +686,13 @@ public class OperationType {
     @IgnorableOnReplayFailed
     public static final short OP_ALTER_RESOURCE = 13557;
 
+    // Physically erase a materialized index parked in the CatalogRecycleBin (e.g. a superseded index
+    // retired by a tablet reshard). The recycle itself is not journaled -- it is rebuilt from the
+    // reshard job's replay and persisted in the recycle-bin image; only the leader-driven erase is.
+    // Must stay below OP_TYPE_EOF (20000): buildIgnorableOperations() System.exit()s on any non-
+    // whitelisted op above it.
+    public static final short OP_ERASE_MATERIALIZED_INDEX = 13558;
```
kevincai, line 696: `OP_ERASE_MATERIALIZED_INDEX` is missing the `@IgnorableOnReplayFailed` annotation that every neighboring erase/drop op carries (`OP_ERASE_DB_V2`=13553, `OP_ERASE_PARTITION_V2`=13554, `OP_DROP_ALL_BROKER_V2`=13555, `OP_DROP_REPOSITORY_V2`=13556, `OP_ALTER_RESOURCE`=13557). `buildIgnorableOperations()` only adds annotated ops to `IGNORABLE_OPERATIONS`, so this op cannot be skipped when `metadata_journal_ignore_replay_failure=true`.

Since `replayEraseMaterializedIndex` is documented as idempotent and null-safe, and a failed erase merely leaves the index installed to retry on the next cycle, this is a best-effort cleanup that should be ignorable like its siblings. Without the annotation, a replay exception on this op hard-aborts the FE instead of being skippable.

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
```diff
@@ -96,6 +97,9 @@ public class CatalogRecycleBin extends LeaderDaemon implements Writable, MemoryT
     private final com.google.common.collect.Table<Long, String, RecycleTableInfo> nameToTableInfo;
     private final Map<Long, RecyclePartitionInfo> idToPartition;
     private final Map<Long, Long> physicalPartitionIdToPartitionId;
+    // Superseded materialized indexes parked here by a tablet reshard (issue #75993), keyed by the
+    // index's own (globally unique) id. Never user-recoverable; erased strictly at the tablet level.
+    private final Map<Long, RecycleMaterializedIndexInfo> idToIndex;
```
kevincai, line 102: This new `idToIndex` map is not reset in `clear()` (the `// for test` helper at line 1699), even though `clear()` resets every other recycle-bin map (`idToDatabase`, `idToTableInfo`, `idToPartition`, `idToRecycleTime`, `enableEraseLater`, `asyncDeleteForPartitions`, ...).

As a result, parked-index entries leak across test scenarios: after `clear()`, `idToRecycleTime` is empty but `idToIndex` still holds entries, so `isMaterializedIndexRecycled()` can return a stale `true`, and `eraseMaterializedIndex()` will find no matching `idToRecycleTime` entry and skip those orphans permanently. Add `idToIndex.clear();` to `clear()`.

fe/fe-core/src/main/java/com/starrocks/persist/OperationType.java
luohaha: Good catch — fixed in 956a693. Added `@IgnorableOnReplayFailed` so it joins its erase/drop siblings in `IGNORABLE_OPERATIONS`. As you noted, `replayEraseMaterializedIndex` is idempotent/null-safe and a failed erase just leaves the index installed to retry next cycle, so it's safe to skip under `metadata_journal_ignore_replay_failure` rather than abort the FE.

fe/fe-core/src/main/java/com/starrocks/catalog/CatalogRecycleBin.java
luohaha: Fixed in 956a693 — added `idToIndex.clear();` to `clear()` alongside the other recycle-bin maps, so parked-index entries no longer leak across test scenarios.

kevincai (approved):

xiangguangyxg (approved):

PR #76198, opened 2026-07-10
[Tool] unstable-case-review: make probation-grant a `grant_probation` input (default off)

wanpengfei-git:
## Why I'm doing:
The review workflow's "Open PR granting probation eligibility" step was hardcoded `if: false`, so turning it on needed a code change each time. Its two original blockers are now resolved — SYNC_PAT is configured on CelerData, and the runner's `/var/lib/ci-tool` is a bind-mount real dir (the per-job `cp` is a true isolated copy, no source pollution).

## What I'm doing:
Replace the hardcoded `if: false` with a `workflow_dispatch` input **`grant_probation`** (default `'false'` — identical off-by-default behavior). Dispatch with `grant_probation=true` to open the probation-eligibility PR; no more code edits to toggle. Same pattern as `dry_run` (promote-demote) and `runs`. In practice it only fires on CelerData (SYNC_PAT + the ubuntu/ai runners live there).

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [x] Tool

(Default is off, same as the previous `if: false`; behavior only changes when a run explicitly passes `grant_probation=true`.)

## Checklist:

- [ ] This is a backport pr

### DIFF
--- a/.github/workflows/unstable-case-review.yml
+++ b/.github/workflows/unstable-case-review.yml
@@ -27,6 +27,11 @@ on:
         required: false
         default: '3'
         type: string
+      grant_probation:
+        description: 'Open a PR moving now-passing blacklisted cases into PROBATION (default false; needs SYNC_PAT, i.e. CelerData)'
+        required: false
+        default: 'false'
+        type: string
 
 permissions:
   checks: write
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #76201, opened 2026-07-10
[Enhancement] Support pipeline engine for stream load and routine load

Hyper-FF:
## Why I'm doing:

Prerequisites for PR #76067

Classic synchronous stream load (the streamLoadPut path) and routine load previously ran on the legacy non-pipeline engine (StreamLoadPlanner + PlanFragmentExecutor), while broker load and transaction stream load already run on the pipeline engine. Let them run on the pipeline engine too, behind two default-off switches so the legacy path stays the fallback.

FE: when Config.enable_pipeline_stream_load / enable_pipeline_routine_load is on, build the plan via LoadPlanner (the pipeline-correct planner) instead of StreamLoadPlanner, pin the scan to the requesting/assigned BE, and materialize a single BE-local TExecPlanFragmentParams (no RPC deploy). A new syncStreamLoad flag on StreamLoadScanNode/LoadPlanner keeps the real load id and skips channel_id so the BE reads the HTTP action LoadStreamMgr pipe. DefaultCoordinator.buildLocalStreamLoadParams builds and registers the fragment instance exec state (Deployer with needDeploy=false) and marks it executing so exec-state/profile reports are accepted, which makes the load profile work. Load-specific query options (compression, log_rejected_record_num, timezone, partial_update_mode, enable_profile) are carried over to match the legacy path.

BE: StreamLoadOrchestrator routes params with is_pipeline to a new in-process pipeline FragmentExecutor path; a finish callback on FragmentContext harvests the result from runtime_state on completion (no coordinator). Routine load reuses this unchanged (its task executor goes through the same orchestrator).

Correctness and safety:
- Bind the load mem-trackers (instance/query) and runtime profile right after prepare(), before the body is received, so upload buffers are charged to the per-load tracker for the whole receive phase (mirrors the legacy start callback); previously they were only set at finish.
- The sync/BE-local path never generates a multi-fragment shuffle plan (LoadPlanner gated on syncStreamLoad), so single-fragment in-process execution always applies, even when eliminate_shuffle_load_by_replicated_storage is off.
- Routine load registers its StreamLoadTask only after planning succeeds (no orphan task on plan failure) and falls back to the legacy planner on any planning error.
- The pinned BE is not rejected on a transient FE heartbeat lag, since it owns the load pipe and is already executing.
- Write error_url before signaling the HTTP thread, to avoid a torn read.

Both switches default off; when off the paths are byte-identical to the legacy behavior. Enable the switches only after all BEs are upgraded (an older BE would ignore the pipeline scan ranges and commit zero rows); this is noted in the docs (en/zh/ja).

Add a SQL regression test (test/sql/test_stream_load_pipeline) covering pipeline CSV/JSON, legacy CSV, and pipeline-vs-legacy parity.

## What I'm doing:

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

luohaha (approved):

stdpain (approved):

PR #76208, opened 2026-07-10
[Enhancement] Record generation version on lake PK index sstables

xiangguangyxg:
## Why I'm doing:

Shared-data (lake) range-distribution tables support tablet split/merge. A version-based incremental copy of physical files needs a reliable per-file version — the tablet metadata version at which the file first became visible. Segment (`rowset.version`, field 10), delvec (`version_to_file` key), and DCG (`DeltaColumnGroupVerPB.versions`) already carry one; the PK-index sstable did not. This adds the open-source infrastructure to populate it reliably.

## What I'm doing:

Add a new field `PersistentIndexSstablePB.generation_version` (ordinal 14) that records, for each physical PK-index sstable, the tablet metadata version at which it first became visible.

- `LakePersistentIndex::commit` is the single choke point that assigns versions, via the static helper `assign_generation_versions(new_meta, prev_meta, generation_version)`. It derives each file's version from the **previous persisted `sstable_meta`** — which the builder still holds before `finalize_sstable_meta` — rather than mutating live in-memory state:
  - a file already recorded in the previous meta keeps its recorded version, so a **legacy / pre-feature file (recorded as `0`) stays `0`** and a version-based consumer must treat `0` conservatively (e.g. lake vacuum retention retains it) — this avoids over-stamping a pre-existing file with a later version and vacuuming it prematurely;
  - a genuinely-new file (absent from the previous meta) is stamped with the generation version;
  - a non-zero value is never overwritten.

  This one choke point covers normal flush, all compaction variants, and `ingest_sst`. Because it reads the previous persisted metadata instead of mutating live state, it never re-stamps across publishes and needs no locking beyond `commit()`'s existing per-index lock. A `DCHECK` guards the resolved version as `> 0`.
- The split/merge PK-memtable flush runs `commit()` at `base_version`, so `flush_pk_memtable` threads the reshard version as an explicit `generation_version` parameter (which otherwise defaults to the builder's publish version). Freshly-flushed sstables are stamped it, and the returned metadata's own version is left untouched.
- Merge builds `sstable_meta` directly (bypasses `commit()`): rebuild outputs are new physical files stamped the merge version at their two emit sites, while projection paths reuse the physical file and inherit the version via `CopyFrom`.
- Use a **new proto field** (ordinal 14) rather than reusing the deprecated field 1, so a stale value written before this feature can never be misread as a generation version.

`0`/unset means unknown. This is open-source infrastructure only; the enterprise external-cluster-snapshot consumer is out of scope.

**Out of scope / follow-up:** cross-cluster **lake replication** (`lake_replication_txn_manager`) is intentionally not touched — it does a full-copy metadata replacement (`incremental_snapshot=false`), so its copied SSTs belong to a full baseline rather than the `version > pre_version` incremental path.

Tests (BE UTs): `assign_generation_versions` unit (new file → generation version; carry-forward from previous meta; **legacy `0` preserved**; non-zero never overwritten); `commit` stamping + no re-stamp on re-commit; `ingest_sst`; reshard fresh-flush → generation version with returned metadata kept at `base_version`; split publish; merge legacy-rebuild → merge version; merge projection inherits. Lake persistent-index + reshard suites pass; `clang-format` and BE module boundaries clean.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/lake_persistent_index.cpp
+++ b/be/src/storage/lake/lake_persistent_index.cpp
@@ -14,6 +14,8 @@
 
 #include "storage/lake/lake_persistent_index.h"
 
+#include <unordered_map>
+
 #include "base/debug/trace.h"
 #include "base/utility/defer_op.h"
 #include "column/chunk_factory.h"
@@ -246,6 +248,10 @@ Status LakePersistentIndex::ingest_sst(const FileMetaPB& sst_meta, const Persist
     sstable_pb.set_filesize(sst_meta.size());
     sstable_pb.set_shared_rssid(rssid);
     sstable_pb.set_shared_version(version);
+    // Record the generation version (PersistentIndexSstablePB.generation_version):
+    // the version at which this ingested sstable becomes visible in this tablet.
+    // Coexists with shared_version (read-time projection) and does not affect it.
+    sstable_pb.set_generation_version(version);
     sstable_pb.set_encryption_meta(sst_meta.encryption_meta());
     // Preserve the shared flag from sst_meta. During tablet split cross-publish,
     // newly-ingested SSTs may be shared with sibling split tablets; losing this flag
@@ -873,7 +879,32 @@ Status LakePersistentIndex::apply_opcompaction(const TabletMetadataPtr& metadata
     return Status::OK();
 }
 
-Status LakePersistentIndex::commit(MetaFileBuilder* builder) {
+void LakePersistentIndex::assign_generation_versions(PersistentIndexSstableMetaPB* new_meta,
+                                                     const PersistentIndexSstableMetaPB& prev_meta,
+                                                     int64_t generation_version) {
+    // Callers resolve |generation_version| to a real publish/reshard version first; a new file must
+    // never be stamped 0 ("unknown"). Benign if it ever slips through in release -- 0 is the
+    // conservative value consumers treat as retain -- so a debug assert suffices.
+    DCHECK_GT(generation_version, 0);
+    std::unordered_map<std::string, int64_t> prev_versions;
+    prev_versions.reserve(prev_meta.sstables_size());
+    for (const auto& s : prev_meta.sstables()) {
+        prev_versions.emplace(s.filename(), s.generation_version());
+    }
+    for (auto& s : *new_meta->mutable_sstables()) {
+        if (s.generation_version() != 0) {
+            continue;
+        }
+        auto it = prev_versions.find(s.filename());
+        if (it != prev_versions.end()) {
+            s.set_generation_version(it->second);
+        } else {
+            s.set_generation_version(generation_version);
+        }
+    }
+}
+
+Status LakePersistentIndex::commit(MetaFileBuilder* builder, int64_t generation_version) {
     if ((too_many_rebuild_files() || too_many_rebuild_rows()) && !_memtable->empty()) {
         // If we have too many files or rows need to be rebuilt,
         // we need to do flush to reduce index rebuild cost later.
@@ -896,6 +927,14 @@ Status LakePersistentIndex::commit(MetaFileBuilder* builder) {
             new_sstable_pb->CopyFrom(sstable_pb);
         }
     }
+    // generation_version == 0 means "use this publish's version"; a non-zero override is passed by
+    // the reshard flush (see UpdateManager::flush_pk_memtable). The builder still holds the
+    // previous persisted sstable_meta here (before finalize_sstable_meta below), which
+    // assign_generation_versions uses to carry existing versions forward.
+    if (generation_version == 0) {
+        generation_version = builder->tablet_meta()->version();
+    }
+    assign_generation_versions(&sstable_meta, builder->tablet_meta()->sstable_meta(), generation_version);
     builder->finalize_sstable_meta(sstable_meta);
     auto [file_cnt, row_cnt] = need_rebuild_counts(*builder->tablet_meta(), sstable_meta);
     _need_rebuild_file_cnt = file_cnt;
--- a/be/src/storage/lake/lake_primary_index.cpp
+++ b/be/src/storage/lake/lake_primary_index.cpp
@@ -109,7 +109,8 @@ Status LakePrimaryIndex::ingest_sst(const FileMetaPB& sst_meta, const Persistent
     }
 }
 
-Status LakePrimaryIndex::commit(const TabletMetadataPtr& metadata, MetaFileBuilder* builder) {
+Status LakePrimaryIndex::commit(const TabletMetadataPtr& metadata, MetaFileBuilder* builder,
+                                int64_t generation_version) {
     TRACE_COUNTER_SCOPE_LATENCY_US("primary_index_commit_latency_us");
     if (_persistent_index == nullptr) {
         return Status::OK();
@@ -117,7 +118,7 @@ Status LakePrimaryIndex::commit(const TabletMetadataPtr& metadata, MetaFileBuild
 
     auto* lake_persistent_index = dynamic_cast<LakePersistentIndex*>(_persistent_index.get());
     if (lake_persistent_index != nullptr) {
-        return lake_persistent_index->commit(builder);
+        return lake_persistent_index->commit(builder, generation_version);
     } else {
         return Status::InternalError("Persistent index is not a LakePersistentIndex.");
     }
--- a/be/src/storage/lake/update_manager.cpp
+++ b/be/src/storage/lake/update_manager.cpp
@@ -235,7 +235,8 @@ void UpdateManager::unload_and_remove_primary_index(int64_t tablet_id) {
 
 DEFINE_FAIL_POINT(skip_lake_pk_index_flush);
 
-StatusOr<TabletMetadataPtr> UpdateManager::flush_pk_memtable(const TabletMetadataPtr& metadata) {
+StatusOr<TabletMetadataPtr> UpdateManager::flush_pk_memtable(const TabletMetadataPtr& metadata,
+                                                             int64_t generation_version) {
     // Test-only escape hatch: reshard unit tests build hand-crafted metadata with no real
     // in-memory PK memtable, so there is nothing to flush; skip the (index-loading) flush so
     // they exercise the metadata merge/split logic without materializing a real index.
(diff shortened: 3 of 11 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/lake/lake_persistent_index.cpp
```diff
@@ -881,7 +885,16 @@ Status LakePersistentIndex::commit(MetaFileBuilder* builder) {
...
+    // Stamp the generation version on each sstable's own PB (set-if-unset): a newly
+    // created file gets the current publish version and keeps it forever after, while
+    // an already-stamped file is left untouched. get_all_sstable_pbs' CopyFrom below
+    // then carries it into the persisted metadata, and it is inherited unchanged across
+    // publishes and tablet split/merge. The reshard flush pre-sets tablet_meta()->version()
+    // to its new_version (see UpdateManager::flush_pk_memtable) so split/merge-flushed
+    // sstables get the reshard publish version here.
+    const int64_t publish_version = builder->tablet_meta()->version();
     for (auto& sstable_fileset : _sstable_filesets) {
+        sstable_fileset->set_version_if_unset(publish_version);
```
xiangguangyxg: H1 - cross-PR interaction with #76209 (potential under-retention of legacy sstables). This set-if-unset loop runs over the whole live fileset and cannot tell a genuinely-new sstable from one loaded from existing metadata. Lake PK sstables never had `version` populated before this PR, so every legacy (pre-upgrade) sstable is unset here and gets stamped with the current publish version P rather than its true (earlier) creation version. #76209 then reads `sstable.version()` as the creation version for vacuum retention (remove_compacted_sst -> retained_by_version). For a manual/external cluster snapshot pinning a pre-upgrade version V < P, the legacy sstable then has interval [P, remove) which excludes V, so it is vacuumed while the snapshot still needs it -> the snapshot becomes unrestorable. The previous filename-based retain handled this correctly because it read the pinned-version metadata. Suggested fix: stamp at the true creation sites (memtable flush / compaction outputs / ingest_sst, which already stamps) so legacy loaded sstables stay unset (0) and #76209 unset->conservative-retain covers them; or stamp legacy-unset files with the earliest version instead of the current publish version. (Dormant in pure OSS where getVacuumRetainVersions is empty; activates with the enterprise snapshot consumer.)

be/src/storage/lake/update_manager.cpp
```diff
@@ -277,7 +277,16 @@ StatusOr<TabletMetadataPtr> UpdateManager::flush_pk_memtable(const TabletMetadat
...
     // here — this flush does not produce its own metadata version; the
     // returned metadata is consumed by the surrounding reshard publish.
+    // This flush runs at base_version, but the freshly-flushed sstables first
+    // become visible to the split/merge children at the reshard publish version.
+    // Temporarily set the metadata version to new_version across commit() so the
+    // generation-version stamp records new_version for those (new) sstables, then
+    // restore base_version. Inherited sstables keep their base-metadata versions
+    // (they are present in the pre-commit sstable_meta), and the returned metadata
+    // is otherwise identical to before -- only the sstable generation stamps change.
+    mutable_metadata->set_version(new_version);
```
xiangguangyxg: M1 - robustness. Temporarily mutating the metadata version around commit() is safe today only because commit() reads the tablet version for nothing except the new generation-version stamp: need_rebuild_counts feeds only the _need_rebuild_* heuristics and finalize_sstable_meta just CopyFroms the sstable_meta. That is a fragile invariant - any future version-dependent logic in commit() would be silently corrupted by this transient bump. Consider a short comment/assert pinning the invariant here, or the more localized alternative of stamping freshly-flushed sstables in tablet_splitter/merge right after set_version(new_version) and leaving flush_pk_memtable version untouched.

xiangguangyxg (commented):
Self-review (generation-version stamping). Overall looks good: the proto change is comment-only (field 1 name/type/ordinal unchanged), set-if-unset is idempotent, and the reshard-flush new_version stamping is well covered by tests. Two inline notes below - one cross-PR interaction with #76209 (under-retention of legacy sstables) worth reconciling before both land, and one robustness note on the temporary version mutation in flush_pk_memtable. Minor test-coverage gaps: the merge projection path (sstable reused, version inherited via CopyFrom) and ingest_sst stamping have no dedicated case.

CelerData-Reviewer:
@codex review

be/src/storage/lake/lake_persistent_index.cpp
xiangguangyxg: Fixed in b323f8a2ab6. Reverted the stamp-and-inherit approach back to prev-metadata-match: `LakePersistentIndex::commit` now stamps via `assign_generation_versions`, which keeps any file already present in the previous persisted `sstable_meta` at its recorded version. A legacy/pre-feature sstable (never stamped, version 0) therefore **stays 0** instead of being over-stamped with the current publish version P, so #76209 `retained_by_version` conservatively retains it and a pre-upgrade pinned snapshot (V < P) is no longer at risk of premature vacuum. Only genuinely-new files (absent from the previous meta) are stamped with the publish version, and a non-zero value is never overwritten. Added `AssignGenerationVersionsTest` asserting `legacy_zero` stays 0.

be/src/storage/lake/update_manager.cpp
xiangguangyxg: Fixed in b323f8a2ab6. Removed the transient metadata-version mutation. The reshard `new_version` is now threaded as an explicit `generation_version` parameter through `LakePrimaryIndex::commit` -> `LakePersistentIndex::commit` (default 0 = use the builder publish version); `flush_pk_memtable` passes `new_version` directly and no longer does `set_version(new_version)` / restore around `commit()`. This removes the fragile invariant on `commit()` reading the tablet version.

CelerData-Reviewer:
@codex review

be/src/storage/lake/lake_persistent_index.cpp
```diff
@@ -873,7 +879,27 @@ Status LakePersistentIndex::apply_opcompaction(const TabletMetadataPtr& metadata
     return Status::OK();
 }
 
-Status LakePersistentIndex::commit(MetaFileBuilder* builder) {
+void LakePersistentIndex::assign_new_versions(PersistentIndexSstableMetaPB* new_meta,
```
kevincai: `DCHECK(new_version > 0)` or return error if `new_version` is 0 by any chance.

kevincai (dismissed):

be/src/storage/lake/lake_primary_index.cpp
```diff
@@ -228,7 +228,7 @@ Status LakePrimaryIndex::ingest_sst(const FileMetaPB& sst_meta, const Persistent
     }
 }
 
-Status LakePrimaryIndex::commit(const TabletMetadataPtr& metadata, MetaFileBuilder* builder) {
+Status LakePrimaryIndex::commit(const TabletMetadataPtr& metadata, MetaFileBuilder* builder, int64_t new_version) {
```
luohaha: Why not use `metadata.version` instead of pass a extra new_version

be/src/storage/lake/lake_persistent_index.cpp
xiangguangyxg: Done in `01a579c` — added `DCHECK_GT(new_version, 0)` at the top of `assign_new_versions`.

`commit()` already resolves the sentinel (`new_version == 0` → `builder->tablet_meta()->version()`, which is always ≥ 1) before calling in, so the assert documents/enforces that invariant for all call sites. I went with `DCHECK` over an error return because even if `0` ever slipped through in release it is benign — a new file stamped `0` reads as "unknown", the conservative value consumers already retain — so there is no correctness path to guard, only a programming invariant to catch in debug/CI.

gensrc/proto/lake_types.proto
```diff
@@ -45,7 +45,15 @@ message PersistentIndexSstableRangePB {
...
-    optional int64 version = 1; // Deprecated
+    // The tablet metadata version at which this physical sstable file first
+    // became visible -- i.e. the new version of the publish (or reshard) that
+    // produced it. Stamped once when the file is first persisted and never
+    // changed afterward; CopyFrom inherits it across tablet split/merge so a
+    // shared file keeps its original value. A non-zero value is never
+    // overwritten. 0/unset means unknown -- e.g. a pre-existing file from before
+    // this field was populated stays 0, so consumers such as lake vacuum
+    // retention must treat 0 conservatively.
+    optional int64 version = 1;
```
luohaha: Don't reuse this, use a new item and new name, e.g. `generate_meta_version`.

be/src/storage/lake/lake_primary_index.cpp
xiangguangyxg: During the reshard PK-memtable flush, `metadata` sits at `base_version`: that flush runs at base_version and deliberately does not produce its own metadata version (`builder.finalize()` is not called). But the freshly-flushed sstables first become visible to the split/merge children at the reshard **publish** version, which differs from `metadata->version()` (= base_version). Using `metadata.version()` would stamp base_version — the wrong visibility version.

So the reshard version is threaded in as an explicit `generation_version` param (default `0` → normal DML resolves to the builder's publish version). This also avoids temporarily mutating the tablet-level metadata version around `commit()`, since the returned metadata must stay at `base_version` for the surrounding reshard publish.

gensrc/proto/lake_types.proto
xiangguangyxg: Done in `757c6b1` — added a new field `optional int64 generation_version = 14` and left the deprecated field 1 (`version`) untouched. All stamping/read code and tests now target `generation_version` (the internal helper/param were also renamed to `assign_generation_versions` / `generation_version` to match). Agreed a fresh ordinal is safer: no risk of misreading a stale value that a pre-deprecation writer may have left in field 1.

CelerData-Reviewer:
@codex review

luohaha (approved):

srlch (approved):

PR #76209, opened 2026-07-10
[BugFix] Retain lake vacuum files by version interval across tablet reshard

xiangguangyxg:
## Why I'm doing:

Shared-data range-distribution tables support tablet split/merge. When a cluster snapshot pins a
pre-reshard version `V` and the partition is then split/merged, the FE vacuum daemon enumerates the
partition's current (post-reshard child) tablets and ships the retained-version set (including `V`)
to the BE. A child tablet's earliest metadata is the reshard publish version (`> V`), so it has no
`{child}_{V}.meta`.

`TabletRetainInfo::init` read `{tablet}_{V}.meta` for every retained version to build its protect set.
For a child, `V` returns `NotFound`, so `init` failed. `vacuum_tablet_metadata` calls `init` under
`RETURN_IF_ERROR` **before any deletion**, so the whole partition's vacuum RPC failed every cycle,
`lastSuccVacuumVersion` never advanced, and expired metadata / data files / txn logs accumulated for
as long as the snapshot existed — an auto-vacuum stall / storage leak.

Naively skipping the `NotFound` version is unsafe: the incremental vacuum can delete a pre-reshard
segment that the snapshot still needs (the grace-time clamp that protects automated snapshots does
not hold for a finished manual / external snapshot), and only-suppressing *shared* file cleanup is
also unsafe because the reshard "identical tablet" path does not mark its inherited files `shared`,
so they route through the regular deleter.

## What I'm doing:

Decide vacuum retention by **version interval** instead of by reading the pinned version's metadata.

A garbage file must be kept iff it was live at a retained version, i.e. there is a retained version
`V` with `create_version <= V < remove_version`, where `remove_version` is the metadata version that
moved the file into `compaction_inputs` / `orphan_files`.

- `TabletRetainInfo` now holds only the retained-version set. It reads **no** per-version metadata, so
  the `NotFound` failure mode cannot occur by construction. It exposes
  `retained_by_version(create_version, remove_version)` (data-file liveness) and `contains_version(v)`
  (tablet-metadata-file retention, unchanged). The old metadata read and the
  `contains_file`/`contains_rowset` name/id sets are removed.
- Segments are retained by their rowset's creation version (`RowsetMetadataPB.version`, preserved
  through split `CopyFrom` and compaction).
- Del files can be transferred by a cloud-native PK compaction onto a higher-versioned output rowset,
  so they carry a new **optional `DelfileWithRowsetId.version`**, stamped at creation and retained per
  file (independent of the containing rowset's version).
- Orphan files carry a new **optional `FileMetaPB.version`**, stamped from the source object's own
  version at each orphaning site (dcg `versions[i]`, delvec `version_to_file` key, sstable `generation_version`,
  idg entry `version`; same-txn transient segment orphans are stamped with the metadata version being
  built). An unset/0 version means "unknown → retain conservatively", never a false delete.
- `collect_garbage_files` uses `retained_by_version(candidate.version(), metadata.version())`. Because
  the decision is keyed on version rather than the per-file `shared` flag, it correctly protects both
  shared and exclusive/identical inherited files.

Note for reviewers: this changes the internal vacuum retain mechanism and adds two backward-compatible
optional fields (`FileMetaPB.version` and `DelfileWithRowsetId.version`) to the storage metadata proto.
There is **no** user-facing SQL / config /
API / metric change, so this is classified as a bug fix (behavior change: No). Range distribution is a
new, new-tables-only feature with no on-disk compatibility burden.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/meta_file.cpp
+++ b/be/src/storage/lake/meta_file.cpp
@@ -191,6 +191,9 @@ void MetaFileBuilder::append_dcg(uint32_t rssid,
             if (dcg_ver.shared_files_size() > 0 && i < dcg_ver.shared_files_size()) {
                 file_meta.set_shared(dcg_ver.shared_files(i));
             }
+            if (i < dcg_ver.versions_size()) {
+                file_meta.set_version(dcg_ver.versions(i));
+            }
             _tablet_meta->mutable_orphan_files()->Add(std::move(file_meta));
         }
     }
@@ -273,6 +276,9 @@ void MetaFileBuilder::apply_opwrite(const TxnLogPB_OpWrite& op_write,
         if (del_meta.has_shared()) {
             del_file_with_rid.set_shared(del_meta.shared());
         }
+        // Freshly created del file: stamp its creation version so lake vacuum can retain it by its
+        // own version after it is later transferred onto a higher-versioned compaction output rowset.
+        del_file_with_rid.set_version(rowset->version());
         rowset->add_del_files()->CopyFrom(del_file_with_rid);
     }
     // if rowset don't contain segment files, still inc next_rowset_id
@@ -280,7 +286,14 @@ void MetaFileBuilder::apply_opwrite(const TxnLogPB_OpWrite& op_write,
     // collect trash files: replaced partial-update segments and their segment-name-keyed .vi files
     for (const auto& orphan_file : orphan_files) {
         DCHECK(is_segment(orphan_file.name()) || is_vector_index(orphan_file.name()));
-        _tablet_meta->mutable_orphan_files()->Add()->CopyFrom(orphan_file);
+        auto* added_orphan = _tablet_meta->mutable_orphan_files()->Add();
+        added_orphan->CopyFrom(orphan_file);
+        // These replaced segment/.vi files are produced by this txn's partial-update rewrite, so
+        // their creation version is the metadata version being built. Respect an accurate version if
+        // the producer already set one.
+        if (!added_orphan->has_version()) {
+            added_orphan->set_version(_tablet_meta->version());
+        }
     }
     if (!_tablet_meta->rowset_to_schema().empty()) {
         auto schema_id = _tablet_meta->schema().id();
@@ -301,6 +314,9 @@ void MetaFileBuilder::apply_column_mode_partial_update(const TxnLogPB_OpWrite& o
         // sibling tablets. The orphan FileMetaPB only carries `shared`, so encode bundling into it
         // too, otherwise vacuum deletes the shared bundle file while a sibling still references it.
         file_meta.set_shared(segment_meta.shared() || segment_meta.has_bundle_file_offset());
+        // These segments are produced and discarded within this txn, so their creation version is
+        // the metadata version being built.
+        file_meta.set_version(_tablet_meta->version());
         _tablet_meta->mutable_orphan_files()->Add(std::move(file_meta));
     }
 }
@@ -502,6 +518,7 @@ void MetaFileBuilder::apply_drop_index(const TxnLogPB_OpDropIndex& op) {
                     file_meta.set_name(entry.index_file());
                     if (entry.has_file_size()) file_meta.set_size(entry.file_size());
                     if (entry.has_shared_file()) file_meta.set_shared(entry.shared_file());
+                    file_meta.set_version(entry.version());
                     _tablet_meta->mutable_orphan_files()->Add(std::move(file_meta));
                 }
             }
@@ -636,6 +653,7 @@ void MetaFileBuilder::remove_compacted_sst(const TxnLogPB_OpCompaction& op_compa
             }
         }
         file_meta.set_shared(shared);
+        file_meta.set_version(input_sstable.generation_version());
         _tablet_meta->mutable_orphan_files()->Add(std::move(file_meta));
     }
 }
@@ -652,7 +670,12 @@ void MetaFileBuilder::remove_lcrm_file(const TxnLogPB_OpCompaction& op_compactio
     // tablet metadata GC process, which runs periodically and handles failures gracefully.
     // This approach is safe, non-blocking, and handles distributed cleanup correctly.
     if (op_compaction.has_lcrm_file()) {
-        _tablet_meta->add_orphan_files()->CopyFrom(op_compaction.lcrm_file());
+        auto* added = _tablet_meta->add_orphan_files();
+        added->CopyFrom(op_compaction.lcrm_file());
+        // The mapper file is produced and orphaned by this compaction and is never referenced by any
+        // visible tablet metadata, so its creation version is the metadata version being built.
+        // Stamping it lets vacuum reclaim it instead of over-retaining it under a covering snapshot.
+        added->set_version(_tablet_meta->version());
     }
 }
 
@@ -737,6 +760,9 @@ Status MetaFileBuilder::apply_opcompaction(const TxnLogPB_OpCompaction& op_compa
                 if (dcg.shared_files_size() > 0) {
                     file_meta.set_shared(dcg.shared_files(i));
                 }
+                if (i < dcg.versions_size()) {
+                    file_meta.set_version(dcg.versions(i));
+                }
                 // Put useless `.cols` files into orphan files
                 _tablet_meta->mutable_orphan_files()->Add(std::move(file_meta));
             }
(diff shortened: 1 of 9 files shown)


CelerData-Reviewer:
@codex review

be/src/storage/lake/vacuum.cpp
```diff
@@ -351,7 +351,7 @@ static Status collect_garbage_files(const TabletMetadataPB& metadata, const std:
     }
 
     for (const auto& file : metadata.orphan_files()) {
-        if (retain_info.contains_file(file.name())) {
+        if (retain_info.retained_by_snapshot(file.version(), metadata.version())) {
```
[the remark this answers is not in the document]
xiangguangyxg: Good catch — fixed in the latest push.

You're right that the replaced partial-update segment/`.vi` orphans were unstamped, so with the switch to interval retention an unset `version` (0) made `retained_by_snapshot(0, current_meta_version)` keep them for the snapshot's lifetime — a reclaim regression vs. the old name-based check (no data loss, since over-retention is always safe, but a snapshot-duration leak).

I stamped them at the consumer. In `RowsetUpdateState::rewrite_segment` the orphaned files are segments of **this txn's own** rowset (`params.op_write.rowset()`) being rewritten to add the unmodified columns — not files inherited from an older version — so their creation version is exactly the metadata version being built. So `MetaFileBuilder::apply_opwrite` / `batch_apply_opwrite` now stamp each copied orphan with `_tablet_meta->version()`:

```cpp
auto* added_orphan = _tablet_meta->mutable_orphan_files()->Add();
added_orphan->CopyFrom(orphan_file);
if (!added_orphan->has_version()) {
    added_orphan->set_version(_tablet_meta->version());
}
```

The interval then becomes `[new_version, new_version)` (they are created and orphaned in the same txn), which is empty, so they are collected immediately — matching the old behavior — even while an older snapshot is pinned. The `!has_version()` guard means that if a future producer sets an *accurate* (older) version we won't clobber it; stamping the current version is safe here only because these are this-txn files (stamping a current version onto a genuinely older orphan would under-retain, so that is deliberately avoided).

CelerData-Reviewer:
@codex review

be/src/storage/lake/meta_file.cpp
```diff
@@ -636,6 +653,7 @@ void MetaFileBuilder::remove_compacted_sst(const TxnLogPB_OpCompaction& op_compa
             }
         }
         file_meta.set_shared(shared);
+        file_meta.set_version(input_sstable.version());
```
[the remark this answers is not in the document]
xiangguangyxg: Intentional and forward-compatible, not a defect.

`PersistentIndexSstablePB.version` is being populated with the SST's creation version by a separate, coordinated change; once that lands these SST orphans carry an accurate creation version here automatically, with no edit to this stamp. Until then the field is unset, so `FileMetaPB.version == 0`, which this retention treats as "creation version unknown → earliest possible" and keeps conservatively.

That is safe by construction: it can only ever *over*-retain, never delete a file a snapshot still needs. The only cost is that such an SST orphan is reclaimed by the periodic full vacuum instead of incremental vacuum — bounded, no data loss. Reserving `0` for unknown/legacy metadata is exactly the intended contract (the same conservative handling applies to any file whose version is not yet stamped).

xiangguangyxg: H1 - cross-PR interaction with #76208. Stamping the orphaned sstable with its generation version is only accurate if input_sstable.version() is the sstable's true creation version. #76208 stamps sstable version set-if-unset in commit(), which assigns legacy (pre-upgrade, previously-unset) sstables the current publish version P rather than their earlier real creation version. A manual/external snapshot pinning V < P would then under-retain such an sstable here (interval [P, remove) excludes V) -> unrestorable snapshot. The fix belongs in #76208 (keep legacy sstables unset so the unset->conservative-retain path covers them). Flagging so the two PRs are reconciled before both land.

be/src/storage/lake/vacuum.cpp
```diff
@@ -323,35 +301,62 @@ static Status collect_garbage_files(const TabletMetadataPB& metadata, const std:
...
+        // A rowset's segments share the rowset's creation version, so retain them together by
+        // rowset.version(). The one compaction path that would break this equality is lake
+        // partial-segment compaction, which carries OLDER segments forward into a higher-versioned
+        // output rowset, so rowset.version() overstates those carried segments' creation version. That
+        // path is config-gated (enable_lake_compaction_use_partial_segments, off by default), non-PK
+        // only, and being retired in favor of parallel compaction, so it is deliberately not
+        // special-cased here. CAUTION: if it is ever enabled, this can under-retain -- delete a carried
+        // segment that an older retained snapshot still needs -- and the fix is to give those carried
+        // segments their own persisted creation version to key on here instead of rowset.version().
+        if (!retain_info.retained_by_version(rowset.version(), metadata.version())) {
```
xiangguangyxg, line 313: Good call documenting the partial-segment-compaction hazard here. Confirming the assessment: it is config-gated (enable_lake_compaction_use_partial_segments, default off), non-PK, and being retired, so keying carried segments on rowset.version() is acceptable for now. Worth keeping this CAUTION prominent - if that path is ever re-enabled, carried segments would need their own persisted creation version to key on here instead of rowset.version().

xiangguangyxg (commented):
Self-review (version-interval vacuum retention). The redesign is correct: keying retention on each file [create, remove) version interval removes the NotFound failure by construction and protects both shared and identical-tablet inherited files (reshard does not mark the latter shared). Retaining del files by their own version and stamping the orphan sites is thorough, and unset(0)->conservative-retain is the right fail-safe. Dependency: this needs #76208 for sstable.version() to be meaningful (remove_compacted_sst). Land #76208 first or together; if this lands alone, sstable orphans read version 0 -> conservatively retained (safe, just over-retains). Two inline notes below.

be/src/storage/lake/meta_file.cpp
xiangguangyxg: Confirmed and agreed — this PR (#76209) stays as-is.

The `unset → 0 → conservative-retain` path here is exactly the safety net for legacy sstables, so this side needs no change. Forcing sstable orphans conservative unconditionally, or dropping the `version` read, would over-retain correctly-stamped *new* sstables and defeat the precision #76208 adds — so the fix does not belong here.

Tracking the fix to #76208: only stamp sstables actually created in the current txn, and leave legacy / pre-upgrade (unset) sstables unset, so this PR's conservative path covers them. The two need to be reconciled before both land — #76208 should not merge ahead of that change.

be/src/storage/lake/vacuum.cpp
```diff
@@ -323,35 +301,62 @@ static Status collect_garbage_files(const TabletMetadataPB& metadata, const std:
...
             if (del_file.shared() && shared_file_deleter != nullptr) {
                 RETURN_IF_ERROR(shared_file_deleter->delete_file(join_path(base_dir, del_file.name())));
             } else {
                 RETURN_IF_ERROR(deleter->delete_file(join_path(base_dir, del_file.name())));
             }
+            deleted_any_file = true;
+        }
+        // Estimate reclaimed bytes at the rowset level (segment_metas may omit per-file size); count a
+        // rowset only when at least one of its files was actually deleted, so a fully-retained rowset
+        // under a snapshot does not inflate the metric.
```
kevincai: may also indicate that this `garbage_data_size` may be over-counts because not all of the files in the rowset get deleted.

kevincai (dismissed):

be/src/storage/lake/vacuum.cpp
xiangguangyxg: Good catch — fixed. `garbage_data_size` now adds `rowset.data_size()` only inside the segment-deletion branch (i.e. only when the rowset's segments are actually deleted), instead of whenever *any* file in the rowset was deleted. So deleting only a del file while the segments are retained no longer inflates the reclaimed-bytes estimate with the full segment payload. Del files aren't part of `rowset.data_size()`, so they legitimately don't contribute; dropped the now-unused `deleted_any_file`. It stays a rowset-level estimate (segment_metas may omit per-file size). `test_vacuum_shared_data_files` still reports 16384.

CelerData-Reviewer:
@codex review

be/src/storage/lake/meta_file.cpp
xiangguangyxg: Reconciliation complete. #76208 merged using a dedicated `PersistentIndexSstablePB.generation_version` field, and its `assign_generation_versions` keeps legacy (pre-upgrade) sstables at `generation_version = 0` (carried forward via `prev_versions`), stamping only genuinely-new files — so the legacy-under-retention hazard is closed on that side.

This PR now reads `generation_version` (rebased onto main to pick up the field): the four sstable-orphan stamps use `sstable.generation_version()` instead of the deprecated `version`. Net: new sstables → precise creation version; legacy sstables → `0` → conservative retain (never a false delete). Rebased head `0589b36656d`.

CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #76215, opened 2026-07-10
[BugFix] Fix Iceberg manifest data-file cache serving an incomplete file set

Youngwb:
## Why I'm doing:
Fix https://github.com/StarRocks/StarRocksTest/issues/11562

The FE Iceberg manifest data-file cache can serve a cached file set that is missing a live data file while still passing the read-side completeness check, so scan planning silently drops the file and the query returns a short result (e.g. `count(*)` under-counts by exactly one data file's rows). It reproduces intermittently right after a commit — when metadata-cache refresh, statistics collection and the first query fill the cache concurrently — and self-heals once the entry is evicted, so a serial re-run passes.

## What I'm doing:

`ManifestReader.fillCacheIfNeeded` accumulated the cached data files into a plain `HashSet`. That set can be mutated by the parallel manifest fill, and a concurrent mutation can leave `size()` reporting the expected live-file count while an element is actually lost. The read-side check `StarRocksIcebergTableScan.isCompleteCachedFiles` only compares `size()` against `existingFilesCount + addedFilesCount`, so a set whose size lies passes the check and the missing file is dropped from planning.

Because data files have unique paths, a *trustworthy* `size() == live count` already implies completeness, so the fix makes the size trustworthy rather than adding a content checksum:

- Accumulate into `ConcurrentHashMap.newKeySet()` so `size()` stays accurate under parallel fill; the existing size check then reliably rejects and refills an incomplete entry.
- Log a warning when a non-empty cache entry fails the completeness check, since that failure path was previously silent.

Added a unit test showing the size-only completeness check is content-blind (accepts an expected-size set that is missing a live file).

Fixes #issue: N/A — race in the FE Iceberg manifest data-file cache; found during Iceberg query testing.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/memory/estimate/Estimator.java
+++ b/fe/fe-core/src/main/java/com/starrocks/memory/estimate/Estimator.java
@@ -492,6 +492,10 @@ public class Estimator {
         if (collection instanceof java.util.LinkedList) {
             return (long) size * LINKED_LIST_NODE_SIZE;
         }
+        if (collection instanceof java.util.concurrent.ConcurrentHashMap.KeySetView) {
+            // Backed by a ConcurrentHashMap; charge the same node + table overhead as the map itself.
+            return (long) size * HASH_NODE_SIZE + hashTableBytes(size);
+        }
         return 0;
     }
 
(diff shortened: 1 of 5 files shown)


CelerData-Reviewer:
@codex review

Youngwb:
@codex review

dirtysalt (approved):
LGTM! 🚀

Astralidea (approved):

PR #76220, opened 2026-07-10
[BugFix] Escape jsonpaths value in SHOW CREATE ROUTINE LOAD output (backport #75755)

srihithg:
## Why I'm doing:

Manual backport of #75755 (squashed commit `94637df`). Mergify's automatic backport failed with a cherry-pick conflict in `RoutineLoadJob.java` and auto-closed the original backport PR (#75770). Resubmitting with the conflict resolved.

The conflict was purely contextual: the three call-site edits (`jsonpaths`, `json_root`, `merge_condition`) applied cleanly, and the only conflicting hunk was the insertion point of the new `escapeForDoubleQuotedSql` helper, which main anchors after an unrelated `columnDescsToSql` method that does not exist on this branch. Resolution keeps only the `escapeForDoubleQuotedSql` helper; no unrelated code was pulled in. The resulting source and test are byte-identical to the merged change.

Original description:

`SHOW CREATE ROUTINE LOAD` emits the `jsonpaths` property value raw into a double-quoted SQL string. A real `jsonpaths` value contains double quotes, e.g. `["$.id", "$.name"]`, which terminate the enclosing `"..."` early and produce non-runnable DDL — copy/paste of the output fails to parse. The same generated DDL is used in replay/migration and tooling paths, so the job definition is effectively unrecoverable from the output. The sibling `json_root` and `merge_condition` properties have the same latent bug (a merge condition can legitimately contain a string literal like `src.op = "delete"`).

## What I'm doing:

Escape backslashes and double quotes in the `jsonpaths`, `json_root`, and `merge_condition` values in `RoutineLoadJob.jobPropertiesToSql()` via a small targeted helper (`\` first, then `"`).

A targeted escape is used instead of `StringEscapeUtils.escapeJava` (the convention used for `enclose`/`escape`) deliberately: `escapeJava` emits `\uXXXX` for non-ASCII characters, which the parser's `escapeBackSlash()` cannot decode. That would turn a CJK jsonpath like `["$.用户名"]` into a silently wrong path (loading NULLs) instead of failing loudly at parse time. The targeted escape round-trips all inputs exactly.

Added `ShowCreateRoutineLoadDDLTest` with round-trip tests: the emitted properties SQL is embedded into a full `CREATE ROUTINE LOAD` statement, parsed back, and the recovered values are asserted equal to the originals.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [x] This is a backport pr

This is a manual backport of #75755.

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
+++ b/fe/fe-core/src/main/java/com/starrocks/load/routineload/RoutineLoadJob.java
@@ -1734,6 +1734,14 @@ public abstract class RoutineLoadJob extends AbstractTxnStateChangeCallback
         return gson.toJson(jobProperties);
     }
 
+    // Escape a value for embedding in a double-quoted SQL string literal: backslash first, then
+    // double quote. Pairs with the parser's escapeBackSlash() so the emitted DDL parses back to
+    // the original value. escapeJava is unsuitable here: it emits \\uXXXX for non-ASCII chars,
+    // which escapeBackSlash() cannot decode, silently corrupting e.g. CJK jsonpaths.
+    private static String escapeForDoubleQuotedSql(String value) {
+        return value.replace("\\", "\\\\").replace("\"", "\\\"");
+    }
+
     public String jobPropertiesToSql() {
         StringBuilder sb = new StringBuilder();
         sb.append("(\n");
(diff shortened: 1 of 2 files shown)


kevincai (approved):

trueeyu:

PR #76224, opened 2026-07-10
[BugFix] Skip load spill clear_parent_path when no remote block used

luohaha:
## Why I'm doing:

`LoadSpillBlockManager::clear_parent_path()` is invoked unconditionally on the
`DeltaWriter` close path. Even when a load spills entirely to local storage and
never touches remote storage, it still issues a remote
`acquire_writable_dir()` followed by a `delete_dir()` against the object store.
For a local-only spill the remote parent directory was never created, so these
remote calls are pure overhead (extra object-store round-trips) with nothing to
clean up.

## What I'm doing:

Track whether any remote block was ever acquired:

- `acquire_block()` sets an atomic `_used_remote_block` flag when the block it
  returns is remote (either `force_remote` or a local-acquire fallback).
- `clear_parent_path()` returns early (no-op) when the flag is still false, so a
  purely local spill skips the remote `acquire_writable_dir` + `delete_dir`
  entirely.

The flag is atomic because blocks may be acquired concurrently by parallel
flush threads. Behavior is unchanged whenever any remote block exists — the
parent directory is still cleaned up exactly as before.

Added a UT (`test_clear_parent_path_skipped_for_local_only`) asserting that a
local-only spill never creates the remote parent directory and that
`clear_parent_path()` is a clean no-op.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Enhancement
- [ ] Feature
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/lake/load_spill_block_manager.cpp
+++ b/be/src/storage/lake/load_spill_block_manager.cpp
@@ -88,6 +88,11 @@ LoadSpillBlockManager::~LoadSpillBlockManager() {
 }
 
 Status LoadSpillBlockManager::clear_parent_path() {
+    // Skip cleanup if no remote block was ever acquired: a purely local spill never creates the
+    // remote parent directory, so acquiring a dir and issuing a remote delete would be wasted work.
+    if (!_used_remote_block.load(std::memory_order_relaxed)) {
+        return Status::OK();
+    }
     // _remote_dir_manager is initialized in init(), skip cleanup if init() was not called or failed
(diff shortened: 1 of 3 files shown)


wyb (approved):

trueeyu:

PR #76227, opened 2026-07-11
[BugFix] Delete load spill blocks before removing the per-load spill dir

luohaha:
## Why I'm doing:

On shared-data primary key loads that spill to remote object storage, every load leaves an orphaned 0-byte `<remote_spill_path>/load_spill/<load_id>/` directory marker behind in the object store. They accumulate one-per-load and are only reclaimed later by GC/vacuum.

Root cause: `LoadSpillBlockManager::clear_parent_path()` (called from `DeltaWriterImpl::release_resources()`) calls `delete_dir()` while the spill blocks are still held by `_block_container`, so their backing files still exist. `delete_dir()` is a delete-empty-dir operation, sees a non-empty directory, returns `"not empty"`, and — because the status is intentionally ignored (best-effort, "cleared by GC later") — the dir marker is never removed. The block files are only deleted afterwards by `~FileBlockContainer()` when `~LoadSpillBlockManager()` resets the container, i.e. after the one-shot `clear_parent_path()` already gave up.

Verified on a branch-3.5 shared-data cluster: `clear_parent_path` logs `Failed to clear load spill parent path ... dir load_spill/<load_id> is not empty`, and the empty `<load_id>/` marker persists across loads (accumulating one per remote-spill load). The spill *data* files themselves are correctly hot-deleted during merge — only the empty dir marker leaks.

## What I'm doing:

Reset `_block_container` at the start of `clear_parent_path()` so the spill block files are deleted synchronously (via `~FileBlockContainer`, created with `skip_parent_path_deletion=true`) **before** `delete_dir()` runs. This honors the intent already documented on `acquire_block()`: the directory is only removed after all spill files are cleaned up.

Added a regression test (`test_clear_parent_path_with_blocks_in_container`) that mirrors the real DeltaWriter flow — the block is held by `block_container()` rather than released by the caller — and asserts the parent directory is deleted. Without the fix, `clear_parent_path()` returns `"not empty"` and the directory survives.

Note: the per-load spill directory was removed entirely on `main` by the zero-LIST load-spill refactor, so this issue and fix apply to **branch-3.5 only** (this PR targets branch-3.5 directly; there is no `main` counterpart to backport from).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

### DIFF
--- a/be/src/storage/lake/load_spill_block_manager.cpp
+++ b/be/src/storage/lake/load_spill_block_manager.cpp
@@ -88,6 +88,11 @@ LoadSpillBlockManager::~LoadSpillBlockManager() {
 }
 
 Status LoadSpillBlockManager::clear_parent_path() {
+    // Release all spill blocks first. Their backing files are removed synchronously by
+    // ~FileBlockContainer() (which was created with skip_parent_path_deletion=true, so it only
+    // deletes the block file, not the parent dir). Without this, delete_dir() below runs while the
+    // block files still exist, sees a non-empty directory, and leaves an orphaned <load_id>/ dir marker.
+    _block_container.reset();
     // _remote_dir_manager is initialized in init(), skip cleanup if init() was not called or failed
     Status status = Status::OK();
     if (_remote_dir_manager != nullptr) {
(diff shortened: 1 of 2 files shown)


luohaha:
Superseded — folding this fix into #76224 (same branch-3.5 load-spill clear_parent_path area).

Issue #63819, opened 2025-10-08
Support Iceberg v3 Capabilities


inviscid:
## Feature request

**Is your feature request related to a problem? Please describe.**
Currently StarRocks does not support Iceberg v3 so we cannot store Geo types.  v3 also comes with several other optimizations that could help improve performance and support greater analytical capabilities.

**Describe the solution you'd like**
Enhance StarRocks to support Iceberg v3.

**Describe alternatives you've considered**
None

**Additional context**
What’s New in Iceberg V3
Apache Iceberg format version 3 introduces a set of capabilities aimed at enhancing flexibility, performance, and expressiveness in data modeling. While V1 and V2 focused on stability and row-level operations, V3 is about expanding the format to accommodate more complex use cases and data types.

New Data Types
V3 brings support for several advanced data types, allowing broader data modeling options:

`timestamp` with timezone (nanosecond precision)
`variant` for semi-structured data, similar to JSON
`geometry` and `geography` for geospatial analytics
`unknown `to support dynamic schemas

These additions help Iceberg cater to domains like IoT, analytics with location data, and systems that rely heavily on semi-structured input.

_Default Column Values_
You can now define default values for columns at the schema level. This simplifies schema evolution and reduces the need for client-side logic to populate values during insert operations.

_Multi-Argument Transforms_
V3 adds support for multi-argument partition and sort transforms. This enables more advanced partitioning strategies, such as bucketing based on multiple columns or composite date functions.

_Row Lineage Tracking_
The new spec allows capturing lineage information for rows. This is especially useful in regulated environments where traceability and auditability are critical.

_Binary Deletion Vectors_
To make row-level delete handling more space-efficient, V3 introduces binary deletion vectors. These can represent deletions compactly and are particularly beneficial when working with high-frequency updates or deletes.

Together, these features mark a significant upgrade in what Iceberg tables can represent and how efficiently they can perform under modern workloads.

PR #76228, opened 2026-07-11
[Feature] Support reading Iceberg v3 merge-on-read deletion vectors

dongyi-kim:
## Why I'm doing:

Iceberg spec v3 replaces v2 position-delete data files with **deletion vectors**: roaring bitmaps
stored in Puffin blobs of type `deletion-vector-v1`. StarRocks currently fail-fast rejects any
Iceberg table that contains a deletion vector, so v3 merge-on-read tables written by Spark / Flink /
other engines cannot be queried at all. This PR adds read (merge-on-read) support for them.

Relates to #63819 (Support Iceberg v3 Capabilities → Binary Deletion Vectors).

## What I'm doing:

Add read support for Iceberg v3 deletion vectors across the FE, BE, and thrift interface:

- **thrift** (`TIcebergDeleteFile`): add a `PUFFIN` value to `THdfsFileFormat` and the DV locator
  fields `referenced_data_file`, `content_offset`, `content_size_in_bytes` (all new/optional).
- **FE**: emit deletion vectors as `PUFFIN` scan-range delete files (with the blob locator) instead
  of failing. Gated by a new session variable `enable_iceberg_v3_deletion_vector` (default on;
  setting it `false` restores the explicit fail-fast).
- **FE — delete-file cache correctness**: fix `DeleteFileWrapper.equals/hashCode` to key on
  `(path, contentOffset)` instead of `path` alone. Iceberg v3 packs multiple deletion vectors into a
  single Puffin file (same `path`, distinct `contentOffset`), so keying on `path` alone collapses
  them into one entry in the manifest cache's `HashSet<DeleteFileWrapper>`; the cached delete-file
  count then never matches the manifest's live-file count, so the delete manifest is re-read on every
  query. This is a pre-existing latent issue that only surfaces once v3 DV tables can actually be
  read — `contentOffset` is null for position/equality deletes, so their behavior is unchanged.
- **BE**: `IcebergDeleteBuilder::build_deletion_vector` reads the `deletion-vector-v1` blob directly
  via `content_offset`/`content_size_in_bytes` (no Puffin footer parsing needed), deserializes its
  roaring64 bitmap, and merges it into the existing `DeletionBitmap` apply path used by v2 position
  deletes — wired into both the Parquet and ORC scanner dispatch.

The DV blob body is binary-compatible with the Delta Lake deletion vector format, so the existing
`roaring64_bitmap_portable_deserialize_safe` primitive is reused.

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Interface/UI changes
- [x] Parameter changes: adds the session variable `enable_iceberg_v3_deletion_vector` (default on)
- [x] Policy changes: queries on Iceberg v3 tables with deletion vectors now succeed instead of failing fast
- [ ] Feature removed
- [x] Miscellaneous: new optional thrift fields (upgrade/downgrade compatible)

## Verification

Built the FE and BE and ran the result against a live lakehouse — Iceberg REST catalog + S3 object
storage + Spark 3.5.3 / Iceberg 1.10.2, which writes real `deletion-vector-v1` Puffin blobs. Test
tables were created as `format-version=3, write.delete.mode=merge-on-read`; their `.delete_files`
metadata confirms `file_format = PUFFIN`.

**Functional — deleted rows correctly excluded on both Parquet and ORC:**

| table | data format | deleted ids | StarRocks returns | count |
|---|---|---|---|---|
| `dv_parquet` | Parquet + DV | 2, 4, 6, 8 | 1, 3, 5, 7, 9, 10 | 6 |
| `dv_orc` | ORC + DV | 3, 6, 9 | 1, 2, 4, 5, 7, 8, 10 | 7 |

- `SET enable_iceberg_v3_deletion_vector = false` → query fails fast; `= true` → correct results.
- Baseline: stock StarRocks (without this PR) on the same table → `ERROR: Iceberg V3 Deletion Vectors
  are not supported`, confirming this PR is what enables the read.

**Correctness at scale:** a 5,000,000-row table with 10% of rows deleted returns `count(*) =
4,500,000`, and querying the deleted keys (`WHERE id % 10 = 0`) returns 0 rows — the deletion vectors
are genuinely applied during the scan, not a metadata shortcut.

**Unit tests:** FE `IcebergV3UnsupportedFeaturesTest` — `Tests run: 15, Failures: 0, Errors: 0`
(scan-range emission, the feature flag, and DeleteFileWrapper DV identity); plus a BE gtest covering
the DV blob parser (`parse_deletion_vector_blob`: valid / empty / bad magic / bad length / merge).

**Performance — v3 deletion vectors on par with v2 position deletes (merge-on-read):** identical
5,000,000-row data written two ways, same 10% deleted, same query set (`sum` / filtered `count` /
`count(distinct)`), median of repeated runs:

| storage latency (injected RTT) | v3 deletion vector | v2 position delete |
|---|---|---|
| 0 ms (local) | ~0.07 s | ~0.07 s |
| +30 ms (remote-cluster-like) | ~0.10 s | ~0.10 s |

Traced at the object store, a v3 query issues exactly one GET per deletion vector (same shape as v2's
one read per position-delete file) with no per-query manifest re-read, so read performance holds up
under network latency.

## Checklist:

- [x] I have added test cases for my new feature
- [x] This pr needs user documentation (Iceberg catalog docs updated in en / zh / ja)
  - [x] I have added documentation for my new feature
- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


dongyi-kim:
recheck

fe/fe-core/src/main/java/com/starrocks/connector/iceberg/DeleteFileWrapper.java
```diff
@@ -53,12 +53,14 @@ public boolean equals(Object o) {
 
         DeleteFileWrapper that = (DeleteFileWrapper) o;
 
-        return deleteFile.path().toString().equals(that.deleteFile.path().toString());
+        // DVs can share a Puffin path; contentOffset (null for non-DV deletes) makes them distinct.
+        return deleteFile.path().toString().equals(that.deleteFile.path().toString())
+                && Objects.equals(deleteFile.contentOffset(), that.deleteFile.contentOffset());
```
dongyi-kim, line 58: Why contentOffset is part of the cache key: Iceberg v3 packs multiple deletion vectors into a single Puffin file (same `path`, distinct `contentOffset`). `DeleteFileWrapper` is the key type for the manifest deleteFile cache (a HashSet keyed by equals/hashCode). Keying on path alone collapses several DVs into one entry, so the cached count never matches the manifest's live-file count and the delete manifest is re-read on every query. `contentOffset()` is null for position/equality deletes, so their behavior is unchanged.

be/src/formats/iceberg/iceberg_delete_builder.cpp
```diff
@@ -67,6 +72,100 @@ StatusOr<std::unique_ptr<RandomAccessFile>> IcebergDeleteBuilder::open_random_ac
...
+        return Status::InternalError(strings::Substitute(
+                "Iceberg deletion vector length mismatch: prefix=$0, expected=$1", blob_length, expected_length));
+    }
+
+    uint32_t magic_number;
+    memcpy(&magic_number, blob + kDvLengthSize, kDvMagicSize);
+    if (!LittleEndian::IsLittleEndian()) {
+        magic_number = LittleEndian::ToHost32(magic_number);
+    }
+    if (magic_number != kDvMagic) {
```
dongyi-kim, line 260: Format note (moved out of the code): the deletion-vector-v1 blob is binary-compatible with the Delta Lake serialized DV. The magic is the 4 bytes {0xD1,0xD3,0x39,0x64}, read as a little-endian uint32 (= 1681511377) on a little-endian host — matching the existing Delta DV reader convention. The 4-byte big-endian length prefix covers magic + bitmap (not the length field itself and not the trailing CRC), which is why the length check is `content_size - length - crc`.

dongyi-kim:
Hi @Youngwb @stephen-shelby — friendly ping on this one 🙏

This PR has been open for about four weeks. I've just **rebased it onto the latest `main` and resolved the conflicts** (it picked up the `IcebergDeleteBuilder` refactor into `IcebergPositionDeleteReader::read_rows` and the scanner move to `connector/hive/scanner/`), so it's mergeable again and CI is re-running on the new head. CLA is signed and the earlier CI was green.

Quick recap of the change (read-only, gated by `enable_iceberg_v3_deletion_vector`, default on):
- Reads Iceberg v3 `deletion-vector-v1` Puffin blobs directly via `content_offset`/`content_size_in_bytes` (no Puffin footer parsing), reusing the existing roaring64 primitive, and applies them through the current deletion-bitmap path for both Parquet and ORC.
- Replaces the previous fail-fast with a PUFFIN scan-range delete file.
- Also fixes a pre-existing `DeleteFileWrapper.equals/hashCode` cache bug (multiple DVs share one Puffin path; keying on path alone collapsed them, forcing the delete manifest to be re-read every query).

It relates to the v3 umbrella #63819. Would you have some time to take a look, or could you suggest a more appropriate reviewer? Happy to address any feedback. Thanks!

PR #76233, opened 2026-07-11
[Enhancement] Extend Sample-Based Tablet Pre-Split to range tables with a rollup index

xiangguangyxg:
## Why I'm doing:

Sample-Based Tablet Pre-Split previously excluded any range-distribution table that
carries a visible secondary materialized index (a rollup) via a hard
`getVisibleIndexMetas().size() != 1` eligibility gate. A bulk load into such a table
therefore got no pre-split: every index stayed a single tablet and the load wrote
serially. This change lets a bulk load into a range table with a rollup run in parallel,
like a hash-distributed table.

## What I'm doing:

Extend pre-split to split BOTH the base index and its rollup, each by boundaries sampled
from that index's own sort key, submitted as ONE combined reshard job.

FE-only. The BE / thrift / proto are unchanged: the External-Boundaries substrate
(`SplitTabletJobFactory.forExternalBoundariesMultiTablet`) already splits a multi-index
reshard job, so the change is entirely on the FE planning side -- resolve every visible
index's single tablet, sample the load source projecting each index's own sort key, plan
each index's boundaries, and feed all indexes' `oldTabletId -> ranges` into one job.

- On a shared-data range-distribution table a synchronous MV cannot be created, so the
  only visible secondary index is a routable range rollup that carries its own (possibly
  divergent) sort key; each index is sampled and planned by its own sort key.
- Split boundaries are advisory: the BE routes every row by its true value, so a
  per-index boundary can only shift where a split lands, never lose or misroute a row.
- Covers single-partition and multi-partition (auto-partition) loads for INSERT-from-FILES
  and Broker Load. Multi-partition planning is transactional per partition (all indexes of
  a partition split together or the partition is skipped), and secondary sample tuples are
  matched to their index by meta id, not by position.
- INSERT-from-table is intentionally skipped when the target has a rollup (its source is
  another table whose columns cannot be mapped to a divergent rollup sort key) -- a
  documented follow-up, not a partial split.
- The single-index (no-rollup) path is preserved exactly, and every sampling / planning /
  submit failure falls back to no-pre-split so the load always completes.

No new config, session variable, metric, thrift, or proto field.

Tests: full `com.starrocks.alter.reshard.presplit` FE unit package (34 classes) -- 624 run,
0 failures, 0 errors, 4 pre-existing skips; checkstyle 0. New coverage for the eligibility
relaxation, per-index planning on both the single- and multi-partition paths, transactional
all-or-nothing per partition, id-matched tuple pairing, the INSERT-from-table descope
(including a resolved-set race regression), and fail-safe fallbacks. A TSP shared-data
end-to-end run (base + rollup both split 1 -> N tablets, external-boundaries fallback = 0)
will be attached as a comment.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/AbstractSqlSampleSubqueryExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/AbstractSqlSampleSubqueryExecutor.java
@@ -165,15 +165,37 @@ abstract class AbstractSqlSampleSubqueryExecutor implements SampleSubqueryExecut
         SampleSpec spec = resolveSampleSpec(request);
         double samplingRate = pickSamplingRate(spec.totalInputBytes());
         int rowLimit = pickRowLimit(request.getSampleByteLimit());
+        List<SecondaryIndexSpec> secondaryIndexSortKeys = request.getSecondaryIndexSortKeys();
         String sampleSql = buildSampleSql(
                 spec.fromClauseSql(), spec.whereClauseSqlOrNull(),
-                spec.sortKeyProjectionIdents(), spec.partitionProjectionIdents(),
+                spec.sortKeyProjectionIdents(), secondaryProjectionIdents(secondaryIndexSortKeys),
+                spec.partitionProjectionIdents(),
                 samplingRate, rowLimit, request.getSeed());
         List<TResultBatch> resultBatches = runSampleQuery(
                 sampleSql, spec.computeResource(), request.getQueryTimeoutSeconds());
-        Iterator<SampleRow> rowIterator =
-                decodeRows(resultBatches, spec.sortKeyColumns(), spec.partitionSourceColumns()).iterator();
+        Iterator<SampleRow> rowIterator = decodeRows(
+                resultBatches, spec.sortKeyColumns(), secondaryIndexSortKeys, spec.partitionSourceColumns())
+                .iterator();
         return new SampleExecution(rowIterator, new Estimates(spec.totalInputBytes(), 0L));
+    }
+
+    /**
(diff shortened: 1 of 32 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
TSP shared-data e2e (1 FE + 3 CN, ubuntu): a range-distribution `DUPLICATE KEY(k1,k2) ORDER BY(k1,k2)` table with a single-column rollup `ORDER BY(k2)`, loaded via `INSERT ... SELECT * FROM FILES(...)`.

Result: the pre-split splits BOTH indexes -- base 1->2 tablets AND rollup 1->2 tablets, `external_boundaries_fallback_total` = 0 on all 3 CNs, `split_count` = 2, meta-tier used, COUNT=3,000,000 with base and rollup MIN/MAX/SUM all exact.

Note on ordering: the rollup split relies on the range-rollup shadow tablet carrying its own schema, which is fixed in **#76235 ([BugFix], prerequisite)**. Without #76235, the rollup falls back to identical (data-safe -- the base still splits and all data is correct); with #76235 both indexes split. This PR (pre-split) is correct and data-safe on its own; #76235 should merge first for the rollup to actually split. (This feature is main-only on the rollup path -- a range table cannot carry a rollup on branch-4.1 since `LakeRangeRollupJob` is main-only, so the multi-index path is dormant on the 4.1 backport and single-index behavior is unchanged there.)

kevincai (approved):

srlch (approved):

PR #76252, opened 2026-07-13
[BugFix] Unschedule global RF timer in PipelineDriver destructor

jenwitteng:
## Why I'm doing:

A BE/CN aborts during shutdown with `std::bad_weak_ptr`:

```
terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
  ... starrocks::pipeline::RunTimerTask(void*)
  ... bthread::TimerThread::Task::run_and_delete()
  ... bthread::TimerThread::run()
```

The global runtime-filter timer (`RFScanWaitTimeout`) is owned solely by
`PipelineDriver::_global_rf_timer`, and it is only unscheduled in
`finalize()`. Queued or blocked drivers that are abandoned when the driver
executor is closed at shutdown are destroyed **without** going through
`finalize()`, so the bthread `TimerThread` — which keeps only a raw pointer to
the task — can still fire on a task whose last owning `shared_ptr` is already
gone. `PipelineTimerTask::doRun()` then calls `shared_from_this()` on the freed
object and throws `std::bad_weak_ptr`, aborting the process.

This is the abandoned-driver counterpart to the query-context-destruction race
fixed in #73082, which unscheduled only from `finalize()`. The throwing
`shared_from_this()` in `doRun()` was introduced when `PipelineTimerTask`
became `enable_shared_from_this` in #72058.

## What I'm doing:

- Extract the unschedule into `_unschedule_global_rf_timer()` and call it from
  **both** `finalize()` and `~PipelineDriver()`.
- The helper moves the `shared_ptr` into a local so the task stays alive across
  `unschedule_and_join()`, closing the window where a concurrently running
  `doRun()` could observe a zero `use_count`.
- Add unit tests: the destructor unschedules a pending global RF timer, and the
  destructor blocks until an in-flight timer task returns.

Verified: both changed translation units compile cleanly with the project
toolchain (gcc-12, `-std=gnu++23`), zero diagnostics.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/exec/runtime/pipeline_driver.cpp
+++ b/be/src/exec/runtime/pipeline_driver.cpp
@@ -102,6 +102,10 @@ PipelineDriver::PipelineDriver()
           _driver_id(0) {}
 
 PipelineDriver::~PipelineDriver() noexcept {
+    // Queued or blocked drivers abandoned when the driver executor is closed during shutdown are
+    // destroyed without going through finalize(), so unschedule the global runtime-filter timer here
+    // as well; otherwise the timer thread may run a task whose owning shared_ptr is already gone.
+    _unschedule_global_rf_timer();
     if (_workgroup != nullptr) {
         _workgroup->decr_num_running_drivers();
     }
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

stdpain (approved):
It's OK for a hot fix

trueeyu (approved):

PR #76253, opened 2026-07-13
[Refactor] Unify position-fallback tabletId-to-bucketSeq fill into one facade helper

xiangguangyxg:
## Why I'm doing:

The "position-fallback tabletId→bucketSeq fill policy" — *if a `RangeColocateScanDispatch` is present
and the scan is range-aligned (`dispatch.computeBucketSeq(selectedIndex) != null`), use that bucketSeq
map; otherwise fall back to position-based bucketSeq (`put(allTabletIds.get(i), i)`)* — was written
twice:

- `OlapScanNode.computeTabletInfo` (a private `fillTabletId2BucketSeq`, early-return form), and
- inline inside `PlanFragmentBuilder.visitPhysicalOlapScan` (nested `if/else` form).

Two independent producers of the same policy can silently drift. Today any divergence is benign — the
dispatch-time alignment guard in `OlapScanNode.getBucketNums()` fails closed (worst case: one path
throws → shuffle fallback, never a wrong result) — but the duplication is unnecessary and easy to
collapse.

## What I'm doing:

Promote the existing `OlapScanNode` helper to a single shared static and route both call sites through
it:

```java
public static void OlapScanNode.fillTabletId2BucketSeq(
        @Nullable RangeColocateScanDispatch dispatch,
        MaterializedIndex selectedIndex,
        List<Long> allTabletIds,
        Map<Long, Integer> target)
```

`OlapScanNode` is the natural home: it owns the `tabletId2BucketSeq` field, its setter, the
`getBucketNums()` alignment guard that validates the built assignment, and every consumer of the map —
and this is where the method already lived (as a private instance method) before the de-dup. The
helper writes into the caller-supplied `target` map without clearing it, preserving the existing
whole-scan accumulation across sub-partitions. HASH / range-non-colocate scans pass a `null` dispatch,
so the shared path reads as `OlapScanNode.fillTabletId2BucketSeq(null, …)` and no longer implies that
every distribution routes through range-colocate. `RangeColocateScanDispatch` is unchanged.

This is a **behavior-preserving extraction**. The two sites were already semantically identical
(early-return vs nested `if/else`, same outcomes); the helper reproduces both exactly for all three
cases, so no plan shape or bucketSeq assignment changes:

| case | result |
|------|--------|
| `dispatch == null` (HASH / range non-colocate) | position-based fill |
| `dispatch != null` and `computeBucketSeq != null` (aligned) | `putAll(dispatch map)` |
| `dispatch != null` and `computeBucketSeq == null` (transiently unaligned) | position-based fallback |

Added 3 focused unit tests covering the three fill cases; the existing range-colocate plan/dispatch
suites stay green.

Backport: `[Refactor]` is not auto-backported. The same duplication exists on branch-4.1 (the
range-colocate feature line), so the 4.1 box below is checked to flag this PR for a **manual**
cherry-pick to branch-4.1 after merge.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [x] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
+++ b/fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
@@ -885,7 +885,7 @@ public class OlapScanNode extends AbstractOlapTableScanNode {
                     scanTabletIds.addAll(allTabletIds);
                 }
 
-                fillTabletId2BucketSeq(dispatch, selectedIndex, allTabletIds);
+                fillTabletId2BucketSeq(dispatch, selectedIndex, allTabletIds, tabletId2BucketSeq);
                 totalTabletsNum += selectedIndex.getTablets().size();
                 selectedTabletsNum += tablets.size();
                 addScanRangeLocations(partition, physicalPartition, selectedIndex, tablets, List.of(), localBeId);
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/planner/OlapScanNode.java
```diff
@@ -885,40 +885,15 @@ private void computeTabletInfo() throws StarRocksException {
                     scanTabletIds.addAll(allTabletIds);
                 }
 
-                fillTabletId2BucketSeq(dispatch, selectedIndex, allTabletIds);
+                RangeColocateScanDispatch.fillTabletId2BucketSeq(
```
kevincai: This refactor makes the code looks unclear, reading from the name, it seems that all colocation must go through `RangeColocateScanDispatch` no matter it is a range distribution or hash distribution. The common code can be extracted, but find a better place and name for it.

xiangguangyxg: Good point, thanks. I moved the extracted helper off `RangeColocateScanDispatch` to a `public static OlapScanNode.fillTabletId2BucketSeq(dispatch, selectedIndex, allTabletIds, target)`.

`OlapScanNode` is the natural home: it owns the `tabletId2BucketSeq` field, its setter, the `getBucketNums()` alignment guard that validates the built assignment, and every consumer — and this was the method's original location (a private instance method) before the de-dup. HASH / range-non-colocate scans now read as `OlapScanNode.fillTabletId2BucketSeq(null, ...)`, so the shared path no longer implies everything goes through range-colocate. `RangeColocateScanDispatch` is now unchanged from main.

I kept it a single `public static` rather than introducing a new class for one method. Updated in `a0e7943256e`.

CelerData-Reviewer:
@codex review

xiangguangyxg:
@codex review

kevincai:
@xiangguangyxg update the PR description accordingly.

kevincai (approved):

xiangguangyxg:
Thanks for the review! Updated the PR description to reflect the new home (`OlapScanNode.fillTabletId2BucketSeq`).

srlch (approved):

xiangguangyxg:
@Mergify backport branch-4.1

PR #76256, opened 2026-07-13
[BugFix] Reject full-column ORDER BY that reorders key columns of a key-derived range-distribution table

xiangguangyxg:
## Why I'm doing:

On a cloud-native range-distribution table whose sort key is key-derived (created without an explicit `ORDER BY`), a full-column `ALTER TABLE ... ORDER BY` that permutes the key columns takes the schema-reorder path (`processReorderColumn`). That reorders the schema — and thus the key-derived range sort key — but the plain lake schema-change job copies each shadow tablet's range verbatim instead of re-sampling. After the flip the sort key is the new key order while the stored boundary tuples are still in the old key-column space, so `RangeDistributionPruner` (FE) and `RangeRouter` (BE) compare new-order key tuples against old-order boundaries: silent wrong pruning and mis-routed loads, with no error. The subset `ORDER BY` form is routed to the range-rewrite job and re-samples, so only the full-column form slips through the `orderedColumns.size() != baseSchema.size()` dispatch. Reported by @banmoy while reviewing #76047.

## What I'm doing:

Reject the corrupting case: a full-column `ORDER BY` that permutes the key columns of a key-derived (`getSortKeyIdxes() == null`) cloud-native range-distribution table now fails with a clear message directing the user to the subset `ORDER BY (<sort-key columns>)` form, which re-samples the K-tablet boundaries. Cases that are already safe are unaffected: a full reorder that keeps the key order (reorders only value columns), tables created with an explicit `ORDER BY` (the sort key is remapped by name through a reorder), and non-range / non-cloud-native tables. Full support (routing the full reorder to a re-sampling rewrite job) is left as a follow-up enhancement.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
@@ -1228,6 +1228,33 @@ public class SchemaChangeHandler extends AlterHandler {
         if (newSchema.size() != targetIndexSchema.size()) {
             throw new DdlException("Reorder stmt should contains all columns");
         }
+        // A full-column ORDER BY on a cloud-native range-distribution table whose sort key is key-derived
+        // (no explicit ORDER BY at creation, so getSortKeyIdxes() == null) reorders the schema and thus the
+        // key-derived range sort key. But this plain schema-change job copies each shadow tablet's range
+        // verbatim, so the stored boundary tuples stay in the OLD key-column space while the new sort key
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

banmoy (approved):

luohaha (approved):

srlch (approved):

PR #76259, opened 2026-07-13
[BugFix] Fix scan-teardown use-after-free on MorselQueueFactory

sevev:
## Why I'm doing:

CN/BE nodes crashed repeatedly with a use-after-free during query teardown.
The crash stack is consistently:

```
IndividualMorselQueueFactory::mark_split_source_morsel_finished
  <- ConnectorChunkSource::_report_split_source_morsel_finished_once
  <- ConnectorChunkSource::close
  <- ~ScanOperator <- ~PipelineDriver <- ~FragmentContext
```

`FragmentContext` destroys its members in reverse declaration order.
`_morsel_queue_factories` was declared *after* `_pipelines`, so it was
destroyed *first*. While `_pipelines` is then being torn down, a scan
operator's `ConnectorChunkSource::close()` (run from `~ScanOperator`)
calls back into the `MorselQueueFactory` via
`ScanOperatorFactory::mark_split_source_morsel_finished()` — dereferencing
an already-freed object.

The callback in `close()` was added by #69095 as a safety net so
split-source completion is still reported when a chunk source exits
through non-EOF paths (cancel / error / limit reached). On those teardown
paths the factory is gone, so the report becomes a UAF. It only reproduces
for lake/connector scans that use split-source morsels and tear the
fragment down without reaching EOF.

## What I'm doing:

Declare `_morsel_queue_factories` before `_pipelines` so it is destroyed
*after* the pipelines and their operators, matching the existing ordering
constraint already documented for `_event_scheduler`. This is a
member-declaration-order fix only; the relative order versus
`_runtime_state` / `_plan` is unchanged.

Add a regression test (`FragmentContextExecRuntimeTest.MorselQueueFactoryOutlivesPipelines`)
that installs a tracking `MorselQueueFactory` and a tracking `Pipeline`
into a `FragmentContext` and asserts the factory is still alive when the
pipeline is destroyed, pinning the destruction order so a future member
reshuffle cannot silently reintroduce the crash.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/exec/runtime/fragment_context.h
+++ b/be/src/exec/runtime/fragment_context.h
@@ -206,6 +206,13 @@ private:
     std::unique_ptr<FragmentDictState> _fragment_dict_state;
     ExecNode* _plan = nullptr; // lives in _runtime_state->obj_pool()
     size_t _next_driver_id = 0;
+    // MorselQueueFactory must outlive the pipelines/operators declared below: a
+    // scan operator's ConnectorChunkSource::close() is invoked from ~ScanOperator
+    // during _pipelines teardown and calls back into the MorselQueueFactory via
+    // ScanOperatorFactory::morsel_queue_factory(). Declaring _morsel_queue_factories
+    // before _pipelines makes it destroyed after them, avoiding a use-after-free on
+    // non-EOF (cancel/error/limit) teardown paths.
+    MorselQueueFactoryMap _morsel_queue_factories;
     // Must outlive PipelineDriver observers owned by _pipelines.
     std::unique_ptr<EventScheduler> _event_scheduler;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

GavinMar (approved):

meegoo (approved):

PR #76267, opened 2026-07-13
[Enhancement] Skip getTable() for root in SHOW ANALYZE STATUS/JOB

dirtysalt:
## Why I'm doing:
`SHOW ANALYZE STATUS` and `SHOW ANALYZE JOB` resolve the underlying table for every row and run a privilege check before including that row in the output. `root` always holds every privilege, so the check is guaranteed to pass — but the `getTable()` call backing it still happens, and for external catalogs (Hive/Iceberg) that can mean a Glue/HMS round trip per row. In BYOC deployments, ops backends poll these `SHOW` statements as `root`, so this is pure overhead on every poll.

## What I'm doing:
- Added `StatisticUtils.isRootUser(context)`.
- In `ShowAnalyzeStatusStmt` and `ShowAnalyzeJobStmt`, root now skips `getTable()` entirely instead of just skipping the privilege check:

```
non-root:  getTable() -> privilege check -> render row (filtered out if table missing/denied)
root:      render row directly from the AnalyzeStatus/AnalyzeJob record, no getTable() at all
```

- Trade-off: for root, a row can still be shown after its table was dropped from the external catalog (previously filtered out via the `getTable()` failure). The column-count-vs-schema comparison used to decide between `ALL` and an explicit column list is skipped too, so root always sees the explicit list when one was recorded.
- Scope is intentionally limited to these two `SHOW` statements (not the `SHOW STATS META` family in `ShowExecutor`), matching where BYOC ops actually poll as root.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/ast/ShowAnalyzeJobStmt.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/ast/ShowAnalyzeJobStmt.java
@@ -26,6 +26,7 @@ import com.starrocks.sql.ast.expression.Expr;
 import com.starrocks.sql.ast.expression.LimitElement;
 import com.starrocks.sql.parser.NodePosition;
 import com.starrocks.statistic.AnalyzeJob;
+import com.starrocks.statistic.StatisticUtils;
 
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

PR #76275, opened 2026-07-13
[BugFix] Keep OR-nested predicates when a branch is only partially bitmap-indexed

Beauuu-0826:
## Why I'm doing:

The bitmap index evaluator erased any column predicate whose bitmap range was applied, even when that predicate sat under an `OR` whose branch was only **partially** covered by bitmap indexes.

Because an `OR`'s bitmap is a union (a superset) of its branch bitmaps, dropping the indexed leaves loses the per-branch correlation and produces false positives. For example, `(a = 1 AND b = 2) OR (a = 2 AND b = 3)` with only `a` bitmap-indexed wrongly matched rows like `(a = 1, b = 3)` and `(a = 2, b = 2)`: the residual predicate collapsed to `(b = 2 OR b = 3)` over the scan range `{a in (1, 2)}`, so any row with `a in (1,2)` and `b in (2,3)` slipped through.

## What I'm doing:

Track a per-node "fully covered" flag (every leaf in the subtree is bitmap-indexed) and thread an `erasable` flag through the seeker. Erasability is preserved across `AND` nodes but revoked when descending into a not-fully-covered `OR`, so predicates below such an `OR` are retained in the residual tree.

Scan-range narrowing is unchanged (the union bitmap remains a valid superset), so partially covered `OR`s keep their filtering benefit without unsafe predicate erasure.

Adds a SQL regression test under `test_scan/test_pushdown_or_predicate` that isolates the bitmap-index path and asserts the bitmap-enabled result matches the bitmap-disabled control.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/rowset/bitmap_index_evaluator.cpp
+++ b/be/src/storage/rowset/bitmap_index_evaluator.cpp
@@ -47,6 +47,8 @@ struct BitmapIndexInitializer {
         }
 
         parent->_ctx.is_node_support_bitmap[&node] = true;
+        // A column predicate backed by a bitmap iterator is a fully-covered leaf.
+        parent->_ctx.is_node_fully_covered[&node] = true;
         auto& col_ctx = _get_or_create_column_context(bitmap_iter, cid, parent_node_ctx, parent_type);
         col_ctx.nodes.emplace_back(&node);
         return true;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

ZiheLiu (approved):

trueeyu (approved):

Beauuu-0826:
@mergify backport branch-3.5

Beauuu-0826:
@mergify backport branch-4.0

PR #76287, opened 2026-07-13
[Feature] Support builtin audit loader to persist audit events into internal table

### LINKED ISSUE #76291: Support a builtin audit loader that persists audit events into an internal table
## Feature request

Analyzing audit data with SQL currently requires installing the external
[fe-plugins-auditloader](https://github.com/StarRocks/fe-plugins-auditloader)
plugin, which has several long-standing pain points:

1. Operational cost: users must manually create the audit database/table,
   download a zip, edit `plugin.conf`, distribute it to every FE, and run
   `INSTALL PLUGIN`; every StarRocks upgrade requires re-validating plugin
   compatibility.
2. Security risk: `plugin.conf` stores a database user and password in
   plaintext (or weakly encrypted) to stream-load audit data back into the
   cluster, which violates common enterprise security baselines.
3. Version skew: the plugin is released independently from the server and
   uses reflection to stay compatible with multiple releases, so it only
   collects the common subset of audit fields, and new fields lag behind.
4. Silent mismatch: a plugin/server version mismatch can silently drop
   audit data.

Build the audit-loading capability into the FE as a native, toggleable
feature:

- A builtin AUDIT plugin persists audit events (queries and connections)
  into an internal table `_statistics_.starrocks_audit_tbl` so audit data
  can be analyzed directly with SQL. No external zip, no plaintext
  credentials: the write path reuses the cluster-internal stream load
  mechanism already used by the `query_history` system table.
- Disabled by default and gated by mutable FE configs:
  `enable_audit_loader` (default `false`),
  `audit_loader_load_interval_seconds` (default 60),
  `audit_loader_batch_max_bytes` (default 50MB, byte-capped per-FE buffer
  that drops and counts events under extreme overload to protect FE
  memory).
- The audit table is created automatically by the leader FE
  (day-partitioned, `partition_live_number=30`, recreated automatically if
  dropped by mistake).
- Conflict protection: while any external dynamic AUDIT plugin (e.g. the
  auditloader zip plugin) is installed, the builtin loader stays inert to
  avoid importing audit data twice, and recovers automatically after the
  external plugin is uninstalled.
- Reliability rules: batches are only removed from the in-memory buffer
  after a successful load (retry on failure); empty batches are never
  submitted; oversized statements are truncated to the 1MB column width
  instead of failing the whole batch.
- Coexists with and is independent of the file-based `fe.audit.log`;
  slow-query classification follows the existing `qe_slow_log_ms` config.

1. Keep using the external fe-plugins-auditloader: leaves the plaintext
   credential and version-skew problems unsolved, and requires per-cluster
   manual deployment.
2. Ship the plugin source inside the main repo but keep the
   `INSTALL PLUGIN` zip flow: still requires credentials in `plugin.conf`
   and manual installation on every FE.
3. Collect `fe.audit.log` with an external log pipeline (Filebeat/Vector
   etc.) into another system: introduces extra infrastructure and loses
   the ability to analyze audit data in StarRocks with plain SQL.

**Additional context**

- Implementation PR: #76287
- The table schema follows the documented layout of the existing
  audit-loader plugin (see the Audit Loader page in the docs), with the
  `stmt` column widened to `VARCHAR(1048576)` and `clientIp` widened to
  `VARCHAR(64)` for IPv6 host-port strings.
- Non-goals for the first version: lossless auditing under extreme
  overload (the buffer is byte-capped by design), a configurable target
  table (the credential-free internal load whitelist requires a fixed
  table), and per-user/expression filtering (may be added later based on
  feedback).

iatman:
Rebuild the fe-plugins-auditloader capability as a native FE feature: a builtin AUDIT plugin persists audit events into `starrocks_audit_db__.starrocks_audit_tbl__` so audit data can be analyzed with SQL, without installing an external plugin zip or storing plaintext credentials in `plugin.conf`. The database and table names match what the external plugin creates, so operators keep their existing habits.

## How it works

- `AuditEventProcessor` dispatches AFTER_QUERY / CONNECTION events to the builtin `AuditLoaderPlugin` (registered in `PluginMgr.initBuiltinPlugins`).
- `exec()` only serializes the event into a JSON row (streamed with Gson's `JsonWriter`) and appends it to a byte-bounded concurrent queue on the local FE. It never blocks the audit worker thread and never throws.
- `AuditLoaderMgr` (a `FrontendDaemon` on a 5s cycle, running on every FE) detects conflicts, ensures the database and table exist (created by the leader, day-partitioned keeping the 30 most recent daily partitions, recreated automatically if dropped), and flushes batches through the internal stream load path in `com.starrocks.summary.StreamLoader`.
- A single `COLUMN_SPECS` list is the source of truth for the 29 columns: the CREATE TABLE statement, the JSON row and the per-column truncation widths are all derived from it, so adding a column is a one-place change.
- The load sends no explicit column list, so JSON keys are mapped to table columns by name. A key without a matching column is ignored rather than failing the batch, which keeps an upgraded cluster loading even before its table gains a newly added column.

## Reliability rules

Several of these exist specifically to avoid defects observed in `QueryHistoryMgr`:

- Never send an empty batch (an empty payload aborts a load transaction).
- Copy-then-remove: rows leave the queue only after the load succeeds.
- A failed flush is reported as a failure even when an earlier batch in the same call landed, so the flush clock is not advanced and the batch still at the head is retried on the next 5s cycle rather than a whole interval later.
- A batch that keeps failing is discarded after `MAX_BATCH_FLUSH_RETRY` (6) attempts, logged with the time range it covered and counted, so one undeliverable batch cannot block everything queued behind it. That takes about 30 seconds when attempts fail immediately and up to 90 seconds when each one runs into the request timeout.
- While the table is missing the flush is skipped entirely without touching the queue, so buffered rows survive and no retry budget is consumed.
- Events beyond the byte cap are dropped and counted, with a rate-limited warning, to protect FE memory.
- Every path that loses an event increments the same counter, surfaced in a periodic status line: buffer size, drops since the last line, total drops, consecutive flush failures.

## Feature boundaries

- **Disabled by default.** Three mutable FE configs: `enable_audit_loader` (false), `audit_loader_load_interval_seconds` (60), `audit_loader_batch_max_bytes` (50MB). Both numeric knobs are clamped at each end with a rate-limited warning: the batch cap to 1.2MB–512MB, the interval to 0–1 day (0 means flushing every daemon cycle).
- **Stays inert while any external dynamic AUDIT plugin is installed**, to avoid importing audit data twice, and recovers automatically after it is uninstalled. The check is deliberately broad: it fires for any dynamic AUDIT plugin regardless of name or target table.
- **Credential-free load whitelist**: `FrontendServiceImpl.checkIsInternalLoad` now accepts exactly two (database, table) pairs — the pre-existing `_statistics_.query_history` and this feature's table. The host / nodeName / alive-FE checks are unchanged; the pair check was hoisted out of the per-FE loop since it does not depend on the FE being examined.
- **Lossy by design under extreme overload.** A flush is triggered once the buffer reaches half the cap, leaving the other half to absorb inflow until the next cycle; beyond that, excess events are dropped and counted.
- **Delivery is at-least-once.** A retry builds a new label, so a batch the BE already committed but whose response was lost is written again and leaves duplicate rows. Making retries idempotent would mean treating a "label already exists" rejection as success, which risks masking real failures, and is deliberately left out.
- **No schema evolution.** An existing table is never altered. Because the load sends no column list, a column added to `COLUMN_SPECS` is ignored by a table that predates it rather than failing every batch; the new field simply stays uncollected until the table is altered or recreated.
- `stmt` is stored as `VARCHAR(1048576)`; longer statements are truncated to 1MB by UTF-8 bytes without splitting a multi-byte character. Every other VARCHAR column is truncated to its own declared width.
- Timestamps use the cluster `time_zone`, not the JVM default, so rows from every FE share one timeline and land in the right daily partition.
- Independent of and coexists with `fe.audit.log`; slow_query classification follows `qe_slow_log_ms`.

## Behaviour outside this feature

`StreamLoader` is shared with `QueryHistoryMgr`, and two fixes there apply to both callers:

- The load label now comes from a process-wide sequence instead of `System.identityHashCode(this)`, which could repeat within the same second and cost a retry to a duplicate-label rejection.
- The request now carries a timeout. It previously had none at all, so a BE that accepted the connection and then stopped answering could wedge the caller's daemon thread indefinitely. The audit path uses a short fixed bound plus a small `max_filter_ratio`; `QueryHistoryMgr` keeps its strict all-or-nothing semantics and uses `Config.stream_load_default_timeout_second`, guarded against that unvalidated config being set to `<= 0`.

## Verification

- **Unit tests**: 131 cases pass, checkstyle 0 violations. `AuditLoaderTest` 45, `FrontendServiceImplTest` 70, `LogUtilTest` 11, `PluginMgrTest` 5. Coverage includes JSON key set pinned to the generated DDL, non-finite doubles, null versus empty-string semantics, cluster time zone, per-column truncation driven by the DDL itself, config clamping at both ends, the retry counter lifecycle, and the whitelist's negative cases.
- **End-to-end on a live cluster**: count equality (50/50, 2400/2400, 4200/4200), special characters round-tripping byte-for-byte (quotes, backslashes, tabs, newlines, control characters, CJK, four-byte emoji, JSON literals), 1100KB statement truncated to exactly 1048576 bytes, all 29 columns populated correctly.
- **Table lifecycle**: DROP TABLE and DROP DATABASE both self-heal within one cycle; buffered rows and rows produced during the outage are all preserved and land afterwards; an operator's `partition_live_number` change is not overwritten; TTL verified to actually drop the oldest partitions (7 → 3, newest kept).
- **Failure paths**: with the BE unavailable, batches are retried and discarded on a measured 30s cadence with the lost time range logged, warnings stay rate-limited to roughly one per minute, the loader never stops on its own, and it recovers without an FE restart. Adding a NOT NULL column to the table breaks every load — the load sets absent columns to NULL rather than to their DEFAULT — and dropping the column restores loading; this is called out in the docs.
- **Overload**: at a sustained rate the old interval-only trigger could not have absorbed, 2400/2400 rows land with zero drops thanks to the half-cap watermark.
- **Memory**: A/B against the pre-change build under an identical workload shows peak heap down 21% (372MB → 294MB) from the streaming serialization, with settled heap returning to baseline and zero full GCs on both sides.
- **Docs**: the three FE configs are documented in en / zh / ja.

## Relationship with the audit error message added in #78798

That change adds an `errorMessage` field to `AuditEvent`, normalized and length-capped by
`audit_log_error_message_max_length`. This feature deliberately does not add a matching column: the
audit table is meant to answer when, who, how long, what statement, which tables and what outcome,
and `errorCode` already identifies the failure. Storing the full message text for every failed
statement would cost far more than it answers.

The practical consequence is that `fe.audit.log` carries the message while the audit table carries
only the code. Resolving a code to its text is better served later by a separate mapping table that
can be joined, which keeps the audit table itself narrow.

## Known limitation

The follower-side branch that warns when the audit table stays invisible for a long time is not exercised at runtime here: an FE is registered by host and binds its ports on the wildcard address, so a second FE cannot be started on one machine, and verifying it needs a multi-host or containerized cluster. The branch only emits a log line and does not touch the data path; it has been confirmed by code review only.

## Why I'm doing:

Audit data is only useful if it can be queried. Today that requires installing the external fe-plugins-auditloader zip and putting a username and password in `plugin.conf` in plaintext. This makes it a native FE capability instead: no plugin to install, no credentials to store, and the load runs over the FE's own internal identity.

## What I'm doing:

Adding a builtin AUDIT plugin plus an `AuditLoaderMgr` daemon that batches audit events in FE memory and stream-loads them into `starrocks_audit_db__.starrocks_audit_tbl__`, with the database and table created and repaired automatically. The feature is off by default and stays inert whenever the external plugin is present.

Fixes #76291

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [x] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

iatman:
## How this compares to the existing self-healing mechanisms

The audit table needs the same kind of upkeep that `TableKeeper` and
`StatisticsMetaManager#refreshStatisticsTable` already provide for other internal tables.
This is what was reused, what was left out, and why.

### Capability comparison

| | `TableKeeper` | `refreshStatisticsTable` | `AuditLoaderMgr` (this PR) |
|---|---|---|---|
| Create database | No — logs a warning and returns | Yes, `CreateDbStmt` in a retry loop | Yes, `CREATE DATABASE IF NOT EXISTS` |
| Create table | Yes, `executeDDL` on SQL text | Yes, `LocalMetastore.createTable(stmt)` from an AST | Yes, `executeDDL` on generated SQL text |
| Replication self-heal | Yes, `correctTable()` | Yes, `alterSystemTableReplicationNumIfNecessary()` | Yes, `correctReplicationNum()` |
| TTL / `partition_live_number` | Yes, `changeTTL()` | No | Set at creation, never corrected (deliberate, see below) |
| Schema evolution | No | Yes, but hardcoded to one column of one table | No (deliberate, see below) |

### Threading

| | Thread | Interval | Role |
|---|---|---|---|
| `TableKeeper` | `table-keeper` | `Config.table_keeper_interval_second` | Leader only |
| `StatisticsMetaManager` | `statistics-meta-manager` | `Config.statistic_manager_sleep_time_sec` | Leader only |
| `AuditLoaderMgr` | `AuditLoader` | 5s flush, maintenance every 12th cycle | **All FE roles** |

The role difference is intrinsic rather than a preference: audit events are produced and
buffered on every FE, so each FE has to flush its own buffer. Metadata writes are still
leader-only, so the leader check moved inside the cycle (`ensureAuditTable` returns false on
a follower, and the replication self-heal is guarded separately) instead of gating the whole
thread at startup.

### Why `TableKeeper` is not reused

1. `TableKeeperDaemon` runs every registered keeper unconditionally, and a keeper creates its
   table whenever it is missing. This feature is disabled by default, so registering it would
   create `starrocks_audit_db__.starrocks_audit_tbl__` on every cluster whether or not audit
   loading is enabled. `TableKeeper` has no enable predicate, so this would mean changing
   infrastructure shared by six other tables.
2. `TableKeeper` does not create databases — its six current users all live in `_statistics_`,
   which `StatisticsMetaManager` creates for them. This feature owns a separate database that
   nobody else creates, so the database bootstrap has to exist here regardless; reusing
   `TableKeeper` would only cover the table half.
3. `TableKeeper` runs leader-only while the flush loop must run on every FE, so reuse would
   mean splitting maintenance into a second thread and keeping state in two places.

If `TableKeeper` ever grows database creation and a conditional-enable predicate, moving the
audit table onto it would be the natural consolidation, and it would also let the existing
`// TODO: add statistic table` be addressed at the same time.

### Deliberately not implemented

- **Schema evolution.** An existing table is never altered. Because the load no longer sends an
  explicit `columns` header, a column added to `COLUMN_SPECS` is ignored by a table that
  predates it rather than failing every batch, so an upgraded cluster keeps collecting audit
  data; the new field simply stays uncollected until the table is altered or recreated. This is
  documented on `COLUMN_SPECS`. The natural implementation is to diff the spec names against the
  live schema and issue `ADD COLUMN` for the missing ones — add only, never drop or modify, and
  skipping shadow columns while a schema change is in flight.
- **TTL correction.** `partition_live_number` is set to 30 at creation but never corrected,
  unlike `TableKeeper.changeTTL()`. The documentation tells operators they can tune it with
  `ALTER TABLE ... SET ("partition_live_number"="N")` and that the system will not overwrite the
  change, so a TTL self-heal would silently revert their setting within a minute.
- **Non-range partition handling in the replication self-heal.** `TableKeeper.alterTableReplicas`
  branches on `isRangePartition()`; this implementation always issues the two range statements
  because the table is created with `PARTITION BY date_trunc('day', timestamp)`. If someone
  recreates the table unpartitioned, `MODIFY PARTITION(*)` fails, is caught, and logs a warning
  once per maintenance cycle — degraded but harmless.
- **Creation logging.** Creating the database or the table is not logged; only failures are, at
  WARN. Combined with the recreation behaviour below, an operator has to read the create time
  from `information_schema` rather than the FE log.

### Operational note verified on a live cluster

The database and table are recreated within one daemon cycle after a `DROP`, which takes the
name back before `RECOVER TABLE` / `RECOVER DATABASE` can run:

```
DROP TABLE starrocks_audit_db__.starrocks_audit_tbl__;   -- 288 rows
                                                          -- recreated 6s later
RECOVER TABLE starrocks_audit_db__.starrocks_audit_tbl__;
ERROR 1050 (42S01): Table 'starrocks_audit_tbl__' already exists
```

Turning `enable_audit_loader` off first makes the recover succeed. The dropped data otherwise
stays in the recycle bin until `Config.catalog_trash_expire_second` elapses. This is documented
on `ensureAuditTable`. The internal statistics database behaves the same way.

gengjun-git (approved):

PR #76288, opened 2026-07-13
[Enhancement] Add orthogonal bitmap aggregate functions

priterzhang:
## Why I'm doing:

Add bitmap aggregate helpers for retention/loss analysis scenarios that need either the actual bitmap result or per-key count arrays from grouped bitmap rows.

## What I'm doing:

- Add `orthogonal_bitmap_intersect(bitmap_column, column_to_filter, filter_values...)` to return the bitmap intersection across selected key groups.
- Add `orthogonal_bitmap_difference(bitmap_column, column_to_filter, filter_values...)` to return the first key bitmap minus the selected-key intersection.
- Add `bitmap_intersect_count_each_column(bitmap_column, column_to_filter, filter_values...)` to return `ARRAY<BIGINT>` counts like `[count(head), count(head & key2), ...]`.
- Add `bitmap_difference_count_each_column(bitmap_column, column_to_filter, filter_values...)` to return `ARRAY<BIGINT>` counts like `[count(head), count(head - key2), ...]`.
- Register these functions in FE/BE aggregate resolution and add SQL regression coverage.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [x] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] I have added SQL regression coverage for the new functions.
- [x] I have run `git diff --check`.


fe/fe-core/src/main/java/com/starrocks/sql/analyzer/FunctionAnalyzer.java
```diff
@@ -422,27 +422,31 @@ private static void analyzeBuiltinAggFunction(String fnName,
                     functionCallExpr.getPos());
         }
 
-        if (fnName.equals(FunctionSet.INTERSECT_COUNT)) {
+        if (fnName.getFunction().equals(FunctionSet.INTERSECT_COUNT)
```
caneGuy, line 425: If fnName is of type string, the compiler will report an error.

be/src/exprs/agg/orthogonal_bitmap_difference.h
```diff
@@ -0,0 +1,176 @@
...
+            old_size += intersect.size();
+            offsets.set(i + 1, old_size);
+        }
+    }
+
+    void finalize_to_column(FunctionContext* ctx, ConstAggDataPtr __restrict state, Column* to) const override {
+        auto& intersect = this->data(state).intersect;
+        auto* col = down_cast<BitmapColumn*>(to);
+
+        auto head_arg_column = ctx->get_constant_column(2);
```
caneGuy, line 155: it may crash or return wrong result here

PR #76301, opened 2026-07-13
[Enhancement] Gate the one-tablet optimization by tablet row count

stephen-shelby:
## Why I'm doing:

When a scan is pruned to a single tablet, StarRocks applies the "one-tablet optimization": it runs the aggregation in a single phase and gathers the output on a single node/instance, skipping the shuffle/exchange a normal distributed plan would use. This is a big win when the tablet is small. But the optimization only keys off the tablet *count* (`<= 1`), never its *size*. When that single tablet is large, the optimization pins the entire scan + aggregation onto one node/instance and serializes it — far slower than a normal two-phase, multi-instance plan. There is currently no guard against applying the optimization to an oversized single tablet.

## What I'm doing:

Add a session variable `one_tablet_opt_max_tablet_rows` (default `10000000`; set to `-1` to disable) that turns the one-tablet optimization off when the single selected tablet's fuzzy row count exceeds the threshold. The variable is user-visible/settable and documented under `docs/{en,zh,ja}/sql-reference/System_variable.md`.

The gate is applied on both the logical and physical sides so the two plans stay consistent:

1. **Logical source** — `LogicalProperty.visitLogicalTableScan` returns `OneTabletProperty.notSupport()` when `Utils.isSelectedSingleTabletTooLarge(...)` holds. This is where the behavior actually changes: the optimizer stops requesting the single-instance / one-phase-aggregation distribution, so `RequiredPropertyDeriver` falls back to the normal path (a gather for non-grouping aggregation, or a hash shuffle for `GROUP BY`).

2. **Physical output guard** — `PlanFragmentBuilder.createOutputFragment` decides whether the final `ResultSink` can be attached directly to the top fragment (valid only when the whole plan runs on a single instance) or needs a dedicated gather `ExchangeNode`. Its single-tablet shortcut fires whenever all scans touch `<= 1` tablet. Once the optimization is disabled, a `GROUP BY` aggregation shuffles the top fragment into multiple instances while the scan still touches only one tablet, so the shortcut would wrongly pin a single-tablet direct output onto a multi-instance fragment (producing multiple result sinks). The new `hasOversizedSingleTablet(...)` guard — which reuses the same `Utils.isSelectedSingleTabletTooLarge(...)` predicate — makes the shortcut step aside in exactly that case, so the output is gathered through an exchange.

The predicate uses the lock-free `Tablet.getFuzzyRowCount()` (the real row count on both `LocalTablet` and `LakeTablet`); an unresolved / not-yet-collected (`0`) row count is treated as "not too large", so the optimization stays on for small/medium tablets and for tablets whose row count has not been reported yet. `maxTabletRows < 0` disables the gate entirely, preserving the previous behavior regardless of tablet size. The gate applies to both cloud-native (lake) and olap native tables.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

ZiheLiu (approved):

satanson (approved):

PR #76307, opened 2026-07-13
[BugFix] Honor CREATE TABLE IF NOT EXISTS on Iceberg REST catalogs

ycra-dev:
## Why I'm doing:

Context: I run StarRocks with an Iceberg (REST) catalog in production. This is a follow-up to #75017 (`Honor CREATE DATABASE IF NOT EXISTS on Iceberg REST catalogs`, already backported to 3.5): I initially reported only the `CREATE DATABASE` case I had hit, and while further auditing the same exception-translation gap I found `CREATE TABLE IF NOT EXISTS` has the identical problem, which this PR fixes.

On Iceberg catalogs (notably the REST catalog), `CREATE TABLE IF NOT EXISTS <tbl>` can fail with an error when the table already exists, instead of being silently ignored as the `IF NOT EXISTS` clause promises.

The root cause is an exception-type mismatch at the StarRocks ↔ Iceberg boundary — the same class of bug already fixed for `CREATE DATABASE IF NOT EXISTS` in #75017. `MetadataMgr.createTable()` pre-checks existence with `tableExists()` and honors `IF NOT EXISTS` there. `tableExists()` is a live catalog probe, so a single sequential statement is correctly handled; but under concurrent creation of the same table there is a TOCTOU window where several statements pass the pre-check (table not yet created) and then race into the actual create. The Iceberg client (notably the REST catalog) then throws its own `org.apache.iceberg.exceptions.AlreadyExistsException` for the losers of the race, which is an unrelated class to StarRocks' `com.starrocks.common.AlreadyExistsException`. That connector-level exception leaks through unchanged, so the upper layer never recognizes it as "already exists" and the statement errors out.

Reproduced on StarRocks 4.1.1 + Iceberg REST catalog: 23 of 24 concurrent `CREATE TABLE IF NOT EXISTS` statements failed with a leaked `org.apache.iceberg.exceptions.AlreadyExistsException: Table already exists: ...` (a single serial statement is correctly a no-op, because the live pre-check catches it).

## What I'm doing:

Translate the connector-level `org.apache.iceberg.exceptions.AlreadyExistsException` thrown by `icebergCatalog.createTable()` into StarRocks' `com.starrocks.common.AlreadyExistsException`, carrying the table name in the message. This mirrors the `CREATE DATABASE IF NOT EXISTS` fix (#75017). The upper layer then recognizes the condition and honors `IF NOT EXISTS` (silently ignore) or reports a clean `ERR_TABLE_EXISTS_ERROR`.

- `IcebergMetadata.createTable()`: wrap the `createTable` call in try/catch and rethrow the Iceberg exception as StarRocks `AlreadyExistsException`.
- Thread the checked `AlreadyExistsException` through the connector chain so the translated exception can propagate: `ConnectorMetadata`, `CatalogConnectorMetadata`, `UnifiedMetadata`.
- `MetadataMgr.createTable()` / `createTemporaryTable()`: catch the translated exception and reuse the existing `IF NOT EXISTS` handling (return `false` when set, otherwise report `ERR_TABLE_EXISTS_ERROR`). The exception is absorbed here, so the DDL/statement executors and other callers are unaffected.
- Added a unit test `testCreateTableTranslatesConnectorAlreadyExists` that mocks the catalog to throw the Iceberg exception and asserts the translated StarRocks exception (type + message) is raised.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/CatalogConnectorMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/CatalogConnectorMetadata.java
@@ -312,7 +312,7 @@ public class CatalogConnectorMetadata implements ConnectorMetadata, DelegatingCo
     }
 
     @Override
-    public boolean createTable(ConnectContext context, CreateTableStmt stmt) throws DdlException {
+    public boolean createTable(ConnectContext context, CreateTableStmt stmt) throws DdlException, AlreadyExistsException {
         return normal.createTable(context, stmt);
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorMetadata.java
@@ -319,7 +319,7 @@ public interface ConnectorMetadata {
         return Lists.newArrayList();
     }
 
-    default boolean createTable(ConnectContext context, CreateTableStmt stmt) throws DdlException {
+    default boolean createTable(ConnectContext context, CreateTableStmt stmt) throws DdlException, AlreadyExistsException {
         throw new StarRocksConnectorException("This connector doesn't support creating tables");
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -337,7 +337,7 @@ public class IcebergMetadata implements ConnectorMetadata {
     }
 
     @Override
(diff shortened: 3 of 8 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

PR #76308, opened 2026-07-13
[BugFix] Honor CREATE TABLE IF NOT EXISTS on Iceberg REST catalogs (backport #76307)

ycra-dev:
## Why I'm doing:

Backport of #76307 (`Honor CREATE TABLE IF NOT EXISTS on Iceberg REST catalogs`) to `branch-3.5`.

On Iceberg catalogs (notably the REST catalog), `CREATE TABLE IF NOT EXISTS <tbl>` can fail with an error when the table already exists, instead of being silently ignored as the `IF NOT EXISTS` clause promises. The Iceberg client throws its own `org.apache.iceberg.exceptions.AlreadyExistsException`, which is unrelated to StarRocks' `com.starrocks.common.AlreadyExistsException`; under concurrent creation of the same table the live `tableExists()` pre-check is bypassed by a TOCTOU race, and the untranslated connector exception leaks through and defeats the `IF NOT EXISTS` clause.

This is the same exception-translation gap already fixed for `CREATE DATABASE IF NOT EXISTS` on this branch in #75019 (backport of #75017).

## What I'm doing:

Translate the connector-level `org.apache.iceberg.exceptions.AlreadyExistsException` thrown by `icebergCatalog.createTable()` into StarRocks' `com.starrocks.common.AlreadyExistsException` in `IcebergMetadata.createTable()`, and honor `IF NOT EXISTS` in `MetadataMgr` (return `false` when set, otherwise report `ERR_TABLE_EXISTS_ERROR`). Same fix as #<MAIN_PR>, manually adapted to the 3.5 connector API:

- 3.5's `createTable` has no `ConnectContext` parameter, and the Iceberg delegate takes 6 args (no `context`, no `sortOrder`); the signatures/call sites differ from the main branch, so this is a manual backport rather than a clean cherry-pick. The translation logic itself is identical.
- Threads the checked `AlreadyExistsException` through the connector chain (`ConnectorMetadata`, `CatalogConnectorMetadata`, `UnifiedMetadata`); `MetadataMgr.createTable()` / `createTemporaryTable()` catch it, so the exception is absorbed and callers are unaffected.
- Added a unit test `testCreateTableTranslatesConnectorAlreadyExists` mirroring the existing `testCreateDbTranslatesConnectorAlreadyExists`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/CatalogConnectorMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/CatalogConnectorMetadata.java
@@ -257,7 +257,7 @@ public class CatalogConnectorMetadata implements ConnectorMetadata {
     }
 
     @Override
-    public boolean createTable(CreateTableStmt stmt) throws DdlException {
+    public boolean createTable(CreateTableStmt stmt) throws DdlException, AlreadyExistsException {
         return normal.createTable(stmt);
     }
 
--- a/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/ConnectorMetadata.java
@@ -259,7 +259,7 @@ public interface ConnectorMetadata {
         return Lists.newArrayList();
     }
 
-    default boolean createTable(CreateTableStmt stmt) throws DdlException {
+    default boolean createTable(CreateTableStmt stmt) throws DdlException, AlreadyExistsException {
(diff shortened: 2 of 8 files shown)


trueeyu:

PR #76310, opened 2026-07-13
[BugFix] Fix Arrow Flight Prepared Statement Forwarding

chris-celerdata:
## Why I'm doing:
When behind a round-robin load balancer, a `CreatePreparedStatement/ClosePreparedStatement` that lands on a different FE than the one holding the session is forwarded with the protobuf message name (`arrow.flight.protocol.sql.ActionCreatePreparedStatementRequest`) as the action type instead of the Flight SQL action type (`CreatePreparedStatement`), so the remote FE rejects it as `Unrecognized request` and every ADBC client that opens with a prepared statement fails.
## What I'm doing:
Passing the correct Flight SQL action-type string from each call site into `forwardActionToRemoteFE` so the forwarded doAction matches the remote FE's dispatcher.

Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
+++ b/fe/fe-core/src/main/java/com/starrocks/service/arrow/flight/sql/ArrowFlightSqlServiceImpl.java
@@ -40,6 +40,7 @@ import com.starrocks.sql.plan.ExecPlan;
 import com.starrocks.system.ComputeNode;
 import com.starrocks.system.SystemInfoService;
 import com.starrocks.thrift.TUniqueId;
+import org.apache.arrow.flight.Action;
 import org.apache.arrow.flight.CallStatus;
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

jaogoy (approved):
Is it better to use a map to map from `request.getDescriptorForType().getFullName()` to `actionType`?

chris-celerdata:
No, there is no need for a map. It would just add more complexity through declaring an additional structure. This is the only place the value is needed.

HangyuanLiu (approved):

CelerData-Reviewer:
@codex review

PR #76316, opened 2026-07-14
[BugFix] Fix SparseRangeIterator::remaining_rows() undercounting after the first sub-range

stephen-shelby:
## Why I'm doing:

`SparseRangeIterator<T>::remaining_rows()` returns the wrong count once the iterator has advanced past its first sub-range. It measures the leading (partially consumed) sub-range against `_ranges[0].end()` — the end of the **first** sub-range — instead of `_ranges[_index].end()`, the end of the **current** sub-range.

The sub-ranges are disjoint and ascending, so whenever `_index > 0` the cursor `_next_rowid` is already at or beyond `_ranges[0].end()`. The `Range` constructor zeroes any range whose `begin >= end`, so that leading term collapses to an empty range and contributes `0`. The rows still remaining in the current sub-range are therefore dropped, and `remaining_rows()` under-reports. Because the result feeds `size_t` subtraction at the call site, the undercount can also underflow to a huge bogus value.

On the current main the only caller is `SegmentIterator::_try_to_update_ranges_by_runtime_filter()`, which computes `runtime_stats_filtered` when a runtime filter arrives mid-scan:

```cpp
size_t prev_size = _range_iter.remaining_rows();
_range_iter = _range_iter.intersection(r, &res);
...
_opts.stats->runtime_stats_filtered += (prev_size - _range_iter.remaining_rows());
```

## What I'm doing:
Measure the leading partial sub-range against the current sub-range's end (_ranges[_index].end()) instead of the first sub-range's end (_ranges[0].end()), so the un-consumed rows of the current sub-range are counted:

- auto range = Range<T>(_next_rowid, _range->_ranges[0].end());
+ auto range = Range<T>(_next_rowid, _range->_ranges[_index].end());
Add a SparseRangeIteratorTest.remaining_rows regression test that advances an iterator across three disjoint sub-ranges ([0,10) [20,40) [50,70)) and checks remaining_rows() at each position. At the _index > 0 checkpoints the pre-fix code returns 20 / 20 / 0 while the correct values are 40 / 35 / 20; the test fails before the fix and passes after.
Fixes #issue

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage_primitive/range.h
+++ b/be/src/storage_primitive/range.h
@@ -468,7 +468,7 @@ inline size_t SparseRangeIterator<T>::remaining_rows() const {
         return 0;
     }
     size_t res = 0;
-    auto range = Range<T>(_next_rowid, _range->_ranges[0].end());
+    auto range = Range<T>(_next_rowid, _range->_ranges[_index].end());
     res += range.span_size();
     for (size_t i = _index + 1; i < _range->_ranges.size(); i++) {
         res += _range->_ranges[i].span_size();
--- a/be/test/storage_primitive/range_test.cpp
+++ b/be/test/storage_primitive/range_test.cpp
@@ -204,4 +204,38 @@ TEST(SparseRangeIteratorTest, intersect_test) {
     }
 }
 
+TEST(SparseRangeIteratorTest, remaining_rows) {
+    // Three disjoint ranges, 10 + 20 + 20 = 50 rows in total.
+    SparseRange<> r1({{0, 10}, {20, 40}, {50, 70}});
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

trueeyu (approved):

stdpain (approved):

PR #76341, opened 2026-07-14
[Enhancement] Support metadata-only add of a trailing sort-key column on shared-data range-distribution tables

xiangguangyxg:
## Why I'm doing:

On shared-data (cloud-native) range-distribution tables, adding a key column currently routes to a full K-tablet data rewrite (O(data)), even when the change only appends a brand-new **trailing** sort-key column with a constant/NULL default. In that case the on-disk composite-key encoding of existing rows stays byte-identical (old segments serve the new column as a uniform default, which is non-discriminating in the per-segment sorted seek and the AGG/UNIQUE merge/aggregate), so the rewrite is unnecessary. This makes a common schema change far more expensive than it needs to be.

## What I'm doing:

Add an O(1) **metadata-only** fast path for `ALTER TABLE ... ADD COLUMN <c> KEY DEFAULT <const>` that appends one or more brand-new trailing sort-key columns on shared-data range-distribution **DUPLICATE / AGGREGATE / UNIQUE** tables, instead of the K-tablet data rewrite:

- Reproject each tablet's boundary tuple by appending one canonical NULL sentinel (`Variant.nullVariant`) per newly added column, preserving unbounded sides — routing-neutral, so every row still maps to the same tablet.
- Push the reprojected per-tablet ranges to the BE over the existing tablet-meta update channel via a new optional `tablet_range` field (thrift `TTabletMetaInfo`, proto `MetadataUpdateInfoPB`, reusing `TTabletRange`/`TabletRangePB`).
- BE apply validates the transition (new = old + exactly N trailing typed NULLs, one per added column; existing sort keys unchanged) and structural well-formedness, archives the old rowset schema with the non-clearing pattern, and writes `TabletMetadataPB.range`. `Rowset::get_seek_range()` decodes a per-rowset or tablet-fallback range with whichever available schema's sort-key arity matches the range's bound arity, so old rowsets stay readable across the arity change — including after a later automatic tablet reshard/split that stamps a current-arity range onto rowsets still carrying the archived (smaller) schema.
- FE flips tablet ranges in place under the table write lock in two phases (all fallible validation before the `FINISHED` journal record; a must-not-fail install after), from a persisted immutable target-range map, with exact-state replay idempotency. The range pruner degrades to a full scan (never drops tablets) on a transient arity mismatch.

Every non-eligible case falls through to the existing data-rewrite / reject paths unchanged: DROP, sort-key type widens, PRIMARY KEY, AGG/UNIQUE key drop, promoting an existing column to a key, a batch combined with other alter clauses, colocate / AUTO_INCREMENT / multi-index / temp-partition tables, and sort-key arities beyond the supported cap.

The result schema is byte-for-byte identical to what the rewrite path produces; only the execution mechanism and cost change. Covered by FE unit tests (routing/classifier, boundary reprojection, in-place flip + idempotency, pruner defense) and BE unit tests (transition/structural validation, non-clearing archival, seek-range decode incl. read-before/after-compaction for DUP/AGG/UNIQUE and the reshard-after-add arity-matching case). Validated end-to-end on a shared-data cluster (1 FE + 3 CN): the metadata-only add on DUP / AGG / UNIQUE (single- and multi-bounded-tablet) with `cloud_native_fast_schema_evolution_v2` both on and off — data conserved, per-tablet boundaries reprojected with the trailing NULL sentinel, routing preserved across reprojected tablet boundaries, and reads remain correct after a subsequent automatic tablet reshard/split of a metadata-only-altered tablet.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

## Compatibility

This is a forward-only on-disk change for shared-data range-distribution tables. After a metadata-only
trailing sort-key ADD (optionally followed by a reshard), a tablet's persisted per-tablet range can have
a sort-key arity larger than an old rowset's archived schema. The new BE reads this correctly via a
prefix-projection in `create_seek_range_from`; a pre-projection BE (e.g. an older 4.1 build) would fail
the exact-arity decode. Downgrade safety is addressed by backporting the read-side projection to
branch-4.1 so a downgraded 4.1 BE tolerates a main-written range. Rolling upgrade is safe by the
standard BE/CN-before-FE order: only a new FE emits the wider range, and it only meets already-upgraded
BEs. The feature is main-only / unreleased.

### DIFF
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -163,9 +163,9 @@ StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
     // range's bound arity; the current sort key is a prefix-preserving superset of the archived one,
     // so the leading column types agree.
     if (_tablet_metadata != nullptr) {
-        const int range_arity = range_pb->has_lower_bound()  ? range_pb->lower_bound().values_size()
-                                : range_pb->has_upper_bound() ? range_pb->upper_bound().values_size()
-                                                             : 0;
+        const int range_arity = range_pb->has_lower_bound()
+                                        ? range_pb->lower_bound().values_size()
+                                        : range_pb->has_upper_bound() ? range_pb->upper_bound().values_size() : 0;
         if (range_arity > 0 && static_cast<int>(range_schema->sort_key_idxes().size()) != range_arity) {
             auto current_schema = GlobalTabletSchemaMap::Instance()->emplace(_tablet_metadata->schema()).first;
             if (static_cast<int>(current_schema->sort_key_idxes().size()) == range_arity) {
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -139,9 +139,6 @@ Rowset::~Rowset() {
 
 StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
     const TabletRangePB* range_pb = nullptr;
-    // _tablet_schema is this rowset's own schema, which for an old rowset may be an archived
-    // historical schema with fewer sort-key columns than the current tablet schema.
-    TabletSchemaPtr range_schema = _tablet_schema;
     if (_metadata->has_range()) {
         range_pb = &_metadata->range();
     } else if (_tablet_metadata != nullptr && _tablet_metadata->has_range()) {
@@ -152,29 +149,23 @@ StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
         return std::optional<SeekRange>{};
     }
 
-    // A tablet range is expressed in whatever tablet sort key was current when the range was written,
-    // which can differ from this rowset's own (possibly archived) schema after a metadata-only trailing
-    // sort-key add (N -> N+1): the tablet-level fallback range is always in the current sort key, and a
-    // reshard that runs AFTER the add stamps a per-rowset range in the current sort key onto an old
-    // rowset that still carries its archived (N) schema. When the range's bound arity does not match
-    // this rowset's schema, decode with the current tablet schema instead -- its arity matches, and the
-    // current sort key is a prefix-preserving superset of the archived one, so the leading column types
-    // agree. A same-arity range decodes identically under either schema.
-    if (_tablet_metadata != nullptr) {
-        const int range_arity = range_pb->has_lower_bound()
-                                        ? range_pb->lower_bound().values_size()
-                                        : range_pb->has_upper_bound() ? range_pb->upper_bound().values_size() : 0;
-        if (range_arity > 0 && static_cast<int>(range_schema->sort_key_idxes().size()) != range_arity) {
-            auto current_schema = GlobalTabletSchemaMap::Instance()->emplace(_tablet_metadata->schema()).first;
-            if (static_cast<int>(current_schema->sort_key_idxes().size()) == range_arity) {
-                range_schema = current_schema;
-            }
-        }
-    }
+    // The SeekRange is used to seek into THIS rowset's segments, which are laid out per _tablet_schema
+    // (for an old rowset, an archived historical schema with fewer sort-key columns than the current
+    // tablet). The SeekRange's field ids are positional in the schema it is decoded with, so we MUST
+    // decode with _tablet_schema for the positions to align with the segments. A range can be written at
+    // a larger (later) sort-key arity than _tablet_schema -- a tablet-level fallback range is in the
+    // current sort key, and a reshard that runs after a metadata-only trailing key add can stamp a
+    // per-rowset range at a later arity onto an old rowset. create_seek_range_from projects such a wider
+    // bound onto _tablet_schema's sort key, comparing the added columns' defaults (what old rows in this
+    // rowset read) -- taken from the current tablet schema -- against the dropped trailing bound values.
+    TabletSchemaCSPtr current_schema =
+            _tablet_metadata != nullptr ? GlobalTabletSchemaMap::Instance()->emplace(_tablet_metadata->schema()).first
+                                        : nullptr;
 
     // Do not use mem_pool here. SeekRange will reference the data owned by protobuf metadata,
     // and metadata lifetime is guaranteed to outlive rowset iterators.
-    ASSIGN_OR_RETURN(auto range, TabletRangeHelper::create_seek_range_from(*range_pb, range_schema, nullptr));
+    ASSIGN_OR_RETURN(auto range,
+                     TabletRangeHelper::create_seek_range_from(*range_pb, _tablet_schema, nullptr, current_schema));
     return std::optional<SeekRange>(std::move(range));
 }
 
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -139,10 +139,18 @@ Rowset::~Rowset() {
 
 StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
     const TabletRangePB* range_pb = nullptr;
+    // _tablet_schema is this rowset's own schema, which for an old rowset may be an archived
+    // historical schema with fewer sort-key columns than the current tablet schema. The explicit
+    // per-rowset range was encoded with the rowset's schema, but the tablet-level fallback range is
+    // encoded with the current tablet sort key, so it must be decoded with the current schema. This
+    // keeps an old rowset readable after a metadata-only trailing sort-key add (N -> N+1), where the
+    // fallback range gains a value while the archived schema still has N sort keys.
+    TabletSchemaPtr range_schema = _tablet_schema;
     if (_metadata->has_range()) {
         range_pb = &_metadata->range();
     } else if (_tablet_metadata != nullptr && _tablet_metadata->has_range()) {
         range_pb = &_tablet_metadata->range();
+        range_schema = GlobalTabletSchemaMap::Instance()->emplace(_tablet_metadata->schema()).first;
     }
 
     if (range_pb == nullptr) {
@@ -151,7 +159,7 @@ StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
 
     // Do not use mem_pool here. SeekRange will reference the data owned by protobuf metadata,
     // and metadata lifetime is guaranteed to outlive rowset iterators.
-    ASSIGN_OR_RETURN(auto range, TabletRangeHelper::create_seek_range_from(*range_pb, _tablet_schema, nullptr));
+    ASSIGN_OR_RETURN(auto range, TabletRangeHelper::create_seek_range_from(*range_pb, range_schema, nullptr));
     return std::optional<SeekRange>(std::move(range));
 }
 
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -140,11 +140,7 @@ Rowset::~Rowset() {
 StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
     const TabletRangePB* range_pb = nullptr;
     // _tablet_schema is this rowset's own schema, which for an old rowset may be an archived
-    // historical schema with fewer sort-key columns than the current tablet schema. The explicit
-    // per-rowset range was encoded with the rowset's schema, but the tablet-level fallback range is
-    // encoded with the current tablet sort key, so it must be decoded with the current schema. This
-    // keeps an old rowset readable after a metadata-only trailing sort-key add (N -> N+1), where the
-    // fallback range gains a value while the archived schema still has N sort keys.
+    // historical schema with fewer sort-key columns than the current tablet schema.
     TabletSchemaPtr range_schema = _tablet_schema;
     if (_metadata->has_range()) {
         range_pb = &_metadata->range();
@@ -155,6 +151,27 @@ StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
 
     if (range_pb == nullptr) {
         return std::optional<SeekRange>{};
+    }
+
+    // A tablet range is expressed in whatever tablet sort key was current when the range was
+    // written, which can differ from this rowset's own (possibly archived) schema after a
+    // metadata-only trailing sort-key add (N -> N+1). Two cases both leave a bound arity that does
+    // not match _tablet_schema: the tablet-level fallback range is always in the current (N+1) sort
+    // key (handled above by switching to the current schema), and a reshard that runs AFTER the add
+    // stamps a per-rowset range in the current (N+1) sort key onto an old rowset that still carries
+    // its archived (N) schema. Decode with whichever available schema's sort-key arity matches the
+    // range's bound arity; the current sort key is a prefix-preserving superset of the archived one,
+    // so the leading column types agree.
+    if (_tablet_metadata != nullptr) {
+        const int range_arity = range_pb->has_lower_bound()  ? range_pb->lower_bound().values_size()
+                                : range_pb->has_upper_bound() ? range_pb->upper_bound().values_size()
+                                                             : 0;
+        if (range_arity > 0 && static_cast<int>(range_schema->sort_key_idxes().size()) != range_arity) {
+            auto current_schema = GlobalTabletSchemaMap::Instance()->emplace(_tablet_metadata->schema()).first;
+            if (static_cast<int>(current_schema->sort_key_idxes().size()) == range_arity) {
+                range_schema = current_schema;
+            }
+        }
     }
 
     // Do not use mem_pool here. SeekRange will reference the data owned by protobuf metadata,
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -146,22 +146,20 @@ StatusOr<std::optional<SeekRange>> Rowset::get_seek_range() const {
         range_pb = &_metadata->range();
     } else if (_tablet_metadata != nullptr && _tablet_metadata->has_range()) {
         range_pb = &_tablet_metadata->range();
-        range_schema = GlobalTabletSchemaMap::Instance()->emplace(_tablet_metadata->schema()).first;
     }
 
     if (range_pb == nullptr) {
         return std::optional<SeekRange>{};
     }
 
-    // A tablet range is expressed in whatever tablet sort key was current when the range was
-    // written, which can differ from this rowset's own (possibly archived) schema after a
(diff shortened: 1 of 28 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

xiangguangyxg:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -4834,6 +4931,146 @@ private static boolean resolveModifyColumnKeyness(OlapTable table, ColumnDef col
...
+
+        List<Column> oldSchema = table.getSchemaByIndexMetaId(baseIndexMetaId);
+        Set<Integer> oldUniqueIds = new HashSet<>();
+        for (Column column : oldSchema) {
+            oldUniqueIds.add(column.getUniqueId());
+        }
+        List<Column> newKeyColumns = new ArrayList<>();
+        for (Column column : newSchema) {
+            if (column.isKey() && !oldUniqueIds.contains(column.getUniqueId())) {
+                newKeyColumns.add(column);
```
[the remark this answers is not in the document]
xiangguangyxg: Fixed in `b2b4845c5b9`. `isMetadataOnlyTrailingKeyAdd` now requires the resolved base schema to add exactly one column total (`newSchema.size() == oldSchema.size() + 1`), not just one new key column. A single `ADD COLUMNS` that co-adds the trailing key with any other new column (AUTO_INCREMENT / generated / non-constant-default value column, etc.) now falls through to the K-tablet data-rewrite path, which materializes it. Added regression test `RangeDistributionGuardTest#testCoAddedValueColumnNotMetadataOnly`.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -4480,6 +4503,80 @@ private AlterJobV2 createFastSchemaEvolutionJobInSharedDataMode(SchemaChangeData
...
+        Column newTrailingColumn = findNewTrailingKeyColumn(olapTable, schemaChangeData);
+        if (newTrailingColumn == null) {
+            return null;
+        }
+        AlterJobV2 job = createFastSchemaEvolutionJobInSharedDataMode(schemaChangeData);
+        if (!(job instanceof LakeTableAsyncFastSchemaChangeJob)) {
+            return null;
+        }
+        LakeTableAsyncFastSchemaChangeJob asyncJob = (LakeTableAsyncFastSchemaChangeJob) job;
+        asyncJob.setTargetRanges(computeTargetRanges(olapTable, newTrailingColumn));
```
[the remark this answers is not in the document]
xiangguangyxg: Intentional, and consistent with the rest of this shared-data range-distribution schema-change series (P1a/P1b/P2): the feature is main-only / unreleased with no downgrade, and mixed-version is handled by the documented **BE/CN-before-FE** rolling-upgrade order. Only a new FE emits `tablet_range`, and the metadata-only routing code only exists on a new FE, so a new FE never sends it to an old BE — all BE/CN are upgraded first. A BE capability/version gate was deliberately omitted across the series for this reason. (As defense in depth, the arity mismatch this could otherwise cause on the read side is now handled in `Rowset::get_seek_range`, which decodes a range with whichever available schema's sort-key arity matches its bound.)

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/storage/lake/rowset.cpp
```diff
@@ -139,19 +139,44 @@ Rowset::~Rowset() {
...
+    // range's bound arity; the current sort key is a prefix-preserving superset of the archived one,
+    // so the leading column types agree.
+    if (_tablet_metadata != nullptr) {
+        const int range_arity = range_pb->has_lower_bound()
+                                        ? range_pb->lower_bound().values_size()
+                                        : range_pb->has_upper_bound() ? range_pb->upper_bound().values_size() : 0;
+        if (range_arity > 0 && static_cast<int>(range_schema->sort_key_idxes().size()) != range_arity) {
+            auto current_schema = GlobalTabletSchemaMap::Instance()->emplace(_tablet_metadata->schema()).first;
+            if (static_cast<int>(current_schema->sort_key_idxes().size()) == range_arity) {
+                range_schema = current_schema;
```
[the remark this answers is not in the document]
xiangguangyxg: Addressed by a later commit in this PR (`a6af28af254`). `Rowset::get_seek_range()` no longer decodes with the current tablet schema, and no longer "only checks N and N+2". It now decodes the range with the rowset's OWN (archived) schema — so the SeekRange's positional field ids align with that rowset's segments — and, when a bound is wider than that schema's sort key (a reshard stamped a later-arity range onto an old rowset), projects the bound onto the rowset's sort-key prefix by comparing the added columns' read-time defaults against the dropped trailing values (adjusting each bound's inclusivity). This handles any archived-vs-range arity gap, not just N/N+2, and the same primitive is used in the tablet-merger DCG rebuild. Covered by unit tests (`create_seek_range_from` projection cases + a `Rowset::get_seek_range` test with a real segment read + compaction using a non-NULL boundary that distinguishes correct projection from naive truncation) and validated on a shared-data cluster: reshard-after-add producing non-NULL trailing boundaries on archived rowsets, with a global per-(k1,k2) diff against a hash-distributed control table = 0.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -4480,6 +4503,82 @@ private AlterJobV2 createFastSchemaEvolutionJobInSharedDataMode(SchemaChangeData
...
+        List<Column> newTrailingColumns = findNewTrailingKeyColumns(olapTable, schemaChangeData);
+        if (newTrailingColumns.isEmpty()) {
+            return null;
+        }
+        AlterJobV2 job = createFastSchemaEvolutionJobInSharedDataMode(schemaChangeData);
+        if (!(job instanceof LakeTableAsyncFastSchemaChangeJob)) {
+            return null;
+        }
+        LakeTableAsyncFastSchemaChangeJob asyncJob = (LakeTableAsyncFastSchemaChangeJob) job;
+        asyncJob.setTargetRanges(computeTargetRanges(olapTable, newTrailingColumns));
```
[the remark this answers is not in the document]
xiangguangyxg: Same as the earlier thread on this concern: intentional and consistent across this shared-data range-distribution schema-change work. The feature is unreleased / main-only with no downgrade, and mixed-version is handled by the documented BE/CN-before-FE rolling-upgrade order: only a new FE emits `tablet_range`, and the metadata-only routing code exists only on a new FE, so a new FE never sends `tablet_range` to an old BE — all BE/CN are upgraded first. A BE capability/version gate was therefore deliberately omitted. As defense-in-depth, the BE read/merge path also decodes a range with the segment's own schema and projects wider bounds, so an arity mismatch degrades gracefully rather than corrupting.

xiangguangyxg:
@codex review

fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
```diff
@@ -417,6 +431,13 @@ public List<TTabletMetaInfo> getTTabletMetaInfoList() {
                     metaInfo.setTablet_schema(tabletSchema);
                 }
 
+                if (tabletRanges != null) {
+                    TabletRange range = tabletRanges.get(tabletId);
+                    if (range != null) {
+                        metaInfo.setTablet_range(range.toThrift());
```
kevincai, line 437: This new call path persists the output of `TabletRange.toThrift()` into the BE tablet-meta apply channel for the first time. `toThrift()` (in `TabletRange.java`, unchanged by this PR) sets `lower_bound_included` / `upper_bound_included` unconditionally, so for an `AllRange` or a one-sided range the thrift struct carries `__isset.*_bound_included = true` even though the corresponding bound is unset. That flows through `convert_t_range_to_pb_range` into the persisted `TabletMetadataPB.range`, leaving `has_lower_bound_included() == true` while `has_lower_bound() == false`. No current consumer is affected (all check `has_*_bound()` first), but the persisted metadata is inconsistent under the convention that an inclusivity flag is set only alongside a present bound. Consider setting the flags in `toThrift()` only when the matching bound is non-null.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -4834,6 +4933,162 @@ private static boolean resolveModifyColumnKeyness(OlapTable table, ColumnDef col
...
+     * (which routes a broader set of key changes to the K-tablet rewrite job).
+     *
+     * <p>Eligible iff ALL of:
+     * <ul>
+     *   <li>the table is a shared-data (cloud-native) range-distribution table;</li>
+     *   <li>the table has exactly one index meta (no rollup / synchronous MV);</li>
+     *   <li>the table is not a colocate table, has no AUTO_INCREMENT column, and has no temp
+     *       partitions;</li>
+     *   <li>the table's keysType is DUP_KEYS, AGG_KEYS, or UNIQUE_KEYS (not PRIMARY_KEYS);</li>
+     *   <li>the base index's resolved schema adds exactly one brand-new key column (a unique id not
```
kevincai: Stale doc: this Javadoc says the classifier accepts "exactly one brand-new key column", but the implementation accepts one-or-more (the guard is `newColumns.isEmpty()` and the loop iterates every added column; see line 5046 "adds one or more columns"). The same "exactly one" inaccuracy appears in two BE contracts added by this PR: `be/src/storage/lake/tablet_range_helper.h:106-108` ("exactly one new trailing column" / "exactly one trailing typed NULL_VALUE appended") and `be/src/storage/lake/txn_log_applier.cpp:95-96` ("the schema arity grows by one"). `validate_range_transition` actually accepts `num_added >= 1`. Please align all three to "one or more".

fe/fe-core/src/main/java/com/starrocks/planner/RangeDistributionPruner.java
```diff
@@ -45,30 +51,53 @@ public class RangeDistributionPruner implements DistributionPruner {
     public RangeDistributionPruner(List<Tablet> tabletsInOrder,
                                    List<Column> rangeDistributionColumns,
                                    Map<String, PartitionColumnFilter> distributionColumnFilters) {
-        this.tabletInOrder = new TreeMap<>();
+        this.rangeDistributionColumns = rangeDistributionColumns;
+        this.distributionColumnFilters = distributionColumnFilters;
+
+        // Capture each tablet's id and range exactly once (never call Tablet#getRange() twice) and
+        // validate arity only after every tablet has been captured, so a concurrent flip can't swap the
```
kevincai: This comment describes a two-pass design ("validate arity only after every tablet has been captured") that does not match the code: the loop below captures each tablet's id/range AND checks its arity (setting `degrade`) in the same single pass; only the TreeMap construction is deferred past the loop. The real safety invariant is that `ids.add(...)` precedes the per-iteration arity check and there is no `break` on mismatch. Suggest rewording to describe the single-pass capture-and-validate, then deferred TreeMap build.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
```diff
@@ -4834,6 +4933,162 @@ private static boolean resolveModifyColumnKeyness(OlapTable table, ColumnDef col
...
+     * applies when loading a {@code TabletSchemaPB} (see {@code tablet_schema.cpp}'s sort-key
+     * resolution): a non-empty {@code sortKeyUniqueIds} locates sort-key columns by unique id;
+     * otherwise a non-empty {@code sortKeyIdxes} locates them by schema position; otherwise the sort
+     * key is the leading run of key columns ({@link Column#isKey()}). Both {@code null} and empty
+     * lists fall through to the next precedence level. This differs from {@link
+     * MetaUtils#getRangeDistributionColumns(OlapTable, long)}, which treats a non-null (but possibly
+     * empty) {@code sortKeyIdxes} as an explicit sort key and never falls back to key columns.
+     */
+    @VisibleForTesting
+    static List<Column> resolveEffectiveSortKeyColumns(List<Column> schema, @Nullable List<Integer> sortKeyUniqueIds,
```
kevincai: Placement suggestion (not a correctness issue): `resolveEffectiveSortKeyColumns` is a pure static function over catalog types (`List<Column>`, `sortKeyUniqueIds`, `sortKeyIdxes`) with no `SchemaChangeHandler` state, is `@VisibleForTesting`, and its own Javadoc contrasts itself with `MetaUtils#getRangeDistributionColumns`. It reads as a sort-key-resolution utility rather than alter-routing logic. Consider moving it to `MetaUtils` (com.starrocks.sql.common) next to `getRangeDistributionColumns`, its natural sibling -- a pure placement move with no logic change.

be/src/storage/lake/txn_log_applier.cpp
```diff
@@ -71,12 +92,32 @@ Status apply_alter_meta_log(TabletMetadataPB* metadata, const TxnLogPB_OpAlterMe
...
+            auto new_schema = TabletSchema::create(alter_meta.tablet_schema());
+            RETURN_IF_ERROR(TabletRangeHelper::validate_range_structural(alter_meta.tablet_range(), *new_schema));
+            RETURN_IF_ERROR(
+                    TabletRangeHelper::validate_range_transition(*metadata, *new_schema, alter_meta.tablet_range()));
+
+            // Archive the pre-alter schema (non-clearing) before installing the new one.
+            archive_current_schema_into_history(metadata);
+
+            metadata->mutable_schema()->CopyFrom(alter_meta.tablet_schema());
+            metadata->mutable_range()->CopyFrom(alter_meta.tablet_range());
```
kevincai, line 114: Downgrade compatibility: this is the first path to persist a `TabletMetadataPB.range` whose sort-key arity (N+1) exceeds the arity of the archived per-rowset schemas (N). New BE tolerates the mismatch via the projection in `create_seek_range_from`, but the pre-PR BE's `create_seek_range_from` requires exact arity (`n != sort_key_idxes.size()` -> `Status::Corruption`). So once this feature is used, downgrading BE would fail scans on affected tablets with Corruption -- effectively a forward-only on-disk change. That seems inconsistent with the checked "No, this PR will not result in a change in behavior" box. Could you confirm the downgrade story (and note the forward-only limitation), or gate the range write behind a version/feature flag?

kevincai (commented):
### Code review

Posting the borderline findings from an automated pass as inline comments (comment/doc accuracy, one placement suggestion, and a downgrade-compatibility question). None block the happy-path correctness; the downgrade-compat item and the stale "exactly one" docs are the most worth addressing.

<sub>- If this review was useful, please react with 👍. Otherwise, react with 👎.</sub>

fe/fe-core/src/main/java/com/starrocks/task/TabletMetadataUpdateAgentTaskFactory.java
xiangguangyxg: Fixed in f72e801f3b1. `convert_t_range_to_pb_range` now persists a `*_bound_included` flag only alongside a present bound (added `&& has_*_bound()`), so a one-sided or Range.all range no longer leaves a stray inclusivity flag in the persisted metadata. I fixed it on the persist path rather than the shared `TabletRange.toThrift()` (which is also on the OlapTableSink routing hot path) since the persisted metadata is what the inconsistency note is about. Regression test: `test_convert_t_range_drops_included_flag_without_bound`.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
xiangguangyxg: Fixed in f72e801f3b1. Aligned all three to "one or more": the classifier Javadoc, the `validate_range_transition` header contract in `tablet_range_helper.h`, and the `txn_log_applier.cpp` comment.

fe/fe-core/src/main/java/com/starrocks/planner/RangeDistributionPruner.java
xiangguangyxg: Fixed in f72e801f3b1. Reworded to describe the actual single-pass capture-and-validate: the id is added before the per-iteration arity check, there is no `break` on mismatch (only `degrade` is flagged), and the TreeMap is built after the loop completes.

fe/fe-core/src/main/java/com/starrocks/alter/SchemaChangeHandler.java
xiangguangyxg: Done in f72e801f3b1. Moved `resolveEffectiveSortKeyColumns` to `MetaUtils` (com.starrocks.sql.common), next to its sibling `getRangeDistributionColumns` -- pure placement move, no logic change; updated the two call sites and the tests.

be/src/storage/lake/txn_log_applier.cpp
xiangguangyxg: Good catch on the forward-only aspect. The feature is unreleased / main-only, and the rolling-upgrade direction is safe (BE/CN upgrade before FE, so a new range is never read by an older node during an upgrade). For downgrade you are right: a pre-projection BE would fail the exact-arity decode on a mixed-arity range. Rather than gate the write, we will backport the read-side compatibility -- `create_seek_range_from` projecting a wider range onto the target segment sort key -- to branch-4.1 so a downgraded 4.1 BE tolerates a main-written range (verifying it against 4.1 rowset-schema machinery; will link the PR here). On the checkbox: it tracks user-facing interface changes (SQL / config / external protocol); an internal forward-only metadata format is not one, but I will call out the forward-only limitation in the PR description.

xiangguangyxg: Downgrade-compat backport opened: StarRocks/starrocks#76537 — it ports only the read-side projection to branch-4.1 so a 4.1 BE tolerates a range written by a newer build (`create_seek_range_from` projects a wider bound onto the segment sort key instead of failing the exact-arity decode). With that on 4.1, downgrading a cluster that used this feature to 4.1 is safe. Rolling upgrade remains safe by the BE/CN-before-FE order, and the PR body now documents the forward-only limitation.

CelerData-Reviewer:
@codex review

kevincai (approved):

srlch (approved):

PR #76340, opened 2026-07-14
[Enhancement] Make the segment iterator reusable across scans

stephen-shelby:
## Why I'm doing:

Today StarRocks builds a brand-new `SegmentIterator` for every scan of a segment, repeating the one-time setup (schema resolution, predicate/index preparation) each time. When the same segment is scanned many times in quick succession — most notably by the upcoming lake prepared-physical-split scan, where one segment is fanned out into many child scans — that rebuild is pure overhead. There is currently no way to reset an existing `SegmentIterator` and re-drive it against a new scan range while keeping its one-time setup intact.

## What I'm doing:

Add the segment-layer machinery to reuse a `SegmentIterator` across scans, plus the precomputed-scan-range plumbing a reused iterator consumes:

- **Reset/reuse**: `SegmentIterator::reset_for_reuse()` + `_init_reused_scan()`, with the one-time setup split out behind `_one_time_setup_done` (and `ScanContext::reset()` / `_init_scan_range_and_context()`). Per-scan state is reset while the expensive one-time setup is preserved.
- **Create/reuse via a caller-owned slot**: `Segment::new_reusable_iterator()` / `new_reusable_segment_iterator()`. An empty slot builds the underlying iterator; a populated slot resets and reuses it.
- **`NonClosingChunkIterator`**: a wrapper whose `close()` is a no-op, so a finishing driver does not tear down the shared reused iterator held in the slot.
- **`SegmentReadOptions::read_state_cache` (`SegmentReadStateCache`)** + `_apply_precomputed_scan_range()` + `_get_prepared_pruned_row_ranges()`: a reused iterator can be handed a precomputed (already page-filtered) scan range and apply it directly instead of re-running the pruning pipeline.
- **`BitmapIndexEvaluator::init()`** idempotency reset so it can be re-initialized on reuse.

This is infrastructure only. Nothing populates the read-state cache or calls `reset_for_reuse()` yet — the lake prepared physical split scan follow-ups do — so there is **no behavior change** on the existing scan path.

## Tests:

`SegmentIteratorTest.ReuseSegmentIteratorAcrossScans` drives a single underlying `SegmentIterator`, kept in a caller-owned slot, across three scans:
1. first scan builds the iterator (empty slot) and reads all rows;
2. second scan reuses the **same instance** (asserted by pointer identity — `reset_for_reuse`, not a rebuild) and still returns the full result set;
[...]

PR #76353, opened 2026-07-14
[Enhancement] Lake prepared physical split scan: seed pruning and prepared read state

stephen-shelby:
Part of #76355

## Why I'm doing:

The lake prepared physical split scan wants to prune each segment **once**, at the seed scan, and share the result (the page-filtered row ranges plus the resolved rowid bounds) with every split child, so the children can skip re-running the page-filter pipeline. This PR adds the lake storage core for that: the cross-child shared read state and the seed build / refine paths that populate it. It builds on the reusable `SegmentIterator` from #76340.

## What I'm doing:

- **`PreparedTabletReadState` / `PreparedSegmentReadState`**: cross-child shared scan state — an atomic-guarded pruned scan range and rowid-bounds cache, plus a coarse split cursor protected by `coarse_range_lock`.
- **Seed path** (`TabletReader`): `build_prepared_tablet_read_state` → `build_initial_coarse_split_tasks` → `prepare_segment_pruned_scan_range_for_split`, which folds every page filter (short-key / zonemap / bloom / bitmap / inverted / vector) into the prepared scan range at the seed, and `refine_initial_coarse_split_and_append_refined_tasks` which emits the refined tasks (using `subtract_sparse_ranges` to compute the remaining coverage).
- **CHILD path**: a split child reuses the prepared segment iterator / read state (`make_segment_read_state_cache` → `SegmentReadStateCache`) so it skips re-pruning.
- **`Rowset::read` refactor**: a shared `do_read` kernel plus a `read_prepared_segment` entry, with `init_segment_read_options` / `set_segment_tablet_range` / `build_segment_schema` extracted for reuse.
- **`LakeSplitContext`** (`split_scan_morsel.h`): carries the prepared state and the rowid-range source between the seed and the split-morsel scheduler.
- Prepared-split profile counters on `OlapReaderStatistics`.

Gated by `TabletReaderParams::enable_prepared_physical_split_scan` (default **off**, set by a follow-up connector) — a no-op when off. The split-morsel queue that consumes these tasks is a follow-up.

## Tests:

- **`LakeTabletReaderSpit.test_prepared_physical_split_end_to_end_equivalence`** — the core correctness contract: the INITIAL_COARSE seed children (which read their coarse range and refine) plus the REFINED children they append reproduce the full-scan result **exactly** — no missing rows, no duplicates. Drives `refine_initial_coarse_split_and_append_refined_tasks`, `subtract_sparse_ranges`, the CHILD `read_prepared_segment` consumption, and the shared prepared state end to end.
- **`LakeTabletReaderSpit.test_prepared_physical_split_gate_off`** — gate off (default) falls back to the regular `PhysicalSplitMorselQueue`; every task is a REGULAR split with no prepared state.
- **`LakeTabletReaderSpit.test_prepared_physical_split_seed`** — the seed produces INITIAL_COARSE tasks carrying the shared, non-null prepared tablet read state.
- **`SubtractSparseRangesTest.CoversAllBranches`** — `subtract_sparse_ranges` (the REFINED coverage math) across empty / exact-cover / left-cut / right-cut / hole / disjoint / rhs-spanning-two-ranges / multiple-cuts.
- **`PreparedReadStateCacheTest.CacheStateMachine`** — the publish / clear / disable truth table for both caches, including the sticky-disabled behavior.

All existing lake `tablet_reader` tests (plain dup read, aggregate, delete-predicate reads, split read) pass through the refactored `do_read`, so the existing read path is unchanged.

## What type of PR is this:

- [ ] BugFix
- [x] Enhancement
- [ ] Feature
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

(The new prepared-split path is gated off by default and has no caller yet — the connector is a follow-up. The only always-on change is the `Rowset::read` → `do_read` refactor, which is behavior-preserving and covered by the existing lake `tablet_reader` tests.)

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/rowset.cpp
+++ b/be/src/storage/lake/rowset.cpp
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <future>
+#include <set>
 #include <unordered_set>
 
 #include "base/debug/trace.h"
@@ -27,6 +28,8 @@
 #include "fs/fs_factory.h"
 #include "runtime/current_thread.h"
 #include "runtime/runtime_env.h"
+#include "runtime/runtime_state.h"
+#include "storage/chunk_helper.h"
 #include "storage/delete_predicates.h"
 #include "storage/lake/column_mode_partial_update_handler.h"
 #include "storage/lake/filenames.h"
@@ -52,6 +55,18 @@
 #include "types/type_descriptor.h"
 
 namespace starrocks::lake {
+
+static SegmentReadStateCache make_segment_read_state_cache(const PreparedSegmentReadState& state) {
+    SegmentReadStateCache cache;
+    if (state.has_pruned_scan_range()) {
+        cache.scan_range = state.pruned_scan_range;
+    }
+    if (state.has_rowid_bounds_cache()) {
+        cache.seek_range_rowid_ranges = &state.seek_ranges_rowid_bounds;
+        cache.tablet_rowid_range = &state.tablet_range_rowid_bounds;
+    }
+    return cache;
+}
 
 Rowset::Rowset(TabletManager* tablet_mgr, int64_t tablet_id, const RowsetMetadataPB* metadata, int index,
                TabletSchemaPtr tablet_schema)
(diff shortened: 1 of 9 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby:
Re: the module-risk briefing (segment/rssid positional alignment) — validated against the diff, the invariant holds:

- \`Rowset::segments()\` leaves a \`nullptr\` placeholder for every lost segment instead of dropping it, so the prepared-segment vector stays positionally aligned with `size == num_segments` (index `i` == metadata position `i`).
- The prepared branch in \`do_read\` guards this explicitly: it returns `InvalidArgument` if `prepared_segments->size() != num_segments()`, skips the `nullptr` placeholders, applies the metadata filter by the same segment position, and derives `shared()` / `.vi` uid state from that metadata position — so a misaligned/compacted vector fails fast rather than silently corrupting `meta_pos`-derived state.
- `PreparedTabletReadState::rowset_segments` is built once from `get_rowsets()` (full, non-segment-range query rowsets → `seg_start == 0`, so `i == meta_pos`) and shared read-only across all INITIAL_COARSE/REFINED children.

Added a unit test that drives the seed + refined children end to end and asserts they reproduce the full scan exactly (no missing/duplicate rows), which would trip on any such misalignment.

(Note: the briefing head-sha is one push behind; the prepared-segment logic is byte-identical on the current head.)

satanson (approved):

meegoo (approved):

PR #76365, opened 2026-07-14
[BugFix] Fix shared-data range-distribution tablet-reshard publish stalls and PK-index corruption

xiangguangyxg:
## Why I'm doing:

Shared-data range-distribution tables that support dynamic tablet reshard (split/merge) can permanently stall publish and corrupt the primary-key index under continuous batched ingestion with concurrent splits. Two independent bugs in the reshard publish path are responsible; both were reproduced on a daily-release shared-data cluster.

## What I'm doing:

Fixes #issue

**1. Identical-tablet reshard skips the PK-index flush (`be/src/storage/lake/tablet_reshard.cpp`)**

`handle_identical_tablet` copies the parent tablet metadata to the new tablet without flushing the parent's in-memory PK-index memtable, unlike `split_tablet` and the merge path. The identical child then inherits an `sstable_meta` that does not cover recently-written rowsets whose overwrite/delete resolution lived only in the memtable. Its cold PK-index rebuild re-derives an index inconsistent with the inherited delete vectors, surfacing as `insert found duplicate key ...` on rebuild, or `delvec inconsistent` when a cross-published `op_write` is re-applied. Fix: flush the PK-index memtable before the metadata copy, mirroring the pre-split flush in `split_tablet` (no-op for non-PK / non-cloud-native-persistent-index tablets).

**2. File-bundling carry-forward excludes touched indexes by physical id (`fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java`)**

The file-bundling carry-forward path excludes already-touched indexes by physical index id. A tablet reshard replaces an index with a new physical id but the same metaId, so a batched pre-reshard publish carries the reshard'd index's child tablets forward as no-op normal tablets while they are also cross-published from their parent; the two tasks self-collide on the per-tablet publish lock and wedge the partition. Fix: key the touched-index set by metaId at both collection sites and in `collectFileBundlingCarryForwardTablets`.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/tablet_reshard.cpp
+++ b/be/src/storage/lake/tablet_reshard.cpp
@@ -28,6 +28,7 @@
 #include "storage/lake/tablet_reshard_helper.h"
 #include "storage/lake/tablet_splitter.h"
 #include "storage/lake/transactions.h"
+#include "storage/lake/update_manager.h"
 #include "storage/lake/vacuum.h" // delete_files_async
 #include "storage/storage_env.h"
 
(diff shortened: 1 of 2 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
Split into two focused PRs: #76367 (BE, identical-tablet flush) and #76368 (FE, carry-forward metaId). Both validated on a shared-data cluster.

PR #76366, opened 2026-07-14
[BugFix] Restore ConnectContext on query-deploy workers

### LINKED ISSUE #76363: ConnectContext of finished queries leaks into query-deploy worker ThreadLocals (AllAtOnceExecutionSchedule.TracerContext)

1. Make sure `enable_connector_deploy_scan_ranges_background` is `true` (the default) so that incremental scan-range deployment runs on the shared `query-deploy` thread pool.
2. Run any external-catalog query (Hive / Iceberg / Paimon / Hudi, ...) that goes through `AllAtOnceExecutionSchedule.DeployScanRangesTask`, e.g. a scan over a table with enough splits to require incremental scan-range deployment.
3. Wait for the query to finish and the client connection to close.
4. Force a full GC (`jcmd <fe_pid> GC.run`) and inspect the heap (`jmap -histo:live`, or a heap dump).

**Observed:** the finished query's `ConnectContext` — together with everything it references (`StmtExecutor`, `ExecPlan`, all scan nodes and their scan ranges, session state) — is still strongly reachable. The GC root is a `query-deploy-N` worker thread's `ThreadLocal`. A JFR recording dumped with `path-to-gc-roots=true` shows `jdk.OldObjectSample` reference chains ending at:

```
root = { description = "Thread Name: query-deploy-N", system = "Threads", type = "Handle Area" }
```

### Root cause

`AllAtOnceExecutionSchedule.TracerContext` swaps the query's `ConnectContext` into the worker thread's ThreadLocal, but `close()` only restores it when the *saved* context is non-null:

```java
TracerContext(Tracers forkedTracers, ConnectContext connectContext) {
    Tracers.set(forkedTracers);
    if (connectContext != null) {
        savedConnectContext = ConnectContext.get();   // null on a pooled worker
        ConnectContext.set(connectContext);
    } else {
        savedConnectContext = null;
    }
}

@Override
public void close() {
    Tracers.close();
    if (savedConnectContext != null) {                // null -> restore skipped
        ConnectContext.set(savedConnectContext);
    }
}
```

On a fresh `query-deploy` worker, `ConnectContext.get()` is `null`, so `savedConnectContext == null` and `close()` skips the restore entirely. The first query that runs a deploy task on each worker leaves its `ConnectContext` in that worker's ThreadLocal **permanently**.

Worse, the leak is self-sustaining: the *next* query on the same worker saves the leaked context (now non-null), swaps in its own, and on close dutifully restores **the leaked first context back**. So each worker permanently pins the ConnectContext of the first query it ever deployed, and it can never be displaced by later queries.

Since `run()` re-submits the task (`start()`) for each incremental deployment round, a single query can seed several cold workers.

After a query finishes, its `ConnectContext` must become unreachable so the plan object graph can be collected. `close()` should restore the worker's ThreadLocal to its prior state — including "empty": when the saved context is null, call `ConnectContext.remove()` instead of skipping the restore.

### Impact

- The `query-deploy` pool has `Config.query_deploy_threadpool_size = max(50, availableProcessors() * 10)` threads, so on a large FE hundreds of workers can each pin one query's full plan graph. For wide external-table scans (many splits / large scan ranges) this reaches multi-GB, GC-immune FE heap retention.
- The retention is invisible to normal metrics (it is live memory, not a cache), and full GC cannot reclaim it. It looks like slow FE memory creep after connector-heavy workloads.

### Real behavior (Required)

`ConnectContext` of completed queries remains strongly reachable from `query-deploy-N` worker ThreadLocals indefinitely (verified via JFR `OldObjectSample` with `path-to-gc-roots=true` and `jmap -histo:live` after forced GC).

- main (verified in current `main` source: `fe/fe-core/src/main/java/com/starrocks/qe/scheduler/dag/AllAtOnceExecutionSchedule.java`, `TracerContext.close()`)

spaces-X:
## Why I'm doing:

`AllAtOnceExecutionSchedule.TracerContext.close()` only restores the saved `ConnectContext` when it is non-null. On a pooled `query-deploy` worker the previous context is always null, so the restore is skipped and the query's `ConnectContext` — together with the whole `ExecPlan` object graph it references (scan nodes, scan ranges, session state) — stays pinned in the worker's ThreadLocal indefinitely.

The leak is self-sustaining: the next query on the same worker saves the leaked context, swaps in its own, and on close restores the leaked first context back. Each worker therefore permanently pins the `ConnectContext` of the first query it ever deployed. With `query_deploy_threadpool_size = max(50, cores * 10)` workers and `enable_connector_deploy_scan_ranges_background` on by default, connector-heavy workloads accumulate GC-immune FE heap retention (multi-GB on large FEs).

Verified with JFR: `jdk.OldObjectSample` reference chains (dump with `path-to-gc-roots=true`) end at `Thread Name: query-deploy-N / Handle Area` for queries that finished long ago, and `jmap -histo:live` after a forced GC still shows the finished query's plan objects.

Fixes #76363

## What I'm doing:

Track whether a context was actually swapped in (`connectContextSwapped`) and restore unconditionally on close: put the saved context back when there was one, otherwise `ConnectContext.remove()` so the worker returns to its pristine state.

The `savedConnectContext != null` branch is kept for correctness in case a worker legitimately had a context before the swap; behavior on the main-thread path (no swap performed) is unchanged.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/qe/scheduler/dag/AllAtOnceExecutionSchedule.java
+++ b/fe/fe-core/src/main/java/com/starrocks/qe/scheduler/dag/AllAtOnceExecutionSchedule.java
@@ -47,22 +47,33 @@ public class AllAtOnceExecutionSchedule implements ExecutionSchedule {
      */
     class TracerContext implements AutoCloseable {
         private final ConnectContext savedConnectContext;
+        private final boolean connectContextSwapped;
 
         TracerContext(Tracers forkedTracers, ConnectContext connectContext) {
             Tracers.set(forkedTracers);
             if (connectContext != null) {
                 savedConnectContext = ConnectContext.get();
                 ConnectContext.set(connectContext);
+                connectContextSwapped = true;
             } else {
                 savedConnectContext = null;
+                connectContextSwapped = false;
             }
         }
 
         @Override
         public void close() {
             Tracers.close();
-            if (savedConnectContext != null) {
-                ConnectContext.set(savedConnectContext);
+            // Restore whenever a context was swapped in: on a pooled query-deploy worker the
+            // previous context is null, and skipping the restore would leave the query's
+            // ConnectContext (and every plan object it references) pinned in the worker's
+            // ThreadLocal indefinitely.
+            if (connectContextSwapped) {
+                if (savedConnectContext != null) {
+                    ConnectContext.set(savedConnectContext);
+                } else {
+                    ConnectContext.remove();
+                }
             }
         }
     }
--- a/fe/fe-core/src/test/java/com/starrocks/qe/scheduler/dag/AllAtOnceExecutionScheduleTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/qe/scheduler/dag/AllAtOnceExecutionScheduleTest.java
@@ -1,0 +1,79 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.qe.scheduler.dag;
+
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

dirtysalt (approved):
LGTM! 🚀

dirtysalt:
@spaces-X thanks for catching this bug.

ZiheLiu (approved):

PR #76368, opened 2026-07-14
[BugFix] Exclude touched indexes by metaId in file-bundling carry-forward

xiangguangyxg:
## Why I'm doing:

In shared-data mode, a file-bundling range-distribution table can permanently wedge publish (a livelock under continuous batched ingestion with concurrent tablet reshards).

The file-bundling carry-forward path excludes already-touched indexes by **physical index id**. A tablet reshard (split/merge) replaces an index with a new physical id but the **same metaId**, so a batched pre-reshard publish treats the reshard'd base index as untouched and carries its child tablets forward as no-op normal tablets while the same children are also cross-published from their parent. The two tasks then self-collide on the per-tablet publish lock, so the aggregate publish never fully succeeds and the partition's publish wedges.

## What I'm doing:

Fixes #issue

Key the touched-index set by **metaId** instead of physical index id — at both collection sites and in `collectFileBundlingCarryForwardTablets` — so a reshard'd index is still recognized as touched and not carried forward.

Validated on a shared-data cluster: with continuous batched load forcing ~930 tablet splits on a `file_bundling=true` range-PK table, VisibleVersion advanced past 3600 with zero publish stalls (the same workload wedges without this fix).

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
+++ b/fe/fe-core/src/main/java/com/starrocks/transaction/PublishVersionDaemon.java
@@ -644,7 +644,7 @@ public class PublishVersionDaemon extends LeaderDaemon {
             }
 
             useAggregatePublish = table.isFileBundling();
-            Set<Long> publishedNormalIndexIds = Sets.newHashSet();
+            Set<Long> publishedNormalIndexMetaIds = Sets.newHashSet();
             for (int i = 0; i < transactionStates.size(); i++) {
                 TransactionState txnState = transactionStates.get(i);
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

luohaha (approved):

sevev (approved):

PR #76392, opened 2026-07-15
[Enhancement] Pre-split base and rollup for INSERT-from-table into a range table

xiangguangyxg:
## Why I'm doing:

Sample-Based Tablet Pre-Split previously skipped `INSERT ... SELECT ... FROM <single OLAP table>` whenever the range-distribution target carried a rollup, so on such a load neither the base index nor the rollup got pre-split (they stayed at one tablet and relied on post-load auto-merge/compaction). PR #76233 shipped multi-index pre-split for FILES/Broker loads but documented INSERT-from-table + rollup as a descope, because the sampler could only project the base sort key and only by target column name, whereas INSERT-from-table must project by the *source* column name via the target->source column remap.

## What I'm doing:

Make INSERT-from-table split every visible index (base plus each rollup):

- `InsertFromTableScanContext` now carries the full target->source column-name map instead of a base-only source-column list. The data-tier executor maps whatever sort key it is handed through that map: `request.getSortKey()` for the base (in the multi-partition single-request flow, and per-index in the single-partition flow), and each rollup spec's sort key for the secondary projection.
- The shared secondary-index projection becomes an overridable executor seam (`AbstractSqlSampleSubqueryExecutor.secondaryIndexProjectionIdents`). The default projects by target column name, so FILES/Broker generate byte-identical SQL; INSERT-from-table overrides it to remap to source names.
- `TablePreSplitSource` now builds one secondary index spec per visible rollup.
- The two guards that skipped multi-index INSERT-from-table are removed.

Safety: single-index (no-rollup) INSERT-from-table is byte-identical (the base sort key maps to the same source names). FILES/Broker and the ALTER-side pre-split path are untouched. Any unmapped sort-key column fails the sample so the load simply proceeds without pre-split (a base-only or no split is data-safe, since the BE routes each row by its true value). Data tier only -- an OLAP source has no Parquet/ORC footer, so the meta tier is already bypassed for this load kind.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/SecondaryIndexSpec.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/SecondaryIndexSpec.java
@@ -15,7 +15,11 @@
 package com.starrocks.alter.reshard.presplit;
 
 import com.starrocks.catalog.Column;
+import com.starrocks.catalog.MaterializedIndexMeta;
+import com.starrocks.catalog.OlapTable;
+import com.starrocks.sql.common.MetaUtils;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -26,4 +30,21 @@ import java.util.List;
  * routed back to the correct index downstream.
  */
 public record SecondaryIndexSpec(long indexMetaId, List<Column> sortKey) {
+
+    /**
+     * Builds one spec per visible rollup index of {@code target} (every visible index except the
+     * base), each carrying its own sort key. Shared by the FILES / Broker / INSERT-from-table
+     * pre-split sources so the "every visible index except base" enumeration lives in one place.
+     */
+    static List<SecondaryIndexSpec> forVisibleRollups(OlapTable target) {
+        List<SecondaryIndexSpec> specs = new ArrayList<>();
+        for (MaterializedIndexMeta meta : target.getVisibleIndexMetas()) {
+            if (meta.getIndexMetaId() == target.getBaseIndexMetaId()) {
+                continue;
+            }
+            specs.add(new SecondaryIndexSpec(meta.getIndexMetaId(),
+                    MetaUtils.getRangeDistributionColumns(target, meta.getIndexMetaId())));
+        }
+        return specs;
+    }
 }
--- a/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/AbstractSqlSampleSubqueryExecutor.java
+++ b/fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/AbstractSqlSampleSubqueryExecutor.java
@@ -45,7 +45,9 @@ import java.util.stream.Collectors;
  *
  * <p>Subclasses implement {@link #resolveSampleSpec} to supply the FROM clause
  * SQL, an optional WHERE predicate, the total input byte count, the compute
- * resource, and the projected column identifier lists. Everything else —
+ * resource, and the base sort-key / partition projection identifier lists; a
+ * subclass may also override {@link #secondaryProjectionIdents} to control how
+ * the secondary indexes' sort keys are projected. Everything else —
  * sampling-rate math, SQL synthesis, BE invocation, JSON row decode — is shared
  * here.
  */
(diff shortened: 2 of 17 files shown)


CelerData-Reviewer:
@codex review

xiangguangyxg:
### Shared-data e2e validation (TSP build 2611, 1 FE + 3 CN, ubuntu)

Range table `DUPLICATE KEY(k1,k2) ORDER BY(k1,k2)` + divergent rollup `ADD ROLLUP r(k1,k2,v) ORDER BY(k2)`, 3,000,000-row OLAP source. All CNs: `tablet_reshard_split_external_boundaries_fallback_total` Δ=0 throughout.

| Case | Source SELECT | Result |
|---|---|---|
| Primary | `SELECT * FROM src` (names match, by-position) | base **1→2** (composite `(k1,k2)` ranges) **and** rollup **1→2** (single `(k2)` ranges); `COUNT`=3,000,000, `SUM(k2)`=4,499,998,500,000 exact; `tier_used{data_tier}`+1; fallback Δ=0 |
| Non-identity remap | `SELECT a, b, c FROM src2` (bare cols, `k1←a`, `k2←b`) | base **1→2** and rollup **1→2** with **source-derived** boundaries; `COUNT`/`SUM(k2)` exact; fallback Δ=0 |
| Renamed source under `*` | `SELECT * FROM src2` (names differ) | **skips** pre-split (by-position `SELECT *` name-match gate); data loaded correctly |
| Aggregation | `SELECT k1,k2,sum(v) ... GROUP BY k1,k2` | **skips** gracefully (no split, no fallback); data loaded |
| Expression | `SELECT k1, k2+1, v` | **skips** gracefully (no split); data loaded |

Both the base index and the divergent rollup pre-split from the same sample, with per-index boundaries computed from the correctly-remapped source columns; every non-eligible source skips without a fallback. FE unit suite (presplit + broker) 636/0/0, checkstyle clean.

CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

fe/fe-core/src/main/java/com/starrocks/alter/reshard/presplit/SecondaryIndexSpec.java
```diff
@@ -26,4 +30,21 @@
...
 public record SecondaryIndexSpec(long indexMetaId, List<Column> sortKey) {
+
+    /**
+     * Builds one spec per visible rollup index of {@code target} (every visible index except the
+     * base), each carrying its own sort key. Shared by the FILES / Broker / INSERT-from-table
+     * pre-split sources so the "every visible index except base" enumeration lives in one place.
+     */
+    static List<SecondaryIndexSpec> forVisibleRollups(OlapTable target) {
+        List<SecondaryIndexSpec> specs = new ArrayList<>();
+        for (MaterializedIndexMeta meta : target.getVisibleIndexMetas()) {
```
kevincai, line 41: This is not protected by the table read lock, and the hashMap is not thread safe, so iterating it may result ConcurrentModificationException, consider patch it for thread safe or properly handle it if the table meta is modified in the middle.

xiangguangyxg: Thanks -- I traced the concurrency path and I believe it is already bounded-safe, so I've left the code as-is; happy to add a lock if you prefer (details below).

- `OlapTable.getVisibleIndexMetas()` returns a freshly-built `ArrayList` (a snapshot), so the `for` loop here iterates a private copy -- it cannot throw `ConcurrentModificationException` from a concurrent meta change.
- This hook runs pre-plan, deliberately outside the planner's `PlannerMetaLocker` read lock (to avoid deadlocking with the reshard daemon's write lock -- see the `InsertPreSplitHook` class doc). The whole hook is wrapped in `catch (Throwable) -> log + proceed without pre-split`, so if a concurrent ALTER does race the lock-free read (e.g. a dropped rollup makes `MetaUtils.getRangeDistributionColumns` throw), it is caught and the load simply proceeds unsplit.
- A stale-but-non-throwing read is also covered downstream: before any tablet is split, `DefaultPreSplitPipeline.preSubmit` re-resolves the visible index-meta-id set under a table READ lock (`currentVisibleIndexMetaIds`) and bails if it no longer matches; the multi-partition path re-checks the same set under its post-pre-create READ lock. So a race can only ever cause a missed pre-split, never a wrong split or a crash.
- Note this lock-free `getVisibleIndexMetas()` read is the pre-existing pre-split-hook pattern (also in `PreSplitTargets.findEligibleTable` and `InsertPreSplitHook`, and the FILES/Broker path merged in #76233); this PR only consolidated it into `forVisibleRollups`.

If you'd still like a hard guarantee, I can wrap the hook's meta reads in a brief `lockTableWithIntensiveDbLock(READ)` that snapshots and releases before the async sample (safe vs the daemon's WRITE, not held across submit). Let me know and I'll add it.

kevincai (approved):

srlch (approved):

PR #76411, opened 2026-07-15
[BugFix] Honor the targeted snapshot's schema and partition spec for Iceberg reads (backport #74711)

GavinMar:
## Why I'm doing:

Iceberg time travel queries (`VERSION AS OF` / `TIMESTAMP AS OF`, tag/branch) resolved the target snapshot only to select data files, while column resolution, the result-set metadata, the descriptor sent to BE, scan planning, and the partition view all used the **latest** table schema/spec. After a schema or partition-spec change the time-travel read diverged from the targeted snapshot:

```sql
CREATE TABLE t (id INT, c INT);            -- snapshot S1 committed with schema (id, c)
ALTER TABLE t RENAME COLUMN c TO c_renamed;
INSERT INTO t VALUES (...);                -- snapshot S2 with schema (id, c_renamed)

SELECT * FROM t VERSION AS OF <S1>;        -- returned column `c_renamed` instead of `c`
SELECT * FROM t VERSION AS OF <S1> WHERE c = 1;  -- failed: `c` could not be resolved
```

The same divergence affected partitioning: a partition field added by later spec evolution was advertised for a snapshot whose files were written under the previous spec. This is contrary to Iceberg semantics — the snapshot's `schema-id` is defined by the spec as "ID of the table's current schema when the snapshot was created", and a snapshot is planned with the specs attached to its manifests (matching `SnapshotUtil.schemaFor` / `DataScan#useSnapshotSchema` and other engines).

## What I'm doing:

Resolve the query period once during analysis and bind the targeted snapshot's metadata — schema and partition specs — through the whole read path:

- `QueryAnalyzer`: after table resolution, resolve the query period to a version range (via the new `QueryPeriodResolver`, extracted from `RelationTransformer` so both layers share one implementation), pin the resolved range on the `TableRelation` so the transformer does not resolve it again, and — whenever time travel resolves a snapshot — replace the table with an immutable per-query copy created by `IcebergTable#withReadMetadata`. It is bound whenever a snapshot is resolved, not only when the schema id changes, because partition evolution can change the spec while leaving the schema id unchanged.
- `IcebergTable`: `withReadMetadata` pins a `SnapshotReadView` (the snapshot schema plus the partition specs its manifests reference, via `IcebergMetadata#getSnapshotSpecs`) and rebuilds the StarRocks-side full schema from it. `getReadSchema()` / `getReadSpec()` feed column/partition resolution and `toThrift` (columns, `iceberg_schema`, partition info) so FE and BE see one consistent view: a single snapshot spec is used directly, multiple specs collapse to unpartitioned so table-level partition-metadata optimizations are disabled (scan planning still uses each file's own spec), and partition fields added after the snapshot are not advertised.
- `IcebergMetadata`: convert pushdown predicates against the read schema, matching the schema the scan binds expressions against. `IcebergScanNode#setupScanRangeLocations` filters required column names against the read schema so snapshot-only names are not dropped.
- `StarRocksIcebergTableScan`: the scan schema is the read schema; rebind specs to it (`scanSpecsById`) for the residual/partition/metrics evaluators, `DeleteFileIndex`, `ManifestGroup`, and manifest filtering, so filter expressions bind consistently, and carry the read schema into remote metadata planning (`MetadataParser`) too. A spec that cannot be rebound degrades to no pruning with the full filter kept as residual. Time-travel reads force local planning as a temporary workaround, since the remote metadata scanner binds predicates against the current specs (removed once it honors the snapshot schema/specs).

Fixing this the same way — resolving a read against the schema the query targets rather than the latest schema — also fixes StarRocksTest#11432: a metadata-only `ALTER TABLE ... ADD COLUMN` advances the current schema without producing a snapshot, and an ordinary current read now resolves the newly added column against the current schema instead of failing with `Cannot find field` until the next write.

Verification:
- New UTs cover schema rebinding, the analyzer path (including the pre-resolved external-table path), the BE descriptor, scan planning with predicates on renamed columns, partition-evolution snapshots (fields added on new and preexisting columns), forced local planning for time travel, and the metadata-only `ADD COLUMN` current read.
- New SQL tests (`test_timetravel_snapshot_schema`) cover tag/branch reads, old/new column-name resolution, predicates on renamed columns, a partitioned table whose partition source column was renamed, partition evolution, and reading a column added after the current snapshot.
- Verified end to end on a real Hive-catalog Iceberg cluster.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/IcebergTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/IcebergTable.java
@@ -128,6 +128,10 @@ public class IcebergTable extends Table {
     private Map<String, String> icebergProperties = Maps.newHashMap();
 
     private org.apache.iceberg.Table nativeTable; // actual iceberg table
+    // Per-query read view of a targeted snapshot for time travel; null means an ordinary read of the
+    // current table state. Bundles the snapshot schema and the partition specs its manifests reference
+    // so the two are always pinned and dropped together.
+    private transient SnapshotReadView readView;
     private List<Column> partitionColumns;
     private Optional<Boolean> hasBucketProperties = Optional.empty();
     private final AtomicLong partitionIdGen = new AtomicLong(0L);
@@ -186,9 +190,8 @@ public class IcebergTable extends Table {
     @Override
     public List<Column> getPartitionColumns() {
         if (partitionColumns == null) {
-            List<PartitionField> partitionFields = this.getNativeTable().spec().fields();
-            Schema schema = this.getNativeTable().schema();
-            partitionColumns = partitionFields.stream().map(partitionField ->
+            Schema schema = getReadSchema();
+            partitionColumns = readSchemaPartitionFields().stream().map(partitionField ->
                     getColumn(getPartitionSourceName(schema, partitionField))).collect(Collectors.toList());
         }
         return partitionColumns;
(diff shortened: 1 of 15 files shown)


dirtysalt (approved):
LGTM! 🚀

PR #76414, opened 2026-07-15
[Tool] Add manual workflow_dispatch to bugfix-version-index

andyziye:
## Why I'm doing:

The `bugfix-version-index` reconcile / reconcile-all sweeps only run on the daily/weekly cron. There is no way to trigger a self-heal on demand (e.g. right after fixing the index or landing a plugin change).

> **Scope note:** this PR is **only** the manual trigger. It does **not** fix the separate issue that this workflow currently fails on the `[ubuntu, ai]` runner because it runs `claude` as **root** (`--dangerously-skip-permissions cannot be used with root`). The obvious `IS_SANDBOX=1` shortcut was **rejected in review** — it merely skips the guard without creating any isolation, and since `pull_request_target` feeds attacker-controllable PR content to a root `claude` holding `CLAUDE_CODE_OAUTH_TOKEN` / `GITHUB_TOKEN` / ContextBase-write plus `--dangerously-skip-permissions`, a prompt injection would be root RCE / credential exfiltration. The real fix (a bubblewrap OS sandbox with `failIfUnavailable` + a network allowlist, or a dedicated non-root user) is a runner-side change handled separately. This dispatch trigger is safe and useful once that lands.

## What I'm doing:

Add a `workflow_dispatch` trigger with a `mode` input (choice: `reconcile` | `reconcile-all`) and extend the `reconcile` / `reconcile-all` job `if` conditions to also fire on the matching manual dispatch. No other job is affected.

celerdata inherits this workflow via the sync pipeline.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [x] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/.github/workflows/bugfix-version-index.yml
+++ b/.github/workflows/bugfix-version-index.yml
@@ -48,6 +48,16 @@ on:
   schedule:
     - cron: '0 19 * * *'    # daily 19:00 UTC (03:00 Asia/Shanghai) → reconcile (pending-only)
     - cron: '0 20 * * 0'    # weekly Sun 20:00 UTC (Mon 04:00 Asia/Shanghai) → reconcile-all (concrete too)
+  workflow_dispatch:        # manual trigger — run a self-heal sweep on demand, don't only wait for cron
+    inputs:
+      mode:
+        description: 'Self-heal sweep to run (same jobs the cron fires)'
+        type: choice
+        required: true
+        default: reconcile
+        options:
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

PR #74838, opened 2026-06-16
[Enhancement] Replace offset implementation in BinaryColumn by AdaptiveOffsets

srlch:
## What I'm doing:
In previous pr, AdaptiveOffsets was introduced but it does not be used at all. In this pr, we replace the original offset implementation in BinaryColumn by AdaptiveOffsets which make the BinaryColumn can represent the data with total size > 4GiB.

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [x] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

PR #76419, opened 2026-07-15
[BugFix] Reject oversized array/map constructor results instead of silent corruption

XinzhuangL:
## Why I'm doing:

`ArrayExpr` and `MapExpr` flatten all children into a single element (key/value) column. When the flattened bytes of one chunk exceed `UINT32_MAX` (~4.3GB), `BinaryColumn`'s uint32 offsets silently wrap around while the bytes buffer keeps growing: every element keeps its correct length but points at bytes ~4GB earlier in the buffer. The query returns silently corrupted `ARRAY<VARCHAR>` / `MAP` values, which can then be persisted by CTAS / INSERT / MV refresh.

Unlike function calls (`function_call_expr.cpp`) and `array_map` / `array_agg`, these two constructor exprs performed no capacity check at all.

**Why this PR targets branch-4.1 instead of main**: on main this is already fixed at the root by #74549 / #74838 (`AdaptiveOffsets` promote `BinaryColumn` offsets to uint64, and `capacity_limit_reached()` was relaxed to 2^64), so oversized results are legal there and this guard is unnecessary (the added UT would even fail on main by design). That rework is an invasive Enhancement and not backportable, while release branches still use `Buffer<uint32_t>` offsets and silently corrupt data, so they need this explicit guard.

## What I'm doing:

Add a `capacity_limit_reached()` check on the constructed column in `ArrayExpr::evaluate_checked` and `MapExpr::evaluate_checked`, so such queries now fail with an explicit `CapacityLimitExceed` error instead of returning wrong data.

Workaround for affected queries: reduce `chunk_size` or split the constructor expression.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5
  - [ ] 3.4

### DIFF
--- a/be/src/exprs/array_expr.cpp
+++ b/be/src/exprs/array_expr.cpp
@@ -70,6 +70,13 @@ public:
         }
 
         auto ptr = ArrayColumn::create(std::move(array_elements), std::move(array_offsets));
+        // Flattening all children into one element column may exceed the 4GB capacity of
+        // BinaryColumn's uint32 offsets within a single chunk, which would silently wrap
+        // and corrupt the data. Fail the query instead.
+        if (auto st = ptr->capacity_limit_reached(); !st.ok()) {
+            return Status::CapacityLimitExceed("array constructor result exceeds column capacity limit, " +
+                                               std::string(st.message()));
+        }
         if (all_const) {
             return ConstColumn::create(std::move(ptr), output_rows);
         }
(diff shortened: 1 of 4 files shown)


satanson (approved):

srlch (approved):

XinzhuangL:
@Mergifyio backport branch-4.0

XinzhuangL:

PR #67567, opened 2026-01-07
[Enhancement] Implement iceberg delete commit operation

Youngwb:
## Why I'm doing:
#66944

## What I'm doing:
This pull request enhances the Iceberg delete file commit flow, ensuring that delete operations are properly committed and validated, and adds robust handling and testing for delete file metadata. The changes introduce a clear distinction between data and delete operations on the FE side, improve the commit callback logic on the BE side, and add comprehensive unit tests for delete file commits.

**Iceberg Delete File Commit Handling:**

- **FE (Frontend) Improvements:**
  - Added logic in `IcebergMetadata` to distinguish delete operations (using `POSITION_DELETES`) and handle them with a new `commitDeleteOperation` method, which uses Iceberg's `RowDelta` for delete file commits. This includes validation of referenced data files, conflict detection, and isolation level handling.
  - Refactored commit and cleanup logic into a reusable `commitWithCleanup` method for both data and delete operations, improving code structure and error handling. 
  - Ensured cache invalidation and metadata refresh after commits to keep the system state consistent.

**BE (Backend) Delete Sink Enhancements:**

- **Commit Callback Logic:**
  - Improved `callback_on_commit` in `IcebergDeleteSink` to construct and record detailed `TIcebergDataFile` metadata when a delete file is committed, including statistics and referenced data file.
  - Updated the file writer callback to set the `referenced_data_file` in the commit result, ensuring correct metadata propagation.
  - Extended `CommitResult` to support setting and storing `referenced_data_file`.

**Testing and Validation:**

- **Unit Test Coverage:**
  - Added comprehensive tests for `callback_on_commit` in `IcebergDeleteSinkTest`, covering cases with empty, partial, complete, and failed commit results, as well as verifying the `referenced_data_file` setter. 

**Dependency and Import Updates:**

- Updated imports in `IcebergMetadata.java` to support new logic for delete file handling and Iceberg APIs. 

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
[...]

PR #67230, opened 2025-12-25
[UT] Fix iceberg cache unstable UT

Youngwb:
## Why I'm doing:
sometimes sr can not query latest snapshot after insert into 
## What I'm doing:
refresh fe cache after insert into finish

Fixes #issue

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [x] UT
- [ ] Doc
- [ ] Tool

Does this PR entail a change in behavior?

- [ ] Yes, this PR will result in a change in behavior.
- [x] No, this PR will not result in a change in behavior.

If yes, please specify the type of change:

- [ ] Interface/UI changes: syntax, type conversion, expression evaluation, display information
- [ ] Parameter changes: default values, similar parameters but with different default values
- [ ] Policy changes: use new policy to replace old one, functionality automatically enabled
- [ ] Feature removed
- [ ] Miscellaneous: upgrade & downgrade compatibility, etc.

## Checklist:

- [ ] I have added test cases for my bug fix or my new feature
- [ ] This pr needs user documentation (for new or modified features or behaviors)
  - [ ] I have added documentation for my new feature or new function
  - [ ] This pr needs auto generate documentation
- [ ] This is a backport pr

## Bugfix cherry-pick branch check:
- [x] I have checked the version labels which the pr will be auto-backported to the target branch
  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4
  - [ ] 3.3

---

> [!NOTE]
> Ensures FE sees the latest Iceberg snapshot right after data commit.
> 
> - In `IcebergMetadata.finishSink`, after `batchWrite.commit()` and `transaction.commitTransaction()`, invoke `refreshTable(dbName, table, new ArrayList<>(), false)` before `asyncRefreshOthersFeMetadataCache(...)` to update FE metadata cache.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4339ec3337375b908f80b9d04e6ce3106c139fd8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #76422, opened 2026-07-15
[BugFix] Invalidate Iceberg table metadata cache after sink commit (backport #67230)

Youngwb:
## Why I'm doing:

On branch-4.0, an Iceberg `INSERT` commits the new snapshot but `finishSink` only invalidates the partition cache and asynchronously refreshes *other* FEs — it never invalidates the committing FE's own cached `Table`. So a read that immediately follows the commit on the same FE resolves its scan against the **stale pre-commit snapshot**.

This is most damaging for a self-referential `INSERT INTO t SELECT * FROM t`: its `SELECT` reads the stale (smaller) snapshot, so it appends fewer rows and **permanently loses data**. Reproduced on a hive Iceberg catalog — the doubling `insert into t select * from t` intermittently committed fewer than 2x rows (e.g. 1,280,000 → 960,000), which cascades into a short `count(*)` and a wrong `c_map != ...` result.

`main` already fixed this: `finishSink` invalidates the local table cache after commit (added in #67230, later refactored into `invalidateCacheAfterCommit`/`commitWithCleanup` by #67567). branch-4.0 never received it — its `finishSink` `finally` still carries the old `// Do we really need that?` comment and only calls `invalidatePartitionCache`.

## What I'm doing:

Invalidate this FE's cached table in the `finishSink` `finally` block after the commit, so a read-after-write sees the new snapshot:

```java
} finally {
    tables.remove(TableIdentifier.of(dbName, tableName));
    icebergCatalog.invalidateCache(dbName, tableName);
}
```

This ports `main`'s post-commit invalidation to branch-4.0 using the existing `invalidateCache` (branch-4.0 has no narrow `invalidateTableCache`; `invalidateCache` is a superset that also clears the manifest/partition caches, which is safe). Verified on a live hive-Iceberg cluster: the doubling ingest that previously lost rows on the first run is now correct across 30/30 runs once the cache is invalidated after commit.

Fixes #issue: N/A — backport of the `finishSink` cache invalidation (#67230 / #67567) missing on branch-4.0; found via Iceberg read-after-write testing.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [x] This is a backport pr

  - [ ] 4.0
  - [ ] 3.5
  - [ ] 3.4

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergMetadata.java
@@ -1609,8 +1609,9 @@ public class IcebergMetadata implements ConnectorMetadata {
             LOG.error("Failed to commit iceberg transaction on {}.{}", dbName, tableName, e);
             throw new StarRocksConnectorException(e.getMessage());
         } finally {
-            // Do we really need that? because partition cache is associated with snapshotId
-            icebergCatalog.invalidatePartitionCache(dbName, tableName);
+            // Invalidate this FE's cached table after commit so a read-after-write sees the new snapshot.
+            tables.remove(TableIdentifier.of(dbName, tableName));
+            icebergCatalog.invalidateCache(dbName, tableName);
         }
     }
 


Astralidea (approved):

Youngwb:
@mergify backport branch-4.0.13

PR #76453, opened 2026-07-16
[BugFix] Reuse inferred CTAS columns for recursive CTE

### LINKED ISSUE #76452: [Bug] Recursive CTE CTAS fails with duplicate column name

1. Enable recursive CTEs for the session:

```sql
SET enable_recursive_cte = true;
```

2. Run a minimal recursive CTE through CTAS:

```sql
CREATE DATABASE IF NOT EXISTS recursive_cte_ctas_repro;
USE recursive_cte_ctas_repro;

CREATE TABLE recursive_result AS
WITH RECURSIVE r(n) AS (
    SELECT 1
    UNION ALL
    SELECT n + 1 FROM r WHERE n < 3
)
SELECT n FROM r;
```

The same failure occurs with `CREATE TEMPORARY TABLE ... AS` when temporary tables are enabled.

Control cases:

- The recursive `WITH RECURSIVE ... SELECT` succeeds without CTAS.
- A separately created table followed by `INSERT INTO ... WITH RECURSIVE ... SELECT` succeeds and inserts `1, 2, 3`.
- Non-recursive CTAS succeeds.

The underlying re-analysis behavior is also reproducible on current `main` at commit `4b09e493bd68dfb7592838bcc7b5d7bc3fa96a35`:

```java
CreateTableAsSelectStmt stmt = (CreateTableAsSelectStmt)
        UtFrameUtils.parseStmtWithNewParser(
                "create table ctas_reanalyze as select 1 as id;", ctx);

// parseStmtWithNewParser has already analyzed the statement once.
Analyzer.analyze(stmt, ctx);
```

Result:

```text
com.starrocks.sql.analyzer.SemanticException:
Getting analyzing error. Detail message: Duplicate column name 'id'.
```

The CTAS statement should succeed, create `recursive_result`, and insert the rows `1`, `2`, and `3`.

Re-analyzing the same CTAS statement should not append another copy of its inferred column definitions.

### Real behavior (Required)

The CTAS statement fails before execution:

```text
ERROR 1060 (42S21): Duplicate column name 'n'
```

### Root cause analysis

The recursive CTE path appears to analyze the same mutable CTAS AST twice:

1. `StmtExecutor` detects a recursive CTE and calls `RecursiveCTEExecutor.splitOuterStmt(parsedStmt)`.
2. `RecursiveCTEExecutor.splitOuterStmt` calls `Analyzer.analyze(stmt, context)` on the complete CTAS statement.
3. `CTASAnalyzer.analyze` infers the query output columns and appends them to `CreateTableStmt.columnDefs` through `addColumnDef`.
4. Normal planning later calls `StatementPlanner.plan`, which invokes `Analyzer.analyze(statement, session)` again on the same statement.
5. `CTASAnalyzer.analyze` appends the inferred columns again. `CreateTableAnalyzer` then sees two columns with the same name and raises error 1060.

Relevant code on current `main`:

- [`StmtExecutor` recursive CTE path](https://github.com/StarRocks/starrocks/blob/4b09e493bd68dfb7592838bcc7b5d7bc3fa96a35/fe/fe-core/src/main/java/com/starrocks/qe/StmtExecutor.java#L979-L996)
- [`RecursiveCTEExecutor.splitOuterStmt`](https://github.com/StarRocks/starrocks/blob/4b09e493bd68dfb7592838bcc7b5d7bc3fa96a35/fe/fe-core/src/main/java/com/starrocks/qe/recursivecte/RecursiveCTEExecutor.java#L90-L118)
- [`CTASAnalyzer` appending inferred columns](https://github.com/StarRocks/starrocks/blob/4b09e493bd68dfb7592838bcc7b5d7bc3fa96a35/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CTASAnalyzer.java#L99-L109)
- [`StatementPlanner` analyzing the statement again](https://github.com/StarRocks/starrocks/blob/4b09e493bd68dfb7592838bcc7b5d7bc3fa96a35/fe/fe-core/src/main/java/com/starrocks/sql/StatementPlanner.java#L281-L291)

### Workaround

Use a non-recursive empty CTAS for schema inference, then insert the recursive query result:

```sql
CREATE TABLE recursive_result AS
SELECT CAST(NULL AS BIGINT) AS n
FROM dual
WHERE 1 = 0;

INSERT INTO recursive_result
WITH RECURSIVE r(n) AS (
    SELECT 1
    UNION ALL
    SELECT n + 1 FROM r WHERE n < 3
)
SELECT n FROM r;
```

```text
4.1.1-14b7e3f
```

The same duplicate-column behavior is reproducible by re-analyzing a CTAS statement on current `main` commit `4b09e493bd68dfb7592838bcc7b5d7bc3fa96a35`.

JacobZheng0927:
## Why I'm doing:

Recursive CTE statements are analyzed once before `RecursiveCTEExecutor` splits them and again during normal planning. For CTAS, both analyses append inferred columns to the outer `CreateTableStmt`, so the second analysis fails with `Duplicate column name`.

## What I'm doing:

- Keep the complete CTAS analysis before recursive CTE splitting so invalid outer CTAS definitions fail before recursive execution and `EXPLAIN CTAS` keeps its semantic checks.
- Reuse columns inferred from the original query when the rewritten CTAS AST is analyzed again, avoiding duplicate columns and preserving the original type and nullability instead of deriving them from the recursive temporary table.
- Add FE regression tests for repeated CTAS analysis, outer CTAS validation, and preservation of inferred `VARCHAR` length.
- Add a SQL integration regression that creates a table from a recursive CTE and verifies rows `1`, `2`, and `3`.

Fixes #76452

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

## Validation:

- `./run-fe-ut.sh --test 'com.starrocks.sql.plan.RecursiveCTETest,com.starrocks.sql.analyzer.CTASAnalyzerTest'` — 30 tests passed.
- `cd fe && mvn checkstyle:check -pl fe-core` — 0 violations.
- `test/.venv/bin/python test/run.py -l --file_filter 'test_recursive_cte'` — SQL integration case discovered successfully; execution is left to CI because no source-build cluster is configured locally.

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CTASAnalyzer.java
+++ b/fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CTASAnalyzer.java
@@ -96,16 +96,20 @@ public class CTASAnalyzer {
         TableName tableNameObject = TableName.fromTableRef(normalizedTableRef);
         CreateTableAnalyzer.analyzeEngineName(createTableStmt, tableNameObject.getCatalog());
 
-        for (int i = 0; i < allFields.size(); i++) {
-            Type type = AnalyzerUtils.transformTableColumnType(allFields.get(i).getType(),
-                    createTableStmt.isOlapEngine(), /* preferStringForVarchar */ false);
-            Expr originExpression = allFields.get(i).getOriginExpression();
-            ColumnDef columnDef = new ColumnDef(finalColumnNames.get(i), new TypeDef(type), false,
-                    null, null, originExpression.isNullable(), ColumnDef.DefaultValueDef.NOT_SET, "");
-            if (isPKTable && keysDesc.containsCol(finalColumnNames.get(i))) {
-                columnDef.setAllowNull(false);
+        // Recursive CTE execution rewrites and analyzes the same CTAS AST again. Preserve the columns
+        // inferred from the original query instead of appending columns derived from the rewritten query.
+        if (createTableStmt.getColumnDefs().isEmpty()) {
+            for (int i = 0; i < allFields.size(); i++) {
+                Type type = AnalyzerUtils.transformTableColumnType(allFields.get(i).getType(),
+                        createTableStmt.isOlapEngine(), /* preferStringForVarchar */ false);
+                Expr originExpression = allFields.get(i).getOriginExpression();
+                ColumnDef columnDef = new ColumnDef(finalColumnNames.get(i), new TypeDef(type), false,
+                        null, null, originExpression.isNullable(), ColumnDef.DefaultValueDef.NOT_SET, "");
+                if (isPKTable && keysDesc.containsCol(finalColumnNames.get(i))) {
+                    columnDef.setAllowNull(false);
+                }
+                createTableStmt.addColumnDef(columnDef);
             }
-            createTableStmt.addColumnDef(columnDef);
         }
 
         // For replication_num, The behavior is the same as creating a table
--- a/fe/fe-core/src/test/java/com/starrocks/sql/analyzer/CTASAnalyzerTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/sql/analyzer/CTASAnalyzerTest.java
@@ -514,6 +514,19 @@ public class CTASAnalyzerTest {
         }
     }
 
+    @Test
+    public void testRepeatedAnalysisReusesInferredColumns() throws Exception {
+        ConnectContext ctx = starRocksAssert.getCtx();
+        String sql = "create table test_repeated_analysis as select cast('abc' as varchar(10)) as c;";
+        CreateTableAsSelectStmt stmt =
+                (CreateTableAsSelectStmt) UtFrameUtils.parseStmtWithNewParser(sql, ctx);
+
+        Analyzer.analyze(stmt, ctx);
+
+        Assertions.assertEquals(1, stmt.getCreateTableStmt().getColumnDefs().size());
+        assertVarcharLength(stmt.getCreateTableStmt().getColumnDefs().get(0).getTypeDef(), 10);
+    }
+
     private static void assertVarcharLength(TypeDef typeDef, int expectedLength) {
         ScalarType scalarType = (ScalarType) typeDef.getType();
         Assertions.assertEquals(PrimitiveType.VARCHAR, scalarType.getPrimitiveType());
--- a/fe/fe-core/src/test/java/com/starrocks/sql/plan/RecursiveCTETest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/sql/plan/RecursiveCTETest.java
@@ -19,6 +19,7 @@ import com.starrocks.common.profile.Tracers;
 import com.starrocks.qe.recursivecte.RecursiveCTEAstCheck;
 import com.starrocks.qe.recursivecte.RecursiveCTEExecutor;
 import com.starrocks.sql.analyzer.SemanticException;
+import com.starrocks.sql.ast.CreateTableAsSelectStmt;
 import com.starrocks.sql.ast.StatementBase;
 import com.starrocks.sql.parser.SqlParser;
 import org.junit.jupiter.api.Assertions;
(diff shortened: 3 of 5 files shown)


CelerData-Reviewer:
@codex review

Smith-Cruise (approved):
LGTM

Smith-Cruise:
Reproduce problem by
```sql
create TEMPORARY table t1 
as WITH RECURSIVE numbers AS (
    SELECT 1 AS n

    UNION ALL

    SELECT n + 1
    FROM numbers
    WHERE n < 5
) select * from numbers;
```

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CTASAnalyzer.java
```diff
@@ -96,16 +96,20 @@ public static void analyze(CreateTableAsSelectStmt createTableAsSelectStmt, Conn
...
-            Type type = AnalyzerUtils.transformTableColumnType(allFields.get(i).getType(),
-                    createTableStmt.isOlapEngine(), /* preferStringForVarchar */ false);
-            Expr originExpression = allFields.get(i).getOriginExpression();
-            ColumnDef columnDef = new ColumnDef(finalColumnNames.get(i), new TypeDef(type), false,
-                    null, null, originExpression.isNullable(), ColumnDef.DefaultValueDef.NOT_SET, "");
-            if (isPKTable && keysDesc.containsCol(finalColumnNames.get(i))) {
-                columnDef.setAllowNull(false);
+        // Recursive CTE execution rewrites and analyzes the same CTAS AST again. Preserve the columns
+        // inferred from the original query instead of appending columns derived from the rewritten query.
+        if (createTableStmt.getColumnDefs().isEmpty()) {
```
stephen-shelby, line 101: This guard is too narrow — the fix doesn't hold for CTAS with more than one column.

It only makes the `columnDefs` list idempotent. Line 162 `Analyzer.analyze(createTableStmt, session)`
still runs on the second pass, and `CreateTableAnalyzer` is not idempotent on the ColumnDef objects
that the first pass already mutated: `analyzeKeysDesc()` stamps `AggregateType.NONE` onto every
non-key column (CreateTableAnalyzer.java:435-441), and the second pass then rejects exactly that
(CreateTableAnalyzer.java:415-428).

Repro — only 2 columns, the most common recursive-CTE shape:

```sql
SET enable_recursive_cte = true;
CREATE TABLE t AS
WITH RECURSIVE cte(path, lvl) AS (
    SELECT CAST('a' AS VARCHAR(50)), CAST(1 AS BIGINT)
    UNION ALL
    SELECT CAST(CONCAT(path, '/a') AS VARCHAR(50)), lvl + 1 FROM cte WHERE lvl < 3
)
SELECT path, lvl FROM cte;
```

before: `Duplicate column name 'path'`
after:  `DUP_KEYS table should not specify aggregate type for non-key column[lvl]`

It fails whenever the derived DUP_KEYS prefix doesn't cover all columns, i.e. 4+ columns
(SHORTKEY_MAX_COLUMN_COUNT = 3), or a leading VARCHAR/STRING column. All three new tests and the
new SQL case use a single column, which is the only shape that passes.

stephen-shelby (approved):

fe/fe-core/src/main/java/com/starrocks/sql/analyzer/CTASAnalyzer.java
Smith-Cruise: You're right

stephen-shelby: approved as hot fix

wyb (approved):

PR #66784, opened 2025-12-16
[Enhancement] Support complex type for paimon table

miomiocat:
## Why I'm doing:
Support complex type for paimon table

## What I'm doing:

Without this PR, querying Paimon complex types may result in the following core dump：
```
*** Aborted at 1750398852 (unix time) try "date -d @1750398852" if you are using GNU date ***
PC: @          0x72eef74 starrocks::parquet::LakeMetaHelper::_is_valid_type(starrocks::parquet::ParquetField const*, starrocks::TIcebergSchemaField const*, starrocks::TypeDescriptor const*) const
*** SIGSEGV (@0x30) received by PID 8 (TID 0x7f2edadfc700) from PID 48; stack trace: ***
    @     0x7f302b40620b __pthread_once_slow
    @          0x7efc4a0 google::(anonymous namespace)::FailureSignalHandler(int, siginfo_t*, void*)
    @     0x7f302f148c65 os::Linux::chained_handler(int, siginfo_t*, void*)
    @     0x7f302f14dc61 JVM_handle_linux_signal
    @     0x7f302f140bd8 signalHandler(int, siginfo_t*, void*)
    @     0x7f302b40f630 (/usr/lib64/libpthread-2.17.so+0xf62f)
    @          0x72eef74 starrocks::parquet::LakeMetaHelper::_is_valid_type(starrocks::parquet::ParquetField const*, starrocks::TIcebergSchemaField const*, starrocks::TypeDescriptor const*) const
    @          0x72eef7d starrocks::parquet::LakeMetaHelper::_is_valid_type(starrocks::parquet::ParquetField const*, starrocks::TIcebergSchemaField const*, starrocks::TypeDescriptor const*) const
    @          0x72ef2d6 starrocks::parquet::LakeMetaHelper::prepare_read_columns(std::vector<starrocks::HdfsScannerContext::ColumnInfo, std::allocator<starrocks::HdfsScannerContext::ColumnInfo> > const&, std::vector<starrocks::parquet::GroupReaderParam::Column, std::allocator<sta�Y�
    @          0x72c0f5b starrocks::parquet::FileReader::init(starrocks::HdfsScannerContext*)
    @          0x70ec6ae starrocks::HdfsParquetScanner::do_open(starrocks::RuntimeState*)
    @          0x70ddf10 starrocks::HdfsScanner::open(starrocks::RuntimeState*)
    @          0x706ab88 starrocks::connector::HiveDataSource::_init_scanner(starrocks::RuntimeState*)
[...]

PR #76455, opened 2026-07-16
[BugFix] Fix native parquet reader crash on incomplete nested lake schema

dontknow9179:
## Why I'm doing:

When the FE ships an incomplete nested lake schema for a complex column — only the top-level field id/name with no children (e.g. Paimon complex columns such as `ARRAY<STRUCT<...>>` before #66784) — the native parquet reader crashes with SIGSEGV while validating the column type.

`LakeMetaHelper::_is_valid_type()` iterates the ARRAY/MAP children using the parquet-file field's children count as the loop bound, but indexes the FE `field_schema->children[idx]` with the same index. The FE `field_schema` children vector can be empty (or shorter) than the file's, so `&field_schema->children[idx]` reads past the end of the vector and produces a null pointer, which is then dereferenced one level down in the STRUCT branch:

```
PC: starrocks::parquet::LakeMetaHelper::_is_valid_type(...)
*** SIGSEGV (@0x30) received by PID ...
```

## What I'm doing:

Bounds-check every index access in the ARRAY/MAP branch against all three vectors (`parquet_field`, `field_schema`, `type_descriptor`). When the FE lake schema does not fully describe the nesting, stop and report the type as invalid instead of indexing out of bounds; the caller already handles an invalid type gracefully by skipping the column.

Note: #66784 fixed the FE side (recursively filling nested children), so the well-formed path no longer produces a truncated schema. This PR hardens the CN side so a truncated schema can never crash the BE.

Adds `LakeMetaHelperTest` with a regression case that reproduces the crash, plus coverage for the complete-schema and scalar paths.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/be/src/formats/parquet/meta_helper.cpp
+++ b/be/src/formats/parquet/meta_helper.cpp
@@ -204,6 +204,11 @@ bool LakeMetaHelper::_is_valid_type(const ParquetField* parquet_field, const TIc
 
     if (parquet_field->type == ColumnType::ARRAY || parquet_field->type == ColumnType::MAP) {
         for (size_t idx = 0; idx < parquet_field->children.size(); idx++) {
+            // Never index past the end of any of the three vectors -- a mismatch means we cannot
+            // validate this nested type, so stop and report it as invalid.
+            if (idx >= field_schema->children.size() || idx >= type_descriptor->children.size()) {
+                break;
+            }
             if (_is_valid_type(&parquet_field->children[idx], &field_schema->children[idx],
                                &type_descriptor->children[idx])) {
                 has_valid_child = true;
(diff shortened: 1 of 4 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

satanson (approved):

wyb (approved):

PR #76456, opened 2026-07-16
[BugFix] Reconnect before get_table_req fallback in Hive getTable()

### LINKED ISSUE #76454: Iceberg Hive catalog: getTable() falls back to get_table_req() on the same Thrift connection after a socket timeout, causing "out of sequence response"
## Steps to reproduce the behavior (Required)

Querying an Iceberg table in a **Hive catalog backed by a Thrift HMS** intermittently fails. The FE leader log shows table metadata refresh failing:

```
WARN (hive-background-iceberg-worker-pool-*) [CachingIcebergCatalog$1.reload()] refresh table <db>.<table> failed
java.lang.RuntimeException: Failed to get table info from metastore <db>.<table>
    at org.apache.iceberg.hive.HiveTableOperations.doRefresh(HiveTableOperations.java:120)
    ...
    at com.starrocks.connector.iceberg.hive.IcebergHiveCatalog.getTable(IcebergHiveCatalog.java:113)
    at com.starrocks.connector.iceberg.CachingIcebergCatalog$1.reload(CachingIcebergCatalog.java:143)
Caused by: org.apache.thrift.TApplicationException: get_table_req failed: out of sequence response: expected 3 but got 2
    at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:89)
    at org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore$Client.recv_get_table_req(ThriftHiveMetastore.java:2095)
    at org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore$Client.get_table_req(ThriftHiveMetastore.java:2082)
    at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.getTable(HiveMetaStoreClient.java:664)   <-- fallback call
    ...
```

Shortly before it, the same connection logs a `get_table` read timeout:

```
WARN [HiveMetaStoreClient.getTable():661] Using get_table() failed, fail over to use get_table_req()
org.apache.thrift.transport.TTransportException: java.net.SocketTimeoutException: Read timed out
    at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:79)
    at org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore$Client.recv_get_table(ThriftHiveMetastore.java:2042)
    at org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore$Client.get_table(ThriftHiveMetastore.java:2028)
    at org.apache.hadoop.hive.metastore.HiveMetaStoreClient.getTable(HiveMetaStoreClient.java:656)   <-- get_table() call
    ...
```

When the metadata refresh fails, the cached Iceberg table becomes unavailable and user queries surface `ERROR 5502 ... Unknown table`.

## Expected behavior (Required)

A slow/timed-out HMS call should fail cleanly (and be retried on a fresh connection by `RetryingMetaStoreClient`), not corrupt the connection and produce `out of sequence response`.

## Root cause

`HiveMetaStoreClient.getTable(catName, dbName, tableName)` calls `get_table()` first and, on **any** exception, falls over to `get_table_req()` **on the same Thrift connection without reconnecting**:

```java
try {
    return client.get_table(dbName, tableName);          // seqid N
} catch (NoSuchObjectException e) {
    throw e;
} catch (Exception e) {                                   // also catches SocketTimeoutException / TTransportException
    LOG.warn("Using get_table() failed, fail over to use get_table_req()", e);
    GetTableRequest req = new GetTableRequest(dbName, tableName);
    req.setCapabilities(version);
    return client.get_table_req(req).getTable();          // seqid N+1, SAME socket, no reconnect
}
```

When `get_table()` (seqid N) read-times-out, its request was already sent; the HMS response arrives late and stays buffered on the socket. The `catch (Exception)` then issues `get_table_req()` (seqid N+1) on that same socket, which reads the stale `get_table` response → Thrift `TApplicationException.BAD_SEQUENCE_ID` → `out of sequence response: expected N+1 but got N`.

Because it happens on the background async metadata-refresh threads (`CachingIcebergCatalog` async reload), the failed refresh leaves the table's cache entry unusable, so user queries then fail with `Unknown table`.

**Why it's intermittent:** `RetryingMetaStoreClient` treats `BAD_SEQUENCE_ID` as retriable and reconnects, so it usually self-heals; it surfaces to the user when the HMS is slow enough that the retry also fails (or when retries are exhausted).

## How other projects handle the same spot

- **Apache Hive** (`HiveMetaStoreClient.getTable`): calls `get_table_req()` directly — no `get_table()`-first fallback, so no same-socket second call.
- **Trino** (`ThriftHiveMetastoreClient`): its `alternativeCall(...)` helper, before trying the next alternative after a failure, does `disconnect(); connect();` with the comment *"Client that threw exception is in an unknown state. We need to open it again to make sure it will respond properly to the next call."*

StarRocks broadened the catch to `Exception` (so it catches transport timeouts) **and** omitted the reconnect between the two calls — that combination is the bug.

## Proposed fix

Reconnect before the `get_table_req()` fallback, so the fallback runs on a clean connection (mirrors Trino's behavior). Alternatively, only fall back for the legitimate "method not supported" case (`TApplicationException.UNKNOWN_METHOD`) and let transport exceptions propagate.

## Real-time flink and StarRocks versions

Observed on the Iceberg Hive catalog path (`iceberg-hive-metastore` client pool). Present on `main` and 3.5.x (the affected `getTable` fallback is identical across branches).

Youngwb:
## Why I'm doing:

Querying an Iceberg table in a Hive catalog (Thrift HMS) intermittently fails with `out of sequence response`, and downstream with `ERROR 5502 ... Unknown table`.

`HiveMetaStoreClient.getTable()` tries `get_table()` first and, on any exception, falls over to `get_table_req()` **on the same Thrift connection without reconnecting**. When `get_table()` read-times-out, its request was already sent and the HMS response arrives late, staying buffered on the socket. The fallback `get_table_req()` then reads that stale response, desyncing the Thrift seqid → `TApplicationException.BAD_SEQUENCE_ID` (`out of sequence response: expected N+1 but got N`). Because it happens on the background Iceberg metadata-refresh threads, the failed refresh leaves the cached table unusable and user queries fail with `Unknown table`.

`RetryingMetaStoreClient` treats `BAD_SEQUENCE_ID` as retriable and reconnects, so it usually self-heals; it surfaces to the user when the HMS is slow enough that the retry also fails.

For reference: Apache Hive's `getTable()` calls `get_table_req()` directly (no fallback), and Trino's thrift metastore client reconnects (`disconnect(); connect();`) before trying an alternative call, with the comment "Client that threw exception is in an unknown state."

## What I'm doing:

Reconnect before the `get_table_req()` fallback so it runs on a clean connection instead of reading `get_table()`'s stale buffered response.

Fixes #76454

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
+++ b/fe/fe-core/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.java
@@ -660,6 +660,10 @@ public class HiveMetaStoreClient implements IMetaStoreClient, AutoCloseable {
             throw e;
         } catch (Exception e) {
             LOG.warn("Using get_table() failed, fail over to use get_table_req()", e);
+            // Reconnect first: get_table() may have failed on a read timeout that left an
+            // undrained response on the socket; reusing it for get_table_req() desyncs the
+            // Thrift seqid ("out of sequence response").
+            reconnect();
             GetTableRequest req = new GetTableRequest(dbName, tableName);
             req.setCapabilities(version);
             return client.get_table_req(req).getTable();
--- a/fe/fe-core/src/test/java/com/starrocks/connector/hive/HiveMetastoreClientTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/connector/hive/HiveMetastoreClientTest.java
@@ -21,12 +21,18 @@ import mockit.Mocked;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
+import org.apache.hadoop.hive.metastore.api.GetTableRequest;
+import org.apache.hadoop.hive.metastore.api.GetTableResult;
 import org.apache.hadoop.hive.metastore.api.NoSuchObjectException;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.api.ThriftHiveMetastore;
 import org.apache.thrift.TException;
 import org.apache.thrift.transport.TSocket;
+import org.apache.thrift.transport.TTransportException;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+
+import java.net.SocketTimeoutException;
 
 public class HiveMetastoreClientTest {
     @Test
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

dirtysalt (approved):
LGTM! 🚀

Astralidea (approved):

PR #76459, opened 2026-07-16
[BugFix] Detect Hive table-not-found by exception type instead of message text

dirtysalt:
## Why I'm doing:
#75490 tightened the exception messages that connector metadata calls (Hive/Delta/Iceberg/etc.) surface to users, dropping the inlined root-cause text from the top-level message and pushing it into the cause chain instead. Two call sites, however, were relying on substring-matching that inlined text (`e.getMessage().contains("NoSuchObjectException")` / `"table not found"`) to detect "the Hive base table was dropped" and treat it as an expected, swallow-able case. Once the message format changed, that detection silently broke: a materialized view whose Hive base table gets dropped no longer gets marked `INACTIVE` on refresh, and refresh failures get logged as noisy errors instead of being handled gracefully. See StarRocksTest#11584.

## What I'm doing:
Stop pattern-matching on exception message text and instead walk the full cause chain by exception type, which is stable regardless of how outer wrapper messages are worded.

- Added `LogUtil.isCausedBy(Throwable, Class<? extends Throwable>)`, a small shared helper that checks whether any exception in the cause chain is an instance of a given type.
- `MvUtils.getTable` / `MvUtils.getTableWithIdentifier`: replaced the message substring check with `LogUtil.isCausedBy(e, NoSuchObjectException.class)` to detect a dropped Hive table and return `Optional.empty()` instead of propagating the exception.
- `AnalyzeStatusSystemTable`: same fix, applied to suppress the (cosmetic) warning log for the same expected "table not found" case.

```
before: SRConnectorException("Failed to get table [...], msg: NoSuchObjectException: ...")
              ^--- substring match on this text, broke after #75490 shortened it

after:  SRConnectorException("Failed to get table [...]")
              └─ cause: InvocationTargetException
                    └─ cause: NoSuchObjectException      <-- matched by type via getCausalChain
```

issue: https://github.com/StarRocks/StarRocksTest/issues/11584

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/system/information/AnalyzeStatusSystemTable.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/system/information/AnalyzeStatusSystemTable.java
@@ -23,6 +23,7 @@ import com.starrocks.catalog.system.SystemId;
 import com.starrocks.catalog.system.SystemTable;
 import com.starrocks.common.MetaNotFoundException;
 import com.starrocks.common.util.DateUtils;
+import com.starrocks.common.util.LogUtil;
 import com.starrocks.qe.ConnectContext;
 import com.starrocks.qe.ShowResultSetMetaData;
 import com.starrocks.server.GlobalStateMgr;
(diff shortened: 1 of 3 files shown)


CelerData-Reviewer:
@codex review

HangyuanLiu (approved):

Youngwb (approved):

PR #76460, opened 2026-07-16
[BugFix] Persist merged parallel-compaction txn log on the regular path

meegoo:
## Why I'm doing:

On a non-file-bundling cloud-native (shared-data) table, a compaction transaction can get stuck in `COMMITTED` forever and block every subsequent transaction on the partition (routine load, inserts, later compactions). The publish daemon 404s on a txn log that was never written (`.../log/<txn>_<tablet>.log` → later "Both txn_log and corresponding tablet_meta missing").

Root cause: parallel lake compaction splits one tablet's compaction (for a single `tablet_id + txn_id`, which has exactly one txn-log location) into subtasks whose partial txn logs are merged into one. The subtasks are forced to `skip_write_txnlog=true` so their partial logs can be held and merged, which is necessary. But the **merged** context was *also* constructed with `skip_write_txnlog=true` unconditionally, which routes the merged log into `CompactResponse.txn_logs` for an aggregator to persist.

That is only correct on the **aggregate / file-bundling** path (FE sets `skip_write_txnlog=true`, and `LakeServiceImpl::aggregate_compact` → `collect_txnlogs` + `write_combined_txn_log` persists a single combined log). On the **regular / non-file-bundling** path FE never sets `skip_write_txnlog` and there is **no aggregator**: the merged log is neither written to object storage by the CN nor read by FE (a repo-wide search confirms no FE code reads `CompactResponse.txn_logs`), so it is silently dropped. FE sees an empty `failed_tablets`, commits the compaction txn, and publish then can never find the log.

This is why on the reported cluster the tablets that took the parallel path were stuck while the tablets that fell back to the serial path (which keeps `skip_write_txnlog=false` and writes the log normally) published fine.

Introduced by #66586 (support parallel compaction). Affects all non-file-bundling cloud-native tables that have a tablet large enough to trigger the parallel path; parallel compaction is on by default (`lake_compaction_max_parallel_default=3`).

## What I'm doing:

Honor the originating request's `skip_write_txnlog` intent for the **merged** context, instead of hardcoding `true`:

- **Aggregate/file-bundling path** (`skip_write_txnlog=true`): unchanged — keep `skip=true` so the merged log is returned inline via `CompactResponse.txn_logs` for the aggregator to combine and persist once.
- **Regular path** (`skip_write_txnlog=false`): build the merged context with `skip=false` and persist the merged log directly via `TabletManager::put_txn_log()` (which normalizes the log before saving), exactly as a serial compaction does in `CompactionTask::execute()`.

Subtasks still force `skip_write_txnlog=true` so their partial logs can be merged into one. The write happens on the subtask-completion (worker) thread, off the `CompactionTaskCallback` mutex, matching the serial path. `op_parallel_compaction` in a standalone txn log is already handled end-to-end by `txn_log_applier` (`apply_parallel_compaction_log`) and by `normalize_txn_log_before_save`, so publish works unchanged.

Fixes the stuck-`COMMITTED` / unpublishable compaction txn on non-file-bundling tables.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/storage/lake/compaction_scheduler.cpp
+++ b/be/src/storage/lake/compaction_scheduler.cpp
@@ -80,6 +80,10 @@ bool CompactionTaskCallback::allow_partial_success() const {
     }
 }
 
+bool CompactionTaskCallback::skip_write_txnlog() const {
+    return _request->has_skip_write_txnlog() && _request->skip_write_txnlog();
+}
+
 Status CompactionTaskCallback::has_error() const {
     std::lock_guard l(_mtx);
     if (_status.ok()) {
--- a/be/src/storage/lake/compaction_scheduler.h
+++ b/be/src/storage/lake/compaction_scheduler.h
@@ -71,6 +71,12 @@ public:
 
     bool allow_partial_success() const;
 
+    // Whether the originating FE request asked the CN to skip persisting the txn log to object
+    // storage and instead return it inline via the RPC response. Only the aggregate/file-bundling
+    // path sets this; on the regular path it is false, meaning the CN must persist the txn log
+    // itself. The parallel compaction manager consults this to decide how to deliver the merged log.
+    bool skip_write_txnlog() const;
+
     void set_last_check_time(int64_t now) {
         std::lock_guard l(_txn_valid_check_mutex);
         _last_check_time = now;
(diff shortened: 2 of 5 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

sevev (approved):

luohaha (approved):

PR #76478, opened 2026-07-16
[BugFix] Fix Finished Queries CSV download escaping and full-statement export

ycra-dev:
## Why I'm doing:

The **Download CSV** button on the FE web UI `/query` (Finished Queries) page produces a corrupted file:

1. **Broken alignment / field splitting.** `download_query()` joins cell text with commas without any escaping, and groups fields using a hard-coded column count of `9` while the table actually renders `11` columns. As a result every row is misaligned, and any SQL statement containing a comma or newline (e.g. a multi-line query) spills across cells and rows.
2. **Truncated statements for heavy queries.** A long SQL statement is rendered truncated on screen (`ui_queries_sql_statement_max_length`, default 128), with the full text kept only in the child element's `title` attribute. The export reads `textContent`, so heavy queries are captured only up to the truncated `... ` prefix.

Example (before), a single multi-line query breaks into multiple physical CSV lines and shifts all following columns; the standard CSV parser reads far more "rows" than there are queries.

## What I'm doing:

Fix `webroot/static/starrocks.js` `download_query()`:

- Escape each field per **RFC 4180** — wrap values containing a comma, double quote, or newline in double quotes and double any inner quotes.
- Derive the column count from the header row (`table th`) instead of the hard-coded `9`.
- Read the full statement from the child element's `title` attribute when present, so heavy (truncated) queries are exported in full.

Verified against a StarRocks 4.1.1 cluster: short, long single-line, and long multi-line queries all export as well-formed CSV — a standard parser reads exactly one record per query, each with 11 columns, with newlines/commas and the full statement preserved.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [x] 3.5

### DIFF
--- a/webroot/static/starrocks.js
+++ b/webroot/static/starrocks.js
@@ -17,15 +17,23 @@
 
 function download_query() {
     let tableThs = Array.from($('table th')).map(v=>{return v.textContent})
-    let tableTds = Array.from($('table td')).map(v=>{return v.textContent})
+    let cellValue = (el)=>{
+        let titled = el.querySelector('[title]')
+        return titled ? titled.getAttribute('title') : el.textContent
+    }
+    let tableTds = Array.from($('table td')).map(cellValue)
     let combine = tableThs.concat(tableTds)
     let csv = ''
-    let colWidth = 9
+    let colWidth = tableThs.length
     let column = []
+    let escapeCsv = (v)=>{
+        v = (v == null ? '' : v).toString()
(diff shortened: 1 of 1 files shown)


CelerData-Reviewer:
@codex review

Issue #59226, opened 2025-05-22
BE hang on signalHandler


sevev:
### Steps to reproduce the behavior (Required)
1. if `enable_core_file_size_optimization` and enable release data_cache before core dump, the BE maybe hang and the stack is like following:
```
Thread 511 (Thread 0x739a97fcd640 (LWP 848) "pip_exec_com"):
#0  0x0000739bde5ef2c0 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x0000739bde5f6002 in pthread_mutex_lock () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x000000000b3dea8e in je_malloc_mutex_lock_slow ()
#3  0x000000000b3d349c in je_ecache_alloc_grow ()
#4  0x000000000b3e0465 in pac_alloc_real ()
#5  0x000000000b3e05c0 in pac_alloc_impl ()
#6  0x000000000b3df485 in je_pa_alloc ()
#7  0x000000000b37af39 in je_arena_extent_alloc_large ()
#8  0x000000000b3dbce5 in je_large_palloc ()
#9  0x000000000b365bbf in je_malloc_default ()
#10 0x0000000008ab8bb8 in my_malloc ()
#11 0x000000000f006f0c in operator new(unsigned long) ()
#12 0x0000000008c9f9f2 in void std::vector<starrocks::LRUHandle*, std::allocator<starrocks::LRUHandle*> >::_M_realloc_insert<starrocks::LRUHandle* const&>(__gnu_cxx::__normal_iterator<starrocks::LRUHandle**, std::vector<starrocks::LRUHandle*, std::allocator<starrocks::LRUHandle*> > >, starrocks::LRUHandle* const&) ()
#13 0x0000000008c9e5ea in starrocks::LRUCache::_evict_from_lru(unsigned long, std::vector<starrocks::LRUHandle*, std::allocator<starrocks::LRUHandle*> >*) ()
#14 0x0000000008c9e7d1 in starrocks::LRUCache::set_capacity(unsigned long) ()
#15 0x0000000008c9e95b in starrocks::ShardedLRUCache::set_capacity(unsigned long) ()
#16 0x000000000671f562 in starrocks::StoragePageCache::set_capacity(unsigned long) ()
#17 0x00000000053c8687 in starrocks::ExecEnv::try_release_resource_before_core_dump() ()
#18 0x0000000008af6d0d in starrocks::failure_handler_after_output_log() ()
[...]

PR #76491, opened 2026-07-16
[BugFix] Force BE exit when the fatal-signal handler hangs after marking the process as crashing

### LINKED ISSUE #76441: BE/CN can get permanently stuck reporting SHUTDOWN ("BE is shutting down") while the process stays alive, if the fatal-signal handler stalls before abort()

**Background / root cause.** `#66212` (2025-12) added a one-way "crashing" flag so the FE learns a BE is not alive during a crash. The flag is set at the very top of the fatal-signal handler, before the process is actually terminated:

```
// be/src/service/failure_handler.cpp
static void failure_handler_after_output_log() {
    ...
        set_process_is_crashing();              // :181  flag ON
        try_release_resources_before_core_dump(...);
        dontdump_unused_pages();                // :192  calls je_mallctl(arena.purge / arena.dontdump)
    ...
}
static void failure_function() {
    dump_trace_info();
    failure_handler_after_output_log();         // :206
    std::abort();                               // :207  the process only dies HERE
}
```

The flag is set-only: `be/src/common/process_exit.h` declares `set_process_is_crashing()` / `is_process_crashing()` but there is **no `clear_process_is_crashing()`** (contrast the sibling flag, which has `clear_frontend_aware_of_exit()`). So if the handler stalls anywhere between `set_process_is_crashing()` and `std::abort()`, the process keeps running with the flag permanently on. This is a known failure mode for the crash path — `#59226` documented the crash handler hanging while re-acquiring a jemalloc mutex, and the `dontdump_unused_pages()` call on line 192 is exactly such a `je_mallctl` call.

Once the flag is stuck on, the heartbeat server answers **every** FE heartbeat with SHUTDOWN forever:

```
// be/src/agent/heartbeat_server.cpp:96
if (process_exit_in_progress() || is_process_crashing()) {
    res = Status::Shutdown("BE is shutting down");
}
```

and nothing recovers it: the main loop only checks `process_exit_in_progress()` and ignores `is_process_crashing()` (`starrocks_be.cpp:431`, so the process does not self-terminate), the FE has no auto-drop for a node stuck in SHUTDOWN, and `/api/health` is a hardcoded `200 OK` (`health_action.cpp:61,66`, body `"To Be Added"`) that never reflects the crashing/exiting state, so a k8s liveness probe pointed at it can never restart the pod.

**Deterministic reproduction.** The natural trigger (a genuine handler stall inside jemalloc) is hard to force on demand, so the steps below reproduce the exact *state the handler leaves behind* — flag on, process alive — by setting the flag the same way line 181 does, then letting the process run. Verified on `starrocks/allin1-ubuntu:4.0-latest` (`select current_version()` → `4.0.10-5b36164`), core limit `unlimited`.

1. Start the cluster and confirm the BE is alive:
```
docker run -d --name sr --ulimit core=-1 --cap-add=SYS_PTRACE --security-opt seccomp=unconfined starrocks/allin1-ubuntu:4.0-latest
# wait for startup, then:
mysql -h127.0.0.1 -P9030 -uroot -e "SHOW PROC '/backends'\G" | grep -E 'Alive|ErrMsg|StatusCode'
#   Alive: true / ErrMsg: (empty) / StatusCode: OK
```

2. (Control) send a real fatal signal — this is the intended behavior and is **not** the bug:
```
kill -SEGV $(pgrep -x starrocks_be)
# process dies within ~1s, supervisor restarts it within ~3s, node returns to OK.
```

3. (Bug) put the process into the state the handler leaves if it stalls after line 181 — set the flag, then leave the process running:
```
apt-get update && apt-get install -y gdb

BE=$(pgrep -x starrocks_be)
gdb -q -batch -p $BE \
  -ex 'call (void)_ZN9starrocks23set_process_is_crashingEv()' \
  -ex 'print (char)_ZN9starrocks19is_process_crashingEv()' \
  -ex 'detach'
#   $1 = 1   (flag is now on; the process was NOT killed and keeps running)
```
(Using the mangled linkage name because `starrocks::set_process_is_crashing` may not resolve in the top frame's context on a release build.)

4. Observe the node from the FE, repeatedly:
```
mysql -h127.0.0.1 -P9030 -uroot -e "SHOW PROC '/backends'\G" | grep -E 'Alive|ErrMsg|StatusCode'
curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8040/api/health
```

A BE/CN that has entered the fatal-signal handler must eventually **terminate** (so the orchestrator restarts it) or otherwise return to a serving state. It must never remain alive while permanently reporting SHUTDOWN with no path back to OK, and `/api/health` should reflect the crashing/shutting-down state so an external liveness probe can act on it.

### Real behavior (Required)

The node stays wedged indefinitely:

- `SHOW PROC '/backends'` (and `SHOW COMPUTE NODES` in shared-data) reports `Alive: false`, `StatusCode: SHUTDOWN`, `ErrMsg: BE is shutting down` — continuously. In the reproduction it held this for **7.5+ minutes with zero OK heartbeats**, and the process uptime never changed (never restarted).
- `/api/health` returns **200 the entire time**, so a k8s liveness probe passes and the pod is never restarted.
- The only recovery is killing the process (which lets the supervisor/orchestrator start a fresh one); on restart the flag is clear and the first OK heartbeat returns the node to `Alive: true` immediately.

In shared-data mode this is especially dangerous: such a node is excluded from scheduling but still counts as a replica, so zombies accumulate silently and, once a warehouse's available CN count reaches zero, queries fail cluster-wide.

This was first noticed on an internal production cluster via BE memory metrics — the affected node's jemalloc memory flat-lined (stopped changing), and `SHOW` then revealed it stuck at "BE is shutting down". The raw production metrics/logs are internal and can't be attached here, which is why the reproduction above isolates the same failure state on a public image so it can be verified independently. A live `gdb -p <be_pid>` + `thread apply all bt` on such a stuck node (before restarting it) would confirm the stall site — expected to be a thread parked in `je_mallctl` / `dontdump_unused_pages`, consistent with `#59226`.

main (`86fe47fb3f0`). The same code shape exists on the 3.5 / 4.0 / 4.1 release branches (introduced by #66212).

ronjay1827:
## Why I'm doing:

Fixes #76441.

PR #66212 made the BE report SHUTDOWN to the FE as soon as the fatal-signal handler runs, by setting a one-way `k_starrocks_be_crashing` flag at the top of the handler (`be/src/service/failure_handler.cpp` `failure_handler_after_output_log`) before the re-raise. The flag has no clear path, and the main loop never checks it, so if the handler stalls between setting the flag and the re-raise the process keeps running with the flag on forever. This is a real failure mode for the crash path: #59226 documented the handler hanging while re-acquiring a jemalloc mutex, and the handler's `dontdump_unused_pages()` / resource-release step is exactly such a `je_mallctl` call. When it happens, the node becomes a zombie that neither the FE (every heartbeat is answered with SHUTDOWN, and there is no auto-drop) nor k8s (`/api/health` is a hardcoded 200) can recover — the pod keeps holding its CPU/memory but the FE schedules no queries to it. We hit this in production on 4.0.4: a group of CNs got stuck this way at once and could only be reclaimed by manually restarting the pods. The three prior point fixes for this handler (#59227, #59340, #59538) each closed one stall; bounding the time is the only fix that also covers future unknown stalls.

## What I'm doing:

Add an opt-in watchdog that guarantees the process eventually dies once the crash flag is set, so an orchestrator can restart it. It does not change the crash handler or where the flag is set — it only observes the flag and enforces a deadline.

- New immutable BE config `process_force_exit_after_crash_handler_hang_second`, default `0` (disabled). Upgrading changes nothing: with the default, the watchdog thread is never launched and the existing crash/core-dump behavior is fully preserved. Setting a positive value (and restarting the node) opts in; the value is read once at startup and the polling thread is only launched when it is positive, so the disabled path costs nothing.
- `be/src/common/crash_watchdog.{h,cpp}` (new): when enabled, a detached process-lifetime thread polls `is_process_crashing()` once per second and force-exits the process once the flag has stayed set for the configured deadline. The force-exit path is allocation/LOG-free and cannot block: it makes stderr non-blocking, does one best-effort diagnostic `write()`, and terminates with `_exit(1)` — deliberately not re-raising `SIGABRT`, so it never depends on a core dump that can be slow or fail on a large-memory host (a hung-state core can still be captured manually with `gcore` during the grace window). The deadline logic is a testable `CrashWatchdog::run_once()` with an injectable terminate callback.
- `be/src/service/failure_handler.cpp`: start the watchdog from `init_runtime_logging_hooks()`, and when it is enabled log a one-line startup warning so the tradeoff (a force-exit can preempt a slow crash cleanup/core dump) is never silent.
- Docs (en/zh/ja) and unit tests: deadline-logic tests with a stub terminate, a disabled-path test (no thread launched by default), an enabled-path test (poller runs, never fires without the flag), and one death test that a wedged handler is force-exited with exit code 1.

Because the watchdog is opt-in and the crash flag stays where #66212 put it, the FE still sees SHUTDOWN heartbeats during the grace window before the process dies, and default deployments are byte-for-byte unchanged in behavior.

Out of scope (follow-up candidates): making `/api/health` reflect the crashing state so a k8s liveness probe can also catch it, and FE-side handling of nodes stuck in SHUTDOWN.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [x] 4.0
  - [ ] 3.5


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

be/src/common/config.h
```diff
@@ -2068,6 +2068,13 @@ CONF_Int32(python_worker_expire_time_sec, "300");
 CONF_mBool(enable_pk_strict_memcheck, "true");
 // Reduce core file size by not dumping jemalloc retain pages
 CONF_mBool(enable_core_file_size_optimization, "true");
+// If the fatal-signal (crash) handler hangs, e.g. a jemalloc deadlock while releasing resources
+// before the core dump (https://github.com/StarRocks/starrocks/issues/59226), force the process to
+// exit after this many seconds so orchestrators can restart it. The crash flag is still set first,
+// so the FE keeps seeing SHUTDOWN heartbeats during the grace window; this only bounds how long a
+// crashing process can linger while alive (https://github.com/StarRocks/starrocks/issues/76441).
+// A value <= 0 disables the watchdog.
+CONF_mInt64(be_exit_after_crash_handler_hang_second, "60");
```
kevincai:
```suggestion
CONF_mInt64(process_force_exit_after_crash_handler_hang_second, "60");
```

docs/en/administration/management/BE_parameters/log_server_meta.md
```diff
@@ -169,6 +169,15 @@ This topic introduces the following types of BE configurations:
...
 - Introduced in: v3.4.0, v3.5.0
 
+### be_exit_after_crash_handler_hang_second
+
+- Default: 60
+- Type: Int
+- Unit: Seconds
+- Is mutable: Yes
+- Description: If the fatal-signal (crash) handler hangs, for example a jemalloc deadlock while releasing resources before the core dump, the BE/CN forces the process to exit after this many seconds so that an orchestrator can restart it. The crash flag is still set first, so the FE keeps seeing `SHUTDOWN` heartbeats during the grace window; this only bounds how long a crashing process can linger while alive. A value of `0` or less disables the watchdog.
+- Introduced in: -
```
kevincai: set to v4.1.5, v4.0.15, won't backport to v3.5

be/src/common/config.h
kevincai: A coredump on a large memory VM, may take relative long time, this may cause the coredump never success at all.

CelerData-Reviewer:
@codex review

be/src/common/config.h
ronjay1827: Rebased + addressed review: `_exit()` instead of SIGABRT (@kevincai), non-blocking diagnostic write (Codex P2).

ronjay1827:
For context: we hit this in production on 4.0.4. A group of CN pods went into permanent SHUTDOWN at once after a SIGSEGV (a spill hash-join build-side use-after-free, itself already fixed by #67040 / #69231). The relevant part is that the fatal-signal handler then hung before the re-raise, leaving the process alive forever answering SHUTDOWN. The logs show the flag was set but no `je_mallctl execute purge` line, so it stalled in `dontdump_unused_pages()` — likely a second concurrently-crashing thread parked by glog's first-crasher guard while holding the jemalloc arena mutex. This happened with #59227/#59340/#59538 all already applied, i.e. a new stall variant, which is exactly why bounding theㅍtime is more robust than point-fixing each trigger. @kevincai

kevincai:
> For context: we hit this in production on 4.0.4. A group of CN pods went into permanent SHUTDOWN at once after a SIGSEGV (a spill hash-join build-side use-after-free, itself already fixed by #67040 / #69231). The relevant part is that the fatal-signal handler then hung before the re-raise, leaving the process alive forever answering SHUTDOWN. The logs show the flag was set but no `je_mallctl execute purge` line, so it stalled in `dontdump_unused_pages()` — likely a second concurrently-crashing thread parked by glog's first-crasher guard while holding the jemalloc arena mutex. This happened with #59227/#59340/#59538 all already applied, i.e. a new stall variant, which is exactly why bounding theㅍtime is more robust than point-fixing each trigger. @kevincai

Understood the situation, but doesn't see it address my concern that this early exit in 60s interrupts the running coredump that may take minutes to complete.

be/src/common/crash_watchdog.cpp
```diff
@@ -0,0 +1,101 @@
...
+    int64_t elapsed_seconds = (now_mono_ns - _first_crashing_mono_ns) / 1000000000LL;
+    return elapsed_seconds >= timeout_seconds;
+}
+
+namespace {
+
+// Runs after the process is already known to be crashing, so the heap may be corrupt: keep this
+// path allocation- and LOG-free, and never let it block. This is the last-resort killer, so it
+// terminates the process directly with _exit() instead of raising SIGABRT: a core dump can be slow
+// or fail on large-memory hosts and must not be relied on here (a hung-state core can still be
```
kevincai, line 50: the coredump is the last resort to capture the crash, it's kind of breaking existing function completely. A compromise solution could be, set the new introduced config var default value to `0`, which means don't enable this feature. When the user is willing to enable this feature, it is for sure for what it may break and could set a proper value based on their own env.

Current default enablement with 60s, break coredump without even notice.

ronjay1827: thanks for the review, a few updates. the config defaults to `0` now so it's opt-in, upgrading won't change the existing crash/coredump behavior unless someone sets a positive timeout, and when enabled it logs a warning at startup so the coredump tradeoff isn't silent.

one more idea, maybe for a follow-up: `/api/health` is still a hardcoded 200 with `msg: "To Be Added"`, so a k8s liveness probe can never notice a wedged node. returning 503 when `is_process_crashing()` is set would let liveness restart it and complement the watchdog on the k8s side. only for the crash case though, not the graceful `process_exit_in_progress()` drain, since that same endpoint also backs readiness/startup. separate pr for that, or fold it in here?

kevincai: yes, we can do that, the `/api/health` is a dummy implementation right now. we can add more info to it to make it useful.

be/src/common/config.h
```diff
@@ -2071,6 +2073,14 @@ CONF_mInt32(python_udf_rpc_timeout_ms, "0");
...
 // Reduce core file size by not dumping jemalloc retain pages
 CONF_mBool(enable_core_file_size_optimization, "true");
+// If the fatal-signal (crash) handler hangs, e.g. a jemalloc deadlock while releasing resources
+// before the core dump (https://github.com/StarRocks/starrocks/issues/59226), force the process to
+// exit after this many seconds so orchestrators can restart it. The crash flag is still set first,
+// so the FE keeps seeing SHUTDOWN heartbeats during the grace window; this only bounds how long a
+// crashing process can linger while alive (https://github.com/StarRocks/starrocks/issues/76441).
+// Disabled by default (0) so an upgrade keeps the existing crash/core-dump behavior unchanged; set a
+// positive value to opt in and force-exit after that many seconds. A value <= 0 keeps it disabled.
+CONF_mInt64(process_force_exit_after_crash_handler_hang_second, "0");
```
kevincai: Rethink about this, will be necessary to be runtime mutable? if turn to a non-mutable config, then the watchdog thread can be conditionally launched base on this configvar.

kevincai:
@ronjay1827 thank your for the update, can you please update the PR description to match the code change?

be/src/service/failure_handler.cpp
```diff
@@ -225,6 +226,14 @@ void init_runtime_logging_hooks() {
         // This symbol may be unavailable on macOS builds using system glog.
         google::InstallFailureHandlerAfterOutputLog(failure_handler_after_output_log);
 #endif
+        // Guarantee the process eventually dies even if the crash handler above stalls, so an
+        // orchestrator can restart it instead of leaving a live node stuck reporting SHUTDOWN.
+        start_crash_watchdog();
+        // Opt-in only (default 0). When enabled, warn once at startup so the tradeoff is explicit:
+        // the force-exit can terminate a crash cleanup or core dump that is slower than the bound.
+        if (config::process_force_exit_after_crash_handler_hang_second > 0) {
```
kevincai: the config is referenced in multiple locations, it is not a problem right now, a better way is: crash_watchdog() is config agnostic, it takes an input parameter to guide its behavior.

be/src/service/failure_handler.cpp
```diff
@@ -225,6 +226,14 @@ void init_runtime_logging_hooks() {
         // This symbol may be unavailable on macOS builds using system glog.
         google::InstallFailureHandlerAfterOutputLog(failure_handler_after_output_log);
 #endif
+        // Guarantee the process eventually dies even if the crash handler above stalls, so an
+        // orchestrator can restart it instead of leaving a live node stuck reporting SHUTDOWN.
+        start_crash_watchdog();
```
kevincai:
```suggestion
        may_start_crash_watchdog();
```

the watchdog may not be started at all if the configuration <= 0.

ronjay1827: Done in the latest push. Renamed to may_start_crash_watchdog() and made it config agnostic: it takes the timeout as a parameter, the config is now read in one place (init_runtime_logging_hooks) and crash_watchdog.cpp no longer includes the config header at all.

kevincai (approved):

trueeyu (approved):

Issue #76485, opened 2026-07-16
[Feature] Allow autoscaled CNs joining mid-query to contribute to already-running queries


pushpavanthar:
## Feature request

**Description**

In shared-data clusters with CN autoscaling (e.g. Kubernetes HPA), compute nodes added while a query is running cannot contribute to that query. Fragment placement is fixed at plan/deploy time:

- Worker selection happens once, at schedule time (`DefaultSharedDataWorkerProvider` / coordinator prepare), from the warehouse membership known at that moment.
- Exchange wiring is baked into deploy params (`TPlanFragmentDestination` lists), and receivers (`DataStreamRecvr`) fix their sender count at creation with no dynamic registration.
- Scan ranges are divided across the deployed instance set up front; a node with no fragment instance has nothing to pull.

With a CPU-average-based autoscaler this becomes a self-defeating loop:

1. A heavy query starts on the current CNs → they pin at 100% CPU.
2. The autoscaler adds CNs → the new nodes cannot join the running query and sit idle.
3. Idle nodes dilute the average CPU below the scale-up target → scaling stops.
4. The query grinds on using only the original CNs while paid capacity idles.

Observed in practice: a scan-heavy query on an autoscaled pool (max 8 CNs) never engaged more than 2 CNs; the added replicas stayed idle until the query finished.

**Describe the solution you'd like**

Allow in-flight queries to use CNs that join the warehouse mid-query. Two possible directions, not mutually exclusive:

1. **Elastic leaf (scan) stages in the streaming engine**: coordinator-side scan-range queue (generalizing the existing incremental scan-range delivery, `enable_connector_incremental_scan_ranges`, to lake tablets), late deployment of additional scan-fragment instances on newly joined CNs, and dynamic sender registration in the exchange layer (versioned sender set + coordinator-driven EOS seal, since `_num_remaining_senders` is currently fixed at receiver creation). Interior hash-partitioned stages would remain fixed (consumer count is baked into partitioning and accumulated operator state).
2. **Stage-by-stage scheduling**: decide each stage's placement and parallelism when the stage becomes ready, from current cluster membership, so new CNs pick up work at the next stage boundary — including interior join/aggregation stages. This is anticipated by the intra-query fault tolerance RFC (#76056), where staged execution with durable exchanges provides these boundaries.

**Describe alternatives you've considered**

[...]

PR #76493, opened 2026-07-16
[Enhancement] Support dynamic exchange sender registration on BE

pushpavanthar:
## Why I'm doing:

Exchange receivers fix their sender count at construction (`per_exch_num_senders` → `_num_remaining_senders`), so a fragment instance created after a query is deployed has no way to join its exchanges. This PR adds the BE-side primitive needed for mid-query CN elasticity (#76485): letting the coordinator register additional senders on running receivers before deploying a late scan-fragment instance. No FE caller yet — that lands in a follow-up; this PR is independently reviewable and fully covered by BE UTs.

## What I'm doing:

- **Proto** (`internal_service.proto`, optional-only fields, fresh ordinals): `PUpdateExchangeSendersRequest {finst_id, node_id, be_number, unregister}` / `PUpdateExchangeSendersResult {status}` and rpc `update_exchange_senders`. Registration is keyed by sender identity (`be_number`), never by delta — **idempotent**, so a retry after an ambiguous RPC timeout cannot double-count, and compensation (unregister of a never-registered sender) is a no-op by construction.
- **`PipelineSenderQueue::try_update_senders`**: under the same lock as EOS accounting — register rejects when `_num_remaining_senders` has reached zero (the exchange source may already have latched `is_finished()`; a registration must never resurrect a finished stream), no-ops on duplicate registration, tracks dynamic senders in a set so unregistration only compensates registrations that actually landed and never double-decrements after a real EOS. `decrement_senders` now decrements under the lock to make the reject-at-zero check sound (it previously decremented outside the eos-set critical section).
- **`DataStreamRecvr::update_senders`**: `NotSupported` on merging receivers (fixed one-queue-per-sender layout); wakes the exchange source via the existing observer path since an unregister may complete the stream. The legacy non-pipeline queue returns `NotSupported` via the base-class default.
- **`DataStreamMgr::update_exchange_senders`**: resolves via existing `find_recvr`; `NotFound` when the receiver is gone (callers treat it as an aborted add).
- **RPC handler** in `internal_service.cpp` following the `transmit_runtime_filter` pattern (query_rpc_pool).

Rolling upgrade (standard BE-before-FE order): old FE never calls the new RPC; a future new-FE → old-BE call gets method-not-found and aborts the add gracefully. Existing exchange traffic is byte-identical.

Tests: new `exchange_dynamic_senders_test.cpp` (11 cases) driving the public `DataStreamMgr` API — register extends the stream, duplicate register no-op, reject-after-finished, unregister unknown/after-EOS no-ops, compensation, merging/non-pipeline `NotSupported`, `NotFound`, cancelled receiver, and a concurrent register-vs-EOS stress case. Verified: protoc syntax, `check_gensrc_schema_compatibility.py` clean, `check_be_module_boundaries.py` clean, changed lines clang-formatted.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/be/src/compute_env/data_stream/data_stream_mgr.cpp
+++ b/be/src/compute_env/data_stream/data_stream_mgr.cpp
@@ -132,6 +132,18 @@ std::shared_ptr<DataStreamRecvr> DataStreamMgr::find_recvr(const TUniqueId& frag
     return {};
 }
 
+Status DataStreamMgr::update_exchange_senders(const PUpdateExchangeSendersRequest& request) {
+    TUniqueId finst_id;
+    finst_id.hi = request.finst_id().hi();
+    finst_id.lo = request.finst_id().lo();
+    std::shared_ptr<DataStreamRecvr> recvr = find_recvr(finst_id, request.node_id());
+    if (recvr == nullptr) {
+        return Status::NotFound("exchange receiver not found: fragment_instance_id=" + print_id(request.finst_id()) +
+                                " node_id=" + std::to_string(request.node_id()));
+    }
+    return recvr->update_senders(request.be_number(), request.unregister());
+}
+
 Status DataStreamMgr::transmit_chunk(const PTransmitChunkParams& request, ::google::protobuf::Closure** done) {
     const PUniqueId& finst_id = request.finst_id();
     // TODO(zc): Use PUniqueId directly
--- a/be/src/compute_env/data_stream/data_stream_mgr.h
+++ b/be/src/compute_env/data_stream/data_stream_mgr.h
@@ -65,6 +65,7 @@ class RuntimeState;
 class MetricRegistry;
 class PUniqueId;
 class PTransmitChunkParams;
+class PUpdateExchangeSendersRequest;
 
 // Singleton class which manages all incoming data streams at a backend node. It
 // provides both producer and consumer functionality for each data stream.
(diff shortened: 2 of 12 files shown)


CelerData-Reviewer:
@codex review

pushpavanthar:
**Changes since the automated review:**

- `eeb229ef9fc` — covers the previously-uncovered `update_exchange_senders` RPC handler (the BE incremental-coverage gap): drives the handler through task submission, the NotFound branch, and closure completion.
- `401eb2152c9` — closes two distributed-ordering races found in a failure-ordering review of the register/unregister protocol:
  1. **Tombstoned unregister**: register and unregister are independent RPCs on a multi-threaded pool. If a compensating unregister executes before the delayed register it compensates, the old no-op path let the late register add a sender that will never deploy, hanging the receiver until query timeout. Unregister of an unknown be_number now tombstones it in `_sender_eos_set` (registration already rejects tombstoned senders); safe because the coordinator reserves dynamic be_numbers monotonically and never reuses them.
  2. **Reject registration after close()**: a register racing a short-circuited receiver (e.g. LIMIT) is no longer silently accepted for a stream nobody will drain.
  Plus `_dynamic_senders` hygiene on EOS. New UTs: `unregister_then_late_register_is_rejected`, `register_after_close_is_rejected`.

PR #76494, opened 2026-07-16
[BugFix] Pre-split: alias sample-query projection so a shared sort-key/partition column doesn't dup

luohaha:
Fixes the Sample-Based Tablet Pre-Split data-tier sampler failing with "Duplicate column name" when a column is both a sort-key column and the partition-expression source (e.g. `__time` on a `date_trunc('month', __time)`-partitioned table), which silently skips pre-split and pins the range table at 1 tablet/partition. Aliases each projected column to a unique positional name (`ps_col_0..N`) in `buildSampleSql`.

Draft opened to obtain a TSP build for a RANGE-3-vs-HASH-3 benchmark (verifying the fix lets an expression-partitioned RANGE table pre-split to ~3 tablets/partition).

### DIFF
--- a/docs/en/administration/management/FE_parameters/shared_lake_other.md
+++ b/docs/en/administration/management/FE_parameters/shared_lake_other.md
@@ -490,6 +490,15 @@ This topic introduces the following types of FE configurations:
 - Description: The Compaction Score threshold that triggers Data Ingestion Slowdown in a shared-data cluster. This configuration only takes effect when `lake_enable_ingest_slowdown` is set to `true`.
 - Introduced in: v3.2.0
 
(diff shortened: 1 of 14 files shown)


luohaha:
Closing: this is not the actual fix for range pre-split not splitting. Live testing showed the sample sub-query with the duplicate `__time` projection analyzes and executes fine (a duplicate output column is legal), so this aliasing is inert. The real root cause is that expression-partitioned tables are forced onto the data-tier full-scan sampler, which exceeds the default 300s `tablet_pre_split_pre_submit_timeout_seconds` on large tables — fixable by config (raise the timeout), no code change. A proper follow-up would make the data-tier sample cheap (leading-column projection / row-group sampling) rather than full-scan.

PR #76496, opened 2026-07-16
[BugFix] Skip unavailable nodes when refreshing dictionary cache

### LINKED ISSUE #76495: Dictionary cache goes CANCELLED cluster-wide when a single BE/CN is down, failing all dictionary_get
## Description

When a single BE/CN is unavailable (crashed or stopped) but still registered, the periodic dictionary cache refresh flips the whole dictionary to `CANCELLED`. Every `dictionary_get()` consumer — plain queries and materialized views alike — is then rejected cluster-wide with `dictionary ... is in CANCELLED state`, for as long as the node stays down. Writes, non-dictionary queries, and non-dictionary MVs keep working, so the dictionary feature has zero tolerance for a single node failure, asymmetric with the rest of the cluster's HA.

## Steps to reproduce

1. `CREATE DICTIONARY` over a table and confirm `dictionary_get(...)` works.
2. Stop or kill one BE (or CN) and leave it down.
3. Run any query (or refresh an MV) that uses `dictionary_get(...)`.

Observed: it fails with `dictionary ... is in CANCELLED state`; the dictionary stays CANCELLED and each refresh retry re-hits the down node with no backoff. It recovers only after the node comes back.

## Root cause

- `DictionaryMgr.fillBackendsOrComputeNodes` builds the refresh broadcast set from `getBackends()` / `getComputeNodes()`, which include dead/decommissioned nodes (no liveness filter).
- `DictionaryMgr.processDictionaryCacheInteranl` aborts on the first node whose RPC throws or returns non-OK (all-or-nothing), so the down node fails the whole refresh (already at the BEGIN phase).
- `RefreshDictionaryCacheWorker.finish()` then sets the single, global dictionary to `CANCELLED`.
- `ExpressionAnalyzer.visitDictionaryGetExpr` rejects any `dictionary_get` whose dictionary is `CANCELLED`.

## Expected

A single down node should not disable the dictionary on the surviving nodes; the dictionary should stay queryable, and the down node should be re-synced when it recovers.

## Proposed fix

Build the broadcast set from `getAvailableBackends()` / `getAvailableComputeNodes()` so dead/decommissioned nodes are excluded. The refresh then succeeds on the available nodes and the dictionary stays queryable; a recovered node is re-included on the next refresh cycle.

Further hardening (follow-up, not in the attached PR): tolerate a node dying mid-refresh instead of aborting the whole broadcast; add backoff between failed refreshes; optionally fall back to the last successful version in the consumer gate.

Youngwb:
## Why I'm doing:

When a single BE/CN is unavailable (crashed or stopped) but still registered, the periodic dictionary cache refresh flips the whole dictionary to `CANCELLED`, and every `dictionary_get()` consumer — plain queries and materialized views alike — is then rejected cluster-wide with `dictionary ... is in CANCELLED state` for as long as the node stays down. Writes, non-dictionary queries, and non-dictionary MVs are unaffected, so the dictionary feature has zero tolerance for a single node failure.

Root cause: `DictionaryMgr.fillBackendsOrComputeNodes` builds the refresh broadcast target set from `getBackends()` / `getComputeNodes()` (no liveness filter), and `processDictionaryCacheInteranl` aborts on the first node whose RPC fails, so one dead node fails the whole refresh and `finish()` sets the dictionary `CANCELLED`; `ExpressionAnalyzer.visitDictionaryGetExpr` then rejects every `dictionary_get` on a `CANCELLED` dictionary.

## What I'm doing:

Build the dictionary refresh broadcast set from `getAvailableBackends()` / `getAvailableComputeNodes()` so dead and decommissioned nodes are excluded. The refresh then succeeds on the available nodes and the dictionary stays queryable; a node that recovers is re-included on the next refresh cycle.

Follow-up hardening tracked in the issue (not in this PR): tolerate a node dying mid-refresh instead of aborting the whole broadcast, add backoff between failed refreshes, and optionally fall back to the last successful version in the consumer gate.

Fixes #76495

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

Internal robustness fix: no change to query results, syntax, or configuration. It only prevents a single down node from cancelling the dictionary on the surviving nodes.

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/DictionaryMgr.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/DictionaryMgr.java
@@ -165,13 +165,13 @@ public class DictionaryMgr implements Writable, GsonPostProcessable {
     }
 
     public static void fillBackendsOrComputeNodes(List<TNetworkAddress> nodes) {
-        List<Backend> backends = GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getBackends();
+        List<Backend> backends = GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getAvailableBackends();
         for (Backend backend : backends) {
             nodes.add(backend.getBrpcAddress());
         }
 
         List<ComputeNode> computeNodes =
-                GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getComputeNodes();
+                GlobalStateMgr.getCurrentState().getNodeMgr().getClusterInfo().getAvailableComputeNodes();
         for (ComputeNode cn : computeNodes) {
             nodes.add(cn.getBrpcAddress());
         }
--- a/fe/fe-core/src/test/java/com/starrocks/catalog/DictionaryMgrTest.java
+++ b/fe/fe-core/src/test/java/com/starrocks/catalog/DictionaryMgrTest.java
@@ -48,8 +48,14 @@ public class DictionaryMgrTest {
     @Mocked
     private SystemInfoService systemInfoService;
 
-    private List<Backend> backends = Arrays.asList(new Backend(1, "127.0.0.1", 1234));
-    private List<ComputeNode> computeNodes = Arrays.asList(new ComputeNode(2, "127.0.0.2", 1235));
+    private Backend availableBackend = new Backend(1, "127.0.0.1", 1234);
+    private Backend unavailableBackend = new Backend(3, "127.0.0.3", 1234);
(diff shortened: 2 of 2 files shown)


CelerData-Reviewer:
@codex review

srlch (approved):

Astralidea (approved):

srlch (approved):

PR #76497, opened 2026-07-16
[Enhancement] Support registering fragment instances after execution DAG finalization

pushpavanthar:
## Why I'm doing:

A query's fragment-instance set is frozen in `ExecutionDAG.finalizeDAG()`: the dense `indexInJob` routing space, instance-id allocation, worker instance counts, and the profile completion barrier (`MarkedCountDownLatch`, fixed size) are all materialized exactly once. That makes it structurally impossible to add a fragment instance to a running query — the FE-side prerequisite for mid-query CN elasticity (#76485), where scan instances are added on compute nodes that join the warehouse while a query runs.

This is the "unfreeze" refactor of that series: pure mechanism, **zero behavior change**, no production caller yet (the orchestration lands in a follow-up).

## What I'm doing:

- **`ExecutionDAG.registerLateInstance(fragment, worker)`**: retains the dense `indexInJob` sequence after `finalizeDAG()` and continues it for late instances (keeping BE→FE report routing by `backend_num` consistent); reuses the existing monotonic instance-id allocator (collision-free by construction); appends to the fragment so existing instances are never reordered — global-runtime-filter component ordinals derived from the original ordering stay valid. The destination fragment's `numSendersPerExchange` is deliberately NOT updated: downstream receivers already hold the plan-time sender count, and the wire-side sender set grows through the BE dynamic sender-registration RPC (#76493) instead.
- **Concurrency without penalizing the common path**: `instanceIdToInstance` / `workerIdToNumInstances` become concurrent maps (point lookups only, no ordered iteration — golden scheduler tests unchanged); `ExecutionFragment.addInstanceLate` publishes a copied instance list (volatile field) so report and cancel threads iterating `getInstances()` never observe a partially updated list. The prepare-time `addInstance` path still mutates in place — no O(n²) copying for wide fragments.
- **`GrowableMarkedLatch`**: replaces the fixed-count `MarkedCountDownLatch` for `QueryRuntimeProfile.profileDoneSignal`. The count is the number of outstanding marks; marks can be added while the query runs and the latch is sealed on completion, so `QueryRuntimeProfile.attachInstance(id)` fails once the query has finished (a late instance must not be deployed then). Same API surface (`markedCountDown`, `countDownToZero`, `getLeftMarks`, listeners, timed `await`). `MarkedCountDownLatch` itself is untouched (other users: load jobs).

Tests: `GrowableMarkedLatchTest` (grow-while-running, seal-on-completion, forced completion + status, unknown-mark, blocking await, one-shot listeners) and `RegisterLateInstanceTest` on the scheduler harness (dense index continuation, fresh id + lookup registration, appended `indexInFragment`, worker accounting, existing instances untouched). Byte-identical proof: `QueryRuntimeProfileTest`, `StartSchedulingTest`, `PhasedScheduleTest`, `ExecutionDAGTest`, `DeployerTest`, `GetNextTest`, `SingleNodeScheduleTest`, `QueryQueueManagerTest`, `IncrementalDeployHiveTest` all pass unchanged.

## What type of PR is this:

- [ ] BugFix
- [ ] Feature
- [x] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [ ] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/common/util/concurrent/GrowableMarkedLatch.java
+++ b/fe/fe-core/src/main/java/com/starrocks/common/util/concurrent/GrowableMarkedLatch.java
@@ -1,0 +1,155 @@
+// Copyright 2021-present StarRocks, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.starrocks.common.util.concurrent;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Multimap;
+import com.starrocks.common.Status;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * A {@link MarkedCountDownLatch} variant whose mark set can still grow while the latch has not
+ * completed. The count is the number of outstanding marks, so there is no fixed initial count:
+ * {@link #addMark} raises the count, {@link #markedCountDown} lowers it, and waiters are released
+ * when no marks remain.
+ *
+ * <p>Once the latch has completed — all marks counted down after at least one was ever added, or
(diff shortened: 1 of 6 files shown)


CelerData-Reviewer:
@codex review

CelerData-Reviewer:
@codex review

pushpavanthar:
**Changes since the automated review:**

- `04cda675245` — addresses the **P2 finding**: `attachInstances` fully builds the profile latch before publishing it, so a concurrent KILL/cancel can no longer observe a half-marked latch.
- `b5d4fb571db` — addresses the **P1 finding** (runtime-filter layout for late instances): rather than recomputing the RF merge layout — which would fix only the new instance's view while already-deployed builders and the merger keep the plan-time count — `registerLateInstance` now **rejects** fragments that build global runtime filters. This keeps the primitive safe for any caller; probe-only RF fragments are unaffected. UT appends an instance to an RF-build fragment and asserts rejection.

Each commit standalone; existing scheduler golden tests unchanged.

PR #76513, opened 2026-07-17
[BugFix] Resolve Iceberg columns via schema.name-mapping.default to fix NULLs from Fabric Warehouse

EraserKing:
## Why I'm doing:

When querying a Microsoft Fabric Warehouse table through its Iceberg endpoint (OneLake exposes Warehouse tables as read-only Iceberg tables via a Delta-to-Iceberg metadata shim), StarRocks returns NULL for columns that
actually contain data.

Root cause: Fabric Warehouse's underlying storage is Delta Lake with column mapping enabled, so the physical Parquet columns are named using Delta's column-mapping IDs (e.g. `col-<uuid>`) instead of the logical column name (e.g. `DateID`), and the Parquet files do not carry embedded Iceberg field-ids. The Iceberg metadata shim correctly publishes a `schema.name-mapping.default` table property (per the Iceberg spec) so readers can resolve each field-id to its physical column name(s) in this
case — but StarRocks' Iceberg connector never reads this property at all.
As a result, column resolution falls back to matching by logical name only, fails to find `col-<uuid>` under the name `DateID`, and the column is silently treated as missing and filled with NULL.

Note this is not specific to Fabric — any Iceberg table backed by Parquet files without embedded field-ids that relies on `schema.name-mapping.default` (e.g. Delta-to-Iceberg conversions in general) hits the same issue.

## What I'm doing:

Fixes #issue

- Add `Column.setPhysicalName()` (the getter already existed, but there was no setter).
- Add `IcebergApiConverter.applyIcebergNameMapping()`, called from `toFullSchemas(Schema, Table)`, which:
  - Reads the `schema.name-mapping.default` table property, if present.
  - Parses it into an Iceberg `NameMapping`.
  - For each converted `Column`, looks up the corresponding Iceberg field-id's `MappedField` and, if the mapping lists an alias name different from the logical column name, sets it as the column's `physicalName`.
- This lets the existing BE `col_physical_name` fallback (in `ParquetMetaHelper`, today only populated by the Delta Lake connector) correctly resolve the column against the Parquet file's actual physical name when the file has no embedded field-id for that column.
- Added unit tests covering: alias resolution from a mapping, no-op when the property is absent, and safe handling of a malformed mapping.

## What type of PR is this:

- [x] BugFix
- [ ] Feature
- [ ] Enhancement
- [ ] Refactor
- [ ] UT
- [ ] Doc
- [ ] Tool

- [ ] Feature removed

## Checklist:

- [ ] This is a backport pr

  - [x] 4.1
  - [ ] 4.0
  - [ ] 3.5

### DIFF
--- a/fe/fe-core/src/main/java/com/starrocks/catalog/Column.java
+++ b/fe/fe-core/src/main/java/com/starrocks/catalog/Column.java
@@ -1087,6 +1087,10 @@ public class Column implements Writable, GsonPreProcessable, GsonPostProcessable
         return physicalName;
     }
 
+    public void setPhysicalName(String physicalName) {
+        this.physicalName = physicalName;
+    }
+
     // return max unique id of all fields
     public int getMaxUniqueId() {
         return Math.max(this.uniqueId, type.getMaxUniqueId());
--- a/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergApiConverter.java
+++ b/fe/fe-core/src/main/java/com/starrocks/connector/iceberg/IcebergApiConverter.java
@@ -330,6 +330,7 @@ public class IcebergApiConverter {
 
     public static List<Column> toFullSchemas(Schema schema, Table table) {
         List<Column> fullSchema = toFullSchemas(schema);
+        applyIcebergNameMapping(fullSchema, schema, table.properties());
         if (table instanceof BaseTable) {
             addMetaColumns(fullSchema);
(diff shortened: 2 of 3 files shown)


CelerData-Reviewer:
@codex review

stephen-shelby (approved):

EraserKing:
Hi, should the P1 be fixed or it's safe to leave as-is? Thanks.

